PR #27047, opened 2024-10-09
cli: repo lint fixes

Rugvip:

Some fixes for the `--successCache` path of `repo lint`. This switches to using globby to enumerate files, so can take `.gitignore` into account. It's slower, but more accurate. Using the git tree sha strategy from `repo test` is tricky still because we need to make sure the lint config for each file is part of the cache key, and we can only get that by enumerating it for each file. Long-term I'm thinking the best solution here is to rely on the new ESLint config format, but for now I think this is fine. I might evaluate using `git` to list files instead of globby though, could be that that's a lot faster.

Other fix is to only include packages that actually have a `lint` script.

### DIFF
--- a/.changeset/rotten-rockets-deny.md
+++ b/.changeset/rotten-rockets-deny.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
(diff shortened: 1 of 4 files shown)


freben (approved):
aaaaalright then :)

deejay1 (approved):

PR #27053, opened 2024-10-09
feat(catalog-backend): Add catalog processing pipeline parameters

deejay1:

Allows for settings the low and high watermarks for the pipeline.


allegro-dbargowski (approved):

adowsky (approved):

awanlin (changes requested):
Hi @deejay1, same with this one can you please add documentation for this change, here would be fine: https://github.com/backstage/backstage/blob/master/docs/features/software-catalog/configuration.md#processing-interval

vinzscam (approved):

deejay1:
@awanlin could you rereview it ?

deejay1:
@freben could you take a look? it would be great if this hit todays release

docs/features/software-catalog/configuration.md
```diff
@@ -192,6 +192,24 @@ Setting this value too low risks exhausting rate limits on external systems that
...
+  Once the number of in-flight tasks reaches this count, more tasks will be loaded in.
+- `highWatermark` defines the maximum number of items to process in parallel.
+
+Keep in mind that the `lowWatermark` has to be less than `highWatermark`.
+
+```yaml title="app-config.yaml"
+catalog:
+  pipeline:
+    lowWatermark: 5
+    highWatermark: 10
```
freben, line 210: Could we make this

```yaml
catalog:
  processing:
    concurrency: { min: 5, max: 10 }
```

instead?

The reason I'm asking is, I've been thinking that it'd be good to move ALL processing loop related things into a `processing` key, and this naming also expresses more clearly what is going on.

Later down the line, we may move `processingInterval` into `processing.interval` etc as well, and maybe have a `processing.enabled` too, but do not do that as part of this PR.

plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
```diff
@@ -129,9 +129,13 @@ export class DefaultCatalogProcessingEngine {
   }
 
   private startPipeline(): () => void {
+    const lowWatermark =
+      this.config.getOptionalNumber('catalog.pipeline.lowWatermark') ?? 5;
+    const highWatermark =
+      this.config.getOptionalNumber('catalog.pipeline.highWatermark') ?? 10;
```
freben, line 135: Should throw an exception if min isn't strictly smaller than max.

I'm also not sure, should you really be permitted to define just one of them? That'll just cause problems if we want to change the defaults - then people's installations may start to break if that violates this check. So, maybe `concurrency` should be optional, but not `min` and `max` inside it?

freben (commented):
Were you adding this just for completeness, or were you really having issues from it not being configurable? We wanted to leave as few configurability parameters as possible in here.

docs/features/software-catalog/configuration.md
deejay1: done

deejay1:
> Were you adding this just for completeness, or were you really having issues from it not being configurable? We wanted to leave as few configurability parameters as possible in here.

Unfortunately we really do need this and we're really not looking forward to forking the catalog due to the need to change two hardcoded parameters, which are a good tunable when working with thousands of entities. Right now every processing run tends to take over 15 minutes and tuning this should allow to reduce that and be more responsive for our users

drodil:
@freben @vinzscam any chance this could make it to 1.32.0?

awanlin (approved):

Rugvip:
@deejay1 what parameters work well for you if you tweak this with Yarn patches? We've picked the current values very arbitrarily and would happily update to whatever is optimal for your setup as that is probably going to benefit most deployments out there.

drodil:
@Rugvip I think it depends on the catalog size and thus I am in favor of making this configurable. Sure the defaults should be optimal values for most use cases.

I think these values are one of those things you can only find out with try and error by running the instance for some time with each change.

Rugvip:
I don't quite see how it could depend on catalog size tbh, we're really just looking for maximum concurrency and throughput right? If one instance grabs too many items that'd be bad, but if they're DB contention or idling that might be bad too.

I'm fairly confident that nobody will benefit from making this configurable in the long runt as it makes the overall setup more inflexible. I'm quite curious if there's a good indication that we could use to adjust these dynamically as well, for example based on how many items we actually manage to pick from the DB on average vs our throughput?

deejay1:
Why does it make the setup more inflexible? For example we consider running only worker nodes of backstage to handle only tasks instead of everything, so we could tune them separately

Issue #22504, opened 2024-01-25
Support Wildcards for all integrations in `UrlReaderProcessor`


benjdlambert:
### 🔖 Feature description

Currently we detect wether a URL for reading is a search query by detecting if there's a wildcard in the URL.

https://github.com/backstage/backstage/blob/4a3a4278d1f3fe0e30ef82cc563d4d660a2428bf/plugins/catalog-backend/src/modules/core/UrlReaderProcessor.ts#L128-L129

This is not a great approach for many reasons, but one being that it doesn't work for non Git URL's of which there are many. One use case is #21955, when reading files from Google Cloud Storage URLs, which of course don't parse into a Git URL.

We attempted to fix that, by parsing that URL as is, and checking the `pathname`:

https://github.com/backstage/backstage/blob/c3249d6/plugins/catalog-backend/src/modules/core/UrlReaderProcessor.ts#L127-L128

But this caused issues with the `AzureURLReader` https://github.com/backstage/backstage/issues/22465, as the actual wildcard is in the `querystring` not the `pathname` e.g  `https://dev.azure.com/my-org/my-project/_git/my-repo?path=/templates/*/*.yaml`

### 🎤 Context

We should support reading from sources consistently, and support wildcards for all `URLReaders`.

### ✌️ Possible Implementation

I think that there could be an option to flip this around a little bit, and have a `isSearchUrl` on the `URLReader` implementations which can be used to detect if the URL which has been provided is a wildcard URL before it is then used here:

https://github.com/backstage/backstage/blob/4a3a4278d1f3fe0e30ef82cc563d4d660a2428bf/plugins/catalog-backend/src/modules/core/UrlReaderProcessor.ts#L131

Something like this:

```ts
if (this.options.reader.isSearchUrl(location))) {
  const limiter = limiterFactory(5);
  const response = await this.options.reader.search(location, { etag });
  const output = response.files.map(async file => ({
    url: file.url,
    data: await limiter(file.content),
  }));
  return { response: await Promise.all(output), etag: response.etag };
}
```

Where the implementation for all the Git based URL readers could be something like:

```ts
public isSearchUrl(location: string) {
  const { filepath } = parseGitUrl(location);
  return filepath?.match(/[*?]/)
 }
 ```
 
 But more specialized `URLReaders` like the Google Cloud Storage one can be something like this:
 
```ts
public isSearchUrl(location: string) {
   const { pathname: filepath } = new URL(location);
   return filepath?.match(/[*?]/)
[...]

PR #27054, opened 2024-10-09
feat(catalog): always use search for UrlReaderProcessor

tcardonne:

> **BREAKING:** The `UrlReaderProccessor` now always calls the `search` method of the `UrlReaders`. Previous behavior was to call the `search` method only if the parsed Git URL's filename contained a wildcard and use `readUrl` otherwise. `UrlReaderService` must implement this logic in the `search` method instead.
>
> This allows each `UrlReaderService` implementation to check whether it's a search URL (that contains a wildcard pattern) or not using logic that is specific to each provider.
>
> In the different `UrlReadersService`, the `search` method have been updated to use the `readUrl` if the given URL doesn't contain a pattern.
For `UrlReaders` that didn't implement the `search` method, `readUrl` is called internally.

See comment by @Rugvip for the reasoning: https://github.com/backstage/backstage/pull/27054#discussion_r1801439398

Closes https://github.com/backstage/backstage/issues/22504

Unblocks https://github.com/backstage/backstage/pull/25704

- [X] ~Added or updated documentation~

### DIFF
--- a/packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.ts
+++ b/packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.ts
@@ -37,7 +37,11 @@ import fetch, { RequestInit, Response } from 'node-fetch';
 import parseGitUrl from 'git-url-parse';
 import { Minimatch } from 'minimatch';
 import { Readable } from 'stream';
-import { NotFoundError, NotModifiedError } from '@backstage/errors';
+import {
+  assertError,
+  NotFoundError,
+  NotModifiedError,
+} from '@backstage/errors';
 import { ReadTreeResponseFactory, ReaderFactory } from './types';
 import { ReadUrlResponseFactory } from './ReadUrlResponseFactory';
 import { parseLastModified } from './util';
@@ -179,6 +183,35 @@ export class GithubUrlReader implements UrlReaderService {
     url: string,
     options?: UrlReaderServiceSearchOptions,
   ): Promise<UrlReaderServiceSearchResponse> {
+    const { filepath } = parseGitUrl(url);
+
+    // If it's a direct URL we use readUrl instead
+    if (!filepath?.match(/[*?]/)) {
+      try {
+        const data = await this.readUrl(url, options);
+
+        return {
+          files: [
+            {
+              url: url,
+              content: data.buffer,
+              lastModifiedAt: data.lastModifiedAt,
+            },
+          ],
+          etag: data.etag ?? '',
+        };
+      } catch (error) {
+        assertError(error);
+        if (error.name === 'NotFoundError') {
+          return {
+            files: [],
+            etag: '',
+          };
+        }
+        throw error;
+      }
+    }
+
     const repoDetails = await this.getRepoDetails(url);
     const commitSha = repoDetails.commitSha;
 
@@ -186,7 +219,6 @@ export class GithubUrlReader implements UrlReaderService {
       throw new NotModifiedError();
     }
 
-    const { filepath } = parseGitUrl(url);
     const { headers } = await this.getCredentials(url, options);
 
     const files = await this.doSearch(
--- a/plugins/catalog-backend/src/processors/UrlReaderProcessor.ts
+++ b/plugins/catalog-backend/src/processors/UrlReaderProcessor.ts
@@ -18,6 +18,7 @@ import { Entity } from '@backstage/catalog-model';
 import { assertError } from '@backstage/errors';
 import limiterFactory, { Limit } from 'p-limit';
 import { LocationSpec } from '@backstage/plugin-catalog-common';
+import parseGitUrl from 'git-url-parse';
 import {
   CatalogProcessor,
   CatalogProcessorCache,
@@ -28,6 +29,7 @@ import {
   processingResult,
 } from '@backstage/plugin-catalog-node';
 import { LoggerService, UrlReaderService } from '@backstage/backend-plugin-api';
+import { Config } from '@backstage/config';
 
 const CACHE_KEY = 'v1';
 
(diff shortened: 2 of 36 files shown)


packages/backend-plugin-api/src/services/definitions/UrlReaderService.ts
```diff
@@ -40,6 +40,12 @@ export interface UrlReaderService {
     options?: UrlReaderServiceReadTreeOptions,
   ): Promise<UrlReaderServiceReadTreeResponse>;
 
+  /**
+   * Returns true if a given URL contains a search pattern such as wildcards
+   * in the path.
+   */
+  isSearchUrl(url: string): boolean;
```
Rugvip, line 47: Idea that came up during the framework SIG (👀 @benjdlambert):

First a bit of context, which is that the entire need for this is this piece of code here: https://github.com/backstage/backstage/blob/e0613e9957909b1ffaefb208721914fd2746c128/plugins/catalog-backend/src/processors/UrlReaderProcessor.ts#L134-L148

We need to branch based on whether it's a search URL or not. But what if we didn't need to do that? Rather than having the need to differentiate between search and direct file URLs, could we make the `search` method essentially short-circuit in the case of a direct URL? We'd still keep the existing API, but in the various implementations of the `search` methods we would check whether it's a search URL or not using logic that is specific to each provider. If it's not a search URL, we simply defer to the `readUrl` method for that implementation and return that single file as the search result. With this approach we can still do proper `etag` handling since that's supported by the `search` method, and I think really the only remaining question about the implementation would be how we'd handle 404s? Should that result in an error or an empty search result?

I do think that approach is pretty neat, and not really a breaking change in the API itself as it's arguably the correct behavior already. The tricky bit is gonna be the `UrlReaderProcessor` though. It's a breaking change to switch over to always calling the `search` method, so I think we may need to have a flaaaag

Rugvip (commented):
Thank you! 🎉 

Continuing the discussion from the framework SIG

packages/backend-plugin-api/src/services/definitions/UrlReaderService.ts
tcardonne: Thanks for the feedback during the SIG and here!

I've pushed on this PR an attempt with the `GithubUrlReader`. I'll continue on other readers but I wanted to start small!

We might need to do some change in the return type of the `search` method: the `readUrl` returns an optional `etag` whereas for the `search` the `etag` is mandatory. For the `GithubUrlReader` this requires to get the repo datails first before doing the short-circuit to `readUrl`.

> how we'd handle 404s? Should that result in an error or an empty search result?
I've moved the handling of the `NotFoundError` error in the `search` method, to return an empty search result in case a direct URL returns a 404.
The current implementation returns an empty list of files if the glob pattern doesn't match anything, so for me it makes sense to return an empty list if a direct URL returns a 404.
In the `UrlReaderProcessor`, if the list is empty and the location isn't optional, then it emits an error. Otherwise it's ignored.
WDYT?

> The tricky bit is gonna be the UrlReaderProcessor though. It's a breaking change to switch over to always calling the search method, so I think we may need to have a flaaaag
How would you implement this flag? Checking for an environment variable that will be removed eventually in a future release?

packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.ts
```diff
@@ -187,6 +191,34 @@ export class GithubUrlReader implements UrlReaderService {
     }
 
     const { filepath } = parseGitUrl(url);
+
+    // If it's a direct URL we use readUrl instead
+    if (!filepath?.match(/[*?]/)) {
```
Rugvip, line 196: We'll want do this check and fork before checking repo details imo, otherwise we'll end up with a lot of extra repo details requests. Along with that we'll also want to use the `etag` that's returned from the `readUrl` call instead of the commit sha. I'm not too sure what to do if the `etag` is missing though, it's a bit unfortunate that the `search` method requires it. We could try throwing an error, although I have a feeling that returning `''` is gonna be the simplest and most compatible solution.

packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.ts
```diff
@@ -187,6 +191,34 @@ export class GithubUrlReader implements UrlReaderService {
...
+          etag: commitSha,
+        };
+      } catch (error) {
+        assertError(error);
+        if (error.name === 'NotFoundError') {
+          return {
+            files: [],
+            etag: commitSha,
+          };
+        }
```
Rugvip, line 217: Yep seen from the point of view of the `search` method this really seems like the only sensible behavior tbh. If we didn't we'd have completely different behaviors for search and exact URLs.

Rugvip (commented):
Thank you! 👍 

Feeling fairly optimistic about this approach

packages/backend-plugin-api/src/services/definitions/UrlReaderService.ts
Rugvip: > How would you implement this flag? Checking for an environment variable that will be removed eventually in a future release?

Let's see if we feel we need it first. It could be a config option, or a separate module. Ideally we're able to ship without that, but I'm not too optimistic about being able to find a solution for that tbh.

tcardonne: Alright, let me know which way you want to go!

tcardonne:
Hi @Rugvip 👋 

I've finished the implementation for other readers. Let me know if there's something to change or a flag to make the change smooth 👍

tcardonne:
Not stale!

plugins/catalog-backend/src/processors/UrlReaderProcessor.ts
```diff
@@ -128,23 +132,13 @@ export class UrlReaderProcessor implements CatalogProcessor {
...
-        data: await this.#limiter(file.content),
-      }));
-      return { response: await Promise.all(output), etag: response.etag };
-    }
+    const response = await this.options.reader.search(location, { etag });
+
+    const output = response.files.map(async file => ({
+      url: file.url,
+      data: await this.#limiter(file.content),
+    }));
```
Rugvip, line 140: Let's move this new behavior behind a config flag so that people can opt-in, and then I think this is good to go 👍 

Let's also log a deprecation warning if the flag is not set to encourage people to move over

tcardonne: Done!

tcardonne:
Just rebased and fixed a conflict, should be good to reopen 👍 

(and happy new year!)

PR #27056, opened 2024-10-09
NFS: Add missing default API's and `DefaultSignInPage`

benjdlambert:
Not sure about `DefaultSignInPage` 💯  but without it you get 401 errors when starting an app out of the box.


Rugvip (approved):
👍

PR #27066, opened 2024-10-10
feat: api-docs pulgin support i18n

mario-mui:

### DIFF
--- a/plugins/api-docs/src/translation.ts
+++ b/plugins/api-docs/src/translation.ts
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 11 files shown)


plugins/api-docs/src/translation.ts
```diff
@@ -0,0 +1,86 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createTranslationRef } from '@backstage/core-plugin-api/alpha';
+
+export const apiDocsTranslationRef = createTranslationRef({
+  id: 'api-docs',
+  messages: {
+    apiDefinitionDialog: {
```
vinzscam, line 22: I haven't followed the approach we have used in other plugins, but I would like to point out something that might not work with this approach.

I see that translation keys are very closely tied to the locations of the components. This means that if a component is renamed or elements are rearranged, the translation keys might need to be refactored as well. If they aren't refactored, they will end up with meaningless keys. Another disadvantage I see is that we'll have a lot of repetition; for example, the word "APIs" is defined three times.

I suggest trying to keep a flat structure, with the name of the key associated with its English value. If, in the future, we see the need for having different translations depending on where a single message is rendered, we can duplicate and create a more specific translation key.

mario-mui: I think we need to care about translation keys, when we need to refactor some conponents, we need to consider those translation keys, in case of leaving keys not use.

we should think  about duplicated keys, maybe flat structure is a good solution.

@Rugvip @benjdlambert Do you have any suggestions?

benjdlambert: There's a load of [documentation](https://backstage.io/docs/plugins/internationalization/) around best practices.

Avoiding [flat keys](https://backstage.io/docs/plugins/internationalization/#flat-keys] is one of the recommendations and think that it's still true for most cases. Things that you expect to be integral to the page could perhaps be hoisted, but it probably makes sense to keep them scoped where possibel.

mario-mui:
ping

PR #27067, opened 2024-10-10
feat: added the ability to disable catalog processing

benjdlambert:
This adds the ability to disable catalog processing with config. 

Was in two minds to just add this ability to the existing `readonly` config, but thought that it could be an unexpected breaking change in behaviour for people that don't allow new locations to be registered through the API, but have Discovery Providers.

I also wanted to move both `processingInterval` and `processingEnabled` to `processing.interval` and `processing.enabled`, but thought best to leave that for now, and revisit the config on Catalog 2.0

### DIFF
--- a/plugins/catalog-backend/config.d.ts
+++ b/plugins/catalog-backend/config.d.ts
@@ -164,7 +164,6 @@ export interface Config {
 
     /**
      * The interval at which the catalog should process its entities.
-     *
      * @remarks
      *
      * Example:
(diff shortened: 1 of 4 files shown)


plugins/catalog-backend/config.d.ts
```diff
@@ -182,5 +181,20 @@ export interface Config {
      * housing catalog-info files.
      */
     processingInterval?: HumanDuration;
```
Rugvip, line 183: Maybe silly, but should setting the interval to 0 disable processing instead?

benjdlambert: Hmm, it's interesting. But it would be `processingInterval: { seconds: 0 }` which is a little odd. Maybe `processingInterval?: HumanDuration | boolean`?

freben: Or even `HumanDuration | false` specifically?

Rugvip (approved):
:+1: :shipit:

alexef:
good that this is now possible, coming after a 3 day incident where we couldn't figure out why github api limits are exhausted after upgrade from 1.30->1.32

PR #27070, opened 2024-10-10
rewrite some more tests to use actual fake catalog instead of mock

freben:


vinzscam:
> As discussed - this does mean that tests should avoid using `jest.resetAllMocks()`.

this looks like a strong assumption.
Let's say I'm testing a component with some mocked apis for the component. If I want to have a difference mock for each test, I would like to invoke `resetAllMocks()` after each test. But now, I can't anymore since it will break the catalog mocks 🤔

freben:
Yeah we spoke about that actually - I called out the same thing. It's very nice though, and it's freaking annoying when you have a perfectly fine and small test, and then when you want to add an assertion on the call, you have to completely rewrite it to use a mock instead, which usually implies much uglier code at the top.

Maybe that's a wrong assumption though, maybe I should lean into using `jest.spyOn` for those tests instead. Let me look at that and maybe revise the tests accordingly.

Rugvip (approved):
Nice! 👍

PR #27088, opened 2024-10-11
promote all backend feature exports to main entry point

Rugvip:

🧹 , removing the need to use `/alpha` when installing feature for the new system. There's still gonna be cases where extension points etc. are only available under `/alpha`, but wanted to get this out of the way.


pjungermann (approved):
There seems to be a circular dependency issue for 2 plugins still.

Approving for the ones related to events-maintainers, and those where I'm one of the CODEOWNERS. 👍

Rugvip:
Find the issue with the circular imports, it was the plugin implementations importing from the package index rather than internally

freben (approved):

PR #27095, opened 2024-10-12
POC Refactor for non-yarn package manager support

tylerd-canva:

It's still pretty rough but I wanted to get some sort of sanity check before I continue down this path and go to the work of actually adding pnpm support.

### Summary of changes so far
* Refactor all the `yarn` commands and `Lockfile` stuff into an interface/abstract class and move it into `@backstage/cli-node`.
* Add a new Role: `root` since `pnpm` doesn't use the `workspaces` field in `package.json`. *probably not the best way to do this but it's what I came up with off the top of my head*
* Kept all the types the same for now, not sure what would need to become more generic until I actually attempt the PNPM support which I think makes sense to become a separate PR

### Things yet to do
* Update API docs once the interfaces in this approach have been approved
* Tests for the non-abstract methods in Lockfile.ts
* Add PNPM support (probably a separate PR)
* try and remove new `chalk` dependency from `@backstage/cli-node` (used by `errors` code which was moved over)
* Move yarn support (and future pnpm support) into their own separate packages? This might be especially useful for....
* create-app support for choosing your package manager at creation time. but if we have a dynamic template we might also need...
* backstage-upgrade-helper support. pick your package manager in the UI so that you get an effective diff

### Open Questions
* `@backstage/eslint-plugin`: https://github.com/backstage/backstage/blob/434b9c19dbd3c61f7b98b7c554ee658683120bed/packages/eslint-plugin/rules/no-undeclared-imports.js#L189
* `@backstage/e2e-test`: https://github.com/backstage/backstage/blob/c35be1d21b33a72c3698e980e97d3526fb35c66b/packages/e2e-test/src/commands/run.ts#L57
* `@backstage/yarn-plugin` -> might need to plan a pnpm version, but this is optional
* `@backstage/repo-tools`: https://github.com/backstage/backstage/blob/2a42f97175f2c04e7a4a60ea24150a8156a24ac3/packages/repo-tools/src/commands/package/schema/openapi/init.ts#L68

## :heavy_check_mark: Checklist


packages/cli-node/src/pacman/yarn/Lockfile2.test.ts
tylerd-canva, line 1: need to combine these tests into the other test file

packages/cli-node/src/pacman/yarn/yarn.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  async loadLockfile(gitRef?: string): Promise<Lockfile> {
+    const lockfilePath = paths.resolveTargetRoot('yarn.lock');
+
+    let lockfileContents: string;
+    if (gitRef) {
+      lockfileContents = await GitUtils.readFileAtRef('yarn.lock', gitRef);
+    } else {
+      lockfileContents = await fs.readFile(lockfilePath, 'utf8');
+    }
```
tylerd-canva, line 69: definitely on the fence about exposing this git ref API versus just exposing the `parse` method on the lockfile

plugins/devtools-backend/package.json
```diff
@@ -38,9 +38,10 @@
     "test": "backstage-cli package test"
   },
   "dependencies": {
-    "@backstage/backend-common": "^0.25.0",
+    "@backstage/backend-common": "workspace:^",
```
tylerd-canva, line 41:
```suggestion
    "@backstage/backend-common": "^0.25.0",
```

oops

tylerd-canva:
the CI / Verify steps are failing because I haven't updated the API docs yet (which I don't think is worth doing until this has had some more eyes on it).

.gitignore
```diff
@@ -75,6 +75,7 @@ typings/
 
 # Node version directives
 .nvmrc
+.tool-versions
```
tylerd-canva, line 78: I use ASDF locally, I hope it was okay to add this.

Rugvip (commented):
Thank you for taking a stab at this! 👍 

Before diving in deeper I've got one question/request: are there any parts of this that you could see that could ship through separate changes in order to break this down a bit? It's going to be a very slow process to iterate on this and merge if we need to tackle it all at once.

For example, the `--alwaysPack` change seems pretty straightforward to split out, and perhaps the `Lockfile` changes too? It'll be a lot quicker to review and merge those separately

tylerd-canva:
> are there any parts of this that you could see that could ship through separate changes in order to break this down a bit

Yes, absolutely. I wanted to get a "working slice" together just to actually prove that it was feasible first, but if you're open to this change overall I would definitely prefer to split it out into smaller incremental bits where possible. There will still probably be one bigger PR at the end to get all the tests working with the new abstraction layer, though.

PR #27097, opened 2024-10-12
feat: add i18n support for scaffolder-react plugin

JounQin:


plugins/catalog-import/src/translation.ts
```diff
@@ -0,0 +1,162 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { createTranslationRef } from '@backstage/frontend-plugin-api';
+
+export const catalogImportTranslationRef = createTranslationRef({
+  id: 'catalog-import',
+  messages: {
+    buttons: {
+      backText: 'Back',
```
jhaals, line 22:
```suggestion
      back: 'Back',
```
Should be fine to name this just `back` as it's under buttons?

plugins/catalog-import/report.api.md
```diff
@@ -140,11 +142,13 @@ export { catalogImportPlugin };
 export { catalogImportPlugin as plugin };
 
 // Warning: (ae-forgotten-export) The symbol "StepperProvider" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "catalogImportTranslationRef" needs to be exported by the entry point index.d.ts
```
jhaals, line 145: This should be exported

plugins/scaffolder-react/src/translation.ts
```diff
@@ -0,0 +1,58 @@
...
+      createButtonText: 'Create',
+      reviewButtonText: 'Review',
+      stepIndexLabel: 'Step {{index, number}}',
+      nextButtonText: 'Next',
+    },
+    templateCategoryPicker: {
+      title: 'Categories',
+    },
+    templateCard: {
+      defaultDescription: 'No description',
```
jhaals, line 44:
```suggestion
      noDescription: 'No description',
```

plugins/scaffolder-react/src/translation.ts
```diff
@@ -0,0 +1,58 @@
...
+      noTemplatesFoundPrefix:
+        'No templates found that match your filter. Learn more about ',
+      noTemplatesFoundLinkText: 'adding templates',
+      noTemplatesFoundSuffix: '.',
+    },
+    templateOutputs: {
+      defaultTitle: 'Text Output',
+    },
+    workflow: {
+      defaultDescription: 'No description',
```
jhaals, line 55:
```suggestion
      noDescription: 'No description',
```

jhaals (changes requested):
Thanks for the PR, I've added some suggestions

plugins/catalog-import/report.api.md
JounQin: I just added it into `alpha.ts` entry like other plugins.

plugins/catalog-import/src/translation.ts
```diff
@@ -0,0 +1,163 @@
...
+      githubIntegration: {
+        title: 'Link to a repository',
+        label: 'GitHub only',
+        descriptionPrefix: 'The wizard discovers all ',
+        descriptionSuffix:
+          ' files in the repository, previews the entities, and adds them to the {{appTitle}} catalog.',
+        prDescriptionPrefix:
+          'If no entities are found, the wizard will prepare a Pull Request that adds an example ',
+        prDescriptionSuffix:
+          ' and prepares the {{appTitle}} catalog to load all entities as soon as the Pull Request is merged.',
```
jhaals, line 49: Try to avoid having prefix and suffix as it's very fragile and might not work well in how sentences are structured in some languages. Try to go with interpolation or worst case leave them untranslated.

plugins/scaffolder-react/src/translation.ts
```diff
@@ -0,0 +1,58 @@
...
+    templateCard: {
+      noDescription: 'No description',
+      chooseButtonText: 'Choose',
+      noTemplatesFoundPrefix:
+        'No templates found that match your filter. Learn more about ',
+      noTemplatesFoundLinkText: 'adding templates',
+      noTemplatesFoundSuffix: '.',
+    },
+    templateOutputs: {
+      defaultTitle: 'Text Output',
```
jhaals, line 52:
```suggestion
      title: 'Text Output',
```

jhaals (changes requested):
Here are some more suggestions :)

plugins/catalog-import/src/translation.ts
JounQin: I'm not using interpolation here because the texts are used with React elements.

Ideally we should use [`Trans`](https://react.i18next.com/latest/trans-component) provided by `react-i18next`, but we are not exposing `i18next` interface in our translation plugin API, so I choose to use `prefix` and `suffix` instead temporarily, it should work for all languages although because they are in pair?

JounQin: @jhaals I tried to implement a `Translation` component at https://github.com/backstage/backstage/pull/27193/files#diff-9a9b8f6706578fbe2e5510acb56ff0b3de1767daf0bc3f2ff1c71ae86796c838R74 but `build:api-reports` runs out of memory...

JounQin: The tsc OOM disappears after upgrading i18next libs.

JounQin:
ping

jhaals (changes requested):
Really need those api-reports as well :) Try running `yarn build:api-reports` from the root of the repo. I tried locally and it works for me with this PR

plugins/catalog-import/src/translation.ts
Rugvip: @JounQin thank you! I'm thinking it's best to move this to a new PR and sort this change out by itself first?

On the direction, at a glance I'm thinking that the alternate usage of the `Trans` component might fit a bit better? https://react.i18next.com/latest/trans-component#alternative-usage-which-lists-the-components-v11.6.0. It avoids the need to list the default message multiple times.

acierto:
Hey @JounQin, maybe you can split this PR to two, as it affects 2 different areas of maintainers.
It can go faster if each PR will affect each area separately.

plugins/catalog-import/src/translation.ts
JounQin: https://github.com/backstage/backstage/pull/27193/files#diff-1eeb293ff108472b37a75835f10938ed211d2fed61c1cbdf56d451dd19c5dcb0R298

I was using `Trans` component inside.

JounQin:
OK, I'll split this PR to two later.

plugins/catalog-import/src/translation.ts
JounQin: Component support is added in #27193, can we migrate later?

JounQin:
ping

JounQin:
bump

JounQin:
bump

plugins/scaffolder-react/src/translation.ts
```diff
@@ -0,0 +1,58 @@
...
+      stepIndexLabel: 'Step {{index, number}}',
+      nextButtonText: 'Next',
+    },
+    templateCategoryPicker: {
+      title: 'Categories',
+    },
+    templateCard: {
+      noDescription: 'No description',
+      chooseButtonText: 'Choose',
+      noTemplatesFoundPrefix:
```
benjdlambert, line 46: Maybe lets have `templatesNotFoundText` and `templatesNotFoundLinkText` as the two fields, and always render the period instead.

benjdlambert (changes requested):
This looks pretty good, just a few questions about one translation.

plugins/scaffolder-react/src/translation.ts
JounQin: Period will be different in different languages, and the text order would be different in Chinese at least, we won't say `aaa. Learn more about [xxx].` but `aaa。访问 [xxx] 学习更多内容。`

And these are just a temporary solution for jsx, it would be replaced by #27193

JounQin: @benjdlambert News on this?

benjdlambert: I just commented in the attached PR https://github.com/backstage/backstage/pull/27193#issuecomment-2636494789

JounQin: I noticed #27193, but that's a new feature, should that block this PR?

benjdlambert: I think so yes, as we don't want to add translations and then immediately break them.  

The messages are the API after all, the `prefix` / `suffix` pattern is something we want to avoid.

JounQin: > Maybe we could leave the problematic translations as they are for now and just undo those ones to get the rest in and then come back to the those translations once we have a way to support these nested translations

Done. I've reverted it back.

benjdlambert:
@JounQin looks like the e2es are failing here? can you have a look into what's going on? :pray:

benjdlambert:
Ping @JounQin, looks like the e2es are still failing here :pray:

JounQin:
@benjdlambert It's passing now.

benjdlambert (approved):
Thanks!

PR #27100, opened 2024-10-14
nerf incremental log levels

freben:
The incremental provider is way too noisy in its logging during normal operation. Especially egregious are the "rest period continuing" messages. Let's pull down most of these to a lower level.


Rugvip (approved):
:+1:

Issue #26970, opened 2024-10-04
🐛 Bug Report: Unable to change `ItemCardHeader` Used for Template Cards with override


awanlin:
### 📜 Description

We want to be able to set the styles for the `ItemCardHeader` used for Template Cards seen from Create. In our custom theme we tried using:

```tsx
BackstageItemCardHeader: {
      styleOverrides: {
        root: {
          backgroundImage: 'unset',
          boxShadow: 'unset',
        },
      },
    },
```    
  
Sadly this doesn't seem to work

### 👍 Expected behavior

Using the `BackstageItemCardHeader` override should work for all Cards.

### 👎 Actual Behavior with Screenshots

Using the `BackstageItemCardHeader` override should work for all Cards but does not seem to work for the Cards when hey are listed under Create.

### 👟 Reproduction steps

There's a lot of steps to this so going to just document it at a high level:

1. Create a Backstage instances
2. Add a custom theme
3. Add an override like this:

```tsx
BackstageItemCardHeader: {
      styleOverrides: {
        root: {
          backgroundImage: 'unset',
          boxShadow: 'unset',
        },
      },
    },
```   

4. Start up the Backstage instance
5. Make sure you are using your custom theme
6. Navigate to Create

**Notice: the Card still have their header image.**

### 📃 Provide the context for the Bug.

As we aren't able to styles these cards the UX is disjointed as some things are styled properly but other things are not

### 🖥️ Your Environment

```text
OS:   Darwin 24.0.0 - darwin/arm64
node: v18.18.2
yarn: 4.5.0
cli:  0.28.0-next.1 (installed)
backstage:  1.32.0-next.1

Dependencies:
  @backstage/app-defaults                                          1.5.12-next.1
  @backstage/backend-app-api                                       1.0.1-next.0
  @backstage/backend-common                                        0.23.3, 0.25.0
  @backstage/backend-defaults                                      0.5.1-next.1
  @backstage/backend-dev-utils                                     0.1.5
  @backstage/backend-openapi-utils                                 0.1.19-next.0, 0.1.18
  @backstage/backend-plugin-api                                    1.0.1-next.0, 0.7.0, 1.0.0
  @backstage/catalog-client                                        1.7.0
  @backstage/catalog-model                                         1.7.0
  @backstage/cli-common                                            0.1.14
  @backstage/cli-node                                              0.2.8
  @backstage/cli                                                   0.28.0-next.1
  @backstage/config-loader                                         1.9.1
[...]

PR #27106, opened 2024-10-14
Fix ItemCardHeader override

jacri901:
Closes https://github.com/backstage/backstage/issues/26970

This fixes a bug where custom themes could not override `ItemCardHeader`.


jacri901:
@Rugvip, I need some more time to fix the things that blocking CI to pass. 

I try to do this
image
 
I'm having problems to generate the api-report. I can build backstage with `yarn build:all` but when I try

camilaibs (changes requested):
Thank you for the fix @jacri901 🎉 
Could you please update the `api-reports` to resolve the conflicts?

jacri901:
> Thank you for the fix @jacri901 🎉 Could you please update the `api-reports` to resolve the conflicts?

Yes sorry, I'll fix that later this week 😃

camilaibs (approved):
Hi @jacri901,
I tested the changes locally and it is working like a charm. So I'm going to approve and merge the pull request! Thanks again for contributing with us 🙌🏻 

image

image

jacri901:
@camilaibs thank you!

Unfortunally I got remined now that there was a test that did not pass. The previous implementation created a variable that was used as a type in the export of 'ItemCardHeaderProps'. However, this may possibly been the casue of why the overrides did not work previously. At least the way of how arguments was used in the 'makesStyles' function.

When I looked around in similar components they do not use types in the export like the previous implementaion of  'ItemCardHeaderProps' did. I tried to find a workaround previously but was unsuccesful. 

I'm open for suggestions 😄

camilaibs:
> @camilaibs thank you!
> 
> Unfortunally I got remined now that there was a test that did not pass. The previous implementation created a variable that was used as a type in the export of `ItemCardHeaderProps`. However, this may possibly been the cause of why the overrides did not work previously. At least the way of how arguments was used in the `makesStyles` function.
> 
> When I looked around in similar components they do not use types in the export like the previous implementaion of `ItemCardHeader` did. I tried to find a workaround previously but was unsuccesful.
> 
> I'm open for suggestions 😄

I think you can try something like this:
```tsx
function ItemCardHeader(props: ItemCardHeaderProps) {
  const { title, subtitle, children } = props;
  const classes = useStyles();
  return (
    <Box className={classNames(classes.root, props.classes?.root)}> 👈🏻 
      ...
    </Box>
  );
}

jacri901:
I have looked at this a bit more.
The suggestion you had @camilaibs makes the test pass but it breaks the override again.

When I looked in /report.api.md `ItemCardGridProps` was the only thing that used `Partial<WithStyles<typeof` `styles>>` like `ItemCardHeaderProps `. I tested overrides with `ItemCardGridProps` and that implementation works, eventhough it is more or less identical to the previous version of `ItemCardHeaderProps` where the override does not work.

awanlin:
Hi @camilaibs @jacri901, any chance you can look into this early in the new year? Would be great to have this fixed. 👍

mpsanchis:
Don't stale 😄

PR #27109, opened 2024-10-14
cli: link to logger service docs in backend plugin template

Rugvip:
🧹, see https://github.com/backstage/backstage/pull/27096#discussion_r1799452550


freben (approved):
🥳

PR #27111, opened 2024-10-14
Fixing issue with `ParamKeys` type inferrence

benjdlambert:
Bonus addition for `@ts-expect-errror` in some test cases as this is now checked in TypeScript.

### DIFF
--- a/.changeset/thin-doors-rule.md
+++ b/.changeset/thin-doors-rule.md
@@ -1,0 +1,5 @@
+---
+'@backstage/core-plugin-api': patch
+---
+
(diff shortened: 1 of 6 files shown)


Rugvip (approved):
:shipit: 🎉 

Bit more context on this one: We found that essentially `const ref: SubRouteRef = subRouteRef` was failing across versions. Not quite sure why this is as the type hasn't changed, but our theory is that TypeScript is able to take some shortcuts in the type checking when they're from the same module, while it otherwise has to do full structural type checking.

PR #27116, opened 2024-10-14
Fix API Report warning validation

benjdlambert:
There was a small issue in the `api-reports` that it didn't build the filename for the report path correctly, which mean that it failed silently, and `0` was the errors in all files.

This fixes the reports so that they actually warn, as well as fixing all the errors that actually came out of the migration to latest API extractor too.


.changeset/pink-sheep-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/repo-tools': patch
+---
+
+Fix issues with warnings not being reported due to bad filename path when reading report
```
Rugvip, line 5:
```suggestion
Fix issues with warnings in API reports not being checked or reported.

Due to the recent version bump of API Extractor you may now see a lot of `ae-undocumented` warnings, these can be ignored using the `-o` option, for example, `backstage-repo-tools api-reports -o ae-undocumented,ae-wrong-input-file-type`.
```

Rugvip (approved):
Nice! 🎉 👍 

Just changeset that could use a lil bit more info

Issue #26708, opened 2024-09-16
🐛 Bug Report: Events HTTP Verification Breaks Due to JSON Parsing Changing Request Body


codingdiaz:
### 📜 Description

The events backend plugin router is using the [express.json()](https://github.com/backstage/backstage/blob/master/plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts#L74) middleware. JSON payloads are [already parsed](https://github.com/backstage/backstage/blob/master/plugins/events-node/src/api/http/validation/RequestDetails.ts#L22) by the time a request verifier gets the payload to verify the request. 

There can be very subtle differences in the parsed request body compared to the raw body which will cause verification to fail for valid events. 

For example this is an example for a difference between a GitHub webhook event and the parsed event:

```diff
<           "score": 5.0
---
>           "score": 5
```

The raw un-parsed body should be sent to request verifiers to ensure proper validation.

### 👍 Expected behavior

The raw un-parsed body should be sent to request verifiers to ensure proper validation.

### 👎 Actual Behavior with Screenshots

A parsed and modified body was sent to a verifier causing a valid event to fail.

### 👟 Reproduction steps

This was a little hard to replicate and discover. I had to find some specific GitHub webhook events that were failing (due to parse differences) and re-send them with some logging.

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

camilaibs:
Thank you for opening the issue, it makes sense.
We are happy to provide feedback in a pull request if there is an opportunity for you to contribute with a fix 🙂

benjdlambert:
I think that this is fixed now in #27121. Closing!

PR #27121, opened 2024-10-14
fix(events,github): fixes signature validation by using raw req body

pjungermann:
Adds raw body information (body as buffer, encoding) to `RequestDetails` to support more request validation use cases.

Additionally, uses the raw body to retrieve the transmitted JSON string unparsed/raw to correctly validate the signature.

Previously, we re-stringified the parsed JSON payload which could lead to different JSON strings.
Those differences can lead to the rejection of requests due to a mismatch with the expected signature.

Fixes: #26709
Relates-to: PR #26884


pjungermann:
Seems like the bump of the browser action in PR #27087 broke the "setup-chrome" step at the E2E tests.

.changeset/shiny-falcons-fly.md
```diff
@@ -0,0 +1,8 @@
+---
+'@backstage/plugin-events-node': minor
```
Rugvip, line 2:
```suggestion
'@backstage/plugin-events-node': patch
```

Think we can keep this as a patch, no breaking changes there

plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
```diff
@@ -87,25 +98,53 @@ export class HttpPostIngressEventPublisher {
...
+      const encoding = parsedContentType.parameters.charset ?? 'utf-8';
+      if (!Buffer.isEncoding(encoding)) {
+        throw new UnsupportedCharsetError(encoding);
+      }
+
+      const bodyString = bodyBuffer.toString(encoding);
+      const bodyParsed =
+        parsedContentType.type === 'application/json'
+          ? JSON.parse(bodyString)
+          : bodyString;
```
Rugvip, line 122: Previously we would just throw an error if the content type wasn't supported, thinking we might as well keep doing that right? i.e. if it's not JSON

Rugvip (commented):
Nice, thank you! 👍

plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
pjungermann: yes, we could. Even for just now and rediscuss later. Just thought to remove this limitation that was not necessarily put in place because we must.

.changeset/shiny-falcons-fly.md
pjungermann: I used `minor` due to the required fields added to `RequestDetails`. However, that should only impact tests and no production code for request validation. Therefore, I guess `patch` might be ok as well. 🤔 

I will adjust it. Let me know if that info changes your stand here.

plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
pjungermann: I will add the rejection of other content types. We can decide later, whether we want to open this up more for other content types (and maybe only specific ones).

codingdiaz:
This will also fix: #26708

pjungermann:
After rebase, there is now
```
/usr/bin/pip install docutils
error: externally-managed-environment
```

benjdlambert:
@pjungermann give it another rebase and hopefully errors be gone.

Rugvip (approved):
Thank you, let's :shipit:! 🎉

benjdlambert (approved):

PR #27124, opened 2024-10-14
Update to Storybook 8

cdedreuille:
Storybook 8

- Update Storybook to version 8 using Vite as builder.
- Removed `storybook-dark-mode` in favour of `@storybook/addon-themes`. This will allow us to use more than two themes.
- Update `dialog.stories.tsx` as a good example on how stories should be written using args.
- Added `@storybook-react` dependency to `core-components` to be used in stories. This is not the ideal solution as we might need to update this dependency in multiple places but this is a good first step to make sure stories are typed correctly.

To make the process work I had to deactivate 3 stories. We will work on them in separate PRs:
- **GroupProfileCard.stories.tsx**: The Storybook team is working on fix - [More details here](https://github.com/storybookjs/storybook/issues/29378)
- **MembersListCard.stories.tsx**: This was not visible on the existing Storybook but need to be addressed.
- **MarkdownContent.stories.tsx**: is failing locally but not on the published version of Storybook. We need to investigate this further.


Rugvip (approved):
Awesome, looks good! 😁

TS config is nice, and I think we may want to re-evaluate the separation of the storybook into a separate workspace. The dependency list is for sure a lot slimmer now, which is big part of the reason we split it out in the first place.

Let's :shipit: 🎉

PR #27130, opened 2024-10-15
try to improve log tests stability

freben:
We were seeing lots of flaky test runs in this package, e.g. [this](https://github.com/backstage/backstage/actions/runs/11329221423/job/31504442372?pr=27107#step:11:52) and similar.

Seems that it's not only windows that has some nondeterminism in these log based tests. Since we explicitly don't really care all that much about the order or presence/non-presence of those extra lines, let's update the test to mirror that fact.


Rugvip (approved):
🎉

PR #27139, opened 2024-10-15
docs/releases: add 1.32

Rugvip:
Release notes for 1.32 📝

### DIFF
--- a/docs/releases/v1.32.0.md
+++ b/docs/releases/v1.32.0.md
@@ -1,0 +1,160 @@
+---
+id: v1.32.0
+title: v1.32.0
+description: Backstage Release v1.32.0
+---
+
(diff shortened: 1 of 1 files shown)


docs/releases/v1.32.0.md
```diff
@@ -0,0 +1,160 @@
...
+                - acme.org
+```
+
+See also the full [identity resolvers documentation](https://backstage.io/docs/auth/identity-resolver) for more details.
+
+### **BREAKING**: CLI Changes
+
+We’ve been doing some cleanup and removed some deprecated commands, updated as follows:
+
+`create`: Use `backstage-cli new` instead
```
benjdlambert, line 40: This has lost the formatting and looks poop 😓

PR #27167, opened 2024-10-16
chunk up refs in `getEntitiesByRefs`

freben:
Sometimes in huge catalogs, callers try to fetch large numbers of entities by refs at the same time. An example of this is in the owner picker when there's an extremely large number of potential owners. This can lead to hitting Express.js payload size limits.

The owner picker itself should be updated to be smarter about what data it fetches as well, of course. But fixing this in the client ensures that all other, hitherto unknown similar problems are fixed at the same time, and that the server isn't suffering from trying to serve overly large requests.

### DIFF
--- a/packages/catalog-client/src/utils.ts
+++ b/packages/catalog-client/src/utils.ts
@@ -24,3 +24,67 @@ export function isQueryEntitiesInitialRequest(
 ): request is QueryEntitiesInitialRequest {
   return !(request as QueryEntitiesCursorRequest).cursor;
 }
+
+/**
+ * Takes a set of entity refs, and splits them into chunks (groups) such that
+ * the total string length in each chunk does not exceed the default Express.js
+ * request body limit of 100 kB (with some margin) when JSON encoded as an
+ * array.
+ */
+export function splitRefsIntoChunks(
+  refs: string[],
+  options?: {
+    // No chunk has more than this many refs, no matter what
+    maxCountPerChunk?: number;
+    // The total string length (taking the extraStringLengthPerRef into account)
+    // of each chunk never exceeds this many characters, no matter what
+    maxStringLengthPerChunk?: number;
+    // Add this many characters to the length of each ref when calculating
+    // (default is 3, since eacn array entry is surrounded by quotes and a
+    // comma)
+    extraStringLengthPerRef?: number;
+  },
+): string[][] {
+  if (!refs.length) {
+    return [];
+  }
+
+  const {
+    maxCountPerChunk = 1000,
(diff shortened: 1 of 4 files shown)


plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
```diff
@@ -57,33 +58,40 @@ export function useFacetsEntities({ enabled }: { enabled: boolean }) {
...
+    // the typical catalog installation
+    const entities = new Array<Entity | undefined>();
+    for (const refsChunk of chunk(entityRefs, 1000)) {
+      Array.prototype.push.apply(
+        entities,
+        await catalogApi
+          .getEntitiesByRefs({ entityRefs: refsChunk })
+          .then(r => r.items)
+          .catch(() => []),
+      );
```
Rugvip, line 72: `entities.push(...newEntities)` plz

freben: But performance!

Hehe, maybe it turns out to be the same thing behind the scenes anyway

freben: Fixed

packages/catalog-client/src/utils.ts
```diff
@@ -24,3 +24,67 @@ export function isQueryEntitiesInitialRequest(
...
+    extraStringLengthPerRef?: number;
+  },
+): string[][] {
+  if (!refs.length) {
+    return [];
+  }
+
+  const {
+    maxCountPerChunk = 1000,
+    maxStringLengthPerChunk = 90 * 2 ** 10,
```
Rugvip, line 54: image

Rugvip (approved):
Nice 👌

vinzscam:
> Sometimes in huge catalogs, callers try to fetch large numbers of entities by refs at the same time. An example of this is in the owner picker when there's an extremely large number of potential owners. This can lead to hitting Express.js payload size limits.
> 
> The owner picker itself should be updated to be smarter about what data it fetches as well, of course. But fixing this in the client ensures that all other, hitherto unknown similar problems are fixed at the same time, and that the server isn't suffering from trying to serve overly large requests.

Sorry, I don't think this is the case for the owner picker, as it only calls `getEntitiesByRefs`, passing only the selected entities, which is usually a low number. 🤔 I'm worried that having such optimization will incentivize people to call `getEntitiesByRefs` with a large number of entities, which isn't a good practice.

PR #27177, opened 2024-10-16
feat: Added custom timeout for oidc provider

goququ:

Hi folks! I have used [this way](https://discord.com/channels/687207715902193673/978977570852835379/1070438453944270998) to handle oauth requests time out. But seems like this is not working now. Now the code was changed and there is no way how to override the timeout.

image

I suggest configuring timeout using config file. Here is my attempt to do this.

### DIFF
--- a/.changeset/early-feet-lay.md
+++ b/.changeset/early-feet-lay.md
@@ -3,3 +3,15 @@
 ---
 
 Added custom timeout setting for oidc provider
+
+Here is an example of how to use a custom timeout with the configuration:
+
+```yaml
+auth:
+  oidc:
+    production:
+      clientId: ${AUTH_GOOGLE_CLIENT_ID}
+      clientSecret: ${AUTH_GOOGLE_CLIENT_SECRET}
+      timeout:
+        seconds: 30
+```
--- a/.changeset/early-feet-lay.md
+++ b/.changeset/early-feet-lay.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-auth-backend-module-oidc-provider': minor
+---
+
+Added custom timeout setting for oidc provider
--- a/plugins/auth-backend-module-oidc-provider/src/authenticator.test.ts
+++ b/plugins/auth-backend-module-oidc-provider/src/authenticator.test.ts
@@ -28,6 +28,7 @@ import { ConfigReader } from '@backstage/config';
 import { JWK, SignJWT, exportJWK, generateKeyPair } from 'jose';
 import { rest } from 'msw';
 import express from 'express';
+import { custom } from 'openid-client';
 
 describe('oidcAuthenticator', () => {
   let implementation: any;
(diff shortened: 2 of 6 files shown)


plugins/auth-backend-module-oidc-provider/src/authenticator.ts
```diff
@@ -84,6 +86,9 @@ export const oidcAuthenticator = createOAuthAuthenticator({
         'The oidc provider no longer supports the "scope" configuration option. Please use the "additionalScopes" option instead.',
       );
     }
+    const httpOptionsProvider = createHttpOptionsProvider({
+      timeout: config.getOptionalNumber('timeout'),
```
deejay1, line 90: you need to add this to `config.d.ts` also

deejay1 (changes requested):

plugins/auth-backend-module-oidc-provider/src/authenticator.ts
deejay1: also it would be good if this would be of type `HumanDuration` as in other places in Backstage

goququ: Hi @deejay1, thanks for your feedback. I have changed the code. Please have a look 🙂

plugins/auth-backend-module-oidc-provider/src/authenticator.ts
```diff
@@ -85,6 +88,16 @@ export const oidcAuthenticator = createOAuthAuthenticator({
       );
     }
 
+    const timeoutHumanDuration = config.getOptional<HumanDuration>('timeout');
```
aramissennyeydd, line 91: Can we update this to https://github.com/backstage/backstage/blob/master/packages/config/src/readDurationFromConfig.ts#L47?

deejay1: like this 
```typescript
const timeoutMilliseconds = config.has('timeout')
      ? durationToMilliseconds(
          readDurationFromConfig(config, { key: 'timeout' }),
        )
      : undefined;
```

goququ: @deejay1 @aramissennyeydd thank you for your comments! Just updated the code. Please have a look. Have a nice one 🙂

deejay1: @goququ please check the errors in the test

deejay1 (approved):

.changeset/early-feet-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-auth-backend-module-oidc-provider': minor
+---
+
+Added custom timeout setting for oidc provider
```
camilaibs, line 5: Could you please add an example here on how the custom config can be used in the `app-config.yaml` file?

plugins/auth-backend-module-oidc-provider/src/authenticator.ts
```diff
@@ -85,6 +89,15 @@ export const oidcAuthenticator = createOAuthAuthenticator({
       );
     }
 
+    const timeoutMilliseconds = config.has('timeout')
```
camilaibs, line 92: Could you please add a test for the new configuration? By doing that, we make sure that the test fails if someone accidentally removes the configuration support 🙏🏻

camilaibs (changes requested):
Thanks for fixing it @goququ !
Left just two minor comments 🙂

.changeset/early-feet-lay.md
goququ: Hi, thank you for your feedback. I have added an example.

plugins/auth-backend-module-oidc-provider/src/authenticator.ts
goququ: Also added some tests. Please have a look. Thanks!

goququ:
@camilaibs Hi. Could you please check the changes one more time? Thanks!

camilaibs (approved):
I apologize for the delay in re-reviewing, it looks great to me, thank you again for your contribution 🙌🏻

plugins/auth-backend-module-oidc-provider/src/authenticator.test.ts
```diff
@@ -169,6 +170,81 @@ describe('oidcAuthenticator', () => {
...
+          callbackUrl: 'https://backstage.test/callback',
+          config: new ConfigReader({
+            metadataUrl: 'https://oidc.test/.well-known/openid-configuration',
+            clientId: 'clientId',
+            clientSecret: 'clientSecret',
+            timeout: '30s',
+          }),
+        });
+      }).toThrow(
+        "Failed to read duration from config, TypeError: Invalid type in config for key 'timeout' in 'mock-config', got string, wanted object",
```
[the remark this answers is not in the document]
goququ: The error message comes from `readDurationFromConfig`, and I can't modify it directly. Honestly, I don't want to catch the error just to tweak the message. Let's leave it as is. 🙂

goququ:
@Rugvip @aramissennyeydd @deejay1

Hi all! Could you please review my changes? Thanks a lot!

Issue #27197, opened 2024-10-16
🚀 Feature: Add `spec.tier` as an optional Component entity descriptor


SocalNick:
### 🔖 Feature description

Similar to `spec.lifecycle`, a Component entity could specify its "tier" in an organization's software ecosystem. Tier would indicate the criticality of the service, e.g. `tier: 0` would be the most critical, `tier: 1` less critical, ... `tier: n` least critical.

### 🎤 Context

Our use case is to be able to quickly see our Components / Services that are most critical to our business. Criticality could be used to add additional production readiness criteria, operational requirements, etc.

### ✌️ Possible Implementation

[I imagine this would be added here as an optional string](https://github.com/backstage/backstage/blob/master/packages/catalog-model/src/kinds/ComponentEntityV1alpha1.ts#L36).

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

aramissennyeydd:
👋 Is this something you've implemented already at your company and are looking to contribute back? If not, I'd recommend you start there with [this guide](https://backstage.io/docs/features/software-catalog/extending-the-model/) as this can be achieved without any upstream changes. Catalog model changes tend to be arduous to get in, as they will affect everybody :)

awanlin:
Curious as to why this has to be on `spec` and not simply using `labels` for this? Labels are already built into the model and a great place for this information, I've seen this exact setup with a few Backstage Adopters.

SocalNick:
@awanlin part of my motivation for establishing this as part of the specification is interoperability w/ Datadog's service catalog. We have Datadog configured to import our Backstage catalog. It's nice that there's interoperability w/ specification fields like lifecycle but we now have a use case to add tier to both Backstage and Datadog.

awanlin:
Thanks for the follow up @SocalNick, still feel like something you can do without it being part of the model though as previously mentioned.

awanlin:
Wait, why can't they pull in labels? That would make a lot more sense to me now that I'm thinking about it as it would open up more options beyond just this.

PR #27212, opened 2024-10-17
Add `spec.tier` as an optional Component entity descriptor

SocalNick:
Similar to `spec.lifecycle`, a Component entity could specify its "tier" in an organization's software ecosystem. Tier would indicate the criticality of the service, e.g. `tier: '0'` would be the most critical, `tier: '1'` less critical, ... `tier: 'n'` least critical.

Our use case is to be able to quickly see our Components / Services that are most critical to our business. Criticality could be used to add additional production readiness criteria, operational requirements, etc.

Resolves https://github.com/backstage/backstage/issues/27197

Signed-off-by: Nicholas Calugar <nicholas.calugar@gusto.com>

### DIFF
--- a/.changeset/fair-shrimps-ring.md
+++ b/.changeset/fair-shrimps-ring.md
@@ -1,0 +1,5 @@
+---
+'@backstage/catalog-model': minor
+---
+
+Add `spec.tier` as an optional Component entity descriptor
--- a/docs/features/software-catalog/descriptor-format.md
+++ b/docs/features/software-catalog/descriptor-format.md
@@ -609,6 +609,13 @@ which the component is a part, e.g. `spotify-ios-app`. This field is optional.
 | ---------------------------------------- | ------------------------------------------ | ----------------------------------------------------------------------------- |
 | [`Component`](#kind-component) (default) | Same as this entity, typically `default`   | [`partOf`, and reverse `hasPart`](well-known-relations.md#partof-and-haspart) |
 
+### `spec.tier` [optional]
+
+The tier of the component, e.g. `0` or `critical`. This field is optional.
+
+The software catalog accepts any tier value, but an organization should
+take great care to establish a proper taxonomy for these.
+
 ### `spec.providesApis` [optional]
 
 An array of [entity references](references.md#string-references) to the APIs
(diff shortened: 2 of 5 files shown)


ssarne:
There is a balance to strike here, between making the metadata minimal and expressive enough. Declaring criticality on a component is quite valuable, while I honestly don't know how granular it needs to be.

At Spotify we have been using "tier" for a couple of years, even if we have called it serviceTier. What we see though is that having one tier value isn't enough, we see different interpretations of it ranging from business criticality to reliability expectations. To clarify this, we rename it to reliabilityTier and let it apply only to reliability aspects and also use it across our different disciplines of backend, client, data, ml and web. Declaring this intent is very useful for teams owning a component.

Over time we have also added securityTier, which we actually derive and don't declare and use to describe and validate security expectations. Similarly we have also added dataValueTier which also became a bit ambiguous because it both handles reliability aspects and data quality aspects. Using reliability tier for the timeliness aspect will be helpful.

We still have conversations on this topic though, can we combine them into one or how many do we need.  =)

ssarne:
Yeah, we have them as fields in the `spec` next to lifecycle
(actually not security tier, separate lookup for historical reasons, likely we will change that)

SocalNick:
Hey folks - anything else I can provide to help land this PR? Thanks!

awanlin:
Hi @SocalNick, I don't this is something that will get moved on quickly. Changes like this take a fair amount of consideration. If you need this then I would follow the recommendation that was posted on the issue your logged: https://github.com/backstage/backstage/issues/27197#issuecomment-2418304715

vinzscam:
Actually you don't need this PR to be merged in order to use this feature in your Backstage. It should work out of the box without this change.

freben:
^ this is correct, but it's recommended to add a custom processor that validates the new field of course. Feel free to model it after the one [for the common model](https://github.com/backstage/backstage/blob/e8b52d4f8f58d6c3f92f4bf6c61b5f68b16a7171/plugins/catalog-backend/src/processors/BuiltinKindsEntityProcessor.ts#L59) (but you only need to validate the "additions", not the whole kind, when you amend an existing one)

SocalNick:
Thanks for all the comments folks. As I mentioned, my motivation is around interoperability w/ the Datadog service specification. Making it officially part of the Backstage specification would make it more likely that they would incorporate this w/ their capability to ingest Backstage service catalog.

PR #27213, opened 2024-10-17
feat(cli): move config to its own "module" in the CLI

aramissennyeydd:

Based on @Rugvip's comment [here](https://github.com/backstage/backstage/pull/24904#pullrequestreview-2342349766).

Tested locally with all 4 config commands and everything looked good 😁 

### DIFF
--- a/packages/cli/src/commands/build/buildFrontend.ts
+++ b/packages/cli/src/commands/build/buildFrontend.ts
@@ -18,7 +18,7 @@ import fs from 'fs-extra';
 import { resolve as resolvePath } from 'path';
 import { buildBundle, getModuleFederationOptions } from '../../lib/bundler';
 import { getEnvironmentParallelism } from '../../lib/parallel';
-import { loadCliConfig } from '../../lib/config';
+import { loadCliConfig } from '../../modules/config/lib/config';
 
 interface BuildAppOptions {
   targetDir: string;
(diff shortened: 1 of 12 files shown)


Rugvip (changes requested):
Nice! 👍 

I've been experimenting a lil bit as well, and tbh I'm kinda thinking we might want to move away from `commander` x) I'd like much better type support for options and args.

Somewhat related to that, could you make this refactor internally within the CLI package instead? We can for example have a `modules` folder or such that mocks a cross-package setup. I'd just like to avoid any form of public API or dependencies to maintain as we iterate on the layout.

Regarding the actual API across commands, I've been thinking that we likely just want the different modules to essentially just receive an `argv` that it parses itself. That'd ensure a very minimal coupling across modules, which I think we can get away with tbh. Then it might be that we need some form of shared (by read-only) context for commands, but that'd be something I think we can discovery as we go along.

aramissennyeydd:
@Rugvip Took a bit of a look at what a commander-less usage would be in [e8ffcf4](https://github.com/backstage/backstage/pull/27213/commits/e8ffcf4ae8e5063cfab257ee78a82f39670dae40). One of the things that I think we'd want to keep is the ability to list what commands are registered across modules - if we do some very light parsing + validation at the framework level, I think we get a bit more control there. I'm not convinced we should just leave that up to the plugins/modules to define.

aramissennyeydd:
Trying out a bit more `new backend`-feel wiring, https://github.com/backstage/backstage/pull/27266 with a separate entry point.

aramissennyeydd:
closing in favor of #27266

PR #27215, opened 2024-10-17
cli: refactor success cache to be additive

Rugvip:

This changes the success cache to store each entry in a marker file rather than all of them in a JSON file. This lets us keep old entries around and increases the chance for cache hits at a slight cost to storage and complexity. Still aiming to keep the implementation very simple, so the cache items are just empty files, all information needed is stored in the filename.

And a bit of a refactor to use the same implementation for the two commands.

### DIFF
--- a/packages/cli/src/lib/cache/SuccessCache.ts
+++ b/packages/cli/src/lib/cache/SuccessCache.ts
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import fs from 'fs-extra';
+import { resolve as resolvePath } from 'node:path';
+
+const DEFAULT_CACHE_BASE_PATH = 'node_modules/.cache/backstage-cli';
+
+const CACHE_MAX_AGE_MS = 7 * 24 * 3600_000;
+
+export class SuccessCache {
+  readonly #path: string;
+
+  constructor(name: string, basePath?: string) {
+    this.#path = resolvePath(basePath ?? DEFAULT_CACHE_BASE_PATH, name);
+  }
+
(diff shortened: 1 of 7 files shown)


packages/cli/src/lib/cache/SuccessCache.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  async read(): Promise<Set<string>> {
+    const state = await fs.stat(this.#path).catch(error => {
+      if (error.code === 'ENOENT') {
+        return undefined;
+      }
+      throw error;
+    });
+    if (!state || !state.isDirectory()) {
```
freben, line 38: you're permissive with non-directories here, but does `fs.ensureDir` handle it gracefully in `write()` below or will it throw if it finds a file?

packages/cli/src/lib/cache/SuccessCache.ts
```diff
@@ -0,0 +1,89 @@
...
+    for (const key of newEntries) {
+      // Remove any existing items with the key we're about to add
+      const trimmedItems = existingItems.filter(item =>
+        item.endsWith(`_${key}`),
+      );
+      for (const trimmedItem of trimmedItems) {
+        await fs.unlink(resolvePath(this.#path, trimmedItem));
+      }
+
+      await fs.writeFile(resolvePath(this.#path, `${now}_${key}`), empty);
```
freben, line 86: you're REALLY avoiding making like a yaml state file or so :)

freben: Come to think of it, i guess this actually is racy isn't it (delete and then create), if you have multiple concurrent builds going on. It's extremely unlikely of course that you'd do that AND have hash matches. :) So I'm not really worried. Otherwise you could do like, write first, then delete everything that matches with an older timestamp than that of the new entry. Or something. But whatever.

Rugvip: yeah seems very unlikely in the first place, and you also wouldn't be running multiple builds of this in the first place. It explicitly prevents mixing with the shard flag for example: https://github.com/backstage/backstage/blob/f98ca6919574a9c877b805ee6a708973ff9ade19/packages/cli/src/commands/repo/test.ts#L290-L294

Rugvip: Also yes, trying to keep it very simple. I did actually put the timestamp in the file content to start but realized that's gonna a be a bit of a performance hit to every time have to read through all the files to clean the cache

packages/cli/src/lib/cache/SuccessCache.ts
Rugvip: Nah it errors if there's a file already. Tbh this bit could error on that too, just felt ugly to call readDir of we know it's not a dir x)

freben (approved):

PR #27218, opened 2024-10-17
fix: disable blur-handler when selecting option from Autocomplete

### LINKED ISSUE #27214: 🐛 Bug Report: `EntityPicker` passes display name when `allowArbitraryValues` is `true` (as it is by default)
### 📜 Description

a while back I contributed a change to Backstage that modifies how `EntityPicker` shows selected value to the user: #26866

unfortunately that end up having a bug when `allowArbitraryValues` is true as in that case it will forward display name as value of field and that causes significant problems down the line.

### 👍 Expected behavior

When `allowArbitraryValues` is true it should pass entityref as it does when `allowArbitraryValues` is false.

image

### 👟 Reproduction steps

1. Go to: https://demo.backstage.io/create/custom-fields
2. Select "Artists"
3. Click *Review*
4. Check field value, it's "Artist" instead of "domain:default/artists".

_No response_

### 🖥️ Your Environment

_No response_

None

zcmander:

This fixes #27214 by introducing another state to the component in order to bypass blur handler if the user select an option from Autocomplete component. The issue here is that freesolo-mode allows freeform-text to the Textfield component and that is treated "as-is" while selected option from Autocomplete should be kept as entity reference. This is really hard thing to do as event happens in the following order:

 * The user selects Option from Autocomplete
 * Title from the selected Option in filled to Textfield complete and user's focus move to the field
 * User click outside of Textfield
 * Another change event is issued and now containing "freeform" input (entity display name as text)

So, this PR introduces another state; when user types something to the text field it known that it should be treated as text meanwhile selecting option from Autocomplete says in entity reference because `blur` effect is ignored as user didn't touch the text field (mode is changed via `keyDown` event).

At best; This is a hack but `EntityPicker` does so much already! I'm also fine with reverting back to old version :)

### DIFF
--- a/.changeset/moody-pandas-marry.md
+++ b/.changeset/moody-pandas-marry.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder': patch
+---
+
+disable blur-handler when selecting option from Autocomplete
--- a/plugins/scaffolder/src/components/fields/EntityPicker/EntityPicker.test.tsx
+++ b/plugins/scaffolder/src/components/fields/EntityPicker/EntityPicker.test.tsx
@@ -113,6 +113,7 @@ describe('<EntityPicker />', () => {
 
       const input = getByRole('textbox');
 
+      fireEvent.keyDown(input, {}); // trigger freeSolo mode
       fireEvent.change(input, { target: { value: 'squ' } });
       fireEvent.blur(input);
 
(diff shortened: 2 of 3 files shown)


acierto:
Hey @zcmander 👋 

I see your point from technical point of view. 

Can you please describe your use case when you actually need to do it? 
I wonder if it is possible to solve it with alternative way.

zcmander:
> Hey @zcmander 👋
> 
> I see your point from technical point of view.
> 
> Can you please describe your use case when you actually need to do it? I wonder if it is possible to solve it with alternative way.

I think our use case is explained quite well in my original PR, here: #26866  Basically it boils down to the fact that entity references are not user-friendly so it would be best to avoid them if we can. That includes `<Textfield>` component in the `<EntityPicker>` scaffolder field that is quite ugly when the actual dropdown shows pretty user-friendly names.

PR #27228, opened 2024-10-17
signals-backend: fan-out signal events in order to support scaled deployments

Rugvip:

By using a unique subscriber ID for each instance all events will be fanned out to all instances and reach users regardless of what instance they're connected to.

👀  @drodil 

### DIFF
--- a/plugins/signals-backend/src/service/SignalManager.ts
+++ b/plugins/signals-backend/src/service/SignalManager.ts
@@ -15,6 +15,7 @@
  */
 import { EventParams, EventsService } from '@backstage/plugin-events-node';
 import { SignalPayload } from '@backstage/plugin-signals-node';
+import crypto from 'crypto';
 import { RawData, WebSocket } from 'ws';
 import { v4 as uuid } from 'uuid';
 import { JsonObject } from '@backstage/types';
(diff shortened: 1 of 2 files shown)


plugins/signals-backend/src/service/SignalManager.ts
```diff
@@ -62,10 +63,18 @@ export class SignalManager {
   }
 
   private constructor(options: SignalManagerOptions) {
+    this.events = options.events;
```
freben, line 66: ...how did this ever work

freben (approved):

plugins/signals-backend/src/service/SignalManager.ts
```diff
@@ -62,10 +63,18 @@ export class SignalManager {
...
     ({ events: this.events, logger: this.logger } = options);
 
+    // Use a unique subscriber ID for each signals instance, in order to fan-out
+    // all events to each signals instance. This ensures that events always
+    // reach users in a scaled deployment.
+    const id = `signals-${crypto.randomBytes(8).toString('hex')}`;
```
drodil, line 72: Hope this is random enough to not collide

Rugvip: 1/18,446,744,073,709,551,615 + birthday paradox, but we won't have too many in flight at once. I think we should be fine x)

plugins/signals-backend/src/service/SignalManager.ts
Rugvip: ah, thought I removed that one

PR #27232, opened 2024-10-17
Patch release of #27227, #27230, #27231

benjdlambert:
This release fixes the following issues:

- Scaffolder forms would not update existing state when going back to update previous values
- Fixes an issue with the `EntityPicker` not setting the correct value in the form
- Fixes some resiliency issues with the new `events-backend` bus


Rugvip (approved):
👍

deejay1 (approved):

Issue #27220, opened 2024-10-17
🚀 Feature: catalog client in coreServices


grantila:
### 🔖 Feature description

Add the CatalogClient into coreServices for use in the backends, and allow an async `getToken()` function to be passed to the CatalogClient. The coreServices catalog client would then create a service token for each request (as is expected in backends).

### 🎤 Context

Today, everywhere in backends when calls to the catalog are made, one has to:
```ts
      const { token } = await this.auth.getPluginRequestToken({
        onBehalfOf: await this.auth.getOwnServiceCredentials(),
        targetPluginId: 'catalog',
      });
```
and then feed that token into each and every catalog call. You're not supposed to cache this either, but create a new one for each call, to ensure it's updated if necessary. This is cumbersome, and bloats code quite a bit.

It would be a fairly simple task to add an optional callback in the CatalogClient to an async `getToken()` which would simplify using the catalog (and have this invoked in every API function it has automatically). And even nicer (and a fairly simple contribution) to add this to the coreServices, so you don't even need to do that.

I get that you want to keep `backend-plugin-api` slim and dependency light, but the catalog client is most likely used (and imported) anyway in a backend, in practically all Backstage instances. If you'd prefer, this _service_ could be a separate package too.

I'm willing to contribute this, if you think it's a good idea. Either just the `getToken()` in the CatalogClient, or a backend service as well.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

freben:
One problem here is the `onBehalfOf`. What token to use depends on the context and the client currently has absolutely no way of knowing which one to use. Is this being done "as the service itself" (pure service token), or for the user (on-behalf-of token)? This is critically important for having the right permissions apply.

I'm not necessarily against adding the catalog client to the core services (personally), but we may want to look into the interface first. Most likely, if it were a core service, it wouldn't accept a token argument but instead a `credentials` argument or similar.

Maybe there's even things we can do in terms of storing the current incoming user identity in request handlers using continuation-local storage so it can be picked up transparently when present. But that's a bit more of a wild random thought, let's not build that upfront :)

Rugvip:
Fixing the credentials bit in #27234

PR #27234, opened 2024-10-17
plugins/catalog-node: add credentials support for catalog service + promote

Rugvip:

This makes it possible to pass credentials objects to the catalog service in the backend. It now has its own type as well, making it a bit less awkward to pass around. Fixes parts of #27220.

It also promotes the catalog service to the main entry point rather than alpha, although we could consider #27220 instead


plugins/catalog-node/report.api.md
```diff
@@ -102,6 +119,94 @@ export type CatalogProcessorResult =
...
+export const catalogServiceRef: ServiceRef<
+  CatalogService,
+  'plugin',
+  'singleton'
+>;
+
+// @public (undocumented)
+export interface CatalogServiceRequestOptions extends CatalogRequestOptions {
+  // (undocumented)
+  credentials?: BackstageCredentials;
```
freben, line 207: Should they really be optional?

Rugvip: yep could discuss going for the breaking change here instead, might not be too bad tbh. Especially since the catalog client will stay around

grantila:
Great! Looks very much like what I've done. However, to not repeat the Api, would it make sense to add support for providing the async `getOptions()` to `CatalogClient` instead, and have the `#getOptions` logic there?

Rugvip:
@grantila how would you structure that though? We can't reference `BackstageCredentials` in a common package but want to keep things type safe

freben (approved):
Nice!

plugins/catalog-node/report.api.md
Rugvip: Updated to only accept and require credentials https://github.com/backstage/backstage/pull/27234/commits/e4e74e720ea85c1d2c3aa3d8044c90244ef8cb10

Bit of an issue with that is that it's hard to support both old and new backend system at the same time with the new service, so we'll need to keep the alpha one around for a while to support that. I think that's an alright tradeoff though

etiennechabert:
Hey @Rugvip, thanks a lot for your contribution!

Sadly I couldn’t really get it to work on my side. I feel like a couple more examples or a bit more context around how BackstageCredentials work — and how we’re supposed to obtain them — would really help.

This doc explains how to get a token as a plugin:
https://backstage.io/docs/backend-system/core-services/auth/#creating-request-tokens

And this method takes that token and returns BackstageCredentials:
https://backstage.io/docs/reference/backend-plugin-api.authservice.authenticate

But when I try that (I’m calling the catalog plugin from another internal plugin), I just hit an auth error.

So right now I’m falling back to making HTTP requests to the Backstage API directly from my plugin, using a static token… which isn’t great.

Let me know if I’m missing something obvious — would love to do it the right way.

freben:
The receiving end uses `coreServices.httpAuth` and its [`credentials` method](https://backstage.io/docs/reference/backend-plugin-api.httpauthservice.credentials) to get the caller credentials.

What's the error? Are you giving it the right (receiving) plugin ID?

etiennechabert:
@freben thanks for the quick reply, and sorry for the lack of details provided.

Here is my full code:
```
    async function resolveSystemTags(entityRef: string): Promise<string[]> {
        const ref = parseEntityRef(entityRef);
        if (ref.kind.toLocaleLowerCase('en-US') === 'system') {
            return [ref.name];
        }

        if (ref.kind.toLocaleLowerCase('en-US') === 'group') {
            const service_credentials = await auth.getOwnServiceCredentials()
            const { token } = await auth.getPluginRequestToken({
                onBehalfOf: service_credentials,
                targetPluginId: 'catalog',
            });
            const credentials = await auth.authenticate(token) # will throw: "Failed plugin token verification"

            const { items } = await catalog.getEntities(
                {
                    filter: {
                        kind: 'system',
                        'relations.ownedBy': [entityRef.toLocaleLowerCase('en-US')],
                    },
                    fields: ['metadata.name'],
                    limit: 500,
                },
                { credentials: credentials }, # works when replacing `credentials` by `service_credentials`
            );
            return items.map(e => e.metadata.name);
        }
        return [];
    }
```
This fails at the `authenticate` step with Error: `Failed plugin token verification`

Funny enough, if I now remove the `authenticate` line, and simply pass my `service_credentials` as `credentials` for the request, it works. 

I really don't understand how and why, based on what I am reading here, this is just the credentials of my plugin, and should not allow me to communicate with the catalog plugin ?
https://backstage.io/docs/auth/service-to-service-auth/#standard-plugin-to-plugin-auth

In general, I feel it would be great to have a complete and direct example explaining how one would communicate with the catalog plugin from another backend plugin, this is probably a very common scenario?

PR #27246, opened 2024-10-18
[test-utils] support matches per breakpoint mock

camilaibs:

Adds the ability to mock a media query per breakpoint and to change the active breakpoint during a test. Usage example:
```ts
const { set } = mockBreakpoint({ 
  initialBreakpoint: 'md',
  queryBreakpointMap: {
    '(min-width:1500px)': 'xl',
    '(min-width:1000px)': 'lg',
    '(min-width:700px)': 'md',
    '(min-width:400px)': 'sm',
    '(min-width:0px)': 'xs'
  },
});
// assertions for when the active breakpoint is "md"
set('lg');
// assertions for when the active breakpoint is "lg"
```


packages/core-components/report-testUtils.api.md
```diff
@@ -3,8 +3,31 @@
 > Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
 
 ```ts
+// @public (undocumented)
+export type Breakpoint = 'xs' | 'sm' | 'md' | 'lg' | 'xl';
```
Rugvip, line 7: Strange, thus isn't exported by MUI but it is defined internally O_o

I'm thinking it might be best to inline these for now, doesn't seem useful to have in the external API

packages/core-components/src/testUtils.ts
```diff
@@ -25,19 +32,118 @@
  *  https://mui.com/material-ui/react-use-media-query/#testing
  *
  * @public
+ *
+ * @xample
```
Rugvip, line 36: `@example` 😁 , same below

packages/core-components/src/testUtils.ts
```diff
@@ -25,19 +32,118 @@
...
+        matches:
+          'matches' in options
+            ? options.matches
+            : breakpoints[mediaQueryString] === breakpoint,
+        media: mediaQueryString,
+        onchange: null,
+        addListener: jest.fn(listener => {
+          mediaQueryListeners.add(listener);
+        }),
+        // deprecated
```
Rugvip, line 111: don't really need these comments tbh, yep they might be but we're reimplementing the API and they're not going anywhere 😅. Think they're on the wrong ones too

packages/core-components/report-testUtils.api.md
```diff
@@ -3,8 +3,31 @@
...
-export function mockBreakpoint(options: { matches: boolean }): void;
+export function mockBreakpoint(
+  options:
+    | {
+        matches: boolean;
+      }
+    | {
+        initialBreakpoint?: Breakpoint;
+        queryBreakpointMap?: Record<string, Breakpoint>;
+      },
```
Rugvip, line 21: Might be worth experimenting with using an overload instead of the union. That lets use have separate documentation for the two for example. Whichever one works the best when looking at the function in the editor and exploring call options

Rugvip (commented):
Nice! 😁 🎉 

Just some small bits

packages/core-components/report-testUtils.api.md
camilaibs: They are exported from MUI, thanks for pointing that out

packages/core-components/src/testUtils.ts
camilaibs: 🤦🏻‍♀️ , fixed 😅.

packages/core-components/src/testUtils.ts
camilaibs: Removed

packages/core-components/report-testUtils.api.md
camilaibs: Done 🙌🏻

Rugvip (approved):
Nice! 🎉

PR #27249, opened 2024-10-18
Add TechDocs index page redirect to Overview page

jroebu14:

This prevents a jarring user experience when importing the `backstage` component and going to TechDocs as one evaluating Backstage might do.

### DIFF
--- a/plugins/techdocs/src/reader/transformers/handleMetaRedirects.tsx
+++ b/plugins/techdocs/src/reader/transformers/handleMetaRedirects.tsx
@@ -20,7 +20,10 @@ import React from 'react';
 import { renderReactElement } from './renderReactElement';
 import { TechDocsRedirectNotification } from '../components/TechDocsRedirectNotification';
 
-export const handleMetaRedirects = (entityName: string): Transformer => {
+export const handleMetaRedirects = (
+  navigate: (to: string) => void,
+  entityName: string,
+): Transformer => {
   const redirectAfterMs = 3000;
 
   const determineRedirectURL = (metaUrl: string) => {
@@ -52,6 +55,7 @@ export const handleMetaRedirects = (entityName: string): Transformer => {
         if (!metaContentParameters || metaContentParameters.length < 2) {
           return dom;
         }
+
         const metaUrl = metaContentParameters[1];
         const redirectURL = determineRedirectURL(metaUrl);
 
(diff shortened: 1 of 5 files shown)


Rugvip (commented):
Is it possible to make this a redirect instead? Thinking we may want to avoid having this be the content at https://backstage.io/docs

jroebu14:
@Rugvip sure! So installing this [plugin](https://github.com/mkdocs/mkdocs-redirects) is ok to achieve that?

jroebu14:
@Rugvip I discovered `mkdocs-redirects `is included in `mkdocs-techdocs-core` so have included the redirect.

Another issue has presented itself though. The redirect works but then a few seconds later there is another redirect to a page that doesn't exist.

https://github.com/user-attachments/assets/eff34f63-8167-4d13-968f-b532f22c094a

plugins/techdocs/src/reader/transformers/handleMetaRedirects.tsx
```diff
@@ -69,15 +68,17 @@ export const handleMetaRedirects = (
...
+            }}
             autoHideDuration={redirectAfterMs}
           />,
           container,
         );
         document.body.appendChild(container);
 
         setTimeout(() => {
-          navigate(redirectURL);
+          window.location.href = redirectURL;
```
jroebu14, line 81: @squid-ney this change was needed to prevent a double redirect happening (can be seen in the video above). Just checking this is OK with you since you added the redirect handling. Thanks!

squid-ney: Hi @jroebu14, I think it would be best to avoid a hard page refresh when redirecting so I'd prefer to use the `navigate`. 

This issue is occurring because the `handleMetaRedirects.tsx` function is getting called more than once on the same page with the redirect tag. It should only be called once for that page content. I took a look at why this bug is _only_ occurring when redirecting the `index.md' page and found an edge case. I believe this PR will solve your issue: https://github.com/backstage/backstage/pull/27505 let me know what you think

jroebu14: Hey @squid-ney really appreciate you investigating this and providing a fix! PR looks great. Thanks! I'll revert the changes to the redirect handling I've made in this PR.

freben (approved):

plugins/techdocs/src/reader/transformers/handleMetaRedirects.tsx
squid-ney: Sounds good, no problem! :)

PR #27252, opened 2024-10-18
catalog-client: fix case sensitive filters in mock api

Rugvip:
🧹, ran into this when trying to query relations with the mock client.


benjdlambert (approved):

freben (approved):
👍

Issue #20658, opened 2023-10-18
🚀 Feature: Support pnpm as alternative package manager to yarn


ourchitectureio:
### 🔖 Feature description

Offer support for the [existing "packageManager" configuration setting for Node projects](https://stackoverflow.com/a/72825528).

Please consider removing the explicit dependency on `yarn` in the "backstage-cli" and support alternative package managers like [pnpm](https://pnpm.io/).

### 🎤 Context

We have a monorepo built with a lot of `pnpm` automation, but the Backstage instance directories including all projects / workspaces require `yarn` and must be excluded with special treatment.

`pnpm` can offer some benefits to teams over `yarn` depending on the features they value most.

### ✌️ Possible Implementation

Read the preferred package manager tool from the "package.json" file and the [configuration setting "packageManager"](https://stackoverflow.com/a/72825528). Support this tool in all automation including "backstage-cli".

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

awanlin:
Some loosely related issues just for history:

- https://github.com/backstage/backstage/issues/11570
- https://github.com/backstage/backstage/issues/8054

There's a lot of customization around the `yarn.lock` file in the `backstage-cli` as you noted, the DevTools plugin also uses it for its Info tab feature

Rugvip:
Would love to add support for more package managers! This is a capacity/priority issue, happy to help guide any contributions in this area. We don't want the CLI to require you to use Yarn, in fact there are many parts of the CLI that should work without it as long as you're using package manager with the workspaces monorepo setup. Things like `version:bump` will for sure not be supported though so we'd need more work to get there.

dbrugger:
Very simple and scoped PR for npm support at https://github.com/backstage/backstage/pull/20670 but fully working to just build everything without yarn which was my main use case. Will take a stab at making it slightly more extensible as that seems the overall interest.

dbrugger:
I am happy to add pnpm to #20670 if someone can test it (see https://github.com/backstage/backstage/pull/20670#discussion_r1369750361)

ESmithaaS:
Please add support for pnpm to the backstage project and corresponding backstage-cli.

[...]

PR #27253, opened 2024-10-18
cli: conditionally include yarnPlugin query param in versions:bump upgrade-helper link

mtlewis:

Depends on https://github.com/backstage/upgrade-helper/pull/16. With this change we automatically link to the right flavor of the diff in upgrade helper based on whether the yarn plugin is installed.

### DIFF
--- a/.changeset/swift-laws-applaud.md
+++ b/.changeset/swift-laws-applaud.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+Update upgrade-helper link in versions:bump command to include yarnPlugin param when the yarn plugin is installed
--- a/.changeset/swift-laws-applaud.md
+++ b/.changeset/swift-laws-applaud.md
@@ -2,4 +2,4 @@
 '@backstage/cli': patch
 ---
 
-Update upgrade-helper link in versions:bump command to include yarnPlugin param when the yarn plugin is installed
+Update upgrade-helper link in `versions:bump` command to include `yarnPlugin` param when the yarn plugin is installed
--- a/packages/cli/src/commands/versions/bump.ts
+++ b/packages/cli/src/commands/versions/bump.ts
@@ -228,7 +228,10 @@ export default async (opts: OptionValues) => {
 
     // Do not update backstage.json when upgrade patterns are used.
     if (pattern === DEFAULT_PATTERN_GLOB) {
-      await bumpBackstageJsonVersion(releaseManifest.releaseVersion);
+      await bumpBackstageJsonVersion(
+        releaseManifest.releaseVersion,
+        hasYarnPlugin,
+      );
     } else {
       console.log(
         chalk.yellow(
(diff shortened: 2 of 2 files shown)


vinzscam (approved):

tylerd-canva:
Hi there! Just a bystander, but I'm working on https://github.com/backstage/backstage/issues/20658 and it may be a nitpick, but could we consider changing the term `yarnPlugin` to `backstageVersionProtocolSupport` or similar? I've got a branch where I've removed nearly all the explicit references to Yarn and it would be really nice to avoid adding any new ones :)

mtlewis:
Hmm - I've got some reservations about that, to be honest. I think generalizing this parameter bakes in the assumption that we could build a plugin for a different package manager that provides _exactly_ the same functionality. Diffs for the upgrade helper are generated by actually running the yarn plugin so that we can guarantee that the right content is shown for each pair of versions. If there were any differences between the yarn and (say) pnpm implementations, the diff wouldn't reflect them.

With that in mind, if we did have a plugin for pnpm (which by the way would be amazing!), I think I'd probably advocated for us considering something like this:
- Generate pnpm-plugin-based diffs for the upgrade-helper to account for any differences between plugins
- Introduce a new `pnpmPlugin` parameter to the upgrade helper
- Detect use of the pnpm plugin in the versions:bump command [as we do for yarn](https://github.com/backstage/backstage/blob/master/packages/cli/src/commands/versions/bump.ts#L472-L499)
- Factor in the pnpm plugin when deciding whether to use `backstage:^` versions (this also allows us to introduce any necessary variations in output based on which plugin is in use)
- Use the `pnpmPlugin` query param in the versions:bump upgrade helper link when the pnpmPlugin has been detected.

Another option might be to consider something like `?backstageProtocol=true&packageManager=yarn` in the upgrade-helper, but it seems like that would be fine to revisit later if we decide we prefer that to separate `yarnPlugin` & `pnpmPlugin` params.

Despite the wall of text, I definitely don't have a strongly-held viewpoint here - happy to discuss further. @vinzscam / @Rugvip, any thoughts on this?

tylerd-canva:
> Another option might be to consider something like ?backstageProtocol=true&packageManager=yarn in the upgrade-helper, but it seems like that would be fine to revisit later if we decide we prefer that to separate yarnPlugin & pnpmPlugin params.

I actually completely agree with this. It's pretty similar to something I mentioned in my PR:

> * create-app support for choosing your package manager at creation time. but if we have a dynamic template we might also need...
> * backstage-upgrade-helper support. pick your package manager in the UI so that you get an effective diff

Suffice to say, I would not have an objection to leaving this as-is for now.

.changeset/swift-laws-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+Update upgrade-helper link in versions:bump command to include yarnPlugin param when the yarn plugin is installed
```
Rugvip, line 5:
```suggestion
Update upgrade-helper link in `versions:bump` command to include `yarnPlugin` param when the yarn plugin is installed
```

mtlewis:
Alright, sounds good @tylerd-canva - appreciate the discussion. The closest thing I've found in pnpm are the [hooks](https://pnpm.io/pnpmfile#hooks) - it looks like you could probably use them for the resolution part, but restoring npm versions during packing might be difficult.

Rugvip (approved):
Thank you! 👍 

Regarding the yarn plugin discussion, I think the plugin is specific enough for now that we should handle that explicitly as it's being done here

PR #27255, opened 2024-10-18
fix: `useEventBus` by propagating config to `DefaultEventsService`

RedlineTriad:

This setting was added by reading from an optional `config` argument. This was never passed, so `auto` was always assumed. https://github.com/backstage/backstage/pull/27227/files#diff-cffd40a187b4fc584f03c968517fbacc110bde6a467f2384bbd92b803d3db47eR374-R379

I noticed this issue because when using the proxy, event service has serious issues, and even seems to break login somehow?
(I think it uses a local `172.*.*.*` IP instead of `localhost` or `127.0.0.1`, and it can't parse IP ranges in `GLOBAL_AGENT_NO_PROXY`)
But I will hunt that issue in a different PR.


freben:
Hi! Note that you'll need to run `yarn build:api-reports` when API surfaces change.

RedlineTriad:
> Hi! Note that you'll need to run `yarn build:api-reports` when API surfaces change.

I know, I forget it every time...

plugins/events-backend/report.api.md
```diff
@@ -18,7 +18,7 @@ import { LoggerService } from '@backstage/backend-plugin-api';
 // @public @deprecated
 export class DefaultEventBroker implements EventBroker {
   // @deprecated
-  constructor(logger: LoggerService, events?: EventsService);
+  constructor(logger: LoggerService, config?: Config, events?: EventsService);
```
Rugvip, line 21: Can't break the API like this. Let's leave the config out of this one and instead hardcode it to `'never'` for the old event broker implementation.

plugins/events-backend/report.api.md
```diff
@@ -29,7 +29,7 @@ export class DefaultEventBroker implements EventBroker {
 
 // @public @deprecated
 export class EventsBackend {
-  constructor(logger: Logger);
+  constructor(logger: Logger, config?: Config);
```
Rugvip, line 32: Same here, can skip this change

Rugvip (commented):
Thank you! 👍 

Just need some tweaks, although I'll get on those so we can ship this asap

Rugvip (approved):
Alright, thank you! Gonna :shipit:

PR #27266, opened 2024-10-19
feat: move config to its own "module"

aramissennyeydd:

Related to https://github.com/backstage/backstage/pull/27213. This adds a more complete entrypoint + plumbing for registration of individual commands.

### DIFF
--- a/packages/cli/src/alpha.ts
+++ b/packages/cli/src/alpha.ts
@@ -15,8 +15,14 @@
  */
 
 import { CliInitializer } from './wiring/CliInitializer';
+import chalk from 'chalk';
 
 (async () => {
+  console.warn(
+    chalk.yellow(
+      'THIS ENTRYPOINT IS IN ALPHA AND MAY CHANGE IN THE FUTURE - DO NOT USE THIS FOR NORMAL DEVELOPMENT',
+    ),
+  );
   const initializer = new CliInitializer();
   initializer.add(import('./modules/config/alpha').then(m => m.default));
   await initializer.run();
--- a/packages/cli/src/alpha.ts
+++ b/packages/cli/src/alpha.ts
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CliInitializer } from './wiring/CliInitializer';
+
+(async () => {
+  const initializer = new CliInitializer();
+  initializer.add(import('./modules/config/alpha').then(m => m.default));
+  await initializer.run();
+})();
--- a/packages/cli/src/wiring/CliInitializer.ts
+++ b/packages/cli/src/wiring/CliInitializer.ts
@@ -14,24 +14,14 @@
  * limitations under the License.
  */
 
-import yargs from 'yargs';
 import { CommandGraph } from './CommandGraph';
 import { CliFeature, InternalCliFeature, InternalCliPlugin } from './types';
 import { CommandRegistry } from './CommandRegistry';
+import { program } from 'commander';
+import { version } from '../lib/version';
+import chalk from 'chalk';
 
 type UninitializedFeature = CliFeature | Promise<CliFeature>;
-
-function checkCommands(
-  nestedYargs: yargs.Argv,
-  argv: Awaited<yargs.Argv['argv']>,
-  numRequired: number,
-) {
-  if (argv._.length < numRequired) {
-    nestedYargs.showHelp();
-  } else {
-    // check for unknown command
-  }
-}
 
 export class CliInitializer {
   private graph = new CommandGraph();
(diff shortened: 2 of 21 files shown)


packages/cli/src/wiring/CliInitializer.ts
```diff
@@ -0,0 +1,101 @@
...
+      _: commandName,
+      $0: binaryName,
+      ...options
+    } = await yargs(process.argv.slice(2)).parse();
+    const command = this.graph.find(commandName.map(String));
+    if (!command) {
+      console.error(chalk.red(`Command not found: "${commandName.join(' ')}"`));
+      const possibleCommands = this.graph.atDepth(commandName.length - 1);
+      if (possibleCommands.length > 0) {
+        console.log('Available commands:');
```
Rugvip, line 65: We should aim to make this a standard help page. I'd actually expect us to be able to build up the command graph using a CLI parser of our choosing, e.g. command or yargs. We'd just avoid using that library to communicate with plugins.

packages/cli/src/alpha.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CliInitializer } from './wiring/CliInitializer';
+
```
Rugvip, line 18: Let's add a very clear `console.log` that this is an experimental entry point for internal development of the CLI that should not be used, just to avoid any kind of confusion about what this entry point is for. As opposed to other experimental entry points I'd also like us to avoid shipping any new features here so that we don't tempt anyone to use this. As soon as we're able to mirror the existing CLI using the new structure we should get rid of this entry point too.

packages/cli/src/wiring/CliInitializer.ts
```diff
@@ -0,0 +1,101 @@
...
+            text += ' [command]';
+          } else {
+            text += ` [options]\t\t${node.command.description}`;
+          }
+          console.log(text);
+        }
+      }
+      process.exit(1);
+    }
+    const parsedOptions = command.schema.parse(options);
```
Rugvip, line 78: I don't think this is gonna work very well in practice tbh, you kinda need to know what the option structure is when you're parsing the args.

For example, parsing these directly with `yargs` gives you different results:

```
yargs().parse('--fix --config app-config.yaml file.txt')
yargs().parse('--config app-config.yaml --fix file.txt')
```

But if you define `fix` as a flag upfront, you get what you'd arguably want:

```
yargs().option('fix', { type: 'boolean' }).parse('--config app-config.yaml --fix file.txt')
```

I'm hoping we'll be able to lean on existing libraries to handle the arg parsing within plugins, that we don't need to know about or handle that at the framework level at all. So to be clear, I think we should pass an array of all raw args that appear after the command itself in input argv to `command.execute`.

Rugvip (commented):
Nice! 👍 I think it makes sense to move this forward. You could mark this one as ready and close #27213 tbh.

packages/cli/src/wiring/CliInitializer.ts
aramissennyeydd: Ah, makes sense - does the current `zod` interface make sense as the way to pass args? Only thing missing would be positional args.

packages/cli/src/wiring/CliInitializer.ts
aramissennyeydd: Do you have an interface in mind when you say array of raw args? Thinking something like `{optionName: string; type:'boolean' | 'string' | 'number', variadic: boolean }`

Rugvip: Plain and simple `string[]`:

```ts
reg.addCommand({
  path: ['config:check'],
  description:
    'Validate that the given configuration loads and matches schema',
  execute: async ({ args }) => {
    const options = yargs()
      .option('lax', { type: 'boolean' })
      /* ... */
      .parse(args);
    const m = await import('./commands/validate');
    await m.default(options);
  },
});
```

aramissennyeydd: Ohhhh, gotcha, summarizing for my understanding - each command gets to use whichever arg parser they want, on the framework end we would also use an arg parser (say `yargs`), but they'd be separate instances. Was a little hungup on how to handle help pages, but I guess it wouldn't matter since they'd just pass the `--help` to the command which would parse it and show its own help. Tree-level help would still be provided by the framework.

Rugvip: Yep, spot on! 👍

aramissennyeydd:
@Rugvip Pushing up what I have, spent a bit of time trying to wrestle `yargs` into something we could use, I don't love it rn - curious to hear your thoughts on my examples. 

Biggest issue is that subcommands aren't completely supported and need to be run as nested functions during parse-time, https://github.com/yargs/yargs/issues/225 https://github.com/yargs/yargs/issues/1024. That pattern doesn't lead well to us being able to recursively nest commands.

Rugvip:
@aramissennyeydd hmm, alright. Could potentially be that it is simpler to stick to `commander` at the root level and `yargs` for parsing options in commands? Dunno if the current code is all that bad though, I doubt it's gonna be that much simpler with commander either 😅. We're also free to experiment with this in the long run so I wouldn't mind settling on something that works for now and consider refactor later, as long as we can keep the commands API intact.

aramissennyeydd:
@Rugvip Bump here - this is in a good place now with the switch to commander

Rugvip (approved):
Yep I think this is looking clean 👌

Let's get this rollin' 😁 

Curious on your thoughts on the best way forward. Two things to explore that come to mind are either some form of shared context CLI plugins to rely on, or to explore ways to make package/repo commands pluggable?

aramissennyeydd:
@Rugvip Shared context is enticing, but I don't think it's valuable yet as we're not sharing code across package boundaries. The pluggability is the next thing I think we have to explore, both how to define the modules for a given CLI and a manifest of dependencies for those modules.

PR #27273, opened 2024-10-21
scaffolder: Rework state handling

benjdlambert:
TLDR: Some of the `liveOmit` and `omitExtraData` things are broken, so this rolls back some of that functionality from being the default.

This reverts some of the merging of the step state from before into just having one model for now, and we will revisit this step state collection later.
Hoping that this should fix a number of scaffolder issues that's popped up since last release.


plugins/scaffolder-react/src/next/components/Stepper/Stepper.tsx
```diff
@@ -163,54 +163,63 @@ export const Stepper = (stepperProps: StepperProps) => {
...
-    if (hasErrors(returnedValidation)) {
-      setErrors(returnedValidation);
-    } else {
+  const handleChange = useCallback(
+    (e: IChangeEvent) => {
       setStepsState(current => {
         const newState = [...current];
         newState[activeStep] = {
-          ...formData,
+          ...e.formData,
```
vinzscam, line 175: I think if you do something like:

```ts
setStepsState(current => {
  if(isEqual(e.formData, current[activeStep]) {
      return current;
  }
  const newState = ...
  return newState;
});
```

it should allow you to remove the JSON.stringify while creating `formData`:

```ts
const formData = useMemo(
    () => stepsState[activeStep],
    [stepsState, activeStep],
)
```

stephenglass:
just wondering if this is related to any of the excessive re-rerendering perf issues observed in #26781 ?

benjdlambert:
@stephenglass it's possible it helps, but this was originally to fix an issue which was caused by #27230 which we noticed internally. I haven't tested the secret field with the latest changes though.

Rugvip (approved):
aight

PR #27277, opened 2024-10-21
fix search collator filter

### LINKED ISSUE #27265: 🐛 Bug Report: `DefaultCatalogCollator` doesn't support list as entity filter
### 📜 Description

`DefaultCatalogCollator`'s filter can be configured using `search.collators.catalog.filter` config key. But, although the filter type is `EntityFilterQuery` and it can be a list of records, when you pass a list to it, it raises the following error:
```
ForwardedError [TypeError]: Module 'catalog-collator' for plugin 'search' startup failed; caused by TypeError: Invalid type in config for key 'search.collators.catalog.filter' in 'app-config.yaml', got array, wanted object
```

### 👍 Expected behavior

It should accept a list as an input. A possible value would be like this:
```yaml
search:
  collators:
    catalog:
      filter:
        - kind: [System, API, Resource, Location, User, Group, Domain, Template]
          spec:
            lifecycle: experimental
        - kind: [Component]
           spec:
             type: service
```

The following error will occur when passing a list as filter:
```
ForwardedError [TypeError]: Module 'catalog-collator' for plugin 'search' startup failed; caused by TypeError: Invalid type in config for key 'search.collators.catalog.filter' in 'app-config.yaml', got array, wanted object
```

### 👟 Reproduction steps

[ ] Enable search plugin
[ ] Pass a list as a filter in the `search.collators.catalog.filter` config field

_No response_

### 🖥️ Your Environment

_No response_

freben:
Fixes #27265

This should be allowed to both be an object and an array.

### DIFF
--- a/.changeset/forty-pets-wave.md
+++ b/.changeset/forty-pets-wave.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-search-backend-module-catalog': patch
+---
+
+Fixed a bug where the `filter` setting of the collator was not permitted to be an array.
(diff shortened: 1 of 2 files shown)


Rugvip (approved):
👍

PR #27287, opened 2024-10-21
remove canvas dev dep

Rugvip:
🧹, checking whether we still need `canvas`


freben (approved):
If it works, then ... good riddance.

PR #27299, opened 2024-10-22
cli: new solution for local workspace linking

Rugvip:

This replaces our existing local development solution for linking in local workspaces. Our existing solution used to possible to make work while we were still on Yarn v1, but it's not more or less impossible to get working properly.

This new solution only does the linking at runtime during local development. You use the `--link <workspace-path>` option for the package `start` command to select an external workspace for linking, and then overrides module resolution to prefer packages in that workspace. The only exception are `react` and `react-dom` packages, which will always be resolved from the target package, which I think is a sensible baseline but we might need to make configurable.

The option works both with Webpack and Rspack, but it is frontend-only for now


benjdlambert (approved):

Rugvip:
I took a stab at implementing this for backend packages too, landed at https://github.com/backstage/backstage/pull/27299/commits/d2e166cbb207d633474931fba34bc5e726524064

packages/cli/src/lib/runner/runBackend.ts
```diff
@@ -119,6 +121,7 @@ export async function runBackend(options: RunBackendOptions) {
         stdio: ['ignore', 'inherit', 'inherit', 'ipc'],
         env: {
           ...process.env,
+          LINKED_WORKSPACE: options.linkedWorkspace,
```
freben, line 124:
```suggestion
          BACKSTAGE_CLI_LINKED_WORKSPACE: options.linkedWorkspace,
```

perhaps?

freben (approved):
Sweet! This makes a lot of sense

packages/cli/src/lib/runner/runBackend.ts
Rugvip: Yep! updated 👍

Issue #16389, opened 2023-02-16
🚀 Feature: Add Environment Concept to the Backstage Entity Model


raffaelespazzoli:
### 🔖 Feature description

I'd like to open a discussion on whether the Backstage entity model should have the concept of Environments.

Here by Environment I mean the place where Components are deployed. Environments are definitely part of the Developers' parlance and their point of view.

In general I have identified three types of environments:
1. static SDLC environments: these are environment that exists statically and that the application has to go through to get to production. Their life-cycle is tied to the application life-cycle.
2. ephemeral environments: these environments exist for a much more limited span of time. They could be used for example to deploy feature branches or hosting resource intense performance tests.
3. personal environments: these are environments dedicated to a single developer.

I'd like to focus initially on the static SLDC environments. It would be nice in my opinion to be able to model that a component is associated with a set of static environments. 

Here are some use cases that can be enabled by this feature:
1. provisioning of these environments
2. possibly auto-configuration of a pipeline.
3. ability for the developer to trigger a deployment to a specific environment.

In terms of modeling the Environment entity would simple have enough information to identify the actual runtime locations a given environment is comprised of. This could be kubernetes cluster/namespaces, AWS VPCs, GCP projects and so on. Notice an environment can be comprised of a combination of the above.

I don't think that Backstage should keep track of what is an environment, nor should it try to orchestrate the environment in any way. This will be left to potential plugins or external systems. This way the environment entity should be relatively static as the rest of the backstage entities.

### 🎤 Context

please see above.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

freben:
Thanks for writing this up.

This could definitely make sense (whether in the core model or as an extension). Possibly in the form of a `kind: Environment` and some set of entity relations to point between them and e.g. components - does that sound close to what you'd say fulfils the needs here?

As a matter of fact, what would the ideal granularity be? Should individual components and resources point to their environments, or would it make sense to think about this at a more overarching level such that e.g. a System or an [Application](https://github.com/backstage/backstage/issues/4873) relates to the environment instead?

[...]

PR #27306, opened 2024-10-22
Change the word "production" to "maintained" as a `spec.lifecycle` value

patricekrakow:

Using the word "production" is quite confusing, because it sounds like a (production) environment, while @freben made it clear in https://github.com/backstage/backstage/issues/16389#issuecomment-2045383992 that "a component is an abstract concept that is often represented by the source repo. No versions, no environments."

### DIFF
--- a/docs/features/software-catalog/descriptor-format.md
+++ b/docs/features/software-catalog/descriptor-format.md
@@ -58,7 +58,7 @@ metadata:
       type: admin-dashboard
 spec:
   type: website
-  lifecycle: production
+  lifecycle: maintained
   owner: artist-relations-team
   system: public-websites
 ```
@@ -93,7 +93,7 @@ This is the same entity as returned in JSON from the software catalog API:
     "uid": "2152f463-549d-4d8d-a94d-ce2b7676c6e2"
   },
   "spec": {
-    "lifecycle": "production",
+    "lifecycle": "maintained",
     "owner": "artist-relations-team",
     "type": "website",
     "system": "public-websites"
@@ -139,7 +139,7 @@ metadata:
   description: The Petstore API
 spec:
   type: openapi
-  lifecycle: production
+  lifecycle: maintained
   owner: petstore@example.com
   definition:
     $text: https://petstore.swagger.io/v2/swagger.json
@@ -522,7 +522,7 @@ metadata:
   description: The place to be, for great artists
 spec:
   type: website
-  lifecycle: production
+  lifecycle: maintained
   owner: artist-relations-team
   system: artist-engagement-portal
   dependsOn:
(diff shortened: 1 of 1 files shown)


hudsonb:
`experimental`, `production`, and `deprecated` are well-known and common lifecycles so I don't think they should be changed.

Parsifal-M:
Hmm the terminology used [here](https://backstage.io/docs/features/software-catalog/descriptor-format#kind-component)

Describes a component as:

A Component describes a software component. It is typically intimately linked to the source code that constitutes the component, and should be what a developer may regard a "unit of software", usually with a distinct **deployable** or **linkable artifact**.

The way I interpret that is that it would be something you deploy to an environment (like production).

freben:
Yeah I'll disagree here too.

- The original intent wasn't necessarily meant as an environment identifier but rather "it's in production" in the sense of "actively used and functional and maintained and probably relied upon"
- This is just a free-form string - any org is free to choose (and enforce) any terminology without the default model changing
- This base recommendation has been established for several years now and probably isn't worth changing unless there's particular reason to, which I don't think there is.

Thanks for your input, but I'll close this. One of the strengths of the catalog is that it's so flexible. The default model is intentionally on the small and generic side, and adopters are actively encouraged to extend and tweak it if needed. So I think this PR isn't needed.

patricekrakow:
@hudsonb and @freben let me have one more/last try ;-) When I see the set of values ( `experimental`, `production`, `deprecated` ), I still have the impression that `production` is an "intruder".

@hudsonb you said you have a different experience, maybe you can find back some references where the value `production` is used next to `experimental` and `deprecated`.

On my side, I have made some (anecdotal) research and found the following sets:

- `experimental`, **`standard_track`**, `deprecated` in <https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#status-information>

- `Planned`, `Experimental`, **`Regular`**, `Deprecated`, `Obsolete` in <https://docs.evolveum.com/midpoint/versioning/feature-lifecycle/>

- `Experimental`, **`General Availability (Production)`**, `Deprecated` in https://developers.miro.com/docs/lifecycle-policy

> I am being honest and also copied this one - I have been tempted to forget about it ;-) - but you can see that `Production` is not on its own.

- `experimental`, **`stable`**, `deprecated`, `removed` in <https://developers.arcgis.com/arcgis-urban-api/guides/features-lifecycle/>

- etc.

On the other hand, I do agree that if you refer to the full documentation:

```markdown
`production` - an established, owned, maintained component
```

and 

```markdown
`production` - an established, owned, maintained API
```

the reference to a lifecycle state and **not** an environment is clear. However, for a (too) quick reader, I think that using another term - `maintained`, `established`, `regular`, `standard_track`,  `general_availability`, `stable`, etc. - might be beneficial.

PR #27307, opened 2024-10-22
Gitlab discovery rate limit

mbenson:

This PR makes, without disruptive changes to e.g. specify retry strategies/limits, an attempt to take into account GitLab's published rate limiting strategy, by checking for the `429` HTTP status and waiting for the duration recommended by the accompanying `Retry-After` response header.

Related links:
* https://docs.gitlab.com/ee/user/gitlab_com/index.html#gitlabcom-specific-rate-limits
* https://docs.gitlab.com/ee/administration/settings/user_and_ip_rate_limits.html#response-headers

### DIFF
--- a/plugins/catalog-backend-module-gitlab/src/lib/client.ts
+++ b/plugins/catalog-backend-module-gitlab/src/lib/client.ts
@@ -48,6 +48,9 @@ interface UserListOptions extends CommonListOptions {
   exclude_internal?: boolean | undefined;
 }
 
+const tooManyRequests = 429;
+export const maxRetry = 5;
+
 export class GitLabClient {
   private readonly config: GitLabIntegrationConfig;
   private readonly logger: LoggerService;
@@ -421,21 +424,39 @@ export class GitLabClient {
       }
     }
 
-    this.logger.debug(`Fetching: ${request.toString()}`);
-    const response = await fetch(
-      request.toString(),
-      getGitLabRequestOptions(this.config),
-    );
-
-    if (!response.ok) {
-      throw new Error(
-        `Unexpected response when fetching ${request.toString()}. Expected 200 but got ${
-          response.status
-        } - ${response.statusText}`,
+    for (let i = 0; ; i++) {
+      this.logger.debug(`Fetching: ${request.toString()}`);
+      const response = await fetch(
+        request.toString(),
+        getGitLabRequestOptions(this.config),
       );
-    }
 
-    return Promise.resolve(response);
+      if (response.status === tooManyRequests && i < maxRetry) {
+        const retryAfter = response.headers.get('Retry-After');
+
+        if (retryAfter) {
+          const delay = isNaN(Number(retryAfter))
+            ? new Date(retryAfter).getTime() - Date.now() // Retry-After is a date
+            : parseInt(retryAfter, 10) * 1000; // Retry-After is in seconds
+
+          this.logger.debug(
+            `Rate limited. Retrying after ${delay / 1000} seconds...`,
+          );
+          await new Promise(resolve => setTimeout(resolve, delay));
+
+          continue;
+        }
+      }
+      if (!response.ok) {
+        throw new Error(
+          `Unexpected response when fetching ${request.toString()}. Expected 200 but got ${
+            response.status
+          } - ${response.statusText}`,
+        );
+      }
+
+      return Promise.resolve(response);
+    }
   }
 }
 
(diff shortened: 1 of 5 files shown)


mbenson:
I thought I had seen an older issue for this. I can no longer find it, but I see that GitHub and BitBucket have also provided a number of users problems with rate limiting. I wonder if this signifies a need for some kind of pluggable retry service; at the same time I don't think the acceptance of this patch would preclude later adoption of such a service.

benjdlambert:
#24860 and #26063 are probably the closest related here

plugins/catalog-backend-module-gitlab/src/lib/client.ts
```diff
@@ -46,6 +46,9 @@ interface UserListOptions extends CommonListOptions {
   exclude_internal?: boolean | undefined;
 }
 
+const tooManyRequests = 429;
+export const maxRetry = 5;
```
camilaibs, line 50: Is it a good idea to make this value customizable via app-config.yaml?

plugins/catalog-backend-module-gitlab/src/lib/client.ts
```diff
@@ -428,20 +413,39 @@ export class GitLabClient {
...
+        request.toString(),
+        getGitLabRequestOptions(this.config),
       );
-    }
 
-    return response.json();
+      if (response.status === tooManyRequests && i < maxRetry) {
+        const retryAfter = response.headers.get('Retry-After');
+
+        if (retryAfter) {
```
camilaibs, line 426: Is it only going to retry after value is returned in the response, just to be clear?

camilaibs (approved):
Looks great to me, left just minor comments 🙂

plugins/catalog-backend-module-gitlab/src/lib/client.ts
mbenson: @camilaibs My thought process was:
* `429` status received (`const tooManyRequests`)
* Look for `Retry-After` response header
* If said header's value is truthy, sleep for the specified duration

So essentially if there is no `Retry-After` header the retry will occur immediately. We could replace this with some default sleep duration. Does that answer your question?

plugins/catalog-backend-module-gitlab/src/lib/client.ts
```diff
@@ -428,20 +413,39 @@ export class GitLabClient {
...
+          const delay = isNaN(Number(retryAfter))
```
vinzscam, line 427: mmm I guess some tests are missing here, as `delay` is currently a boolean

vinzscam (changes requested):

plugins/catalog-backend-module-gitlab/src/lib/client.ts
mbenson: actually, it's either a Date or a number; the `boolean` was just the intro to a ternary expression 😄

vinzscam: sorry I'm stupid, looks good 👍

plugins/catalog-backend-module-gitlab/src/lib/client.ts
```diff
@@ -428,20 +413,39 @@ export class GitLabClient {
...
+            ? new Date(retryAfter).getTime() - Date.now() // Retry-After is a date
```
benjdlambert, line 428: Should we be having a `maxDelay` or something here, it feels a little unfortunate for consumers of this API to not fail straight away and potentially block anything more than a few seconds to be honest. 

Take the catalog for instance, if you have 1000 components and all those reads are delayed 15minutes, you're going to lock up the catalog processing waiting for these reads to resolve, rather than failing straight away and it being re-processed at a later time.

benjdlambert (changes requested):
Little concern that I think we want to address here

plugins/catalog-backend-module-gitlab/src/lib/client.ts
mbenson: I'll update to satisfy the bot and maybe we can discuss at an upcoming SIG.

mbenson:
Since #27850 was merged, it seems that some form of this should eventually be accepted. Shall we remove `stale`? What is the best way to move forward?

webark:
The other PR caused issues for us due to artifically throttling us in ways that was unexpected and unwanted. :/ I appreciate the sentiment of it, but I would much perfer a mechanism like this which will handle these concerns is a much more flexible way.

PR #27308, opened 2024-10-22
cli: fix tree sha reading for repo test

Rugvip:

Turns out `git ls-tree` doesn't always return the objects in the same order as the input, so this is a lil bit more resilient x)

### DIFF
--- a/packages/cli/src/commands/repo/test.ts
+++ b/packages/cli/src/commands/repo/test.ts
@@ -54,26 +54,36 @@ interface GlobalWithCache extends Global {
  * Use git to get the HEAD tree hashes of each package in the project.
  */
 async function readPackageTreeHashes(graph: PackageGraph) {
-  const pkgs = Array.from(graph.values());
+  const pkgs = Array.from(graph.values()).map(pkg => ({
+    ...pkg,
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/repo/test.ts
```diff
@@ -54,26 +54,35 @@ interface GlobalWithCache extends Global {
...
-  const pkgs = Array.from(graph.values());
+  const pkgs = Array.from(graph.values()).map(pkg => ({
+    ...pkg,
+    path: relativePath(paths.targetRoot, pkg.dir),
+  }));
   const output = await runPlain(
     'git',
     'ls-tree',
-    '--object-only',
+    '--format="%(path)=%(objectname)"',
```
freben, line 64: Any file/dir with an equals sign breaks this right? Can we put the known-good thing first (the hash) and then split smartly on just the FIRST match of the separator?

freben (approved):
Approve but suggesting a tweak

packages/cli/src/commands/repo/test.ts
Rugvip: Yep, fix'd

PR #27309, opened 2024-10-22
fix(backend-dynamic-feature-service): add `resolvePackagePath` support in the `CommonJSModuleLoader`.

davidfestal:

The `resolvePackagePath` function of the `backend-plugin-api` Core backstage package, tries to resolve the `package.json` of a package by package name, to find out the local folder of the plugin.
This works out-of-the-box for static plugins which are in the main monorepo structure of the application.
But for the dynamic backend plugins which are loaded for an external place (as sub-folders of the `dynamic-plugins-root`), the default `require.resolve` algorithm isn't sufficient.

This PR overrides the CommonJS `resolveFileName` logic, in order to also search such a `package.json` file in the list of loaded backend dynamic plugins.

It also allows customizing this search, in order to support cases when the name of the dynamic plugin (after some repackaging for example) doesn't match exactly the name expected by the code that called `resolvePackagePath`.

Fixing this backend dynamic plugin limitation related to `resolvePackagePath` is important for backend dynamic plugins which use the database, since database migration scripts systematically use `resolvePackagePath`.

_**NOTE:** this is indeed a contribution to upstream, in a more generic shape, of [a fix](https://github.com/janus-idp/backstage-showcase/commit/3522f3d81e70683c87bf2c93f5a4b4b42171608f) which is in place in the [`janus-idp/backstage-showcase`](https://github.com/janus-idp/backstage-showcase) (Red hat Developer Hub upstream project) since 7 months._

### DIFF
--- a/packages/backend-dynamic-feature-service/report.api.md
+++ b/packages/backend-dynamic-feature-service/report.api.md
@@ -69,6 +69,32 @@ export interface BaseDynamicPlugin {
 }
 
 // @public (undocumented)
+export class CommonJSModuleLoader implements ModuleLoader {
+  constructor(options: CommonJSModuleLoaderOptions);
+  // (undocumented)
+  bootstrap(
+    backstageRoot: string,
+    dynamicPluginsPaths: string[],
+    scannedPluginManifests: Map<string, ScannedPluginManifest>,
+  ): Promise<void>;
+  // (undocumented)
+  load(packagePath: string): Promise<any>;
+  // (undocumented)
+  readonly options: CommonJSModuleLoaderOptions;
+}
+
+// @public (undocumented)
+export type CommonJSModuleLoaderOptions = {
+  logger: LoggerService;
+  dynamicPluginPackageNameSuffixes?: String[];
+  customResolveDynamicPackage?: (
+    logger: LoggerService,
+    searchedPackageName: string,
+    scannedPluginManifests: Map<string, ScannedPluginManifest>,
+  ) => string | undefined;
+};
+
+// @public (undocumented)
 export type DynamicPlugin = FrontendDynamicPlugin | BackendDynamicPlugin;
 
 // @public (undocumented)
@@ -264,7 +290,11 @@ export type LegacyPluginEnvironment = {
 // @public (undocumented)
 export interface ModuleLoader {
   // (undocumented)
-  bootstrap(backstageRoot: string, dynamicPluginPaths: string[]): Promise<void>;
+  bootstrap(
+    backstageRoot: string,
+    dynamicPluginPaths: string[],
+    scannedPluginManifests?: Map<string, ScannedPluginManifest>,
+  ): Promise<void>;
   // (undocumented)
   load(id: string): Promise<any>;
 }
--- a/packages/backend-dynamic-feature-service/src/loader/CommonJSModuleLoader.ts
+++ b/packages/backend-dynamic-feature-service/src/loader/CommonJSModuleLoader.ts
@@ -16,17 +16,35 @@
 import { ModuleLoader } from './types';
 import { LoggerService } from '@backstage/backend-plugin-api';
 import path from 'path';
+import { ScannedPluginManifest } from '../scanner';
 
+/**
+ * @public
+ */
+export type CommonJSModuleLoaderOptions = {
+  logger: LoggerService;
+  dynamicPluginPackageNameSuffixes?: String[];
+  customResolveDynamicPackage?: (
+    logger: LoggerService,
+    searchedPackageName: string,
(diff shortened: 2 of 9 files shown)


packages/backend-dynamic-feature-service/src/loader/CommonJSModuleLoader.ts
```diff
@@ -44,11 +62,80 @@ export class CommonJSModuleLoader implements ModuleLoader {
...
       });
-      this.logger.debug(
+      this.options.logger.debug(
         `Overriding node_modules search path for dynamic plugin ${from} to: ${filtered}`,
       );
       return filtered;
     };
+
+    const oldResolveFileName = this.module._resolveFilename;
+    this.module._resolveFilename = (
```
Rugvip, line 72: Probably best if we could avoid this tbh. I wouldn't necessarily be against modifying the behavior of `resolvePackagePath` tbh, depending on what that'd look like. Especially if we can make it work with a future bundled layout for backend packages where they might all be mixed under a single directory

davidfestal: I agree that having the `resolvePackagePath` be more flexible would be great.

However, without this fix, the `CommonJSModuleLoader` default implementation of the dynamic backend plugin `ModuleLoader` is nearly useless, since it just doesn't support database-enabled dynamic plugins.
That's why we've been using a copied & fixed version of the CommonJSModuleLoader since 7 months on the Janus side (see the note above).

And it seems to me that, even when/if the `resolvePackagePath` utility is enhanced, this fix would still be necessary to support any dynamic backend plugin using the old version of `resolvePackagePath`.

davidfestal: As mentioned in the framework SIG call, I added a comment just above the `_resolveFileName` overriding, in order to provide context, and the fact that it might not be necessary anymore, when `resolvePackagePath` is made more flexible (or an equivalent, more flexible utility is provided).

schultzp2020:
+1 for the changes

/lgtm
/approve

Rugvip (commented):
Alright so as discussed in the SIG, let's have some comments with the context for this code close to the implementation to make it clear that this is something we're aiming to remove. Other than that I'm wondering if we can simplify the external API now that we've fixed this gap in the module loading?

packages/backend-dynamic-feature-service/report.api.md
```diff
@@ -68,6 +68,32 @@ export interface BaseDynamicPlugin {
   version: string;
 }
 
+// @public (undocumented)
+export class CommonJSModuleLoader implements ModuleLoader {
```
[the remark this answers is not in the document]
davidfestal: I agree that, having been fixed for the default use-case, the `CommonJSModuleLoader` could still be considered the default implementation of the `ModuleLoader`, not exported, and just used when no module loader is explicitly provided.

However, this would force rewriting a complete `CommonJSModuleLoader` (most probably duplicating code) as soon as dynamic plugin package resolution rules would be slightly different (what happened for us on RHDH).

This is why, through  the `CommonJSModuleLoaderOptions` (`dynamicPluginPackageNameSuffixes`, and `customResolveDynamicPackage`), as mentioned in the PR description, 
> it also allows customizing this search, in order to support cases when the name of the dynamic plugin (after some repackaging for example) doesn't match exactly the name expected by the code that called `resolvePackagePath`.

By design the dynamic plugins manager **_do not enforce or expect any specific pre-requisite about how the dynamic plugin should be packaged_** (apart from being a Javascript package with a `package.json`). Thus it is not excluded that, in some use-case, dynamic plugin packages would be repackaged with a derived package name (typically a suffixed name), or even embedded as a dependency inside a wrapper package for some specific needs.
Indeed the feedback of RHDH production usage proved that both are useful according to the use-case.

So here's the purpose of exporting `CommonJSModuleLoader`:
- still have it as a default value when no module loader is provided
- but also provide a basis implementation that can be configured to implement custom package resolution rules, according to the dynamic plugin packaging approaches that should be supported.

davidfestal:
> Alright so as discussed in the SIG, let's have some comments with the context for this code close to the implementation to make it clear that this is something we're aiming to remove.

I added this comment: https://github.com/backstage/backstage/pull/27309/files#diff-536979b34bfce39b3660e00e3a933de7fdedc29f7e1540c9336d2fdd51248810R71

> Other than that I'm wondering if we can simplify the external API now that we've fixed this gap in the module loading?

See my comment [above](https://github.com/backstage/backstage/pull/27309#discussion_r1824241282)

Rugvip (approved):
Alright, let's :shipit: 👍 

I'll open a new issue to discuss the way forward for `resolvePackagePath`

PR #27317, opened 2024-10-23
Patch release of #27291, #27273, #27255, #27308

benjdlambert:
This release fixes the following issues:

- Scaffolder form state not being tracked correctly, this reverts some of the `liveOmit` and `omitExtraData` functionality that was recently release. We're gonna look at shipping this a better way soon 🤞 
- Fix `events.useEventBus` by propagating config to `DefaultEventsService`
- Small fix for `sha` tree treading for `repo test` command


Rugvip (commented):
Quite a lot of unrelated patch bumps, not sure where they're from but I think it's best if we trim those out

PR #27326, opened 2024-10-24
feat(events): allow setting a timeout for event bus polling

kuangp:

Added a configurable option to timeout event bus polling requests instead of having it hang indefinitely by default. We have a scenario where our instance is deployed behind network routers that timeout API requests after 30s but this does not seem to properly close the hanging polling request (I suspect due to headers already being flushed) - this results in our subscriptions being orphaned since they continue to hang indefinitely and never attempt to fetch new events. This change should address this scenario by having a hard timeout to retrigger the poll - this should also be useful in cases where the event plugin backend may be terminated ungracefully preventing the request stream from closing properly as well. 
 
### DIFF
--- a/plugins/events-node/src/api/DefaultEventsService.ts
+++ b/plugins/events-node/src/api/DefaultEventsService.ts
@@ -22,7 +22,11 @@ import {
   RootConfigService,
 } from '@backstage/backend-plugin-api';
 import { EventParams } from './EventParams';
-import { EventsService, EventsServiceSubscribeOptions } from './EventsService';
+import {
+  EVENTS_NOTIFY_TIMEOUT_HEADER,
+  EventsService,
+  EventsServiceSubscribeOptions,
+} from './EventsService';
 import { DefaultApiClient } from '../generated';
 import { ResponseError } from '@backstage/errors';
 
@@ -214,10 +218,28 @@ class PluginEventsService implements EventsService {
             // immediately again
 
             lock.release();
-            // We don't actually expect any response body here, but waiting for
-            // an empty body to be returned has been more reliable that waiting
-            // for the response body stream to close.
-            await res.text();
+
+            const notifyTimeoutHeader = res.headers.get(
+              EVENTS_NOTIFY_TIMEOUT_HEADER,
+            );
+
+            // Add 1s to the timeout to allow the server to potentially timeout first
+            const notifyTimeoutMs =
+              notifyTimeoutHeader && !isNaN(parseInt(notifyTimeoutHeader, 10))
+                ? Number(notifyTimeoutHeader) + 1_000
+                : null;
+
+            await Promise.race(
+              [
+                // We don't actually expect any response body here, but waiting for
+                // an empty body to be returned has been more reliable that waiting
+                // for the response body stream to close.
+                res.text(),
+                notifyTimeoutMs
+                  ? new Promise(resolve => setTimeout(resolve, notifyTimeoutMs))
+                  : null,
+              ].filter(Boolean),
+            );
           } else if (res.status === 200) {
             const data = await res.json();
             if (data) {
--- a/plugins/events-node/src/api/EventsService.ts
+++ b/plugins/events-node/src/api/EventsService.ts
@@ -55,3 +55,8 @@ export type EventsServiceSubscribeOptions = {
  * @public
  */
 export type EventsServiceEventHandler = (params: EventParams) => Promise<void>;
+
+/**
+ * @public
+ */
+export const EVENTS_NOTIFY_TIMEOUT_HEADER = 'backstage-events-notify-timeout';
--- a/.changeset/old-taxis-exist.md
+++ b/.changeset/old-taxis-exist.md
@@ -1,0 +1,11 @@
+---
+'@backstage/plugin-events-backend': patch
+'@backstage/plugin-events-node': patch
+---
+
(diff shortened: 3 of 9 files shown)


camilaibs (approved):
Thanks for opening the pull request 🎉 
Left two minor comments, please take a look 🙂

plugins/events-node/src/api/DefaultEventsService.ts
```diff
@@ -227,7 +228,14 @@ class PluginEventsService implements EventsService {
           // immediately again
           if (res.status === 202) {
             lock.release();
-            await res.body?.getReader()?.closed;
+            await Promise.race([
+              res.body?.getReader()?.closed,
```
Rugvip, line 232: If this is hanging we'll likely want to do more than ignore it. Probably best to close the previous request if we hit the timeout.

plugins/events-node/config.d.ts
```diff
@@ -30,5 +30,11 @@ export interface Config {
      * will never be disabled, even if the events backend returns a 404.
      */
     useEventBus?: 'never' | 'always' | 'auto';
+    /**
+     * Timeout in milliseconds for how long to wait for the hanging polling request
+     * on the event bus until triggering a new request. By default, the request will
+     * hang indefinitely until the bus ends the response to signal a new event.
+     */
+    eventBusPollingTimeoutMs?: number;
```
Rugvip, line 38: I think it'd be ideal to control this from the backend, to keep things simple in a split deployment. The backend uses 55s, but we could make that configurable so that you can bring it down to 25 or so. See https://github.com/backstage/backstage/blob/a096e9ff2d553c7ead7df62794b33babff1d3f37/plugins/events-backend/src/service/hub/createEventBusRouter.ts#L145

What I'm thinking we could to do control this form the backend is to communicate the timeout in a response header, and default to 55s if no header is present.

Beyond that I'm not exactly sure what the effective timeout should be given some known timeout in the backend. I'd imagine it's ideal to typically be closing connections from the frontend and backend first, but not entirely sure which one. My guess would be that letting things close in the backend might produce cleaner logs so we could for example add 1s to the client-side timeout.

Rugvip (commented):
Thank you! 👍 

Couple of ideas to refine the implementaiton

Rugvip (changes requested):
Just to be clear :grin:

Rugvip (commented):
It's possible that this is due to https://github.com/backstage/backstage/pull/27407/commits/1c7d10eea1bb34d353230c877781a534a47fee5b as well, waiting for the body to close seems to hang in many situations, and just reading it out as text seems to be a lot more resilient. It'll be part of the patch release for that fix so you can try it out.

kuangp:
> It's possible that this is due to [1c7d10e](https://github.com/backstage/backstage/commit/1c7d10eea1bb34d353230c877781a534a47fee5b) as well, waiting for the body to close seems to hang in many situations, and just reading it out as text seems to be a lot more resilient. It'll be part of the patch release for that fix so you can try it out.

unfortunately that fix didn't address my issue but I updated this change based on your comments!

plugins/events-node/config.d.ts
kuangp: updated! hopefully I implemented it in line with what you described

plugins/events-node/src/api/EventsService.ts
```diff
@@ -55,3 +55,8 @@ export type EventsServiceSubscribeOptions = {
  * @public
  */
 export type EventsServiceEventHandler = (params: EventParams) => Promise<void>;
+
+/**
+ * @public
+ */
+export const EVENTS_NOTIFY_TIMEOUT_HEADER = 'x-backstage-events-notify-timeout';
```
kuangp, line 62: Not sure if there's already some convention in this project for custom Backstage headers - let me know if I should change this

plugins/events-node/src/api/DefaultEventsService.ts
kuangp: I'm assuming this involves passing an Abort controller signal to the underlying fetch call? If so, I'm not too sure how to go about that here since we're using a `DefaultApiClient` that is auto generated through the openapi spec and I'm unfamiliar with how to extend that interface - mind providing some guidance? 🙏

kuangp:
cc: @Rugvip no rush - just wanted to make sure this didn't fall off your radar

plugins/events-node/src/api/EventsService.ts
Rugvip: ah, only thing is to not use the x- prefix as that's generally discouraged nowadays

kuangp: ah ok cool - thanks for informing me, was not aware of the updated standards

kuangp:
Hey @Rugvip sorry to ping again and I know you guys are probably busy with KubeCon this week but was hoping to have this in for the next release if possible 🙏

kuangp:
Hey @Rugvip another gentle nudge here when you get a chance 🙏

Rugvip (approved):
Yep sorrry, it looks good, let's :shipit:!

If you want to try it out in your internal instance you can generate patches by running this in a local clone of the Backstage repo on latest `master`:

```sh
yarn backstage-repo-tools generate-patch plugins/events-backend --target ../internal-backstage-project
yarn backstage-repo-tools generate-patch plugins/events-node --target ../internal-backstage-project
```

PR #27332, opened 2024-10-24
docs/backend-system: add backend common deprecation section to migration docs

### LINKED ISSUE #27279: 🐛 Bug Report: The deprecation process is poorly executed
### 📜 Description

This is not a code issue. This is a people/communication/documentation problem.

I understand that `@backstage/backend-common` is deprecated. The process to deprecate it has provided a poor developer experience. 

The primary APIs we were using were `errorHandler` for `express` and `getVoidLogger` for unit testing. I searched for those things directly which resulted in 404s on the documentation site as well as 404s in the GitHub repo. Ultimately I had to search Discord to find other people asking the same questions and looking for the same breadcrumbs to be able to get the system upgraded. This process should be so much better and more convenient.

### 👍 Expected behavior

The Backstage team provides a _clear path_ for developers to follow to replace the deprecated APIs in their code with the new APIs and the where to import those new APIs from.

There is nothing but dead ends available to upgrading developers. The package itself has an incomplete deprecation message. Looks like it was _almost_ going to be useful.

image

The documentation where I would expect to find statements like "replace `errorHandler` with `X` from `@backstage/new-package-name`" is just a [404 now](https://backstage.io/docs/reference/backend-common.errorhandler/). [This](https://backstage.io/docs/reference/backend-common.getrootlogger/) would be a great place to provide a diff of the upgrade process (before and after with the new API location).

image

Maybe find the [old package in source](https://github.com/backstage/backstage/tree/master/packages/backend-common) and look for that documentation there? Nope. Another dead end. 

image

### 👟 Reproduction steps

1. Try to find an upgrade path from the deprecated APIs provided in `@backstage/backend-common`.

_No response_

### 🖥️ Your Environment

_No response_

None

Rugvip:
Fixes #27279

### DIFF
--- a/docs/backend-system/building-backends/08-migrating.md
+++ b/docs/backend-system/building-backends/08-migrating.md
@@ -219,7 +219,7 @@ These are the deprecation messages for the most common replacements:
 - `createLegacyAuthAdapters` - Migrate to use the new backend system and auth services instead.
 - `errorHandler` - Use `MiddlewareFactory.create.error` from `@backstage/backend-defaults/rootHttpRouter` instead.
 - `getRootLogger` - This function will be removed in the future. If you need to get the root logger in the new system, please check out this documentation: https://backstage.io/docs/backend-system/core-services/logger
-- `getVoidLogger` - This function will be removed in the future. If you need to mock the root logger in the new system, please use `mockServices.logger.mock()` from `@backstage/test-utils` instead.
+- `getVoidLogger` - This function will be removed in the future. If you need to mock the root logger in the new system, please use `mockServices.logger.mock()` from `@backstage/backend-test-utils` instead.
 - `legacyPlugin` - Fully use the new backend system instead.
 - `loadBackendConfig` - Please migrate to the new backend system and use `coreServices.rootConfig` instead, or the [@backstage/config-loader#ConfigSources](https://backstage.io/docs/reference/config-loader.configsources) facilities if required.
 - `loggerToWinstonLogger` - Migrate to use the new `LoggerService` instead.
--- a/docs/backend-system/building-backends/08-migrating.md
+++ b/docs/backend-system/building-backends/08-migrating.md
@@ -210,6 +210,25 @@ argument of `createBackend`.
 
 :::
 
+## Removing `@backstage/backend-common`
+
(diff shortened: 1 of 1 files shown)


docs/backend-system/building-backends/08-migrating.md
```diff
@@ -210,6 +210,25 @@ argument of `createBackend`.
...
+## Removing `@backstage/backend-common`
+
+The `@backstage/backend-common` package has been deprecated as part of moving to the new backend system, an you will need to replace existing usage of it. All exports from the package have been marked as deprecated in the last few releases of the package, and each export has its own deprecation message that explains how to replace that particular export.
+
+These are the deprecation messages for the most common replacements:
+
+- `createLegacyAuthAdapters` - Migrate to use the new backend system and auth services instead.
+- `errorHandler` - Use `MiddlewareFactory.create.error` from `@backstage/backend-defaults/rootHttpRouter` instead.
+- `getRootLogger` - This function will be removed in the future. If you need to get the root logger in the new system, please check out this documentation: https://backstage.io/docs/backend-system/core-services/logger
+- `getVoidLogger` - This function will be removed in the future. If you need to mock the root logger in the new system, please use `mockServices.logger.mock()` from `@backstage/test-utils` instead.
```
freben, line 222:
```suggestion
- `getVoidLogger` - This function will be removed in the future. If you need to mock the root logger in the new system, please use `mockServices.logger.mock()` from `@backstage/backend-test-utils` instead.
```

freben (approved):

docs/backend-system/building-backends/08-migrating.md
Rugvip: Ah shit, that's what's in the published deprecation message 😅. Alright, good to have it fixed here at least

planeiii-te:
@Rugvip Thanks for updating the documentation with these pointers! One export that was missed was the `Git` helper. Here's a [dead docs link](https://backstage.io/docs/reference/backend-common.git.log) referencing it and here a [a few references](https://github.com/search?type=code&q=%2Fimport.*%5Cbgit%5Cb.*+from+.%40backstage%5C%2Fbackend-common%2F) in the wild.

PR #27341, opened 2024-10-24
catalog: Add `entity_ref` to the `final_entities` table

benjdlambert:
We're hoping to move away from needing to join the `refresh_state` table with a lot of the read path SQL requests to improve performance, mainly across read-replicas.

### DIFF
--- a/plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
+++ b/plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
@@ -38,6 +38,13 @@ exports.up = async function up(knex) {
         .where('r.entity_id', knex.raw('f.entity_id')),
     })
     .table('final_entities as f');
+
+  // SQLite does not support ALTER COLUMN.
+  if (!knex.client.config.client.includes('sqlite3')) {
+    await knex.schema.alterTable('final_entities', table => {
+      table.text('entity_ref').notNullable().alter({ alterNullable: true });
+    });
+  }
 };
 
 /**
--- a/plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
+++ b/plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// @ts-check
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function up(knex) {
+  await knex.schema.alterTable('final_entities', table => {
+    table
+      .text('entity_ref')
+      .nullable()
+      .comment(
+        'The entity reference of the entity that was created from the catalog processing',
+      );
+  });
+
+  // Set the default entity_ref in final_entities from the refresh_state table
+  await knex.raw(
+    `UPDATE final_entities SET entity_ref = refresh_state.entity_ref FROM refresh_state WHERE refresh_state.entity_id = final_entities.entity_id`,
+  );
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = async function down(knex) {
+  await knex.schema.alterTable('final_entities', table => {
+    table.dropColumn('entity_ref');
+  });
+};
--- a/plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
+++ b/plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
@@ -26,8 +26,6 @@ exports.up = async function up(knex) {
       .string('entity_ref')
       .nullable()
       .comment('The entity reference of the entity');
-
-    table.index('entity_ref', 'entity_ref_idx');
   });
 
   await knex
@@ -39,12 +37,12 @@ exports.up = async function up(knex) {
     })
     .table('final_entities as f');
 
-  // SQLite does not support ALTER COLUMN.
-  if (!knex.client.config.client.includes('sqlite3')) {
-    await knex.schema.alterTable('final_entities', table => {
-      table.string('entity_ref').notNullable().alter({ alterNullable: true });
+  await knex.schema.alterTable('final_entities', table => {
+    table.string('entity_ref').notNullable().alter({ alterNullable: true });
+    table.unique(['entity_ref'], {
+      indexName: 'final_entities_entity_ref_uniq',
     });
-  }
+  });
 };
 
 /**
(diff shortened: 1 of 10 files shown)


plugins/catalog-backend/src/tests/migrations.test.ts
Rugvip, line 1: 👌

plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
```diff
@@ -0,0 +1,51 @@
...
+// @ts-check
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function up(knex) {
+  await knex.schema.alterTable('final_entities', table => {
+    table
+      .text('entity_ref')
```
freben, line 26:
```suggestion
      .string('entity_ref')
```

especially since we are very likely to want to index on this in the future. Also that's what the type is in `refresh_state`

plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
```diff
@@ -0,0 +1,51 @@
...
+      .nullable()
+      .comment(
+        'The entity reference of the entity that was created from the catalog processing',
```
freben, line 29:
```suggestion
        'The entity reference of the entity',
```

plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
```diff
@@ -669,11 +658,8 @@ export class DefaultEntitiesCatalog implements EntitiesCatalog {
   }
 
   async entityAncestry(rootRef: string): Promise<EntityAncestryResponse> {
-    const [rootRow] = await this.database<DbRefreshStateRow>('refresh_state')
-      .leftJoin<DbFinalEntitiesRow>('final_entities', {
-        'refresh_state.entity_id': 'final_entities.entity_id',
-      })
-      .where('refresh_state.entity_ref', '=', rootRef)
+    const [rootRow] = await this.database<DbFinalEntitiesRow>('final_entities')
```
freben, line 661: Should we really change this function at all? Now you're changing its semantics to deal with already-stitched things, which makes it miss for example entities that have processing errors. The bodies will also be the stitched ones, which wasn't the case before.

Essentially I think if we want to make any changes to this function then maybe it should be in a separate PR and probably as a breaking change even (shunting it to a 2.0 worst case).

plugins/catalog-backend/src/tests/migrations.test.ts
```diff
@@ -395,4 +395,79 @@ describe('migrations', () => {
...
+          entity_ref: 'k:ns/n',
+          unprocessed_entity: '{}',
+          processed_entity: '{}',
+          errors: '[]',
+          next_update_at: knex.fn.now(),
+          last_discovery_at: knex.fn.now(),
+        })
+        .into('refresh_state');
+
+      // Insert a simple URL before the migration
```
freben, line 421: url? 🤔

plugins/catalog-backend/src/tests/migrations.test.ts
```diff
@@ -395,4 +395,79 @@ describe('migrations', () => {
...
+          hash: '3f5a4d6ba8507be297bb7cd87c4b55b63e3f4c14',
+          stitch_ticket: '52367ed7-120b-405f-b7e0-cdd90f956312',
+          final_entity: '{}',
+        })
+        .into('final_entities');
+
+      await migrateUpOnce(knex);
+
+      // verify that the entity_ref column has been added
+      const columnInfo = await knex('final_entities').columnInfo();
```
freben, line 434: TIL!

can you do this exact same but inverse check before `migrateUpOnce`, just to demonstrate that it went from not defined to defined?

plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
benjdlambert: I'm not sure I understand what you mean, the code always stitched with `final_entities`, it's just a reverse lookup with the `refresh_state` table from `entity_ref` to the actual `final_entities` row.

It's always used stitched data as far as I can tell?

freben: Ah I missed the join and only looked here locally. OK!

plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
```diff
@@ -0,0 +1,59 @@
...
+ * @returns { Promise<void> }
+ */
+exports.up = async function up(knex) {
+  await knex.schema.alterTable('final_entities', table => {
+    table
+      .string('entity_ref')
+      .nullable()
+      .comment('The entity reference of the entity');
+
+    table.index('entity_ref', 'entity_ref_idx');
```
freben, line 30:
```suggestion
    table.index('entity_ref', 'final_entities_entity_ref_idx');
```

plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = async function down(knex) {
+  await knex.schema.alterTable('final_entities', table => {
+    table.dropIndex([], 'entity_ref_idx');
```
freben, line 56:
```suggestion
    table.dropIndex([], 'final_entities_entity_ref_idx');
```

(or `dropUnique` if you like the comment above)

plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
freben: As a matter of fact ... how do you feel about

```ts
    table.unique(['entity_ref'], {
      indexName: 'final_entities_entity_ref_uniq',
    });
```

and then `.dropUnique` below? It's not _strictly_ necessary at this point, but on the other hand it's absolutely vital that it holds true, and when we refactor how processing happens, maybe we'll be glad that we have the constraint in place.

Could introduce that later, sure. But does it hurt to be stringent?

plugins/catalog-backend/src/database/operations/stitcher/performStitching.ts
```diff
@@ -212,11 +213,12 @@ export async function performStitching(options: {
       final_entity: JSON.stringify(entity),
       hash,
       last_updated_at: knex.fn.now(),
+      entity_ref: entityRef,
     })
     .where('entity_id', entityId)
     .where('stitch_ticket', stitchTicket)
     .onConflict('entity_id')
-    .merge(['final_entity', 'hash', 'last_updated_at']);
+    .merge(['final_entity', 'hash', 'last_updated_at', 'entity_ref']);
```
freben, line 221: this SHOULD be impossible to change after the fact - could probably revert this line

plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
```diff
@@ -698,16 +684,14 @@ export class DefaultEntitiesCatalog implements EntitiesCatalog {
       const parentRows = await this.database<DbRefreshStateReferencesRow>(
         'refresh_state_references',
       )
-        .innerJoin<DbRefreshStateRow>('refresh_state', {
-          'refresh_state_references.source_entity_ref':
-            'refresh_state.entity_ref',
-        })
+
```
freben, line 687:
```suggestion

```

plugins/catalog-backend/migrations/20241024104700_add_entity_ref_to_final_entities.js
freben: Updated to be a unique constraint

jhaals (approved):

PR #27345, opened 2024-10-25
Update Google Auth plugin to add Frontend Visibility header

dlaird-ovo:

When running locally I found that the following frontend code was not returning the `google` keyword

```
const providersConfig = configApi.getOptionalConfig('auth.providers');
const configuredProviders = providersConfig?.keys() || [];
```

It was however present and could be seen via the DevTools plugin, and by the backend code.
I tried different config blocks
```
auth:
  environment: development
  providers:
    github:
      development:
        <blah>
    chicken:
      development:
        <blah>
    cfaccess:
      <blah>
```

which did return all keywords.  I reviewed the differences with the Github Auth provider and the Google Auth provider and spotted this difference which I believe will resolve my issue

### DIFF
--- a/.changeset/few-hornets-rush.md
+++ b/.changeset/few-hornets-rush.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-auth-backend-module-google-provider': patch
+---
+
+Fix visibility of config for use in front end code
--- a/plugins/auth-backend-module-google-provider/config.d.ts
+++ b/plugins/auth-backend-module-google-provider/config.d.ts
@@ -18,6 +18,7 @@ export interface Config {
   /** Configuration options for the auth plugin */
   auth?: {
     providers?: {
+      /** @visibility frontend */
       google?: {
         [authEnv: string]: {
           clientId: string;


Rugvip (approved):
Great, thank you! 👍

Rugvip:
> Would be great to patch the 1.32.x release as well if this is the correct fix :-)

Would you be happy to try out the new `generate-patch` command from #27331 instead? You'd check out the latest version of this repo and run the following in the repo root:

```sh
yarn backstage-repo-tools generate-patch plugins/auth-backend-module-google-provider --target ../path-to-your-internal-project
```

We'd like to try out that approach a bit to gauge how viable it is. It does require you to be using Yarn 2+ in your project though

dlaird-ovo:
I am running Yarn 4.5.x and when I run - (0.10.0) it does not have that option - so does it only work for Yarn 2.x? (just confirming 2+ does not mean 4.5.x should work)?

dlaird-ovo:
Ah sorry realised my target/source were wrong way around :-) 
Now have a patch in my tree - will try it out - 👍

PR #27348, opened 2024-10-25
repo-tools: more reliable shelling to git for generate-patch

Rugvip:
🧹 , followup for #27331, turns out plain `git` can vary quite a bit 😁


freben (approved):
wow

PR #27355, opened 2024-10-25
fix(catalog): fix search by `displayName` when searching User and Group entities in the catalog table

### LINKED ISSUE #27354: 🐛 Bug Report: Searching User or Group entities by displayName in the catalog table doesn't work
### 📜 Description

When searching in the catalog table an entity of `kind: User` or `kind: Group` using the value of its `spec.profile.displayName` no results are returned

### 👍 Expected behavior

When searching in the catalog table it is possible to search for entities of `kind: Component` by their `metadata.title`
I would expect the same to be possible for `kind: User` and `kind: Group` but with `spec.profile.displayName` as they don't have `metadata.title`.
Moreover, using global search it is possible to search users and groups by `displayName`

Failing to search by displayName in the catalog table : 
Capture d’écran 2024-10-25 à 12 20 48

Working with the global search : 
Capture d’écran 2024-10-25 à 12 21 11

Expected behavior : 
Capture d’écran 2024-10-25 à 12 23 15

### 👟 Reproduction steps

1. Go to [demo site](https://demo.backstage.io/catalog?filters%5Bkind%5D=user&filters%5Bnamespace%5D=default&filters%5Buser%5D=all&limit=20)
2. Search user by its `displayName` (actually it's the value displayed on each rows). If you put just first word it will usually work because the `metadata.name` is something like `amelia.park` so be sure to put more than just the first word
3. No results displayed

_No response_

### 🖥️ Your Environment

_No response_

Julien-Hery:

This PR fixes #27354. When searching entities of `kind: User` or `kind: Group` in the catalog table by their `spec.profile.displayName` it doesn't work.

When the SQL query for the catalog is built, something like this is created : 
```sql
select search.entity_id from search
where key in ('metadata.name', 'metadata.title', 'spec.profile.displayName')
and (value like '%amelia p%')
``` 
but in the database the value of `key` column is in lowercase so `spec.profile.displayName` will never match because it needs to be `spec.profile.displayname`.

I also added a searchable hidden column to the catalog table with the value of the `displayName` (like it's done for other fields)


plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
```diff
@@ -422,7 +422,11 @@ export class DefaultEntitiesCatalog implements EntitiesCatalog {
       } else {
         const matchQuery = db<DbSearchRow>('search')
           .select('search.entity_id')
-          .whereIn('key', textFilterFields)
+          // textFilterFields must be lowercased to match searchable keys in database, i.e. spec.profile.displayName -> spec.profile.displayname
+          .whereIn(
+            'key',
+            textFilterFields.map(field => field.toLocaleLowerCase('en-US')),
```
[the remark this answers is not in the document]
Julien-Hery: Not a catalog maintainer myself but I think in SQL it's possible to do something like `WHERE 'your_value' ILIKE ANY(your_array)` which could be translated to something `.whereRaw('key ILIKE ANY(?)', [value])` with Knex. 
I don't think knex comes with a built in `WHERE ... ILIKE ANY ...`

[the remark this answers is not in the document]
Julien-Hery: From what I can see `ILIKE ANY` is Postgres specific so it's probably not a good idea to use it as a raw query 
I also tried using the array directly in the `.whereILike` but I get the following error from `better-sqlite` : `Expected second argument to be a string`

I guess mapping to lowercase is the best option even if I agree using ilike would have been clearer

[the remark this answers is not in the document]
camilaibs: Got it, thanks for responding so quickly. I haven't tried it myself, was just curious. Let's see what the @backstage/catalog-maintainers think 🙂

vinzscam (commented):
@Julien-Hery have you enabled pagination on your catalog?

Julien-Hery:
> @Julien-Hery have you enabled pagination on your catalog?

@vinzscam I'm not sure to understand, I used the catalog that is on Backstage repo which looks like this : 

```jsx
<Route
  path="/catalog"
  element={<CatalogIndexPage pagination={{ mode: 'offset', limit: 20 }} />}
/>
```

It has pagination settings, should I try something else ? 🙂

plugins/catalog/src/components/CatalogTable/columns.tsx
```diff
@@ -180,6 +180,16 @@ export const columnFactories = Object.freeze({
       searchable: true,
     };
   },
+  createDisplayNameColumn(options?: {
```
vinzscam, line 183: what's the purpose of adding this column? 🤔

plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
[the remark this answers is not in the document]
vinzscam: good catch @Julien-Hery!

plugins/catalog-react/src/filters.ts
```diff
@@ -94,6 +94,7 @@ export class EntityTextFilter implements EntityFilter {
     const partialMatch = this.toUpperArray([
       entity.metadata.name,
       entity.metadata.title,
+      (entity.spec?.profile as { displayName?: string })?.displayName,
```
vinzscam, line 97: I don't think this change is needed as paginated tables don't rely on this method.

```suggestion
```

vinzscam (changes requested):
the changes for making the fields lowercase look good to me. But I'm not sure the new column and the change to `EntityTextFilter` are needed.

plugins/catalog/src/components/CatalogTable/columns.tsx
Julien-Hery: @vinzscam from what I saw the filtering is made by two different things : backend filter (fixed with the lowercasing in the SQL query) and "frontend" filter (native filtering from the table). If I fix only one of them it doesn't work as expected

I tried again to remove it in case I missed something but it really doesn't work without this, it's related to the search feature of `@material-table`

plugins/catalog-react/src/filters.ts
Julien-Hery: @vinzscam You are right it seems that this one can be removed 🤔 Is there a way to use non-paginated table ? Should we keep it for non-paginated table ?

plugins/catalog/src/components/CatalogTable/columns.tsx
vinzscam: mmm I think we have an issue as the `@material-table` search shouldn't be used at all when using pagination

Julien-Hery: @vinzscam So what do we do ? Do you want to check on your side or open an issue about it ? Because for now with my PR it works and fixes a bug so maybe the issue with material-table could be fixed in a separate PR 🙂

vinzscam (changes requested):
alright here we are. So, the `OffsetPaginatedCatalogTable` has a filtering issue at the moment, which is going to be fixed by #27129. Once #27129 is merged, I believe only the changes made by this PR to `catalog-backend` will be enough to fix the issue when searching for groups.

Julien-Hery:
> alright here we are. So, the `OffsetPaginatedCatalogTable` has a filtering issue at the moment, which is going to be fixed by #27129. Once #27129 is merged, I believe only the changes made by this PR to `catalog-backend` will be enough to fix the issue when searching for groups.

@vinzscam I updated the PR and the changelog, it works with the backend part only now :)

vinzscam (approved):
let's go! 🚀

PR #27359, opened 2024-10-25
refactor(backend-dynamic-feature-service): Allow providing config-aware root logger options

davidfestal:

When enabling the dynamic feature in the backend application with the 1-liner `dynamicPluginsFeatureLoader`, the root logger is overridden in order to also hide secret values defined in config schemas of dynamic plugins.
However we want users of this feature loader to be able to customize as much as possible the behavior of this overridden root logger.
This was already available with the `transports`, `level`, and `format` option values.

However there are cases where we want to define the above options based on the application config (think of a rotating logging mechanism that should be driven by configuration for example).

This PR changes these logger-related options, and gathers them under a single `logger` option value which is a function taking a `Config`.

_**NOTE:** it is required to switch the Janus backstage-showcase backend application to the one-liner activation of the dynamic backend plugins support, while still providing audit logging through config-based custom transports and format._

### DIFF
--- a/.changeset/bright-singers-do.md
+++ b/.changeset/bright-singers-do.md
@@ -1,0 +1,10 @@
+---
+'@backstage/backend-dynamic-feature-service': minor
+---
+
+**BREAKING** The `dynamicPluginsFeatureLoader` options related to the root logger behavior (`transports`, `level`, `format`) are now gathered under a single `logger` option which is a function taking an optional `Config` argument and returning the logger options.
+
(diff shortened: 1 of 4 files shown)


davidfestal:
cc @schultzp2020

schultzp2020 (approved):
/approve
/lgtm

+1 to the breaking changes because we will need to add the auditor service here as well.

Rugvip (approved):
:shipit:

PR #27361, opened 2024-10-25
two more catalog faq entries

freben:
These have been asked a number of times


docs/features/software-catalog/faq.md
```diff
@@ -42,3 +42,29 @@ First, performance. As is described [here](#can-i-call-the-catalog-itself-from-i
...
+
+To a large degree, this boils down to user experience. The catalog is an eventually consistent system that constantly tries to mirror external realities. Users make changes in catalog-info files, or things update in external systems, and those changes get streamed in and settle over time inside the catalog. But throwing an error in a processor, instantly aborts processing of that entity and stops its ingestion. Now imagine being a large organization where these changes happen maybe hundreds of times per day. Owners of catalog-info files will constantly be surprised by their files "breaking" in ingestion, maybe a very long time after they were initially created - they were valid at their time of creation and haven't been touched since! This is very frustrating and slows down your users because things break silently for reasons out of your control.
```
Parsifal-M, line 52: Do we need the word `eventually` here? or should the catalog just be a consistent system? 😅

Rugvip:
image

I should've kept reading 😂

Parsifal-M (commented):
Awesome read! 👍

docs/features/software-catalog/faq.md
freben: As long as the "storage model" essentially is external systems - as opposed to controlling the lifecycle of its data entirely within its own database and edited directly there - we cannot possibly be directly consistent. Changes stream in with some lag. And regarding the data validity, you can always for example have someone edit your LDAP data over in the HR department and suddenly a file somewhere on GitHub is "invalid" for some definition of invalid - but only actually a short while later when the LDAP sync has run :)

Parsifal-M: Ah I see what you mean now, you mean that eventually post-streaming of data it becomes consistent, not that when you first start with backstage your catalog is inconsistent but later will become consistent 😅

freben: Yep its output eventually becomes consistent with whatever you pointed at to be consumed, as it changes over time.

Parsifal-M (approved):

PR #27367, opened 2024-10-25
Add autocomplete to GitlabRepoUrlPicker

swnia:

This is the code from #25842 with little adjustments. As the other one seemed stale, I opened a new PR.

### DIFF
--- a/plugins/scaffolder-backend-module-bitbucket-cloud/src/autocomplete/autocomplete.ts
+++ b/plugins/scaffolder-backend-module-bitbucket-cloud/src/autocomplete/autocomplete.ts
@@ -25,7 +25,7 @@ export async function handleAutocompleteRequest({
   resource: string;
   token: string;
   context: Record<string, string>;
-}): Promise<{ results: { title: string; id?: string }[] }> {
+}): Promise<{ results: { title?: string; id: string }[] }> {
   const client = BitbucketCloudClient.fromConfig({
     host: 'bitbucket.org',
     apiBaseUrl: 'https://api.bitbucket.org/2.0',
@@ -92,12 +92,15 @@ export async function handleAutocompleteRequest({
           'Missing workspace and/or repository context parameter',
         );
 
-      const results: { title: string }[] = [];
+      const results: { title: string; id: string }[] = [];
 
       for await (const page of client
         .listBranchesByRepository(context.repository, context.workspace)
         .iteratePages()) {
-        const names = [...page.values!].map(p => ({ title: p.name! }));
+        const names = [...page.values!].map(p => ({
+          title: p.name!,
+          id: p.name!,
+        }));
         results.push(...names);
       }
 
--- a/plugins/scaffolder-backend-module-bitbucket-cloud/src/autocomplete/autocomplete.ts
+++ b/plugins/scaffolder-backend-module-bitbucket-cloud/src/autocomplete/autocomplete.ts
@@ -34,12 +34,11 @@ export async function handleAutocompleteRequest({
 
   switch (resource) {
     case 'workspaces': {
-      const results: { title: string; id: string }[] = [];
+      const results: { title?: string; id: string }[] = [];
 
       for await (const page of client.listWorkspaces().iteratePages()) {
         const slugs = [...page.values!].map(p => ({
-          title: p.slug!,
-          id: p.uuid!,
+          id: p.slug!,
         }));
         results.push(...slugs);
       }
@@ -50,14 +49,13 @@ export async function handleAutocompleteRequest({
       if (!context.workspace)
         throw new InputError('Missing workspace context parameter');
 
-      const results: { title: string; id: string }[] = [];
+      const results: { title?: string; id: string }[] = [];
 
       for await (const page of client
         .listProjectsByWorkspace(context.workspace)
         .iteratePages()) {
         const keys = [...page.values!].map(p => ({
-          title: p.key!,
-          id: p.uuid!,
+          id: p.key!,
         }));
         results.push(...keys);
       }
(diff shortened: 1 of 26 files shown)


plugins/scaffolder-node/report-alpha.api.md
```diff
@@ -23,6 +23,7 @@ export type AutocompleteHandler = ({
 }) => Promise<{
   results: {
     title: string;
+    id?: string;
```
benjdlambert, line 26: I think that we should make this `id` required, and name be optional.

Name would be the human readable version that's rendered in the UI if it's available, otherwise it falls back to the `id` instead.

it's technically a breaking change, but because this stuff is on alpha, it's fine for us to release this as a `minor` change.

I see that you've already updated the bitbucket one too so that's great, but I think it makes sense to make that change now.

benjdlambert (changes requested):
Hey :wave: thanks for picking this up, cc @fedy97 too that this is in flight. 

Just a quick high level comment that I think we want to tackle now :pray:

plugins/scaffolder-node/report-alpha.api.md
swnia: I think we would need to define what the ID holds then. E.g. for [this structure](https://github.com/backstage/backstage/pull/27367/files#diff-a3a6fd3b5458cb6adb4e14446e4fbae8be3226c7dc7dadbd64e66abd957eef99R100) there is no `uuid`. Also, the gitlab `publish:mr` [action](https://github.com/backstage/backstage/blob/a096e9ff2d553c7ead7df62794b33babff1d3f37/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabMergeRequest.ts) requires `group` and `title` and not the projects id.

Therefore, I do not know what the whole impact of that change would be.

swnia: Hey @benjdlambert, do you really want to make this change now? Should the dependant modules be updated right away? How would you want to go about this?

nia-potato:
hi @benjdlambert can we merge this as a minor change anytime soon? Our organization would also benefit greatly from this feature.

plugins/scaffolder-node/report-alpha.api.md
benjdlambert: Hey @swnia, sorry for the delay in getting back, had a week off last week and feeling pretty sick the beginning of this week.

I think that we should make this change. I think that it was a bit of an oversight at the beginning not to have `id` and `title?` in the first place, and the title is what is rendered in the dropdown but the ID is what is used to call the value unless it's special.

I think it's OK to extend if needed, but that should be on a per autocomplete basis, and perhaps cast in the usage of the `.autocomplete()` call in the field extensions themselves.

`scaffolderClient.autocomplete<GitlabProjects>()` or something?

swnia: The RepoUrlPicker does not seem to be using the `.autocomplete()` call... I'll see what changes are necessary to make it work for Gitlab.

benjdlambert: It's used in the `GitlabRepoUrlPicker` though right? It would be there that we could cast if necessary.

swnia: Yes, my search over the code was too narrow which 🤦 So, yes, it's [here](https://github.com/backstage/backstage/pull/27367/files#diff-8c91e7f5b7e1bfd5878797e0bc07aab845b75c00a0df03caee9b0f2b68b69548R86)

swnia:
@benjdlambert I think I updated all references to the data structure to include the id. Ideally, the `availableRepos` ([here](https://github.com/backstage/backstage/pull/27367/files#diff-08f7af11daad3526af6ada407063fe070af2f83404f72996390d4d5c4db29c02L23)) would include the ID, too. Then, it could be used in templates directly. What is your take on it?

swnia:
I'm not sure why the checks fail. When I run `yarn build:api-reports plugins/scaffolder*`, git detects no changes in the files.

plugins/scaffolder-backend-module-bitbucket-cloud/src/autocomplete/autocomplete.ts
```diff
@@ -34,10 +34,13 @@ export async function handleAutocompleteRequest({
...
   switch (resource) {
     case 'workspaces': {
-      const results: { title: string }[] = [];
+      const results: { title: string; id: string }[] = [];
 
       for await (const page of client.listWorkspaces().iteratePages()) {
-        const slugs = [...page.values!].map(p => ({ title: p.slug! }));
+        const slugs = [...page.values!].map(p => ({
+          title: p.slug!,
+          id: p.uuid!,
```
benjdlambert, line 42: So I think that we don't want to introduce an ID if we don't need it.

The ID should be the same as the title today, it's just that the title might be something other than the ID if there's a better representation of it.

Don't want to make any braking changes if we can avoid it. The ID would be used in the frontend as the `onChange` value, but the `title` would be whats rendered in dropdowns.

Does this make sense?

benjdlambert (changes requested):

plugins/scaffolder-backend-module-bitbucket-cloud/src/autocomplete/autocomplete.ts
swnia: I think it does.

Hopefully, I interpreted that correctly.

benjdlambert:
@benjidotsh can you try out the bitbucket changes here to make sure things look good for you too?

swnia:
@benjdlambert are there any tests specifically that should be performed? Is something else missing for this PR to be merged?

benjidotsh:
@swnia since you're also touching Bitbucket's implementation, I think @benjdlambert just wants to verify whether that still works.

I tried testing it out, but I'm struggling getting the code to work in my instance at all, probably due to a version mismatch.

benjidotsh:
Ok, I finally got it to work after merging the latest changes into this branch.

Looks good! ✅

nia-potato:
@benjdlambert  can we merge this now that all checks are done?

benjdlambert (approved):
OK, let's go with this for now, might wanna revisit some of this autocomplete API, but we can get to that before the next mainline.

PR #27372, opened 2024-10-25
docs(sofware-templates): improves instruction for --no-node-snapshot

sonikro:

I'm just making it clear in the docs that the best practice is to append flags to the NODE_OPTIONS environment variable, since other [debugging tools](https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_how-can-i-set-nodeoptions) rely on that variable, and overriding it directly as it was suggested before was breaking the debugging capabilities.


freben (approved):
Alright!

PR #27377, opened 2024-10-27
Create port.yaml

fokolo:

I added a new plugin we have created.


Rugvip (changes requested):
Sorry this looks very incomplete to me. I'd doesn't  seem to integrate with existing Backstage features.

It also has an unauthenticated endpoint in the backend plugin that lets you fetch an access token: https://github.com/port-labs/backstage-plugin/blob/ac9d85d3ea61fc20dec4191996fbf91bed840a81/backend-plugin/src/service/router.ts#L100

That seems like an extremely bad idea and we can't recommend anyone to use this in it's current form.

fokolo:
Thanks for your feedback :) 

I’ll address the unauthenticated endpoint issue – that’s definitely a mistake. Could you clarify what you meant about the following?

> I'd doesn't seem to integrate with existing Backstage features properly.

Rugvip:
@fokolo Looking at things like this: https://github.com/port-labs/backstage-plugin/blob/ac9d85d3ea61fc20dec4191996fbf91bed840a81/frontend-plugin/src/features/HomePage.tsx#L3-L13

fokolo:
Thank you so much for the feedback!

We fixed you have mentioned. looking forward for your thoughts :)

fokolo:
@Rugvip Can you please review again? and approve or give more comments?

Thanks a lot!

PR #27385, opened 2024-10-29
Added catalog-webhook plugin & image for the plugin directory

jgentes:
Hopefully this is an easy PR. Added a new public plugin that provides the ability to push catalog changes to a remote URL.


Rugvip:
Two quick improvements to consider to make sure this is more broadly usable:

- https://github.com/ForgedApps/backstage-catalog-webhook-plugin/blob/61a1afe16b917a5d186305f20bdbcc02e37552b8/src/webhookProcessor.ts#L42-L45: could this use the [task scheduler](https://backstage.io/docs/backend-system/core-services/scheduler) instead? It lets it scale across multiple instances and is gonna be more robust than `setInterval`
- https://github.com/ForgedApps/backstage-catalog-webhook-plugin/blob/61a1afe16b917a5d186305f20bdbcc02e37552b8/src/cache.ts#L4: could this use the [cache service](https://backstage.io/docs/backend-system/core-services/cache) instead? same thing here - helps it scale across instances and avoids local filesystem access

jgentes:
> Two quick improvements to consider to make sure this is more broadly usable:
> 
> * https://github.com/ForgedApps/backstage-catalog-webhook-plugin/blob/61a1afe16b917a5d186305f20bdbcc02e37552b8/src/webhookProcessor.ts#L42-L45: could this use the [task scheduler](https://backstage.io/docs/backend-system/core-services/scheduler) instead? It lets it scale across multiple instances and is gonna be more robust than `setInterval`
> * https://github.com/ForgedApps/backstage-catalog-webhook-plugin/blob/61a1afe16b917a5d186305f20bdbcc02e37552b8/src/cache.ts#L4: could this use the [cache service](https://backstage.io/docs/backend-system/core-services/cache) instead? same thing here - helps it scale across instances and avoids local filesystem access

Thanks @Rugvip these are good suggestions, I wasn't aware of those - I'll do a quick refactor!

jgentes:
Hi @Rugvip, I made both changes. Felt good to get rid of the node-persist dependency. Thanks for the smart suggestions, feels better now :)

jgentes:
Hi folks, let me know if there are any enxt steps I need to take to add my plugin to the directory, thanks!

Rugvip (approved):
Nice, thank you! 👍

Issue #27346, opened 2024-10-25
🚀 Feature: Able to pass --quiet for eslint to suppress rules marked as warnings


Glenf:
### 🔖 Feature description

In `backstage-cli repo lint` a new option to pass `--quiet` globally to suppress eslint rules marked as warnings.

### 🎤 Context

We are introducing some global lint rule changes in our setup and would like to keep our CI running. These rules are introduced as warnings at first, and aren't mandatory for old code but are required for new changes. We have pre-commit hooks in place for new changes. 

As linting is our gate in CI, failure is blocking. The `--quiet` option would allow our CI runs to pass and give us "annoying squiggly lines" in editor as a reminder to keep on cleaning.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

Glenf:
I dug a bit deeper into this. It seems that eslint node.js API allows passing flags to it, but this has been introduced in eslint 9. 

So options for this are to overwrite `max-warnings` (to have -1 when passing `--quiet` or upgrading to newer eslint. 
https://github.com/backstage/backstage/blob/7c26bd2334b5d075127c358638a1f81f8e573b6d/packages/cli/src/commands/repo/lint.ts#L201

camilaibs:
Hi @Glenf, would you be interested in giving it a try and open a pull request 🙂 ?

Glenf:
Yes, I can do a PR for this.

camilaibs:
Awesome 🎉 , I assigned the issue to you.
Thanks in advance for the contribution 🙌🏻

camilaibs:
Hi @Glenf, just so you know, we are at a reduced capacity this week, so it might take us a bit longer to review your pull request 🙂 .

PR #27387, opened 2024-10-29
feat(lint): Add `--quiet` option to repo lint

Glenf:
With `--quiet`, eslint ruled marked as warn do not cause non-zero exit code. This is useful in enviroments, where new new eslint rules are introduced but cannot be enforced immediately. 

Implements backstage/backstage#27346

### DIFF
--- a/.changeset/giant-shoes-reflect.md
+++ b/.changeset/giant-shoes-reflect.md
@@ -1,0 +1,6 @@
+---
+'@backstage/cli': minor
+---
+
+Add `--quiet` flag for repo lint to reduce logging on rules marked as warnings in ESLint.
+This does not affect how max-warnings is handled.
--- a/docs/tooling/cli/03-commands.md
+++ b/docs/tooling/cli/03-commands.md
@@ -97,6 +97,7 @@ Options:
   --successCache            Enable success caching, which skips running tests for unchanged packages that were successful in the previous run
   --successCacheDir <path>  Set the success cache location, (default: node_modules/.cache/backstage-cli)
   --fix                     Attempt to automatically fix violations
+  --quiet                   Report errors only. Rules marked as warnings are not logged out, unless max-warnings threshold is reached.
 ```
 
 ## repo test
--- a/packages/cli/cli-report.md
+++ b/packages/cli/cli-report.md
@@ -453,6 +453,7 @@ Options:
   --successCache
   --successCacheDir <path>
   --max-warnings <number>
+  --quiet
   --fix
   -h, --help
 ```
--- a/packages/cli/src/modules/lint/commands/repo/lint.ts
+++ b/packages/cli/src/modules/lint/commands/repo/lint.ts
@@ -233,9 +233,11 @@ export async function command(opts: OptionValues, cmd: Command): Promise<void> {
     if (runFailed) {
       console.log(chalk.red(`Lint failed in ${relativeDir}`));
       failed = true;
+    } else if (sha) {
+      outputSuccessCache.push(sha);
+    }
 
-      // When doing repo lint, only list the results if the lint failed to avoid a log
-      // dump of all warnings that might be irrelevant
+    if (failed || !opts.quiet) {
       if (resultText) {
         if (opts.outputFile) {
           errorOutput += `${resultText}\n`;
@@ -244,8 +246,6 @@ export async function command(opts: OptionValues, cmd: Command): Promise<void> {
           console.log(resultText.trimStart());
         }
       }
-    } else if (sha) {
-      outputSuccessCache.push(sha);
     }
   }
 
--- a/packages/cli/src/modules/lint/index.ts
+++ b/packages/cli/src/modules/lint/index.ts
@@ -66,6 +66,10 @@ export function registerRepoCommands(command: Command) {
       '--max-warnings <number>',
       'Fail if more than this number of warnings. -1 allows warnings. (default: -1)',
     )
+    .option(
+      '--quiet',
+      'Report errors only. Rules marked as warnings are not logged out, unless max-warnings threshold is reached.',
+    )
     .option('--fix', 'Attempt to automatically fix violations')
     .action(lazy(() => import('./commands/repo/lint'), 'command'));
 }


packages/cli/src/commands/index.ts
```diff
@@ -70,6 +70,10 @@ export function registerRepoCommand(program: Command) {
       'Set the success cache location, (default: node_modules/.cache/backstage-cli)',
     )
     .option('--fix', 'Attempt to automatically fix violations')
+    .option(
+      '--quiet',
+      'Report errors only. Rules marked as warnings do not cause non-zero exit.',
+    )
```
Rugvip, line 76: Hmm, I don't think a `--quiet` flag should ever cause a change in behavior, only reduced logging. I'd actually think that this use-case is something that might already be covered by the `--max-warnings` flags for the individual `package lint`? That flag is picked up by `repo lint` too and lets you increase the max allowed warnings for each package.

Glenf: As I understood this, `--max-warnings` cannot be set for `repo lint` but must be set in package level. 

In our case, we are introducing some rules globally that will cause issues (warnings) with existing packages. I'd want to introduce this rule and start fixing these issues but it is not possible to do it in one go. And because of this, our CI will fail. 

I'd prefer not to set `--max-warnings` in package level, we do have quite a few of those.

Rugvip: @Glenf the problem with adding this at the repo level is that it will break linting at the package level. Linting from the root might not lead to an error while running `lint` in a package will.

I'd really recommend setting the max warning per package as it also gives you a view into which packages have committed to the new lint rules and which still need some work.

Glenf: Out of curiosity, why are eslint rules marked as `warn` considered errors? 

And another question, if Backstage would use ESLint 9, would `--quiet` option supported by it be a better approach then?

Glenf: I did a little test on this and it seems that `--max-warnings` behaviour is different than what ESLint documents. Adding `--max-warnings=-1`, which should ignore all warnings in ESLint, does not have any effect and reports warnings as errors. This is also how the wrapper is implemented. 

Should I create a separate issue out of this?

Glenf: I can do a PR for those. I have a version ready from my testing but it needs a small cleanup.

Glenf:
I did another pass on this one. Now behaviour doesn't change if `--quiet` is passed. 

As an example, here's the output on default values when there are warnings on a package:
repo lint pkg max-warnings default : 0

Here's the same with `max-warnings -1` in package:
repo lint pkg max-warnings -1

And finally with `max-warnings 1` in package level:
repo lint pkg max-warnings 1

package.json
```diff
@@ -28,8 +28,8 @@
     "docker-build": "yarn tsc && yarn workspace example-backend build && yarn workspace example-backend build-image",
     "fix": "backstage-cli repo fix --publish",
     "postinstall": "husky || true",
-    "lint": "backstage-cli repo lint --since origin/master",
-    "lint:all": "backstage-cli repo lint",
+    "lint": "backstage-cli repo lint --since origin/master --quiet",
```
Glenf, line 31: I'm not sure if this is necessary. Everything should work even without this change, because earlier every `warn` was error.

Glenf:
Not stale. I'd be interested if the changes made are enough or if there's something else to be made for this to get merged?

Glenf:
@Rugvip Are the changes made for this PR enough to get this change forward?

package.json
```diff
@@ -84,6 +84,9 @@
     ]
   },
   "prettier": "@backstage/cli/config/prettier",
+  "jest": {
+    "rejectFrontendNetworkRequests": true
```
awanlin, line 88: What's the reason for this change? Seems unrelated 🤔

package.json
```diff
@@ -28,8 +28,8 @@
     "docker-build": "yarn tsc && yarn workspace example-backend build && yarn workspace example-backend build-image",
     "fix": "backstage-cli repo fix --publish",
     "postinstall": "husky || true",
-    "lint": "backstage-cli repo lint --since origin/master",
-    "lint:all": "backstage-cli repo lint",
+    "lint": "backstage-cli repo lint --since origin/master --quiet",
+    "lint:all": "backstage-cli repo lint --quiet",
```
awanlin, line 32: Do we actually want this change? This would be for this repo only and not something that gets shipped. I'm not sure we want to change what has been working well for the project itself.

.changeset/giant-shoes-reflect.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/cli': minor
+---
+
+Add --quiet flag for repo lint to reduce logging on rules marked as warnings in ESLint.
```
awanlin, line 5:
```suggestion
Add `--quiet` flag for repo lint to reduce logging on rules marked as warnings in ESLint.
```

awanlin (changes requested):
Thanks for the contribution @Glenf, mostly reviewing the docs and left a few comments. Still would like @Rugvip to give a follow up review as well. 👍

package.json
Glenf: I'm also interested to know if this is something that's wanted. I can remove this, as there isn't really any change in how this project reports things.

package.json
Glenf: Just me being bad with git. Fixed.

package.json
Glenf: Removed the `--quiet` portion from `lint` scripts.

Glenf:
Not stale

packages/cli/src/commands/repo/lint.ts
```diff
@@ -233,9 +233,11 @@ export async function command(opts: OptionValues, cmd: Command): Promise<void> {
     if (runFailed) {
       console.log(chalk.red(`Lint failed in ${relativeDir}`));
       failed = true;
+    } else if (sha) {
+      outputSuccessCache.push(sha);
+    }
 
-      // When doing repo lint, only list the results if the lint failed to avoid a log
-      // dump of all warnings that might be irrelevant
+    if (failed || !opts.quiet) {
```
vinzscam, line 240: shouldn't this be: 🤔 

```suggestion
    if (failed && !opts.quiet) {
```

Glenf: Logic here is that `failed` will be logged out to console no matter what `opts.quiet` value is. But warnings will not be logged to console if `opts.quiet` is `true`.

Glenf:
Any chance for this one to get merged? Some other changes needed?

Glenf:
I rebased and updated this change. Any comment on this PR on merge possibilities? @Rugvip @awanlin

Glenf:
Now that the default value in `--max-warnings` is changed (https://github.com/backstage/backstage/pull/28699), would there be more interest to in this PR?

.changeset/giant-shoes-reflect.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/cli': minor
```
Rugvip, line 2:
```suggestion
'@backstage/cli': patch
```

Rugvip (commented):
Yep this is indeed back on the radar a bit more now!

I do wonder though, is this redundant at this point? The existing behavior is already quite silent, but the PR now changes the default to be verbose and adds the flag to bring it back to the previous behavior, is that intended?

.changeset/giant-shoes-reflect.md
Glenf: My original thought was to keep it quite verbose on local development but silence it through the flag. e.g. notify developer about potential issues but keep CI logs silent. But I'll check the current verbosity level and what these change.

PR #27391, opened 2024-10-29
Added a `permission-module` template

awanlin:

Added a `permission-module` template to make getting started with the Permissions Framework easier. This will create a backend module that include a very basic Permission Policy which you can then extend. 

The intention is to follow up this PR with a refactor of the current Permissions Framework docs to use this as it will simplify them and make them easier to follow as the boiler plate is handled by this template.

### DIFF
--- a/packages/cli/templates/permission-module/package.json.hbs
+++ b/packages/cli/templates/permission-module/package.json.hbs
@@ -1,0 +1,42 @@
+{
+  "name": "{{name}}",
+  "description": "The {{id}} module for @backstage/plugin-permission-backend",
+  "version": "{{pluginVersion}}",
+  "main": "src/index.ts",
+  "types": "src/index.ts",
+  "license": "{{license}}",
+{{#if privatePackage}}
+  "private": {{privatePackage}},
+{{/if}}
+  "publishConfig": {
+{{#if npmRegistry}}
+    "registry": "{{npmRegistry}}",
+{{/if}}
+    "access": "public",
+    "main": "dist/index.cjs.js",
+    "types": "dist/index.d.ts"
+  },
+  "backstage": {
+    "role": "backend-plugin-module"
+  },
+  "scripts": {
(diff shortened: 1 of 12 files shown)


awanlin:
@backstage/documentation-maintainers just popping this on your radar, no docs changes here but there will be as a follow up. 👍

packages/cli/templates/permission-module/src/customPermissionPolicy.ts
```diff
@@ -0,0 +1,13 @@
...
+} from '@backstage/plugin-permission-common';
+import { PermissionPolicy } from '@backstage/plugin-permission-node';
+
+export class CustomPermissionPolicy implements PermissionPolicy {
+  async handle(): Promise<PolicyDecision> {
+      // This acts as an allow all policy
+      // See https://backstage.io/docs/permissions/writing-a-policy for how to write your own custom policy
+    return { result: AuthorizeResult.ALLOW };
+  }
+}
```
Rugvip, line 13: I think it would be very useful to provide a basic test as well, maybe even both for this and the module. Something that shows a pattern for how to create additional tests.

packages/cli/templates/permission-module/package.json.hbs
```diff
@@ -0,0 +1,42 @@
...
+  "publishConfig": {
+{{#if npmRegistry}}
+    "registry": "{{npmRegistry}}",
+{{/if}}
+    "access": "public",
+    "main": "dist/index.cjs.js",
+    "types": "dist/index.d.ts"
+  },
+  "backstage": {
+    "role": "backend-plugin-module"
```
Rugvip, line 20: We could populate the `pluginPackage` here as well, to make sure that's correct upfront

Rugvip (commented):
Good idea! 👏 🎉 

Definitely looking forward to #26812, lots of unfortunate boilerplate here 😅

PR #27410, opened 2024-10-31
fix(core-components): content component can now be rendered as other element than article

roope-t:

Content component can now be passed an element string as a prop to render it as other element than article.

### DIFF
--- a/.changeset/thirty-needles-pay.md
+++ b/.changeset/thirty-needles-pay.md
@@ -1,0 +1,5 @@
+---
+'@backstage/core-components': patch
+---
+
+Added a way to render <Content /> component as other element instead of article.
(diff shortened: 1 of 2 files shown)


benjdlambert:
hey :wave: sorry this seemed to have fallen through the cracks a little bit.

What's the use case here? What would you want to change it to instead? 

Thinking a little more, I wonder if we want to follow the mui `as` prop instead of having `element`?

roope-t:
Hey, no worries.

Yeah, I can rename it to "as". The use case here is that we have a plugin that lists articles from a headless CMS, so I think it is semantically weird that we have an article inside an article.

vinzscam (approved):
nice

benjdlambert:
Apologies again that this went stale and got closed, been a busy few months with KubeCon and a lot of things happening around in the project.

I wanna unfortunately push back a bit on this, as there's a lot of things happening in this space.

It's most likely going to be solved with #27726 which is currently underway, so want to try a fold this into that instead of making changes to the existing `core-components` package.

PR #27414, opened 2024-10-31
Backstage Yarn Plugin Docs adjustments

awanlin:

This update the documentation around the Backstage Yarn plugin.

While adding this to the Demo site I ran into an error in the Docker build as the `backstage.json` file is needed. I've updated the related Dockerfile and Docker docs to include this.

Also, I feel that having this as the lead off to the "Keeping Backstage Updated" page dives right into a complex topic. It makes more sense for this to be a little latter.


awanlin:
@aramissennyeydd @Parsifal-M seems were weren't tagged on this, could it be because I'm a member?

aramissennyeydd:
@awanlin I don't think so, I get tagged for my OpenAPI tooling PRs on my personal account still. ex, https://github.com/backstage/backstage/pull/23610. I think it's because there's still an error on https://github.com/backstage/backstage/blob/master/.github/CODEOWNERS#L12, maybe we don't have the right perms still?

Rugvip:
Ah yep, `@backstage/documentation-maintainers` didn't have write access, fix'd

docs/deployment/docker.md
```diff
@@ -85,6 +85,7 @@ WORKDIR /app
 # Copy files needed by Yarn
 COPY --chown=node:node .yarn ./.yarn
 COPY --chown=node:node .yarnrc.yml ./
+COPY --chown=node:node backstage.json ./
```
Rugvip, line 88: @mtlewis do you think it would make sense and be possible to have the plugin gracefully handle a missing `backstage.json`? I'm thinking that it's just use whatever is in `yarn.lock` instead, all information you need would be there already right? Bumping the version is then more of a dev/CI-time thing where `backstage.json` is present, and the production build no longer requires it.

Rugvip (commented):
I'd like to explore https://github.com/backstage/backstage/pull/27414/files#r1824614263 or similar first tbh, it seems a bit unfortunate that this would be a production dependency.

awanlin:
> I'd like to explore https://github.com/backstage/backstage/pull/27414/files#r1824614263 or similar first tbh, it seems a bit unfortunate that this would be a production dependency.

Yup, I'm cool with that. Should I scale this back to just the "Keeping Backstage Updated" docs changes then? That way if this isn't done before the main release in November there's some tips about this.

Parsifal-M:
Hey 👋 

Interesting, we had the same issue but we use the [Minimal Hardened Image for Backstage](https://github.com/backstage/backstage/blob/master/contrib/docker/minimal-hardened-image/README.md) I was going to add something here as a note as well, I didn't think to check if it was also causing an issue in the other docker docs.

[Discord thread for some extra reference](https://discord.com/channels/687207715902193673/1300786038444392458)

mtlewis:
Will take a deeper look at this soon, but just to firstly point out that the reason the yarn plugin section comes first is to keep it ahead of the section about bumping with the backstage-cli. Otherwise, somebody following this doc as they work through an upgrade will likely do the upgrade with the cli, and _then_ hit the section about the plugin when it’s too late to use it. If we’re going to swap the order, I think it makes sense to telegraph the existence of the plugin a bit in the earlier section.

mtlewis (approved):
LGTM after the recent changes - from my perspective the pragmatic thing here is to keep all the required information explicitly in the container by including backstage.json. Also happy with the ordering adjustment given the new callout about the Backstage plugin.

Rugvip (approved):
:shipit:

Issue #27403, opened 2024-10-30
🐛 Bug Report: Getting Started seems broken at the moment


ericfrederich:
### 📜 Description

Following [getting started](https://backstage.io/docs/getting-started/) guide I am running into errors on Ubuntu 24.04 under WSL.

### 👍 Expected behavior

I shouldn't get an error running `npx @backstage/create-app@latest`

### 👎 Actual Behavior with Screenshots

Note: I re-ran the `npx @backstage/create-app@latest` command to get this screenshot because I had already closed the terminal.
![image](https://github.com/user-attachments/assets/190d599c-5765-4b7f-bccd-0a47ba4c1c59)

### 👟 Reproduction steps

Here's my complete history from a fresh install of Ubuntu-24.04
```
    1  sudo nano -w /etc/wsl.conf 
    2  echo "${LOGNAME} ALL=(ALL) NOPASSWD:ALL" | sudo tee /etc/sudoers.d/${LOGNAME}
    3  sudo apt update
    4  sudo apt upgrade -y
    5  sudo apt install -y build-essential docker.io
    6  curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.1/install.sh | bash
    7  echo "here in a new terminal after installing nvm"
    8  nvm install 22
    9  npm install --global yarn
   10  mkdir src && cd src
   11  npx @backstage/create-app@latest
   12  cat /tmp/xfs-8cf2eabe/build.log | sed "s#${LOGNAME}#ubuntu-user#g" > /mnt/c/Users/${LOGNAME}/Desktop/build.log
   13  history | sed "s#${LOGNAME}#ubuntu-user#g" > /mnt/c/Users/${LOGNAME}/Desktop/history
```

### 📃 Provide the context for the Bug.

I get an error with the following log.
[build.log](https://github.com/user-attachments/files/17577664/build.log)

Here's what seems to be the actual compile-time error.
```
../src/isolate/class_handle.h: In lambda function:
../src/isolate/class_handle.h:324:39: error: there are no arguments to ‘RuntimeTypeError’ that depend on a template parameter, so a declaration of ‘RuntimeTypeError’ must be available [-fpermissive]
  324 |                                 throw RuntimeTypeError(detail::CalleeName(info)+ " must be called with `new`");
      |                                       ^~~~~~~~~~~~~~~~
```

### 🖥️ Your Environment

This was done using a fresh install of Ubuntu 24.04 on WSL2.

[...]

PR #27415, opened 2024-10-31
Updated Getting Started Node link

awanlin:

Updated the Getting Started Node JS Active LTS link to point to the Node JS version policy as it has some helpful clarifications. Also added a small sentence trying to highlight the timing of support for when a new Node version has become active LTS in Backstage.

### DIFF
--- a/docs/overview/versioning-policy.md
+++ b/docs/overview/versioning-policy.md
@@ -166,10 +166,7 @@ following schedule for determining the [Node.js releases](https://nodejs.org/en/
 
 - At any given point in time we support exactly two adjacent even-numbered
   releases of Node.js, for example v12 and v14.
-- Once a new Node.js release becomes _Active LTS_ we switch to support that
-  release and the previous one. The switch is not immediate but done as soon
-  as possible. You can find the Node.js version supported by each release
-  in the `engines` field in the root `package.json` of a new app.
+- Once a new Node.js release becomes _Active LTS_ we switch to support that release and the previous one. The switch is not immediate but done as soon as possible. This support will come in the first Backstage main release line release after a Node.js release becomes _Active LTS_. You can find the Node.js version supported by each release in the `engines` field in the root `package.json` of a new app.
 
 When we say _Supporting_ a Node.js release, that means the following:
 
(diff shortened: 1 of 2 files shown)


awanlin:
This might be one way to help improve the issue reported here: https://github.com/backstage/backstage/issues/27403

docs/overview/versioning-policy.md
```diff
@@ -166,10 +166,7 @@ following schedule for determining the [Node.js releases](https://nodejs.org/en/
 
 - At any given point in time we support exactly two adjacent even-numbered
   releases of Node.js, for example v12 and v14.
-- Once a new Node.js release becomes _Active LTS_ we switch to support that
-  release and the previous one. The switch is not immediate but done as soon
-  as possible. You can find the Node.js version supported by each release
-  in the `engines` field in the root `package.json` of a new app.
+- Once a new Node.js release becomes _Active LTS_ we switch to support that release and the previous one. The switch is not immediate but done as soon as possible. This support will come in the first Backstage main release line release after a Node.js release becomes _Active LTS_. You can find the Node.js version supported by each release in the `engines` field in the root `package.json` of a new app.
```
Rugvip, line 169: > This support will come in the first Backstage main release line release after a Node.js release becomes _Active LTS_.

This isn't necessarily true since it depends on the timing of things. If there are unresolved blocking issues or the releases fall very close to each other we make exceptions. I'd prefer to keep this as a best effort with an intended target.

I think a fairly big improvement could actually be to simply list the current supported versions here? Now that we have versioned docs for stable/next releases we can make that update properly.

awanlin: Ok, cool. I can do that, in the past we've tried to avoid it as it means another place we need to make sure to update but you are right, now that things are versioned it's a bit easier.

TimKnight-DWP: Yeah that sounds great @awanlin and @Rugvip

thenets: That's great. Listing the supported versions and linking this page in the Getting Started page would be perfect.

Rugvip (approved):
Thanks! 👍

Issue #17063, opened 2023-03-24
🐛 Bug Report: New relation type for same source entity and target entity pair not reflected


alecjacobs5401:
### 📜 Description

Defining two unique relationships between two entities results in one of the Entities not having the relation reflected in the `final_entity` post processing.

Given the following scenario:
A `Group` is defined as follows and is ingested into the Software catalog
```yaml
apiVersion: backstage.io/v1alpha1
kind: Group
metadata:
  name: test-group
spec:
  type: team
  children: []
```

Some time later, a new `User` entity is defined that has a [custom defined relation type](https://backstage.io/docs/features/software-catalog/extending-the-model#adding-a-new-relation-type) of `company-riskOwner`
```yaml
apiVersion: backstage.io/v1alpha1
kind: User
metadata:
  name: test-user
spec:
  company-riskOwner: group:default/test-group
```

After `user:default/test-user` is ingested in the Catalog, we have two Entities that have proper relations emitted and defined in the respective final entities: `group:default/test-group` has a  relation `type` of `company-riskOwnedBy` with a target of `user:default/test-user` and  `user:default/test-user` has a relation `type` of `company-riskOwnerOf` with a target of `group:default/test-group`

Now, some time later, the `group:default/test-group` Entity definition is updated to include a `members` definition as follows:
```yaml
apiVersion: backstage.io/v1alpha1
kind: Group
metadata:
  name: test-group
spec:
  type: team
  members: [test-user]
  children: []
```

After the processing pipeline finishes, `group:default/test-group` now has **two** relations defined:
1. relation `type` of `company-riskOwnedBy` with a target of `user:default/test-user`
2. relation `type` of `hasMember` with a target of `user:default/test-user`

_however_, `user:default/test-user` still only has **one** relation defined in the finaly entity: `user:default/test-user` has a relation `type` of `company-riskOwnerOf` with a target of `group:default/test-group`. 

The `memberOf` relation is missing.

### 👍 Expected behavior

After processing finishes, the `user:default/test-user` Entity should also have two relations defined in its `final_entity`:
1. `user:default/test-user` has a relation `type` of `company-riskOwnerOf` with a target of `group:default/test-group`
2. `user:default/test-user` has a relation `type` of `memberOf` with a target of `group:default/test-group`

### 👎 Actual Behavior with Screenshots

The `memberOf` relation type is missing on the User entity

### 👟 Reproduction steps

See steps above

### 📃 Provide the context for the Bug.

[...]

PR #27417, opened 2024-10-31
Fixes #27325 Stitch entity for which target of relationship has changed

### LINKED ISSUE #27325: 🐛 Bug Report: Relationships not properly updated in Entities generated from custom provider/processor
### 📜 Description

When modifying relationships via a `delta` mutation, the final entity has incorrect relationships. In this specific test case (which mimics our real use case), we have the following:

1. A provider that emits a "raw" (in this case, `rawComponent`) entity initially via a `full` mutation, but over the course of the day via `delta` mutations.
2. A processor converts that raw entity into a final component entity, forming custom relationships.
3. The relationships are not properly updated in the final entity (in the test case below, only one side of the relationship is updated).

### 👍 Expected behavior

Relationships should be properly updated as a result of `delta` mutations, for both sides of the relationship.

In the test code given below, a "test" provider and processor emit entities. It simulates both a `full` mutation and a `delta` mutation as follows:

1. During a simulated `full` mutation, emits a single group entity (as owner) and three components: `a`, `b`, and `c`. The provider emits `rawComponent` entities that are converted into `component` entities by the processor. As a result of the simulated `full` mutation, `a` has a `testDependsOn` and `testDependencyOf` relationship with `b`.

2. During a simulated `delta` mutation, the providers emits new `rawComponent` entities that cause the processor to switch the relationship from between `a` and `b` to between `a` and `c`. As a result only one side of the relationship is updated, `a` has a `testDependsOn` with `b` still.

Interestingly enough, the problem persists even if in Step 2 a `full` mutation is done to switch the relationships. Also interestingly, if the processor-emitted component has `spec.dependsOn` (to generate the built-in dependsOn relationship), the bug no longer happens!

### 👟 Reproduction steps

I've created a `test` provider in a fork of the Backstage repo, you can see the code here: https://github.com/backstage/backstage/compare/master...jslott2sigma:backstage:relationships-not-being-updated

The test catalog module exposes two endpoints: `/full` to populate the entities initially and `delta` to switch the relationship from between `a` and `b` to between `a` and `c`.

Simply run `yarn dev` to launch the sample app. I am happy to investigate this further if pointed in the right direction and submit a fix.

_No response_

### 🖥️ Your Environment

_No response_

jslott2sigma:

Fixes #27325. Fix a bug where if the target of the relationship changes, the entity now be restitched.

### DIFF
--- a/.changeset/short-pots-remember.md
+++ b/.changeset/short-pots-remember.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Fixed an issue where entities would not be marked for restitching if only the target of a relationship changed.
--- a/plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.test.ts
+++ b/plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.test.ts
@@ -464,6 +464,111 @@ describe('DefaultCatalogProcessingEngine', () => {
     await engine.stop();
   });
 
+  it('should stitch both the previous and new sources when relation target changes', async () => {
+    const engine = new DefaultCatalogProcessingEngine({
+      config: new ConfigReader({}),
+      logger: mockServices.logger.mock(),
+      processingDatabase: db,
+      knex: {} as any,
+      orchestrator: orchestrator,
+      stitcher: stitcher,
+      createHash: () => hash,
+      pollingIntervalMs: 100,
+    });
+
+    db.transaction.mockImplementation(cb => cb((() => {}) as any));
+
+    const entity = {
+      apiVersion: '1',
+      kind: 'k',
+      metadata: { name: 'me', namespace: 'ns' },
+    };
+    const processableEntity = {
+      entityRef: 'foo',
+      id: '1',
+      unprocessedEntity: entity,
+      resultHash: '',
+      state: [] as any,
+      nextUpdateAt: DateTime.now(),
+      lastDiscoveryAt: DateTime.now(),
+    };
+
+    db.listParents.mockResolvedValue({ entityRefs: [] });
+    db.getProcessableEntities
+      .mockResolvedValueOnce({
+        items: [processableEntity],
+      })
+      .mockResolvedValueOnce({
+        items: [processableEntity],
+      });
+    db.updateProcessedEntity
+      .mockImplementationOnce(async () => ({
+        previous: { relations: [] },
+      }))
+      .mockImplementationOnce(async () => ({
+        previous: {
+          relations: [
+            {
+              originating_entity_id: '',
+              type: 't',
+              source_entity_ref: 'k:ns/other1',
+              target_entity_ref: 'k:ns/me',
+            },
+          ],
+        },
+      }));
+
+    orchestrator.process
+      .mockResolvedValueOnce({
+        ok: true,
+        completedEntity: entity,
+        relations: [
+          {
+            type: 't',
+            source: { kind: 'k', namespace: 'ns', name: 'other1' },
+            target: { kind: 'k', namespace: 'ns', name: 'me' },
+          },
+        ],
+        errors: [],
+        deferredEntities: [],
+        state: {},
+        refreshKeys: [],
+      })
+      .mockResolvedValueOnce({
+        ok: true,
+        completedEntity: entity,
+        // change just the target of the relationship to a new entity,
+        // leaving the source and relation type the same.
+        // see: https://github.com/backstage/backstage/issues/27325
+        relations: [
+          {
+            type: 't',
+            source: { kind: 'k', namespace: 'ns', name: 'other1' },
+            target: { kind: 'k', namespace: 'ns', name: 'newtarget' },
+          },
+        ],
+        errors: [],
+        deferredEntities: [],
+        state: {},
+        refreshKeys: [],
+      });
+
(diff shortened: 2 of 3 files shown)


.changeset/short-pots-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Stitch entities for which the target of a relationship has changed
```
aramissennyeydd, line 5:
```suggestion
Fixed an issue where entities would not be marked for restitching if only the target of a relationship changed.
```

plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.test.ts
```diff
@@ -464,6 +464,106 @@ describe('DefaultCatalogProcessingEngine', () => {
...
+        previous: { relations: [] },
+      }))
+      .mockImplementationOnce(async () => ({
+        previous: {
+          relations: [
+            {
+              originating_entity_id: '',
+              type: 't',
+              source_entity_ref: 'k:ns/other1',
+              target_entity_ref: 'k:ns/me',
```
aramissennyeydd, line 516: Can we adjust the entity names here to make it a little clearer what a failure on this test case would mean? I added this test locally to master to see if it would fail (it did), but the assertion is a little difficult to understand what's wrong at a glance.

plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
```diff
@@ -295,7 +295,7 @@ export class DefaultCatalogProcessingEngine {
                 });
               oldRelationSources = new Map(
                 previous.relations.map(r => [
-                  `${r.source_entity_ref}:${r.type}`,
+                  `${r.source_entity_ref}:${r.type}:${r.target_entity_ref}`,
```
aramissennyeydd, line 298: nit: do we want to have something like
```suggestion
                  `${r.type}:${r.source_entity_ref}->${r.target_entity_ref}`,
```
here? that's a bit more clear wrt directionality and type

aramissennyeydd (approved):
Thanks for this! 🎉 

Can you move the `fixes #27325` into the PR description? GH uses that to automatically tag the PR to the issue.

.changeset/short-pots-remember.md
jslott2sigma: @aramissennyeydd I believe "Fixes #27325" is already in the PR description and the commit description. Perhaps did you mean the `changeset` md file?

plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
jslott2sigma: I like it!

plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.test.ts
jslott2sigma: I renamed the new target to `newtarget` but I also added comments to the tests which I think is a good use of comments to explain the purpose of the test. Please let me know if you have other suggestions, thanks!

.changeset/short-pots-remember.md
aramissennyeydd: I only see it in the title on my end (and commit description), I think it's fine to leave as is.

freben: Nope, he means at the top of the page here, editing the description text (that now has like "Hey, I just made a Pull Request" to have `Fixes #27325` on a line of its own, because that means when the PR gets merged, that issue gets auto-closed as well (and linked to the PR)

plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
```diff
@@ -295,7 +295,7 @@ export class DefaultCatalogProcessingEngine {
                 });
               oldRelationSources = new Map(
                 previous.relations.map(r => [
-                  `${r.source_entity_ref}:${r.type}`,
+                  `${r.source_entity_ref}:${r.type}->${r.target_entity_ref}`,
```
freben, line 298: You know what - at this point, the Map with the unique keys really doesn't have a purpose anymore. The source-type-target triplet is the _entire relation_ basically. So I'd suggest using a map of source refs instead!

```ts
let oldRelationSources: Set<string>();
// ...
oldRelationSources = new Set(previous.relations.map(r => r.source_entity_ref));
}
```

and then the same thing with the new sources, making that a set too.

Then the refs that need stitching is `oldRelationSources.symmetricDifference(newRelationSources)`.

jslott2sigma: @freben Thanks so much for the comments! However, I'm confused. The way the code worked before #17190 seems to be what you describe (that PR fixed issue #17063). If the type of the relation changed between old and new, then I believe the source would not appear in the symmetric difference between old and new and not get stitched according to your suggested code change above.

Am I misunderstanding something? Thanks!

freben: Ahh. I think you're quite right, I over simplified things. I'll make another review as soon as I get the chance

PR #27419, opened 2024-11-01
DevTools - Removed Deprecated Code

awanlin:

Removed support for what is known as the legacy backend, please use the New Backend System.


awanlin:
I'm not sure how to resolve the CI error:

```text
Incorrect type dependencies in @backstage/plugin-devtools-backend:
  Move from dependencies to devDependencies: @types/express

At least one package had incorrect type dependencies
```

When I run `yarn lint:type-deps` locally I get no results 🤔

freben:
@ahhhndre That check works on the dist output. If you run a `yarn tsc:full` and then `yarn backstage-cli repo build --all`, you'll see that the command starts working and gives the same output.

You really just need to move that dependency in `plugins/devtools-backend/package.json` to be a dev dependency instead. It used to be a regular dependency only because its types were exposed publicly. Now that they aren't (as you can see in the api report diff), it should be moved to a dev dependency instead since you currently only need its types.

freben (approved):

PR #27430, opened 2024-11-01
feat: add utility types for auth providers

schultzp2020:


plugins/auth-backend-module-oidc-provider/report.api.md
```diff
@@ -37,15 +38,8 @@ export type OidcAuthResult = {
...
 export namespace oidcSignInResolvers {
-  const emailLocalPartMatchingUserEntityName: SignInResolverFactory<
-    unknown,
-    | {
-        allowedDomains?: string[] | undefined;
-      }
-    | undefined
-  >;
-  const emailMatchingUserEntityProfileEmail: SignInResolverFactory<
-    unknown,
```
schultzp2020, line 48: These are no longer needed as `module.ts` already includes `commonSignInResolvers`.

Rugvip (changes requested):
Thank you, but I think we should avoid this. The [OpenID spec is very explicit about the use of `preferred_username`](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims):

> The RP MUST NOT rely upon this value being unique

plugins/auth-backend-module-oidc-provider/report.api.md
Rugvip: Indeed! 👍 perhaps you could update the PR to only include this change?

JessicaJHee:
Hey @Rugvip! 

I read over the docs and it mentions in [section 5.7](https://openid.net/specs/openid-connect-core-1_0.html#ClaimStability) that "the sub (subject) and iss (issuer) Claims from the ID Token, used together, are the only Claims that an RP can rely upon as a stable identifier for the End-User". i.e. any other identifiers wouldn't be reliable.

By this logic, the current [commonSignInResolvers](https://github.com/backstage/backstage/blob/1d478d27f08bd6fa8681ba3b454fa9bcfc172e4c/plugins/auth-node/src/sign-in/commonSignInResolvers.ts) that uses email to resolve users would also be considered unreliable. Imo in this case, adding this optional resolver which has similar downsides to the existing resolvers should be okay.

What do you think?

Rugvip:
@JessicaJHee yep you're right, it's not quite as simple as following the spec completely here. What we have on our end in the end is a reflection of the current state in the catalog, that's what makes something like `email` reasonably safe to use since even though you might have different users with the same email over time, that'll be reflected in the catalog and in the end mapped to the correct user identities.

That said I'm just a bit uncertain at this point about the intention of the `preferred_username` claim. It sounds like something that could potentially be up to the end user to pick by themselves, but also in some systems essentially just be the plain username. It's a bit of a strange judgement call where the spec definitely has a strong opinion, but it's ultimately comes down to what implementations actually do. For the `email` claim I think it's safe to expect that users can not provide arbitrary emails, but is the same also true for `preferred_username`? on idea tbh 😐

schultzp2020:
@Rugvip I've updated the PR.

PR #27436, opened 2024-11-02
Cleaning up event subscriptions after the max age window

acierto:

During the test of the issue with WebSockets (when some of WS messages are lost) on the deployment with multiple nodes I found that the logic which has to distribute events across the nodes doesn't work always correctly.

Namely, if you'll run your cluster when in your database is yet no events and no subscriptions, WS sends the request to store the subscription, and it stored. But after few seconds cleanup job will remove it, because `event_bus_events` is empty yet. 
So this logic will work only if *subscription* and *event(s)* will be saved in the database within few seconds.

Though instead it should wait with cleaning up as the comment properly says:

        // No events left, remove all subscribers. This can happen if no events
        // are published within the max age window.

This PR just does what it says.

### DIFF
--- a/.changeset/silly-kiwis-destroy.md
+++ b/.changeset/silly-kiwis-destroy.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-events-backend': patch
+---
+
+Cleaning up event subscriptions after the max age window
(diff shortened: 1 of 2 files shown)


Rugvip (approved):
Nice, thank you! :tada:

Issue #26990, opened 2024-10-07
🐛 Bug Report: Owner name is overflowing in AboutContent Component of Catalog Plugin 


its-mitesh-kumar:
### 📜 Description

In the AboutContent component of the Catalog plugin, longer owner names is causing layout issues due to text overflow. To maintain a consistent and user-friendly design, we need to truncate owner names that exceed a specific length and display the full name upon hover. 

### 👍 Expected behavior

Instead of displaying the full owner name, names longer than a specified length should be truncated and the full name shown on hover.

### 👎 Actual Behavior with Screenshots

We can see owner name text is overflowing .
<img width="1710" alt="Screenshot 2024-10-04 at 3 04 07 PM" src="https://github.com/user-attachments/assets/c62a05df-4286-4f42-82ac-b4fded532158">

### 👟 Reproduction steps

Create Catalog Plugin having longer owner name 

### 🖥️ Your Environment

Not environment specific .

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

benjdlambert:
Ah yep, this looks like it should have some additional properties.

Perhaps `text-overflow: ellipsis` and `overflow: hidden`. Do you want to raise a PR with a suggested fix? :pray:

its-mitesh-kumar:
Yes I am working on it , I raise a PR soon .

its-mitesh-kumar:
PR : https://github.com/backstage/backstage/pull/27044

PR #27442, opened 2024-11-04
Catalog react owner text overflow

its-mitesh-kumar:
PR for the issue :[26990](https://github.com/backstage/backstage/issues/26990)

With this fix the text will not overflow as we can see in the below screenshot .

Screenshot 2024-11-04 at 2 33 21 PM

### DIFF
--- a/plugins/catalog-react/src/components/EntityDisplayName/EntityDisplayName.tsx
+++ b/plugins/catalog-react/src/components/EntityDisplayName/EntityDisplayName.tsx
@@ -77,7 +77,16 @@ export const EntityDisplayName = (
   );
 
   // The innermost "body" content
-  let content = <>{primaryTitle}</>;
+  let content = (
+    <div
+      style={{
+        overflow: 'hidden',
+        wordBreak: 'break-word',
+      }}
+    >
+      {primaryTitle}
+    </div>
+  );
 
   // Optionally an icon, and wrapper around them both
   content = (
--- a/.changeset/healthy-planets-confess.md
+++ b/.changeset/healthy-planets-confess.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-react': patch
+---
+
+Fixed bug in `EntityDisplayName` where text was overflowing.


plugins/catalog-react/src/components/EntityDisplayName/EntityDisplayName.tsx
```diff
@@ -77,7 +77,16 @@ export const EntityDisplayName = (
   );
 
   // The innermost "body" content
-  let content = <>{primaryTitle}</>;
+  let content = (
+    <div
+      style={{
+        overflow: 'hidden',
+        wordBreak: 'break-word',
+      }}
```
freben, line 85: This should go into the `makeStyles` above, under a new key like `content` or so.

its-mitesh-kumar: @freben moved these styles under *makeStyles*

its-mitesh-kumar:
@freben @vinzscam comment were addressed last week only , please re-review it .

plugins/catalog-react/src/components/EntityDisplayName/EntityDisplayName.tsx
```diff
@@ -77,7 +81,7 @@ export const EntityDisplayName = (
   );
 
   // The innermost "body" content
-  let content = <>{primaryTitle}</>;
+  let content = <div className={classes.breakTextOverflow}>{primaryTitle}</div>;
```
freben, line 84: I think we have a problem actually. This feature requires a CSS block context, while this component requires to function in inline contexts. Thinking more about this, I think maybe EntityDisplayName itself cannot support this feature at all, but rather, the surrounding div that the display name is rendered in, needs to be the one with the overflow handling. Does that make sense?

its-mitesh-kumar: @freben I have made the changes requested by you .

freben (approved):

PR #27495, opened 2024-11-05
Made it possible to construct `mockServices.database` with a given knex instance

freben:
This enables

```ts
mockServices.database({ knex });
// and
mockServices.database.factory({ knex });
```

so you don't have to do

```ts
{
  getClient: async () => knex,
}
// and
createServiceFactory({
  service: coreServices.database,
  deps: {},
  factory: () => ({ getClient: async () => knex }),
})
```

This goes very well together with doing a `TestDatabases` creation right before it.

### DIFF
--- a/.changeset/little-rivers-deny.md
+++ b/.changeset/little-rivers-deny.md
@@ -1,0 +1,5 @@
+---
+'@backstage/backend-test-utils': minor
+---
+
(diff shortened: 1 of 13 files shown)


Rugvip (approved):
Yep, nice! 👍

PR #14568, opened 2022-11-10
[PRFC] Surface ability for Kubernetes Plugin to integrate with the Permission Framework

RubenV-dev:
## Hey, I just made a Pull Request!

I am submitting this PRFC to open a discussion on integrating the kubernetes plugin with the permission framework. The suggested changes by no means covers the multiple scenarios that these plugin endpoints can handle but it does set a foundation for future contributions to build upon.

## Need
The Kubernetes Plugin in its current state does not integrate with the permission framework so anyone with a properly set up kubernetes authprovider or even service account token in their config can access resources via the following endpoints:
* api/kubernetes/resources/workloads/query
* api/kubernetes/resources/custom/query
* api/kubernetes/clusters

## Proposal
I propose we leverage the permission framework in the Kubernetes plugin which allows for finer-grained control over access to kubernetes resources via definitive and conditional policy decisions returned by a permission policy.

We are effectively allowing backstage users the ability to restrict the kubernetes endpoints so that it is safer to configure kubernetes clusters.

This is being done by:
* Introducing a PermissionEvaluator in the KubernetesEnvironment type which is consumed by the kubernetesBuilder.
* Introducing permission api calls  in the endpoint handlers for cluster, workloads and custom resources.
* Introducing Kubernetes Condition exports and rules that can be used by our permission policies to control what is shown to the user.

This would allow users to setup sample policies [here](https://github.com/backstage/backstage/blob/master/packages/backend/src/plugins/permission.ts) like:

Basic Permission for Cluster - Definitive policy decision
```typescript
class DenyAllKubernetesClusterRead implements PermissionPolicy {
  async handle(request: PolicyQuery): Promise<PolicyDecision> {
    if (isPermission(request.permission, kubernetesClusterReadPermission)) {
      return { result: AuthorizeResult.DENY };
    }
    return { result: AuthorizeResult.ALLOW };
  }
}
```
Which allows the /api/kubernetes/clusters endpoint to return a 403 response if receiving a definitive DENY response from the permission api endpoint.

ResourcePermission for Cluster Resources - Definitive policy decision
```typescript
class K8sPermissionPolicy implements PermissionPolicy {
  async handle(request: PolicyQuery): Promise<PolicyDecision> {
    if (
      isPermission(
        request.permission,
        kubernetesCustomResourcesReadPermission,
      ) ||
      isPermission(
        request.permission,
        kubernetesWorkloadResourcesReadPermission,
      )
    ) {
      return {
        result: AuthorizeResult.DENY,
      };
[...]

PR #27499, opened 2024-11-05
feat(kubernetes): introduce kubernetes permission

### LINKED ISSUE #27620: 🚀 Feature: Add permissions to Kubernetes plugins
### 🔖 Feature description

Restrict access to other Kubernetes endpoints besides the proxy endpoint via Permissions Framework.

### 🎤 Context

We would like to use permissions with Kubernetes plugins.

New permission like `kubernetes.resource` permission could be added.

dzemanov:

Introduces a simple way to restrict access to the other kubernetes endpoints besides the proxy endpoint using Persmission Framework.

The `/clusters` endpoint is now protected by the `kubernetes.clusters.read` permission.
The `/services/:serviceId` endpoint is now protected by the `kubernetes.resources.read` permission.
The `/resources` endpoints are now protected by the `kubernetes.resources.read` permission.

### Changes
Users can create permission policy like this:
```
class KubernetesDenyAllResourcePermissionPolicy implements PermissionPolicy {
  async handle(request: PolicyQuery): Promise<PolicyDecision> {
    if (request.permission.name === 'kubernetes.clusters.read') {
      return {
        result: AuthorizeResult.DENY,
      };
    }
    return { result: AuthorizeResult.ALLOW };
  }
}
```

The endpoints `/resources/workloads/query`, `/resources/custom/query`, `/services/:serviceId` and `/clusters` return  403 Forbidden even with correct credentials.

### Setup
- Enable permission framework as shown [here](https://backstage.io/docs/permissions/getting-started)
- Deploy a service to cluster and add it to backstage locations
- Setup `kubernetes-backend` as shown [here](https://backstage.io/docs/features/kubernetes/installation) and [here](https://backstage.io/docs/features/kubernetes/configuration) 
- Try out the endpoints with KubernetesDenyAllResourcePermissionPolicy policy (You can add custom permission as shown [here](https://backstage.io/docs/permissions/getting-started)) 

### Fixes
#27620

### UI changes
kubernetes-permission
kubernetes-cluster-permission


dzemanov:
WDYT @freben @Rugvip @vinzscam about adding a new permission to kubernetes for remaining endpoints? I can see there was already a fine-grained approach to solve the need for permissions [here](https://github.com/backstage/backstage/pull/14568).

plugins/kubernetes-backend/src/auth/requirePermission.ts
```diff
@@ -0,0 +1,49 @@
...
+      ],
+      {
+        credentials: await httpAuth.credentials(req),
+      },
+    )
+  )[0];
+
+  if (decision.result === AuthorizeResult.DENY) {
+    const err = new NotAllowedError('Unauthorized');
+    throw err;
```
vinzscam, line 47: this is safer in case we switch to a resource permission in the future, which can return `CONDITIONAL`
```suggestion
  if (decision.result === AuthorizeResult.ALLOW) {
    return;
  }
  throw new NotAllowedError('Unauthorized');
```

vinzscam (changes requested):
Hi @dzemanov, this looks like a great idea!
The only recommendation I have is to think about the possible use cases so that we can create permissions that match real scenarios. For example, right now the `kubernetesResourcePermission` is used for authorizing the `/resources/*`, `/clusters` and `/services` endpoints.

From a user perspective, would it make sense to have different permissions for those endpoints? Or would it never happen that adopters would allow only a subset of these endpoints?

dzemanov:
@vinzscam thanks for the review!
Yeah, it makes sense to add separate permissions that can be used if needed.

plugins/kubernetes-backend/src/service/KubernetesBuilder.test.ts
```diff
@@ -779,6 +795,9 @@ metadata:
     expect(response.body).toMatchObject({
       permissions: [
         { type: 'basic', name: 'kubernetes.proxy', attributes: {} },
+        { type: 'basic', name: 'kubernetes.resource', attributes: {} },
+        { type: 'basic', name: 'kubernetes.service', attributes: {} },
```
mclarke47, line 799: should these be plural?

dzemanov: Sure, I can rename the permissions to plural if it makes more sense.

dzemanov: But these are singular, because permission names are:
`kubernetes.service` permission and `kubernetes.resource` permission.

dzemanov: @mclarke47 updated.

plugins/kubernetes-backend/src/auth/requirePermission.ts
```diff
@@ -0,0 +1,52 @@
...
+      [
+        {
+          permission: permissionRequired,
+        },
+      ],
+      {
+        credentials: await httpAuth.credentials(req),
+      },
+    )
+  )[0];
```
mclarke47, line 46: Will this ever be empty?

dzemanov: I would say no, the description of the [method](https://github.com/backstage/backstage/blob/master/packages/backend-plugin-api/src/services/definitions/PermissionsService.ts#L54-#L55) says `The returned array has the same number of items, in the same order, as the given requests.` It is used this way in a couple of other places in Backstage, e.g [1](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/service/AuthorizedRefreshService.ts#L30-#L40), or [2](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/service/AuthorizedLocationAnalyzer.ts#L38-#L47).

dzemanov:
@vinzscam I divided the endpoint permissions, however, I think it would make more sense to add [/services/{service-id}](https://github.com/backstage/backstage/blob/master/plugins/kubernetes-backend/src/service/KubernetesBuilder.ts#L394-#L412) endpoint under `kubernetes.resources` permission, since that is the only one endpoint under `services`, which is deprecated and is doing the same as [/resources/workloads/query](https://github.com/backstage/backstage/blob/master/plugins/kubernetes-backend/src/routes/resourcesRoutes.ts#L64-#L74)

dzemanov:
Not stale.
@mclarke47 @vinzscam PTAL?

Rugvip (commented):
Thank you! 🎉 👍 

As discussed during the framework SIG, please add the frontend implementation as well

plugins/kubernetes-common/src/permissions.ts
```diff
@@ -24,8 +24,28 @@ export const kubernetesProxyPermission = createPermission({
...
+  attributes: {},
+});
+
+/** This permission is used to check access to the /clusters endpoint
+ * @public
+ */
+export const kubernetesClustersPermission = createPermission({
+  name: 'kubernetes.clusters',
+  attributes: {},
+});
```
vinzscam, line 41: perhaps we can be more explicit:

```suggestion
export const kubernetesResourcesReadPermission = createPermission({
  name: 'kubernetes.resources.read',
  attributes: {
    action: 'read',
  },
});

/** This permission is used to check access to the /clusters endpoint
 * @public
 */
export const kubernetesClustersReadPermission = createPermission({
  name: 'kubernetes.clusters.read',
  attributes: {
    action: 'read',
  },
});
```

plugins/kubernetes-cluster/src/components/KubernetesClusterContent/KubernetesClusterContent.tsx
```diff
@@ -57,7 +59,15 @@ const ContentGrid = () => {
  * @public
  */
 export const KubernetesClusterContent = () => {
-  return (
+  const hasKubernetesClustersPermission = useKubernetesClustersPermission();
+
+  return !hasKubernetesClustersPermission ? (
```
vinzscam, line 64: mmm I think this will cause a glitch where `WarningPanel` will be visible while the permission is loading

vinzscam (changes requested):
awesome stuff! I've just found two small improvements

dzemanov:
@vinzscam PTAL at the changes when you will have time.

vinzscam (approved):
amazing work @dzemanov 🙏

dzemanov:
This will need to also be approved by @backstage/kubernetes-maintainers. @mclarke47 do you think you can take a look?

dzemanov:
@vinzscam do you think this is ok to be merged?

mclarke47:
Should we add any docs?

dzemanov:
@mclarke47 I updated the documentation.

dzemanov:
@Rugvip @freben is this one alright to go?

.changeset/real-foxes-film.md
```diff
@@ -0,0 +1,10 @@
+---
+'@backstage/plugin-kubernetes-backend': minor
+'@backstage/plugin-kubernetes-cluster': minor
+'@backstage/plugin-kubernetes-common': minor
+'@backstage/plugin-kubernetes': minor
+---
+
+**BREAKING** The `/clusters` endpoint is now protected by the `kubernetes.clusters.read` permission.
+**BREAKING** The `/services/:serviceId` endpoint is now protected by the `kubernetes.resources.read` permission.
+**BREAKING** The `/resources` endpoints are now protected by the `kubernetes.resources.read` permission.
```
benjdlambert, line 10:
```suggestion
---
'@backstage/plugin-kubernetes-backend': patch
'@backstage/plugin-kubernetes-cluster': patch
'@backstage/plugin-kubernetes-common': patch
'@backstage/plugin-kubernetes': patch
---

The `/clusters` endpoint is now protected by the `kubernetes.clusters.read` permission.
The `/services/:serviceId` endpoint is now protected by the `kubernetes.resources.read` permission.
The `/resources` endpoints are now protected by the `kubernetes.resources.read` permission.
```

benjdlambert (changes requested):
I'm thinking that we ship this as a non-breaking change. We recommend that when people write their own policies that the default catch all decision is to `ALLOW` otherwise I'd say that it could be a breaking change. 

https://backstage.io/docs/permissions/writing-a-policy

Parsifal-M:
Hey 👋 

Was just curious, we don't (currently) use the K8s plugin so maybe its more clear for people that use it, but are we set on the message logged if a user does not have permission to access? If I applied it to our setup, they would not know who to contact if we said "administrators" in fact they'd probably contact the team that delivers the clusters and not the team who owns Backstage 😅. 

Maybe for a later iteration, but would be nice if you could either be able to customise that message or make it more generic (You don't have permission to view this)

Sorry, was trying to hold off on sticking my nose in at the finish line but was just curious about this 🙏

dzemanov:
> Hey 👋
> 
> Was just curious, we don't (currently) use the K8s plugin so maybe its more clear for people that use it, but are we set on the message logged if a user does not have permission to access? If I applied it to our setup, they would not know who to contact if we said "Backstage administrator" I have a feeling sometimes its not clear to the end user what backstage even is 😆 its just what ever we call it (cockpit) in our case.
> 
> Maybe for a later iteration, but would be nice if you could either be able to customise that message or make it more generic (You don't have permission to view this)
> 
> Sorry, was trying to hold off on sticking my nose in at the finish line but was just curious about this 🙏

You are right. Using Backstage in the message can also cause confusion. In that case, I would suggest going with the `contact your portal administrator to give you...`, if @mclarke47 agrees. We have in RHDH configuration value:
```
flavor: rhdh
```
but I haven't noticed something like that in Backstage. Maybe it would be good to add, WDYT @Rugvip? Though not sure if that should be part of this PR.

mclarke47:
@dzemanov I think we'd likely want to use i18n for this

benjdlambert:
@dzemanov yep I think that @mclarke47 is right, we'd want to move the kubernetes plugin over to using i18n and then have these messages configurable by providing your own translations if needed.

I'm thinking of merging this as is though for now, and we can take that in a followup as it's not a simple fix and I think that the wording is OK as is.

benjdlambert (approved):

dzemanov:
@benjdlambert  Oh, pardon, I have updated the changes to use internalization. I can revert but I think these are alright.

benjdlambert:
Oh nevermind - great! Will have a look through just after lunch!

Rugvip (approved):
lgtm 👍

benjdlambert (approved):
Nice! Thanks @dzemanov and all involved :tada:

mclarke47:
Ty @dzemanov !

PR #27544, opened 2024-11-07
Use core Select instead of Checkbox for filters on Search page

jroebu14:

On the Search page, use Select from core-components and update Lifecycle filter to use Select instead checkboxes.

### Before

https://github.com/user-attachments/assets/ff67dbc5-ad8c-434e-962b-22c7ad00963f

### After

https://github.com/user-attachments/assets/c26eb081-496f-4adf-bdab-60158e79a880

### DIFF
--- a/.changeset/healthy-years-begin.md
+++ b/.changeset/healthy-years-begin.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-search-react': patch
+---
+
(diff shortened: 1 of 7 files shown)


vinzscam:
this will mean an extra click 🤔

jroebu14:
> this will mean an extra click 🤔

We are also providing the all option there. So its more of a stylistic preference rather than UX and the UI is a bit more consistent.

Issue #26554, opened 2024-09-09
🐛 Bug Report: Backstage Catalog – Groups – backstage is able to identify only child groups of gitlab, parent groups it’s not identifying


vinaykv23:
### 📜 Description

•	Backstage  - "version": "1.28.0" 
•	@backstage/plugin-catalog-backend-module-gitlab - "^0.3.20" 
•	@backstage/plugin-catalog-backend-module-gitlab-org - "^0.0.4" 

### 👍 Expected behavior

Backstage Catalog should identify/discover parent groups in gitlab

### 👎 Actual Behavior with Screenshots

Backstage Catalog is not identify/discover parent groups in gitlab

### 👟 Reproduction steps

Gitlab groups setup

"W1SC Platform" is parent group

![image](https://github.com/user-attachments/assets/62f36ed8-d609-415b-b9be-cb1bc45775ac)

"backstage-poc" is parent group

![image](https://github.com/user-attachments/assets/0813b303-f255-4442-b9a3-d20d8d86d67e)

Child Groups under "W1SC Platform"  we are able to see in backstage

![image](https://github.com/user-attachments/assets/ce548c8b-e635-43c7-9736-91d94af8b2ce)

![image](https://github.com/user-attachments/assets/4efd133b-a558-454a-9cee-a33bd9b6bfeb)

Parent group "W1SC Platform" its not listing

![image](https://github.com/user-attachments/assets/16492e59-4ff2-457a-b1b6-d1c4ffdd3bd5)

Parent group "backstage-poc" its not listing

![image](https://github.com/user-attachments/assets/b6f8e8cb-b493-4be8-bbd6-44bdff7b02e6)

For logged in user also its not able to assign the parent groups, for my user I have access to parent group

![image](https://github.com/user-attachments/assets/33873804-0610-4c3f-917c-e9b7ebd59324)

Below is the app-config.yaml

```yaml
integrations:
  gitlab:
    - host: gitlab.corp.org.com
      apiBaseUrl: https://gitlab.corp.org.com/api/v4
      token: ${GITLAB_TOKEN}
 
catalog:
  providers:
    gitlab:
      org-platform:
        host: gitlab.corp.org.com
        # branch: main # Optional. Used to discover on a specific branch
        # fallbackBranch: master
        allowInherited: true # Allow groups to be ingested even if there are no direct members.
        schedule: # Same options as in TaskScheduleDefinition. Optional for the Legacy Backend System.
[...]

PR #27575, opened 2024-11-10
extend filter logic to include parent group

hghtwr:

Extends the filter logic to include parent groups fetched from Gitlab for #26554 

Basically, it's rather simple: For cases that use `listGroups`, it's enough to adopt the filter logic for `shouldProcessGroups` to also include the parent group. 
For cases using `listDescendantGroups` (SaaS + Selfhosted with restricted users), we need to also push the org group to the results. To achieve this, we need to fetch the parent group first from the Gitlab API. That's why I had to implement `getGroupByPath` as alternative to fetching the group by id. 

I had to adopt the tests a bit, so I'd request a thorough review. I tested it on Gitlab SaaS, hopefully I get the time to test it non self-hosted as well, but the logic should be the same.

grafik
grafik
grafik
grafik


jhaals:
Make sure to mark this as ready for review once you're done otherwise it won't be picked up by us :)

freben (approved):
👍

PR #27581, opened 2024-11-11
chore(scaffolder-backend-module-gitlab): Upgrade to gitbeaker ^41.2.0

mpsanchis:

Upgrade gitbeaker to `^41.2.0` (major version) in `scaffolder-backend-module-gitlab`.

First split of https://github.com/backstage/backstage/pull/27360 as requested.


CptnFizzbin:
do you know if this update would fix the error messages from GitLab (as I reported in #27600)

For example when trying to create a branch via the API, if there are branch name rules configured, the error message is added to a `description` field instead of the error's `message`

mpsanchis:
Hi @CptnFizzbin 
I haven't tried it, I'm not sure. In fact this PR is a "prerequisite" for me, because I want to add a feature that only newer versions of Gitbeaker support. I have done the bare minimum to do the upgrade (ideally I would have only modified `package.json`s). Feel free to give it a try and let me know 😄

plugins/scaffolder-backend-module-gitlab/report.api.md
```diff
@@ -76,11 +76,17 @@ export const createGitlabProjectDeployTokenAction: (options: {
 }) => TemplateAction<
   {
     name: string;
+    scopes: (
+      | 'read_repository'
+      | 'read_registry'
+      | 'write_registry'
+      | 'read_package_registry'
+      | 'write_package_registry'
+    )[];
```
freben, line 85: Hm. Is this a good thing being the specific strings? Is this truly the only possible scopes now and forever? :)

plugins/scaffolder-backend/report.api.md
```diff
@@ -354,7 +354,11 @@ export const createPublishGitlabMergeRequestAction: (options: {
     sourcePath?: string | undefined;
     targetPath?: string | undefined;
     token?: string | undefined;
-    commitAction?: 'auto' | 'update' | 'delete' | 'create' | 'skip' | undefined;
+    commitAction?: 'auto' | 'update' | 'delete' | 'create' | 'skip' | undefined
+    /**
+     * @public
+     * @deprecated use import from `@backstage/plugin-scaffolder-backend-module-github` instead
```
freben, line 360: Wait where did this come from? :) I don't think I see it elsewhere in the code and this is an autogenerated file eh?

freben (commented):
Alright, looking good, just minor questions

plugins/scaffolder-backend-module-gitlab/report.api.md
mpsanchis: Up to you guys :) Having it as string allows "unwanted" behaviour for now, but of course the user will quickly realize.

I'm happy to leave it as a `string`. Should I cast it `as DeployTokenScope` in `createGitlabProjectDeployTokenAction` when calling `api.DeployTokens.create`?

CptnFizzbin: might be good to list out the known good ones, and have `| string` tagged onto the end. Would allow intellisense to suggest these values, while also providing an escape hatch (and type hints) that there could be more options in the future.

freben: Yep that's a good middle ground. Will still need a cast internally but is useful to the end user.

plugins/scaffolder-backend/report.api.md
freben: Nah leave it in if that's how it got generated. Curious thing tho 🤔

freben:
seems to need a rebase now too after the other PR got merged

plugins/scaffolder-backend-module-gitlab/report.api.md
mpsanchis: Can you give me a hand on this, guys? TS considers some types as "subtypes" of others. For instance:
```typescript
let a: string | "read_repository" | "read_registry" | "write_registry" | "read_package_registry" | "write_package_registry";
```
is equivalent (in VSCode suggestions) to:
```typescript
let a: string;
```

With `zod`, these types would be generated with:
```typescript
scopes: z.array(
  z.union([
    z.enum([
      'read_repository',
      'read_registry',
      'write_registry',
      'read_package_registry',
      'write_package_registry',
    ]),
    z.string()
  ]),
  { description: 'Scopes' },
```

But I get no benefit with intellisense, and the API report will declare `string[]`.

I'm good with reverting it to `string[]`. It looks like the only viable alternative.

mpsanchis: I'm pushing it with `string[]` for now. If you like it, we continue like that. Otherwise let me know and I change it to `DeployTokenScope[]` (or maybe a 3rd option?)

CptnFizzbin: leaving it as `string[]` is probably our best bet

plugins/scaffolder-backend/report.api.md
mpsanchis: Now it wasn't added anymore by the api report generator.. weird 🤷

mpsanchis:
@CptnFizzbin & @freben it seems that CI is green now 😄 Do we stay with the current state? Or do you want something else from my side?

.changeset/shy-teachers-count.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend-module-gitlab': minor
+---
+
+Upgraded gitbeaker to ^41.2.0
```
freben, line 5:
```suggestion
**BREAKING**: Upgraded the `gitbeaker` library to version 41. As part of this, the `scopes` parameter to the `gitlab:projectDeployToken:create` is no longer optional, so you will have to pass it a value (for example `['read_repository']`).
```

PR #27603, opened 2024-11-12
CatalogTable title -- properly capitalize Kind facets

tylerd-canva:

This is actually my second attempt at this after trying to add `label` to the `EntityKindFilter` itself (see https://github.com/backstage/backstage/pull/27590).

I am still learning React so I'm sure there's some improvements to be made here, especially in how `useAllKinds()` is called from CatalogTable.tsx.

### Before
before screenshot

### After
after screenshot

There is one problem, which is how `pluralize` deals with casing -- for `API` it matches the case instead of giving us `APIs`.

problematic APIS


plugins/catalog-react/src/components/EntityKindPicker/kindFilterUtils.ts
```diff
@@ -14,48 +14,21 @@
...
-    return items;
-  }, [catalogApi]);
-
-  return { loading, error, allKinds: allKinds ?? [] };
-}
-
 /**
  * Filter and capitalize accessible kinds.
  */
 export function filterKinds(
```
tylerd-canva, line 20: I wonder if it would be better to accept the `allowedKinds` and `forcedKinds` optional parameters in `useAllKinds()` instead and do the filtering there?

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
```diff
@@ -168,7 +171,10 @@ export const CatalogTable = (props: CatalogTableProps) => {
     },
   ];
 
-  const currentKind = filters.kind?.value || '';
+  let currentKind = filters.kind?.value || '';
+  if (!loadingKinds && currentKind) {
+    currentKind = allKinds[currentKind] || currentKind;
+  }
```
tylerd-canva, line 177: the problem here is that the `useAllKinds()` hook results in a loading state which means there's two titles that get renders, for example `Owned components (1)` and then `Owned Components (1)`. Does anyone know the best way to deal with this?

PR #27639, opened 2024-11-14
feature: update example app to use 'discoveryFeatureLoader'

schultzp2020:


freben (commented):
Interesting! Let me ask the team how "ready" we are for this to be the primary mode of consumption. Especially since it means that basically all installation docs need to be updated - if you both use this discovery loader AND add a thing "by hand" with `backend.add`, I believe there will be errors at startup. And things get a bit more "implicit" rather than seeing in front of yourself what's going on, so there's a balancing act. We'll have a chat and get back to ya.

vinzscam:
Shall we start by documenting this as an experimental feature in the docs?
It could also be a breaking change if adopters are conditionally enabling some plugins under certain conditions.

Issue #27631, opened 2024-11-13
🚀 Feature: Capitalize Kind Facets in Catalog Table Title


tylerd-canva:
### 🔖 Feature description

Use the properly capitalized version of the entity Kind in the Catalog Table. This is what it *should* look like:

![mockup with capitalized Components title](https://github.com/user-attachments/assets/3525ae43-5681-48e8-9527-3236cd763fe4)

### 🎤 Context

Currently the default CatalogTable shows a dynamic title based on the Kind and Type filters. Kind is a special case because there is a properly capitalized version (eg `Component`) which shows up in the Filter/Picker but under the hood we use a lowercase version (eg `component`). Unfortunately the current title construction only has access to the filter value so you get the lowercase version in the title.

![existing behavior with lowercase components title](https://github.com/user-attachments/assets/2fa4f6e5-86f1-4f2f-b248-3fa686306259)

### ✌️ Possible Implementation

I tried two different approaches for this before opening this issue:
* https://github.com/backstage/backstage/pull/27590 adds a `label` attribute to the `EntityKindFilter` and then pulls that if it is set when constructing the table title.
* https://github.com/backstage/backstage/pull/27603 exports the `useAllKinds()` hook and uses that to retrieve the labels for a given Kind filter value when constructing the table title.

Both of them work functionally, but I think they both have similar issues at the React level (which I am far from an expert with) and both seem to have test-time issues which I haven't been able to fully resolve.

I am looking for advice on which of these two approaches is preferred, or if there is an even better third option (likely) that I haven't considered.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

tylerd-canva:
Related question, is there ongoing work to change how the CatalogTable toolbar/search/title works as implied in this TODO?

https://github.com/backstage/backstage/blob/13180c52602fbd3c6c8a719711ddbf7fa35d52e2/plugins/catalog/src/components/CatalogTable/CatalogTable.tsx#L174

cc @timbonicus

vinzscam:
which version of Backstage are you using? It looks correct on demo.backstage.io

tylerd-canva:
[...]

PR #27648, opened 2024-11-14
Capitalize Kind facet in CatalogTable title

tylerd-canva:

Addresses https://github.com/backstage/backstage/issues/27631

This is actually my second attempt at this after trying to add `label` to the `EntityKindFilter` itself (see https://github.com/backstage/backstage/pull/27590).

I am still learning React so I'm sure there's some improvements to be made here, especially in how `useAllKinds()` is called from CatalogTable.tsx.

### Before
before screenshot

### After
after screenshot

There is one problem, which is how `pluralize` deals with casing -- for `API` it matches the case instead of giving us `APIs`.

problematic APIS

### DIFF
--- a/plugins/catalog-react/src/components/EntityKindPicker/kindFilterUtils.ts
+++ b/plugins/catalog-react/src/components/EntityKindPicker/kindFilterUtils.ts
@@ -19,12 +19,12 @@ import useAsync from 'react-use/esm/useAsync';
 import { catalogApiRef } from '../../api';
 
 /**
- * Fetch and return all availible kinds.
+ * Fetch and return all available kinds.
  */
 export function useAllKinds(): {
   loading: boolean;
   error?: Error;
-  allKinds: string[];
+  allKinds: Map<string, string>;
 } {
   const catalogApi = useApi(catalogApiRef);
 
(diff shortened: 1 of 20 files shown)


plugins/catalog-react/src/components/EntityKindPicker/kindFilterUtils.ts
```diff
@@ -14,48 +14,21 @@
...
-    return items;
-  }, [catalogApi]);
-
-  return { loading, error, allKinds: allKinds ?? [] };
-}
-
 /**
  * Filter and capitalize accessible kinds.
  */
 export function filterKinds(
```
tylerd-canva, line 20: We could, in theory, move the filtering functionality here into the `useAllKinds` hook directly, which might reduce a bit of complexity.

plugins/catalog/src/components/CatalogTable/OffsetPaginatedCatalogTable.tsx
tylerd-canva, line 1: note this change is only necessary because of the bug I've fixed separately in https://github.com/backstage/backstage/pull/27632

tylerd-canva, line 1: this change is only needed until https://github.com/backstage/backstage/pull/27632 is merged (the catalog in `yarn dev` does not have a visible title without this change)

tylerd-canva:
Just had a thought about the pluralization problem. I haven't worked with it yet, but maybe this (and actually this whole PR) could be solved by the EntityPresentationAPI? Of course in this case it's Entity _facets_ we need to present, but if we could extend it slightly to support this use case we could get the proper display form of a Kind facet in both plural and singular forms from there rather than relying on the native `pluralize` library.

plugins/catalog/src/components/CatalogTable/CatalogTable.test.tsx
```diff
@@ -53,10 +55,24 @@ const entities: Entity[] = [
...
-  const mockApis = TestApiRegistry.from([
-    starredEntitiesApiRef,
-    new MockStarredEntitiesApi(),
-  ]);
+  const mockApis = TestApiRegistry.from(
+    [starredEntitiesApiRef, new MockStarredEntitiesApi()],
+    [
+      catalogApiRef,
+      {
+        getEntityFacets: jest.fn().mockResolvedValue({
```
freben, line 63: This is one of those places where [a `catalogApiMock`](https://github.com/backstage/backstage/blob/51b2573091a1f8807c8180c8a9c56eb9eb0779f4/plugins/catalog-react/src/testUtils/catalogApiMock.ts#L61) with a single entity in it would simplify things.

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
```diff
@@ -97,8 +98,10 @@ export const CatalogTable = (props: CatalogTableProps) => {
     emptyContent,
   } = props;
   const { isStarredEntity, toggleStarredEntity } = useStarredEntities();
-  const entityListContext = useEntityList();
 
+  const { allKinds } = useAllKinds();
+
+  const entityListContext = useEntityList();
```
freben, line 104: This does make me uneasy. Especially for users with huge catalogs, we now make this extra work/request mandatory and happen on every catalog table everywhere. And additionally, it's work that's already being done elsewhere (by the actual kind filter) so it's duplucated and wasteful.

You already are reading `filters.kind?.value`. Is the problem that that one doesn't have the right casing? Can we make it have the right case in the first place instead? Or have an additional special field `originalCaseValue` if we're forced to keep the old behavior for some reason - but I don't thing we are.

tylerd-canva: I have reworked this significantly and it no longer requires any additional calls.

PR #27676, opened 2024-11-15
slightly speed up the refresh loop

freben:
This does two things to the `getProcessableEntities` call:

* No longer run it in a transaction. This really doesn't matter for this use case.
* Do not fetch the `processed_entity` and `last_discovery_at` columns because these are of no interest to the processing system

This in total ends up being of comparable speed for small batch sizes since there's so much other overhead there. But in high speed scenarios like when catching up with a backed up queue, it's 10-15% faster at least in the local perf test.

### DIFF
--- a/plugins/catalog-backend/src/database/DefaultProcessingDatabase.ts
+++ b/plugins/catalog-backend/src/database/DefaultProcessingDatabase.ts
@@ -198,40 +198,46 @@ export class DefaultProcessingDatabase implements ProcessingDatabase {
   }
 
   async getProcessableEntities(
-    txOpaque: Transaction,
+    maybeTx: Transaction | Knex,
     request: { processBatchSize: number },
   ): Promise<GetProcessableEntitiesResult> {
-    const tx = txOpaque as Knex.Transaction;
+    const knex = maybeTx as Knex.Transaction | Knex;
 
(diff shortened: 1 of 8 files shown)


plugins/catalog-backend/src/database/DefaultProcessingDatabase.ts
```diff
@@ -198,40 +198,46 @@ export class DefaultProcessingDatabase implements ProcessingDatabase {
...
 
   async getProcessableEntities(
-    txOpaque: Transaction,
+    maybeTx: Transaction | Knex,
     request: { processBatchSize: number },
   ): Promise<GetProcessableEntitiesResult> {
-    const tx = txOpaque as Knex.Transaction;
-
-    let itemsQuery = tx<DbRefreshStateRow>('refresh_state').select();
+    const knex = maybeTx as Knex;
```
Rugvip, line 204: `Knex | Knex.Transaction` for safety?

Rugvip (approved):
Nice! :+1:

plugins/catalog-backend/src/database/DefaultProcessingDatabase.ts
freben: Indeed! Fix'd

PR #27680, opened 2024-11-15
BEP 0012-device-auth

phclark:

### DIFF
--- a/beps/0012-device-auth/README.md
+++ b/beps/0012-device-auth/README.md
@@ -67,7 +67,8 @@ Best effort support for implementation of refresh token should be included, but 
 The following are out of scope for this BEP:
 
 - Access delegation, using OAuth providers, such as the ability to acquire a Github token when running a template task.
-- Implementation of refresh flow, as we expect future design of the auth signin to make this easier to implement.
+- Implementation of refresh flow, as we expect future design of the auth sign-in to make this easier to implement.
+- A client implementation for CLIs, including a TypeScript API for CLI developers to use. This requires support for profiles or similar in the CLI, and is considered a future effort.
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


beps/0012-device-auth/README.md
```diff
@@ -0,0 +1,276 @@
...
+If there is any particular feedback to be gathered during the rollout, this should be described here as well.
+-->
+
+## Dependencies
+
+<!--
+List any dependencies that this work has on other BEPs or features.
+-->
+
+## Alternatives
```
secustor, line 270: Do you considered simply adding an extension point which allows to call the `/userinfo` endpoint of an identity provider` with the provided token?  

That way the same data would be available like on the "normal" login. The token would be set in the `auth-backend` and mapped to the user. 

This would have less risk as the Oauth2 device flow is maintained by dedicated products.

Rugvip: That's roughly how the regular auth flow works today, but I don't quite see how it helps us with device auth? Device auth requires more of an ability for clients to initialize a decoupled auth flow that users can then approve, rather than initializing auth directly. In this flow the "login / fetch token" step would need to be the device auth flow, initialized towards Backstage. There would then be a second client that completes it towards Backstage and/or the IdP.

secustor: > That's roughly how the regular auth flow works today, but I don't quite see how it helps us with device auth? 

Yes, that is the reason why I'm proposing it. The IdP would implement the actual authorization part and Backstage would have not to implement it as it its not the core functionality. IMO this should be offloaded as much to dedicated tools e.g. Keycloak or commercial offerings. 

> Device auth requires more of an ability for clients to initialize a decoupled auth flow that users can then approve, rather than initializing auth directly. In this flow the "login / fetch token" step would need to be the device auth flow, initialized towards Backstage. There would then be a second client that completes it towards Backstage and/or the IdP.

I think this is the question should Backstage fully implement the Device Flow or should the externalAccess JWKS option be extended to make use of the permission system. 

```diff
backend:
  auth:
    externalAccess:
      - type: jwks
        options:
          url: https://another-example.com/.well-known/jwks.json
          issuer: https://example.com
+         userInfoPath: /userinfo
```

Alternatively a `externalAccess` type `oidc`: 

```
backend:
  auth:
    externalAccess:
      - type: oidc
         options:
           url: https://another-example.com/
```

I think in the end both would work. So also happy if Backstage implements the DeviceAuth flow, but I simply think this will pull a lot of resources security wise 😄

PR #27687, opened 2024-11-16
Change create-app yarn test to use `--since`

tylerd-canva:

Just a proposal. IMHO the behaviour of `yarn test` should match `yarn lint`.

### DIFF
--- a/.changeset/perfect-tomatoes-complain.md
+++ b/.changeset/perfect-tomatoes-complain.md
@@ -1,0 +1,5 @@
+---
+'@backstage/create-app': patch
+---
+
+Change 'test' package script to use --since just like 'lint'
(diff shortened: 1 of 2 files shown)


tylerd-canva:
Actually based on https://github.com/backstage/backstage/blob/3371d2023135da1d50ac4c1a77cfe1eb58cac5f4/CONTRIBUTING.md?plain=1#L143 I think this change is correct and should be applied here as well perhaps https://github.com/backstage/backstage/blob/ecdb1ce0b66149a7051cd6635153ab0070494994/package.json#L52

tylerd-canva:
On second thought... I don't think this is working properly. Is there a known issue here? I made a trivial change in one of my packages locally and it claims there are no changes.

Rugvip (changes requested):
The `test` and `lint` scripts are primarily intended for local development. The `test` script already has "since"-like behavior built-in via Jest, which `lint` doesn't have and why we use `--since` there.

Not opposed to adding something for CI though, could for example be `:changed` scripts for both?

Rugvip:
Oh and yep the docs aren't really accurate, `yarn test` tests pending changes that are not yet committed, rather than changes since master

tylerd-canva:
@Rugvip thank you for the explanation! I will rework this.

PR #27689, opened 2024-11-16
app-backend: serve templated index.html from memory

Rugvip:

This switches the templating of `index.html` in the app backend to happen in memory rather than being written to the filesystem. This means you no longer need to use the `app.disableConfigInjection` flag when using a read-only filesystem, allowing for config injection there too.

This also removes the double config injection that was present for a few releases as we move from injecting config into the static config to instead have it in `index.html`.

### DIFF
--- a/.changeset/five-tables-remain.md
+++ b/.changeset/five-tables-remain.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-app-backend': patch
+---
+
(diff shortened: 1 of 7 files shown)


freben (approved):
Neat. Less work than I had thought

Rugvip:
@freben yeah same. I remember it being tricky last time I looked at it, but I think that was before switching to templating `index.html

PR #27690, opened 2024-11-17
Added sorted prop to EntityTypePicker

matdtr:
This is my first PR hope that i did all good. 
This add the possibility to sort the values inside `EntityTypePicker`. 

Default behaviour:

Screenshot 2024-11-17 alle 16 58 41

With   `<EntityTypePicker sorted />`  (the value `all` is always first):

Screenshot 2024-11-17 alle 17 15 14


matdtr:
not stale

matdtr:
@Rugvip @vinzscam for visibility

awanlin (changes requested):
Thanks for this @matdtr, I'd like to get feedback from the Catalog Maintainers but my thinking is to avoid the prop entirely and this should be the default. As someone who is generally pro "opt-in" I think in this case it's not the right direction and having them sorted by default will be a better experience for everyone.

freben:
Oh sorry I hadn't seen this. Indeed - `all` at the top followed by sorted server side entries would be great as the default without a prop. Thanks! We can treat it as a bug that it wasn't sorted in the first place.

matdtr:
Hi @freben @awanlin removed the prop and sorted by default. Let me know if needed something else

freben (approved):
Thanks! Gonna force this one through so it gets in today's release, hope that's ok @ahhhndre

awanlin:
Fine with me, thanks for merging @freben and for the contribution to fix this @matdtr 🚀

Issue #23035, opened 2024-02-16
🐛 Bug Report: Human Duration in Config Adds Not Overrides


awanlin:
### 📜 Description

If I have a config like this in my `app-config.yaml`:

```yaml
schedule:
  frequency: { hours: 6 }
  timeout: { minutes: 5 }
```

And like this in my `app-config.local.yaml`:

```yaml
schedule:
  frequency: { minutes: 3 }
  timeout: { minutes: 5 }
```

I ended up with a schedule like this in my database:

![Screenshot 2024-02-16 at 1 31 53 PM](https://github.com/backstage/backstage/assets/67169551/861f32c4-f82a-4213-924e-0601872c7e07)

When I would prefer it to be:

![Screenshot 2024-02-16 at 1 21 10 PM](https://github.com/backstage/backstage/assets/67169551/ee451aca-7009-409a-9e1b-5bd0398e5ca5)

**Note:** this is using `HumanDuration`

### 👍 Expected behavior

Should result in this:

![Screenshot 2024-02-16 at 1 21 10 PM](https://github.com/backstage/backstage/assets/67169551/ee451aca-7009-409a-9e1b-5bd0398e5ca5)

### 👎 Actual Behavior with Screenshots

Ended up with a schedule like this:

![Screenshot 2024-02-16 at 1 31 53 PM](https://github.com/backstage/backstage/assets/67169551/861f32c4-f82a-4213-924e-0601872c7e07)

### 👟 Reproduction steps

1. Create a Backstage instance
2. Setup any scheduled configuration in Backstage that uses `HumanDuration`, for example most of the Entity Providers.
3. Make the schedule in `minutes` and set it to 3 in your `app-config.local.yaml`
4. Make the schedule in `hours` and set it to 6 in your `app-config.yaml`

**Notice:** in your database the schedule is for 6 hours and 3 minutes

### 📃 Provide the context for the Bug.

When doing local development this makes it confusing as to when the schedule will run. As a work around if you use one unit, say all minutes then it worked as expected.

### 🖥️ Your Environment

```text
OS:   Darwin 23.3.0 - darwin/arm64
node: v18.18.2
yarn: 3.5.0
cli:  0.25.2-next.3 (installed)
backstage:  1.23.0-next.3

Dependencies:
  @backstage/app-defaults                                          1.5.0-next.3
  @backstage/backend-app-api                                       0.5.10, 0.5.11-next.3
[...]

PR #27691, opened 2024-11-17
implement support for string form human durations in config

freben:
Entering durations in config has been a bit iffy. You need to use nested objects (as if your yaml wasn't already indented enough as is), you have to learn the format, and you're subject to #23035.

This PR enables `readDurationFromConfig` to support [the `ms` library](https://github.com/vercel/ms)'s human friendly string formats,

```yaml
timeout: 2 minutes
```

as well as [ISO8601 duration](https://en.wikipedia.org/wiki/ISO_8601#Durations) strings

```yaml
frequency: P1D
```

in addition to the previous raw

```yaml
ttl:
  seconds: 3
```

Why all the formats? Well, ultimately what I thought would be the most useful was to get `ms` in there, just like we sometimes use `yn` for boolean things in scripts etc. It's a nice understandable format to use as the new standard. The ISO format is there basically mostly because the scheduler already supported that and wanted to continue to do so.

For now, this is only on the config side of things. We could make it so that `durationToMilliseconds` supports string forms too, if we want, but it didn't feel reasonable to do right now.

Should probably update the docs later as well, to highlight the `ms` format as the preferred one, if we agree that this was a good approach.

### DIFF
--- a/.changeset/clean-pears-roll.md
+++ b/.changeset/clean-pears-roll.md
@@ -1,0 +1,5 @@
+---
+'@backstage/config': minor
+---
+
+Make `readDurationFromConfig` support both ISO and ms formats as well, to make it easier to enter time as an end user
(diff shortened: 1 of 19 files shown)


Rugvip (approved):
Nice 🎉 

:shipit:

Issue #27149, opened 2024-10-15
🚀 Feature: Paginated TechDocs Index


awanlin:
### 🔖 Feature description

It would be awesome to have a paginated TechDocs index much like the one for the Catalog. If there is an existing issue for this feel free to share that, noting came up in my searching.

### 🎤 Context

As Adopters expand their Catalog with entities you often run into issues with performance. The paginated Catalog index is a great help for this but that is not available for the TechDocs index

### ✌️ Possible Implementation

There must be a way to use the underlying API used for the Catalog index to be used to drive the TechDocs index. The TechDocs index is just a list of all entities with the [TechDocs annotation](https://backstage.io/docs/features/software-catalog/well-known-annotations#backstageiotechdocs-ref).

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

aramissennyeydd:
👋 Interesting - it looks like the techdocs table is a completely new table element, https://github.com/backstage/backstage/blob/master/plugins/techdocs/src/home/components/Tables/DocsTable.tsx#L67. It uses the same underlying `EntityListContext` that the catalog page does, so it wouldn't be awful to either add support for pagination to the `DocsTable` component. Longer term though, I'm curious if there are plans to merge the 2 tables?

hyb175:
Hello! I hope to work on this issue as part of contribfest

awanlin:
Thanks @hyb175, I've assigned this to you. You can look at the CatalogIndexPage in the Catalog plugin for some examples as to how best approach this 👍

MihaiTabara:
Found in triaging. I think this is done, please holler if needs re-opening.

PR #27694, opened 2024-11-18
[Feature] Add Pagination to Tech Docs Table

hyb175:

This PR is to address issue https://github.com/backstage/backstage/issues/27149 - Add pagination support to TechDocs index page.
Screenshot 2024-11-18 at 12 23 52 AM
Screenshot 2024-11-18 at 12 24 05 AM

### DIFF
--- a/plugins/techdocs/src/home/components/Tables/EntityListDocsTable.tsx
+++ b/plugins/techdocs/src/home/components/Tables/EntityListDocsTable.tsx
@@ -24,14 +24,22 @@ import {
   TableProps,
   WarningPanel,
 } from '@backstage/core-components';
+import { configApiRef, useApi, useRouteRef } from '@backstage/core-plugin-api';
+import { RELATION_OWNED_BY } from '@backstage/catalog-model';
 import {
+  getEntityRelations,
+  humanizeEntityRef,
   useEntityList,
   useStarredEntities,
 } from '@backstage/plugin-catalog-react';
 import { DocsTable } from './DocsTable';
+import { OffsetPaginatedDocsTable } from './OffsetPaginatedDocsTable';
+import { CursorPaginatedDocsTable } from './CursorPaginatedDocsTable';
 import { actionFactories } from './actions';
 import { columnFactories } from './columns';
 import { DocsTableRow } from './types';
+import { toLowerMaybe } from '../../../helpers';
+import { rootDocsRouteRef } from '../../../routes';
 
 /**
  * Props for {@link EntityListDocsTable}.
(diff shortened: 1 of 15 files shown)


plugins/techdocs/src/home/components/Tables/EntityListDocsTable.tsx
```diff
@@ -65,6 +84,53 @@ export const EntityListDocsTable = (props: EntityListDocsTableProps) => {
     ),
   ];
 
+  const documents = entities.map(entity => {
```
squid-ney, line 87: Hi, would it be possible for `CursorPaginatedDocsTable` and `OffsetPaginatedDocsTable` to use `DocsTable` so that the entity mapping logic is not duplicated? I noticed this same logic is also present in `DocsTable`  [here](https://github.com/backstage/backstage/blob/5550aa743348cec674e993d5d46298a99e6435b6/plugins/techdocs/src/home/components/Tables/DocsTable.tsx#L74).

awanlin:
This is awesome, thanks so much @hyb175, cool to see your ContribFest contribution!

plugins/techdocs/src/home/components/Tables/EntityListDocsTable.tsx
hyb175: Sounds good!
I initially was going to remove the old <DocsTable /> and rename the <EntityListDocsTable /> to <DocsTable /> but decided against it for the breaking API changes.

hyb175: Hi! After looking at the code a little more closely, the `<DocsTable />` component seems rather limiting to me and a little awkward to be used to support pagination, and if it does then`<EntityListDocsTable />` provides little value.
I have a proposal - is it fair to deprecate the usage of the old `<DocsTable />` and promote the usage of the new `<EntityListDocsTable />` (as the new `<DocsTable />` perhaps).
If the main concern here is the duplication of the data mapping, perhaps extracting the logic in a little helper locally could help address that.
Let me know what you think!

hyb175:
Rebased to add `actions` back to the paginated table.
Extract `defaultColumns` and `entitiesToDocsMapper` out of the table components

plugins/techdocs/src/home/components/Tables/EntityListDocsTable.tsx
johnphilip283: After looking at the signatures of both these components, it seems to make sense to leave both components as they are and allow folks to use whichever one they'd like to avoid breaking changes - this looks great!

johnphilip283 (approved):

camilaibs (approved):

PR #27695, opened 2024-11-18
cli: inject config via index.html for rspack

Rugvip:

This switches the config injection for Rspack builds and the dev server to be done via `index.html` instead, just like the `app-backend` now does. I would've liked to switch over the WebPack path as well, but for some reason it doesn't update correctly across reloads so keeping that as is for now.

Long-term I'm hoping that we can completely move over to Rspack and do all config injection straight into `index.html` both in the CLI and at runtime.

### DIFF
--- a/.changeset/metal-carrots-brush.md
+++ b/.changeset/metal-carrots-brush.md
@@ -1,0 +1,9 @@
+---
+'@backstage/cli': patch
+---
+
+When using the experimental Rspack flag the app build and dev server now injects configuration via a `<script type="backstage.io/config">...</script>` tag in `index.html` rather than the `process.env.APP_CONFIG` definition, which will now be defined as an empty array instead.
+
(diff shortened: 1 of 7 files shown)


benjdlambert (approved):
Nice! Why do we need the `frontend-defaults` changeset? Just to pull in the new package?

freben (approved):

Rugvip:
@benjdlambert because of https://github.com/backstage/backstage/blob/9f0c7f70668ad7744e835cdd0abacd08d564a932/packages/frontend-defaults/src/createApp.tsx#L21 😅

benjdlambert:
@Rugvip yaha, that's an unfortunate thing to have to remember for all packages that do imports like that.

I wonder if we can do something to the changesets action to try and autodetect this stuff too.

PR #27697, opened 2024-11-18
WIP: Canon

cdedreuille:


packages/canon/src/components/stack/sprinkles.css.ts
```diff
@@ -0,0 +1,37 @@
...
+  defaultCondition: 'xs',
+  properties: {
+    ...commonProperties,
+  },
+  shorthands: {
+    ...boxShorthands,
+    direction: ['flexDirection'],
+    items: ['alignItems'],
+    justify: ['justifyContent'],
+  },
```
Rugvip, line 31: early thought on this bit: could we rely on an existing solution for the styling API as well? I think I'd prefer if we did not need to maintain and document our own API for shorthands and such, since that's a very universal problem and not specific to Backstage at all

PR #27699, opened 2024-11-18
Feat: Additional test coverage and component structure for TemplateCard

kalleericson:

`TemplateCard` in `plugin-scaffolder-react` is too intricate and monolithic. This PR breaks it up into smaller, exposed, components to improve testability and reusability.

Also adds test-selectors to all its important nodes, to allow downstream projects to test the component properly.

**No logic or layout change, strictly code maintenance**

### DIFF
--- a/plugins/scaffolder-react/src/next/components/TemplateCard/TemplateCard.tsx
+++ b/plugins/scaffolder-react/src/next/components/TemplateCard/TemplateCard.tsx
@@ -15,68 +15,33 @@
  */
 
 import { RELATION_OWNED_BY } from '@backstage/catalog-model';
-import { MarkdownContent, UserIcon } from '@backstage/core-components';
-import {
-  IconComponent,
-  useAnalytics,
-  useApp,
-} from '@backstage/core-plugin-api';
-import {
-  EntityRefLinks,
-  getEntityRelations,
-} from '@backstage/plugin-catalog-react';
+import { IconComponent, useAnalytics } from '@backstage/core-plugin-api';
+import { getEntityRelations } from '@backstage/plugin-catalog-react';
 import { TemplateEntityV1beta3 } from '@backstage/plugin-scaffolder-common';
-import Box from '@material-ui/core/Box';
 import Card from '@material-ui/core/Card';
 import CardActions from '@material-ui/core/CardActions';
 import CardContent from '@material-ui/core/CardContent';
-import Chip from '@material-ui/core/Chip';
 import Divider from '@material-ui/core/Divider';
-import Button from '@material-ui/core/Button';
 import Grid from '@material-ui/core/Grid';
 import { makeStyles, Theme } from '@material-ui/core/styles';
-import LanguageIcon from '@material-ui/icons/Language';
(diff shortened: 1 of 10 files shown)


kalleericson:
Hey @benjdlambert ! 

Since the `TemplateCard` takes very limited props, the only real way to differentiate different types or categories of cards in lists of components seem to be to carbon copy the entire Card. Breaking it up like this, would allow downstream projects to reuse a lot of the logic and structure, whilst still making local differentiations. 

This might be swinging a bit too far in the "exposed"-direction, admittedly, but to expose types such as the AdditionalLinks and such would also go a long way in building for consistent downstream cases.

I find that there's a lot of need throughout Backstage for better test selectors. We use a lot of end-to-end tests, and it's currently a pain to make the scripts reliable and non-hacky. So I would argue a selector structure like suggested here would be nice to consider more broadly.  

I figured since it's clearly marked `alpha`, that it lowered the risk of exposing components - even those under development still? If that's a misunderstanding, I'm all ears.

acierto:
Hey @kalleericson,

What stops you to define your own component and supply it via `EXPERIMENTAL_TemplateListPageComponent`?

kalleericson:
Hi @acierto !

Nothing is _stopping_ us from using `EXPERIMENTAL_TemplateListPageComponent`, apart from a general approach to try and use the upstream components to the maximum extent. To carbon the `TemplateCard` _just_ to add test selectors, would add a lot of unnecessary complexity and required sync in upcoming upstream updates.

If this PR is taking the concept of breaking the card down into smaller components too far, I'll be happy to roll back all but the added selectors? A bit reluctantly, considering code complexity, but it would help us out a lot.

benjdlambert:
@kalleericson I'm a little reluctant to break apart this component any further and re-export them as building blocks for other usages. I don't think we're there yet. I'd much rather wait for something like #27726 to settle before exposing smaller pieces of components and patterns.

Happy to move forward with the selectors and tests though :pray:

plugins/scaffolder-react/src/next/components/TemplateCard/TemplateCard.tsx
```diff
@@ -15,81 +15,42 @@
...
+  url: string;
+}
+
 /**
  * The Props for the {@link TemplateCard} component
  * @alpha
  */
 export interface TemplateCardProps {
   template: TemplateEntityV1beta3;
-  additionalLinks?: {
```
benjdlambert, line 87: Can we keep this type inline for now? Don't really wanna add too much more API surface exposed out of the package :pray:

benjdlambert (changes requested):
Thanks for this! Just a small change about the types and I think it's good :Tada:

plugins/scaffolder-react/src/next/components/TemplateCard/TemplateCard.tsx
kalleericson: Sure thing, you're the boss :)

benjdlambert (approved):
Nice! Thanks for this :tada:

PR #27701, opened 2024-11-18
fix: make schedule in catalog.providers.github required

djamaile:

Seems the code expects this config to be set else it will error. So, making it required in the schema. Same has been done for `githubOrg`

### DIFF
--- a/.changeset/late-mails-pull.md
+++ b/.changeset/late-mails-pull.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-github': patch
+---
+
+Make schedule config required as is expected from the code
--- a/plugins/catalog-backend-module-github/config.d.ts
+++ b/plugins/catalog-backend-module-github/config.d.ts
@@ -114,9 +114,9 @@ export interface Config {
               visibility?: Array<'private' | 'internal' | 'public'>;
             };
             /**
-             * (Optional) TaskScheduleDefinition for the refresh.
+             * TaskScheduleDefinition for the refresh.
              */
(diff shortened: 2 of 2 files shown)


freben:
Hm. Is this true? https://github.com/backstage/backstage/blob/6a5de4df3b669f561875a4d6be5c9e1315b98ca7/plugins/catalog-backend-module-github/src/providers/GithubEntityProviderConfig.ts#L98

freben:
As long as you can call fromConfig as a user, you can pass in schedules from the outside and not have to give it in config, as far as i read things https://github.com/backstage/backstage/blob/6a5de4df3b669f561875a4d6be5c9e1315b98ca7/plugins/catalog-backend-module-github/src/providers/GithubEntityProvider.ts#L104

UsainBloot:
> As long as you can call fromConfig as a user, you can pass in schedules from the outside and not have to give it in config, as far as i read things
> 
> https://github.com/backstage/backstage/blob/6a5de4df3b669f561875a4d6be5c9e1315b98ca7/plugins/catalog-backend-module-github/src/providers/GithubEntityProvider.ts#L104

But in the instance of the new Backend system it would mean that you would always have to provide a schedule as there is no fallback without having to destructure the module and call `fromConfig` on the provider.

To keep the schema optional but still support the new Backend system, wouldn't it be best to have a default schedule defined here?
https://github.com/backstage/backstage/blob/6a5de4df3b669f561875a4d6be5c9e1315b98ca7/plugins/catalog-backend-module-github/src/providers/GithubEntityProviderConfig.ts#L98C1-L102C17

djamaile:
Closing this in favor of https://github.com/backstage/backstage/pull/27753

Issue #24600, opened 2024-05-02
🚀 Feature: Home Page Quick Start Card


nikolarAutodesk:
### 🔖 Feature description

A Quick Start card on the home page can be helpful for new users to get up to speed on their companies internal developer portal.
![genericCard](https://github.com/backstage/backstage/assets/146387693/cf2b21e1-65ce-40a0-90f4-5db567c5f052)

https://github.com/backstage/backstage/assets/146387693/b7dff080-d490-465d-a478-6f48961c3e42

### 🎤 Context

Our team frequently receives questions about our new IDP and requests for help with onboarding. So, our product and design team created this Quick Start card to link users to the most important information when they first land on the IDP.

We thought this might be helpful for other Backstage adopters, but we want to get an opinion on this before submitting a PR contributing to the [[@backstage/plugin-home](https://github.com/backstage/backstage/blob/master/plugins/home/api-report.md)](https://github.com/backstage/backstage/blob/master/plugins/home/api-report.md)

### ✌️ Possible Implementation

The following displays all of the features of the card we have implemented on our end. We would contribute this as a basic QuickStart card, where a developer can pass in props to populate the card specifically for their company

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

jhaals:
friendly ping to @backstage/home-maintainers

emmaindal:
Hello @nikolarAutodesk, thanks for opening this issue! Looks like a really good addition to the group of optional Homepage cards adopters can use for Backstage. I'll add a "help wanted" label to the issue so feel free to let us know if you are planning to work on it and we can assign you to the issue.

nikolarAutodesk:
I'd like to work on this during KubeCon Contribfest

awanlin:
Thanks @nikolarAutodesk, I've assigned it to you 🚀

nikolarAutodesk:
Hi @awanlin I am not sure what happened, but my [PR](https://github.com/backstage/backstage/pull/27705) did not automatically request a reviewer. Was there a process change? Do I need to request specific people?

nikolarAutodesk:
closing this issue since the related PR has been merged: https://github.com/backstage/backstage/pull/27705

PR #27705, opened 2024-11-19
[plugins/home] Add Home Page Quick Start Card

nikolarAutodesk:

**_Related Issue picked up during KubeCon Contribfest_** https://github.com/backstage/backstage/issues/24600

Adding a QuickStart card that can be used to display content which can be considered helpful for users onboarding to the Catalog. This card is customizable so that a developer can pass in props to populate the card specifically for their company. This also includes overridable styles. 

Screenshot 2024-11-18 at 5 45 44 PM

**Fun fact**: I was at the Contribfest session at KubeCon sitting at one of the back tables in a grey sweater 👋  I regret not introducing myself to the maintainers I never got a chance to stop by the booth!


awanlin:
Hi @backstage/home-maintainers, can you give this a review please?

plugins/home/src/homePageComponents/QuickStart/styles.ts
```diff
@@ -0,0 +1,68 @@
...
+    },
+    cardTitleIcon: {
+      verticalAlign: 'bottom',
+      marginLeft: '-4px',
+    },
+    videoContainer: {
+      borderRadius: '10px',
+      width: '100%',
+      height: 'auto',
+      background: '#000000',
```
emmaindal, line 55: do you think it makes sense to use the theme for this color instead?

plugins/home/src/homePageComponents/QuickStart/styles.ts
```diff
@@ -0,0 +1,68 @@
...
+  | 'imageSize'
+  | 'contentModal';
+
+export const useStyles = makeStyles(
+  (theme: Theme) => ({
+    link: {
+      display: 'inline-flex',
+      alignItems: 'center',
+      textDecoration: 'none',
+      color: `${theme.palette.link}!important`,
```
emmaindal, line 35: do we need to use the important flag here? would love to avoid it if possible

plugins/home/src/homePageComponents/QuickStart/styles.ts
```diff
@@ -0,0 +1,68 @@
...
+      '&:hover': {
+        background: 'transparent!important',
```
emmaindal, line 37: same here, if we can avoid using the important flag that would be great.

plugins/home/src/homePageComponents/QuickStart/QuickStart.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+import ContentImage from './static/backstageSystemModel.png';
+
+export default {
+  title: 'Plugins/Home/Components/QuickStartCard',
+  decorators: [
+    (Story: ComponentType<PropsWithChildren<{}>>) =>
+      wrapInTestApp(
+        <TestApiProvider apis={[]}>
+          <Story />
+        </TestApiProvider>,
```
emmaindal, line 30: if you are not using any apis I thiiink we should be able to just get rid of this wrapper, can you try that?

plugins/home/src/homePageComponents/QuickStart/ContentModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export const ContentModal = (props: ContentModalProps) => {
+  const { modalContent, linkContent } = props;
+  const styles = useStyles();
+  const [open, setOpen] = useState(false);
+
+  return (
+    <div className={styles.linkText} data-testid="content-modal-container">
+      <Link
+        to="#"
```
emmaindal, line 37: is it possible to skip this property when the component "button" is used?

plugins/home/src/homePageComponents/QuickStart/Content.tsx
```diff
@@ -0,0 +1,84 @@
...
+export const Content = (props: QuickStartCardProps): JSX.Element => {
+  const styles = useStyles();
+  return (
+    <>
+      <ContentModal
+        modalContent={props.image}
+        linkContent={props.modalTitle || 'Onboarding'}
+      />
+      <Typography variant="body1" paragraph>
+        {props.cardDescription || "Get started with Backstage's Catalog"}
```
emmaindal, line 60:
```suggestion
        {props.cardDescription || "Get started with Backstage"}
```

What do you think about using this as default instead? just thinking that its the home page and not specifically the catalog page so might be valuable to keep it more generic

plugins/home/src/homePageComponents/QuickStart/Content.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+import { Content } from './Content';
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+import { renderInTestApp, TestApiProvider } from '@backstage/test-utils';
+import ContentImage from './static/backstageSystemModel.png';
+
+describe('<QuickStartCard />', () => {
+  const Wrapper = ({ children }: { children?: React.ReactNode }) => (
+    <TestApiProvider apis={[]}>{children}</TestApiProvider>
```
emmaindal, line 25: same here, I don't think you need to use the test api wrapper if you don't have any apis

emmaindal (commented):
Really great work!! I've added a couple of comments from my side, thank you so much for the contribution 🎉

plugins/home/src/homePageComponents/QuickStart/ContentModal.tsx
nikolarAutodesk: I'm not sure I understand the question 🤔 
It looks like the prop `to` is required in the [LinkProps definition](https://github.com/backstage/backstage/blob/7d8777dbba10cabe0629e8f239806231519f5062/packages/core-components/src/components/Link/Link.tsx#L103).
Should I use something other than `Link`?

plugins/home/src/homePageComponents/QuickStart/styles.ts
nikolarAutodesk: sorry about that! All the style changes make sense. I should have self reviewed better!

plugins/home/src/homePageComponents/QuickStart/QuickStart.stories.tsx
nikolarAutodesk: looks like we still need `wrapInTestApp`. Without it we get a "missing context provider" error

plugins/home/src/homePageComponents/QuickStart/ContentModal.tsx
emmaindal: Oh my bad, didn't realised this was the Link component from backstage core components instead of mui. I'll continue to review your changes during the day!

emmaindal (approved):
Amazing, thank you so much for this! Great write up of changeset and storybook exampels as well. Really good work!

nikolarAutodesk:
Thank you for the review @emmaindal! How do we get this PR merged? Do I need more reviews?

Issue #27707, opened 2024-11-19
🐛 Bug Report: Redis Client builds incorrect cache keys with useRedisSets set to false


eipc16:
### 📜 Description

CacheService incorrectly resolves cache keys for Redis connector with useRedisSets set to false. The keys are resolved like this:
- `sets:namespace:<random-plugin-id>:<pluginId>:<client-key>`

### 👍 Expected behavior

The correct form with `useRedisSets` should look something like this:
- `sets:namespace:<pluginId>:<client-key>`

### 👎 Actual Behavior with Screenshots

Before is a result of SCAN command executes on our Redis instance (the base64 client key is trimmed, but the prefix is the important one):

![image](https://github.com/user-attachments/assets/f4c483ed-b616-464d-81c9-3d6633619b41)

### 👟 Reproduction steps

1. Configure redis connector for CacheService with useRedisSets set to false
2. Start the service
3. SSH into your Redis instance, notice that cache keys have format like `sets:namespace:<random-plugin-id>:<pluginId>:<client-key>` instead of `sets:namespace:<pluginId>:<client-key>`

### 📃 Provide the context for the Bug.

This is a bug in the @keyv/redis library that is used in Redis CacheService implementation.

Tickets in the keyv repo:
- https://github.com/jaredwray/keyv/issues/1174
- https://github.com/jaredwray/keyv/issues/1142

Because of this we have two options:
- using the `useRedisSets` option that causes the cache to grow infinitely
- deal with cache misses, because of the inconsistent cache key resolution

### 🖥️ Your Environment

```
OS:   Linux 6.8.0-48-generic - linux/x64
node: v20.17.0
yarn: 4.5.1
cli:  0.28.2 (installed)
backstage:  1.32.5

Dependencies:
  @backstage/app-defaults                                          1.5.12
  @backstage/backend-app-api                                       0.7.9, 1.0.1
  @backstage/backend-common                                        0.22.0, 0.23.3, 0.24.1, 0.25.0
  @backstage/backend-defaults                                      0.5.2
  @backstage/backend-dev-utils                                     0.1.5
  @backstage/backend-openapi-utils                                 0.2.0
  @backstage/backend-plugin-api                                    0.6.21, 0.7.0, 0.8.1, 1.0.1
  @backstage/backend-tasks                                         0.5.27, 0.6.1
  @backstage/backend-test-utils                                    1.0.2
  @backstage/catalog-client                                        1.7.1
  @backstage/catalog-model                                         1.7.0
[...]

PR #27716, opened 2024-11-19
Fixed cache key resolution for redis store with useRedisSets set to false

eipc16:
Resolves https://github.com/backstage/backstage/issues/27707.

Changes:
- bumped keyv to v5
- bumped @keyv/redis to v4
- bumped @keyv/memcache to v2

I had to also bump the keyv library to use the `useKeyPrefix` option. With provided configuration cache keys stay the same as when useRedisSets were enabled (<plugin-id>:<encoded-key>). I've added BREAKING to the changesets since this changes how the keys are built with `useRedisSets=false` option.

The keyv upgrade also forced the update of `@keyv/memcache` to v2


packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -139,12 +139,12 @@ export class CacheManager {
...
 
   private createRedisStoreFactory(): StoreFactory {
-    const KeyvRedis = require('@keyv/redis');
+    const KeyvRedis = require('@keyv/redis').default;
     let store: typeof KeyvRedis | undefined;
     return (pluginId, defaultTtl) => {
       if (!store) {
         store = new KeyvRedis(this.connection, {
-          useRedisSets: this.useRedisSets,
+          keyPrefixSeparator: ':',
```
eipc16, line 147: In v4 the library by default uses `::` as a separator, this should make sure that the keys do not change for default configuration

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -157,7 +157,7 @@ export class CacheManager {
         ttl: defaultTtl,
         store,
         emitErrors: false,
-        useRedisSets: this.useRedisSets,
+        useKeyPrefix: false,
```
eipc16, line 160: It's quiet confusing, but this option disables appending additional keyprefix. Without this we get doubled prefix (e.g. techdocs:techdocs:<key>).

This is described here:
- https://github.com/jaredwray/keyv/tree/main/packages/redis#migrating-from-v3-to-v4.

---

Also, the same configuration causes issues for memcache implementation (it has doubled prefix), but since the same happens on master I did not want to change it in this MR.

package.json
```diff
@@ -110,6 +110,8 @@
     "@spotify/eslint-plugin": "^15.0.0",
     "@spotify/prettier-config": "^15.0.0",
     "@techdocs/cli": "workspace:*",
+    "@types/cacheable-request": "^8.3.6",
+    "@types/memjs": "^1.3.3",
```
eipc16, line 114: This fixes broken yarn build:api-docs after keyv and @keyv/memcache upgrade.

- `@types/memjs` issue is probably caused by references to older @keyv/memcache package in pinned @backstage/backend-common package. I think it will be safe to remove after getting rid of @backstage/backend-common references.

- `@types/cacheable-request` is used by `got` package that is used by some other dependencies in the repo. This way the easiest way I found to fix it.

Would appreciate if you have any suggestions on how to handle it better :)

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -139,37 +139,39 @@ export class CacheManager {
   }
 
   private createRedisStoreFactory(): StoreFactory {
-    const KeyvRedis = require('@keyv/redis');
-    let store: typeof KeyvRedis | undefined;
+    const KeyvRedis = require('@keyv/redis').default;
+    const stores: Record<string, typeof KeyvRedis> = {};
```
eipc16, line 143: During initialization the Keyv instead sets the namespace on the provider adapter: https://github.com/jaredwray/keyv/blob/main/packages/keyv/src/index.ts#L181

This means that reusing a single adapter / connection is not possible for multiple plugins, because each new initialization would override the previous namespace.

This seems to be the case in older versions of the library as well: https://github.com/jaredwray/keyv/blob/2024-05-04/packages/keyv/src/index.js#L63. So I have no idea how it worked before regardless of the useRedisSets option :thinking:

package.json
Rugvip: as long as these aren't actually needed this is an alright place to put these deps to unbreak things, since it won't affect published packages

packages/backend-test-utils/package.json
```diff
@@ -53,8 +53,8 @@
     "@backstage/plugin-auth-node": "workspace:^",
     "@backstage/plugin-events-node": "workspace:^",
     "@backstage/types": "workspace:^",
-    "@keyv/memcache": "^1.3.5",
-    "@keyv/redis": "^2.5.3",
+    "@keyv/memcache": "^2.0.1",
+    "@keyv/redis": "^4.0.1",
     "@types/express": "^4.17.6",
```
Rugvip, line 58: are these deps actually used? could they be removed instead? I'd assume some code changes are needed otherwise

Rugvip (commented):
Great stuff, thank you! :tada: 

Just a tiny Q and then we can :shipit:

packages/backend-test-utils/package.json
eipc16: There are used here:
- memcache: https://github.com/backstage/backstage/blob/master/packages/backend-test-utils/src/cache/memcache.ts
- redis: https://github.com/backstage/backstage/blob/master/packages/backend-test-utils/src/cache/redis.ts

I don't think any code changes are needed since they are only used to create a test connection without any additional options. And there are tests covering these test connections like: https://github.com/backstage/backstage/blob/master/packages/backend-test-utils/src/cache/redis.test.ts

Rugvip (approved):
Alright, let's :shipit:!

Thank you! 🎉

tasiorek27:
hello @Rugvip @eipc16 it looks like after introducing this change we are facing connection interruptions when using Azure Redis Cache.
image

Based on the Redis metrics I can see that cache is used but there is a lot of interruptions. We've checked connectivity and we didn't detect any issues in this area.
image

our config wasn't change and is as below:
```  cache:
    store: redis
    connection: rediss://:${REDIS_CACHE_PRIMARY_KEY}@${REDIS_HOST_AND_PORT}
```

Do you have any suggestions?

PR #27718, opened 2024-11-19
compute deltas based on hashes just like full mutations

freben:
Compute add/update/remove splits of `'delta'` mutations from providers.

* Delta with all-new data:  10x faster
* Delta with the same data again leading to no changes: 100x faster
* Delta with 100% changes to existing entities: no change (still slow - opportunity for future improvement)

One of the reasons that this change was made, is that anything that does delta mutations - including very large incremental runs - used to exclusively be served as upsert calls internally. These happen one-by-one instead of in batches AND also invalidate the processing timer unconditionally, pushing them to the top of the queue. This basically made the `unprocessed_cache` column entirely ineffective for delta mutations.

One side effect of this is that a provider that tries to issue a delta mutation _of an entity that belongs to another provider_ will not get a conflict error thrown if the entity had no changes. This is very much a corner case though, and should not cause issues in practice.

Implementation-wise, it feels slightly icky. I would like to rewrite this better sometime, by for example batch dumping all additions (and their hashes) into a transaction-temporary table and then performing a single query that does the update atomically based on hash comparisons. But for now this is less work and easier to make work for all engines.

### DIFF
--- a/.changeset/itchy-rats-pump.md
+++ b/.changeset/itchy-rats-pump.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Compute deltas more efficiently, which generally leads to less wasted processing cycles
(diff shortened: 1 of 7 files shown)


Rugvip (approved):
:shipit:

Issue #26392, opened 2024-08-30
🐛 Bug Report: Search queries do not respect quotes


starryeyez024:
### 📜 Description

When executing a search in Backstage, it should respect quote marks which encapsulate a string, and only return results that match that full string. For example, if you search for "row your boat" it should only return entities with "row your boat" in the name or description. 
Also the same results are returned in the quick results modal as are shown on the full results page.

### 👍 Expected behavior

If you search for "digital acceleration team" it should only return entities with the full string "digital acceleration team" in the name or description.

### 👎 Actual Behavior with Screenshots

If you search for "digital acceleration team", it will return entities which only have "accelerate" in the name.
Adding quote marks around a sequence of words makes no difference.

<img width="1211" alt="no_quotes" src="https://github.com/user-attachments/assets/4d77e34f-8490-48b2-ae38-49ba14ecd0dc">
<img width="1245" alt="quotes" src="https://github.com/user-attachments/assets/0dbaef06-c201-400b-b3de-544aaad538f0">

### 👟 Reproduction steps

Try searching for a string with and without quotes and notice there is no distinction. 

### 🖥️ Your Environment

 Backstage | 1.28.2  

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

benjdlambert:
Ping @backstage/search-maintainers

githoober:
@starryeyez024  Did you end up somehow pinging @backstage/search-maintainers? I could not find how to do it myself.

awanlin:
Hi @starryeyez024, Ben was doing just that in his comment and you did just now. The team you just mentioned would get an email. As they are pretty small team they may not have been able to get to this. Let me see what I can do to get their input on this.

starryeyez024:
Thanks, I didn't realize I should do that.

awanlin:
No worries, I don't think it was expected of you @starryeyez024 👍 

While I have you what "search engine" are you using Postgres or Elasticsearch? I'm pretty sure supporting this for Postgres will require some changes specific to it's implementation.

starryeyez024:
We are using Postgresql.

githoober:
[...]

PR #27720, opened 2024-11-19
Update elasticsearch translator to handle phrase searches.

squid-ney:

Update the `ElasticSearchSearchEngine` translator to handle phrase queries. Partially resolves https://github.com/backstage/backstage/issues/26392

### DIFF
--- a/plugins/search-backend-module-elasticsearch/src/engines/ElasticSearchSearchEngine.ts
+++ b/plugins/search-backend-module-elasticsearch/src/engines/ElasticSearchSearchEngine.ts
@@ -248,6 +248,7 @@ export class ElasticSearchSearchEngine implements SearchEngine {
       });
 
     const esbQueries = [];
+    // https://regex101.com/r/Lr0MqS/1
     const phraseTerms = term.match(/"[^"]*"/g);
 
     if (isBlank(term)) {
--- a/plugins/search-backend-module-elasticsearch/src/engines/ElasticSearchSearchEngine.ts
+++ b/plugins/search-backend-module-elasticsearch/src/engines/ElasticSearchSearchEngine.ts
@@ -246,18 +246,39 @@ export class ElasticSearchSearchEngine implements SearchEngine {
           'Failed to add filters to query. Unrecognized filter type',
         );
       });
-    const esbQuery = isBlank(term)
-      ? esb.matchAllQuery()
-      : esb
-          .multiMatchQuery(['*'], term)
-          .fuzziness('auto')
-          .minimumShouldMatch(1);
+
+    const esbQueries = [];
+    const phraseTerms = term.match(/"[^"]*"/g);
+
+    if (isBlank(term)) {
+      const esbQuery = esb.matchAllQuery();
+      esbQueries.push(esbQuery);
+    } else if (phraseTerms && phraseTerms.length > 0) {
+      let restTerm = term;
+      for (const phraseTerm of phraseTerms) {
+        restTerm = restTerm.replace(phraseTerm, '');
+        const esbPhraseQuery = esb
+          .multiMatchQuery(['*'], phraseTerm.replace(/"/g, ''))
+          .type('phrase');
+        esbQueries.push(esbPhraseQuery);
+      }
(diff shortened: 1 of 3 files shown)


plugins/search-backend-module-elasticsearch/src/engines/ElasticSearchSearchEngine.ts
```diff
@@ -246,18 +246,39 @@ export class ElasticSearchSearchEngine implements SearchEngine {
...
+      esbQueries.push(esbQuery);
+    }
+
     const pageSize = query.pageLimit || 25;
     const { page } = decodePageCursor(pageCursor);
 
     let esbRequestBodySearch = esb
       .requestBodySearch()
-      .query(esb.boolQuery().filter(filter).must([esbQuery]))
+      .query(esb.boolQuery().filter(filter).should(esbQueries))
```
squid-ney, line 281: From [Lucene Boolean Clause docs](https://lucene.apache.org/core/5_0_0/core/org/apache/lucene/search/BooleanClause.Occur.html)
> public static final [BooleanClause.Occur](https://lucene.apache.org/core/5_0_0/core/org/apache/lucene/search/BooleanClause.Occur.html) SHOULD
Use this operator for clauses that should appear in the matching documents. For a BooleanQuery with no MUST clauses one or more SHOULD clauses must match a document for the BooleanQuery to match.

As long as one of the queries in the `esbQueries` match for a document, it will be included. I think this behavior is sensible. For example, if the search term is: `"foo bar" test` , then we would want the search result to include documents that match on either the phrase query or the regular term query or both.

plugins/search-backend-module-elasticsearch/src/engines/ElasticSearchSearchEngine.ts
```diff
@@ -246,18 +246,39 @@ export class ElasticSearchSearchEngine implements SearchEngine {
           'Failed to add filters to query. Unrecognized filter type',
         );
       });
-    const esbQuery = isBlank(term)
-      ? esb.matchAllQuery()
-      : esb
-          .multiMatchQuery(['*'], term)
-          .fuzziness('auto')
-          .minimumShouldMatch(1);
```
squid-ney, line 254: I don't think `minimumShouldMatch(1)` is needed. See the following explanation:
According to the [`multi_match` query docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#type-best-fields): 
(note: `best_fields` is the default type)
> "The best_fields type generates a [match query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html) for each field and wraps them in a [dis_max](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-dis-max-query.html) query, to find the single best matching field."

Looking at the [Disjunction max query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-dis-max-query.html):

> Returns documents matching one or more wrapped queries, called query clauses or clauses.

In my opinion, this makes the `minimumShouldMatch` parameter set to 1 redundant.

plugins/search-backend-module-elasticsearch/src/engines/ElasticSearchSearchEngine.ts
```diff
@@ -246,18 +246,39 @@ export class ElasticSearchSearchEngine implements SearchEngine {
...
+
+    const esbQueries = [];
+    const phraseTerms = term.match(/"[^"]*"/g);
```
johnphilip283, line 251:
```suggestion
    // https://regex101.com/r/Lr0MqS/1
    const phraseTerms = term.match(/"[^"]*"/g);
```
Whenever I have to use a regex, I always leave a comment to this website where you can see what groups are matched on a regex you've written - it helps with gaining context while reading / reviewing code!

squid-ney: Good tip, I'll add!

johnphilip283 (approved):

PR #27727, opened 2024-11-19
yarn-plugin: move Backstage version to parameters in bound descriptors

mtlewis:

The previous version of this code transformed the version range `backstage:^` to `backstage:<version-from-backstage.json>` when writing entries in yarn.lock. This does seem to be something that the yarn plugin API supports, but it turns out to cause issues in some third-party tools which assume that version ranges will be consistent between package.json and yarn.lock.

To fix this issue, we're now leveraging the descriptor binding feature in Backstage (which I didn't know about before) to append a descriptor parameter containing the Backstage version. This keeps the selector of the descriptors consistent between package.json and yarn.lock entries, which will hopefully mean that this plugin no longer breaks other tools. Given the similarity between the naming of the APIs, I suspect this is closer to the intended usage of the [Resolver#bindDescriptor](https://www.yarnpkg.cn/api/interfaces/yarnpkg_core.resolver.html#binddescriptor) method.

cf. https://github.com/snyk/nodejs-lockfile-parser/issues/256

### DIFF
--- a/packages/yarn-plugin/src/resolver/BackstageResolver.test.ts
+++ b/packages/yarn-plugin/src/resolver/BackstageResolver.test.ts
@@ -96,7 +96,7 @@ describe('BackstageResolver', () => {
 
   describe('bindDescriptor', () => {
     describe('with range "backstage:^"', () => {
-      it('returns a descriptor basedwith a version range for the current Backstage version', () => {
+      it('returns a descriptor with a version range for the current Backstage version', () => {
         expect(
           backstageResolver.bindDescriptor(
             structUtils.makeDescriptor(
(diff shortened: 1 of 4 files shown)


packages/yarn-plugin/src/resolver/BackstageResolver.test.ts
```diff
@@ -107,25 +107,25 @@ describe('BackstageResolver', () => {
         ).toEqual(
           structUtils.makeDescriptor(
             structUtils.makeIdent('backstage', 'core'),
-            'backstage:1.23.45',
+            'backstage:^::v=1.23.45',
```
mtlewis, line 110: The parameter name here is up for discussion - I figured using `v` to keep it brief was the best way to keep things readable, but happy to discuss alternatives!

freben:
@mtlewis Is this feature (expressing the versions like this) supported by all v4+ versions of yarn? Or is it something that now requires to be above a certain minor version of yarn

mtlewis:
Good callout Freben! I wouldn't have expected this to be a new feature but just to double check, I went back to the earliest version of yarn we support (4.1.1) and verified that the new version of the plugin still works.

PR #27735, opened 2024-11-20
Fixed grammar mistakes in the signin identity docs

AmbrishRamachandiran:
Fixed grammar mistakes in the signin identity docs


Rugvip (changes requested):
Sorry this doesn't seems like an obvious improvement to me

PR #27741, opened 2024-11-20
backend-defaults: export DefaultHttpAuthService

jhaals:
This pull request introduces the `DefaultHttpAuthService` class to allow for custom token extraction logic by supplying a custom function for extracting tokens from the incoming http request.


packages/backend-defaults/src/entrypoints/httpAuth/httpAuthServiceFactory.ts
```diff
@@ -71,23 +70,54 @@ type RequestWithCredentials = Request & {
...
+/**
+ * @public
+ * Options for creating a DefaultHttpAuthService.
+ */
+export interface DefaultHttpAuthServiceOptions {
+  auth: AuthService;
+  discovery: DiscoveryService;
+  pluginId: string;
+  // Optionally override the default token extraction logic
+  extractTokenFromRequest?: (req: Request) => string | undefined;
```
benjdlambert, line 82:
```suggestion
  getTokenFromRequestHandler?: (req: Request) => string | undefined;
```

```suggestion
  getTokenFromRequest?: (req: Request) => string | undefined;
```

```suggestion
  getToken?: (req: Request) => string | undefined;
```

Are any of these better?

packages/backend-defaults/report-httpAuth.api.md
```diff
@@ -3,9 +3,50 @@
...
+}
+
+// @public
+export interface DefaultHttpAuthServiceOptions {
+  // (undocumented)
+  auth: AuthService;
+  // (undocumented)
+  discovery: DiscoveryService;
+  // (undocumented)
+  extractTokenFromRequest?: (req: Request_2) => string | undefined;
```
freben, line 45: Maybe make it an object as the return type, that way we can put more contextual info in there if needed (what if it's actually not easy to know JUST by looking at the token value what service it originated from? for non-jwks etc)

packages/backend-defaults/src/entrypoints/httpAuth/httpAuthServiceFactory.ts
```diff
@@ -71,23 +70,54 @@ type RequestWithCredentials = Request & {
...
-class DefaultHttpAuthService implements HttpAuthService {
+/**
+ * @public
+ * Options for creating a DefaultHttpAuthService.
+ */
+export interface DefaultHttpAuthServiceOptions {
+  auth: AuthService;
+  discovery: DiscoveryService;
+  pluginId: string;
+  // Optionally override the default token extraction logic
```
Rugvip, line 81: Might as well make this a `/** */` doc while at it 😁 

Could you also document this new customizability at https://github.com/backstage/backstage/blob/master/docs/backend-system/core-services/http-auth.md#configuring-the-service?

jhaals:
Renamed argument to `getTokenFromRequest`. Updated it to return an optional token which provides more flexibility.

Rugvip (approved):
Let's :shipit:

Rugvip (changes requested):
Oh, nope, realized we still need `docs/*` 😅

docs/backend-system/core-services/http-auth.md
```diff
@@ -102,6 +102,45 @@ or [contribute](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.
 
 :::
 
+### Custom token extraction logic
+
+In some cases, you might want to customize how tokens are extracted from incoming requests. It might be that you want to extract tokens from a different location in the request. To support this you supply your own slightly modified httpAuth service. The `DefaultHttpAuthService` class is exported from the `@backstage/backend-defaults` package and it's `create` method can be used to pass in a custom `getTokenFromRequest` extraction function.
```
Parsifal-M, line 107:
```suggestion
In some cases, you might want to customize how tokens are extracted from incoming requests. It might be that you want to extract tokens from a different location in the request. To support this you can supply your own slightly modified httpAuth service. The `DefaultHttpAuthService` class is exported from the `@backstage/backend-defaults` package and it's `create` method can be used to pass in a custom `getTokenFromRequest` extraction function.
```

Sorry I am not trying to be pedantic 😅  it also does make sense without the `can`

jhaals: We _can_ add it 😄

docs/backend-system/core-services/http-auth.md
```diff
@@ -102,6 +102,45 @@ or [contribute](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.
 
 :::
 
+### Custom token extraction logic
+
+In some cases, you might want to customize how tokens are extracted from incoming requests. It might be that you want to extract tokens from a different location in the request. To support this you can supply your own slightly modified httpAuth service. The `DefaultHttpAuthService` class is exported from the `@backstage/backend-defaults` package and it's `create` method can be used to pass in a custom `getTokenFromRequest` extraction function.
```
Rugvip, line 107:
```suggestion
In some cases, you might want to customize how tokens are extracted from incoming requests. It might be that you want to extract tokens from a different location in the request. To support this you can supply your own slightly modified httpAuth service. The `DefaultHttpAuthService` class is exported from `@backstage/backend-defaults/httpAuth` and it's static `create` method can be used to pass in a custom `getTokenFromRequest` extraction function.
```

Rugvip (approved):
:shipit:

Issue #27742, opened 2024-11-20
🚀 Feature(scaffolder): Add `templateExecutePermission`


CptnFizzbin:
### 🔖 Feature description

With the current set of permissions, it's hard to filter the template list to ones that the user can actually run without also hiding them from the catalog as a whole. This feature adds a `scaffolder.template.execute' permission that policy writers can hook into to restrict who has access to a run a template.

This would also allow the template index page to dynamically filter templates to ones the user is actually able to run.

### 🎤 Context

At my company, we have some template actions that need to be restricted to a specific group, which can be done via the existing `actionExecutePermission`, however it still allows the user to attempt to run the template, but it just skips over those actions.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!
```[tasklist]
### Tasks
- [x] add new permission and rules
- [ ] (optimization) fetch templates via scaffolder router instead of catalog
- [ ] https://github.com/backstage/backstage/pull/27748
```

awanlin:
Please make sure to document this - https://backstage.io/docs/features/software-templates/authorizing-scaffolder-template-details - as this has created a large number of permissions you need to have just for a person to be able to run a template. 

Also, worth being aware of these changes and the discussion in the PR - https://github.com/backstage/backstage/pull/25969

CptnFizzbin:
Yep, that's another part I'm hoping to resolve as well. Possibly by merging the existing permissions into the new one.

And thanks for the link to the other PR, i'll have a look

CptnFizzbin:
Actually, that might be a good question to ask now:

Should the existing `templateParameterReadPermission` and `templateStepReadPermission` be deprecated in favour of a unified `templateExecutePermission`? What effects would that have?

awanlin:
Might be worth bringing this up at the Framework SIG on the 26th and/or the Scaffolder SIG on the 29th. Calendar is here: https://calendar.google.com/calendar/u/0/embed?src=c_qup9gbhn9sqpuao6trttd8mk5s@group.calendar.google.com

This would help coordinate this better IMO.

CptnFizzbin:
I'm not familiar with what the Framework or Scaffolder SIGs are.

[...]

PR #27748, opened 2024-11-20
feature(Scaffolder): add `templateExecutePermission`

CptnFizzbin:
Part of #27742

This PR adds a new permission for Scaffolder templates to allow policy writers to control execution of a template, and allow for the template index page to be filtered to only templates the user can execute.

The new permission is designed to supersede the `parameterRead` and `stepRead` permissions. These existing permissions prevent unauthorized use of an action or custom field, however they don't prevent a user from running a template. This can lead to errors in the template, and confusion from users (eg, template succeeded, but the github:publish action was blocked).

TODO:

- [ ] Optimize fetching templates and checking permissions
- [ ] Add extentionPoint for custom rules
- [ ] Add Documentation


CptnFizzbin:
Finished changes to the code, and just need to double check on Monday that the documentation I added for the new extension point works. In the meantime, I think this PR is ready for review.

plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
```diff
@@ -140,94 +140,14 @@ const isActionAuthorized = createConditionAuthorizer(
...
 
   constructor(private readonly options: NunjucksWorkflowRunnerOptions) {
     this.defaultTemplateFilters = createDefaultFilters({
       integrations: this.options.integrations,
     });
   }
 
-  private readonly tracker = scaffoldingTracker();
-
-  private isSingleTemplateString(input: string) {
```
CptnFizzbin, line 152: issue: need to revert reordered code

plugins/scaffolder-backend/src/service/router.test.ts
```diff
@@ -230,14 +235,12 @@ describe('createRouter', () => {
...
         .spyOn(permissionApi, 'authorizeConditional')
-        .mockImplementation(async () => [
-          {
-            result: AuthorizeResult.ALLOW,
-          },
-          {
-            result: AuthorizeResult.ALLOW,
-          },
-        ]);
+        .mockImplementation(async permissionQueries => {
```
CptnFizzbin, line 238: comment: these changes to the permissionApi mocks will likely conflict with #25969

plugins/scaffolder-react/src/next/components/TemplateGroups/useTemplateEntityList.ts
```diff
@@ -0,0 +1,52 @@
...
+import { stringifyEntityRef } from '@backstage/catalog-model';
+import { useApi } from '@backstage/frontend-plugin-api';
+import { useEntityList } from '@backstage/plugin-catalog-react';
+import { AuthorizeResult } from '@backstage/plugin-permission-common';
+import { permissionApiRef } from '@backstage/plugin-permission-react';
+import { templateExecutePermission } from '@backstage/plugin-scaffolder-common/alpha';
+import useSWR from 'swr';
+
+export const useTemplateEntityList = () => {
+  const { loading, error: listError, entities } = useEntityList();
```
CptnFizzbin, line 26: comment: in a future PR, entities will be fetched from the scaffolderApi instead of the catalogApi. The scaffolder endpoint will more or less act as a proxy to the catalogApi, and apply the `templateExecutePermission` check there.

Rugvip:
Just had a discussion around this with @benjdlambert, and an idea that came up is to rely on the existing catalog entity read permission instead. If the user doesn't have permission to read a template entity, they would also not be allowed to execute the template, and the template should not show up in the frontend either. I feel that would cover the high-level template permissions, and then we could look into improving the UX of the action permissions separately. What do you think?

CptnFizzbin:
For us, we still wanted the template to be visible in the catalog. It would allow for us to add cards to the entity page that would explain why the user isn't allowed to run the template, and who to contact to get permissions.

plugins/scaffolder-backend/src/deprecated.ts
```diff
@@ -84,3 +94,29 @@ export const fetchContents = fetchContentsNode;
  * @deprecated Import from `@backstage/plugin-catalog-backend-module-scaffolder-entity-model` instead
  */
 export const ScaffolderEntitiesProcessor = ScaffolderEntitiesProcessorModule;
+
+/**
+ * @public
+ * @deprecated Import from `@backstage/plugin-scaffolder-node/alpha` instead
+ */
```
benjdlambert, line 101:
```ts
import { myType as derpType } from '...';

/** @deprecated **/
export type TemplatePermissionRuleInput<..> = derpType<..>;
```

plugins/scaffolder-backend/src/service/index.ts
```diff
@@ -14,4 +14,31 @@
  * limitations under the License.
  */
```
benjdlambert, line 16: rename this file to `deprecated.ts`

plugins/scaffolder-backend/test-results/.last-run.json
```diff
@@ -0,0 +1,4 @@
+{
```
benjdlambert, line 1: Think we can drop this :pray:

plugins/scaffolder-common/report-alpha.api.md
```diff
@@ -12,12 +12,18 @@ export const actionExecutePermission: ResourcePermission<'scaffolder-action'>;
 // @alpha
 export const RESOURCE_TYPE_SCAFFOLDER_ACTION = 'scaffolder-action';
 
+// @alpha
+export const RESOURCE_TYPE_SCAFFOLDER_ENTITY = 'scaffolder-entity';
```
benjdlambert, line 16:
```suggestion
export const RESOURCE_TYPE_SCAFFOLDER_TEMPLATE_ENTITY = 'scaffolder-template-entity';
```

plugins/scaffolder-common/src/permissions.ts
```diff
@@ -123,6 +141,12 @@ export const templateManagementPermission = createPermission({
   attributes: {},
 });
 
+/**
+ * List of the scaffolder permissions that are associated with template entity.
+ * @alpha
+ */
+export const scaffolderEntityPermissions = [templateExecutePermission];
```
benjdlambert, line 148:
```suggestion
export const scaffolderTemplateEntityPermissions = [templateExecutePermission];
```

plugins/scaffolder-node/report-alpha.api.md
```diff
@@ -27,12 +50,307 @@ export type AutocompleteHandler = ({
...
+  TemplateParametersV1beta3 | TemplateEntityStepV1beta3,
+  {},
+  'scaffolder-template',
+  {
+    tag: string;
+  }
+>;
+
+// @public (undocumented)
+export const hasTaggedAction: PermissionRule<
```
benjdlambert, line 235: Don't think we should export these permission rules directly, use the namespace below.

CptnFizzbin:
@benjdlambert in the Scaffolder SIG, you had mentioned wanting a BEP for proposing a change to the scaffolder to pull templates from the Scaffolder API instead of the Catalog API, or was it for another issue?

benjdlambert:
@CptnFizzbin :wave: 

Thinking a BEP for more the proposal of this change really, but you can include that change to read from the Scaffolder API for grabbing templates which support the disabled use case in a nice way.

Higher level descriptions on the motivation behind this change, and then I can go ahead and add more more details around the changes to the `backstage:permissions` thing too seperately.

plugins/scaffolder-node/src/permissions/rules/scaffolderTemplateRules.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * @public
+ */
+export const createTemplatePermissionRule = makeCreatePermissionRule<
+  TemplateEntityStepV1beta3 | TemplateParametersV1beta3,
+  {},
+  typeof RESOURCE_TYPE_SCAFFOLDER_TEMPLATE
+>();
+
+const hasTag = createTemplatePermissionRule({
```
vinzscam, line 34: mmm what's the reason for moving all the rules over to `scaffolder-node`?

CptnFizzbin: ~~it's to follow prior work in the Catalog plugin, where rules and conditions are in the `-node` module~~

Annnd... double checking that, turns out I misread something at some point and they are *actually* in the `-backend` module. I'll move them back.

CptnFizzbin:
Note: I don't know who is supposed to resolve a comment on a PR here. Should the reviewer or should I resolve threads?

CptnFizzbin:
Test failure looks to be unrelated, and I'm unable to trigger a rerun

CptnFizzbin:
as pointed out by @vinzscam in the BEP proposal, there is an `actionExecutePermission` that also controls execution after template substitution for each step. The proposed `templateExecutePermission` is unable to run at that point since the nunjucks processor doesn't have the full context of the template, just the step that's being executed.

CptnFizzbin:
going to close this PR for now while I go back to the drawing board.

PR #27752, opened 2024-11-21
Fix api-reports space-after-comma bug

### LINKED ISSUE #27551: 🐛 Bug Report: Repo Tools - API Reports - Process swallowed if args are not correct
### 📜 Description

If you have a space in the command like this, the space is in the args being sent to the `-o` flag:

 `backstage-repo-tools api-reports -o ae-wrong-input-file-type, ae-undocumented --validate-release-tags`

The command will only output:

```text
# Compiling TypeScript
```

### 👍 Expected behavior

I "think" the behaviour in this case should be an error that the command is wrong and that perhaps saying it got unexpected arguments.

If you have a space in the command like this, the space is in the args being sent to the `-o` flag:

 `backstage-repo-tools api-reports -o ae-wrong-input-file-type, ae-undocumented --validate-release-tags`

The command will only output:

```text
# Compiling TypeScript
```

### 👟 Reproduction steps

This works best when you have changes that would need the API Reports to be updated.
In this repo or the Community Plugins repo update a `build:api-reports:only` script to pass args to the `-o` flag with a space between them like this:

`backstage-repo-tools api-reports -o ae-wrong-input-file-type, ae-undocumented --validate-release-tags`

Then run `yarn build:api-reports` notice that it does not generate a new API Report

**Notice:** the API Reports are not updated/generated

I made the mistake of adding a space when I was upgrading the [Azure DevOps plugins to `1.32.x`](https://github.com/backstage/community-plugins/pull/1563). As this does not cause any errors it was merged in. Then in follow up PRs where I expected the API Reports to change I [noticed this issue and fixed it](https://github.com/backstage/community-plugins/pull/1803).

This is problematic as it won't get picked up by the CI and you can easily move forward without these reported being updated.

### 🖥️ Your Environment

This repo or the Community Plugins repo

jescalada:

Fixes #27551 (error when executing `yarn build:api-reports` and `-o` flag has a space after the comma without double quote wrapping.

Steps to execute:

- Change the `build:api-reports:only` script in `package.json` to `NODE_OPTIONS=--max-old-space-size=8192 backstage-repo-tools api-reports -o ae-wrong-input-file-type, ae-undocumented --validate-release-tags`
- Execute `yarn build:api-reports`

Now, it should throw an error and stop the execution, as expected:

image

### DIFF
--- a/packages/repo-tools/src/commands/api-reports/api-reports.ts
+++ b/packages/repo-tools/src/commands/api-reports/api-reports.ts
@@ -45,7 +45,15 @@ export const buildApiReports = async (paths: string[] = [], opts: Options) => {
   const runTsc = opts.tsc;
   const allowWarnings = parseArrayOption(opts.allowWarnings);
   const allowAllWarnings = opts.allowAllWarnings;
-  const omitMessages = parseArrayOption(opts.omitMessages);
+  const omitMessagesRaw = opts.omitMessages;
+
+  if (omitMessagesRaw && omitMessagesRaw.endsWith(',')) {
+    throw new Error(
+      `Invalid value for --omit-messages: ${omitMessagesRaw}\nMust be a comma-separated list of strings without spaces or wrapped in quotations.`,
+    );
+  }
+
+  const omitMessages = parseArrayOption(omitMessagesRaw);
 
   const isAllPackages = !paths?.length;
   const selectedPackageDirs = await resolvePackagePaths({


packages/repo-tools/src/commands/api-reports/api-reports.ts
```diff
@@ -45,7 +45,15 @@ export const buildApiReports = async (paths: string[] = [], opts: Options) => {
   const runTsc = opts.tsc;
   const allowWarnings = parseArrayOption(opts.allowWarnings);
   const allowAllWarnings = opts.allowAllWarnings;
-  const omitMessages = parseArrayOption(opts.omitMessages);
+  const omitMessagesRaw = opts.omitMessages;
+
+  if (omitMessagesRaw && omitMessagesRaw.endsWith(',')) {
```
freben, line 50: Hm.

I mean the input was truly malformed; having a space there (in _any_ command line parsing system) means it was two separate arguments. If you'd passed in `-o "a, b"` with quotes, then I'd just say that `parseArrayOption` should instead be amended with a `.filter(Boolean)` to strip out accidental extra commas etc so that it's nicely forgiving. But if there's an unquoted space, the command correctly treats it as a folder argument to make reports for.

And THERE is actually the real bug.

```shell
yarn build:api-reports:only directoryThatDoesNotExist packages/cli
```

will happily generate reports for `packages/cli` but NOT complain about the first bad argument. If you could make a fix instead for that problem, it'd solve your problem as well as another really confusing thing that's been around forever. 🙏

Rugvip: 👍, I think it'd be reasonable to throw if a provided directory does not exist

jescalada:
@freben @Rugvip I got a new fix out which should catch not only malformed flags, but also invalid directories. I also updated the tests to cover those specific cases! As you mentioned before, malformed flags get parsed into `paths` instead.

freben (approved):
Looks good to me!

PR #27756, opened 2024-11-21
[Core] Enable structured request logging

camilaibs:

Log request and response metadata so they can be used for filtering.

image


packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.test.ts
```diff
@@ -44,7 +47,7 @@ describe('MiddlewareFactory', () => {
       warn: jest.fn(),
       error: jest.fn(),
       debug: jest.fn(),
-      child: () => childLogger,
+      child: jest.fn(() => childLogger),
```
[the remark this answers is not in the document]
camilaibs: Oh, so clean, thanks for sharing that 🙌🏻

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
```diff
@@ -137,18 +159,25 @@ export class MiddlewareFactory {
...
+        meta = getLogMeta(tokens, req, res);
+        return getLogMessage(tokens, req, res);
+      },
+      {
+        stream: {
+          write(message: string) {
+            const middlewareLogger = logger.child({
+              type: 'incomingRequest',
+              ...meta,
+            });
```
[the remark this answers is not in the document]
camilaibs: I didn't know that it is possible, looks much better now 🙂

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
```diff
@@ -45,6 +46,27 @@ import {
...
+) {
+  return {
+    date: tokens.date(req, res, 'clf') ?? '-',
+    method: tokens.method(req, res) ?? '-',
+    url: tokens.url(req, res) ?? '-',
+    httpVersion: tokens['http-version'](req, res) ?? '-',
+    status: tokens.status(req, res) ?? '-',
+    contentLength: tokens.res(req, res, 'content-length') ?? '-',
+    referrer: tokens.referrer(req, res) ?? '-',
+    userAgent: tokens.req(req, res, 'user-agent') ?? '-',
```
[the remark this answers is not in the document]
camilaibs: Good point, I used the same default value that Morgan uses for undefined token values, but I agree with you 🙂 . Please take a look on the last change I made.
Regarding what keys to add, I'm adding the same keys we already have in the message, but I think referrer is a important one to track the request origin.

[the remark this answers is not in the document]
camilaibs: We have been using user-agent to know if that is a service-to-service request, maybe it is not ideal.

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.test.ts
```diff
@@ -27,33 +27,23 @@ import express from 'express';
 import createError from 'http-errors';
 import request from 'supertest';
 import { MiddlewareFactory } from './MiddlewareFactory';
-import { ConfigReader } from '@backstage/config';
+import { mockServices } from '@backstage/backend-test-utils';
+
+jest.useFakeTimers();
+jest.setSystemTime(new Date('2024-11-20T00:00:00Z'));
```
freben, line 33: can also do `jest.useFakeTimers({ now: new Date('...') })`

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.test.ts
```diff
@@ -254,5 +244,57 @@ describe('MiddlewareFactory', () => {
...
+
+      await request(app).get('/').expect(200);
+
+      expect(logger.info).toHaveBeenCalledWith(
+        expect.stringContaining(
+          '[20/Nov/2024:00:00:00 +0000] "GET / HTTP/1.1" 200 11 "-" "-"',
+        ),
+        {
+          type: 'incomingRequest',
+          date: '20/Nov/2024:00:00:00 +0000',
```
freben, line 259: Hm can we get UTC datetimes? Also, this one is maybe not useful since every emitted log entry has a timestamp itself, right?

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.test.ts
```diff
@@ -254,5 +244,57 @@ describe('MiddlewareFactory', () => {
...
+          method: 'GET',
+          url: '/',
+          status: '200',
```
freben, line 262: could this be a number?

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
```diff
@@ -137,18 +159,26 @@ export class MiddlewareFactory {
...
-      '[:date[clf]] ":method :url HTTP/:http-version" :status :res[content-length] ":referrer" ":user-agent"';
-    return morgan(customMorganFormat, {
-      stream: {
-        write(message: string) {
-          logger.info(message.trimEnd());
+    const logger = this.#logger;
+    let meta: Record<string, string | undefined> = {};
+    return morgan(
+      (tokens: TokenIndexer, req: IncomingMessage, res: ServerResponse) => {
+        meta = getLogMeta(tokens, req, res);
```
freben, line 166: i gotta ask - is this a common pattern? it's a bit unexpected to mutate external state and then reuse it inside the stream thing. Feels like we can't for sure know that the stream is called exactly immediately after the first function? Could it end up being the case that the stream sends the wrong, old, meta?

Extreme thought ... what if we returned `JSON.stringify({ message: ..., a, b, c })` from the first function and then deserialize it in the stream? ;) Or is that just too crazy

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.test.ts
camilaibs: I converted to number, people probably expect that 🙂

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.test.ts
camilaibs: Yes we can, just updated that.

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
camilaibs: 🤔 , not sure if it is a common patter as I'm doing it for the first time, but agree that we don't have control of when the stream method will be called. I updated the code with your suggest 🙌🏻

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.test.ts
camilaibs: Nice, code update 🙌🏻

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.test.ts
```diff
@@ -254,5 +241,57 @@ describe('MiddlewareFactory', () => {
...
+          '[20/Nov/2024:00:00:00 +0000] "GET / HTTP/1.1" 200 11 "-" "-"',
+        ),
+        {
+          type: 'incomingRequest',
+          date: 'Wed, 20 Nov 2024 00:00:00 GMT',
+          method: 'GET',
+          url: '/',
+          status: 200,
+          httpVersion: '1.1',
+          contentLength: '11',
```
freben, line 263: what about thisssss as a number :) haha, sorry i missed it the first time around

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.test.ts
```diff
@@ -254,5 +241,57 @@ describe('MiddlewareFactory', () => {
...
+
+      await request(app).get('/').expect(200);
+
+      expect(logger.info).toHaveBeenCalledWith(
+        expect.stringContaining(
+          '[20/Nov/2024:00:00:00 +0000] "GET / HTTP/1.1" 200 11 "-" "-"',
+        ),
+        {
+          type: 'incomingRequest',
+          date: 'Wed, 20 Nov 2024 00:00:00 GMT',
```
freben, line 258: i don't see the conversion being done - i meant, if we keep this at all (i argue that maybe we should just not have this at all), maybe it should have been ISO format `2024-11-20T00:00:00Z` so that it's standard formatted and easily parseable

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
```diff
@@ -45,6 +46,28 @@ import {
...
+  req: IncomingMessage,
+  res: ServerResponse,
+) {
+  const status = Number(tokens.status(req, res));
+  return {
+    date: tokens.date(req, res, 'web'),
+    method: tokens.method(req, res),
+    url: tokens.url(req, res),
+    httpVersion: tokens['http-version'](req, res),
+    status: isNaN(status) ? undefined : status,
```
freben, line 64:
```suggestion
    status: isFinite(status) ? undefined : status,
```

even more strict!

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
```diff
@@ -137,18 +160,27 @@ export class MiddlewareFactory {
...
+        const message = getLogMessage(tokens, req, res);
+        return JSON.stringify({ meta, message });
+      },
+      {
+        stream: {
+          write(json: string) {
+            const { meta, message } = JSON.parse(json);
+            logger.info(message.trimEnd(), {
+              type: 'incomingRequest',
+              ...Object.entries(meta).reduce((rest, [key, value]) => {
```
freben, line 176: it's actually the case that `JSON.stringify` throws away keys that have an undefined value, so you can skip the entire reduce.

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.test.ts
camilaibs: Oh I think your are suggesting iso
https://github.com/expressjs/morgan/blob/efd6bff1adae9418f12a8827614336c72e2c4300/index.js#L273

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.test.ts
camilaibs: 🙌🏻

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
camilaibs: ✅

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
camilaibs: Ah-a, fixed 🙏🏻

freben (approved):

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
```diff
@@ -45,6 +46,29 @@ import {
...
+
+function getLogMeta(
+  tokens: TokenIndexer,
+  req: IncomingMessage,
+  res: ServerResponse,
+) {
+  const status = Number(tokens.status(req, res));
+  const contentLength = Number(tokens.res(req, res, 'content-length'));
+  return {
+    date: tokens.date(req, res, 'iso'),
```
Rugvip, line 61: Should we skip the date perhaps? or is it useful somehow even if we already have the log timestamp?

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
```diff
@@ -137,18 +161,25 @@ export class MiddlewareFactory {
...
-    return morgan(customMorganFormat, {
-      stream: {
-        write(message: string) {
-          logger.info(message.trimEnd());
+    const logger = this.#logger;
+    return morgan(
+      (tokens: TokenIndexer, req: IncomingMessage, res: ServerResponse) => {
+        const meta = getLogMeta(tokens, req, res);
+        const message = getLogMessage(tokens, req, res);
+        return JSON.stringify({ meta, message });
```
Rugvip, line 169: Hmm, this seems unfortunate, surely there must be a way to wire this up so that we don't this?

Rugvip (commented):
Nice! 👍

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
Rugvip: Actually, I realized that this basically does half the work of formatting the message already, I think it's worth expanding these `tokens.*` calls, construct the message ourselves, and get rid of `morgan` altogether.

camilaibs: I'll give it a try

camilaibs: I removed morgan usage, ptal

.changeset/five-goats-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-defaults': patch
+---
+
+Log request and response metadata as logging fields so they can be used for filtering log messages.
```
Rugvip, line 5: let's mention the change in date format too

Rugvip (approved):
yep I think that's it :+1:

freben (approved):

PR #27766, opened 2024-11-21
events-node: clarify purpose of subscriber ID

Rugvip:
🧹


freben (approved):
Ah thanks, that helped

pjungermann (approved):
looks good to me, thank you!

PR #22089, opened 2024-01-03
ADR: Align on Zod version to safely support exposing Zod schemas from plugins

HHogg:
## Hey, I just made a Pull Request!

.. to add an ADR around Zod. 

#### :heavy_check_mark: Checklist

- [X] All your commits have a `Signed-off-by` line in the message. ([more info](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md#developer-certificate-of-origin))

PR #27771, opened 2024-11-21
Move Scaffolder API to OpenAPI

### LINKED ISSUE #23819: 🚀 Feature: Backend client for Scaffolder
### 🔖 Feature description

A Scaffolder backend client, similar to [`catalog-client`](https://github.com/backstage/backstage/blob/629612dd11636ff82cce943e727e608b8bd0d54a/packages/catalog-client/README.md), that can be imported and used on the backend.

### 🎤 Context

We're looking to trigger (and possibly expose via APIs) scaffolding jobs on the backend, and would like to be able to do that using a Scaffolder backend client. Another useful purpose for this is being able to trigger automated tests to test software template changes.

This can possibly be approached in a way similar to `catalog-client` and start by creating an OpenAPI spec for Scaffolder APIs.

solimant:

Closes #23819

Server and client generated using:
```
yarn backstage-repo-tools package schema openapi generate --server --client-package plugins/scaffolder-common
```


plugins/scaffolder-backend/src/schema/openapi.yaml
```diff
@@ -0,0 +1,1055 @@
...
+          description: |
+            The output is an object where template authors can pull out information from template actions and return them in a known standard way.
+        templateInfo:
+          $ref: '#/components/schemas/TemplateInfo'
+          description: Some information about the template that is stored on the task spec.
+        user:
+          type: object
+          properties:
+            entity:
+              $ref: '../../../catalog-backend/src/schema/openapi.yaml#/components/schemas/UserEntity'
```
Rugvip, line 517: To be honest I think it's best to keep this one opaque. As an adopter of Backstage you can extend the `UserEntity` schema anyway, so there's kinda no point aiming to model this imo.

Rugvip (commented):
Thank you for looking at this! 👍 

Is the goal to wire up the router etc with this as well in this PR? I'd really be needed to ensure that the schema is correct and remains so.

plugins/scaffolder-backend/src/schema/openapi.yaml
solimant: Fair point. I'll adjust here, and remove the addition from the Catalog OpenAPI schema.

solimant (commented):
> Is the goal to wire up the router etc with this as well in this PR? I'd really be needed to ensure that the schema is correct and remains so.

@Rugvip - Yes, that's the goal, and I'm working on making more commits for that. And, yes, the schema needs to be correct 100%. That's the reason why I started with that since everything else is pretty much a matter of code generation.

solimant:
@aramissennyeydd - I'm cleaning up a few things, and I noticed that we've got some schemas defined in OpenAPI and the corresponding generated TypeScript type, and also defined separately (manually) in other places. For example, `Entity` defined in [`packages/catalog-model/src/entity/Entity.ts`]( https://github.com/solimant/backstage/blob/93cc245ddc4e371812424bafe544d62bcd34f54e/packages/catalog-model/src/entity/Entity.ts#L28) and in [`plugins/catalog-backend/src/schema/openapi/generated/models/Entity.model.ts`](https://github.com/solimant/backstage/blob/93cc245ddc4e371812424bafe544d62bcd34f54e/plugins/catalog-backend/src/schema/openapi/generated/models/Entity.model.ts#L27). Is that intentional?

solimant:
@benjdlambert - On a separate note, I'm seeing an issue with properties that have special characters, such as `backstage:permissions` and `EXPERIMENTAL_recovery`, where the generated TypeScript type is omitting those special characters, which seems related to [OpenAPITools/openapi-generator/issues/15261](https://github.com/OpenAPITools/openapi-generator/issues/15261). How do we feel about manually adjusting those properties to match?

solimant:
@Rugvip - because I've come across a few Scaffolder types that use `JsonObject` and similar types from `@backstage/types` (e.g. [`TaskStep`](https://github.com/solimant/backstage/blob/93cc245ddc4e371812424bafe544d62bcd34f54e/plugins/scaffolder-common/src/TaskSpec.ts#L76)), how do you feel about defining the types in `@backstage/types` in OpenAPI spec, and then repurpose `*.ts` files under [`packages/types/src](https://github.com/solimant/backstage/tree/93cc245ddc4e371812424bafe544d62bcd34f54e/packages/types/src) to export the generated TypeScript types? The benefit here is that we can cross-reference those common types in other OpenAPI spec YAMLs across the application.

CC: @benjdlambert, @aramissennyeydd

aramissennyeydd:
@solimant 

> For example, Entity defined in [packages/catalog-model/src/entity/Entity.ts](https://github.com/solimant/backstage/blob/93cc245ddc4e371812424bafe544d62bcd34f54e/packages/catalog-model/src/entity/Entity.ts#L28) and in [plugins/catalog-backend/src/schema/openapi/generated/models/Entity.model.ts](https://github.com/solimant/backstage/blob/93cc245ddc4e371812424bafe544d62bcd34f54e/plugins/catalog-backend/src/schema/openapi/generated/models/Entity.model.ts#L27). Is that intentional?

Yup this is intentional, the duplication here is minimal and the OpenAPI version should be looser than the catalog one.

> @Rugvip - because I've come across a few Scaffolder types that use JsonObject and similar types from @backstage/types (e.g. [TaskStep](https://github.com/solimant/backstage/blob/93cc245ddc4e371812424bafe544d62bcd34f54e/plugins/scaffolder-common/src/TaskSpec.ts#L76)), how do you feel about defining the types in @backstage/types in OpenAPI spec, and then repurpose *.ts files under [`packages/types/src](https://github.com/solimant/backstage/tree/93cc245ddc4e371812424bafe544d62bcd34f54e/packages/types/src) to export the generated TypeScript types? The benefit here is that we can cross-reference those common types in other OpenAPI spec YAMLs across the application.

Similarly here, I don't think we gain much if we deduplicate the types and could start running into issues where we expect x version of the shared types and get y.

> @benjdlambert - On a separate note, I'm seeing an issue with properties that have special characters, such as backstage:permissions and EXPERIMENTAL_recovery, where the generated TypeScript type is omitting those special characters, which seems related to https://github.com/OpenAPITools/openapi-generator/issues/15261. How do we feel about manually adjusting those properties to match?

I can take a look into a fix here - in the meantime, can you just add these as an `additionalProperties: {type: string}` to allow them on requests and responses. That will give us an initial bit of schema definition and we can add them back once this is fixed.

plugins/scaffolder-backend/src/schema/openapi.yaml
```diff
@@ -0,0 +1,1057 @@
...
+                - type: string
+                - type: boolean
+            backstage:permissions:
+              $ref: '#/components/schemas/TemplatePermissionsV1beta3'
+          required:
+            - action
+      description: Step that is part of a Template Entity.
+    TemplateEntityV1beta3:
+      allOf:
+        - $ref: '../../../catalog-backend/src/schema/openapi.yaml#/components/schemas/Entity'
```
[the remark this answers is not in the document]
aramissennyeydd: nope - just copy this over. I haven't figured out the stitching between specs yet. This kind of route may fail if you were you install this package on your local instance.

solimant:
@aramissennyeydd - for the Scaffolder client, I'm thinking to create the following:
1. Create `plugins/scaffolder-common/src/ScaffolderClient.ts` where we `export class ScaffolderClient implements ScaffolderApi {...}`
2. Generate the client code under `plugins/scaffolder-common/client/`

What do you think? Or should we create a new package called `packages/scaffolder-client` similar to `packages/catalog-client` in structure?

solimant:
I should've clarified the part about `ScaffolderClient`. What I meant is to relocate `plugins/scaffolder/src/api.ts` to `plugins/scaffolder-common/src/ScaffolderClient.ts`, and also `plugins/scaffolder-react/src/api/types.ts` to `plugins/scaffolder-common/src/api.ts`.

aramissennyeydd:
@solimant The `ScaffolderClient` idea makes sense, we should just use whatever folder the `yarn backstage-repo-tools package schema openapi generate --client <folder>` outputs to (I believe `/client`).

`scaffolder-common` also makes sense, and we can re-export from `scaffolder-react` so it isn't breaking. For the `CatalogClient`, I kept the same API interface for the handwritten client, but changed out the internals to use the generated client - we can probably do the same here. Also helps to ensure the types match up fully.

solimant:
Please remove the "stale" label.

awanlin:
Hi @solimant, the best way to avoid this going stale while you wait for reviews is to rebase. Your branch is currently 690 commits behind.

plugins/scaffolder-node/src/tasks/types.ts
```diff
@@ -37,7 +37,8 @@ export type TaskStatus =
   | 'completed'
   | 'failed'
   | 'open'
-  | 'processing';
+  | 'processing'
```
solimant, line 40: I feel like the types in this file could also be relocated `@backstage/plugin-scaffolder-common`..? For example, `TaskStatus` is pretty much the same as `ScaffolderTaskStatus`.

solimant (commented):
@aramissennyeydd - I've made the necessary changes for code generation, and some refactoring for existing code. Please have a look and let me know what you think. I've also posted some comments where I'd appreciate feedback. Happy to address any requests. Thank you 🙏🏼

plugins/scaffolder/src/api.ts
```diff
@@ -14,377 +14,12 @@
...
-
-    if (!response.ok) {
-      throw await ResponseError.fromResponse(response);
-    }
-
-    const { results } = await response.json();
-    return { results };
-  }
-}
+export class ScaffolderClient extends _ScaffolderClient {}
```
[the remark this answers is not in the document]
aramissennyeydd: We could probably just do a `export {ScaffolderClient} from '@backstage/plugin-scaffolder-common'` here instead. Example, https://github.com/backstage/backstage/blob/2d49ad89ba39fd78b0c44aa19386c41698274d2e/plugins/auth-backend/src/providers/oauth2-proxy/index.ts#L18-L24

plugins/scaffolder-backend/src/schema/openapi.yaml
```diff
@@ -0,0 +1,1319 @@
...
+      description: Construct a type with a set of properties K of type T
+    ScaffolderScaffoldOptions:
+      type: object
+      properties:
+        templateRef:
+          type: string
+        values:
+          type: object
+          additionalProperties:
+            $ref: '#/components/schemas/JsonValue'
```
aramissennyeydd, line 403: By default, all values of a payload should be JSON encode-able, we could probably just do `additionalProperties: true` here.

plugins/scaffolder-backend/src/schema/openapi.yaml
```diff
@@ -0,0 +1,1319 @@
...
+        secrets:
+          type: object
+          additionalProperties:
+            type: string
+      required:
+        - templateRef
+        - values
+      description: The input options to the `scaffold` method of the `ScaffolderClient`.
+    Schema:
```
aramissennyeydd, line 412: We can probably skip this for now, we shouldn't be trying to 100% document the API to start. We should be aiming for broad strokes to make sure that the shape makes sense for the Scaffolder and then we can expand as needed. A good example here is the catalog, most of the entity request/response types are just `JsonObject`.

plugins/scaffolder-backend/src/schema/openapi.yaml
```diff
@@ -0,0 +1,1319 @@
...
+          additionalProperties:
+            type: string
+        - type: object
+          properties:
+            backstageToken:
+              type: string
+      description: TaskSecrets
+    TaskSpec:
+      $ref: '#/components/schemas/TaskSpecV1beta3'
+    TaskSpecV1beta3:
```
aramissennyeydd, line 642: See my comment above, we can skip this for now

aramissennyeydd (commented):
Thanks for this! I think we should skip the `Template` entity modeling, that should remove the issue with `EXPERIMENTAL_` and `backstage:permissions` and align this OpenAPI spec with the catalog spec. It's adding a bunch of complexity to this spec and means maintaining those definitions in a few places, if there's desire to add this later we can revisit in a separate PR 🙂 .

A bit more context on why some of these things should stay at the application layer and not live in the spec in this PR, https://github.com/backstage/backstage/pull/22089.

packages/template-openapi-plugin-client/generator/src/main/java/org/openapitools/codegen/TypeScriptBackstageGenerator.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.openapitools.codegen.*;
+import org.openapitools.codegen.languages.TypeScriptClientCodegen;
+import org.openapitools.codegen.model.*;
+import io.swagger.models.properties.*;
+
+import java.util.*;
+import java.io.File;
+
+public class TypeScriptBackstageGenerator extends TypeScriptClientCodegen implements CodegenConfig {
```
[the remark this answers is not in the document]
aramissennyeydd: I don't love adding this yet, I'd like to try skipping the classes that need special treatment for now and seeing how far that gets us.

plugins/scaffolder/src/api.ts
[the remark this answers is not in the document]
solimant: I had attempted to do that, but reverted because I didn't get a nice IntelliSense experience for the deprecation. So, if I do this:
```ts
export {
  /**
   * @deprecated use import from {@link @backstage/plugin-scaffolder-common#ScaffolderClient} instead as this has now been moved.
   */
  ScaffolderClient 
} from '@backstage/plugin-scaffolder-common';
```
Anywhere that imports  `ScaffolderClient` from `@backstage/plugin-scaffolder-react` will show up like this in VS Code, which is fine:
image
The problems are:
1. it doesn't mention the `@deprecated` comment, which it would otherwise show like this:
   image

2. ⌘+clicking the `ScaffolderClient` takes me to the export from `@backstage/plugin-scaffolder-common`:
   image
   image
as opposed to taking me to the deprecated export, like so:
   image

If I'm being too picky here, I'm happy to match the example you shared.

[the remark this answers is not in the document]
aramissennyeydd: Hmm, can you try https://github.com/backstage/backstage/blob/821f065ee9dc781de74c1cfa7ce38b751f1e35b2/packages/backend-common/src/deprecated/index.ts#L119-L130? That's the only example I have of re-exporting and marking as deprecated.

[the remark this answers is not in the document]
solimant: Thank you for looking into it and finding that example. What this will do, though, is that it will keep the old implementation of `ScaffolderClient`. How about this example?
https://github.com/solimant/backstage/blob/104f83587b922fadbcae22396ea03398e0b63f8e/plugins/catalog-backend-module-github/src/deprecated.ts#L33-L47
It uses the same pattern of having the deprecated type extend the new type.

solimant:
@aramissennyeydd - thank you very much for the review 🙏🏼 I've taken care of the `Template` entity modeling changes you've mentioned. Please let me know if you'd like me to tone down other places.

plugins/scaffolder-common/src/api.ts
```diff
@@ -0,0 +1,285 @@
...
+
+/**
+ * A single action example
+ *
+ * @public
+ */
+export type ActionExample = {
+  description: string;
+  example: string;
+};
```
aramissennyeydd, line 68: Are we able to pull these from the spec as well?

aramissennyeydd (commented):
Looks a lot better 😁, I think the next step is to get CI passing. 

We should also add inline Jest schema testing for more confidence in the changeset. You can follow the catalog as an example for how to do that, https://github.com/backstage/backstage/blob/768e84d5e94d43432b9b59bebe8e7acbbebd9ca4/plugins/catalog-backend/src/service/createRouter.test.ts#L96.

plugins/scaffolder-common/src/api.ts
solimant: These two specifically? Or other API types in [`plugins/scaffolder-common/src/api.ts`](https://github.com/backstage/backstage/pull/27771/files/233b9e97f969c84194c1707f2d037f04b2d2657c#diff-1b5ad46439cb89690c8103b84f339965bc7fbb60b08f33f66e347ebe1e78b26f) as well?

solimant (commented):
Thank you very much, @aramissennyeydd 🙏🏼 I'll work on CI and the inline Jest schema testing.
I was also hoping to add a `ScaffolderService` and a `scaffolderServiceRef`, similar to [`CatalogService`](https://github.com/backstage/backstage/blob/3fb2f7c69077d75cc4dcf37e9c9e255b53264a15/plugins/catalog-node/src/catalogService.ts#L58) and [`catalogServiceRef`](https://github.com/backstage/backstage/blob/3fb2f7c69077d75cc4dcf37e9c9e255b53264a15/plugins/catalog-node/src/catalogService.ts#L303). Although, it might be more appropriate to have them in their own PR. What do you think? If you're fine either way, I already have the changes and can push them here for review.

plugins/scaffolder-common/src/api.ts
aramissennyeydd: The other types in this file as well :)

aramissennyeydd:
@solimant I think that makes sense to add in a separate PR, the core maintainers would be the ones to take a look at that PR.

plugins/scaffolder-common/src/api.ts
solimant: Sounds good 👍🏼 For types in `scaffolder-node`, such as [`TaskStatus`](https://github.com/backstage/backstage/blob/233b9e97f969c84194c1707f2d037f04b2d2657c/plugins/scaffolder-node/src/tasks/types.ts#L35) and [`SerializedTask`](https://github.com/backstage/backstage/blob/233b9e97f969c84194c1707f2d037f04b2d2657c/plugins/scaffolder-node/src/tasks/types.ts#L55), should we keep those defined there? Or also pull them from the spec through `scaffolder-common`?

solimant: @aramissennyeydd - I couldn't pull beyond `ActionExample` and `ScaffolderTaskStatus` because most of the other types are using a little more "involved" types. For example, `ScaffolderTask` uses `TaskSpec` (which we're excluding from the spec), `Action` uses `JSONSchema7`, `ScaffolderScaffoldOptions` uses `JsonValue`, etc.
I'm happy to still move forward with pulling these types from the spec. Just wanted to confirm our direction.

aramissennyeydd: Let's keep the more complex ones here, any of the simple types would be great to pull directly from the spec though :)

plugins/scaffolder-backend/src/schema/openapi.yaml
```diff
@@ -0,0 +1,1342 @@
+openapi: 3.0.3
```
solimant, line 1: Will clean up this file once everything is finalized.

packages/backend-openapi-utils/src/schema/response-body-validation.ts
```diff
@@ -69,15 +69,18 @@ export class ResponseBodyParser
       const jsonContentType = Object.keys(contentTypes).find(contentType =>
         contentType.split(';').includes('application/json'),
       );
-      if (!jsonContentType) {
+      const eventStreamContentType = Object.keys(contentTypes).find(
```
solimant, line 72: Had to adjust this a little to accommodate for event-stream content type.

plugins/scaffolder-common/src/ScaffolderClient.ts
```diff
@@ -0,0 +1,418 @@
...
+            secrets: request.secrets,
+            directoryContents: request.directoryContents,
+          },
+        },
+        options,
+      ),
+    );
+  }
+
+  private streamLogsEventStream({
```
solimant, line 249: I couldn't use the generated API client for this one because it's not generic enough -- it uses `fetchEventSource` from `@microsoft/fetch-event-source`.

solimant (commented):
@aramissennyeydd - I've managed to fix CI, and I think this is at a good state for another review. Please let me know if you see anything that's odd, and I'll be happy to address it. I've left some comments below for feedback.

Also, should I take care of refactoring deprecated type references so we're using the new types instead?

plugins/scaffolder-backend/src/schema/openapi.yaml
```diff
@@ -0,0 +1,1342 @@
...
+      responses:
+        '200':
+          description: Ok
+          content:
+            text/event-stream:
+              schema:
+                type: string
+      parameters:
+        - $ref: '#/components/parameters/eventsAfter'
+        - $ref: '#/components/parameters/taskId'
```
aramissennyeydd, line 1168: How do you feel about skipping this for now and adding in a followup? Everything else is looking good in this PR, but it looks like OpenAPI generator support for SSE is shaky at best and worth a separate discussion. related to https://github.com/backstage/backstage/pull/27771#discussion_r1900277890 and https://github.com/backstage/backstage/pull/27771#discussion_r1900275747

aramissennyeydd (commented):
Looking good! Let's skip the SSE endpoint for now - happy to discuss in a followup PR, but I don't want to hold up review here for it. I'd much rather have 1 undocumented route that have a completely missing spec :)

Re: deprecated types, let's skip that for now as well, there's already a bunch of moving parts here that have the potential to cause unintended issues.

plugins/scaffolder-common/src/ScaffolderClient.test.ts
```diff
@@ -359,21 +365,18 @@ describe('api', () => {
 
     it('should list tasks with limit and offset', async () => {
       server.use(
-        rest.get(
-          `${mockBaseUrl}/v2/tasks?limit=5&offset=0`,
```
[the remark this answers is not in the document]
solimant: I made the change because this warning was being thrown during `yarn test`:
```
  console.warn
    [MSW] Found a redundant usage of query parameters in the request handler URL for "GET http://backstage/api/v2/tasks?limit=5&offset=0". Please match against a path instead and access query parameters in the response resolver function using "req.url.searchParams".

      366 |     it('should list tasks with limit and offset', async () => {
      367 |       server.use(
    > 368 |         rest.get(
          |              ^
      369 |           `${mockBaseUrl}/v2/tasks?limit=5&offset=0`,
      370 |           (_req, res, ctx) => {
      371 |             return res(

      at Object.warn (../../../node_modules/msw/lib/index.js:370:11)
      at RestHandler.warn [as checkRedundantQueryParameters] (../../../node_modules/msw/lib/index.js:1224:14)
      at new checkRedundantQueryParameters (../../../node_modules/msw/lib/index.js:1208:10)
      at Object.get (../../../node_modules/msw/lib/index.js:2217:12)
      at Object.get (ScaffolderClient.test.ts:368:14)
```

But, sure, I'll revert, although I don't think this test (and likely some others) are written correctly. `listTasks` is supposed to return an object with:
```ts
{
  tasks: ScaffolderTask[];
  totalTasks?: number;
}
```
But the mock is returning an array:
```ts
[
  {
    createdBy: null,
  },
  {
    createdBy: null,
  },
]
```

plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -681,6 +681,7 @@ export async function createRouter(
       await taskBroker.retry?.(taskId);
       res.status(201).json({ id: taskId });
     })
+    // @ts-ignore - Skipping SSE for now
```
solimant, line 684: @aramissennyeydd - I've added this `ts-ignore` because after [skipping SSE](https://github.com/backstage/backstage/pull/27771/commits/cdc271c4b9211b730398d6a2406077c6c3cd104f), we're now missing the generated type for SSE from the [`EndpointMap`](https://github.com/backstage/backstage/pull/27771/commits/cdc271c4b9211b730398d6a2406077c6c3cd104f). Is there a better?

plugins/scaffolder-common/src/ScaffolderClient.test.ts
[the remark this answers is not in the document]
aramissennyeydd: Ah, yeah, that seems like a pre-existing error, definitely something to look at after this goes in 😁

plugins/scaffolder-backend/src/service/router.ts
aramissennyeydd: We should be able to just cast `router as Router`, using `import Router from 'express-promise-router'` and then use this as normal. The chaining is a bit annoying here, but I think we can interrupt it, ie
```
.get('route-above')

(router as Router).get('this route')

router.get('rest-of-chain')
```

plugins/scaffolder-common/client/src/schema/openapi/generated/apis/Api.client.ts
aramissennyeydd, line 1: Hmm, I think the prefix is wrong here, this should just be `plugins/scaffolder-common/src/schema/openapi/apis/Api.client.ts`, there's an extra `client` in here.

solimant: Sure, I can change that, but I was following the [docs](https://backstage.io/docs/openapi/01-getting-started/#generating-a-typed-client-from-a-spec):
> ## Generating a typed client from a spec
>
> From your current backend plugin directory, run `yarn backstage-repo-tools package schema openapi generate --client-package <plugin-client-directory>`. `<plugin-client-directory>` is a new directory and npm package that you should create. The general pattern is to add a new entry point to your plugin's common package, `plugins/<plugin-name>-common/client`. You should add this command to your `package.json` for future use.

plugins/scaffolder-backend/src/service/router.ts
solimant: Ah! I didn't think about breaking the chaining 💡

solimant:
@aramissennyeydd - I've addressed the most recent set of requested changes, but please let me know if there's anything else. Thank you for taking the time 🙏🏼
@benjdlambert, @Rugvip - would greatly appreciate your feedback here as well 🙏🏼

solimant:
Hi @Rugvip, @benjdlambert 👋🏼  - Just wanted to follow up on this to see if there are any change requests I can address.

aramissennyeydd (approved):
Looks good on my end!

plugins/scaffolder/src/alpha/components/TemplateEditorPage/DryRunResults/DryRunResultsList.tsx
```diff
@@ -119,7 +119,7 @@ async function downloadDirectoryContents(
   directoryContents: {
     path: string;
     base64Content: string;
-    executable: boolean;
+    executable?: boolean;
```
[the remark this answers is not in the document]
solimant: In `router.ts`, `executable` is set from `file.executable`:
https://github.com/backstage/backstage/blob/f2644a1b5bafe8b427f42891491f5ccfeadf3988/plugins/scaffolder-backend/src/service/router.ts#L867-L875

`file` is of type `SerializedFile`, which has `executable` as an optional boolean:
https://github.com/backstage/backstage/blob/04559e342286e4b4360d054d708c0ec4bc5df76b/plugins/scaffolder-node/src/files/types.ts#L20-L25

Therefore, the response from the backend has it as optional. And so, the change on the frontend in `DryRunResultsList.tsx` is ensuring it matches that.

solimant:
Hi @benjdlambert, @Rugvip 👋🏼 I'm happy to refactor the PR if it will make the review process simpler. Just let me know.

aramissennyeydd:
@solimant If you haven't already, I think this would be a great PR to discuss during the [Scaffolder SIG](https://github.com/backstage/community/tree/main/sigs/sig-scaffolder) 😁

solimant:
@aramissennyeydd - done! Thank you for the tip 🙏🏼

benjdlambert:
Hey @solimant :wave: Did you manage to get any time to look into better e2e tests? :pray:

solimant:
Hi @benjdlambert - I haven't yet, but I will. From the time we discussed this in the SIG meeting, these are the docs for testing, correct?
https://backstage.io/docs/plugins/new-backend-system/#testing
Anything else I should look at that can help?

benjdlambert:
@solimant sorry I missed your comment as it was hidden behind the collapsed items.

Yep I think those docs should be a good start to be able to spin up the Scaffolder backend at least and start hitting it with some requests. :pray:

solimant:
@aramissennyeydd - could you please help reopen the PR? Looks like I was too slow to catch it before it was closed by the bot. Thank you 🙏🏼

solimant:
I've made more commits to the PR branch to resolve conflicts, but looks like they won't show up here until the PR is reopened, so just a gentle reminder 🙏🏼

awanlin:
Hi @solimant, sadly now that you've push commits we can't reopen this PR any more, you'll have to open a new one, sorry about that.

solimant:
@awanlin - thank you for taking a look. I've reset my commit history to the most recent commit SHA before the PR was closed. Could you please take another look to see if this helps?

benjdlambert:
@solimant done :tada:

solimant:
I'm seeing random unit test failures. Is this a known issue? The failing tests are unrelated (mostly from `FAIL plugins/kubernetes-backend/src/service/KubernetesFetcher.test.ts`) and they fail with a message saying "Test environment has been torn down":

image

solimant:
Opened #29923 to add a tests foundation for the Scaffolder.

benjdlambert:
@solimant hopefully we should be able to get this in the release for next week.

I'm just wrapping up the last bits of breaking changes with the scaffolder to allow for the `catalogServiceRef` to be used everywhere in the actions instead of the `catalogApi` and then I think we can merge the tests, and then this :pray:

benjdlambert:
@solimant now we've merged the tests, do you wanna rebase this :pray: would like to get it in the next release tomorrow to do some testing before the mainline release :pray:

solimant:
@benjdlambert - of course! I'm on it 👍🏼

plugins/scaffolder-common/report.api.md
```diff
@@ -7,15 +7,386 @@ import { Entity } from '@backstage/catalog-model';
...
+  id: string;
+  description?: string;
+  schema?: {
+    input?: JSONSchema7;
+    output?: JSONSchema7;
+  };
+  examples?: ActionExample[];
+};
+
+// @public @deprecated
```
benjdlambert: No point exporting this as deprecated straight away, might as well just remove it from the exports entirely.

plugins/scaffolder-common/src/ScaffolderClient.ts
benjdlambert: I think that it's mostly fine to do that. The SSE one is a little special.

solimant:
@benjdlambert - all good now. Please have a look when you get a chance. Happy to address anything.

solimant:
Btw, do we need to include documentation updates or callouts?

awanlin:
> Btw, do we need to include documentation updates or callouts?

Hi @solimant, we added a feature to the docs to use the spec to generate them here's an example for the Catalog's `getEntities` endpoint: https://backstage.io/docs/features/software-catalog/api/get-entities. There are some manual changes needed for this to work though. Like this: https://github.com/backstage/backstage/blob/71e0cc444ac1761ffad6d98b84a9b75deeeda17d/microsite/sidebars.ts#L15. Outside of that as this isn't really documented right now I don't think there's anything that needs to be updated.

@aramissennyeydd any chance you can help with this? I'm not sure all the changed needed to add the Scaffolder OpenAPI to the docs.

aramissennyeydd:
@awanlin How do we feel about a fast follow? This PR is already chunky enough as is - happy to help there.

awanlin:
> @awanlin How do we feel about a fast follow? This PR is already chunky enough as is - happy to help there.

Sure, I'm good with that, don't want to block this for the docs bit that's more technical in nature.

plugins/scaffolder-backend/src/ScaffolderPlugin.test.ts
```diff
@@ -737,9 +737,10 @@ describe('scaffolderPlugin', () => {
     });
 
     // Retry the task by ID
-    response = await request(server).post(
-      `/api/scaffolder/v2/tasks/${taskId}/retry`,
-    );
+    response = await request(server)
+      .post(`/api/scaffolder/v2/tasks/${taskId}/retry`)
+      .send({});
```
benjdlambert, line 742: Hmm is this a breaking change with the OpenAPI spec?

plugins/scaffolder-backend/src/scaffolder/tasks/taskRecoveryHelper.ts
```diff
@@ -28,7 +28,7 @@ export const trimEventsTillLastRecovery = (
   if (recoveredEventInd >= 0) {
     const ind = events.length - recoveredEventInd - 1;
     const { recoverStrategy } = events[ind].body as {
-      recoverStrategy: TaskRecoverStrategy;
+      recoverStrategy?: TaskRecoverStrategy;
```
benjdlambert, line 31: Do we even need this typing now?

plugins/scaffolder-backend/src/schema/openapi/generated/models/TaskSecrets.model.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// ******************************************************************
+// * THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY. *
+// ******************************************************************
+
+/**
+ * TaskSecrets
+ * @public
+ */
+export interface TaskSecrets {
```
benjdlambert, line 25: Hmm, I'm not sure that we should keep this in the public API?

There should be no way to set this from the outside as it should be just inferred from the Authorization header, and there should be no way to get it from a task either.

plugins/scaffolder-backend/src/service/router.test.ts
```diff
@@ -977,7 +991,7 @@ describe('scaffolder router', () => {
 
   describe('GET /v2/tasks/:taskId/eventstream', () => {
     it('should return log messages', async () => {
-      const { router, taskBroker } = await createTestRouter();
+      const { unwrappedRouter: router, taskBroker } = await createTestRouter();
```
benjdlambert, line 1051: What's the need to use the `unwrappedRouter` here? Does this not work when using the OpenAPI Router directly?

plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -373,8 +372,10 @@ export async function createRouter(
             description: template.metadata.description,
             'ui:options': template.metadata['ui:options'],
             steps: parameters.map(schema => ({
-              title: schema.title ?? 'Please enter the following information',
-              description: schema.description,
+              title:
+                (schema.title as string) ??
```
benjdlambert, line 407: What's the need for the explicit cast here in `schema.title` and `schema.description`? What other types can they be here that don't get nullish coerced?

plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -696,8 +696,10 @@ export async function createRouter(
         await auditorEvent?.fail({ error: err });
         throw err;
       }
-    })
-    .get('/v2/tasks/:taskId/eventstream', async (req, res) => {
+    });
+  (router as express.Router).get(
```
benjdlambert, line 751: Ah, is this because we don't define the eventstream endpoint in OpenAPI?

plugins/scaffolder-common/report.api.md
```diff
@@ -7,15 +7,386 @@ import { Entity } from '@backstage/catalog-model';
...
+export type ActionExample = ScaffolderUsageExample;
+
 // @public
 export const isTemplateEntityV1beta3: (
   entity: Entity,
 ) => entity is TemplateEntityV1beta3;
 
+// @public
+export type ListActionsResponse = Array<Action>;
```
benjdlambert, line 34: I'm a little unsure if this package is the best place for this. I don't think there's a precedent right now for where to put the OpenAPI clients that are generated and their types.

`catalog` -> `catalog-client`
`events` -> `events-node` 
`search` -> `missing`

Just wanna chat to other @backstage/maintainers here and chat about options and the best path forward for this.

benjdlambert: Another thing is I think that a lot of these types could be re-used in the `scaffolder-backend` / `scaffolder-node` too and cleaned up

benjdlambert (changes requested):
Hey @solimant!

Thanks for the continued effort on this, just had a flick through and there's some comments that I think we want to address before we can ship this.

Don't think anything is too controversial though. Also wanna chat about the location for the `scaffolder-client` and where all those types should be stored. Like should we be creating a seperate package for this, or are we happy putting it all in `-common`?

Will get back to you when I chat to the maintainers later.

plugins/scaffolder-common/report.api.md
solimant: > I'm a little unsure if this package is the best place for this

So, as far as location goes, the docs say they should be in common:
> The general pattern is to add a new entry point to your plugin's common package, `plugins/<plugin-name>-common/client`.

> a lot of these types could be re-used in the `scaffolder-backend` / `scaffolder-node`

That's true, and it's pretty much why I thought putting in an isomorphic package would make sense. Unless we're saying that the types wouldn't be re-used on the frontend..?

plugins/scaffolder-backend/src/ScaffolderPlugin.test.ts
solimant: Ah, yeah, the spec marks the body as required because of the following part in the router where it tries to spread the `secrets` from the `body`:
https://github.com/backstage/backstage/blob/9ad11f65556aa2880ab030b213e0e13cd81ae6e6/plugins/scaffolder-backend/src/service/router.ts#L683-L691

plugins/scaffolder-backend/src/scaffolder/tasks/taskRecoveryHelper.ts
solimant: It's because `body`'s type is being change to:
```diff
-  body: JsonObject;
+  body: {
+    message: string;
+    stepId?: string;
+    status?: TaskStatus;
+  } & JsonObject;
```
And apparently `recoverStrategy` is expected to be in the `JsonObject` portion.

plugins/scaffolder-backend/src/service/router.test.ts
solimant: It's because routes that are a bit involved and not quite typical like the other routes, such as `/v2/tasks/:taskId/eventstream`, weren't captured in the OpenAPI spec. I'm happy to give it another attempt, but might need some help.

@aramissennyeydd - I don't recall if we discussed this particular one before and decided to skip. Any thoughts?

plugins/scaffolder-backend/src/service/router.ts
solimant: It's because `schema.title` is of type `JsonValue`, and `steps` now expects `title` to be `string`. Hence the explicit casting. Why is `schema.title`'s type `JsonValue`? Because `schema` is of type `TemplateParametersV1beta3`, which is declared as:
https://github.com/backstage/backstage/blob/9ad11f65556aa2880ab030b213e0e13cd81ae6e6/plugins/scaffolder-common/src/TemplateEntityV1beta3.ts#L150-L152
Therefore, any undeclared properties are assumed to be in the `JsonObject` part, which would give them the type `JsonValue`.

plugins/scaffolder-backend/src/service/router.ts
solimant: Precisely 👍🏼 I had made an attempt to define it, but failed. Any help would be great.

plugins/scaffolder-node/src/tasks/types.ts
solimant: @benjdlambert - what do you think? This touches on the topic of reorganizing/centralizing types.

solimant (commented):
@benjdlambert - thank you very much for the review, and my apologies for the delayed response. I have addressed all your questions and change requests. Happy to address anything further 🙏🏼

benjdlambert:
@aramissennyeydd do you know of a way we can define the EventStream endpoint in OpenAPI? :pray:

plugins/scaffolder-backend/src/service/router.test.ts
aramissennyeydd: @benjdlambert My thinking there is that that could be a followup - the tooling is generally only set up to support JSON endpoints, so supporting an SSE/EventStream endpoint could be a significant chunk of work. Propagating the types down to the express router is probably another big chunk of work.

benjdlambert:
OK - will have another look through this next week and see how we're feeling. It's a little too close to mainline release to say that we can ship this next week, but will see what the possibilities are. :pray:

benjdlambert:
@solimant looks like there's some tests failing in this. We've got a mainline release today, would like to try and get this merged ASAP after that if it's possible. Wondering if this is something to do with the permissions work that got merged recently: https://github.com/backstage/backstage/pull/29202

aramissennyeydd:
Yup, I think the error message could use some help here - basically, we don't have documented errors for 403s for those 2 routes.

aramissennyeydd:
@solimant Looks like there's an API report failure still happening but everything else looks good 🎉

solimant:
> @solimant Looks like there's an API report failure still happening but everything else looks good 🎉

Thank you, @aramissennyeydd, I've updated the API reports.
In case you missed them, could you please also confirm that the changes I made to `typescript-backstage-client/modelEnum.mustache` and `typescript-backstage-server/modelEnum.mustache` are good? I made those to overcome the following error:
```
com.samskivert.mustache.MustacheParseException: Inverted section missing close tag 'stringEnums' @ line 15
```

sennyeya:
(personal account) yeah that looks fine

aramissennyeydd:
@benjdlambert How are we feeling on this one?

plugins/scaffolder-common/package.json
```diff
@@ -61,10 +61,20 @@
...
+    "@backstage/errors": "workspace:^",
+    "@backstage/integration": "workspace:^",
     "@backstage/plugin-permission-common": "workspace:^",
-    "@backstage/types": "workspace:^"
+    "@backstage/types": "workspace:^",
+    "@microsoft/fetch-event-source": "^2.0.1",
+    "@types/json-schema": "^7.0.9",
+    "cross-fetch": "^4.0.0",
+    "json-schema": "^0.4.0",
+    "msw": "^1.0.0",
```
Rugvip: This one needs to be a dev dep

plugins/scaffolder-backend/src/service/router.ts
benjdlambert: @solimant does this help? https://github.com/OAI/OpenAPI-Specification/issues/396

Otherwise I wonder if we can just model it simply like this:


```
...

event: log
data: {"id":106,"isTaskRecoverable":false,"taskId":"86475da8-1186-40ac-98f1-3b85f1a75b2f","body":{"message":"Finished step Notify","stepId":"notify","status":"completed"},"type":"log","createdAt":"2025-07-22T11:52:05.000Z"}

event: completion
data: {"id":107,"isTaskRecoverable":false,"taskId":"86475da8-1186-40ac-98f1-3b85f1a75b2f","body":{"message":"Run completed with status: completed","output":{"links":[{"url":"https://github.com/x/x.git","title":"GitHub Repository"},{"url":"https://github.com/x/x/actions/workflows/pages.yml","title":"Github Action Workflow"}],"text":[{"title":"Reminder to Delete Repository","content":"**Note:** You will need to manually delete the resulting repository once you are done.\n"}]}},"type":"completion","createdAt":"2025-07-22T11:52:05.000Z"}
```

aramissennyeydd: @benjdlambert From my comment above, https://github.com/backstage/backstage/pull/27771#discussion_r2163712643 - can we do this in a follow up? It should be possible to do that as incremental spec coverage later on. (part of this is that I'm not confident the tooling will handle any non-JSON endpoints rn)

benjdlambert: Yeah OK. Happy to address this in a followup, but ideally before 1.42.

.changeset/clean-chairs-sit.md
```diff
@@ -0,0 +1,10 @@
+---
+'@backstage/backend-openapi-utils': minor
```
benjdlambert, line 2:
```suggestion

```

This should probably be in another changeset, and also only a `patch`

plugins/scaffolder-node/src/tasks/types.ts
benjdlambert: Yeah, think these types might need a cleanup, but happy to take that before the mainline release if that's OK? Going to try and ship this ASAP.

plugins/scaffolder/src/components/OngoingTask/OngoingTask.tsx
```diff
@@ -223,7 +223,7 @@ function OngoingTaskContent(props: {
   const [, { execute: triggerRetry }] = useAsync(async () => {
     if (taskId) {
       analytics.captureEvent('retried', 'Template has been retried');
-      await scaffolderApi.retry?.(taskId);
+      await scaffolderApi.retry?.({ taskId });
```
benjdlambert: I think that this is a breaking change to the client which we can't make yet, this needs updating in our wrapping client to take just `taskId`.

There's a bigger piece of work which is to make sure that all the functions of the scaffolder API follow this pattern `({ taskId })` for example, but rather make that change as a bigger breaking change instead.

benjdlambert (changes requested):
Just some minor bits and then we can ship I think.

.changeset/clean-chairs-sit.md
benjdlambert: There's also a changeset needed for `repo-tool` as well?

benjdlambert:
@solimant I created a separate changeset for the `-react` package with some additional callouts to all the deprecations.

I think that covers everything for now, but if you're willing to cleanup the types in a followup PR those deprecations will need to also be called out explicitly in another changeset too :pray:

Approving and auto-merge on! Thanks a lot for the effort here all round from everyone involved :tada: @aramissennyeydd  too 👍

benjdlambert (approved):

PR #27793, opened 2024-11-22
Techdocs - Add option for custom iframe attributes

### LINKED ISSUE #27538: 🐛 Bug Report: allowedIframeHosts sanitizes properties in the iframe
### 📜 Description

when using `allowedIframeHosts` for something like [embedding a structurizr diagram](https://docs.structurizr.com/cloud/embed) the backstage sanitizer seems to be removing the `apiKey` property of the iframe, thus causing the iframe to fail. 

### 👍 Expected behavior

I expect that having a markdown file like this:

```markdown
<html>
<body>
<iframe id="myEmbeddedDiagram" src="https://myurl.com/embed/1?diagram=Diagram1&diagramSelector=false&iframe=myEmbeddedDiagram" apiKey="VALUE HERE" width="100%" marginwidth="0" marginheight="0" frameborder="0" scrolling="no" allowfullscreen="true"></iframe>
<script type="text/javascript" src="https://myurl.com/static/js/structurizr-embed.js"></script>
<p></p>
</body>
</html>
```

Should result in all of the provided iframe values coming through on the other side, especially if this host is in my `allowedIframeHosts` 

From looking at other Issues it seems like this was done on purpose to remove vulnerabilities, but in this case it completely obviates structurizr's main purpose. 

It removed everything except the `id`,`width`, and `src` properties from the iframe tag.

image

### 👟 Reproduction steps

put techdocs markdown file with an iframe in it and any random properties in your repo. 

_No response_

### 🖥️ Your Environment

```
OS:   Darwin 23.6.0 - darwin/arm64
node: v18.19.0
yarn: 1.22.19
cli:  0.28.0 (installed)
backstage:  1.32.2
```

snowe2010:

- **Add allowedAttributes config to allowedIframeHosts**
- **Allow using 'old' style sanitizer config**
- **Add documentation around this feature**

Allows adding iframe attributes that might be non-standard, without filtering them out with the sanitizer. 
Each iframe host can have a different config.

Resolves #27538 


.changeset/tasty-moles-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-techdocs': minor
+---
+
+Add a new config for `allowedIframeHosts` to allow settings custom allowed attributes
```
alexlorenzi, line 5: Sorry for the nitpick,  but it might be useful to include an example of the updated config in the changelog to make it easier for folks to understand the change, rather than jumping into the docs. What you have in the how-to could be replicated here.

plugins/techdocs/src/reader/transformers/html/transformer.ts
```diff
@@ -41,13 +45,87 @@ export const useSanitizerTransformer = (): Transformer => {
...
+                      !allowedAttributes.includes(attr.name.toLocaleLowerCase())
+                    ) {
+                      currNode.removeAttribute(attr.name);
+                    }
+                  }
+                }
+              }
+            });
+          }
+        }
```
alexlorenzi, line 128: This is looking good so far!  I'm wondering if it might make more sense to move this logic into the `removeUnsafeIframes` function?  It seems like a natural fit, and you could use the existing `data` parameter to check if the attributes are allowed as well. What do you think?

plugins/techdocs/src/reader/transformers/html/hooks/iframes.ts
```diff
@@ -37,6 +37,22 @@ const isSafe = (node: Element, hosts: string[]) => {
...
+ */
+export const isHost = (node: Element, check: string) => {
+  const src = node.getAttribute('src') || '';
+  try {
+    const { host } = new URL(src);
+    return host.includes(check);
+  } catch {
+    return false;
+  }
+};
```
alexlorenzi, line 54: Hey, just want to make sure I'm understanding this right.  Is the main difference between this function and `isSafe` that this one works on a single host, while `isSafe` uses an array?  

Also, is it super important that we're checking for a substring here, or could we potentially reuse `isSafe` by passing the host in as a single-item array (like `isSafe(node, [host])`)?  Maybe we could even update `isSafe` to handle both arrays and single strings?

alexlorenzi (changes requested):
Hey, sorry about the delay

The logic in these changes looks good overall! 👍 Just had a couple of thoughts about whether we could simplify things a bit by reusing some existing code.  What do you think?

snowe2010:
Oh sorry, I'd been looking into these comments but got distracted and haven't finished going over all of them yet.

.changeset/tasty-moles-wonder.md
snowe2010: Will do!

snowe2010:
I'm still working on it! got sick and haven't had time to test and push.

johnphilip283:
Hi @snowe2010 - just doing a cleanup of some PRs and want to see if this was still something you were interested in contributing?

snowe2010:
@johnphilip283 Hi john, yes I'll still contribute if it's desired, but was out of town for a while due a death in the family. Have had trouble catching back up with everything.

johnphilip283:
> @johnphilip283 Hi john, yes I'll still contribute if it's desired, but was out of town for a while due a death in the family. Have had trouble catching back up with everything.

Oh man, I'm quite sorry to hear that - please take all the time you need. No rush at all for this - it's still a nice bit of functionality that could be supported by TechDocs better at the moment, but no one has an explicit need for it just yet, so no pressure to get it over the line.

rajibq:
Can this be re-opened? We have a techdocs which embeds some videos with iframe, but because DOMPurify is stripping out  `allow` or `allowfullscreen`, the full screen button on videos fails to work.

PR #27797, opened 2024-11-23
correctly measure stitching queue length

freben:
It doesn't matter whether the `next_stitch_at` value is in the past or not. If it's still non-null, it's either due for stitching or undergoing stitching right now. The only case where it's in the future, is right after it gets picked up for a stitch attempt, since that's a retry mechanism that ensures that some other worker also tries to stitch it if the current one dies.

The problem is, there seems to be some bug where entities can sometimes end up staying indefinitely in the stitch queue, in deferred stitching mode. If that happens, we definitely want to see the ever-growing queue size rise in this metric.

Yes, this number will now include _currently_ stitching entities. But that's not a bad thing, I think.

The metric of how far behind we are on stitching, is instead covered by the `catalog.stitching.queue.delay` metric, so no information is lost by this change.

### DIFF
--- a/.changeset/odd-worms-peel.md
+++ b/.changeset/odd-worms-peel.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Correctly report stitching queue length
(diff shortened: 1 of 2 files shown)


Rugvip (approved):
Yep makes sense :+1:

PR #27799, opened 2024-11-24
Be more aggressive in dequeueing entities for stitching

freben:
Be extra careful to dequeue things from the stitching queue when stitching finishes without exceptions. You may want to review this with ignoring whitespace changes, because that's mostly what it consists of.

The real reason for removing it, is that I wanted to remove a source of "permanent" stitching errors that would just stay around indefinitely, for little value, while making the stitching loop spin infinitely.

### DIFF
--- a/.changeset/neat-insects-cough.md
+++ b/.changeset/neat-insects-cough.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Be more aggressive in dequeueing entities for stitching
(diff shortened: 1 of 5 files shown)


Rugvip (changes requested):
#8345 was a security fix, ensuring that you can't override important entities or fields, e.g. template entities or sign-in resolver lookups

freben:
@Rugvip  Ah alright, reverted that then.

Rugvip (approved):

PR #27800, opened 2024-11-24
Compute some metrics using search table facet aggregations

freben:
This metric takes an unnecessary amount of resources to compute. Better to leverage the search table indices. This is basically the facets query structure, actually. This of course mans that the metric now represents how many entities that exist in the final entities, not in the refresh state. Arguably, this is more correct.

Example prometheus output:

```
catalog_entities_count{kind="api"} 6
catalog_entities_count{kind="component"} 14
catalog_entities_count{kind="domain"} 3
catalog_entities_count{kind="group"} 8
catalog_entities_count{kind="location"} 14
catalog_entities_count{kind="resource"} 1
catalog_entities_count{kind="system"} 3
catalog_entities_count{kind="template"} 6
catalog_entities_count{kind="user"} 17
```

### DIFF
--- a/plugins/catalog-backend/src/database/metrics.ts
+++ b/plugins/catalog-backend/src/database/metrics.ts
@@ -16,9 +16,8 @@
 
 import { Knex } from 'knex';
 import { createGaugeMetric } from '../util/metrics';
-import { DbRefreshStateRow, DbRelationsRow, DbLocationsRow } from './tables';
+import { DbRelationsRow, DbLocationsRow, DbSearchRow } from './tables';
 import { metrics } from '@opentelemetry/api';
-import { parseEntityRef } from '@backstage/catalog-model';
 
 export function initDatabaseMetrics(knex: Knex) {
   const seenProm = new Set<string>();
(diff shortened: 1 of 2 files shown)


plugins/catalog-backend/src/database/metrics.ts
```diff
@@ -30,23 +29,22 @@ export function initDatabaseMetrics(knex: Knex) {
...
       async collect() {
-        const result = await knex<DbRefreshStateRow>('refresh_state').select(
-          'entity_ref',
-        );
-        const results = result
-          .map(row => row.entity_ref.split(':')[0])
-          .reduce((acc, e) => acc.set(e, (acc.get(e) || 0) + 1), new Map());
+        const results = await knex<DbSearchRow>('search')
+          .where('key', '=', 'kind')
+          .whereNotNull('original_value')
```
Rugvip, line 34: I'd expect it to be best to use the case normalized value rather than the original one?

Rugvip (commented):
nice :+1: , this is arguably more accurate as well

plugins/catalog-backend/src/database/metrics.ts
Rugvip: We are? where's that?

Feels like it would be safer to use `value` either way?

freben: Just by virtue of the schemas using strict enums and we use ajv to match against those schemas.

https://github.com/backstage/backstage/blob/5c703d973459caebff85326bad875f5354d828ab/packages/catalog-model/src/schema/kinds/API.v1alpha1.schema.json#L40

Also I guess it's a matter of how dearly we want these metric values to actually be correct instead of forcibly lowercased.

Rugvip (approved):

PR #27804, opened 2024-11-24
Fix issues with type incompatibility between major versions of @types/express

mtlewis:

Ran into this while looking into the failures on https://github.com/backstage/backstage/pull/27802. PR includes a couple of fixes:

1. Make all dependencies on @types/express more restrictive. Using `*` doesn't make sense, since [`@types` packages follow major versions of the packages they target](https://github.com/DefinitelyTyped/DefinitelyTyped?tab=readme-ov-file#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library). Since we restrict express to v4, we should restrict the corresponding types package in the same way.
2. Stop adding `*` dependencies for @types packages in templated plugins. This was causing e2e test failures (and also of course leaking `*` dependencies into templated plugins in the real world).

### DIFF
--- a/packages/cli/src/lib/version.ts
+++ b/packages/cli/src/lib/version.ts
@@ -90,12 +90,6 @@ export function createPackageVersionProvider(lockfile?: Lockfile) {
     }
 
     const lockfileEntries = lockfile?.get(name);
-    if (
-      name.startsWith('@types/') &&
-      lockfileEntries?.some(entry => entry.range === '*')
-    ) {
-      return '*';
-    }
 
     for (const specifier of ['^', '~', '*']) {
       const range = `workspace:${specifier}`;
--- a/.changeset/cold-icons-worry.md
+++ b/.changeset/cold-icons-worry.md
@@ -1,0 +1,6 @@
+---
+'@backstage/cli': patch
+---
+
+Remove special-casing for `@types` packages when generating dependency entries
+during templating
--- a/.changeset/odd-apples-explain.md
+++ b/.changeset/odd-apples-explain.md
@@ -1,0 +1,8 @@
+---
+'@backstage/plugin-auth-node': patch
+'@backstage/plugin-permission-backend': patch
+'@backstage/plugin-devtools-backend': patch
+'@backstage/plugin-signals-backend': patch
+---
+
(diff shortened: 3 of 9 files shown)


Rugvip (approved):
Yep, thanks :+1: 

as a sidenote it's interesting that express v5 types are marked as stable while express v5 is still on the next channel

packages/cli/src/lib/version.ts
```diff
@@ -90,12 +90,6 @@ export function createPackageVersionProvider(lockfile?: Lockfile) {
     }
 
     const lockfileEntries = lockfile?.get(name);
-    if (
-      name.startsWith('@types/') &&
-      lockfileEntries?.some(entry => entry.range === '*')
-    ) {
-      return '*';
```
mtlewis, line 97: This path is only being hit in one template right now (assuming it's not used anywhere else other than as a result of `versionQuery` expressions in templates). I'm proposing pulling out as a blunt first try to fix the e2e test issue.

Rugvip: I'm thinking it might be best if we filter out * from the logic below that finds a range that satisfies the target version, and only use it if it is the only option?

need a changeset for this change too btw

mtlewis: > I'm thinking it might be best if we filter out * from the logic below that finds a range that satisfies the target version, and only use it if it is the only option?

That's on top of this change, right? If so, should we perhaps keep that separate in the hope to keep this change somewhat simple? The change in dab696a already fixes the issue at hand in the code, and I'm running out of time to put into this today.

> need a changeset for this change too btw

Yep, ofc. Will do that once we're definite on the change being made.

Rugvip: No the risk is introduced by the fact that we're no longer exiting on `*`. I think it's fine though because the logic is wrong in a way where `*` won't be prioritized as it picks the lowest compatible range instead of the highest.

:shipit: 🤷

Rugvip (approved):
Nice! 👍 

Thank you for looking into this 🎉

packages/cli/src/lib/version.ts
mtlewis: :+1: changeset added.

PR #27806, opened 2024-11-25
EntityOwnerPicker: fix scroll reset

vinzscam:

Fixing an issue causing `EntityOwnerPicker` to reset scrolling when more elements are loaded.

Before:
https://github.com/user-attachments/assets/1120f983-a7d8-477d-8fa3-8a5c34847992

After:
https://github.com/user-attachments/assets/2e72eca1-775a-45b5-b0b5-b5f15c5871bd

### DIFF
--- a/plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
+++ b/plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
@@ -93,10 +93,6 @@ export type EntityOwnerPickerProps = {
   mode?: 'owners-only' | 'all';
 };
 
-const Popper = React.memo(function Popper({ children }: PopperProps) {
-  return <div>{children as ReactNode}</div>;
-});
-
 function RenderOptionLabel(props: { entity: Entity; isSelected: boolean }) {
   const classes = useStyles();
   const isGroup = props.entity.kind.toLocaleLowerCase('en-US') === 'group';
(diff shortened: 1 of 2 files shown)


plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
```diff
@@ -92,6 +93,10 @@ export type EntityOwnerPickerProps = {
   mode?: 'owners-only' | 'all';
 };
 
+const Popper = React.memo(function Popper({ children }: PopperProps) {
+  return <div>{children as ReactNode}</div>;
+});
```
Rugvip, line 98: `popperProps` are no longer being forwarded to the `div`, could that break things?

Also wondering whether it might be enough to define this component outside, i.e. is `React.memo` actually needed?

Rugvip (approved):
Nice! 👍 

Two Qs

plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
vinzscam: I think you are right, it should be enough without memo, but let me double check. 
I've stopped forwarding the other props because they were all related to the Popper react component and unrelated to the returned div, so we were getting some warnings for that

vinzscam: yep, works! Thx for spotting this

camilaibs (approved):

PR #27814, opened 2024-11-25
Ensure that the scheduled worker task doesn't run at an unreasonably high frequency

freben:
This is a bit hand wavy, but protects the scheduler in aggressive setups

### DIFF
--- a/plugins/catalog-backend-module-incremental-ingestion/src/module/WrapperProviders.ts
+++ b/plugins/catalog-backend-module-incremental-ingestion/src/module/WrapperProviders.ts
@@ -119,12 +119,17 @@ export class WrapperProviders {
         connection,
       });
 
-      const frequency = Duration.isDuration(burstInterval)
+      let frequency = Duration.isDuration(burstInterval)
         ? burstInterval
         : Duration.fromObject(burstInterval);
-      const length = Duration.isDuration(burstLength)
+      if (frequency.as('milliseconds') < 5000) {
+        frequency = Duration.fromObject({ seconds: 5 }); // don't let it be silly low, to not overload the scheduler
+      }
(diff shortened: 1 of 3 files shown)


plugins/catalog-backend-module-incremental-ingestion/src/module/WrapperProviders.ts
```diff
@@ -119,12 +119,17 @@ export class WrapperProviders {
         connection,
       });
 
-      const frequency = Duration.isDuration(burstInterval)
+      let frequency = Duration.isDuration(burstInterval)
         ? burstInterval
         : Duration.fromObject(burstInterval);
-      const length = Duration.isDuration(burstLength)
+      if (frequency.as('milliseconds') < 5000) {
```
Rugvip, line 125: Is there a risk that 5 is a bit high?

Rugvip (approved):

plugins/catalog-backend-module-incremental-ingestion/src/module/WrapperProviders.ts
freben: So what happens is that it'll take up to five seconds longer than you thought between bursts - that is, the interval that's meant to be an actual rest time between intense many-page pagination runs. I felt like resting for less than five seconds, e.g. to stave off rate limits or not exhaust resources, kinda wouldn't achieve much most likely 🤷 I hope that this duration most of the time is set to like on the order of a minute or several minutes, but not sure

freben: Let's merge and experiment I guess, happy to lower it later on if needed

PR #27816, opened 2024-11-25
deprecate IncrementalCatalogBuilder, stop using old event interface

freben:
Found some old EventSubscriber remnants, and a missed deprecation in alpha


Rugvip (approved):
👍

Issue #25595, opened 2024-07-11
Ability to control the refresh token max Age


bbasu:
### 📜 Description

We use backstage for a customer facing portal, the auth idp is Okta.

Backstage now has the ability to configure the backstage access token duration (minimum being 10mins), however, the refresh token duration is hard coded to 1000 days. For modern browsers the http secure cookie has a max 400 days duration, hence the refresh token cookie that is getting set is for 400 days.

We would want to control this so that our application does force logouts after a smaller duration (like 6hrs), I was looking through the code and I could find the implementation here:

```plugins\auth-backend\src\lib\oauth\OAuthAdapter.ts```

![image](https://github.com/backstage/backstage/assets/14269816/959cbf8e-961d-4d2c-9f9b-81105e1ce6a0)

We tried overriding the whole plugin, however after moving to the new BE plugin system this code does not seem to be invoked. We found another place where this same implementation is defined:

```plugins\auth-node\src\oauth\OAuthCookieManager.ts```

In the new BE system, the oktaAuthenticator imported from @backstage/plugin-auth-backend-module-okta-provider seems to be using the 
```createOAuthProviderFactory```  from auth-node which is internally using the ```OAuthCookieManager.```

![image](https://github.com/backstage/backstage/assets/14269816/7c16d109-0640-4f40-b5fb-aaa1e65a2e77)

Can you please tell me whats the cleanest way to override this piece of code so that we can control the refresh token max age.

### 👍 Expected behavior

We would want to control this so that our application does force logouts after a smaller duration (like 6hrs), 

### 👎 Actual Behavior with Screenshots

Has no ability to control the max age of refresh token.

### 👟 Reproduction steps

Create a new backstage app

Run the backstage app

Observe the refresh token duration in the cookie after refresh is invoked.

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

Rugvip:
I think is something we can add as a configuration parameter in the auth Backend tbh, I see that it could be useful. Although I do want to note that there are a few other solutions out there, like https://backstage.io/docs/auth/autologout/ or configuring the refresh token expiration in the identity provider itself for those that support that.

[...]

PR #27821, opened 2024-11-26
support user configuration of auth cookie max age

JessicaJHee:

This PR adds the `auth.providers.providerId.cookieMaxAge` config for auth providers to allow users to configure the `refresh-token` and `granted-scopes` cookie lifespan.
* this was previously hard coded to 1000 days which is longer than the 400 day limit by most modern browsers, it would be nice to be able to configure this
---
Addresses this issue: https://github.com/backstage/backstage/issues/25595
* Referenced [this comment](https://github.com/backstage/backstage/issues/25595#issuecomment-2228444477) under the issue for guidance on the implementation 

**TODO:** update documentation for auth providers once the config is accepted


plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
```diff
@@ -99,6 +99,9 @@ export function createOAuthRouteHandlers<TProfile>(
   const callbackUrl =
     config.getOptionalString('callbackUrl') ??
     `${baseUrl}/${providerId}/handler/frame`;
+  const cookieMaxAge = config.has('cookieMaxAge')
```
Rugvip, line 102: I'm thinking we may want to keep this more oriented towards end outcome rather than the implementation. By that I mean a name such as `sessionMaxAge`, `sessionAgeLimit`, or `sessionDuration` could fit a bit better? One thing I specifically have in mind is the possibility of moving session storage to the backend in the future, at which point we'd be invalidating the DB entry rather than clearing the cookie. What do you think?

plugins/auth-node/src/oauth/OAuthCookieManager.ts
```diff
@@ -16,8 +16,9 @@
 
 import { CookieOptions, Request, Response } from 'express';
 import { CookieConfigurer } from '../types';
+import { HumanDuration, durationToMilliseconds } from '@backstage/types';
 
-const THOUSAND_DAYS_MS = 1000 * 24 * 60 * 60 * 1000;
+const FOUR_HUNDRED_DAYS_MS = 400 * 24 * 60 * 60 * 1000;
```
Rugvip, line 21: Sure this might be the actual max in most cases, but does it hurt to keep it at 1000? The intention here is just for it to last for a long time. I'd rather keep as is tbh

Rugvip (commented):
Thank you! 🎉 

Just some minor bite shedding, almost good to go 👍

plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
JessicaJHee: I agree! I think `sessionDuration` would be the most intuitive in terms of the outcome. 

I'm thinking we can still communicate that this configuration is tied to the cookie lifespan in the documentation like this, WDYT?

```
- `sessionDuration` (optional): Lifespan of the refresh token cookie and granted scopes cookie
  issued during the authentication process.
```

plugins/auth-node/src/oauth/OAuthCookieManager.ts
JessicaJHee: Sure can do that, either way works. The value will be [capped by the browser](https://developer.chrome.com/blog/cookie-max-age-expires#example) anyways 👍

plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
Rugvip: yep sounds good! :+1:

JessicaJHee:
@Rugvip I've made the requested changes, PTAL thank you!

.changeset/serious-ads-leave.md
```diff
@@ -0,0 +1,22 @@
...
+'@backstage/plugin-auth-backend-module-gitlab-provider': minor
+'@backstage/plugin-auth-backend-module-google-provider': minor
+'@backstage/plugin-auth-backend-module-oauth2-provider': minor
+'@backstage/plugin-auth-backend-module-auth0-provider': minor
+'@backstage/plugin-auth-backend-module-oidc-provider': minor
+'@backstage/plugin-auth-backend-module-okta-provider': minor
+'@backstage/plugin-auth-node': minor
+---
+
+Adds the auth.providers.providerId.sessionDuration config for auth providers to allow users to configure the refresh-token and granted-scopes cookie lifespan.
```
Rugvip, line 22:
```suggestion
Added `auth.providers.<providerId>.sessionDuration` config for auth providers to allow the lifespan of user sessions to be configured.
```

Thinking it's best to keep it a bit more high level rather than describe the current implementation? I'd want to keep this config in place and effect even if it for example is tied to session TTL for a DB implementation.

Same thing goes for the documentation entries

JessicaJHee: That makes sense! Will make the changes 👍

vinzscam (approved):
nice!

JessicaJHee:
@vinzscam Thanks for the review! Is there anything else I need to get this PR in?

JessicaJHee:
@vinzscam I re-triggered the DCO checks and it's now passed, do you mind approving this PR again when you get a chance? Thank you!

Rugvip (approved):
Thank you! 👍 

Let's :shipit:

PR #27838, opened 2024-11-26
Add gitlab scaffolder action to migrate groups and projects (bulk_imports)

mpsanchis:

     
 New action that uses [bulk_imports](https://docs.gitlab.com/ee/api/bulk_imports.html) to migrate a repository. Currently limited to `project_entity` (only repositories, no groups), since that is what I have tested. It can be extended in the future.

### DIFF
--- a/.changeset/purple-scissors-switch.md
+++ b/.changeset/purple-scissors-switch.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend-module-gitlab': patch
+---
+
+Added action `gitlab:group:migrate` to migrate projects using `bulk_import`
--- a/plugins/scaffolder-backend-module-gitlab/report.api.md
+++ b/plugins/scaffolder-backend-module-gitlab/report.api.md
@@ -89,20 +89,6 @@ export const createGitlabProjectDeployTokenAction: (options: {
 >;
 
 // @public
-export const createGitlabProjectMigrateAction: (options: {
-  integrations: ScmIntegrationRegistry;
-}) => TemplateAction<
-  {
-    destinationAccessToken: string;
-    destinationUrl: string;
-    sourceAccessToken: string;
-    sourceFullPath: string;
-    sourceUrl: string;
-  },
-  JsonObject
->;
-
-// @public
 export const createGitlabProjectVariableAction: (options: {
   integrations: ScmIntegrationRegistry;
 }) => TemplateAction<
--- a/plugins/scaffolder-backend-module-gitlab/report.api.md
+++ b/plugins/scaffolder-backend-module-gitlab/report.api.md
@@ -89,6 +89,20 @@ export const createGitlabProjectDeployTokenAction: (options: {
 >;
 
 // @public
+export const createGitlabProjectMigrateAction: (options: {
+  integrations: ScmIntegrationRegistry;
+}) => TemplateAction<
+  {
+    destinationAccessToken: string;
+    destinationUrl: string;
+    sourceAccessToken: string;
+    sourceFullPath: string;
+    sourceUrl: string;
+  },
+  JsonObject
+>;
+
+// @public
 export const createGitlabProjectVariableAction: (options: {
   integrations: ScmIntegrationRegistry;
 }) => TemplateAction<
--- a/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabProjectMigrate.test.ts
+++ b/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabProjectMigrate.test.ts
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2023 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { ConfigReader } from '@backstage/config';
+import { ScmIntegrations } from '@backstage/integration';
+import { TemplateAction } from '@backstage/plugin-scaffolder-node';
+import { createGitlabProjectMigrateAction } from './gitlabProjectMigrate';
+import { createMockActionContext } from '@backstage/plugin-scaffolder-node-test-utils';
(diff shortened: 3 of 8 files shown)


mpsanchis:
Note: this is the second PR of the split I did of https://github.com/backstage/backstage/pull/27360

The first PR was already merged: https://github.com/backstage/backstage/pull/27581

plugins/scaffolder-backend-module-gitlab/src/actions/gitlabProjectMigrate.test.ts
```diff
@@ -0,0 +1,98 @@
...
+import { createRootLogger } from '@backstage/backend-common';
+import { ConfigReader } from '@backstage/config';
+import { ScmIntegrations } from '@backstage/integration';
+import { TemplateAction } from '@backstage/plugin-scaffolder-node';
+import { createGitlabProjectMigrateAction } from './gitlabProjectMigrate';
+import { createMockActionContext } from '@backstage/plugin-scaffolder-node-test-utils';
+import { createMockDirectory } from '@backstage/backend-test-utils';
+
+// Make sure root logger is initialized ahead of FS mock
+createRootLogger();
```
benjdlambert, line 25: Why do we need to do this? Ideally we should drop all dependencies on `backend-common` now as the package is deprecated.

.changeset/purple-scissors-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend-module-gitlab': patch
+---
+
+Added action gitlabProjectMigrate to migrate projects using bulk_import
```
benjdlambert, line 5:
```suggestion
Added action gitlabProjectMigrate to migrate projects using `bulk_import`
```

benjdlambert (changes requested):
Nice! Thanks for this, just a few little bits :pray:

plugins/scaffolder-backend-module-gitlab/src/actions/gitlabProjectMigrate.test.ts
mpsanchis: Hi @benjdlambert !
I didn't look so much into it, but I found the same call other 6 times, where `import { createMockDirectory } from '@backstage/backend-test-utils'` was used. I built the tests to look similar to others.

If you know a quick hint on how to replace this, I'd appreciate that 😄

mpsanchis: Hi @benjdlambert 

What should I do with this? Could you give me hand?

Thanks!

benjdlambert: I think you can just remove it right from these tests, they should also be removed from the other ones, but it's not a prio from this PR. Do the tests fail if this is not called?

mpsanchis: Seems that it works without 😄 Let me know if you want more changes 🚀

plugins/scaffolder-backend-module-gitlab/report.api.md
```diff
@@ -88,6 +88,20 @@ export const createGitlabProjectDeployTokenAction: (options: {
   }
 >;
 
+// @public
+export const createGitlabProjectMigrateAction: (options: {
```
benjdlambert, line 92: I'm wondering if we should just stop exporting these actions individually now, as theres no real need with the new backend system, that you just add the module and it registers them for you. It's just going to be removed pretty soon with the removal of all the old backend system work.

benjdlambert (changes requested):
Nice! this is looking pretty good, just a note about exporting the creator.

Let's remove the export for that, and re-run the API Reports.

plugins/scaffolder-backend-module-gitlab/report.api.md
mpsanchis: Hi @benjdlambert , happy new year 🎉 

I am going to update my PR (and rebase, since I'm already changing stuff), trying to limit the surface area of `scaffolder-backend-module-gitlab`. I wanted to directly NOT export the actions, but I found the following uses in `scaffolder-backend`:
* `createGitlabRepoPushAction` is imported in `createBuiltinActions`.
* `createPublishGitlabAction` and `createPublishGitlabMergeRequestAction` are imported in `deprecated`.

My knowledge of Backstage's architecture is still limited (I'm a pretty new user), so I am not sure 100% what's the direction you want to take. As you'll see in my PR, I've limited the exported actions of  `scaffolder-backend-module-gitlab` to those three I mentioned above.

Let me know if this is what you meant, or if not, please be more explicit with the changes and I'll be happy to make them 😄

plugins/scaffolder-backend-module-gitlab/report.api.md
```diff
@@ -9,104 +9,6 @@ import { JsonObject } from '@backstage/types';
 import { ScmIntegrationRegistry } from '@backstage/integration';
 import { TemplateAction } from '@backstage/plugin-scaffolder-node';
 
-// @public
-export const createGitlabGroupEnsureExistsAction: (options: {
```
benjdlambert, line 13: I don't think we want to remove the existing exports as this would be a breaking change that we're not ready for yet.

benjdlambert (changes requested):
So I think that we need to keep the existing exports for now as we don't want to make the breaking change just yet.

plugins/scaffolder-backend-module-gitlab/report.api.md
benjdlambert: Just left another review, I think we're on the right track, but we don't want to remove the existing exports - not just yet at least.

`createBuiltinActions` will most likely be replaced or removed with `scaffolder@2.0.0` as this previously was used to add actions to the default installed, but now you install the modules separately for the desired behaviour and they're automatically registered, which is what you've done in the PR with this code:

```ts
        scaffolder.addActions(
          createGitlabGroupEnsureExistsAction({ integrations }),
          createGitlabProjectMigrateAction({ integrations }),
```

benjdlambert: Exactly, remove this https://github.com/backstage/backstage/pull/27838/files#diff-89c5699ed2a0b1d6c222d31b3d0e30c534af054dc6dec41817b43486d2192bb5R18 and just import it from the file directly and include `export * from './actions'` again :pray:

mpsanchis: Anything else I can do, Ben?

mpsanchis:
@benjdlambert thanks for the contribution. I've fixed up your commit with my first one so that CI doesn't complain (careful if you pull)

benjdlambert (approved):

benjdlambert:
Nice thanks! :tada:

PR #25366, opened 2024-06-24
Make useFacetsEntities fetches full entities from entity references

fnandocl:
EntityOwnerPicker will then display `title` or `displayName` if present for `mode='owners-only'`.

From the demo app,
BEFORE:
Screenshot from 2024-06-24 09-49-31

AFTER:
Screenshot from 2024-06-24 09-36-37

Fixes #25348

Issue #12247, opened 2022-06-23
Catalog(Page) pagination


vinzscam:
Today adopters with 5k+ might encounter issues when viewing the `/catalog` page.
The issues are related to how the entities are fetched by the frontend. By default (almost) all the entities present in the database are fetched and returned in one shot.
This means that in some cases the loading time of the CatalogPage is high. This issue has been pointed out by dozens of adopters. In some cases the issue wasn’t only affecting loading time of the CatalogPage, but has affected the entire usability of the UI. Since a lot of UI components depend on having all the entities downloaded and available in the frontend, in some cases the UI freezes or becomes unresponsive.

In order to fix the issue mentioned above, pagination on the catalog’s entities needs to be enforced.
At the moment, the entities are fetched from the `GET /entities` endpoint, which returns an array of results.
The endpoint supports offset pagination, meaning that on each request, it is possible to provide offset, limit and cursor.
However, whenever a request is sent, only the array of results is sent back to the caller, while all the pagination data gets discarded. Due to this limitation, it is not possible to send subsequent requests to the same endpoint, meaning that it isn’t possible to use pagination.

Having a fully supported paginated endpoint is crucial to provide a good user experience of the CatalogPage.

## Proposed solution

In order to not introduce breaking changes to the current endpoint, I propose to create a new version of the `/entities` endpoint which returns a paginated response of the entities.

I’ve been investigating the proposed solution and I came to a conclusion: this wouldn’t be enough.

Unfortunately, during the years, a lot of logic has been built on top of the CatalogPage.
As a result, a lot of filters depend on having all the entities available on the frontend in order to work properly.
So, together with the new endpoint, *all the frontend filters need to be rebuilt*.
Some of them were fairly easy to fix, like owners, lifecycle and tags.
Others, like starred or owned entities, were a bit more complicated.

During the last (6) months I’ve been working on this and I reached out to a good solution, proving that it is possible to make the CatalogPage work with 30k+ entities.

The amount of work has been massive. However I think it is possible to split the work into small pull requests, in order to ship this as soon as possible.

I will use this issue to track the work that needs to be done, in order to implement a better Catalog Page.

- [x] Create new paginated entities endpoint #12246
- [ ] Decoupling filters from having all the entities available in the frontend
  - [x] EntityOwnerPicker: #17514
  - [x] EntityProcessingStatusPicker: #17889
  - [x] EntityLifecyclePicker: #17888
  - [x] EntityOwnerPicker: #17514, #18199
[...]

PR #27839, opened 2024-11-26
Entity owner picker owners only rollback

vinzscam:

This PR rolls back #25366. Unfortunately, the approach used in #25366 does not work with a large catalog, as `getEntitiesByRefs` doesn't scale. #27167 attempts to address this issue by ensuring that `getEntitiesByRefs()`
 doesn't fail anymore, but the latency of fetching all the entities is still very high.

In this PR, we roll back to using `getEntityFacets()` to fetch the owners. This means that we won't be able to fetch the details of the entities, such as `displayName` or `title`. This tradeoff was decided to move toward #12247 but was probably overlooked when #25366 was merged. In case an adopter wishes to see more details of the entities, make sure to pass `ownerPickerMode="all"` to `CatalogIndexPage`:

```jsx
<CatalogIndexPage ownerPickerMode="all" />
```

 
### DIFF
--- a/plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
+++ b/plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
@@ -17,8 +17,7 @@ import { useApi } from '@backstage/core-plugin-api';
 import useAsyncFn from 'react-use/esm/useAsyncFn';
 import { catalogApiRef } from '../../api';
 import { useState } from 'react';
-import { Entity } from '@backstage/catalog-model';
-import get from 'lodash/get';
+import { Entity, parseEntityRef } from '@backstage/catalog-model';
 
 type FacetsCursor = {
   start: number;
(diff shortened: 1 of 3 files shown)


plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
```diff
@@ -52,37 +49,30 @@ export function useFacetsEntities({ enabled }: { enabled: boolean }) {
...
-      .then(resp =>
-        resp.items
-          .filter(entity => entity !== undefined)
-          .map(entity => entity as Entity)
+      .getEntityFacets({ facets: [facet] })
+      .then(response =>
+        response.facets[facet]
+          .map(e => e.value)
+          .map(ref => {
+            const { kind, name, namespace } = parseEntityRef(ref);
```
freben, line 59: so this will actually create a broken entity (potentially wrong casing for fields, definitely the kind)

freben (approved):
Sure, if we need to. But this control is in dire need of some more proper improvement, like virtualization to only render the entries in the viewport or so, and using only refs as the data model + the renderer using the catalog-node presentation api helper components

vinzscam:
> Sure, if we need to. But this control is in dire need of some more proper improvement, like virtualization to only render the entries in the viewport or so, and using only refs as the data model + the renderer using the catalog-node presentation api helper components

sure, but I don't think it's a problem right now as we already display a subset of the owners, adding more to the DOM as the user scrolls.

PR #27845, opened 2024-11-26
Added missing scaffolder management permission

awanlin:

Added `templateManagementPermission` to `scaffolderTemplatePermissions` array so it will get pickup by the `permissionIntegrationRouter`

### DIFF
--- a/plugins/scaffolder-common/src/permissions.ts
+++ b/plugins/scaffolder-common/src/permissions.ts
@@ -130,7 +130,6 @@ export const templateManagementPermission = createPermission({
 export const scaffolderTemplatePermissions = [
   templateParameterReadPermission,
   templateStepReadPermission,
-  templateManagementPermission,
 ];
 
 /**
@@ -157,4 +156,5 @@ export const scaffolderPermissions = [
   ...scaffolderTemplatePermissions,
   ...scaffolderActionPermissions,
   ...scaffolderTaskPermissions,
+  templateManagementPermission,
 ];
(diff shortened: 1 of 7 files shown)


acierto (approved):

plugins/scaffolder-common/src/permissions.ts
```diff
@@ -156,5 +157,4 @@ export const scaffolderPermissions = [
   ...scaffolderTemplatePermissions,
   ...scaffolderActionPermissions,
   ...scaffolderTaskPermissions,
-  templateManagementPermission,
```
vinzscam, line 159: mmm this used to be correct 🤔 
I think the problem should be somewhere else then

vinzscam (changes requested):

plugins/scaffolder-common/src/permissions.ts
vinzscam: So I think the way to go would be to pass `permissions: scaffolderPermissions` here: https://github.com/backstage/backstage/blob/c34fd4ed4ab9ac2e90d51c988d259b5e4be4f0b6/plugins/scaffolder-backend/src/service/router.ts#L448
but I'm not sure whether the integration router can properly handle duplicates at this point, as `scaffolderPermissions` contains resources permissions as well.
Short term another way would be to pass `permissions: [...scaffolderPermissions, templateManagementPermission]`

awanlin: I'll give this a try, with the current fix there was no duplicates based on my testing.

vinzscam: your fix should work. The issue is to group the new permission under `scaffolderTemplatePermissions` as they are resource permissions of type `scaffolder-template`

vinzscam (approved):
alright I've added the fix, it should be ready to go now 🚀

PR #27846, opened 2024-11-26
CLI: have module federation settings take in account all the exports defined in the `package.json`

davidfestal:

This PR enhances the behavior of the experimental support for module federation in the backstage CLI:
Until now exposed modules are a fixed value set to `src/index`.
With this PR, in the case of a remote container (not the main application), would add also look into the `package.json` of the built package, for an `exports` entry and, if it exists, use it to include alternate imports (like `./alpha`) into the  module federation exposed modules.
The change is compatible with the existing behavior, but only fixes the existing limitation that, for example, prevented accessing the `alpha` definitions in the module federation scenario.

### DIFF
--- a/.changeset/shaggy-tomatoes-travel.md
+++ b/.changeset/shaggy-tomatoes-travel.md
@@ -1,0 +1,7 @@
+---
+'@backstage/cli': patch
+---
+
+Enhance the behavior of the experimental support for module federation in the backstage CLI,
+by using the `package.json` exports (when present) to complete the list of exposed modules.
+This allows, for example, using exported `alpha` definitions through module federation.
--- a/packages/cli/src/commands/build/buildFrontend.ts
+++ b/packages/cli/src/commands/build/buildFrontend.ts
@@ -19,6 +19,7 @@ import { resolve as resolvePath } from 'path';
 import { buildBundle, getModuleFederationOptions } from '../../lib/bundler';
 import { getEnvironmentParallelism } from '../../lib/parallel';
 import { loadCliConfig } from '../../modules/config/lib/config';
+import { BackstagePackageJson } from '@backstage/cli-node';
 
 interface BuildAppOptions {
   targetDir: string;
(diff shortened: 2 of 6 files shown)


davidfestal:
Any idea whether this pretty simple and straightforward PR could get some review soon ?
cc @christophe-f

freben (approved):
Looks good to me - @Rugvip for final check?

Rugvip (changes requested):
Hmm, is it specifically things like `/alpha` that you're looking to make available? As in additional entry points that have features we'd want to load.

I think it's best to rely on the feature discovery for this, see https://github.com/backstage/backstage/blob/d849865b6fafdf7d7c8eb5b96e7d5fd933105e16/packages/cli/src/lib/packager/productionPack.ts#L177-L193. That way we can make sure we only include actual features in the remote bundle, rather than entry points like `/testUtils` or `/package.json`.

davidfestal:
> Hmm, is it specifically things like `/alpha` that you're looking to make available? As in additional entry points that have features we'd want to load.
> 
> I think it's best to rely on the feature discovery for this, see
> 
> https://github.com/backstage/backstage/blob/d849865b6fafdf7d7c8eb5b96e7d5fd933105e16/packages/cli/src/lib/packager/productionPack.ts#L177-L193
> 
> . That way we can make sure we only include actual features in the remote bundle, rather than entry points like `/testUtils` or `/package.json`.

Thanks for the pointer, I'm looking into it.

davidfestal:
> > Hmm, is it specifically things like `/alpha` that you're looking to make available? As in additional entry points that have features we'd want to load.
> > I think it's best to rely on the feature discovery for this, see
> > https://github.com/backstage/backstage/blob/d849865b6fafdf7d7c8eb5b96e7d5fd933105e16/packages/cli/src/lib/packager/productionPack.ts#L177-L193
> > 
> > . That way we can make sure we only include actual features in the remote bundle, rather than entry points like `/testUtils` or `/package.json`.
> 
> Thanks for the pointer, I'm looking into it.

@Rugvip I added commit c8dbf11b814a8286a2576d0bcd9e602212125c6d? Is it what you were expecting ?

davidfestal:
@Rugvip Any other review comment apart from the one I fixed ? The merging is still blocked by requested changes.

PR #27848, opened 2024-11-26
Added dependency check to create-app

awanlin:

Added dependency check before creating the app to ensure needed dependencies are in place such as Node, Yarn, and Python.

I'm very open to refactoring but feel this would be very helpful to have in place based on the issues Adopters have had. 

### DIFF
--- a/packages/create-app/src/createApp.ts
+++ b/packages/create-app/src/createApp.ts
@@ -31,9 +31,12 @@ import {
   tryInitGitRepository,
   readGitConfig,
   fetchYarnLockSeedTask,
+  tryCommandForVersion,
 } from './lib/tasks';
 
 const DEFAULT_BRANCH = 'master';
+// Uses the same 'N.x' format as GitHub Actions node-version matrices for easy find-and-replace
+const SUPPORTED_NODE_VERSIONS = ['22.x', '24.x'];
 
 export default async (opts: OptionValues): Promise<void> => {
   const answers: Answers = await inquirer.prompt([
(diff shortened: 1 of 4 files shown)


freben:
Hi! Just wanted to say that the tests are unhappy on this one.

packages/create-app/src/createApp.ts
```diff
@@ -75,6 +77,57 @@ export default async (opts: OptionValues): Promise<void> => {
...
+    Task.error(yarn.error);
+    hasPrerequisiteError = true;
+  }
+
+  if (python.error) {
+    Task.log(chalk.yellow(`Warning: ${python.error}`));
+    hasPrerequisiteError = true;
+  }
+
+  if (major < Math.min(...NODE_LTS_VERSIONS)) {
```
freben: I believe we should make this strict on majors, ie not allow 21 either.

And even better: use the actual exact LTS as a minimum per major (iron was 20.9.0 right? and jod was 22.11.0), because there were some rare features that exist in 22.11 that weren't in 22.0 i believe.

packages/create-app/src/createApp.ts
```diff
@@ -75,6 +77,57 @@ export default async (opts: OptionValues): Promise<void> => {
     ? resolvePath(paths.targetDir, opts.path)
     : resolvePath(paths.targetDir, answers.name);
 
+  // Prerequisite check
+  let hasPrerequisiteError = false;
+  const [major, minor, patch] = process.versions.node.split('.').map(Number);
+  const yarn = await tryCommandForVersion('yarn -v');
+  const python = await tryCommandForVersion('python3 --version');
```
freben: do we require that the binary `python3` exists? or can it happen that there are aliases `python`->`python3`? i really am unsure

packages/create-app/src/createApp.ts
awanlin: Yes, I missed that, we don't support odd number versions of Node and this would allow that.

How would I be able to find which minor version was used when a version became LTS? Is this in the Node docs somewhere?

packages/create-app/src/createApp.ts
awanlin: Good call out, forgot about that. Will give this another think to see what I can come up with to handle this case. 👍

packages/create-app/src/createApp.ts
freben: Ah you know what, I probably was wrong here. I probably read too much into the blog entries like [this one](https://nodejs.org/en/blog/release/v22.16.0); there are several 22.x blogs that have the same `(LTS)` string on them.

freben: So maybe it's perfectly fine to go by `N.0.0` for now as the standard and _if_ we need to be stricter than that because of a missing/known bad feature in a certain minor release, we can leverage that as needed later down the line (if the version comparison code supports it; maybe the `semver` library can help here)

packages/create-app/src/createApp.ts
awanlin: Updated to handle both

packages/create-app/src/createApp.ts
awanlin: Updated to not allow odd versions

awanlin:
Hi @freben, brought this one back to life, got it up to date and addressed your comments from your last review, any chance you can give it another look, please?

packages/create-app/src/createApp.ts
```diff
@@ -31,11 +31,78 @@ import {
...
+      `Node version ${major} is not a supported LTS version. Please use a supported version (${SUPPORTED_NODE_VERSIONS.join(
+        ' or ',
+      )}).`,
+    );
+    hasPrerequisiteError = true;
+  }
+
+  if (hasPrerequisiteError) {
+    Task.log(
+      'It seems that something went wrong when validating the prerequisites 🤔',
```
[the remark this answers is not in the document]
awanlin: Already in the code in other places.

packages/create-app/src/createApp.ts
```diff
@@ -31,11 +31,78 @@ import {
...
+    );
+    Task.log(
+      'For help with setting up prerequisites, see https://backstage.io/docs/getting-started/#prerequisites',
+    );
+
+    Task.error('🔥  Failed to validate needed prerequisites!');
```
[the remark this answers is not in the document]
awanlin: Already in the code in other places.

packages/create-app/src/lib/tasks.ts
```diff
@@ -336,3 +336,20 @@ export async function fetchYarnLockSeedTask(dir: string) {
...
+ *
+ * @param command - command to run
+ * @returns object - version with the version found and N/A if not found, error contains any error and undefined otherwise
+ */
+export async function tryCommandForVersion(command: string) {
+  try {
+    const result = await exec(command);
+
+    const version = (result.stdout || result.stderr).trim();
+    return { version: version || 'N/A', error: undefined };
```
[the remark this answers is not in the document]
freben: or better `result.stdout?.trim() || result.stderr?.trim() || ''` I would say. Defends against both nullish values and empty strings.

freben (approved):

Issue #16692, opened 2023-03-03
💬 RFC: Standardization on entity provider clientside throttling


bforbis:
### 🔖 Need

Ensuring that backstage doesn't become a problem for external systems when fetching data requires careful consideration on concurrency control / maximum throughput. Without clientside throttling support, backstage could hit an external system fast enough that it could cause degradation of services or temporary outages.

Take the Bitbucket server discovery provider for example. The current crawler implementation right now will fetch from every single repository in an organization using a simple `for await (const repo of repositories) {}` loop. Without throttle control, this could send a massive amount of requests against bitbucket in large organizations.

### 🎉 Proposal

Entity providers should begin to standardize on strategies for adding configurable support for throttling requests against external systems. One possibility would be to add functionality from either the [bottleneck](https://github.com/SGrondin/bottleneck) or [limiter](https://github.com/jhurliman/node-rate-limiter) projects and add standard properties to the task scheduler configurations that can be used by entity provider implementations.

An example of how either of these might be configured in a provider could be:

```ts
BitbucketServerEntityProvider.fromConfig(env.config, {
  schedule: {
    frequency: { hours: 6 },
    timeout: { minutes: 15 },
  },
  // Using limiter API
  limiter: {
    tokensPerInterval: 5,
    interval: 'second',
  },
  // Using bottleneck API
  bottleneck: {
    maxConcurrent: 5,
    minTime: 200
  },
});
```

### 〽️ Alternatives

As @taras suggests in the comments, providers could be standardized on using "Incremental Entity Providers" which may have some form of throttle control built into them by controlling burst and rests.

### ❌ Risks

It is unlikely to be able to update all entity providers to have uniform throttling features at once, so the standardization won't be there initially. We would also need to be extra vigilant that contributors adhere to the throttling standard.

### 👀 Have you spent some time to check if this RFC has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

taras:
@bforbis another option worth looking into is to refactor Bitbucket Entity Provider to an Incremental Entity Provider. Then the provider will automatically backoff when it hits rate limit. We could even extend Incremental Ingestion backend to support proactively backing-off before hitting the rate limit. You can read more about it here https://backstage.io/blog/2023/01/31/incremental-entity-provider

[...]

PR #27850, opened 2024-11-26
feat: Throttle calls to Bitbucket Server API

kjjuno:

This adds some throttling to the bitbucket server api calls. This helps to avoid flooding the server with too much traffic, and potentially taking down the server.

### DIFF
--- a/.changeset/cool-days-sparkle.md
+++ b/.changeset/cool-days-sparkle.md
@@ -1,0 +1,6 @@
+---
+'@backstage/plugin-catalog-backend-module-bitbucket-server': minor
+'@backstage/backend-defaults': minor
+---
+
+Throttles Bitbucket Server API calls
--- a/packages/backend-defaults/package.json
+++ b/packages/backend-defaults/package.json
@@ -169,6 +169,7 @@
     "node-fetch": "^2.7.0",
     "node-forge": "^1.3.1",
     "p-limit": "^3.1.0",
+    "p-throttle": "^4.1.1",
     "path-to-regexp": "^8.0.0",
     "pg": "^8.11.3",
     "pg-connection-string": "^2.3.0",
--- a/packages/backend-defaults/src/entrypoints/urlReader/lib/BitbucketServerUrlReader.ts
+++ b/packages/backend-defaults/src/entrypoints/urlReader/lib/BitbucketServerUrlReader.ts
@@ -46,7 +46,7 @@ const throttle = pThrottle({
 });
 
 const throttledFetch = throttle(
-  async (url: fetch.RequestInfo, options?: fetch.RequestInit) => {
+  async (url: RequestInfo, options?: RequestInit) => {
     return await fetch(url, options);
   },
 );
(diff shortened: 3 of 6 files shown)


cal5barton:
@Rugvip as discussed in the Framework SIG these are the new proposed changes, please review and add anyone else that needs to have eyes on this so we can push this forward. Thanks!

freben (approved):
Sorry I rudely rebased this to try to get it in for today's release.

The two separate plimits of course means we can reach 2 rps instead of 1. But it's a start!

goququ:
Hi all! Great news. Will be cool if something like this will be implemented in gitlab discovery process also

webark:
Was this inteded to be eventually configurable? We handle our throttling seperatly, so adding in this in this central place kind of tanked a lot of our performance.. And it's impossible to modify this at all without just completely overwriting the whole URL reader.. :(

freben:
@webark Sorry to hear that. And this new way of doing things is not sufficient to replace what you had there before? Alright, would you mind opening an issue and/or pr to see if we can improve on that?

webark:
@freben No. :/ Each bitbucket server key gets its own throttle that's configurable, and we have a lower threshold for personal tokens than the backstage token. Those are still much higher. I think it's like 1000 with a refill rate of 60 per second, and personally tokens we do 100 with a refill rate of 10 per second.

For ingestion, we use the incremental, and we moved off of locations a while ago, so there are no network calls to bitbucket (or anything but our cache) during the processing loop.

We also have places where we are using the url reader to fetch things on the users behalf. So like displaying a changelog, readme, configuration, or a contributing file on an entity detail page. Here the one call a second gets very noticeable. We could move these calls off of the url reader, but then we loose the unified interface between github, bitbucket, azure, etc.

If the value was configurable, then we could just set that to be the value we need, or optionally disable it altogether.

I'll look into getting a PR up to configure this setting though.

bforbis:
Relates to #16692. @webark would having the rate limit at the provider level be better than at the reader level? Sounds like you want different limits for user triggered traffic vs automatic discovery traffic?

webark:
Yea. I don't think there's should be a blanket artificial throttle (one that is not driven by a 429) outside of a place like the provider or processor or somewhere that is already iterating over a dataset.

We use the "incremental ingestors" and have enjoyed the kind of mechanism that is uses. Its interface for controlling its burst length and interval is a bit confusing at first (we initially thought the interval was between bursts rather that the interval of starting bursts for example). But how it allows you flexibility in how to process through the data set. Like with bitbucket server, we have about 7000 repos in about 400 projects, so we iterate over the projects, doing a search for catalog files (or other file types you care about processing) per project, and then iterating through those responses to get the final set of entities to ingest. And we can do that due to the flexibility that is available with the cursor you can provide.

webark:
@knowacki23 we just patched that throttle library to be a no op, and it allows up to bypass the throttle. 

All for the proposal 👍

knowacki23:
@webark do you have a link to the patch?

PR #27853, opened 2024-11-27
[plugin-techdocs] Add props to increase flexibility and reusability 

nikolarAutodesk:

This PR aims to add optional props to `TechDocsCustomHome` that surfaces existing props and new props which should make it easier for adopters to reuse existing Backstage code without having to write a bunch of custom logic. 
For us at Autodesk, this will remove an entire forklift of our techdocs page. 

### Changes

- Add optional props to `TechDocCustomHome` to allow for more flexibility to:
  - Adjust filter 
  - Modify or hide page header/subtitle 
  - Hide support button 
  - modify props already available from `TechDocsIndexPage`
- Add new Grid option called `InfoCardGrid` which is a more customizable card option for the Docs grid.
  -  this new grid option avoids use of CardMedia, surfaces overridable classNames, and allows for doc links to be customized -- see in video below, links can go to techdocs or external sites (aka: confluence, company docs..) 
- Expose existing `CustomDocsPanel` so that it can be used independently if desired.
  -  we use this exact component in 2 other places in our Backstage instance, we wanted to surface this to avoid duplication (we are betting we are not the only ones too 😉 )
  
Here are screenshots/records with all of the new props modified:
Screenshot 2024-12-02 at 8 49 23 PM
Screenshot 2024-12-02 at 8 49 38 PM
Screenshot 2024-12-03 at 4 31 10 PM

https://github.com/user-attachments/assets/cc5c388c-9cd9-4945-bd4c-22f9bde10258

Here are screenshots of the default behavior (unchanged):
Screenshot 2024-12-03 at 4 06 07 PM
Screenshot 2024-12-03 at 4 06 15 PM


plugins/techdocs/src/home/components/TechDocsCustomHome.tsx
```diff
@@ -153,7 +172,7 @@ export const TechDocsCustomHome = (props: TechDocsCustomHomeProps) => {
     error,
   } = useAsync(async () => {
     const response = await catalogApi.getEntities({
-      filter: {
+      filter: filter || {
```
awanlin, line 175: This should probably be merging these together. Not having the TechDocs annotation essentially means you have the same functionality as the Catalog index page.

awanlin (commented):
Hi @nikolarAutodesk, I know this is still Draft but wanted to mention a couple of things worth considering:

1. There's a PR to add pagination to this area, might be worth looking to see if there's any conflicts between the changes - https://github.com/backstage/backstage/pull/27694
2. Given the work being done for the [New Frontend System](https://backstage.io/docs/frontend-system/architecture/index) do these changes make sense at this time? Would it be better to implement this in that system instead?

I left one small comment on a part that caught my eye but did not look too closely at anything else.

nikolarAutodesk:
Hi @awanlin thank you for the initial look, I updated that suggestion. 

1. I will have merge conflicts with https://github.com/backstage/backstage/pull/27694 when it merges, but there is no conflicting logic. We are just adding props in the same place. 
2. I took a look at the new frontend system and though we will be able to extend the tech-docs plugin I do think this PR is still valuable as is. This PR reduces the amount of code adopters need to slightly modify the techdocs page. It should give users more flexibility without having to write an extension. But I am open to thoughts if anyone disagrees!

plugins/techdocs/src/home/components/TechDocsCustomHome.tsx
```diff
@@ -119,15 +147,21 @@ const CustomPanel = ({
```
nikolarAutodesk, line 147: A question for maintainers:
We have another use case where we would like to filter documentation sections based on a config file (in order to prevent teams at the company just registering their docs in any section in their yaml files). I was thinking to pass the config along with the entity for optional use in line 144 above like so:
```
const configApi: ConfigApi = useApi(configApiRef);
return (
      typeof config.filterPredicate === 'function' &&
      config.filterPredicate(entity, configApi)
    );
```
But I wasn't sure if that was an anit-pattern or if there is a better recommended way. Any tips?

plugins/techdocs/src/home/components/Grids/InfoCardGrid.tsx
```diff
@@ -0,0 +1,94 @@
...
+  linkContent?: string | JSX.Element;
+  linkDest?: (entity: Entity) => string;
+};
+
+/**
+ * Component which accepts a list of entities and renders a info card for each entity
+ *
+ * @public
+ */
+export const InfoCardGird = (props: InfoCardGirdProps) => {
```
squid-ney:
```suggestion
export const InfoCardGrid = (props: InfoCardGridProps) => {
```

nikolarAutodesk: Oh I think you might be reviewing an old commit, that has been updated. Thank you!

plugins/techdocs/src/home/components/Grids/InfoCardGrid.tsx
```diff
@@ -0,0 +1,99 @@
...
+      namespace: toLowerMaybe(entity.metadata.namespace ?? 'default', config),
+      kind: toLowerMaybe(entity.kind, config),
+      name: toLowerMaybe(entity.metadata.name, config),
+    });
+  };
+
+  if (!entities) return null;
+  return (
+    <ItemCardGrid data-testid="info-card-container">
+      {!entities?.length
```
johnphilip283, line 78: should this check here also handle the empty list case and return null early? I don't see a situation where we'd want to return the grid container with no items in it.

plugins/techdocs/src/home/components/TechDocsCustomHome.tsx
```diff
@@ -120,16 +148,22 @@ const CustomPanel = ({
...
   return (
     <>
-      <ContentHeader title={config.title} description={config.description}>
-        {index === 0 ? (
-          <SupportButton>
-            Discover documentation in your ecosystem.
-          </SupportButton>
-        ) : null}
-      </ContentHeader>
+      {config.panelProps?.showHeader !== false && (
```
johnphilip283, line 151:
```suggestion
      {!!config.panelProps?.showHeader && (
```

plugins/techdocs/src/home/components/TechDocsCustomHome.tsx
```diff
@@ -120,16 +148,22 @@ const CustomPanel = ({
...
+        <ContentHeader title={config.title} description={config.description}>
+          {index === 0 && config.panelProps?.showSupport !== false && (
```
johnphilip283, line 153:
```suggestion
          {index === 0 && !!config.panelProps?.showSupport && (
```

plugins/techdocs/src/home/components/TechDocsPageWrapper.tsx
```diff
@@ -34,16 +37,18 @@ export type TechDocsPageWrapperProps = {
...
+    `Documentation available in ${
+      configApi.getOptionalString('organization.name') ?? 'Backstage'
+    }`;
 
   return (
     <PageWithHeader
-      title="Documentation"
-      subtitle={generatedSubtitle}
+      title={title || 'Documentation'}
+      subtitle={showSubtitle ? generatedSubtitle : undefined}
```
johnphilip283, line 51: With these props, could we also follow the approach I mentioned above, where we explicitly pass the react components we want to render as props, but just keep the existing defaults when those props aren't passed in?

johnphilip283 (commented):
Thanks so much for these changes - it will be very useful to have these components be decoupled. I've noted some changes to the props that we'd like to have, but this is a great addition.

plugins/techdocs/src/home/components/TechDocsCustomHome.tsx
nikolarAutodesk: At our company, we allow teams to manage & register their own catalog-info.yaml files. So they can technically add whatever they want to their yaml files.
We want a way to control our docs page sections (potentially through a config file), because if we filter on tags or another field in the yaml file - any team can add the required filter to their yaml file. 
So say our filterPredicate is:
```
filterPredicate: entity =>
          entity?.metadata?.tags?.includes('golden-path') ?? false
```
any team can add this to their yaml file:
```
metadata:
  tags: 
    - 'golden-path'
```
and it would show up in our docs section for anyone to view.

plugins/techdocs/src/home/components/TechDocsCustomHome.tsx
nikolarAutodesk: updated 👍

plugins/techdocs/src/home/components/TechDocsCustomHome.tsx
nikolarAutodesk: updated 👍

plugins/techdocs/src/home/components/Grids/InfoCardGrid.tsx
nikolarAutodesk: updated

plugins/techdocs/src/home/components/DefaultTechDocsHome.tsx
```diff
@@ -46,15 +46,28 @@ export type DefaultTechDocsHomeProps = TechDocsIndexPageProps;
  * @public
  */
 export const DefaultTechDocsHome = (props: TechDocsIndexPageProps) => {
-  const { initialFilter = 'owned', columns, actions, ownerPickerMode } = props;
+  const {
+    initialFilter = 'owned',
+    columns,
+    actions,
+    ownerPickerMode,
+    showHeader = true,
```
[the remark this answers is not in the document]
nikolarAutodesk: Hi @johnphilip283 thank you for the review. I am a little confused by what you mean here and the last comment. I was trying different ways of this & realized I need clarification. Can you please provide an example of what you mean?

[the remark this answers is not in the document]
nikolarAutodesk: Actually, @johnphilip283 I _think_ I made the updates with the props as you requested above. Please take a look 🙏

nikolarAutodesk:
Hi @awanlin @Rugvip @johnphilip283 sorry to poke, I'm concerned this PR got lost in the weeds, were hoping to get this PR merged in soon to update our own code 🤞 🙏 Can I get some eyes on this please?

docs/features/techdocs/how-to-guides.md
```diff
@@ -145,18 +149,51 @@ const techDocsTabsConfig = [
...
       path="/docs"
-      element={<TechDocsCustomHome tabsConfig={techDocsTabsConfig} />}
+      element={
+        <TechDocsCustomHome
+          tabsConfig={techDocsTabsConfig}
+          filter={{
+            kind: ['Location', 'Resource', 'Component'],
+            'metadata.annotations.featured-docs': CATALOG_FILTER_EXISTS,
+          }}
+          CustomPageWrapper={({ children }: React.PropsWithChildren<{}>) => (<PageWithHeader title="Docs" themeId="documentation">{children}</PageWithHeader>)}
```
johnphilip283, line 194: small nitpick - would you be able to define these props as variables above and then pass them in? this is a bit hard to read at first glance.

plugins/techdocs/src/home/components/Grids/InfoCardGrid.tsx
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  if (!entities || !entities?.length) return null;
+  return (
+    <ItemCardGrid data-testid="info-card-container">
+      {entities.map(entity => (
+        <InfoCard
+          key={entity.metadata.name}
+          data-testid={entity?.metadata?.title}
+          title={entity?.metadata?.title || entity?.metadata?.name}
```
johnphilip283, line 82: this might be a good use of `useEntityPresentation()` from https://backstage.io/docs/reference/plugin-catalog-react.useentitypresentation/ - just to standardize how this entity is presented.

johnphilip283:
The PR looks good to go - just left two small comments. Once they're resolved, I can approve and merge!

plugins/techdocs/src/home/components/Grids/InfoCardGrid.tsx
nikolarAutodesk: Thanks for the re-review! 
It looks like `useEntityPresentation()` is a [hook that deals with individual entities](https://github.com/backstage/backstage/blob/8b4ba413a94df01d14aebdd6f6e5ab0b9b165cd0/plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx#L90). 
Do you prefer that I add a component that deals with individual entities which uses InfoCard? 
Or should I leave this as is?

nikolarAutodesk: Just want to point out that this is also how [DocsCardGrid](https://github.com/backstage/backstage/blob/8b4ba413a94df01d14aebdd6f6e5ab0b9b165cd0/plugins/techdocs/src/home/components/Grids/DocsCardGrid.tsx#L59) gets the title, so it is consistent with that

johnphilip283: There's a good example of how to use the entity presentation API for multiple entities [here](https://github.com/backstage/backstage/blob/8b4ba413a94df01d14aebdd6f6e5ab0b9b165cd0/plugins/scaffolder/src/components/fields/EntityPicker/EntityPicker.tsx#L97) that might be useful to borrow from!

nikolarAutodesk: Thanks @johnphilip283 I updated both InfoCardGrid and DocsCardGrid to use the presentation API

johnphilip283 (approved):

Rugvip (approved):
Approving the `yarn.lock` changes

PR #27876, opened 2024-11-27
googleAuth: `includeGrantedScopes` to persist scopes across refreshes

benjdlambert:
Need to pass this through otherwise on refreshes of both the token and the browser can lose the scopes that have already been requested which can lead to popups for auth all the time across tabs and across pages.


jhaals (approved):

Rugvip (approved):
:shipit:

camilaibs (approved):

Issue #27840, opened 2024-11-26
🐛 Bug Report: [Unprocessed Entities] doesn't render timestamps correctly for postgres DB


knowacki23:
### 📜 Description

When using pg database (postgres:16.1-alpine) instead of getting correct timestamps for `LAST DISCOVERY AT` and `NEXT REFRESH AT` we are getting `Invalid DateTime`.
Tested locally with following config in app-config.yaml:
```
backend:
  database:
    client: pg
    connection:
      host: localhost
      port: 5432
      user: backstage
      password: backstage
```
database used: `postgres:16.1-alpine`

Can we add handling of different time formats to [convertTimeToLocalTimezone](https://github.com/backstage/backstage/blob/master/plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx#L95-L107)?

### 👍 Expected behavior

It should correctly display `LAST DISCOVERY AT` and `NEXT REFRESH AT` timestamps.
Like it does for `better-sqlite3` in-memory database:
![image](https://github.com/user-attachments/assets/c45bcba8-c065-4f7d-ae86-05a509005067)
config:
```
backend:
  database:
    client: better-sqlite3
    connection: ':memory:'
```

### 👎 Actual Behavior with Screenshots

`Invalid DateTime` instead of actual date for postgres DB.
![image](https://github.com/user-attachments/assets/8c91e2bc-ebce-4260-8cc2-5e59d32be052)

### 👟 Reproduction steps

1. set database in your backstage instance to `pg`
2. start database
3. start backstage with `yarn dev`
4. go to http://localhost:3000/devtools/unprocessed-entities

### 📃 Provide the context for the Bug.

After some debugging I found following error
```
{
    "reason": "unparsable",
    "explanation": "the input \"2024-11-26T14:57:09.131Z\" can't be parsed as format yyyy-MM-dd hh:mm:ss"
}
```
which was caused by [convertTimeToLocalTimezone](https://github.com/backstage/backstage/blob/master/plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx#L95-L107) function.
The problem is that when using PG database the timestamp format differs from the one expected in the function above.
Format returned from PG:
```
2024-11-26T14:57:09.131Z
```
and from what I understand it expects to get timestamp in following format:
```
'yyyy-MM-dd hh:mm:ss',
    {
      zone: 'UTC',
    },
```

### 🖥️ Your Environment

I'm running Backstage  v1.33.4.
[...]

PR #27881, opened 2024-11-27
Added support for ISO timestamps to convertTimeToLocalTimezone

jacri901:
Related to [#27840](https://github.com/backstage/backstage/issues/27840)

Fixed a bug where timestamps form postgres database did not work. Now a valid `DateTime` is displayed as seen below.
image

### DIFF
--- a/plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
+++ b/plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
@@ -93,13 +93,17 @@ const RenderErrorContext = ({
  * easily understand the times.
  */
 const convertTimeToLocalTimezone = (strDateTime: string | Date) => {
-  const dateTime = DateTime.fromFormat(
+  let dateTime = DateTime.fromFormat(
     strDateTime.toLocaleString(),
     'yyyy-MM-dd hh:mm:ss',
     {
       zone: 'UTC',
     },
   );
+
+  if (!dateTime.isValid) {
+    dateTime = DateTime.fromISO(strDateTime.toLocaleString());
+  }
 
   const dateTimeLocalTz = dateTime.setZone(DateTime.local().zoneName);
 
--- a/.changeset/fifty-tigers-suffer.md
+++ b/.changeset/fifty-tigers-suffer.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-unprocessed-entities': minor
+---
+
+Fixed a bug in convertTimeToLocalTimezone where there was no support for ISO dates


jacri901:
@Rugvip, @camilaibs I don't understand why this action fails?

plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
```diff
@@ -93,14 +93,18 @@ const RenderErrorContext = ({
...
+  let dateTime = DateTime.fromFormat(
     strDateTime.toLocaleString(),
     'yyyy-MM-dd hh:mm:ss',
     {
       zone: 'UTC',
     },
   );
 
+  if (!dateTime.isValid) {
+    dateTime = DateTime.fromISO(strDateTime.toLocaleString());
```
camilaibs, line 105: Quick question: Is it possible to use another format besides ISO? If yes, how about adding a configuration to `app-config.yaml` so the user can specify the date format string there?

jacri901: The problem is that timestamps like "2024-11-26T14:57:09.131Z" that includes T and Z cannot be parsed with `toLocaleString`. I tried this but it did not work for me. 

```typescript

'yyyy-MM-ddThh:mm:ss.SSSZ',
    {
      zone: 'UTC',

```

I don't know what things could trigger `FailedEntities` from the first place and what types of time formats to expect. The documentation from Luxton shows that `fromISO` support a wide range of formats. It may be possible to replace the entire `toLocalString()` part with only `fromISO` but I don't know if that could break functionality with something else. See Luxton [documentation](https://moment.github.io/luxon/#/parsing?id=iso-8601)

I assume that, providing the input format in `app-config.yaml` would need some extra parsing implemented in this case.
However, the output format feels more convient to specify in the that file. I could create a separate issue/PR for it if sounds useful.

I'm thinking of this 
https://github.com/backstage/backstage/blob/master/plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx?plain=1#L106

jacri901:
@Rugvip and @camilaibs. This PR is ready to be reviewed and merged if wanted. Let me know if you have any more questions.

benjdlambert:
@momiskill @jacri901 sorry this PR landed over the dreaded holidays and vacation time so seemed to slip through the cracks whilst folks where out on vacations!

Reopened!

plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
```diff
@@ -93,14 +93,18 @@ const RenderErrorContext = ({
  * easily understand the times.
  */
 const convertTimeToLocalTimezone = (strDateTime: string | Date) => {
-  const dateTime = DateTime.fromFormat(
+  let dateTime = DateTime.fromFormat(
```
freben, line 96: This code confuses me (what was there before).

Why would we ever stringify _in a local unknown format and then try to parse it again_ when the input was already a Date that has perfect strict semantics already? :D

This should all be changed around so that IF the type is string, then try to parse the two well known formats (the one that was here before `yyyy-MM-dd hh:mm:ss`, as well as ISO, no `toLocaleString` necessary at all. But if it's NOT string, then it's a Date, so use `DateTime.fromJSDate(strDateTime)` on it instead.

Chris-Lee-Ca:
Bringing attention to this PR

freben:
I can try to reopen it, but the original author needs to follow up. Otherwise, others may try to take a stab as well at fixing it.

plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
freben: This function should also have some tests added to show what is being covered.

PR #27900, opened 2024-11-28
Updated docs of search

AmbrishRamachandiran:
Updated Note design

### DIFF
--- a/docs/features/search/getting-started.md
+++ b/docs/features/search/getting-started.md
@@ -301,11 +301,15 @@ indexBuilder.addCollator({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


AmbrishRamachandiran:
Commenting for avoid stale

awanlin (changes requested):
Thanks for the contribution @AmbrishRamachandiran, but I don't think this makes sense to change as it's all focused on the Legacy Backend.

Would you be open to re-writing this in the context of the New Backend System? The entire section under the Backend heading needs to be re-done: https://backstage.io/docs/features/search/getting-started#backend

PR #27909, opened 2024-11-28
simplify catalog testing by making the mock type match all interfaces

freben:
This mock facility used to be tricky to leverage, since there's this duality of `CatalogService` / `CatalogApi`. Instead of making several mock functions (like having both `catalogServiceMock` and `catalogApiMock`, I opted to try this instead. The other options lead to a ton of typescript errors in already written tests.


Rugvip (commented):
Hmm, shouldn't `catalog-react` export the mock for `CatalogApi`?

freben:
@Rugvip Oh, yeah, it does, for frontend code. But there's a bunch of _backend_ code out there that accepts `CatalogApi` as the input type, since `CatalogService` was invented very recently. At least to bide us over, this smooths the transition.

camilaibs (approved):

PR #27911, opened 2024-11-28
fix: ensure tags are handled by the generated clients

aramissennyeydd:

Discord thread: https://discord.com/channels/687207715902193673/1034089724664610938/1309619723729768448

By default, the openAPI generator creates an API file per tag, and then places operations with those tags in the various files - duplicating if necessary. My initial testing didn't use any tags, so I didn't stumble on this behavior. This PR updates our rendering to use SupportingFiles instead of the API templates to directly access `apiInfo` and render all tags into a single API class instead of separating it.


packages/repo-tools/src/commands/package/schema/openapi/generate/client.ts
```diff
@@ -37,12 +39,11 @@ async function generate(
...
-  const additionalProperties = clientAdditionalProperties
-    ? `--additional-properties=${clientAdditionalProperties}`
-    : '';
-  mkdirpSync(resolvedOutputDirectory);
+  const additionalProperties = toGeneratorAdditionalProperties({
+    initialValue: clientAdditionalProperties,
+  });
 
-  await fs.mkdirp(resolvedOutputDirectory);
+  await fs.emptyDir(resolvedOutputDirectory);
```
Rugvip, line 46: Oh that's a handy one 👍

packages/catalog-client/src/schema/openapi/generated/apis/Api.client.ts
Rugvip, line 1: I'd expect this one to be replacing `DefaultApi.client.ts` in this PR, same with the server. It's not a net new thing right?

Rugvip (commented):
Ah, nice! 👍 

Had to read up on what tags are 😅 

Need changesets for the catalog changes too

Rugvip (commented):
and getting ridd of https://github.com/backstage/backstage/blob/master/plugins/events-node/src/generated/index.ts with friends right?

aramissennyeydd:
@Rugvip We use the generated server router in `createEventBusRouter` in the events-backend, so we'd want to keep those files as well

Rugvip (approved):
Oh I see what's going on ._.

GitHub is showing `plugins/events-backend/src/schema/openapi.generated.ts` as moved to `plugins/events-backend/src/schema/openapi/generated/router.ts` rather than deleted, which got me confused.

Yep let's :shipit:! 🎉

kevspl:
@aramissennyeydd Hey there, quick bug I found with this. If you have tags enabled for everything in your OpenAPI yaml spec, the `plugins/plugin-backend/src/schema/openapi/generated/apis/index.ts` file will still try to export `export * from './DefaultApi.server';` which won't exist since everything will exist in a tagged file.

PR #27915, opened 2024-11-28
Implement `/entities` in terms of `queryEntities`

freben:
In very large catalogs, it is a risky proposition to allow callers to use the `/entities` endpoint indiscriminately. It reads every single entity out of the database in a single operation which is heavy for the database, consumes huge amounts of heap space for the service, and may even run into limits where the json serialization throws an exception because the payload hits hard restriction limits.

To get around this problem, implement the `/entities` endpoint on top of `queryEntities` based pagination, and stream out pages of JSON data to the response as they arrive.

On my machine, a 65k entity catalog that gives a 77MB json response on the endpoint, is curled out in 2.2 seconds. The old code is decidedly faster at 1.8 seconds. These timings will obviously vary depending on circumstances.

### DIFF
--- a/plugins/catalog-backend/src/service/createRouter.ts
+++ b/plugins/catalog-backend/src/service/createRouter.ts
@@ -19,7 +19,6 @@ import {
   ANNOTATION_LOCATION,
   ANNOTATION_ORIGIN_LOCATION,
   Entity,
-  parseEntityRef,
   parseLocationRef,
   stringifyEntityRef,
 } from '@backstage/catalog-model';
@@ -42,8 +41,10 @@ import { parseEntityFacetParams } from './request/parseEntityFacetParams';
 import { parseEntityOrderParams } from './request/parseEntityOrderParams';
 import { LocationService, RefreshService } from './types';
 import {
+  createEntityArrayJsonStream,
   disallowReadonlyMode,
   encodeCursor,
+  expandLegacyCompoundRelationRefsInResponse,
   locationInput,
   validateRequestBody,
 } from './util';
@@ -170,17 +171,15 @@ export async function createRouter(
         // For other read-the-entire-world cases, use queryEntities and stream
         // out results.
 
-        // Imitate the httpRouter behavior of pretty-printing in development
-        const prettyPrint = process.env.NODE_ENV === 'development';
-
         // The write lock is used for back pressure, preventing slow readers
         // from forcing our read loop to pile up response data in userspace
         // buffers faster than the kernel buffer is emptied.
         // https://nodejs.org/api/http.html#http_response_write_chunk_encoding_callback
         let writeLock: DeferredPromise | undefined;
         const controller = new AbortController();
+        const signal = controller.signal;
         req.on('end', () => {
-          controller.abort();
+          controller.abort(new Error('Client closed connection'));
           writeLock?.resolve();
           writeLock = undefined;
         });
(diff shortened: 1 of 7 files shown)


plugins/catalog-backend/src/service/createRouter.ts
```diff
@@ -135,24 +137,139 @@ export async function createRouter(
...
+          writeLock?.resolve();
+          writeLock = undefined;
+        });
+
+        res.setHeader('Content-Type', 'application/json; charset=utf-8');
+        res.status(200);
+        res.flushHeaders();
+        res.write(prettyPrint ? '[\n' : '[');
+
+        const limit = 10000;
```
camilaibs, line 197: Is there an opportunity to make this limit configurable?

plugins/catalog-backend/src/service/createRouter.ts
```diff
@@ -135,24 +137,139 @@ export async function createRouter(
...
+              await writeLock;
+            }
+
+            if (controller.signal.aborted) {
+              res.end();
+              return;
+            }
+
+            if (!res.write(data)) {
+              // The kernel buffer is full. Create the lock but do not await it
```
vinzscam, line 254: Will `data` be lost in this case? 🤔

camilaibs (approved):
So much good stuff, left minor suggestions, please take a look 🙂

plugins/catalog-backend/src/service/createRouter.ts
freben: No, this is just an optimization. We just stream out blocks as fast as the reader consumes them, instead of as fast as we can produce them (which fills up our memory when the consumer is slow). This should actually deal with tricky scenarios where large numbers of consumers simultaneously consume the entire database, even if they are relatively slow.

plugins/catalog-backend/src/service/createRouter.ts
freben: I try to keep the number of options to a minimum, just because they risk causing maintenance burden and confusion :) The exact value here isn't super important; it tries to strike a balance between memory use and streaming performance. The only reason that this number might get problematic, is if individual entities are so huge that even 10k of them are hard to fit in memory. But at that point you probably have huge issues 🤔

plugins/catalog-backend/src/service/createRouter.ts
```diff
@@ -135,24 +137,139 @@ export async function createRouter(
...
+          );
+
+          if (result.items.length) {
+            // TODO(freben): This is added as a compatibility guarantee, until we can be
+            // sure that all adopters have re-stitched their entities so that the new
+            // targetRef field is present on them, and that they have stopped consuming
+            // the now-removed old field
+            // TODO(jhaals): Remove this in April 2022
+            for (const entity of result.items) {
+              if (entity.relations) {
```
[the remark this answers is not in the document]
freben: Yeah we could do that. This is all code that we want to throw away (as the comment suggests, we haven't been super successful in doing that 😅) and is copy pasted from the old implementation. Should be possible to dedupe and break out a bit

plugins/catalog-backend/src/service/createRouter.ts
```diff
@@ -135,24 +137,139 @@ export async function createRouter(
...
+            res.setHeader('link', `<${url.pathname}${url.search}>; rel="next"`);
+          }
+
+          res.json(entities);
+          return;
         }
 
-        // TODO(freben): encode the pageInfo in the response
-        res.json(entities);
+        // For other read-the-entire-world cases, use queryEntities and stream
```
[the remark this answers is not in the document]
freben: Not sure 🤔 The regular http request metrics may suffice here, probably

plugins/catalog-backend/src/service/createRouter.ts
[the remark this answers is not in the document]
freben: Fixed!

plugins/catalog-backend/src/service/createRouter.ts
```diff
@@ -135,24 +138,94 @@ export async function createRouter(
...
+
+              signal.throwIfAborted();
+
+              expandLegacyCompoundRelationRefsInResponse(result.items);
+              if (!responseStream.send(result.items)) {
+                // The kernel buffer is full. Create the lock but do not await it
+                // yet - we can better spend our time going to the next round of
+                // the loop and read from the database while we wait for it to
+                // drain.
+                writeLock = createDeferred();
```
vinzscam, line 216: I wonder if doing `res.once('drain', () => { });` and somehow awaiting for the drain event could simplify things 🤔

plugins/catalog-backend/src/service/createRouter.ts
vinzscam: nice nice

plugins/catalog-backend/src/service/createRouter.ts
freben: Updated!

plugins/catalog-backend/src/service/createRouter.ts
```diff
@@ -135,24 +138,92 @@ export async function createRouter(
...
+              if (!responseStream.send(result.items)) {
+                // The kernel buffer is full. Create the lock but do not await it
+                // yet - we can better spend our time going to the next round of
+                // the loop and read from the database while we wait for it to
+                // drain.
+                locks.writeLock = createDeferred();
+                res.once('drain', () => {
+                  locks.writeLock?.resolve();
+                  delete locks.writeLock;
+                });
```
vinzscam, line 214: what about doing the following so that we won't need any other locks spread in the code?

```suggestion
                const drainLock = createDeferred();
                res.once('drain', () => drainLock.resolve());
                await drainLock;
```

of course we won't read the next iteration of entities, but the code will be much simpler.

freben: Yeah exactly

freben: we could do a whole dance where we both subscribe to end and drain, and unsubscribe from end, etc - but it's gonna be a bit of a bigger piece of code then and it'll aggressively subscribe and unsubscribe at a high rate, not sure if worth it

Issue #24590, opened 2024-05-02
🐛 Bug Report: Random HTTP failures with Node20


drodil:
### 📜 Description

Sometimes catalog (or other backend) calls fail due to `socket hang up` error. For example one of our processors tries to fetch entities from the catalog and it randomly fails as follows:

```
Catalog processing error occurred for component:default/my-component owned by group:default/developers: InputError: Processor MyProcessor threw an error while postprocessing; caused by FetchError: request to http://localhost:7007/api/catalog/entities?fields=kind,metadata.namespace,metadata.name&filter=kind%3DComponent failed, reason: socket hang up
```

Might have some relation to this or other issues linked to this: https://github.com/node-fetch/node-fetch/issues/1735 / https://github.com/nodejs/node/issues/47130

### 👍 Expected behavior

Sockets should not hang up

### 👎 Actual Behavior with Screenshots

Sockets hang up

### 👟 Reproduction steps

1. Use NodeJS 20
2. Add a processor that fetches catalog entities

### 📃 Provide the context for the Bug.

```ts
    const credentials = await this.auth.getOwnServiceCredentials();
    const { token } = await this.auth.getPluginRequestToken({
      onBehalfOf: credentials,
      targetPluginId: 'catalog',
    });

    const components = await this.client.getEntities(
      {
        filter: {
          kind: 'Component',
        },
        fields: ['kind', 'metadata.namespace', 'metadata.name'],
      },
      { token },
    );
```

### 🖥️ Your Environment

OS:   Darwin 23.4.0 - darwin/arm64
node: v20.12.0
yarn: 1.22.19
cli:  0.26.4 (installed)
backstage:  1.26.4

Dependencies:
  @backstage/app-defaults                                            1.5.4
  @backstage/backend-app-api                                         0.6.2, 0.7.1
  @backstage/backend-common                                          0.21.7
  @backstage/backend-defaults                                        0.2.17
  @backstage/backend-dev-utils                                       0.1.4
  @backstage/backend-openapi-utils                                   0.1.10
  @backstage/backend-plugin-api                                      0.6.17
  @backstage/backend-tasks                                           0.5.22
  @backstage/backend-test-utils                                      0.3.7
  @backstage/catalog-client                                          1.6.4
  @backstage/catalog-model                                           1.4.5
  @backstage/cli-common                                              0.1.13
  @backstage/cli-node                                                0.2.5
  @backstage/cli                                                     0.26.4
  @backstage/config-loader                                           1.8.0
[...]

PR #27925, opened 2024-11-29
recommend undici

freben:
Now that node 20 is the minimum, we should move to using the native (undici) fetch in the backend. This may also partially help with [some issues](https://github.com/backstage/backstage/issues/24590) with the use of other libraries.

### DIFF
--- a/docs/architecture-decisions/adr013-use-fetch.md
+++ b/docs/architecture-decisions/adr013-use-fetch.md
@@ -2,7 +2,7 @@
 id: adrs-adr013
 title: 'ADR013: Proper use of HTTP fetching libraries'
 # prettier-ignore
-description: Architecture Decision Record (ADR) for the proper use of fetchApiRef, node-fetch, and cross-fetch for data fetching.
+description: Architecture Decision Record (ADR) for the proper use of fetchApiRef, native fetch, and cross-fetch for data fetching.
 ---
 
 ## Context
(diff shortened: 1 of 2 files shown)


docs/architecture-decisions/adr013-use-fetch.md
```diff
@@ -12,13 +12,13 @@ support burden of keeping said package up to date.
 
 ## Decision
 
-Backend (node) packages should use the `node-fetch` package for HTTP data
-fetching. Example:
+Backend (node) packages should use the native `fetch` for HTTP data fetching,
+and importing `undici` where necessary. Example:
```
Rugvip, line 16: Why would one need to do this? thinking it's good to have an example or just not mention the option

freben: Maybe it's not worth mentioning in here honestly. It's for stuff like `setGlobalDispatcher` and `ProxyAgent` usage etc ... and maybe the only use case really is in the corporate proxy document. So ... remove?

Rugvip (approved):
👍 , :shipit:

Rugvip (commented):
Although there's a bit of a Q of whether we should ship this as a new ADR to replace the old one?

Parsifal-M:
> Although there's a bit of a Q of whether we should ship this as a new ADR to replace the old one?

I was wondering if you should keep a record of We used to do X and now we do Y becuase...

Instead of replacing it :)

Rugvip:
@Parsifal-M yep that's what I mean, replacing in the sense that the old one would be marked as invalid and point to the new one

freben:
And yes this would apply to plugins as well. They don't particularly have to hurry to do so, but I had planned actually posting this on discord and notifying the community-plugins maintainers because it's a potential migration that's of common interest. The only risk of deviating from this, is mainly #24590 and perhaps being surprised that our proxy instructions will stop being sufficient for you etc. But for some time to come, we'll probably try to be nice and not assume that node-fetch has gone away instantly in the ecosystem.

Rugvip:
> Is it valuable to do so?
> 
> Not saying it isn't, but after a while (much like migration dirs) I'm not sure it's going to be particularly easily readable. Is the information sufficiently available in git history instead?

Kinda just thinking about https://backstage.io/docs/architecture-decisions/#superseding-an-adr, but I can't say it's a very explicitly documented process anyway. The value of creating a new ADR is that we could record what changed and why we're now recommending native fetch, kinda captures the actual decision taken now a bit better I'd think.

freben:
Alright, updated to use a new one!

PR #27926, opened 2024-11-29
create-app: add lock seed for nwsapi

Rugvip:
🧹, bad release and fixes are on the way. In the meantime let's use the older version.


freben (approved):
No changeset because expecting it to be fixed before release anyway?

Rugvip:
No changeset because there's no change to the published package

Issue #23801, opened 2024-03-22
🚀 Feature: creating custom cli template


hubofgitongithub:
### 🔖 Feature description

Documentation and mechanism to easily plug into the templating mechanism of the backstage-cli. Or being able to override existing templates to create a new plugin.

### 🎤 Context

Our Backstage for instance has been migrated to MUI 5, new front-end plugins are still generated with old material-ui 4 dependencies. This is just one example of us wanting to be more in control of what is generated for people when they run a backstage-cli template, without having to fork the cli package.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

Rugvip:
Sorry, missed this one! This is something we want but not working on immediately in the maintainer team. @minkimcello has been looking at this though

awanlin:
Hi @minkimcello, any updates on this one? Came up recently on Discord again and I've found myself looking for something like this recently as well.

Rugvip:
To outline a bit more of our idea for the way forward here: Convert the `new` templates into standalone definitions that work like simplified scaffolder templates. We'd also make sure that they're portable so that you can for example link to templates on GitHub as well as in the local file system/repo. We then also add a configuration somewhere where you can list the templates that you want to use in your own repo, likely also with the ability to set default/fixed parameters and so on. Overall this would allow adopters to pick the specific templates that apply to their own project and setup, along with the ability to create custom templates for internal needs.

brendaHuang-2023:
Thanks @awanlin @Rugvip. This is a fairly common use case for us, since we have a steady stream of contributors who need to onboard to our backstage instance and create new plugins.

brendaHuang-2023:
@minkimcello Thanks a lot for helping out. We'd like to have the new plugin set up with MUI 5 and also we use a different eslint config. So it'd be nice to be able to override the default eslint config somehow.

Rugvip:
@minkimcello wanted to check in on your progress as we're looking to work on/ship some of this next week. Would love a draft PR to be able to start up early review/tweaks.

Issue #26548, opened 2024-09-09
🚀 Feature: Customize Backstage cli command that create plugins


brendaHuang-2023:
### 🔖 Feature description

We'd like to run `backstage-cli new --select plugin` and have the option to generate a plugin that uses MUI 5 and custom eslint config. 

### 🎤 Context

We have different eslint config than what the command currently generates, and we use MUI 5 instead of Material UI 4, which is the default of Backstage. Currently we have to manually update files after we generate the plugin. This is a common use case for us, since we have a steady stream of contributors who need to generate new plugins. 

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

Rugvip:
Yep, thanks! I re-opened the old one, closing this one as a duplicate

PR #27934, opened 2024-11-29
Declarative CLI Templates

minkimcello:

There's been a bit of interests for adopters to be able to create custom templates for the `backstage-cli new` command. I've replaced all of the existing template factories with a more flexible approach that allows you to define your own CLI templates with yaml files.

If you'd like to see how the new system works, I added a [new page](https://github.com/backstage/backstage/blob/05049c3fd9d1edc4dfe5b3d0957dd6797d7414ee/docs/tooling/cli/04-templates.md) in the docs on how to configure CLI options and how to create new templates.

One thing I [left out](https://github.com/backstage/backstage/blob/517d1d62666073008645f41350ee1f5340563fe8/packages/cli/src/lib/new/templateSelector.ts#L68-L70) from the scope of this PR is the ability to add remote templates. We'll do a follow-up PR for that feature.

---

- The Discord discussion for this feature can be found [here](https://ptb.discord.com/channels/687207715902193673/1222154056844181674).
- Previous PR that was closed but you can still find some of Patrick's early feedback - #26812
- Resolves issues #23801 and #26548

---


freben:
Just writing a comment to mention I was here and noted the test errors.

packages/cli/templates/scaffolder-module.yaml
```diff
@@ -0,0 +1,14 @@
+description: A module exporting custom actions for @backstage/plugin-scaffolder-backend
+template: ./scaffolder-module
+targetPath: plugins
+backendModulePrefix: true
```
mtlewis, line 4: I wonder if we could find a way to generalize this a bit - perhaps by adding support for variables interpolated from the template inputs? Might end up looking something like:

```yaml
variables:
  - suffix: scaffolder-backend-module-{{ moduleId }}
```

packages/cli/templates/scaffolder-module.yaml
```diff
@@ -0,0 +1,14 @@
...
+prompts:
+  - id: id
+    prompt: Enter the ID of the plugin [required]
+    default: scaffolder
+    validate: backstage-id
+  - moduleid
```
mtlewis, line 10: nit: I appreciate these are case-insensitive, but perhaps worth aiming to keep the casing right wherever we can to help with debugging issues stemming from places where the same string is used in a case-sensitive context (e.g. in the file immediately below).

docs/tooling/cli/04-templates.md
```diff
@@ -0,0 +1,139 @@
...
+### Additional Actions
+
+Additional actions are useful for running tasks. For example, if you want your newly created package to be automatically added as a dependency to `packages/backend/package.json` so that your users won't have to do so manually, you could add `install-backend` as one of the `additionalActions` to your template.
+
+There are four `additionalActions`:
+
+|        Action         | Description                                                            |
+| :-------------------: | :--------------------------------------------------------------------- |
+|  `install-frontend`   | Installs your new package/plugin as a dependency to `packages/app`     |
+| `add-frontend-legacy` | Adds your new frontend plugin to `packages/app/src/App.tsx`            |
```
mtlewis, line 137: Do these actions need to be separate? Perhaps a single `add-to-frontend` action (and the same for backend) can do the whole task of integrating the plugin into the local app and backend packages.

mtlewis (commented):
This is super exciting!

packages/cli/src/commands/new/new.ts
```diff
@@ -70,25 +66,72 @@ export default async (opts: OptionValues) => {
...
+    plugin: template.plugin ?? true,
+  });
+
+  const moduleVar =
+    options.moduleId ??
+    `${camelCase(options.id)}Module${camelCase(
+      options.moduleId,
+    )[0].toUpperCase()}${camelCase(options.moduleId).slice(1)}`; // used in default-backend-module template
+  const extensionName = `${upperFirst(camelCase(options.id))}Page`; // used in default-plugin template
+  const pluginVar = `${camelCase(options.id)}Plugin`; // used in default-backend-plugin and default-plugin template
```
Rugvip, line 84: What do you think of having some form of helper to help do this in the template instead? Something like `{{ prompts.pluginId + '-plugin' | camelCase }}`.

We have been talking about moving the scaffolder to https://liquidjs.com/, and could potentially start with these templates. Happy to leave that as a follow-up though

packages/cli/src/lib/new/additionalActions.ts
```diff
@@ -0,0 +1,184 @@
...
+            [options.name]: `^${options.version}`,
+          },
+        },
+      );
+    });
+  }
+}
+
+async function addBackend({ name }: AdditionalActionsOptions) {
+  if (await fs.pathExists(paths.resolveTargetRoot('packages/backend'))) {
```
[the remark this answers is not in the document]
Rugvip: Exactly, I think that could be handled by skipping the `add-backend` action though. I do expect that we'll add the ability to override template options as part of installing custom templates later on.

packages/cli/src/lib/new/prompts.ts
```diff
@@ -0,0 +1,152 @@
...
+  const answers = await inquirer.prompt(
+    prompts.map((prompt: ConfigurablePrompt) => {
+      if (typeof prompt === 'string') {
+        switch (prompt) {
+          case 'id':
+            return pluginIdPrompt();
+          case 'moduleId':
+            return moduleIdIdPrompt();
+          case 'npmRegistry':
+            return npmRegistryPrompt();
```
Rugvip, line 110: This should probably just be a configuration option right? defaulted to whatever is in `.yarn.yml` / `.npmrc` too

packages/cli/src/lib/new/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export function populateOptions(
+  prompts: Record<string, string>,
+  template: Template,
+): Options {
+  return {
+    id: prompts.id ?? '',
+    private: !!prompts.private,
+    baseVersion: prompts.baseVersion ?? '0.1.0',
+    owner: prompts.owner ?? '',
```
Rugvip, line 57: Thinking it's better to keep this as optional rather than default to `''`, or is it more convenient to give it a value?

Rugvip (commented):
Awesome stuff! 😁 🎉 

Bunch of comments around the design, in particular I think it would be best if we at least in a migration phase could keep support for supplying the existing command line options.

Also open to not fully settling and implementing all of these discussions, but to merge after the most important bits are settled some time after the main line release on Tuesday.

docs/tooling/cli/04-templates.md
```diff
@@ -0,0 +1,138 @@
...
+
+The behavior of the `backstage-cli new` command is configurable through your root `package.json`, and you can also create and add custom CLI templates to suit your needs.
+
+## Basic Configuration
+
+```json
+{
+  "name": "root",
+  "backstage": {
+    "cli": {
```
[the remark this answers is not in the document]
Rugvip: yep sounds good! 👍

docs/tooling/cli/04-templates.md
```diff
@@ -0,0 +1,138 @@
...
+      "defaults": true,
```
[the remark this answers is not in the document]
Rugvip: yep I like the idea of shipping the array in create-app, I think it's a good starting point and it'll be visible in release notes / upgrade helper

docs/tooling/cli/04-templates.md
```diff
@@ -0,0 +1,138 @@
...
+Your first step in creating your own CLI template is composing your yaml file:
+
+```yaml
+# ./custom-template.yaml
+description: Description of my CLI template
+template: ./template # required
+targetPath: packages # required
+
+suffix: hello
+backendModulePrefix: true
```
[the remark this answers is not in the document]
Rugvip: it's kinda tricky because the `package.json` isn't valid JSON before it's been templated, but to template it we need to values. We could template with placeholder values, but it seems a bit sketchy to go down that path, I think being explicit in the template declaration is better.

docs/tooling/cli/04-templates.md
```diff
@@ -0,0 +1,138 @@
...
+```diff
+{
+  // ...
+  "backstage": {
+    "cli": {
+      // ...
++       "templates": [
++         {
++           "id": "my custom CLI template",
++           "target": "./custom-template.yaml"
```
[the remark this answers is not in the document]
Rugvip: ah yes, good point, let's keep as is 👍

docs/tooling/cli/04-templates.md
```diff
@@ -0,0 +1,138 @@
...
+# ./custom-template.yaml
+description: Description of my CLI template
+template: ./template # required
+targetPath: packages # required
+
+suffix: hello
+backendModulePrefix: true
+
+prompts:
+  - id # required
```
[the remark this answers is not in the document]
Rugvip: Yep and yep! 👍

docs/tooling/cli/04-templates.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Additional Actions
+
+Additional actions are useful for running tasks. For example, if you want your newly created package to be automatically added as a dependency to `packages/backend/package.json` so that your users won't have to do so manually, you could add `install-backend` as one of the `additionalActions` to your template.
+
+There are four `additionalActions`:
+
+|        Action         | Description                                                            |
+| :-------------------: | :--------------------------------------------------------------------- |
+|  `install-frontend`   | Installs your new package/plugin as a dependency to `packages/app`     |
```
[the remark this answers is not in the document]
Rugvip: Could go with the verbose route:

```yaml
additionalActions:
  - action: install
      path: packages/app
```

Another option might be similar to what we do in the new frontend system where array items are either strings or single-key objects:

```yaml
additionalActions:
  - install:
      path: packages/app
```

packages/cli/cli-report.md
```diff
@@ -177,13 +177,6 @@ Options:
 Usage: backstage-cli new [options]
 
 Options:
-  --select <name>
-  --option <name>=<value>
-  --scope <scope>
-  --npm-registry <URL>
-  --baseVersion <version>
-  --license <license>
-  --no-private
```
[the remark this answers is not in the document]
Rugvip: Yep, just thinking that we should at minimum keep supporting the existing options, at least the most used ones

docs/features/software-catalog/external-integrations.md
```diff
@@ -72,9 +72,11 @@ putting all extensions like this in a backend module package of their own in the
 `plugins` folder of your Backstage repo:
 
 ```sh
-yarn new --select backend-module --option id=catalog
+yarn new
 ```
 
+Select `backend-module` and enter `catalog` for the id.
```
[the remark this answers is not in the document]
Rugvip: Yep, the hard-coded options I think should all be moved over to the global template config, and then we could keep something more generic to supply any kind of values. I think the existing `--option <name>=<value>` might handle that?

packages/cli/templates/default-backend-module/package.json.hbs
```diff
@@ -1,7 +1,7 @@
 {
   "name": "{{name}}",
-  "description": "The {{moduleId}} backend module for the {{pluginId}} plugin.",
-  "version": "{{packageVersion}}",
+  "description": "The {{moduleId}} backend module for the {{id}} plugin.",
+  "version": "{{pluginVersion}}",
```
Rugvip, line 4: I think it's best to keep `packageVersion`, feels more correct imo

packages/cli/templates/default-backend-module/README.md.hbs
```diff
@@ -1,5 +1,5 @@
 # {{name}}
 
-The {{moduleId}} backend module for the {{pluginId}} plugin.
+The {{moduleId}} backend module for the {{id}} plugin.
```
Rugvip, line 3: I think it's best to keep `pluginId` too, much more explicit and feels safer. It's especially what we stick to in the backend system with `pluginId` / `moduleId`. I think this is even more important in this case for modules where you have both the `moduleId` and `pluginId`.

Rugvip (commented):
Just a couple more in-flight comments. Please mark any threads as resolved that you think are handled @minkimcello

packages/cli/templates/default-backend-module/README.md.hbs
Rugvip: Yep there's certainly a lot of built-in cleanup work from the variation in the existing templates in this PR. I'm just aiming to help guide that 😁

docs/tooling/cli/04-templates.md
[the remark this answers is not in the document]
vinzscam: perhaps just `prefix`? 🤔

Rugvip:
@minkimcello No worries, but I'll pick this up and try to get it ready to shim an early version in the `next` release on Tuesday,. Really looking to get this completely done in time for the next mainline release which is just under two weeks out.

Parsifal-M (commented):
So looking forward to this 🤩  really cool!

I did a very quick read over some of the docs just trying to help move it a long 👍

docs/tooling/cli/03-commands.md
```diff
@@ -253,35 +253,33 @@ Restores the changes made by the prepack command
 ## new
 
 The `new` command opens up an interactive guide for you to create new things
-in your app. If you do not pass in any options it is completely interactive, but
+in your app. If you do not pass in any options it is completely interactive, but in your app.
 it is possible to pre-select what you want to create using the `--select` flag,
 and provide options using `--option`, for example:
```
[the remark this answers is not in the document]
Rugvip: 👍 , just the start of that line somehow being duplicated 😅

docs/tooling/cli/03-commands.md
```diff
@@ -253,35 +253,33 @@ Restores the changes made by the prepack command
 ## new
 
 The `new` command opens up an interactive guide for you to create new things
-in your app. If you do not pass in any options it is completely interactive, but
+in your app. If you do not pass in any options it is completely interactive, but in your app.
```
freben: this seems off? 🤔

docs/tooling/cli/04-templates.md
```diff
@@ -0,0 +1,149 @@
...
+    "cli": {
+      "new": {
+        "templates": ["./templates/custom-plugin"]
+      }
+    }
+  }
+}
+```
+
+Each entry in the `templates` array should be relative path that points to a directory containing a `portable-template.yaml` file. If the path starts with `./` it will be use as is, otherwise it will be resolved as a module within `node_modules`.
```
freben:
```suggestion
Each entry in the `templates` array should be relative path that points to a directory containing a `portable-template.yaml` file. If the path starts with `./` it will be used as is, otherwise it will be resolved as a module within `node_modules`.
```

docs/tooling/cli/04-templates.md
```diff
@@ -0,0 +1,149 @@
...
+
+Next, add any other files you want to be part of the template to the same directory. All files will be copied as is, except any files with a `.hbs` extension. They will be treated as [Handlebars](https://handlebarsjs.com/) templates and will be rendered with the values from the `portable-template.yaml` file as well as additional prompts such as `pluginId`. For example, you could create a `src/index.ts` file with the following content:
+
+```typescript title="in templates/custom-plugin/src/index.ts.hbs"
+export function getPluginId() {
+  return '{{ pluginId }}';
+}
+```
+
+If you'd like to see more examples, you can find all the default templates and its yaml files [here](https://github.com/backstage/backstage/tree/master/packages/cli/templates).
```
freben, line 130:
```suggestion
If you'd like to see more examples, you can find all the default templates and their yaml files [here](https://github.com/backstage/backstage/tree/master/packages/cli/templates).
```

package.json
```diff
@@ -1,11 +1,21 @@
 {
   "name": "root",
-  "version": "1.36.0-next.0",
+  "version": "1.36.0-next.2",
```
freben, line 3: really?

packages/cli/src/commands/new/new.ts
```diff
@@ -14,17 +14,66 @@
...
+    private: isPrivate,
+    ...otherGlobals
+  } = opts;
+
+  const prefilledParams = parseParams(rawArgOptions);
+
+  let pluginInfix: string | undefined = undefined;
+  let packagePrefix: string | undefined = undefined;
+  if (scope) {
+    const backstagePrefix = scope.startsWith('backstage') ? '' : 'backstage-';
```
freben, line 45: should there be an `@` here or no?

packages/cli/src/lib/new/preparation/loadPortableTemplateConfig.ts
```diff
@@ -0,0 +1,159 @@
...
+    license: overrides.license ?? config?.globals?.license ?? defaults.license,
+    version: overrides.version ?? config?.globals?.version ?? defaults.version,
+    private: overrides.private ?? config?.globals?.private ?? defaults.private,
+    publishRegistry:
+      overrides.publishRegistry ??
+      config?.globals?.publishRegistry ??
+      defaults.publishRegistry,
+    packageNamePrefix:
+      overrides.packageNamePrefix ??
+      config?.globals?.namePrefix ??
```
freben, line 119: any particular reason that the config globals are differently named? (here and below)

packages/cli/src/lib/new/preparation/loadPortableTemplateConfig.ts
```diff
@@ -0,0 +1,159 @@
...
+      defaults.packageNamePrefix,
+    packageNamePluginInfix:
+      overrides.packageNamePluginInfix ??
+      config?.globals?.namePluginInfix ??
+      defaults.packageNamePluginInfix,
+  };
+}
+
+function resolveLocalTemplatePath(pointer: string, basePath: string): string {
+  if (pointer.startsWith('.')) {
```
freben, line 129: ...or path.sep? or some form of isabsolute that also doesn't break on package names?

freben (approved):
Alright! Nothing blocking really

package.json
Rugvip: Not quite sure why GitHub is showing this, it's the current version O_o

packages/cli/src/commands/new/new.ts
Rugvip: Nop the old scope was defined without `@`

packages/cli/src/lib/new/preparation/loadPortableTemplateConfig.ts
Rugvip: I added the `package*` prefix to all internal representation for extra context. Had them wrapped in a separate object for a while too but felt it was a bit nicer to flatten things out in the end. On the other hand I think `packageNamePrefix` etc. is a bit too verbose for config

packages/cli/src/lib/new/preparation/loadPortableTemplateConfig.ts
Rugvip: I threw in a check to reject absolute paths for now

PR #27947, opened 2024-11-29
GitlabOrgDiscoveryEntityProvider restrictUsersToGroup should use the entire groupPath

cthtrifork:

GithubOrg.restrictUsersToGroup should use the entire groupPath as it calls `GitLabClient.listGroupMembers` in https://github.com/backstage/backstage/blob/master/plugins/catalog-backend-module-gitlab/src/lib/client.ts#L117

which uses:

```
return this.pagedRequest(
      `/groups/${encodeURIComponent(groupPath)}/members/all`,
      options,
    );
```

current behaviour:

```yaml
catalog:
  providers:
    gitlab:
      myProviderId:
        host: gitlab.com  # Identifies one of the hosts set up in the integrations
        fallbackBranch: main  # Optional. Fallback to be used if there is no default branch configured at the Gitlab repository. It is only used, if `branch` is undefined. Uses `master` as default
        group: my-org/kubernetes  # Optional. Group and subgroup (if needed) to look for repositories. If not present the whole instance will be scanned
        ...
       orgEnabled: true  # import users
        restrictUsersToGroup: true #
```

which would invoke
`https://gitlab.com/api/v4/groups/kubernetes/members/all`

and return 0 users as the org is missing...

instead of

`https://gitlab.com/api/v4/groups/my-org%2Fkubernetes/members/all`

which returns the expected users (I have tested it)

### DIFF
--- a/.changeset/blue-dragons-shout.md
+++ b/.changeset/blue-dragons-shout.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-gitlab': patch
+---
+
+restrictUsersToGroup should use the entire groupPath
--- a/.changeset/blue-dragons-shout.md
+++ b/.changeset/blue-dragons-shout.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-backend-module-gitlab': patch
 ---
 
-restrictUsersToGroup should use the entire groupPath
+`restrictUsersToGroup` should use the entire group path when getting members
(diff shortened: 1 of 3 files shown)


cthtrifork:
I think this is a patch change, as the code was only being used for getting members referenced in the gitlab client.
It is also only executed when `restrictUsersToGroup: true` and I believe this has never worked as intended since `2.0`

cthtrifork:
@benjdlambert Hi, any way to prioritise this? The following bug prevents our organization from logging into backstage, unless their user have been manually created in a catalog-info.yaml file.

benjdlambert (approved):

PR #27948, opened 2024-11-30
Update `fetch` ADR regarding `msw`

freben:
Turns out that `msw` 1.x (which has been deprecated for a long time but is in extensive use) is not compatible with native fetch.


benjdlambert:
Might be worth trying https://codemod.com/registry/msw-2-upgrade-recipe and see how much that can automate for us in all of our repos to get us there

benjdlambert (approved):

Rugvip (approved):
👍

PR #27949, opened 2024-11-30
add more data to deployment.json for k8s

sebalaini:

### DIFF
--- a/.changeset/brown-cobras-battle.md
+++ b/.changeset/brown-cobras-battle.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-kubernetes': patch
+---
+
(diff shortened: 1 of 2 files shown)


mclarke47:
Can you update the test also

sebalaini:
> Can you update the test also

Actually, it looks like it's used only in the KubernetesContent.test, and in there I can't test this data, I should add this data to the other json file `plugins/kubernetes-react/src/__fixtures__/1-deployments.json` but again, is not used for the test of the component where I could see this data, is used for the Cluster.test where I can't test this data either.

mclarke47:
If it's not used in the test and doesn't change the test then many we don't need to update it by itself?

mclarke47 (approved):

PR #27950, opened 2024-11-30
remove most usages of node-fetch

freben:
https://backstage.io/docs/next/architecture-decisions/adrs-adr014/

There are a few left still, that had trickier tests.


Rugvip (approved):
Nice! :tada:

PR #27957, opened 2024-12-02
fix nulls in by-query

freben:
The root query accidentally included entities where the `final_entity` column was null.


vinzscam (approved):
🙏

benjdlambert (approved):

PR #27960, opened 2024-12-02
Store the refresh state timing in a separate unlogged table

freben:
As part of the ongoing catalog optimisation efforts, let's try to move out the `next_update_at` timestamp column from `refresh_state` into its own table. Also, make that table `UNLOGGED` on postgres, making it not contribute to disk I/O at all.

This experiment looks to reduce WAL pressure and lock contention, leaving the `refresh_state` table completely unchanged unless entity data actually does change. We have debated whether to make it possible to externalize this queue into a separate system such as redis, but let's start with this as a baseline and see if we feel that that extra complexity is worth it after things settle.

For now, the deferred stitching related columns stay in `refresh_state` for two reasons: losing state in them has actual impact (risks data loss), and they only change whenever the processed output and stitched output need to change anyway.


benjdlambert:
Probably wanna update the otel metrics and alike right now?

PR #27961, opened 2024-12-02
Stop accepting traffic before shutdown

camilaibs:

Immediately cease accepting traffic when a termination event occurs, rather than waiting until plugins and services are stopped.

### DIFF
--- a/.changeset/green-carpets-smell.md
+++ b/.changeset/green-carpets-smell.md
@@ -3,3 +3,5 @@
 ---
 
 The `RootLifecycleService` now has a new `addBeforeShutdownHook` method, and hooks added through this method will run immediately when a termination event is received.
+
+The backend will not proceed with the shutdown and run the `Shutdown` hooks until all `BeforeShutdown` hooks have completed.
--- a/.changeset/green-carpets-smell.md
+++ b/.changeset/green-carpets-smell.md
@@ -1,0 +1,5 @@
+---
+'@backstage/backend-plugin-api': patch
+---
+
+The `RootLifecycleService` now has a new `addBeforeShutdownHook` method, and hooks added through this method will run immediately when a termination event is received.
--- a/docs/backend-system/core-services/root-http-router.md
+++ b/docs/backend-system/core-services/root-http-router.md
@@ -50,8 +50,16 @@ The `app-config.yaml` file provides configurable options that can be adjusted to
 backend:
   lifecycle:
     # (Optional) The maximum time that paused requests will wait for the service to start, before returning an error (defaults to 5 seconds).
+    # Supported formats:
+    # - A string in the format of '1d', '2 seconds' etc. as supported by the `ms` library.
+    # - A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
+    # - An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.
     startupRequestPauseTimeout: { seconds: 10 }
     # (Optional) The maximum time that the server will wait for stop accepting traffic, before returning an error (defaults to 30 seconds).
+    # Supported formats:
+    # - A string in the format of '1d', '2 seconds' etc. as supported by the `ms` library.
+    # - A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
+    # - An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.
     shutdownRequestPauseTimeout: { seconds: 20 }
 ```
 
--- a/docs/backend-system/core-services/root-http-router.md
+++ b/docs/backend-system/core-services/root-http-router.md
@@ -55,12 +55,12 @@ backend:
     # - A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
     # - An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.
     startupRequestPauseTimeout: { seconds: 10 }
-    # (Optional) The maximum time that the server will wait for stop accepting traffic, before returning an error (defaults to 30 seconds).
+    # (Optional) The minimum time that the HTTP server will delay the shutdown of the backend. During this delay health checks will be set to failing, allowing traffic to drain (defaults to 0 seconds).
     # Supported formats:
     # - A string in the format of '1d', '2 seconds' etc. as supported by the `ms` library.
     # - A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
     # - An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.
-    shutdownRequestPauseTimeout: { seconds: 20 }
+    serverShutdownTimeout: { seconds: 20 }
 ```
 
 ### Via Code
--- a/docs/backend-system/core-services/root-http-router.md
+++ b/docs/backend-system/core-services/root-http-router.md
@@ -42,6 +42,21 @@ createBackendPlugin({
 
 ## Configuring the service
 
+### Via `app-config.yaml`
+
+The `app-config.yaml` file provides configurable options that can be adjusted to meet your `RootHttpRouterService` specific requirements:
+
+```yaml
+backend:
+  lifecycle:
+    # (Optional) The maximum time that paused requests will wait for the service to start, before returning an error (defaults to 5 seconds).
+    startupRequestPauseTimeout: { seconds: 10 }
+    # (Optional) The maximum time that the server will wait for stop accepting traffic, before returning an error (defaults to 30 seconds).
+    shutdownRequestPauseTimeout: { seconds: 20 }
+```
+
+### Via Code
+
 There's additional options that you can pass to configure the root HTTP Router service. These options are passed when you call `createBackend`.
 
 - `indexPath` - optional path to forward all unmatched requests to. Defaults to `/api/app` which is the `app-backend` plugin responsible for serving the frontend application through the backend.
--- a/packages/backend-app-api/src/wiring/BackendInitializer.ts
+++ b/packages/backend-app-api/src/wiring/BackendInitializer.ts
@@ -452,7 +452,7 @@ export class BackendInitializer {
     const rootLifecycleService = await this.#getRootLifecycleImpl();
 
     // Root services like the health one need to immediatelly be notified of the shutdown
-    await rootLifecycleService.preShutdown();
+    await rootLifecycleService.beforeShutdown();
 
     // Get all plugins.
     const allPlugins = new Set<string>();
@@ -480,7 +480,7 @@ export class BackendInitializer {
   async #getRootLifecycleImpl(): Promise<
     RootLifecycleService & {
       startup(): Promise<void>;
-      preShutdown(): Promise<void>;
+      beforeShutdown(): Promise<void>;
       shutdown(): Promise<void>;
     }
   > {
--- a/packages/backend-app-api/src/wiring/BackendInitializer.ts
+++ b/packages/backend-app-api/src/wiring/BackendInitializer.ts
@@ -65,9 +65,18 @@ const instanceRegistry = new (class InstanceRegistry {
     if (!this.#registered) {
       this.#registered = true;
 
-      process.addListener('SIGTERM', this.#exitHandler);
-      process.addListener('SIGINT', this.#exitHandler);
-      process.addListener('beforeExit', this.#exitHandler);
+      process.addListener('SIGTERM', () => {
+        this.#exitHandler();
+      });
+      process.addListener('SIGINT', () => {
+        this.#exitHandler();
+      });
+      process.addListener('SIGQUIT', () => {
+        process.exit(0);
+      });
+      process.addListener('beforeExit', () => {
+        this.#exitHandler();
+      });
     }
 
     this.#instances.add(instance);
@@ -79,12 +88,18 @@ const instanceRegistry = new (class InstanceRegistry {
 
   #exitHandler = async () => {
     try {
-      const results = await Promise.allSettled(
-        Array.from(this.#instances).map(b => b.stop()),
-      );
-      const errors = results.flatMap(r =>
-        r.status === 'rejected' ? [r.reason] : [],
-      );
+      // This signals to the healthcheck service that the process is shutting down
+      process.exitCode = 0;
+
+      const results = await Promise.race([
+        // Give the backend 30 seconds to shut down, then force exit
+        new Promise(resolve => setTimeout(resolve, 30000)),
+        Promise.allSettled(Array.from(this.#instances).map(b => b.stop())),
+      ]);
+
+      const errors = Array.isArray(results)
+        ? results.flatMap(r => (r.status === 'rejected' ? [r.reason] : []))
+        : [];
 
       if (errors.length > 0) {
         for (const error of errors) {
--- a/packages/backend-app-api/src/wiring/BackendInitializer.ts
+++ b/packages/backend-app-api/src/wiring/BackendInitializer.ts
@@ -65,18 +65,9 @@ const instanceRegistry = new (class InstanceRegistry {
     if (!this.#registered) {
       this.#registered = true;
 
-      process.addListener('SIGTERM', () => {
-        this.#exitHandler();
-      });
-      process.addListener('SIGINT', () => {
-        this.#exitHandler();
-      });
-      process.addListener('SIGQUIT', () => {
-        process.exit(0);
-      });
-      process.addListener('beforeExit', () => {
-        this.#exitHandler();
-      });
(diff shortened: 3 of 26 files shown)


packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -74,12 +83,18 @@ const instanceRegistry = new (class InstanceRegistry {
...
   #exitHandler = async () => {
     try {
-      const results = await Promise.allSettled(
-        Array.from(this.#instances).map(b => b.stop()),
-      );
-      const errors = results.flatMap(r =>
-        r.status === 'rejected' ? [r.reason] : [],
-      );
+      // This signals to the healthcheck service that the process is shutting down
+      process.exitCode = 0;
```
Rugvip, line 87: Let's avoid communication via any form of globals if possible

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -74,12 +83,18 @@ const instanceRegistry = new (class InstanceRegistry {
...
+
+      const results = await Promise.race([
+        // Give the backend 30 seconds to shut down, then force exit
+        new Promise(resolve => setTimeout(resolve, 30000)),
+        Promise.allSettled(Array.from(this.#instances).map(b => b.stop())),
+      ]);
```
Rugvip, line 93: I don't think this is quite the place to handle this kind of force exit logic tbh, better have that live within the backend instances instead imo

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -74,12 +83,18 @@ const instanceRegistry = new (class InstanceRegistry {
...
-      const errors = results.flatMap(r =>
-        r.status === 'rejected' ? [r.reason] : [],
-      );
+      // This signals to the healthcheck service that the process is shutting down
+      process.exitCode = 0;
+
+      const results = await Promise.race([
+        // Give the backend 30 seconds to shut down, then force exit
+        new Promise(resolve => setTimeout(resolve, 30000)),
+        Promise.allSettled(Array.from(this.#instances).map(b => b.stop())),
```
Rugvip, line 92: Tbh, I'm kinda wondering if all we need to do is add a wait to the `stop()` after we have called the shutdown hooks. Basically we'd make it a `backend.postShutdownWait` configuration or something like that, just a silly little wait for a certain amount of time after the shutdown hooks have run successfully. The goal isn't really to drain traffic but to make sure that we won't receive any new one, so I think that'd be alright.

packages/backend-plugin-api/src/services/definitions/RootLifecycleService.ts
```diff
@@ -23,4 +23,6 @@ import { LifecycleService } from './LifecycleService';
  *
  * @public
  */
-export interface RootLifecycleService extends LifecycleService {}
+export interface RootLifecycleService extends LifecycleService {
+  addPreShutdownHook(hook: () => void): void;
```
Rugvip, line 27: Naming wise just wanna float some options, these are all for the pattern `add...Hook`:

```
PreShutdown
PrepareShutdown
Drain
TrafficDrain
ConnectionDrain
Close
BeforeShutdown
```

Rugvip: Based on discussion on Discord we're moving forward with `addBeforeShutdownHook`

packages/backend-defaults/src/entrypoints/rootHealth/rootHealthServiceFactory.test.ts
```diff
@@ -51,28 +51,22 @@ describe('DefaultRootHealthService', () => {
...
-
+    it(`should return a 503 response if the server is stopping`, async () => {
       const service = new DefaultRootHealthService({
-        lifecycle,
+        lifecycle: mockServices.rootLifecycle.mock(),
       });
 
-      mockServerStartedFn();
-      mockServerStoppedFn();
+      process.exitCode = 1;
```
Rugvip, line 59: This was removed right?

packages/backend-app-api/src/wiring/BackendInitializer.ts
camilaibs: We decided to implement a new before shutdown lifecycle hook, it will be called before plugin and services are stopped.

packages/backend-plugin-api/src/services/definitions/RootLifecycleService.ts
camilaibs: Name updated 🙌🏻

packages/backend-defaults/src/entrypoints/rootHealth/rootHealthServiceFactory.test.ts
camilaibs: Oops, removed 🙏🏻

packages/backend-app-api/src/wiring/BackendInitializer.ts
camilaibs: We realized that it is already handled by the plugins lifecycle middleware:
https://github.com/backstage/backstage/blob/2b6efe53674316b90f0f11a04e417bad78b64631/packages/backend-defaults/src/entrypoints/httpRouter/http/createLifecycleMiddleware.ts#L97-L101

.changeset/green-carpets-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-plugin-api': patch
+---
+
+The `RootLifecycleService` now has a new `addBeforeShutdownHook` method, and hooks added through this method will run immediately when a termination event is received.
```
Rugvip, line 5:
```suggestion
The `RootLifecycleService` now has a new `addBeforeShutdownHook` method, and hooks added through this method will run immediately when a termination event is received.

The backend will not proceed with the shutdown and run the `Shutdown` hooks until all `BeforeShutdown` hooks have completed.
```

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
```diff
@@ -120,7 +120,7 @@ const rootHttpRouterServiceFactoryWithOptions = (
         },
       });
 
-      lifecycle.addShutdownHook(() => server.stop());
+      lifecycle.addBeforeShutdownHook(() => server.stop());
```
Rugvip, line 123: The `server.stop` needs to be part of the shutdown hook, we still want to receive traffic during the before shutdown phase. We'll want to add a separate `BeforeShutdown` hook that ensures that traffic is drained , along with the optional wait period that can be added through configuration. The traffic draining logic can be moved over from the plugin scoped HTTP router service.

Rugvip (commented):
Nice! 👍 

Looks good, just needs the graceful router shutdown handling. I think it's best to keep that part of this PR.

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
camilaibs: Something like this https://github.com/backstage/backstage/pull/27961/commits/cbb499d102fa4942b989fcdf6b48bfcd262ff40c?

packages/backend-defaults/src/entrypoints/rootHttpRouter/createLifecycleMiddleware.ts
```diff
@@ -0,0 +1,108 @@
...
+  });
+
+  lifecycle.addBeforeShutdownHook(async () => {
+    state = 'down';
+
+    for (const item of waiting) {
+      clearTimeout(item.timeout);
+      item.next(new ServiceUnavailableError('Service is shutting down'));
+    }
+    waiting.clear();
```
Rugvip, line 81: Best move this to the final shutdown phase too

packages/backend-defaults/src/entrypoints/rootHttpRouter/createLifecycleMiddleware.ts
```diff
@@ -0,0 +1,108 @@
...
+ * @public
+ */
+export interface LifecycleMiddlewareOptions {
+  lifecycle: RootLifecycleService;
+  /**
+   * The maximum time that paused requests will wait for the service to start, before returning an error.
+   *
+   * Defaults to 5 seconds.
+   */
+  requestPauseTimeout?: HumanDuration;
```
Rugvip, line 35: Ah, we didn't discuss the need for this one so far but given that we're adding config here I think it makes sense to expose this too. Iirc there's been some situations where it would be useful for people to be able to bump this a bit.

Perhaps we can make it even more explicit in the naming, thinking something like `startupRequestPauseTimeout`?

Don't forget to document these in `docs/backend-system/` as well! 🙏

docs/backend-system/core-services/root-http-router.md
```diff
@@ -42,6 +42,21 @@ createBackendPlugin({
...
+
+The `app-config.yaml` file provides configurable options that can be adjusted to meet your `RootHttpRouterService` specific requirements:
+
+```yaml
+backend:
+  lifecycle:
+    # (Optional) The maximum time that paused requests will wait for the service to start, before returning an error (defaults to 5 seconds).
+    startupRequestPauseTimeout: { seconds: 10 }
+    # (Optional) The maximum time that the server will wait for stop accepting traffic, before returning an error (defaults to 30 seconds).
+    shutdownRequestPauseTimeout: { seconds: 20 }
```
Rugvip, line 55: The naming here is a bit too symmetrical imo, because these two don't function in the same way.

I'm thinking simply `shutdownRequestPauseTimeout` -> `shutdownDelay` or something like that? It's not really a timeout.

Regarding both of these I'm a bit torn on `backend.lifecycle.*`, some other options would be `backend.*` or `backend.http.*`

Rugvip: So as discussed let's stick to `backend.lifecycle.*`

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.test.ts
```diff
@@ -198,4 +203,71 @@ describe('rootHttpRouterServiceFactory', () => {
...
+
+    const beforeShutdownPromise = (lifecycle as any)
+      .beforeShutdown()
+      .then(() => {
+        return (lifecycle as any).shutdown();
+      });
+
+    await request(app!).get('/test').expect(200, {
+      status: 'ok',
+    });
```
Rugvip, line 265: As Ben mentioned, let's also add assertions that the liveness and readiness probes return expected results as well

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
```diff
@@ -30,6 +32,26 @@ import {
...
+    return undefined;
+  }
+  if (typeof value === 'number') {
+    return { milliseconds: value };
+  }
+  if (typeof value === 'string') {
+    return {
+      milliseconds: parseInt(value, 10),
+    };
+  }
```
Rugvip, line 52: I don't think we want to support these, should be enough with just the built-in functionality from `readDurationFromConfig`

camilaibs: Oh sorry, I didn't realize that `readDurationFromConfig ` supports string, and in a better way than I implemented it 🤦🏻‍♀️ . Just fixed it.

docs/backend-system/core-services/root-http-router.md
Rugvip: @camilaibs this still needs a fix ^

camilaibs: Fixed 🙏🏻

packages/backend-defaults/config.d.ts
```diff
@@ -28,6 +28,29 @@ export interface Config {
...
+       * Defaults to 5 seconds
+       * Supported formats:
+       * - A string in the format of '1d', '2 seconds' etc. as supported by the `ms`
+       *   library.
+       * - A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
+       * - An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.
+       */
+      startupRequestPauseTimeout?: string | HumanDuration;
+      /**
+       * The maximum time that the server will wait for stop accepting traffic, before returning an error.
```
Rugvip, line 43:
```suggestion
       * The minimum time that the HTTP server will delay the shutdown of the backend. During this delay health checks will be set to failing, allowing traffic to drain.
```

docs/backend-system/core-services/root-http-router.md
```diff
@@ -42,6 +42,29 @@ createBackendPlugin({
...
+    # - A string in the format of '1d', '2 seconds' etc. as supported by the `ms` library.
+    # - A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
+    # - An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.
+    startupRequestPauseTimeout: { seconds: 10 }
+    # (Optional) The maximum time that the server will wait for stop accepting traffic, before returning an error (defaults to 30 seconds).
+    # Supported formats:
+    # - A string in the format of '1d', '2 seconds' etc. as supported by the `ms` library.
+    # - A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
+    # - An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.
+    shutdownRequestPauseTimeout: { seconds: 20 }
```
Rugvip, line 63: This needs an update

packages/backend-defaults/src/entrypoints/rootHttpRouter/createLifecycleMiddleware.ts
```diff
@@ -0,0 +1,124 @@
...
+ * limitations under the License.
+ */
+
+import { RootLifecycleService } from '@backstage/backend-plugin-api';
+import { ServiceUnavailableError } from '@backstage/errors';
+import { HumanDuration, durationToMilliseconds } from '@backstage/types';
+import { RequestHandler } from 'express';
+
+export const DEFAULT_STARTUP_REQUEST_PAUSE_TIMEOUT = { seconds: 5 };
+export const DEFAULT_SERVER_SHUTDOWN_TIMEOUT = { seconds: 30 };
```
Rugvip, line 23: The default for this one should be 0, most often I don't think it's needed. We definitely don't want it to be above 0 for local development by default

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
```diff
@@ -114,6 +140,7 @@ const rootHttpRouterServiceFactoryWithOptions = (
           app.use(middleware.compression());
           app.use(middleware.logging());
           app.use(healthRouter);
+          app.use(lifecycleMiddleware);
```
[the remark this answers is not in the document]
camilaibs: Question, should it be added to the middleware factory 🤔 ?

packages/backend-defaults/src/entrypoints/rootHttpRouter/createLifecycleMiddleware.ts
```diff
@@ -0,0 +1,124 @@
...
+export interface LifecycleMiddlewareOptions {
+  lifecycle: RootLifecycleService;
+  /**
+   * The maximum time that paused requests will wait for the service to start, before returning an error.
+   *
+   * Defaults to 5 seconds.
+   */
+  startupRequestPauseTimeout?: HumanDuration;
+  /**
+   * The maximum time that the server will wait for stop accepting traffic, before returning an error.
```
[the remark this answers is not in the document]
camilaibs: Applied @Rugvip suggestion 🙏🏻

packages/backend-defaults/src/entrypoints/rootHealth/rootHealthServiceFactory.ts
```diff
@@ -29,7 +29,7 @@ export class DefaultRootHealthService implements RootHealthService {
     options.lifecycle.addStartupHook(() => {
       this.#isRunning = true;
     });
-    options.lifecycle.addShutdownHook(() => {
+    options.lifecycle.addBeforeShutdownHook(() => {
```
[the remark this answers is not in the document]
camilaibs: Great idea, just updated the implementation 🙌🏻

Rugvip (approved):
Let's :shipit: 🎉 

I think middleware factory would be best, but also fine and best to handle in a followup

PR #27963, opened 2024-12-02
BEP: Template Execution Permissions

CptnFizzbin:
# Summary

Companies are able to create custom actions and fields for their templates, and will want to restrict who can execute a
template action to specific groups of users.

With the existing permissions `stepReadPermission` and `parameterReadPermission`, the template sections can be
restricted, but still allows the template to run. This can result in "successful" templates that silently fail. For
example, a template could be written to scaffold out a Backstage plugin, but the user is not allowed to read the
`publish:github` action. The template would succeed, but the push to github would be skipped.

Using `catalogEntityReadPermission` can restrict execution, however this would also hide the entity from the catalog for
the user. The entity page for templates can be used by companies to provide additional information about the template,
and if a user doesn't have permission to run the template, information on why and how to get permission can be displayed
on that page.

Adding a `templateExecutePermission` will allow policy writers to explicitly restrict execution of a template based on
built in rules, and any custom rules created by the writer. An additional entity `spec` section for permissions would
provide a dedicated section to provide data to built in or custom permission rules.

### DIFF
--- a/beps/0012-template-execution-permissions/README.md
+++ b/beps/0012-template-execution-permissions/README.md
@@ -1,0 +1,262 @@
+---
+title: Template Execution Permissions
+status: implementable
+authors:
+  - '@cptnfizzbin'
+owners:
+  - '@cptnfizzbin'
+project-areas:
+  - scaffolder
+  - permissions
+creation-date: 2024-12-02
+---
+
+<!--
+**Note:** When your BEP is complete, all these pre-existing comments should be removed
+-->
+
+# BEP: Template Execution Permissions
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Design Details](#design-details)
+- [Release Plan](#release-plan)
+- [Dependencies](#dependencies)
+- [Alternatives](#alternatives)
+
+## Summary
+
+<!--
+The summary of the BEP is a few paragraphs long and give a high-level overview of the features to be implemented. It should be possible to read *only* the summary and understand what the BEP is proposing to accomplish and what impact it has for users.
+-->
+
+Companies are able to create custom actions and fields for their templates, and will want to restrict who can execute a
+template action to specific groups of users.
+
+With the existing permissions `stepReadPermission` and `parameterReadPermission`, the template sections can be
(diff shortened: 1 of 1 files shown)


CptnFizzbin:
Initial implementation: #27748

vinzscam:
mmm I'm not sure the `templateExecutePermission` gives a better control over the existing `actionExecutePermission`.
Let's say you restrict the execution of templates via `templateExecutePermission`.
A user could create a new template stripping out the information used for permissioning the template which will allow them to circumvent `templateExecutePermission`.
On the other side, `actionExecutePermission` is more secure, as it won't allow the execution of certain actions if some conditions aren't met in any template.
So in order to secure your template, an adopter must end up replicating the same rules for `templateExecutePermission` to `actionExecutePermission`. So my question is: do we really need a new `templateExecutePermission`? If our goal is to upfront disallow the execution of a template if some actions aren't allowed, the scaffolder could expose an endpoint for authorizing the template upfront, authorizing the `actionExecutePermission` for all actions, so that you only need to target `actionExecutePermission`.

CptnFizzbin:
There isn't an `actionExecutePermission` so I assume you meant the `stepReadPermission` instead. 

Updating the permissions check to scan all actions and fields using the `stepRead` and `parameterRead` permissions is an option, however for the use case of when a template author wants to limit the whole template to an specific user group, an `templateExecute` resource permission would still be needed to allow rules to read a top level permissions section.

Ideally, `stepRead` and `parameterRead` would both be deprecated in favour of a unified `templateExecutePermission` since that resource would include access to check all steps and fields. (it cleans up the terminology as well since actions are also known as steps, and fields also known as parameters)

vinzscam:
> There isn't an `actionExecutePermission` so I assume you meant the `stepReadPermission` instead.

This is the permission I was referring to: 
https://github.com/backstage/backstage/blob/master/plugins/scaffolder-common/src/permissions.ts#L39

https://backstage.io/docs/features/software-templates/authorizing-scaffolder-template-details#authorizing-actions

CptnFizzbin:
huh. I missed that while going through the code, although it has the same issues (and similar functionality) as `stepReadPermission` since it's evaluated on a per action level and doesn't really allow for restricting/allowing a template on a whole.

Though it does highlight that if a `templateExecutePermission` is added, the permission would need to be re-evaluated after value substitution for each step

PR #27970, opened 2024-12-03
chore: add synergy plugin to microsite

jiteshy:

This PR attempts to add a new plugin - Synergy to the microsite. It includes the associated YAML file with all relevant details and the static asset for the logo.


freben (approved):
Hi!

Cool plugin! Happy to add.

Some future ideas for the plugin while looking through the code:

* Could add some installation instructions in the README files of the packages; just some small "how to backend.add this thing" type blurb for example for the backend one
* You say that you only support the new backend system (which is perfectly fine!) but you still export the raw router too - feel free to remove that export from your index to not have users be tempted to "peek" at what's effectively closed interfaces
* The backend reads config but doesn't have a corresponding `config.d.ts` and then referencing that in its `package.json` - [see here](https://backstage.io/docs/conf/defining) - you _especially_ want to mark tokens with secret visibility so they don't leak
* You seem to only grab the first provider key from config, what about the rest? :)
* The backend package's dependencies on the auth backend and guest module can be dev dependencies (it's only used in the dev entrypoint)
* Your backend sets an auth policy of unauthenticated on its entire API surface, is that _really_ what you want to do? It opens it up for any user to call it willy nilly with an unauthed curl

jiteshy:
> Hi!
> 
> Cool plugin! Happy to add.
> 
> Some future ideas for the plugin while looking through the code:
> 
> * Could add some installation instructions in the README files of the packages; just some small "how to backend.add this thing" type blurb for example for the backend one
> * You say that you only support the new backend system (which is perfectly fine!) but you still export the raw router too - feel free to remove that export from your index to not have users be tempted to "peek" at what's effectively closed interfaces
> * The backend reads config but doesn't have a corresponding `config.d.ts` and then referencing that in its `package.json` - [see here](https://backstage.io/docs/conf/defining) - you _especially_ want to mark tokens with secret visibility so they don't leak
> * You seem to only grab the first provider key from config, what about the rest? :)
> * The backend package's dependencies on the auth backend and guest module can be dev dependencies (it's only used in the dev entrypoint)
> * Your backend sets an auth policy of unauthenticated on its entire API surface, is that _really_ what you want to do? It opens it up for any user to call it willy nilly with an unauthed curl

Thanks @freben for the thorough feedback 👍  I will look through these and fix wherever applicable.

PR #27984, opened 2024-12-04
Add OpenFGA permission plugin backstage to market place

AmbrishRamachandiran:
Add OpenFGA permission plugin backstage to market place:

**This plugin wraps around the Backstage Permission Framework and uses the OPENFGA client to evaluate policies. It will send a request to OPENFGA with the permission and identity information, OPENFGA will then evaluate the policy and return a decision, which is then passed back to the Permission Framework.**

image

Here the complete documentation about the plugin : https://github.com/Infosys/openfga-plugin-backstage/blob/main/plugins/openfga/README.md

A webinar submitted to cncf on this: **https://youtu.be/wWFbLPvwOyQ?si=_odKh_afogm8Ta5K**

### DIFF
--- a/microsite/data/plugins/openfga-permission.yaml
+++ b/microsite/data/plugins/openfga-permission.yaml
@@ -1,0 +1,10 @@
+---
+title: OpenFGA Permission
+author: Infosys
+authorUrl: https://github.com/Infosys
+category: Authentication/Authorization
+description: Manage your Backstage permissions with OpenFGA
+documentation: https://github.com/Infosys/openfga-plugin-backstage/blob/main/plugins/openfga/README.md
+iconUrl: https://github.com/Infosys/openfga-plugin-backstage/blob/main/plugins/openfga/src/docs/openfgalogo.png
+npmPackageName: '@infosys_ltd/openfga-plugin-backstage'
+addedDate: '2024-12-04'


AmbrishRamachandiran:
Please take a look ! commenting for avoid stale

AmbrishRamachandiran:
> I'm a bit confused by the code. Some feedback on the plugin itself that I recommend fixing:
> 
> The installation instructions tell users to add `import { permissionModuleACatalogPolicy } from '../../../plugins/openfga/src/module';`, that doesn't seem right; nobody will copy this actual code
> 
> `filter: { 'kind=component': CATALOG_FILTER_EXISTS }` should be `filter: { kind: 'component' }`, but is it really just for components?
> 
> If you do `fields: ['metadata.name']` and handle plain names, you will never work with catalogs that use other namespaces than `default`. You should _at least_ do `fields: ['kind', 'metadata.namespace', 'metadata.name']` and `stringifyEntityRef` them to handle full refs - which is the canonical universal identifier for entities.
> 
> Your client.ts doesn't inject `fetchApi`; [it's best practice](https://backstage.io/docs/architecture-decisions/adrs-adr013) to always use the fetch api to ensure that people who customize it still can get their metrics and behaviors etc as they expect.
> 
> Your policy is named `a-name-policy` which is confusing to users
> 
> There's a config section that has comments saying that you should modify those settings by hand in code, which of course isn't an option for an adopter who just imports your module
> 
> The actual policy hard codes an entity name etc, I am not sure what's up with that, does this code really do what the end user thinks it does?



**Query :** 

The installation instructions tell users to add import { permissionModuleACatalogPolicy } from '../../../plugins/openfga/src/module';, that doesn't seem right; nobody will copy this actual code

**Answer:** 

To enable the policy that will be used can be found in **plugins/openfga/src/module** thats why we added it in the instruction like to use this once they added openfga plugin they need to add the module in the backend file.  

**similar reference :  The OPA Plugin**

image


**Query:**

`filter: { 'kind=component': CATALOG_FILTER_EXISTS }` should be `filter: { kind: 'component' }`, but is it really just for components?
 If you do `fields: ['metadata.name']` and handle plain names, you will never work with catalogs that use other namespaces than `default`. You should _at least_ do `fields: ['kind', 'metadata.namespace', 'metadata.name']` and `stringifyEntityRef` them to handle full refs - which is the canonical universal identifier for entities.
 
 
 **Answer:** 
 
 Currently we are hardcoded the default entity into the policy.ts because we are yet to discover how we can dynamically load the entities in the policy files and we are opened it for the contributors/adopter who can modify it based on their requirement .
 
 
 
image


The Basic motive/idea of this plugin is we can use openfga and backstage to handle the policies dynamically , here in this plugin we prove that we can control the policies using openfga Api with support of backstage core plugins.

**Query:**

The actual policy hard codes an entity name etc, I am not sure what's up with that, does this code really do what the end user thinks it does?


**Answer:**

In the webinar that we presented in the cncf we mentioned about why we need to use this plugin? how it can be used based on some dynamic modifications etc


A webinar submitted to cncf on this: https://youtu.be/wWFbLPvwOyQ?si=_odKh_afogm8Ta5K

AmbrishRamachandiran:
Commenting to avoid stale

AmbrishRamachandiran:
@vinzscam @freben can you please review it?

AmbrishRamachandiran:
> > I'm a bit confused by the code. Some feedback on the plugin itself that I recommend fixing:
> > The installation instructions tell users to add `import { permissionModuleACatalogPolicy } from '../../../plugins/openfga/src/module';`, that doesn't seem right; nobody will copy this actual code
> > `filter: { 'kind=component': CATALOG_FILTER_EXISTS }` should be `filter: { kind: 'component' }`, but is it really just for components?
> > If you do `fields: ['metadata.name']` and handle plain names, you will never work with catalogs that use other namespaces than `default`. You should _at least_ do `fields: ['kind', 'metadata.namespace', 'metadata.name']` and `stringifyEntityRef` them to handle full refs - which is the canonical universal identifier for entities.
> > Your client.ts doesn't inject `fetchApi`; [it's best practice](https://backstage.io/docs/architecture-decisions/adrs-adr013) to always use the fetch api to ensure that people who customize it still can get their metrics and behaviors etc as they expect.
> > Your policy is named `a-name-policy` which is confusing to users
> > There's a config section that has comments saying that you should modify those settings by hand in code, which of course isn't an option for an adopter who just imports your module
> > The actual policy hard codes an entity name etc, I am not sure what's up with that, does this code really do what the end user thinks it does?
> 
> **Query :**
> 
> The installation instructions tell users to add import { permissionModuleACatalogPolicy } from '../../../plugins/openfga/src/module';, that doesn't seem right; nobody will copy this actual code
> 
> **Answer:**
> 
> To enable the policy that will be used can be found in **plugins/openfga/src/module** thats why we added it in the instruction like to use this once they added openfga plugin they need to add the module in the backend file.
> 
> **similar reference : The OPA Plugin**
> 
> image
> 
> **Query:**
> 
> `filter: { 'kind=component': CATALOG_FILTER_EXISTS }` should be `filter: { kind: 'component' }`, but is it really just for components? If you do `fields: ['metadata.name']` and handle plain names, you will never work with catalogs that use other namespaces than `default`. You should _at least_ do `fields: ['kind', 'metadata.namespace', 'metadata.name']` and `stringifyEntityRef` them to handle full refs - which is the canonical universal identifier for entities.
> 
> **Answer:**
> 
> Currently we are hardcoded the default entity into the policy.ts because we are yet to discover how we can dynamically load the entities in the policy files and we are opened it for the contributors/adopter who can modify it based on their requirement .
> 
> image
> 
> The Basic motive/idea of this plugin is we can use openfga and backstage to handle the policies dynamically , here in this plugin we prove that we can control the policies using openfga Api with support of backstage core plugins.
> 
> **Query:**
> 
> The actual policy hard codes an entity name etc, I am not sure what's up with that, does this code really do what the end user thinks it does?
> 
> **Answer:**
> 
> In the webinar that we presented in the cncf we mentioned about why we need to use this plugin? how it can be used based on some dynamic modifications etc
> 
> A webinar submitted to cncf on this: https://youtu.be/wWFbLPvwOyQ?si=_odKh_afogm8Ta5K

@freben please let us know the queries are addressed?

freben:
> similar reference : The OPA Plugin

But that's not the same at all. They instruct the user to do `backend.add(import('their npm package'))` which is correct. Then they effectively say, "if you are curious about what the default policy looks like, you can peek in the code here in our repo what it looks like". They do NOT tell users to "`import` this file from your local file system", because that file does not exist, so the code will fail.

And I still don't understand what you mean [in this file](https://github.com/Infosys/openfga-plugin-backstage/blob/main/plugins/openfga/src/openFgaConfig.ts), the customer won't have your source code, will they? They don't clone your repo, they install your package from NPM.

It's also only a frontend package, that has instructions for installation into the backend. Frontend and backend packages can't import each other across those worlds. There's just so much that seems to not be quite right with this code, and for a security-related feature, I prefer not to add it to the plugins repository in its current shape.

AmbrishRamachandiran:
Don't stale we are working on the fixes

AmbrishRamachandiran:
@freben 

We have renamed the policy to permissionModuleCatalogPolicy to make it clear and intuitive for users. We have also removed all hard-coded values and provided clear configuration instructions. Users can now configure the policy through the app-config.yaml file without needing to modify any code directly.

We have made the following key changes based on your feedback:

- Updated installation instructions to reference the npm package correctly.
- Injected fetchApi to allow for customization and consistency.
- Renamed the policy and removed hard-coded values, providing clear configuration instructions.
- we have created different plugins packages , **front end plugin** [@infosys_ltd/openfga-plugin-backstage](https://www.npmjs.com/package/@infosys_ltd/openfga-plugin-backstage)
**common package** [@infosys_ltd/backstage-plugin-openfga-common](https://www.npmjs.com/package/@infosys_ltd/backstage-plugin-openfga-common)
**plugin module** [@infosys_ltd/backstage-plugin-permission-backend-module-catalog-policy](https://www.npmjs.com/package/@infosys_ltd/backstage-plugin-permission-backend-module-catalog-policy)

These changes address the concerns raised and enhance the usability and security of the OpenFGA Plugin for Backstage. We appreciate your guidance and look forward to your feedback.

AmbrishRamachandiran:
@freben looking forward to your feedback please review the restructured plugins and let us know if any changes need to be done for approval.

freben:
`@infosys_ltd/backstage-plugin-permission-backend-module-catalog-policy` should not be `catalog-policy` at the end right? It's an `openfga-policy`, that happens to currently have rules in place for the catalog but there may be more in the future.

But the bigger problem is, a user can't really install that policy. If they do, the effect will be that their instance is basically broken. ALL operations in their entire Backstage instance will be denied, except on the catalog where they can see everything and perform every action except perhaps not delete an example entity that does not exist.

I am closing this for now. I recommend running your policy and solution in your own production environment for some time to see that everything runs smoothly, and then maybe open a new PR.

PR #27985, opened 2024-12-04
tighten the types in the incremental provider some more

freben:
Happened to see that there was some more opportunity to move to the proper types. Gonna piggyback on the changeset from #27972


Rugvip (approved):
Kinda needs a changeset right?

freben:
see description, piggybacks on a different just merged thing

PR #27989, opened 2024-12-04
avoid extra order by in by-entities when not asked for

freben:
If we avoid the extra `ORDER BY` on the `search` table when the user didn't specify an ordering, we can gain a significant amount of performance. For those bare queries, there's still a final ordering on `entity_id` to keep the result set stable.

The majority of the change here, is dealing with the fact that there no longer is any default for `sortField`, so it can be undefined.

You may want to review this while ignoring whitespace.

### DIFF
--- a/.changeset/stale-hats-exist.md
+++ b/.changeset/stale-hats-exist.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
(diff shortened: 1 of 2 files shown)


vinzscam (approved):
🚀

benjdlambert (approved):

PR #27990, opened 2024-12-04
Fix check for --no-node-snapshot

brianphillips:

There is a helper to ensure that, when running the scaffolder plugin in node >=v20, that the `--no-node-snapshot` is specified. The helper checks the `NODE_OPTIONS` environment variable but it is also a valid option to specify `--no-node-snapshot` on the `node` command line as a literal argument. Options passed to the `node` binary can be found in `process.execArgv` but the code previously was checking just `process.argv` which is only for arguments provided to the script being executed by `node`.


benjdlambert (approved):
Nice catch thanks!

PR #28012, opened 2024-12-06
[docs] Update provider.md for newest provider

jolies93:
Corrected the following that we encountered when using this documentation:
- Added to the callback url path - without this correction we received errors
- Removed the SCOPES - this wouldn't initialize for us, got errors until we removed it

We believe this may have something to do with backstage version or dependency changes - the url hadn't been changed in 3 years and the SCOPES were added 9 months ago

### DIFF
--- a/docs/auth/atlassian/provider.md
+++ b/docs/auth/atlassian/provider.md
@@ -46,9 +46,9 @@ auth:
       development:
         clientId: ${AUTH_ATLASSIAN_CLIENT_ID}
         clientSecret: ${AUTH_ATLASSIAN_CLIENT_SECRET}
-        audience: "https://api.atlassian.com"                                                  # Optional: Specifies the intended recipient of tokens
-        callbackUrl: "https://backstage.example.com/api/auth/atlassian/handler/frame"          # Optional: Must match the redirect URL set in Atlassian OAuth settings.
-        additionalScopes:                                                                      # Optional: Additional permissions requested from Atlassian.
+        audience: "https://api.atlassian.com"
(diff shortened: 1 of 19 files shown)


docs/auth/atlassian/provider.md
```diff
@@ -46,18 +46,16 @@ auth:
...
         clientSecret: ${AUTH_ATLASSIAN_CLIENT_SECRET}
-        scope: ${AUTH_ATLASSIAN_SCOPES}
         signIn:
           resolvers:
             # See https://backstage.io/docs/auth/atlassian/provider#resolvers for more resolvers
             - resolver: usernameMatchingUserEntityName
 ```
 
-The Atlassian provider is a structure with three configuration keys:
+The Atlassian provider is a structure with two configuration keys:
```
awanlin, line 55:
```suggestion
The Atlassian provider is a structure with the following configuration keys:
```

docs/auth/atlassian/provider.md
```diff
@@ -46,18 +46,16 @@ auth:
       development:
```
awanlin, line 46: Would it make sense to include all the config here? There are some items that are missing, though optional, that would be good to see here. 

https://github.com/backstage/backstage/blob/master/plugins/auth-backend-module-atlassian-provider/config.d.ts

awanlin (changes requested):
Thanks for the updates @jolies93, left a few small comments.

docs/auth/atlassian/provider.md
jolies93: @awanlin - added the optional parameters - thanks for the link that made it easy!

PR #28013, opened 2024-12-06
Add new Container component in Canon

cdedreuille:

CleanShot 2024-12-06 at 17 54 38@2x

### DIFF
--- a/packages/canon/src/components/Container/Docs.mdx
+++ b/packages/canon/src/components/Container/Docs.mdx
@@ -17,7 +17,11 @@ import { containerProperties } from './sprinkles.css';
 </Text>
 
 <Title type="h2">Installation</Title>
-<Source code={`yarn install @backstage/canon/container`} language="bash" dark />
+<Source
+  code={`import { Container } from "@backstage/canon";`}
+  language="tsx"
+  dark
+/>
 
 <Title type="h2" style={{ marginBottom: '16px' }}>
   API reference
--- a/packages/canon/src/components/Container/Docs.mdx
+++ b/packages/canon/src/components/Container/Docs.mdx
@@ -1,0 +1,74 @@
+import { Meta, Unstyled, Source } from '@storybook/blocks';
+import * as ContainerStories from './Container.stories';
+import { Title, Text } from '../../../docs/components';
+import { PropsTable, getBoxProps } from '../../../docs/components/PropsTable';
+import { spacingProperties } from '../../layout/sprinkles.css';
+import { containerProperties } from './sprinkles.css';
+
+<Meta of={ContainerStories} />
+
+<Unstyled>
+
+<Title type="h1">Container</Title>
+
+<Text>
+  The container component let you use our default max-width and center the
+  content on the page.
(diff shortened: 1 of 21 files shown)


packages/canon/src/components/Container/Docs.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+<Title type="h1">Container</Title>
+
+<Text>
+  The container component let you use our default max-width and center the
+  content on the page.
+</Text>
+
+<Title type="h2">Installation</Title>
+<Source code={`yarn install @backstage/canon/container`} language="bash" dark />
```
Rugvip, line 20: Hold up, this doesn't actually work 😅 

Needs to be an import or install of `@backstage/canon`

packages/canon/src/components/Container/Docs.mdx
```diff
@@ -0,0 +1,74 @@
...
+  dark
+/>
+
+<Title type="h3">Responsive padding & margin</Title>
+<Text>
+  The Container component also supports responsive values, making it easy to
+  create responsive designs.
+</Text>
+<Source
+  code={`<Container paddingY={{ xs: 'sm', md: 'md' }}>
```
Rugvip, line 65: Would this be "slightly wider content on xs", or something else? that's what I'd expect anyway.

packages/canon/src/components/Grid/Docs.mdx
```diff
@@ -22,11 +22,11 @@ import { spacingProperties } from '../../layout/sprinkles.css';
   The grid component is made of two parts: the grid container and the grid item. Import all parts and piece them together.
 </Text>
 <Source
-  code={`import * as Grid from "@backstage/canon/grid";
+  code={`import { Grid, GridItem } from "@backstage/canon/grid";
```
Rugvip, line 25: How come? I do like `Grid.Item`. If you wanna keep that you can use `export Object.assign(Grid, { Item: GridItem })`

Rugvip (approved):
🚢  & 🇮🇹

packages/canon/src/components/Container/Docs.mdx
cdedreuille: Oh you're right. Radix is actually creating dedicated packages for each of their component. Not sure what I was thinking here :)

packages/canon/src/components/Grid/Docs.mdx
cdedreuille: Ah you're right. I'll change that.

packages/canon/src/components/Container/Docs.mdx
cdedreuille: This is not the content I'm changing here but the padding on top and bottom. This component essentially set a max-width and center the content. You can only change the width of the content at a top level in the CSS variable.

packages/canon/src/components/Container/Docs.mdx
cdedreuille: I fixed that.

PR #28016, opened 2024-12-07
implement single-item fetches to go by refs instead of filters

freben:
Felt silly to do this via the search table unnecessarily 😅


Rugvip (approved):
Yup 😅

PR #28019, opened 2024-12-07
backend-app-api: enable service overrides over feature loaders

Rugvip:

This adds the ability to override service factories from feature loaders by explicitly adding a different factory via `backend.add(serviceFactory)`. You still can't add conflicting service factories across feature loaders, that'll lead to an error.

The purpose of this is to be able to use a feature loader to install common service factories across multiple backends, but still be able to install overrides in individual backends without the need for config flags, separate imports, etc.

### DIFF
--- a/docs/backend-system/architecture/07-feature-loaders.md
+++ b/docs/backend-system/architecture/07-feature-loaders.md
@@ -75,7 +75,7 @@ export default createBackendFeatureLoader({
 
 ### Overriding service factories
 
-Service factories registered by feature loaders have lower priority by ones added directly via `backend.add`. This allows you to use a feature loader for a larger number of service implementations, but still override individual services.
+Service factories registered by feature loaders have lower priority than ones added directly via `backend.add`. This allows you to use a feature loader for a larger number of service implementations, but still override individual services.
 
 The ordering in which different feature loaders or service factories are added does not matter. There is also no priority between feature loaders, if two different feature loaders add a factory for the same service, the backend will fail to start.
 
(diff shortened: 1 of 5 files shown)


packages/backend-app-api/src/wiring/ServiceRegistry.ts
```diff
@@ -191,6 +191,16 @@ export class ServiceRegistry {
     }
   }
 
+  hasBeenAdded(ref: ServiceRef<any>) {
+    if (ref.id === coreServices.pluginMetadata.id) {
+      return true;
+    }
+    if (ref.multiton) {
```
freben, line 198: This may be unfortunate to have here right? Unexpected by future callers. Could it be better to add this multiton check on the calling side?

Rugvip: Hmm, maybe. It currently mirrors logic from a bit further up. I'll try it out

docs/backend-system/architecture/07-feature-loaders.md
```diff
@@ -73,6 +73,29 @@ export default createBackendFeatureLoader({
 });
 ```
 
+### Overriding service factories
+
+Service factories registered by feature loaders have lower priority by ones added directly via `backend.add`. This allows you to use a feature loader for a larger number of service implementations, but still override individual services.
```
freben, line 78:
```suggestion
Service factories registered by feature loaders have lower priority than ones added directly via `backend.add`. This allows you to use a feature loader for a larger number of service implementations, but still override individual services.
```

PR #28052, opened 2024-12-09
Canon - Remove vanilla extract in favour of pure CSS

cdedreuille:


Rugvip (approved):
So still sprinkles? but no extract

cdedreuille:
> So still sprinkles? but no extract

Yes for now. Sprinkles is the function to generate all class utilities at build time. I explored a bit how to build our own but this will require too much time for now. All the other ones were CSS-in-JS library but since we moved to pure CSS we don't need those.

One thing though that I would like to mention is that I'm exploring using Lightning CSS to bundle our CSS and make sure that it is backward compatible with old browsers and to minify it at build time.

https://lightningcss.dev/

PR #28060, opened 2024-12-09
notifications: refactor backend routes to avoid using root route

Rugvip:

Some work towards #24963, making sure the notifications plugin uses similar API design to other plugins.

I did consider updating the `POST /update` endpoint to be `PATCH /notifications` instead as well @backstage/notifications-maintainers , with body switching to `[{ id: '...', ...changeFields }, ...moreUpdates ]` instead. Lemme know if you think that's a better path, but felt it was best to keep the update simply for now.

### DIFF
--- a/.changeset/mighty-birds-rest.md
+++ b/.changeset/mighty-birds-rest.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-notifications': minor
+---
+
+Switched to using the new `/notifications` endpoints. Be sure to update the `notifications` plugin backend before deploying this frontend plugin change.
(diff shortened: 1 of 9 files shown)


drodil:
Looks good! Btw, is there any better way to deprecate API endpoints than just leave them and remove them at some point in the future?

drodil:
Also related #28050

freben (approved):

Rugvip:
@drodil `/v1/*`, `/v2/*` endpoints is what we'd typically use, and deprecate and remove full chunks of those. I think we'll want to put everything under `/v1/` eventually but thinking it's fine to experiment a bit in this area still.

Gonna ship this one to get it in for the next release to make it easier to try out.

PR #28061, opened 2024-12-10
Do not add /index.html for html pages in getEntityDocs

faeshaas:

When use_directory_urls: false is used in the mkdocs configuration techdocs still adds /index.html at the end of the url which breaks the GET request of the page. 


faeshaas:
@freben Can you have a look?

faeshaas:
Please review

jboeijenga:
@Rugvip or @freben: What is needed to get this this PR reviewed and merged?

plugins/techdocs/src/client.ts
```diff
@@ -171,7 +171,9 @@ export class TechDocsStorageClient implements TechDocsStorageApi {
     const url = `${storageUrl}/${namespace}/${kind}/${name}/${path}`;
 
     const request = await this.fetchApi.fetch(
-      `${url.endsWith('/') ? url : `${url}/`}index.html`,
+      `${
+        url.endsWith('/') || url.endsWith('.html') ? url : `${url}/`
+      }index.html`,
```
johnphilip283, line 176: this line still keeps the `index.html` suffix to the end of the request, regardless of if the original url ends with `.html`.

faeshaas: Resolved it 👍

plugins/techdocs/src/client.ts
```diff
@@ -171,7 +171,9 @@ export class TechDocsStorageClient implements TechDocsStorageApi {
     const url = `${storageUrl}/${namespace}/${kind}/${name}/${path}`;
 
     const request = await this.fetchApi.fetch(
-      `${url.endsWith('/') ? url : `${url}/`}index.html`,
+      `${
+        url.endsWith('/') || url.endsWith('.html') ? url : `${url}/index.html`
+      }`,
```
squid-ney, line 176: Hi! I believe this would break in the case `use_directory_urls: false` because it would return just `/` (directory url) when it should return `/index.html`.  I think you could get the configuration value for `use_directory_urls` following a similar pattern as this function: [validateMkdocsYaml](https://github.com/backstage/backstage/blob/8de3d2d53d82dd2ec23359be944050c7e98fe565/plugins/techdocs-node/src/stages/generate/helpers.ts#L272). You could then construct the request url based on that configuration.

faeshaas: Thanks for noticing this! I solved it a bit different and also tried to make the statement a bit better readable. I think now all use_directory_urls use-cases are covered, feel free to share your feedback on it 👍

johnphilip283: Hm, while this new implementation does technically work, it is a little unclear to someone reading the code why this is implemented this way and has a silent coupling to the `use_directory_urls` configuration parameter. I would recommend going with the above approach proposed by @squid-ney so that it's clear how that configuration is tied to the implementation. Also, please include an updated test that tests out both scenarios where `use_directory_urls` is both false and true to make sure the appropriate url is being supplied.

faeshaas: I agree, it makes more sense and makes it way better readable. Added the implementation including tests as suggested.

faeshaas: As you can see there is an issue regarding dependencies because the current dependency between @backstage/plugin-techdocs-node & @backstage/plugin-techdocs is not allowed.

Is this maybe the reason this was not done before as well? Because in the 'old' implementation there was a check if it ends with a '/' or not, indeed with no relation to the use_directory_urls. With this dependency issue I'm starting to doubt the previous implementation was done like this on purpose? 

What are your thoughts? @johnphilip283 and @squid-ney

faeshaas: Any feedback on above?

faeshaas:
Still waiting for feedback before I can make it work. I can also revert the last change and do it the 'ugly' way which will functionally work but is a bit less readable.

@johnphilip283 @squid-ney

faeshaas:
Because of the lack of feedback and the fact this PR is already open for months. I decided to revert my latest change and set it back to the point where the pipeline succeeded and also the implementation was working as verified by @johnphilip283. It might not be the best readable piece of code but it does do the trick and it solves a nasty bug. So please approve so it can be merged.

squid-ney:
Hi @faeshaas, so sorry for the late response. I tried out your solution locally and set the example-documentation to use `use_directory_urls: false`. While the actual html file is fetched correctly, the css stylessheet is not fetched correctly and results in a bug where the documentation appears to never load and the loading bar spins infinitely. 
Screenshot 2025-02-18 at 2 23 01 PM
Screenshot 2025-02-18 at 2 23 19 PM

The html page includes this link to the stylesheet: `<link rel="stylesheet" href="assets/stylesheets/main.3cba04c6.min.css">`. And since it is a relative path it appends it to the current url which is in this case is `http://localhost:3000/docs/default/component/documented-component/sub-page.html` which results in a faulty url: `http://localhost:7007/api/techdocs/static/docs/default/component/documented-component/sub-page.html/assets/stylesheets/main.3cba04c6.min.css`. This url should be: `http://localhost:7007/api/techdocs/static/docs/default/component/documented-component/assets/stylesheets/main.3cba04c6.min.css`. 

Do you run into this bug as well?

squid-ney:
I agree, looks like TechDocs does not support `use_directory_urls:false` and there are assumptions baked in that assume a directory file structure. I think we will need a more robust solution here vs a quick bugfix. Let's go with the more readable explicit solution proposed by John.

faeshaas:
Yes, I was able to reproduce the same behavior locally and I also assume there are some things hard-coded into the generation related to the use_directory_urls: true setting. But I was not yet able to find the root-cause of this issue. 

And as response on your other comment. I also like the readable, explicit solution more. But it will cause a not allowed dependency between @backstage/plugin-techdocs-node & @backstage/plugin-techdocs which I'm not sure how to solve or work around it.. 

So if you have any ideas on both issues feel free to share them. I will also have another look to see if there is a nice, solid way to fix this issue.

faeshaas:
So I was able to fix the issue you mentioned @squid-ney. In the [getBaseUrl](https://github.com/backstage/backstage/pull/28061/files#diff-b74fd6baa66b6b7c837933ee5daf530343f37ca59baea6041323e208f5b375d6R263) there is a similar check on the url if it ends with a '/' which doesn't take the use_directory_urls option into account. I did a quickfix to verify this is the problem by just checking the url and ignore it when it ends with '.html' so it will not add a '/' to it which causes the issue. 

This seems to work as expected. So to fix it in a neath way I can again apply the nice fix as proposed by @johnphilip283 but still I have the dependency problem which makes me unable to even push it. I can test it locally but I would like to get some advice on how to deal with this dependency.

squid-ney:
Thanks for fixing the bug. Yes, we would need to expose the `use_directory_urls` setting to the FE, one way could be through techdocs metadata.

However,  I'd like to understand your use case for `use_directory_urls:false`, is there a limitation with the storage solution you are using? Perhaps you could create your own implementation of [TechDocsStorageApi](https://github.com/backstage/backstage/blob/55114cfeb7045e3e5eeeaf67546b58964f4adcc7/plugins/techdocs/src/api.ts#L33) with the current logic from this PR. This would unblock you and you could implement the faster approach without worrying about other breaking changes. I'm not sure `use_directory_urls :false` is something we want to support on the default TechDocsStorageApi implementation, but hearing more about your use case would help make that decision.

faeshaas:
Let me try to explain our use-case. 
When refering from 1 md file to another using a relative path it will not work when you try to refer to a file on a level higher. In our case these relative links are generated in our Mermaid diagrams.

But as simple example:
File1 = /docs/default/app1/folder1/document1.md
File2 = /docs/default/app1/folder2/document2.md

When refering to document2.md in document1 we use something like "<a href="../folder2/document2>Document 2</a>" in document1.md

The combination of having enables use_directory_urls and TechDocs in all items in the navigation bar a '/' is added for every page. This means the Techdocs page for document1 is this: http://localhost:3000/docs/default/app1/folder1/document1/

Because this last '/' in the url Techdocs/Mkdocs treat this as a folder. So the link I created in document1 is pointing to: http://localhost:3000/docs/default/app1/folder1/folder2/document2

As you can see this is the wrong link because it should be: http://localhost:3000/docs/default/app1/folder2/document2

The reason for this is because of this last '/' which is added to the url cause mkdocs to think it is a folder. When removing this '/' it will work as expected. 
In order to workaround this we investigated using the use_directory_urls=false to avoid this problem because the '/' is not added in the end and we could use default mkdocs functionality to fix it.

That's where the fun started and we figured out this configuration might not be supported in Techdocs.

Also good to mention, we use the following mkdocs plugins. 
  - monorepo
  - same-dir
  - awesome-pages
  
I did not investigate if one of these plugins might cause this problem by adding the '/' on items in the navigation bar. That would be another solution direction. But our first thought was we would be able to fix it in Techdocs.

Hopefully this explaination makes our use-case a bit clear

squid-ney:
Hi @faeshaas thanks for detailing your use case more. A few follow up questions:
- Just want to confirm that these links are generated by mermaid and you don't have control to set them? If you are able to override the links, maybe changing the relative path `<a href="../folder2/document2>Document 2 ` to point to the actual file `<a href="../folder2/document2.md>Document 2 ` fixes it? Also, anchor tags can be tricky to get to work with TechDoc relative links. If you use this markdown style linking: `[Relative link to Docs Home](../../index.md)`, that may fix it.
- Are you using the [backstage-plugin-techdocs-addon-mermaid](https://www.npmjs.com/package/backstage-plugin-techdocs-addon-mermaid)?
- I'm not familiar with mermaid generating doc links, can you explain this more?
- Curious, if you don't use the `same-dir` plugin and follow the more common `/docs` `mkdocs.yaml` structure, do you still have broken links?

After hearing more about this scenario, I'm hesitant that the right fix is to make changes to the `TechDocsStorageClient`. We can think of some other solutions based on answers to the above questions, but if you'd like to quickly unblock yourself then I'd recommend creating your own implementation of [TechDocsStorageApi](https://github.com/backstage/backstage/blob/55114cfeb7045e3e5eeeaf67546b58964f4adcc7/plugins/techdocs/src/api.ts#L33).

faeshaas:
The issue is not really the Mermaid part of it, I can also reproduce this issue in a simple project with 2 md files living in a different folder where one links to another. 
I tried your first suggestion and played around a bit in the way links are created. I removed all plugins and created a docs folder. Some variantions I tried:

image

When using the same-dir plugin and awesome-pages it gives the same result:
image

So there seems to be 1 option that is working in both cases which is in markdown:
[This link will work in Techdocs and it will link correctly in plain md](../folder2/test.md)

Because the Mermaid output is generated in HTML the option that we would like to use is:
<a href="../folder2/test.md">This link will not work in Techdocs but will link correctly in plain md</a>

But this one will not work because, let's get the circle complete again, it will add '/index.html' behind the link so it points to "../folder2/test.md/index.html" which does not exist.

This Mermaid output we generate our self these links we also generate. So we are fully in control in changing these but as stated above. There is currenty no variantion in HTML which will work in markdown and in Techdocs. We need to support both.

squid-ney:
Thanks for sharing more!
TechDocs uses MkDocs which uses the [Python-Markdown](https://python-markdown.github.io/) library to generate HTML pages from markdown. We are limited in what mkdocs supports, here is their documentation on linking:
https://www.mkdocs.org/user-guide/writing-your-docs/#linking-to-pages

If changing the format of your generated links causes a lot of manual work, then I'd suggest a good option could be to create a TechDocs content addon. This way you could access and manipulate a TechDocs site's DOM. Here is the documentation for content addons: https://backstage.io/docs/features/techdocs/addons/#addons-in-the-content-location You also could build this outside of the open source repository and use it directly for your own TechDocs.

PR #28074, opened 2024-12-10
break circular imports

freben:


plugins/config-schema/src/components/SchemaView/SchemaView.tsx
```diff
@@ -15,12 +15,22 @@
...
+);
+const MatchView = React.lazy(() =>
+  import('./MatchView').then(m => ({ default: m.MatchView })),
+);
+const ObjectView = React.lazy(() =>
+  import('./ObjectView').then(m => ({ default: m.ObjectView })),
+);
+const ScalarView = React.lazy(() =>
+  import('./ScalarView').then(m => ({ default: m.ScalarView })),
+);
```
Rugvip, line 32: so these are a bit strange, how did you arrive at this? 😁

freben: I'm not sure! Haha :)

Like, these are truly circularly importing each other and the build warns about that. I just wanted to fix it somehow. Is this approach a problem?

vinzscam: perhaps try to merge the recursive components in a single file?

freben: Sure. But if not merging all that into one file, I'd still be curious if the lazy thing is unexpected or doesn't work well 🤔

vinzscam (approved):

Rugvip (approved):

Rugvip (approved):

PR #28076, opened 2024-12-10
Dynamic frontend features based on module federation and the new frontend system

davidfestal:

This PR contributes support for the dynamic frontend features, which completes the already-existing support for backend dynamic features provided by package [backend-dynamic-feature-service](/backstage/backstage/tree/master/packages/backend-dynamic-feature-service).

Dynamic frontend features are based on both the module federation support and the new frontend system.

In more details:
- On the backend it exposes the module federation assets of any frontend dynamic plugin found in the dynamic plugins root folder, as well as a list of all frontend dynamic plugins module federation manifests
- On the frontend side, it provides a `CreateAppFeatureLoader` which loads the exposed module federation remote modules through the module federation runtime library.

##### Testing

This PR can be more easily tested using the content of the temporary PR #28580 which based on this one, and provides updated frontend and backend application to test it, as well as a detailed description of steps to test the current PR.


freben:
Hi! Feel free to rebase this while waiting for us to get back up to speed.

davidfestal:
> Hi! Feel free to rebase this while waiting for us to get back up to speed.

Thanks @freben.
This is done, and I also updated the dependencies to module federation packages to 0.8.0 as has been done in the CLI in the meantime.

davidfestal:
@Rugvip Any idea when you would be able to review this PR ?
During the Framework SIG meeting on January 7th, when this PR was presented for review (after demoing it on Dec 2024), there seemed to be a general agreement on it with other Core maintainers.
But since you expressed the need to test it manually before approving, I created a second [temporary PR](https://github.com/backstage/backstage/pull/28580) whose content makes it much easier to quickly test the current PR. See the new `Testing` sub-section of the current PR description.

Rugvip:
@davidfestal thank you! Yep just need to get to it, there's quite a couple of big ones on the way

davidfestal:
Any news ?

packages/frontend-defaults/src/discovery.ts
```diff
@@ -81,7 +81,12 @@ export function getAvailableFeatures(config: Config): FrontendFeature[] {
   );
 }
 
-function isBackstageFeature(obj: unknown): obj is FrontendFeature {
+/**
+ * Checks whether a given value is a valid backstage frontend feature.
+ *
+ * @public
+ */
+export function isBackstageFeature(obj: unknown): obj is FrontendFeature {
```
freben, line 89: Hm if we are exporting this, I wonder if it belongs better in `@backstage/frontend-plugin-api`. @Rugvip?

packages/frontend-dynamic-feature-loader/README.md
```diff
@@ -0,0 +1,32 @@
...
+Backstage frontend feature loader to load new frontend system plugins exposed as module federation remotes.
+The frontend feature loader provided in this package works hand-in-hand with the server of frontend plugin module federation remotes server which is part of backend dynamic feature service in package `@backstage/backend-dynamic-feature-service`.
+
+**NOTE: The [new frontend system](https://backstage.io/docs/frontend-system/) that this package is relaying upon is in alpha, and we do not yet recommend using it for production deployments**
+
+## Usage
+
+- To enable this loader, you should:
+
+  - Enable the backend dynamic features in your backend application, as explained in `packages/backend-dynamic-feature-service/README.md#how-it-works`
```
freben, line 12: probably worth making that a clickable link

packages/frontend-dynamic-feature-loader/package.json
```diff
@@ -0,0 +1,76 @@
+{
+  "name": "@backstage/frontend-dynamic-feature-loader",
+  "version": "0.0.1-next.0",
```
freben, line 3:
```suggestion
  "version": "0.0.0",
```

and combine it with a `minor` level changeset

packages/frontend-dynamic-feature-loader/package.json
```diff
@@ -0,0 +1,76 @@
...
+    "msw": "^1.0.0",
+    "react": "^18.0.2",
+    "react-dom": "^18.0.2",
+    "react-router-dom": "^6.3.0"
+  },
+  "peerDependencies": {
+    "@types/react": "^16.13.1 || ^17.0.0 || ^18.0.0",
+    "react": "^16.13.1 || ^17.0.0 || ^18.0.0",
+    "react-dom": "^16.13.1 || ^17.0.0 || ^18.0.0",
+    "react-router-dom": "6.0.0-beta.0 || ^6.3.0"
```
freben, line 68:
```suggestion
    "@types/react": "^17.0.0 || ^18.0.0",
    "react": "^17.0.0 || ^18.0.0",
    "react-dom": "^17.0.0 || ^18.0.0",
    "react-router-dom": "^6.3.0"
```

since https://github.com/backstage/backstage/pull/28579

packages/frontend-dynamic-feature-loader/report.api.md
```diff
@@ -0,0 +1,18 @@
...
+import { init } from '@module-federation/enhanced/runtime';
+
+// @public
+export function dynamicFrontendFeaturesLoader(
+  options?: DynamicFrontendFeaturesLoaderOptions,
+): CreateAppFeatureLoader;
+
+// @public (undocumented)
+export type DynamicFrontendFeaturesLoaderOptions = {
+  moduleFederation: Omit<Parameters<typeof init>[0], 'name' | 'remotes'>;
```
freben, line 16: I wonder if this is risky (passthrough of a type). Even small bumps of the module-federation library might bring breaking changes for users of this function now right? Maybe it's OK, no strong opinion, but we can't "shield" end users from changing underlying interfaces now, especially in the case of parameters where we would have preferred to provide a default or where usage of the parameter actually can lead to breakages in our implementation. Any opinions there?

packages/backend-dynamic-feature-service/src/schema/openapi.yaml
```diff
@@ -0,0 +1,90 @@
...
+      responses:
+        '200':
+          description: ''
+          content:
+            application/json:
+              schema:
+                type: object
+                properties: {}
+                additionalProperties:
+                  type: string
```
freben, line 82: I'll assume this will need to be evolved later right? I would have expected something less "raw" that explains and names fields, like

```json
{
  "pluginManifests": [
    {
      "pluginId": "foo",
      ...
}
```

packages/backend-dynamic-feature-service/src/schemas/schemas.ts
```diff
@@ -152,8 +157,16 @@ export const dynamicPluginsSchemasServiceFactory = Object.assign(
...
-  schemaLocator: (pluginPackage: ScannedPluginPackage) => string = () =>
-    path.join('dist', 'configSchema.json'),
+  schemaLocator: (
+    pluginPackage: ScannedPluginPackage,
+  ) => string = pluginPackage =>
+    path.join(
+      'dist',
+      PackageRoles.getRoleInfo(pluginPackage.manifest.backstage.role)
+        .platform === 'node'
+        ? 'configSchema.json'
```
freben, line 167: Is this logic sound? In the base case, package.json points out the schema file (which can have any name). And should the logic be inverted such that "the special case" of the dotfile is checked instead?

packages/backend-dynamic-feature-service/src/server/plugin.ts
```diff
@@ -0,0 +1,57 @@
...
+          await createRouter({
+            logger,
+            discovery,
+            metadata,
+            dynamicPlugins,
+          }),
+        );
+        httpRouter.addAuthPolicy({
+          path: `/`,
+          allow: 'unauthenticated',
```
freben, line 52: Is this really needed? Do we want or need them unauthed? Can we use cookie auth like techdocs has instead?

packages/frontend-dynamic-feature-loader/config.d.ts
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export interface Config {
+  /**
+   * @visibility frontend
+   */
+  dynamicPlugins?: {};
```
freben, line 21: another common pattern is `true` as a boolean for the intent of "enable with defaults" , but maybe there's going to be many more fields here later on so it doesn't make sensed to introduce that?

packages/frontend-dynamic-feature-loader/src/loader.ts
```diff
@@ -0,0 +1,196 @@
...
+        error(`Failed initializing module federation`, err);
+        return { features: [] };
+      }
+
+      const features = (
+        await Promise.all(
+          Object.entries(frontendPluginManifests).map(
+            async ([name, manifestLocation]) => {
+              // eslint-disable-next-line no-console
+              console.info(
```
freben, line 124: i'll assume these are meant to be kind of temporary, but would it be better to only log once complete or in a `catch` if it failed? Saying "i'm about to ..." is less useful when we assume that they normally always succeed

freben (commented):
OK, made a more thorough review

packages/frontend-dynamic-feature-loader/src/loader.ts
```diff
@@ -0,0 +1,196 @@
...
+                `Loading dynamic plugin '${name}' from '${manifestLocation}'`,
+              );
+              let manifest: Manifest;
+              try {
+                const response = await fetch(manifestLocation);
```
Rugvip, line 129: I remember in some of the previous testing as we were implementing support in the CLI the manifests didn't work particularly well in practice. Can't remember fully what the issue was though. I do wonder if it makes sense to rely on the hardcoded remote entry point instead? Especially since it avoids this extra fetch.

Rugvip (commented):
Alright, finally got around to digging into this one a bit deeper 😅. Couple of bits here and there but main concern is the plugin addition and whether it might be better to ship this as an experimental addition to the `app` plugin backend instead.

packages/frontend-defaults/src/discovery.ts
davidfestal: I'd be happy to move it there if it's better.

packages/frontend-dynamic-feature-loader/src/loader.ts
```diff
@@ -0,0 +1,196 @@
...
+        return { features: [] };
+      }
+
+      try {
+        init({
+          ...options?.moduleFederation,
+          name: appPackageName
+            .replaceAll('@', '')
+            .replaceAll('/', '__')
+            .replaceAll('-', '_'),
```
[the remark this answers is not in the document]
davidfestal: Isn't it consistent with what is done in the CLI when running with the module federation support: https://github.com/backstage/backstage/blame/1a03dc6f69341df1f4b20a53b60d8f09cfaa2e05/packages/cli/src/lib/bundler/moduleFederation.ts#L42 ?

In addition we default to the 'app' name if none is mentioned in the frontend config.

packages/frontend-dynamic-feature-loader/src/loader.ts
```diff
@@ -0,0 +1,196 @@
...
+      return 'dynamic-plugins-loader';
+    },
+
+    async load({ config }) {
+      const dynamicPLuginsConfig = config.getOptionalConfig('dynamicPlugins');
+      if (!dynamicPLuginsConfig) {
+        return {
+          features: [],
+        };
+      }
```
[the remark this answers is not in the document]
davidfestal: The overall dynamic plugins functionality is enabled by the `dynamicPlugins` config object.
If dynamic plugins are not enabled at all (not even on the backend side), we obviously don't even want to try loading frontend plugins.

packages/frontend-dynamic-feature-loader/src/loader.ts
davidfestal: I'd rather depend on the standard, and most flexible module federation way, which is: use the remote manifest in order to take into account all the options that were set when the module federation remote was built.

packages/frontend-dynamic-feature-loader/report.api.md
```diff
@@ -0,0 +1,18 @@
+## API Report File for "@backstage/frontend-dynamic-feature-loader"
+
+> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
+
+```ts
+import { CreateAppFeatureLoader } from '@backstage/frontend-defaults';
```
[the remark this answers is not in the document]
davidfestal: I'd gladly contribute, but I would need additional guidance and details though, in order to know where to implement it, and what exactly you have in mind.

packages/backend-dynamic-feature-service/src/server/plugin.ts
```diff
@@ -0,0 +1,57 @@
...
+} from '@backstage/backend-plugin-api';
+import { createRouter } from './router';
+import { dynamicPluginsServiceRef } from '@backstage/backend-dynamic-feature-service';
+
+/**
+ * frontendRemotesServerPlugin backend plugin
+ *
+ * @internal
+ */
+export const frontendRemotesServerPlugin = createBackendPlugin({
```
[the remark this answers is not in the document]
davidfestal: > I don't see this one being used anywhere?

It is used in the feature loader: https://github.com/backstage/backstage/pull/28076/files#diff-5da3f59e4424a7b4fd7d7825be308dde884efbe526f21b9ef8500510be2410b7R68.

That's the whole point: have the feature loader install everything required to have dynamic plugins work, based on the presence of the `dynamicPlugins` configuration object.

> Can't have a plugin being exported by a service package though

That's why the plugin is only used internally by the feature loader, and not even exported as a public API from the service package.

> You can use the root router to add routes to the backend instance though, although that will not be discoverable from the frontend.

I guess I can try to do that if the current design is really a blocking problem, since the frontend uses the generated code from the openapi yaml definition.

packages/frontend-dynamic-feature-loader/src/schema/openapi/generated/apis/DefaultApi.client.ts
```diff
@@ -0,0 +1,87 @@
...
+  public async getManifests(
+    // @ts-ignore
+    request: GetManifests,
+    options?: RequestOptions,
+  ): Promise<TypedResponse<{ [key: string]: string }>> {
+    const baseUrl = await this.discoveryApi.getBaseUrl(pluginId);
+
+    const uriTemplate = `/manifests`;
+
+    const uri = parser.parse(uriTemplate).expand({});
```
Rugvip, line 77: `${backend.baseUrl}/.backstage/dynamic-features/manifests` here instead + switch the backend to use `rootHttpRouter` instead

packages/frontend-dynamic-feature-loader/config.d.ts
davidfestal: There are other already other mandatory fields here on the backend side.
The main addition here is that we add the frontend visibility.

davidfestal: There are already other mandatory fields here on the backend side.
The main addition here is that we add the frontend visibility.

freben:
Just popping by to note that I was going through old pull requests and saw this one. There are conflicts and actionable comments above still, right, am I understanding this correctly?

davidfestal:
> There are conflicts and actionable comments above still, right, am I understanding this correctly?

Sure. I had to rework part of the structure to fix some of the review comments, and also rebase it on top of main.
I isolated to changes fixing the review comments in the last 3 commits.

packages/frontend-dynamic-feature-loader/README.md
davidfestal: Done in commit 7452c175b70c1df851251dfa7adc582d529bb5de

packages/frontend-dynamic-feature-loader/package.json
davidfestal: Done in commit 7452c175b70c1df851251dfa7adc582d529bb5de

packages/frontend-dynamic-feature-loader/package.json
davidfestal: Done in commit 7452c175b70c1df851251dfa7adc582d529bb5de

packages/frontend-dynamic-feature-loader/report.api.md
davidfestal: I changed this in commit 7452c175b70c1df851251dfa7adc582d529bb5de to use publicly exported Module Federation types.

packages/backend-dynamic-feature-service/src/schema/openapi.yaml
davidfestal: I completed this schema to now return the real description of a Module Federation remote as expected by the Module Federation Runtime.
It's much more structured.

packages/backend-dynamic-feature-service/src/schemas/schemas.ts
davidfestal: You're right.
This should simply point to the compiled schema of the module federation remote for frontend plugins, and to the same file name for compiled schemas of backend plugins for the sake of consistency.
This has been modified in commit [f7886f6](https://github.com/backstage/backstage/pull/28076/commits/f7886f62fdaee092c85d068ceecde51e4bfb99cd#diff-69befd61a07507b883f67e6c0dec7618f5c026d3b70958e6c1d5dda1ccc20722)

packages/backend-dynamic-feature-service/src/server/plugin.ts
davidfestal: I didn't require this anymore when using a service that adds a router to the rootHttpRouter.
Did I miss something ?

packages/frontend-dynamic-feature-loader/src/loader.ts
davidfestal: I changed them to debug calls, and kept the log level for the message related to load completion.

davidfestal: Fixed in commit 4335b04da0394e2eb126146392ed527c301580a1

packages/frontend-dynamic-feature-loader/package.json
```diff
@@ -0,0 +1,76 @@
...
+  "dependencies": {
+    "@backstage/config": "workspace:^",
+    "@backstage/core-app-api": "workspace:^",
+    "@backstage/core-plugin-api": "workspace:^",
+    "@backstage/errors": "workspace:^",
+    "@backstage/frontend-app-api": "workspace:^",
+    "@backstage/frontend-defaults": "workspace:^",
+    "@backstage/frontend-plugin-api": "workspace:^",
+    "@backstage/types": "workspace:^",
+    "@backstage/version-bridge": "workspace:^",
```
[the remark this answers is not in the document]
davidfestal: Done

packages/frontend-dynamic-feature-loader/report.api.md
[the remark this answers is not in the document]
davidfestal: This has been done, and the `frontend-dynamic-feature-loader` now uses the new `createFrontendFeatureLoader ` function.

packages/frontend-dynamic-feature-loader/package.json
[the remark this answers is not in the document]
davidfestal: Done in commit 7452c175b70c1df851251dfa7adc582d529bb5de

packages/frontend-dynamic-feature-loader/src/loader.ts
davidfestal: Still I the ability to configure / customize the module federation asset serving through a plugin gives the ability to use the Javascript directly instead of the manifest if desired.
However using the manifest is still the default because it's more powerful and, combined with the configurability mentioned above, can allows things like the change of the public path to accommodate CDN use-cases.

packages/frontend-dynamic-feature-loader/src/schema/openapi/generated/apis/DefaultApi.client.ts
davidfestal: Done in commit f7886f62fdaee092c85d068ceecde51e4bfb99cd

davidfestal:
@freben @Rugvip Afaict I adressed all your review comments, either by restructuring some parts, fix details, or just by providing an answer.

davidfestal:
Any idea whether it could be included into the `1.38.0 release ?

scripts/verify-local-dependencies.js
```diff
@@ -77,6 +77,7 @@ const roleRules = [
       '@backstage/core-compat-api',
       '@backstage/dev-utils',
       '@backstage/frontend-defaults',
+      '@backstage/frontend-dynamic-feature-loader',
```
Rugvip, line 80: Hmm, is this really needed? can be cleaned up in a followup though

Rugvip (approved):
Alright, looks good! 👍 

Let's :shipit:

scripts/verify-local-dependencies.js
davidfestal: Oh, sure. I also guess we can remove it.
I'll create a PR for it.

PR #28082, opened 2024-12-10
Canon - Table component

cdedreuille:
In this PR I'm introducing the first version of our Table component. I also cleaned up the CSS structure to introduce a better normalised version by default that will set everything to standard for us with the right font.

The Table component will be done in multiple steps which I'll share in following PRs

CleanShot 2024-12-13 at 17 21 40@2x


Rugvip (approved):
:shipit: 

might be a bit overkill with a story for each of these

cdedreuille:
> :shipit:
> 
> might be a bit overkill with a story for each of these

Maybe... I tried to be smart here :) But you're right I might bring everything back to one.

PR #28085, opened 2024-12-10
feat(GitHubOrgEntityProvider): adds throttling and rate limit handling to GithubOrgEntityProvider

sonikro:

When using the **GithubOrgEntityProvider** in organizations with over 14000 users on a GitHub Organization, the EntityProvider was failing 100% of the time with Secondary Rate Limit errors.

This fix adds the official [throttling](https://github.com/octokit/plugin-throttling.js/tree/main) plugin to the **graphql** client responsible for fetching the data.

- I extracted the code to instantiate the GraphQL Client into a separate function, so we can re-use this in the other EntityProviders later

I tested this package locally and not only the rate limit consumption went down significantly, but when a rate limit error happened, it successfully retried and moved on. 

**PS:** Using octokit versions on this monorepo is getting complicated, as there are so many different versions installed. I spent most of my time trying to find the exact octokit versions I could use that would not break types everywhere. 


sonikro:
Something seems flaky in the Windows E2E test. Can someone retry that ?

sonikro:
Not stale

koalaty-code:
@benjdlambert or @camilaibs, can this get some 👀 ?

sonikro:
Hi there @benjdlambert, @camilaibs, just following up to see if there is anything we can do to get this merged. Our instance is failing the synchronization process 100% of the time now and we had to add some ugly workarounds while this fix isn't released. Let me know if there is anything else I can do.

Thanks, and happy new year :)

awanlin:
Hi @sonikro, can you rebase while you wait for a review, I think that should address the CI failure.

Also, have you looked at using the Events support for this provider? https://backstage.io/docs/integrations/github/org#events-support

sonikro:
> Hi @sonikro, can you rebase while you wait for a review, I think that should address the CI failure.
> 
> Also, have you looked at using the Events support for this provider? https://backstage.io/docs/integrations/github/org#events-support

Hi @awanlin , thanks for the reply. I just rebased the branch.

We are using the Events Support for this provider, however the original open-source code for handling **membership** events causes a full load of every single user of the organization, for every single event, as seen here:

https://github.com/backstage/backstage/blob/8b4ba413a94df01d14aebdd6f6e5ab0b9b165cd0/plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts#L448-L453

So we actually had to override that behavior with our own behavior that only fetches data from a single user. However our current implementation has some concurrency problems, since sometimes we have GitHub Teams that are synched from External Identity Providers, and we can get hundreds of membership events to the same team, at the same time. Since this trigger the `connection.applyMutation` in parallel, and as far as I know, there isn't an easy way to "lock" the entities, we sometimes miss data due to the race condition. Which was ok, since our cronjob would reconciliate any missing members when it ran.

benjdlambert:
I was a little bit concerned about implementing any sort of Rate Limiting or Throttling as part of the processing loop, but I think that keeping this stuff in the providers for now makes somewhat sense. 

When implementing Rate Limiting and Throttling in the catalog as part of the URLReaders, I think that we're going to want to have the ability to throw an error with a `RetryAt` time or something, as we don't want to block processing by waiting for the retry time. This way the processors can run for other entities that might not be blocked on waiting.

benjdlambert (approved):
I went ahead and cleaned up some of the typescript `any` types and it looks like the casting is no longer required either.

Let's give this a go and see where we land.

Rugvip (approved):
Thank you! 🎉

freben (approved):
Looks good!

awanlin:
Hi @sonikro, thanks for the follow up, looks like we got this merged!

As for the events part, could you log an issue about this:

> We are using the Events Support for this provider, however the original open-source code for handling membership events causes a full load of every single user of the organization, for every single event


Would be good to have that logged so that if someone from the Community can help improve it they can. 👍

PR #28098, opened 2024-12-11
fix: breaking API changes comment

aramissennyeydd:

I think I was missing a few pieces when I originally tried to get this working. I used the removed Uffizzi workflow as a base, but I think I removed too much - this should be the right balance of credential isolation and actually working 😁.

 What's the best way to test this?

Failing runs: https://github.com/backstage/backstage/actions/workflows/api-breaking-changes-comment.yml

Message:
```
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
```

### DIFF
--- a/.github/workflows/api-breaking-changes-comment.yml
+++ b/.github/workflows/api-breaking-changes-comment.yml
@@ -20,6 +20,7 @@ jobs:
       git-ref: ${{ steps.event.outputs.GIT_REF }}
       pr-number: ${{ steps.event.outputs.PR_NUMBER }}
       action: ${{ steps.event.outputs.ACTION }}
+      comment-cache-key: ${{ steps.hash.outputs.COMMENT_FILE_HASH }}
     steps:
       - name: Harden Runner
         uses: step-security/harden-runner@0080882f6c36860b6ba35c610c98ce87d4e2f26f # v2.10.2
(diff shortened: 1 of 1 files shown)


Rugvip (commented):
Coming back to this is sure is a quite elaborate setup for posting a feedback comment 😅 

We do have the existing https://github.com/backstage/backstage/blob/master/.github/workflows/automate_merge_message.yml that does almost the same thing, except that it doesn't need to edit an existing comment. We do have that functionality over in https://github.com/backstage/actions/blob/b3c1841fd69e1658ac631afafd0fb140a2309024/changeset-feedback/postFeedback.ts though, and it might simply be worth having a pretty big inline script rather than the split triggering.

Not sure that the split with a separate `add-comment` job is needed either, if we just aim to slim this down a lil bit? I'd expect it to be fine to merge this all into a single job and avoid the cache,

aramissennyeydd:
@Rugvip I had originally cloned this from the Uffizzi workflow which was at the time the best example of that split comment flow. I agree this is a rather complex set up for a feedback comment, I'm happy to rewrite this to reduce complexity in the future, but I likely won't get to it for a bit 😕. Do we want to move forward on this PR as is or should we wait until I can update the full workflow?

PR #28107, opened 2024-12-11
use `useApi` instead of `apiHolder` for `formFieldsApi`

benjdlambert:
Just a small refactor out the back of #28059. Going to hop in the back of that changeset too. I ended up picking this way as there was a load of tests that broke and wanted to get the functionality in the `-next` release to test.

### DIFF
--- a/plugins/scaffolder-react/src/hooks/useCustomFieldExtensions.ts
+++ b/plugins/scaffolder-react/src/hooks/useCustomFieldExtensions.ts
@@ -35,7 +35,7 @@ export const useCustomFieldExtensions = <
   // Get custom fields created with FormFieldBlueprint
   const formFieldsApi = useApi(formFieldsApiRef);
   const [{ result: blueprintFields }, methods] = useAsync(
-    formFieldsApi?.getFormFields,
+    formFieldsApi.getFormFields,
     [],
   );
   useMountEffect(methods.execute);
--- a/plugins/scaffolder-react/src/hooks/useCustomFieldExtensions.ts
+++ b/plugins/scaffolder-react/src/hooks/useCustomFieldExtensions.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import { useAsync, useMountEffect } from '@react-hookz/web';
-import { useApiHolder, useElementFilter } from '@backstage/core-plugin-api';
+import { useApi, useElementFilter } from '@backstage/core-plugin-api';
 import { formFieldsApiRef } from '../next';
 import { FieldExtensionOptions } from '../extensions';
 import {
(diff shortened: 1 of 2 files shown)


plugins/scaffolder/src/components/Router/Router.test.tsx
```diff
@@ -26,26 +30,37 @@ import {
...
 } from '@backstage/plugin-scaffolder-react';
 import { TemplateListPage, TemplateWizardPage } from '../../alpha/components';
+import { formFieldsApiRef } from '@backstage/plugin-scaffolder-react/alpha';
 
 jest.mock('../../alpha/components', () => ({
   TemplateWizardPage: jest.fn(() => null),
   TemplateListPage: jest.fn(() => null),
 }));
 
+const renderInApp = (element: React.ReactElement, opts?: TestAppOptions) =>
```
acierto, line 40: Maybe this name (`renderInApp `) can be a bit more informative to highlight what the difference is between `renderInApp` and `renderInTestApp` without the need to go to read the code of it.

plugins/scaffolder-react/src/hooks/useCustomFieldExtensions.ts
```diff
@@ -33,10 +33,9 @@ export const useCustomFieldExtensions = <
   outlet: React.ReactNode,
 ) => {
   // Get custom fields created with FormFieldBlueprint
-  const apiHolder = useApiHolder();
-  const formFieldsApi = apiHolder.get(formFieldsApiRef);
+  const formFieldsApi = useApi(formFieldsApiRef);
   const [{ result: blueprintFields }, methods] = useAsync(
-    formFieldsApi?.getFormFields ?? (async () => []),
+    formFieldsApi?.getFormFields,
```
freben, line 38: Now you can remove the question mark. But wasn't that intentional, that the api holder was used so that the API was explicitly optional to provide?

freben (approved):

plugins/scaffolder-react/src/hooks/useCustomFieldExtensions.ts
benjdlambert: Na, there's a default implementation shipped with the `plugin` anyways, which just returns `[]`.

In the new frontend system it will automatically add things to that array, but in the old system for backwards compat it just returns empty. You can of course move away from `JSX` and use that API if you want and provide your own implementation, but it's probably not worth it.

plugins/scaffolder/src/components/Router/Router.test.tsx
benjdlambert: Yeah, for sure. Let me fix!

benjdlambert:
@acierto does this look better? :pray:

acierto:
@benjdlambert Yup, well defined in name 👍

acierto (approved):

PR #28109, opened 2024-12-11
Move startup lifecycle back to plugin service

camilaibs:

Follow-up of https://github.com/backstage/backstage/pull/27961

Fix server response time by moving the lifecycle startup hooks back to the plugin lifecycle service.

### DIFF
--- a/.changeset/fast-tools-fix.md
+++ b/.changeset/fast-tools-fix.md
@@ -2,7 +2,7 @@
 '@backstage/backend-defaults': minor
 ---
 
-**BREAKING CHANGE**: The `LifecycleMiddlewareOptions.startupRequestPauseTimeout` has been removed. Use the `backend.lifecycle.startupRequestPauseTimeout` setting in your `app-config.yaml` file to customize how the `createLifecycleMiddleware` function should behave. Also the root config service is required as an options when calling the `createLifecycleMiddleware` function:
+**BREAKING PRODUCERS**: The `LifecycleMiddlewareOptions.startupRequestPauseTimeout` has been removed. Use the `backend.lifecycle.startupRequestPauseTimeout` setting in your `app-config.yaml` file to customize how the `createLifecycleMiddleware` function should behave. Also the root config service is required as an option when calling the `createLifecycleMiddleware` function:
 
 ```diff
 - createLifecycleMiddleware({ lifecycle, startupRequestPauseTimeout })
--- a/.changeset/fast-tools-fix.md
+++ b/.changeset/fast-tools-fix.md
@@ -1,0 +1,10 @@
+---
+'@backstage/backend-defaults': minor
+---
+
+**BREAKING CHANGE**: The `LifecycleMiddlewareOptions.startupRequestPauseTimeout` has been removed. Use the `backend.lifecycle.startupRequestPauseTimeout` setting in your `app-config.yaml` file to customize how the `createLifecycleMiddleware` function should behave. Also the root config service is required as an options when calling the `createLifecycleMiddleware` function:
+
+```diff
+- createLifecycleMiddleware({ lifecycle, startupRequestPauseTimeout })
++ createLifecycleMiddleware({ config,  lifecycle })
+```
--- a/packages/backend-defaults/src/entrypoints/httpRouter/http/createLifecycleMiddleware.ts
+++ b/packages/backend-defaults/src/entrypoints/httpRouter/http/createLifecycleMiddleware.ts
@@ -14,7 +14,11 @@
  * limitations under the License.
  */
 
-import { LifecycleService } from '@backstage/backend-plugin-api';
+import {
+  RootConfigService,
+  LifecycleService,
+} from '@backstage/backend-plugin-api';
+import { readDurationFromConfig } from '@backstage/config';
 import { ServiceUnavailableError } from '@backstage/errors';
 import { HumanDuration, durationToMilliseconds } from '@backstage/types';
 import { RequestHandler } from 'express';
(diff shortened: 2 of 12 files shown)


packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
```diff
@@ -140,21 +125,34 @@ const rootHttpRouterServiceFactoryWithOptions = (
...
+          key: 'backend.lifecycle.serverShutdownDelay',
+        });
+      }
+
+      lifecycle.addBeforeShutdownHook(async () => {
+        const timeoutMs = durationToMilliseconds(serverShutdownDelay);
+        return await new Promise(resolve => {
+          setTimeout(resolve, timeoutMs);
+        });
+      });
```
Rugvip, line 147: While at it it feels like it makes sense to move this within the `if` block above and make `serverShutdownDelay` a `const`

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
```diff
@@ -130,7 +112,10 @@ const rootHttpRouterServiceFactoryWithOptions = (
         logger,
         lifecycle,
         healthRouter,
-        lifecycleMiddleware,
+        // TODO: delete this when the deprecated lifecycleMiddleware is removed
+        lifecycleMiddleware: (_req, _res, next) => {
+          next();
+        },
```
Rugvip, line 118: This was added in the previous PR right? It's not been released in mainline yet so we can remove it straight away and update the existing changeset

Rugvip (changes requested):
👍 

Let's update the existing changesets from the previous PR as well, so that we don't have those going out with incorrect instructions in the mainline release

packages/backend-defaults/src/entrypoints/httpRouter/httpRouterServiceFactory.ts
```diff
@@ -57,6 +59,7 @@ export const httpRouterServiceFactory = createServiceFactory({
     });
 
     router.use(createAuthIntegrationRouter({ auth }));
+    router.use(createLifecycleMiddleware({ lifecycle }));
```
Rugvip, line 62: This needs to be hooked up to the `startupRequestPauseTimeout` config right? I think it would be best to pass `config` in here and read it within the creator instead, same as the `createCredentialsBarrier` does

Rugvip (commented):
Nice! 👍 

Just minor config thing left I think

packages/backend-defaults/src/entrypoints/httpRouter/httpRouterServiceFactory.ts
camilaibs: Oh, right. Good catch! Just added it back.

packages/backend-defaults/src/entrypoints/httpRouter/httpRouterServiceFactory.ts
```diff
@@ -56,7 +60,17 @@ export const httpRouterServiceFactory = createServiceFactory({
       config,
     });
 
+    let startupRequestPauseTimeout: HumanDuration | undefined;
+    if (config.has('backend.lifecycle.startupRequestPauseTimeout')) {
+      startupRequestPauseTimeout = readDurationFromConfig(config, {
+        key: 'backend.lifecycle.startupRequestPauseTimeout',
+      });
+    }
```
Rugvip, line 68: We try to keep the factory function clean, let's move this config reading to happen within `createLifecycleMiddleware` instead

camilaibs: I just updated the logic. I had to change the middleware signature API, as it was exported from the backend defaults package.

packages/backend-defaults/src/entrypoints/httpRouter/http/createLifecycleMiddleware.ts
```diff
@@ -25,15 +29,20 @@ export const DEFAULT_TIMEOUT = { seconds: 5 };
...
-}
+export type LifecycleMiddlewareOptions =
+  | {
+      lifecycle: LifecycleService;
+      /**
+       * The maximum time that paused requests will wait for the service to start, before returning an error.
+       *
+       * Defaults to 5 seconds.
+       */
+      startupRequestPauseTimeout?: HumanDuration;
```
Rugvip, line 40: ah but I think we should remove this completely :sweat_smile:, there's not that much use to being able to set this via code is there?

camilaibs: It is a breaking change, right 🤔 ?

.changeset/fast-tools-fix.md
```diff
@@ -0,0 +1,10 @@
+---
+'@backstage/backend-defaults': minor
+---
+
+**BREAKING CHANGE**: The `LifecycleMiddlewareOptions.startupRequestPauseTimeout` has been removed. Use the `backend.lifecycle.startupRequestPauseTimeout` setting in your `app-config.yaml` file to customize how the `createLifecycleMiddleware` function should behave. Also the root config service is required as an options when calling the `createLifecycleMiddleware` function:
```
Rugvip, line 5:
```suggestion
**BREAKING PRODUCERS**: The `LifecycleMiddlewareOptions.startupRequestPauseTimeout` has been removed. Use the `backend.lifecycle.startupRequestPauseTimeout` setting in your `app-config.yaml` file to customize how the `createLifecycleMiddleware` function should behave. Also the root config service is required as an options when calling the `createLifecycleMiddleware` function:
```

See https://backstage.io/docs/overview/versioning-policy#changes-that-are-not-considered-breaking

packages/backend-defaults/src/entrypoints/httpRouter/http/createLifecycleMiddleware.ts
```diff
@@ -81,6 +80,17 @@ export function createLifecycleMiddleware(
     waiting.clear();
   });
 
+  let startupRequestPauseTimeout: HumanDuration = DEFAULT_TIMEOUT;
+
+  if (
+    'config' in options &&
```
Rugvip, line 86: 🧹

Rugvip (approved):
Two small bits, then :shipit: 😁

PR #28120, opened 2024-12-12
stop autobumping of css-loader

freben:
is there something better to link to here as an explanation?

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -78,5 +78,10 @@
       matchPackageNames: ['yn'],
       allowedVersions: '<5.0.0',
     },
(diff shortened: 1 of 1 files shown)


benjdlambert:
Hmm, upon diffing further I wonder if we could actually use the `restore 6.x behaviour` block for now? 

https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md#-breaking-changes

maybe this would allow us to keep the existing asset types.

vinzscam (approved):

Rugvip (changes requested):
Yep I think bumping to latest with updated configuration is the way to go

PR #28121, opened 2024-12-12
catalog-backend: avoid JSON parsing when possible

Rugvip:

`JSON.parse` and `JSON.stringify` is a significant source of synchronous work in the catalog, which blocks the main thread and slows down the entire backend. This change avoids JSON deserialization and serialization in many situations when reading entities, especially for list endpoints.

### DIFF
--- a/plugins/catalog-backend/src/catalog/types.ts
+++ b/plugins/catalog-backend/src/catalog/types.ts
@@ -58,7 +58,7 @@ export type EntitiesRequest = {
  */
 export type EntitiesResponseItems =
   | {
-      type: 'objects';
+      type: 'object';
       entities: (Entity | null)[];
     }
   | {
--- a/plugins/catalog-backend/src/catalog/types.ts
+++ b/plugins/catalog-backend/src/catalog/types.ts
@@ -52,8 +52,22 @@ export type EntitiesRequest = {
   credentials: BackstageCredentials;
 };
 
+/**
+ * Encapsulates either a deserialized or serialized entities to be sent in a response.
+ * @internal
+ */
+export type EntitiesResponseItems =
+  | {
+      type: 'objects';
+      entities: (Entity | null)[];
+    }
+  | {
+      type: 'raw';
+      entities: (string | null)[];
+    };
+
 export type EntitiesResponse = {
-  entities: Entity[];
+  entities: EntitiesResponseItems;
   pageInfo: PageInfo;
 };
 
@@ -86,7 +100,7 @@ export interface EntitiesBatchResponse {
    * The list of entities, in the same order as the refs in the request. Entries
    * that are null signify that no entity existed with that ref.
    */
-  items: Array<Entity | null>;
+  items: EntitiesResponseItems;
 }
 
 export type EntityAncestryResponse = {
@@ -224,7 +238,7 @@ export interface QueryEntitiesResponse {
   /**
    * The entities for the current pagination request
    */
-  items: Entity[];
+  items: EntitiesResponseItems;
 
   pageInfo: {
     /**
--- a/plugins/catalog-backend/src/service/CatalogBuilder.ts
+++ b/plugins/catalog-backend/src/service/CatalogBuilder.ts
@@ -115,6 +115,7 @@ import {
   UrlReaderService,
   SchedulerService,
 } from '@backstage/backend-plugin-api';
+import { entitiesResponseToObjects } from './response';
 
 /**
  * This is a duplicate of the alpha `CatalogPermissionRule` type, for use in the stable API.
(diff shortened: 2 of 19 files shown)


plugins/catalog-backend/src/service/util.ts
```diff
@@ -148,75 +148,21 @@ export function decodeCursor(encodedCursor: string) {
...
+    for (const relation of entity.relations as any) {
+      if (!relation.targetRef && relation.target) {
+        // This is the case where an old-form entity, not yet stitched with
+        // the updated code, was in the database
+        relation.targetRef = stringifyEntityRef(relation.target);
+      } else if (!relation.target && relation.targetRef) {
+        // This is the case where a new-form entity, stitched with the
+        // updated code, was in the database but we still want to produce
+        // the old data shape as well for compatibility reasons
+        relation.target = parseEntityRef(relation.targetRef);
```
freben, line 163: Do you think that the sheer amount of `parseEntityRef` going on in one single linear job (when returning huge numbers of highly connected entities) could be a problem in and of itself, causing a large part of the blockage?

plugins/catalog-backend/src/service/response/write.ts
```diff
@@ -0,0 +1,95 @@
...
+    const needsDrain = !res.write(prefix + entity, 'utf8');
+    if (needsDrain) {
+      await new Promise<void>(resolve => {
+        const cont = () => {
+          res.off('drain', cont);
+          res.off('close', cont);
+          resolve();
+        };
+        res.on('drain', cont);
+        res.on('close', cont);
```
freben, line 90: as mentioned elsewhere - do we actually want to retain the old behavior of handling closure with throwing and aborting immediately?

plugins/catalog-backend/src/catalog/types.ts
```diff
@@ -52,8 +52,22 @@ export type EntitiesRequest = {
   credentials: BackstageCredentials;
 };
 
+/**
+ * Encapsulates either a deserialized or serialized entities to be sent in a response.
+ * @internal
+ */
+export type EntitiesResponseItems =
```
freben, line 59: this type is slightly wonky in that not all endpoints are able to respond with nulls etc - but maybe it's ok, or add a template param `<TNullable extends boolean = false>` and the corresponding changes inside, or whatever ...

plugins/catalog-backend/src/service/CatalogBuilder.ts
```diff
@@ -485,6 +486,10 @@ export class CatalogBuilder {
       discovery,
     });
 
+    const enableRawJson = config.getOptionalBoolean(
+      'catalog.enableRawJsonResponses',
```
freben, line 490: Since the end result for the caller is still the same old json, maybe the parameter name is a bit confusing. But I can't think of anything better really

freben (commented):
yeah it's a little bit sad that the dual representations and nullability permeated a bit much, but it is likely worth it

plugins/catalog-backend/src/catalog/types.ts
Rugvip: yep I had a param for a while but it doesn’t help too much. In practice we still need to support null and passing through non-nullable items in some places doesn't really do much more than avoiding some type fiddling in a few places, but requires more code in other

plugins/catalog-backend/src/service/CatalogBuilder.ts
Rugvip: yeah kinda the best I could think of :sweat_smile:, maybe something like `disableRelationsResponseCompatability` instead? that way it doesn't need as much explaining either

plugins/catalog-backend/src/service/util.ts
Rugvip: Hmm, maybe, let's leave that for now though

plugins/catalog-backend/src/service/response/write.ts
Rugvip: Refactored to properly close. Didn't bother figuring out a test for it at this point but did manual verification to make sure an interrupted request is handled correctly.

plugins/catalog-backend/src/catalog/types.ts
Rugvip: Eh, leaving this for a potential later internal refactor

plugins/catalog-backend/src/service/CatalogBuilder.ts
Rugvip: Switched to `catalog.disableRelationsCompatibility` instead.

benjdlambert (approved):

PR #28129, opened 2024-12-13
feat: add a new system metadata service

aramissennyeydd:

Experimental work for #25431. This PR adds a new `SystemMetadataService` and adds an HTTP API for accessing (and testing) both the instance metadata service and new system metadata service. I also changed the config for tracking instance URLs to a separate root key - I think this should help with the case of a user having `catalog.mysite.com` as their catalog plugin URL, we'd enforce that the instance URLs listed have `.backstage/*` URLs that are reachable.

It also adds an example of how to combine both of them to list all installed features across the entire system. This would likely end up being its own custom discovery plugin or similar. Happy to separate that out into a separate PR.

Open questions:
- [ ] What's the best way to host the HTTP routes for instance + system metadata service such that they can handle overrides? `rootHttpRouter`?

Example call to `/.backstage/systemMetadata/features/installed`
```
{
  "auth": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    }
  ],
  "app": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    }
  ],
  "catalog": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    },
    {
      "internalUrl": "http://localhost:7008/",
      "externalUrl": "http://localhost:7008/"
    }
  ],
  "events": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    }
  ],
  "devtools": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    }
  ],
  "kubernetes": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    }
  ],
  "permission": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    },
    {
      "internalUrl": "http://localhost:7008/",
      "externalUrl": "http://localhost:7008/"
    }
  ],
  "proxy": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    }
  ],
  "scaffolder": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    }
  ],
  "techdocs": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    }
  ],
  "signals": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    }
  ],
  "notifications": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    }
  ],
  "search": [
    {
      "internalUrl": "http://localhost:7007/",
      "externalUrl": "http://localhost:7007/"
    }
  ]
}
```


packages/backend-defaults/src/entrypoints/systemMetadata/systemMetadataServiceFactory.ts
```diff
@@ -0,0 +1,41 @@
...
+} from '@backstage/backend-plugin-api';
+import { DefaultSystemMetadataService } from './lib/DefaultSystemMetadataService';
+import { systemMetadataServiceRef } from '@backstage/backend-plugin-api/alpha';
+
+/**
+ * Metadata about an entire Backstage system, a collection of Backstage instances.
+ *
+ * @alpha
+ */
+export const systemMetadataServiceFactory = createServiceFactory({
```
freben, line 29: Tempting to bikeshed some name that doesn't collide with catalog kinds. Like `cluster`, `deployment`, etc.

packages/backend-plugin-api/report-alpha.api.md
```diff
@@ -46,5 +52,26 @@ export const instanceMetadataServiceRef: ServiceRef<
...
+
+// @alpha
+export const systemMetadataServiceRef: ServiceRef<
+  SystemMetadataService,
+  'plugin',
+  'singleton'
+>;
+
+// @alpha (undocumented)
+export type Target =
```
freben, line 69: This may warrant a longer more descriptive name like `InstanceDiscoveryUrl` or whatever.

packages/backend-defaults/src/entrypoints/systemMetadata/lib/DefaultSystemMetadataService.ts
```diff
@@ -0,0 +1,64 @@
...
+  public static create(pluginEnv: {
+    logger: LoggerService;
+    config: RootConfigService;
+  }) {
+    return new DefaultSystemMetadataService(pluginEnv);
+  }
+
+  listInstances() {
+    const endpoints =
+      this.config.getOptionalConfigArray('discovery.instances') ?? [];
```
freben, line 46: This needs a `config.d.ts` addition.

I'm sort of pondering what the purpose and scope of this method is. Is it "just a bunch of unordered urls that you may want to make some further calls to"? And is that clear enough? I guess we'll document that it's meant to be at the granularity not really of pods, but of the ingresses that front those pods. What I'm getting at, I think, is whether this service is meant to be able to serve as the underpinning of discovery (because all pods are returned here and discovery can round-robin return among those) or if it's the other way around that this service is higher level and might even depend on service discovery.

And I'm also wondering - do we need `.instances`? Should we look at `.endpoints` + `backend.baseUrl` instead? And then for the latter just "crawl" a number of steps up looking for the root level `.backstage` endpoint. From this we can build a union of roots.

packages/backend-plugin-api/report-alpha.api.md
```diff
@@ -18,6 +18,12 @@ export type BackendFeatureMeta =
       moduleId: string;
     };
 
+// @alpha (undocumented)
+export interface BackstageInstance {
+  // (undocumented)
+  url: Target;
```
freben, line 24: And this might be all of the metadata for the foreseeable future? It won't dig any deeper than that.

Maybe it'd be nice for consumers to have two members `internalBaseUrl` and `externalBaseUrl` (which may or may not be set to the same value)? At this level, the caller typically knows what they want to access and needing to do `typeof` checks as a consumer might be undesirable.

packages/backend-split/src/experimental/features.http
freben, line 1: What's this file?

packages/backend-split/src/experimental/instanceMetadata.ts
```diff
@@ -18,22 +18,32 @@ import {
...
             .join(', ')}`,
         );
+
+        const router = Router();
+
+        router.get('/features/installed', (_, res) => {
+          res.json({ items: instanceMetadata.getInstalledFeatures() });
+        });
+
+        httpRouter.use('/.backstage/instanceInfo', router);
```
freben, line 46:
```suggestion
        httpRouter.use('/.backstage/instanceMetadata/v1', router);
```

i think we started doing versioning for services. and probably makes sense to name like the service itself.

packages/backend-split/src/experimental/systemMetadata.ts
```diff
@@ -0,0 +1,93 @@
...
+            const instance = result.value.instance;
+            const installedFeatures = result.value.response
+              .items as BackendFeatureMeta[];
+            for (const feature of installedFeatures) {
+              if (feature.type === 'plugin') {
+                if (!pluginByInstance[feature.pluginId]) {
+                  pluginByInstance[feature.pluginId] = [];
+                }
+                pluginByInstance[feature.pluginId].push(
+                  `${instance.url}/api/${feature.pluginId}`,
```
freben, line 81: I think this one is a bit interesting, in the sense that a plugin doesn't necessarily have a rest api at all. Maybe that's not a concern at all and the caller just needs to figure it out.

packages/backend-split/src/experimental/systemMetadata.ts
```diff
@@ -0,0 +1,93 @@
...
+                );
+              }
+            }
+          }
+          res.json(pluginByInstance);
+        });
+
+        httpRouter.use('/.backstage/systemInfo', router);
```
freben, line 89:
```suggestion
        httpRouter.use('/.backstage/systemMetadata/v1', router);
```

packages/backend-split/src/experimental/systemMetadata.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        const router = Router();
+
+        router.get('/instances', async (_, res) => {
+          res.json(await systemMetadata.listInstances());
+        });
+
+        router.get('/features/installed', async (_, res) => {
+          const instances = await systemMetadata.listInstances();
+          const featurePromises = await Promise.allSettled(
```
freben, line 51: Is it safe enough to throw away failures?

Just wondering what the impact on callers will be if there's flakiness and features seem to pop in and out of existence. Maybe not a concern for now 🤔

packages/backend-split/src/index.ts
```diff
@@ -0,0 +1,37 @@
...
+backend.add(import('@backstage/plugin-catalog-backend'));
+
+backend.add(
+  import('@backstage/plugin-permission-backend-module-allow-all-policy'),
+);
+backend.add(import('@backstage/plugin-permission-backend'));
+
+backend.add(import('./experimental/instanceMetadata'));
+backend.add(import('./experimental/systemMetadata'));
+backend.add(systemMetadataServiceFactory);
```
freben, line 35: in the end, we don't want to have to install this one by hand of course

packages/backend-split/src/experimental/systemMetadata.ts
```diff
@@ -0,0 +1,93 @@
...
+} from '@backstage/backend-plugin-api';
+import {
+  BackendFeatureMeta,
+  systemMetadataServiceRef,
+} from '@backstage/backend-plugin-api/alpha';
+import Router from 'express-promise-router';
+
+// Example usage of the instance metadata service to log the list of instances and a small HTTP API.
+export default createBackendPlugin({
+  pluginId: 'system-metadata',
```
freben, line 28: Maybe a stupid question - can't the service depend on `rootHttpRouter` and register this itself instead of making a wrapper plugin necessary?

freben (commented):
Alright, a bunch of thoughts :)

packages/backend-defaults/src/entrypoints/systemMetadata/lib/DefaultSystemMetadataService.ts
```diff
@@ -0,0 +1,64 @@
...
+      } else {
+        this.logger.warn(
+          `Instance ${endpoint.get(
+            'target',
+          )} is missing a 'baseUrl' property. This is required for the system metadata service.`,
+        );
+      }
+    }
+    this.logger.info(`Found ${instances.length} instances.`);
+    return Promise.resolve(instances);
```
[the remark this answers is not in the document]
aramissennyeydd: completely :)

packages/backend-defaults/report-alpha.api.md
```diff
@@ -0,0 +1,32 @@
...
+    config: RootConfigService;
+  }): DefaultSystemMetadataService;
+  // (undocumented)
+  listInstances(): Promise<BackstageInstance[]>;
+}
+
+// @alpha
+export const systemMetadataServiceFactory: ServiceFactory<
+  SystemMetadataService,
+  'plugin',
```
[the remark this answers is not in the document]
aramissennyeydd: This definitely doesn't make sense, had forgotten this was the default scope - updated to root :)

packages/backend-plugin-api/report-alpha.api.md
aramissennyeydd: I think so for now at least - once we have the URL, we can talk directly to the instance in question and get health/instance/any other information through its REST API.

packages/backend-split/src/experimental/systemMetadata.ts
aramissennyeydd: Yeah, my goal here is to return a mapping of plugin -> instance w/ plugin installed for usage later on. Maybe being explicit with that makes more sense, and we don't need to prepare the plugin URLs here.

packages/backend-split/src/experimental/systemMetadata.ts
aramissennyeydd: I think we can skip this for now - long term, my thinking is that this should be a background job that tries to fill a cache of some kind that we can use across instances.

packages/backend-split/src/index.ts
aramissennyeydd: just while we're experimental 😁

packages/backend-split/src/experimental/systemMetadata.ts
aramissennyeydd: Yes, it can - I think I was causing issues for myself with the non-`root` scoping. Let me see if I can squeeze this into the service.

aramissennyeydd: Hmm, do we want to add this to `rootHttpRouter` instead? Pushing an implementation where the router is hosted from the service directly.

packages/backend-split/src/experimental/features.http
```diff
@@ -0,0 +1,7 @@
+// Test requests for the split backend to test the system metadata service :)
+// The 2 below requests should return the same data.
+GET http://localhost:7007/.backstage/systemMetadata/v1/features/installed
+
+###
+
+GET http://localhost:7008/.backstage/systemMetadata/v1/features/installed
```
aramissennyeydd, line 7: Going to work on test cases to replace this file.

packages/backend-plugin-api/report-alpha.api.md
aramissennyeydd: Removed in favor of just sticking this on the Backstage instance definition below, I agree with your point below that having to call `typeof` when consuming this is ugly.

packages/backend-defaults/src/entrypoints/systemMetadata/lib/DefaultSystemMetadataService.ts
aramissennyeydd: Will update - my end goal with this is to have discovery use this service for where to route to. By default, we'd route to local plugins, and then figure out a heuristic for further down range ones. That is probably worth a separate BEP though 😁.

I think we do, the prickly case that I wasn't able to solve is the `catalog.my-company.com` that doesn't expose any other routes to public. We can't really crawl for the instanceInfo URL because they likely are routing from `catalog.my-company.com` -> `<my-pod>/api/catalog` directly. My attempt here is to move that format to truly custom and make it more standard to host 1 pod to 1 URL to allow for better routability. Consumers can always override the default systemMetadataService/discovery implementation to handle their custom work.

packages/backend-defaults/src/entrypoints/systemMetadata/systemMetadataServiceFactory.ts
aramissennyeydd: Any ideas are very welcome, I think I started at deployment, but it had overlap with k8s which might be confusing.

freben:
@aramissennyeydd note the reference to https://github.com/backstage/backstage/pull/28621 above which i found might be interesting here later down the line

freben:
bah, misclick

aramissennyeydd:
👋 This should be ready for another review, I fixed the test cases and added a proper test for the instance scraping logic.

packages/backend-defaults/report-alpha.api.md
```diff
@@ -0,0 +1,34 @@
...
+  constructor(options: { logger: LoggerService; config: RootConfigService });
+  // (undocumented)
+  static create(pluginEnv: {
+    logger: LoggerService;
+    config: RootConfigService;
+  }): DefaultSystemMetadataService;
+  // (undocumented)
+  introspect(): Promise<{
+    instances: BackstageInstance[];
+  }>;
```
aramissennyeydd, line 23: @Rugvip Thoughts on this?

freben:
Hm it seems a new test failure crept in.

aramissennyeydd:
@freben Fixed! Was the same error we were chatting about last week

freben:
ah but needs rebase 😅

aramissennyeydd:
@freben Missed your comment 😅  - conflicts resolved now!

packages/backend-defaults/src/entrypoints/systemMetadata/lib/createSystemMetadataRouter.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  logger.info(
+    `Instances in this system: ${JSON.stringify(await getInstances())}`,
+  );
+
+  const router = Router();
+
+  router.get('/instances', async (_, res) => {
+    res.json(await getInstances());
+  });
```
Rugvip: Never return arrays in responses ~imo~, could keep it wrapped in `instances` as it is, or list it in `items`.

packages/backend-plugin-api/report.api.md
```diff
@@ -721,6 +730,14 @@ export interface ServiceRefOptions<
   scope?: TScope;
 }
 
+// @public (undocumented)
+export interface SystemMetadataService {
+  // (undocumented)
+  introspect(): Promise<{
+    instances: BackstageInstance[];
+  }>;
```
Rugvip: I'm a bit worried about how much of a snapshot this API is. For the instance metadata I feel it's fine because it's not mutable anyway, but for system I think it might be best to build it around it being an observable state instead? Thinking that it's a bit too simple to end up in a place where you need to run around and restart instances for them to pick up the latest meta to properly enable features. Then again it does make things a fair bit more complex ofc, but I'm thinking that might be needed for things to behave correctly.

packages/backend-test-utils/src/services/mockServices.ts
```diff
@@ -538,4 +539,20 @@ export namespace mockServices {
       subscribe: jest.fn(),
     }));
   }
+
+  export namespace systemMetadata {
+    /**
+     * Creates a functional mock factory for the
+     * {@link @backstage/backend-plugin-api#coreServices.systemMetadata}.
+     */
+    export const factory = () => systemMetadataServiceFactory;
```
Rugvip: Thinking it might be best to have a proper mock implementation here that doesn't attempt network calls? Ideally it's even configurable to be able to try different discovered features, but can leave that as a future addition too.

packages/backend-plugin-api/src/services/definitions/coreServices.ts
```diff
@@ -277,4 +277,15 @@ export namespace coreServices {
   export const urlReader = createServiceRef<
     import('./UrlReaderService').UrlReaderService
   >({ id: 'core.urlReader' });
+
+  /**
+   * Read information about your current Backstage deployment.
+   * @public
```
Rugvip, line 284: Happy to move the instance to stable as well, but either way we'd want to keep them in sync

packages/backend-plugin-api/report.api.md
aramissennyeydd: @Rugvip Makes sense, do you see each service-level thing being individually subscribe-able or should be it just be the entire snapshot? ie `.instances`, `.features`, `.whatnot` or should they all sit in a single object?

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -149,8 +150,31 @@ function createInstanceMetadataServiceFactory(
...
+          .join(', ')}`,
+      );
+
+      const router = Router();
+
+      router.get('/features/installed', (_, res) => {
+        res.json({ items: instanceMetadata.getInstalledFeatures() });
+      });
+
+      httpRouter.use('/.backstage/instanceMetadata/v1', router);
```
Rugvip: Had the thought that maybe this should be just `metadata`, since the fact that we're calling an API kind of implies that we're talking to an instance anyway. I don't dislike the current naming either, just wonder if there are any other similar thoughts.

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -149,8 +150,31 @@ function createInstanceMetadataServiceFactory(
...
+        `Installed features on this instance: ${instanceMetadata
+          .getInstalledFeatures()
+          .join(', ')}`,
+      );
+
+      const router = Router();
+
+      router.get('/features/installed', (_, res) => {
+        res.json({ items: instanceMetadata.getInstalledFeatures() });
+      });
```
Rugvip: Random though: I think it might be useful to also include the timestamp of when the instance was started, to help resolve conflicts in the metadata. That in turn made me think that maybe this is best designed as a single "fat" one-stop-shop endpoint where we put anything one might want to know. Just `/info` or something.

packages/backend-defaults/config.d.ts
```diff
@@ -1164,5 +1164,15 @@ export interface Config {
        */
       plugins: string[];
     }>;
+
+    /**
+     * A list of deployed Backstage instances that can be crawled for discovery.
```
Rugvip: Not immediately obvious to me: is this supposed to be individual instances, i.e. pods? or service name pointing to multiple instances of the same class, i.e. deployments. That question also got me thinking about whether there's value in supporting custom DNS lookups so that we can resolve all A/AAAA/SRV records and reach each individual instance for metadata collection. For the purpose of detecting if a feature is supported I would for example expect that all pods in a deployment report that they support the feature.

packages/backend-plugin-api/src/services/definitions/SystemMetadataService.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface BackstageInstance {
+  internalUrl: string;
+  externalUrl: string;
+}
+
+/**
+ * @public
+ */
+export interface SystemMetadataService {
+  instances(): Observable<BackstageInstance[]>;
```
Rugvip: Hmm, not sure about observable, especially for backend. Another option is `AsyncIterator`, similar to

https://github.com/backstage/backstage/blob/72d019d663cbc19035e82ebecf74b207faacab26/packages/config-loader/src/sources/types.ts#L77

It sucks to produce, but is nice to consume.

Honestly a simple callback might do it in this case though? We're not looking for reporting either completions or errors anyway?

Curious to here why you chose this though. Honestly the main drawback of observables for me is that most people can't help themselvs and end up writing way to complicated code 😅

packages/backend-test-utils/src/services/simpleMock.ts
```diff
@@ -39,6 +39,7 @@ export function simpleMock<TService>(
     const mock = mockFactory();
     if (partialImpl) {
       for (const [key, impl] of Object.entries(partialImpl)) {
+        console.log(key, impl, mock);
```
Rugvip: 👀

packages/backend-plugin-api/src/services/definitions/SystemMetadataService.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { Observable } from '@backstage/types';
+
+/**
+ * @public
+ */
+export interface BackstageInstance {
+  internalUrl: string;
+  externalUrl: string;
+}
```
Rugvip: I tried using this over in https://github.com/backstage/backstage/pull/31049, couple of thoughts:

The list of instances and just getting their URLs is awkward from a plugin perspective. We'd need a service built on top of this for feature-discovery like use-cases. It does bring me back to the BEP though and raises questions about the value of the intermediate implementation ._. As in is there enough value in these additions by themselves or do we need to go further?

It's possible that the observable route is a bit redundant as well, I couldn't find any place where it could reasonable act the way we'd expect with an observable or similar. Any consumption of other features is generally directly in request paths or scheduled. We probably need to refine this interface a bit more before knowing for sure though.

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -149,8 +150,31 @@ function createInstanceMetadataServiceFactory(
...
+      logger.info(
+        `Installed features on this instance: ${instanceMetadata
+          .getInstalledFeatures()
+          .join(', ')}`,
+      );
+
+      const router = Router();
+
+      router.get('/features/installed', (_, res) => {
+        res.json({ items: instanceMetadata.getInstalledFeatures() });
```
Rugvip: Bit of a sidetrack: I think a better shape for listing installed features is probably this:

```ts
getInstalledPlugins(): Array<{
  pluginId: string
  modules: Array<{
    moduleId: string
  }>
}>
```

Couple of reasons:

- `getInstalledFeatures` is lying a bit, it doesn't list service factories, which are technically features
- Wherever I've used the service so far, I've always ended up with a `.filter(f => f.type === 'plugin')...`
- This is closer to the real structure of backend features, modules belong to plugins

Think we could consider ^ for promoting it to a stable core service.

In general I could see stabilizing the instance metadata service and making that information available on an endpoint as a potentially good next incremental step too. I think that might make it easier to ship some things and start experimenting/building with them, which has already happened a bit with the instance metadata.

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -149,8 +150,31 @@ function createInstanceMetadataServiceFactory(
...
+
+      logger.info(
+        `Installed features on this instance: ${instanceMetadata
+          .getInstalledFeatures()
+          .join(', ')}`,
+      );
+
+      const router = Router();
+
+      router.get('/features/installed', (_, res) => {
```
Rugvip: Starting to get a bit worried about the security of this, especially if we start adding more information. There are all public and unauthenticated atm? Even with something like the gateway plugin the main instance will still have this exposed. I think we may need to move these to a separate port so that we can add information here a bit more freely.

packages/backend-defaults/config.d.ts
aramissennyeydd: Yeah, it's ambiguous - with the current implementation, it's based on regular URLs from config which means we don't enumerate all pods, only the one we hit from the load balancer. I think URLs are the right level here, because as you say all pods in a deployment should report the same data, but it also means that our instance metadata collection could be off if that isn't true.

packages/backend-plugin-api/src/services/definitions/SystemMetadataService.ts
aramissennyeydd: I wanted to stick with a subscribe-based interface, a callback is likely the right in between here 😅 - we just want a way to get future data about updates.

packages/backend-app-api/src/wiring/BackendInitializer.ts
aramissennyeydd: a separate port would make routing a lot more difficult as we couldn't leverage existing URLs from discovery as easily - maybe a good thing but would make adopting this harder?

we could make these authenticated + toggleable to start? so for a gateway deployment you could turn this off?

packages/backend-app-api/src/wiring/BackendInitializer.ts
aramissennyeydd: Fair, and external clients likely don't care which service factories you have installed - I'll try and pull this out.

aramissennyeydd:
Chatted with @Rugvip on this, going to update this to match the following
1. Update this PR to _just_ provide plugins defined in discovery config. Things like `*` will not be supported.
 - could provide a way to define plugins alongside a wildstar so that they are picked up.
2. Migrate OpenAPI schema discovery to this new system.
3. Followup PR for dynamic discovery that would do the heavy lifting of dynamic system discovery,
  - opt in, on a separate port to protect against misconfiguration
  - authentication would still need to be figured out, should we move discovery out of the pluginAuth system and just use pluginAuth? (chicken and egg-y) would allow us to just say add another instance of type `catalog` and use the existing certificates/authentication

packages/backend-defaults/package.json
```diff
@@ -214,6 +214,7 @@
     "@types/yauzl": "^2.10.0",
     "aws-sdk-client-mock": "^4.0.0",
     "better-sqlite3": "^12.0.0",
+    "get-port": "^5.1.1",
```
Rugvip, line 217:
```suggestion
```

Don't see this being used anywhere? remnant from an old version? Will try to suggest this away :p

yarn.lock
```diff
@@ -2915,6 +2915,7 @@ __metadata:
     express-promise-router: "npm:^4.1.0"
     express-rate-limit: "npm:^7.5.0"
     fs-extra: "npm:^11.2.0"
+    get-port: "npm:^5.1.1"
```
Rugvip, line 2918:
```suggestion
```

.changeset/old-cats-shake.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/backend-defaults': minor
```
Rugvip:
```suggestion
'@backstage/backend-defaults': patch
```

This one is actually still on 0.x, and no breaking change

Rugvip (approved):
Nice, thank you! Let's :shipit:, just a few small fixes

packages/backend-defaults/package.json
aramissennyeydd: the power of port 0!

PR #28133, opened 2024-12-13
catalog-backend: fix for not waiting for response buffer to drain as needed

Rugvip:
Followup fix for #28121. I think the Node.js docs tricked me a bit here, claiming that additional writes would be buffered internally. Turns out that's not at all what happens and it seems the rest of the writes get lost instead until waiting for darin.


benjdlambert (approved):
:chefskiss:

freben (approved):
Only worry I have is, does close really fire AFTER the actual closure, or do we have to subscribe unconditionally once to that event before attempting to send anything?

Rugvip:
@freben added https://github.com/backstage/backstage/pull/28133/commits/224bcc88377e999eb09e057db2f6ba37acc9db86 as an extra check to make sure we don't wait for a `'drain'` that will never be emitted and leak.

PR #28136, opened 2024-12-13
Canon - Text + Heading components

cdedreuille:

### DIFF
--- a/packages/canon/src/components/Heading/Heading.tsx
+++ b/packages/canon/src/components/Heading/Heading.tsx
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { forwardRef } from 'react';
+import { HeadingProps } from './types';
+import { useTheme } from '../../theme/context';
+import { getResponsiveValue } from '../../utils/getResponsiveValue';
+
+export const Heading = forwardRef<HTMLParagraphElement, HeadingProps>(
+  (props, ref) => {
+    const { children, variant = 'title1', ...restProps } = props;
+    const { breakpoint } = useTheme();
+    const responsiveVariant = getResponsiveValue(variant, breakpoint);
+
+    console.log(breakpoint);
+    console.log(responsiveVariant);
+
+    return (
+      <p
+        ref={ref}
+        {...restProps}
+        className={`text ${
+          responsiveVariant ? `text-${responsiveVariant}` : ''
+        }`}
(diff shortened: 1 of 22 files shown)


packages/canon/docs/components/Text/styles.css
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-.text {
+.sb-text {
```
Rugvip, line 17: we're gonna need the inverse solution to this, as in we can't require these plain class names to be unused

packages/canon/src/components/Heading/Heading.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { getResponsiveValue } from '../../utils/getResponsiveValue';
+
+export const Heading = forwardRef<HTMLParagraphElement, HeadingProps>(
+  (props, ref) => {
+    const { children, variant = 'title1', ...restProps } = props;
+    const { breakpoint } = useTheme();
+    const responsiveVariant = getResponsiveValue(variant, breakpoint);
+
+    console.log(breakpoint);
+    console.log(responsiveVariant);
```
Rugvip, line 29: :eyes:

packages/canon/src/components/Heading/Heading.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    return (
+      <p
```
Rugvip, line 32: I'd kind of expected matching h* elements here. Correct levels here is gonna be important for accessibility right?

packages/canon/src/theme/context.tsx
```diff
@@ -17,28 +17,59 @@
...
+    if (isBreakpointXl) return 'xl';
+    if (isBreakpointLg) return 'lg';
+    if (isBreakpointMd) return 'md';
+    if (isBreakpointSm) return 'sm';
+    return 'xs';
+  })();
+
   return (
-    <ThemeContext.Provider value={{ icons: combinedIcons }}>
+    <ThemeContext.Provider value={{ icons: combinedIcons, breakpoint }}>
```
Rugvip, line 72: hmm, not sure we want the theme provider to contain this kind of state, are there any other options here?

Rugvip (commented):
quite a bit of duplication here so I'll review again when other things are merged, couple of comments though

packages/canon/docs/components/Text/styles.css
cdedreuille: Yes. I'm planning to do this for all components in a separate PR.

packages/canon/src/components/Heading/Heading.tsx
cdedreuille: This is now cleaned

packages/canon/src/components/Heading/Heading.tsx
cdedreuille: This is now fixed. By default it will follow the variant to assign the right tag but if you want something specific for accessibility reasons you can use the `as` prop.

packages/canon/src/theme/context.tsx
cdedreuille: Breakpoints is quite a big part of the process across the system to allow for responsive values like `variant={{ xs: 'small', md:'medium' }}`. The idea is to have a set list of keys but to allow for users to update them if they want to.

Why do you think we should not have it in the provider?

Rugvip: I'm worried about the complexity of having this provider be stateful with a lot of built-in logic. Can leave as is for now an keep an eye on it

Rugvip (approved):
:shipit:

PR #28140, opened 2024-12-13
Update CodeOwnersProcessor helper

isabeltomb:

We noticed a component from our org that was imported from github ended up as an unprocessed entity due to the error: `Processor CodeOwnersProcessor threw an error while preprocessing; caused by TypeError: Cannot read properties of undefined (reading 'match')`

Screenshot 2024-12-02 at 9 27 57 AM

Debugging in the backstage demo app revealed that the error was happening in one of the helper functions for the processor, a bug in a conditional check when the function is called allowed for `undefined` to be passed into the function in certain circumstances causing the function to throw an error.
 
With the existing code:
`return match ? normalizeCodeOwner(match.owners[0]) : undefined;`
the type of the `match` object is:
```
export interface CodeOwnersEntry {
	pattern: string
	owners: Array<string>
}
```
the ternary operator is only checking if the `match` object exists and then calling `normalizeCodeOwner()` on the first item in `match.owners` however `match` can exist with an empty `owners` array which is what will cause the error that we're seeing. In order to fix this the condition in the ternary operator can be updated to add a check that will make sure that `match.owner` isn't empty.

The stack trace on the unprocessed entity object in our database also shows that the error is happening in `normalizeCodeOwners`:

`TypeError: Cannot read properties of undefined (reading 'match')\n    at normalizeCodeOwner...`

### DIFF
--- a/.changeset/twenty-laws-tie.md
+++ b/.changeset/twenty-laws-tie.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-backend': patch
 ---
 
-Updated condition in resolveCodeOwner to fix a bug where normalizeCodeOwner could potentially be called with an invalid arg causing an error in CodeOwnersProcessor.
+Updated condition in `resolveCodeOwner` to fix a bug where `normalizeCodeOwner` could potentially be called with an invalid argument causing an error in `CodeOwnersProcessor`
--- a/.changeset/twenty-laws-tie.md
+++ b/.changeset/twenty-laws-tie.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/twenty-laws-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Updated condition in resolveCodeOwner to fix a bug where normalizeCodeOwner could potentially be called with an invalid arg causing an error in CodeOwnersProcessor.
```
benjdlambert, line 5:
```suggestion
Updated condition in `resolveCodeOwner` to fix a bug where `normalizeCodeOwner` could potentially be called with an invalid argument causing an error in `CodeOwnersProcessor`
```

benjdlambert (changes requested):
Nice, thanks for this! Just a small changeset thing to fix :tada:

benjdlambert (approved):

benjdlambert:
Thanks! :tada:

PR #28141, opened 2024-12-13
Update provider.md

jolies93:
Corrected the Callback URL since we encountered this when implementing.  Also removed the scopes config item - including this resulted in errors for us.

I had submitted a PR before, and I addressed two feedback items - one was wording and the other was a request to include the optional configurations in the instructions.  Both are addressed.  Apologize I had to make a new PR because I had issues with the other being open so long and I didn't rebase it correctly - better to start fresh.

Changes to documentation to include configuration corrections

### DIFF
--- a/docs/auth/atlassian/provider.md
+++ b/docs/auth/atlassian/provider.md
@@ -28,7 +28,7 @@ Name your integration and click on the `Create` button.
 
 Settings for local development:
 
-- Callback URL: `http://localhost:7007/api/auth/atlassian`
+- Callback URL: `http://localhost:7007/api/auth/atlassian/handler/frame`
 - Use rotating refresh tokens
 - For permissions, you **must** enable `View user profile` for the currently
   logged-in user, under `User identity API`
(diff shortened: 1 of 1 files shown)


Parsifal-M (commented):
Hey! 👋 

I think it looks good and its clear, I do not however use this provider...

Just out of curiosity which was the original PR? Just for some added context 😅

jolies93:
> Hey! 👋
> 
> I think it looks good and its clear, I do not however use this provider...
> 
> Just out of curiosity which was the original PR? Just for some added context 😅

Hello and thanks for the review - the original PR is here (closed): https://github.com/backstage/backstage/pull/28012

I had moved the definitions and descriptions for those elements from being on the same line as the example to below so that it matched the general format - I figured perhaps that's what prettier didn't like.

awanlin (approved):
Thanks for the follow up on this @jolies93 🚀

awanlin (changes requested):
Ah, looks like Prettier isn't happy, can you take a look at that, please?

awanlin (approved):

PR #28142, opened 2024-12-15
only emulate entities with by-query if `disableRelationsCompatibility` is set

freben:
This changes the optimization to become an opt-in, giving it at least another release cycle to mature before it becomes default.

Intentionally no changeset; piggybacks on #28121


Rugvip (approved):
Can remove the `disableRelationsCompatibility` check further down now as well

freben:
Ah yeah. Updated that bit and improved the tests now.

PR #28143, opened 2024-12-15
log `BackendFeatureMeta` nicer

freben:
before:

Screenshot 2024-12-15 at 21 48 14

after:

Screenshot 2024-12-15 at 21 56 29


aramissennyeydd (approved):
🙇

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -111,13 +111,33 @@ function createInstanceMetadataServiceFactory(
...
-              return { type: 'plugin', pluginId: feature.pluginId };
+              return Object.defineProperty(
+                {
+                  type: 'plugin',
+                  pluginId: feature.pluginId,
+                },
+                'toString',
+                {
+                  enumerable: false,
+                  configurable: false,
```
Rugvip, line 122: Let's leave it configurable imo, little reason to lock that down here imo.

Rugvip (approved):

packages/backend-app-api/src/wiring/BackendInitializer.ts
freben: Alright, updated!

PR #28145, opened 2024-12-15
Canon - Document + improve responsive behaviour

cdedreuille:

- Added a new page for responsive design
- Move responsive function to the provider
- Improve responsive mechanism for `Heading` and `Text`
- Improve responsive mechanism for `Button`
- Improve docs for `Button`
- Renamed `ThemeProvider` into `CanonProvider`
- Rename classes for Button to include the `cn` prefix

### DIFF
--- a/packages/canon/src/components/Button/Button.tsx
+++ b/packages/canon/src/components/Button/Button.tsx
@@ -16,21 +16,7 @@
 
 import React, { forwardRef } from 'react';
 import { Icon } from '../Icon';
-import type { IconNames } from '../Icon/types';
-
-/**
- * Properties for {@link Button}
- *
- * @public
- */
-export interface ButtonProps {
-  size?: 'small' | 'medium';
-  variant?: 'primary' | 'secondary' | 'tertiary';
-  children: React.ReactNode;
-  disabled?: boolean;
-  iconStart?: IconNames;
-  iconEnd?: IconNames;
-}
+import { ButtonProps } from './types';
 
 /** @public */
 export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
--- a/packages/canon/src/components/Button/Button.tsx
+++ b/packages/canon/src/components/Button/Button.tsx
@@ -17,6 +17,7 @@
 import React, { forwardRef } from 'react';
 import { Icon } from '../Icon';
 import { ButtonProps } from './types';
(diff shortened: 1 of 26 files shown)


packages/canon/src/components/Button/styles.css
```diff
@@ -14,69 +14,70 @@
...
+    padding-right: 8px;
+    height: 40px;
+  }
 
-.button-content {
-  display: flex;
-  align-items: center;
-  gap: var(--canon-spacing-xs);
-}
+  & .cn-button-content {
```
Rugvip, line 73: Thinking it may be best to have this be standalone so that it's possible to override with a simple `.cn-button-content { ... }`  rather than needing to include `.cn-button` too. In general I'd say it's very important to form some strict rules around how specific we get with selectors, in order to make it simple to override styles if we go with the CSS approach

packages/canon/src/components/Button/Button.tsx
```diff
@@ -44,15 +31,28 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
+          `cn-button-${responsiveSize}`,
+          `cn-button-${responsiveVariant}`,
+        ].join(' ')}
       >
         <span
-          className={['button-content', iconStart && iconEnd ? 'icon-both' : '']
+          className={[
+            'cn-button-content',
+            iconStart && iconEnd ? 'cn-button-content-icon-both' : '',
+          ]
```
Rugvip, line 55: Gonna need to introduce more explicit separators here tbh, or I think we're gonna end up with naming clashes quite soon. Stuff like `cn-button__content` for example

Rugvip (approved):
👍

packages/canon/src/components/Button/styles.css
cdedreuille: Oh you're right. Making sure these styles are flat. I'll do the update.

packages/canon/src/components/Button/Button.tsx
cdedreuille: Yes let's validate together the naming convention and I can do a separate PR for all components.

packages/canon/src/components/Button/styles.css
cdedreuille: Ok this is fixed now.

PR #28171, opened 2024-12-17
emit database connection pool metrics

freben:
Experimenting with a variation of https://github.com/backstage/backstage/pull/28113 to discuss among the maintainers.

This results in metrics on the form

```
# HELP backend_database_pool_connections_count Number of database pool connections
# TYPE backend_database_pool_connections_count gauge
backend_database_pool_connections_count{plugin="auth",state="used"} 0
backend_database_pool_connections_count{plugin="auth",state="free"} 2
backend_database_pool_connections_count{plugin="auth",state="pending"} 0
backend_database_pool_connections_count{plugin="app",state="used"} 0
backend_database_pool_connections_count{plugin="app",state="free"} 1
backend_database_pool_connections_count{plugin="app",state="pending"} 0
backend_database_pool_connections_count{plugin="catalog",state="used"} 0
backend_database_pool_connections_count{plugin="catalog",state="free"} 4
backend_database_pool_connections_count{plugin="catalog",state="pending"} 0
...
```

### DIFF
--- a/packages/backend-defaults/src/entrypoints/database/DatabaseManager.ts
+++ b/packages/backend-defaults/src/entrypoints/database/DatabaseManager.ts
@@ -53,8 +53,6 @@ export type DatabaseManagerOptions = {
  * Testable implementation class for {@link DatabaseManager} below.
  */
 export class DatabaseManagerImpl {
-  private didRegisterMetrics: boolean = false;
-
   constructor(
     private readonly config: Config,
     private readonly connectors: Record<string, Connector>,
@@ -72,6 +70,8 @@ export class DatabaseManagerImpl {
         await this.shutdown({ logger: options.rootLogger });
       });
     }
+
+    this.registerMetrics();
   }
 
   /**
@@ -188,7 +188,6 @@ export class DatabaseManagerImpl {
       );
     }
 
-    this.registerMetricsOnFirstConnection();
     return clientPromise;
   }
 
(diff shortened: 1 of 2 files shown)


packages/backend-defaults/src/entrypoints/database/DatabaseManager.ts
```diff
@@ -185,6 +188,7 @@ export class DatabaseManagerImpl {
       );
     }
 
+    this.registerMetricsOnFirstConnection();
```
benjdlambert, line 191: Why couldn't this just go in the constructor? Are we trying to be lazy about things and not overreport? Just a little unfortunate to have to have this extra state check to see if the metrics have been registered.

freben: Yeah that was a remnant from an earlier iteration where it made more sense to make it lazy but here there's no reason to. Updated!

freben:
For transparency:

We've been discussing this and are hesitant to move forward with the current approach. Pool events are going to happen at a frequency that's far higher than the metrics scraping frequency. If they are scraped every 30 seconds, a gauge won't capture that the pool was entirely saturated for 29 seconds between samples. A meter isn't quite right either, since if no changes happen, it's not quite the type of application that a meter is meant for. So maybe a completely different approach of in-memory accounting of the number of microseconds of saturation is better, or so. Needs some investigation

PR #28172, opened 2024-12-17
Canon - Update Base UI dependency ✨

cdedreuille:
Base UI is finally out ✨ We are updating the package to their new name. We can't wait to continue this path forward.

- Update Base UI to the new package
- Update Storybook
- Move `vanilla-extract/sprinkles` as dev dependency


Rugvip (approved):
🎉

PR #28175, opened 2024-12-17
Canon - Field + Input components

cdedreuille:

### DIFF
--- a/packages/canon/src/components/Input/Input.tsx
+++ b/packages/canon/src/components/Input/Input.tsx
@@ -13,16 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import React from 'react';
 
-/* Components */
-@import '../components/Button/styles.css';
-@import '../components/Stack/styles.css';
-@import '../components/Inline/styles.css';
-@import '../components/Grid/styles.css';
-@import '../components/Container/styles.css';
-@import '../components/Icon/styles.css';
-@import '../components/Checkbox/styles.css';
-@import '../components/Table/styles.css';
-@import '../components/Text/styles.css';
-@import '../components/Heading/styles.css';
-@import '../components/Field/styles.css';
+import { Input as InputPrimitive } from '@base-ui-components/react/input';
+import clsx from 'clsx';
+
+const Input = React.forwardRef<
+  React.ElementRef<typeof InputPrimitive>,
+  React.ComponentPropsWithoutRef<typeof InputPrimitive>
+>(({ className, ...props }, ref) => (
+  <InputPrimitive
+    ref={ref}
+    className={clsx('canon-input', className)}
+    {...props}
+  />
+));
+Input.displayName = InputPrimitive.displayName;
+
+export { Input };
--- a/packages/canon/.storybook/preview.tsx
+++ b/packages/canon/.storybook/preview.tsx
@@ -75,7 +75,6 @@ const preview: Preview = {
           },
         },
       },
-      defaultViewport: 'small',
     },
   },
   decorators: [
--- a/packages/canon/package.json
+++ b/packages/canon/package.json
@@ -37,26 +37,26 @@
     "test": "backstage-cli package test"
   },
   "dependencies": {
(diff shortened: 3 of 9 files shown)


packages/canon/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,80 @@
...
+    className,
+    label,
+    description,
+    errorMatch = 'customError',
+    errorForceShow = false,
+    errorMessage = 'An error occurred',
+    validationMode = 'onChange',
+    size = 'md',
+    placeholder,
+    required = false,
```
Rugvip, line 36: Hmm, feeling we might need a bit more of a pattern for how we forward props if that's not provided by Base UI. Perhaps some of the more specific things like `errorForceShow` should be forwarded with a full `errorProps` or something like that instead? Or even a more general `props: { error: { ... } }` kind of pattern.

packages/canon/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,80 @@
...
+          'canon-input-control-sm': size === 'sm',
+          'canon-input-control-md': size === 'md',
+        })}
+      />
+
+      <Field.Error
+        className="canon-input-error"
+        match={errorMatch}
+        forceShow={errorForceShow}
+        // children={errorForceShow ? 'An error occurred' : null}
```
Rugvip, line 66: Should this be here or not? 😁

packages/canon/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+'use client';
+
+import React, { forwardRef } from 'react';
+import clsx from 'clsx';
+import { Field } from '@base-ui-components/react/field';
+import type { InputProps } from './types';
+
+/** @public */
+const Input = forwardRef<HTMLDivElement, InputProps>((props, ref) => {
```
Rugvip, line 25: Sup with this one vs https://mui.com/base-ui/react-input/?

packages/canon/src/components/Input/styles.css
```diff
@@ -0,0 +1,79 @@
...
+  border-color: var(--canon-border-hover);
+}
+
+.canon-input-control:focus-visible {
+  outline-color: var(--canon-border-selected);
+  outline-width: 0px;
+  border-color: var(--canon-border-selected);
+}
+
+.canon-input-control[data-invalid] {
```
Rugvip, line 56: QQ: are all of these various things documented somewhere? In the Base UI docs I see what seems to be an alternate `.*--error` pattern instead

Rugvip (approved):
Hmm, this one seems a tricky. Approved to merge and unblock things if you want but I find the interaction with Base UI to be a bit tricky as pointed out in a few comments

packages/canon/src/components/Input/styles.css
Rugvip: Was looking at the old docs, this is documented in the new docs: https://base-ui.com/react/components/input

canon-docs/public/components.css
```diff
@@ -277,3 +277,67 @@
...
+  border-color: var(--canon-error);
+}
+
+.canon-input-control-sm {
+  height: 2rem;
+}
+
+.canon-input-control-md {
+  height: 2.5rem;
+}
```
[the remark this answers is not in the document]
cdedreuille: Space tokens are made to for padding and margin only, not to control the height of some elements.

packages/canon/src/components/Input/Input.tsx
cdedreuille: We now split Field and Input so this solved this issue.

packages/canon/src/components/Input/Input.tsx
cdedreuille: It should not. After discussing we split Field and Input

cdedreuille:
@Rugvip I updated the `Field` and `Input` component to follow Base UI structure. I think we are good now. All docs have been updated as well.

PR #28179, opened 2024-12-18
move catalog collator to use cursor paged streaming

freben:


Rugvip (approved):

freben:
@backstage/discoverability-maintainers fyi

tikabom (approved):

Issue #24493, opened 2024-04-24
[Meta] New Backend System | Release 1.0


camilaibs:
## Table of Content

  - [Glossary](#glossary)
  - [Context](#context)
  - [Deadline](#deadline)
  - [Monitoring](#monitoring)
  - [Contributing](#contributing)
  - [Scope](#scope)
  - [Nice to Have](#nice-to-have)
  - [Out-of-Scope](#out-of-scope)

## Glossary <a href="#glossary" id="glossary"/>

<dl>
  <dt>NBS</dt>
  <dd>Abbreviation of "New backend System"</dd>
</dl>

## Context <a href="#context" id="context"/>

It is the stable release of the new backend system along with a deprecation of the old system and a huge push towards a stable plugin ecosystem built on top of the new backend system. The stable release is vital to ensure that adopters have confidence in the new system.

## Deadline <a href="#deadline" id="deadline"/>

We **estimate** that preparations for the stable version will take 2-3 months, with work starting this month, May 2024.
The stable version will be released in August-September releases, and the legacy system will be deprecated and no longer updated after that.
Next, we intend to completely remove the legacy system by the end of December 2024, although we may be able to remove it sooner if we move forward and complete the previous steps of preparation and deprecation faster.

## Monitoring <a href="#monitoring" id="monitoring"/>

This issue is mainly created for tracking the work to be done prior to the release, see a comprehensive list of tasks and their respective status in the [Scope](#scope) session.

## Contributing <a href="#contributing" id="contributing"/>

We are always open to contributions 🙏🏻 , so please comment here if you would like to work on some of the scoped tasks.

## Scope <a href="#scope" id="scope"/>

Excluding this milestone dependencies, it is primarily cleanup, review of APIs and documentation. There is also some amount of work related to ensuring the deprecation of the `@backstage/backend-common`, as well as any remaining plugin migrations of core features.

> [!Note]
Let us know by commenting in this issue if there's anything you would like us to fix or add before the release 🙌🏻 .

### Documentation

Documentation pages that need to be updated are tracked in the [external issue](https://github.com/backstage/backstage/issues/24195) below:

```[tasklist]
### Tasks
- [ ] https://github.com/backstage/backstage/issues/24195
```

### Framework

Here are tasks enumerated after a first review of the framework packages API:

#### Defaults and Common

```[tasklist]
### Tasks
- [ ] https://github.com/backstage/backstage/issues/24803
- [ ] https://github.com/backstage/backstage/issues/24851
[...]

PR #28182, opened 2024-12-18
remove old backend support in the catalog collator

freben:
The end of life of the old backend system [is upon us](https://github.com/backstage/backstage/issues/24493#deadline).


plugins/search-backend-module-catalog/src/module.ts
```diff
@@ -60,7 +59,7 @@ export const catalogCollatorExtensionPoint =
  *
  * @public
  */
-export default createBackendModule({
+export const searchBackendModule = createBackendModule({
```
emmaindal, line 62: any reason to not follow the naming patterns here? https://backstage.io/docs/backend-system/architecture/naming-patterns. I'd guess this would end up being `searchModuleCatalogCollator` or something like that?

freben: Good point, renamed! It isn't exported with that name, but indeed this was just me forgetting about the pattern.

benjdlambert (approved):

PR #28186, opened 2024-12-18
🧹  use `mockServices` more

freben:
Happened to come upon this pattern around the repo while doing other stuff. No functional changes, just moving from explicit `jest.Mocked<FooService>` to `mockServices.foo`


camilaibs (approved):
:tada

PR #28189, opened 2024-12-18
🧹  remove backend-common from notifications

freben:
This turned out to be an easy one

### DIFF
--- a/plugins/notifications-backend/dev/index.ts
+++ b/plugins/notifications-backend/dev/index.ts
@@ -20,13 +20,13 @@ import {
   createBackendPlugin,
 } from '@backstage/backend-plugin-api';
 import { notificationService } from '@backstage/plugin-notifications-node';
-import { errorHandler } from '@backstage/backend-common';
 import {
   notificationSeverities,
   NotificationSeverity,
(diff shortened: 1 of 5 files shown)


plugins/notifications-backend/dev/index.ts
```diff
@@ -96,7 +100,7 @@ const notificationsDebug = createBackendPlugin({
           });
           res.status(200).send({ status: 'ok' });
         });
-        router.use(errorHandler());
+        router.use(middleware.error());
```
camilaibs, line 103: Could we completely get rid of this middleware?

camilaibs (approved):

plugins/notifications-backend/dev/index.ts
freben: If I understood the debate earlier correctly, the default "root" handler doesn't get the logging context right, so it doesn't look obvious that the log line comes from this particular plugin. But I'm happy if that's not the case so we can delete these!

drodil (approved):

mareklibra (approved):

benjdlambert (approved):

Issue #28203, opened 2024-12-18
🚀 Feature: Techdocs from confluence


cb-jarunmadhesh:
### 🔖 Feature description

A new feature that enables Backstage TechDocs to source documentation directly from Confluence.

### 🎤 Context

Many organizations, including my company, rely heavily on Confluence for documentation because it is user-friendly and makes it easy to create and maintain content. Migrating or duplicating content to Markdown-based systems can be a cumbersome and time-consuming process.

Supporting Confluence as a TechDocs source would allow Backstage to better align with enterprise workflows and reduce barriers to adoption. My company uses Confluence for its documentation, and having TechDocs integration for Confluence would be immensely beneficial both for us and for other organizations with similar needs.

While there are existing tools, such as a scaffolder to migrate Confluence docs to Markdown on Github, this approach eliminates the need for manual migration and the challenge of maintaining duplicate documentation across Confluence and TechDocs.

### ✌️ Possible Implementation

The Preparer could be extended to support reading directly from Confluence by introducing a new Confluence URL Reader. This would allow TechDocs to accept and process Confluence URLs directly. 

I have a poc implementation and would be happy to raise a PR if this feature request is approved.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

awanlin:
Hi @cb-jarunmadhesh, On the surface this sounds like a good idea for some, but I worry that this will end up as a poor experience. 

Do you have a working implementation of this? If so can you share screen shots of what this looks like? Will this just bring in a single page or a whole set of pages? Would this require Adopters to follow some specific convention to make it work? How would the navigation work? What happens when pages have plugins being used? How does the HTML look and will it work with existing themes properly?

cb-jarunmadhesh:
Hi @awanlin,
Yes, I understand. Let’s ensure this is implemented properly:

1. Yes, I do have a working example. PFA a screen recording.
2. It'll fetch the page along with its child pages and their attachments. In Confluence, a page can have child pages. While converting to Markdown, I’m creating a Markdown file and a folder with the same name, which will contain the child pages, representing the directory structure.
[...]

PR #28230, opened 2024-12-19
Techdocs from confluence

cb-jarunmadhesh:

Implemented a new URL reader for confluence for techdocs to accept and process Confluence URLs
#28203 


packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
```diff
@@ -0,0 +1,266 @@
...
+      const children = await this.fetchChildPages(pageId);
+      for (const childPage of children) {
+        const childPageContent = await this.loadConfluencePage(
+          childPage.id,
+          `${path}/${title.toLocaleLowerCase().replaceAll(' ', '-')}`,
+        );
+        docItems.push(...childPageContent);
+      }
+    } catch (error) {
+      console.log(error);
```
alexlorenzi, line 141: Might have left this in by mistake

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
```diff
@@ -0,0 +1,266 @@
...
+    private readonly deps: {
+      treeResponseFactory: ReadTreeResponseFactory;
+    },
+  ) {}
+
+  async readUrl(
+    _url: string,
+    _options?: UrlReaderServiceReadUrlOptions,
+  ): Promise<UrlReaderServiceReadUrlResponse> {
+    throw new Error('Method not implemented.');
```
alexlorenzi, line 64: Maybe add a note in here for why it wasn't implemented, and if you plan on doing it in the future

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  parseUrl(url: string) {
+    const regex =
+      /https:\/\/((?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,6})\/wiki\/spaces\/([A-Za-z0-9_-]+)\/pages\/(\d+)\/([A-Za-z0-9-_+]+)/;
+    const [_string, host, space, pageId, title] = url.match(regex)!;
```
alexlorenzi, line 70: Are we safe to assume it'll match the URL? Maybe a check with an error message just in case

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
```diff
@@ -0,0 +1,266 @@
...
+    let markdownToPublish = markdown;
+    for (const attachment of attachments) {
+      const attachmentTitle = attachment.title.replace(/ /g, '-');
+      const attachmentTitleInUrl = attachment.title.replace(/ /g, '%20');
+      const regex = attachmentTitleInUrl.includes('.pdf')
+        ? new RegExp(`(\\[.*?\\]\\()(.*?${attachmentTitleInUrl}.*?)(\\))`, 'gi')
+        : new RegExp(
+            `(\\!\\[.*?\\]\\()(.*?${attachmentTitleInUrl}.*?)(\\))`,
+            'gi',
+          );
```
alexlorenzi, line 160: This is a little dense to read, maybe break it out a bit or add some comments

packages/integration/src/confluence/config.ts
```diff
@@ -0,0 +1,82 @@
...
+  config: Config,
+): ConfluenceIntegrationConfig {
+  const host = config.getString('host');
+  const apiToken = config.getString('apiToken');
+
+  if (!isValidHost(host)) {
+    throw new Error(
+      `Invalid Confluence integration config, '${host}' is not a valid host`,
+    );
+  }
```
alexlorenzi, line 58: Given the assumptions of the URL shape above, maybe you should check that the URL shape matches your expected confluence one in addition to the isValidHost check. This would give the operator faster feedback.

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
cb-jarunmadhesh: Yes, removed it in e0582bc

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
cb-jarunmadhesh: Added some comments to describe it. could you check if that's fine

packages/integration/src/confluence/config.ts
cb-jarunmadhesh: sure, that'd be better. done in e0582bc

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
cb-jarunmadhesh: Yeah, confluence page urls - from what I checked- are in this format `https://<company-domain>/wiki/spaces/<space-id>/pages/<page-id>/<page-title>`. 

Added a check to validate the url before parsing it.  [e0582bc](https://github.com/backstage/backstage/pull/28230/commits/e0582bca60b4884e79ff4ccc332996de4302bf07)

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
cb-jarunmadhesh: Both, readUrl and search, are not required in this case. We can implement readUrl if needed in the future.

cb-jarunmadhesh: Added the same in the Error message as well

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
```diff
@@ -0,0 +1,273 @@
...
+    const init = { headers: { Authorization: apiToken } };
+
+    const response = await this.fetchResponse(url, init);
+    const apiResponse = await response.json();
+
+    const childPages: ChildPage[] = apiResponse.results;
+    return childPages;
+  }
+
+  private htmlToMardDown(html: string) {
```
CasperJ, line 189: typo: MardDown => Markdown

cb-jarunmadhesh: Sure, will fix it. 
Let me know if there's anything else you'd like me to address. I can include it in this fix before committing

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
```diff
@@ -0,0 +1,273 @@
...
+    const init = { headers: { Authorization: apiToken } };
+
+    const response = await this.fetchResponse(url, init);
+    const apiResponse = await response.json();
+
+    const pageHtml = apiResponse.body.export_view.value;
+    const title = apiResponse.title;
+    const lastModifiedAt = apiResponse.version.createdAt;
+    return {
+      page: this.htmlToMardDown(pageHtml),
```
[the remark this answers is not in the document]
cb-jarunmadhesh: Fixed it in 8d35b97

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
cb-jarunmadhesh: Fixed it in 8d35b97

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
```diff
@@ -0,0 +1,273 @@
...
+
+  async readUrl(
+    _url: string,
+    _options?: UrlReaderServiceReadUrlOptions,
+  ): Promise<UrlReaderServiceReadUrlResponse> {
+    throw new Error('This method will be implemented in the future'); //+
+  }
+
+  parseUrl(url: string) {
+    // url format: https://<company-domain>/wiki/spaces/<space-id>/pages/<page-id>/<page-title>
```
awanlin, line 68: Can we add support for on-prem? That's going to come up right away if this moved forward. There are some examples here: https://github.com/backstage/backstage/blob/master/plugins/scaffolder-backend-module-confluence-to-markdown/src/actions/confluence/helpers.ts

awanlin (changes requested):
Thanks for the contribution @cb-jarunmadhesh, I left a comment regarding on-prem support. I'll also note that there is no documentation in this PR. You can create a `confluence` folder here: https://github.com/backstage/backstage/tree/master/docs/integrations. Then add the docs as needed.

I also left a comment on the issue you logged for this that would be really great to have answered 👍

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
cb-jarunmadhesh: Added support for on-perm @awanlin ..and added docs for confluence integration as well.
Could you please check them both?

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
```diff
@@ -0,0 +1,310 @@
...
+  parseUrl(url: string): {
+    spacekey: string;
+    title: string;
+    titleWithSpaces: string;
+  } {
+    let spacekey: string | undefined = undefined;
+    let title: string | undefined = undefined;
+    let titleWithSpaces: string | undefined = '';
+    const params = new URL(url);
+    if (params.pathname.split('/')[1] === 'display') {
```
zcmander, line 74: Could you add support for tinyurls (`https://mycompany.atlassian.net/x/3RE_Kg`) or page id (`https://mycompany.atlassian.net/pages/viewpage.action?pageId=1234567890`) URLs?

Apparently in some cases (cloud i think) it might be some like this: `<mysite>.atlassian.net/l/cp/XXXX` but I haven't seen that as I haven't used cloud confluence at all.

cb-jarunmadhesh: Hi, how can I get the page ID and space ID from a tiny URL! I use Cloud Confluence exclusively, but I’ve never encountered URLs like https://mycompany.atlassian.net/pages/viewpage.action?pageId=1234567890 or <mysite>.atlassian.net/l/cp/XXXX. Is there a way to configure Confluence to generate these types of URLs?

Also, just curious—why do Confluence URLs come in so many different formats?

cb-jarunmadhesh: Hi, how can I get the page ID and space ID from a tiny URL!
I use cloud confluence only, but I’ve never encountered URLs like https://mycompany.atlassian.net/pages/viewpage.action?pageId=1234567890 or <mysite>.atlassian.net/l/cp/XXXX. Is there a way to configure confluence to generate these types of URLs?

Also, just curious—why do confluence urls come in so many different formats?

docs/integrations/confluence/techdocs.md
```diff
@@ -0,0 +1,22 @@
...
+To use this integration, add configuration to your root `app-config.yaml`:
+
+```yaml
+integrations:
+  confluence:
+    - host: mycompany.atlassian.net
+      apiToken: ${ATLASSIAN_API_TOKEN}
+```
+
+:warning: Token generated from atlassian cannot be used here directly. Please provide the encoded token here. [Refer](https://developer.atlassian.com/cloud/jira/platform/basic-auth-for-rest-apis/)
```
awanlin, line 22: Can we change this to use the raw token instead? There's precedence for this as the current [Confluence to markdown scaffolder action](https://github.com/backstage/backstage/tree/master/plugins/scaffolder-backend-module-confluence-to-markdown#auth-methods) and the [Confluence search collator](https://github.com/backstage/community-plugins/tree/main/workspaces/confluence/plugins/search-backend-module-confluence-collator#auth-methods) both work this way. This would make it easier for people to add this while using their existing tokens.

awanlin (changes requested):
Thanks for adding the docs, I think we might need a few changes to get them to show up properly but I'm short on time to test it out and share the needed changes, will do so soon.

I did notice one item that I think should be changed and left a comment about it. 👍

packages/backend-defaults/src/entrypoints/urlReader/lib/ConfluenceUrlReader.ts
zcmander: Unfortunately, I've no idea how to fetch those from share-links. But you can generate ones for yourself, see docs: https://support.atlassian.com/confluence-cloud/docs/share-a-page-or-blog-post/

I've seen confluence to generate `viewpage.action?pageId=xx` link in case the title of the page contains some special characters that would not url encode correctly (or human-friendly or something...) but I'm not sure what characters trigger that effect.

docs/integrations/confluence/techdocs.md
cb-jarunmadhesh: @awanlin Updated to use the raw token. Sorry for the delayed response.

cb-mohitbhatia: Hi @awanlin - can you please help us with review and merge this PR, we are waiting for this integration. Thank you!

freben:
I have a slightly naive question. Is this a typical way of making techdocs integrations? The integrations package and the url reader are pretty low level facilities that are meant for very general use and reading binary data from files. It seems to me, that the process of extracting confluence into techdocs would be a very non-general process involving complex data and metadata and attachments and transformations and stuff. A task that the url reader to my mind sounds like a poor fit for. I wonder if there's some other mechanism of achieving what we're gunning for here, that doesn't involve the url reader at all.

cb-jarunmadhesh:
@freben Confluence ain't that complicated like you say. Even if it is, there's a way to export the processed html page and we use that html to convert to markdown here. I'm not sure what you meant by complex data and metadata and attachments and transformations though, but we are able to fetch the content, attachments and child pages; and not able to fetch the plugins only, with this approach.

freben:
@cb-jarunmadhesh What I meant was not in complexity of technical implementation - but complexity in terms of semantics.

The url reader takes an url and returns a binary buffer of the file contents. For a file that's trivial, semantically. And it's something that hundreds of plugins can lean against and leverage. It's intended for having an unbounded scope of usage.

If I am understanding things right, this PR uses that same interface to return a "baked" version of a remote resource. And it does so in such a way that its scope of usability is exactly one precise type of consumer, ie having zero reusability. Someone else might have a completely conflicting view of what it means to "read a confluence page" - what's the output format going to be? A zip file of the contents and attachments in HTML form? Just the raw page as markdown? A json confluence API response that you then follow up with doing more work based off? Or something else?

It's almost like hot patching into the `fs` module of node, and saying that "oh if you read a wav file you don't get the bits back, you get an mp3 compressed version of it instead". Ok that was a poor analogy, but url reader being on the same sort of semantic level as the fs module or perhaps `fetch`, might at least be somewhat illustrative.

Again, with the caveat that I may be misunderstanding things and that this actually opens up for good use cases.

cb-jarunmadhesh:
@freben So, this is what I understand: we want URLReader to read a URL and return whatever it is in its "true form"—if it's a video, we get a video; if it's a photo, we get a photo, in whatever format it's encoded in. Right?

For a Confluence page, I see it as just that—a page, an HTML page—and that's what I'm trying to fetch here. Plus, I'm downloading any attachments linked to the page—whether it's a .jpeg, .png, .mp4, or .pdf—and keeping them in their original format, linking them to the HTML page. I'm also checking if the page has child pages and fetching those as HTML pages as well.

But here's the catch: I'm converting them to .md. Why? Because it was easier for me to process and establish connections in markdown. Can we do it without converting to .md? Yes. But do we need to? For the TechDocs application, we don’t have to.

So, We can wither start with step one and build something useful for TechDocs, then tweak it as needed when other applications come up. or we let it not serve TechDocs itself.

If you think URLReader isn’t the best place for this, I’m open to suggestions. Please let me know a better solution.
Should we make it a separate plugin? Personally, I think that’d be a stretch thou.

If implemented, this would help us a lot. I’d say something is better than nothing, but you’re the maintainer—your call!

vrabbi:
Any chances of getting this merged? This would be extremely helpful for us as well

cb-jarunmadhesh:
@awanlin @freben Any comment on this?

KrishanNagarro:
@cb-jarunmadhesh , i tried to run this locally, but it fails and says that mkdocs.yaml not found.

PR #28253, opened 2024-12-20
fix: Omit collecting `backend-common` `config.d.ts` if `backend-defaults` is detected

### LINKED ISSUE #28170: 🐛 Bug Report: Database connection objects no longer pass schema validation
### 📜 Description

Looks like this was caused by https://github.com/backstage/backstage/pull/27901

```

Error: Configuration does not match schema
--
  |  
  | Config must have required property 'instance' { missingProperty=instance } at /backend/database/connection
  | Config must have required property 'type' { missingProperty=type } at /backend/database/connection
  | Config must NOT have additional properties { additionalProperty=host } at /backend/database/connection
  | Config must NOT have additional properties { additionalProperty=port } at /backend/database/connection
  | Config must NOT have additional properties { additionalProperty=user } at /backend/database/connection
  | Config must NOT have additional properties { additionalProperty=host } at /backend/database/connection
  | Config must NOT have additional properties { additionalProperty=port } at /backend/database/connection
  | Config must NOT have additional properties { additionalProperty=user } at /backend/database/connection
  | Config must be string { type=string } at /backend/database/connection
  | Config must match a schema in anyOf {  } at /backend/database/connection


```

My config looks like this:

```
  database:
    client: pg
    connection:
      host: ${DB_HOST}
      port: ${DB_PORT}
      user: ${DB_USERNAME}
      password: ${DB_PASSWORD}
```
It's trying to validate it against the new CloudSQL schema:

```
          {
            /**
             * The specific config for cloudsql connections
             */
            type: 'cloudsql';
            /**
             * The instance connection name for the cloudsql instance, e.g. `project:region:instance`
             */
            instance: string;
          }
```

instead of the more flexible type:
```
          {
            /**
             * Password that belongs to the client User
             * @visibility secret
             */
            password?: string;
            /**
             * Other connection settings
             */
            [key: string]: unknown;
          }
```

### 👍 Expected behavior

`yarn backstage-cli config:check --strict` should pass. This was not listed as a breaking change so I'm assuming I shouldn't need to change my config at all. I'm happy to do so if it was simply mislabeled though.

`yarn backstage-cli config:check --strict` fails with above description

### 👟 Reproduction steps

I assume: create a new backstage app and setup postgres then run `yarn backstage-cli config:check --strict`

This is new in 1.34.0

### 🖥️ Your Environment

_No response_

benjdlambert:
Fixes #28170

### DIFF
--- a/packages/config-loader/src/schema/collect.ts
+++ b/packages/config-loader/src/schema/collect.ts
@@ -153,18 +153,17 @@ export async function collectConfigSchemas(
   ]);
 
   const tsSchemas = await compileTsSchemas(tsSchemaPaths);
+  const allSchemas = schemas.concat(tsSchemas);
 
-  return schemas
-    .concat(tsSchemas)
-    .filter(
-      ({ packageName }, _, original) =>
-        !(
-          packageName === '@backstage/backend-common' &&
-          original.some(
-            ({ packageName: p }) => p === '@backstage/backend-defaults',
-          )
-        ),
-    );
+  const isMissingBackendDefaults = !allSchemas.some(
+    ({ packageName }) => packageName === '@backstage/backend-defaults',
+  );
+
+  // Filter out the backend-common schema unless the backend-defaults schema is missing
+  return allSchemas.filter(
+    ({ packageName }) =>
+      packageName !== '@backstage/backend-common' || isMissingBackendDefaults,
+  );
 }
 
 // This handles the support of TypeScript .d.ts config schema declarations.
--- a/packages/config-loader/src/schema/collect.ts
+++ b/packages/config-loader/src/schema/collect.ts
@@ -159,7 +159,10 @@ export async function collectConfigSchemas(
     ({ packageName }) => packageName === '@backstage/backend-defaults',
   );
 
-  // Filter out the backend-common schema unless the backend-defaults schema is missing
+  // Filter out the backend-common schema unless the backend-defaults schema is missing.
+  // This was causing issue with merging of the schemas.
+  // This should be removed when we have no need for backend-common anymore, and other
+  // packages are no longer depending on it.
   return allSchemas.filter(
     ({ packageName }) =>
       packageName !== '@backstage/backend-common' || isMissingBackendDefaults,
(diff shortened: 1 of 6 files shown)


camilaibs (approved):

packages/config-loader/src/schema/collect.ts
```diff
@@ -151,14 +154,26 @@ export async function collectConfigSchemas(
...
-  return schemas.concat(tsSchemas);
+  return schemas
+    .concat(tsSchemas)
+    .filter(
+      ({ packageName }, _, original) =>
+        !(
+          packageName === '@backstage/backend-common' &&
+          original.some(
+            ({ packageName: p }) => p === '@backstage/backend-defaults',
+          )
```
Rugvip, line 165: Bit cleaner to extract this into an earlier `hasBackendDefaults` I think

packages/config-loader/src/schema/collect.ts
```diff
@@ -45,7 +45,7 @@ export async function collectConfigSchemas(
   packagePaths: string[],
 ): Promise<ConfigSchemaPackageEntry[]> {
   const schemas = new Array<ConfigSchemaPackageEntry>();
```
Rugvip, line 47: Thinking `& { packageName: string }` here instead to avoid the public API change

packages/config-loader/src/schema/collect.ts
```diff
@@ -150,15 +153,26 @@ export async function collectConfigSchemas(
...
+  const isMissingBackendDefaults = !allSchemas.some(
+    ({ packageName }) => packageName === '@backstage/backend-defaults',
+  );
 
-  return schemas.concat(tsSchemas);
+  // Filter out the backend-common schema unless the backend-defaults schema is missing
+  return allSchemas.filter(
+    ({ packageName }) =>
+      packageName !== '@backstage/backend-common' || isMissingBackendDefaults,
+  );
```
Rugvip, line 166:
```
if (hasBackendDefaults) {
  return allSchemas.filter(...)
}
return allSchemas
```

thinking that's the most clear way

Rugvip (approved):

PR #28266, opened 2024-12-21
add error handler middleware in the plugin router

freben:
This may be mildly controversial.

On the upside, few plugins were using the error handler middleware, so their errors will now be more precisely logged and reported, tagged with the plugin ID. And we could now remove ALL `MiddlewareFactory` usage in the entire `plugins` directory which surely is a nice simplification.

On the downside, this is now a "forced" behaviour, and the `httpRouter` does not yet have a `configure` like the `rootHttpRouter` does. Those who relied on errors bubbling all the way down to the root http router for custom error handling, will now have to hoist their error handling up to the plugin level. This is therefore marked as a breaking change.

This also does mean that in the response body you'll have the "local" route, i.e. without the `/api/{pluginId}` prefix. This is a matter of taste but I kinda would not have minded if it were the full path including the prefix.

### DIFF
--- a/.changeset/shiny-walls-press.md
+++ b/.changeset/shiny-walls-press.md
@@ -2,6 +2,6 @@
 '@backstage/backend-defaults': minor
 ---
 
-**BREAKING**: Ensure that an error handler middleware exists at the end of each plugin `httpRouter` handler chain. This makes it so that exceptions thrown by plugin routes are caught and encoded in the standard error format.
+Ensure that an error handler middleware exists at the end of each plugin `httpRouter` handler chain. This makes it so that exceptions thrown by plugin routes are caught and encoded in the standard error format.
 
 If you were using the standard `MiddlewareFactory` just to put an `error` middleware in you router, you can now remove that at your earliest convenience since it's redundant. If you have custom error handlers in your plugin router, those will continue to function as previously. If you were relying on thrown errors propagating all the way down to the root HTTP router, you will find that they no longer do that, and may want to hoist your error handling up to the plugin level instead.
--- a/.changeset/shiny-walls-press.md
+++ b/.changeset/shiny-walls-press.md
@@ -1,0 +1,7 @@
+---
+'@backstage/backend-defaults': minor
+---
+
+**BREAKING**: Ensure that an error handler middleware exists at the end of each plugin `httpRouter` handler chain. This makes it so that exceptions thrown by plugin routes are caught and encoded in the standard error format.
+
+If you were using the standard `MiddlewareFactory` just to put an `error` middleware in you router, you can now remove that at your earliest convenience since it's redundant. If you have custom error handlers in your plugin router, those will continue to function as previously. If you were relying on thrown errors propagating all the way down to the root HTTP router, you will find that they no longer do that, and may want to hoist your error handling up to the plugin level instead.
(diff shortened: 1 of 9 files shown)


Rugvip (commented):
Hmm, not sure I understand the benefit? Is it so that we can add plugin-specific meta to the error handling?

plugins/catalog-backend-module-incremental-ingestion/src/router/routes.ts
```diff
@@ -253,12 +248,6 @@ export class IncrementalProviderRouter {
       },
     );
 
-    const middleware = MiddlewareFactory.create({
-      logger: this.logger,
-      config: this.config,
-    });
-    router.use(middleware.error());
```
Rugvip, line 260: This is already redundant and we can get rid of it either way right?

freben:
When we lean on the root error handler, there are two outward visible differences:

- The reported URL is the full one including the prefix
- The log uses the root logger, so isn't tagged with the plugin id (or other customer specific plugin level adaptations)

I think maybe we should either do what's in this PR, or just stop ever putting these middlewares inside plugin code.

plugins/catalog-backend-module-incremental-ingestion/src/router/routes.ts
freben: Except that when you do, you get less informative error output.

.changeset/shiny-walls-press.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/backend-defaults': minor
+---
+
+**BREAKING**: Ensure that an error handler middleware exists at the end of each plugin `httpRouter` handler chain. This makes it so that exceptions thrown by plugin routes are caught and encoded in the standard error format.
```
Rugvip, line 5: Hmm, what is this actually breaking? 😁 I'd expect this to at most be **BREAKING PRODUCERS**?

packages/backend-defaults/src/entrypoints/httpRouter/httpRouterServiceFactory.ts
```diff
@@ -63,9 +73,15 @@ export const httpRouterServiceFactory = createServiceFactory({
     router.use(credentialsBarrier.middleware);
     router.use(createCookieAuthRefreshMiddleware({ auth, httpAuth }));
 
+    const pluginRoutes = PromiseRouter();
+    router.use(pluginRoutes);
+
+    const middleware = MiddlewareFactory.create({ config, logger });
```
Rugvip, line 79: Perhaps for later, but I think we should probably move the error handler out of the `MiddlewareFactory`, given that it's used in a couple of different places. This one is really only intended for the root HTTP service.

Rugvip (approved):
Alright, yep I think simply having the plugin ID in there is a good enough improvement tbh.

.changeset/shiny-walls-press.md
freben: Mm, maybe it was a bit strict, removed it entirely.

packages/backend-defaults/src/entrypoints/httpRouter/httpRouterServiceFactory.ts
freben: Agreed, it's a bit sneaky to take it out of where it's clearly exported by the root router. We should keep it in mind at the very least.

PR #28267, opened 2024-12-21
🧹  remove usages of some backend-common helpers

freben:
This is on top of #28266. It targets `notFoundHandler`, `errorHandler` etc from the old `backend-common`.

### DIFF
--- a/plugins/app-backend/src/service/router.ts
+++ b/plugins/app-backend/src/service/router.ts
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-import { notFoundHandler } from '@backstage/backend-common';
 import {
   DatabaseService,
   resolvePackagePath,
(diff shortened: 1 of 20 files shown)


plugins/app-backend/src/service/router.ts
```diff
@@ -299,7 +298,9 @@ async function createEntryPointRouter({
   if (staticFallbackHandler) {
     staticRouter.use(staticFallbackHandler);
   }
-  staticRouter.use(notFoundHandler());
+  staticRouter.use((_req: Request, res: Response) => {
+    res.status(404).end();
```
camilaibs, line 302: Should we use `MiddlewareFactory.notfound` by convention? I wondered if it would be useful to have a `createTestRouter` or something that installs the standard middlewares by default... Nothing to block this pull request, just a possible idea of ​​follow-up 🙂 .

camilaibs (approved):

plugins/app-backend/src/service/router.ts
freben: Here it turned out to be complex to keep the middleware factory version. I honestly forget why but I had it here first and then removed it. Probably because of the dependency change needed.

PR #28308, opened 2024-12-26
cli: add support for native ESM in Node.js packages

### LINKED ISSUE #8242: CLI: Support dynamic imports in backend code
## Feature Suggestion

We should add support for `import()` in backend code, that is dynamic `import()` expressions that do **not** got transpiled into a `require()` call but stay as is. By supporting dynamic imports, we'll make it possible to import ESM modules into Node.js CommonJS modules, and move forward with overall migration towards using the new module syntax in Node.js.

Dynamic imports have been supported in Node.js LTS since 14, so it is safe to pass them through in our build process as we currently only support Node.js 14 and 16. Note that we do not want to start transpiling all our Node.js code to ESM yet, as this would break existing usage of our packages. We rather want to add support for `import()`s in our CommonJS build process.

## Possible Implementation

We'll need to update both the build and bundle transpilation in the CLI. The building, which is what creates a production build of a single package, is handled by rollup. The configuration for that lives [here](https://github.com/backstage/backstage/blob/ca9176903623879d6c6cd41c6e5ea0c6ab083464/packages/cli/src/lib/builder/config.ts#L66), and we'll want to update our CommonJS output configuration to pass through dynamic imports.

The bundling, which we primarily used for our backend development setup, is done by webpack, and reuses a lot of our frontend webpack config. The transpilation setup for that lives [here](https://github.com/backstage/backstage/blob/ca9176903623879d6c6cd41c6e5ea0c6ab083464/packages/cli/src/lib/bundler/transforms.ts#L52), but to be honest I don't know whether this already transforms things like we want to or not, I imagine it actually already does. The last piece would be the webpack backend config itself, which lives [here](https://github.com/backstage/backstage/blob/ca9176903623879d6c6cd41c6e5ea0c6ab083464/packages/cli/src/lib/bundler/config.ts#L232), and again I'm not actually sure whether this needs updating or not, but I would imagine it does because of the `target: 'node'`

### LINKED ISSUE #12218: backend: full support for ESM packages
## Feature Suggestion

Full ESM support, including the package `"exports"` field with the module format specifiers, across all aspects of development.

## Possible Implementation

We currently support ESM in the frontend and tests. Backend support is more tricky as we're executing code in Node.js, which is a lot more picky about module formats. We likely want a combination of build tooling additions and refactors. The optimal solution would ofc be if there's no impact on existing code at all, since in the end our source is mostly ESM.

A couple of dependencies:

- [ ] Support for dynamic imports, #8242
- [x] Support for the exports field, #7926

Rugvip:

Happy holidays! 🎁 😁 

Closes #8242, closes #12218

This adds support for ESM in Node.js code. It keeps existing behavior mostly as is, and uses the explicit `.mts` and `.cts` extension to opt-in to new behavior, as well as `"type": "module"` in `package.json.

It keeps the same TypeScript module resolution, although it seems to work pretty well with `"bundle"` too. I figured it was cleanest to enable `allowImportingTsExtensions`, which lets you import for example `myModule.mts` instead of `myModule.mjs` when `.mts` is what exists on the file system. The downside of using it is that editors might get confused at times, but it makes the resolution logic simpler and imo makes more sense when not using the build tooling provided by TypeScript itself.

The most important breaking change is that `import(...)` now behaves differently in Node.js code, even when importing a CommonJS module. This is because the existing interop layer is removed, and we can't rely on the existing one in Node.js, because it breaks existing dependencies. It's simple to fix though, existing `import(...)`s can be replaced with a `require(...) as typeof import(...)`, since that is what it was currently being transformed to.

Jest support is ... weird. I've tried a couple of different approaches, including manually detecting `"type": "module"` like in the node transform. It doesn't work all too well though, because the decision to load things via CommonJS or ESM is done quite early on in the Jest module loading system. In the end I settled on the approach of letting the tested package decide whether `.ts` files should be treated as modules or not, which means that you can't have internal dependencies on non-module package.


docs/tooling/cli/02-build-system.md
Parsifal-M, line 1: Think it looks good to me from a docs point of view :smile: 

Not sure I fully understand the benefits but I probably need to do some googling on the `commonJS` vs `ES Modules` in Node :smiley:

packages/cli/config/nodeTransformHooks.mjs
```diff
@@ -0,0 +1,282 @@
...
+
+  const packageJson = JSON.parse(await readFile(packageJsonPath, 'utf8'));
+  return {
+    ...resolved,
+    format: packageJson.type ?? DEFAULT_MODULE_FORMAT,
+  };
+}
+
+/**
+ * Find the closes package.json file from the given path.
```
freben, line 149:
```suggestion
 * Find the closes package.json file from the given path.
```

freben (approved):
Complex stuff!

plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.test.ts
```diff
@@ -92,9 +92,12 @@ describe('NunjucksWorkflowRunner', () => {
     );
   }
 
-  beforeEach(() => {
+  beforeEach(async () => {
     mockDir.clear();
 
+    // This one is ESM-only
+    stripAnsi = await import('strip-ansi').then(m => m.default);
```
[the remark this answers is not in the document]
Rugvip: Yep something like that

packages/cli/config/nodeTransform.cjs
```diff
@@ -76,3 +80,8 @@ addHook(
   },
   { extensions: ['.js', '.cjs'], ignoreNodeModules: true },
 );
+
+// Register module hooks, used by "type": "module" in package.json, .mjs and
+// .mts files, as well as dynamic import(...)s, although dynamic imports will be
+// handled be the CommonJS hooks in this file if what it points to is CommonJS.
+Module.register('./nodeTransformHooks.mjs', pathToFileURL(__filename));
```
fordneild, line 87: Im getting `TypeError: Module.register is not a function` when trying to update my backstage to this version
Have you seen this error before?

PR #28314, opened 2024-12-28
Add ajv and immer peer dependencies to plugin-scaffolder-react

andrcatros:

While setting up Backstage to run with yarn pnp, I found that `plugin-scaffolder-react` has 2 peer dependencies (inherited from its dependencies) that are not explicitly declared:

- `ajv` - a peer dependency of `ajv-errors` 
- `immer` - a peer dependency of `use-immer` 

Peer dependency versions from `yarn.lock`:
- [ajv-errors peer dependencies](https://github.com/backstage/backstage/blob/a51536e54649b36190a5312aafce5f623af928e0/yarn.lock#L21984)
- [use-immer peer dependencies](https://github.com/backstage/backstage/blob/a51536e54649b36190a5312aafce5f623af928e0/yarn.lock#L45887)


benjdlambert:
Should these not just be moved to actual dependencies instead of peer deps? Don't really wanna have to force the backstage integrators to have to define these packages in their `package.json`. Probably makes more sense for the package that consumes these to define them explicitly.

benjdlambert:
@andrcatros friendly ping here?

benjdlambert (approved):
I ended up moving the dependencies from the `peerDependencies` block to the real deps as after chats with other maintainers it's probably not worth just pushing the problem up the chain.

Thanks! :pray:

PR #28315, opened 2024-12-28
repo-tools: add SQL reports

Rugvip:

Same as API reports and the CLI reports, but this time for the SQL schema based on migration files. I've been wishing for a way to get an overview of the current schema for plugins without needing to set up a Postgres DB and inspect the schema from there. I also figured this is a good opportunity to add some basic built-in testing for down migrations by making sure that the schema is the same when migrating back down.

Just doing this for Postgres for now to keep it simple, and figured that's the most important one anyway, especially for down migrations. The reports include a high level description of the schema, but doesn't go into all the details such as foreign key constraints or index types. We can add more things over time if they prove to be useful.

It's behind the `--sql-reports` flag for now, as there's an optional peer dependency on `embedded-postgres`. I'm a bit unsure how well that one works on different systems, so wanted to roll this out only in the main repo for now. This is also a bit of a trial run for whether it would make sense to switch all development and testing to use `embedded-postgres` instead of SQLite.

On top of #28308 since `embedded-postgres` is ESM-only.


.github/vale/config/vocabularies/Backstage/accept.txt
```diff
@@ -276,6 +276,7 @@ nohoist
 nonces
 noop
 npm
+Nullable
```
freben, line 279:
```suggestion
nullable
```

Otherwise only that casing is allowed

packages/repo-tools/package.json
```diff
@@ -84,17 +86,22 @@
     "@backstage/types": "workspace:^",
     "@types/is-glob": "^4.0.2",
     "@types/node": "^20.16.0",
-    "@types/prettier": "^2.0.0"
+    "@types/prettier": "^2.0.0",
+    "embedded-postgres": "17.2.0-beta.15"
```
freben, line 90: Did the beta have something in particular you needed, or could we be on the mainline instead?

packages/repo-tools/src/commands/api-reports/sql-reports/getPgSchemaInfo.ts
```diff
@@ -0,0 +1,88 @@
...
+import { SchemaIndexInfo, SchemaInfo, SchemaSequenceInfo } from './types';
+
+export async function getPgSchemaInfo(knex: Knex): Promise<SchemaInfo> {
+  const { rows: tableNames } = await knex.raw<{ rows: { name: string }[] }>(`
+    SELECT table_name as name
+    FROM information_schema.tables
+    WHERE
+      table_schema = 'public'
+      AND table_type = 'BASE TABLE'
+      AND table_name NOT LIKE 'knex_migrations%'
```
freben, line 27: Could maybe even make it `%knex%` (same below)

packages/repo-tools/src/commands/api-reports/sql-reports/runSqlExtraction.ts
```diff
@@ -0,0 +1,243 @@
...
+    user: 'postgres',
+    password: 'password',
+  };
+  const pg = new EmbeddedPostgres({
+    databaseDir: './data/db',
+    ...basePgOpts,
+    port,
+    persistent: false,
+    onError(_messageOrError) {
+      // console.error('EmbeddedPostgres error:', messageOrError);
```
freben, line 57: Work in progress?

packages/repo-tools/src/commands/api-reports/sql-reports/runSqlExtraction.ts
```diff
@@ -0,0 +1,243 @@
...
+        logApiReportInstructions();
+      }
+      throw new Error(`Report ${reportPath} is out of date`);
+    }
+  }
+}
+
+function prettyReport(content: string): string {
+  try {
+    const prettier = require('prettier') as typeof import('prettier');
```
freben, line 233: So in this esm world, it's not even possible to await import a cjs library, right?

packages/repo-tools/src/commands/api-reports/sql-reports/runSqlExtraction.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    const config = prettier.resolveConfig.sync(cliPaths.targetRoot) ?? {};
```
freben, line 235: In the next versions of prettier I believe the sync version is removed. To lower future migration hurdles, if it's possible, consider using the async versions of its api

freben (commented):
Neat!

packages/repo-tools/src/commands/api-reports/sql-reports/runSqlExtraction.ts
Rugvip: It is, but it works poorly with default exports, you end up with `.default.default`. Same with named exports unless you are importing a CommonJS module that is handled with the interop layer. In the end `require(...)` is generally simpler and more or less does the same thing anyway.

packages/repo-tools/package.json
Rugvip: There aren't any mainline releases, they're all marked as beta for now: https://www.npmjs.com/package/embedded-postgres?activeTab=versions

packages/repo-tools/src/commands/api-reports/sql-reports/runSqlExtraction.ts
Rugvip: Needs to be explicit no-op, but I'll clean these up

packages/repo-tools/src/commands/api-reports/sql-reports/getPgSchemaInfo.ts
Rugvip: Eh, I like to keep it somewhat explicit

packages/repo-tools/src/commands/api-reports/sql-reports/runSqlExtraction.ts
freben: Mmm isn't that because of that interop setting I enabled somewhere 🤔 ok then

packages/repo-tools/src/commands/api-reports/sql-reports/runSqlExtraction.ts
Rugvip: I plan a followup to refactor the API reports code, prolly best to handle as part of that. This is copied from a different place but didn't wanna do the whole refactor in this PR.

packages/repo-tools/src/commands/api-reports/sql-reports/runSqlExtraction.ts
Rugvip: I think that one just decides the behavior of imports in general, but these dynamic imports no longer go via any rollup compat layer at all and are just left as they are for Node. Node does have a similar interop layer, but we can't use it by default because it breaks dependencies

benjdlambert:
What's the motivation here for using `embedded-postgres` vs something like [`pglite`](https://pglite.dev/) instead?

PGLite is a lot smaller, and a lot faster with some of my simple experiments to use that instead of `embedded-postgres` for this use case.

Just thinking that it would also just be a complete drop in replacement for sqlite too, at a fraction of the cost.

Rugvip:
@benjdlambert I used `embedded-postgres` here because of it being a bit of a trial for #28318. I find it to be a much better fit for a dev DB because you don't have any of the connection constraints that might mess with perf and things like listeners, and you don't need to worry about restoring the DB after restarts since it's managed by the parent process.

I'll prolly try to switch in pglite for this still tho to see how it works

Rugvip:
Took @benjdlambert's pglite experiment and moved over fully to that, so this is now on top of `master` since it no longer needs ESM support

packages/repo-tools/src/commands/api-reports/sql-reports/runSqlExtraction.ts
```diff
@@ -0,0 +1,215 @@
...
+  // Since we're passing this as the client we need to replace the `config.client` with `pg` afterwards
+  class WrappedClientPgLite extends ClientPgLite {
+    constructor(config: any) {
+      super({ ...config, client: 'pg' });
+    }
+  }
+
+  // Patch this since it's not included in knex-pglite, but required by our knex code
+  // ClientPgLite.prototype.config = { client: 'pg' };
+
```
benjdlambert, line 45: Don't need this anymore right since we have the above wrapping class?

benjdlambert (changes requested):
Just one small nit, but looks great! :tada:

packages/repo-tools/src/commands/api-reports/sql-reports/runSqlExtraction.ts
Rugvip: Ah yep that was the first attempt 😅

packages/repo-tools/src/commands/api-reports/sql-reports/runSqlExtraction.ts
```diff
@@ -0,0 +1,212 @@
...
+      const knex = Knex({
+        client: WrappedClientPgLite,
+        dialect: 'postgres',
+        connection: {
+          database,
+        },
+      });
+
+      await knex.raw(`CREATE DATABASE "${database}"`);
```
freben, line 78:
```suggestion
      await knex.raw('CREATE DATABASE ??', [database]);
```

not that it matters much

freben (approved):
<3

benjdlambert (approved):

PR #28317, opened 2024-12-29
Canon website

cdedreuille:
We are introducing a new documentation site for Canon. As documentation is a big part of the development journey, we decided to branch out from our Storybook and build our own dedicated website based on NextJs.

To avoid duplicating code, we are trying the use of embed stories from Storybook and Chromatic. The website will work for both light and dark mode as well as the V1 and V2 of our theme.

CleanShot 2024-12-31 at 10 35 28@2x


packages/repo-tools/src/commands/peer-deps/peer-deps.ts
```diff
@@ -104,6 +104,11 @@ export default async ({ fix }: { fix: boolean }) => {
   const packagesWithRelevantDependencies = [];
 
   for (const pkg of packages) {
+    if (pkg.dir.includes('canon-website')) {
```
mtlewis, line 107: Rather than special-casing the canon-website specifically, should we perhaps filter out packages that don't have a Backstage role? We could rename `isStandaloneApplication` to `isBackstagePackage`, invert the logic, and add the clause in there.

packages/canon/src/components/Heading/Heading.tsx
```diff
@@ -14,10 +14,13 @@
  * limitations under the License.
  */
 
+'use client';
```
mtlewis, line 17: Backstage doesn't do any server-side rendering, so this directive doesn't make sense when this component is used in production. It seems a little counter-productive to have changes in components that are motivated strictly by the Canon website - perhaps we can move this directive further up the tree into website code to get around this need?

packages/canon/src/components/Button/styles.css
```diff
@@ -35,18 +35,18 @@
 
   &:hover {
     background-color: transparent;
-    box-shadow: inset 0 0 0 1px var(--canon-outline-focus);
+    box-shadow: inset 0 0 0 1px var(--canon-border-focus);
```
mtlewis, line 38: Just picking a random spot for this comment really: it seems like there's a good bit of code in here that's not specifically to do with the Canon website. I think it might be worth splitting the change up to make it more digestible, given the overall diff is so large.

packages/canon/src/components/Heading/Heading.tsx
cdedreuille: I could but I think this is good practice anyway. All Base UI components have this directive as well. The more we can make this library React 19 ready the better. https://github.com/mui/base-ui/blob/master/packages/react/src/accordion/root/AccordionRoot.tsx

packages/canon/src/components/Button/styles.css
cdedreuille: All the changes made to Canon were actually made for the website to work since we are using these components in practice. I could split it but some part of the site will just not work as expected.

packages/canon/src/components/Heading/Heading.tsx
cdedreuille: I removed all 'use client' directives. I'll do that as a separate PR.

packages/repo-tools/src/commands/peer-deps/peer-deps.ts
cdedreuille: We removed this change to avoid adding the new website as a workspace.

canon-docs/public/core.css
Rugvip, line 1: As far as I can tell these CSS files will easily drift from the ones in `packages/canon`. If you update them at the source they won't actually be updated here until someone goes and runs `dev-all` or `watch-css` and commits those changes.

Is it not possible to in some way import the CSS directly from `packages/canon` and avoid the build step? I think it would be fine to skip the CSS bundling step too right, assuming Next.js can handle that?

Rugvip (commented):
Nice! 👍 

Just a few project structure bits, but main concern is the CSS setup.

canon-docs/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "jsx": "preserve",
+    "incremental": true,
+    "plugins": [
+      {
+        "name": "next"
+      }
+    ],
+    "paths": {
+      "@/*": ["./src/*"]
+    }
```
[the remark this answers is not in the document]
cdedreuille: This is basic Next.js setup. Would you be ok if we keep that one as this is actually quite handy.

canon-docs/public/core.css
cdedreuille: I 100% agree but I could not find a way right now as for iframes we need to have them at the root and not import it from the system. I can explore this more though.

canon-docs/README.md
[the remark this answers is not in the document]
cdedreuille: Cleaned up 🧹

canon-docs/next.config.mjs
```diff
@@ -0,0 +1,14 @@
+import createMDX from '@next/mdx';
+
+const nextConfig = {
```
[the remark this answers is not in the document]
cdedreuille: Done

canon-docs/.gitignore
```diff
@@ -0,0 +1,36 @@
...
+
+# local env files
+.env*.local
+
+# vercel
+.vercel
+
+# typescript
+*.tsbuildinfo
+next-env.d.ts
```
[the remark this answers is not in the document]
cdedreuille: Cleaned up 🧹

.prettierignore
```diff
@@ -1,6 +1,9 @@
 .yarn
 dist
 microsite
+canon-docs/.next
+canon-docs/public
+canon-docs/out
```
Rugvip, line 6:
```suggestion
canon-docs/dist
```

canon-docs/.gitignore
```diff
@@ -0,0 +1,4 @@
+# next.js
+/.next/
+/dist/
```
Rugvip, line 3: this should already be applied project-wide

Rugvip (approved):
minor stuff so pre-approved :shipit:

PR #28318, opened 2024-12-30
cli: experimental support for using embedded-postgres as dev DB

Rugvip:

Does what is says on the tin. This is on top of #28308 because it needs support for module-only packages, and is a parallel to #28315.

Some rough benchmarking shows that this slows down the initial startup time of the backend somewhat, while restarts are about the same speed for a small DB, but noticeably faster with a large DB. Of course performance isn't really the main concern with moving in this direction as long as it's good enough. The biggest upside would be to eventually be able to drop the need for supporting SQLite and thereby simplifying DB code a lot.

### DIFF
--- a/.changeset/thin-elephants-joke.md
+++ b/.changeset/thin-elephants-joke.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+The `package start` command now supports an experimental `EXPERIMENTAL_DEV_DB` env flag that can be set to enable the use of `embedded-postgres` as the database for local development, rather than SQLite. For this to work the desired version of the `embedded-postgres` package must be installed in your project, typically as a `devDependency`.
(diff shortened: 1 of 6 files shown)


vinzscam (approved):
🚀

Rugvip:
Gonna try moving this one over to `pglite` as well, same as in #28315

Rugvip:
Tried `pglite`, it's very slow atm and the IPC times out when fetching the old DB state: [rugvip/pglite](https://github.com/backstage/backstage/tree/rugvip/pglite)

Leaning back towards this most likely being the better option unless we can find a way to re-use the same `pglite` instance across all plugins.

szubster:
Does `embedded-postgres` support postgres extensions?
Especially interested in `pgvector`, which we use in one of our internal plugins.
This now means we run postgres in a container. `pglite` does support extensions.

PR #28359, opened 2025-01-03
feat: Improve JSON format of OpenAPI definition, allow YAML format

GabDug:

Hello :wave:!

-  **Improve JSON format of OpenAPI definition, allow YAML format**
    - By default, OpenAPI definition will now have JSON with spaces instead of minified JSON, for better visualization in the UI.
    - Configuration `catalog.providers.backstageOpenapi.definitionFormat` was also added, and can take values `json` (default) or `yaml`, so organizations can use their preferred/standard format.
    - Adding spaces to the JSON _could_ be considered breaking changes.
- Moved to new PR #28364 
  - **docs: Use a valid configuration as example**
      - Previous configuration used an entity name with a space. 

Have a great day!

### DIFF
--- a/plugins/catalog-backend-module-backstage-openapi/config.d.ts
+++ b/plugins/catalog-backend-module-backstage-openapi/config.d.ts
@@ -33,7 +33,7 @@ export interface Config {
          * The format of the definition.
          * @defaultValue json
          */
-        definitionFormat: 'json' | 'yaml';
+        definitionFormat?: 'json' | 'yaml';
       };
     };
   };
--- a/plugins/catalog-backend-module-backstage-openapi/config.d.ts
+++ b/plugins/catalog-backend-module-backstage-openapi/config.d.ts
@@ -29,6 +29,11 @@ export interface Config {
          * Properties to override on the final entity object.
          */
         entityOverrides?: object;
+        /**
+         * The format of the definition.
+         * @defaultValue json
+         */
+        definitionFormat: 'json' | 'yaml';
       };
     };
   };
--- a/.changeset/nine-hounds-rest.md
+++ b/.changeset/nine-hounds-rest.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-backstage-openapi': minor
+---
+
+feat: Improve JSON format of OpenAPI definition, allow YAML format
--- a/plugins/catalog-backend-module-backstage-openapi/README.md
+++ b/plugins/catalog-backend-module-backstage-openapi/README.md
@@ -28,6 +28,7 @@ catalog:
         - catalog
         - events
         - search
+      definitionFormat: 'yaml"' # Optional, defaults to 'json'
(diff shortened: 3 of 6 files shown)


aramissennyeydd (commented):
👋 Thanks for this! Would this be a better fit as a separate catalog processor that takes an API spec and ensures that all API entities are the same format (YAML/JSON) and performs the JSON pretty-printing. I'd imagine this has been a request for other providers that return API specs, and centralizing + decoupling that logic would give you quite a bit more bang for your buck IMO, what do you think?

Either way, the doc updates are very welcome 🙇 !

GabDug:
Hey 👋 
You are right that to standardize, a processor would probably be better suited, so we may drop the YAML/JSON configuration.

However, I believe the definition should still be pretty-printed by default:  the default `EntityApiDefinitionCard` does not provide formatting and we should have a reasonable display by default, without a processor or custom Card.

Here's the current display:
Current OpenAPI display with the definition on one big JSON line

Tell me what you think about it :)

I'll split the documentation change so it can be merged sooner.

aramissennyeydd:
@GabDug Hmm, I'm ok to merge this to make this more readable for now, but I'd love to fix this more generally, as I think this is a weird edge case that affects all default `JSON.stringify` JSON definitions. YAML isn't affected because the new lines are _generally_ required. How would you feel about enabling the `wrapLongLines` option for the underlying library that renders the raw JSON? example, https://react-syntax-highlighter.github.io/react-syntax-highlighter/demo/

aramissennyeydd:
👋 I'm still interested to get this in as-is, the default JSON formatting is rough 😅

GabDug:
Hey! 

Sorry @aramissennyeydd, I have not been very active.

> How would you feel about enabling the wrapLongLines option for the underlying library that renders the raw JSON? example

This does not renders well when the JSON is minified without any line return.

I'll rebase the MR, let me know if you have other changes in mind, or if it's good enough for now.

aramissennyeydd (approved):

plugins/catalog-backend-module-backstage-openapi/config.d.ts
```diff
@@ -29,6 +29,11 @@ export interface Config {
          * Properties to override on the final entity object.
          */
         entityOverrides?: object;
+        /**
+         * The format of the definition.
+         * @defaultValue json
+         */
+        definitionFormat: 'json' | 'yaml';
```
benjdlambert, line 36: Does this need to be `definitionFormat?: 'json' | 'yaml'` otherwise it might fail schema validation?

GabDug: Good catch, thanks!

GabDug:
```suggestion
        definitionFormat?: 'json' | 'yaml';
```

benjdlambert (approved):
Nice thanks! :tada:

PR #28379, opened 2025-01-06
feat(catalog): always use search for UrlReaderProcessor

tcardonne:

Copy of PR #27054 that was closed because it was stale. There's some context and discussions in it.

> The `UrlReaderProccessor` now always calls the `search` method of the `UrlReaders`. Previous behavior was to call the `search` method only if the parsed Git URL's filename contained a wildcard and use `readUrl` otherwise. `UrlReaderService` must implement this logic in the `search` method instead.
>
> This allows each `UrlReaderService` implementation to check whether it's a search URL (that contains a wildcard pattern) or not using logic that is specific to each provider.
>
> In the different `UrlReadersService`, the `search` method have been updated to use the `readUrl` if the given URL doesn't contain a pattern.
For `UrlReaders` that didn't implement the `search` method, `readUrl` is called internally.

See comment by @Rugvip for the reasoning: https://github.com/backstage/backstage/pull/27054#discussion_r1801439398

Closes https://github.com/backstage/backstage/issues/22504

Unblocks https://github.com/backstage/backstage/pull/25704

- [X] ~Added or updated documentation~


.changeset/tiny-spoons-repeat.md
```diff
@@ -0,0 +1,15 @@
+---
+'@backstage/backend-defaults': patch
+'@backstage/plugin-catalog-backend': minor
+---
+
+**BREAKING**: The `UrlReaderProccessor` accepts a new config flag `catalog.useUrlReadersSearch` to always call the `search` method of `UrlReaders`.
```
vinzscam, line 6: this doesn't look like breaking at the moment, right? 🤔

vinzscam (changes requested):

.changeset/tiny-spoons-repeat.md
tcardonne: Yes you're right, I removed it!

tcardonne:
Hi @Rugvip @vinzscam,

Is there anything I can do to move this forward? I wasn't able to attend the last framework SIG to talk about this PR 😞 

I'd like to start adapting https://github.com/backstage/backstage/pull/28387 on top of this one but I can't start yet because we use it to patch our instance. Also it's becoming hard for us to keep them without conflicts and it makes the upgrading process a bit hard (as we have to patch 4 packages)

Thanks!

packages/backend-defaults/src/entrypoints/urlReader/lib/AwsS3UrlReader.ts
```diff
@@ -357,8 +362,33 @@ export class AwsS3UrlReader implements UrlReaderService {
...
   }
 
-  async search(): Promise<UrlReaderServiceSearchResponse> {
-    throw new Error('AwsS3Reader does not implement search');
+  async search(
+    url: string,
+    options?: UrlReaderServiceSearchOptions,
+  ): Promise<UrlReaderServiceSearchResponse> {
+    try {
+      const data = await this.readUrl(url, options);
```
vinzscam, line 370: I wonder if we should throw an error if we detect a non-direct URL here. Otherwise, it will return zero files, which is incorrect. The same applies to other readers who didn't implement search previously.

```suggestion
    if (isSearchUrl(url)) {
      throw new Error('Unsupported url');
    }
    try {
      const data = await this.readUrl(url, options);
```

vinzscam (changes requested):
apologies for the waiting time @tcardonne. Let's try to ship this asap

packages/backend-defaults/src/entrypoints/urlReader/lib/AwsS3UrlReader.ts
tcardonne: Yes I guess you're right @vinzscam!

I updated the readers for AWS Code Commit, AWS S3, Fetch, Gerrit, Gitea and Harness

.changeset/tiny-spoons-repeat.md
```diff
@@ -0,0 +1,15 @@
+---
+'@backstage/backend-defaults': patch
```
Rugvip, line 2: Could you split up this changeset with one for `backend-defaults` and one for `plugin-catalog-backend`? only mentioning the changes that apply to each of those pacakges?

plugins/catalog-backend/src/processors/UrlReaderProcessor.test.ts
```diff
@@ -31,6 +31,8 @@ import {
 import { defaultEntityDataParser } from '../util/parse';
 import { UrlReaderProcessor } from './UrlReaderProcessor';
 import { UrlReaders } from '@backstage/backend-defaults/urlReader';
+import { UrlReaderService } from '@backstage/backend-plugin-api';
+import { mockApis } from '@backstage/test-utils';
```
Rugvip, line 35: Use the config mock from `mockServices` from `@backstage/backend-test-utils` instead

plugins/catalog-backend/src/processors/UrlReaderProcessor.ts
```diff
@@ -47,14 +48,25 @@ export class UrlReaderProcessor implements CatalogProcessor {
...
     },
   ) {
     this.#limiter = limiterFactory(5);
+
+    this.#useUrlReadersSearch =
+      this.options.config?.getOptionalBoolean('catalog.useUrlReadersSearch') ||
+      false;
+    if (!this.#useUrlReadersSearch) {
+      this.options.logger.warn(
+        'UrlReaderProcessor uses the legacy readUrl/search behavior which will be removed in a future release. Set catalog.useUrlReadersSearch to true to adopt the new behavior.',
```
Rugvip, line 67: 👍 

Given how strong of a recommendation this is, let's update the `app-config.yaml` that ships with `@backstage/create-app` to set this flag as well, otherwise we ship with this warning out of the box:

https://github.com/backstage/backstage/blob/612817b0ce5cd945cc7a9ca16a73f478d8aa296c/packages/create-app/templates/default-app/app-config.yaml.hbs#L74

Rugvip (changes requested):
Awesome stuff! 🎉 👌 

Just a few small nits and we're good to ship

.changeset/tiny-spoons-repeat.md
tcardonne: Done 👍

plugins/catalog-backend/src/processors/UrlReaderProcessor.ts
tcardonne: Done 👍 

I added a small comment to give better context on the option

Rugvip (approved):
Awesome stuff, thank you! 😁 🎉 

:shipit:

PR #28391, opened 2025-01-07
Canon - Add 'use client' directive when needed

cdedreuille:
This is a small one but some components are using hooks and in order to make them work with React 19 we have to add the 'use client' directive.


Rugvip (commented):
no need for this anymore right? since we're not aiming for React 19 atm

cdedreuille:
> no need for this anymore right? since we're not aiming for React 19 atm

We still need it as these directives are needed on React 18 as well. This is still good practice to have them available if we are using hooks.

Rugvip (approved):
alright :shipit:

PR #28393, opened 2025-01-07
yarn-plugin: include entries for both `backstage:` and `npm:` ranges in lockfile

mtlewis:

Unfortunately the recent changes in https://github.com/backstage/backstage/pull/28130 reintroduced the issue previously fixed by https://github.com/backstage/backstage/pull/27727. It turns out that we actually need _two_ ranges present in the lockfile when using the plugin:
* A `backstage:^` range, which allows tools like the Snyk lockfile parser to match packages in `package.json` with entries in the lockfile.
* The corresponding `npm:` range, ensuring that package versions don't get unlocked when translating `backstage:^` ranges during npm pack (in particular when building dist workspaces).

This PR adds a failing test to confirm the issue with lockfile parsing in the current plugin implementation, then resolves the issue by introducing a new resolver to add the appropriate `npm:`-range-based dependencies to packages using `backstage:` ranges.

As part of this change, I've also documented the current architecture of the plugin to aid in future maintenance of this system.

### DIFF
--- a/.changeset/wise-pillows-smile.md
+++ b/.changeset/wise-pillows-smile.md
@@ -1,0 +1,5 @@
+---
+'yarn-plugin-backstage': patch
+---
+
+Add both `npm:` and `backstage:` ranges to the lockfile to ensure compatibility with tools that parse the lockfile and ensure dependencies stay locked when building dist workspaces.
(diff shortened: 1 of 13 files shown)


mtlewis:
Just flagging a couple of potential follow-ups that I sadly wasn’t able to get to in this PR:

1. Test suite for the `BackstageNpmResolver`. It’s already tested indirectly by the entry point test suite, but adding a dedicated suite would let us cover more edge cases.
2. Extra entry point test for dependency deduping. There are a couple of code paths which are only hit during dedupe, so it would be good to cover it in the tests.

benjidotsh:
Not stale - we still need this PR to get the plugin to work with Snyk 😅

vinzscam:
alright PR is green 🎉
I had to add a patch https://github.com/backstage/backstage/pull/28393/commits/a0cfdfa866b15f56c632e374411d7641fa745b4e to fix an issue in `@yarnpkg/plugin-npm` and raised upstream: https://github.com/yarnpkg/berry/pull/6764

Rugvip (approved):
Couple bits but pre-approved 👍

PR #28419, opened 2025-01-09
Updated docs of versioning policy

AmbrishRamachandiran:
Updated docs of versioning policy full stop is missing in one of the sentence


benjdlambert:
Just a quick note that it would be appreciated if you could collect a lot of these small changes into a single PR rather than creating super tiny ones. There's a lot of overhead that goes into reviewing PR's so would be nice if we didn't have to go through lots of small tiny doc changes. :pray:

benjdlambert (approved):

PR #27028, opened 2024-10-08
Pass region to getDefaultCredentialsChain

KaemonIsland:
### Summary of Changes

This pull request updates the `getDefaultCredentialsChain` function to accept a `region` parameter, passed via the configuration, which sets the region for `fromNodeProviderChain`. This change prevents the region from defaulting to `us-east-1` when `getDefaultCredentialsChain` is called without specifying a region.

### Why This Change is Needed

After upgrading to the new Backend System, we encountered connectivity issues with our S3 bucket. Our configuration file (`app-config.yaml`) is straightforward, with environment variables managing settings like ARN, REGION, and BUCKET. This setup worked with the previous backend, but following the migration guide for the upgrade, connections started failing.

Here’s our current `app-config.yaml`:

```yaml
techdocs:
  builder: 'external'
  publisher:
    awsS3:
      bucketName: ${AWS_BUCKET_NAME}
      region: ${AWS_REGION}
      roleArn: ${AWS_ROLE_ARN}
    type: 'awsS3'
aws:
  mainAccount:
    region: ${AWS_REGION}
    roleArn: ${AWS_ROLE_ARN}
```

Upon investigating, I found [this related issue](https://github.com/backstage/backstage/issues/16884), which describes a similar problem. Implementing the suggested changes resolved our issue, so I’m proposing this update to incorporate the fix directly. This way, we avoid importing additional files to modify a few lines of code, which would also ensure compatibility for other users.

### Technical Context

As noted in the [@aws-sdk/credential-provider-node](https://www.npmjs.com/package/@aws-sdk/credential-provider-node) documentation, a region must be explicitly passed if it’s not `us-east-1`. If omitted, the SDK defaults to using `us-east-1` as the region:

```js
const { getDefaultRoleAssumerWithWebIdentity } = require("@aws-sdk/client-sts");
const { defaultProvider } = require("@aws-sdk/credential-provider-node");
const { S3Client, GetObjectCommand } = require("@aws-sdk/client-s3");

const provider = defaultProvider({
  roleAssumerWithWebIdentity: getDefaultRoleAssumerWithWebIdentity({
    // Explicit region passing required for non-us-east-1
    region: "eu-west-1"
  }),
});
```

The key takeaway is that the region must be explicitly passed unless you're using `us-east-1`. Importantly, the AWS SDK now provides a wrapper for this in the `@aws-sdk/credential-providers` package, which simplifies the process by eliminating the need to import individual methods like `getDefaultRoleAssumerWithWebIdentity()`.

[...]

PR #28420, opened 2025-01-09
Fix aws.mainAccount account ID check during fallback in DefaultAwsCredentialsManager

cflee:

Given that:

- there is no entry within `aws.accounts[]` with accountId that matches the query, and 
- the `aws.mainAccount.region` config field is not provided, 

When calling `DefaultAwsCredentialsManager#getCredentialProvider()` with an accountId/arn to lookup a credential provider, 

Then a `Region is missing` error will be returned instead of the expected `There is no AWS integration that matches ${accountId}. Please add a configuration for this AWS account.` error. 

This is confusing when troubleshooting config issues.

This unexpected error comes from sending a GetCallerIdentityCommand to STSClient to retrieve the main account's account ID, but the call fails due to an undefined region being provided to the STSClient. There is no way to provide or configure this region, as `mainAccountCredProvider.stsRegion` is always unset.

There appear to be tests that cover this behavior, but this issue is hidden by the STSClient mock always returning a successful value for GetCallerIdentityCommand calls, even when an undefined region is not provided to the STSClient constructor.

Key changes:

- Set a default region of `us-east-1` when calling STS to check the account ID of the main account when considering whether to fall back to the main account during credential provider lookup. This aligns the behavior with the existing implementation of `getDefaultCredentialsChain()`. If the call to STS completes and confirms that the account id is not the desired one, then the "no AWS integration that matches accountId" error message will be returned as expected.
- Set the region for calling STS from the `aws.mainAccount.region` config field, similar to the existing implementation for reading `aws.accounts[].region` and passing it to `fromNodeProviderChain()` #27028 

### DIFF
--- a/.changeset/slow-drinks-enjoy.md
+++ b/.changeset/slow-drinks-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@backstage/integration-aws-node': patch
+---
+
+Fixed bug in DefaultAwsCredentialsManager where aws.mainAccount.region has no effect on the STS region used for account ID lookup during credential provider lookup when falling back to the main account, and it does not default to us-east-1
--- a/packages/integration-aws-node/src/DefaultAwsCredentialsManager.test.ts
+++ b/packages/integration-aws-node/src/DefaultAwsCredentialsManager.test.ts
@@ -500,5 +500,21 @@ describe('DefaultAwsCredentialsManager', () => {
         },
       });
     });
+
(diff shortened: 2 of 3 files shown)


vinzscam (approved):
thank you!

cflee:
Hmm, I don't think this PR is stale, just waiting for an approval. Please let me know if there's something that is waiting on me.

PR #28423, opened 2025-01-09
root: switch foreach scripts to --jobs unlimited

Rugvip:
🧹, quick followup for #28296, switching to `--jobs` since I find that more informative. I also found `unlimited` to be more semantically correct rather than `2`, and safe in case more packages are added.


awanlin (approved):
Thanks for this @Rugvip, I had settled on `-j 2` as the safe route but I do think this is better and more clear 👍

benjdlambert (approved):

PR #28433, opened 2025-01-10
cfaccess: fallback to identity email when constructing the user profile

tylerd-canva:

Noticed this while updating our dependencies and setting up some Service Tokens -- they don't have an `email` in their claims so the current profile transform doesn't work properly.

### DIFF
--- a/plugins/auth-backend-module-cloudflare-access-provider/src/authenticator.ts
+++ b/plugins/auth-backend-module-cloudflare-access-provider/src/authenticator.ts
@@ -38,7 +38,7 @@ export function createCloudflareAccessAuthenticator(options?: {
     async defaultProfileTransform(result: CloudflareAccessResult) {
       return {
         profile: {
-          email: result.claims.email ?? result.cfIdentity.email,
+          email: result.cfIdentity.email,
           displayName: result.cfIdentity.name,
         },
       };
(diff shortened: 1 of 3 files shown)


plugins/auth-backend-module-cloudflare-access-provider/src/authenticator.ts
```diff
@@ -38,7 +38,7 @@ export function createCloudflareAccessAuthenticator(options?: {
     async defaultProfileTransform(result: CloudflareAccessResult) {
       return {
         profile: {
-          email: result.claims.email,
+          email: result.claims.email ?? result.cfIdentity.email,
```
Rugvip, line 41: Will `cfIdentity` always contain the correct email, should we use that instead? Always nice to be explicit in this area 😁

Rugvip (commented):
Thank you! just a tiny Q

plugins/auth-backend-module-cloudflare-access-provider/src/authenticator.ts
tylerd-canva: Good point. As far as I can tell, yes, the `cfIdentity` struct will contain the email in both cases so that's simpler.

benjdlambert (approved):
Nice thanks! :tada:

Issue #22244, opened 2024-01-15
💬 RFC: Simplify and Standardize the approach for Pull Request and Merge Requests in Scaffolder Actions


benjdlambert:
### 🔖 Need

We have a large amount of actions that deal with upstream VCS providers like GitHub, Bitbucket, GitLab etc. There's the ability to create Pull Requests / Merge Requests from Branches, and these actions implement this in different ways. Some of the VCS providers implement standlone rest API's / GraphQL API's to be able to create Pull Requests or Merge Requests without checking out the code and operating with the Git Protocol, you can create commits with file changes through the API which in turn becomes a PR (looking at you GitHub).

This can lead to some issues like https://github.com/backstage/backstage/issues/17188, and sadly sometimes this approach is not available with the provider API's. (looking at you Bitbucket / GitLab).

Of course we want this functionality to be available for as many providers as possible, so thinking that theres chance to align the approach for all providers.

### 🎉 Proposal

We take advantage of the `isomorphic-git` library and prefer to use native git protocols instead of using the provider API's where this makes sense.

PR/MR actions will follow the following workflow:

- Clone the repository
- Create a Branch
- Make the required changes
  - Add/Remove/Edit files
- Create a commit
- Push the commit to the branch
- Use the provider specific API's to create a PR from the HEAD of the changed branch towards the base branch

We should be able to make this a non-breaking change, as we should have all the required permissions* and inputs required to make this a reality.

### 〽️ Alternatives

Open for other suggestions here, but we could just leave it as is today and accept the fact that bigger pull requests could fail because of rate limits, and different providers have different abilities.

### ❌ Risks

* This approach could be much slower on bigger repositories and for smaller changes. Thinking that we could lean towards falling back to this approach if there's a count of file changes above a certain threshold. Maybe we can just play this by ear though.
* Need to confirm that the user token permissions should be enough as is what happens today. Just want to make sure that we don't need to create forks of repositories, because that could be unfortunate.

### 👀 Have you spent some time to check if this RFC has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

acierto:
@benjdlambert What do you think about including here also idempotency and aborting functionality?

benjdlambert:
[...]

PR #28440, opened 2025-01-10
Support empty repo and skip creation of a repository if it already exists

### LINKED ISSUE #28427: 🚀 Feature: Possible publish:gitlab improvements
### 🔖 Feature description

We have 2 different usecase which are at this moment not completely supported by `publish:gitlab` but could be with simple changes.

- The action always expects a sourcePath or uploads the entire workspace. 
I propose to **add an input `initialize_repo_with_workspace`** which receives a boolean (default true to keep backwards compatibility). 
This would bypass the sourcePath/workspace usage to be uploaded. It allows the usage to be created empty. This would also allow the usage of the `initialize_with_readme` flag to only create a README.md file.
- To have a less conditional templating/steps, it would help us if this action would support an existence check before trying the create. Similar to `gitlab:group:ensureExists` action we where thinking to create a `gitlab:project:ensureExists`, but it would be 95% copy of this action or calling this action internally. We propose to **add an input `skip_existing`** that if it is set to true will first do a check. If the project already exists, do not do an action. A new output `created` could indicate if creation was needed or not. If skip_existing (=default) is false and the repo exists already we keep the current behaviour of a failing template.

### 🎤 Context

**initialize_repo_with_workspace:**
I think this better supports the gitlab rest api and makes the action more flexible without compromising a lot.

**skip_existing:**
We have a k8s-config repository containing kustomize components/files/... per domain. 
In this repo there is 1 folder per system normally. Not all domains have apps/systems -> not all domains need a k8s-config repo.
Whenever we create a new system in a domain, we could push the default components/overlays creating the domain configuration files. After this action, we would know for sure the default config repo exists and we can follow up with a MR action to create the system configuration files through MR. 

- add 2 boolean flags `initialize_repo_with_workspace` and `skip_existing` to the input
- add 1 output `created`
```typescript
...
+const result = await client.Groups.allProjects(owner, { search: repo });
+
+const project = result.find(
+    searchPathElem => searchPathElem.path === repo,
+);

+if (!skip_existing || (skip_existing && !project)) {
     ctx.logger.info(`Creating repo ${repo} in namespace ${owner}.`);
     const { id: projectId, http_url_to_repo } = await client.Projects.create({
        namespaceId: targetNamespaceId,
        name: repo,
        visibility: repoVisibility,
        ...(topics.length ? { topics } : {}),
        ...(Object.keys(settings).length ? { ...settings } : {}),
      });
     ...
+     if (initialize_repo_with_workspace) {
         const commitResult = await initRepoAndPush({
            dir: getRepoSourceDirectory(ctx.workspacePath, ctx.input.sourcePath),
            remoteUrl: http_url_to_repo as string,
            defaultBranch,
            auth: {
              username: 'oauth2',
              password: token,
            },
            logger: ctx.logger,
            commitMessage: gitCommitMessage ? gitCommitMessage : config.getOptionalString('scaffolder.defaultCommitMessage'),
            gitAuthorInfo,
          });
 +         ctx.output('commitHash', commitResult?.commitHash);
 +     }
      ...
      ctx.output('remoteUrl', remoteUrl);
      ctx.output('repoContentsUrl', repoContentsUrl);
      ctx.output('projectId', projectId);
      ctx.output('created', true);
+} else if (skip_existing && project) {
+    ctx.logger.info(`Repo ${repo} already exists in namespace ${owner}.`);
+    const { id: projectId, http_url_to_repo, default_branch } = project;
+    const remoteUrl = (http_url_to_repo as string).replace(/\.git$/, '');
+    ctx.output('remoteUrl', remoteUrl);
+    ctx.output('repoContentsUrl', `${remoteUrl}/-/blob/${default_branch}`);
+    ctx.output('projectId', projectId);
+    ctx.output('created', false);
+}
 ```
 
We did not provide a PR immediately as we first want to know if there would be value, what path we should take and we do not have a lot knowledge (yet?) in writing the corresponding test cases.

### PR DESCRIPTION
**Pull request by Jenson3210**

Fixes #28427

### DIFF
--- a/plugins/scaffolder-backend-module-gitlab/src/actions/gitlab.ts
+++ b/plugins/scaffolder-backend-module-gitlab/src/actions/gitlab.ts
@@ -43,7 +43,8 @@ export function createPublishGitlabAction(options: {
     defaultBranch?: string;
     /** @deprecated in favour of settings.visibility field */
     repoVisibility?: 'private' | 'internal' | 'public';
-    sourcePath?: string;
+    sourcePath?: string | boolean;
+    skipExisting?: boolean;
     token?: string;
     gitCommitMessage?: string;
     gitAuthorName?: string;
@@ -124,8 +125,14 @@ export function createPublishGitlabAction(options: {
           sourcePath: {
             title: 'Source Path',
             description:
-              'Path within the workspace that will be used as the repository root. If omitted, the entire workspace will be published as the repository.',
-            type: 'string',
+              'Path within the workspace that will be used as the repository root. If omitted or set to true, the entire workspace will be published as the repository. If set to false, the created repository will be empty.',
+            type: ['string', 'boolean'],
+          },
+          skipExisting: {
+            title: 'Source Path',
+            description:
+              'Do not publish the repository if it already exists. The default value is false.',
+            type: ['boolean'],
           },
           token: {
             title: 'Authentication Token',
@@ -321,6 +328,10 @@ export function createPublishGitlabAction(options: {
             title: 'The git commit hash of the initial commit',
             type: 'string',
           },
+          created: {
+            title: 'Whether the repository was created or not',
+            type: 'boolean',
+          },
         },
       },
     },
@@ -337,6 +348,7 @@ export function createPublishGitlabAction(options: {
         settings = {},
         branches = [],
         projectVariables = [],
+        skipExisting = false,
       } = ctx.input;
       const { owner, repo, host } = parseRepoUrl(repoUrl, integrations);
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: plugins/scaffolder-backend-module-gitlab/src/actions/gitlab.ts [hunk 1]
```diff
@@ -124,8 +125,14 @@ export function createPublishGitlabAction(options: {
           sourcePath: {
             title: 'Source Path',
             description:
-              'Path within the workspace that will be used as the repository root. If omitted, the entire workspace will be published as the repository.',
-            type: 'string',
+              'Path within the workspace that will be used as the repository root. If omitted or set to true, the entire workspace will be published as the repository. If set to false, the created repository will be empty.',
```
benjdlambert, line 128: I think this is fine as it is to merge, but I'd really encourage the use of a `gitlab:repo:create` action instead here to create an empty repo rather than using this action to do it.

Maybe something to think about when #22244 comes up

plugins/scaffolder-backend-module-gitlab/src/actions/gitlab.ts
```diff
@@ -124,8 +125,14 @@ export function createPublishGitlabAction(options: {
...
+            type: ['string', 'boolean'],
+          },
+          skipExisting: {
+            title: 'Source Path',
```
benjdlambert, line 132:
```suggestion
            title: 'Skip if repository exists',
```

plugins/scaffolder-backend-module-gitlab/src/actions/gitlab.ts
Jenson3210: Is this action part of an external package? I do not find such an action. 
Or you meant to say that we should have raised a PR that creates this actions?

Jenson3210: sounds good!

Jenson3210: I think then this one is also good to go, or are we waiting for other reviews here? 
Just to know if I prepare my 1.35.0 bump by removing our custom version of this one.

benjdlambert (approved):
Nice - let's go with this for now, really wanna get round to refactoring some of these actions to clean them up a little bit as they're becoming pretty big blocks of code in all places and pretty hard to maintain.

Thanks!

Jenson3210:
@GabDug after looking a bit into the api this is indeed going to cause issues in your case then. On our side we do not use personal projects, so I "forgot" about them.

PR #28444, opened 2025-01-12
refactor: replace @keyv/redis with @keyv/valkey

### LINKED ISSUE #27397: 🚀 Feature: Cache Service support for Valkey
### 🔖 Feature description

Valkey is an open source drop-in replacement for Redis (which Backstage already supports via `keyv`). Keyv has now added a [dedicated package](https://www.npmjs.com/package/@keyv/valkey) for Valkey support so it would be good to add support for this in Backstage as well.

### 🎤 Context

Since Redis is no longer open source, a lot of major vendors are moving support into Valkey instead. At this stage Valkey is a drop-in replacement, but we can assume that in the future the feature sets will diverge. By adding support now, it can be added as a simple clone of the existing Redis implementation and the supported features can evolve with in line with the product rather than being a more complex integration later. It also assists those who have moved their infrastructure over to running Valkey since they can trust that there will not be an unexpected issue arise from changes in the Redis/Valkey feature parity status.

Should be almost a copy+paste job on the existing Redis support.

Lucifergene:

Replaced @keyv/redis package with @keyv/valkey since Redis is no longer open-source and Valkey is the drop-in replacement.

Closes #27397 

### DIFF
--- a/.changeset/blue-actors-compare.md
+++ b/.changeset/blue-actors-compare.md
@@ -3,4 +3,4 @@
 '@backstage/backend-defaults': patch
 ---
 
-Replace @keyv/redis package with @keyv/valkey since Redis is no longer open-source and Valkey is the drop-in replacement for it.
+Replace `@keyv/redis` package with `@keyv/valkey` since Redis is no longer open-source and Valkey is the drop-in replacement for it.
--- a/.changeset/blue-actors-compare.md
+++ b/.changeset/blue-actors-compare.md
@@ -1,0 +1,6 @@
+---
+'@backstage/backend-test-utils': patch
+'@backstage/backend-defaults': patch
+---
+
+Replace @keyv/redis package with @keyv/valkey since Redis is no longer open-source and Valkey is the drop-in replacement for it.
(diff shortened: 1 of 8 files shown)


benjdlambert (approved):
Nice! Thanks for this! :pray:

benjdlambert:
I'm a little concerned with what's going to happen here now that we've had to remove this line in order to get the tests to pass: https://github.com/backstage/backstage/pull/28444/files#diff-42975462070406316e4534ce0579d1d12d54fc5cd62e239d1ae676c1e290b473L161

It looks like if we don't supply a `keyPrefix` then all keys are shared between plugins which is wrong anyways.

This is probably gonna be shipped after the mainline release today, as it's a bit of a risk, but @Lucifergene can you check to see what the keys look like if you apply the latest changes locally? Is the upgrade to this package effectively going to be a cache buster too so that all the keys have different names?

benjdlambert:
@Lucifergene sorry, accidentally automerged this. Wanna move the conversation over to #28465 instead.

PR #28445, opened 2025-01-12
Canon - Add custom theme capability

cdedreuille:
I'm adding the capability to add your custom theme to our docs. Everything is saved to local storage and you can see your changes on every pages. For better visibility you can also reduce the editor. This will come back as normal if you refresh.

This PR also fix some eslint issues and add some extra metadata.

https://github.com/user-attachments/assets/2f641bef-0e9c-4137-8d18-83c2ee1c5942

### DIFF
--- a/canon-docs/src/components/CustomTheme/customTheme.tsx
+++ b/canon-docs/src/components/CustomTheme/customTheme.tsx
@@ -7,10 +7,43 @@ import styles from './styles.module.css';
 import { usePlayground } from '@/utils/playground-context';
 import { AnimatePresence, motion } from 'framer-motion';
 import { Icon } from '../../../../packages/canon';
+import { createTheme } from '@uiw/codemirror-themes';
+import { tags as t } from '@lezer/highlight';
 
 const defaultTheme = `:root {
   --canon-accent: #000;
 }`;
+
+const myTheme = createTheme({
+  theme: 'light',
+  settings: {
+    background: 'var(--canon-surface-1)',
+    backgroundImage: '',
+    foreground: '#6182B8',
+    caret: '#5d00ff',
+    selection: '#036dd626',
+    selectionMatch: '#036dd626',
+    lineHighlight: '#8a91991a',
+    gutterBackground: '#fff',
+    gutterForeground: '#8a919966',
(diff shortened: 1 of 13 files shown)


canon-docs/src/components/CustomTheme/customTheme.tsx
```diff
@@ -0,0 +1,147 @@
...
+    if (customTheme) {
+      localStorage.setItem('customThemeCss', customTheme);
+      updateStyleElement(customTheme);
+      setSavedMessage('Saved!');
+      setTimeout(() => setSavedMessage('Save'), 1000);
+    }
+  };
+
+  const handleChange = useCallback((val: string) => {
+    setCustomTheme(val);
```
vinzscam, line 102: shouldn't this invoke `setSelectedTheme` from `usePlayground` instead?

canon-docs/src/components/CustomTheme/customTheme.tsx
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }, [selectedThemeName]);
+
+  useEffect(() => {
+    setIsClient(true);
+  }, []);
+
+  const handleSave = () => {
+    if (customTheme) {
+      localStorage.setItem('customThemeCss', customTheme);
```
vinzscam, line 94: perhaps we can move any `localStorage` interaction to the playground context so that everything is centralized

canon-docs/src/components/CustomTheme/customTheme.tsx
cdedreuille: No, `customTheme` is the actual css we get from localStorage. `selectedTheme` is just the name of the theme `light` or `dark`.

canon-docs/src/components/CustomTheme/customTheme.tsx
cdedreuille: I'm trying to avoid putting everything to the context. This logic only applies to this component with no need in other parts of the code. For now I think it is best to keep it within this component unless we want to reuse it elsewhere.

Rugvip (approved):
ye

PR #28449, opened 2025-01-13
Harmonize `CatalogTable`

Andy2003:

Harmonize `CatalogTable` by: 

- Show pagination text for `OffsetPagination`
- Use same `OffsetPaginatedCatalogTable` also as fallback if no pagination is set
- Do not show paging if there is only one page

### DIFF
--- a/.changeset/funny-papayas-tell.md
+++ b/.changeset/funny-papayas-tell.md
@@ -1,0 +1,9 @@
+---
+'@backstage/plugin-catalog': minor
+---
+
+Harmonize `CatalogTable`
+
+- Show pagination text for `OffsetPagination`
+- Use same `OffsetPaginatedCatalogTable` also as fallback if no pagination is set
+- Do not show paging if there is only one page
--- a/.changeset/funny-papayas-tell.md
+++ b/.changeset/funny-papayas-tell.md
@@ -2,7 +2,4 @@
 '@backstage/plugin-catalog': minor
 ---
 
-Harmonize `CatalogTable`
-
-- Show pagination text for `OffsetPagination`
-- Do not show paging if there is only one page
+Show the pagination text for the offset-paginated catalog table, and remove the pagination bar from the top of the `CatalogTable` when pagination is enabled.
--- a/.changeset/funny-papayas-tell.md
+++ b/.changeset/funny-papayas-tell.md
@@ -5,5 +5,4 @@
 Harmonize `CatalogTable`
 
 - Show pagination text for `OffsetPagination`
-- Use same `OffsetPaginatedCatalogTable` also as fallback if no pagination is set
 - Do not show paging if there is only one page
--- a/plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
+++ b/plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
@@ -236,7 +236,6 @@ export const CatalogTable = (props: CatalogTableProps) => {
       subtitle={subtitle}
       options={options}
       data={rows}
-      clientPagination={paginationMode === 'none'}
     />
   );
 };
--- a/plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
+++ b/plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
@@ -23,6 +23,7 @@ import {
 } from '@backstage/catalog-model';
 import {
   CodeSnippet,
+  FavoriteToggleIcon,
   TableColumn,
   TableProps,
   WarningPanel,
@@ -47,7 +48,6 @@ import { CursorPaginatedCatalogTable } from './CursorPaginatedCatalogTable';
 import { defaultCatalogTableColumnsFunc } from './defaultCatalogTableColumnsFunc';
 import { useTranslationRef } from '@backstage/core-plugin-api/alpha';
 import { catalogTranslationRef } from '../../alpha';
-import { FavoriteToggleIcon } from '@backstage/core-components';
 
 /**
  * Props for {@link CatalogTable}.
@@ -236,6 +236,7 @@ export const CatalogTable = (props: CatalogTableProps) => {
       subtitle={subtitle}
       options={options}
       data={rows}
+      clientPagination={paginationMode === 'none'}
     />
   );
 };
--- a/plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
+++ b/plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
@@ -23,7 +23,6 @@ import {
 } from '@backstage/catalog-model';
 import {
   CodeSnippet,
-  Table,
   TableColumn,
   TableProps,
   WarningPanel,
@@ -47,7 +46,7 @@ import { OffsetPaginatedCatalogTable } from './OffsetPaginatedCatalogTable';
 import { CursorPaginatedCatalogTable } from './CursorPaginatedCatalogTable';
 import { defaultCatalogTableColumnsFunc } from './defaultCatalogTableColumnsFunc';
 import { useTranslationRef } from '@backstage/core-plugin-api/alpha';
-import { catalogTranslationRef } from '../../alpha/translation';
+import { catalogTranslationRef } from '../../alpha';
 import { FavoriteToggleIcon } from '@backstage/core-components';
 
 /**
@@ -194,13 +193,20 @@ export const CatalogTable = (props: CatalogTableProps) => {
       .join(' ');
 
   const actions = props.actions || defaultActions;
-  const options = {
+  const options: TableProps['options'] = {
+    paginationPosition: 'both',
     actionsColumnIndex: -1,
     loadingType: 'linear' as const,
     showEmptyDataSourceMessage: !loading,
     padding: 'dense' as const,
     ...tableOptions,
   };
+
+  if (paginationMode !== 'cursor' && paginationMode !== 'offset') {
+    entities.sort(refCompare);
+  }
+
+  const rows = entities.map(toEntityRow);
 
   if (paginationMode === 'cursor') {
     return (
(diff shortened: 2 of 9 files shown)


benjdlambert:
@vinzscam do you wanna have a peek through this? :pray:

plugins/catalog/src/components/CatalogTable/OffsetPaginatedCatalogTable.tsx
```diff
@@ -64,7 +66,6 @@ export function OffsetPaginatedCatalogTable(
         setLimit(pageSize);
       }}
       totalCount={totalItems}
-      localization={{ pagination: { labelDisplayedRows: '' } }}
```
vinzscam, line 67: are you sure the pagination text would be correct? I think it was removed because the data was wrongly interpreted by the underlying table.

vinzscam (changes requested):

plugins/catalog/src/components/CatalogTable/OffsetPaginatedCatalogTable.tsx
Andy2003: The cursor-based table was separated from the offset-based table in commit 78475c33a4842c1a09f0be4fa2dc236eb7534054. It makes no sense to display the paging text for the cursor table. However, it makes sense for the offset. The removal was simply forgotten at the time.

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
```diff
@@ -213,48 +218,24 @@ export const CatalogTable = (props: CatalogTableProps) => {
...
 
-  const rows = entities.sort(refCompare).map(toEntityRow);
-  const pageSize = 20;
-  const showPagination = rows.length > pageSize;
-
+  // else use offset paging
   return (
-    <Table<CatalogTableRow>
-      isLoading={loading}
+    <OffsetPaginatedCatalogTable
```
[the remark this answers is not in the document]
Andy2003: There are only cursor or offset based tables. Even if paging-mode is not explicitly set, the offset variant is used at the end. The separation, as it was until now, only makes the table behave differently in the case that paging is explicitly activated than if no explicit paging mode is specified.

plugins/catalog/src/components/CatalogTable/OffsetPaginatedCatalogTable.tsx
vinzscam: the Offset table contains some customization to make it work with server side paginated data. The normal `<Table />` uses client side paginated data, which behaves differently.

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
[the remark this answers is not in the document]
vinzscam: the Offset table contains some customization to make it work with server side paginated data. The normal `<Table />` uses client side paginated data, which behaves differently.

plugins/catalog/src/components/CatalogTable/OffsetPaginatedCatalogTable.tsx
Andy2003: Yes I've tested it:

image

image

see also: https://github.com/material-table-core/core/blob/master/src/components/MTablePagination/index.js#L107-L116

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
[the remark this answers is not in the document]
Andy2003: I've tested both with pagination mode `offset` and `none`, they behave as expected:

- `offset`: server side paging
  
  image

- `none`: client side paging

  image

plugins/catalog/src/components/CatalogTable/OffsetPaginatedCatalogTable.tsx
```diff
@@ -28,43 +28,49 @@ export function OffsetPaginatedCatalogTable(
...
-      page={page}
-      onPageChange={newPage => {
-        setPage(newPage);
-      }}
-      onRowsPerPageChange={pageSize => {
-        setLimit(pageSize);
-      }}
-      totalCount={totalItems}
-      localization={{ pagination: { labelDisplayedRows: '' } }}
+      {...(clientPagination
```
vinzscam, line 66: I would prefer to keep a duplicated `<Table />` as it was before as it's make things easier if more things need to be adjusted.

vinzscam (changes requested):
going through the proposed changes one by one:

> `- Show pagination text for `OffsetPagination`

  I think this is great

> `- Use same `OffsetPaginatedCatalogTable` also as fallback if no pagination is set
`

I feel like we should keep things as they are as this won't fix any bug or reduce any complexity. Defining an extra `<Table />` in `CatalogTable` allows us to better split the logic for server side and client side paginated table.

> `- Do not show paging if there is only one page`

if it's possible to change the number of items per page via the UI we should always show the paging information (right now if a user select to display a higher number of items per page than the items that are available, it won't be possible to change the limit anymore).

plugins/catalog/src/components/CatalogTable/OffsetPaginatedCatalogTable.tsx
Andy2003: I disagree, the separation has already led to the `CatalogTable` behaving differently several times, e.g:

- currently: missing display of the pagination text
- https://github.com/backstage/backstage/pull/27129 Here, for example, the toolbar was not built into ?`CatalogTable`, which was also fixed in the next PR
- Missing title https://github.com/backstage/backstage/pull/27632

vinzscam: I agree that this led us to miss some common props, but the advantage has been that we have introduced 3 different ways of managing and displaying data without affecting each other. Now we have a more complex `OffsetPaginatedCatalogTable` that handles server side and client side logic. How can we make sure that this change won't break the existing logic?  Also in the future the goal is to sunset the client side paginated table, which is pretty straight forward at the moment as the logic isn't mixed with any of the two new tables.

Andy2003: > I agree that this led us to miss some common props, but the advantage has been that we have introduced 3 different ways of managing and displaying data without affecting each other.

I don't see it as an advantage that the server and client-side paging were handled in different places, as explained above, this has led to inconsistencies in the code on several occasions. With my adjustments, the paging behavior is now harmonized and always the same, whether server or client side.

> Now we have a more complex OffsetPaginatedCatalogTable that handles server side and client side logic. How can we make sure that this change won't break the existing logic?

It's just an refactoring, that is even reducing the total lines of code. The added "complexity" is just an early return in the hook and a ternary for setting the server side config on the Table.  You can simply compare the previous invocation of the client-side-paged table:

image

with the now called code of the `OffsetPaginatedCatalogTable`

image

the final call of `Table` is ultimately exactly the same as it was before, with the following adjustments:

- fix of paging text
- harmonized pages size selection (options for 5 and 10 are added)
- always visibility of the paging bar as requested by you in https://github.com/backstage/backstage/pull/28449#pullrequestreview-2559283912

> Also in the future the goal is to sunset the client side paginated table, which is pretty straight forward at the moment as the logic isn't mixed with any of the two new tables.

It is still easy to remove the client-side paging, because you can explicitly find the paging mode `none` here (which was not the case before). And in the end you only have to remove `clientPagination`, which is 2 places in this file, that's it.

Andy2003:
@vinzscam what do you think about my [latest commment](https://github.com/backstage/backstage/pull/28449#discussion_r1922413538)?

Andy2003:
up

awanlin:
Hi @Andy2003, I'm not sure why there are various API report files being changed in this PR. Perhaps your can rebase and revert those changes?

awanlin:
Hi @vinzscam, can you give this one another look, please?

Andy2003:
> Hi @Andy2003, I'm not sure why there are various API report files being changed in this PR. Perhaps your can rebase and revert those changes?

@awanlin This does not work (see https://github.com/backstage/backstage/actions/runs/13374155759/job/37349562189?pr=28449)

```
*************************************************************************************
* You have uncommitted changes to the public API or reports of a package.           *
* To solve this, run `yarn build:api-reports` and commit all md file changes.       *

Error: API Extractor completed with 0 errors and 1 warnings

*************************************************************************************

Error: Process completed with exit code 1.
```

I have no idea why my changes result in api-report changes

awanlin:
Yeah, that's really odd. 🤔 

I don't have any solutions but will reach out and see what I can figure out.

Andy2003:
up

Andy2003:
up again

awanlin (changes requested):
Hi @Andy2003, this fell off my radar and I've just got back from vacation and BackstageCon/KubeCon in London UK, busy catching up.

~~It does look like the CI is failing due to API Reports not being updated, details on how to fix that here: https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md#api-reports~~

Sorry, didn't re-read everything to get caught up :( Will reach out to the maintainers to help us resolve this.

awanlin:
Hi @Andy2003, can you try to rebase and then generate the API reports again? Just want to see if we get lucky and some other change might have resolve this for us.

plugins/catalog/src/components/CatalogTable/OffsetPaginatedCatalogTable.tsx
vinzscam: > It's just an refactoring, that is even reducing the total lines of code. The added "complexity" is just an early return in the hook and a ternary for setting the server side config on the Table

I understand that the refactoring is correct, but it doesn't necessarily improve readability, as I've already spent some time understanding how things work here. In my opinion, the refactoring poses more risks than benefits for the adopters. Therefore, I suggest excluding it from the PR and prioritizing the fix for displaying the pagination text in the offset table.

Andy2003:
> Hi @Andy2003, can you try to rebase and then generate the API reports again? Just want to see if we get lucky and some other change might have resolve this for us.

seems to be fixed

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
```diff
@@ -194,7 +194,8 @@ export const CatalogTable = (props: CatalogTableProps) => {
       .join(' ');
 
   const actions = props.actions || defaultActions;
-  const options = {
+  const options: TableProps['options'] = {
+    paginationPosition: 'both',
```
vinzscam, line 198: I think this shouldn't be here, otherwise it will change the behavior of the client-side paginated table.
But it's still a good suggestion to remove `paginationPosition` from the two cursor/offset paginated tables so that they are all consistent. (I think it was added there during the development of the tables and was later forgotten.)

```suggestion
```

.changeset/funny-papayas-tell.md
```diff
@@ -0,0 +1,8 @@
+---
+'@backstage/plugin-catalog': minor
+---
+
+Harmonize `CatalogTable`
+
+- Show pagination text for `OffsetPagination`
+- Do not show paging if there is only one page
```
vinzscam, line 8:
```suggestion
Show the pagination text for the offset-paginated catalog table, and remove the pagination bar from the top of the `CatalogTable` when pagination is enabled.
```

vinzscam (commented):
thank you 🙏, added a minor suggestion and then we can 🚢

vinzscam (approved):
🚢

freben (approved):

PR #28465, opened 2025-01-14
refactor: replace `@keyv/redis` with `@keyv/valkey`

benjdlambert:
Original PR here: https://github.com/backstage/backstage/pull/28444

Want to do some further testing with this, and really work out if it should be a complete replacement, or if we should be supporting both and having different configuration.

### DIFF
--- a/.changeset/blue-actors-compare.md
+++ b/.changeset/blue-actors-compare.md
@@ -1,0 +1,6 @@
+---
+'@backstage/backend-test-utils': patch
+'@backstage/backend-defaults': patch
+---
+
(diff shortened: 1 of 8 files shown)


benjdlambert:
@Lucifergene how would you feel about repurposing your original PR to support valkey as a different option in config to configure rather than replacing the original redis implementation?

freben:
Mm that's likely to be preferable, i think

leinad87:
I arrive here by https://github.com/backstage/backstage/issues/27397 . I can help with this issue too

PR #28466, opened 2025-01-14
restore the search collator types to not break the api

freben:
This reverts parts of https://github.com/backstage/backstage/pull/28182 to ensure that we don't break the catalog api unnecessarily.


Rugvip (approved):
:pray:

PR #28468, opened 2025-01-14
Canon - Update spacing tokens

cdedreuille:
- **Improve spacing tokens** -  We added a new `--canon-space` token that is set to 4px (0.25rem) by default. If you change this value alone, it will affect all the other spacing scale values. We also have a new scale naming from 1 to 9.
- **Change `[data-theme='light']` into `:root`** - We make sure to have a good default for all tokens no matter the theme you use, avoiding duplications.


Rugvip (approved):
Yep I like the new naming a lot more too. Much easier to get a sense to the relation between two spacings. Very hard to tell that `-2xl` is twice the spacing of `-md` for example, while with `-12` vs `-6` it's trivial.

PR #28487, opened 2025-01-15
fix(scaffolder): resolve bug in BitbucketRepoBranchPicker crashing the scaffolder

benjidotsh:
This PR resolves a bug in the `BitbucketRepoBranchPicker` component that crashed the scaffolder.

This bug seems to have been introduced in #27367, which I was asked to validate, but it seems like I forgot to test out the `BitbucketRepoBranchPicker`. 🤡

Ideally, I think this fix should be released as a hotfix instead of waiting for next month's release since the current `BitbucketRepoBranchPicker` renders every template that uses this component unusable.

### DIFF
--- a/.changeset/proud-hornets-cheer.md
+++ b/.changeset/proud-hornets-cheer.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder': patch
+---
+
+Fixed a bug in the BitbucketRepoBranchPicker component that crashed the scaffolder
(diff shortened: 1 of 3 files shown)


benjdlambert (approved):

benjdlambert:
Thanks @benjidotsh!

benjdlambert:
Will probably just hold off on a patch straight away, wanna see if there's any other things that need a patch release and bundle them together.

PR #28491, opened 2025-01-15
Improve knip reports by using a single config with workspaces

Pike:
Also move the dependency from the workspace into repo-tools.

The run with a configured workspace is slower, so I had to bump the timeout.


Pike:
This only includes a changeset for repo-tools, as I'm only changing the knip-report files in the other packages. Given that those are not exposed on npm, I don't rev those packages.

Actually addressing the unused deps would be in follow-up PRs.

CC @freben, we quickly talked about the knip false-positives on discord. Not quite yet at the point where there are none, but much closer.

Pike:
PS: the following packages don't have reports at all yet, not sure if it's good to add them in this PR:

```
packages/backend-common
packages/frontend-defaults
packages/yarn-plugin
packages/canon
packages/frontend-internal
packages/scaffolder-internal
packages/opaque-internal
plugins/auth-backend-module-bitbucket-provider
plugins/catalog-unprocessed-entities-common
plugins/proxy-node
plugins/app
plugins/user-settings-common
plugins/auth-backend-module-bitbucket-server-provider
plugins/notifications-backend-module-email
plugins/catalog-backend-module-gitlab-org
plugins/auth-backend-module-azure-easyauth-provider
plugins/catalog-backend-module-logs
plugins/techdocs-common
plugins/scaffolder-backend-module-notifications
plugins/auth-backend-module-auth0-provider
plugins/auth-react
```

packages/repo-tools/src/commands/knip-reports/knip-extractor.ts
```diff
@@ -55,41 +56,47 @@ function logKnipReportInstructions() {
...
-    ],
+async function generateKnipConfig({ knipConfigPath }: KnipConfigOptions) {
+  const knipConfig: KnipConfig = {
+    workspaces: {
+      '.': {},
+      '{packages,plugins}/*': {
+        entry: [
+          'dev/index.{ts,tsx}',
+          'src/index.{ts,tsx}',
+          'src/alpha.{ts,tsx}',
```
freben, line 67: Would be nice to generate this based on the actual package.json files. Some packages have MANY entrypoints (notably backend-defaults).

Regarding extensions, we're adding more in #19855

packages/frontend-test-utils/knip-report.md
```diff
@@ -1,2 +1,15 @@
...
+| Name                      | Location     | Severity |
+| :------------------------ | :----------- | :------- |
+| @backstage/version-bridge | package.json | error    |
+| zod                       | package.json | error    |
+
+## Unlisted dependencies (1)
+
+| Name               | Location                          | Severity |
+| :----------------- | :-------------------------------- | :------- |
+| @internal/frontend | src/app/createExtensionTester.tsx | error    |
```
freben, line 14: Would be neat to not list the internal ones. They are marked as such in package.json, which is a better way of detect them than by name.

packages/repo-tools/src/commands/knip-reports/knip-extractor.ts
Pike: entrypoints work, I just cross-checked backend-defaults.

This fixes alpha, we only have two packages (I think), which don't expose an existing `src/alpha.ts`:

```bash
grep -L ./alpha `ls p*/*/src/alpha.ts|sed -e s'|src/alpha.ts|package.json|'`
packages/cli/package.json
plugins/scaffolder-backend-module-sentry/package.json
```

The scaffolder one is empty, not sure about the cli one?

This won't cover the `dev` directories for the dev server, though. 

I tested dropping `src/index.{ts,tsx}`, but that triggered more false positives. I didn't dig in deep, but repo-tools is one of them. It only exposes the dist cjs, which one may not have. This might need some more dedicated research.

Re extensions, is that the right PR? That's 2023?

packages/frontend-test-utils/knip-report.md
Pike: Yeah, I think this is something we need to discuss. This might be a topic to bring up with the knip folks upstream.

I wonder if a mono-repo and `"private": true,` in package.json would be enough of a signal. Though, looking at https://github.com/search?q=repo%3Awebpro-nl%2Fknip+private+path%3A%2F%5Epackages%5C%2Fknip%5C%2Ffixtures%5C%2F%2F&type=code it seems that that signal is already used in their test fixtures.

The 

```json
  "backstage": {
    "role": "web-library",
    "inline": true
  },
```

part would only be available to plugins, which need to be upstream. And I don't see a way to extend ignored dependencies from those. Another thing to check.

Should I just add something name-based for now?

packages/repo-tools/src/commands/knip-reports/knip-extractor.ts
freben: Ahh no I meant https://github.com/backstage/backstage/pull/28308

freben: What I meant with the entrypoints is, does it understand that for example [this](https://github.com/backstage/backstage/blob/master/packages/backend-defaults/src/entrypoints/auth/index.ts) file is an entirely separate entrypoint too, and that any dependencies it uses need to be taken into account as well, in addition to the root index?

packages/frontend-test-utils/knip-report.md
freben: Yeah it's the Backstage specific "online" concept I meant indeed.

Maybe let's leave it as a future improvement. It's still a lot better than before either way.

packages/repo-tools/src/commands/knip-reports/knip-extractor.ts
Pike: Yes, excerpt from running knip --debug:

```bash
[packages/backend-defaults] Finding package.json entry paths
{
  patterns: [
    'packages/backend-defaults/package.json',
    'packages/backend-defaults/src/entrypoints/userInfo/index.ts',
    'packages/backend-defaults/src/entrypoints/urlReader/index.ts',
    'packages/backend-defaults/src/entrypoints/scheduler/index.ts',
    'packages/backend-defaults/src/entrypoints/rootLogger/index.ts',
    'packages/backend-defaults/src/entrypoints/rootLifecycle/index.ts',
    'packages/backend-defaults/src/entrypoints/rootHttpRouter/index.ts',
    'packages/backend-defaults/src/entrypoints/rootHealth/index.ts',
    'packages/backend-defaults/src/entrypoints/rootConfig/index.ts',
    'packages/backend-defaults/src/entrypoints/permissions/index.ts',
    'packages/backend-defaults/src/entrypoints/permissionsRegistry/index.ts',
    'packages/backend-defaults/src/entrypoints/logger/index.ts',
    'packages/backend-defaults/src/entrypoints/lifecycle/index.ts',
    'packages/backend-defaults/src/entrypoints/httpRouter/index.ts',
    'packages/backend-defaults/src/entrypoints/httpAuth/index.ts',
    'packages/backend-defaults/src/entrypoints/discovery/index.ts',
    'packages/backend-defaults/src/entrypoints/database/index.ts',
    'packages/backend-defaults/src/entrypoints/cache/index.ts',
    'packages/backend-defaults/src/entrypoints/auth/index.ts',
    'packages/backend-defaults/src/index.ts'
  ],
...
  paths: [
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/package.json',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/userInfo/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/urlReader/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/scheduler/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/rootLogger/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/rootLifecycle/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/rootHttpRouter/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/rootHealth/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/rootConfig/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/permissions/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/permissionsRegistry/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/logger/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/lifecycle/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/httpRouter/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/httpAuth/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/discovery/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/database/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/cache/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/entrypoints/auth/index.ts',
    '/Users/54717/src/backstage/backstage/packages/backend-defaults/src/index.ts'
  ]
}
```

packages/frontend-test-utils/knip-report.md
Pike: I went with a regexp on the name for now.

packages/repo-tools/src/commands/knip-reports/knip-extractor.ts
Pike: I rebased this ontop of https://github.com/backstage/backstage/pull/28308 now that that's merged, and I needed some adjustment for the tests.

Otherwise, this works like expected, I think.

Pike:
> PS: the following packages don't have reports at all yet, not sure if it's good to add them in this PR:

I back-filled the missing reports now, too. Only canon is not done, because that has a storybook setup that differs from the rest, so I'd recommend that to be done in a follow-up.
I also enabled the techdocs-cli-embed one again, I couldn't reproduce the problem that was suggested in the comment to disable it.

@freben , this should be ready for another look.

packages/repo-tools/src/commands/knip-reports/knip-extractor.ts
```diff
@@ -170,15 +185,9 @@ export async function runKnipReports({
         ),
       ),
     );
+    await cleanKnipConfig({ knipConfigPath });
```
freben, line 188: should this be in a `finally` block so that it always happens?

packages/repo-tools/package.json
```diff
@@ -74,6 +74,7 @@
     "just-diff": "^6.0.2",
     "knex": "^3.0.0",
     "knex-pglite": "^0.11.0",
+    "knip": "^5.42.0",
```
freben, line 77: The thing about having it in the root instead of here, is that we don't impose the cost of installing knip if you don't use it. The feature isn't really "external" fully yet; it's a bit hidden away and it's mostly us that use it inside the main repo, except some brave souls.

I don't have a very strong opinion here. What do you think?

packages/repo-tools/src/commands/knip-reports/knip-extractor.ts
```diff
@@ -55,41 +58,43 @@ function logKnipReportInstructions() {
...
-      'node_modules/**',
-      'dist/**',
-      '{fixtures,migrations,templates}/**',
-    ],
     ignoreDependencies: [
+      // these is reported as a referenced optional peerDependencies
+      // TBD: investigate what triggers these
+      '@types/react',
+      '@types/jest',
+      '@internal/.*', // internal packages are not published and inlined
```
freben, line 87: The trouble with this is, `@internal` is a very common prefix and not always meaning it isn't published. Actually I don't know if the package name can be used for anything truly useful here 🤔 But yeah, given that his feature really isn't truly fully usable for external people yet (that would require some more configurability etc probably), this is still true for our own repo and some common cases, so let's keep it as is for now.

packages/repo-tools/src/commands/knip-reports/knip-extractor.ts
Pike: The way the command fails is that it spits out the invocation of `knip`, including the config file. If you then want to debug that, we can't really, unless the file is there.

That's why I changed this to leave the config file around when errors were hit.

Future scenario might even look like putting most of the config into a knip plugin, and then only the stuff specific to this instance of backstage like storyboard and internal packages would go into a knip config checked in to this repo. And other instances would do the same thing.

But we'll need a stable state for at least two backstages before that, I think.

packages/repo-tools/package.json
Pike: I don't have a strong opinion, I tested it the other way around and it doesn't seem to cause any harm. Happy to go either way.

freben: Yepyep it's not overly large, we can keep this.

packages/repo-tools/src/commands/knip-reports/knip-extractor.ts
freben: Hm OK. I see it's already in our gitignore - that's nice.

PR #28494, opened 2025-01-15
Feature(Scaffolder): Use PermissionsRegistryService

CptnFizzbin:
Updates the scaffolder plugin to allow for use of the new PermissionsRegistryService, and restores the ability to add custom rules with the new backend framework.

follow up to #28053

### DIFF
--- a/.changeset/late-moose-care.md
+++ b/.changeset/late-moose-care.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend': minor
+---
+
+`permissionRules` option in `createRouter()` is deprecated. Please upgrade to the new backend system and use `coreServices.permissionsRegistry` to add custom rules
--- a/docs/permissions/plugin-authors/03-adding-a-resource-permission-check.md
+++ b/docs/permissions/plugin-authors/03-adding-a-resource-permission-check.md
@@ -135,7 +135,7 @@ export const todoListPermissionResourceRef = createPermissionResourceRef<
   TodoFilter
 >().with({
   pluginId: 'todolist',
-  type: TODO_LIST_RESOURCE_TYPE,
+  resourceType: TODO_LIST_RESOURCE_TYPE,
 });
 
 export const isOwner = createPermissionRule({
--- a/docs/permissions/plugin-authors/03-adding-a-resource-permission-check.md
+++ b/docs/permissions/plugin-authors/03-adding-a-resource-permission-check.md
@@ -141,7 +141,7 @@ export const todoListPermissionResourceRef = createPermissionResourceRef<
 export const isOwner = createPermissionRule({
   name: 'IS_OWNER',
   description: 'Should allow only if the todo belongs to the user',
-  resourceType: todoListPermissionResourceRef,
+  resourceRef: todoListPermissionResourceRef,
   paramsSchema: z.object({
     userId: z.string().describe('User ID to match on the resource'),
   }),
--- a/plugins/scaffolder-backend/report.api.md
+++ b/plugins/scaffolder-backend/report.api.md
@@ -49,6 +49,7 @@ import { Logger } from 'winston';
 import { PermissionEvaluator } from '@backstage/plugin-permission-common';
 import { PermissionRule } from '@backstage/plugin-permission-node';
 import { PermissionRuleParams } from '@backstage/plugin-permission-common';
+import { PermissionsRegistryService } from '@backstage/backend-plugin-api';
 import { PermissionsService } from '@backstage/backend-plugin-api';
 import { RESOURCE_TYPE_SCAFFOLDER_ACTION } from '@backstage/plugin-scaffolder-common/alpha';
 import { RESOURCE_TYPE_SCAFFOLDER_TEMPLATE } from '@backstage/plugin-scaffolder-common/alpha';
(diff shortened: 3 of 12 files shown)


freben:
@CptnFizzbin Just noting that this is not building correctly.

.changeset/late-moose-care.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend': major
```
freben, line 2: We should try to not make a major breaking change out of this. It's better if we deprecate old things and try to roll many such breakages into a single future major release. There's already some of that type of thing lined up.

Could you instead mark the old exports and `permissionRules` field as deprecated? Then the underlying code could take both the new service AND old input into account, and throwing an exception if someone tries to supply the same thing through both methods.

plugins/scaffolder-node/src/extensions.ts
```diff
@@ -0,0 +1,163 @@
...
+
+import { createExtensionPoint } from '@backstage/backend-plugin-api';
+import {
+  TaskBroker,
+  TemplateAction,
+  TemplateFilter,
+  TemplateGlobal,
+} from '@backstage/plugin-scaffolder-node';
+
+export * from './tasks/alpha';
```
freben, line 25: this shouldn't be here, right

vinzscam:
Hey @CptnFizzbin, just a small suggestion to make reviews easier - try to avoid moving or refactoring stuff unless it’s really needed for the changes. It can make the review take longer and might risk the PR getting delayed. Thanks! 😊

CptnFizzbin:
@vinzscam Yep. This PR was a continuation of #28053 which was originally for adding the extension point. The moved files were to update the structure to follow the convention for permissions that I saw in the Catalog plugin.

plugins/scaffolder-backend/report.api.md
```diff
@@ -74,7 +74,7 @@ import { ZodTypeDef } from 'zod';
 // @public @deprecated (undocumented)
 export type ActionContext<TInput extends JsonObject> = ActionContext_2<TInput>;
 
-// @public (undocumented)
+// @public @deprecated (undocumented)
```
acierto, line 77: Is there any place for users to find the rectification after numbers of these types are deprecated?

CptnFizzbin: for the `*RulesInput` types, they won't have a direct replacement other than using the `create*PermissionRule` exports. The `PermissionsRegistry` should then handle all of that for us.

.changeset/late-moose-care.md
CptnFizzbin: Downgraded the changes to minor, and added a flag to `createRouter` to disable the legacy `permissionsIntegrationRouter` when using the new Backend framework.

Rugvip:
@CptnFizzbin I've opened #28677, aiming to close those gaps

plugins/scaffolder-backend/src/permissions/rules/scaffolderActionRules.ts
```diff
@@ -14,53 +14,48 @@
...
 
-export const hasActionId = createActionPermissionRule({
+const hasActionId = createPermissionRule({
   name: 'HAS_ACTION_ID',
-  resourceType: RESOURCE_TYPE_SCAFFOLDER_ACTION,
   description: `Match actions with the given actionId`,
+  resourceType: scaffolderActionPermissionResourceRef,
   paramsSchema: z.object({
     actionId: z.string().describe('Name of the actionId to match on'),
   }),
```
CptnFizzbin, line 61: @Rugvip ran into an issue using the new `resourceRef`s. When passing it as the resourceType here, TypeScript gets confused and resolves the permission rule to `PermissionRule<unknown, unknown, string, undefined>` breaking everything downstream of this

CptnFizzbin: realized my mistake. need to use `resourceRef` not `resourceType`

docs/permissions/plugin-authors/03-adding-a-resource-permission-check.md
```diff
@@ -135,7 +135,7 @@ export const todoListPermissionResourceRef = createPermissionResourceRef<
   TodoFilter
 >().with({
   pluginId: 'todolist',
-  type: TODO_LIST_RESOURCE_TYPE,
+  resourceType: TODO_LIST_RESOURCE_TYPE,
 });
```
CptnFizzbin, line 139: question: should this get renamed to `resourceRef` instead of `type`? I think almost all of the other permission rule stuff uses the `resourceRef` term

freben: It's a bit confusing with the nomenclature here.

This is indeed a type name, that's used to form what now is dubbed a resource ref, but the thing is, inside the framework you'll also see resource refs being what we call individual objects keyed by ID strings (e.g. in the catalog, a resource ref may actually be an entity ref).

Not sure if that helps with your question

CptnFizzbin: Oh! Nvm, my brain thought it saw a place that asked for a `resourceRef`, but the property name was `resourceType`

awanlin (approved):
Docs look good, thanks for keeping them up to date @CptnFizzbin 🚀

CptnFizzbin:
I'll need to get back around to working on this. Ran into some issues trying to update the tests to the new registry service. It's not the nicest API to mock

benjdlambert:
@vinzscam can you take a peek throught this? :pray:

freben:
@CptnFizzbin Hey just checking in - did we ever get around to finishing this one off? Would love to have it in

CptnFizzbin:
Sorry, I haven't had bandwidth to work on this.

PR #28501, opened 2025-01-15
Canon - Transform and bundle CSS

cdedreuille:
We are now transforming and bundling CSS to be available in the `dist` folder.


Rugvip (approved):
Yep let's :shipit:, no super strong opinions here for now, especially with the new direction for CSS. But as mentioned I think it's better to have a bespoke setup for this rather than adding things to the CLI build tooling.

PR #28506, opened 2025-01-15
Canon - Removing docs from Storybook

cdedreuille:
This is a bit of a cleaning PR but as we moved our docs to a new site in `/canon-docs` we don't need all the mdx files in Storybook. Storybook will remain as our testing platform.


Rugvip (approved):
✨

PR #28511, opened 2025-01-15
Canon - Renaming all component class names

cdedreuille:
In this PR I'm cleaning up CSS class naming following this pattern:

```css
--canon-[component]--[prop]-[value]
```

I also fixed missing props and used `clsx` for class names everywhere.


packages/canon/src/components/Checkbox/styles.css
```diff
@@ -33,7 +15,25 @@
   }
 }
 
-.checkbox-indicator {
+.canon-checkboxLabel {
```
Rugvip, line 18:
```suggestion
.canon-CheckboxLabel {
```

Rugvip (approved):
👍

PR #28518, opened 2025-01-16
catalog: cleanup `refresh_state_references` for old providers

benjdlambert:
This PR will remove all `refresh_state_references` for a given entity when it's being written to the DB.

This means that things are cleaned up correctly when a provider emits an entity with a `locationKey` which replaces one from a different provider without a `locationKey`.

### DIFF
--- a/plugins/catalog-backend/src/tests/integration.test.ts
+++ b/plugins/catalog-backend/src/tests/integration.test.ts
@@ -971,5 +971,14 @@ describe('Catalog Backend Integration', () => {
         .where({ target_entity_ref: 'component:default/component-1' })
         .select(),
     ).resolves.toEqual([]);
+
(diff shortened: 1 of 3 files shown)


plugins/catalog-backend/src/tests/integration.test.ts
```diff
@@ -55,6 +55,8 @@ import { mockServices } from '@backstage/backend-test-utils';
 import { LoggerService } from '@backstage/backend-plugin-api';
 import { DatabaseManager } from '@backstage/backend-common';
 import { entitiesResponseToObjects } from '../service/response';
+import { DbRefreshStateReferencesRow } from '../database/tables';
+import { textChangeRangeIsUnchanged } from 'typescript';
```
Rugvip, line 59: 🤌

PR #28521, opened 2025-01-16
Hide text output button if only one text output is defined

yu-kkurii:

Our team has recently started using text outputs in scaffolder templates to show users information they should be aware of after running a template. Our users appreciate this extra info, but the buttons that text outputs add to the button box have been confusing our users a lot, because, unlike links, these buttons don't seem to do anything. This PR proposes to hide the button if there is only one output defined. 

Example template output:
```yaml
  output:
    text:
      - title: Your next steps
        content: >
          # Success!
          
          Text
```

Before
Screenshot 2025-01-16 at 10 32 07

After
Screenshot 2025-01-16 at 10 32 18

### DIFF
--- a/.changeset/brown-cycles-beg.md
+++ b/.changeset/brown-cycles-beg.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-scaffolder-react': patch
 ---
 
-Adds an option to hide text output buttons
+Hide text output button if only one is present
--- a/.changeset/brown-cycles-beg.md
+++ b/.changeset/brown-cycles-beg.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-react': patch
+---
+
+Adds an option to hide text output buttons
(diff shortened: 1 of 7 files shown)


benjdlambert:
So I think that the behaviour exists so that if you have multiple `text` outputs, you would be able to switch between the different text outputs by clicking the buttons.

I'd be fine to repurpose this PR into always hiding the buttons if there's only one `text` output instead of having a `showButton` prop.

yu-kkurii:
@benjdlambert do you think the text output buttons could also be combined into a button switch group? That way it would be a bit more clear switch buttons switch between these outputs

benjdlambert:
@yu-kkurii i mean maybe? Do you wanna send some examples of what you think is a better design and we can see how it feels?

I made the output components, and I'm not a designer so happy for improvements and suggestions!

yu-kkurii:
After a discussion we decided it would probably be best to just leave it as is! I believe the PR should now be ready

acierto (approved):

Issue #28490, opened 2025-01-15
🐛 Bug Report: GithubUrlReader does not cache blobs


novellizator:
### 📜 Description

It doesn't send the request with the necessary header, so a rate limit credit is consumed

### 👍 Expected behavior

It should use the cache, not to hit the github so hard...

### 👎 Actual Behavior with Screenshots

https://github.com/backstage/backstage/blob/116cd46adca300328e0344cb5504b3bf955c228f/packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.ts#L272 "init" does not contain the etag

### 👟 Reproduction steps

just try tracking the rate limits how they are decreasing rapidly

### 👀 Have you spent some time to check if this bug has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Maybe

novellat:
The quickest solution is to make GithubUrlReader stateful by adding a cache for the files.

Maybe the cleanest solution would be to change the interface of the search() function so that a cache is supplied. But that would require way more changes.

freben:
Hm.

[Here](https://github.com/backstage/backstage/blob/116cd46adca300328e0344cb5504b3bf955c228f/plugins/catalog-backend/src/processors/UrlReaderProcessor.ts#L64) is the code that does the actual reading of locations. It already has a cache builtin. It is meant to pass along the previous etag from the last call, and expects a `NotModifiedError` back if the remote hasn't changed.

My hunch is that specifically search (ie when there are wildcards in the url) doesn't properly make use of the etag. So instead of making a cache in the `GithubUrlReader`, could you look into why the etag handling isn't working? My expectation would be that the etag would be a placeholder for the hash of the tip of the main branch of that repo, or something like that, so that the read request to github only happens if there has been any changes at all to the repo since last time.

novellat:
Yes, I think, this is the very case.
The search function: 
https://github.com/backstage/backstage/blob/116cd46adca300328e0344cb5504b3bf955c228f/plugins/catalog-backend/src/processors/UrlReaderProcessor.ts#L136C24-L136C76 

[...]

PR #28526, opened 2025-01-16
Update GithubUrlReader.ts

novellat:

### DIFF
--- a/packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.ts
+++ b/packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.ts
@@ -72,6 +72,16 @@ export class GithubUrlReader implements UrlReaderService {
     });
   };
 
+  public cacheStats() {
+    return {
+      cacheSize: Object.keys(this.githubFileCache).length,
+      cacheHits: this.cacheHits,
+      cacheMisses: this.cacheMisses,
+      // this is not reliable, since different requests have different rate limits, but mostly you get the first type
+      rateLimitRemaining: this.rateLimitsRemaining,
+    };
+  }
+
   constructor(
     private readonly integration: GithubIntegration,
     private readonly deps: {
@@ -84,7 +94,16 @@ export class GithubUrlReader implements UrlReaderService {
         `GitHub integration '${integration.title}' must configure an explicit apiBaseUrl or rawBaseUrl`,
       );
     }
+
+    setInterval(() => {
+      console.log('GithubUrlReader Stats', this.cacheStats());
+    }, 1000 * 20);
   }
+
+  private githubFileCache: GithubFileCache = {};
+  private cacheHits: number = 0;
+  private cacheMisses: number = 0;
+  private rateLimitsRemaining: number | undefined = undefined;
 
   async read(url: string): Promise<Buffer> {
     const response = await this.readUrl(url);
@@ -269,7 +288,10 @@ export class GithubUrlReader implements UrlReaderService {
       return matching.map(item => ({
         url: pathToUrl(item.path!),
         content: async () => {
-          const blob: GhBlobResponse = await this.fetchJson(item.url!, init);
+          const blob: GhBlobResponse = await this.fetchCachedJson(
+            item.url!,
+            init,
+          );
           return Buffer.from(blob.content, 'base64');
         },
       }));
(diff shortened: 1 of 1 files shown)


awanlin:
Hi @novellat, can you please update the title and the description with some more details, please? Just makes it easier to understand what this change is all about. 👍

novellat:
@awanlin: Hey Andre. In the original issue (https://github.com/backstage/backstage/issues/28490#issuecomment-2595422615) I've discussed a problem in the catalog processor with Freben - whereby the current processor is unnecessarily overwhelming the github api, which is then hitting rate limits.

I've proposed several possible solutions and this is basically a POC I whipped out in a few hours to fix it in our company. 

I think I should have capacity to make a higher quality solution if you guys are interested. Just give me directions ;)

awanlin:
Hi @novellat, that's helpful, it's just that reading this "Update GithubUrlReader.ts" doesn't really tell me any of that. I thought maybe this was just a typo fix when I spotted it, which it is not at all. If you could update the title to be more clear about the change that would be great! Thanks for updating the description. 👍

novellat:
@awanlin, updated. Please let me know if there's anythign else I can do. 

Theoretically speaking we might even close this, since it's not produciton ready and I only made it to showcase one possible solution.

packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.ts
```diff
@@ -122,28 +146,21 @@ export class GithubUrlReader implements UrlReaderService {
...
+      headers: credentials.headers,
     });
+    return Buffer.from(blob.content, 'base64');
+  }
 
-    return ReadUrlResponseFactory.fromNodeJSReadable(response.body, {
-      etag: response.headers.get('ETag') ?? undefined,
-      lastModifiedAt: parseLastModified(response.headers.get('Last-Modified')),
-    });
+  async readUrl(
```
novellat, line 155: i'm deleting a whole bunch of stuff from readUrl. Honestly I have no idea what it was for - studying the code if seems it wasn't very useful. Either way, my implementation behaves equivalently.

packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.ts
```diff
@@ -363,4 +388,42 @@ export class GithubUrlReader implements UrlReaderService {
...
+      };
+      this.cacheMisses++;
+      return json;
+    } catch (error) {
+      if (error instanceof NotModifiedError) {
+        this.cacheHits++;
+        return cachedFile?.contents;
+      }
+      this.logger.info(
+        `GithubUrlReader: Error fetching ${url.toString()} ${error.name} ${
```
novellat, line 420: mostly the 404s, it's good to know about them

packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.ts
```diff
@@ -58,34 +58,58 @@ export type GhBlobResponse =
...
-  static factory: ReaderFactory = ({ config, treeResponseFactory }) => {
+  static factory: ReaderFactory = ({ config, treeResponseFactory, logger }) => {
     const integrations = ScmIntegrations.fromConfig(config);
     const credentialsProvider =
       DefaultGithubCredentialsProvider.fromIntegrations(integrations);
     return integrations.github.list().map(integration => {
       const reader = new GithubUrlReader(integration, {
         treeResponseFactory,
         credentialsProvider,
+        logger,
```
novellat, line 69: GithubUrlReader can supply a lot of important information relevant to the catalog...

novellat:
@awanlin @vinzscam @benjdlambert So I've updated the PR. As mentioned in https://github.com/backstage/backstage/issues/28490#issuecomment-2614518814 I've managed to get the github api rate limits consumption down 98%. Please let me know if there's anything I can do...

Rugvip (changes requested):
The catalog already has its own cache for the file contents, this would cause us to cache the contents twice. We settled on not returning the existing content on cache hits for URL readers and I think we want to keep that as is still.

novellat:
Ok, fair. We're using these changes and it saved us from hitting github rate limits which renders the whole backstage project unusable for us. If anyone has the same problem, they can use my changes as well.

Issue #27902, opened 2024-11-28
💬 RFC: Support for Postgres Flexible server Connectivity using Microsoft Entra authentication (Database Connectivity)


aswathysen:
### 🔖 Need

Currently connectivity to database is through Postgres password and server information. Is there a way to connect using Microsoft Entra authentication (Under a tenant) for better security. Is there a solution already discussed or supported for this request. 

### 🎉 Proposal

Currently the Postgres flexible server connectivity is using PG admin and password (Database Connectivity). Is there a way to connect using Microsoft Entra authentication. If so, is there any supporting documents available. Or can this be worked on for allowing database to connect using Microsoft Entra authentication.

### 👀 Have you spent some time to check if this RFC has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

Issue #22535, opened 2024-01-26
🚀 Feature: Support for rotating Postgres password / managed identities


stuartsaid:
### 🔖 Feature description

Add support in the new backend core Postgres database for database passwords that expire/are rotated during runtime, such as for usage with Azure Managed Identities where a token is used for the password.

### 🎤 Context

I'd like to switch away from username & password authentication of our Postgres database to Managed Identities in Azure to make the database more secure. At present there does not appear to be any support for this and would mean we need to extend the [Database Manager](https://github.com/backstage/backstage/blob/afe7af72fdf41ff0db44633255880ca622cbc5a7/packages/backend-common/src/database/DatabaseManager.ts#L230)

### ✌️ Possible Implementation

Allow for providing a custom function to fetch the token and expirationChecker to the underlying Knex connection. Knex supports this via a connection function param as detailed here: https://knexjs.org/guide/#configuration-options

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

vinzscam:
Thanks for reporting this 🙏. There has been some effort in the past toward supporting dynamic database credentials, but nothing landed in production. Indeed, it would be a nice addition.

iorlas:
Well, it is critical for Azure environment, especially in enterprise industry. Up.

szboynono:
Hey just wanna keep this feature request up, we are needing that as well in our org...just need to have it to do what techdoc uploading to azure blob is doing.

iniinikoski:
Hi, just linking issues together: https://github.com/backstage/backstage/issues/17466 - this is also about rotating Postgres password but also the user, as that one is about support for Hashicorp Vault. No idea if this would be the same implementation but.

aramissennyeydd:
Also interested in a version of this for IAM auth to RDS - open to picking it up, but wanted to talk a bit about what this would look like in the new backend system. 

My initial thought was to use https://github.com/backstage/backstage/pull/25593 to either 
1. move the db connectors to db connector service modules 
2. move just the knex initialization logic to separate service modules

[...]

PR #28532, opened 2025-01-16
Add support for Entra authentication on Azure Database for PostgreSQL

mike-schenk:

This adds support for passwordless Entra authentication on Azure Database for PostgreSQL by setting `database.connection.type` to `azure`.

Addresses #27902 and #22535

### DIFF
--- a/packages/backend-defaults/config.d.ts
+++ b/packages/backend-defaults/config.d.ts
@@ -15,7 +15,6 @@
  */
 
 import { HumanDuration } from '@backstage/types';
-import { AzureTokenCredentialConfig } from './src/entrypoints/database/connectors/postgres';
 
 export interface Config {
   app: {
@@ -429,7 +428,39 @@ export interface Config {
             /**
              * Optional Azure token credential configuration
              */
-            tokenCredential?: AzureTokenCredentialConfig;
+            tokenCredential?: {
+              /**
+               * How early before an access token expires to refresh it with a new one.
+               * Defaults to 5 minutes
+               * Supported formats:
+               * - A string in the format of '1d', '2 seconds' etc. as supported by the `ms`
+               *   library.
+               * - A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
+               * - An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.
+               */
+              tokenRenewalOffsetTime?: string | HumanDuration;
+            } & (
+              | {
+                  type: 'ClientSecretCredential';
+                  clientId: string;
+                  /**
+                   * @visibility secret
+                   */
+                  clientSecret: string;
+                  tenantId: string;
+                }
+              | {
+                  type: 'ManagedIdentityCredential';
+                  /**
+                   * The client ID of a user-assigned managed identity.
+                   * If not provided, the system-assigned managed identity is used.
+                   */
+                  clientId?: string;
+                }
+              | {
+                  type?: undefined | 'DefaultAzureCredential';
+                }
+            );
           }
         | {
             /**
--- a/packages/backend-defaults/config.d.ts
+++ b/packages/backend-defaults/config.d.ts
@@ -426,9 +426,15 @@ export interface Config {
              */
             type: 'azure';
             /**
-             * How many milliseconds before token expiration to refresh the token Defaults to 180,000 (3 minutes)
+             * How early before an access token expires to refresh it with a new one.
+             * Defaults to 5 minutes
+             * Supported formats:
+             * - A string in the format of '1d', '2 seconds' etc. as supported by the `ms`
+             *   library.
+             * - A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
+             * - An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.
              */
-            allowedClockSkewMs?: number;
+            tokenRenewalOffsetTime?: string | HumanDuration;
           }
         | {
             /**
--- a/packages/backend-defaults/config.d.ts
+++ b/packages/backend-defaults/config.d.ts
@@ -422,6 +422,16 @@ export interface Config {
         | string
         | {
             /**
+             * The specific config for Azure database for PostgreSQL with Entra authentication
+             */
+            type: 'azure';
+            /**
+             * How many milliseconds before token expiration to refresh the token Defaults to 180,000 (3 minutes)
+             */
+            allowedClockSkewMs?: number;
+          }
+        | {
+            /**
              * The specific config for cloudsql connections
              */
             type: 'cloudsql';
--- a/packages/backend-defaults/config.d.ts
+++ b/packages/backend-defaults/config.d.ts
@@ -15,6 +15,7 @@
  */
 
 import { HumanDuration } from '@backstage/types';
+import { AzureTokenCredentialConfig } from './src/entrypoints/database/connectors/postgres';
 
 export interface Config {
   app: {
@@ -426,15 +427,9 @@ export interface Config {
              */
             type: 'azure';
             /**
-             * How early before an access token expires to refresh it with a new one.
-             * Defaults to 5 minutes
-             * Supported formats:
-             * - A string in the format of '1d', '2 seconds' etc. as supported by the `ms`
-             *   library.
-             * - A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
-             * - An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.
+             * Optional Azure token credential configuration
              */
-            tokenRenewalOffsetTime?: string | HumanDuration;
+            tokenCredential?: AzureTokenCredentialConfig;
           }
         | {
             /**
--- a/packages/backend-defaults/src/entrypoints/database/connectors/postgres.test.ts
+++ b/packages/backend-defaults/src/entrypoints/database/connectors/postgres.test.ts
@@ -199,7 +199,7 @@ describe('postgres', () => {
           client: 'pg',
           connection: {
             type: 'azure',
-            allowedClockSkewMs: 60_000, // 1 minute
+            tokenRenewalOffsetTime: '1 minute',
             user: 'user@contoso.com',
             database: 'other_db',
             port: 5423,
@@ -208,7 +208,7 @@ describe('postgres', () => {
       );
       let connectionResult = await configResult.connection();
 
-      jest.useFakeTimers({ now: tokenExpirationTimestamp - 120_000 });
+      jest.useFakeTimers({ now: tokenExpirationTimestamp - 90_000 });
       let expirationResult = await connectionResult.expirationChecker();
       expect(expirationResult).toBe(false);
 
@@ -216,20 +216,16 @@ describe('postgres', () => {
       expirationResult = await connectionResult.expirationChecker();
       expect(expirationResult).toBe(true);
 
-      jest.useFakeTimers({ now: tokenExpirationTimestamp - 30_000 });
-      expirationResult = await connectionResult.expirationChecker();
-      expect(expirationResult).toBe(true);
-
       jest.useFakeTimers({ now: tokenExpirationTimestamp });
       expirationResult = await connectionResult.expirationChecker();
       expect(expirationResult).toBe(true);
 
+      // Check the default tokenRenewalOffsetTime of 5 minutes
       configResult = await buildPgDatabaseConfig(
         new ConfigReader({
           client: 'pg',
           connection: {
             type: 'azure',
-            allowedClockSkewMs: 0,
             user: 'user@contoso.com',
             database: 'other_db',
             port: 5423,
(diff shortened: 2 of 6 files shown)


sanderaernouts:
Thanks for this PR @mike-schenk 👍. This was on my list of things I wanted to contribute to Backstage for a while but I didn't get around to it yet.

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -121,6 +129,41 @@ export async function buildPgDatabaseConfig(
...
+      'https://ossrdbms-aad.database.windows.net',
+    );
+
+    const connectionConfig = {
+      ...sanitizedConfig.connection,
+      password: token.token,
+      expirationChecker: () => {
+        /* return true if the token is near to or past its expiration. */
+        const isExpired =
+          token.expiresOnTimestamp - allowedClockSkewMs <= Date.now();
```
sanderaernouts, line 153: The default refresh offset in The Microsoft authentication library is [5 minutes](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/d28085905644510fbcfacfed98ed1b43014b67c1/lib/msal-common/src/client/RefreshTokenClient.ts#L54). Maybe you should use 5 minutes as a constant here and then subtract the `allowedClockSkewMs` as an additional offset (on top of the default 5 minutes).
```suggestion
          token.expiresOnTimestamp - fiveMinutes - allowedClockSkewMs <= Date.now();
```
or rename the `allowedClockSkewMs` (configuration) variable to something like `tokenExpirationOffset`
```suggestion
          token.expiresOnTimestamp - tokenExpirationOffset <= Date.now();
```

packages/backend-defaults/config.d.ts
```diff
@@ -420,6 +420,16 @@ export interface Config {
...
         | string
+        | {
+            /**
+             * The specific config for Azure database for PostgreSQL with Entra authentication
+             */
+            type: 'azure';
+            /**
+             * How many milliseconds before token expiration to refresh the token Defaults to 180,000 (3 minutes)
+             */
+            allowedClockSkewMs?: number;
```
sanderaernouts, line 431: Maybe you can use `Duration | HumanDuration` here instead similarly to how this is done in the [SchedulerServiceTaskScheduleDefinition](https://github.com/backstage/backstage/blob/d651ce74623869a27964472644969766d0e9fbb9/packages/backend-plugin-api/src/services/definitions/SchedulerService.ts#L107C1-L107C37)
```suggestion
            allowedClockSkew?: Duration | HumanDuration;
```

Then user could write for example:
```yaml
...
connection:
  type: 'azure'
  allowedClockSkew: 
    minutes: 10
...
```

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -121,6 +129,41 @@ export async function buildPgDatabaseConfig(
...
+  const { DefaultAzureCredential } = require('@azure/identity');
+
+  // By default get a new token starting three minutes before the old one expires
+  const allowedClockSkewMs =
+    config.connection.allowedClockSkewMs !== undefined
+      ? config.connection.allowedClockSkewMs
+      : 180000;
+
+  async function getConnectionConfig() {
+    const credential = new DefaultAzureCredential();
```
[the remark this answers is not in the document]
mike-schenk: Are you suggesting we should allow the credential type to be specified in configuration? That seems reasonable.

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
mike-schenk: Thanks for the reference to the MSAL library. I see its configuration has an optional property named `tokenRenewalOffsetSeconds` which defaults, as you said, to 5 minutes. What if I adopted that name and its default value for our configuration? I don't think we need _both_ that offset as well as a clock skew offset.

packages/backend-defaults/config.d.ts
mike-schenk: I didn't notice that `HumanDuration` type before. That looks like a good improvement for the configuration. I see there are two other ways `HumanDuration` is used in this `Config` interface. One uses `string | HumanDuration` and the other uses `number | HumanDuration | string`. And I see that the implementation code in both use the `readDurationFromConfig` function. I think I'll go with the `string | HumanDuration` approach for the config.

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
sanderaernouts: That makes sense. I feel it's also bit more descriptive and applicable than `allowedClockSkewMs`. Normally the offset is also meant to have a chance to recover for failures when renewing the token.

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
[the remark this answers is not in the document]
sanderaernouts: Yes, I hope that you can reuse the types from the Azure integration (which is Azure DevOps) then it should be relatively easy to do. I have a PR open in that integration myself to add support for federated credentials, which you would then also get "for free" in the PostgreSQL database connection (#28437).

packages/backend-defaults/config.d.ts
sanderaernouts: I only noticed it the other day myself. What would the `string` look like? Something like though `00:05:00`?

freben: Correct, string and HumanDuration.

If a string, it gets parsed by the `ms` library, which supports human text forms like "20ms" or "1 minute" etc

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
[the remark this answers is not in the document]
mike-schenk: I'm not sure I can use those types as-is since `PersonalAccessTokenCredentials` isn't an option for database authentication, but I'll see if I can find ways to share some of the code.

I'd still like to give users the ability to configure `DefaultAzureCredential` if they choose.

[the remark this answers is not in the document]
mike-schenk: Looking around the codebase, I see there is a fair amount of near-duplication for handling Azure credentials:

- The [Azure Blob Storage "SCM integration"](https://github.com/backstage/backstage/blob/a5286b5cbd63e22aa961890c53803a22737758b2/packages/integration/src/azureBlobStorage/DefaultAzureCredentialsProvider.ts#L45) supports only `ClientSecretCredential` and `DefaultCredential`.
- The [MS Graph catalog backend plugin](https://github.com/backstage/backstage/blob/a5286b5cbd63e22aa961890c53803a22737758b2/plugins/catalog-backend-module-msgraph/src/microsoftGraph/client.ts#L92) supports only `ClientSecretCredential` and `DefaultCredential`.
- The [Azure Blob Storage TechDocs publisher](https://github.com/backstage/backstage/blob/a5286b5cbd63e22aa961890c53803a22737758b2/plugins/techdocs-node/src/stages/publish/azureBlobStorage.ts#L117) supports only `DefaultCredential` and storage account key.
- The [Kubernetes plugin backend](https://github.com/backstage/backstage/blob/a5286b5cbd63e22aa961890c53803a22737758b2/plugins/kubernetes-backend/src/auth/AzureIdentityStrategy.ts#L41) supports only `DefaultAzureCredential` (and, BTW, has a hard-coded 15 minute token renewal offset).

I think what I'd like to do in this PR is to copy what the first two are doing.

I think a future improvement would be to introduce a shared package, modeled after `integration-aws-node`, to support all the variety of `TokenCredential` implementations, and then use it in the places above, in this database config, and in your Azure DevOps integration. Special cases would still need to be handled in the different respective configs such as account keys for Blob storage and PATs for DevOps repository access.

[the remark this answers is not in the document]
sanderaernouts: That sounds like a good idea. hen I created the credential providers for Azure DevOps I did not think to separate them properly for re-use in other parts.

I think the main concern would be to choose a configuration format that doesn't change because that would mean users have to migrate their configuration. W But it would be nice in my opinion to have a shared library that supports different `TokenCredential` types.

sanderaernouts:
@mike-schenk do you want me to have another look or are you working on some more changes?

mike-schenk:
> @mike-schenk do you want me to have another look or are you working on some more changes?

@sanderaernouts, I think I'm done. Go ahead and respond to what's here.

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -121,6 +138,119 @@ export async function buildPgDatabaseConfig(
...
+  | {
+      type: 'ManagedIdentityCredential';
+      /**
+       * The client ID of a user-assigned managed identity.
+       * If not provided, the system-assigned managed identity is used.
+       */
+      clientId?: string;
+    }
+  | {
+      type: undefined | 'DefaultAzureCredential';
```
freben, line 168:
```suggestion
      type?: undefined | 'DefaultAzureCredential';
```

needs to be truly optional right? 🤔

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -121,6 +138,119 @@ export async function buildPgDatabaseConfig(
...
+    ManagedIdentityCredential,
+    ClientSecretCredential,
+  } = require('@azure/identity');
+
+  // By default get a new token starting five minutes before the old one expires
+  let tokenRenewalOffsetMs = 300_000;
+  const configReader = new ConfigReader(config);
+  if (configReader.has('connection.tokenCredential.tokenRenewalOffsetTime')) {
+    tokenRenewalOffsetMs = durationToMilliseconds(
+      readDurationFromConfig(new ConfigReader(config), {
```
freben, line 184:
```suggestion
      readDurationFromConfig(configReader, {
```

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -121,6 +138,119 @@ export async function buildPgDatabaseConfig(
...
+        );
+        break;
+      case 'DefaultAzureCredential':
+      case undefined:
+        credential = new DefaultAzureCredential();
+        break;
+      default:
+        throw new Error(
+          `Unknown token credential type: ${
+            (tokenCredentialConfig as any).type
```
freben, line 221:
```suggestion
            (tokenCredentialConfig as any)?.type
```

just defensively against null :)

freben (commented):
Nice, almost done!

mike-schenk:
@freben, I think I've addressed all your review comments. Thank you. Please have another look.

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -121,6 +138,123 @@ export async function buildPgDatabaseConfig(
...
+      }),
+    );
+  }
+
+  let credential: TokenCredential;
+  if (configReader.has('connection.tokenCredential')) {
+    const tokenCredentialConfig = configReader.get(
+      'connection.tokenCredential',
+    ) as AzureTokenCredentialConfig;
+    switch (tokenCredentialConfig.type) {
```
sanderaernouts, line 199: Does this mean the `type` field has to be specified in the configuration file? In the Azure integration the this `type` inferred based on the fields you configure. For example if you pass `tenant_id` + `client_id` + `client_secret` we assume that it is a `ClientSecretCredential`, if you pass `tenant_id` + `clien_id` we assume it is a `ManagedIdentityCredential`

packages/backend-defaults/src/entrypoints/database/connectors/postgres.test.ts
```diff
@@ -138,6 +140,114 @@ describe('postgres', () => {
       });
     });
 
+    it('uses the correct config when using azure', async () => {
```
sanderaernouts, line 143: I'd expect test cases here for the client secret and managed identity credential configurations as well

sanderaernouts:
@mike-schenk are you still working on this?

fcamgz:
Any update on this? If there's any outstanding work I would like to help out and see this feature gets merged 🙌

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
mike-schenk: My personal preference is to require the type to be explicit, but I can see that it could be inferred at least for now until something changes and it becomes ambiguous.

mike-schenk:
> @mike-schenk are you still working on this?

> Any update on this? If there's any outstanding work I would like to help out and see this feature gets merged 🙌

I would appreciate a bit of help on this. I have not had the time to complete the last requests which are noted above, including:

* test cases here for the client secret and managed identity credential configurations
* Possibly changing the `type` field to be inferrable.

mike-schenk:
> 👋 Did you test this with recreating connections after the credentials expire? I was pretty convinced looking into [#22535 (comment)](https://github.com/backstage/backstage/issues/22535#issuecomment-2513157330) that knex doesn't actually respect the `expirationChecker` and a lot of DBMS providers don't invalidate existing connections with expired credentials. Instead, when the pool needs to create a new connection, it fails because the credentials never updated (`connection`/`expirationChecker` was never called again).

In the specific case of the knex connection pool and PostgreSQL on Azure, here is what I have observed:
- If a token credential expires while an open connection remains in the pool and then that connection is retrieved from the pool to be re-used, the connection is still open, valid, and usable.
- The `expirationChecker` _is_ properly called by knex when it needs to create a _new_ connection to add to the pool.

In other words, already-pooled connections don't get reset, and don't need to be reset when credentials expire. And unless knex needs to create a new connection, it won't call `expirationChecker` because it will return a pooled connection that has remained open.

PR #28533, opened 2025-01-16
Canon - fix docs styling

cdedreuille:
Fixing the way we bundle CSS to be shown on the docs at build time


Rugvip (approved):
yep this is what I saw referring to in the initial docs PR :sweat_smile:

PR #28550, opened 2025-01-19
docs(backstage): Update README to include policy management setup

abhipsnl:
- Added documentation to the README outlining the process of managing policies in Backstage using a ConfigMap.
- Explained how to deploy Backstage with ConfigMap mounts for policy management and configure `app-config.yaml` to read from the mounted policy file.
- Included step-by-step instructions for setting up the policy management without ScmAuth and RBAC.
- Updated the README to provide clarity on how to customize and verify policy configurations in Backstage.

This document change complements the feature implementation for flexible policy management in Backstage.

### DIFF
--- a/docs/deployment/k8s.md
+++ b/docs/deployment/k8s.md
@@ -537,3 +537,71 @@ For production purposes, this image tag will generally be a full-fledged URL
 pointing to a container registry where built Docker images are hosted. This can
 be hosted internally in your infrastructure, or a managed one offered by a cloud
 provider.
+
+### Backstage [RBAC Policy](https://github.com/backstage/community-plugins/blob/main/workspaces/rbac/plugins/rbac-backend/docs/permissions.md) using configmap
+
+This setup allows you to configure a ConfigMap will store the policy that Backstage will use. This policy will be mounted into the Backstage deployment, and the app-config.yaml will be configured to read the policy from the mounted path.
+
+1. Create a ConfigMap with Policy
+
+Example: Create a file called policy-configmap.yaml:
+
+```
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: backstage-policy-config
+  namespace: backstage
+data:
+  policy.csv: |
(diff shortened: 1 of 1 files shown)


Parsifal-M:
Hey! 👋 

Thanks for this! but would it not make more sense to add this to the [community plugins repo](https://github.com/backstage/community-plugins/tree/main/workspaces/rbac)?

A few concerns I have with this is that this might splinter the documentation in two places in the future, also its not a given that everyone uses this community plugin for RBAC so it could cause confusion, personally I am thinking it would be better to add these docs to the community repo for this plugin and make a reference to the the k8s deployment docs

WDYT?

abhipsnl:
> Hey! 👋
> 
> Thanks for this! but would it not make more sense to add this to the [community plugins repo](https://github.com/backstage/community-plugins/tree/main/workspaces/rbac)?
> 
> A few concerns I have with this is that this might splinter the documentation in two places in the future, also its not a given that everyone uses this community plugin for RBAC so it could cause confusion, personally I am thinking it would be better to add these docs to the community repo for this plugin and make a reference to the the k8s deployment docs
> 
> WDYT?

I had the same thought. From my perspective, the plugin documentation should focus on how to configure and use the plugin itself, keeping it standalone and deployment-agnostic. On the other hand, the Kubernetes deployment documentation provides a broader context, helping users understand how to incorporate the RBAC plugin seamlessly into their deployment. This way, users who are specifically looking to leverage RBAC as a plugin can find clear guidance within the deployment documentation. but may be we can make this change more clear?

awanlin:
Hi @abhipsnl, I strongly agree with @Parsifal-M on this topic. This should live with the plugin's documentation.

Can you please raise a PR to add this there?

Thank!

awanlin:
Talked this over with the other Docs maintainers and we agree this makes more sense with the plugins existing documentation here: https://github.com/backstage/community-plugins/tree/main/workspaces/rbac/plugins/rbac-backend/docs. If you tag me in the PR to add it I can help review and get it merged 👍

Issue #18710, opened 2023-07-18
🐛 Bug Report: Component Owner filter not resetting properly for Group and User kinds


roman-shandurenko:
### 📜 Description

I have an issue with EntityOwnerPicker (Owner filter) and User/Group kind.

When I open a catalog, go to "Component" page, and select an owner filter, the Component list gets filtered.
But if I then change `Kind` to User/Group, the Owner filter gets disappeared and the table shows nothing. 
It means that a filter is still applied even though the filter [is not available for User/Group kinds](https://github.com/backstage/backstage/blob/master/plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx#L111-L116).

### 👍 Expected behavior

If the filter is not available for some kinds, it should not be applied or should be reset.
Another solution would be to update `EntityOwnerPicker` to accept `allowedKinds` so users can deliberately display the filter for `Group` and `Users` (or make it as a default behaviour).

### 👎 Actual Behavior with Screenshots

The filter is disappeared when you change the `Kind` to `Group` or `User`

<img width="597" alt="image" src="https://github.com/backstage/backstage/assets/101107402/68c03316-8218-46ee-96bc-921f6742cf4a">

---

<img width="853" alt="image" src="https://github.com/backstage/backstage/assets/101107402/04654987-49cf-4523-a28a-f13d8d11503a">

### 👟 Reproduction steps

1. Choose a value from "Owner" filter. It generates the following URL: https://demo.backstage.io/catalog?filters%5Bkind%5D=component&filters%5Buser%5D=all&filters%5Bowners%5D=group%3Abackstage%2Fmaintainers

---

2. Change `Kind` to "Group" or "User". It generates the following url: https://demo.backstage.io/catalog?filters%5Bkind%5D=group&filters%5Buser%5D=all&filters%5Bowners%5D=group%3Abackstage%2Fmaintainers

### 📃 Provide the context for the Bug.

It's not a good UX and might be cumbersome for users to not see anything.

### 🖥️ Your Environment

**Plugin**: https://github.com/backstage/backstage/tree/master/plugins/catalog-react

**backstage.json**
```json
{
  "version": "1.15.0"
}
```

**packages/app/package.json**
```
"@backstage/plugin-catalog-react": "^1.7.0"
```

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

[...]

PR #28561, opened 2025-01-20
fix(EntityOwnerPicker): fix entity owner picker behavior (#18710)

Jona-Soerensen:

This PR should fix the EntityOwnerPicker bug described in the issue #18710
Based on the comments of the linked issue, I adjusted the `availableOwners` in the EntityOwnerPicker.tsx file to return only the related OwnerEntities after every EntityKind change.

- Updated EntityOwnerPicker Component to dynamically update the available Owners based on the selected Entity kind.
- Introduced `selectedKind` Prop: Added a new (optional) selectedKind prop to the useFetchEntities hook to specify the kind of entities to fetch.
- refactored useFacetEntities Hook: Refactored to return EntityOwners whenever there’s a change in the selected Entity kind.

I encountered some challenges in adapting the tests to the new logic. As a result, a few tests have been temporarily skipped. Any help is welcome here :)

https://github.com/user-attachments/assets/6b7748af-78e1-4766-9f40-e1b5148a06c2

### DIFF
--- a/.changeset/real-news-marry.md
+++ b/.changeset/real-news-marry.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-react': patch
+---
+
+Fixed the `EntityOwnerPicker`component to dynamically update the `availableOwners` based on the selected Entity kind.
--- a/plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
+++ b/plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
@@ -28,7 +28,7 @@ import { makeStyles } from '@material-ui/core/styles';
 import CheckBoxIcon from '@material-ui/icons/CheckBox';
 import CheckBoxOutlineBlankIcon from '@material-ui/icons/CheckBoxOutlineBlank';
 import React, { useEffect, useMemo, useState } from 'react';
-import { useEntityList } from '../../hooks/useEntityListProvider';
+import { useEntityList } from '../../hooks';
 import { EntityOwnerFilter } from '../../filters';
 import { useDebouncedEffect } from '@react-hookz/web';
 import PersonIcon from '@material-ui/icons/Person';
@@ -145,10 +145,13 @@ export const EntityOwnerPicker = (props?: EntityOwnerPickerProps) => {
   const [{ value, loading }, handleFetch, cache] = useFetchEntities({
     mode,
     initialSelectedOwnersRefs: selectedOwners,
+    selectedEntityKind: filters.kind?.value,
   });
   useDebouncedEffect(() => handleFetch({ text }), [text, handleFetch], 250);
 
-  const availableOwners = value?.items || [];
+  const availableOwners = useMemo(() => {
+    return value?.items || [];
+  }, [value]);
 
   // Set selected owners on query parameter updates; this happens at initial page load and from
   // external updates to the page location.
@@ -166,6 +169,12 @@ export const EntityOwnerPicker = (props?: EntityOwnerPickerProps) => {
         : undefined,
     });
   }, [selectedOwners, updateFilters]);
+
+  // useEffect(() => {
+  //   if (availableOwners.length === 0) {
+  //     setSelectedOwners([]);
+  //   }
+  // }, [availableOwners]);
 
   if (
     ['user', 'group'].includes(
--- a/plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
+++ b/plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
@@ -170,11 +170,11 @@ export const EntityOwnerPicker = (props?: EntityOwnerPickerProps) => {
     });
   }, [selectedOwners, updateFilters]);
 
-  // useEffect(() => {
-  //   if (availableOwners.length === 0) {
-  //     setSelectedOwners([]);
-  //   }
-  // }, [availableOwners]);
+  useEffect(() => {
+    if (availableOwners.length === 0) {
+      setSelectedOwners([]);
+    }
+  }, [availableOwners]);
 
   if (
     ['user', 'group'].includes(
--- a/plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
+++ b/plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
@@ -16,8 +16,13 @@
 import { useApi } from '@backstage/core-plugin-api';
 import useAsyncFn from 'react-use/esm/useAsyncFn';
 import { catalogApiRef } from '../../api';
-import { useState } from 'react';
-import { Entity, parseEntityRef } from '@backstage/catalog-model';
+import { useCallback, useEffect, useState } from 'react';
+import {
+  Entity,
+  parseEntityRef,
+  stringifyEntityRef,
+} from '@backstage/catalog-model';
+import { EntityFilterQuery } from '@backstage/catalog-client';
 
 type FacetsCursor = {
   start: number;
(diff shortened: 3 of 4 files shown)


freben:
Note that tests are failing.

Jona-Soerensen:
Updated the code in order to pass all tests.
I had to update two already existing tests to check only for `toHaveBeenCalledWith` rather than `toHaveBeenLastCalledWith` because of the implemented updateFilter reset.

Jona-Soerensen:
@freben @vinzscam 
Just checking to see if there’s anything else needed from my side for this PR, or if we’re just waiting for the next steps. Let me know! 😊

plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
```diff
@@ -165,7 +168,15 @@ export const EntityOwnerPicker = (props?: EntityOwnerPickerProps) => {
...
-  }, [selectedOwners, updateFilters]);
+  }, [selectedOwners, updateFilters, filters.kind?.value]);
+
+  useEffect(() => {
+    if (availableOwners.length === 0) {
+      updateFilters({
+        owners: undefined,
+      });
+    }
+  }, [availableOwners, updateFilters]);
```
freben, line 179: Could also have made `availableOwners.length` the dependency here, which means that whenever the LENGTH changes (and goes to zero) the thing fires. That's safe against the reference equality changing, and also perhaps even semantically more correct

plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
```diff
@@ -41,40 +41,63 @@ type FacetsInitialRequest = {
...
-  const [facetsPromise] = useState(async () => {
-    if (!enabled) {
+  const fetchFacetsEntities = useCallback(async (): Promise<Entity[]> => {
+    try {
+      if (selectedEntityKind) {
+        const facetResponseFiltered = await catalogApi.getEntityFacets({
+          facets: ['relations.ownedBy'],
+          filter: { kind: [selectedEntityKind] },
+        });
+        if (facetResponseFiltered.facets['relations.ownedBy'].length === 0) {
```
freben, line 63: I'm trying to wrap my head around this one. So first you query by the selected kind. And if THAT has no owners you return the empty array. Otherwise, you throw away the response and query for all kinds again? Or am I just reading it wrong?

I would have expected something along the lines of the inverse:

```ts
if (facetResponseFiltered.facets['relations.ownedBy'].length) {
  return facetResponseFiltered;
}
```

plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
```diff
@@ -41,40 +41,63 @@ type FacetsInitialRequest = {
...
+  selectedEntityKind?: string;
+}) {
   const catalogApi = useApi(catalogApiRef);
+  const [facetsPromise, setFacetsPromise] = useState<Promise<Entity[]>>(
+    Promise.resolve([]),
+  );
 
-  const [facetsPromise] = useState(async () => {
-    if (!enabled) {
+  const fetchFacetsEntities = useCallback(async (): Promise<Entity[]> => {
```
freben, line 56: It feels slightly convoluted now with a `useCallback` and `useEffect`; could this be a single `useMemo` instead that depends on `[catalogApi, selectedEntityKind, enabled]`? Which evaluates to a `Promise<Entity[]>` still, of course.

freben (commented):
Pulling in @vinzscam

vinzscam (changes requested):
can I suggest to skip the filters at call time? So whenever we detect that kind group or user has been selected we make  omit the owner filter from `queryEntities()` or `getEntities()`? That way we don't have to touch react

plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
Jona-Soerensen: I took your suggestion and replaced the availableOwners dependency with `availableOwners.length`

plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
Jona-Soerensen: The challenge here is twofold. First, for a given selected entity kind, I need to return **all owners** and not just a subset of the available owners. In the current implementation in Backstage, the `EntityOwnerPicker` lists all owners and not just the available owners of the selected kind. 

Second, once I’ve received a `facetResponse`, I can’t perform further local filtering by kind to reduce API calls. Essentially, I need the API call to reflect the selected entity kind, so that I get a complete list of owner references for that kind.

While I understand the suggestion of simply returning the filtered response if it has owners, the issue is that we need to ensure that the filtered response includes all the owner references.

Hope I got your question right :)

Jona-Soerensen:
> Can I suggest skipping the filters at call time? Whenever we detect that a specific group or user has been selected, we ensure not to pass down the owner filter to `queryEntities()`/`getEntities()`. This way we don't have to touch React.

When you say “skipping the filters at call time” do you mean that if we detect that a `group` or `user` kind has been selected, we completely avoid passing any filters to the api? So instead of updating the filter in React, we intercept the group and user entities (as proposed in the original bug ticket https://github.com/backstage/backstage/issues/18710#issuecomment-1663824019) and reset the filter accordingly.

My understanding is that this approach would allow us to avoid extra React state updates for the owner filter and the backend would always return all owners. However, as suggested in the bug report, we intended to intercept group and user entities directly in the EntityOwnerPicker to clear the owner filter when it doesn’t apply.

Jona-Soerensen:
Hey, just wanted to ping you again about this PR. @vinzscam could you please elaborate your last comment?
Please let me know if there’s anything else you need from me or any remaining changes you’d like to see. Thanks!

plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
```diff
@@ -41,19 +41,37 @@ type FacetsInitialRequest = {
...
+  enabled,
+  selectedEntityKind,
+}: {
+  enabled: boolean;
+  selectedEntityKind?: string;
+}) {
   const catalogApi = useApi(catalogApiRef);
-
-  const [facetsPromise] = useState(async () => {
+  const facetsPromise = useMemo<Promise<Entity[]>>(() => {
```
freben, line 52:
```suggestion
  const facetsPromise = useMemo<Promise<Entity[]>>(async () => {
```

and then you can greatly simplify everything below

plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
```diff
@@ -41,19 +41,37 @@ type FacetsInitialRequest = {
...
-      return [];
+      return Promise.resolve([]);
     }
-    const facet = 'relations.ownedBy';
 
-    return catalogApi
-      .getEntityFacets({ facets: [facet] })
-      .then(response =>
-        response.facets[facet]
+    return (async (): Promise<Entity[]> => {
```
freben, line 57: you don't need this extra wrapping now

plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
```diff
@@ -41,19 +41,37 @@ type FacetsInitialRequest = {
...
+      try {
+        if (selectedEntityKind) {
```
freben, line 59: no need for this extra `if`. Just make a single query always, and make it have

```ts
filter: selectedEntityKind ? { kind: selectedEntityKind } : undefined,
```

plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
```diff
@@ -41,19 +41,37 @@ type FacetsInitialRequest = {
...
+          const facetResponseFiltered = await catalogApi.getEntityFacets({
+            facets: ['relations.ownedBy'],
+            filter: { kind: [selectedEntityKind] },
+          });
+
+          if (facetResponseFiltered.facets['relations.ownedBy'].length === 0) {
```
freben, line 65: same with this `if`, the `map` below will handle it fine anyway

plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
Jona-Soerensen: Had to revert this one because the useEffect wont update if you want to switch from e.g. groups to users because both have avaiableOwner.length of 0.

Jona-Soerensen:
Refactored useFacetsEntities & EntityOwnerPicker to perform a single API call and display only the available owners.

- Consolidated the owner retrieval logic into a single API call

- Now, the picker only shows the **available owners** (based on the API response) instead of all owners

- Adjusted the reset behavior for group and user types by changing the useEffect in the EntityOwnerPicker

Open Question: Can the OwnerPicker only dispaly the available owners? 😃

vinzscam:
> When you say “skipping the filters at call time” do you mean that if we detect that a group or user kind has been selected, we completely avoid passing any filters to the api? So instead of updating the filter in React, we intercept the group and user entities (as proposed in the original bug ticket https://github.com/backstage/backstage/issues/18710#issuecomment-1663824019) and reset the filter accordingly.

yes exactly, basically accordingly filtering them out in all the api calls in [useEntityListProvider.tsx](https://github.com/backstage/backstage/blob/36d9578211e6634ff42ebb7f6fa818e53226beb8/plugins/catalog-react/src/hooks/useEntityListProvider.tsx#L263).

vinzscam:
> Refactored useFacetsEntities & EntityOwnerPicker to perform a single API call and display only the available owners.
> 
> * Consolidated the owner retrieval logic into a single API call
> * Now, the picker only shows the **available owners** (based on the API response) instead of all owners
> * Adjusted the reset behavior for group and user types by changing the useEffect in the EntityOwnerPicker
> 
> Open Question: Can the OwnerPicker only dispaly the available owners? 😃

the `EntityOwnerPicker` displays the available owners only if the `mode` prop is set to `owners-only`.

.changeset/real-news-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-react': patch
+---
+
+Fixed the `EntityOwnerPicker`component to dynamically update the `availableOwners` based on the selected Entity kind.
```
vinzscam, line 5: mmm I'm not sure this is the fix for #18710.
It seems the issue we want to address is that if I have already selected an owner and then switch the kind to users or groups, the owner filter shouldn't prevent me from seeing the available groups or users. However, the "available owners" feature should remain as it is.

This means that if an adopter sets `<CatalogIndexPage ownerPickerMode='all' />` in their `App.tsx`, we should display all users/groups present in the catalog. On the other hand, if they set `<CatalogIndexPage ownerPickerMode='owners-only' />`, we should only display users/groups that are assigned as the owner of at least one entity.

vinzscam (changes requested):

.changeset/real-news-marry.md
Jona-Soerensen: Thank you. I just created a new [PR](https://github.com/backstage/backstage/pull/29678) wich just adds a filter for the "user" & "groups" entity kind inside the useEntityListProvider. This fixes the Bug without touching the state management.
https://github.com/backstage/backstage/pull/29678

Jona-Soerensen:
Created a new PR for the requested changes. 
https://github.com/backstage/backstage/pull/29678

camilaibs:
Hi @Jona-Soerensen 👋🏻 , should this pull request be closed in favor of [this](https://github.com/backstage/backstage/pull/29678) other one?

Jona-Soerensen:
Yes! Let’s go ahead and close this PR in favor of the other one 🚀

benjdlambert:
Thanks @Jona-Soerensen :tada:

PR #28564, opened 2025-01-21
Debounce rerender when user resize quickly

toantd90:

When users quickly resize the catalog page window, it causes the <Filter /> component to rerender. Rerendering triggers several unnecessary API calls. This PR adds some delays to prevent unnecessary rerender and API calls.

### DIFF
--- a/plugins/catalog-react/src/components/CatalogFilterLayout/CatalogFilterLayout.tsx
+++ b/plugins/catalog-react/src/components/CatalogFilterLayout/CatalogFilterLayout.tsx
@@ -23,6 +23,7 @@ import Typography from '@material-ui/core/Typography';
 import useMediaQuery from '@material-ui/core/useMediaQuery';
 import { Theme, useTheme } from '@material-ui/core/styles';
 import FilterListIcon from '@material-ui/icons/FilterList';
+import useDebounce from 'react-use/esm/useDebounce';
 import { catalogReactTranslationRef } from '../../translation';
 import { useTranslationRef } from '@backstage/core-plugin-api/alpha';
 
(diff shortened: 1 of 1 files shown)


vinzscam (changes requested):
mmm I think a good improvement would be to avoid re-fetching the data when the user resizes the screen. I feel like this logic adds extra logic that doesn't bring a real benefit.

toantd90:
@vinzscam Thank you for your review. 

I couldn't find a solution that avoids re-fetching the data when the user resizes the screen quicj=k.

The filter will get pass as {props.children} [here](https://github.com/backstage/backstage/blob/master/plugins/catalog-react/src/components/CatalogFilterLayout/CatalogFilterLayout.tsx#L71) and [here](https://github.com/backstage/backstage/blob/master/plugins/catalog-react/src/components/CatalogFilterLayout/CatalogFilterLayout.tsx#L77) .

When the screen width falls below a certain [threshold](https://github.com/backstage/backstage/blob/master/plugins/catalog-react/src/components/CatalogFilterLayout/CatalogFilterLayout.tsx#L46), the application will render a Drawer instead of maintaining the filters alongside the Catalog list. All filter components that we passed [here](https://github.com/sg-stargate/dev-portal/blob/main/plugins/catalog-fork/src/components/CatalogPage/DefaultCatalogPage.tsx#L149) will get unmounted and then remounted to restore their initial state. From a React perspective, this unmounting and remounting occurs because the structure of the components changes. When React mounts the Filter component from scratch, it triggers the `useEffect` in each component, leading to data fetching. This behavior is expected.

Most users don't frequently resize their windows. I only notice unnecessary data fetching when there's an issue with the API itself. However, this is an area where we can improve. We could reduce the delay time to 200 ms to strike a balance between a smooth user interface and minimizing API calls.

If you have any better suggestions, please share them!

freben:
I still wonder. What does one do to trigger this condition?

First you grab an edge of the window, right? Then you ... shake it side to side to make it quickly go between small and large size? Or does it happen if you move the mouse smoothly from large to small window? I tried this on the demo site and yes it happens exactly once that it reloads itself as i made the window smaller (and again when I made it larger). Making that happen more than five times per second (making it more heavy than the limit you set in this pr) seems to require some effort.

freben:
It would seem that the real solution is to somehow make those effects not run again in the first place. I'm surprised react does that just because of the parent tree change - it really shouldn't, since the children are created "outside" of this component and just passed through unmodified. Something else is going on here

PR #28570, opened 2025-01-21
Canon - Fix responsive capacity

cdedreuille:
- Removed the capacity to edit the list of breakpoints from the provider.
- Reduce the number of re-renders across multiple breakpoints.
- Update list of of breakpoints to clarify the. initial value (Inspired from Radix theme)

### DIFF
--- a/packages/canon/src/utils/getBreakpoint.ts
+++ b/packages/canon/src/utils/getBreakpoint.ts
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2025 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { useMediaQuery } from '../hooks/useMediaQuery';
+import type { Breakpoint } from '../types';
+
+export const breakpoints: { name: string; id: Breakpoint; value: number }[] = [
+  { name: 'Initial', id: 'initial', value: 0 },
+  { name: 'Extra Small', id: 'xs', value: 640 },
+  { name: 'Small', id: 'sm', value: 768 },
+  { name: 'Medium', id: 'md', value: 1024 },
+  { name: 'Large', id: 'lg', value: 1280 },
+  { name: 'Extra Large', id: 'xl', value: 1536 },
+];
+
+export const getBreakpoint = (): Breakpoint => {
+  const queries = breakpoints.map(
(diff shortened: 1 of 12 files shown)


packages/canon/src/utils/getBreakpoint.ts
```diff
@@ -0,0 +1,42 @@
...
+  { name: 'Large', id: 'lg', value: 1280 },
+  { name: 'Extra Large', id: 'xl', value: 1536 },
+];
+
+export const getBreakpoint = (): Breakpoint => {
+  const queries = breakpoints.map(
+    breakpoint => `(min-width: ${breakpoint.value}px)`,
+  );
+
+  const matches = queries.map(query => useMediaQuery(query));
```
Rugvip, line 33: You can't call a hook in a loop like this, and you'd need to signal that `getBreakpoint` is a gook as well by naming it `useBreakpoint`.

packages/canon/src/utils/getResponsiveValue.ts
```diff
@@ -0,0 +1,45 @@
...
+  const currentBreakpoint = getBreakpoint();
+
+  if (typeof value === 'object') {
+    const index = breakpoints.findIndex(
+      breakpoint => breakpoint.id === currentBreakpoint,
+    );
+
+    for (let i = index; i >= 0; i--) {
+      if (value[breakpoints[i].id]) {
+        // console.log(value[breakpoints[i].id]);
```
Rugvip, line 31: 🧹

packages/canon/src/utils/getBreakpoint.ts
```diff
@@ -0,0 +1,41 @@
...
+export const breakpoints: { name: string; id: Breakpoint; value: number }[] = [
+  { name: 'Initial', id: 'initial', value: 0 },
+  { name: 'Extra Small', id: 'xs', value: 640 },
+  { name: 'Small', id: 'sm', value: 768 },
+  { name: 'Medium', id: 'md', value: 1024 },
+  { name: 'Large', id: 'lg', value: 1280 },
+  { name: 'Extra Large', id: 'xl', value: 1536 },
+];
+
+export const getBreakpoint = (): Breakpoint => {
```
Rugvip, line 28: Needs to be called `useBreakpoint` and be placed in `hooks` since it uses a hook itself.

packages/canon/src/utils/getBreakpoint.ts
```diff
@@ -0,0 +1,41 @@
...
+  const matches = breakpoints.map(breakpoint => {
```
Rugvip, line 29: Probably worth a comment to point out that this is safe given the static array of breakpoints defined just above. I do think it would be cleaner to let `useMediaQuery` take an array in some way though.

packages/canon/src/utils/getResponsiveValue.ts
cdedreuille: Removed the console.log

packages/canon/src/utils/getBreakpoint.ts
cdedreuille: Done

packages/canon/src/utils/getBreakpoint.ts
cdedreuille: Just added it.

Rugvip (approved):
:shipit: 👍

Issue #26662, opened 2024-09-12
🐛 Bug Report: Techdocs retains scroll state after navigating in entity view


lintaod-hub:
### 📜 Description

When users navigating via hyperlinks in Techdocs under catalog entity tab, the scroll level stays the same after the navigation.

### 👍 Expected behavior

Scroll level should reset to top of the page after navigating to a new page.

### 👎 Actual Behavior with Screenshots

![techdocs](https://github.com/user-attachments/assets/6b39097a-ac29-442f-82f9-2afa93a46b60)

### 👟 Reproduction steps

Using the demo site as example: 
1. Visit https://demo.backstage.io/catalog/default/component/backstage/docs/plugins/.
2. Scroll down to section `Creating a plugin` and click on the hyperlink with text `here`.
3. The new page is opened with scroll level same as previous page.

### 🖥️ Your Environment

```
OS:   Darwin 23.6.0 - darwin/arm64
node: v18.20.2
yarn: 1.22.19
cli:  0.26.11 (installed)
backstage:  1.29.2

Dependencies:
  @backstage/app-defaults                                          1.5.9
  @backstage/backend-app-api                                       0.8.0
  @backstage/backend-common                                        0.23.3
  @backstage/backend-defaults                                      0.4.1
  @backstage/backend-dev-utils                                     0.1.4
  @backstage/backend-openapi-utils                                 0.1.15
  @backstage/backend-plugin-api                                    0.6.21, 0.7.0, 0.8.0
  @backstage/backend-tasks                                         0.5.27
  @backstage/backend-test-utils                                    0.4.4
  @backstage/catalog-client                                        1.6.5, 1.6.6
  @backstage/catalog-model                                         1.5.0, 1.6.0
  @backstage/cli-common                                            0.1.14
  @backstage/cli-node                                              0.2.7
  @backstage/cli                                                   0.26.11
  @backstage/config-loader                                         1.8.1
  @backstage/config                                                1.2.0
  @backstage/core-app-api                                          1.14.1, 1.14.2
  @backstage/core-compat-api                                       0.2.7, 0.2.8
  @backstage/core-components                                       0.13.10, 0.14.9, 0.14.10
  @backstage/core-plugin-api                                       1.9.3
  @backstage/dev-utils                                             1.0.36
  @backstage/e2e-test-utils                                        0.1.1
  @backstage/errors                                                1.2.4
  @backstage/eslint-plugin                                         0.1.8
[...]

PR #28577, opened 2025-01-21
fix(techdocs): scroll bar retains position after changing page

alexlorenzi:

As reported in #26662

Override the BackstagePage default styles to avoid a "double scrollbar" situation that stopped the page from fully scrolling to the top of the page when we navigated to a new page

Before (Double Scrollbar)

https://github.com/user-attachments/assets/a703ebb1-6c5c-4630-8dbb-5de53403e704

After 

https://github.com/user-attachments/assets/5aa7b280-1fa1-4c9c-8a44-328eba9b28a7

### DIFF
--- a/.changeset/chatty-turkeys-brake.md
+++ b/.changeset/chatty-turkeys-brake.md
@@ -1,0 +1,5 @@
+---
+'@backstage/core-components': patch
+---
+
+Added `classNames` prop to the `Page` component
--- a/.changeset/bright-pets-agree.md
+++ b/.changeset/bright-pets-agree.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-techdocs': patch
+---
+
+Fixed double scrollbar issue that would appear on the Entity TechDocs view page that would stop the page from full scrolling to the top when navigating to a new page
--- a/packages/core-components/src/layout/Page/Page.tsx
+++ b/packages/core-components/src/layout/Page/Page.tsx
@@ -51,7 +51,7 @@ type Props = {
 
 export function Page(props: Props) {
   const { themeId, className, children } = props;
-  const styles = useStyles();
+  const classes = useStyles();
   return (
     <ThemeProvider
       theme={(baseTheme: Theme) => ({
@@ -59,7 +59,7 @@ export function Page(props: Props) {
         page: baseTheme.getPageTheme({ themeId }),
       })}
     >
-      <main className={classNames(styles.root, className)}>{children}</main>
+      <main className={classNames(classes.root, className)}>{children}</main>
     </ThemeProvider>
   );
 }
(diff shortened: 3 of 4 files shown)


Rugvip (changes requested):
Hmm, not a precedent that we'd want to set to do these kind of theme overrides.

Would it work to instead add a `className` prop to the `Page` and style it like that?

alexlorenzi:
@Rugvip 
> Would it work to instead add a `className` prop to the `Page` and style it like that?

Yeah, that's the issue. The `Page` component doesn't allow styling, so I need to use the Theme.
My only other thought to avoid this was to do something like
```javascript
...
document?.querySelector('header')?.scrollIntoView();
document.querySelector('.techdocs-reader-page > main')?.scrollTo(0,0)
...
```

here https://github.com/backstage/backstage/blob/93e7ac2e735aede3634a6c6cb64480cc50dcab05/plugins/techdocs/src/reader/components/TechDocsReaderPageContent/TechDocsReaderPageContent.tsx#L101-L111

But that seems a bit tightly coupled with the classname across distant files

Rugvip:
@alexlorenzi yep but I think it's fine to add a `className` prop to the `Page` component

alexlorenzi:
https://github.com/user-attachments/assets/307c29ce-da00-454a-9a43-0a444c358477

Seems to be some strange behaviour if I use the `className` to attach the styles. It loads on the first render of the page, but then when I switch techdoc pages it doesn't remain. I guess due to how we remove and add the DOM for the content... If I use override it with the `style` tag instead it works though, if that's an acceptable alternative?

Rugvip:
@alexlorenzi `className` would be better tbh, or `classes`. Could you push those changes?

alexlorenzi:
Override via that classes didn't have any effect either. Pushed the classname version as that's the most straight forward.

alexlorenzi:
@Rugvip - Switched this to use the styled component API instead, which seems to solve the issue of the CSS getting removed when switching between TechDoc pages. So now we're able to pass the styles down via the classnames. Also has the benefit of being able to remove the odd placeholder div that was there to allow us to insert a class that some javascript hooked onto, and join it with the Page component.

awanlin:
Hi @alexlorenzi @Rugvip, I know this has been an issue in TechDocs for some time now, would really like to see this fixed. I've re-opened this PR seeing as it got closed as stale, what else is needed to move it forward?

Rugvip (commented):
👍 , just needs changesets for each of the packages

.changeset/chatty-turkeys-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/core-components': patch
```
alexlorenzi, line 2: Wasn't sure if this was enough to warrent a minor bump?

Rugvip (approved):
Nice! 👍 

Let's :shipit:

Issue #28080, opened 2024-12-10
🚀 Feature: Update to React 19


schultzp2020:
### 🔖 Feature description

Requirements

- [x] Deprecate React 16
- [ ] Migrate to the [new JSX transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)
  - React 17 and 18 support the new transform
- [ ] Follow the [migration guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide) 
- [ ] Update peer dependency ranges to include React 19

Blockers

- [ ] MUI v4 does not support React 18 and 19

### 🎤 Context

We need to update to the latest version. 

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

hairmare:
there is a potential duplicate issue here: https://github.com/backstage/backstage/issues/24554

awanlin:
Added the `no stale` label as we do intend to do this upgrade but it will take time as there are a few moving parts to tackle first.

Thanks for keeping this moving forward @schultzp2020 🚀

awanlin:
I've updated this in the ticket description but posting here as well:

> This work is blocked by Material UI (MUI) v4 which does not support newer versions of React. Our decision as a project to move away from MUI in favor of our own design system for Backstage - [Backstage UI  (BUI)](https://ui.backstage.io/) - will help us address this. Once we have fully migrated to BUI we will be able to move forward with removing MUI v4 and begin this upgrade. We've create a [BUI migration tracker to help track the progress](https://github.com/backstage/backstage/issues/31467).

PR #28579, opened 2025-01-21
chore: deprecate old react versions

schultzp2020:

Removed older versions of React packages as a preparatory step for upgrading to React 19. This commit does not introduce any functional changes, but removes dependencies on previous React versions, allowing for a cleaner upgrade path in subsequent commits.

This PR is marked as a patch because it’s reasonable to assume that downstream users are using React 17 or React 18. A subsequent commit will include code changes that necessitate at least a minor version bump.

Related https://github.com/backstage/backstage/issues/28080

### DIFF
--- a/packages/app-defaults/package.json
+++ b/packages/app-defaults/package.json
@@ -55,10 +55,10 @@
     "react-router-dom": "^6.3.0"
   },
   "peerDependencies": {
-    "@types/react": "^16.13.1 || ^17.0.0 || ^18.0.0",
-    "react": "^16.13.1 || ^17.0.0 || ^18.0.0",
-    "react-dom": "^16.13.1 || ^17.0.0 || ^18.0.0",
-    "react-router-dom": "6.0.0-beta.0 || ^6.3.0"
+    "@types/react": "^17.0.0 || ^18.0.0",
+    "react": "^17.0.0 || ^18.0.0",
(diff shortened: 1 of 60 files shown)


packages/app-defaults/package.json
```diff
@@ -50,15 +50,15 @@
     "@testing-library/jest-dom": "^6.0.0",
     "@testing-library/react": "^16.0.0",
     "@types/react": "^18.0.0",
-    "react": "^18.0.2",
-    "react-dom": "^18.0.2",
-    "react-router-dom": "^6.3.0"
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
+    "react-router-dom": "^6.0.0"
```
Rugvip, line 55: Kinda curious why we'd lower the ranges on these and the peer deps?

Rugvip (commented):
Nice, thank you! 🎉 

Some Qs and wanna discuss this with @backstage/maintainers a bit too

packages/app-defaults/package.json
Rugvip: Let's keep the most common minimum versions for now, so `18.0.2` for react and `6.3.0` for react-router, and use those across the entire project.

freben (approved):

Rugvip (approved):

PR #28586, opened 2025-01-22
Update 02-build-system.md

gusega:
documentation how to run jest tests in intellij

Update part of the doc how to run jest tests in intellij

### DIFF
--- a/.github/vale/config/vocabularies/Backstage/accept.txt
+++ b/.github/vale/config/vocabularies/Backstage/accept.txt
@@ -513,6 +513,4 @@ zoomable
 zsh
 scrollable
 severities
-Intellij
 intellij
-IntelliJ
--- a/.github/vale/config/vocabularies/Backstage/accept.txt
+++ b/.github/vale/config/vocabularies/Backstage/accept.txt
@@ -515,3 +515,4 @@ scrollable
 severities
 Intellij
 intellij
+IntelliJ
--- a/.github/vale/config/vocabularies/Backstage/accept.txt
+++ b/.github/vale/config/vocabularies/Backstage/accept.txt
@@ -513,3 +513,5 @@ zoomable
 zsh
 scrollable
 severities
+Intellij
+intellij
--- a/docs/tooling/cli/02-build-system.md
+++ b/docs/tooling/cli/02-build-system.md
@@ -596,36 +596,16 @@ Caching is used sparingly throughout the Backstage build system. It is always us
 
 For your productivity working with unit tests it's quite essential to have your debugging configured in IDE. It will help you to identify the root cause of the issue faster.
 
-#### Preface
+We cannot execute tests with just raw `jest`, because there are a few concerns such as module transforms that need to be in place for the `jest` runtime to be happy.
+Therefore, we delegate to the Backstage CLI to wrap the jest run for us, since it knows how to put those things in place.
+This aligns things so that your in-IDE test runs work the same way as your CI and manual command line test runs do.
 
-At the moment it's not possible to run jest tests via jest command line. That is, if you run the following command from the project root: 
-```bash
-NODE_OPTIONS='--no-node-snapshot --experimental-vm-modules' node node_modules/jest-cli/bin/jest.js --config node_modules/@backstage/cli/config/jest.js
-```
-
-You will get an error `SyntaxError: Unexpected token 'export'` complaining about one of the imports from the `node_modules/@backstage/cli/config/jest.js` config file. 
-But why does jest work then when running `backstage-cli repo test`?
-The reason is that `packages/cli/bin/backstage-cli` first executes `require('@backstage/cli/config/nodeTransform.cjs');` that probably transforms config files so that jest can digest them.
-Even simply adding `require('@backstage/cli/config/nodeTransform.cjs');` to the beginning of `node_modules/jest/bin/jest.js`, will make jest work in the script above.
-
-Happily, there is a better way to do it, because `backstage-cli repo test` is simply a starter of the `jest-cli` with extra parameters. That is, you can run jest from the command line via `backstage-cli repo test` like this:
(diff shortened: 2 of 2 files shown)


.github/vale/config/vocabularies/Backstage/accept.txt
```diff
@@ -513,3 +513,6 @@ zoomable
 zsh
 scrollable
 severities
+Intellij
+intellij
+IntelliJ
```
freben, line 518:
```suggestion
intellij
```

if you have an all-lowercase version, it automatically means "accept any mixed case too"

docs/tooling/cli/02-build-system.md
```diff
@@ -596,22 +596,53 @@ Caching is used sparingly throughout the Backstage build system. It is always us
...
+```
+or
+```bash
+NODE_OPTIONS='--no-node-snapshot --experimental-vm-modules' node node_modules/@backstage/cli/bin/backstage-cli repo test
+```
+
+
+`backstage-cli` adds `--passWithNoTests --watch --workerIdleMemoryLimit=1000M --config node_modules/@backstage/cli/config/jest.js` jest options.
+
+With that in mind, let's configure IDE's to run `backstage-cli` in the role of jest.
```
freben, line 623:
```suggestion
With that in mind, let's configure IDEs to run `backstage-cli` in the role of jest.
```

maybe that passes the spell checker

docs/tooling/cli/02-build-system.md
```diff
@@ -596,22 +596,53 @@ Caching is used sparingly throughout the Backstage build system. It is always us
 
 For your productivity working with unit tests it's quite essential to have your debugging configured in IDE. It will help you to identify the root cause of the issue faster.
 
+#### Preface
+
+At the moment it's not possible to run jest tests via jest command line. That is, if you run the following command from the project root:
```
freben, line 601: All of this is a lot of text and background and technical details. Not sure it helps the reader. 🤔 Could it be shortened to just convey the high level point? Something like

```
We cannot execute tests with just raw `jest`, because there are a few concerns such as module transforms that need to be in place for the runtime to be happy. Therefore, we delegate to the Backstage CLI to wrap the jest run for us, since it knows how to put those things in place. This aligns things so that your in-IDE test runs work the same way as your CI and manual command line test runs do.
```

That's basically all that you need to know at this point. The rest of this section can be skipped IMO, which also makes it less text to maintain since all of the technicality in there is subject to change over time.

docs/tooling/cli/02-build-system.md
```diff
@@ -596,22 +596,33 @@ Caching is used sparingly throughout the Backstage build system. It is always us
...
 
-2. Now you can run any tests by clicking on green arrow located on `describe` or `it`.
+    Until intellij maintainers resolved the issue, create a jest configuration manually. Happily, intellij will pre-fill the configuration from the template. The only thing you need to do is provide path to the test file. Note, that after intellij runs test in the file you can click on the individual tests from the run panel and re-run them, this time intellij will create a correct jest run configuration.
 
 #### VS Code
 
 ```jsonc
 {
-  "jest.jestCommandLine": "node_modules/.bin/jest --config node_modules/@backstage/cli/config/jest.js",
+  "jest.jestCommandLine": "node_modules/.bin/backstage-cli --config node_modules/@backstage/cli/config/jest.js",
```
freben, line 625:
```suggestion
  "jest.jestCommandLine": "yarn test",
```

right? The user should already have a properly set-up `yarn test` in their `package.json` which likely invokes `backstage-cli` etc properly. And, then, likely also including all of the NODE_OPTIONS etc.

[This](https://github.com/backstage/backstage/blob/ed8b541ea90ad17361dc9f395a0b0a5005508ddc/package.json#L52) is what our current setup looks like.

This is the only changed VScode setting I have locally.

Screenshot 2025-01-28 at 15 15 43


With this I can run tests, but actually not debug them - but that is probably a separate issue.

docs/tooling/cli/02-build-system.md
```diff
@@ -596,22 +596,33 @@ Caching is used sparingly throughout the Backstage build system. It is always us
...
-- In "Jest package" you have to point to relative path of jest module (it will be suggested by IntelliJ), i.e. `~/proj/backstage/node_modules/jest`
-- In "Jest config" point to your jest configuration file, use absolute path for that, i.e. `--config /Users/user/proj/backstage/packages/cli/config/jest.js --runInBand`
+2.  Currently, intellij has an issue that if you right-click on a jest test and press "run", intellij will create a playwright run configuration instead of jest configuration, see [WEB-67720](https://youtrack.jetbrains.com/issue/WEB-67720/Jest-test-runs-as-playwright-test).
 
-2. Now you can run any tests by clicking on green arrow located on `describe` or `it`.
+    Until intellij maintainers resolved the issue, create a jest configuration manually. Happily, intellij will pre-fill the configuration from the template. The only thing you need to do is provide path to the test file. Note, that after intellij runs test in the file you can click on the individual tests from the run panel and re-run them, this time intellij will create a correct jest run configuration.
```
freben, line 619:
```suggestion
    Until intellij maintainers resolve the issue, create a jest configuration manually. Happily, intellij will pre-fill the configuration from the template. The only thing you need to do is provide a path to the test file. Note, that after intellij runs test in the file you can click on the individual tests from the run panel and re-run them, this time intellij will create a correct jest run configuration.
```

docs/tooling/cli/02-build-system.md
gusega: I don't use VSCode, so I just put changes similar to what I did for intellij. If it works in VSCode great, I will accept your suggestion. For debuging, probably the configuration below works, as it's more or less what I have for intellij (and it was there before, so someone might have tested it).

freben (approved):

PR #28612, opened 2025-01-23
refactor(CatalogTable): allow overriding default components

kuangp:

Proposing this change as a workaround to address an issue caused by https://github.com/backstage/backstage/pull/27632/files which overrode the default Material table Toolbar component with the new custom [CatalogTableToolbar](https://github.com/backstage/backstage/blob/master/plugins/catalog/src/components/CatalogTable/CatalogTableToolbar.tsx) implementation. This resulted in breaking/removing some built-in Material table features related to the native toolbar such as [`columnsButton`](https://github.com/material-table-core/core/blob/master/src/components/MTableToolbar/index.js#L137), [`exportMenu`](https://github.com/material-table-core/core/blob/master/src/components/MTableToolbar/index.js#L205), etc. which are not currently implemented by the custom `CatalogTableToolbar`. This change allows consumers to revert back to using the Material toolbar component if they were relying on these missing features. 

Also open to suggestions for a better solution here - the new `CatalogTableToolbar` component seems useful since it allows persisting search terms to the URL but unfortunately lacks the rest of the functionality of the Material table toolbar and it doesn't seem like we are able to override just the search component of the built-in toolbar while reusing/inheriting the rest of the existing implementation 🙁. Alternatively `CatalogTableToolbar` can fork over the rest of the features of the built-in toolbar but not sure if that's a good option. Another option is to propose a change to the Material table project to allow the search component in the toolbar to be overridden 🤔  

### DIFF
--- a/plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
+++ b/plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
@@ -67,6 +67,7 @@ export interface CatalogTableProps {
    */
   title?: string;
   subtitle?: string;
+  components?: TableProps<CatalogTableRow>['components'];
 }
 
 const refCompare = (a: Entity, b: Entity) => {
@@ -95,6 +96,7 @@ export const CatalogTable = (props: CatalogTableProps) => {
     tableOptions,
     subtitle,
     emptyContent,
+    components,
   } = props;
   const { isStarredEntity, toggleStarredEntity } = useStarredEntities();
   const entityListContext = useEntityList();
@@ -216,6 +218,7 @@ export const CatalogTable = (props: CatalogTableProps) => {
         data={entities.map(toEntityRow)}
         next={pageInfo?.next}
         prev={pageInfo?.prev}
+        components={components}
       />
     );
   } else if (paginationMode === 'offset') {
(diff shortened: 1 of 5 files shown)


kuangp:
@Rugvip 👈 gentle nudge

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
```diff
@@ -249,6 +253,7 @@ export const CatalogTable = (props: CatalogTableProps) => {
       }}
       components={{
         Toolbar: CatalogTableToolbar,
+        ...(components ?? {}),
```
vinzscam, line 256: I think `CatalogTableToolbar` shouldn't be used here as it isn't supposed to be used for client side paginated table.
Would removing the line above fix your issue @kuangp?

vinzscam (changes requested):

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
kuangp: Yes! that would fix it for us since we don't currently use server side pagination - I'll remove this as a "workaround" for now however the original issue I described will still exist for those who do use server side pagination and wish to use built-in Material Table toolbar features (in addition to the inconsistency between the table impls since it would've been nice to have search query URL persistence here as well) but I guess than can be revisited in a subsequent change

vinzscam (approved):
🙏

PR #28619, opened 2025-01-24
frontend-plugin-api: add support for multiple attachTos

Rugvip:

Draft for now while we discuss whether this is needed. This allows for extensions to be cloned through definition of multiple attachment points. The purpose of this is to provide an alternative solution for when extensions are used in multiple parts of the app, for example [scaffolder field extensions](https://github.com/backstage/backstage/blob/78aff31f0dc96399e3783e8f20f8d420b6745a5d/plugins/scaffolder-react/src/next/blueprints/FormFieldBlueprint.tsx#L35-L45) or TechDocs addons. Our current approach is to use Utility APIs for these extensions, but looking into it a bit more I'm not sure that's the best or most intuitive solution.

This also is a bit of a trial of cloning extensions in general, where we could potentially allow for extensions to be cloned through config as well. This PR does not implement that.

### DIFF
--- a/.changeset/three-glasses-sell.md
+++ b/.changeset/three-glasses-sell.md
@@ -1,0 +1,6 @@
+---
+'@backstage/frontend-plugin-api': patch
+'@backstage/frontend-app-api': patch
+---
+
+Add support for defining multiple attachment points for extensions and blueprints.
--- a/docs/frontend-system/architecture/20-extensions.md
+++ b/docs/frontend-system/architecture/20-extensions.md
@@ -337,3 +337,21 @@ const routableExtension = createExtension({
   },
 });
 ```
+
+## Multiple attachment points
+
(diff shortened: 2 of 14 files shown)


PeaWarrior:
This would be an amazing addition! We're actually running into a usecase for this where we need to attach TechDocs addons to separate pages such as the standalone techdoc pages and entity doc pages.

freben (approved):

freben:
Random thought: I wonder if there's some interplay with aliases / redirects here. Being able to say that "things that attach themselves to A, should also be attached to B", which enables you to make e.g. a secondary closed source catalog root based page, that automatically gets all filters on it without you having to make overrides for all of them one by one.

Rugvip:
@freben yep I think this does open up for useful new features and a can of worms at the same time. There's for sure more that can be built on top of this but I think its best to take it slow

PR #28621, opened 2025-01-24
Add configuration to permit backend startup failure

timbonicus:
Adds a `backend.startup` configuration to support startup options for backend plugins. The only initial option is an `optional` flag — when set, startup failures for that plugin are permitted rather than being fatal to the entire backend process (as they are today).

The startup failures will still be logged, but the rest of the backend will not be prevented from starting up. This can be useful for troubleshooting, or for environments where application availability is more critical than a given plugin starting up.


.changeset/twenty-carrots-jog.md
```diff
@@ -0,0 +1,17 @@
...
+---
+
+Added a configuration to permit backend plugin failures on startup:
+
+```yaml
+backend:
+  ...
+  startup:
+    plugin-x:
+      optional: true
```
freben, line 12: In a couple of places we have added an extra `plugins:` nesting level under the parent (`startup` in this case). This lets you add defaults and more options under `startup`. Would that make sense here?

freben (commented):
Hm. I wonder if this has an overlap with https://github.com/backstage/backstage/pull/28129, which could be returning feature _state_ as well - including if it's starting / started / failed startup / is shutting down, and the relevant error thrown if relevant.

.changeset/twenty-carrots-jog.md
Rugvip: Yep probably best in this case too, so that we could potentially do things like this?

```
backend:
  startup:
    defaults:
      optional: true
    plugins:
      catalog:
        optional: false
```

packages/backend-app-api/config.d.ts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { BackendStartupOptions } from './src';
```
Rugvip, line 17: This doesn't work in practice, you'll need to repeat the types here

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -336,6 +336,21 @@ export class BackendInitializer {
...
+
+    const startupOptions =
+      rootConfig
+        ?.getOptionalConfig('backend.startup')
+        ?.get<BackendStartupOptions>() ?? {};
+
+    // Gather backend pluginIds marked as optional, since these should not cause a startup failure
+    const optionalBackends = Object.entries(startupOptions)
+      .filter(([_, value]) => value?.optional)
+      .map(([key]) => key);
```
Rugvip, line 352: I think it would be nice to encapsulate this a little bit, just something simple like `const isPluginOptional = this.#getPluginOptionalityPredicate()`?

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -336,6 +336,21 @@ export class BackendInitializer {
...
 
+    const rootConfig = await this.#serviceRegistry.get(
+      coreServices.rootConfig,
+      'root',
+    );
+
+    const startupOptions =
+      rootConfig
+        ?.getOptionalConfig('backend.startup')
+        ?.get<BackendStartupOptions>() ?? {};
```
Rugvip, line 347: Avoid using `get()` since it doesn't give as good feedback for invalid config or will allow invalid config. Switching this to `.getOptionalBoolean(...)` would give a clean error message if there's an unexpected value in config.

freben (approved):
OK I'm happy now in terms of code.

Only question I have is whether the "optional" wording explains the scope well, and booleans can be iffy (hah!). Could just as well mean "does not have to be installed" to someone who looks at the config file. Something like `onPluginBootFailure: 'abort' | 'continue'` might have been clearer and not risking to "leak" into other concerns. But I won't block on it.

timbonicus:
@freben I like that better too, updated in https://github.com/backstage/backstage/pull/28621/commits/42e07c36fb27fd750b285960768a7e078343aa26. Thanks for the suggestion!

webark:
@timbonicus followed this pattern and added an option to allow for individual modules to be permitted to fail without causing the whole plugin to fail.

https://github.com/backstage/backstage/pull/29348

PR #28626, opened 2025-01-27
Canon - Remove client directive

cdedreuille:
I'm removing the 'use client' directive in Canon and fixed all pages in our docs to make sure we can still take advantage of server components in some cases.


Rugvip (commented):
Looks good, but now that we're publishing the package it's time to start writing changesets :grin: 

Most in-depth guide to that is over here: https://github.com/backstage/backstage/blob/master/REVIEWING.md#changesets

cdedreuille:
> Looks good, but now that we're publishing the package it's time to start writing changesets 😁
> 
> Most in-depth guide to that is over here: https://github.com/backstage/backstage/blob/master/REVIEWING.md#changesets

Ah you're right. I just added a patch for this update.

Rugvip (approved):
:shipit:

PR #28628, opened 2025-01-27
cli: more options for the new command

vinzscam:

This PR adds 3 new options to the cli `new` command:
-  `--plugin-prefix`: custome the prefix of the new plugin, which by default is `backstage-plugin-`.
- `--skip-install`: skips the installation of dependencies after creating a new plugin
- `-allowed-types`: Defines which plugin types should be available in the interactive prompt. This is useful if some types aren't applicable so that adopters could specify the allowed types in their `new` script in their `package.json`

### DIFF
--- a/.changeset/chilly-glasses-burn.md
+++ b/.changeset/chilly-glasses-burn.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+Added `--allowedOptions` to `new` command. This is useful for limiting the options shown in the interactive prompt.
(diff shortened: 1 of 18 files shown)


Rugvip (changes requested):
I think we should make sure that these requirements work their way into #27934, but I'd like to avoid moving this forward on its own due to the conflicts with #27934

vinzscam:
> I think we should make sure that these requirements work their way into #27934, but I'd like to avoid moving this forward on its own due to the conflicts with #27934

it's fine, just give #27934 priority 👍
I will deal with the conflicts

Rugvip:
Gonna close this one as it was all included in #27934

PR #28630, opened 2025-01-27
Canon - Update props structure

cdedreuille:
This PR restructure the way we handle props and css classes.
- Update CSS utility classes for `padding` and `margin`
- Rename padding and margin props to use short names like `p` or `px`
- Allow for custom values to be added
- Update breakpoints class name structure


.changeset/serious-gifts-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/canon': patch
```
Rugvip, line 2: Big set of breaking changes here so this should be `minor` and you'll want to prefix the message with `**BREAKING**:`

packages/canon/report.api.md
```diff
@@ -214,6 +289,49 @@ export type FlexDirection = 'row' | 'column';
 // @public (undocumented)
 export type FlexWrap = 'wrap' | 'nowrap' | 'wrap-reverse';
 
+// @public (undocumented)
+export const gapPropDefs: {
+  gap: {
+    type: 'enum | string';
+    className: string;
+    customProperties: '--gap'[];
```
Rugvip, line 297: Do these need to be part of the public API?

packages/canon/src/components/Box/Box.tsx
```diff
@@ -16,23 +16,35 @@
...
-  const classNames = clsx('canon-Box', utilityClassNames, className);
+  const propDefs = {
+    ...spacingPropDefs,
+    ...widthPropDefs,
+    ...heightPropDefs,
+    ...positionPropDefs,
+    ...displayPropDefs,
+    ...boxPropDefs,
+  };
+  const { className, style } = extractProps(props, propDefs);
```
Rugvip, line 40: Does look clean 👌 

Might even inline the `propDefs` into the `extractProps` call

packages/canon/src/components/Container/types.ts
```diff
@@ -19,11 +19,11 @@ import { SpaceProps } from '../../types';
...
-  marginTop?: SpaceProps['marginTop'];
-  paddingY?: SpaceProps['paddingY'];
-  paddingBottom?: SpaceProps['paddingBottom'];
-  paddingTop?: SpaceProps['paddingTop'];
+  my?: SpaceProps['my'];
+  mb?: SpaceProps['mb'];
+  mt?: SpaceProps['mt'];
+  py?: SpaceProps['py'];
+  pb?: SpaceProps['pb'];
+  pt?: SpaceProps['pt'];
```
Rugvip, line 27: ✨

packages/canon/src/css/utilities/height.css
```diff
@@ -0,0 +1,86 @@
...
+  min-height: var(--min-height);
+}
+
+.cu-max-h {
+  max-height: var(--max-height);
+}
+
+/* Breakpoint xs */
+@media (min-width: 640px) {
+  .xs\:cu-h {
```
Rugvip, line 15: How come these `\` are needed?

Rugvip (approved):
Nice! 🎉 couple small bits

Looks good overall, I didn't dig deep into the API and will leave that for trying it out as a whole a bit later.

packages/canon/src/css/utilities/height.css
cdedreuille: `:` and`.` can't be use without a backslash on css. This pattern is commonly used with Tailwind, Radix, ... This would not be something we should use for common css classes but utility classes that's fine.

packages/canon/report.api.md
cdedreuille: No you're right. I'll refactor that in a following PR. I was mostly trying to get the warnings out but I need to clean this up a little.

packages/canon/src/css/utilities/height.css
Rugvip: ah alright, so it's actually part of the class name rather than a pseudo class

PR #28631, opened 2025-01-27
createGitlabProjectMigrateAction returns migration id

mpsanchis:

The action `createGitlabProjectMigrateAction`, belonging to `scaffolder-backend-module-gitlab`, adds the `migrationId` to its outputs.

Ideally, I wanted to have the project ID in the output, so that future actions could act on the project (e.g.: migrate a project, and then modify its settings, or open an MR). However, the [bulk import API](https://docs.gitlab.com/ee/api/bulk_imports.html) does not return the project ID. Instead, it returns a migration object which has a status. It is unfortunate, but all that subsequent actions can do is poll Gitlab to check if migration has status `finished` (and hopefully, that'll mean that the project is already there).

### DIFF
--- a/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabProjectMigrate.ts
+++ b/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabProjectMigrate.ts
@@ -89,6 +89,10 @@ export const createGitlabProjectMigrateAction = (options: {
             title: 'URL to the newly imported repo',
             type: 'string',
           },
+          migrationId: {
+            title: 'Id of the migration that imports the project',
+            type: 'number',
+          },
         },
       },
     },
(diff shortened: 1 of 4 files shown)


plugins/scaffolder-backend-module-gitlab/src/actions/gitlabProjectMigrate.ts
```diff
@@ -134,8 +138,13 @@ export const createGitlabProjectMigrateAction = (options: {
         access_token: sourceAccessToken,
       };
 
+      let migrationId: number;
       try {
-        await api.Migrations.create(sourceConfig, migrationEntity);
+        const migrationStatus = await api.Migrations.create(
```
benjdlambert, line 143:
```suggestion
        const { id: migrationId } = await api.Migrations.create(
```

benjdlambert: And then also call `ctx.output` in this `try/catch` instead of calling it with `undefined`?

benjdlambert (changes requested):
Just a small nit :pray:

plugins/scaffolder-backend-module-gitlab/src/actions/gitlabProjectMigrate.ts
mpsanchis: 🫡 yes, sir! Change made

benjdlambert:
@mpsanchis looks like the tests are failing as the call to gitlab returns `undefined` so destructuring `{ id }` throws an errror. Can you fix these up please? :pray:

benjdlambert (approved):
nice! thank you! :pray: :tada:

PR #28633, opened 2025-01-27
Adds createWhenEmpty option to publish:github:pull-request action

tjuedema:

This PR expands the functionality of the `publish:github:pull-request` action by connecting the `createWhenEmpty` option from the action's underlying plugin to the action's template parameters.

The behavior of the `createWhenEmpty` option is decribed [here](https://github.com/gr2m/octokit-plugin-create-pull-request#:~:text=By%20default%2C%20a%20pull%20request%20is%20created%2C%20even%20if%20no%20files%20have%20been%20changed.%20To%20prevent%20an%20empty%20pull%20request%2C%20set%20options.createWhenEmpty%20to%20false.%20If%20no%20pull%20request%20has%20been%20created%2C%20octokit.createPullRequest()%20resolves%20with%20null.). 

> By default, a pull request is created, even if no files have been changed. To prevent an empty pull request, set options.createWhenEmpty to false. If no pull request has been created, octokit.createPullRequest() resolves with null.

This option is useful to have available in the use case where a template is called multiple times on a repository, and subsequent invocations of the template would otherwise produce empty pull requests.

### DIFF
--- a/.changeset/thick-moose-do.md
+++ b/.changeset/thick-moose-do.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-scaffolder-backend-module-github': patch
 ---
 
-Added oneOf output schema to publish:github:pull-request scaffolder action to allow for non-breaking addition of createWhenEmpty option
+Added `oneOf` output schema to `publish:github:pull-request` scaffolder action to allow for non-breaking addition of `createWhenEmpty` option
--- a/.changeset/thick-moose-do.md
+++ b/.changeset/thick-moose-do.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend-module-github': patch
+---
+
+Added oneOf output schema to publish:github:pull-request scaffolder action to allow for non-breaking addition of createWhenEmpty option
--- a/.changeset/thick-moose-do.md
+++ b/.changeset/thick-moose-do.md
@@ -1,5 +1,0 @@
----
-'@backstage/plugin-scaffolder-backend-module-github': patch
----
-
-Added `oneOf` output schema to `publish:github:pull-request` scaffolder action to allow for non-breaking addition of `createWhenEmpty` option
--- a/.changeset/wise-cycles-sparkle.md
+++ b/.changeset/wise-cycles-sparkle.md
@@ -3,4 +3,4 @@
 '@backstage/plugin-scaffolder': patch
 ---
 
-Added conditional rendering of oneOf output schemas on the Installed Actions page for scaffolder actions
+Added conditional rendering of `oneOf` output schemas on the Installed Actions page for scaffolder actions
--- a/.changeset/wise-cycles-sparkle.md
+++ b/.changeset/wise-cycles-sparkle.md
@@ -1,0 +1,6 @@
+---
+'@backstage/plugin-scaffolder-backend': patch
+'@backstage/plugin-scaffolder': patch
+---
+
+Added conditional rendering of oneOf output schemas on the Installed Actions page for scaffolder actions
--- a/plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts
+++ b/plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts
@@ -256,29 +256,53 @@ export const createPublishGithubPullRequestAction = (
             type: 'boolean',
             title: 'Create When Empty',
             description:
-              'Set whether to create pull request when there are no changes to commit. The default value is true',
+              'Set whether to create pull request when there are no changes to commit. The default value is true. If set to false, remoteUrl is no longer a required output.',
           },
         },
       },
       output: {
-        required: [],
-        type: 'object',
-        properties: {
-          targetBranchName: {
-            title: 'Target branch name of the merge request',
-            type: 'string',
+        oneOf: [
+          {
+            required: ['remoteUrl'],
+            type: 'object',
+            properties: {
+              targetBranchName: {
+                title: 'Target branch name of the merge request',
+                type: 'string',
+              },
+              remoteUrl: {
+                type: 'string',
+                title: 'Pull Request URL',
+                description: 'Link to the pull request in Github',
+              },
+              pullRequestNumber: {
+                type: 'number',
+                title: 'Pull Request Number',
+                description: 'The pull request number',
+              },
+            },
           },
-          remoteUrl: {
-            type: 'string',
-            title: 'Pull Request URL',
-            description: 'Link to the pull request in Github',
+          {
+            required: [],
+            type: 'object',
+            properties: {
+              targetBranchName: {
+                title: 'Target branch name of the merge request',
+                type: 'string',
+              },
+              remoteUrl: {
+                type: 'string',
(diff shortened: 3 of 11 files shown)


tjuedema:
It appears that something unrelated to my contribution is causing the E2E checks to fail. Can a maintainer take a look?

plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts
```diff
@@ -253,10 +254,16 @@ export const createPublishGithubPullRequestAction = (
...
+            type: 'boolean',
+            title: 'Create When Empty',
+            description:
+              'Set whether to create pull request when there are no changes to commit. The default value is true',
+          },
         },
       },
       output: {
-        required: ['remoteUrl'],
+        required: [],
```
freben, line 266: Was this unintentional? Looks unrelated at least.

plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts
```diff
@@ -419,6 +428,11 @@ export const createPublishGithubPullRequestAction = (
         }
         const response = await client.createPullRequest(createOptions);
 
+        if (createWhenEmpty === false && !response) {
+          ctx.logger.info('No changes to commit, pull request was not created');
+          return;
```
freben, line 433: Is it OK for this one to just ignore making any outputs and still report that it succeeded?

plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts
tjuedema: This is intentional. If createWhenEmpty is set to false and there are no changes to be PR'd, no PR will be created and therefore no remoteUrl will be output. Making the createWhenEmpty option available necessitates that the remoteUrl output be an optional output.

plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts
tjuedema: The action is still considered successful in its execution. Setting the createWhenEmpty parameter to false basically translate to "create a PR if it is needed, pass through if not".

plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts
benjdlambert: I think this change in its form is technically breaking as users would expect this to always return. We could however split this into a `oneOf` schema instead, that when you provide the new option that the return URL is optional.

Or we just simplify that and build then URL at runtime regardless of calling the repo create call. 

Could argue that the first option even though more complex is better than trying to guess what the output might be, and might even over time be more complex.

koalaty-code: It is technically breaking in terms of the schema. However, it would not break any existing templates as the new `createWhenEmpty` option defaults to `true` which will throw an error if we don't have a pull request.

I favor the `oneOf` schema if that doesn't add too much overhead for the consumer to understand.

> Or we just simplify that and build then URL at runtime regardless of calling the repo create call.

Are you thinking we could set `remoteUrl` to `https://${host}/${owner}/${repo}/pulls` if no pull request is opened? We currently send Backstage notifications to the owning group when a pull request is opened. We would like to do this conditionally based on a pull request being opened. This option wouldn't allow for that.

benjdlambert: My understanding is that `remoteUrl` is actually not the PR link, that's what the `prNumber` is output for.

https://github.com/backstage/backstage/blob/8fb1f2ae9812084c40495d2941aa5cbf4444210b/plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts#L454

I thought that `remoteUrl` should be for example `https://github.com/backstage/backstage`, but happy to be proven wrong on that, and if it is, then it should definitely be renamed anyways in a separate todo / breaking change.

If it is just the `repoUrl` then that's pretty easy to build always, even if a little bit icky for the most part.

koalaty-code: `remoteUrl` is described as "Link to the pull request in Github": https://github.com/backstage/backstage/blob/8fb1f2ae9812084c40495d2941aa5cbf4444210b/plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts#L273-L277

I've confirmed it is indeed a link to the pull request.

benjdlambert: I think we should update the json schema to be `oneOf` then in this case, that if you provide `createWhenEmpty` as false, then the `remoteUrl` is no longer required.

tjuedema: I've updated the output schema to utilize oneOf to allow remoteUrl to be an optional output. In the process of doing this, I notice that the Installed Actions page was not rendering oneOf output schemas. I've included a change in this PR to the scaffolder plugin  that will allow for these schemas to be displayed:
Screenshot 2025-02-04 at 2 47 01 PM

benjdlambert: I think that this is probably fine to treat as a separate issue and we fix in a followup PR :pray:

.changeset/thick-moose-do.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend-module-github': patch
+---
+
+Added oneOf output schema to publish:github:pull-request scaffolder action to allow for non-breaking addition of createWhenEmpty option
```
freben, line 5:
```suggestion
Added `oneOf` output schema to `publish:github:pull-request` scaffolder action to allow for non-breaking addition of `createWhenEmpty` option
```

.changeset/wise-cycles-sparkle.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-scaffolder-backend': patch
+'@backstage/plugin-scaffolder': patch
+---
+
+Added conditional rendering of oneOf output schemas on the Installed Actions page for scaffolder actions
```
freben, line 6:
```suggestion
Added conditional rendering of `oneOf` output schemas on the Installed Actions page for scaffolder actions
```

freben (approved):
OK by me, even though the output oneOf with its duplication is a bit sad. @benjdlambert ?

plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts
benjdlambert: Ended up back tracking on this a little bit, as it's still a breaking change either way as you can't switch the output based on a given input, which is an unfortunate restriction of the framework, maybe something to think about for the future.

So i just updated this PR to be breaking for now, as in practice it not breaking, only when you provide the new option.

I kept the changes for rendering though as they're still useful, but just not to this change in the PR so much anymore :pray:

benjdlambert (approved):

PR #28634, opened 2025-01-27
Canon - Merge Stack + Inline into Flex

cdedreuille:
We are merging the `Stack` component with the `Inline` component. The idea is to simplify the API, getting closer to what engineers are used to. This will also allow for the `direction` prop to be responsive and change the direction if needed.


Rugvip (approved):
Yep! 👍

.changeset/early-peaches-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/canon': patch
```
Rugvip, line 2: Big breaking change and needs to be marked as such

cdedreuille: I changed it to a minor and added the Breaking prefix following the other PR. Since the code was actually not accessible by anyone, nobody will have to migrate to the new component.

PR #28642, opened 2025-01-28
catalog-backend: allow processors to include location key when emitting entities

Rugvip:

This adds a new option that lets you explicitly set the location key of entities emitted during processing.

Bonus change that was required was to include the parents for all related entities in the processing cache key, or the entities would not get updated after being overridden. That's what the second test is about.

### DIFF
--- a/.changeset/hip-hotels-type.md
+++ b/.changeset/hip-hotels-type.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Fixed an bug in the entity processing caching that would prevent entities that were emitted during processing to be restored after being overridden.
--- a/plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
+++ b/plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
@@ -215,7 +215,12 @@ export class DefaultCatalogProcessingEngine {
               const { entityRefs: parents } =
                 await this.processingDatabase.transaction(tx =>
                   this.processingDatabase.listParents(tx, {
-                    entityRef,
+                    entityRefs: [
+                      entityRef,
+                      ...result.deferredEntities.map(e =>
+                        stringifyEntityRef(e.entity),
+                      ),
+                    ],
                   }),
                 );
 
--- a/plugins/catalog-backend/src/processing/ProcessorOutputCollector.ts
+++ b/plugins/catalog-backend/src/processing/ProcessorOutputCollector.ts
@@ -137,7 +137,9 @@ export class ProcessorOutputCollector {
         },
       };
 
-      this.deferredEntities.push({ entity, locationKey: location });
+      const locationKey =
+        i.locationKey === undefined ? location : i.locationKey ?? undefined;
+      this.deferredEntities.push({ entity, locationKey });
     } else if (i.type === 'location') {
       const entity = locationSpecToLocationEntity({
         location: i.location,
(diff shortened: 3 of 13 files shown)


plugins/catalog-backend/src/processing/ProcessorOutputCollector.ts
```diff
@@ -137,7 +137,9 @@ export class ProcessorOutputCollector {
         },
       };
 
-      this.deferredEntities.push({ entity, locationKey: location });
+      const locationKey =
+        i.locationKey === undefined ? location : i.locationKey ?? undefined;
```
freben, line 141: What does the `?? undefined` at the end do? Does it risk negating any intentional nulls?

.changeset/hip-hotels-type.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Fixed an bug in the entity processing caching that would prevent entities that were emitted during processing to be restored after being overridden.
```
freben, line 5: Am I right in understanding that this will re-stitch too? At least to RUN the stitcher and then bail out with `unchanged`? But then again, probably mostly for Location entities themselves since most kinds don't emit children 🤔 So maybe it's benign enough of an event that we don't have to call it out here.

plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
```diff
@@ -215,7 +215,12 @@ export class DefaultCatalogProcessingEngine {
               const { entityRefs: parents } =
                 await this.processingDatabase.transaction(tx =>
                   this.processingDatabase.listParents(tx, {
-                    entityRef,
+                    entityRefs: [
+                      entityRef,
+                      ...result.deferredEntities.map(e =>
+                        stringifyEntityRef(e.entity),
```
freben, line 221: This makes me realize that the stableStringify function still is sensitive to array order; i wonder if in particular discovery locations are at risk of emitting entities in a stochastic order. If so, it might make sense to order this array and `result.deferredEntities` as well by ref 🤔 Being basically explicit in code about "the order of these do not constitute an effective change" so to speak. Could be done separately perhaps especially if we want to order `result.deferredEntities` too.

freben (commented):
Needs changesets

plugins/catalog-backend/src/processing/ProcessorOutputCollector.ts
Rugvip: after this point `null` is represented as `undefined`, so it's just converting :sweat_smile:

.changeset/hip-hotels-type.md
Rugvip: yep expecting this to trigger stitch but hit the cache

plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
Rugvip: Thinking separate change for that if we want to do it?

Rugvip:
> Needs changesets

it has changesets

freben:
sorry, i meant api reports :D

plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
freben: Mm

PR #28653, opened 2025-01-29
[cli] Improve parallelism limits in build-workspace yarn pack

iamEAP:
## What / Why

Seeing occasional errors like this while running `backstage-cli build-workspace --alwaysYarnPack` in a decently sized monorepo in CI.

```
Repacking @foo/backstage-plugin-bar into dist workspace
➤ YN00ZZ: Calling the "prepack" lifecycle script
undefined:135
    "@types/jest

SyntaxError: Unterminated string in JSON at position 4707 (line 135 column 17) (when parsing /home/runner/work/org/repo/plugins/bar/package.json)
    at JSON.parse (<anonymous>)
    at t.loadFile (/home/runner/work/org/repo/.yarn/releases/yarn-4.6.0.cjs:140:122594)
    at async t.fromFile (/home/runner/work/org/repo/.yarn/releases/yarn-4.6.0.cjs:140:122278)
    at async t.tryFind (/home/runner/work/org/repo/.yarn/releases/yarn-4.6.0.cjs:140:122001)
    at async iE.setup (/home/runner/work/org/repo/.yarn/releases/yarn-4.6.0.cjs:205:9258)
```

Think it makes sense to swap out the hard-coded parallel work limiter in favor of existing utils in the CLI.

### DIFF
--- a/.changeset/cli-melancholy-in-advance.md
+++ b/.changeset/cli-melancholy-in-advance.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+The packing process when running `build-workspace` with the `--alwaysYarnPack` flag now respects the `BACKSTAGE_CLI_BUILD_PARALLEL` environment variable, defaulting parallel work limits based on CPU availability.
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -130,7 +130,6 @@
     "node-libs-browser": "^2.2.1",
     "npm-packlist": "^5.0.0",
     "ora": "^5.3.0",
-    "p-limit": "^3.1.0",
     "p-queue": "^6.6.2",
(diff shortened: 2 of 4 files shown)


Rugvip (approved):
Makes sense either way regardless of errors 👍 

I'm not sure this is gonna fix the errors you're seeing though, as I'd expect that to be a race when you have two or more `pack`s running at once. It's probably the case that the only way to avoid the error is to not run `pack` in parallel at all.

iamEAP:
> _It's probably the case that the only way to avoid the error is to not run pack in parallel at all._

Could be. Sure hope not, though! 😅 

In any event, at least we can control this now via `BACKSTAGE_CLI_BUILD_PARALLEL`.

PR #28677, opened 2025-01-30
permissions: add resources refs and rule accessor

Rugvip:

This is on top of #28400, which hasn't been released yet. It fills in the gap identified by @CptnFizzbin in https://github.com/backstage/backstage/pull/28494#issuecomment-2607634349, and adds two new concepts:

- `PermissionResourceRef`: a new ref type that encapsulates the types and constants associated with a permissioned resource that supports conditional decisions.
- `PermissionRuleAccessor`: a callback that can be used to fetch permission rules by name, accessed via the new `permissionsRegistry.getRulesAccessor(...)`.

A bunch of methods have been updated to accept these new types in the permission system, including `createConditionAuthorizer` and `createConditionTransformer`.

No breaking changes here other than compared to #28400. In most places I used overloads to support the old signatures, and deprecated the old variant.

### DIFF
--- a/packages/backend-plugin-api/src/services/definitions/PermissionsRegistryService.ts
+++ b/packages/backend-plugin-api/src/services/definitions/PermissionsRegistryService.ts
@@ -142,7 +142,7 @@ export interface PermissionsRegistryService {
   ): void;
 
   /**
-   * Returns a lookup function that can be used to look up rules for the provided resource by name.
+   * Returns the set of registered rules for this resource.
    *
    * @remarks
    *
--- a/packages/backend-plugin-api/src/services/definitions/PermissionsRegistryService.ts
+++ b/packages/backend-plugin-api/src/services/definitions/PermissionsRegistryService.ts
@@ -18,7 +18,7 @@ import { Permission } from '@backstage/plugin-permission-common';
 import {
   PermissionResourceRef,
   PermissionRule,
-  PermissionRuleAccessor,
+  PermissionRuleset,
 } from '@backstage/plugin-permission-node';
 
 /**
@@ -148,7 +148,7 @@ export interface PermissionsRegistryService {
    *
    * Primarily intended for use with {@link @backstage/plugin-permission-node#createConditionAuthorizer} and {@link @backstage/plugin-permission-node#createConditionTransformer}.
    */
-  getRuleAccessor<TResourceType extends string, TResource, TQuery>(
+  getPermissionRuleset<TResourceType extends string, TResource, TQuery>(
     resourceRef: PermissionResourceRef<TResource, TQuery, TResourceType>,
-  ): PermissionRuleAccessor<TResource, TQuery, TResourceType>;
+  ): PermissionRuleset<TResource, TQuery, TResourceType>;
 }
--- a/packages/backend-plugin-api/src/services/definitions/PermissionsRegistryService.ts
+++ b/packages/backend-plugin-api/src/services/definitions/PermissionsRegistryService.ts
@@ -15,7 +15,11 @@
  */
 
 import { Permission } from '@backstage/plugin-permission-common';
-import { PermissionRule } from '@backstage/plugin-permission-node';
+import {
+  PermissionResourceRef,
+  PermissionRule,
+  PermissionRuleAccessor,
+} from '@backstage/plugin-permission-node';
 
 /**
  * Prevent use of type parameter from contributing to type inference.
@@ -33,11 +37,12 @@ type NoInfer<T> = T extends infer S ? S : never;
 export type PermissionsRegistryServiceAddResourceTypeOptions<
   TResourceType extends string,
   TResource,
+  TQuery,
 > = {
   /**
-   * The identifier for the resource type.
+   * The {@link @backstage/plugin-permission-node#PermissionResourceRef} that identifies the resource type.
    */
-  resourceType: TResourceType;
+  resourceRef: PermissionResourceRef<TResource, TQuery, TResourceType>;
 
   /**
    * Permissions that are available for this resource type.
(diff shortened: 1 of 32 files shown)


plugins/permission-node/report.api.md
```diff
@@ -268,6 +350,13 @@ export type PermissionRule<
   toQuery(params: NoInfer_2<TParams>): PermissionCriteria<TQuery>;
 };
 
+// @public
+export type PermissionRuleAccessor<
```
freben, line 354: Naming suggestion: `PermissionRuleset` with `getByName`

Rugvip: Yep good call on wrapping this up a bit, will refactor. Naming-wise I like ruleset more, I'll switch to that for now.

Rugvip: Updated. I went with `getRuleByName` to be a bit more explicit and to have the type itself be a bit more strict since I don't think it's worth making it an opaque type at this point.

vinzscam (changes requested):
sorry at this point I have trouble understanding the advantages of having the new `PermissionRegistryService`. I thought the goal was to automatically include the `createPermissionIntegrationRoute` so that plugin author's won't forget it. The error ratio of committing such a mistake was limited to plugins exposing basic permissions. If a plugin exposes resource permission the plugin author can't forget it, as it is required for resolving conditional decisions.
Now, plugins authors would need to:
- manually call `permissionsRegistry.addPermissions()`
- manually invoke `permissionsRegistry.addResourceType` (Which is the same as using `createPermissionIntegrationRoute`)
- define an extra `PermissionResourceRef` for each resource

I feel like we are adding more complexity on top of a system which was already complex

Rugvip:
@vinzscam no the main point of the new registry service was to avoid the need for each backend plugin to have to build its own extension points for modules to add additional permissions, rules, and resource types.

The purpose of `PermissionResourceRef` is to have a single way to encapsulate types and constants associated with a resource type in the backend, rather than having it split across things like `createCatalogPermissionRule` with `CatalogPermissionRule`, `RESOURCE_TYPE_CATALOG_ENTITY`, `Entity` and `EntitiesSearchFilter`. More generally it makes it fairly straight-forward to add more utilities related to permission resources in the permissions system without requiring all plugins to add new exports.

Rugvip:
@vinzscam alright updated as discussed

vinzscam (approved):
👍

packages/backend-plugin-api/src/services/definitions/PermissionsRegistryService.ts
```diff
@@ -59,7 +68,9 @@ export type PermissionsRegistryServiceAddResourceTypeOptions<
    * resolve conditional decisions except when requesting resources directly
    * from the plugin.
    */
-  getResources?(resourceRefs: string[]): Promise<Array<TResource | undefined>>;
+  getResources?(
+    resourceRefs: string[],
```
freben, line 72: This one is still meant to take just strings, not resource ref objects? Sorry maybe it's obvious but I was not sure about this part of the system

packages/backend-plugin-api/src/services/definitions/PermissionsRegistryService.ts
```diff
@@ -122,10 +133,22 @@ export interface PermissionsRegistryService {
...
     options: PermissionsRegistryServiceAddResourceTypeOptions<
       TResourceType,
-      TResource
+      TResource,
+      TQuery
     >,
   ): void;
+
+  /**
+   * Returns a lookup function that can be used to look up rules for the provided resource by name.
```
freben, line 145:
```suggestion
   * Returns the set of registered rules for this resource.
```

plugins/permission-node/src/types.ts
```diff
@@ -66,3 +66,22 @@ export type PermissionRule<
...
 };
+
+/**
+ * A function that can be used to look up permission rules by name for a particular resource type.
+ *
+ * @remarks
+ *
+ * Accessed via {@link @backstage/backend-plugin-api#PermissionsRegistryService.getPermissionRuleset}.
+ *
+ * Will throw an error if a rule with the provided name does not exist.
```
freben, line 77: maybe put this on the method instead

plugins/permission-node/src/integration/createPermissionResourceRef.ts
```diff
@@ -0,0 +1,56 @@
...
+    with<TPluginId extends string, TResourceType extends string>(options: {
+      pluginId: TPluginId;
+      resourceType: TResourceType;
+    }): PermissionResourceRef<TResource, TQuery, TResourceType, TPluginId> {
+      return {
+        $$type: '@backstage/PermissionResourceRef',
+        pluginId: options.pluginId,
+        resourceType: options.resourceType,
+        TQuery: null as TQuery,
+        TResource: null as TResource,
```
freben, line 52: some day maybe put getters that throw an exception on these

freben (approved):
Some thoughts but preapproving

packages/backend-plugin-api/src/services/definitions/PermissionsRegistryService.ts
Rugvip: Yep this is essentially a data loader. These refs aren't refs but something like entity string refs.

plugins/permission-node/src/integration/createPermissionResourceRef.ts
Rugvip: kinda stopped doing that since it messes with test assertion, `null` is the new throw

packages/backend-plugin-api/src/services/definitions/PermissionsRegistryService.ts
freben: oh that's not confusing at all :D

freben (approved):

PR #28680, opened 2025-01-30
docs: add new JSX Transform guide

schultzp2020:

- Adds new JSX Transform guide
- Updates the React 18 guide to include React 16 deprecation.


docs/tutorials/react18-migration.md
```diff
@@ -1,75 +1,127 @@
...
+Be sure to update your lockfile after modifying your `package.json` files.
+
+:::
 
 ### TypeScript Errors
 
 When upgrading to React 18 you are likely to see a fair number of TypeScript type errors. A summary of the breaking changes can be found in the [Pull Request that introduced them](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210). A codemod is also provided to help with the migration.
 
-Run `yarn tsc:full` to asses the damage.
+Run `yarn tsc:full` to access the damage.
```
Parsifal-M, line 124: I think you do actually mean `assess` right? like to make an assessment of the damage?

docs/tutorials/react18-migration.md
```diff
@@ -1,75 +1,127 @@
...
-// highlight-add-next-line
-ReactDOM.createRoot(document.getElementById('root')!).render(<App />);
-```
-
-Once these steps are done you should be able to run your app and see it working as before, except now using React 18.
+### Upgrading to React 18
+
+#### Backstage Instance
+
+To migrate a Backstage instance to React 18, follow these three steps:
```
Parsifal-M, line 23: I think we can just say follow these steps 😅, since reading it it is technically 3 + 1 at the end

docs/tutorials/jsx-transform-migration.md
```diff
@@ -0,0 +1,477 @@
...
+description: A guide to migrating your project to the New JSX Transform
+---
+
+The Backstage core libraries are in the process of deprecating React 16 and evaluating the adoption of React 19. In React 19, the [New JSX Transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), introduced in September 2020, is now [mandatory](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#installing). This requires a modification in how React is imported into components to ensure compatibility with React 19.
+
+## Migration Process
+
+### Using the Codemod
+
+The codemod referenced in the [Introducing the New JSX Transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) article is no longer functional. However, we have identified a working solution detailed below.
```
Parsifal-M, line 13: Should we mention this if it doesn't work anyway? 💭

docs/tutorials/jsx-transform-migration.md
```diff
@@ -0,0 +1,477 @@
+---
+id: jsx-transform-migration
+title: Transitioning to the New JSX Transform
+description: A guide to migrating your project to the New JSX Transform
+---
+
+The Backstage core libraries are in the process of deprecating React 16 and evaluating the adoption of React 19. In React 19, the [New JSX Transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), introduced in September 2020, is now [mandatory](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#installing). This requires a modification in how React is imported into components to ensure compatibility with React 19.
```
Parsifal-M, line 7: Just wondering as it might confuse some people (like me 😆) we should be safe to be on React 18, but before we move to 19 we should do this? When we say that we are evaluating the adoption of React 19 does that potentially mean I shouldn't jump the gun and do these steps?

Just trying to wrap my head around should I be doing this now? or should I wait for an all green 🤔

Parsifal-M (commented):
Hey! 👋 

Awesome stuff!

Just a few thoughts while reading!

docs/tutorials/jsx-transform-migration.md
Parsifal-M: You're right I think its fine as long as it mentions to not use it because its not working 😅

szubster:
This, unfortunately, does not work by default for adopters.
tsconfig and eslint config included in backstage-cli both complain (in different ways) when removing React import.

szubster:
I think at least the eslint rule should be changed in backstage cli.
Otherwise, I need to add it in multiple packages and add instructions to people creating new plugins in our repo, to change the eslint.
It's not a good experience. The rule can be removed in backstage repo itself.
Overriding it in root eslintrc does not work, because of the way inheritance works in eslint.

Pike:
I see that this is a hard problem while we're transitioning.

One idea I just had was if we can pull all of this into a subcommand of `backstage-cli migrate` to make it easier to reapply when creating new plugins from old templates?

That would also be easier to explain in inner-source communities.

I also spotted that a healthy amount of our eslint configs are still in the deprecated `@spotify` repo.

docs/tutorials/jsx-transform-migration.md
```diff
@@ -0,0 +1,571 @@
...
+
+   ```tsx
+   import { default as React } from 'react';
+   ```
+
+### Updating Configuration Files
+
+To ensure compatibility when using the `backstage-cli`, you must modify both the `tsconfig.json` and `eslintrc.js` files. Failure to do so may result in errors.
+
+#### TypeScript Configuration (TSConfig)
```
freben, line 491: You may want to add the word "tsconfig" to `.github/vale/config/vocabularies/Backstage/accept.txt` or put this in backticks

docs/tutorials/jsx-transform-migration.md
```diff
@@ -0,0 +1,571 @@
...
+
+While a codemod for the New JSX Transform was originally introduced in the [Introducing the New JSX Transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) article, it is no longer functional. A working solution, inspired by the original, is detailed below:
+
+1. **Create the transform file**
+
+   Create a file named `transform.js` in the root directory of your Backstage project.
+
+   ```js
+   /**
+    * (c) Facebook, Inc. and its affiliates. Confidential and proprietary.
```
freben, line 34: This ... makes me wonder if we really can have a copy of it here :) If we can really (from a license perspective) have it, it might be better to put it in contrib, and link to that here. The document just becomes so hard to read with all this code

freben (commented):
Alright, some notes. Gonna pull in @Rugvip for the timing and process

aramissennyeydd (commented):
Nice!

freben:
Just noting here that it seems we'd be moving that file into contrib, if anything

docs/tutorials/jsx-transform-migration.md
```diff
@@ -0,0 +1,124 @@
...
+        message:
+          'Default React import not allowed. If you need a global type that conflicts with a React named export (e.g., `MouseEvent`), use `globalThis.MouseHandler`.',
+        selector:
+          "ImportDeclaration[source.value='react'] :matches(ImportDefaultSpecifier, ImportNamespaceSpecifier)",
+      },
+    ],
+    // highlight-add-end
+  },
+});
+```
```
Rugvip, line 119: Hmm, I'm thinking it might be better to just leave this to be a warning for now and not instruct adopters to add these rules? It's fail anyway due to the updated TS config

Rugvip: 👍, could you update the PR asap? 🙏, looking to ship the release soon

Rugvip (approved):
Thanks!

PR #28695, opened 2025-01-31
Fix: Default feature type information on exports

HHogg:
Fixes an issue where default feature type information wasn't being added to the package.json/exports before publishing for packages where exports didn't exist before hand, by also adding this information to the backstage metadata.

### DIFF
--- a/packages/cli-node/report.api.md
+++ b/packages/cli-node/report.api.md
@@ -12,6 +12,9 @@ export type BackstagePackage = {
   packageJson: BackstagePackageJson;
 };
 
+// @public (undocumented)
+export type BackstagePackageFeatureType = (typeof packageFeatureType)[number];
+
 // @public
 export interface BackstagePackageJson {
   // (undocumented)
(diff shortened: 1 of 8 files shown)


packages/cli-node/report.api.md
```diff
@@ -112,6 +116,14 @@ export type LockfileDiffEntry = {
   range: string;
 };
 
+// @public
+export const packageFeatureType: readonly [
+  '@backstage/BackendFeature',
+  '@backstage/BackstagePlugin',
```
Rugvip, line 122: do we really need those one? it's for the old frontend system right?

Rugvip (approved):
alright let's :shipit:, just a small bonus Q

packages/cli-node/report.api.md
HHogg: I'm not sure what needs to go here, it's what ever the types are for all of the frontend/backend plugins and modules. We need to keep in mind to capture any old stuff from where packages may not have been migrated yet?

Issue #26973, opened 2024-10-04
🚀 Feature: Ability to Group various tabs together in the Entity view in the Catalog


awanlin:
### 🔖 Feature description

It would be helpful to have a way to group various tabs together in the Entity view in the Catalog

### 🎤 Context

As your Adoption of Backstage grows so does the number of plugins and tabs that show up on an Entity in the Catalog. Currently there is a horizontal scroll when this happens which isn't the best UX as it leads to potentially helpful information being hidden.

### ✌️ Possible Implementation

One way this could be done is by using a dropdown for the grouping like this:

![Screenshot 2024-10-03 at 7 21 17 AM](https://github.com/user-attachments/assets/fdcd32ba-e525-4ba2-b14b-621dd4537b11)

And then in action:

![Screenshot 2024-10-03 at 7 21 07 AM](https://github.com/user-attachments/assets/c61fa3cf-a270-4ba5-8178-e7bb5fbca4bc)

This could be done by extending the `EntityLayout` with an `EntityLayout.Group`.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

awanlin:
I'll note this is something we have built internally at Spotify and would like to contribute but would like to make it clear there is no timeline on when that might happen. Wanted to log this for others to know about and help us get an idea of how much it would be used by the larger Backstage Community. 👍

stephenglass:
> I'll note this is something we have built internally at Spotify and would like to contribute but would like to make it clear there is no timeline on when that might happen. Wanted to log this for others to know about and help us get an idea of how much it would be used by the larger Backstage Community. 👍

This is definitely a feature that my org would be interested in.

nia-potato:
same here

Guazi:
Willing and able to help here but not sure how to reproduce your use case with the standard dev environment since these are the only tabs I get:

![image](https://github.com/user-attachments/assets/8dbac0f0-3948-4114-bbce-045baa897f1c)


From what I can tell, `export const EntityLayout` would need to be modified in `EntityLayout.tsx` by expanding:

```
return [
            {
              path: elementProps.path,
              title: elementProps.title,
              children: elementProps.children,
[...]

PR #28701, opened 2025-02-01
[NFS] Entity Page Content Groups

camilaibs:

Closes: https://github.com/backstage/backstage/issues/26973

There is now the option to group multiple entity contents in the same tab on the new frontend system entity page.
> [!Important]
> - Tabs with more than one entity content will be rendered as a dropdown.
> - Entity contents determine the order in which the groups are rendered, so a group will be rendered as soon as its first entity content is detected.

image

### Usage examples

Disabling all default groups:

```diff
# app-config.yaml
app:
  extensions:
    # Pages
+   - page:catalog/entity:
+       config:
+         groups: []
```

Creating a new group via configuration:
```diff
# app-config.yaml
app:
  extensions:
    # Pages
+   - page:catalog/entity:
+       config:
+         groups:
+           - custom:
+               title: Custom
```

Overriding a default group title via configuration:
```diff
# app-config.yaml
app:
  extensions:
    # Pages
+   - page:catalog/entity:
+       config:
+         groups:
+           - documentation:
+               # this is overriding the default "documentation" group title to "Docs"
+               title: Docs
```

Setting an optional default group to an entity content when creating the extension:
```diff
const entityKubernetesContent = EntityContentBlueprint.make({
  name: 'kubernetes',
  params: {
    defaultPath: '/kubernetes',
    defaultTitle: 'Kubernetes',
+   defaultGroup: 'deployment',
    filter: 'kind:component,resource',
    loader: () =>
      import('./KubernetesContentPage').then(m =>
        compatWrapper(<m.KubernetesContentPage />),
      ),
  },
});
```

Disassociating an entity content from a default group:

```diff
# app-config.yaml
app:
  extensions:
    # Entity page content
-   - entity-content:kubernetes/kubernetes
+   - entity-content:kubernetes/kubernetes:
+       config:
+         group: false
```

Associating an entity content with a different default or custom group than the one defined via code when the extension was created:
```diff
# app-config.yaml
app:
  extensions:
    # Entity page content
    - entity-content:catalog/overview
    - entity-content:api-docs/definition
-   - entity-content:api-docs/apis
+   - entity-content:api-docs/apis:
+       config:
+         group: documentation # associating this extension with a default group id
-   - entity-content:techdocs
+   - entity-content:techdocs:
+       config:
+         group: documentation
-   - entity-content:kubernetes/kubernetes
+   - entity-content:kubernetes/kubernetes:
+       config:
+         group: custom # associating this extension with a custom group id, the group should have previously been created via entity page configuration
```

### DIFF
--- a/.changeset/thin-steaks-trade.md
+++ b/.changeset/thin-steaks-trade.md
@@ -1,0 +1,45 @@
+---
+'@backstage/plugin-catalog': patch
+---
+
+Update the default entity page extension component to support grouping multiple entity content items in the same tab.
+
+Disable a default group via configuration:
+
+```diff
+# app-config.yaml
+app:
+  extensions:
+    # Pages
++   - page:catalog/entity:
++       config:
++         groups:
++           deployment: false
+```
+
+Change a default group title via configuration:
+
+```diff
+# app-config.yaml
+app:
+  extensions:
+    # Pages
++   - page:catalog/entity:
++       config:
++         groups:
++           deployment: Infrastructure # this is overriding the default group title
+```
+
+Create a custom entity content group via configuration:
+
+```diff
+# app-config.yaml
+app:
+  extensions:
+    # Pages
++   - page:catalog/entity:
++       config:
++         groups:
++           # <id>: <Title>
++           custom: Custom
+```
--- a/.changeset/thin-steaks-trade.md
+++ b/.changeset/thin-steaks-trade.md
@@ -4,7 +4,7 @@
 
 Update the default entity page extension component to support grouping multiple entity content items in the same tab.
 
-Disable a default group via configuration:
+Disable all default groups:
 
 ```diff
 # app-config.yaml
@@ -13,11 +13,10 @@ app:
     # Pages
 +   - page:catalog/entity:
 +       config:
-+         groups:
-+           deployment: false
++         groups: []
 ```
 
-Change a default group title via configuration:
+Create a custom list of :
 
 ```diff
 # app-config.yaml
@@ -27,19 +26,7 @@ app:
 +   - page:catalog/entity:
 +       config:
 +         groups:
-+           deployment: Infrastructure # this is overriding the default group title
-```
-
-Create a custom entity content group via configuration:
-
-```diff
-# app-config.yaml
-app:
-  extensions:
-    # Pages
-+   - page:catalog/entity:
-+       config:
-+         groups:
-+           # <id>: <Title>
-+           custom: Custom
++           # This array of groups completely replaces the default groups
++           - custom:
++               title: 'Custom'
 ```
--- a/packages/app-next/app-config.yaml
+++ b/packages/app-next/app-config.yaml
@@ -12,6 +12,18 @@ app:
     # - apis.plugin.graphiql.browse.gitlab: true
     # - graphiql-endpoint:graphiql/gitlab: true
 
+    # Pages
+    - page:catalog/entity:
+        config:
+          groups:
+            # example disabling a default group
+            development: false
+            # example overriding a default group title
+            documentation: Docs
+            # example adding a new group
+            # Format <GROUP_ID>: <GROUP_TITLE>
+            custom: Custom
+
     # Entity page cards
     - entity-card:catalog/about
     - entity-card:catalog/labels
@@ -27,23 +39,33 @@ app:
     - entity-card:api-docs/provided-apis
     - entity-card:api-docs/providing-components
     - entity-card:api-docs/consuming-components
-
-    # - entity-card:azure-devops/readme
-    # Entity page content
-    - entity-content:api-docs/definition
-    - entity-content:api-docs/apis
-    - entity-content:techdocs
-    # - entity-content:azure-devops/pipelines
-    # - entity-content:azure-devops/pull-requests
-    # - entity-content:azure-devops/git-tags
-
     # Org Plugin
     - entity-card:org/group-profile
     - entity-card:org/members-list
     - entity-card:org/ownership
     - entity-card:org/user-profile
+    # - entity-card:azure-devops/readme
 
-    - entity-content:kubernetes/kubernetes
+    # Entity page contents
+    - entity-content:catalog/overview:
+        config:
+          # associating with a custom group
+          group: custom
+    - entity-content:api-docs/definition
+    - entity-content:api-docs/apis:
+        config:
+          # example associating with a default group
+          group: documentation
+    - entity-content:techdocs:
+        config:
+          group: documentation
+    - entity-content:kubernetes/kubernetes:
+        config:
+          # example disassociating with a default group
+          group: false
+    # - entity-content:azure-devops/pipelines
+    # - entity-content:azure-devops/pull-requests
+    # - entity-content:azure-devops/git-tags
 
   # scmAuthExtension: >-
   #   createScmAuthExtension({
--- a/packages/app-next/app-config.yaml
+++ b/packages/app-next/app-config.yaml
@@ -17,12 +17,13 @@ app:
         config:
           groups:
             # example disabling a default group
-            development: false
+            - development: false
             # example overriding a default group title
-            documentation: Docs
+            - documentation:
+                title: Docs
             # example adding a new group
-            # Format <GROUP_ID>: <GROUP_TITLE>
-            custom: Custom
+            - custom:
+                title: Custom
 
     # Entity page cards
     - entity-card:catalog/about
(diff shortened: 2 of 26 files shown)


packages/core-components/package.json
```diff
@@ -69,6 +69,7 @@
     "@types/react-sparklines": "^1.7.0",
     "ansi-regex": "^6.0.1",
     "classnames": "^2.2.6",
+    "clsx": "^2.1.1",
```
Rugvip, line 72: We can stick to using `classnames` rather than adding this right?

Rugvip (commented):
Nice! 👍 

As discussed let's refactor to avoid changes to the old frontend system

plugins/catalog/src/components/EntityLayout/EntityLayout.tsx
```diff
@@ -65,17 +65,28 @@ import { EntityContextMenu } from '../EntityContextMenu/EntityContextMenu';
...
+};
+
+const groupDataKey = 'plugin.catalog.entityLayoutGroup';
+const Group: (props: EntityLayoutGroupProps) => null = () => null;
+attachComponentData(Group, groupDataKey, true);
 
+const routeDataKey = 'plugin.catalog.entityLayoutRoute';
 const Route: (props: EntityLayoutRouteProps) => null = () => null;
-attachComponentData(Route, dataKey, true);
+attachComponentData(Route, routeDataKey, true);
```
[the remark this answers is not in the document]
camilaibs: Done

packages/core-components/package.json
camilaibs: Yes, removed.

packages/app-next/app-config.yaml
```diff
@@ -12,6 +12,13 @@ app:
     # - apis.plugin.graphiql.browse.gitlab: true
     # - graphiql-endpoint:graphiql/gitlab: true
 
+    # Pages
+    - page:catalog/entity:
+        config:
+          groups:
+            - id: documentation
```
camilaibs, line 19: Example creating or overriding a default group title

packages/app-next/app-config.yaml
```diff
@@ -27,12 +34,17 @@ app:
...
     - entity-card:api-docs/consuming-components
-
     # - entity-card:azure-devops/readme
+
     # Entity page content
+    - entity-content:catalog/overview
     - entity-content:api-docs/definition
-    - entity-content:api-docs/apis
-    - entity-content:techdocs
+    - entity-content:api-docs/apis:
```
camilaibs, line 42: Example setting a group to a tab via config

plugins/catalog-react/src/alpha/blueprints/extensionData.tsx
```diff
@@ -32,3 +32,19 @@ export const entityFilterExpressionDataRef =
   createExtensionDataRef<string>().with({
     id: 'catalog.entity-filter-expression',
   });
+
+/**
+ * @alpha
+ * Default entity content groups.
+ */
+export const defaultEntityContentGroups = {
```
camilaibs, line 40: Open to suggestions on what the default groups should be

plugins/catalog/src/alpha/pages.tsx
```diff
@@ -62,21 +65,57 @@ export const catalogEntityPage = PageBlueprint.makeWithOverrides({
...
+        // Config groups can override default group titles
+        const groups: Record<string, string> = {
+          ...defaultEntityContentGroups,
+          ...config.groups?.reduce(
+            (rest, group) => ({ ...rest, [group.id]: group.title }),
+            {},
+          ),
+        };
+
+        const tabs = inputs.contents.reduce<
```
camilaibs, line 93: Does it make sense to show groups in the order their entity contents appear?

.changeset/thin-steaks-trade.md
```diff
@@ -0,0 +1,45 @@
...
+
+```diff
+# app-config.yaml
+app:
+  extensions:
+    # Pages
++   - page:catalog/entity:
++       config:
++         groups:
++           deployment: false
```
Rugvip, line 17: Seems a bit tricky to have this being an object wrt ordering. Do you think it would be horrible to always keep this as an array that has to define all the available groups?

Or maybe this can be the group configuration while there's a separate `groupOrder` or smth that just list the IDs. Or perhaps we just leave this as is but also support array for completely overriding all groups, and with the object form it might not be possible to add new groups?

.changeset/thin-steaks-trade.md
```diff
@@ -0,0 +1,45 @@
...
+
+```diff
+# app-config.yaml
+app:
+  extensions:
+    # Pages
++   - page:catalog/entity:
++       config:
++         groups:
++           deployment: Infrastructure # this is overriding the default group title
```
Rugvip, line 30: Kinda feeling it's worth having a `title: ` key in there for readability tbh, so you'd need to have

```yaml
groups:
  deployment:
    title: Infrastructure
```

Happy to discuss though

.changeset/thin-steaks-trade.md
Rugvip: Replying to my initial comment myself here 😁 

I just realized that there is never any config merging going on here, so we can actually rely on the object key order for the ordering of the groups, at least somewhat. The main issue I guess would be for those that want to completely replace all default groups with their own, that'd be a bit cleaner to do with an array, which i turn takes us back to maybe supporting both array and object form

Rugvip (commented):
Ran out of time today so just posting the halfway review of the config structure from the docs ._.

.changeset/thin-steaks-trade.md
camilaibs: Done.

.changeset/thin-steaks-trade.md
camilaibs: I'm not sure I understand what you mean by array and object support. I've made some changes, please take a look if that's what you mean. One downside is that if people wanted to mix default and custom groups, they would have to recreate all the default groups via configuration (and I'm not sure I like that rewording):

What if we had two separate categories for the groups? Something like:

```yaml
# disable all default groups
groups:
  defaults: false
  # create custom groups:
  customs:
    - doc:
        title: 'Documentation'
    - env:
        title: 'Environment'
```

```yaml
# keeps default groups and rename the title of some of them
groups:
  defaults:
    - documentation:
        title: 'Docs'
    - infrastructure:
        title: 'Infra'
  # create custom groups:
  customs:
    - other:
        title: 'Other'

File: plugins/catalog-react/src/alpha/blueprints/EntityContentBlueprint.ts [hunk 9]
```diff
@@ -40,30 +42,35 @@ export const EntityContentBlueprint = createExtensionBlueprint({
...
       defaultPath,
       defaultTitle,
+      defaultGroup,
       filter,
       routeRef,
     }: {
       loader: () => Promise<JSX.Element>;
       defaultPath: string;
       defaultTitle: string;
+      defaultGroup?: keyof typeof defaultEntityContentGroups;
```
Rugvip, line 73: Noting for followup: I think we should allow `| string` here too, to allow for organic creation of new groups as well as custom groups internally

plugins/catalog-react/src/alpha/blueprints/EntityContentBlueprint.ts
```diff
@@ -91,5 +99,9 @@ export const EntityContentBlueprint = createExtensionBlueprint({
     } else if (typeof filter === 'function') {
       yield entityFilterFunctionDataRef(filter);
     }
+
+    if (group) {
+      yield entityContentGroupDataRef(group);
```
Rugvip, line 104: Note for followup: might be that we want to simply skip yielding if `group === false` rather than including `false` as a potential value of the data.

plugins/catalog-react/src/alpha/blueprints/extensionData.tsx
Rugvip: Yep I think this is a good starting point but might need another pass before this hits mainline

plugins/catalog/src/alpha/pages.tsx
Rugvip: Probably not tbh, but leaving this as a thing to potentially tweak

Rugvip (approved):
Awesome stuff 👏 

Made a few notes for followups and things to iterate on, but let's :shipit:

plugins/catalog-react/src/alpha/blueprints/EntityContentBlueprint.ts
Rugvip: Followup in https://github.com/backstage/backstage/pull/28916

plugins/catalog-react/src/alpha/blueprints/EntityContentBlueprint.ts
Rugvip: Followup in https://github.com/backstage/backstage/pull/28917

PR #28705, opened 2025-02-03
feat(scaffolder): experimental flag to wait for task completions

drodil:

Enabling the `EXPERIMENTAL_gracefulShutdown` flag in the scaffolder config will make the scaffolder block the shutdown process until all running tasks have completed. This is useful when there is a need to ensure that all tasks have completed before the scaffolder is shut down.

### DIFF
--- a/.changeset/tidy-oranges-allow.md
+++ b/.changeset/tidy-oranges-allow.md
@@ -1,0 +1,12 @@
+---
+'@backstage/plugin-scaffolder-backend': minor
+---
+
+Added experimental flag for scaffolder to wait for running tasks to complete on shutdown
+
+Enabling the `EXPERIMENTAL_gracefulShutdown` flag in the scaffolder config will make the
+scaffolder block the shutdown process until all running tasks have completed. This is useful
+when there is a need to ensure that all tasks have completed before the scaffolder is shut down.
+
+Please note, that the `TaskWorker` `stop` method is now asynchronous and awaited for the
+tasks to complete when the experimental flag is enabled.
--- a/plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
+++ b/plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
@@ -44,6 +44,7 @@ export type TaskWorkerOptions = {
   permissions?: PermissionEvaluator;
   logger?: Logger;
   auditor?: AuditorService;
+  gracefulShutdown?: boolean;
 };
 
 /**
@@ -74,6 +75,7 @@ export type CreateWorkerOptions = {
   concurrentTasksLimit?: number;
   additionalTemplateGlobals?: Record<string, TemplateGlobal>;
   permissions?: PermissionEvaluator;
+  gracefulShutdown?: boolean;
 };
 
 /**
@@ -108,6 +110,7 @@ export class TaskWorker {
       concurrentTasksLimit = 10, // from 1 to Infinity
       additionalTemplateGlobals,
       permissions,
+      gracefulShutdown,
     } = options;
 
     const workflowRunner = new NunjucksWorkflowRunner({
@@ -127,6 +130,7 @@ export class TaskWorker {
       concurrentTasksLimit,
       permissions,
       auditor,
+      gracefulShutdown,
     });
   }
 
@@ -156,8 +160,13 @@ export class TaskWorker {
     })();
   }
 
-  stop() {
+  async stop() {
     this.stopWorkers = true;
+    if (this.options?.gracefulShutdown) {
+      while (this.taskQueue.size > 0) {
+        await new Promise(resolve => setTimeout(resolve, 1000));
+      }
+    }
   }
 
   protected onReadyToClaimTask(): Promise<void> {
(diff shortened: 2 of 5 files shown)


plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
```diff
@@ -156,8 +156,15 @@ export class TaskWorker {
     })();
   }
 
-  stop() {
+  stop(options?: { waitForTaskCompletion?: boolean }) {
```
benjdlambert, line 159: I wonder if we should just have this as part of the `CreateWorkerOptions` and pass through to the constructor instead?

Can be optional in that instead?

benjdlambert (changes requested):
Hey :wave: thanks for this!

Wondering if we should just make this interface a little simpler and move the options to `create` instead?

plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
drodil: true, will change!

drodil:
Related discussion: https://discord.com/channels/687207715902193673/1334501396590104677

plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
```diff
@@ -158,6 +162,13 @@ export class TaskWorker {
 
   stop() {
     this.stopWorkers = true;
+    if (this.options?.gracefulShutdown) {
```
freben, line 165: Is `stop` externally visible public api? If not, should it not be blocking (made async), and the caller block on it too?

drodil: Yep, it's visible 😟

freben: @benjdlamben how would you feel about making it async in a minor release, saying "please await if you care about when tasks finish"? This really should not break anybody right 🤔

benjdlambert: yeah, I mean it's non-breaking to do it that way, it really should be connected to the `lifecycleService` anyways I feel.

drodil: @freben you mean there's no need for the config value but instead make it the default behaviour for the `stop` function?

freben: No I was only referring to making the method async and blocking (and the caller awaiting that), nothing else.

Spawning a little free floating piece of work won't have any effect in practise, I think.

drodil: I changed it now to async. Also mention in the changeset and changed it to minor version update.

benjdlambert (approved):
Nice work @drodil!

If this works pretty good in testing, I'd also be inclined to just have this as the default moving forward, feels like there shouldn't really be a case where you wouldn't want this?

.changeset/tidy-oranges-allow.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+Added experimental flag for scaffolder to wait for running tasks to complete on shutdown
+
+Enabling the `EXPERIMENTAL_gracefulShutdown` flag in the scaffolder config will make the
+scaffolder block the shutdown process until all running tasks have completed. This is useful
+when there is a need to ensure that all tasks have completed before the scaffolder is shut down.
+
+Please note, that the `TaskWorker` `stop` method is now asynchronous and awaited for the
+tasks to complete when the experimental flag is enabled.
```
acierto, line 12: Hey, did you test it on k8s redeploy scenario?

benjdlambert: This should work pretty nicely with the new liveness and readiness checks with the new backend system, it's what I was gonna test internally too.

https://github.com/backstage/backstage/pull/27961

[these docs](https://backstage.io/docs/backend-system/core-services/root-http-router/#via-app-configyaml) might also be super relevant to configure, as the backend will just force quit I think if that's not configured to a longer timeout.

acierto: That would be cool if it already works 👍, last time I played with that I was losing the connection to the database immediately once I hit "kubectl rollout restart deployment"

drodil:
> Nice work @drodil!
> 
> If this works pretty good in testing, I'd also be inclined to just have this as the default moving forward, feels like there shouldn't really be a case where you wouldn't want this?

Thanks @benjdlambert - I think it would make sense to make this the default in the future but would like to test this in our environment with the upcoming Backstage release. Would it be a good idea to leave this behind the experimental config for now and perhaps remove it for 1.37.0?

benjdlambert:
@drodil yep fine to ship and test and remove later :)

.changeset/tidy-oranges-allow.md
acierto: Oh I see, that's a culprit, as we are actually using a proxy sidecar to DB

benjdlambert: Yeah you can configure those sidecars to wait I believe, or at least have a `preStop` command of `sleep` is a bit of hack to make sure it's the last thing to be shut down

benjdlambert: What's the sidecar you're using? We could just add support for the connector in knex like we have for cloud-sql-proxy too.

But ye, documenting this is also a good shout

drodil:
@freben @benjdlambert just came to my mind, would it make sense to have a similar kind of graceful exit for the scheduled tasks? I think they will be aborted immediately when the lifecycle shutdown is sent.

benjdlambert:
Hmm, immediate thought is most probably yes. Can't think of a reason why? :pray:

Issue #27884, opened 2024-11-27
🚀 Feature: Add autocompletion from github to the repourlpicker


GersonTf:
### 🔖 Feature description

Similar to https://github.com/backstage/backstage/pull/27367#issuecomment-2480083403 for gitlab
and https://github.com/backstage/backstage/pull/25132#issuecomment-2258270240 for bitbucket

It would be nice to also support github autocompletion of repos for the scaffolder.

I am creating the issue to notify that I want to work on this but I need to do more research before adding a better description, I will update this description in the future.

CC @benjdlambert I am tagging you since I see you involved in both the bitlab and bitbucket issue, this is my first time contributing so any support will be appreciated =) 

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

benjdlambert:
@GersonTf I think the reason the bitbucket common module exists is because of the shared code between bitbucket server and bitbucket cloud. You don't have to create another module for this, I think it's fine to add to the existing github module for the scaffolder backend :tada:

GersonTf:
Hi @benjdlambert @Rugvip 

I am already working on this and I am close to an implementation but I have a question. I see that the gitlab implementation is leveraging API calls to this [library](https://github.com/jdalrymple/gitbeaker)` import { Gitlab, GroupSchema } from '@gitbeaker/rest';`

After research, I found a similar library for [github](https://github.com/octokit/rest.js) called oktokit/rest. 
Would it be OK for me to use it on backstage?

benjdlambert:
I think it's most likely that theres duplicates of the `octokit` library. I would look at the current version that's being used in the scaffolder github module. I don't think you'll need to install any more libraries, as the existing ones in the module should be enough.

I'd also look into copying a lot of the logic from something like the `github:repo:create` module which I think has some special management for the tokens.

https://github.com/backstage/backstage/blob/8743beff4d843cc0fe0896a04407369f2db5c902/plugins/scaffolder-backend-module-github/src/actions/githubRepoCreate.ts#L180-L186

benjidotsh:
@GersonTf are you still working on this? We also want this functionality and if not, I don't mind taking over the work. 😄

[...]

PR #28706, opened 2025-02-03
feat(scaffolder): add support for autocompletion to `GitHubRepoPicker`

benjidotsh:
This PR adds support for autocompletion to `GitHubRepoPicker`, resolving #27884.

This required a small refactor of `getOctokitOptions`, moving it to a more generic `util.ts` file outside of the `actions` directory and adding support for only providing `host`. It also clears @benjdlambert's `TODO` in the code. 😄 

I have also combined the `owners` and `repositories` resources into a `repositoriesWithOwner` resource to avoid pulling in multiple sources (since owners can be organizations and users) and keeping it as simple as possible. This also benefits the rate limit since we only have to do 1 call for both resources.

### DIFF
--- a/.changeset/fast-rabbits-unite.md
+++ b/.changeset/fast-rabbits-unite.md
@@ -2,7 +2,7 @@
 '@backstage/plugin-scaffolder-backend-module-github': patch
 ---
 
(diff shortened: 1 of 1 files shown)


benjidotsh:
Not sure why it's saying `You have uncommitted changes to the public API or reports of a package.` for `backstage/plugin-scaffolder-backend`...

benjdlambert (approved):
Nice! think that this is looking pretty good! :tada:

PR #28707, opened 2025-02-03
feat(gerrit): support glob discovery

tcardonne:

This PR adds support for wildcards in the Gerrit discovery configuration:

```diff
catalog:
  providers:
    gerrit:
      all:
        host: gerrit.company.com
        query: 'state=ACTIVE&type=CODE'
+       catalogPath: '**/catalog-info.{yml,yaml}'
```

To support this, the `GerritUrlReader` now implements the `search` method for pattern URLs. This search, just like the `GithubUrlReader` downloads the tree of the repository using Gitiles API. It then iterates and filters the result using `minimatch`.

This was tested locally using our Gerrit instance and it allows to find all Backstage catalog manifests in our 500+ repositories including one monorepo with ~800 catalog files. It takes less than one minute to ingest all of these 🚀 

The `GerritUrlReader` are still skipped but locally they are passing!

### DIFF
--- a/packages/backend-defaults/src/entrypoints/urlReader/lib/GerritUrlReader.ts
+++ b/packages/backend-defaults/src/entrypoints/urlReader/lib/GerritUrlReader.ts
@@ -277,7 +277,6 @@ export class GerritUrlReader implements UrlReaderService {
       this.integration.config.gitilesBaseUrl,
       getGitilesAuthenticationUrl(this.integration.config),
     );
-    console.log(treeUrl);
 
     const treeResponse = await fetch(treeUrl, {
       ...getGerritRequestOptions(this.integration.config),
--- a/packages/backend-defaults/src/entrypoints/urlReader/lib/GerritUrlReader.ts
+++ b/packages/backend-defaults/src/entrypoints/urlReader/lib/GerritUrlReader.ts
@@ -43,6 +43,8 @@ import {
   assertError,
 } from '@backstage/errors';
 import { ReadTreeResponseFactory, ReaderFactory } from './types';
+import { Minimatch } from 'minimatch';
+import { getGitilesAuthenticationUrl } from '@backstage/integration';
 
 /**
  * Implements a {@link @backstage/backend-plugin-api#UrlReaderService} for files in Gerrit.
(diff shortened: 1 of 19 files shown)


tcardonne:
With https://github.com/backstage/backstage/pull/28379 merged this PR can be opened 🚀 

This PR has been attempted before (https://github.com/backstage/backstage/pull/25704) but it required to move the "wildcard" pattern detection inside each URL readers.

cc @benjdlambert @Rugvip @vinzscam as you were involved in the previous PR 🙏

plugins/catalog-backend-module-gerrit/src/providers/GerritEntityProvider.ts
```diff
@@ -178,10 +183,44 @@ export class GerritEntityProvider implements EntityProvider {
...
+      throw new Error(`Failed to get project's HEAD for ${project}, ${e}`);
+    }
+
+    if (!response.ok) {
+      throw await ResponseError.fromResponse(response);
+    }
+
+    // Gerrit responds with something like `refs/heads/master`
+    const projectHeadResponse = (await parseGerritJsonResponse(
+      response as any,
```
benjdlambert, line 218: Do we need this cast here?

benjdlambert (changes requested):
Hey :wave: sorry this has taken a bit to get round to - few small bits

packages/backend-defaults/src/entrypoints/urlReader/lib/GerritUrlReader.ts
```diff
@@ -260,4 +266,66 @@ export class GerritUrlReader implements UrlReaderService {
...
+    url: string,
+    options?: UrlReaderServiceReadTreeOptions,
+  ): Promise<UrlReaderServiceSearchResponse['files']> {
+    const { path, basePath } = parseGitilesUrlRef(this.integration.config, url);
+
+    const treeUrl = `${basePath}/?format=JSON&recursive`.replace(
+      this.integration.config.gitilesBaseUrl,
+      getGitilesAuthenticationUrl(this.integration.config),
+    );
+    console.log(treeUrl);
```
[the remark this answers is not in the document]
tcardonne: Woops!

plugins/catalog-backend-module-gerrit/src/providers/GerritEntityProvider.ts
tcardonne: Yes you're right, we don't. Good catch!

I removed the cast here and on another similar call (which I copied) 👍

awanlin (approved):
Docs looks good, thanks @tcardonne 👍

benjdlambert (approved):
nice, let's go with this and iterate.

We've been speaking a little bit about breaking out some of the integrations module for a while, so that they would be in their respective provider modules, like `-github` and `-gerrit` etc, but that's not for now, just something to bear in mind in the future.

:pray:

tcardonne:
Thanks @benjdlambert 🎉

> We've been speaking a little bit about breaking out some of the integrations module for a while, so that they would be in their respective provider modules, like `-github` and `-gerrit` etc, but that's not for now, just something to bear in mind in the future.

Yes, for the Gerrit integration I think there's some cleaning/refactoring to do. There's some leftovers of the firsts implementations (ie, Gitiles wasn't mandatory at first but it's now mandatory, but some code and comments still handle the case of a missing Gitiles installation). Also the tests are currently skipped.

If I find some time I may try to do some cleaning. Should I wait until they are breaked apart or it's ok to start on this?

As we discussed in the last SIG Framework meeting and as we're heavily using the Gerrit integration, I can help by being codeowner of the integration 👍

PR #28708, opened 2025-02-03
Backend rate limit

drodil:

uses express-rate-limit together with custom store that uses CacheService to save the request rate for each client. configuration allows controlling almost all possible configuration available in the express-rate-limit library.

reopened #26725 


packages/backend-defaults/src/entrypoints/rootHttpRouter/http/RateLimitStoreFactory.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  private static redis(storeConfig: Config): Store {
+    const connectionString = storeConfig.getString('connection');
+    const KeyvRedis = require('@keyv/redis').default;
+    const keyv = new KeyvRedis(connectionString);
+    return new RedisStore({
+      sendCommand: async (...args: string[]) => {
+        const client = await keyv.getClient();
```
freben, line 51: is it safe/performant to do this for every single command? Like it doesn't make duplicates and churns memory etc right

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
```diff
@@ -227,6 +231,85 @@ export class MiddlewareFactory {
...
+        return (
+          Boolean(req.ip && ipAllowList.includes(req.ip)) ||
+          Boolean(
+            req.socket.remoteAddress &&
+              ipAllowList.includes(req.socket.remoteAddress),
+          )
+        );
+      },
+      validate: {
+        trustProxy: false,
```
freben, line 305: should this be false always?

freben (commented):
OK some more thoughts.

I pondered one thing while reading this. We have analytics based on service token principals. It would be kinda nice to base rate limits on principals rather than IPs. Some principals are going to be deployed to hundreds of pods, some to only one. This will of course only apply to authed traffic, but maybe there's a world where we extract the "bucket key" as either the principal if it exists, and otherwise the IP?

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/RateLimitStoreFactory.ts
drodil: Yep, checked the source. First time it will check if the client is connected and if not, will connect it. Otherwise it will just return the same instance.

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/MiddlewareFactory.ts
drodil: Yep, the app setting controls this here: https://github.com/backstage/backstage/pull/28708/files/a70c2429dbf419130f9ea158455473947102760b#diff-86ad1b6a694dd250823aee39d410428dd837c9d9a04ca8c33bd1081fbe3f22afR120

.changeset/famous-terms-rescue.md
```diff
@@ -0,0 +1,15 @@
+---
+'@backstage/backend-defaults': patch
+---
+
+Added new rate limit middleware to allow rate limiting requests to the backend
+
+If you are using the `configure` callback of the root HTTP router service and do NOT call `applyDefaults()` inside it, please see [the relevant changes](https://github.com/backstage/backstage/pull/26725/files#diff-86ad1b6a694dd250823aee39d410428dd837c9d9a04ca8c33bd1081fbe3f22af) that were made, to see if you want to apply them as well to your custom configuration.
```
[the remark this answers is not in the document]
drodil: True.

drodil:
> OK some more thoughts.
> 
> I pondered one thing while reading this. We have analytics based on service token principals. It would be kinda nice to base rate limits on principals rather than IPs. Some principals are going to be deployed to hundreds of pods, some to only one. This will of course only apply to authed traffic, but maybe there's a world where we extract the "bucket key" as either the principal if it exists, and otherwise the IP?

Hm. It might be possible to get the principal out of the request and use it as key here: https://github.com/backstage/backstage/pull/28708/files#diff-a74a65743502b54c75cd0766ea9840ac4251a89786564a07a1a3a9002e179c7aR290

But that would probably also mean that principals can be added to the current `ipAllowList`. Have to think about this.

drodil:
> Hm. For postgres limiting, it made me wonder if this could be an always per-plugin thing, and it using the actual regular plugin database service just like e.g. the scheduler does

Yeah, that could be possible and why not even with Redis using the CacheService? But it differs quite a lot from the current PR scope so I would suggest that the plugin-specific rate limiting is another PR unless it makes sense to only allow it per plugin and not for the whole backend?

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/RateLimitStoreFactory.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  private static postgres(storeConfig: Config): Store {
+    const connection = storeConfig.get('connection') as any;
+    const isConnectionString =
+      typeof connection === 'string' || connection instanceof String;
+    const connectionOptions = isConnectionString
+      ? parsePgConnectionString(connection as string)
+      : connection;
+    return new PostgresStore(connectionOptions, 'rl');
+  }
```
Rugvip, line 65: Bit torn on this one tbh. I think I'd expect to be able to use the current DB configuration when using this one? One potential path forward for that though could be that we add `database` or something like that as a store in the future if we want to, and wire that up to the plugin DB.

Would you use this one at this point? I'm thinking it might be simplest just to skip it for now since it's an extra postgres config to deal with otherwise.

packages/backend-defaults/config.d.ts
```diff
@@ -609,6 +617,68 @@ export interface Config {
...
 
+    /**
+     * Rate limiting options. Defining this as `true` will enable rate limiting with default values.
+     */
+    rateLimit?:
+      | true
+      | {
+          store?:
+            | {
+                client: 'redis';
```
Rugvip, line 628: Thinking that `type` might be a better name for this key? Similar to for example `backend.auth.pluginKeyStore`

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
```diff
@@ -87,6 +87,10 @@ const rootHttpRouterServiceFactoryWithOptions = (
       const logger = rootLogger.child({ service: 'rootHttpRouter' });
       const app = express();
 
+      const trustProxy = config.getOptional<boolean | number>(
+        'backend.trustProxy',
+      );
```
Rugvip, line 92: Not a super safe way to read this value, since it could end up being anything at runtime depending on what people put in config. Adding a full check for the type and then using the bool/number reader is a bit much to inline here though since they is intended as a piece of code that is easy to copy. A middleground would be if we could wrap this up somehow, but tbh I'm not seeing any great way to do that atm other than adding something new like `configureExpressApp(app, config)` but I don't love the idea of that.

We'll need to update https://backstage.io/docs/backend-system/core-services/root-http-router#via-code as well though

Rugvip (commented):
Nice! 👍 

Happy to add this is an optional capability and I think this is more or less it, just a couple of things to figure out

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
drodil: I changed it to be just a boolean for now. I think the `Config` requires some way to get either one without copy-pasting always the same type check code.

packages/backend-defaults/config.d.ts
drodil: Sure. Also was wondering if we should have the automatic config for the store in case it's not defined; first check the `cache.store` for `redis` connection, and then use `postgresql` from `database.client` and lastly use memory store. But this is something to improve later.

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/RateLimitStoreFactory.ts
drodil: Removed support for the postgres for now.

.changeset/famous-terms-rescue.md
```diff
@@ -0,0 +1,27 @@
...
+
+```yaml
+backend:
+  rateLimit:
+    global: false # This will disable the global rate limiting
+    plugin:
+      catalog:
+        window: 6s
+        incomingRequestLimit: 100
+```
```
awanlin, line 27: We should be documenting this config somewhere beyond just this changeset but I'm not exactly sure the best place. 🤔

awanlin (changes requested):
Thanks @drodil, would like to have this new config documented somewhere beyond just the changeset.

drodil:
Not stale

freben:
Waiting for some docs right? Just checking

drodil:
Yeah, should just figure out a place to put those :D

Parsifal-M:
> Yeah, should just figure out a place to put those :D

Heya! 👋 

I think maybe somewhere around here? https://backstage.io/docs/backend-system/core-services/http-router (if you feel it would be better somewhere else also fine!)

At least for now?

drodil:
Not stale, just waiting for time to finish this

freben:
Added https://github.com/backstage/backstage/pull/29562 as a discussion piece so we at least ensure that that bit lands. But don't rebase just yet, I don't know whether the other maintainers agree with adding it separately

drodil:
@freben thanks for the update! Will check this once back from holiday next week!

drodil:
rebased

drodil:
Not stale, needs a rebase again

.changeset/famous-terms-rescue.md
drodil: Added docs to microsite in http router section

drodil:
Not stale

drodil:
@awanlin the docs are there, WDYT?

docs/backend-system/core-services/http-router.md
```diff
@@ -68,6 +68,61 @@ For those routes you will also have to specify `allowLimitedAccess: true` when
 using the [`auth`](./auth.md) and [`httpAuth`](./http-auth.md) services to
 access the incoming credentials.
 
+## Rate limiting
+
+Rate limiting allows you to limit the amount of requests users can send to you backend.
+This is useful for blocking various network attacks, such as DDOS, if your instance does not
+have additional firewall configured to handle this.
```
Parsifal-M: Maybe we don't need this part? I think its clear with just the first sentence, then its just up to you why you'd want to do that. WDYT?

docs/backend-system/core-services/http-router.md
```diff
@@ -68,6 +68,61 @@ For those routes you will also have to specify `allowLimitedAccess: true` when
...
+
+To configure the default rate limiting, add the following to your config:
```
Parsifal-M:
```suggestion
To enable rate limiting, add the following to your config:
```

docs/backend-system/core-services/http-router.md
drodil: Yeah, makes sense, will change

packages/backend-defaults/src/lib/RateLimitStoreFactory.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  private static redis(options: { store: Config; prefix?: string }): Store {
+    const { store, prefix } = options;
+    const connectionString = store.getString('connection');
+    const KeyvRedis = require('@keyv/redis').default;
+    const keyv = new KeyvRedis(connectionString);
+    return new RedisStore({
+      prefix,
+      sendCommand: async (...args: string[]) => {
+        const client = await keyv.getClient();
```
freben, line 53: i hope this doesn't generate a new client for every request and piles up connections and leads to garbage collection churn 🤔

freben (approved):
Nice! Let's merge and iterate.

PR #28726, opened 2025-02-05
poc: use case for instance metadata

aramissennyeydd:

Quick POC following the discussion of #28129 in the SIG. Goal here is to align on a use case to drive design.

My thinking is that this could be a new DevTools page with the ability to list plugin tasks and for each task, see the last run date, task status, and be able to manually trigger the task. AFAIK those aren't currently supported from the scheduler service interface, so it would be an additional set of steps to add.

Screenshot 2025-02-04 at 10 50 19 PM

### DIFF
--- a/packages/app/src/components/devtools/CustomDevToolsPage.tsx
+++ b/packages/app/src/components/devtools/CustomDevToolsPage.tsx
@@ -18,6 +18,7 @@ import {
   ConfigContent,
   ExternalDependenciesContent,
   InfoContent,
+  ListSchedulesContent,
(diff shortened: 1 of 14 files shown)


aramissennyeydd:
Planning to update this PR to be the frontend for https://github.com/backstage/backstage/pull/29553

Parsifal-M:
Just QQ! Would this look better if we used the [CodeSnippet](https://backstage.io/docs/reference/core-components.codesnippet/) component? I have not checked but I assume the "thin" version would also support JSON? it would then also give a copy button etc

Just a little thought 😅

Issue #22318, opened 2024-01-18
🐛 Bug Report: GithubOrgEntityProvider may fail to read Users and Groups depending on the size of the organization


sonikro:
### 📜 Description

When running the **GithubOrgEntityProvider** to populate **Users** and **Groups** based on Teams from a specific **GitHub** Organization, I'm getting intermittent timeout errors from Github. (github.com). The problem happens on this query:

https://github.com/backstage/backstage/blob/def682a59d66291a7c56de2f46f730ffd25bb037/plugins/catalog-backend-module-github/src/lib/github.ts#L186-L220

And it's happening because the organization I'm using has a huge number of Teams and Members inside of those teams, which causes **100** Teams to fetch too much data at once. It takes more than 10s for github.com to respond, and several times, Github will throw a 5xx error saying it couldn't respond in time.

This is an intermittent error, it comes and goes, since it depends on github.com response time. However, lowering the number of teams down to 50 allowed me to get a consistent response back.

### 👍 Expected behavior

Users and Groups should've been populated based on GitHub Teams

### 👎 Actual Behavior with Screenshots

> GithubOrgEntityProvider:github refresh failed, HttpError: Unknown error: {"data":null,"errors":[{"message":"Something went wrong while executing your query. This may be the result of a timeout, or it could be a GitHub bug. Please include <REMOVED>` when reporting this issue."}]} Unknown error: {"data":null,"errors":[{"message":"Something went wrong while executing your query. This may be the result of a timeout, or it could be a GitHub bug. Please include `<REMOVED>` when reporting this issue."}]}

Request Body:

```graphql
{"query":"
    query teams($org: String!, $cursor: String) {
      organization(login: $org) {
        teams(first: 100, after: $cursor) {
          pageInfo { hasNextPage, endCursor }
          nodes {
            slug
            combinedSlug
            name
            description
            avatarUrl
            editTeamUrl
            parentTeam { slug }
            members(first: 100, membership: IMMEDIATE) {
              pageInfo { hasNextPage }
              nodes {
                avatarUrl,
                bio,
                email,
                login,
                name,
                organizationVerifiedDomainEmails(login: $org)
               }
            }
          }
        }
      }
    }","variables":{"org":"<REMOVED>"}}
```

### 👟 Reproduction steps

1. Setup GithubOrgEntityProvider to a org with hundreds of teams and each team might have thousands of users
2. Run the scheduled job to fetch Users and Groups
3. Wait for exception to happen

[...]

PR #28741, opened 2025-02-06
fix(catalog-backend-module-github): decrease number of repos fetched

Roohn:

This PR is sort of a continuation of [this issue](https://github.com/backstage/backstage/issues/22318) that was solved in [this PR](https://github.com/backstage/backstage/pull/22523) by the same solution as this PR.

The change I made was to decrease the amount of repositories that are fetched by the `getOrganizationRepositories` query. It was resulting in a timeout on the Github side. After cloning and testing I found out that lowering it from 50 to 40 solves the issue. 

The error I got: 
```
Something went wrong while executing your query. This may be the result of a timeout, or it could be a GitHub bug. Please include `F0B2:197BE6:517000:536DC1:67A35A7C` when reporting this issue.\"}]}","name":"HttpError","plugin":"catalog","request":{"body":"{\"query\":\"\\n    query repositories($org: String!, $catalogPathRef: String!, $cursor: String) {\\n      repositoryOwner(login: $org) {\\n        login\\n        repositories(first: 50, after: $cursor) {\\n          nodes {\\n            name\\n            catalogInfoFile: object(expression: $catalogPathRef) {\\n              __typename\\n              ... on Blob {\\n                id\\n                text\\n              }\\n            }\\n            url\\n            isArchived\\n            isFork\\n            visibility\\n            repositoryTopics(first: 100)...........
```

### DIFF
--- a/.changeset/green-kings-cheat.md
+++ b/.changeset/green-kings-cheat.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-github': patch
+---
+
+Decreased number of repositories fetched per page by GraphQL query, due to timeouts when running against big organizations.
--- a/plugins/catalog-backend-module-github/src/lib/github.ts
+++ b/plugins/catalog-backend-module-github/src/lib/github.ts
@@ -26,9 +26,9 @@ import {
 } from './defaultTransformers';
 import { withLocations } from './withLocations';
 
+import { LoggerService } from '@backstage/backend-plugin-api';
 import { DeferredEntity } from '@backstage/plugin-catalog-node';
 import { Octokit } from '@octokit/core';
-import { LoggerService } from '@backstage/backend-plugin-api';
 import { throttling } from '@octokit/plugin-throttling';
 // Graphql types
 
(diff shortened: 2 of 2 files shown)


mhitkaohp:
Hi @Roohn. I am facing the same issue (I guess we are just around some threshold, after months with no such problem the calls to GH to retrieve repositories recently started to fail more and more often, now failing almost every time). Did you find some other solution, resp. why did you close the PR?

ojathelonius:
@mhitkaohp we recently started facing the same consistent issue as well. Decreasing the page size did not help, even with very low numbers e.g. 5 or 10 instead of 100. You can try it yourself, even the simplest GraphQL queries will eventually timeout:

```
query users {
    organization(login: "your_org") {
        membersWithRole(first: 20) {
            pageInfo {
                hasNextPage
                endCursor
            }
            nodes {
                login
            }
        }
    }
}
// sometimes takes >10s
```

I'm starting to think this might be an issue with GitHub itself... Lowering the pagination number to minimal values will cause rate limitation issues, although these are somewhat worked around using the throttled Octokit client. I am considering converting the problematic GraphQL calls with plain REST calls which are way more performant.

Issue #27347, opened 2024-10-25
🐛 Bug Report: Techdocs Memory Leak


pcgqueiroz:
### 📜 Description

The `techdocs-backend` plugin is experiencing a memory leak when integrated with the `search-backend-module-techdocs`. In a production setup with a dedicated pod for `techdocs-backend`, the `/static/docs` endpoint is queried by the `search-backend-module-techdocs` every 10 minutes. With over 3000 entities to process, memory consumption increases progressively until the JavaScript heap is exhausted, resulting in a crash.

### 👍 Expected behavior

The `techdocs-backend` plugin should handle repeated `/static/docs` queries and large volumes of entities without excessive memory consumption or memory leaks, allowing stable operation in production.

### 👎 Actual Behavior with Screenshots

The memory usage of the `techdocs-backend` plugin gradually increases each time it processes requests from the `search-backend-module-techdocs`, eventually causing the JavaScript heap to overflow and crash.

![image](https://github.com/user-attachments/assets/24f66fda-2ca6-43b9-9966-353bedccfc39)

```log
<--- Last few GCs --->
[1:0x65c76b0] 20854057 ms: Scavenge 2021.6 (2074.2) -> 2019.2 (2076.2) MB, 11.9 / 0.0 ms  (average mu = 0.941, current mu = 0.519) allocation failure;
[1:0x65c76b0] 20854079 ms: Scavenge 2023.1 (2076.2) -> 2021.2 (2079.2) MB, 14.9 / 0.0 ms  (average mu = 0.941, current mu = 0.519) allocation failure;
[1:0x65c76b0] 20855608 ms: Scavenge 2026.6 (2079.2) -> 2024.2 (2097.2) MB, 1527.2 / 0.0 ms  (average mu = 0.941, current mu = 0.519) allocation failure;
<--- JS stacktrace --->
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0xb87bc0 node::Abort() [node]
 2: 0xa96834  [node]
 3: 0xd687f0 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xd68b97 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xf462a5  [node]
[...]

PR #28742, opened 2025-02-06
feat(techdocs): generator timeout config

drodil:

this allows specifying techdocs generator timeout in configuration that prevents techdocs generation to take infinite time in some cases like with huge `plantuml` diagrams and local `mkdocs` renderer.

it does not fix the potential memory leak #27347 but might improve the situation so that sockets get closed in case the generating is blocking the system.

the timeout works both with the default docker generator as well as with the local `mkdocs`.

### DIFF
--- a/.changeset/slow-dots-applaud.md
+++ b/.changeset/slow-dots-applaud.md
@@ -1,0 +1,10 @@
+---
+'@backstage/plugin-techdocs-backend': patch
+'@backstage/plugin-techdocs-node': patch
+---
+
+Support timeout in the techdocs generator
+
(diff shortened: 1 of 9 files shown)


drodil:
After investigation, it looks like the memory leak is not related to generating getting stuck but instead the static cache middleware. Anyways I think this feature should be in place as the generation can get stuck or at least take very long time blocking the CPU.

squid-ney:
Hi @drodil, since this timeout is managed by docker and not TechDocs are there any logs or errors coming surfaced to the techdocs-backend if the timeout is reached? Also, could you make a documentation update?

PR #28748, opened 2025-02-06
Documenting the experimental feature of "Retries and Recovery" in Scaffolder

acierto:

Documenting the experimental feature of "Retries and Recovery" in Scaffolder


docs/features/software-templates/experimental.md
```diff
@@ -13,7 +13,64 @@ Please leave feedback on these features in the [Backstage Discord](https://disco
...
 ## Retries and Recovery
 
-### TODO
+Do you have long-running tasks, and can they recover and proceed after Backstage redeploy?
+
+Then you definitely will benefit from enabling the experimental scaffolder task recovery feature.
+
+Whenever you do redeploy, on startup there will be a check of all tasks in "processing" state that you identified in your template as they are capable of starting over.
+
+This is an example of how you can do it:
```
Parsifal-M, line 22: Looking forward to trying this! Just wondering, is it better to try and explain to the reader why this was implemented (to solve the problem of) instead of a nice feature you can enable? I think it adds more weight and context.

When I read it what I am imagining is that I may use this if I have slow/long scaffolder tasks and if someone redeploys the Backstage app it will fail/lose any and all progress (the problem) and then this would be the solution right? LMK if I am misunderstanding 😅

docs/features/software-templates/experimental.md
```diff
@@ -13,7 +13,64 @@ Please leave feedback on these features in the [Backstage Discord](https://disco
...
+apiVersion: scaffolder.backstage.io/v1beta3
+kind: Template
+metadata:
+  name: recoverable-template
+spec:
+  EXPERIMENTAL_recovery:
+    EXPERIMENTAL_strategy: startOver
+```
+
+And to enable the recovery feature itself, you can do it by adding this snippet into your app-config.yaml file:
```
Parsifal-M, line 34:
```suggestion
To enable the recovery feature, add this snippet into your app-config.yaml file:
```

docs/features/software-templates/experimental.md
```diff
@@ -13,7 +13,64 @@ Please leave feedback on these features in the [Backstage Discord](https://disco
...
+
+```yaml
+scaffolder:
+  EXPERIMENTAL_recoverTasks: true
+```
+
+By default, all the tasks that are in a processing state and have a heartbeat of more than 30 seconds will be automatically recovered.
```
Parsifal-M, line 41:
```suggestion
By default, the tasks that are in a processing state and have a heartbeat of more than 30 seconds will be automatically recovered.
```

docs/features/software-templates/experimental.md
```diff
@@ -13,7 +13,64 @@ Please leave feedback on these features in the [Backstage Discord](https://disco
...
+This implies that the task's status will shift to "Open," initiating its execution from the beginning. So you have to be sure that your tasks can run multiple times.
```
Parsifal-M, line 42: Could you help me understand this part a little more? Tasks will shift to an Open status if they where in processing and had a heartbeat of more than 30 seconds post redeploy/interruption? and then re-initialise execution? 💭

Parsifal-M (commented):
Heya! 👋 

Looks like a really cool feature! Just some small comments! LMK what you think? 👏

docs/features/software-templates/experimental.md
acierto: Hey @Parsifal-M thank you for your feedback. That's quite valid point. 
Do you think pointing to BEP would be sufficient here? Just to not duplicated the same information here?

https://github.com/backstage/backstage/tree/master/beps/0004-scaffolder-task-idempotency

docs/features/software-templates/experimental.md
acierto: Maybe I have to rephrase it if it's confusing 😅 

The gist is that after Backstage restart all tasks which were running would be in "processing" state. 
Though they will be doing nothing.

The default behaviour now in Backstage to change the status to "failed" when certain timeout of no heartbeat reached. 

In case when you want to make the process of running tasks seamless to users (no failed tasks due to Backstage restart), and not add extra work of checking that no running tasks now in the queue, so Backstage can be restarted, the recovery feature can be enabled.

When such task "in limbo" (processing state, no heartbeat for a certain time) is found, we will start executing the task from the start (that what happens if to change the status of task to "Open").

docs/features/software-templates/experimental.md
```diff
@@ -13,7 +13,64 @@ Please leave feedback on these features in the [Backstage Discord](https://disco
...
+You can look at how to incorporate [checkpoints](https://backstage.io/docs/features/software-templates/writing-custom-actions#using-checkpoints-in-custom-actions-experimental) into your custom actions to achieve that.
+
+In case 30 seconds of no heartbeat time is not appropriate for your case to restart the task, you can customize it for your needs with the configuration:
```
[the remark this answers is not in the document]
acierto: I'm not sure 😅 

It's a terminology from the task in scaffolder

Screenshot 2025-02-07 at 19 53 39

I wouldn't use another term for it here.
Maybe better to add some extra doc to explain how scaffolder engine works internally.
And from here to point to the doc.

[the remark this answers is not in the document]
Parsifal-M: Ah yeah, that makes sense then 😄 let's keep it as is otherwise will confuse people

docs/features/software-templates/experimental.md
Parsifal-M: Oh I see, I think if we explain this behaviour in that context it will be easier to understand 👍  WDYT? Like a TLDR version 😅

docs/features/software-templates/experimental.md
Parsifal-M: Hey! Hmm, good point I can't think of any major issues with pointing to the BEP as reference at least, only maybe later on if things start to drift slightly 😅  but I think it should be fine

Having read your other explanation further down in another [comment](https://github.com/backstage/backstage/pull/28748#discussion_r1947013146) I think this is also probably fine as is 😃

docs/features/software-templates/experimental.md
```diff
@@ -13,7 +13,66 @@ Please leave feedback on these features in the [Backstage Discord](https://disco
...
+```yaml
+scaffolder:
+  EXPERIMENTAL_workspaceSerialization: true
+```
+
+By default, the serialized workspace will be stored in your database.
+
+If you work with large files, it might not be the best option for you.
+
+At this moment we support the integration with the GCP bucket; to switch the serialization to this provider, you can with:
```
freben, line 67:
```suggestion
At the moment we also support integration with Google GCS; to switch the serialization to this provider, you can do so with:
```

docs/features/software-templates/experimental.md
```diff
@@ -13,7 +13,66 @@ Please leave feedback on these features in the [Backstage Discord](https://disco
...
+
+```yaml
+scaffolder:
+  EXPERIMENTAL_workspaceSerializationProvider: gcpBucket
+  EXPERIMENTAL_workspaceSerializationGcpBucketName: name-of-your-bucket
```
freben, line 72: Ooh these two were probably unfortunately named. The product name is `GCS` or Cloud Storage; GCP is the entire platform. Like S3 vs AWS. I'd recommend changing these two in code and perhaps providing transparent backwards compatibility with these config key names too but mark them deprecated.

Even better perhaps:

```yaml
scaffolder:
  EXPERIMENTAL_workspaceSerializationProvider:
    type: gcs
    bucketName: name-of-your-bucket
```

then you can trivially add more providers with a bit nicer structure (it's also a pattern we use in several other places) and TypScript being more like `{ type: a, ...} | { type: b, ...}`

freben (approved):
Preapproving but some suggestions along the way

docs/features/software-templates/experimental.md
```diff
@@ -13,7 +13,66 @@ Please leave feedback on these features in the [Backstage Discord](https://disco
 
 ## Retries and Recovery
 
-### TODO
+Do you have long-running tasks, and can they recover and proceed after Backstage redeploy?
```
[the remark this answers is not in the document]
acierto: If to use current scaffolder pre-built actions, then the tasks are not recoverable till we introduce everywhere checkpoints. 

If they the actions themselves, most probably they should be able to answer on this question.

Do you ask it to give more context here or/and to rephrase this sentence?

docs/features/software-templates/experimental.md
acierto: Make sense. 

Yeah, I was thinking of platform naming and then the feature. 

Ok, I'll change to this structure!

docs/features/software-templates/experimental.md
acierto: Included the visualization of it, usually it's easier to grasp :)

docs/features/software-templates/experimental.md
acierto: @freben Just realised that if we will change to 

```
scaffolder:
  EXPERIMENTAL_workspaceSerializationProvider:
    type: gcs
    bucketName: name-of-your-bucket
```

we can do it only without deprecation, as this variable is already used as a string.

This feature is experimental, so what is better to do, to use alternative name to `EXPERIMENTAL_workspaceSerializationProvider` or to introduce a breaking change?

docs/features/software-templates/experimental.md
acierto: @benjdlambert Is it ok to add visuals to docs or better to avoid it?

docs/features/software-templates/experimental.md
benjdlambert: I think that we can rename this when we move off experimental, let's not worry about this too much right now.

docs/features/software-templates/experimental.md
benjdlambert: I think we can skip for now, I reworded some things and hopefully it's little better.

Thinking that we can introduce diagrams later, but would rather do it with draw.io or whatever we've used in other places so that they can be edited over time rather than a png 👍

benjdlambert:
Gonna merge this and iterate - wanna get something on the docs site for the mainline release that goes out today :tada:

PR #28753, opened 2025-02-07
Allow disabling the header divider on homepage cards

kunickiaj:

Some homepage cards would like to define their own header for example, so that a title can be set dynamically.

Removing the divider with the divider prop provides a way to put everything in the content area without an extraneous divider above.

The divider is usually not easily visible using the default theme, however customized themes that may use a larger border-radius this divider becomes visible and distracting.

### Default Theme
The divider is visible, but blends into the border, (All content including the "header" is in the content section)
Screenshot 2025-02-06 at 8 59 39 AM

Homepage Card that uses the title prop, and all content in the content area, no extraneous divider.
Screenshot 2025-02-10 at 9 55 49 AM

### Modified Theme (border radius)
The divider becomes visible. Note that rather than using the `title` field we are still using the content area because we have a more complex header than a simple title. It is dynamic and has multiple sections

Screenshot 2025-02-06 at 9 00 50 AM

### Using the noDivider prop from this PR
Screenshot 2025-02-06 at 9 03 37 AM

### From the default widgets with and without noDivider
Screenshot 2025-02-10 at 1 22 55 PM
Screenshot 2025-02-10 at 1 23 18 PM

Verified using `<HomePageRandomJoke noDivider />`

### DIFF
--- a/.changeset/late-hairs-kneel.md
+++ b/.changeset/late-hairs-kneel.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-home-react': patch
+---
+
+Don't render header divider on homepage cards if no title was specified.
--- a/plugins/home-react/src/extensions.tsx
+++ b/plugins/home-react/src/extensions.tsx
@@ -47,7 +47,9 @@ export type RendererProps = { title?: string } & ComponentParts;
 /**
  * @public
  */
-export type CardExtensionProps<T> = ComponentRenderer & { title?: string } & T;
+export type CardExtensionProps<T> = ComponentRenderer & {
+  title?: string;
+} & T;
 
 /**
  * @public
@@ -150,7 +152,7 @@ function CardExtension<T>(props: CardExtensionComponentProps<T>) {
   }
 
   const cardProps = {
-    ...(title && { title }),
+    ...(title && { title, divider: !!title }),
     ...(Settings && !isCustomizable
       ? {
           action: (


.changeset/late-hairs-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-home-react': minor
+---
+
+Allow disabling the header divider on homepage cards
```
awanlin, line 5: If I'm understanding this correctly existing cards would need to be updated as well to support this? I tried to play around with these but not luck: https://github.com/backstage/backstage/blob/4de4602e941397dfd6129219dec69d89e80ff17c/packages/app/src/components/home/HomePage.tsx#L96. If I'm on the right page we should probably mention that here.

awanlin (changes requested):
Thanks for this, code change itself makes sense to me, left one comment about the changeset. Also, can you update the PR with screen shots showing with and without the divider. Super handy to have for later reference.

.changeset/late-hairs-kneel.md
kunickiaj: Existing cards should not need any updates, as it's an opt-in feature. It is also not a visible change unless you're modified the theme in certain ways. Let me add some screenshots.

awanlin (approved):
Thanks for the follow up @kunickiaj 👍

tikabom (commented):
prefer the prop to be renamed to `hideDivider`.

kunickiaj:
@tikabom updated prop name to `hideDivider`

.changeset/late-hairs-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-home-react': minor
```
vinzscam, line 2:
```suggestion
'@backstage/plugin-home-react': patch
```

kunickiaj:
> > prefer the prop to be renamed to `hideDivider`.
> 
> one argument to use `divider` is to be consistent with the `<InfoCard />` component (used under the hood), which uses `divider` defaulted to `true` 🤷‍♂️

yeah, let me see if i can make that consistent.

freben:
Just a question on this one.

Is it not the case that this should be tied to the title? If you don't supply a title, should the divider also not be drawn just as a side effect?

kunickiaj:
> Just a question on this one.
> 
> Is it not the case that this should be tied to the title? If you don't supply a title, should the divider also not be drawn just as a side effect?

I missed the forest for the trees... that's an excellent point and simplifies this PR a lot 🤣

awanlin (approved):
Very nice!

vinzscam (approved):
🙏

Issue #28745, opened 2025-02-06
🐛 Bug Report: Incorrect number of search results sent in search analytics event


drapalaa:
### 📜 Description

The `search` event captured by the AnalyticsApi when the user enters a search term contains an incorrect number of search results.

### 👍 Expected behavior

The `search` event should be sent with the number of search results corresponding to the entered term.

### 👎 Actual Behavior with Screenshots

The `search` event is sent with the number of search results received for the previously entered term.

![Image](https://github.com/user-attachments/assets/1e2a4e61-d6b6-47e8-90b5-46a4d0f6308f)

![Image](https://github.com/user-attachments/assets/c9a350c7-50b0-4aa2-91a2-5e2cb458f24a)

### 👟 Reproduction steps

1. Click the Search button in the sidebar to open the search modal.
2. Type a term in the search bar to get some search results.
3. Type another term to get a different number of search results.

### 📃 Provide the context for the Bug.

Probably the bug was introduced here: #21911

### 👀 Have you spent some time to check if this bug has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

benjdlambert:
ping @backstage/search-maintainers

PR #28769, opened 2025-02-07
fix: capturing the correct number of search results in the search ana…

drapalaa:

Closes: #28745

The `search` analytics event was captured with the number of search results received for the previous term.
I have made a change so that the number of search results is the value received for the term entered.

### DIFF
--- a/plugins/search-react/src/context/SearchContext.test.tsx
+++ b/plugins/search-react/src/context/SearchContext.test.tsx
@@ -436,7 +436,7 @@ describe('SearchContext', () => {
       </TestApiProvider>
     );
 
-    it('captures analytics events if enabled in app', async () => {
+    it('captures analytics events for non-empty term', async () => {
       searchApiMock.query
         .mockResolvedValueOnce({
           results: [],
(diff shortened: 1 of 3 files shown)


awanlin (approved):
Thanks @drapalaa, this makes sense to me after a bit of validation.

@emmaindal, could you give this a look as well, you made changes in this area in the past.

plugins/search-react/src/context/SearchContext.test.tsx
```diff
@@ -421,31 +421,34 @@ describe('SearchContext', () => {
...
+          {children}
+        </SearchContextProvider>
+      </TestApiProvider>
+    );
 
-      searchApiMock.query.mockResolvedValue({
-        results: [],
-        numberOfResults: 3,
-      });
+    it('captures analytics events if enabled in app', async () => {
```
johnphilip283, line 439: thanks for this fix! would you mind adding/updating a test matching this exact scenario - i.e. two separate search events happening with a non-empty term where the order and payload to the analytics API is as expected?

drapalaa: Hi @johnphilip283. I have updated the test so that it checks events for two non-empty terms.

johnphilip283 (approved):
Thanks for the fix!

PR #28782, opened 2025-02-10
Add support for providing values and labels to the search filters

grantila:
## Add support for labels to search page/modal filters

Currently the search filters, e.g. `SearchFilter.Select` only allows an array of strings as values. This means that filters for entities, groups, domains, etc. can only show the raw value that is being filtered.

This PR forwards the support from MUI's `Select` and `Checkbox` to separate the underlying `value` from its rendered `label` to the `SearchFilter.Select` and `SearchFilter.Checkbox`. Custom filters can then e.g. render the `title` of an entity rather than its `name`:

#### Before

Skärmavbild 2025-02-10 kl  09 34 07

#### After

Skärmavbild 2025-02-10 kl  09 34 32

### DIFF
--- a/packages/app/src/components/search/SearchModal.tsx
+++ b/packages/app/src/components/search/SearchModal.tsx
@@ -148,15 +148,18 @@ export const SearchModal = ({ toggleModal }: { toggleModal: () => void }) => {
                   values={async () => {
                     // Return a list of entities which are documented.
                     const { items } = await catalogApi.getEntities({
-                      fields: ['metadata.name'],
+                      fields: ['metadata.name', 'metadata.title'],
                       filter: {
                         'metadata.annotations.backstage.io/techdocs-ref':
                           CATALOG_FILTER_EXISTS,
                       },
                     });
 
-                    const names = items.map(entity => entity.metadata.name);
-                    names.sort();
+                    const names = items.map(entity => ({
+                      value: entity.metadata.name,
+                      label: entity.metadata.title ?? entity.metadata.name,
+                    }));
+                    names.sort((a, b) => a.label.localeCompare(b.label));
                     return names;
                   }}
                 />
--- a/packages/app/src/components/search/SearchPage.tsx
+++ b/packages/app/src/components/search/SearchPage.tsx
@@ -97,15 +97,18 @@ const SearchPage = () => {
                     values={async () => {
                       // Return a list of entities which are documented.
                       const { items } = await catalogApi.getEntities({
-                        fields: ['metadata.name'],
+                        fields: ['metadata.name', 'metadata.title'],
                         filter: {
                           'metadata.annotations.backstage.io/techdocs-ref':
                             CATALOG_FILTER_EXISTS,
                         },
                       });
 
-                      const names = items.map(entity => entity.metadata.name);
-                      names.sort();
+                      const names = items.map(entity => ({
+                        value: entity.metadata.name,
+                        label: entity.metadata.title ?? entity.metadata.name,
+                      }));
+                      names.sort((a, b) => a.label.localeCompare(b.label));
                       return names;
                     }}
                   />
--- a/plugins/search-react/src/components/SearchFilter/SearchFilter.Autocomplete.tsx
+++ b/plugins/search-react/src/components/SearchFilter/SearchFilter.Autocomplete.tsx
@@ -25,6 +25,7 @@ import Autocomplete, {
 import { useSearch } from '../../context';
 import { useAsyncFilterValues, useDefaultFilterValue } from './hooks';
 import { SearchFilterComponentProps } from './SearchFilter';
+import { ensureFilterValueWithLabel, FilterValueWithLabel } from './types';
 
 /**
  * @public
@@ -55,7 +56,9 @@ export const AutocompleteFilter = (props: SearchAutocompleteFilterProps) => {
   const asyncValues =
     typeof givenValues === 'function' ? givenValues : undefined;
   const defaultValues =
-    typeof givenValues === 'function' ? undefined : givenValues;
+    typeof givenValues === 'function'
+      ? undefined
+      : givenValues?.map(v => ensureFilterValueWithLabel(v));
   const { value: values, loading } = useAsyncFilterValues(
     asyncValues,
     inputValue,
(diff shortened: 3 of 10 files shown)


packages/app/src/components/search/SearchModal.tsx
```diff
@@ -148,15 +148,18 @@ export const SearchModal = ({ toggleModal }: { toggleModal: () => void }) => {
...
                         'metadata.annotations.backstage.io/techdocs-ref':
                           CATALOG_FILTER_EXISTS,
                       },
                     });
 
-                    const names = items.map(entity => entity.metadata.name);
-                    names.sort();
+                    const names = items.map(entity => ({
+                      value: entity.metadata.name,
+                      label: entity.metadata.title ?? entity.metadata.name,
```
awanlin, line 160: Can we use the Presentation API - it was built for handling this type of things - instead? More details here: https://github.com/backstage/backstage/pull/17474. This would go for the `SearchPage` as well 👍

awanlin (changes requested):
Thanks for this @grantila, overall this makes a lot of sense to me, one comment I left is that we should try and use the Presentation API for this.

packages/app/src/components/search/SearchPage.tsx
```diff
@@ -98,15 +98,18 @@ const SearchPage = () => {
...
                           'metadata.annotations.backstage.io/techdocs-ref':
                             CATALOG_FILTER_EXISTS,
                         },
                       });
 
-                      const names = items.map(entity => entity.metadata.name);
-                      names.sort();
+                      const names = items.map(entity => ({
+                        value: entity.metadata.name,
+                        label: entity.metadata.title ?? entity.metadata.name,
```
johnphilip283, line 110: same note as above wrt/ usage of the Presentation API!

plugins/search-react/src/components/SearchFilter/SearchFilter.tsx
```diff
@@ -123,21 +126,23 @@ export const CheckboxFilter = (props: SearchFilterComponentProps) => {
       fullWidth
       data-testid="search-checkboxfilter-next"
     >
-      {label ? <FormLabel className={classes.label}>{label}</FormLabel> : null}
-      {values.map((value: string) => (
+      {formLabel ? (
+        <FormLabel className={classes.label}>{formLabel}</FormLabel>
+      ) : null}
```
johnphilip283, line 131:
```suggestion
      {!!formLabel  && (
        <FormLabel className={classes.label}>{formLabel}</FormLabel>
      )}
```

plugins/search-react/src/components/SearchFilter/SearchFilter.Autocomplete.tsx
```diff
@@ -64,18 +67,25 @@ export const AutocompleteFilter = (props: SearchAutocompleteFilterProps) => {
   );
   const { filters, setFilters } = useSearch();
   const filterValue =
-    (filters[name] as string | string[] | undefined) || (multiple ? [] : null);
+    ensureFilterValueWithLabel(
+      filters[name] as string | string[] | undefined,
+    ) || (multiple ? [] : null);
```
johnphilip283, line 72: nitpick: this took me a minute to parse what was going on here with the ternary + complicated typing in one line - would you mind splitting this up a bit to make it a little easier to read?

plugins/search-react/src/components/SearchFilter/SearchFilter.tsx
grantila: This wasn't changed by me, apart from a variable name change (causing line breaks), but I can change it.

plugins/search-react/src/components/SearchFilter/SearchFilter.Autocomplete.tsx
grantila: Wasn't technically changed by me, but I'll extract out half of it to another variable for somewhat more clarity.

packages/app/src/components/search/SearchModal.tsx
grantila: I think this could make sense, but it's not straight forward to change this without some extracting of this logic elsewhere.

Using the presentation API for this within this async function could be done in two ways:
 * Use a component instead of string as label (and hoping MUI renders it correctly - its API only supports strings as labels). This would not allow for ordering the names alpha-numerically.
 * Use the presentation API's `forEntity` and wait for each promise to resolve, then take the resulting array (, sort it) and return that.

I have now done the latter - extracted this fetching of items to a separate function (hook), re-used in both the search page and modal.

packages/app/src/components/search/SearchPage.tsx
grantila: See above

grantila:
@awanlin @johnphilip283 Do you think this would be good to go, from your point of view?

grantila:
Let me know if there's anything else I can do to proceed with this

packages/app/src/components/search/use-techdocs-entities.ts
```diff
@@ -0,0 +1,61 @@
...
+        'metadata.annotations.backstage.io/techdocs-ref': CATALOG_FILTER_EXISTS,
+      },
+    });
+
+    const presentationEntities = await Promise.all(
+      items.map(
+        entity => presentationApi.forEntity(stringifyEntityRef(entity)).promise,
+      ),
+    );
+    const entityNameByRef = new Map(
```
vinzscam, line 44: cc @awanlin, I think we should roll back the previous change, as using the presentation API this way doesn't scale. For example, if we receive 1000 entities, we'll end up sending approximately 1000 requests back to the server to retrieve essentially the same data we already have.

I propose rolling back [#76738ec4dd7270516a301e74e7cea01fd9775d49](https://github.com/backstage/backstage/pull/28782/commits/76738ec4dd7270516a301e74e7cea01fd9775d49).

Since we're discussing this, @grantila, do you think it would make sense to add a limit to `getEntities`, or would that introduce side effects?

vinzscam (changes requested):

packages/app/src/components/search/use-techdocs-entities.ts
awanlin: That's fine with me, thanks for catching this @vinzscam 👍 

Sorry for the extra work on this @grantila

johnphilip283 (approved):
thanks for the changes!

vinzscam (approved):
:shipit:

awanlin (approved):
🚀

PR #28784, opened 2025-02-10
Storybookcontributiondocs

deepthi-28:
Note design changed 

image

image


awanlin:
Hi @deepthi-28, we appreciate the contributions but as these changes are very low priority they won't get reviewed as quickly as other changes. Also, it seems you are going over all the documentation looking for these changes, it would be better to just submit one PR with all the changes instead of many PRs with a single change.

deepthi-28:
Ok Thank you.

PR #28787, opened 2025-02-10
feat(catalog-react): entityOwnerPicker fetches full entities in 'owne…

Teemukoivumaa:
instead of returning just a partial entity without title / displayName information, fetch the full entity and return that instead so that it works like the 'all'-mode

Before the EntityOwnerPicker with the 'owners-only'-mode returned just the entity name and that was used for the picker. This PR modifys the functionality so that it fetches the full entity information. 

Before:
https://github.com/user-attachments/assets/d3cccbbf-eaa0-498e-bea5-7e79ff658b04

After:
https://github.com/user-attachments/assets/1b0ccc08-18fe-4457-981d-d661f7dcfa95

### DIFF
--- a/plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
+++ b/plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
@@ -13,11 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { useApi } from '@backstage/core-plugin-api';
+import { ProfileInfo, useApi } from '@backstage/core-plugin-api';
 import useAsyncFn from 'react-use/esm/useAsyncFn';
 import { catalogApiRef } from '../../api';
 import { useState } from 'react';
-import { Entity, parseEntityRef } from '@backstage/catalog-model';
+import { Entity } from '@backstage/catalog-model';
 
 type FacetsCursor = {
   start: number;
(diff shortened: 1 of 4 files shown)


Teemukoivumaa:
Not sure if this is a major change, but it suggested that automatically :)

plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
```diff
@@ -44,36 +44,37 @@ type FacetsInitialRequest = {
...
+
+      const catalogResponse = await catalogApi.getEntitiesByRefs({
+        entityRefs: entityRefs,
+      });
+
+      const entities = catalogResponse.items.filter(
+        (item): item is Entity => item !== undefined,
+      );
+
+      return entities.sort(
```
freben, line 66: if this is the sort you are performing, i'd recommend just plain sorting the ref strings that you get out of `response.facets` above instead - MUCH cheaper and gets the same result

plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
```diff
@@ -44,36 +44,37 @@ type FacetsInitialRequest = {
...
-          ),
-      )
-      .catch(() => []);
+    try {
+      const entityRefs = await catalogApi
+        .getEntityFacets({ facets: [facet] })
+        .then(response => response.facets[facet].map(e => e.value));
+
+      const catalogResponse = await catalogApi.getEntitiesByRefs({
+        entityRefs: entityRefs,
```
freben, line 59: Hm. There can easily be tens of thousands of users in the facets response in a big org. Just a heads up that this can get LARGE and take a nontrivial amount of time.

One thing you can add here is `fields: ['kind', 'metadata.name', ...]` for all of the things that you actually are filtering on. That makes it a little bit less bad at least.

Teemukoivumaa: Hmm yeah need to think that, on our instance I get 2078 entities for the `relations.ownedBy`.

Tested the field thing quickly and it looks promising! I'll continue testing to get this a bit smaller

freben: You were still working on this right?

Teemukoivumaa: Yeah sorry I was on holiday and pretty busy with work. I'll try to push my changes tomorrow/today :)

Teemukoivumaa:
Whoops didn't mean to do that..😅

plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
Teemukoivumaa: Can you check the current implementation to see if that's what you were after? The `catalogApi.getEntitiesByRefs` doesn't return the entityref, so needed to do it like that. Works the same as before 👍🏼

plugins/catalog-react/src/components/EntityOwnerPicker/useFacetsEntities.ts
```diff
@@ -44,36 +44,43 @@ type FacetsInitialRequest = {
...
+          if (!entity) return null;
+          return entity;
+        })
+        .filter(entity => entity !== null);
+
+      return ownerEntities.sort(
+        (a, b) =>
+          entityRefs.indexOf(a.metadata.name) -
+          entityRefs.indexOf(b.metadata.name),
+      );
```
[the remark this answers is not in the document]
Teemukoivumaa: What I'm trying to do is return them sorted by kind and entity title, like it was done before my changes 😅

Teemukoivumaa:
Not stale

vinzscam (changes requested):
sorry but unfortunately this approach doesn't scale for large catalog. It was already proposed before and we had to rollback. See #25366 and #27839

Teemukoivumaa:
Damn :/

PR #28793, opened 2025-02-10
Fixing small issue that breaks client open API generation

jefmsmit:

There is a small bug in the open API generation code for generating client code. Specifically, a command is trying to run lint on the directory of the generated file instead of just the file itself.

### DIFF
--- a/packages/repo-tools/src/commands/package/schema/openapi/generate/client.ts
+++ b/packages/repo-tools/src/commands/package/schema/openapi/generate/client.ts
@@ -83,14 +83,15 @@ async function generate(
   );
 
   const parentDirectory = resolve(resolvedOutputDirectory, '..');
+  const indexFile = resolve(parentDirectory, 'index.ts');
 
   await fs.writeFile(
-    resolve(parentDirectory, 'index.ts'),
+    indexFile,
     `// 
     export * from './generated';`,
   );
 
-  await exec(`yarn backstage-cli package lint --fix ${parentDirectory}`, [], {
+  await exec(`yarn backstage-cli package lint --fix ${indexFile}`, [], {
     signal: abortSignal?.signal,
   });
 
--- a/.changeset/long-cars-raise.md
+++ b/.changeset/long-cars-raise.md
@@ -1,0 +1,5 @@
+---
+'@backstage/repo-tools': minor
+---
+
+Fixing a client generation issue due to a misconfigured lint command.


packages/repo-tools/src/commands/package/schema/openapi/generate/client.ts
```diff
@@ -83,14 +83,15 @@ async function generate(
...
 
   await fs.writeFile(
-    resolve(parentDirectory, 'index.ts'),
+    indexFile,
     `// 
     export * from './generated';`,
   );
 
-  await exec(`yarn backstage-cli package lint --fix ${parentDirectory}`, [], {
+  await exec(`yarn backstage-cli package lint --fix ${indexFile}`, [], {
```
aramissennyeydd, line 94: I believe this was intentional, we want to lint everything (for this repo that adds copyright comments and fixes up the templates for lint rules). There are custom rules set for the generated content, https://github.com/backstage/backstage/blob/b78b2b0ed677e9677dd4ff581bfa0f46c09b0f65/packages/cli/config/eslint-factory.js#L169-L185. 

What's the issue this is causing?

aramissennyeydd: Are there any lint errors in the directory? Passing a directory to ESLint/`backstage-cli`'s wrapper should be acceptable.

aramissennyeydd: The fix here should probably be to properly log the lint errors and let consumers decide if they want to update their eslint config to ignore them or fix them manually.

jefmsmit: In my use case, the only file in the directory is the newly generated file that needs to be lint checked. I tried to use the command from the terminal and it does not work if you give it a directory.

jefmsmit: @aramissennyeydd you are correct, the tool does work with a directory as the path. Yes, there are other lint errors in that directory when I run `yarn backstage-cli package lint --fix <path to directory>` that are not fixable with `--fix`. Investigating a bit further.

jefmsmit: This was just me with a misconfigured eslint, sorry about that. Closing.

aramissennyeydd: All good, glad we got you sorted 😁! If you're still interested in a contribution here, better logs would be much appreciated.

PR #28795, opened 2025-02-10
Custom catalog filtering within TechDocs collator

byan1197:
- Added a custom collator filter extension point to account for niche catalog setups.

### DIFF
--- a/.changeset/blue-camels-fry.md
+++ b/.changeset/blue-camels-fry.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-search-backend-module-techdocs': patch
+---
+
+Added an extension point that allows for custom entity filtering during document collation.
--- a/plugins/search-backend-module-techdocs/src/collators/DefaultTechDocsCollatorFactory.ts
+++ b/plugins/search-backend-module-techdocs/src/collators/DefaultTechDocsCollatorFactory.ts
@@ -68,6 +68,7 @@ export type TechDocsCollatorFactoryOptions = {
   legacyPathCasing?: boolean;
   entityTransformer?: TechDocsCollatorEntityTransformer;
   documentTransformer?: TechDocsCollatorDocumentTransformer;
+  entityFilter?: (entity: Entity[]) => Entity[];
 };
 
 type EntityInfo = {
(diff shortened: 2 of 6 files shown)


plugins/search-backend-module-techdocs/src/collators/DefaultTechDocsCollatorFactory.ts
```diff
@@ -177,81 +185,83 @@ export class DefaultTechDocsCollatorFactory implements DocumentCollatorFactory {
...
-              // https://github.com/node-fetch/node-fetch/issues/665
-              const searchIndex = await Promise.race([
-                searchIndexResponse.json(),
-                new Promise((_resolve, reject) => {
-                  setTimeout(() => {
-                    reject('Could not parse JSON in 5 seconds.');
-                  }, 5000);
-                }),
-              ]);
+            // todo(@backstage/techdocs-core): remove Promise.race() when node-fetch is 3.x+
```
drodil, line 223: Not related to this change but is this still needed as the `node-fetch` was removed? @freben ?

.changeset/blue-camels-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-search-backend-module-techdocs': patch
+---
+
+Added an extension point that allows for custom entity filtering during document collation.
```
awanlin, line 5: Is there a reason we did this as an extension over having it in config like we do for the Catalog Collator?

byan1197: Just to have a bit more control of what's happening during the filtering! From the catalog response, we'd like to filter for something on the annotation level that requires some extra checks, I think this would be the least invasive way to do it without overhauling the catalog request.

alexlorenzi: It allows for filtering that Catalog doesn't support... for example, where `backstage.io/techdocs-ref != 'N/A'` which is a use case we have internally

awanlin: Thanks @byan1197 and @alexlorenzi, that helps! I had just updated the filter docs for the Catalog Collator and I was thinking it would be nice for them to be consistent but I can understand that would not work well for your use case.

awanlin (changes requested):
Thanks for adding this @byan1197, will be handy for others! Can we add some documentation about this new extension point, please? Perhaps here: https://github.com/backstage/backstage/blob/master/docs/features/search/collators.md

squid-ney (approved):

awanlin (approved):
Awesome! Thanks for adding details in the docs. 🚀

PR #28798, opened 2025-02-11
Deprecate `createTemplateAction` with old `JSONSchema` and `zod` declarations

benjdlambert:
Replaces #26957


plugins/scaffolder-backend-module-confluence-to-markdown/src/actions/confluence/confluenceToMarkdown.ts
```diff
@@ -66,7 +66,6 @@ export const createConfluenceToMarkdownAction = (options: {
               'Paste your Confluence url. Ensure it follows this format: https://{confluence+base+url}/display/{spacekey}/{page+title} or https://{confluence+base+url}/spaces/{spacekey}/pages/1234567/{page+title} for Confluence Cloud',
             items: {
               type: 'string',
-              default: 'Confluence URL',
```
benjdlambert, line 69: This is actually not part of the `Schema` type, wondering if it should be and that the type is wrong.

Should the type actually be `JSONSchema7` instead from `json-schema`, thinking that this could break actions otherwise. It looks like the types were not inferred properly before, and now they are. @freben

.changeset/dull-olives-itch.md
Rugvip, line 1: Potentially split this one

plugins/scaffolder-node/report.api.md
```diff
@@ -442,6 +509,7 @@ export type TaskStatus =
 export type TemplateAction<
   TActionInput extends JsonObject = JsonObject,
   TActionOutput extends JsonObject = JsonObject,
+  TSchemaType extends 'v1' | 'v2' = 'v1',
```
Rugvip, line 512: Would defaulting to `'v2'` work too? Makes it a bit trickier to use this one directly

Rugvip (approved):
TS ✨ 

looks good to go, with some minor bits

freben (approved):

plugins/scaffolder-node/report.api.md
benjdlambert: unfortunately not, as it looks like it breaks any usage today of just `const: TemplateAction<any, any> = createTemplateAction(...)` which seems to be in a lot of tests.

do you think it's worth doing it anyways?

Rugvip: no I think it's alright to leave for when we remove v1

plugins/scaffolder-node/src/actions/types.ts
```diff
@@ -21,86 +21,154 @@ import { TaskSecrets } from '../tasks';
...
-
-  /**
-   * Task information
-   */
-  task: {
-    id: string;
-  };
+  TSchemaType extends 'v1' | 'v2' = 'v1',
+> = TSchemaType extends 'v2'
+  ? {
```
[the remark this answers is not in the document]
benjdlambert: Maybe, but really wanted to avoid the version in exported types to be honest.

drodil:
As having tens of custom actions it would be nice to have support to ship also codemods with backstage releases for this kind of changes :)

PR #28799, opened 2025-02-11
fix(core-components): Fix misalignment and uppercase text for sidebar dropdown button

bestrand:

Fixed a bug in the `SidebarSubmenuItem` component within the `core-components` package that caused the dropdown button to be misaligned in the sidebar and the button text to appear in uppercase due to the default `<Button>` behavior. Also added an example dropdown button to the 'Home' submenu for reference/example and future UI testing.

Since the `<Button>` component already was modified I added the specific styling directly.

**Before:** 

Screenshot 2025-02-11 at 13 28 13

The same can be seen in the Storybook sample ([Link](https://backstage.io/storybook/?path=/story/layout-sidebar--sample-scalable-sidebar)):

image

**After:** 

image

### DIFF
--- a/packages/app/src/components/Root/Root.tsx
+++ b/packages/app/src/components/Root/Root.tsx
@@ -132,25 +132,6 @@ export const Root = ({ children }: PropsWithChildren<{}>) => (
               to="catalog?filters[kind]=user"
               icon={useApp().getSystemIcon('kind:user')}
             />
-            <SidebarDivider />
-            <SidebarSubmenuItem
-              title="Templates"
-              icon={useApp().getSystemIcon('kind:template')}
-              dropdownItems={[
-                {
-                  title: 'Documentation',
-                  to: 'catalog?filters[kind]=template&filters[type]=documentation',
-                },
-                {
-                  title: 'Service',
-                  to: 'catalog?filters[kind]=template&filters[type]=service',
-                },
-                {
-                  title: 'Website',
-                  to: 'catalog?filters[kind]=template&filters[type]=website',
-                },
-              ]}
-            />
           </SidebarSubmenu>
         </SidebarItem>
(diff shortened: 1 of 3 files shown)


packages/core-components/src/layout/Sidebar/SidebarSubmenuItem.tsx
```diff
@@ -185,6 +185,10 @@ export const SidebarSubmenuItem = (props: SidebarSubmenuItemProps) => {
               classes.item,
               isActive ? classes.selected : undefined,
             )}
+            style={{
+              textTransform: 'none',
+              justifyContent: 'flex-start',
+            }}
```
freben, line 191: Hm. We avoid using `style`, and instead prefer class names in the `makeStyles` above. Mixing these gets really iffy.

freben (changes requested):

packages/core-components/src/layout/Sidebar/SidebarSubmenuItem.tsx
bestrand: Thanks for info! Moved it to makeStyles now.

packages/app/src/components/Root/Root.tsx
```diff
@@ -132,6 +132,25 @@ export const Root = ({ children }: PropsWithChildren<{}>) => (
...
+                {
+                  title: 'Service',
+                  to: 'catalog?filters[kind]=template&filters[type]=service',
+                },
+                {
+                  title: 'Website',
+                  to: 'catalog?filters[kind]=template&filters[type]=website',
+                },
+              ]}
+            />
```
freben, line 153:
```suggestion
```

Just removing this as unrelated for now.

freben (approved):

Issue #17188, opened 2023-03-31
🐛 Bug Report: `publish:github:pull-request` fails with `HttpError: You have exceeded a secondary rate limit.`


rmartine-ias:
### 📜 Description

The scaffolder `publish:github:pull-request` action fails to make a PR with the error `Pull request creation failed; caused by HttpError: You have exceeded a secondary rate limit. Please wait a few minutes before you try again.`

We think this is because the underlying plugin [used by backstage](https://github.com/backstage/backstage/blob/master/plugins/scaffolder-backend/src/scaffolder/actions/builtin/publish/githubPullRequest.ts#L41) to create pull requests is inefficient in its queries in some way that scales with the [number of files](https://github.com/backstage/backstage/blob/391b7fbd28405d00b38769867b5cd4a7973ebcbf/plugins/scaffolder-backend/src/scaffolder/actions/builtin/publish/githubPullRequest.ts#L274) in the `sourcePath`.

Here is what I think is happening:

backstage is [sending](https://github.com/backstage/backstage/blob/391b7fbd28405d00b38769867b5cd4a7973ebcbf/plugins/scaffolder-backend/src/scaffolder/actions/builtin/publish/githubPullRequest.ts#L274) [`type: File`](https://github.com/gr2m/octokit-plugin-create-pull-request/blob/10e9a309ce5f21a0ab9ef3d2372c5d927f9ebc83/src/types.ts#L32)s to `createPullRequest`, which calls [`createTree`](https://github.com/gr2m/octokit-plugin-create-pull-request/blob/10e9a309ce5f21a0ab9ef3d2372c5d927f9ebc83/src/compose-create-pull-request.ts#L103), which [runs](https://github.com/gr2m/octokit-plugin-create-pull-request/blob/db6b15fa1522a2426b88edbd0c20041065a33776/src/create-tree.ts#L24) a `Promise.all` on `Object.keys(changes.files)` (it makes a new promise for each file passed to it, and fires them all at the same time, if I understand correctly). Each `Promise` [calls](https://github.com/gr2m/octokit-plugin-create-pull-request/blob/db6b15fa1522a2426b88edbd0c20041065a33776/src/create-tree.ts#L81) `valueToTreeObject`. Annotated and condensed:

```javascript
export async function valueToTreeObject(
  ...
  value: string | File // It's a File!
) {
  ...
  if (typeof value === "string") {
    // Nope!
  }
  
  // This is what gets called!
  const { data } = await octokit.request(
[...]

PR #28816, opened 2025-02-12
feat: allow signing git commits from scaffolder

### LINKED ISSUE #25934: 🚀 Feature: allow signing git commits from scaffolder action
### 🔖 Feature description

For example, Github scaffolder action `github:repo:push` uses the helper function here to make a commit using isomorphic-git underneath: https://github.com/backstage/backstage/blob/master/plugins/scaffolder-node/src/actions/gitHelpers.ts#L23 

The other git related scaffolder actions use the same helper but it doesn't support signing the commit.

### 🎤 Context

Some repositories might require that the commit is signed with a GPG key and using scaffolder actions to create these commits will fail in the pre-receive hook if signing support is missing. 

We need a new secret value to config/scaffolder that Backstage users can insert their private GPG key to be used for signing. Additionally, all scaffolder actions using the helper need to support a new parameter that determines whether the commit should be signed or not.

Integrating this to isomorphic-git is quite straightforward (https://github.com/backstage/backstage/blob/master/plugins/scaffolder-node/src/scm/git.ts#L145)

```ts
// @ts-ignore
import { pgp } from '@isomorphic-git/pgp-plugin';

return git.commit({ fs, dir, message, author, committer, signingKey, onSign: pgp.sign });
```

None

drodil:

adds support to sign git commits with configured PGP key in scaffolder actions. configuration can be done either by integration or by using the default signing key in the scaffolder config. note that this pgp-plugin is used for signing and that it is limited to using RSA keys and signatures made with SHA1 hashing algorithm.

this change does not support github commit signing as it cannot be done the same way due to GitHub app and REST API being used.

closes #25934

refers to #26333

### DIFF
--- a/packages/integration/config.d.ts
+++ b/packages/integration/config.d.ts
@@ -62,6 +62,10 @@ export interface Config {
         tenantId?: string;
         personalAccessToken?: string;
       }[];
+      /**
+       * PGP signing key for signing commits.
+       */
+      commitSigningKey?: string;
     }>;
 
     /**
@@ -94,6 +98,10 @@ export interface Config {
        * @visibility secret
        */
       appPassword?: string;
+      /**
+       * PGP signing key for signing commits.
+       */
+      commitSigningKey?: string;
     }>;
 
     /** Integration configuration for Bitbucket Cloud */
@@ -108,6 +116,10 @@ export interface Config {
        * @visibility secret
        */
       appPassword: string;
+      /**
+       * PGP signing key for signing commits.
+       */
+      commitSigningKey?: string;
     }>;
 
     /** Integration configuration for Bitbucket Server */
@@ -137,6 +149,10 @@ export interface Config {
        * @visibility frontend
        */
       apiBaseUrl?: string;
+      /**
+       * PGP signing key for signing commits.
+       */
+      commitSigningKey?: string;
     }>;
 
     /** Integration configuration for Gerrit */
@@ -172,6 +188,10 @@ export interface Config {
        * @visibility secret
        */
       password?: string;
+      /**
+       * PGP signing key for signing commits.
+       */
+      commitSigningKey?: string;
     }>;
 
     /** Integration configuration for GitHub */
@@ -233,6 +253,10 @@ export interface Config {
          */
         allowedInstallationOwners?: string[];
       }>;
+      /**
+       * PGP signing key for signing commits.
+       */
+      commitSigningKey?: string;
     }>;
 
     /** Integration configuration for GitLab */
@@ -269,6 +293,10 @@ export interface Config {
        * @visibility frontend
        */
       baseUrl?: string;
+      /**
+       * PGP signing key for signing commits.
+       */
+      commitSigningKey?: string;
     }>;
 
     /** Integration configuration for Google Cloud Storage */
@@ -349,6 +377,10 @@ export interface Config {
        * @visibility secret
        */
       password?: string;
+      /**
+       * PGP signing key for signing commits.
+       */
+      commitSigningKey?: string;
     }>;
     /** Integration configuration for Harness Code */
     harness?: Array<{
--- a/packages/integration/config.d.ts
+++ b/packages/integration/config.d.ts
@@ -64,6 +64,7 @@ export interface Config {
       }[];
       /**
        * PGP signing key for signing commits.
+       * @visibility secret
        */
       commitSigningKey?: string;
     }>;
@@ -100,6 +101,7 @@ export interface Config {
       appPassword?: string;
       /**
        * PGP signing key for signing commits.
+       * @visibility secret
        */
       commitSigningKey?: string;
     }>;
@@ -118,6 +120,7 @@ export interface Config {
       appPassword: string;
       /**
        * PGP signing key for signing commits.
+       * @visibility secret
        */
       commitSigningKey?: string;
     }>;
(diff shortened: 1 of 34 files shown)


packages/integration/src/gitlab/config.ts
```diff
@@ -54,6 +54,11 @@ export type GitLabIntegrationConfig = {
    * If no baseUrl is provided, it will default to `https://${host}`
    */
   baseUrl: string;
+
```
benjdlambert, line 57: Think that these should all be `@visibility secret` right?

dotboris:
I wanted to chime in to say that signing GitHub commits can work using the PGP method but only when backstage is using the personal access token of a regular user for authentication to GitHub.

When your backstage instance is authenticated with a GitHub app which is, in my opinion, the right way to authenticate backstage to GitHub, then the PGP method can't work.

I'm mentioning this because you could technically support GitHub in this PR but it won't work in all scenarios and might encourage people to use personal access tokens which I think is a bad idea. It's a hard choice because either you support GitHub and it only works some of the time which is janky and will generate questions from users or you don't support GitHub which leaves behind some users who could have benefited on this.

I want to clarify that I'm not jumping in here requesting that you add GitHub support to your PR. I wanted to give more context. I honestly feel that this PR which supports everything but GitHub is a fine solution and a good starting point that can be built on top of for GitHub.

benjdlambert:
> I wanted to chime in to say that signing GitHub commits can work using the PGP method but only when backstage is using the personal access token of a regular user for authentication to GitHub.

How would this work though? We'd have to create a commit using the Github API rather than using `isomorphic-git` at all, as we can't get the `signingKey` from the users account through the GItHub API right?

There's the option to use the `users` oauth token and pass that into the `publish:github` action, so is there a way to use that oauth token or is that not enough?

dotboris:
There's 3 scenarios.

**Scenario 1: Personal Access Token Auth**

If backstage is configured with a personal access token:

```yaml
# app-config.yaml
integrations:
  github:
    - host: github.com
      token: ${GITHUB_TOKEN}
```

then that token grants access to a regular user which can configure commit signature keys. In that case, the backstage admin can create a PGP key for that github user and configure it in `app-config.yaml`. Then when the scaffolder is run, it can use `isomorphic-git` with the PGP key to create signed commits.

There's one caveat here. If you attempt to use the [user's OAuth token](https://backstage.io/docs/features/software-templates/writing-templates/#using-the-users-oauth-token) in your template, then this won't work.

**Scenario 2: GitHub App Auth**

There's no place in GitHub's app management UI to define a PGP key. So a backstage admin can try to configure a signing key in backstage and that key can be used in the scaffolder to sign the commits but the signature will not be recognized by GitHub.

In this scenario, the only right way to sign a commit is through the GitHub API.

**Scenario 3: Using the user's OAuth token**

https://backstage.io/docs/features/software-templates/writing-templates/#using-the-users-oauth-token

The scaffolder allows a template to request a token that lets it perform actions on behalf of the user running the current template. This is done through the GitHub OAuth app configured for the GitHub authentication provider.

When you use that token, you get the same situation as scenario 2 where you can use a PGP key to sign the commits just fine but GitHub won't recognize them.

In theory, this is where you'd need to use the GitHub API to sign the commits. I've personally had issues getting that to work and I'm not convinced if it's even possible. That's a whole other story we don't need to get into here.

The only scenario where signing git commits through `isomorphic-git` on GitHub is scenario 1.

benjdlambert:
@dotboris thanks for the detailed explanation, and it's exactly what I thought.

I'm not sure that it's going to be possible to do verified commits and PGP signing when using a users oauth token in GitHub, and I think that it's an acceptable workaround that if you want to do verified commits, you'll need to be using `integrations` config and a dedicated PAT from a user account that can be trusted.

It's unfortunate, but creating commits using the GitHub API is just not scalable as we've noticed with things like https://github.com/backstage/backstage/issues/17188

chromey:
So since it was apparently established that not supporting GH in this PR is acceptable, is it ok if we move forward with this and look into GH separately? This issue is currently preventing us from adopting the templating feature because we have an enterprise-wide requirement for signed commits.

dotboris:
> So since it was apparently established that not supporting GH in this PR is acceptable, is it ok if we move forward with this and look into GH separately? This issue is currently preventing us from adopting the templating feature because we have an enterprise-wide requirement for signed commits.

We have a similar requirement in my org as well. We use the template engine but we rely on the `publish:github:pull-request` which does sign commits correctly. When we create repos, we use the `github:repo:push` action which does not sign commit unfortunately. In our case, the commit signature is enforced at the PR level with org-wide rulesets. These don't apply to the initial commit so it works out for us. The pattern we use goes like this:

1. Create the repo
2. Push the initial commit with a minimal `CODEOWNERS` file
3. Create a PR with the full contents of the repository

packages/integration/config.d.ts
```diff
@@ -62,6 +62,11 @@ export interface Config {
         tenantId?: string;
         personalAccessToken?: string;
       }[];
+      /**
+       * PGP signing key for signing commits.
+       * @visibility secret
+       */
+      signingKey?: string;
```
benjdlambert, line 69: wonder if we should rename this to `commitSigningKey` instead? The PGP key isn't used for anything else right?

benjdlambert (commented):
Hey :wave: just a small thing about the naming but happy to ship this and we create a different issue for GitHub Commit Signing and follow up? :pray:

packages/integration/config.d.ts
drodil: Renamed to `commitSigningKey` and `defaultCommitSigningKey` accordingly

benjdlambert (approved):
Nice! let's give this a go! :tada:

PR #28832, opened 2025-02-12
fix(techdocs): /sync response when entity not found

alexlorenzi:

The the fetch-event-source mentions that if you don't re-throw the error in the `onerror` handler then it'll automatically retry the request. This was causing infinite requests for 404's

https://github.com/Azure/fetch-event-source/tree/main
```
    onerror(err) {
        if (err instanceof FatalError) {
            throw err; // rethrow to stop the operation
        } else {
            // do nothing to automatically retry. You can also
            // return a specific retry interval here.
        }
    }
```

### DIFF
--- a/.changeset/good-flowers-promise.md
+++ b/.changeset/good-flowers-promise.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-techdocs-backend': patch
+---
+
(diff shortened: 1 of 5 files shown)


alexlorenzi:
Guess only thought here is are there valid errors that we might want to retry for?

byan1197 (approved):

PeaWarrior (approved):

Issue #28831, opened 2025-02-12
🐛 Bug Report: events are rejected when are not JSON/Buffer


angeliski:
### 📜 Description

After the fix https://github.com/backstage/backstage/pull/27121 we introduced some break changes:
- The body now **needs** to be a buffer ([ref](https://github.com/backstage/backstage/pull/27121/files#diff-db929a4e919ad28705615d83233455afa61b37d822f6f80242fa078628209d04R114)) 
- We are rejecting **everything** is not `application/json` ([ref](https://github.com/backstage/backstage/pull/27121/files#diff-db929a4e919ad28705615d83233455afa61b37d822f6f80242fa078628209d04R126))

This is an assumption we made, but isn't really true for all use cases.
We use the event system to handle [slack interactions](https://api.slack.com/interactivity/handling#payloads), but Slack always send the content-type as `application/x-www-form-urlencoded`, what make impossible to match the `application/json` requirement. And we use the  `bodyParser.urlencoded({ extended: true })` so we don't have a Buffer when the `HttpPostIngressEventPublisher.ts` receives the request, what results in an error.

### 👍 Expected behavior

We should just propagate the event as received or provide a way to allow the user to custom the requirements (like content-type)

### 👎 Actual Behavior with Screenshots

We receive a error to process the slack event:
```
[...]

PR #28834, opened 2025-02-13
feat(events-backend,events-node) add a parser option to events ingress

angeliski:

Hey @pjungermann

I open this draft to discuss a solution to #28831
It's not a final version, just an idea to provide a way to override the parse to specific endpoints, like that:
```typescript
export const eventsModuleYourFeature = createBackendModule({
  pluginId: 'events',
  moduleId: 'your-feature',
  register(env) {
    // [...]
    env.registerInit({
      deps: {
        // [...]
        events: eventsExtensionPoint,
        // [...]
      },
      async init({ /* ... */ events /*, ... */ }) {
        // [...]
        events.addHttpPostIngress({
          topic: 'your-topic',
          parser: async (req, _topic) => {
            return {
              bodyParsed: JSON.stringify(req.body),
              bodyBuffer: Buffer.from(JSON.stringify(req.body)),
              encoding: 'utf-8'
            };
          }
        });
      },
    });
  },
});

```

WDYT?  

### DIFF
--- a/.changeset/ten-impalas-hope.md
+++ b/.changeset/ten-impalas-hope.md
@@ -1,6 +1,6 @@
 ---
-'@backstage/plugin-events-backend': minor
-'@backstage/plugin-events-node': minor
+'@backstage/plugin-events-backend': patch
+'@backstage/plugin-events-node': patch
 ---
 
 add `addHttpPostBodyParser` to events extension to allow body parse customization. This feature will enhance flexibility in handling HTTP POST requests in event-related operations.
--- a/.changeset/ten-impalas-hope.md
+++ b/.changeset/ten-impalas-hope.md
@@ -1,0 +1,6 @@
+---
+'@backstage/plugin-events-backend': minor
+'@backstage/plugin-events-node': minor
+---
+
+add `addHttpPostBodyParser` to events extension to allow body parse customization. This feature will enhance flexibility in handling HTTP POST requests in event-related operations.
--- a/plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
+++ b/plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
@@ -103,7 +103,7 @@ export class HttpPostIngressEventPublisher {
 
     router.post(path, async (request, response) => {
       const requestContentType = contentType.parse(request);
-      const bodyParser = this.bodyParsers[requestContentType.type];
+      const bodyParser = this.bodyParsers[requestContentType.type ?? ''];
 
       if (!bodyParser) {
         throw new UnsupportedMediaTypeError(requestContentType.type);
@@ -111,6 +111,7 @@ export class HttpPostIngressEventPublisher {
 
       const { bodyParsed, bodyBuffer, encoding } = await bodyParser(
         request,
+        requestContentType,
         topic,
       );
 
--- a/plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
+++ b/plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
@@ -16,35 +16,18 @@
 
 import { LoggerService } from '@backstage/backend-plugin-api';
 import { Config } from '@backstage/config';
-import { CustomErrorBase } from '@backstage/errors';
 import {
   EventsService,
+  HttpBodyParser,
   HttpPostIngressOptions,
   RequestValidator,
 } from '@backstage/plugin-events-node';
 import contentType from 'content-type';
 import express from 'express';
 import Router from 'express-promise-router';
+import { defaultHttpBodyParsers } from './body-parser';
 import { RequestValidationContextImpl } from './validation';
-
-class UnsupportedCharsetError extends CustomErrorBase {
-  name = 'UnsupportedCharsetError' as const;
-  statusCode = 415 as const;
-
-  constructor(charset: string) {
-    super(`Unsupported charset: ${charset}`);
-  }
-}
-
-class UnsupportedMediaTypeError extends CustomErrorBase {
-  name = 'UnsupportedMediaTypeError' as const;
-  statusCode = 415 as const;
-
-  constructor(mediaType?: string) {
-    super(`Unsupported media type: ${mediaType ?? 'unknown'}`);
-  }
-}
-
+import { UnsupportedMediaTypeError } from './errors';
 /**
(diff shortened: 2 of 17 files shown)


plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
```diff
@@ -45,6 +46,39 @@ class UnsupportedMediaTypeError extends CustomErrorBase {
   }
 }
 
+const defaultHttpDefaultBodyParser: HttpDefaultBodyParser = async (
```
pjungermann, line 49:
```suggestion
const defaultHttpBodyParser: HttpBodyParser = async (
```

I guess it is not a "default body parser", but a default implementation of a "body parser". However, maybe I miss something.

plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
```diff
@@ -45,6 +46,39 @@ class UnsupportedMediaTypeError extends CustomErrorBase {
...
+  if (
+    !parsedContentType.type ||
+    parsedContentType.type !== 'application/json'
+  ) {
+    throw new UnsupportedMediaTypeError(parsedContentType.type);
+  }
+
+  const encoding = parsedContentType.parameters.charset ?? 'utf-8';
+  if (!Buffer.isEncoding(encoding)) {
+    throw new UnsupportedCharsetError(encoding);
```
pjungermann, line 71: We could leave a hint here to provide a custom body parser.

pjungermann (commented):
Thank you @angeliski. I think your approach can work.

Providing a body parser for a topic endpoint has the benefit that it can be customized a lot.

I wonder whether we should provide a body parser per content type, though. We could make that pluggable via an extension point. Also, we could add an implementation for additional content types like `application/x-www-form-urlencoded`.

What do you think?

angeliski:
I like the idea @pjungermann 

 Here are some rules we can follow for the implementation:
-  We'll provide a default body parser based on the content type.
- We'll have two extension points to override the body parser: by content type and by topic.
- The topic override takes precedence over the content type override if both are defined.


In this way you can define a global parser for a specific content type or change just for an topic.

WDYT?

I will try to implement that soon

pjungermann:
The "by topic" option **might** not be needed assuming that the body of a certain content type will be parsed accordingly. Additional processing could be done at other levels like the event routers extracting a sub-topic and sending the event forward. E.g., by subscribing to such an event and creating a new one, etc.

The "by topic" option assumes that there is topic-specific body parsing necessary beyond the content type.

By chance, do you have a use case in mind already for it?

For the default body parser, we could add support for a set of content types, step by step increasing coverage as needed.
I'm also wondering whether we could reuse i.e. the body parsers by express or similar.

angeliski:
Hey @pjungermann 

I think, a put an implementation in place. Please give me some feedback so we can ship that (I am not really happy with the names I choose, so be my guest to suggest better options)

.changeset/ten-impalas-hope.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-events-backend': minor
+'@backstage/plugin-events-node': minor
```
pjungermann, line 3:
```suggestion
'@backstage/plugin-events-backend': patch
'@backstage/plugin-events-node': patch
```

As those are still non-stable packages, `patch` should be enough here as there is no breaking change.

plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
```diff
@@ -108,32 +102,17 @@ export class HttpPostIngressEventPublisher {
...
-      const bodyBuffer: Buffer = requestBody;
-      const parsedContentType = contentType.parse(request);
-      if (
-        !parsedContentType.type ||
-        parsedContentType.type !== 'application/json'
-      ) {
-        throw new UnsupportedMediaTypeError(parsedContentType.type);
-      }
+      const requestContentType = contentType.parse(request);
+      const bodyParser = this.bodyParsers[requestContentType.type];
```
pjungermann, line 106: As of the previous code, `parsedContentType.type` can be undefined which we should handle here, too.

plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
```diff
@@ -108,32 +102,17 @@ export class HttpPostIngressEventPublisher {
...
 
-      const bodyString = bodyBuffer.toString(encoding);
-      const bodyParsed =
-        parsedContentType.type === 'application/json'
-          ? JSON.parse(bodyString)
-          : bodyString;
+      const { bodyParsed, bodyBuffer, encoding } = await bodyParser(
+        request,
+        topic,
+      );
```
pjungermann, line 115: We could pass the parsed content type here as well. This would remove the need to parse it inside of `bodyParser(..)` again.
From the interface perspective, it seems not necessary, though.
The decision could depend on how heavy or lightweight the parsing is.

plugins/events-backend/src/service/http/errors.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export class UnsupportedMediaTypeError extends CustomErrorBase {
+  name = 'UnsupportedMediaTypeError' as const;
+  statusCode = 415 as const;
+
+  constructor(mediaType?: string) {
+    super(
+      `Unsupported media type: ${
+        mediaType ?? 'unknown'
+      }. You need to provide a custom body parser for this media type using events extension.`,
```
pjungermann, line 35:
```suggestion
      }. You need to provide a custom body parser for this media type using the EventsExtensionPoint.`,
```

pjungermann (commented):
Thank you @angeliski for your effort and contribution!

It looks very good to me. I left a few nitpick comments and suggestions.

plugins/events-backend/src/service/http/HttpPostIngressEventPublisher.ts
angeliski: I didn't find a way the type could be undefined [here](https://github.com/jshttp/content-type/blob/master/index.js#L106), but in any case I put a fallback just to be sure

pjungermann (approved):

angeliski:
Hey @vinzscam @Rugvip Can you help me here with a review?

Rugvip (approved):
Yep looks good! :sparkles: 

Thank you :+1:, let's :shipit:

PR #28845, opened 2025-02-13
[NFS] Entity Page Headers

camilaibs:
> [!WARNING]
> Merge after https://github.com/backstage/backstage/pull/28758

Now the default entity page header can be customized via new frontend system extensions.

Setting up a completely different default header component:
```jsx
import { EntityHeaderBlueprint } from '@backstage/plugin-catalog-react/alpha';

EntityHeaderBlueprint.make({
  name: 'my-default-header',
  params: {
    loader: () => import('./MyDefaultHeader').then(m => <m.MyDefaultHeader />),
  },
});
```

Disabling the header via configuration:
```yaml
# app-config.yaml
app:
  extensions:
    - entity-header:app/my-default-header: false
```


.changeset/nervous-cups-happen.md
```diff
@@ -0,0 +1,98 @@
...
+Changing a header default filter via configuration:
+
+```yaml
+# app-config.yaml
+app:
+  extensions:
+    - entity-header:app/my-template-header:
+        config:
+          # Using this custom header with components instead
+          filter: 'kind:component'
```
camilaibs, line 97: Maybe in the future we can also support setting a title/subtitle expression via config file. Something like:
```yaml
# app-config.yaml
app:
  extensions:
    - entity-header:app/my-template-header:
        config:
          title: '{{entity.metadata.name}}'
          subtitle: '{{entity.metadata.description}}'
```

.changeset/nervous-cups-happen.md
```diff
@@ -0,0 +1,98 @@
...
+
+EntityHeaderBlueprint.make({
+  name: 'my-default-header',
+  params: {
+    // The `FavoriteEntity` icon button is added by default
+    // You can also completely override the default title
+    // title: <MyTitleComponent />
+    title: { actions: [<CopyEntityNameToClipboard />] },
+    // A subtitle element is also supported.
+    // subtitle: <MySubtitleComponent />
```
camilaibs, line 45: Not sure if these should be lazy loaded, maybe accept both options?

```jsx
title: { actions: ReactNode | async () => Promise<ReactNode> }
```

plugins/catalog-react/report-alpha.api.md
```diff
@@ -323,6 +324,102 @@ export const EntityContentBlueprint: ExtensionBlueprint<{
...
+      }
+    | {
+        defaultFilter?: string | ((entity: Entity) => boolean) | undefined;
+        title?:
+          | ReactNode
+          | {
+              actions: ReactNode[];
+            };
+        subtitle?: ReactNode;
+      };
```
Rugvip, line 344: Oh this is an interesting design, will have to ponder it a bit. It certainly slims down the API surface a bit and reduces the complexity of the tree, but on the other hand it limits the ability to provide extensions from plugins that add elements to the header.

camilaibs: Let's discuss it 🙂

.changeset/nervous-cups-happen.md
camilaibs: Not relevant anymore.

.changeset/nervous-cups-happen.md
camilaibs: Not relevant anymore.

Rugvip (approved):
Let's :shipit: 👍 

Thinking we may make the header itself an extension provided by the catalog directly in the future, but that doesn't change this API much other than potentially removing the blueprint.

Issue #23303, opened 2024-02-28
🐛 Bug Report: AuthorizedSearchEngine fails with Payload too Large


drodil:
### 📜 Description

When searching with a bigger `pageLimit` than 25, the AuthorizedSearchEngine fails to call apply conditions for all entities.

### 👍 Expected behavior

The permission evaluation should not fail even with a bigger `pageLimit`

### 👎 Actual Behavior with Screenshots

Response from `/api/catalog/.well-known/backstage/permissions/apply-conditions`

```
Request failed with status 500 Unexpected response from plugin upstream when applying conditions. Expected 200 but got 400 - Bad Request
Request failed with status 500 There was a problem performing the search query: Request failed with 500 Error
```

Client side failure reason: `HTTP response 400 - Bad Request`
Server side failure reason: `Exception PayloadTooLargeError`

### 👟 Reproduction steps

1. Enable permissions
2. Modify the `pageLimit` for search queries to `100`
3. Try to search for catalog items

### 📃 Provide the context for the Bug.

This fix changed the behavior so that the `pageLimit` can be defined from the query (like with default search engine): https://github.com/backstage/backstage/commit/3ed0eac5b4fc6e48f50395d61adc4a6a88423f21

### 🖥️ Your Environment

OS:   Darwin 23.3.0 - darwin/arm64
node: v18.15.0
yarn: 1.22.19
cli:  0.25.2 (installed)
backstage:  1.23.3

Dependencies:
  @backstage/app-defaults                                            1.5.0
  @backstage/backend-app-api                                         0.5.13
  @backstage/backend-common                                          0.17.0, 0.19.9, 0.20.1, 0.21.2
  @backstage/backend-dev-utils                                       0.1.4
  @backstage/backend-openapi-utils                                   0.1.5
  @backstage/backend-plugin-api                                      0.6.12
  @backstage/backend-tasks                                           0.5.17
  @backstage/backend-test-utils                                      0.3.2
  @backstage/catalog-client                                          1.6.0
  @backstage/catalog-model                                           1.4.4
  @backstage/cli-common                                              0.1.13
  @backstage/cli-node                                                0.2.3
  @backstage/cli                                                     0.25.2
  @backstage/config-loader                                           1.6.2
  @backstage/config                                                  1.1.1
  @backstage/core-app-api                                            1.12.0
  @backstage/core-compat-api                                         0.2.0
  @backstage/core-components                                         0.12.5, 0.13.10, 0.14.0
[...]

Issue #20188, opened 2023-09-27
🐛 Bug Report: 413 when too many entities are assigned


henrysachs:
### 📜 Description

When a user is owner of too many entities (in my case groups) the corresponding JWT Token is growing too big and the browser is rejecting these Requests with a 413. 

### 👍 Expected behavior

TBH I'm not quite sure what I'm expecting. Maybe some Docs on how to increase the maximum header size because Backstage itself can't mitigate this.

### 👎 Actual Behavior with Screenshots

Chrome DevTools show that each request is rejected with a 413 (Request headers too large)  Will add screenshots after creating of the issue

### 👟 Reproduction steps

I had a custom gitlab instance and used the [gpt](https://gitlab.com/gitlab-org/quality/performance) from GitLab to create 1000 Groups. What I didn't think about was that my user is now owner of all of them and the [GitLab Discovery](https://backstage.io/docs/integrations/gitlab/discovery/) is creating Backstage groups for all of them. Probably someone don't need GitLab to reproduce this and could just create 1000 Group Entities via a `Groups.yaml`. 

### 📃 Provide the context for the Bug.

I wasn't able to use backstage anymore as any request was failing

### 🖥️ Your Environment

```
OS:   Darwin 22.6.0 - darwin/arm64
node: v18.18.0
yarn: 3.6.1
cli:  0.22.13 (installed)
backstage:  1.18.1

Dependencies:
  @backstage/app-defaults                                          1.4.3
  @backstage/backend-app-api                                       0.5.3
  @backstage/backend-common                                        0.19.5
  @backstage/backend-defaults                                      0.2.3
  @backstage/backend-dev-utils                                     0.1.1
  @backstage/backend-openapi-utils                                 0.0.4
  @backstage/backend-plugin-api                                    0.6.3
  @backstage/backend-tasks                                         0.5.8
  @backstage/catalog-client                                        1.4.4
  @backstage/catalog-model                                         1.4.2
  @backstage/cli-common                                            0.1.12
  @backstage/cli-node                                              0.1.4
  @backstage/cli                                                   0.22.13
  @backstage/config-loader                                         1.5.0
  @backstage/config                                                1.1.0
  @backstage/core-app-api                                          1.10.0
  @backstage/core-components                                       0.13.5
  @backstage/core-plugin-api                                       1.6.0
  @backstage/dev-utils                                             1.0.21
  @backstage/errors                                                1.2.2
  @backstage/eslint-plugin                                         0.1.3
[...]

PR #28853, opened 2025-02-14
fix(permissions): split condition requests into chunks if necessary

drodil:

this should fix too large payloads being sent to the apply-conditions endpoint in the permission framework leading to 413 payload too large issues.

related to #23303 #28836 #20188 #28837

### DIFF
--- a/plugins/permission-backend/src/service/PermissionIntegrationClient.ts
+++ b/plugins/permission-backend/src/service/PermissionIntegrationClient.ts
@@ -41,6 +41,8 @@ const responseSchema = z.object({
   ),
 });
 
+const MAX_ITEMS_PER_REQUEST = 50;
+
 export type ResourcePolicyDecision = ConditionalPolicyDecision & {
   resourceRef: string;
 };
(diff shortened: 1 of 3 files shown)


plugins/permission-backend/src/service/PermissionIntegrationClient.ts
```diff
@@ -71,30 +73,44 @@ export class PermissionIntegrationClient {
...
+
+      if (!response.ok) {
+        throw await ResponseError.fromResponse(response);
+      }
+
+      return responseSchema.parse(await response.json());
+    });
 
-    return result.items;
+    const results = await Promise.all(requests);
```
vinzscam, line 113: an alternative solution could be to group resourceRefs together so that we don't send duplicates of the same permission. WDYT?

vinzscam (changes requested):

plugins/permission-backend/src/service/PermissionIntegrationClient.ts
drodil: Do you mean by `resourceType`? Wouldn't that still make it possible for the request too large?

Grouping by `resourceRef` on the other hand may introduce way too many requests?

vinzscam: so instead of sending:

```js
[
  {
    permission: {
      name: 'catalog.entity.read',
      attributes: {
        action: 'read',
      },
      resourceType: 'catalog-entity',
    },
    resourceRef: 'ref-1'
  },
  {
    permission: {
      name: 'catalog.entity.read',
      attributes: {
        action: 'read',
      },
      resourceType: 'catalog-entity',
    },
    resourceRef: 'ref-2'
  },
  {
    permission: {
      name: 'catalog.entity.read',
      attributes: {
        action: 'read',
      },
      resourceType: 'catalog-entity',
    },
    resourceRef: 'ref-3'
  },
  {
    permission: {
      name: 'catalog.entity.read',
      attributes: {
        action: 'read',
      },
      resourceType: 'catalog-entity',
    },
    resourceRef: 'ref-4'
  },
  {
    permission: {
      name: 'catalog.entity.read',
      attributes: {
        action: 'read',
      },
      resourceType: 'catalog-entity',
    },
    resourceRef: 'ref-5'
  }        
]
```
  
we send:

```js
[
  {
    permission: {
      name: 'catalog.entity.read',
      attributes: {
        action: 'read',
      },
      resourceType: 'catalog-entity',
    },
    resourceRefs: ['ref-1', 'ref-2', 'ref-3', 'ref-4', 'ref-5']
  }
]
```

drodil: Would make more sense, true, but doesn't this require also support to the receiving end to handle the `resourceRefs` instead `resourceRef`?

vinzscam: yyyep 😅

drodil: Hm. That might not work unless a complete refactoring of the handling; the ID is sent from the client for each of these permission requests and that is used to map each response. Identifying multiple results from the `apply-conditions` endpoint is not possible otherwise. One option would be to remove also the ID and use the `resourceRef` for mapping. But anyways, this approach requires a lot more coding that I probably won't have time to do (at least now).

vinzscam: mmm I don't think it should be a problem mapping the id. It's just a value that is sent back by the permission backend

skeletonz28:
hi, is there any feedback on this @drodil @vinzscam ?
We have also run into this issue, and one of our devs has had to hotwire an extension to set a new limit 
This is in regards to the permissions-node plugin, which is referenced on this PR

drodil:
There was a discussion here https://discord.com/channels/687207715902193673/1371763741556867153 about having a dataloader in the `PermissionClient` which would actually fix this too if the correct `batchSize` is used. But no idea if there's going to be any work on that.

drodil:
Additionally, there is this config flag that can help with the issue: `permission.EXPERIMENTAL_enableBatchedRequests`

drodil:
@skeletonz28 this should help: https://github.com/backstage/backstage/pull/29983

drodil:
@skeletonz28, so in which request are you getting 413? Just wondering if it's permission related at all 🤔

PR #28881, opened 2025-02-17
chore: migrate to yarn v4

secustor:

Replaces yarn v3 with v4 and add volta helper

Upgrading as v3 fails to dedupe https://github.com/backstage/backstage/pull/28880 , but v4 succeeds

Is blocked by an PR which adds the new options `yarn-install` https://github.com/backstage/actions/pull/163


.yarnrc.yml
```diff
@@ -1,13 +1,9 @@
...
 
 nodeLinker: node-modules
 
 npmRegistryServer: 'https://registry.npmjs.org/'
-
-plugins:
-  - path: .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
-    spec: '@yarnpkg/plugin-workspace-tools'
-
-yarnPath: .yarn/releases/yarn-3.8.1.cjs
```
Sarabadu, line 13: FYI, You can still use path with yarn 4. In my case using only corepack caused some issues on CI.

Sarabadu: looking on the failing CI, some install gets changes...
have in minnd you need to run a full yarn install, without `-mode=update-lockfile`

Sarabadu: just dropping some  issues i encounter on the community.plugins to have in mind

secustor: Thanks for the input. 

> FYI, You can still use path with yarn 4. In my case using only corepack caused some issues on CI.

This should be addressed with my PR to the `yarn-install` action from `backstage/actions`

> looking on the failing CI, some install gets changes...
have in minnd you need to run a full yarn install, without -mode=update-lockfile

Yes this is why the lockfile is regenerated mostly. 

> yarn --cwd  stopped working on yarn 4 too.. so instead of yarn --cwd  ./path command now need to run yarn ./path command

Yes, addressed with using [`working-directory` ](https://github.com/backstage/backstage/pull/28881/files#diff-e0f03fe606d54c0a66c7afc5ba8c9452e10023b66772486085510fdbd0403513R36)

package.json
```diff
@@ -147,7 +150,11 @@
     "typedoc": "^0.27.6",
     "typescript": "~5.2.0"
   },
-  "packageManager": "yarn@3.8.1",
+  "packageManager": "yarn@4.6.0",
+  "volta": {
+    "node": "22.14.0",
+    "yarn": "4.6.0"
+  },
```
Rugvip, line 157: What's the benefit of Volta?

.yarnrc.yml
Rugvip: I'd prefer to keep using `yarnPath`, less headaches that way

Rugvip (commented):
Thank you! 👍 

Two initial bits for direction

.yarnrc.yml
secustor: You mean in this PR or generally? 

In my experience it having the code checked in rather creates headache in the upgrade process. 

[Yarn itself seems to agree. ](https://yarnpkg.com/blog/release/4.0#installing-yarn)

> To that end we used to recommend using the [yarnPath](https://yarnpkg.com/configuration/yarnrc#yarnPath) setting pointing to a checked-in binary, but this pattern increased friction more than we liked - many people didn't like the idea of adding a binary to their repository, however small. We listened, and worked conjointely with Node.js on a project called [Corepack](https://nodejs.org/api/corepack.html). Corepack is a tool shipped with Node.js 16+ that will automatically select the right package manager version to run depending on the project you're working on.

package.json
secustor: Interjects calls to the tools and links them to the specified version. 
The result is that devs can switch between projects / branches and always have the correct version. 

If a dev has not the Volta shim installed this is noop, but I would recommend it as that way it is easier in case of behaviour differences between versions.

secustor:
still valid

Issue #28786, opened 2025-02-10
🐛 Bug Report: SimpleStepper backStep action fails together with a higher initial activeStep


lexGPT:
### 📜 Description

When I use the SimpleStepper Component with a initial activeStep higher than 0 (example 3) and use the backAction button it fails to go to the previous action and brings the whole stepper in an unusable state. 

### 👍 Expected behavior

When I have an initial activeStep higher than 0 I can still use the back button to go to previous steps. 

### 👎 Actual Behavior with Screenshots

![Image](https://github.com/user-attachments/assets/509c179b-3816-4a70-8c36-d840ce12bd81)

![Image](https://github.com/user-attachments/assets/b3a7b0d9-74c2-4c6d-871c-b82fc4d46dd9)

![Image](https://github.com/user-attachments/assets/2adb07f9-68b8-4267-94d0-4ed421fc711c)

### 👟 Reproduction steps

1. Go to https://backstage.io/storybook/?path=/story/navigation-simplestepper--default&args=activeStep:2. 
2. Make sure activeStep is set to 2 and reload. 
3. Click on Back
4. First step starts showing instead of the second
5. Click on Next
6. Every step becomes disabled and stepper is unusable

### 📃 Provide the context for the Bug.

I'm trying to build a wizard to guide users through our backstage instance in a home plugin component. I want to make the active step persistant between page visits through the storage api but this makes it very easy for users to brick their wizard. 

### 🖥️ Your Environment

yarn run v1.22.22
$ /Users/lfelix/Projects/Devex/devex-it-explorer/node_modules/.bin/backstage-cli info
OS:   Darwin 24.3.0 - darwin/arm64
node: v20.18.1
yarn: 1.22.22
cli:  0.29.5 (installed)
backstage:  1.34.0

Dependencies:
  @backstage/app-defaults                                          1.5.16
  @backstage/backend-app-api                                       1.1.1
  @backstage/backend-common                                        0.23.3, 0.24.1, 0.25.0
  @backstage/backend-defaults                                      0.6.2, 0.7.0
  @backstage/backend-dev-utils                                     0.1.5
  @backstage/backend-openapi-utils                                 0.4.1
  @backstage/backend-plugin-api                                    0.6.21, 0.7.0, 0.8.1, 1.1.1
  @backstage/backend-tasks                                         0.6.1
  @backstage/backend-test-utils                                    1.2.1
  @backstage/catalog-client                                        1.9.1
[...]

PR #28887, opened 2025-02-18
fix(core-components): SimpleStepper back button works with activeStep set higher than 0

lexGPT:

Fixes this issue. https://github.com/backstage/backstage/issues/28786

The back button functionality looks at the stepHistory array and uses the pop functionality to determine the previous step it was on. It was always populated by a single 0 entry. This meant that if you don't start at the first step it empties the array and cannot determine the correct previous step anymore. 

This step fixes this issue by populating the stepHistory array with integers in numerical order stopping at the activeStep given. This makes the functionality work for steppers where the flow is in numerical order. It won't take into account if certain steps are skipped. If that is a problem maybe we can opt to add an override options to the stepper so users can figure out the correct initial step history array out for their own use case.

### DIFF
--- a/.changeset/beige-chairs-suffer.md
+++ b/.changeset/beige-chairs-suffer.md
@@ -2,4 +2,4 @@
 '@backstage/core-components': minor
 ---
 
-SimpleStepper back button now works with activeStep property set higher than 0
+`SimpleStepper` back button now works with `activeStep` property set higher than 0
--- a/.changeset/beige-chairs-suffer.md
+++ b/.changeset/beige-chairs-suffer.md
@@ -1,0 +1,5 @@
+---
+'@backstage/core-components': minor
+---
+
+SimpleStepper back button now works with activeStep property set higher than 0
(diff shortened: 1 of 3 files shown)


lexGPT:
@benjdlambert, @vinzscam. I believe I did all I needed to do before it can be reviewed. Could you confirm this and maybe also an estimation when it can be reviewed and hopefully merged in?

.changeset/beige-chairs-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/core-components': minor
+---
+
+SimpleStepper back button now works with activeStep property set higher than 0
```
benjdlambert, line 5:
```suggestion
`SimpleStepper` back button now works with `activeStep` property set higher than 0
```

benjdlambert (changes requested):
Just some small bits in the changeset to make the spell check bot happy

freben (approved):
Perfect!

benjdlambert (approved):

PR #28892, opened 2025-02-18
feat(auth): enable persistency of scopes for Bitbucket Cloud

benjidotsh:
This PR enables persistency of scopes for Bitbucket Cloud, preventing the OAuth request dialog from popping up every time you want to do something with Bitbucket Cloud.

I wasn't sure whether this was a breaking change since the scaffolder complains about scopes when autocompleting projects and creating repositories if you're missing scopes in `@backstage/integration-react`.

I'm also not 100% sure if I understand how the scopes work: why does it only complain about the scopes as soon as the token gets refreshed (I guess?) and not when initially authenticating? Are the scopes only used for refreshes?

### DIFF
--- a/packages/integration-react/src/api/ScmAuth.ts
+++ b/packages/integration-react/src/api/ScmAuth.ts
@@ -239,11 +239,13 @@ export class ScmAuth implements ScmAuthApi {
       'pullrequest',
       'snippet',
       'issue',
+      'project',
     ];
     const repoWriteScopes = options?.scopeMapping?.repoWrite ?? [
       'pullrequest:write',
       'snippet:write',
       'issue:write',
+      'repository:admin',
     ];
     return new ScmAuth('bitbucket', bitbucketAuthApi, host, {
       default: defaultScopes,
--- a/packages/integration-react/src/api/ScmAuth.ts
+++ b/packages/integration-react/src/api/ScmAuth.ts
@@ -245,7 +245,6 @@ export class ScmAuth implements ScmAuthApi {
       'pullrequest:write',
       'snippet:write',
       'issue:write',
-      'repository:admin',
     ];
     return new ScmAuth('bitbucket', bitbucketAuthApi, host, {
       default: defaultScopes,
--- a/.changeset/chatty-days-wonder.md
+++ b/.changeset/chatty-days-wonder.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-auth-backend-module-bitbucket-provider': patch
+---
+
+Enabled persistency of scopes for Bitbucket Cloud.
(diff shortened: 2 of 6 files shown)


benjidotsh:
Tests still seem to be failing on the `techdocs-module-addons-contrib` plugin 🤔

packages/integration-react/src/api/ScmAuth.ts
```diff
@@ -239,11 +239,13 @@ export class ScmAuth implements ScmAuthApi {
       'pullrequest',
       'snippet',
       'issue',
+      'project',
     ];
     const repoWriteScopes = options?.scopeMapping?.repoWrite ?? [
       'pullrequest:write',
       'snippet:write',
       'issue:write',
+      'repository:admin',
```
Rugvip, line 248: I think this is a bit excessive for base write scopes, probably best to keep this as an explicitly requested additional scope instead or local override in your setup

Rugvip (commented):
Looks alright as such, with the comment about the new scopes

I'm wondering if this is really needed though, does Bitbucket cloud not persist and track the existing scopes on their side? Most OAuth providers do, it's mostly just GitHub being a bit strange. If you keep needing to log in it might also be something that's up with your configuration, it's certainly not intended anyway. Curious if any other Bitbucket users have any opinions here (@pjungermann 👀 😁)

packages/integration-react/src/api/ScmAuth.ts
benjidotsh: I see where you're coming from, but it's required for certain actions like the one for creating a repository, so that would require the user to provide additional scopes for built-in functionality. Is that something we want to do? `repoWriteScopes` is only being used for the scaffolder anyways, so maybe we should simply rename it to `scaffolderScopes`.

pjungermann: Yeah, `repository:admin` might be a bit excessive as one of the default scopes. Not everyone has this permission for all repositories, and for many operations, it might not be needed at many operations.

At the `RepoUrlPicker` in software templates, we use
```yaml
            requestUserCredentials:
              secretsKey: USER_OAUTH_TOKEN
              additionalScopes:
                bitbucket:
                  - pipeline:write
                  - repository:admin
```
We get asked to log in there. I've tested it, we get asked each time. Nothing is "stored".

`repository:admin` is required for creating the repository using `publish:bitbucketCloud`.

We request `pipeline:write` additionally, as we have another action that modifies configuration via a custom scaffolder action.

benjidotsh: > Not everyone has this permission for all repositories, and for many operations, it might not be needed at many operations.

Good point! I've removed the scope.

pjungermann (approved):

benjidotsh:
Not stale 😄 @vinzscam @Rugvip can you guys take a final look please?

PR #28894, opened 2025-02-18
Made gitlab:repo:push action idempotent. 

acierto:

Made gitlab:repo:push action idempotent. 


plugins/scaffolder-backend-module-gitlab/src/actions/gitlabRepoPush.ts
```diff
@@ -181,15 +186,22 @@ export const createGitlabRepoPushAction = (options: {
...
+        const commitId = await ctx.checkpoint({
+          key: `commit.create.${repoID}.${branchName}`,
+          fn: async () => {
+            const commit = await api.Commits.create(
+              repoID,
+              branchName,
+              ctx.input.commitMessage,
+              actions,
+            );
+            return commit.id;
```
benjdlambert, line 198: I'm wondering if we should actually try to encourage returning the entire `create` call here rather than just ID. 

It kind of limits what you can do with the action moving forward. If I was to change this action to also output something else form the `api.Comits.create` method, then retries would instantly become broken right?

benjdlambert: To clarify, I think it's much better to do 

```ts
const { id: commitId } = await ctx.checkpoint({
    key: `commit.create.${repoID}.${branchName}`,
    fn: async () => {
      return api.Commits.create(
        repoID,
        branchName,
        ctx.input.commitMessage,
        actions,
      );
    }
})
```

acierto: That won't work as the result of `Commits.create` is not JsonValue 😅 

I'm not really sure why it should be broken. 
As a result of checkpoint belongs to a specific task.

This can happen only if user has some on-going tasks and decides to upgrade Backstage. 
But I think we decided to skip it for now, as we can also solve it by versioning.

Also by keeping the entire object can be space inefficient.

benjdlambert: Add to docs on writing checkpoints that when you change the return type of the checkpoint itself you should also change the ID to cache bust?

PR #28930, opened 2025-02-20
catalog-react: rename entity card "area" to "type"

Rugvip:

Followup for https://github.com/backstage/backstage/pull/28758#discussion_r1961776060

It's possible to work around the `variant` conflict but it would be better to use a different naming. In searching for options I realized that where we landed with the semantics of this a card "type" kinda makes sense. It even mirrors the catalog model a bit where `entity-card` is the kind of extension while it then has different "types" that are unique to the kind. Can't remember why we didn't go with this from the beginning tbh, but figured we should try it 😁 

Piggybacking on existing changeset

### DIFF
--- a/.changeset/ninety-teachers-tease.md
+++ b/.changeset/ninety-teachers-tease.md
@@ -7,7 +7,7 @@ Introduces a new `EntityContentLayoutBlueprint` that creates custom entity conte
 The layout components receive card elements and can render them as they see fit. Cards is an array of objects with the following properties:
 
 - element: `JSx.Element`;
-- area: `"peek" | "info" | "full" | undefined`;
+- type: `"peek" | "info" | "full" | undefined`;
 
 ### Usage example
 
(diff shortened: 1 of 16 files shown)


freben:
I'm honestly a little torn about this one.

> expected size and where it will be rendered

The description mixes concerns a bit, but really my thought is mostly that "area" matches the CSS Grid mindset and size doesn't really come into play from the card's POV. A custom entity page layout might just as well have named their areas "left" and "right" for example, it's just labels. Hmmm. "peek" is a bit odd too, "main" and "aside" would have been two better names in my mind. I dunno, so much 🚲 🏠

MihaiTabara (approved):

Rugvip:
@freben that's actually exactly the reason I want this rename, because I don't think it should be similar to a CSS grid. Only defining the area doesn't quite give enough direction for how the cards should be implemented imo, and it makes it very hard to create a layout for the entity pages that is both functional and visually appealing.

That said I wouldn't be surprised if much of this doesn't quite work out in practice, but from a design PoV I think we should flip around our approach to start with a very restrictive system where we then relax certain bits, rather than starting with a very flexible system and then finding things we need to restrict.

benjdlambert (approved):

Issue #28905, opened 2025-02-19
🐛 Bug Report: 1.36.0 Auditor logging `[object Object]` and is way too verbose


drodil:
### 📜 Description

The new auditor service in 1.36.0 has issue with logging like follows:

```
[0] 2025-02-19T07:03:34.790Z catalog info catalog.entity-facets isAuditEvent=true eventId=entity-facets severityLevel=low actor=[object Object] request=[object Object] meta=undefined status=initiated
```
Additionally the auditor service is way too verbose; for example fetching entity facets log each request. 

### 👍 Expected behavior

The audit log should contain proper information about the event. Having `[object Object]` brings no value.

IMHO the audit log should only contain actions done by the user tor example registering/deleting entity, starting scaffolder task and so on - not every data fetch.

### 👎 Actual Behavior with Screenshots

See above.

### 👟 Reproduction steps

1. Update to backstage 1.36.0
2. Go to catalog page
3. Check auditor logs

### 👀 Have you spent some time to check if this bug has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

Rugvip:
I suggest we log read events work a servering level of low or medium, and map those to debug log level by default.

drodil:
Good idea @Rugvip 👍

Rugvip:
Is that something you want to work on @schultzp2020? otherwise I can pick it up

Rugvip:
@schultzp2020 sorry going with #28934 because we need to ship this as a patch so it can't be a minor bump. Let's discuss #28932 and what configuration to add next week.

PR #28932, opened 2025-02-20
feat: add `backend.auditor.severityLogLevelMappings` to map severity levels to log levels

schultzp2020:

Adds `backend.auditor.severityLogLevelMappings` to map severity levels to log levels.

references https://github.com/backstage/backstage/issues/28905

### DIFF
--- a/.changeset/rotten-windows-fly.md
+++ b/.changeset/rotten-windows-fly.md
@@ -1,0 +1,5 @@
+---
+'@backstage/backend-defaults': patch
+---
+
+Adds `backend.auditor.severityLogLevelMappings` to map severity levels to log levels.
--- a/packages/backend-defaults/config.d.ts
+++ b/packages/backend-defaults/config.d.ts
@@ -83,6 +83,32 @@ export interface Config {
         };
 
     /**
+     * Options used by the default auditor service.
+     */
+    auditor?: {
+      /**
+       * Defines how audit event severity levels are mapped to log levels.
+       * This allows you to control the verbosity of audit logs based on the
+       * severity of the event. For example, you might want to log 'low' severity
+       * events as 'debug' messages, while logging 'critical' events as 'error'
+       * messages. Each severity level ('low', 'medium', 'high', 'critical')
+       * can be mapped to one of the standard log levels ('debug', 'info', 'warn', 'error').
+       *
+       * By default, audit events are mapped to log levels as follows:
+       * - `low`: `debug`
+       * - `medium`: `info`
+       * - `high`: `info`
+       * - `critical`: `info`
+       */
+      severityLogLevelMappings?: {
+        low?: 'debug' | 'info' | 'warn' | 'error';
+        medium?: 'debug' | 'info' | 'warn' | 'error';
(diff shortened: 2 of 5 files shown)


packages/backend-defaults/config.d.ts
```diff
@@ -82,6 +82,16 @@ export interface Config {
...
         };
 
+    /**
+     * Options used by the default auditor service.
+     */
+    auditor?: {
+      /**
+       * The lowest log level that will be recorded. Defaults to "medium".
+       */
+      logLevel?: 'low' | 'medium' | 'high' | 'critical';
```
Parsifal-M, line 92: Just my two cents I think there's two things to consider which are verbosity and severity (log level)

Maybe you do want everything logged, but you just want it to be on a lower verbosity.

Wondering how we could maybe tackle that 🤔 💭

Rugvip: Hmm, yep I agree, and I think taking that even further I don't think we want a log level config for the default auditor at all. All we need is really config for a log level mapping right? and then we leave the actual log level filtering to the logger service. For example something like this:

```
backend:
  auditor:
    severityLogLevels:
      low: info
      critical: error
```

The above config would leave the log level for medium and high as they are by default, but set the log level of low severity to info, and critical severity to error.

What do you think @schultzp2020?

.changeset/rotten-windows-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-defaults': minor
```
Rugvip, line 2:
```suggestion
'@backstage/backend-defaults': patch
```

We generally stick to patch bumps for non-breaking changes on 0.x packages

packages/backend-defaults/src/entrypoints/auditor/auditorServiceFactory.ts
```diff
@@ -32,21 +35,40 @@ import { DefaultAuditorService } from './DefaultAuditorService';
...
+        'info',
+      high:
+        auditorConfig?.getOptionalString('severityLogLevelMappings.high') ??
+        'info',
+      critical:
+        auditorConfig?.getOptionalString('severityLogLevelMappings.critical') ??
+        'info',
+    } as Record<
+      AuditorServiceEventSeverityLevel,
+      'debug' | 'info' | 'warn' | 'error'
```
Rugvip, line 63: Hmm, in this case it might actually be good to verify that this is indeed what we're getting from config, since we'll be trying to call `undefined` otherwise

Rugvip (commented):
Looks good! 👍  two small bits

packages/backend-defaults/src/entrypoints/auditor/auditorServiceFactory.ts
Rugvip: I'm thinking about this case:

```yaml
backend:
  auditor:
    severityLogLevelMappings:
      low: definitelynotaloglevel
```

Rugvip (approved):
Thanks! 👍 

Let's :shipit:

PR #28942, opened 2025-02-22
RFC add external token decorator service

Sarabadu:

Currently the Default auth service provides a `pluginTokenHandlerDecoratorServiceRef`  to add custom logic when issueing and validation tokens.

This Pr add the similar functionality to the external tokens handlers this is mostly useful for the jwks tokens as the possible validations of the token can be not sufficient for some token issuers.

## Changes

This PR introduces a new `externalTokenTypeHandlersRef` service  as multiton that enables custom external token validation logic alongside the existing authentication system.

- `createExternalTokenHandler()` simplifies creating type-safe custom token handlers  
- `ExternalTokenHandler` is now a unique interface for all the default external token handlers, and also exported for use in the extension

### Current implementation
The current implementation uses a multiton service allowing user to add custom handlers like this:

```ts
const customHandler = createExternalTokenHandler({
  type: 'custom',
  initialize({ options }) {
    return { apiKey: options.getString('apiKey') };
  },
  async verifyToken(token, context) {
    // Custom validation logic
    return { subject: 'custom:user' };
  },
});

const customExternalTokenHandlers = createServiceFactory({
  service: externalTokenTypeHandlersRef,
  deps: {},
  async factory() {
    return customHandler // this is a multiton so this implementation will be added to all the handlers array 
  },
});

```

### Use case 

Using Entra Id app registration is possible to create a token for an specific app inside your tenant, but validation the issuer and the audience is not always sufficient, and validation of custom token fields as `roles` could be necessary to ensure the token have the rigth privileges.

**BREAKING**: The deprecated backend.auth.keys configuration has been removed. Users must migrate to the backend.auth.externalAccess format. Migration guide included in documentation.

### Alternatives

We could split the `ExternalTokenHandler` between static and  `JWKSHandler` and decorate the logic each type


Sarabadu:
Im changing this as ready for review to bring some discussion around it

Im still not 100%  on this implementation
- pro: it follows the same logic and shape as the plugin handler decorator
- con: this implementation decorates, legacy, static and jwsk tokens in the same place but most probably the logic aplies to one type
- con: for checking some custom claims, a new decoded is needed as the decoded token is not exposed.

as example this is the current implementation to validate azure `roles` claims inside the token
```ts
  const decoratedPluginTokenHandler = createServiceFactory({
  service: externalTokenHandlerDecoratorServiceRef,
  deps: {},
  async factory() {
    return (defaultImplementation: ExternalTokenHandler) => {
      return new (class CustomHandler implements ExternalTokenHandler {
        async verifyToken(token: string): Promise<
          | {
              subject: string;
              accessRestrictions?: BackstagePrincipalAccessRestrictions;
            }
          | undefined
        > {
          const result = await defaultImplementation.verifyToken(token); 
          const decoded = decodeJwt<JWTPayload & { roles?: string[] }>(token); 

          if (!result?.subject.startsWith('external:some-prefix')) { // the prefix seems to be the only value available to filter which tokens need this validation (without re-implementing the default implementation)
            return result;
          }

          if (decoded.roles?.includes('some-app-role')) {
            return result;
          }

          throw new AuthenticationError('Illegal token missing required role');
        }
      })();
    };
  },
});

```

benjdlambert:
Shall we maybe bring this up in the Framework SIG?

Rugvip:
Re multiton services: https://github.com/backstage/backstage/blob/45e7b21f665f9095691886d402cf8d4d90c3b3bf/packages/backend-defaults/src/entrypoints/urlReader/urlReaderServiceFactory.ts#L41-L45

Sarabadu:
Still working on it in

Sarabadu:
not stale

packages/backend-defaults/src/entrypoints/auth/external/types.ts
```diff
@@ -17,17 +17,25 @@
...
 >;
 
+/**
+ * @public
+ * This interface is used to handle external tokens.
+ * It is used by the auth service to verify tokens and extract the subject.
+ */
 export interface TokenHandler {
-  add(options: Config): void;
+  add?(options: Config): TokenHandler;
```
Rugvip, line 34: Let's split into internal/public so that we can remove `add` for the public one

packages/backend-defaults/src/entrypoints/auth/authServiceFactory.test.ts
```diff
@@ -450,8 +455,113 @@ describe('authServiceFactory', () => {
...
+
+      const customPluginTokenHandler = createServiceFactory({
+        service: externalTokenHandlersServiceRef,
+        deps: {},
+        async factory() {
+          return {
+            custom: (options: Config) => {
+              customConfig(options);
+              return customHandler.add(options);
+            },
```
Rugvip, line 534:
```suggestion
          return {
            type: 'custom',
            factory: (options: Config) => {
              customConfig(options);
              return customHandler.add(options);
            },
```

another pattern to consider

Sarabadu:
Hey @Rugvip, I pivot a bit on this one

#### What I've changed:
- The handlers configurations have now a the [{type, factory} shape](https://github.com/backstage/backstage/pull/28942/files#diff-9adc03d6c03035f2776d2a7b2aa3cd7109153aa1ed1c637ce98537bf1ee0b9daR48-R55) as we discussed previously
- The factory gets now an Array of Config object, so there is one Factory execution for each tokenHandler "type"
- I also did a quite awful wrapper for the Legacy old config (Not sure for how long was the deprecation there to consider removing that support)

#### The current logic for adding new handlers would look like:

```ts
      const customPluginTokenHandler = createServiceFactory({
        service: externalTokenTypeHandlersRef,
        deps: {},
        async factory() {
          return {
            type: 'custom',
            factory: (configs: Config[]) => { // factory gets all config for this type
              return new CustomHandler(configs); // or configs.map(c => CustomHandler(c) it can return multiple instances or single instance handling many configs 
            },
          };
        },
      });
```


I've pending update the documentation with this new changes and add more test probably, but would love to get some feedback in the meantime

Sarabadu:
@Rugvip can we re-open this one? I couldn't work for a while but slowly getting back. will try to rebase and see why the CI is not happy

packages/backend-defaults/src/entrypoints/auth/external/ExternalTokenHandler.ts
```diff
@@ -16,19 +16,63 @@
...
+};
+
+/**
+ * @public
+ * This service is used to decorate the default plugin token handler with custom logic.
+ */
+export const externalTokenTypeHandlersRef = createServiceRef<TokenTypeHandler>({
+  id: 'core.auth.externalTokenHandlers',
+  multiton: true,
+  // defaultFactory // :pepe-think: seems like is not possible to use defaultFactory with multiton
```
Sarabadu: 🤔  seems like is not possible to use defaultFactory with multiton, should we look on this? maybe in. a separated PR anyways

Sarabadu:
Not stale

Sarabadu:
Don't stale me now 🎵
Don't stale me now 🎶

packages/backend-defaults/src/entrypoints/auth/plugin/PluginTokenHandler.ts
```diff
@@ -46,7 +46,7 @@ type Options = {
 
 /**
  * @public
- * Issues and verifies {@link https://backstage.iceio/docs/auth/service-to-service-auth | service-to-service tokens}.
+ * Issues and verifies {@link https://backstage.io/docs/auth/service-to-service-auth | service-to-service tokens}.
```
freben, line 49: 🍦

packages/backend-defaults/src/entrypoints/auth/external/types.ts
```diff
@@ -15,20 +15,41 @@
...
+ * This interface is used to handle external tokens.
+ */
+export interface TokenTypeHandler {
+  type: string;
+  /**
+   * A factory function that takes all token configuration for a given type
+   * and returns a TokenHandler or an array of TokenHandlers.
+   */
+  factory: (
+    configs: import('@backstage/config').Config[],
```
freben: What's up with these dynamic imports? We rarely do that, would prefer regular import at the top

packages/backend-defaults/report-auth.api.md
```diff
@@ -44,5 +59,24 @@ export const pluginTokenHandlerDecoratorServiceRef: ServiceRef<
   'singleton'
 >;
 
+// @public
+export interface TokenHandler {
```
freben: The main type should probably be called `ExternalTokenHandler`.

packages/backend-defaults/report-auth.api.md
```diff
@@ -44,5 +59,24 @@ export const pluginTokenHandlerDecoratorServiceRef: ServiceRef<
...
+        subject: string;
+        allAccessRestrictions?: AccessRestrictionsMap;
+      }
+    | undefined
+  >;
+}
+
+// @public
+export interface TokenTypeHandler {
+  factory: (configs: Config[]) => TokenHandler | TokenHandler[];
```
freben: Hm. I'm not sure we need this type at all. Could we take inspiration from the `Authenticator` concept [found in the auth backend itself](https://github.com/backstage/backstage/blob/e60b8719863b68da7216f63462d2166188c63eef/plugins/auth-node/src/oauth/types.ts#L86)?

Something along the lines of

```ts
export interface ExternalTokenHandler<TContext> {
  type: string;
  initialize(ctx: { options: Config }): TContext;
  verifyToken(token: string, ctx: TContext): Promise<
    | { subject: string; }
    | undefined
  >;
}
```

and then, we could actually do

```ts
export function createExternalTokenHandler<TContext>(
  handler: ExternalTokenHandler<TContext>
): ExternalTokenHandler<unknown>;
```

To help TypeScript get the types right. This is the function that each type of service calls in its factory method, to help create a handler for a new custom type of external token.

Note how this version of the code does not deal with `accessRestrictions`. The framework already knows how to read those and they are uniform, so the actual handlers here are unaware of their existence. Crucially, this design does _not_ need to export an `AccessRestrictionsMap` at all, so that can be removed above. At a later point we could even deprecate the public version of `readAccessRestrictionsFromConfig`, but that should probably be deferred a bit.

This `initialize` receives only the `options` subkey contents sinceit has no business looking outside of the options object at that point.

This makes each individual handler self contained in a useful way, and the framework has a kind of convenient way of re-calling all `initialize` when config changes, if it wants to be smart about that.

packages/backend-defaults/report-auth.api.md
Sarabadu: naming is hard, we already have a [ExternalTokenHandler](https://github.com/backstage/backstage/pull/28942/files/92f1dc8a37e4d0f8d13fe0b1f0898bf1b96baedb#diff-9adc03d6c03035f2776d2a7b2aa3cd7109153aa1ed1c637ce98537bf1ee0b9daR82) that contains the logic for all the ExternalTokenHandler... but I will try to come with a different naming if it makes more sense.

packages/backend-defaults/report-auth.api.md
Sarabadu: Thanks for review it @freben, I will give this a try. 
Just to check I follow correctly:
in my current implementation I was accepting an array of options because looking on the static handler having some cleverness on [handling](https://github.com/backstage/backstage/pull/28942/files/92f1dc8a37e4d0f8d13fe0b1f0898bf1b96baedb#diff-11000d8211383bbc7758cada29a778d809d6642f17bb7774e7fce28a1b25baa0R66) all the options at once, so thought there is a valid usage of having 1 handler for all options of the same type

But if I understand correctly having each handler self contained, means one handler instance per each config option, right? if we have 3 statics one and 2 legacy I will have 5 handlers instances

I like the idea of having it self contained and to remove the `AccessRestrictionsMap`  from each handles and instead do it in a upper level, I will give it a try refactoring all the internal types to match this too

packages/backend-defaults/report-auth.api.md
freben: Only internally right? The external api is more important after all.

packages/backend-defaults/report-auth.api.md
freben: Only one handler instance per type (the one that each service factory returns). Then the framework makes three contexts (at startup) if there are three `type: custom` blocks. Then it calls the single instance three times (per token), passing in the respective contexts one by one

freben: Regarding that optimization - yeah it tries to not decode the token more than once right? We could skip that optimization if it becomes too complex, and then the implementation can be updated with a little local ExpiryMap that all contexts share that maps token strings to `Promise<decoded type>` for a small amount of time.

Sarabadu: ohh, ok so the instead of 3 handler instances you mean 3 context instances and the framework will have some kind of 
`{type:string, context:any}[]`. We are loosing the "shared" context anyway between block, right? 

On top level ExternalAllTypesHandler creation,  we call for each option block, the type `initialize` 
and then on verify we iterate over the `{type:string, context:any}[]`, and verify each of them

Sarabadu: yes the optimization in the legacy handler avoids to decode the token for each entry, and in the static one, smaller optimization, uses the map.get instead of iterate over all options

Sarabadu: @freben I'm working on this approach.

WDYT about removing the [old configuration](https://github.com/backstage/backstage/pull/28942/files#diff-9adc03d6c03035f2776d2a7b2aa3cd7109153aa1ed1c637ce98537bf1ee0b9daR111) as part of this? 
If I follow correctly has been deprecated for more than a year

Sarabadu:
@freben I did the proposed implementation...

look much more cleaner, but it is less optimised as now all the tokes goes one by one.

I hope this is closer to what you had in mind, I remove the olg configuration key on this one, let me know if is better to keep that

docs/backend-system/architecture/03-services.md
```diff
@@ -223,6 +223,35 @@ export const customFooServiceFactory = createServiceFactory({
 
 This allows you to provide more advanced options for the service implementation that couldn't be expressed through static configuration. It also gives users of the service implementation access to other services through dependency injection, which can be useful for their customizations.
 
+## Multiton
+
+By default the service reference will point to a singleton instance of the service. This mean if a new service factory uses this reference will override the previous one. This is the most common use-case, but in some cases you may want to have multiple instances of the same service.
+For some services, is desirable to extend the functionality instead of overriding it. For example, some services could have many handler to address a specific event, and you may want to add a new handler instead of overriding the previous one. In this case, you can use the `multiton` option when creating the service reference:
```
Parsifal-M:
```suggestion
By default the service reference will point to a singleton instance of the service. This mean if a new service factory uses this reference it will override the previous one. This is the most common use-case, but in some cases you may want to have multiple instances of the same service.
For some services, it is desirable to extend the functionality instead of overriding it. For example, some services could have many handlers to address specific events, and you may want to add a new handler instead of overriding the previous one. In this case, you can use the `multiton` option when creating the service reference:
```

Parsifal-M (commented):
Just tiny bit on the docs, I think though from a docs point of view its looking great!

freben:
Hm there's lots of commented out stuff in here. Can you go through the diff visually and clean it up a bit so that it's in more of a final state?

I'd also not deprecate the old key config in the same PR. It's a big, breaking move that can (should) be done all on its own separate from other changes. I'm not comfortable mixing it in with the rest.

.changeset/thirty-rules-press.md
```diff
@@ -0,0 +1,26 @@
+---
+'@backstage/backend-defaults': minor
+---
+
+Add a new `externalTokenHandlersServiceRef` to allow custom external token validations
+
+BREAKING CHANGE: The `backend.auth.keys` config has been removed. Please migrate to the new `backend.auth.externalAccess` config as described in the documentation: https://backstage.io/docs/auth/service-to-service-auth
```
freben: Just noting that as per the previous comment, I still think we should revert this and do the breakage some other time in a dedicated PR (that's easier to review and revert)

packages/backend-defaults/src/entrypoints/auth/authServiceFactory.test.ts
```diff
@@ -41,7 +44,7 @@ const mockDeps = [
       backend: {
         baseUrl: 'http://localhost',
         auth: {
-          keys: [{ secret: 'abc' }],
+          // keys: [{ secret: 'abc' }],
```
freben: 👀

packages/backend-defaults/src/entrypoints/auth/authServiceFactory.test.ts
```diff
@@ -450,8 +453,88 @@ describe('authServiceFactory', () => {
...
+      const customLogic = jest.fn();
+      const customConfig = jest.fn();
+      const deps = [
+        discoveryServiceFactory,
+        mockServices.rootConfig.factory({
+          data: {
+            backend: {
+              baseUrl: 'http://localhost',
+              auth: {
+                // keys: [{ secret: 'abc' }w],
```
freben: 👀

packages/backend-defaults/src/entrypoints/auth/external/ExternalAuthTokenHandler.ts
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * @public
+ * This service is used to decorate the default plugin token handler with custom logic.
+ */
+export const externalTokenTypeHandlersRef = createServiceRef<
+  ExternalTokenHandler<unknown>
+>({
+  id: 'core.auth.externalTokenHandlers',
+  multiton: true,
+  // defaultFactory // :pepe-think: seems like is not possible to use defaultFactory with multiton
```
freben: 👀 haha, delete

No. But that's fine, you'll just create a default instance if the ref resolves to an empty array. Same thing, basically. This is not an API that we expect anyone from the outside to resolve and use anyway.

packages/backend-defaults/src/entrypoints/auth/external/ExternalAuthTokenHandler.ts
```diff
@@ -0,0 +1,160 @@
...
+import { AccessRestrictionsMap, ExternalTokenHandler } from './types';
+import { readAccessRestrictionsFromConfig } from './helpers';
+
+const NEW_CONFIG_KEY = 'backend.auth.externalAccess';
+const OLD_CONFIG_KEY = 'backend.auth.keys';
+let loggedDeprecationWarning = false;
+
+/**
+ * @public
+ * This service is used to decorate the default plugin token handler with custom logic.
```
freben: Hmm isn't it more like "Custom handlers for external access tokens". Not "plugin" tokens either way

packages/backend-defaults/src/entrypoints/auth/external/ExternalAuthTokenHandler.ts
```diff
@@ -0,0 +1,160 @@
...
+        };
+      },
+    );
+
+    // Load the old keys too
+    const legacyConfigs = config.getOptionalConfigArray(OLD_CONFIG_KEY) ?? [];
+
+    if (legacyConfigs.length && !loggedDeprecationWarning) {
+      loggedDeprecationWarning = true;
+      // :pepe-think: this message was here for more than a year, and replacing this simplifies things a lot
```
freben: 👀 remove

packages/backend-defaults/src/entrypoints/auth/external/types.ts
```diff
@@ -15,20 +15,39 @@
...
+export interface ExternalTokenHandler<TContext> {
+  type: string;
+  initialize(ctx: { options: Config }): TContext;
+  verifyToken(
+    token: string,
+    ctx: TContext,
+  ): Promise<{ subject: string } | undefined>;
 }
+
+// /**
```
freben: 👀

packages/backend-defaults/src/entrypoints/auth/external/ExternalAuthTokenHandler.ts
```diff
@@ -0,0 +1,160 @@
...
+
+const NEW_CONFIG_KEY = 'backend.auth.externalAccess';
+const OLD_CONFIG_KEY = 'backend.auth.keys';
+let loggedDeprecationWarning = false;
+
+/**
+ * @public
+ * This service is used to decorate the default plugin token handler with custom logic.
+ */
+export const externalTokenTypeHandlersRef = createServiceRef<
```
freben:
```suggestion
export const externalTokenHandlersServiceRef = createServiceRef<
```

packages/backend-defaults/src/entrypoints/auth/index.ts
```diff
@@ -19,4 +19,11 @@ export {
   pluginTokenHandlerDecoratorServiceRef,
 } from './authServiceFactory';
 
+export { createExternalTokenHandler } from './external/helpers';
+
+export { externalTokenTypeHandlersRef } from './external/ExternalAuthTokenHandler';
+
+export type { ExternalTokenHandler } from './external/types';
+export type { AccessRestrictionsMap } from './external/types';
```
freben: We shouldn't export this one

docs/auth/service-to-service-auth.md
```diff
@@ -414,9 +436,13 @@ Each entry has one or more of the following fields:
 
 ## Adding custom or logic for validation and issuing of tokens
 
-The `pluginTokenHandlerDecoratorServiceRef` can be used to decorate the existing token handler without having to re-implement the entire `AuthService` implementation.
+The `pluginTokenHandlerDecoratorServiceRef` and `externalTokenTypeHandlersRef` can be used to extend the existing token handler without having to re-implement the entire `AuthService` implementation.
```
freben:
```suggestion
The `pluginTokenHandlerDecoratorServiceRef` and `externalTokenHandlersServiceRef` can be used to extend the existing token handler without having to re-implement the entire `AuthService` implementation.
```

docs/auth/service-to-service-auth.md
```diff
@@ -439,3 +465,109 @@ const decoratedPluginTokenHandler = createServiceFactory({
   },
 });
 ```
+
+### Adding custom ExternalTokenHandler
+
+The `externalTokenTypeHandlersRef` can be used to add custom external token handlers to the default implementation.
```
freben:
```suggestion
The `externalTokenHandlersServiceRef` can be used to add custom external token handlers to the default implementation.
```

docs/auth/service-to-service-auth.md
```diff
@@ -439,3 +465,109 @@ const decoratedPluginTokenHandler = createServiceFactory({
...
+        accessRestrictions:
+          - plugin: events
+```
+
+And we can implement the custom token handler like this:
+
+```ts
+import {
+  ExternalTokenHandler,
+  externalTokenTypeHandlersRef,
```
freben:
```suggestion
  externalTokenHandlersServiceRef,
```

docs/auth/service-to-service-auth.md
```diff
@@ -439,3 +465,109 @@ const decoratedPluginTokenHandler = createServiceFactory({
...
+  createExternalTokenHandler,
+} from '@backstage/backend-defaults/auth';
+import { createServiceFactory } from '@backstage/backend-plugin-api';
+
+const customExternalTokenHandlers = createServiceFactory({
+  service: externalTokenTypeHandlersRef,
```
freben:
```suggestion
  service: externalTokenHandlersServiceRef,
```

packages/backend-defaults/src/entrypoints/auth/external/ExternalAuthTokenHandler.ts
Sarabadu: just a note on "This is not an API that we expect anyone from the outside to resolve and use anyway."
Im adding documentation about multiton [here](https://github.com/backstage/backstage/pull/28942/files?diff=split#diff-896a57047c94e1bdfdbc2e80703868b190f3ce87084ed48debab4fbe4668ee07R226) is ok to add docs if we are expecting to not be used?

plugins/catalog-react/report-alpha.api.md
Sarabadu, line 1: This is not related but it was failing on CI

Sarabadu:
Thanks @freben for the review, and sry for the mess and delay on this.
I think I cleaned 🧹 all now, and bring back the old configuration to avoid the breaking change.

One Api report from the catalog-react was complaining on CI and appear after rebase

packages/backend-defaults/src/entrypoints/auth/external/jwks.ts
```diff
@@ -15,85 +15,72 @@
...
+    const algorithms = readStringOrStringArrayFromConfig(options, 'algorithm');
+    const issuers = readStringOrStringArrayFromConfig(options, 'issuer');
+    const audiences = readStringOrStringArrayFromConfig(options, 'audience');
+    const subjectPrefix = options.getOptionalString('subjectPrefix');
+    const url = new URL(options.getString('url'));
     const jwks = createRemoteJWKSet(url);
-    const allAccessRestrictions = readAccessRestrictionsFromConfig(config);
-
-    this.#entries.push({
+    const allAccessRestrictions = readAccessRestrictionsFromConfig(options);
```
freben, line 50: You forgot to remove the restrictions usage in this one.

packages/backend-defaults/src/entrypoints/auth/external/ExternalAuthTokenHandler.ts
```diff
@@ -0,0 +1,172 @@
...
+    externalTokenHandlers?: ExternalTokenHandler<unknown>[];
+  }): ExternalAuthTokenHandler {
+    const {
+      ownPluginId,
+      config,
+      externalTokenHandlers: customHandlers,
+      logger,
+    } = options;
+
+    const handlersTypes = [...defaultHandlers, ...(customHandlers ?? [])];
```
freben: It may be worth going through these one by one and

* if there exists a custom handler for type X, do not add the same X from default handlers (in effect, you can then "replace" the implementation of builtins if you like)
* if several custom handlers for the same type X are registered, throw an error

Sarabadu: Good point @freben I did the adjustment. I've decided to not allow replacement of builtin handlers for now as if you need to have something custom it is simply updating the the type in the config too.

The only maybe good case for it is to "throw if any one try to use legacy" for example... but not sure if it worthy. let me know if you have anything else on mind that maybe make that more relevant

.changeset/thirty-rules-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-defaults': minor
+---
+
+Add a new `externalTokenHandlersServiceRef` to allow custom external token validations
```
Rugvip, line 5: Just to double check, PR body mentions that `backend.auth.keys` is removed, but I assume that was rolled back so it's no longer a breaking change?

In that case this is all good. It could be nice to add a link to the docs here for easy access, but no biggie.

docs/backend-system/architecture/03-services.md
Rugvip, line 1: Thank you! 🎉

.changeset/thirty-rules-press.md
Sarabadu: yes, thats was rolled back

PR #28950, opened 2025-02-24
[NFS]: Default Sticky Entity Page Layout

camilaibs:

Now the default entity page layout can be customized with cards of 3 types: `info`, `peek` and `full`.

- Cards of type info will be rendered in a right sticky area;
- Peek cards are rendered at the top of the main content area;
- Full cards and cards with undefined type are rendered as they were before, in the main area, below the peek cards.

So if you want to keep the layout as it was before, you don't need to do anything. But if you want to experiment with the card types and see how they are rendered, here is an example configuring the about card to be rendered as a card of type `info`:

```diff
app:
  extensions:
    # Entity page cards
+   - entity-card:catalog/about:
+       config:
+         type: info
```


plugins/catalog/src/alpha/DefaultEntityContentLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+              </Grid>
+            ))}
+          </Grid>
+        </Grid>
+      ) : null}
+      {peekCards.length > 0 ? (
+        <Grid xs={12} md={infoCards.length ? 8 : undefined} item>
+          <Grid container spacing={3}>
+            {peekCards.map((card, index) => (
+              <Grid key={index} className={classes.card} xs={12} md={6} item>
```
Rugvip, line 113: Bit large right? Thinking `HorizontalScrollGrid` might actually be the best start here and let the cards pick the size for now.

plugins/catalog/src/alpha/DefaultEntityContentLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+              </Grid>
+            ))}
+          </Grid>
+        </Grid>
+      ) : null}
+      {fullCards.length > 0 ? (
+        <Grid xs={12} md={infoCards.length ? 8 : undefined} item>
+          <Grid container spacing={3}>
+            {fullCards.map((card, index) => (
+              <Grid key={index} className={classes.card} xs={12} md={6} item>
```
Rugvip, line 124: Kinda imagining that these will always be 12, regardless of size. Maybe if you expand to widescreen we'd have some side by side, but that'd be `xxxl={6}`

plugins/catalog/src/alpha/DefaultEntityContentLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+      alignSelf: 'stretch',
+      '& > *': {
+        height: '100%',
+        minHeight: 400,
+      },
+    },
+  },
+}));
+
+const entityWarningContent = (
```
Rugvip, line 50: Another one to discuss potentially, bet let's leave as is for now

.changeset/orange-turtles-join.md
Rugvip, line 1: Thinking it would be best to replace the contents of `.changeset/real-hounds-mate.md` (wtf changesets 😅) instead?

plugins/catalog/src/alpha/DefaultEntityContentLayout.tsx
camilaibs: I used `md={6}` for visual backwards compatibility. Wouldn't it be a breaking change if I changed this for everyone using the default layout?

Rugvip: ah, arguably not or even the other way around I'd say. Currently we're kinda switching to 3 columns instead of 2 in total, but if we keep these cards at 12 it'd be more like 1.2 + 0.8 columns :p

camilaibs: In my mind, we only switch to 3 columns if people add types to the cards, but if they don't, the layout will remain 2 columns. However, I updated it to the size you suggested.

.changeset/orange-turtles-join.md
camilaibs: Done!

plugins/catalog/src/alpha/DefaultEntityContentLayout.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const { cards } = props;
+  const classes = useStyles();
+
+  const infoCards = cards.filter(card => card.type === 'info');
+  const peekCards = cards.filter(card => card.type === 'peek');
+  const fullCards = cards.filter(card => !card.type || card.type === 'full');
+  return (
+    <Grid container spacing={3}>
+      {entityWarningContent}
+      {infoCards.length > 0 ? (
```
Rugvip, line 95: Let's render the info cards above the full cards on `xs`, right now they appear lasts, but I think they should have higher prio

plugins/catalog/src/alpha/DefaultEntityContentLayout.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+const useStyles = makeStyles(theme => ({
+  peekCard: {
+    flex: '0 0 auto',
+  },
+  [theme.breakpoints.up('sm')]: {
+    infoArea: {
+      order: 1,
+      position: 'sticky',
+      top: -16,
```
Rugvip, line 44: I thinking a css grid would simplify the layout tbh, MUI grid isn't super good at this kind of stuff :p

Rugvip (approved):
Nice! 👌

Couple of things but pre-approved

Tried it out with a couple of peek cards to see what it looks like, in case anyone else is curious:

image

plugins/catalog/src/alpha/DefaultEntityContentLayout.tsx
Rugvip: Hmm, that's not what I saw, iirc it was `peek` > `full` > `info`

PR #28951, opened 2025-02-24
catalog-react: add support for entity predicates for filtering in new frontend system

### LINKED ISSUE #22390: 🚀 Feature: Support more advanced entity card filtering in new frontend system
### 🔖 Feature description

Support multiple `kind` and `type` filter combinations for entity cards in config.

An example is to show an entity card on a Component entity page but only for the Website type but also show the same card on System entity pages regardless of type.

### 🎤 Context

Currently, entity card filter config does not allow for filtering the `type` for one specific `kind` when you specify multiple `kind` values e.g.

```yaml
    - entity-card:catalog/depends-on-resources:
        config:
          filter: kind:component,system type:website
```

In the above example, System entity pages will not show the card because the `type` filter applies to both `component` and `system` but there is no `website` type for a System.

An example of how this could be expressed in config could be something like:

```yaml
filter: kind:component&type:website kind:system     
```

Open to better ideas though.

None

Rugvip:

This adds support for a new entity filtering format for the new frontend system called entity predicates. It's a JSON-based syntax loosely based on the [MongoDB query predicates](https://www.mongodb.com/docs/manual/reference/operator/query/exists/). Bunch more information on what it looks like in the doc additions in this PR, but here's a fairly simple example:

```yaml
filter:
  kind: component
  spec.type:
    $in: [service, website]
```

And here's a more advanced one:

```yaml
filter:
  $all:
  - kind: component
  - $any:
    - spec.type: service
    - spec.type: website
      metadata.annotations.acme.org/discovery-name: { $exists: true }
  - $not:
      spec.lifecycle: production
  - relations:
      $contains:
        type: ownedBy
        target: group:default/admins
```

🤞 that both of those are somewhat self-explanatory

The goal of this syntax right now is to use it for filtering in the new frontend system, completely replacing our existing string-based syntax. Longer term we might choose to adopt this syntax for filtering of entities via the catalog REST API as well, and it's therefore designed to be fairly constrained to be possible to transform to SQL. No immediate plans on adding that right now though. We did consider a string-based syntax in addition to the JSON-based one, but figured that at least for now it doesn't seem to be needed. A string-based syntax also brings some fairly large security risks for the backend API, so unless we feel it's needed in some other context it's unlikely we'll ever add it.

Closes #22390

### DIFF
--- a/docs/features/software-catalog/catalog-customization.md
+++ b/docs/features/software-catalog/catalog-customization.md
@@ -502,3 +502,201 @@ const routes = (
   </FlatRoutes>
 );
 ```
+
+## New Frontend System
+
+This section of the documentation explains how to create and configure catalog extensions in the [new frontend system](../../frontend-system/index.md).
+
+:::warning Warning
+
+This section is a work in progress.
+
+:::
+
+### Entity filters
+
+Many extensions that attach within the catalog entity pages accept a `filter` configuration. The purpose of the `filter` configuration is to select what entities the extension should be applied to or be present on. Many of these extension will have a default filter defined, but you can override it by providing your own. When defining filters in code you can use either a predicate function or a entity predicate query, while in configuration you can only use an entity predicate query.
+
+### Entity predicate queries
+
+The entity predicate syntax is a minimal JSON-based query language for filtering catalog entities. It is loosely inspired by the [MongoDB query syntax](https://www.mongodb.com/docs/manual/tutorial/query-documents/), behaving roughly the same way but with a different set of operators.
+
+The most simple entity predicate is an object expression with key-value mappings where the key is the full dot-separated path to the value in the entity, and the value is the value to do a case insensitive match against. Each entry in this object is evaluated separately, but all of them must match for the overall predicate to result in a match. For example, the following will match any component entities of the type `service`:
+
+```json
+{
+  "filter": {
+    "kind": "component",
+    "spec.type": "service"
+  }
+}
+```
+
+Or when utilizing YAML syntax:
+
+```yaml
+filter:
+  kind: component
+  spec.type: service
+```
+
+In addition to this basic syntax, entity predicates support logical operators that can be nested and applied around these object expressions. For example, the following will match all components entities that are of type `service` or `website`:
+
+```json
+{
+  "filter": {
+    "$all": [
+      {
+        "kind": "component"
+      },
+      {
+        "$any": [{ "spec.type": "service" }, { "spec.type": "website" }]
+      }
+    ]
+  }
+}
+```
+
+Or when utilizing YAML syntax:
+
+```yaml
+filter:
+  $all:
+    - kind: component
+    - $any:
+        - spec.type: service
+        - spec.type: website
+```
+
+Finally, entity predicates also support value operators that can be used in place of the values in the object expression. For example, the following is a simpler way to express the previous example:
+
+```json
+{
+  "filter": {
+    {
+      "kind": "component",
+      "spec.type": { "$in": ["service", "website"] }
+    },
+  }
+}
+```
+
+Or when utilizing YAML syntax:
+
+```yaml
+filter:
+  kind: component
+  spec.type:
+    $in: [service, website]
+```
+
+### Entity predicate logical operators
+
+The following section lists all logical operators for entity predicates.
+
+#### `$all`
+
+The `$all` operator has the following syntax:
+
+```json
+{ $all: [ { <expression1> }, { <expression2> }, ...] }
+```
+
+The `$all` operator evaluates to `true` if all expressions within the provided array evaluate to `true`. This includes an empty array, which means that `{ "$all": [] }` always evaluates to `true`.
+
+```yaml title="Example usage of $all"
+filter:
+  $all:
+    - kind: component
+    - $not:
+      spec.type: service
+```
+
+#### `$any`
+
+The `$any` operator has the following syntax:
+
+```json
+{ $any: [ { <expression1> }, { <expression2> }, ...] }
+```
+
+The `$any` operator evaluates to `true` if at least one of the expressions within the provided array evaluate to `true`. This includes an empty array, which means that `{ "$any": [] }` always evaluates to `false`.
+
+```yaml title="Example usage of $any"
+filter:
+  $any:
+    - kind: component
+    - metadata.annotations.github.com/project-slug: { $exists: true }
+```
+
+#### `$not`
+
+The `$not` operator has the following syntax:
+
+```json
+{ $not: { <expression> } }
+```
+
+The `$not` operator inverts the result of the provided express. If the expression evalutes to `true` then `$not` will evaluate to false, and the other way around.
+
+```yaml title="Example usage of $not"
+filter:
+  $not:
+    kind: template
+```
+
+### Entity predicate value operators
+
+The following section lists all value operators for entity predicates.
+
+#### `$exists`
+
+The `$exists` operator has the following syntax:
+
+```json
+{ field: { $exists: <boolean> } }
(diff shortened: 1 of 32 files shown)


packages/config-loader/src/sources/transform/include.ts
```diff
@@ -48,7 +50,9 @@ export function createIncludeTransform(
     }
     // Check if there's any key that starts with a '$', in that case we treat
     // this entire object as an include description.
-    const [includeKey] = Object.keys(input).filter(key => key.startsWith('$'));
+    const [includeKey] = Object.keys(input).filter(key =>
+      INCLUDE_KEYS.includes(key),
+    );
```
Rugvip, line 55: This is a tricky one that could use some discussion. In config we currently use `$` as a prefix for keys that should be processed by the config loader. This fixes it but mixing different usages of `$` prefixes still has the risk to cause confusion. One option could be to choose a different prefix for the predicates, or choose a completely different structure that doesn't require prefixing.

docs/features/software-catalog/catalog-customization.md
```diff
@@ -502,3 +502,201 @@ const routes = (
...
+```
+
+The `$all` operator evaluates to `true` if all expressions within the provided array evaluate to `true`. This includes an empty array, which means that `{ "$all": [] }` always evaluates to `true`.
+
+```yaml title="Example usage of $all"
+filter:
+  $all:
+    - kind: component
+    - $not:
+      spec.type: service
```
freben, line 612:
```suggestion
        spec.type: service
```

docs/features/software-catalog/catalog-customization.md
```diff
@@ -502,3 +502,201 @@ const routes = (
...
+
+#### `$not`
+
+The `$not` operator has the following syntax:
+
+```json
+{ $not: { <expression> } }
+```
+
+The `$not` operator inverts the result of the provided express. If the expression evalutes to `true` then `$not` will evaluate to false, and the other way around.
```
freben, line 640:
```suggestion
The `$not` operator inverts the result of the provided express. If the expression evaluates to `true` then `$not` will evaluate to false, and the other way around.
```

docs/features/software-catalog/catalog-customization.md
```diff
@@ -502,3 +502,201 @@ const routes = (
...
+
+#### `$exists`
+
+The `$exists` operator has the following syntax:
+
+```json
+{ field: { $exists: <boolean> } }
+```
+
+The `$exists` operator will evaluate to `true` if the existence of the value it matches against matches the provided boolean. That is `{ $exists: true }` will evaluate to `true` if and only if the value is defined, and `{ $exists: false }` will evalute to `true` if and only if the value is `undefined`.
```
freben, line 660:
```suggestion
The `$exists` operator will evaluate to `true` if the existence of the value it matches against matches the provided boolean. That is `{ $exists: true }` will evaluate to `true` if and only if the value is defined, and `{ $exists: false }` will evaluate to `true` if and only if the value is not defined.
```

plugins/catalog-react/report-alpha.api.md
```diff
@@ -324,6 +321,54 @@ export interface EntityContentLayoutProps {
...
+
+// @alpha (undocumented)
+export type EntityPredicateExpression = {
+  [KPath in string]: EntityPredicateValue;
+} & {
+  [KPath in `$${string}`]: never;
+};
+
+// @alpha (undocumented)
+export type EntityPredicatePrimitive = string | number | boolean;
```
freben, line 346: Hmm with the assumption that it gets stringified if made into a backend query later on, and that it somehow gets coerced or doesn't run into precision problems when running in-memory? I'm just thinking about what happens if you pass in `1/3` as the number, or if you check with equality with `4` but the entity contained `"4"`. Maybe we need to be very explicit here to not fall into a trap of implicit unclear behaviors 🤔

plugins/catalog-react/report-alpha.api.md
```diff
@@ -324,6 +321,54 @@ export interface EntityContentLayoutProps {
...
+
+// @alpha
+export function entityPredicateToFilterFunction<T extends JsonValue>(
```
freben, line 349: these probably should go into -common - especially the types - would it make sense to have them there from the start?

plugins/catalog-react/report-alpha.api.md
```diff
@@ -324,6 +321,54 @@ export interface EntityContentLayoutProps {
...
+    }
+  | {
+      $in: EntityPredicatePrimitive[];
+    }
+  | {
+      $contains: EntityPredicateExpression;
+    };
+
+// @alpha
+export function evaluateEntityPredicate(
```
freben, line 367: is it ok to force people to do `entityPredicateToFilterFunction(p)(value)` instead?

plugins/catalog-react/src/alpha/predicates/evaluateEntityPredicate.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Evaluate a single value against a predicate value.
+ *
+ * @internal
+ */
+function evaluatePredicateValue(
+  filter: EntityPredicateValue,
+  value: JsonValue | undefined,
+): boolean {
+  if (typeof filter !== 'object' || filter === null || Array.isArray(filter)) {
```
freben, line 84: slightly torn about null and what its meaning is - and how it translates to sql in the end

docs/features/software-catalog/catalog-customization.md
```diff
@@ -502,3 +502,201 @@ const routes = (
...
+```
+
+The `$contains` operator will evaluate to `true` if the value it is matched against is an array, and at least one of the elements in the array fully matches the provided expression. If the value matched against is not an array, or if the array is empty, the operator will always evaluate to `false`.
+
+The expression used to match against the array can be any valid entity predicate expression, including logical operators and value operators.
+
+```yaml title="Example usage of $contains"
+filter:
+  relations:
+    $contains:
```
freben, line 698: Hm how would we translate this to sql? I guess this is at the point where we actually could not support it in a general sense right?

freben (approved):
Approving with the mindset that it's limited to the new frontend system and could still see some tweaking, but looking overall good

plugins/catalog-react/report-alpha.api.md
Rugvip: ah yep let's clean this up

plugins/catalog-react/src/alpha/predicates/evaluateEntityPredicate.ts
Rugvip: Yep, I made it not match instead - like other invalid vales. It's already not allowed in types or schema.

docs/features/software-catalog/catalog-customization.md
Rugvip: Yep we'd need to update the storage

plugins/catalog-react/report-alpha.api.md
Rugvip: Hmm, yep it does although at the same time it increases the exposure a bit. Moving the evaluation seems a bit tricky since I don't know if we really want people using that in the backend right now. On the other hand moving just the types doesn't seem very useful either. Thinking that it might be safest to keep it all in -react for now?

plugins/catalog-react/report-alpha.api.md
Rugvip: Yep might be that we want all comparisons to be string based in the end, but on the other hand it's pretty likely we end up with some JSON-based filtering and then we have more strict types.

I think string vs number comparisons should probably always be handled though, so `4 = '4'` now.

PR #28962, opened 2025-02-25
Add Facets.cloud backstage plugin to directory

vivekvdwivedi:
Hi,

I would like to add Facets.cloud backstage plugin to the repository. 

The npm package is published at https://www.npmjs.com/package/@facets-cloud/backstage-plugin-backend

And the plugin code is available at https://github.com/Facets-cloud/facets-backstage-plugin


microsite/data/plugins/facets-cloud.yaml
```diff
@@ -0,0 +1,11 @@
+title: Facets.cloud Platform
```
itainathaniel, line 1: I think you forgot the YAML document separator:
```suggestion
---
title: Facets.cloud Platform
```

freben (approved):
Hi! Thanks for listing yourselves, and apologies for my slow review time. Happy to merge. Some small notes as I do so:

* You should add a config.d.ts file (and reference it in your package.json) that declares your configuration settings. See [here](https://backstage.io/docs/conf/defining)
* Hooks like saveUserConfig etc may enjoy using useAsync, for example [the one from the react-use library](https://github.com/streamich/react-use/blob/master/docs/useAsync.md) but there are more. Makes it a breeze to do the state/error handling bridging from pure promises to the react world
* It seems a bit concerning that the user's sign in credentials for your service are just posted and retrieved verbatim. I'd encourage some form of oauth flow or so, where the service doesn't need to persist the credentials themselves but rather a (revokable etc) token

That being said ,welcome!

Issue #28709, opened 2025-02-03
🚀 Feature: introduce `dangerouslyAllowSignInWithoutUserInCatalog` config for auth


JessicaJHee:
### 🔖 Feature description

During testing/development (in non-production environments), it would be nice to have a config that allows a user to sign in without the need for the user entity to be present in the catalog.

### 🎤 Context

I've read over [Sign-In without Users in the Catalog](https://backstage.io/docs/auth/identity-resolver/#sign-in-without-users-in-the-catalog) and understand that it's possible to implement a custom resolver with this option. However, it would require a lot of duplicate code/logic to apply this to all resolvers. 

I also tried looking into extending/wrapping the existing resolvers but couldn't find a clean way to do so.

### ✌️ Possible Implementation

Wrap the following in a try-catch block: https://github.com/backstage/backstage/blob/bc9a1100a69a52a798f30f68a9ef3fb7c097b3f9/plugins/auth-backend/src/lib/resolvers/CatalogAuthResolverContext.ts#L166-L183

Throw an error similar to 
https://github.com/backstage/backstage/blob/bc9a1100a69a52a798f30f68a9ef3fb7c097b3f9/plugins/auth-node/src/sign-in/readDeclarativeSignInResolver.ts#L70
by default, and issue a token based on the identifier in the query if `dangerouslyAllowSignInWithoutUserInCatalog: true`.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

Rugvip:
Just to repeat the outcome in the framework SIG, we're looking at moving forward by adding config here: https://github.com/backstage/backstage/blob/bc9a1100a69a52a798f30f68a9ef3fb7c097b3f9/plugins/auth-node/src/sign-in/commonSignInResolvers.ts#L84

MihaiTabara:
Closing this issue as the PR was merged. Thank you!

PR #28967, opened 2025-02-25
introduce dangerouslyAllowSignInWithoutUserInCatalog auth resolver config

JessicaJHee:

This PR introduces the `dangerouslyAllowSignInWithoutUserInCatalog` option for resolvers. This will allow users to sign in even if their user entity has not been synced into the catalog. 

~~Currently, the logic is in the `handleSignInUserNotFound` method which issues a Backstage token based on the `id` used in each resolver.~~
This PR modifies the `signInWithCatalogUser` method to take in an additional (optional) parameter `dangerouslyAllowSignInWithoutUserInCatalog?: boolean | undefined`

Related issue for more context: https://github.com/backstage/backstage/issues/28709

### Resolvers not supporting this config at the moment:

- [gitlab](https://github.com/backstage/backstage/blob/master/packages/backend-legacy/src/plugins/auth.ts#L70-L80) and [onelogin](https://github.com/backstage/backstage/blob/master/packages/backend-legacy/src/plugins/auth.ts#L105-L115) from legacy backend
- [common resolvers](https://github.com/backstage/backstage/blob/master/plugins/auth-backend/src/providers/resolvers.ts)  in `auth-backend`
    - used by [okta](https://github.com/backstage/backstage/blob/master/plugins/auth-backend/src/providers/okta/provider.ts#L60-L87)
- [saml](https://github.com/backstage/backstage/blob/master/plugins/auth-backend/src/providers/saml/provider.ts#L196-L213)

I’m unsure if support is still declared for these use cases, if so I’ll look into implementing this config with these as well.


docs/auth/identity-resolver.md
```diff
@@ -403,6 +407,22 @@ async signInResolver({ profile }, ctx) {
 }
 ```
 
+##### Using the `dangerouslyAllowSignInWithoutUserInCatalog` Option
```
awanlin, line 410: Can we include some clear details on the impact of turning this on? Just to help those who might use this better understand why it's prefixed with `dangerouslyAllow` 👍

awanlin (changes requested):
Thanks @JessicaJHee, I'm just focused on the docs, leaving the code changes for others. Left one comment around that aspect.

docs/auth/identity-resolver.md
JessicaJHee: Sounds good and thanks for taking a look! I've added a simple explaination, I'm conflicted on how much (technical) detail to add here, what do you think?

docs/auth/identity-resolver.md
```diff
@@ -402,6 +406,30 @@ async signInResolver({ profile }, ctx) {
...
+          resolvers:
+            - resolver: emailLocalPartMatchingUserEntityName
+              dangerouslyAllowSignInWithoutUserInCatalog: true
+```
+
+:::warning
+Enabling this option in production poses security risks.
+:::
+
+This option grants access to unexpected users who haven’t been onboarded into Backstage. It also breaks integration with permission frameworks like the RBAC plugin, since there is no user entity to associate with the signed-in user, potentially leading to unauthorized access to resources.
```
Parsifal-M, line 431: Hello! 👋 

Let me know if I am misunderstanding, but enabling this will still mean that users are authenticated right? they still go through the authentication that is in place, we just don't additionally check if they exist in the catalog (which is what I gathered from the sig) 

I think as we don't assign them anything as a fallback (which could be cool thing to do maybe later on maybe as config) we can simply say that the permissions framework or just permissions will not be applied to these users at all as they technically do not exist in the catalog.

If the first statement is true though and I understood correctly I'd definitely mention that part somewhere 😄 

Also this is really cool and I think definitely something we'd use!

Thanks!

JessicaJHee: Hi! Thank you so much for your feedback, you are right and users will still be authenticated. I've updated the docs to make this clearer :)

I've also added more detail on how this interacts with permissions (tested with the RBAC plugin), let me know how it looks!

Parsifal-M: Nice! LGTM! 👏

JessicaJHee:
Hi @Rugvip! We discussed this PR in last week's framework sig and there was no additional feedback from the rest of the team. If you can please take another look at this PR that'd be amazing!

freben:
Hi @JessicaJHee! Sorry I saw this one was still pending, and noticed that unfortunately there's a yarn.lock conflict. While we land back from kubecon and all the crazy stuff around that, would you mind rebasing this so it's green when we get to it? Thanks

docs/auth/identity-resolver.md
```diff
@@ -402,6 +406,36 @@ async signInResolver({ profile }, ctx) {
 }
 ```
 
+##### Using the `dangerouslyAllowSignInWithoutUserInCatalog` Option
+
+Another way to bypass this requirement is to enable the `dangerouslyAllowSignInWithoutUserInCatalog` option for resolvers.
+Users will still be authenticated as usual but this config will bypass the check that ensures the user is present in the catalog.
+If the user entity is not found in the catalog, a Backstage user token will still be issued based on the identifying information available at the resolver level.
```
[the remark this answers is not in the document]
JessicaJHee: Was there a whitespace change suggested to this?

JessicaJHee:
@Rugvip Thank you so much for the review! I've addressed your feedback, but I have 2 final questions:

1. Since we’re adding the `NotFoundError` check in the `signInWithCatalogUser` method, would it be safe to remove the try catch [here](https://github.com/backstage/backstage/blob/master/plugins/auth-node/src/sign-in/readDeclarativeSignInResolver.ts#L59-L72)?

2. Currently, I have this as the error message displayed when the user is not found, please lmk if it needs changes:
  "Login failed; caused by Error: Failed to sign-in, unable to resolve user identity. Please verify that your catalog contains the expected User entities that would match your configured sign-in resolver. For non-production environments, manually provision the user or disable the user provisioning requirement by setting the `dangerouslyAllowSignInWithoutUserInCatalog` option."

plugins/auth-node/src/types.ts
```diff
@@ -160,6 +160,8 @@ export type AuthResolverContext = {
    */
   signInWithCatalogUser(
     query: AuthResolverCatalogUserQuery,
+    fallbackUserRef?: string | undefined,
+    dangerouslyAllowSignInWithoutUserInCatalog?: boolean | undefined,
```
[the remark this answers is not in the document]
JessicaJHee: Sounds good! I agree that's much cleaner 👍

JessicaJHee:
Updates from the framework sig discussion:
> 1. Since we’re adding the NotFoundError check in the signInWithCatalogUser method, would it be safe to remove the try catch [here](https://github.com/backstage/backstage/blob/master/plugins/auth-node/src/sign-in/readDeclarativeSignInResolver.ts#L59-L72)?
* No, we should keep the try-catch in cases where a [custom resolver](https://backstage.io/docs/auth/identity-resolver/#building-custom-resolvers) is supplied 

> 2. ... For non-production environments, manually provision the user or disable the user provisioning requirement by setting the dangerouslyAllowSignInWithoutUserInCatalog option."
* We should refrain from mentioning the config in the error message as it doesn't make sense to expose the `dangerouslyAllowSignInWithoutUserInCatalog` config to end users. (i.e. keep the message the same as [here](https://github.com/backstage/backstage/blob/master/plugins/auth-node/src/sign-in/readDeclarativeSignInResolver.ts#L70))
* I've moved the suggestion to apply the `dangerouslyAllowSignInWithoutUserInCatalog` config to shown in the logs instead [here](https://github.com/backstage/backstage/blob/c30d1a9963c33fb96b388563e4a10d553e1921d1/plugins/auth-backend/src/lib/resolvers/CatalogAuthResolverContext.ts#L172C11-L174C9).

docs/auth/identity-resolver.md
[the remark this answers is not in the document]
freben: I don't see anything wrong too 🤔

awanlin (dismissed):
Added a comment around the vale error 👍

plugins/auth-backend/src/lib/resolvers/CatalogAuthResolverContext.ts
```diff
@@ -139,19 +139,59 @@ export class CatalogAuthResolverContext implements AuthResolverContext {
     return { entity: result };
   }
 
-  async signInWithCatalogUser(query: AuthResolverCatalogUserQuery) {
-    const { entity } = await this.findCatalogUser(query);
+  async signInWithCatalogUser(
+    query: AuthResolverCatalogUserQuery,
+    options?: {
+      dangerousEntityRefFallback?:
```
freben, line 145: Suggestion:

```ts
dangerouslyAllowFallbackForMissingEntity?:
  | {
      entityRef: string | {
        kind?: string;
        namespace?: string;
        name: string;
      }
    }
  | undefined;
```

That makes it "a whole fallback" as a concept, and the one required field in it (for now) is the entityRef.

freben (commented):
Alright, very close now!

One thought I had throughout is, I wonder what a sensible fallback really is in some of these resolvers.

- The full email (where applicable) is sort of ok, we could take only what's before the `@` sign but that's probably not really any better.
- Then there are some that are like UUIDs and stuff like that. Maybe there's something nicer in the profile to pick from that is more human-consumable, but it's sort of hard to know for sure for all of these third parties.

You don't have to make any changes because of this last comment, I'm just curious about your thoughts on it.

JessicaJHee:
@freben Thanks so much for the review! 

> I wonder what a sensible fallback really is in some of these resolvers.

In the current implementation, I initially leaned toward using whatever identifier the resolver used to locate the user entity, even if that meant falling back to something not very human-readable, since it was the most unique/reliable reference.

But after thinking about it more, especially given that the fallback user isn’t a “real” user entity, it might make more sense to prioritize readability like you said. For example, we could use the user’s `metadata.name` as the fallback, but set it in a different namespace like `development` or `fallback`. That way, it looks much nicer (like in the settings page), and we avoid the risk of clashing with any real user with the same `metadata.name` in the default namespace.

What do you think about that approach?

Rugvip (approved):
Thank you, looks great, let's :shipit:! 🎉 

I pushed a change to split up the changesets a bit for explanations from `auth-node` and `auth-backend`'s PoV, so that we can get this into today's release

Issue #14574, opened 2022-11-11
📁 Catalog Improvement Meta Issue


Rugvip:
### 🔖 Feature description

There are a lot of ideas around different ways that we can improve the Software Catalog. I've created this issue as a place to track all of these ideas at to make it easier to get an overview of what's already being discussed. It is also a place where we can synchronize the work between members of the community that want to help drive the catalog forward.

Please suggest things to add if you think that something is missing! 🙏 

## Model

- Better docs for extending the catalog: #14383
- There should be a more streamlined API for defining custom catalog models.
- Transitive relations: #7109

## Features

- Ingestion errors are hard to catch and need to be surfaced somehow.
- There is no good UX flow for moving or renaming entities
- Event-based catalog consumption: #8219
- Entity modification timestamps: #14181
- Multi-fetch: #12254, #14354
- Catalog federation: #4888
- Catalog onboarding page: #4741
- Entity statuses: #2292

## Ingestion

- Event-driven ingestion: #8232, #12155, #13931, #12156, #13575, #13531
- Ingestion UX: #11445
- Entity validation: #9245
- Refer to default branch: #9086
- Completely remove orphaned entities as a concept. Related: #7860
- Disallow entity deletions: #8616
- Disallow rogue locations: #13959
- Entity location deduplication: #14042

## Performance

- Catalog processing can be slow and not scale beyond a certain point
- Improved pagination: #12247, #12246
- Facets endpoint is slow: #13655
- Catalog is slow with large number of entities: #12829

## Bugs

- Provided entities not being deleted: #12883
- Unexpected sticky relations: #7872
- Conflicting entities are not detected: #10862

### 🎤 Context

.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

dtuite:
Excellent list @Rugvip .

Based on our experience rolling out Backstage with a few dozen customers, the items I'd highlight from your list are (in order)...

1.  Improved pagination (we estimate this will cut ~8 seconds (aka. 85%) off some page loads)
2. Ingestion errors are hard to catch and need to be surfaced somehow (seems like ~50% of customers ask "I can't see my component in the catalog..." at some point).
[...]

PR #28968, opened 2025-02-25
(WIP) feat(catalog): add support for automatically cleaning up locations

### LINKED ISSUE #28638: 🚀 Feature: Catalog: Enable eager deletion for nonexistent catalog entities
### 🔖 Feature description

When currently a catalog gets regularly deleted from the repository (so the api requests return 404) the catalog entity stays in the database.
This leads to desynchronization between the source of truth and the catalog.

### 🎤 Context

In our case we have multiple actors editing the repository, which includes editing the entities and also deleting them.
However that doesn't reflect on the backstage catalog.

Another, even bigger problem is that those deleted entities are still being queries for and thus consuming github rate limits.

Add a flag in the app-config
catalog.enableDeletionOfNonExistentEntities or catalog.externalSourceOfTruth

so that when the API request returns 404, the entity gets deleted (or at least soft deleted...)

benjidotsh:
This PR resolves #28638 by introducing a new app config property called `locationNotFoundStrategy`. By setting `locationNotFoundStrategy` to `delete`, locations containing not found errors will be deleted during processing.

In combination with setting `orphanStrategy` to `delete`, this can be useful when you want an entity to be automatically removed after removing the associated repository from your SCM provider.

This PR is marked as WIP because I'm not sure if this is the best way to do it, so feel free to discuss. As soon as a decision has been made, I will add documentation, tests and changesets.

### DIFF
--- a/plugins/catalog-backend/config.d.ts
+++ b/plugins/catalog-backend/config.d.ts
@@ -173,6 +173,12 @@ export interface Config {
         };
 
     /**
+     * The strategy to use for locations that return a 404 when fetching the
+     * catalog-info file. The default value is "keep".
+     */
+    locationNotFoundStrategy?: 'keep' | 'delete';
+
+    /**
      * The interval at which the catalog should process its entities.
      * @remarks
      *
--- a/plugins/catalog-backend/src/database/operations/util/deleteLocation.ts
+++ b/plugins/catalog-backend/src/database/operations/util/deleteLocation.ts
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2025 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


benjdlambert:
Maybe on to bring up at the [Framework SIG](https://github.com/backstage/community/tree/main/sigs/sig-framework)?

benjidotsh:
@benjdlambert sure, I'll be there!

Rugvip:
Added this to #14574. Quick summary from the framework SIG discussion:

We want it to be a smooth experience to create new entities via the scaffolder and immediately be able to see them in the catalog, even when using discovery. One ways to achieve that right now is to use event-based ingestion, but the other is to also register the new locations via the locations API. That however leads to this issue, where you end up with a lot of locations lying around that point to entities that have been removed, unless you also have a process around cleaning that up.

We found it best to not move this PR forward, simply because it's very dangerous to ever enable this flag. There are other ways to hit 404s, for example if you access token is not valid for the location. If that happens you'd usually just end up with ingestion errors in the catalog, but with this enabled you might end up with the entire catalog deleted.

Workarounds that we discussed for now is to use event-based ingestion and avoid registering new entities in the catalog via the API. @schultzp2020 also had the idea to do cleanup similar to this but to gate it via some checks, for example that the location has 404'd for some extended period of time.

PR #28971, opened 2025-02-25
Feature: [Catalog] Remove entities from old EntityProviders by default

### LINKED ISSUE #18067: 🚀 Feature: Remove entities from old EntityProviders by default
### 🔖 Feature description

Deleting(uninstalling) an entityProvider or changing the EntityProvider name cause Entities to remain in the catalog unless manually removed. This causes hard to debug issues for users that for example move to a new Entity Provider or change their Entity Provider name leading to reference issues as the `source_key`(EntityProvider name) in the `refresh_state_references` table still remain. This manifest itself as deletion/sync issues as the Entity is "owned" by a provider which is in reality no longer in use so the entity will not be managed any longer.

Change the default startup behaviour of the Catalog to remove entities from EntityProviders that are no longer installed.
This will work for most users but will be undesirable if you shard your ingestion and have a subset of EntityProviders installed on some backends only. This could be mitigated by having a config flag to disable this behaviour.

Initially we could add said config flag but default it to false to retain the old behaviour and potentially avoid having to introduce a breaking change in the catalog.

UsainBloot:
## Context
When switching between different catalog providers, entities related to that provider will remain in the catalog unless manually removed or a custom cleanup provider is created.

This change enables the ability to automatically clean up entities that belong to entity providers which are no longer configured, the next time the catalog is started.

Fixes #18067

💬 _Let me know if you would like me to change some of the names of the config, variables and functions to better fit the catalog terminology._

## Changes
- Removes all entities from entity providers which are no longer added to the catalog
- Adds opt-in `catalog.orphanProviderStrategy: delete` to enable feature

### DIFF
--- a/.changeset/young-carpets-love.md
+++ b/.changeset/young-carpets-love.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Added opt-in ability to evict entities from the catalog whose provider is no longer configured
--- a/.changeset/young-carpets-love.md
+++ b/.changeset/young-carpets-love.md
@@ -1,5 +1,0 @@
----
-'@backstage/plugin-catalog-backend': patch
----
-
-Added opt-in ability to evict entities from the catalog whose provider is no longer configured
--- a/plugins/catalog-backend/config.d.ts
+++ b/plugins/catalog-backend/config.d.ts
@@ -216,5 +216,13 @@ export interface Config {
      * This flag is temporary and will be enabled by default in future releases.
      */
     useUrlReadersSearch?: boolean;
+
+    /**
+     * Evicts entities from the catalog that are no longer referenced by any
+     * added entity providers.
+     *
+     * Defaults to false.
+     */
+    evictOrphanedEntityProviders?: boolean;
   };
 }
--- a/plugins/catalog-backend/config.d.ts
+++ b/plugins/catalog-backend/config.d.ts
@@ -218,10 +218,7 @@ export interface Config {
     useUrlReadersSearch?: boolean;
 
     /**
-     * Evicts entities from the catalog that are no longer referenced by any
-     * added entity providers.
-     *
-     * Defaults to false.
+     * Evicts entities from the catalog that are no longer referenced by entity providers added to the catalog.
      */
     evictOrphanedEntityProviders?: boolean;
   };
--- a/plugins/catalog-backend/config.d.ts
+++ b/plugins/catalog-backend/config.d.ts
@@ -156,6 +156,13 @@ export interface Config {
     orphanStrategy?: 'keep' | 'delete';
 
     /**
+     * The strategy to use for entities that are referenced by providers that are orphaned,
+     * i.e. entities with no providers currently configured in the catalog. The default value is
+     * "keep".
+     */
+    orphanProviderStrategy?: 'keep' | 'delete';
+
+    /**
      * The strategy to use when stitching together the final entities.
      */
     stitchingStrategy?:
@@ -216,10 +223,5 @@ export interface Config {
      * This flag is temporary and will be enabled by default in future releases.
      */
     useUrlReadersSearch?: boolean;
-
-    /**
-     * Evicts entities from the catalog when their related entity provider no longer exists.
-     */
-    evictOrphanedEntityProviders?: boolean;
   };
 }
--- a/plugins/catalog-backend/config.d.ts
+++ b/plugins/catalog-backend/config.d.ts
@@ -218,7 +218,7 @@ export interface Config {
     useUrlReadersSearch?: boolean;
 
     /**
-     * Evicts entities from the catalog that are no longer referenced by entity providers added to the catalog.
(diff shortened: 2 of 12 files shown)


plugins/catalog-backend/src/database/DefaultProviderDatabase.ts
```diff
@@ -199,6 +199,20 @@ export class DefaultProviderDatabase implements ProviderDatabase {
     }
   }
 
+  async listReferenceSourceKeys(txOpaque: Transaction): Promise<string[]> {
```
Rugvip, line 202: Given that we need to add a new DB call either way, perhaps we should make it `pruneOrphanedSourceKeys(ts, knownKeys)` and have that do the delete straight away?

plugins/catalog-backend/config.d.ts
```diff
@@ -216,5 +216,10 @@ export interface Config {
      * This flag is temporary and will be enabled by default in future releases.
      */
     useUrlReadersSearch?: boolean;
+
+    /**
+     * Evicts entities from the catalog when their related entity provider no longer exists.
+     */
+    evictOrphanedEntityProviders?: boolean;
```
Rugvip, line 223: Potential alternative here: should `orphanStrategy: 'delete'` be the switch instead? I think it fits pretty well but is of course a risk at it affects existing users more directly. Probably want to avoid it because of that, but figured it's worth a discussion.

.changeset/young-carpets-love.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
```
Rugvip, line 2: `minor` for this one imo, since we're introducing a new feature and it's 1.0

.changeset/young-carpets-love.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Added opt-in ability to evict entities from the catalog whose provider is no longer configured
```
Rugvip, line 5: Best link to docs or mention the new config key here

Rugvip (commented):
Nice! 👍, couple of things

plugins/catalog-backend/config.d.ts
UsainBloot: I can imagine wanting finer grain control over orphans. 

It's reasonable to expect you might want to keep orphaned entities as they have accidentally been deleted in the source as that is out of your direct control. However, you have control over the providers configured, so would want to remove entities for providers that are no longer configured. 

I do prefer this configuration name though, so I might change it to `orphanedProviderStrategy: 'delete'`

.changeset/young-carpets-love.md
UsainBloot: Resolved in [ca9c51b](https://github.com/backstage/backstage/pull/28971/commits/ca9c51ba1bb30ea326e323c37e1ba469222a2452)

.changeset/young-carpets-love.md
UsainBloot: Resolved in [ca9c51b](https://github.com/backstage/backstage/pull/28971/commits/ca9c51ba1bb30ea326e323c37e1ba469222a2452)

plugins/catalog-backend/config.d.ts
UsainBloot: I've pushed config proposal [077493b](https://github.com/backstage/backstage/pull/28971/commits/077493b896e45a48c42b66951891b761e178a92d)

plugins/catalog-backend/src/database/DefaultProviderDatabase.ts
UsainBloot: I can do, but I think it is important to log when a provider is being cleaned. No other methods for database's include an option for a logger so didnt want to break that approach with this change.

UsainBloot: Also, by having these all in a single transaction, wouldn't this roll back all deletions if a particular provider failed?

plugins/catalog-backend/src/service/CatalogBuilder.ts
```diff
@@ -645,6 +646,16 @@ export class CatalogBuilder {
 
     await connectEntityProviders(providerDatabase, entityProviders);
```
Rugvip, line 647: I wonder if it might be best to evict before we connect the new ones? That way any synchronous ingestion will avoid conflicting with orphaned entities, otherwise you'd need a second instance or restart to reconcile that

Rugvip (approved):
Nice! 👍 

Minor bit on the ordering but otherwise I think it looks good. @freben wanna have a final look as well before merge?

freben (commented):
Yep looks great. 🙏  Agree that we could do the cleanup before connect.

plugins/catalog-backend/src/service/CatalogBuilder.ts
UsainBloot: Resolved in [abd0ab7](https://github.com/backstage/backstage/pull/28971/commits/abd0ab7901aff9539992c03e17764acd8622fc61)

PR #28982, opened 2025-02-27
auth-backend: deprecations everywhere

Rugvip:
🧹, one extra round of deprecations. Technically all of these exports are indirectly deprecated already as they're not useful without `createRouter`, which is marked as deprecated. Figured it's best to be very clear here though and give it one more release.


awanlin (approved):
Thanks!!! 🚀

freben (approved):
yesss

vinzscam (approved):

PR #28994, opened 2025-02-28
add deprecated tag check for createRouter

patroswastik:

- Added a specific check for the `@deprecated` tag through `router.ts` files to ensure consistency with `index.ts`. 
- Will only display `❌ createRouter is NOT deprecated` when both the conditions have been met


packages/repo-tools/src/commands/lint-legacy-backend-exports/lint-legacy-backend-exports.ts
```diff
@@ -75,9 +80,23 @@ function verifyIndex(pkg: string, packageJson?: BackstagePackageJson) {
       .find(tag => tag.getName() === 'deprecated');
   }
 
+  let routerCreateRouterDeprecated = undefined;
```
freben, line 83: Instead of doing this explicit file munging, can't we modify the check on the lines above, to be not just `createRouterExport has a deprecated tag`, but also `if createRouterExport is a re-export from a different, place, then does THAT thing have a deprecated tag`? Traversing up the tree so to speak, instead of hard coding what file to look into

packages/repo-tools/src/commands/lint-legacy-backend-exports/lint-legacy-backend-exports.ts
```diff
@@ -46,10 +46,15 @@ function verifyIndex(pkg: string, packageJson?: BackstagePackageJson) {
   console.log(`Verifying ${pkg}`);
   const tsPath = path.join(pkg, 'src/index.ts');
   const sourceFile = project.getSourceFile(tsPath);
+
+  const tsRouterPath = path.join(pkg, 'src/service/router.ts');
+  const routerFile = project.getSourceFile(tsRouterPath);
```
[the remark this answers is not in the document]
patroswastik: I see, so this could be handled by moving this particular section after the `if (!sourceFile)` condition which makes sense. [Obviously, based on the fact that if hardcoded approach is used]

packages/repo-tools/src/commands/lint-legacy-backend-exports/lint-legacy-backend-exports.ts
patroswastik: I noticed something tricky in the code. The `createRouterDeprecated` variable is only having undefined or empty values. Currently, the presence of deprecated tags is determined by checking if `createRouterDeprecated` has any value, even if that value is empty (but not undefined). This behavior is due to the `getJsDocTags` method, which returns an array of empty objects, though the length of the array is correct for the tags that actually exist in the block.

Also, I wanted to update you that I am able to traverse up the tree now. However, before proceeding further will need your suggestion on the above scenario. If that's good then I can proceed with it and raise a PR for your review.

freben: Hm. I'm not sure I understand what the question was. I am not intimately familiar with this code from the start, so I'm just curious what your suggested solution will look like :)

patroswastik: Apologies for the delay; got caught up with some personal matters. I've implemented the solution as per your instructions. Now, I'm traversing up the tree to locate where `createRouter` is declared, and at that point, it checks if the deprecated tag exists. This approach ensures that it will work as intended, even if the file structure changes. Could you please take a look at it?

.changeset/rich-phones-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/repo-tools': minor
+---
+
+Checking up the files where createRouter has been declared and check if @deprecated tag exists. If it does not exist then only the message will appear.
```
freben, line 5:
```suggestion
Checking up the files where `createRouter` has been declared and check if `@deprecated` tag exists. If it does not exist then only the message will appear.
```

.changeset/rich-phones-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/repo-tools': minor
```
freben, line 2:
```suggestion
'@backstage/repo-tools': patch
```

just a small fix

freben (approved):

Issue #28773, opened 2025-02-08
Refactor the form state handling in the scaffolder plugin


benjdlambert:
### 🔖 Feature description

There's been a few bugs coming up with the collection of the scaffolder form state, and it's really hard to make any changes to the form state handling because of the lack of test coverage that we have with different templates and form interactivity.

Ideally we'd want to rebuild a lot of the state tracking to ensure that the form steps are collected separately, but then collected together to provide a running state which can be computed on each step.

This might help us support things like #16275 in the future to have better state management.

### 🎤 Context

https://github.com/backstage/backstage/pull/27273
https://github.com/backstage/backstage/pull/27230
possibly: https://github.com/backstage/backstage/issues/23836

### ✌️ Possible Implementation

As above

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

Aarav2002:
Hey! I have started exploring Backstage and would love to contribute to this issue.

kurtaking:
not stale

JounQin:
So what's the next step? We need a high level design here?

ganievs:
Up

benjdlambert:
Hopefully can get around to this ticket in the coming weeks :pray:

christoph-jerolimov:
@benjdlambert are you working on this or are you open for help?

nicoBruhn:
not stale

PR #28995, opened 2025-02-28
fix: secrets are properly validated with pattern in scaffolder template

Nikunj0601:

This PR solves the issue #23836. Secrets are currently validated with pattern by their masked value '***' instead of actual secret. So secrets are accessed in formData for validation and then deleted. 

Screenshot before:
image

after:
image

for both input is `testtest` and pattern: `^[a-z]{1}[a-z0-9]{2,7}$`


plugins/scaffolder-react/src/next/components/Stepper/Stepper.tsx
```diff
@@ -186,32 +189,31 @@ export const Stepper = (stepperProps: StepperProps) => {
...
-          analytics.captureEvent('click', `Next Step (${stepNum})`);
-          return stepNum;
-        });
+        setIsValidating(false);
+        return;
       }
+
+      const cleanedFormData = { ...formData };
+      Object.keys(secrets).forEach(secretKey => {
+        delete cleanedFormData[secretKey];
```
benjdlambert, line 205: I think unfortunately, this only works for top level secret keys.

Imagine you have a schema like this:

```yaml
parameters:
  test:
    type: object
    required: [thing]
    properties:
      thing:
        ui:field: SecretField
```

Then this solution doesn't work right?

I think it's part of a bigger refactor of how we deal with the form state and secrets #28773

benjdlambert, line 205: I think unfortunately, this only works for top level secret keys.

Imagine you have a schema like this:

```yaml
parameters:
  test:
    type: object
    required: [thing]
    properties:
      thing:
        ui:field: SecretField
```

Then this solution doesn't work right?

I think it's part of a bigger refactor of how we deal with the form state and secrets #28773 

I'd expect `${{ secrets.test.thing }}` to work here.

benjdlambert (changes requested):
Hey :wave: 

Thanks for trying to take a stab at this!

Bit of a problem I can see though at the minute that I think we wanna fix but it might require a lot of other work.

plugins/scaffolder-react/src/next/components/Stepper/Stepper.tsx
Nikunj0601: I was looking into this with object schema type so it is working currently but not with `${{ secrets.test.thing }}` instead it is working for `${{ secrets.thing }}` cause the way secrets are currently stored is 
```
secrets: {
   thing: 'secretvalue',
}
```
```
formData: {
    "test": {
        "thing": "testtest"
    }
}
```

So, to make it work, I can think of two options: change the object storing to accommodate `${{ secrets.test.thing }}` or leave it as it is and in formData recursively check for secretKey to delete in formData after validation is completed.

I had one more question so when we click on create tasks api is called so for that payload do we need value of all the properties in formData even though one is secret?

Nikunj0601: I was checking on recursively delete or mask secrets approach and it was working on the frontend side but it was throwing error from backend because in validate function only payloads property `values` but in which value of secret will either masked or deleted and it will throw an error. I was wondering should we need the value of secret field as it is or not?

benjdlambert: Yeah, that's also the problem right, is that we need to split the schema apart into a secret schema to validate them individually. It all get a little bit messy and confusing, but maybe it's what we have to do unless we come up with a better way to define a secret schema or something?

Nikunj0601: I have pushed some changes in API to have a correct value in the secret field for the validation to be properly completed. Let me know if I have not understood it properly or if there are any changes to handle it in a better way.

Nikunj0601: @benjdlambert any feedback on the approach or it is not something you meant earlier.

Nikunj0601: @benjdlambert can you review the changes when you have a moment? I'd appreciate your feedback! Thanks!

benjdlambert: @Nikunj0601 sorry - haven't had chance to look at this yet - busy prepping for KubeCon. It's on my list though I haven't forgotten!

Nikunj0601:
not stale

Nikunj0601:
not stale

plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -604,7 +604,51 @@ export async function createRouter(
         );
 
         for (const parameters of [template.spec.parameters ?? []].flat()) {
-          const result = validate(values, parameters);
+          const validateWithSecrets = (
```
vinzscam, line 607: Thanks! Can you add some tests for this corner case? I'm not familiar with the shape of the data, so it will help me understand the problem 🙏

Perhaps we can also move these two new functions outside of the main route so that they are defined once.

plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -604,7 +604,51 @@ export async function createRouter(
...
+            parameterSchema: TemplateParametersV1beta3,
+            secrets: any,
+          ) => {
+            const valuesCopy = structuredClone(formValues);
+            console.log('parameters', parameterSchema);
+
+            const replaceMaskedSecrets = (
+              obj: any,
+              schema: any,
+              secretObj: any,
```
vinzscam, line 618: let's use proper typings here

vinzscam (changes requested):

plugins/scaffolder-backend/src/service/router.ts
Nikunj0601: I have added the test case and did moved those function outside. So currently when input is secret then it is masked value but actual value is sent in secrets field of payload. However while validating with condition it is using that masked value which was failing so i am replacing masked value with actual value from secrets field and then validating so it can poperly validate the input with conditions.

vinzscam (approved):
It looks good to me but I leave it to @backstage/scaffolder-maintainers for a final review 👍

benjdlambert:
Need to have a think around this first before merging I think - just wanna cover some edge cases to make sure this works as expected :pray:

Nikunj0601:
> Need to have a think around this first before merging I think - just wanna cover some edge cases to make sure this works as expected 🙏

Let me know if I have missed anything as I am not associated with any company which uses backstage so I might not know in detail about usecases of scaffolder templates.

awanlin:
Hi @benjdlambert, just want to give you a nudge on this as the PR got closed but this seems like it's close to being able to ship.

benjdlambert:
@awanlin @Nikunj0601 still haven't had time to get around to looking into this yet, also it's looking like #28773 is also becoming more relevant so not sure how this all fits together yet. Will try and get some time before next mainline to look at this. :pray: Apologies for the delay, we're focused on a lot of NFS stuff right now.

Nikunj0601:
!not-closed

PR #29015, opened 2025-03-01
kubernetes-backend: Removed old backend system support

freben:
The time has come. Also, this removes the `@backstage/backend-common` usages.


mclarke47 (approved):
I'll need to trust you on some of these!

freben:
@mclarke47 I think my main ask here is, are you comfortable with these interfaces going away? Notably, the builder had

```
  setAuthStrategyMap(authStrategyMap: {
    [key: string]: AuthenticationStrategy_2;
  }): void;
  setClusterSupplier(clusterSupplier?: KubernetesClustersSupplier_2): this;
  setDefaultClusterRefreshInterval(refreshInterval: Duration): this;
  setFetcher(fetcher?: KubernetesFetcher_2): this;
  setObjectsProvider(objectsProvider?: KubernetesObjectsProvider_2): this;
  setProxy(proxy?: KubernetesProxy): this;
  setServiceLocator(serviceLocator?: KubernetesServiceLocator_2): this;
```

etc. I see there are already extension points for a lot of this, not sure about the refresh interval (maybe that's inside the supplier these days). If you think the deletions are OK, I'm happy to go ahead.

mclarke47:
I think I'd need to look into this more deeply in that case

benjdlambert:
@mclarke47 gentle ping :pray:

mclarke47:
I think I might need some help to validate this for our setup

benjdlambert:
So I don't think that this is going to work for some use cases.

I think that it's wise for us to convert some of the extension points into `serviceRefs` instead, which then give us the ability to get access to the default implementations of some of the dependencies without having to export a load of things from the plugin as right now it's impossible to create a `ClusterLocator` outside of the plugin without copying a load of code.

internally we have a requirement that the `ServiceLocator` needs access to the the `ClusterSupplier` which I think is pretty standard, but there's no way to create your own `ServiceLocator` without having access to the implementation of the `ClusterSupplier` but with extension points this is only possible if you create your own `ClusterSupplier` in the same scope.

Moving these to `ServiceRefs` would mean that you could depend on the default implementation of the `ClusterSupplier` when overriding the `ServiceLocator` with your own.

freben:
I'll close this for now - better to revive when we have the time to do the above adjustments!

PR #29018, opened 2025-03-01
scaffolder: use test databases instead of DatabaseManager

freben:
This is on top of #29017.

Just lightly rewrote the tests to use `TestDatabases` instead of the `DatabaseManager` from `@backstage/backend-common`. Ignore whitespace while reviewing. These tests were unfortunately hard coded to only run against sqlite and did not work well with actual databases yet. So I've retained that limitation for now and we can rewrite the tests better in a followup PR. The goal here was to just get rid of backend-common.

Also fixes a small bug in `DatabaseTaskStore` where sometimes the returned count was a string.

### DIFF
--- a/.changeset/little-ways-happen.md
+++ b/.changeset/little-ways-happen.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend': patch
+---
+
+Fixed a bug where the task count could sometimes be a string instead of a number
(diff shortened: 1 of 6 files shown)


freben:
The build error is due to the scaffolder not shutting down fast enough. Will have to look into enabling the graceful shutdown feature in tests.

acierto (approved):

PR #29020, opened 2025-03-01
frontend-plugin-api: add new DialogApi

Rugvip:

This adds a new Utility API in the new frontend system called `DialogApi`. It's purpose is to be able to show dialogs in the React tree, in particular from code that lives outside of the React tree, for example in even handlers for menu or sidebar items.

The intention of this is to help drive additions like #28673 forward, where we need a way for extensions to render dialogs as part of the react tree that aren't otherwise outputting any React components.

The `show` method shows a regular dialog that can be dismissed by clicking the backdrop or escape, while `showModal` shows a dialog that can only be closed by returning a result. I figured that separating those two use-cases into separate functions ended up being the cleanest API, as I don't really se a big need for additional options, at least not ones that are used frequently. An option would of course be `dialogApi.show({ element: ..., modal: true })`, but it just doesn't look quite as nice when formatted x). Happy to discuss that though, and especially if there are any other options that we think might be needed upfront.

I kept this only in the new frontend system for now to leave a bit more room for iteration and avoid needing to change existing apps.

Two simple examples of the usage of this API, taken from the added docs:

Example with inline dialog content:

```tsx
const dialog = dialogApi.showModal<boolean>(
  <DialogContent>
    <DialogTitle>Are you sure?</DialogTitle>
    <DialogActions>
      <Button onClick={() => dialog.close(true)}>Yes</Button>
      <Button onClick={() => dialog.close(false)}>No</Button>
    </DialogActions>
  </DialogContent>
);
const result = await dialog.result();
```

Example with separate dialog component:

```tsx
function CustomDialog({ dialog }: { dialog: DialogApiDialog<boolean> }) {
  return (
    <DialogContent>
      <DialogTitle>Are you sure?</DialogTitle>
      <DialogActions>
        <Button onClick={() => dialog.close(true)}>Yes</Button>
        <Button onClick={() => dialog.close(false)}>No</Button>
      </DialogActions>
    </DialogContent>
  )
}
const result = await dialogApi.showModal(CustomDialog).result();
```

### DIFF
--- a/plugins/app/src/extensions/DialogDisplay.tsx
+++ b/plugins/app/src/extensions/DialogDisplay.tsx
@@ -63,11 +63,13 @@ function DialogDisplay({
           deferred.resolve(result);
           setDialogs(ds => ds.filter(d => d.dialog.id !== id));
         },
-        update(elementOrComponent) {
+        update(ElementOrComponent) {
           const element =
-            typeof elementOrComponent === 'function'
-              ? elementOrComponent({ dialog })
-              : elementOrComponent;
+            typeof ElementOrComponent === 'function' ? (
+              <ElementOrComponent dialog={dialog} />
+            ) : (
+              ElementOrComponent
+            );
           setDialogs(ds =>
             ds.map(d => (d.dialog.id === id ? { dialog, element } : d)),
           );
@@ -76,7 +78,7 @@ function DialogDisplay({
           return deferred;
         },
       };
-      const element = options.component({ dialog });
+      const element = <options.component dialog={dialog} />;
       setDialogs(ds => [...ds, { dialog, element }]);
       return dialog;
     });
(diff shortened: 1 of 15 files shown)


plugins/app/src/extensions/DialogDisplay.tsx
```diff
@@ -0,0 +1,129 @@
...
+          return deferred;
+        },
+      };
+      const element = <options.component dialog={dialog} />;
+      setDialogs(ds => [...ds, { dialog, element }]);
+      return dialog;
+    });
+  }, [dialogApi]);
+
+  if (dialogs.length > 0) {
```
freben, line 87: there may be something to say here about stacking; just rendering the very last (removing the previous one from the jsx tree when a "parent" triggers a child modal) may be too harsh? Not sure, but I believe Dialog itself is stackable?

freben (approved):
Pretty neat how you got the connect in there.

plugins/app/src/extensions/DialogDisplay.tsx
Rugvip: yep not sure tbh, figured to keep it minimal for now and I especially wanted to avoid double backdrops

PR #29029, opened 2025-03-03
workflows/deploy_microsite: handle missing gen-api-docs in stable docs

Rugvip:
🧹, stable docs don't support the `get-api-docs` command yet, so we need to ignore errors due to the missing command

FYI @aramissennyeydd


aramissennyeydd (approved):
Ah interesting, nice find!

freben (approved):
Yep or the even more canonical `||:` (shells are weird)

Rugvip:
@freben I think `|| true` more broadly understood :p

PR #29031, opened 2025-03-03
fix(core-components): Avoid Layout Shift for DismissableBanner when using a storageApi with latency (e.g. user-settings-backend)

GabDug:

Properly handle the `unknown` state of the storageApi, and wait for a value before opening the banner.

There's a tradeoff here: this change may lead to some Layout Shift if the banner has not been dismissed, but once it has been dismissed, you won't have any.

IMO, it's a better user experience than to have a layout shift on every visit of a page, once you have dismissed the banner. This was especially annoying, as if the latency is ~low, you get a flash of information that disappears almost instantly. 

### DIFF
--- a/.changeset/serious-chefs-provide.md
+++ b/.changeset/serious-chefs-provide.md
@@ -1,0 +1,7 @@
+---
+'@backstage/core-components': patch
+---
+
+Avoid Layout Shift for DismissableBanner when using a storageApi with latency (e.g. user-settings-backend)
+
+Properly handle the `unknown` state of the storageApi. There's a trade-off: this may lead to some Layout Shift if the banner has not been dismissed, but once it has been dismissed, you won't have any.
--- a/.changeset/serious-chefs-provide.md
+++ b/.changeset/serious-chefs-provide.md
@@ -2,6 +2,6 @@
 '@backstage/core-components': patch
 ---
 
-Avoid Layout Shift for DismissableBanner when using a storageApi with latency (e.g. user-settings-backend)
+Avoid Layout Shift for `DismissableBanner` when using a `storageApi` with latency (e.g. `user-settings-backend`)
 
-Properly handle the `unknown` state of the storageApi. There's a trade-off: this may lead to some Layout Shift if the banner has not been dismissed, but once it has been dismissed, you won't have any.
+Properly handle the `unknown` state of the `storageApi`. There's a trade-off: this may lead to some Layout Shift if the banner has not been dismissed, but once it has been dismissed, you won't have any.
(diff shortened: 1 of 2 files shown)


.changeset/serious-chefs-provide.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/core-components': patch
+---
+
+Avoid Layout Shift for DismissableBanner when using a storageApi with latency (e.g. user-settings-backend)
+
+Properly handle the `unknown` state of the storageApi. There's a trade-off: this may lead to some Layout Shift if the banner has not been dismissed, but once it has been dismissed, you won't have any.
```
freben, line 7:
```suggestion
Properly handle the `unknown` state of the `storageApi`. There's a trade-off: this may lead to some Layout Shift if the banner has not been dismissed, but once it has been dismissed, you won't have any.
```

packages/core-components/src/components/DismissableBanner/DismissableBanner.tsx
```diff
@@ -102,23 +102,36 @@ export const DismissableBanner = (props: Props) => {
   const classes = useStyles();
   const storageApi = useApi(storageApiRef);
   const notificationsStore = storageApi.forBucket('notifications');
-  const rawDismissedBanners =
-    notificationsStore.snapshot<string[]>('dismissedBanners').value ?? [];
+  const dismissedBannersSnapshot =
+    notificationsStore.snapshot<string[]>('dismissedBanners');
```
freben, line 106: I believe that the right way of consuming the storage API is in one single place via `useObservable`: https://github.com/backstage/backstage/blob/7612daf25a6756da6f09db377584fb8fe8032f69/plugins/home/src/components/CustomHomepage/CustomHomepageGrid.tsx#L107

GabDug: Good catch, thank you!

packages/core-components/src/components/DismissableBanner/DismissableBanner.tsx
```diff
@@ -102,23 +102,34 @@ export const DismissableBanner = (props: Props) => {
...
   const [dismissedBanners, setDismissedBanners] = useState(
-    new Set(rawDismissedBanners),
+    new Set(observedItems.value ?? []),
   );
-
-  const observedItems = useObservable(
-    notificationsStore.observe$<string[]>('dismissedBanners'),
+  const [loadingSettings, setLoadingSettings] = useState(
+    observedItems.presence === 'unknown',
   );
```
freben, line 115: One more thing - this doesn't need a useState at all right? It's only a computed property of `observedItems`. So you could probably remove basically all code below and instead just do like `const loadingSettings = observedItems.presence === 'unknown';` and then use that directly where needed.

GabDug: It's the first approach I've tried, but it still resulted in some layout shift.

I'll see if I can improve this.

freben:
Now it just wants you to remove the unused imports.

freben (approved):

PR #29032, opened 2025-03-03
workflows/deploy_microsite: completely disable stable API doc gen

Rugvip:
Sledgehammer version of #29029


freben (approved):

freben (commented):
could remove the extra yarn install too if you want to speed things up, right?

Rugvip:
eh, if it's gonna be a problem it'll be a problem 😅

PR #29034, opened 2025-03-03
cli: migrate info command do module

Rugvip:

Migrating the `info` command to module in the CLI. (👀  @aramissennyeydd)

Added a fix for the bare `import(...)`s of CommonJS as well, that ends up crashing at runtime because CommonJS compat is amazing. Situation we're in right now is to use `require` for cjs and `import` for esm, except when we have wrappers that enable `import` like we do in the backend system or the `lazy` helper 🤦 

Btw, not loving yargs for this after trying it a bit more. Might do another one and experiment with the one used in Yarn, https://github.com/arcanis/clipanion. @benjdlambert mentioned that one seems pretty good

### DIFF
--- a/packages/cli/src/modules/info/alpha.ts
+++ b/packages/cli/src/modules/info/alpha.ts
@@ -13,20 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import yargs from 'yargs';
+import { createCliPlugin } from '../../wiring/factory';
 
-import { CliInitializer } from './wiring/CliInitializer';
-import chalk from 'chalk';
-
-(async () => {
-  console.warn(
-    chalk.yellow(
-      'THIS ENTRYPOINT IS IN ALPHA AND MAY CHANGE IN THE FUTURE - DO NOT USE THIS FOR NORMAL DEVELOPMENT',
-    ),
-  );
-  const initializer = new CliInitializer();
-  initializer.add(import('./modules/info/alpha'));
-  initializer.add(import('./modules/config/alpha'));
-  initializer.add(import('./modules/build/alpha'));
-  initializer.add(import('./modules/migrate/alpha'));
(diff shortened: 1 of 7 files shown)


packages/cli/src/modules/info/commands/info.ts
aramissennyeydd, line 1: You may need to remove `packages/cli/src/commands/info.ts`, I don't see that removal/rename in the changeset here.

packages/cli/src/modules/info/alpha.ts
```diff
@@ -0,0 +1,33 @@
...
+  pluginId: 'info',
+  init: async reg => {
+    reg.addCommand({
+      path: ['info'],
+      description: 'Show helpful information for debugging and reporting bugs',
+      execute: async ({ args }) => {
+        yargs().parse(args);
+        const { default: command } =
+          require('./commands/info') as typeof import('./commands/info');
+        await command();
```
aramissennyeydd, line 29: 😬, this is ... not great syntax-wise - can we use the `lazy()` wrapper at all here?

aramissennyeydd (approved):
Nice! Only a few commands left 😁

packages/cli/src/modules/info/commands/info.ts
Rugvip: Oh derp, yep 😅

packages/cli/src/modules/info/alpha.ts
Rugvip: Yep but it's not great with `lazy` either:

```ts
const action = await lazy(() => import('./commands/info'), 'default');
await action();
```

Right now `require` really is the most straightforwad way for lazy imports of CommonJS, and I think it'll stay that way untill sync hooks are supported in the new Node.js module system.

freben (approved):

PR #29040, opened 2025-03-04
fix: Catalog entity page header overflow fix

k-parihar:

fixed `CatalogEntityPage` header where the title overflow caused the right-side icons to extend beyond the Header.
### Before
image

### After fix
image

fixes : #28933 

### DIFF
--- a/packages/core-components/src/layout/Header/Header.tsx
+++ b/packages/core-components/src/layout/Header/Header.tsx
@@ -59,6 +59,7 @@ const useStyles = makeStyles(
     },
     leftItemsBox: {
       maxWidth: '100%',
+      width: '80%', // space for the right items
       flexGrow: 1,
     },
     rightItemsBox: {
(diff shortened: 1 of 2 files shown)


MihaiTabara:
Found in triaging - proposing to re-open this. Apologies @k-parihar for having this PR fall through the cracks of review.

packages/core-components/src/layout/Header/Header.tsx
```diff
@@ -59,6 +59,7 @@ const useStyles = makeStyles(
     },
     leftItemsBox: {
       maxWidth: '100%',
+      width: '80%', // space for the right items
```
Rugvip, line 62: Would be best to have the right items take up the content that they need, rather than a fixed with. This'll cause trouble both if more items are added and on more narrow screens

packages/core-components/src/layout/Header/Header.tsx
```diff
@@ -70,6 +71,10 @@ const useStyles = makeStyles(
       wordBreak: 'break-word',
       fontSize: theme.typography.h3.fontSize,
       marginBottom: 0,
+      '& > div > span > span': {
+        display: 'unset',
+        // to fix title overflow issue
+      },
```
Rugvip, line 77: This needs to be applied more directly to the element, the selector is drilling a bit too deep

PR #29049, opened 2025-03-04
Introduce a `createFrontendFeatureLoader`

davidfestal:

Introduced a `createFrontendFeatureLoader()` function, as well as a `FrontendFeatureLoader` interface, to gather several frontend plugins, modules or feature loaders in a single exported entrypoint and load them, possibly asynchronously.
This new feature, very similar to the `createBackendFeatureLoader()` already available on the backend, supersedes the previous `CreateAppFeatureLoader` type which has been deprecated.

### DIFF
--- a/packages/frontend-defaults/report.api.md
+++ b/packages/frontend-defaults/report.api.md
@@ -54,8 +54,7 @@ export function createPublicSignInApp(options?: CreateAppOptions): {
 
 // @public (undocumented)
 export function discoverAvailableFeatures(config: Config): {
-  features: FrontendFeature[];
-  featureLoaders?: FrontendFeatureLoader[];
+  features: (FrontendFeature | FrontendFeatureLoader)[];
 };
 
 // @public (undocumented)
--- a/packages/frontend-defaults/report.api.md
+++ b/packages/frontend-defaults/report.api.md
@@ -7,7 +7,8 @@ import { Config } from '@backstage/config';
 import { ConfigApi } from '@backstage/frontend-plugin-api';
 import { CreateAppRouteBinder } from '@backstage/frontend-app-api';
 import { ExtensionFactoryMiddleware } from '@backstage/frontend-plugin-api';
-import { FrontendFeature } from '@backstage/frontend-app-api';
+import { FrontendFeature } from '@backstage/frontend-plugin-api';
+import { FrontendFeatureLoader } from '@backstage/frontend-plugin-api';
 import { JSX as JSX_2 } from 'react';
 import { default as React_2 } from 'react';
 import { ReactNode } from 'react';
@@ -17,7 +18,7 @@ export function createApp(options?: CreateAppOptions): {
   createRoot(): JSX_2.Element;
 };
 
-// @public
+// @public @deprecated
 export interface CreateAppFeatureLoader {
   getLoaderName(): string;
   load(options: { config: ConfigApi }): Promise<{
@@ -38,7 +39,11 @@ export interface CreateAppOptions {
     | ExtensionFactoryMiddleware
     | ExtensionFactoryMiddleware[];
   // (undocumented)
-  features?: (FrontendFeature | CreateAppFeatureLoader)[];
+  features?: (
+    | FrontendFeature
+    | FrontendFeatureLoader
+    | CreateAppFeatureLoader
+  )[];
   loadingComponent?: ReactNode;
 }
 
@@ -50,12 +55,17 @@ export function createPublicSignInApp(options?: CreateAppOptions): {
 // @public (undocumented)
 export function discoverAvailableFeatures(config: Config): {
   features: FrontendFeature[];
+  featureLoaders?: FrontendFeatureLoader[];
 };
 
 // @public (undocumented)
 export function resolveAsyncFeatures(options: {
   config: Config;
-  features?: (FrontendFeature | CreateAppFeatureLoader)[];
+  features?: (
+    | FrontendFeature
+    | FrontendFeatureLoader
+    | CreateAppFeatureLoader
+  )[];
 }): Promise<{
   features: FrontendFeature[];
 }>;
--- a/packages/frontend-defaults/src/createApp.tsx
+++ b/packages/frontend-defaults/src/createApp.tsx
@@ -19,6 +19,8 @@ import {
   ConfigApi,
   coreExtensionData,
   ExtensionFactoryMiddleware,
+  FrontendFeature,
+  FrontendFeatureLoader,
 } from '@backstage/frontend-plugin-api';
 // eslint-disable-next-line @backstage/no-relative-monorepo-imports
 import { defaultConfigLoaderSync } from '../../core-app-api/src/app/defaultConfigLoader';
@@ -27,7 +29,6 @@ import { overrideBaseUrlConfigs } from '../../core-app-api/src/app/overrideBaseU
 import { ConfigReader } from '@backstage/config';
 import {
   CreateAppRouteBinder,
-  FrontendFeature,
   createSpecializedApp,
 } from '@backstage/frontend-app-api';
 import appPlugin from '@backstage/plugin-app';
@@ -38,6 +39,7 @@ import { resolveAsyncFeatures } from './resolution';
  * A source of dynamically loaded frontend features.
  *
  * @public
+ * @deprecated Use the {@link @backstage/frontend-plugin-api#createFrontendFeatureLoader} function instead.
  */
 export interface CreateAppFeatureLoader {
   /**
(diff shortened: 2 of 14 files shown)


packages/frontend-plugin-api/src/wiring/createFrontendFeatureLoader.ts
```diff
@@ -0,0 +1,83 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ConfigApi } from '../apis/definitions';
+import { FrontendFeature } from './types';
+
+/** @public */
+export interface CreateFrontendFeatureLoaderOptions {
+  name: string;
```
Rugvip, line 22: If possible I'd prefer to skip the name and use an approach similar to this instead: https://github.com/backstage/backstage/blob/45e7b21f665f9095691886d402cf8d4d90c3b3bf/packages/backend-plugin-api/src/wiring/createBackendFeatureLoader.ts#L51

Just looking to avoid requiring users to name things if it's not really needed x)

packages/frontend-plugin-api/src/wiring/createFrontendFeatureLoader.ts
```diff
@@ -0,0 +1,83 @@
...
+  load(deps: { config: ConfigApi }): Promise<{
+    features: FrontendFeature[];
+  }>;
```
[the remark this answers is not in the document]
davidfestal: Done

davidfestal:
@Rugvip I implemented all the changes we discussed during the last Framework SIG call.
Do you think you could give a deeper look, and tell is something else is required to merge ?

packages/frontend-defaults/report.api.md
```diff
@@ -32,6 +32,8 @@ export interface CreateAppOptions {
     config: ConfigApi;
   }>;
   // (undocumented)
+  featureLoaderRecursionDepth?: number;
```
Rugvip, line 35: Meh, I don't think this is worth an option, let's just hardcode it at somewhere around 5-10 and leave it at that?

packages/frontend-defaults/src/createApp.tsx
```diff
@@ -87,29 +94,93 @@ export function createApp(options?: CreateAppOptions): {
...
+      (acc, item) => {
+        if ('$$type' in item) {
+          acc[0].push(item);
+        } else {
+          acc[1].push(item);
+        }
+        return acc;
+      },
+      [[], []],
+    );
```
Rugvip, line 113: Bit of a nit, but I think a `let a; let b; for (...) { if (...) { a.push(...) } else { b.push(...) }` would be a bit easier to read here. Same thing below

packages/frontend-defaults/src/createApp.tsx
```diff
@@ -87,29 +94,93 @@ export function createApp(options?: CreateAppOptions): {
...
+          return acc;
+        },
+        [[], []],
+      );
+      loadedFeatures.push(...otherFeatures);
+      for (const featureLoader of featureLoaders) {
+        if (featureLoaderNames.some(l => l === featureLoader)) {
+          throw new Error(
+            `Added several instances of feature loader ${featureLoader.description}`,
+          );
```
Rugvip, line 155: I think a `continue` here would be fine tbh

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
```diff
@@ -200,6 +202,15 @@ export function createSpecializedApp(options?: {
...
 }): { createRoot(): JSX.Element } {
   const config = options?.config ?? new ConfigReader({}, 'empty-config');
+  const featureLoaders = options?.features?.filter(
+    isInternalFrontendFeatureLoader,
+  );
+  if (featureLoaders && featureLoaders.length > 0) {
+    throw new Error(
+      `FrontendFeatureLoader is an async construct. It should be passed to createApp(), not to createSpecializedApp().`,
+    );
+  }
```
Rugvip, line 212: So this is the kind of issue I was alluding the regarding the new meaning of the `FrontendFeature` type. Tbh, I think we should keep `FrontendFeature` where it is and as it is, and make `createApp` accept `FrontendFeature | CreateAppFeatureLoader | FrontendFeatureLoader` instead. I at least can't think of any reason right now why that wouldn't be enough and I think it's better to avoid this confusion and keep it type safe.

Rugvip (commented):
Thanks! Couple more bits bits but then good to go. Leaving it to be a race with #29058 😅

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
davidfestal: So I guess that the `loader` method of the `FrontendFeatureLoader` itself should then return `FrontendFeature | FrontendFeatureLoader` ?
Having a single Type for anything that can be loaded by the feature loader (including nested feature loaders themselves) is quite useful.
The same remark applies to the discovery `getAvailableFeatures()` method, which we expect to possibly return feature loaders as well.

davidfestal: Is a change in the types returned by `FrontendFeatureLoader.loader` and `getAvailableFeatures ()` the direction you would recommend ?

packages/frontend-defaults/src/createApp.tsx
davidfestal: What would then be the option you would recommend to at least log a warning ?

packages/frontend-defaults/report.api.md
davidfestal: Quite useful for tests though :-)

davidfestal: Which is why I made it optional

packages/frontend-defaults/src/createApp.tsx
davidfestal: Fixed

packages/frontend-defaults/report.api.md
davidfestal: Fixed.

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
davidfestal: @Rugvip Please could you provide your thought here, before I update the PR and rebase it on top of https://github.com/backstage/backstage/pull/29058 ?

Rugvip: yep, essentially any point in the code that requires a sync logic would use `sync`, while any point that can do async as well would have `sync | async`

davidfestal:
@Rugvip I rebased this PR on top of https://github.com/backstage/backstage/pull/29058, and removed the `FrontendFeatureLoader` from the `FrontendFeature` type, in order to restore type-safety related to sync / async application creation.
Since I had fixed all your other comments, that should hopefully be in a good shape for merge, which would allow me to come back to fixing the other review comment on the pending PR #28076 about *__dynamic frontend plugins__*.

davidfestal:
> @Rugvip I rebased this PR on top of #29058, and removed the `FrontendFeatureLoader` from the `FrontendFeature` type, in order to restore type-safety related to sync / async application creation. Since I had fixed all your other comments, that should hopefully be in a good shape for merge, which would allow me to come back to fixing the other review comment on the pending PR #28076 about _**dynamic frontend plugins**_.

cc @iamEAP

packages/frontend-defaults/src/discovery.ts
```diff
@@ -57,6 +60,7 @@ function readPackageDetectionConfig(config: Config) {
  */
 export function discoverAvailableFeatures(config: Config): {
   features: FrontendFeature[];
+  featureLoaders?: FrontendFeatureLoader[];
```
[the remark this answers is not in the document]
davidfestal: Fixed

packages/frontend-defaults/src/resolution.ts
```diff
@@ -16,30 +16,93 @@
...
+    if (featuresOrLoaders.length === 0) {
+      return;
+    }
+
+    const featureLoaders: FrontendFeatureLoader[] = [];
+    for (const item of toLoad) {
+      if (isBackstageFeatureLoader(item)) {
+        featureLoaders.push(item);
+      } else {
+        loadedFeatures.push(item);
```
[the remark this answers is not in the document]
davidfestal: Fixed

davidfestal:
> Two minor bits and then we'll be good to merge 👍

@Rugvip The 2 minor bits are fixed.

davidfestal:
@Rugvip Any idea why the DCO check, which was successful when submitting this PR, is now back in "Expected" ?

packages/frontend-defaults/src/resolution.ts
```diff
@@ -16,30 +16,98 @@
...
+    featuresOrLoaders: (FrontendFeature | FrontendFeatureLoader)[],
+    recursionDepth: number,
+  ) {
+    if (featuresOrLoaders.length === 0) {
+      return;
+    }
+
+    for (const featureOrLoader of featuresOrLoaders) {
+      if (isBackstageFeatureLoader(featureOrLoader)) {
+        if (alreadyMetFeatureLoaders.some(l => l === featureOrLoader)) {
```
Rugvip, line 71: `.includes` would work here instead, or even better make it a `Set` and use `.has`. Merging as is though as this is minor

Rugvip (approved):
Nice, thank you! 👍 

All good, let's :shipit:

Rugvip:
I see DCO on all commits so the check is just stuck, merging

PR #29053, opened 2025-03-05
add catalog entity lifecycle events

acutlip-wss:

I would like be able to write Backstage plugins which react to entity changes using events. To facilitate the creation of such plugins this PR adds a new concept of a "catalog entity lifecycle" event. When entities are upserted or deleted an event containing the action type and a list of entity refs is published.

This functionality is disabled by default and can be enabled via a new catalog configuration option.

**N.B.:** While entity finalization is centralized in the catalog stitcher, entity deletion occurs in many different places. This may be a source of future bugs should additional entity deletion locations be added.


freben:
Hey! I haven't made a full review, but my gut feeling tells me that it's almost impossible to build this entirely based on just node level code. For example, there are several ways that cascading deletes and similar will happen in the database itself and a solution like this will struggle to cover those correctly. In my opinion, ideally these events should even be sent if an operator goes into the database and issues raw sql delete commands.

I've started a branch some time ago that entirely relies on database level triggers instead, so that detected deletions and updates and additions get persisted to an event queue that code can pump onto a message bus after the fact. That solution will be robust against all possible code changes and shenanigans, I felt.

Any thoughts on that idea?

PR #29057, opened 2025-03-05
chore: Move `createExtensionDataContainer()` to internal, shared location

iamEAP:
## What / Why

Have an upcoming PR where I'd like to introduce the concept of an extension factory middleware.  In order to implement this concept, I need access to `createExtensionDataContainer()` in `@backstage/frontend-app-api`, in addition to where it lives today in `@backstage/frontend-plugin-api`.

This PR lifts and shifts the relevant code to `@internal/frontend` so that it can be inlined in both packages.

### DIFF
--- a/packages/frontend-internal/src/wiring/createExtensionDataContainer.ts
+++ b/packages/frontend-internal/src/wiring/createExtensionDataContainer.ts
@@ -16,33 +16,11 @@
 
 import {
   AnyExtensionDataRef,
+  ExtensionDataContainer,
   ExtensionDataRef,
   ExtensionDataValue,
-} from './createExtensionDataRef';
+} from '@backstage/frontend-plugin-api';
 
-/** @public */
-export type ExtensionDataContainer<UExtensionData extends AnyExtensionDataRef> =
-  Iterable<
-    UExtensionData extends ExtensionDataRef<
-      infer IData,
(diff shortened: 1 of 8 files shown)


packages/frontend-plugin-api/report.api.md
```diff
@@ -982,27 +983,7 @@ export interface ExtensionBoundaryProps {
...
-  > & {
-    get<TId extends UExtensionData['id']>(
-      ref: ExtensionDataRef<any, TId, any>,
-    ): UExtensionData extends ExtensionDataRef<infer IData, TId, infer IConfig>
-      ? IConfig['optional'] extends true
-        ? IData | undefined
-        : IData
-      : never;
-  };
+export { ExtensionDataContainer };
```
iamEAP, line 986: This is unfortunate and I was unable to find a quick way for the full type to be reproduced here. Tips welcome.

Rugvip: That it including `@internal/frontend` iirc, but maybe that causes trouble

iamEAP: Yeah, I think it should work.  One sec.

Rugvip (approved):
Looks good! 👍 

Let's :shipit:

PR #29058, opened 2025-03-05
Expose frontend feature discovery and resolution logic used in `createApp`

iamEAP:
## What / Why

When creating a custom app loader built on `createSpecializedApp()`, it can be useful to reuse the logic for discovering and resolving frontend features. This PR proposes wrapping resolution logic in a `resolveAsyncFeatures` function that's exposed by `@backstage/frontend-defaults`, and also renaming and exporting `getAvailableFeatures()` as `discoverAvailableFeatures()`.


TheSabby (approved):

.changeset/mira-color-like.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/frontend-defaults': minor
```
Rugvip, line 2:
```suggestion
'@backstage/frontend-defaults': patch
```

packages/frontend-defaults/report.api.md
```diff
@@ -40,4 +41,10 @@ export interface CreateAppOptions {
 export function createPublicSignInApp(options?: CreateAppOptions): {
   createRoot(): React_2.JSX.Element;
 };
+
+// @public (undocumented)
+export function resolveFeatures(options: {
+  config: Config;
+  providedFeatures?: (FrontendFeature | CreateAppFeatureLoader)[];
```
Rugvip, line 48: let's call this one `features` imo

Rugvip (commented):
Yep 👍, just a few small bits

@davidfestal FYI, just moving things around but your PR will need a rebase on top of this.

davidfestal (commented):
Should this include the move of the related tests which are currently in the `createApp.tsx` file ?
A related question is: should this function be in a separated file ?

packages/frontend-defaults/src/discovery.ts
```diff
@@ -94,3 +97,29 @@ function isBackstageFeature(obj: unknown): obj is FrontendFeature {
   }
   return false;
 }
+
+/** @public */
+export async function resolveFeatures(options: {
+  config: Config;
+  features?: (FrontendFeature | CreateAppFeatureLoader)[];
+}): Promise<FrontendFeature[]> {
+  const discoveredFeatures = getAvailableFeatures(options.config);
```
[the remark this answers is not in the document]
Rugvip: Ah yep for sure, let's split them apart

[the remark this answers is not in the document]
iamEAP: Great! I've kept them separated, split out `resolveFeatures` to its own file and added relevant tests.  ...I kept the equivalent tests in `createApp`, since I think it's still important to exercise those expectations in the higher context.

davidfestal (approved):

packages/frontend-defaults/report.api.md
```diff
@@ -43,4 +44,13 @@ export interface CreateAppOptions {
 export function createPublicSignInApp(options?: CreateAppOptions): {
   createRoot(): React_2.JSX.Element;
 };
+
+// @public (undocumented)
+export function getAvailableFeatures(config: Config): FrontendFeature[];
+
+// @public (undocumented)
+export function resolveFeatures(options: {
```
Rugvip, line 52: let's workshop these names a bit.

For the first one I'm thinking this kinda thing:

- `getDiscoveredFeatures`
- `discoverAvailableFeatures`
- `resolveConfiguredFeatures`

For the second:

- `resolveAsyncFeatures`
- `resolveFeatureLoaders`

For both of them let's wrap up the `FrontendFeature[]` as `{ features: FrontendFeature[] }` too. Both of these are lower level plumbing so we don't need to ergonomic benefits of avoiding that wrapping.

packages/frontend-defaults/report.api.md
```diff
@@ -43,4 +44,17 @@ export interface CreateAppOptions {
 export function createPublicSignInApp(options?: CreateAppOptions): {
   createRoot(): React_2.JSX.Element;
 };
+
+// @public (undocumented)
+export function discoverAvailableFeatures(config: Config): {
+  features: FrontendFeature[];
```
Rugvip, line 50: Oooh, I realized this one should probably support feature loaders too right?

packages/frontend-defaults/report.api.md
iamEAP: Went with `discoverAvailableFeatures()` and `resolveAsyncFeatures()`.

packages/frontend-defaults/report.api.md
iamEAP: Should it?  I don't think it explicitly works right now, since `discoveredFeatures` doesn't get passed through feature loader resolution logic.  I guess we could?  ...But is it event possible for `window['__@backstage/discovered__']` to be populated with async feature loaders?

Rugvip: Yep right now we don't support it, but I'm thinking we'd want to prepare consumers of this function to support it. I'm a little bit torn on whether it's a good idea to discover feature loaders, but it is how it works in the backend system currently so thinking we should stick to that.

davidfestal: I'll do that when rebasing my PR on top of this one.

davidfestal:
@Rugvip @iamEAP I don't really understand the point in this last commit (https://github.com/backstage/backstage/pull/29058/commits/36e387e4028cf0c922a11c687bcc85ea843101a9), since  `CreateAppFeatureLoader` is expected to be deprecated by PR https://github.com/backstage/backstage/pull/29049 and replaced by real `FrontendFeatureLoader`s created by the `createFrontendFeatureLoader()` function, according to what we had discussed during the last Framework SIG.

I'm a bit confused by the fact that the content of this PR seems to overlap quite much with the [already-available PR related to `createFrontendFeatureLoader()`](https://github.com/backstage/backstage/pull/29049) on which you proposed that I would work as a pre-requisite to merge my PR about dynamic frontend plugins...

iamEAP:
Sorry @davidfestal, I haven't been following all of the developments related to feature loaders.  Happy to back out the change related to supporting feature loaders in `discoverAvailableFeatures()` if it makes your change stack easier to manage.  ...Which I've done now.

packages/frontend-defaults/report.api.md
Rugvip: Alright let's add this after

Rugvip (approved):
Alright good to go, thank you! 👍

packages/frontend-defaults/report.api.md
davidfestal: I'm doing it right now while rebasing my PR on top of this one.

PR #29061, opened 2025-03-05
frontend-plugin-api: sort extension definition in plugin type

Rugvip:

Worth it? 😅

Main purpose is to make API reports easier to browse but especially get rid of the flakiness where we currently have where changes in one package can affect the API reports in a another. I think a better fix really would be to implement this with API Extractor instead, i.e. make sure object keys are always sorted, but not quite ready for that adventure. At least if we do that in the future this would be pretty simple to clean up.

Performance wise the `UnionToArray` is a bit slow for large arrays but other operations are fairly fast. A `tsc:full` on this branch takes is `50.11s user 4.63s system 151% cpu 36.150 total`, while before it's `49.69s user 5.56s system 153% cpu 35.908 total`, so no worries there.

### DIFF
--- a/.changeset/stale-frogs-wonder.md
+++ b/.changeset/stale-frogs-wonder.md
@@ -1,0 +1,5 @@
+---
+'@backstage/frontend-plugin-api': patch
+---
+
+The extensions map for plugins created with `createFrontendPlugin` is now sorted alphabetically by ID in the TypeScript type.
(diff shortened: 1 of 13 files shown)


freben (approved):
Uhh sure :) Like, if we hit compiler limits doing this later down the line there might be a bit confusion. I'd say at least add a doc comment on `MakeSortedExtensionsMap` explaining the rationale and outcome.

PR #29068, opened 2025-03-06
Introduce an extension factory middleware

iamEAP:
## What / Why

I believe we'll find a variety of use-cases unlocked if we introduce a way to modify extensions en-masse as they are being instantiated. A naive example of what would be possible:

```tsx
const app = createApp({
  features,
  *extensionFactoryMiddleware(originalFactory, context) {
    const output = originalFactory();
    yield* output;
    const element = output.get(coreExtensionData.reactElement);

    const owner = getOwnerForPlugin(context.node.spec.source?.id);

    if (element) {
      // Associate analytics events within this extension to a plugin "owner"
      yield coreExtensionData.reactElement(
        <AnalyticsContext attributes={{ owner }>
          {element}
        </AnalyticsContext>,
      );
    }
  },
});
```

Another use-case would be swapping out extensions entirely under some circumstances, e.g. for A/B testing (for which access to metadata about the extension tree would be helpful).


packages/frontend-app-api/report.api.md
```diff
@@ -29,10 +32,19 @@ export function createSpecializedApp(options?: {
   features?: FrontendFeature[];
   config?: ConfigApi;
   bindRoutes?(context: { bind: CreateAppRouteBinder }): void;
+  apis?: ApiHolder;
+  extensionFactoryMiddleware?: ExtensionFactoryMiddleware;
 }): {
+  apis: ApiHolder;
   createRoot(): JSX_2.Element;
+  tree: AppTree;
```
Rugvip, line 40: Forgot about this bit and was thinking it could make sense to wrap things up a bit more potentially. Then I had the thought that maybe we should go in the opposite direction, where we only return `{ apis, tree }` here and grab the root from the tree instead. Or going even further, adding `apis` into the `tree` and just returning `{ tree }`

Either way, happy to merge as is but I might tinker a bit with this in a followup.

packages/frontend-app-api/src/tree/instantiateAppNodeTree.ts
```diff
@@ -289,12 +294,30 @@ export function createAppNodeInstance(options: {
...
+      const outputDataValues = options.extensionFactoryMiddleware
+        ? createExtensionDataContainer(
+            options.extensionFactoryMiddleware(
+              ({ config: configOverride } = {}) =>
+                createExtensionDataContainer(
+                  internalExtension.factory(
+                    configOverride
+                      ? {
+                          ...context,
+                          config: configOverride,
```
Rugvip, line 312: The type indicates that that we can override inputs as well right? i.e. https://github.com/backstage/backstage/blob/4c46e99e2eac332e93243e0e656c97487548ac64/packages/frontend-plugin-api/src/wiring/createExtension.ts#L407-L412

I can take care of getting that sorted if you want though.

Rugvip (approved):
Nice! 🎉 

Some optional bits and followup investigation but good to merge for me

PR #29074, opened 2025-03-06
Added workspace:template and workspace:template:file actions to complement respective fetch:* actions

mbenson:

Added `workspace:template` and `workspace:template:file` actions to complement respective fetch:\* actions

### DIFF
--- a/.changeset/heavy-onions-swim.md
+++ b/.changeset/heavy-onions-swim.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend': patch
+---
+
+Added workspace:template and workspace:template:file actions to complement respective fetch:\* actions
--- a/.changeset/heavy-onions-swim.md
+++ b/.changeset/heavy-onions-swim.md
@@ -1,5 +1,5 @@
 ---
-'@backstage/plugin-scaffolder-backend': patch
+'@backstage/plugin-scaffolder-backend': minor
 ---
 
-Added workspace:template and workspace:template:file actions to complement respective fetch:\* actions
+Added `workspace:template` and `workspace:template:file` actions to complement respective `fetch:*` actions
--- a/plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/template.ts
+++ b/plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/template.ts
@@ -131,24 +131,26 @@ export function createFetchTemplateAction(options: {
       },
     },
     supportsDryRun: true,
-    handler: createTemplateActionHandler({
-      resolveTemplate: async ctx => {
-        ctx.logger.info('Fetching template content from remote URL');
+    handler: ctx =>
+      createTemplateActionHandler({
+        ctx,
+        resolveTemplate: async () => {
+          ctx.logger.info('Fetching template content from remote URL');
 
-        const workDir = await ctx.createTemporaryDirectory();
-        const templateDir = resolveSafeChildPath(workDir, 'template');
+          const workDir = await ctx.createTemporaryDirectory();
+          const templateDir = resolveSafeChildPath(workDir, 'template');
 
-        await fetchContents({
-          baseUrl: ctx.templateInfo?.baseUrl,
-          fetchUrl: ctx.input.url,
-          outputPath: templateDir,
-          token: ctx.input.token,
-          ...options,
-        });
+          await fetchContents({
+            baseUrl: ctx.templateInfo?.baseUrl,
+            fetchUrl: ctx.input.url,
+            outputPath: templateDir,
+            token: ctx.input.token,
+            ...options,
+          });
 
-        return templateDir;
-      },
-      ...options,
-    }),
+          return templateDir;
+        },
+        ...options,
+      }),
   });
 }
--- a/plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/template.ts
+++ b/plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/template.ts
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-import { extname } from 'path';
-import { UrlReaderService } from '@backstage/backend-plugin-api';
-import { resolveSafeChildPath } from '@backstage/backend-plugin-api';
-import { InputError } from '@backstage/errors';
+import {
+  resolveSafeChildPath,
+  UrlReaderService,
+} from '@backstage/backend-plugin-api';
 import { ScmIntegrations } from '@backstage/integration';
 import {
   createTemplateAction,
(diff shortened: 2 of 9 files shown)


mbenson:
@reviewers: this is the first pass at new actions to template workspace content, as discussed during March 7 Scaffolder SIG. This could bear discussion on where the new actions, and shared code, should live. Thanks!

stephenglass:
What is the difference between these workspace:* actions versus the fetch actions? As a user, what would be some scenario when I would prefer one over the other?

mbenson:
> What is the difference between these workspace:* actions versus the fetch actions? As a user, what would be some scenario when I would prefer one over the other?

The only difference is that, rather than fetching the Nunjucks template from the location of the scaffolder template, the Nunjucks template is pulled from a specified location already in the scaffolding context workspace. The immediate need driving this is to mitigate service request limits against git hosting services in the case that a given template should be subjected to interpolation multiple times. Rather than calling `fetch:template` in an `each` loop, the template author would be able to do a single `fetch` without interpolation and _then_ use `workspace:template` with `each`.

Additionally, in the case of the proposed `workspace:template` action, the deprecated `copyWithoutRender` option has _not_ been carried over from `fetch:template`.

plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/template.ts
```diff
@@ -147,201 +131,24 @@ export function createFetchTemplateAction(options: {
...
-            );
-          }
-        }
-      }
-
-      ctx.logger.info(`Template result written to ${outputDir}`);
-    },
+    handler: createTemplateActionHandler({
+      resolveTemplate: async ctx => {
+        ctx.logger.info('Fetching template content from remote URL');
```
acierto, line 136: Something like `Fetching template file ${tmpFilePath} content from remote URL ${ctx.input.url}`.

It's very helpful during troubleshooting when something goes wrong

plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/templateActionHandler.ts
```diff
@@ -0,0 +1,272 @@
...
+): {
+  outputDir: string;
+  copyOnlyPatterns?: string[];
+  renderFilename: boolean;
+  extension: string | false;
+} {
+  const targetPath = ctx.input.targetPath ?? './';
+  const outputDir = resolveSafeChildPath(ctx.workspacePath, targetPath);
+
+  if (ctx.input.copyWithoutRender && ctx.input.copyWithoutTemplating) {
```
acierto, line 214: Maybe it would be better to keep all this extensive list of checks in a separate method?

acierto (dismissed):

plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/templateActionHandler.ts
mbenson: Again, this represents code that was migrated pretty much as-is from `fetch:template` but again LMK if you want me to pursue this here.

plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/templateActionHandler.ts
```diff
@@ -0,0 +1,272 @@
...
+  }
+  if (
+    ctx.input.templateFileExtension &&
+    (copyOnlyPatterns || ctx.input.cookiecutterCompat)
+  ) {
+    throw new InputError(
+      'Fetch action input extension incompatible with copyWithoutRender/copyWithoutTemplating and cookiecutterCompat',
+    );
+  }
+  let extension: string | false = false;
```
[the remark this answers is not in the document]
mbenson:
```
          templateFileExtension: {
            title: 'Template File Extension',
            description:
              'If set, only files with the given extension will be templated. If set to `true`, the default extension `.njk` is used.',
            type: ['string', 'boolean'],
          },
```

plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/template.ts
acierto: Hi @mbenson 👋  I see, then yeah, you can ignore it.
It would make sense then to create a new PR to improve logging in general, not related to this PR

mbenson:
not stale

mbenson:
not stale

plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/workspaceTemplateFile.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Templates variables into a single workspace file, placing the result into another location in the workspace.
+ * @public
+ */
+export function createWorkspaceTemplateFileAction(options: {
+  integrations: ScmIntegrations;
+  additionalTemplateFilters?: Record<string, TemplateFilter>;
+  additionalTemplateGlobals?: Record<string, TemplateGlobal>;
+}) {
+  return createTemplateAction<{
```
benjdlambert, line 37: Whilst we're here, can we move this to the new format for a bit of dogfooding?

```ts
createTemplateAction({
  schema: {
    input: {
      sourcePath: (z) => z.string().describe('Path in workspace to source file.')
...
})
```

benjdlambert:
@mbenson I had a go at refactoring a little bit to try and remove some of the typescript inference.

Typescript can't really infer things that come from another context like that pretty well, but thinking that the refactor looks ok? 

Do you have any objections to going in this direction instead?

.changeset/heavy-onions-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend': patch
```
freben, line 2:
```suggestion
'@backstage/plugin-scaffolder-backend': minor
```

plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/workspaceTemplateFile.ts
```diff
@@ -0,0 +1,76 @@
...
+            .optional()
+            .describe('Values to pass to the templating engine.'),
+        cookiecutterCompat: z =>
+          z
+            .boolean()
+            .optional()
+            .describe(
+              'Enable features to maximise compatibility with templates built for fetch:cookiecutter',
+            ),
+        replace: z =>
```
freben, line 58: this feels like it'd have made sense as an enum `ifTargetPathExists: 'ignore' | 'overwrite' | 'error'` instead, but that's too late now perhaps. Same in the other action

freben (approved):

plugins/scaffolder-backend/src/scaffolder/actions/builtin/fetch/workspaceTemplateFile.ts
benjdlambert: yeah maybe something for another breaking change in the future.

.changeset/heavy-onions-swim.md
mbenson: Hi @benjdlambert ... I see you just merged. 😄 I guess you can appreciate that I was attempting in my original approach to avoid the creation of a new handler function for each action invocation, but if you feel all right with it then it's okay with me. I will say that in the current structure the `createTemplateActionHandler` function doesn't really so much _create_ anymore as just perform the action.

benjdlambert: yeah exactly, but appreciate the motivation to keep it clean. Thinking it's fine for now, and if we run into problems happy to revisit the refactor a little bit more.

PR #29090, opened 2025-03-07
bump to typescript 5.6

Rugvip:

Encountered two issues in libraries: `ast-types` and `expires-map`.

For `ast-types` I figured a patch is the best approach, https://github.com/benjamn/ast-types/issues/948 as been open for a long time. We already ship a TS version where this is broken in create-app, so no big impact from patching it here.

For `expires-map` I first tried to switch to `quick-lru`, but it had the exact same issue 😂. Figured a quick lil inline implementation was the way to go instead and got rid of the dep.

### DIFF
--- a/plugins/catalog/src/apis/EntityPresentationApi/DefaultEntityPresentationApi.ts
+++ b/plugins/catalog/src/apis/EntityPresentationApi/DefaultEntityPresentationApi.ts
@@ -28,7 +28,6 @@ import {
 } from '@backstage/plugin-catalog-react';
 import { durationToMilliseconds, HumanDuration } from '@backstage/types';
 import DataLoader from 'dataloader';
-import ExpiryMap from 'expiry-map';
 import ObservableImpl from 'zen-observable';
 import {
   createDefaultRenderer,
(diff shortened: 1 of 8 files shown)


plugins/catalog/src/apis/EntityPresentationApi/DefaultEntityPresentationApi.ts
```diff
@@ -149,6 +148,45 @@ interface CacheEntry {
   entity: Entity | undefined;
 }
 
+// Simply expory map for the data loader, which only expects a map that implements set, get, and delete and clear
```
freben, line 151:
```suggestion
// Simple expiry map for the data loader, which only expects a map that implements set, get, and delete and clear
```

plugins/catalog/src/apis/EntityPresentationApi/DefaultEntityPresentationApi.ts
```diff
@@ -149,6 +148,45 @@ interface CacheEntry {
...
+    return result;
+  }
+
+  get(key: K) {
+    if (!this.has(key)) {
+      return undefined;
+    }
+    const timestamp = this.#timestamps.get(key)!;
+    if (Date.now() - timestamp > this.#ttlMs) {
+      this.delete(key);
```
freben, line 173: not a huge deal, but maybe delete the timestamp too to clean up

freben (approved):

Issue #27384, opened 2024-10-28
🐛 Bug Report: GithubOrgEntityProvider Secondary Rate Limit


Lucas-Desouza:
### 📜 Description

We started hitting secondary rate limits on loading users and teams using the GithubOrgEntityProvider and GithubMultiOrgEntityProvider.

### 👍 Expected behavior

should be able to import org users and teams

### 👎 Actual Behavior with Screenshots

is currently failing and hitting GitHub secondary rate limits.

sample logs 

```
info  Task worker starting: GithubOrgEntityProvider:production:refresh, {"version":2,"cadence":"PT60M","timeoutAfterDuration":"PT15M"}
info  Task worker starting: GithubOrgEntityProvider:production:refresh, {"version":2,"cadence":"PT60M","timeoutAfterDuration":"PT15M"}
info  Task worker starting: GithubOrgEntityProvider:production:refresh, {"version":2,"cadence":"PT60M","timeoutAfterDuration":"PT15M"}
info  Task worker starting: GithubOrgEntityProvider:production:refresh, {"version":2,"cadence":"PT60M","timeoutAfterDuration":"PT15M"}
info  Task worker starting: GithubOrgEntityProvider:production:refresh, {"version":2,"cadence":"PT60M","timeoutAfterDuration":"PT15M"}
info  Reading GitHub users and teams
error  GithubOrgEntityProvider:production refresh failed, HttpError: You have exceeded a secondary rate limit. Please wait a few minutes before you try again. If you reach out to GitHub Support for help, please include the request ID ***. You have exceeded a secondary rate limit. Please wait a few minutes before you try again. If you reach out to GitHub Support for help, please include the request ID ***.
```

### 👟 Reproduction steps

Happens on task run after starting

### 📃 Provide the context for the Bug.

logs provided above.

### 🖥️ Your Environment

We are currently on backstage v1.28.4

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes. We may have a fix locally that we can contribute. Needs some polish.

Lucas-Desouza:
[...]

PR #29112, opened 2025-03-07
ThrottledOctokit everywhere

johnp789:

This updates both GitHubEntityProvider and GitHubMultiOrgEntityProvider to use the ThrottledOctokit client like GitHubOrgEntityProvider currently does.  At least in the the case of GitHubMultiOrgEntityProvider, it has resolved issues with secondary rate limits in local testing.

### DIFF
--- a/plugins/catalog-backend-module-github/src/providers/GithubEntityProvider.ts
+++ b/plugins/catalog-backend-module-github/src/providers/GithubEntityProvider.ts
@@ -31,13 +31,13 @@ import {
 
 import { LocationSpec } from '@backstage/plugin-catalog-common';
 
-import { graphql } from '@octokit/graphql';
 import * as uuid from 'uuid';
 import {
   GithubEntityProviderConfig,
   readProviderConfigs,
 } from './GithubEntityProviderConfig';
 import {
+  createGraphqlClient,
   getOrganizationRepositories,
   getOrganizationRepository,
   RepositoryResponse,
@@ -219,7 +219,7 @@ export class GithubEntityProvider implements EntityProvider, EventSubscriber {
     );
   }
 
-  private async createGraphqlClient() {
+  private async createConfiguredGraphqlClient() {
     const organization = this.config.organization;
     const host = this.integration.host;
     const orgUrl = `https://${host}/${organization}`;
@@ -228,9 +228,10 @@ export class GithubEntityProvider implements EntityProvider, EventSubscriber {
       url: orgUrl,
     });
 
-    return graphql.defaults({
-      baseUrl: this.integration.apiBaseUrl,
+    return createGraphqlClient({
+      baseUrl: this.integration.apiBaseUrl!,
       headers,
+      logger: this.logger,
     });
   }
 
@@ -238,7 +239,7 @@ export class GithubEntityProvider implements EntityProvider, EventSubscriber {
   private async findCatalogFiles(): Promise<Repository[]> {
     const organization = this.config.organization;
     const catalogPath = this.config.catalogPath;
-    const client = await this.createGraphqlClient();
+    const client = await this.createConfiguredGraphqlClient();
 
     const { repositories: repositoriesFromGithub } =
       await getOrganizationRepositories(client, organization, catalogPath);
@@ -599,7 +600,7 @@ export class GithubEntityProvider implements EntityProvider, EventSubscriber {
     if (this.config.validateLocationsExist) {
       const organization = this.config.organization;
       const catalogPath = this.config.catalogPath;
-      const client = await this.createGraphqlClient();
+      const client = await this.createConfiguredGraphqlClient();
 
       const repositoryFromGithub = await getOrganizationRepository(
         client,
--- a/.changeset/dirty-bugs-attend.md
+++ b/.changeset/dirty-bugs-attend.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-github': patch
+---
+
+Use ThrottledOctokit in GithubMultiOrgEntityProvider and GithubEntityProvider.
--- a/plugins/catalog-backend-module-github/src/providers/GithubEntityProvider.test.ts
+++ b/plugins/catalog-backend-module-github/src/providers/GithubEntityProvider.test.ts
@@ -38,6 +38,7 @@ import {
 
 jest.mock('../lib/github', () => {
   return {
+    createGraphqlClient: jest.fn(),
     getOrganizationRepositories: jest.fn(),
   };
 });
--- a/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.test.ts
+++ b/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.test.ts
@@ -22,7 +22,7 @@ import {
   DefaultEventsService,
   EventsService,
 } from '@backstage/plugin-events-node';
-import { graphql } from '@octokit/graphql';
+import { createGraphqlClient } from '../lib/github';
 import {
   GithubMultiOrgEntityProvider,
   withLocations,
@@ -30,7 +30,10 @@ import {
 import { LoggerService } from '@backstage/backend-plugin-api';
 import { mockServices } from '@backstage/backend-test-utils';
 
-jest.mock('@octokit/graphql');
+jest.mock('../lib/github', () => ({
+  ...jest.requireActual('../lib/github'),
+  createGraphqlClient: jest.fn(),
+}));
 
 const getAllInstallationsMock = jest.fn();
 jest.mock('@backstage/integration', () => ({
@@ -53,7 +56,7 @@ describe('GithubMultiOrgEntityProvider', () => {
 
     beforeEach(() => {
       mockClient = jest.fn();
-      (graphql.defaults as jest.Mock).mockReturnValue(mockClient);
+      (createGraphqlClient as jest.Mock).mockReturnValue(mockClient);
 
       entityProviderConnection = {
         applyMutation: jest.fn(),
@@ -191,7 +194,7 @@ describe('GithubMultiOrgEntityProvider', () => {
           },
         });
 
-      (graphql.defaults as jest.Mock).mockReturnValue(mockClient);
+      (createGraphqlClient as jest.Mock).mockReturnValue(mockClient);
 
       await entityProvider.read();
 
@@ -479,7 +482,7 @@ describe('GithubMultiOrgEntityProvider', () => {
           },
         });
 
-      (graphql.defaults as jest.Mock).mockReturnValue(mockClient);
+      (createGraphqlClient as jest.Mock).mockReturnValue(mockClient);
 
       const githubCredentialsProvider: GithubCredentialsProvider = {
         getCredentials: mockGetCredentials,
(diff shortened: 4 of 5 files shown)


johnp789:
I have backed away from trying to introduce throttling for the REST usages of Octokit, because I couldn't find a way to get all the dependencies aligned.  This branch running locally is able to ingest the users, groups, and repos from a fairly large GitHub organization.

packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.ts
```diff
@@ -365,30 +366,46 @@ export class GithubUrlReader implements UrlReaderService {
...
   ): Promise<Response> {
     const urlAsString = url.toString();
-    const response = await fetch(urlAsString, init);
-
-    if (!response.ok) {
-      let message = `Request failed for ${urlAsString}, ${response.status} ${response.statusText}`;
+    let attempts = 0;
+    const maxAttempts = 5;
+    let response;
+    while (attempts < maxAttempts) {
```
camilaibs, line 372: I'd like to hear @backstage/maintainers' thoughts on whether we should keep a retry logic here, use a `node-fetch` equivalent that supports retries or another preferable approach here.

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -760,10 +801,17 @@ export const createGraphqlClient = (args: {
     },
   });
 
-  const client = octokit.graphql.defaults({
-    headers,
```
camilaibs, line 764: Do you know how headers were used before? I just wanted to understand the effects of removing them.

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -714,44 +749,50 @@ export const createReplaceEntitiesOperation =
...
+ * Creates a Octokit Client with Throttling
  */
-export const createGraphqlClient = (args: {
-  headers:
-    | {
-        [name: string]: string;
-      }
-    | undefined;
+export const createOctokitClient = (args: {
+  token: string;
```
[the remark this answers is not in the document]
johnp789: The Octokit README says, "Authentication is optional for some REST API endpoints accessing public data, but is required for GraphQL queries," and so far, this is only used for GraphQL queries, but it is possible that someday `createOctokitClient` could be used to create a client for public REST API requests.  I'll make `token` optional for `createOctokitClient` but required for `createGraphqlClient`.

plugins/catalog-backend-module-github/src/lib/github.ts
johnp789: From what I can tell, `headers` was only used to supply the token in the form of an `Authorization: Bearer <token>` header.  Plain Octokit core.js wants the token itself passed as `auth`, rather than a header.

plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts
```diff
@@ -497,16 +500,18 @@ export class GithubOrgEntityProvider implements EntityProvider {
       throw new Error('Not initialized');
     }
 
+    const logger = this.options.logger;
     const organizationTeamTransformer =
       this.options.teamTransformer || defaultOrganizationTeamTransformer;
     const { name, html_url: url, description, slug } = event.team;
     const org = event.organization.login;
-    const { headers } = await this.credentialsProvider.getCredentials({
+    const { token } = await this.credentialsProvider.getCredentials({
```
camilaibs, line 508: It seems like the token should be required for this to work (as per [this](https://github.com/backstage/backstage/pull/29112#discussion_r2023136425) comment). Is there some implicit configuration we request from the adopter now (and we should write documentation for)? Should we also throw an explicit error when the token is undefined?

johnp789: I've pushed a commit which switches back to getting headers, not tokens, like the original.

camilaibs:
> I can't see how my last commit would cause the E2E Linux checks to fail---is it possible that the failures are caused by some kind of outage in the yarn package repository? I don't see an option to re-run the failed checks, other than pushing a placeholder commit.

These seem like flaky tests. I'm re-triggering the jobs to see if they succeed this time.

freben:
Heads up that we are looking at doing the inverse of this in #29584 - throttling and retries at the low level has severe implications in holding up the processing queue and leading to knock-on problems elsewhere. So while I'm not saying to close this PR, we'll be holding off and thinking about it a bit

johnp789:
> Heads up that we are looking at doing the inverse of this in #29584 - throttling and retries at the low level has severe implications in holding up the processing queue and leading to knock-on problems elsewhere. So while I'm not saying to close this PR, we'll be holding off and thinking about it a bit

Without throttling, the Backstage GitHub catalog provider fails for us with rate-limit errors every time `getOrganizationRepositories` is called, even on the first run of the provider on startup, so a longer interval does not help.  Once the GitHub principal used by Backstage has hit a rate limit, every GitHub API call the backend makes fails until the rate-limit cooling-off period has elapsed.  For example, manually refreshing an entity in the UI fails if the rate limit has been exceeded recently.  Any GitHub API requests associated with processing a GitHub webhook event would fail, too.

GitHub's documentation has [some guidance](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#handle-rate-limit-errors-appropriately) on handling rate-limiting, and the Octokit throttling plugin is said to implement these best practices.

Rugvip:
@johnp789 throttling to avoid secondary (burst) rate-limits is alright. The solution to hitting primary rate limits is to make fewer requests. For the catalog that can be done by introducing more caching or increasing processing intervals, and in order to counteract the potentially degraded user experience of that we recommend using event-based updates. Throttling is a potential solution, but in practice it doesn't work very well with the current catalog implementation so it's something we want to avoid.

johnp789:
> @johnp789 throttling to avoid secondary (burst) rate-limits is alright. The solution to hitting primary rate limits is to make fewer requests. For the catalog that can be done by introducing more caching or increasing processing intervals, and in order to counteract the potentially degraded user experience of that we recommend using event-based updates. Throttling is a potential solution, but in practice it doesn't work very well with the current catalog implementation so it's something we want to avoid.

Increasing processing intervals to 24 hours still fails every time for our 25,000+ repository GitHub organization, so increasing processing intervals alone does not work for us.  In my local testing, this branch is able to complete an ingestion pass, and I don't notice any user experience degradation, so I'm interested to learn what specifically I should look for there.

benjdlambert:
I wonder if it makes sense to add this rate limiting to the `Providers` only, and nothing that touches the processing loop, as that can be controlled by the `processingInterval`.

I guess what you're running into here is most likely the initial ingestion or run, rather than rate limits caused by the Catalog.

@johnp789 does that align with what you're seeing?

johnp789:
> I wonder if it makes sense to add this rate limiting to the `Providers` only, and nothing that touches the processing loop, so not the `UrlReader` and not the `Processors`, as that can be controlled by the `processingInterval`.
> 
> I guess what you're running into here is most likely the initial ingestion or run, rather than rate limits caused by the Catalog.
> 
> @johnp789 does that align with what you're seeing?

Anytime `getOrganizationRepositories` or `getOrganizationTeams` is called, either on the initial ingestion or later periodic provider runs, we encounter rate limits and then every GitHub API call fails, whether it is in a provider, URL reader, or a processor.  As far as I can tell, the changes in this PR do not slow anything down, unless we are already approaching GitHub rate limits or encountering retryable errors.  I'd like to check for any usability issues caused by running this branch, but I haven't seen anything so far.

By the way, GitHub's documentation has a warning: "Continuing to make requests while you are rate limited may result in the banning of your integration."  I'm concerned that we could end up in this situation.

MarcFogleman:
Im gonna echo @johnp789  sentiment. We have an 8000+ repo org and the current github plugin flat out does not work due to rate limits. The filter in the config doesn't actually reduce the number of calls, it just filters results after the fact. Our initial solution was to create a custom github integration that let us do some very opinionated changes to how the GH API is call, but was hacky and gross. Our whole team has been very excited for this PR.

brentg-telus:
We've been waiting for this feature for almost 2 years. My company has over 10000 repositories and we've had to put together a hacky solution as well. Why not offer both solutions? Keep the current provider/processor as is and give us a secondary rate-limit friendly alternative?

dlaird-ovo:
I will also confirm that 6K+ repos we hit the Github API rate limits and we can increase the processign interval - but each time it scans the entire repo list we hit the limit and then all other GitHub integrations break as the rate limit has been hit. 
Because all the filtering is done 'post' the actual query - we cant use topics , or other means to pre-reduce the list.

freben:
For clarity for those following along, there's some ongoing discussion in https://github.com/backstage/backstage/issues/27384

johnp789:
@freben, I've removed the contentious bit in the URL reader.  My local testing with this looked good, although I suspect it is still possible to have the URL reader encounter rate-limiting errors which aren't handled.

plugins/catalog-backend-module-github/src/providers/GithubEntityProvider.ts
```diff
@@ -469,8 +474,6 @@ export class GithubEntityProvider implements EntityProvider, EventSubscriber {
    * @param event - The repository archived event.
    */
   private async onRepoArchived(event: RepositoryArchivedEvent) {
-    if (this.config.filters.allowArchived) return;
```
freben, line 472: Why was this removed?

plugins/catalog-backend-module-github/src/providers/GithubEntityProvider.ts
```diff
@@ -552,7 +555,7 @@ export class GithubEntityProvider implements EntityProvider, EventSubscriber {
    *
    * Creates new entities for the repository if it matches the filters.
    *
-   * @param event - The repository transferred event.
+   * @param event - The repository unarchived event.
```
freben, line 558: Ah this should be changed back :)

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -632,22 +638,48 @@ export async function queryWithPaging<
...
   let cursor: string | undefined = undefined;
   for (let j = 0; j < 1000 /* just for sanity */; ++j) {
-    const response: Response = await client(query, {
-      ...variables,
-      cursor,
-    });
+    let attempts = 0;
+    const maxRetries = 5;
+    let response: Response | undefined = undefined;
+    while (attempts < maxRetries) {
```
freben, line 659: I thought octokit had a builtin retry plugin to use?

packages/backend-defaults/report-urlReader.api.md
```diff
@@ -316,6 +316,7 @@ export class GithubUrlReader implements UrlReaderService {
       treeResponseFactory: ReadTreeResponseFactory;
       credentialsProvider: GithubCredentialsProvider;
     },
+    logger?: LoggerService | undefined,
```
freben, line 319: Can you move this into `deps` instead? This is a common two-argument form that we use a lot, with one primary thing and the rest in an options object

freben (commented):
Alright, I'm terribly sorry this took so long to get to, but here we are!

PR #29114, opened 2025-03-07
feat(catalog-backend-module-gitlab): added option to disable polling

### LINKED ISSUE #28954: 🚀 Feature: GitlabDiscoveryEntityProvider - When events enabled, disable polling
### 🔖 Feature description

https://github.com/backstage/backstage/blob/b5a82087a7164879ca6c6717329b3e29cc20c3dd/plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts#L148

When events are enabled and the module available, the provider automatically subscribes to the `gitlab.push` topic and processes events. If we have the events packages installed and configured, it would be good to disable the polling mechanism.

I'm more than happy to do the work on this myself but open to thoughts / ideas on how the config should look, if a valid scenario.

### 🎤 Context

On certain GitLab instances, we're seeing the APIs being rate-limited (HTTP 429s) from the refresh / poll mechanism consuming the available requests, given our instance has 1,000s of repositories we're moving to the event based provider instead as it's not feasible to continuously scan all repositories, some of which are not active anymore.

TL;DR:- Being rate limited on GitLab API calls.

Whilst it's possible to automatically infer this, it might be safer to explicitly add a configuration option to choose between the schedule poll vs events as the data source.

Or an additional flag to disable the scheduled refresh / polling if not needed.

I could try not providing the scheduler config so it didn't run every X period but there is a check to throw an Error if not provided so didn't bother testing it.

jneate:

New configuration option added to disable polling, when enabled the push event is more robust for modified files by performing a delta add mutation before calling refresh.

This should resolve #28954 

### DIFF
--- a/.changeset/thirty-bags-invite.md
+++ b/.changeset/thirty-bags-invite.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-gitlab': minor
+---
+
+Added new configuration option to disable polling which when enabled turns off the schedule and makes processing modified files more robust by performing a delta add mutation before triggering the catalog refresh.
--- a/docs/integrations/gitlab/discovery.md
+++ b/docs/integrations/gitlab/discovery.md
@@ -137,7 +137,7 @@ to the catalog configuration.
 
 :::note Note
 
-If you are using the New Backend System, the `schedule` has to be setup in the config, as shown below.
+If you are using the New Backend System, the `schedule` has to be setup in the config (unless disable polling is true), as shown below.
 
 :::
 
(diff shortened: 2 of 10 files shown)


jneate:
Hi - is there any update on the review of the PR?

jneate:
@friederbluemle any chance of some eyes on this?

jneate:
any chance of a review on this?

freben:
There's no need for a separate flag. You can define a schedule as

```yaml
schedule:
  frequency:
    trigger: manual
  timeout: 1h  # doesn't matter really
```

and then it never runs any scheduled ingestion.

But in general we recommend still enabling these ingestions, only at a lower frequency. Since events can be lossy, it's good to every now and then run a "healing" scan that can correct for any errors in the deltas.

I'll close the PR again since there's an existing alternative solution.

PR #29129, opened 2025-03-10
root: bump to TS 5.6 + 5.8

iamEAP:
## What / Why

Been a minute since we've bumped TS.


Rugvip:
🙈 https://github.com/backstage/backstage/pull/29090

iamEAP:
Great minds!

PR #29137, opened 2025-03-10
Add Canon styles

cdedreuille:
We are adding Canon to the different instances of Backstage as well as improving our local development flow with a new watch mode. We are installing Canon in these packages:
- packages/app
- packages/app-next
- packages/create-app


.changeset/common-candies-kiss.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/create-app': minor
```
Rugvip, line 2: `patch` for this one, we only consider the CLI command itself when it comes to stability of `create-app`

.changeset/common-candies-kiss.md
```diff
@@ -0,0 +1,7 @@
...
+'example-app-next': minor
```
Rugvip, line 3: This package is not published so no need to include

.changeset/common-candies-kiss.md
```diff
@@ -0,0 +1,7 @@
...
+'@backstage/canon': minor
```
Rugvip, line 4: Best have a separate changeset for canon that explains the changes there instead, the new CSS layout etc.

packages/app/src/App.tsx
```diff
@@ -86,6 +86,8 @@ import {
   UserNotificationSettingsCard,
 } from '@backstage/plugin-notifications';
 
+import '@backstage/canon/css/styles.css';
```
Rugvip, line 89: Just to discuss this a bit, how do you feel about importing here vs `src/index.tsx`?

packages/app/src/App.tsx
```diff
@@ -86,6 +86,8 @@ import {
...
+
```
Rugvip, line 90: Bit of a weird one, but we also need to add this to `src/index-public-experimental.tsx` in both `app` and `app-next`.

packages/canon/css/box.css
Rugvip, line 1: Hmm, I'm thinking it's best if we skip minification of the CSS at this point, leaving it in the more readable form instead. Bundlers later on will take care of that anyway, and it avoids the need for source maps that bloat the package.

packages/canon/scripts/build-css.mjs
```diff
@@ -58,23 +58,53 @@ if (!fs.existsSync(distDir)) {
...
+const watchMode = args.includes('--watch');
+
+async function buildCSS(logs = true) {
+  // Bundle and transform files
+  cssFiles.forEach(file => {
+    let { code: bundleCode } = bundle({
+      filename: file.path,
+    });
+
+    let { code, map } = transform({
```
Rugvip, line 72: We could do this kinda thing to prettify the code after transform: https://github.com/backstage/backstage/blob/547e41da5ac497e5a606c08c8fb57b429687d5f7/packages/repo-tools/src/commands/api-reports/common/tryRunPrettier.ts#L25-L31

That was we can keep the CSS pretty and avoid the prettier ignore

Rugvip (commented):
Main bits for tomorrow are the changesets and imports, the CSS build process we can iterate on after the release too, as long as it's functional ofc

.changeset/common-candies-kiss.md
cdedreuille: Ok this is now a patch.

.changeset/common-candies-kiss.md
cdedreuille: Ok I removed it.

.changeset/common-candies-kiss.md
cdedreuille: Done

packages/app/src/App.tsx
cdedreuille: You're right it could be installed in `index.tsx`. I tried it and it works.

packages/app/src/App.tsx
cdedreuille: Done

packages/canon/css/box.css
cdedreuille: I'm not sure what would be the benefit at this point. If we want to avoid adding the map files we can just remove them but I would keep the minification.

cdedreuille: Ok I updated the script and remove all sourcemap for now. It might not even be needed. I kept the file minified still.

packages/canon/scripts/build-css.mjs
cdedreuille: The idea is to keep it minified in production. The CSS is quite extensive so this will help loading time. I'm not sure we will ever have to actually read the file directly. The minified process doesn't change any names so if anyone want to read it they can prettify it pretty easily.

packages/canon/css/box.css
Rugvip: can you remove the minification as well? makes it easier to read the css

packages/canon/scripts/build-css.mjs
Rugvip: it will be minified in production later on when the app is built anyway, we don't ship minimized JS code either

packages/canon/css/box.css
cdedreuille: Done

packages/canon/scripts/build-css.mjs
cdedreuille: Ok. I removed the minify step

Rugvip (approved):

PR #29149, opened 2025-03-11
Paraphrasing few sentences

binishma-ss:
Paraphrasing few sentences


Parsifal-M:
Hey 👋 

I am not sure I see the benefit of this? Maybe you can explain why this is better than the original?

Thanks!

freben:
Agreed - it's probably less useful to make pull request changing things that are "a matter of taste" rather than actual errors.

PR #29153, opened 2025-03-11
theme: install Canon theme be default

Rugvip:

This adds the `@backstage/canon` theme as a default theme in `UnifiedThemeProvider`, in addition to adding the `cssLinks` and `themeDataAttribute` props.

It also adds support for `?raw` imports in our WebPack config, as documented here: https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax

One issue I ran into is that `packages/canon/css` is not available on checkout. We'll need to move over be committed to the repo for this to work well, with a CI check to make sure it doesn't go stale.

### DIFF
--- a/.changeset/chilled-shrimps-yell.md
+++ b/.changeset/chilled-shrimps-yell.md
@@ -1,0 +1,7 @@
+---
+'@backstage/theme': minor
+---
+
+Added the baseline theme from `@backstage/canon` to `UnifiedThemeProvider`, in addition to the new `cssLinks` and `themeDataAttribute` props.
+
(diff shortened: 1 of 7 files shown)


Rugvip:
Moving forward with #29137 instead, reasoning being that it's a lot less complicated to have all styles available when loading the app, rather than needing to have them added through the theme provider. With MUI our current setup was a bit more alright because you still had the base styling and could add any additional customizations needed for app loading/error screens manually. With canon the CSS is required and the page will be broken without it, so let's make sure it's always there.

If anyone has a different use-case for `?raw` I'm happy to consider merging that separately.

PR #29154, opened 2025-03-11
fix(techdocs): prevent to throw an error when the SCM provider is not supported

aladjadj:

At present, the plugin lacks support for multiple Source Control Management (SCM) providers within a single Backstage instance when one of these providers is not supported. When encountering an unsupported SCM in the edit_url, the plugin generates an error.

This Pull Request (PR) addresses this issue by dividing the component. The new structure prevents the plugin from loading when the Techdocs are sourced from an unsupported repository. This approach ensures smoother operation and better handling of diverse SCM scenarios within Backstage.

### 🐛 Issue
Error
TypeError

Message
Failed to execute 'containsNode' on 'Selection': parameter 1 is not of type 'Node'.

Stack Trace
TypeError: parameter 1 is not of type 'Node'.
    at https://backstage.example.com/static/module-backstage.b2517ed5.js:618:1729
    at ul (https://backstage.example.com/static/module-react-dom.b4fe5cf3.js:16:24313)
    at Ct (https://backstage.example.com/static/module-react-dom.b4fe5cf3.js:16:42455)
    at qs (https://backstage.example.com/static/module-react-dom.b4fe5cf3.js:16:34578)
    at Ln (https://backstage.example.comstatic/module-react-dom.b4fe5cf3.js:24:1590)
    at MessagePort.Zn (https://backstage.example.comstatic/module-react-dom.b4fe5cf3.js:24:1980)

### DIFF
--- a/plugins/techdocs-module-addons-contrib/src/ReportIssue/ReportIssueContent.tsx
+++ b/plugins/techdocs-module-addons-contrib/src/ReportIssue/ReportIssueContent.tsx
@@ -110,11 +110,7 @@ export const ReportIssueAddonContent = ({
     mainContent!.prepend(feedbackContainer);
   }
 
-  if (
-    !selection ||
-    !selection.containsNode(mainContent, true) ||
-    selection.containsNode(feedbackContainer, true)
-  ) {
+  if (!selection || !selection.containsNode(mainContent, true)) {
     return null;
   }
 
--- a/plugins/techdocs-module-addons-contrib/src/ReportIssue/ReportIssueContent.tsx
+++ b/plugins/techdocs-module-addons-contrib/src/ReportIssue/ReportIssueContent.tsx
@@ -78,7 +78,7 @@ export const ReportIssueAddonContent = ({
     if (
       !selection ||
       !selection.containsNode(mainContent, true) ||
-      selection?.containsNode(feedbackContainer, true)
+      selection.containsNode(feedbackContainer, true)
     ) {
       return;
     }
--- a/plugins/techdocs-module-addons-contrib/src/ReportIssue/ReportIssueContent.tsx
+++ b/plugins/techdocs-module-addons-contrib/src/ReportIssue/ReportIssueContent.tsx
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2022 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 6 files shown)


squid-ney:
Hi @aladjadj , I noticed that your code change has to do with the text selection aspect of the `ReportIssue` addon, but the changeset describes it as fixing an error related to the `edit_uri`. Could you explain a bit more about in what scenario you get the UI error?

aladjadj:
> I noticed that your code change has to do with the text selection aspect of the `ReportIssue` addon, but the changeset describes it as fixing an error related to the `edit_uri`. Could you explain a bit more about in what scenario you get the UI error?

Hello @squid-ney, yes, of course, we have a Gerrit instance as SCM. Gerrit can now use a direct URL to edit content (cf 1), but it is not supported by ReportIssue. If we activate edit_url and/or edit_url_template in the mkdocs, the add-on adds a trace error at the bottom of the page (cf 2) when we click on the page content.

> 1
> image

> 2 
> image

squid-ney:
Hmm, I don't think this change will fix your issue. The previous check `!selection ||` should cover the case of `selection` being undefined/null which would cause the code to return and the `.containsNode` would not get evaluated.  Also, the error seems to be about the `mainContent` parameter since the error reads `parameter 1 is not of type 'Node'`. I think that this is more likely an edge case around `mainContent` not being defined or not being the expected type. 🤔

aladjadj:
Thx for your analyse, Let me check more tomorrow, and trying to add tests ton catch the error.

plugins/techdocs-module-addons-contrib/src/ReportIssue/ReportIssueContent.tsx
```diff
@@ -0,0 +1,131 @@
...
+    });
+  }, [selection, mainContent, feedbackContainer]);
+
+  if (!selection) return null;
+
+  if (!feedbackContainer) {
+    feedbackContainer = document.createElement('div');
+    feedbackContainer.setAttribute('id', ADDON_FEEDBACK_CONTAINER_ID);
+    mainContent!.prepend(feedbackContainer);
+  }
```
tcardonne, line 113: Opportunistic suggestion to fix some issue kind of related:

```suggestion
  if (!feedbackContainer) {
    feedbackContainer = document.createElement('div');
    feedbackContainer.setAttribute('id', ADDON_FEEDBACK_CONTAINER_ID);
    mainContent!.prepend(feedbackContainer);
  }

  if (
    !selection ||
    !selection.containsNode(mainContent, true) ||
    selection.containsNode(feedbackContainer, true)
  ) {
    return null;
  }
```

We want to early return when there's no selection or the selected text isn't in the mainContent area, just like in the useEffect.

This avoids displaying the component when we select something in the navbars:

image

It also fixes a bug if you select something in the content, then something in the nav bar, the tooltip appears at the "old" location

image

tcardonne (approved):

aladjadj:
@tcardonne, the suggestion code introduces test failure due to the zone where the `fireSelectionChangeEvent` triggers the change. Let's try to fix that together next week.

tcardonne (approved):

squid-ney:
Thanks for your contribution! Could you update the changeset to be more reflective of the changes in this PR? Here is the contributing guidance for writing changesets: [Writing changesets](https://github.com/backstage/backstage/blob/618ad32279717e700fedce8f7fed4ed23888617b/CONTRIBUTING.md?plain=1#L186)

squid-ney (approved):
🚀

PR #29160, opened 2025-03-11
Catalog: Implement EntityContextMenuItemBlueprint to extend the entity page's context menu

marknotfound:
Accidentally blew up this branch, new PR here: https://github.com/backstage/backstage/pull/29515


marknotfound:
@Rugvip I've integrated the new dialog API with the context menu item blueprint - let me know if this looks like what you had in mind.

plugins/catalog/src/alpha/pages.tsx
```diff
@@ -73,6 +73,9 @@ export const catalogEntityPage = PageBlueprint.makeWithOverrides({
       EntityContentBlueprint.dataRefs.filterExpression.optional(),
       EntityContentBlueprint.dataRefs.group.optional(),
     ]),
+    extraContextMenuItems: createExtensionInput([
```
freben, line 76: I didn't expect this one to be named "extra". I'd expect that context menu items instead be made first class citizens and that the plugin ships with a few predefined ones (that also get turned into extensions that use that exact blueprint). That way, the end user can rearrange and remove items in any way that they see fit.

plugins/catalog/src/alpha/pages.tsx
```diff
@@ -73,6 +73,9 @@ export const catalogEntityPage = PageBlueprint.makeWithOverrides({
...
+      coreExtensionData.reactElement,
```
freben, line 77: Would it make sense for this type to instead have like a title and an optional icon? And then perhaps a callback onclick function so to speak.

plugins/catalog/src/alpha/pages.tsx
```diff
@@ -89,14 +92,18 @@ export const catalogEntityPage = PageBlueprint.makeWithOverrides({
...
+
         type Groups = Record<
           string,
           { title: string; items: Array<(typeof inputs.contents)[0]> }
         >;
 
         const header = inputs.header?.get(
           EntityHeaderBlueprint.dataRefs.element,
-        ) ?? <EntityHeader />;
+        ) ?? <EntityHeader extraMenuItems={extraMenuItems} />;
```
freben, line 106: then this would probably be more something along the lines of `<EntityHeader contextMenuItems={items} />` where the props of the header has `contextMenuItems?: ContextMenuItem[] | undefined` and where if it's passed in, the default behaviour that it used to have (of rendering a couple of default ones in a fixed set) is disabled, and instead exactly the given items are rendered.

Bonus points for making dividers a possible type of menu item somehow :)

.changeset/proud-dots-fry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-catalog-react': minor
+'@backstage/plugin-catalog': minor
+---
+
+Adds `EntityContextMenuItemBlueprint` to enable extending the entity page's context menu
```
camilaibs, line 6: As we do not have a documentation for this yet, could you please add usage examples here?

plugins/catalog/src/alpha/pages.tsx
camilaibs: +1

plugins/catalog-react/src/alpha/blueprints/EntityContextMenuItemBlueprint.tsx
```diff
@@ -0,0 +1,104 @@
...
+    }
+  | {
+      title: string;
+      icon: JSX.Element;
+      href: string;
+    };
+
+/** @alpha */
+export type FactoryDialogParams = {
+  dialogLoader: () => Promise<
```
camilaibs, line 49: Did you have a chance to discuss this parameter with @Rugvip? I'm curious if you guys considered an `onClick` parameter that would receive the apis?

plugins/catalog-react/src/alpha/blueprints/EntityContextMenuItemBlueprint.tsx
```diff
@@ -0,0 +1,104 @@
...
+  kind: 'entity-context-menu-item',
+  attachTo: { id: 'page:catalog/entity', input: 'extraContextMenuItems' },
+  output: [coreExtensionData.reactElement],
+  *factory(params: EntityContextMenuItemParams, { node, apis }) {
+    const loaderFactory = () => {
+      if ('loader' in params) {
+        return params.loader;
+      }
+
+      if ('dialogLoader' in params) {
```
camilaibs, line 73: Have you tried having a `useOnClick` instead? I'm not sure if it's possible but I believe it would open up the possibility of doing more than just opening modals on click.

camilaibs (commented):
Thanks again for opening this pull request @marknotfound 🙏🏻

plugins/catalog-react/src/alpha/blueprints/EntityContextMenuItemBlueprint.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+      icon: <span>Test</span>,
+    },
+    {
+      title: 'TestDialog',
+      dialogLoader: async () => () => <div>test dialog</div>,
+      icon: <span>Test</span>,
+    },
+  ];
+
+  it.each(data)('should return an extension with sane defaults', params => {
```
camilaibs, line 47: Missing tests interacting with the menu item, rendering it and checking if all params are being used.

plugins/catalog/src/alpha/pages.tsx
marknotfound: @freben @camilaibs I've gone ahead and renamed this to `contextMenuItems` and reimplemented the "Copy Entity URL" menu item as an extension, but I'm unsure how of a good way to actually use that extension which does not

A. Add a duplicate "Copy Entity URL" menu item to the menu in the new FE system. One would appear from the extension, and one would appear from the hardcoded `<MenuItem />` in `EntityContextMenu.tsx`
or
B. Remove the existing menu item from the legacy FE system. If we remove the hardcoded `<MenuItem />` then this works fine in the new system, but removes the menu item from the legacy system.

Any thoughts on how to approach this? Is there some way to detect if the NFS is in use and optionally render the hardcoded menu item if not?

plugins/catalog-react/src/alpha/blueprints/EntityContextMenuItemBlueprint.tsx
marknotfound: I've moved to the useOnClick approach you described and it seems to me we don't need to pass apis to it since hooks can be called directly from `useOnClick` as needed

plugins/catalog-react/report-alpha.api.md
```diff
@@ -381,6 +419,29 @@ export type EntityPredicateValue =
       $contains: EntityPredicateExpression;
     };
 
+// @alpha (undocumented)
+export type FactoryDialogParams = {
+  useOnClick: () => React_2.MouseEventHandler<HTMLLIElement>;
```
Rugvip, line 424: Hmm, this is a quite interesting pattern. It locks us into React a bit more and of course prevents us from being able to access the value directly, but that's probably alright. It's really convenient to be able to use hooks. It should also let us skip the `useHref` variant I'm thinking, recommending `useNavigate` instead.

One thought I had is that it might be better to group these together a bit more, think `useProps`. There could be shared state between these hooks that makes it more convenient to define it as `() => { onClick: MouseEventHandler, title: string, disabled?: boolean }` instead?

I am a bit torn on the general concept though, in reality there shouldn't really be that many hooks to use. We rely more heavily on Utility APIs for global app comms and that could be what we go for here as well. That said we do have a few important contexts that might make this worthwhile, `useEntity` and `useRouteRef` are both very tricky to translate to Utility APIs and much more natural as hooks. In the end I'm thinking we should try this hook pattern in some form as the benefits likely outweigh the downsides.

Side note: We prolly want `HTMLElement` here rather than `HTMLLIElement` since that's unnecessarily specific, or maybe it even makes sense to remove altogether and just make it a `() => void`. I would be a fan of making it `() => Promise<void>` too, since being able to do async work can be very useful, as long as we handle that properly in the implementation.

camilaibs: > It should also let us skip the useHref variant I'm thinking, recommending `useNavigate` instead.

How to decide when to use a link or a button for a menu item in this case? Semantically, it seems more appropriate to use links for navigation and a button if you want to open a dialog, for example. Maybe we can infer this from the return of `useNavigate` function, if it returns a string, we use a link and the string as the path to navigate or if it returns a function, we use a button and the function as the click callback...

I prefer the suggestion of having a `useProps` so that it returns any props we want to pass to the component.

I think it can accept an object or a function that returns the props object. And if we don't want to have the "use" prefix in the blueprint parameter name, maybe something like this would work:

```tsx

EntityContentMenuItemBlueprint.make({
  name: 'my-item',
  params: {
    props: {
      // ....
    }
  }
});
```
or 
```
EntityContentMenuItemBlueprint.make({
  name: 'my-item',
  params: {
    props: function useProps() {
      // the hook code goes here.
    }
  }
});
```

marknotfound: I do agree a `useProps` style hook is a cleaner API here. I had the same thought after realizing I needed to implement `useIsDisabled`, but didn't want to change up the whole API just before another round of reviews. If we go with @camilaibs latter approach (`params: { useProps: (): MenuItemProps => {} }`) it gives us a lot of flexibility to use an `onClick` handler, `href`, or anything else. It also exposes the rest of the `MenuItem` props. Does it make sense to expose all of those props or is it better to restrict the surface to the common use cases (onClick, disabled, href) ?

marknotfound:
@camilaibs @Rugvip I've implemented the `useProps` approach in the latest commit (https://github.com/backstage/backstage/pull/29160/commits/72cc755e8547157bb4174783365812817669a31b), PTAL

plugins/catalog-react/report-alpha.api.md
```diff
@@ -381,6 +423,29 @@ export type EntityPredicateValue =
...
+      useHref: () => string;
+      useIsDisabled?: () => boolean;
+    }
+  | {
+      useTitle: () => string;
+      icon: React_2.JSX.Element;
+      href: string;
+      useIsDisabled?: () => boolean;
+    };
+
```
camilaibs, line 448: These parameters are no longer needed with useProps, right?

marknotfound: Correct, thanks for flagging. I've removed them in the latest commit.

plugins/catalog/src/components/EntityContextMenu/EntityContextMenu.tsx
```diff
@@ -142,37 +145,55 @@ export function EntityContextMenu(props: EntityContextMenuProps) {
...
+                }}
+              >
+                <ListItemIcon>
+                  <FileCopyTwoToneIcon fontSize="small" />
+                </ListItemIcon>
+                <ListItemText
+                  primary={t('entityContextMenu.copyURLMenuTitle')}
+                />
+              </MenuItem>
+            </>
```
camilaibs, line 188: I believe you don't need to render the default items here because you are already providing them as default context menu item extensions.

marknotfound:
Thanks again for your review @camilaibs ! I think I've addressed everything in the [latest commit](https://github.com/backstage/backstage/pull/29160/commits/657389a3d43aae1c6d9a9b3b705e84e9562a3b89). Please let me know what you think.

plugins/catalog/src/components/EntityContextMenu/EntityContextMenu.tsx
camilaibs: Ah, you're right, I thought we duplicated this component so we could introduce changes without having to worry about backwards compatibility with the legacy system. If I'm not mistaken, @Rugvip recommended me to do this when I was working on Entity groups and header extensions. Should we duplicate this one too @Rugvip?

camilaibs (approved):
Looks good to me 🎉 . Thanks again for addressing all the comments @marknotfound 🙏🏻

plugins/catalog/src/components/EntityContextMenu/EntityContextMenu.tsx
Rugvip: It depends, whatever is simplest without impacting the new or old design. In this case I think this is likely the way to go.

plugins/catalog-react/report-alpha.api.md
```diff
@@ -74,6 +74,23 @@ export const catalogReactTranslationRef: TranslationRef<
...
+// @alpha (undocumented)
+export const contextMenuItemComponentDataRef: ConfigurableExtensionDataRef<
+  ContextMenuItemComponent,
+  'catalog.contextMenuItemComponent',
+  {}
+>;
+
+// @alpha (undocumented)
+export type ContextMenuItemProps = {
+  onClose: () => void;
```
Rugvip, line 91: Hmm, this one is a bit awkward, it'd really be something like `onItemClick` or such right? and then that triggers the menu to close. I wonder if we could perhaps handle this with a React context instead? So that we can make these plain React elements with the core data ref instead. The context would be private and consumed by the item wrapper in the blueprint. It'd need to use our singleton utility as there might be duplicates of `catalog-react` installed, see for example https://github.com/backstage/backstage/blob/65454876fb21ec1a8cda1dc556aa1f2af3d874fd/plugins/catalog-react/src/hooks/useEntity.tsx#L36-L38

plugins/catalog-react/report-alpha.api.md
```diff
@@ -74,6 +74,23 @@ export const catalogReactTranslationRef: TranslationRef<
...
+export type ContextMenuItemComponent = (
+  props: ContextMenuItemProps,
+) => React_2.JSX.Element;
+
+// @alpha (undocumented)
+export const contextMenuItemComponentDataRef: ConfigurableExtensionDataRef<
+  ContextMenuItemComponent,
+  'catalog.contextMenuItemComponent',
+  {}
+>;
```
Rugvip, line 87: We've got a pattern where you can expose these via `dataRefs` of the blueprint instead. Let's move to that if we end up needing this data ref.

plugins/catalog-react/src/alpha/blueprints/EntityContextMenuItemBlueprint.tsx
```diff
@@ -0,0 +1,97 @@
...
+  *factory(params: EntityContextMenuItemParams, { node }) {
+    const loader = async (): Promise<ContextMenuItemComponent> => {
+      return ({ onClose }) => {
+        const { title, ...menuItemProps } = params.useProps();
+        let handleClick = undefined;
+
+        if ('onClick' in menuItemProps) {
+          handleClick = () => {
+            const result = menuItemProps.onClick();
+            if (result instanceof Promise) {
```
Rugvip, line 76: `instanceof` is kinda sketchy due to separate JavaScript contexts having their own instances and causing this check to fail. Best switch this to `'then' in result`. Alternatively we could just `await` here too.

That said it does make me think about the actual behavior here, what should actually happen while waiting for the promise? Maybe my initial thought on making the callback async was silly and we should always just close the menu? Otherwise if we want to keep it open I feel we'd need to at least disable all buttons, and maybe showing some form of progress indicator if it takes time.

plugins/catalog/src/alpha/contextMenuItems.tsx
```diff
@@ -0,0 +1,137 @@
...
+          }
+        }, [copyState, alertApi, t]);
+
+        return {
+          title: t('entityContextMenu.copyURLMenuTitle'),
+          onClick: async () => {
+            copyToClipboard(window.location.toString());
+          },
+        };
+      },
```
Rugvip, line 67: Yep I kinda like this pattern. I wonder if we should allow all props to be defined both at top-level and `useProps` 🤔, we can leave that for later iteration though as I think we need a bit of use before it can reach it's final form

Rugvip (commented):
Yep I like 👍 

Main bit before merge is to switch to context for the `onClose`, although if you promise a followup we could ship this in today's `next` 😁

.changeset/proud-dots-fry.md
marknotfound: Added!

PR #29175, opened 2025-03-12
module: move misc repo tasks into a maintenance module

aramissennyeydd:

One of the last modularization pieces. This one shows a few signs of commands not being in quite the right place as we have cross module imports starting to appear. For packing for example, that could either be a build or maintenance task, and I could see arguments for either way. My goal with this PR is to do the initial migration, moving the commands around afterwards should be straightforward.


freben (approved):
yep we can always tweak the structure after

Issue #28818, opened 2025-02-12
🚀 Feature: Scaffolder actions to support rollback


drodil:
### 🔖 Feature description

Add optional rollback possibility to scaffolder actions. This rollback function can in some cases revert the changes made by the action, for example closing/deleting a pull request or deleting a created repository, etc. Of course, it's not possible to rollback all actions such as removing files thus this functionality depends on the action.

### 🎤 Context

Sometimes when running scaffolder templates, the template can create some resources but fail in the next step. This can cause unnecessary and unwanted resources to be left alive. 

### ✌️ Possible Implementation

Add a new optional `rollback` async callback to `TemplateActionOptions` that receives the same input as `handler`. The rollback functions should be called from the `scaffolder` in reverse order of the steps if the run fails starting from the failed action. Another option is to allow the user to manually click `rollback` from the scaffolder UI - I think this should be configurable.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

kurtaking:
Bumping some by saying that I am interested in this functionality.

benjdlambert:
There's a BEP already started for this, but it kind of fizzled out a little bit. https://github.com/backstage/backstage/tree/master/beps/0006-scaffolder-action-rollback

Looking for someone to help drive design a little bit and build ontop of the existing BEP :pray:

kurtaking:
> There's a BEP already started for this, but it kind of fizzled out a little bit. https://github.com/backstage/backstage/tree/master/beps/0006-scaffolder-action-rollback
> 
> Looking for someone to help drive design a little bit and build ontop of the existing BEP 🙏

hey @benjdlambert 👋🏼 I've opened https://github.com/backstage/backstage/pull/29196 to help with this effort.

kurtaking:
I've made enough progress in #29196 to start soliciting feedback.

drodil:
Thanks @kurtaking ! I wrote some comments to the PR.

kurtaking:
Recently added a possible solution to ensuring a more secured rollback context.

drodil:
@kurtaking any chance to get this finalized? We have yet another use-case for this :D

vrabbi:
Really looking for this as well

Issue #23104, opened 2024-02-20
🚀 Feature: Scaffolder Clean Up


jpdeburgo:
### 🔖 Feature description

You should add a way to clean up the resources that are created by the scaffolder if one of the steps fails and is unable to run through successfully.

### 🎤 Context

My organization uses the scaffolder to create GitHub repos, various Azure resources, and Kubernetes deployments, among many other things. Sometimes, one of the steps fails. When this happens the user has to go to all the resources that were created by the scaffolder and delete them manually before attempting to rerun the scaffolder with all of the same input values.

### ✌️ Possible Implementation

- Keep track of all the tasks that were run successfully 
     - I believe this is already being done with the `TaskManager`
- Make a "catch" or "onFail" function for each built-in action 
     - It may be easier to make users implement these if they want this capability since there are a bunch of built-in actions
- Give custom actions the ability to have a "catch" or "onFail" function
- When an action fails, go through all of the completed actions and run the "catch" or "onFail" function
     - The "catch" or "onFail" function can be optional, so it will only run if it is there
     - Or there can be a value passed into the template that will run these "catch" or "onFail" functions when a task fails 

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

jpdeburgo:
> Would some of this be covered by BEP 0004: https://github.com/backstage/backstage/tree/master/beps/0004-scaffolder-task-idempotency?

Yes. Some cases would be covered, but there are other cases when it would not be covered. I think it would depend on what caused the action to fail. If the action failed because of a bug in the template, user input, or some other reason that will cause the action to fail every time then BEP 0004 would not cover it. I think that this should be added to BEP 0004, though. I think it would make sense to try to rerun a failed action a certain number of times and if it never runs successfully then it should run the cleanup functions for each action.

Lucas-Desouza:
> Would some of this be covered by BEP 0004: https://github.com/backstage/backstage/tree/master/beps/0004-scaffolder-task-idempotency?

[...]

Issue #22660, opened 2024-02-02
🚀 Feature: Allow output for failed templates


drodil:
### 🔖 Feature description

When a template fails in scaffolder run for some reason, allow the template to output something like a link to reporting an issue or something else.

### 🎤 Context

When the template run fails, there's no information for the user on what to do and who to contact. Allowing output to be shown also for failed templates could help with this. 

### ✌️ Possible Implementation

Add new possibility to templates:

```yaml
failOutput:
  links:
    - ...
  text:
    - ...
```

or add new property to existing output 

```yaml
output:
  links:
    - title: Title
      url: https://github.com/issues
      onFailure: true
  text:
    - ...
```

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

benjdlambert:
I think that this also is another use case of have `if:` in the outputs part. Similar to how Github actions run things on failure @acierto.

Think that this solves a few things here. Solves people wanting to be able to run steps even if something fails, and potentially showing outputs with something like this:

```yaml
if: 
  failure: true
```

or something like this?

drodil:
@benjdlambert thanks, this sounds fine for me as long as we can output something when the template fails 👍

drodil:
Ohh, and this could be used also to `if` other things in the output for example `if: {{ steps.myStep.output.isDone }}`, or would it? 😄

benjdlambert:
Yeah exactly. it's pretty flexible if we copy some of that approach for sure.

drodil:
Not stale

drodil:
Not stale

drodil:
Not stale

drodil:
Not stale, still something that would be needed, @acierto please reopen, thanks!

dfl-aeb:
Still relevant imo. Could we reopen it?

Issue #11460, opened 2022-05-11
Software Templates: Rollback previous steps in case of an error


omerfarukdogan:
Steps should have a way to rollback their changes if a subsequent step fails during the task.

## Feature Suggestion

Currently, steps are not informed about errors in subsequent steps. Therefore, there is no way to rollback the changes applied in a step if it is completed successfully but a subsequent step fails.

## Possible Implementation

Introduce a new field (such as `errorHandler`) in `TemplateAction`. If a step fails, `errorHandler` functions of previous actions are called by the scaffolder in reverse order.

## Context

We have a template which has multiple steps. Each step creates entities in external systems. When a step fails, these systems become inconsistent. We need a way to inform the completed steps that the task is failed so they can rollback their changes.

benjdlambert:
Hmm not sure that we're looking to introduce this sort of complexity as yet to the scaffolder. What are the things that you are seeing that is failing? Are they meant to fail? Could you not retry them in the action rather than breaking the flow of the scaffolder and exiting when maybe retrying the thing in the action if it might fail could work around this?

omerfarukdogan:
Hello @benjdlambert,

In our case, we have to notify an external system whenever a repository is created. The external system could be unavailable, or have an error which requires manual intervention. Retrying until it is successful may not be a good idea in such situations.

Is Software Templates planned to be limited to repository creation only? Because there could be more steps which are required for a new application to be available, such as provisioning a VM, creating a namespace on Kubernetes, or creating a database.

jhaals:
I agree with Ben that rollbacks introduce a fair bit of complexity and that it require careful consideration before implementation, probably something that require an RFC with more details on how such a system could work.

Software templates can definitely be used to create other things then just repositories as you and others already do.
There's definitely pitfalls with rollbacks which can lead to some big security issues if the scaffolder rollback mean deletion of the resource, some form of takeover. Poorly written rollback logic could lead to deletion of already existing resources prior to scaffolding.

PR #29196, opened 2025-03-12
[BEP 0006] Scaffolder Action Rollback (via checkpoints)

kurtaking:

I've added additional information to the Scaffolder Action Rollback BEP. The goal here is to help drive this to completion.

### Questions to answer

- [x] What happens when a single rollback function fails?
- [x] How are tasks which provide rollback functions tracked?
- [x] What order are rollback functions executed in?
- [x] How can someone writing a template disable a rollback function for a given action?
- [x] What security issues are introduced if the scaffolder rollback means the deletion of the resource?
- [x] Can rollback logic be prevented from deleting already existing resources prior to scaffolding?
- [x] Should rollbacks be an _opt-in_ only approach, rather than requiring the template writer to set `rollback: false`?
- [x] What does a runtime validation mechanism to ensure rollbacks only operate on resources that were actually created by the action look like?

### Outstanding RFC

- [x] Are there experimental or upcoming "scaffolder v2" work that would conflict with this?

### Related issues

- https://github.com/backstage/backstage/issues/28818
- https://github.com/backstage/backstage/issues/11460
- https://github.com/backstage/backstage/issues/23104
- https://github.com/backstage/backstage/issues/22660


beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -69,16 +225,25 @@ This section should describe the rollout process for any new features. It must t
...
+   - Describe the rollback feature
+   - Provide usage examples
+   - Explain best practices
+
 ## Dependencies
 
-<!--
-List any dependencies that this work has on other BEPs or features.
--->
+No known dependencies at this time.
```
kurtaking, line 245: todo: are there any experimental features or scaffolder v2 details that could be dependencies?

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -31,36 +32,191 @@ Introducing the rollback to scaffolder actions provides the mean to come back to
...
+In case of a failure, the workflow runner will invoke the rollback function for each of the completed actions.
+
+```ts
+for (const { action, ctx, step } of [
+  ...this.completedActionsWithRollback,
+].reverse()) {
+  if (action.rollback) {
+    // ...
+    taskLogger.info(`Rolling back step ${step.id} (${step.name})`);
+    await action.rollback(ctx);
```
drodil, line 194: Just wondering what happens if the rollback itself fails? Should it be ignored or does it stop the next rollbacks as well?

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -31,36 +32,191 @@ Introducing the rollback to scaffolder actions provides the mean to come back to
...
+- Modifying existing actions to implement rollback (this proposal only provides the framework)
+- Adding automatic rollback triggers for all possible error scenarios
 
 ## Proposal
 
 Rollback is going to be performed:
 
-- when user manually decide to perform this action.
-- task has to be recovered and TaskRecoverStrategy set to 'rollback'
+- when a user manually decide to perform this action.
```
drodil, line 53: Does this mean there can be a button on the task page to rollback a completed task even if it has finished?

kurtaking: I'm thinking it would be a rollback option (button) presented to the end user only when a template failure occurs. We might not want to start by providing the end user the ability to do this though.

beps/0006-scaffolder-action-rollback/README.md
kurtaking: I lean towards having the worker complete as many of the rollbacks as possible, so if one fails, we continue, and give a report at the end. 

In our use case, if we create 5 resources, I'd rather get 4/5 cleaned up rather than 0/5 if one was to fail.

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -31,36 +32,191 @@ Introducing the rollback to scaffolder actions provides the mean to come back to
...
+Mitigate the issue of manual clean up from partially created resources during template execution failures. When a template execution fails after creating a number of resources but before completing, the system is left in an inconsistent state. Currently, there is no standardized way to clean up these partially created resources, leading to orphaned resources in external systems requiring manual cleanup.
 
 ### Goals
 
-- We will extend action's API to be able to run rollback for failed tasks
-- Rollback will be optional
+- Provide a standardized, optional rollback mechanism for scaffolder actions
+- Enable proper cleanup of resources created in third-party systems when template execution fails
+- Improve the developer experience by reducing manual intervention for cleanup after failures
+- Allow for targeted rollbacks of specific actions rather than only full template rollbacks
```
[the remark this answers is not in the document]
kurtaking: Are you referring to the installed actions [page](https://demo.backstage.io/create/actions)? Having a visual indicator the action supports rollback would definitely be beneficial.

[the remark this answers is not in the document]
drodil: Yes, exactly that 👍

beps/0006-scaffolder-action-rollback/README.md
drodil: Agree. Or maybe this could be configurable and default to that behaviour? But anyways, it probably should be part of this BEP

kurtaking: I've updated the content to be more explicit about what happens when a single rollback fails and mentioned the idea of overriding with something like `dangerouslyEnableRollback`.

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -31,36 +32,191 @@ Introducing the rollback to scaffolder actions provides the mean to come back to
...
+    taskLogger.info(`Rolling back step ${step.id} (${step.name})`);
+    await action.rollback(ctx);
+    taskLogger.info(`Successfully rolled back step ${step.id} (${step.name})`);
+
+    // ...
+  }
+}
+```
+
+### Disable rollback for a step
```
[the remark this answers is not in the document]
kurtaking: Updated to take an opt-in approach.

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -69,16 +326,26 @@ This section should describe the rollout process for any new features. It must t
...
+- Possibly experimental task execution checkpoints ([docs](https://backstage.io/docs/features/software-templates/writing-custom-actions/#using-checkpoints-in-custom-actions-experimental))
 
 ## Alternatives
 
-<!--
-What other approaches did you consider, and why did you rule them out? These do
-not need to be as detailed as the proposal, but should include enough
-information to express the idea and why it was not acceptable.
--->
+- Create a separate action that is called as a parameter to the template step
```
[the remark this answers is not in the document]
kurtaking: I've added a reference in the PR description. I may copy the section into the BEP once finalized.

beps/0006-scaffolder-action-rollback/README.md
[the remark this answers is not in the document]
kurtaking: I've update the goals list to include this.

beps/0006-scaffolder-action-rollback/README.md
drodil: Good job, makes sense!

kurtaking:
👋🏼 @acierto - I'm tagging you because I came across your name in several discussions related to scaffolder rollbacks. I thought you might be interested in taking a look at the PR.

kurtaking:
I plan on spending time learning more about the checkpoints, but from initial discovery, we should be able to leverage the feature to track rollbacks. It will provide idempotency to the rollback handler and preserve state between restarts. That opens up possibilities for functionality such as conditional rollbacks based on the resource state.

drodil (commented):
Just couple of notes, looking good!

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -31,36 +37,281 @@ Introducing the rollback to scaffolder actions provides the mean to come back to
...
+
+### Securing Rollbacks
+
+We will provide a helper function that encapsulates the logic of verifying the resource and executing the rollback. Core security checks can be performed in this function to ensure the resource is
+valid and can be rolled back. There is a possible opportunity to allow others to extend this function to provide custom security checks.
+
+By default, we can handle the following:
+
+1. Verify we reached the point of setting output
+1. Parse resource details from the output and validate them against the input
```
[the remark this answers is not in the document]
kurtaking: The syntax I am using handles incremental numbering the list for me. Second table row in these [docs](https://www.markdownguide.org/basic-syntax/#ordered-lists).

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -31,36 +37,281 @@ Introducing the rollback to scaffolder actions provides the mean to come back to
...
+    try {
+      // 4. Execute the actual rollback with the validated resource
+      ctx.logger.info(`Rolling back ${resource.type} with id ${resource.id}`);
+      await action.rollback(ctx);
+      ctx.logger.info(`Successfully rolled back ${resource.type} with id ${resource.id}`);
+    } catch (error) {
+      ctx.logger.error(
+        `Failed to rollback ${resource.type} with id ${resource.id}: ${error}`,
+      );
+      ctx.logger.info('Continuing with next rollback...');
```
acierto, line 133: This part might be configurable. Maybe by using a different strategy on a task level.

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -31,36 +37,281 @@ Introducing the rollback to scaffolder actions provides the mean to come back to
...
+valid and can be rolled back. There is a possible opportunity to allow others to extend this function to provide custom security checks.
+
+By default, we can handle the following:
+
+1. Verify we reached the point of setting output
+1. Parse resource details from the output and validate them against the input
+1. Execute the actual rollback with the validated resource
+
+```typescript
+export function createSecureRollback
```
acierto, line 84: I think this method can be simplified in terms of input.
Tasks should be idempotent, and for that we use checkpoints.
I'm slowly adding them into built-in actions, you can see them in any of gitlab actions and some of github's too.

The results of checkpoints are already stored in the database.
If we need something extra for rollback, i'd store it there.

and even maybe we can do rollback in the scope of checkpoint?

Some of the actions are huge, like for example 'github:environment:create'.

There we create multiples things: users, policy tags, policy branches, secrets, etc.

If we will keep rollback to the checkpoint, it would be cleaner to see what to rollback.

Like an atomic stuff to create/delete.

acierto: @benjdlambert  wdyt?

kurtaking: I spent a small amount of time looking into checkpoints. I'm happy[ I was on the right track](https://github.com/backstage/backstage/pull/29196#issuecomment-2725844405). I will definitely spend time diving in on this. Looking forward to the feedback from others.

beps/0006-scaffolder-action-rollback/README.md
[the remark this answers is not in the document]
awanlin: I don't like, lol. Just find that when you are working in your editor or looking at the raw Markdown it leads to confusion.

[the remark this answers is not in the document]
kurtaking: Ha, but then you have to redo the numbering if you reorder the list 🥲

beps/0006-scaffolder-action-rollback/README.md
kurtaking: Following up on some other feedback - this would likely be configurable at a `TaskRecoverStrategy` level.

beps/0006-scaffolder-action-rollback/README.md
kurtaking: @acierto - are you referring to the `TaskRecoverStrategy` [type](https://github.com/backstage/backstage/blob/403bb43d3dbcbe35ddc23a6530eddb7ceabcff9f/plugins/scaffolder-common/src/TaskSpec.ts#L54)?

kurtaking: I'm now leaning towards this being a separate mechanism. Recovery and rollbacks are related, but they are distinct concerns.

acierto: I was thinking more of creating a new one `TaskRollbackStrategy`:`StopOnFailure`, `ContinueOnFailure`, more cases?.

benjdlambert:
Sorry - not had chance to have a look at this yet, but it's on my radar! Will try and get some time over the next few days. I'll be out at the Scaffolder SIG tomorrow I think as I'm getting on a boat to Finland, but could be a good place to bring this up and have a chat about it :pray:

drodil:
@benjdlambert welcome to Finland! 🤗

benjdlambert:
@drodil thanks! Good to be back! I'm gonna be bowling at Talihalli - gonna be great 😂

kurtaking:
> Sorry - not had chance to have a look at this yet, but it's on my radar! Will try and get some time over the next few days. I'll be out at the Scaffolder SIG tomorrow I think as I'm getting on a boat to Finland, but could be a good place to bring this up and have a chat about it 🙏

> @benjdlambert welcome to Finland! 🤗

slight flex, no big deal 🤣 have fun!

kurtaking:
I've revised the implementation to focus on leveraging checkpoints rather than handling at the action level. Would love another round of feedback from others.

kurtaking:
@acierto, I opened a [PR](https://github.com/backstage/backstage/pull/29353) that aims to improve type safety around checkpoints. The goal is to make it a bit cleaner before enhancing with rollbacks.

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -1,84 +1,303 @@
...
+    const repository = await githubClient.repos.create({ owner, repo });
+    return {
+      repositoryId: repository.id,
+      owner,
+      repo,
+    };
+  },
+  async data => {
+    // Rollback function using saved checkpoint data
+    const { owner, repo } = data;
```
benjdlambert, line 166: I wonder if it's worth calling out this with an example of using the `data` here, I was a little confused that this was like the initial input to the action, and not the response of the `fn`.

Maybe just add like `ctx.logger('Removing repository, { id: data.repository.id }')` or something?

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -1,84 +1,303 @@
...
+): Promise<T>;
+```
+
+This allows the checkpoint to save data during forward execution that can be used to properly rollback the operation if needed.
+
+Example usage:
+
+```typescript
+// Repository creation with rollback
+await ctx.checkpoint(
```
benjdlambert, line 153: This needs updating right? Shouldn't it be something like:

```ts
ctx.checkpoint({
  key: 'repository-created',
  fn: () => {...},
  rollback: () => {..},
})
```

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -1,84 +1,303 @@
...
+The checkpoint API will be extended to include an optional rollback function that can utilize the data saved by the checkpoint:
+
+```typescript
+/**
+ * Extended Checkpoint API with rollback support
+ */
+async checkpoint<T>(
+  key: string,
+  fn: () => Promise<T>,
+  rollback?: (data: T) => Promise<void>
```
benjdlambert, line 143: I think it might make sense to wrap this up in an object too

```ts
rollback?: (opts: { data: T }) => Promise<void>
```

Makes it easier to evolve moving forward without breaking the API.

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -1,84 +1,303 @@
...
+
+- Rollbacks are executed in proper LIFO order across all actions
+- Each rollback operation is properly logged and tracked
+- **Failures in one rollback don't prevent other rollbacks from being attempted**
+
+```typescript
+// Pseudo-code for rollback execution
+for (const checkpoint of completedCheckpointsWithRollback.reverse()) {
+  try {
+    await checkpoint.rollback(checkpoint.data);
```
benjdlambert, line 215:
```suggestion
    await checkpoint.rollback?.(checkpoint.data);
```

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -1,84 +1,303 @@
...
+metadata:
+  name: example-template
+spec:
+  parameters:
+    # ...
+  steps:
+    - id: github-repo
+      name: Create GitHub Repository
+      action: github:repo:create
+      rollbackEnabled: true # Enable rollback for this action
```
benjdlambert, line 242: Is this just for testing for feature flagging? The `app-config.yaml` option seems fine right to opt into this behavior, do we really need to flag it per action?

benjdlambert: As a side note, should this be done automatically or should we have a button for rolling back in the UI?

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -1,84 +1,303 @@
...
+
+For checkpoints that require additional security checks during rollback, we provide a secure rollback helper function. This function will ensure that the resources being rolled back match what was originally created by validating the checkpoint data.
 
 ```typescript
-const createPublishGitHubAction = createTemplateAction({
-  id: 'publish:github',
-  async handler() {},
-  async rollback() {},
-});
+export function createSecureCheckpointRollback<T>(
```
benjdlambert, line 263: I'm wondering if we should be encouraging resources are decorated with the `taskId` or something like that.

Take for example creating a `GitHub` repository, I know you can add attributes to the repository, and wondering if we should be adding something like `backstage.io/scaffolder-resource: ${taskId}` or something to all of the things it creates so that we can do the checking in the rollback function to make sure that these things were created by the scaffolder.

I'm thinking that if we implement it like the below as a higher order function, then it's not a strong requirement to tackle this straight away.

benjdlambert (changes requested):
Sorry for the delay here in the review! Thanks for picking this up and driving it - it's looking great! Just a few comments for cleanup and some questions :pray:

beps/0006-scaffolder-action-rollback/README.md
```diff
@@ -1,84 +1,303 @@
...
+Introducing rollback capabilities to scaffolder checkpoints provides a systematic approach to recover from template execution failures, allowing resources to be cleanly returned to their initial state.
+
+## Problem Statement
+
+When a template execution fails after creating a number of resources but before completing, the system is left in an inconsistent state. Currently, there is no standardized way to clean up these partially created resources, leading to orphaned resources in external systems requiring manual cleanup.
 
 ## Motivation
 
-The aim is to help to mitigate the issue of manual clean up the partially created resources.
+Mitigate the issue of manual cleanup from partially created resources during template execution failures. By leveraging the existing checkpoint system as the foundation for rollback capabilities, we can provide a more granular and atomic approach to resource cleanup based on well-defined operation boundaries.
```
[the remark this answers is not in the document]
kurtaking: > Is it really worth removing team access and removing the branch protection rules, when simply removing the repository will do all of that for you in one API call?

No, I don't believe so. I came to the same conclusion with others when talking through it out loud in the SIG.

> Which kind of led me to thinking that maybe it should just be at the action level, but then I again it does give you more flexibility in the checkpoints to be able to roll these back, as you don't have to recreate clients etc as they're all in the same scope.

I started at the action level, which we could _start_ with. The intriguing thing about integrating immediately with checkpoints is that they already contain everything needed (same scope). I was working through the concept of a `secureRollback` helper utility that would force validation to prevent bad actors, but that's covered by the checkpoint already having access to the current state of the task execution.

> I also guess that you don't have to provide a rollback for a checkpoint that it's state will be cleaned up later on in the rollback flow when it runs the checkpoints in reverse order.

Exactly, I think it would be optional, and then _hopefully_ the creator of the action can determine if it's needed or not. An edge case is using actions in the embedded scaffolder action, where maybe you are adding a team or modifying branch protection rules without creating/deleting a repo as part of the workflow. That's a thought and not based on anything we are doing internally today.

beps/0006-scaffolder-action-rollback/README.md
benjdlambert: > The thought behind allowing an override for each action is perhaps one company does not allow deletion of repo under any circumstance. What if they wanted to disable that, but could leverage rollbacks else where?

Wondering if it should be `rollback: false` then an opt into disabling instead?

kurtaking:
Hoping to get back on this soon. I took a break while BackstageCon was going and then have been out sick for a week. 

@benjdlambert - in the mean time, this [PR](https://github.com/backstage/backstage/pull/29353) is ready for review which relates somewhat to this

kurtaking:
Hoping we can finish https://github.com/backstage/backstage/pull/29353 early this week 🙏🏼

beps/0006-scaffolder-action-rollback/README.md
kurtaking: Added in https://github.com/backstage/backstage/pull/29196/commits/24c7b7093bb2ddb3fd5a7f03533e5ebd34d72351

beps/0006-scaffolder-action-rollback/README.md
kurtaking: Updated in https://github.com/backstage/backstage/pull/29196/commits/24c7b7093bb2ddb3fd5a7f03533e5ebd34d72351

beps/0006-scaffolder-action-rollback/README.md
kurtaking: Updated in https://github.com/backstage/backstage/pull/29196/commits/24c7b7093bb2ddb3fd5a7f03533e5ebd34d72351

beps/0006-scaffolder-action-rollback/README.md
kurtaking: Ah, yes. Fixed in https://github.com/backstage/backstage/pull/29196/commits/ba1b8c129d69d480231b85b480fa37fb3604e6bf

kurtaking:
@benjdlambert - 

https://github.com/backstage/backstage/pull/29353 has been merged and I've addressed your previous rounds of code review. However, I've fallen behind on updates to the scaffolder and could use some help regrouping on this.

kurtaking:
Not a top priority for me at the moment, but not stale.

kurtaking:
not stale

drodil:
Not stale?

kurtaking:
> Not stale?

No, but it's at the bottom of the list of todos. Happy for anyone else to help finish it, otherwise it's going to come after the metrics work I am doing.

drodil:
Can we get this reopened?

freben:
Sure!

vrabbi:
Any updates on when this may land? Really looking forward to this funcyionality as we have multiple use cases that need this

vrabbi:
Any updates?

kurtaking:
> Any updates?

@vrabbi - I think there is little missing from the BEP to call it good to go, but I'm struggling to find the bandwidth these days. As of now, I have no plans of working on the actual implementation and need help if needed sooner.

PR #29200, opened 2025-03-13
Canon: Text styles updates

cdedreuille:

### DIFF
--- a/packages/canon/src/components/Link/Link.tsx
+++ b/packages/canon/src/components/Link/Link.tsx
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


packages/canon/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,55 @@
...
+    className,
+    ...restProps
+  } = props;
+
+  // Get the responsive values for the variant and weight
+  const responsiveVariant = useResponsiveValue(variant);
+  const responsiveWeight = useResponsiveValue(weight);
+
+  return (
+    <a
```
Rugvip, line 39: We'll need to be able to render links using `Link` from `react-router-dom` as well in some form

Rugvip (approved):

packages/canon/src/components/Link/Link.tsx
cdedreuille: Ah you're right. I'll have to implement the `as` prop so you can add your component.

Issue #27421, opened 2024-11-01
🚀 Feature: Scaffolder task list permissions


drodil:
### 🔖 Feature description

The scaffolder task read permissions do not allow the creation of complex permissions. Currently, it's only on/off with the `taskReadPermission`, and no possibility to add conditions.

### 🎤 Context

We would like to limit the listed tasks so that:

1. The user can see his own runs
2. The template owner can see all runs of that template
3. If the user doesn't have read access to the template entity, the task should be hidden

The number 3 is not really that big of an issue as the template name seems to be hidden in the task list if the user does not have access to it but it still allows user to see the output + logs of the run.

### ✌️ Possible Implementation

Add additional conditions for the `taskReadPermission`. At least the following rules should be available:

* `isTaskCreator` - should filter based on the `createdBy` column using the current user entity ref
* `hasCreatedBy` - should filter based on the `createdBy` column in the `SerializedTask`
* `hasTemplateEntityRef` -  should filter based on the `spec.templateInfo.entityRef` in the `spec` column of `SerializedTask`. This requires some JSON query magic
* `hasTemplateAccess` - should also filter by template entityRef but include only template references the user has access to. Probably requires a catalog call to get the entity refs.

Additionally, some other rules that could be done:

* `hasStatus` - should filter by task status

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

stephenglass:
PR #25969 is relevant however there are still some changes requested

04kash:
Hi @benjdlambert @drodil, I'm currently working on a PR based on https://github.com/backstage/backstage/pull/25969, and I’d like to try incorporating the rules mentioned in this issue as well. Could I be assigned to it?

acierto:
Hi @04kash I assigned it on you 👍

drodil:
@04kash any updates?

04kash:
> @04kash any updates?
Working on addressing the comments left by @vinzscam on https://github.com/backstage/backstage/pull/25969 and will then look into adding additional conditions!

04kash:
[...]

PR #29202, opened 2025-03-13
Scaffolder task granular permissions

04kash:

Converted the `scaffolder.task.read` and `scaffolder.task.cancel permissions` into resource permissions, and added the following rules:
`hasCreatedBy`: Restricts access to tasks based on task creators; this can be be used by conditional policies to restrict access to tasks to only their owners.
`hasTemplateEntityRefs`: Match tasks with the given template entity refs; this can be used by conditional policies to allow template owners to view all runs of their templates.

Related to: https://github.com/backstage/backstage/issues/27421

Example:
In this example, I have two templates, `template:default/message-logger` owned by `user:default/04kash` and `template:default/sleep-dummy` owned by `user:development/guest`. Both users have run each template once. This is what the task list looks like for each user:

user:default/04kash:

Screenshot from 2025-03-12 18-22-30

user:development/guest:

Screenshot from 2025-03-12 18-22-52

Permission Policy used:
```ts
export class CustomPermissionPolicy implements PermissionPolicy {
  private readonly catalogClient: CatalogClient;

  constructor(catalogClient: CatalogClient) {
    this.catalogClient = catalogClient;
  }

  async getUserOwnedTemplates(userRef: string): Promise<string[]> {
    if (!userRef) return [];

    const response = await this.catalogClient.getEntities({
      filter: {
        kind: ['Template'],
        'relations.ownedBy': [userRef],
      },
    });

    return response.items.map(item => `${item.kind.toLocaleLowerCase()}:${item.metadata.namespace}/${item.metadata.name}`);
}

  async handle(
    request: PolicyQuery,
    user?: PolicyQueryUser,
  ): Promise<PolicyDecision> {
    if (isPermission(request.permission, taskReadPermission)) {
      const userOwnedTemplates = await this.getUserOwnedTemplates(
        user?.info.userEntityRef || '',
      );

      return createScaffolderTaskConditionalDecision(request.permission, {
        anyOf: [
          scaffolderTaskConditions.hasCreatedBy({
            createdBy: user?.info.userEntityRef ? [user?.info.userEntityRef] : [],
          }),
          scaffolderTaskConditions.hasTemplateEntityRefs({
            templateEntityRefs: userOwnedTemplates,
          }),
        ],
      });
    }
    return { result: AuthorizeResult.ALLOW };
  }
}
```

### DIFF
--- a/.changeset/weak-bags-behave.md
+++ b/.changeset/weak-bags-behave.md
@@ -1,0 +1,11 @@
+---
+'@backstage/plugin-scaffolder-backend': minor
+'@backstage/plugin-scaffolder-common': minor
+'@backstage/plugin-scaffolder-react': minor
+'@backstage/plugin-scaffolder-node': minor
+'@backstage/plugin-scaffolder': minor
+---
+
+BREAKING : Added two new scaffolder rules for `scaffolder.task.read` and `scaffolder.task.cancel` to allow for conditional permission policies such as restricting access to tasks and task events based on creators (`hasCreatedBy`) and granting template owners visibility into all runs of their templates (`hasTemplateEntityRefs`).
+
+BREAKING: Removed requirement to have both `scaffolder.task.read` and `scaffolder.task.cancel` permissions to cancel tasks.
--- a/docs/features/software-templates/authorizing-scaffolder-template-details.md
+++ b/docs/features/software-templates/authorizing-scaffolder-template-details.md
@@ -176,7 +176,9 @@ class ExamplePermissionPolicy implements PermissionPolicy {
 
 ### Authorizing scaffolder tasks
 
-The scaffolder plugin also exposes permissions that can restrict access to tasks, task logs, task creation, and task cancellation. This can be useful if you want to control who has access to these areas of the scaffolder.
+The scaffolder plugin also exposes permissions that can restrict access to tasks, task logs, task creation, and task cancellation. This can be useful if you want to control who has access to these areas of the scaffolder. You can also restrict access to tasks to only their owners or admin users and allow template owners to view all runs of their templates.
+
+The following is a simple example of how to do this:
 
 ```ts title="packages/src/backend/plugins/permissions.ts"
 /* highlight-add-start */
@@ -185,34 +187,117 @@ import {
   taskCreatePermission,
   taskReadPermission,
 } from '@backstage/plugin-scaffolder-common/alpha';
+import {
+  AuthorizeResult,
+  isPermission,
+  PolicyDecision,
+} from '@backstage/plugin-permission-common';
+import {
+  PermissionPolicy,
+  PolicyQuery,
+  PolicyQueryUser,
+} from '@backstage/plugin-permission-node';
+import {
+  createScaffolderTaskConditionalDecision,
+  scaffolderTaskConditions,
+} from '@backstage/plugin-scaffolder-backend/alpha';
+import { CatalogClient } from '@backstage/catalog-client';
 /* highlight-add-end */
 
 class ExamplePermissionPolicy implements PermissionPolicy {
+  private readonly catalogClient: CatalogClient;
+
+  constructor(catalogClient: CatalogClient) {
+    this.catalogClient = catalogClient;
+  }
+
+  // Fetches all templates owned by the user from the catalog API.
+  async getUserOwnedTemplates(userRef: string): Promise<string[]> {
+    if (!userRef) return [];
+
+    const response = await this.catalogClient.getEntities({
+      filter: {
+        kind: ['Template'],
+        'relations.ownedBy': [userRef],
+      },
+    });
+
+    return response.items.map(
+      item =>
+        `${item.kind.toLocaleLowerCase()}:${item.metadata.namespace}/${
+          item.metadata.name
+        }`,
+    );
+  }
+
   async handle(
     request: PolicyQuery,
     user?: PolicyQueryUser,
   ): Promise<PolicyDecision> {
     /* highlight-add-start */
-    if (isPermission(request.permission, taskCreatePermission)) {
-      if (user?.info.userEntityRef === 'user:default/spiderman') {
+    if (isPermission(request.permission, taskReadPermission)) {
+      // Allow admin1 to read any task
+      if (user?.info.userEntityRef === 'user:default/admin1') {
         return {
           result: AuthorizeResult.ALLOW,
         };
       }
+
+      // Retrieve templates that the user owns
+      const userOwnedTemplates = await this.getUserOwnedTemplates(
+        user?.info.userEntityRef || '',
+      );
+
+      // Allow users to read tasks they created or task runs of templates they own
+      return createScaffolderTaskConditionalDecision(request.permission, {
+        anyOf: [
+          scaffolderTaskConditions.hasCreatedBy({
+            createdBy: user?.info.userEntityRef
+              ? [user?.info.userEntityRef]
+              : [],
+          }),
+          scaffolderTaskConditions.hasTemplateEntityRefs({
+            templateEntityRefs: userOwnedTemplates,
+          }),
+        ],
+      });
+    }
+
+    if (isPermission(request.permission, taskCreatePermission)) {
+      const userArray = ['user:default/spiderman', 'user:default/admin1'];
+      const allowed = userArray.some(
+        allowedUser => user?.info.userEntityRef === allowedUser,
+      );
+      if (allowed) {
+        return {
+          result: AuthorizeResult.ALLOW,
+        };
+      }
+      return {
+        result: AuthorizeResult.DENY,
+      };
     }
     if (isPermission(request.permission, taskCancelPermission)) {
+      // Allow spiderman to cancel only his tasks
       if (user?.info.userEntityRef === 'user:default/spiderman') {
+        return createScaffolderTaskConditionalDecision(
+          request.permission,
+          scaffolderTaskConditions.hasCreatedBy({
+            createdBy: user?.info.userEntityRef
+              ? [user?.info.userEntityRef]
+              : [],
+          }),
+        );
+      }
+      // Allow admin1 to cancel any task
+      if (user?.info.userEntityRef === 'user:default/admin1') {
         return {
           result: AuthorizeResult.ALLOW,
         };
       }
-    }
-    if (isPermission(request.permission, taskReadPermission)) {
-      if (user?.info.userEntityRef === 'user:default/spiderman') {
-        return {
-          result: AuthorizeResult.ALLOW,
-        };
-      }
+      return {
+        result: AuthorizeResult.DENY,
+      };
     }
     /* highlight-add-end */
 
@@ -225,11 +310,21 @@ class ExamplePermissionPolicy implements PermissionPolicy {
 
 In the provided example permission policy, we only grant the `spiderman` user permissions to perform/access the following actions/resources:
 
-- Read all scaffolder tasks and their associated events/logs.
-- Cancel any ongoing scaffolder tasks.
+- Read scaffolder tasks and their associated events/logs for tasks created by `spiderman`.
+- Read scaffolder task runs of templates owned by `spiderman`
+- Cancel ongoing scaffolder tasks created by `spiderman`.
 - Trigger software templates, which effectively creates new scaffolder tasks.
 
-Any other user would be denied access to these actions/resources.
+On the other hand, the `admin1` user is granted:
+
+- Read access to all scaffolder tasks and their associated events/logs.
+- The ability to cancel any ongoing scaffolder task.
+- The ability to trigger software templates.
+
+All other users are granted permissions to perform/access the following actions/resources:
+
+- Read scaffolder tasks and their associated events/logs for tasks created by the user.
+- Read scaffolder task runs of the templates owned by the user
 
 Although the rules exported by the scaffolder are simple, combining them can help you achieve more complex use cases.
 
--- a/plugins/scaffolder-backend/src/scaffolder/tasks/DatabaseTaskStore.ts
+++ b/plugins/scaffolder-backend/src/scaffolder/tasks/DatabaseTaskStore.ts
@@ -35,6 +35,7 @@ import {
   SerializedTask,
   SerializedTaskEvent,
   TaskEventType,
+  TaskFilter,
   TaskSecrets,
   TaskStatus,
 } from '@backstage/plugin-scaffolder-node';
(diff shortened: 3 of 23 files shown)


docs/features/software-templates/authorizing-scaffolder-template-details.md
```diff
@@ -176,7 +176,9 @@ class ExamplePermissionPolicy implements PermissionPolicy {
 
 ### Authorizing scaffolder tasks
 
-The scaffolder plugin also exposes permissions that can restrict access to tasks, task logs, task creation, and task cancellation. This can be useful if you want to control who has access to these areas of the scaffolder.
+The scaffolder plugin also exposes permissions that can restrict access to tasks, task logs, task creation, and task cancellation. This can be useful if you want to control who has access to these areas of the scaffolder. You can also restrict access to tasks to only their owners or admin users and allow template owners to view all runs of their templates.
```
Parsifal-M, line 179: Heya! 👋 

I am just thinking out loud if there is a better way to phrase this as there is no such thing technically as an `admin` user in Backstage right? Maybe I'm overthinking it, but I could imagine if you're going over these docs you could make an assumption that there is an admin role somewhere in Backstage 😅

docs/features/software-templates/authorizing-scaffolder-template-details.md
```diff
@@ -225,11 +310,21 @@ class ExamplePermissionPolicy implements PermissionPolicy {
...
 
-- Read all scaffolder tasks and their associated events/logs.
-- Cancel any ongoing scaffolder tasks.
+- Read scaffolder tasks and their associated events/logs for tasks created by `spiderman`.
+- Read scaffolder task runs of templates owned by `spiderman`
+- Cancel ongoing scaffolder tasks created by `spiderman`.
 - Trigger software templates, which effectively creates new scaffolder tasks.
 
-Any other user would be denied access to these actions/resources.
+On the other hand, the `admin1` user is granted:
```
Parsifal-M, line 318: `spiderman` and `admin1`.. how about [venom](https://en.wikipedia.org/wiki/Venom_(character))! Kidding, maybe we make these just Alice and Bob or John and Jane to keep it generic? Unless Bowie makes some more friends!

Parsifal-M (commented):
Awesome!

Just some comments on docs side! 👏

docs/features/software-templates/authorizing-scaffolder-template-details.md
04kash: Hmm yeah makes sense. I've rephrased this to the following:

> You can configure policies to restrict access to tasks to only their creators, while also granting specific users the permission to view all tasks. Additionally, template owners can be granted permissions to view all runs of their templates.

lmk what you think!

docs/features/software-templates/authorizing-scaffolder-template-details.md
Parsifal-M: I think it looks great! Just want to make sure @benjdlambert  is also on board (I think bowie needs a sidekick)

docs/features/software-templates/authorizing-scaffolder-template-details.md
Parsifal-M: Awesome! 👏

.changeset/weak-bags-behave.md
```diff
@@ -0,0 +1,9 @@
+---
+'@backstage/plugin-scaffolder-backend': minor
+'@backstage/plugin-scaffolder-common': minor
+'@backstage/plugin-scaffolder-react': minor
+'@backstage/plugin-scaffolder-node': minor
+'@backstage/plugin-scaffolder': minor
+---
+
+BREAKING : Added two new scaffolder rules for `scaffolder.task.read` and `scaffolder.task.cancel` to allow for conditional permission policies such as restricting access to tasks and task events based on creators (`hasCreatedBy`) and granting template owners visibility into all runs of their templates (`hasTemplateEntityRefs`).
```
acierto, line 9: Instead of `scaffolder.task.cancel` would be better to use `scaffolder.task.write`

acierto:
@vinzscam  Can you please have a look at it?

04kash:
Hi @drodil @vinzscam 👋, could you please take a look at this?

plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -472,6 +526,18 @@ export async function createRouter(
...
         rules: actionRules,
       },
+      {
+        resourceType: RESOURCE_TYPE_SCAFFOLDER_TASK,
+        permissions: scaffolderTaskPermissions,
+        rules: taskRules,
+        getResources: async resourceRefs => {
+          return Promise.all(
+            resourceRefs.map(async taskId => {
+              return await taskBroker.get(taskId);
```
drodil, line 536: Would it make sense to introduce a new method to get tasks by array of ids? This will introduce quite many queries to database..

04kash: Agreed — having a separate method makes sense so we can fetch all tasks in a single DB query. I’ll go ahead and add that in this afternoon👍

drodil: Or maybe modify the `get` to allow passing single id or array of ids 🤔

plugins/scaffolder-node/report.api.md
```diff
@@ -402,6 +403,7 @@ export interface TaskBroker {
       order: 'asc' | 'desc';
       field: string;
     }[];
+    permissionFilters?: PermissionCriteria<TaskFilters>;
```
Rugvip, line 406: potentially something like `filters?: TaskFilters` here instead to avoid leaking permission system types in the scaffolder API

plugins/scaffolder-node/report.api.md
```diff
@@ -493,6 +495,25 @@ export interface TaskContext {
 // @public
 export type TaskEventType = 'completion' | 'log' | 'cancelled' | 'recovered';
 
+// @public
+export type TaskFilter = {
```
Rugvip, line 499: If possible it would be best to leave these as internal types, but that might not be possible seeing as `TaskBroker` is exported

plugins/scaffolder-backend/src/scaffolder/tasks/DatabaseTaskStore.ts
```diff
@@ -197,6 +206,63 @@ export class DatabaseTaskStore implements TaskStore {
...
+    }
+
+    if (this.isTaskFilter(filter)) {
+      const values: string[] = compact(filter.values) ?? [];
+
+      if (filter.property === 'createdBy') {
+        query.whereIn('created_by', [...new Set(values)]);
+      }
+
+      if (filter.property === 'templateEntityRefs' && values.length > 0) {
```
Rugvip, line 231: As discussed in framework SIG: let's split out things related to this `templateEntityRefs` filter and rule addition into a separate PR, as the rest is more straight-forward

plugins/scaffolder-backend/src/service/router.ts
04kash: > Or maybe modify the get to allow passing single id or array of ids 🤔
Hmm, in that case would the function signature be something like this: `async get(taskIdOrIds: string | string[]): Promise<SerializedTask | SerializedTask[]>` and all current usages of the function would have to be typecasted to `SerializedTask`?
I feel that maybe having a separate method for this would be better.

plugins/scaffolder-backend/src/scaffolder/tasks/DatabaseTaskStore.ts
04kash: I've removed the stuff related to templateEntityRefs from this PR and opened a separate PR for them: https://github.com/backstage/backstage/pull/29624

plugins/scaffolder-node/report.api.md
04kash: hmm, I'm not very sure on how to go about doing that since I'm using [a ConditionTransformer to get the filters](https://github.com/backstage/backstage/pull/29202/files/cc8ef39038c6c2444a49d7561179837f24166cf5#diff-0f3ac61a12b2443bfec912c2932eb97037dda42dab7a48c4300fa5c4435822acR781-R787). Any ideas?

plugins/scaffolder-backend/src/service/router.ts
04kash: I've added a separate method for now

plugins/scaffolder-backend/src/scaffolder/tasks/DatabaseTaskStore.ts
```diff
@@ -197,6 +206,53 @@ export class DatabaseTaskStore implements TaskStore {
...
+
+  private parseFilter(
+    filter: PermissionCriteria<TaskFilters>,
+    query: Knex.QueryBuilder,
+    db: Knex,
+    negate: boolean = false,
+  ): Knex.QueryBuilder {
+    // handle not criteria
+    if (isNotCriteria(filter)) {
+      console.log(`reached here: ${JSON.stringify(filter)}`);
```
drodil, line 221: Extra console log

plugins/scaffolder-backend/src/scaffolder/tasks/DatabaseTaskStore.ts
```diff
@@ -209,18 +265,35 @@ export class DatabaseTaskStore implements TaskStore {
...
+              ...(permissionFilters ? [permissionFilters] : []),
+            ],
+          }
+        : permissionFilters;
+
+    if (combinedPermissionFilters) {
+      this.parseFilter(combinedPermissionFilters, queryBuilder, this.db);
     }
 
+    console.log(`Parse FIlters: ${queryBuilder}`);
```
drodil, line 295: Here too

plugins/scaffolder/src/components/OngoingTask/OngoingTask.tsx
```diff
@@ -198,118 +202,120 @@ export const OngoingTask = (props: {
...
-          onStartOver={startOver}
-          onRetry={triggerRetry}
-          onToggleLogs={setLogVisibleState}
-          onToggleButtonBar={setButtonBarVisibleState}
-          taskId={taskId}
-        />
-      </Header>
-      <Content className={classes.contentWrapper}>
-        {taskStream.error ? (
+    <RequirePermission permission={taskReadPermission} resourceRef={taskId}>
```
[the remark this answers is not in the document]
04kash: Ah yeah you're right, in that case we can also remove getResources from [here](https://github.com/backstage/backstage/pull/29202/files/e7abdc753459d6610591837991fa3e2a655ebe6b#diff-0f3ac61a12b2443bfec912c2932eb97037dda42dab7a48c4300fa5c4435822acR533) I think

plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -472,6 +526,12 @@ export async function createRouter(
         permissions: scaffolderActionPermissions,
         rules: actionRules,
       },
+      {
+        resourceType: RESOURCE_TYPE_SCAFFOLDER_TASK,
+        permissions: scaffolderTaskPermissions,
+        rules: taskRules,
+        getResources: resourceRefs => taskBroker.getTasks(resourceRefs),
```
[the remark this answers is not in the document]
04kash: I had originally added the getResources method to use `RequirePermission` in the `OngoingTask` page. Since that wasn't needed, I've removed this method as well

plugins/scaffolder-backend/src/service/rules.ts
```diff
@@ -129,10 +133,45 @@ function buildHasProperty<Schema extends z.ZodType<JsonPrimitive>>({
...
+export const createTaskPermissionRule = makeCreatePermissionRule<
+  SerializedTask,
+  {
+    key: string;
+    values: any;
+  },
+  typeof RESOURCE_TYPE_SCAFFOLDER_TASK
+>();
+
+export const hasCreatedBy = createTaskPermissionRule({
```
vinzscam, line 145: as catalog already has `isEntityOwner` I would suggest `isTaskOwner`
but up to @backstage/scaffolder-maintainers 

```suggestion
export const isEntityOwner = createTaskPermissionRule({
```

plugins/scaffolder-node/report.api.md
vinzscam: yeah it would be nice to re-use the existing

```ts
filters?: {
  createdBy?: string | string[];
  status?: TaskStatus | TaskStatus[];
}
```

by making it more generic.

vinzscam (changes requested):
this is exciting 🚀

04kash:
Hmm, I'm not sure why the tests related to kubernetes-backend are failing 🤔

plugins/scaffolder-node/report.api.md
Rugvip: Discussed this a bit more in the framework SIG, and while it's unfortunate to have these types be used here, the actual problem is really that these are exported in the first place. Afaik the only real option to the current implementation is to make a complete copy of the type structure, which I don't really see how it could be better. It would of course be best if the existing filters could be used instead, but they're not expressive enough either, so we do need to add something. I think in the end it's best to keep as is, and aim towards moving this type to be internal very soon.

plugins/scaffolder-backend/src/service/rules.ts
04kash: Discussed this in the Scaffolder SIG and we decided to go ahead with renaming this to `isTaskOwner` :+1:

04kash:
Hi @vinzscam @Rugvip, looks like all review comments have been addressed, could you give it another look?

Parsifal-M (approved):
Hey! 👋 

Approving from the docs side of things!

Thanks!

plugins/scaffolder-common/report-alpha.api.md
```diff
@@ -23,22 +26,26 @@ export const scaffolderPermissions: (
...
+  | BasicPermission
+  | ResourcePermission<'scaffolder-task'>
+)[];
 
 // @alpha
 export const scaffolderTemplatePermissions: ResourcePermission<'scaffolder-template'>[];
 
 // @alpha
-export const taskCancelPermission: BasicPermission;
+export const taskCancelPermission: ResourcePermission<'scaffolder-task'>;
```
benjdlambert, line 42: @vinzscam are these breaking changes?

plugins/scaffolder-backend/src/service/rules.ts
```diff
@@ -129,10 +133,45 @@ function buildHasProperty<Schema extends z.ZodType<JsonPrimitive>>({
   });
 }
 
+export const createTaskPermissionRule = makeCreatePermissionRule<
+  SerializedTask,
+  {
+    key: string;
+    values: any;
```
vinzscam: shouldn't this be `string[]` 🤔 

```suggestion
    values: string[];
```

plugins/scaffolder-common/report-alpha.api.md
vinzscam: this might affect adopters that provide custom components such as `<OngoingTask />`, as `usePermission` now requires `resourceRef` to be passed:

```diff
  usePermission({
      permission: taskCancelPermission,
+     resourceRef: taskId,
  });
```

but it's not a broad use case

plugins/scaffolder-backend/src/service/rules.ts
04kash: Yup I replaced 
```
 {
    key: string;
    values: any;
 }
  ```

with `TaskFilter`

PR #29215, opened 2025-03-14
docs: fix example location annotations in IncrementalEntityProvider

Erob711:

Both `ANNOTATION_LOCATION` and `ANNOTATION_ORIGIN_LOCATION` are expected to be a string location reference, which makes these examples a bit misleading. 

### DIFF
--- a/docs/features/software-catalog/external-integrations.md
+++ b/docs/features/software-catalog/external-integrations.md
@@ -1161,9 +1161,11 @@ export class MyIncrementalEntityProvider
   implements IncrementalEntityProvider<Cursor, Context>
 {
   private readonly token: string;
+  private readonly mySource: string;
 
-  constructor(token: string) {
+  constructor(token: string, mySource: string) {
     this.token = token;
+    this.mySource = mySource;
   }
 
   getProviderName() {
@@ -1181,6 +1183,7 @@ export class MyIncrementalEntityProvider
     cursor: Cursor = { page: 1 },
   ): Promise<EntityIteratorResult<Cursor>> {
     const { apiClient } = context;
+    const location = `${this.getProviderName()}:${mySource}`;
 
     // call your API with the current cursor
     const data = await apiClient.getServices(cursor);
@@ -1200,8 +1203,8 @@ export class MyIncrementalEntityProvider
           name: item.name,
           annotations: {
             // You need to define these, otherwise they'll fail validation
-            [ANNOTATION_LOCATION]: this.getProviderName(),
-            [ANNOTATION_ORIGIN_LOCATION]: this.getProviderName(),
+            [ANNOTATION_LOCATION]: location,
+            [ANNOTATION_ORIGIN_LOCATION]: location,
           },
         },
         spec: {
--- a/docs/features/software-catalog/external-integrations.md
+++ b/docs/features/software-catalog/external-integrations.md
@@ -1157,6 +1157,12 @@ export class MyIncrementalEntityProvider
 The last step is to implement the actual `next` method that will accept the cursor, call the API, process the result and return the result.
 
 ```ts
+import {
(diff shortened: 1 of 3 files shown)


docs/features/software-catalog/external-integrations.md
```diff
@@ -1200,8 +1200,8 @@ export class MyIncrementalEntityProvider
           name: item.name,
           annotations: {
             // You need to define these, otherwise they'll fail validation
-            [ANNOTATION_LOCATION]: this.getProviderName(),
```
Parsifal-M, line 1203: Hiya! 👋 

Just curious because you mentioned its not clear what this should be, but this would end up as [MyIncrementalEntityProvider](https://github.com/backstage/backstage/blob/e873f22a935bd5c415f93adfeda68e46b4e8a5c1/docs/features/software-catalog/external-integrations.md?plain=1#L1146)

The `ANNOTATION_LOCATION` and `ANNOTATION_ORIGIN_LOCATION` don't strictly have to point to a url.

A potentially good example is https://github.com/backstage/backstage/blob/0afecf46cdf2d62bbec701daaca6b5e207992acf/plugins/catalog-backend-module-gcp/src/providers/GkeEntityProvider.ts#L157

Just wondering if this helps clarify it in any way or how we can make it clearer in the docs 🙏

Erob711: Ah, I must have misunderstood the possible formats of those annotations, then. My original confusion with the doc remains, but perhaps it comes from elsewhere.

Context: When I run my incremental provider with `this.getProviderName()` as that value, entities do not seem to get handed off from my provider into the catalog. Peaking into the unprocessed entities with `api/catalog/entities/unprocessed/failed`, I ended up seeing an error message on every entity, in this form:
```js
{
"name": "TypeError",
"message": "Unable to parse location ref 'MyCustomProvider', expected '<type>:<target>', e.g. 'url:https://host/path'"
}
```

Erob711: So I ended up switching the values of those annotations to something like `url:...` the service I was scraping entites from, and that cleared up the failed entities. Hence, what took me to this doc 😄

Parsifal-M: Hmm interesting though looking at 
https://github.com/backstage/backstage/blob/85df833fe33c45cf2ad8e0322dafb118b89a5ea8/packages/catalog-model/src/location/helpers.ts#L42-L47

I think maybe we should update this to be like some of the other examples further up, or like the example I pasted.

Just wondering what @freben thinks here 🧙

Erob711: The example you showed makes a ton of sense to me! Perhaps all these IncrementalEntityProvider docs are missing is this added piece where it looks like the string ref is being built with `this.getProviderName()` in mind (also from the example you provided):
https://github.com/backstage/backstage/blob/0afecf46cdf2d62bbec701daaca6b5e207992acf/plugins/catalog-backend-module-gcp/src/providers/GkeEntityProvider.ts#L130

I could try to format the example more like this, and see how that shapes out?

freben: yeah these still need the type prefix - here's an example from the demo site:

```
backstage.io/managed-by-location: url:https://github.com/backstage/backstage/tree/master/packages/catalog-model/examples/components/petstore-component.yaml
backstage.io/managed-by-origin-location: url:https://github.com/backstage/backstage/blob/master/packages/catalog-model/examples/all-components.yaml
```

But in this example, are they even URLs? Should the annotations be ``` `${this.getProviderName()}:${mySource}` ``` as their key instead? They are documented here https://backstage.io/docs/features/software-catalog/well-known-annotations#backstageiomanaged-by-location

PR #29247, opened 2025-03-17
Canon: Add render prop to Link

cdedreuille:

To allow this Link component to work with React Router, I added a render prop using `useRender` from BaseUI [Base UI useRender](https://base-ui.com/react/utils/use-render). Here's an example how it will work:

```tsx
import { Link as RouterLink ) from 'react-router';

<Link render={<RouterLink to="/catalog" />}>
  Hello world
</Link>
```


packages/canon/src/components/Link/types.ts
```diff
@@ -30,4 +30,5 @@ export interface LinkProps
   weight?: 'regular' | 'bold' | Partial<Record<Breakpoint, 'regular' | 'bold'>>;
   className?: string;
   style?: CSSProperties;
+  render?: (props: Omit<LinkProps, 'render'>) => ReactNode;
```
Rugvip, line 33: I think we wanna make it `component` or such. render is a bit of an older pattern afaik and isn't as explicit about whether it's ok to call hooks in the provided function

cdedreuille: This is to follow what Base UI is doing. I had the same impression as you but I would prefer a consistent pattern rather than two different behaviour.

https://base-ui.com/react/handbook/composition

Rugvip: wattafakk, that looks a bit different though with the intended use being to pass an element with some constraints on the underlying component implementation.

I have to assume that Base UI is doing some trickery and cloning or re-rendering with the component type or something like that. I agree we should stick to what Base UI does though, including the ability to pass am element.

Still feel this leaves the confusion around whether it's alright or not to use hooks in the render function though, at this point I assume not 🤷

packages/canon/src/components/Link/Link.tsx
```diff
@@ -14,42 +14,59 @@
...
+      style,
+      children,
+      to,
+      ...restProps,
+    };
+
+    if (render) {
+      // If render is a component type, wrap it in memo to prevent unnecessary re-renders
+      if (typeof render === 'function' && !render.length) {
+        const MemoizedComponent = memo(
```
vinzscam, line 56: mmm but `MemoizedComponent` will be different on each re-render. I think `memo` shouldn't be used inside a render method. Also as `...linkProps` is spread, it might cause a re-render anyway...

vinzscam (dismissed):

Rugvip (changes requested):
yep need to iterate on this a bit

cdedreuille:
@Rugvip @vinzscam I just pushed a new update for this one. Base UI created a special `useRender` hook to help use replicate their behaviour. Now we can pass any component to the render function like this:

```tsx
import { Link as RouterLink ) from 'react-router';

<Link render={<RouterLink to="/catalog" />}>
  Hello world
</Link>
```

Rugvip (approved):
Nice! 🎉, let's :shipit:

Issue #17712, opened 2023-05-09
🚀 Feature: Add light mode to microsite(backstage.io)


jhaals:
### 🔖 Feature description

Our Microsite is built on top of [Docusaurus](https://docusaurus.io/) which supports both dark and light mode. Documentation and other generated pages already look good/alright when light mode is enabled but our custom pages is does not currently support light mode and therefore those pages remain dark with a light menu bar which does not look very pleasing.

It would be fantastic if anyone from the community with CSS skills would be willing to help out updating our styles to support both light and dark mode 🙏 

More info can be obtained at
https://docusaurus.io/docs/api/themes/configuration
https://docusaurus.io/docs/styling-layout

jhaals:
This is the only remaining task from the previous meta issue #12019

PR #29268, opened 2025-03-18
Added theme change feature to docs page

AmbrishRamachandiran:
Added theme change feature to docs page:

Light :

image

Dark:

image

### DIFF
--- a/microsite/docusaurus.config.ts
+++ b/microsite/docusaurus.config.ts
@@ -344,7 +344,7 @@ const config: Config = {
 
     colorMode: {
       defaultMode: 'dark',
-      disableSwitch: true,
+      disableSwitch: false, // Enable the theme switch
     },
     navbar: {
       logo: {


microsite/docusaurus.config.ts
```diff
@@ -344,7 +344,7 @@ const config: Config = {
 
     colorMode: {
       defaultMode: 'dark',
-      disableSwitch: true,
+      disableSwitch: false, // Enable the theme switch
```
benjdlambert, line 347: This isn't just as simple as this. I think there needs to be care taken to ensure that the all pages support light and dark mode. We probably need to get different image assets for both simply the Backstage logo, as well as other images embedded in different pages.

I've attached a few screenshots of what's currently broken, but there might be more. I'm pretty sure that a lot of the fonts and styles do not pass basic accessibility tests because there's not enough contrast between the text and the white background in many places too.



Screenshot 2025-03-20 at 08 25 07
Screenshot 2025-03-20 at 08 25 27
Screenshot 2025-03-20 at 08 25 33
Screenshot 2025-03-20 at 08 25 39

Maybe we should re-open #17712 if you want to take some time to go through this?

benjdlambert (changes requested):

PR #29269, opened 2025-03-18
catalog: Set the default refresh interval to 24hrs

benjdlambert:
Internally we've had a lot of success increasing the default refresh interval to much longer, giving the processing loop time to breathe.

We're encouraging that you setup events support for providers for example [like this](https://backstage.io/docs/integrations/github/discovery#events-support) to enable push support for updates, and just use the processing interval as a reconcilliation loop.

This should reduce the amount of Rate Limits that Catalogs run into, as it will naturally load balance these refreshes throughout the day.

### DIFF
--- a/.changeset/ten-days-slide.md
+++ b/.changeset/ten-days-slide.md
@@ -1,0 +1,15 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Updating the default processing time to be 24 hours. This means that entities will refresh once every ~24 hours, but we encourage the use of the events support for different providers to refresh more often and be reactive to changes outside the refresh loop.
+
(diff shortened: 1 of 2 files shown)


MihaiTabara (approved):

Rugvip (commented):
Hmm, feeling we'd need to ship event support by default if we wanted to do this? Right now this would kinda break the out-of-the-box experience ?

benjdlambert:
@Rugvip in what way would it break?

UsainBloot:
I do agree with @Rugvip that an event strategy should be out of the box.

Currently an adopter would only see changes to their entities once a day unless they manually refresh, which gives a false impression into the capabilities of the catalog.

PR #29270, opened 2025-03-18
Sentence corrected with unnecessary commas removed

deepthi-28:
Document updated by removing semicolon and full stop.


Parsifal-M:
Heya 👋 

Its actually grammatically correct the way it is, without getting too grammar nerdy on it, its a way of emphasising something, could be interesting to read (if you like) https://twominenglish.com/staccato-sentence/ 

We're emphasising **Sometimes** then providing an explanation, like a dramatical effect 😄 

WDYT? 😄

PR #29283, opened 2025-03-18
Made "github:deployKey:create" action idempotent

acierto:

Made "github:deployKey:create" action idempotent

### DIFF
--- a/plugins/scaffolder-backend-module-github/src/actions/githubDeployKey.ts
+++ b/plugins/scaffolder-backend-module-github/src/actions/githubDeployKey.ts
@@ -126,20 +126,35 @@ export function createGithubDeployKeyAction(options: {
 
       const client = new Octokit(octokitOptions);
 
-      await client.rest.repos.createDeployKey({
-        owner: owner,
-        repo: repo,
-        title: deployKeyName,
-        key: publicKey,
+      await ctx.checkpoint({
+        key: `create.deploy.key.${owner}.${repo}.${publicKey}`,
+        fn: async () => {
+          await client.rest.repos.createDeployKey({
+            owner: owner,
+            repo: repo,
+            title: deployKeyName,
+            key: publicKey,
+          });
+        },
       });
-      const publicKeyResponse = await client.rest.actions.getRepoPublicKey({
-        owner: owner,
-        repo: repo,
+
+      const { key, keyId } = await ctx.checkpoint({
+        key: `get.repo.public.key.${owner}.${repo}`,
+        fn: async () => {
+          const publicKeyResponse = await client.rest.actions.getRepoPublicKey({
+            owner: owner,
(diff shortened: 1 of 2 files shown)


plugins/scaffolder-backend-module-github/src/actions/githubDeployKey.ts
```diff
@@ -152,12 +167,17 @@ export function createGithubDeployKeyAction(options: {
...
       );
 
-      await client.rest.actions.createOrUpdateRepoSecret({
-        owner: owner,
-        repo: repo,
-        secret_name: privateKeySecretName,
-        encrypted_value: encryptedBase64Secret,
-        key_id: publicKeyResponse.data.key_id,
+      await ctx.checkpoint({
+        key: `create.or.update.repo.secret.${owner}.${repo}.${keyId}`,
```
freben, line 171: sorry for asking but, how can the secret itself not be part of the key? isn't the `key` part of what can change until the next run?

acierto: If user will change something, it will be a new task.

We need a unique key here with some extras if for example in a template will be multiple actions, so we won't cache the result of first 'github:deployKey:create' and reuse it for all subsequent 'github:deployKey:create's

plugins/scaffolder-backend-module-github/src/actions/githubDeployKey.ts
```diff
@@ -126,20 +126,35 @@ export function createGithubDeployKeyAction(options: {
...
+            title: deployKeyName,
+            key: publicKey,
+          });
+        },
       });
-      const publicKeyResponse = await client.rest.actions.getRepoPublicKey({
-        owner: owner,
-        repo: repo,
+
+      const { key, keyId } = await ctx.checkpoint({
```
benjdlambert, line 141: Does this even need to be a checkpoint? 

This is just getting the key right? No harm in re-running this?

plugins/scaffolder-backend-module-github/src/actions/githubDeployKey.ts
benjdlambert, line 171: Same for this too? This doesn't matter if it's re-run?

benjdlambert (commented):
Some quick comments

plugins/scaffolder-backend-module-github/src/actions/githubDeployKey.ts
acierto: True, no harm, I was thinking also to reduce unnecessary calls to a provider, if we can cache them.
Then re-run will be much faster

benjdlambert (approved):
Let's go with this for now, and we can come back to to this and change if we need to.

PR #29308, opened 2025-03-19
Add Slack Notification Processor

kunickiaj:

Adds a new Slack notification processor for use with the notifications plugin.
Updates the stale PR https://github.com/backstage/backstage/pull/23991

This is a Slack App implementation we have been using for quite some time. It relies on the annotations described in the readme. DM, public and private Slack channels are supported. Depending on your app's granted scopes, you may be able to post to any public channel without first inviting the app. Private channels require membership and a channel ID. Public channels with restricted posting will need the App added to the allowlist for the channel.

Example notification:

image


plugins/notifications-backend-module-slack/README.md
```diff
@@ -0,0 +1,49 @@
+# @backstage/plugin-notifications-backend-module-slack
+
+The Slack backend module for the notifications plugin.
+
+## Getting Started
```
awanlin, line 5: Can we add this as a section in this docs here: https://github.com/backstage/backstage/blob/master/docs/notifications/processors.md

plugins/notifications-backend-module-slack/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/backstage/backstage",
+    "directory": "plugins/notifications-backend-module-slack"
+  },
+  "license": "Apache-2.0",
+  "main": "src/index.ts",
+  "types": "src/index.ts",
+  "files": [
+    "dist"
```
awanlin, line 24: You'll want to add the `config.d.ts` to this array

awanlin: You'll also want to add this property: `"configSchema": "config.d.ts"`

plugins/notifications-backend-module-slack/src/config.d.ts
awanlin, line 1: Pretty sure this file should live one level up form here.

plugins/notifications-backend-module-slack/src/config.d.ts
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+export interface Config {
+  notifications?: {
+    processors?: {
+      slack?: Array<{
+        /**
+         * Slack Bot Token. Usually starts with `xoxb-`.
+         * @visibility backend
```
awanlin, line 22:
```suggestion
         * @visibility secret
```
Making an assumption that this needs to be secured.

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -0,0 +1,285 @@
...
+    this.messagesSent = meter.createCounter(
+      'notifications.processors.slack.sent.count',
+      {
+        description: 'Number of messages sent to Slack successfully',
+      },
+    );
+    this.messagesFailed = meter.createCounter(
+      'notifications.processors.slack.error.count',
+      {
+        description: 'Number of messages that failed to send to Slack',
```
awanlin, line 97: These metrics are a nice addition, should we document them somewhere?

awanlin (changes requested):
Thanks for picking this up again @kunickiaj, just a few comments around docs and the config. 👍

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
kunickiaj: added to readme

awanlin (approved):
Wow, thanks for the quick follow up @kunickiaj 🚀

awanlin:
Hi @mareklibra, can you give this a look, please?

docs/notifications/processors.md
```diff
@@ -104,4 +104,32 @@ Apart from STMP, the email processor also supports the following transmissions:
...
+
+### Slack Processor
+
+Slack processor is used to send notifications to users and channels in Slack. To install the Slack processor, add the `@backstage/plugin-notifications-backend-module-slack` package to your backend.
+
+```bash
+yarn workspace backend add @backstage/plugin-notifications-backend-module-slack
+```
+
+Add the email processor to your backend:
```
drodil, line 117: email -> slack?

docs/notifications/processors.md
```diff
@@ -104,4 +104,32 @@ Apart from STMP, the email processor also supports the following transmissions:
...
+backend.add(import('@backstage/plugin-notifications-backend-module-slack'));
+```
+
+To configure the Slack processor, you need to add the following configuration to your `app-config.yaml`:
+
+```yaml
+notifications:
+  processors:
+    slack:
+      - token: xoxb-XXXXXXXXX
```
drodil, line 132: Maybe instruct where to get the token? I am not slack user so don't know 😆

plugins/notifications-backend-module-slack/README.md
```diff
@@ -0,0 +1,56 @@
...
+- `slack.com/channel-name` for public channel messages
+- `slack.com/channel-id` for public or private channel messages
+
+Slack prefers use of ID over name and `slack.com/channel-id` is the recommended annotation.
+
+### Observability
+
+The processor includes the following counter metrics if you are exporting metrics using OpenTelemetry:
+
+- `notifications.processors.slack.sent.count` - The number of messages sent
```
drodil, line 55: These would be nice to add also to email processor, nice!

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -0,0 +1,285 @@
...
+      },
+    );
+
+    const throttle = pThrottle({
+      limit: 10,
+      interval: durationToMilliseconds({ minutes: 1 }),
```
drodil, line 103: Should these be configurable like in email processor or is the slack api strict with these?

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  getName(): string {
+    return 'SlackNotificationProcessor';
+  }
+
+  async processOptions(
+    options: NotificationSendOptions,
+  ): Promise<NotificationSendOptions> {
+    if (options.recipients.type !== 'entity') {
```
drodil, line 139: Broadcast support can be done later like in email processor, or WDYT?

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -0,0 +1,285 @@
...
+    this.logger.debug(`Sending DM notification: ${JSON.stringify(payload)}`);
+
+    // batch it up
+    await this.sendNotifications([payload]);
+  }
+
+  async getEntities(
+    entityRefs: readonly string[],
+  ): Promise<(Entity | undefined)[]> {
+    const { token } = await this.auth.getPluginRequestToken({
```
drodil, line 230: Could probably use cacheService here (and other entity getters) like in email processor to prevent unnecessary calls to catalog. But can be added later if this becomes an issue or WDYT?

plugins/notifications-backend-module-slack/src/lib/util.ts
```diff
@@ -0,0 +1,92 @@
...
+export function toSlackBlockKit(payload: NotificationPayload): KnownBlock[] {
+  const { description, link, severity, topic } = payload;
+  return [
+    {
+      type: 'section',
+      text: {
+        type: 'mrkdwn',
+        text: description ?? 'No description provided',
+      },
+      accessory: {
```
drodil, line 53: If there's no link, should this be discarded? Don't know the slack api but just thinking empty string as url is interesting

plugins/notifications-backend-module-slack/src/lib/util.ts
```diff
@@ -0,0 +1,92 @@
...
+      type: 'context',
+      elements: [
+        {
+          type: 'plain_text',
+          text: `Severity: ${severity ?? 'normal'}`,
+          emoji: true,
+        },
+        {
+          type: 'plain_text',
+          text: `Topic: ${topic ?? 'N/A'}`,
```
drodil, line 73: Also is this necessary block if there's no topic?

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -0,0 +1,285 @@
...
+  async postProcess(
+    notification: Notification,
+    options: NotificationSendOptions,
+  ): Promise<void> {
+    if (options.recipients.type === 'broadcast' || !notification.user) {
+      return;
+    }
+
+    const entityRefs = [options.recipients.entityRef].flat();
+    if (entityRefs.some(e => parseEntityRef(e).kind === 'group')) {
```
drodil, line 205: This actually cannot ever be group, it's always user. See https://github.com/backstage/backstage/blob/master/plugins/notifications-backend/src/service/router.ts#L535

I think there should be possibility to add some processor specific metadata to notification in `processOptions` which could in this case indicate that slack notification is already sent to group. This though requires some changes to the notification-backend 🤔

drodil (commented):
Good job, looks nice! Just a couple of things to consider, let me know what you think!

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
drodil: Yes, the groups (and actually any other entities as well) are expanded before calling the `preProcess` and `postProcess` so basically this check is unnecessary. But AFAIK there's currently no way to pass processor specific data between the `processOptions`, `preProcess` and `postProcess` which could be used in this case. As said, this would require some additional changes to the `notifications-backend` itself.

Do you think you could live with sending both group and individual messages at this point or do you want to also make the necessary changes to `notifications-backend`? My proposal for this would be to add the following to `Notification` type:

```ts
metadata: JsonObject
```
and this would not be saved to the database so it would only live during the sending phase. WDYT?

plugins/notifications-backend-module-slack/README.md
```diff
@@ -0,0 +1,56 @@
...
+### Slack Configuration
+
+To use this you'll need to create a Slack App or use an existing one. It should have at least the following scopes:
+`chat:write`, `users:read`, `im:write` (for direct message support).
+
+Additionally you may include scopes `chat:write.public` in order to send messages to public channels your app is not
+a member of.
+
+These scopes are under OAuth & Permissions. You will also want to save the Bot User OAuth Token. This will be needed
+in the following step to configure `app-config.yaml`.
```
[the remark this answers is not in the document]
drodil: I think it should be here, @awanlin WDYT?

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -0,0 +1,285 @@
...
+  }) {
+    const { auth, catalog, logger, slack } = options;
+    this.logger = logger;
+    this.catalog = catalog;
+    this.auth = auth;
+    this.slack = slack;
+
+    const meter = metrics.getMeter('default');
+    this.messagesSent = meter.createCounter(
+      'notifications.processors.slack.sent.count',
```
freben, line 89: another possibility: `notifications.processors.slack.sent`, which for example in Grafana you might graph as `notifications_processors_slack_sent_total` (which is how the exporter amends it), and then you could have the label `result: 'ok'` (or `'failed'`). Doing that instead of two individual metrics, lets you nicely co-graph them and compute failure rate etc.

Up to you though

plugins/notifications-backend-module-slack/README.md
```diff
@@ -0,0 +1,56 @@
...
+Multiple instances can be added in the `slack` array, allowing you to have multiple configurations if you need to send
+messages to more than one Slack workspace. Org-Wide App installation is not currently supported.
+
+### Entity Requirements
+
+Entities must be annotated with one of the following annotations:
+
+- `slack.com/user-id` for direct messages
+- `slack.com/channel-name` for public channel messages
+- `slack.com/channel-id` for public or private channel messages
```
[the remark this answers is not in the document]
kunickiaj: This is definitely an interesting idea. The reason for my current decision is likely related to some internal details. We Typically have up to two channels (slack channel and slack alerts channel), the latter of which is used for alerts, but not necessarily notifications like the ones that would come from backstage. I do like the idea of bot-notify though as it can be applied to users as well. We are annotating users with user-id because i thought it was clearer.

I suppose then the url could be: channel, dm, or user profile where the last segment is and ID that is messagable.

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -0,0 +1,285 @@
...
+    entityRefs: readonly string[],
+  ): Promise<(Entity | undefined)[]> {
+    const { token } = await this.auth.getPluginRequestToken({
+      onBehalfOf: await this.auth.getOwnServiceCredentials(),
+      targetPluginId: 'catalog',
+    });
+
+    const response = await this.catalog.getEntitiesByRefs(
+      {
+        entityRefs: entityRefs.slice(),
```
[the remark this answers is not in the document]
kunickiaj: This was added due to:

`The type 'readonly string[]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.ts(4104)`

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
kunickiaj: @drodil 
> This actually cannot ever be group, it's always user. See [master/plugins/notifications-backend/src/service/router.ts#L535](https://github.com/backstage/backstage/blob/master/plugins/notifications-backend/src/service/router.ts?rgh-link-date=2025-03-20T17%3A44%3A28Z#L535)

This is definitely how it works for us today -- notification.user is always a user, but options.recipients.entityRef **may** contain a group. This is how I've been determining whether or not we already sent a notification because the user comes from an expanded group.

I think we should aim to make the change you propose but we can do that separately. This iteration of the processor will only send a user a message if the notificationoptions also had a user as the recipient.

freben (approved):
Alright, approving with optional comments for future improvement

drodil (approved):
Looks great! Approved but agreeing with optional points by @freben

freben:
Let's merge so it gets in the release, then iterate in a followup!

PR #29310, opened 2025-03-19
Added EntityOrderFilter to sort entities by different fields

mahendra8956:

Added EntityOrderFilter to sort entities by different fields/columns. This new filter allows users to specify the order in which entities are displayed in the catalog, replacing the default sorting behavior that only used `metadata.name`.

The EntityOrderFilter provides flexibility to sort by any entity field and control the sort direction. This enables more customized views of the catalog based on user preferences or specific use cases.

### DIFF
--- a/plugins/catalog-react/src/filters.ts
+++ b/plugins/catalog-react/src/filters.ts
@@ -338,13 +338,13 @@ export class EntityErrorFilter implements EntityFilter {
  * @public
  */
 export class EntityOrderFilter implements EntityFilter {
-  constructor(readonly value: EntityOrderQuery) {}
+  constructor(readonly values: [string, 'asc' | 'desc'][]) {}
 
   getOrderFilters(): EntityOrderQuery {
-    return this.value;
+    return this.values.map(([field, order]) => ({ field, order }));
   }
 
-  filterEntity(_: Entity): boolean {
-    return true;
+  toQueryValue(): string[] {
+    return this.values.flat();
   }
 }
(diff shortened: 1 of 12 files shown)


vinzscam:
thanks @mahendra8956! I will have a look into the details in a bit but the direction looks good! This is actually something that the columns of the table can re-use to apply a different sorting 👍

plugins/catalog-react/src/filters.ts
```diff
@@ -331,3 +332,19 @@ export class EntityErrorFilter implements EntityFilter {
     return error !== undefined && this.value === error;
   }
 }
+
+/**
+ * Sort entities by a given field/column.
+ * @public
+ */
+export class EntityOrderFilter implements EntityFilter {
+  constructor(readonly value: EntityOrderQuery) {}
```
vinzscam, line 341: If possible, it might be useful to make the filter serializable by allowing the constructor to accept a `string`
 or `string[]` as a value. This way, we can serialize the filter in the query parameters and reconstruct it on page load.
 
```suggestion
  constructor(readonly value: [string, 'asc' | 'desc'][]) {}
```

vinzscam (changes requested):
a few minor comments, but this is great. Thank you 🙏

mahendra8956:
Hey @vinzscam 
I have made the suggested changes. Please take a look and let me know if you need further adjustments!

benjdlambert:
@vinzscam do you wanna check over this again? :pay:

vinzscam (approved):
amazing, thank you 🚀
It's a bit late to include it in 1.38 but we'll make sure to merge it after the main release

Issue #28981, opened 2025-02-27
🐛 Bug Report: Readme section on component page shows Error 500 when no "/blob/" in gitlab url


mkoralewski:
### 📜 Description

There is 500 error in readme section of component page when the URL doesn't contain "/blob/"

![Image](https://github.com/user-attachments/assets/1982d376-9a77-40b6-ad4a-8a10d8f6506f)

It seems to be related to this line:

https://github.com/backstage/backstage/blob/master/packages/integration/src/gitlab/core.ts#L120

### 👍 Expected behavior

It should show readme correctly having gitlab url like:
https://<gitlab host>/<group>/<subgroup>/<repository>/

### 👎 Actual Behavior with Screenshots

In developer tools we can see that backstage is doing GET request 
https://<backstage host>/api/readme/component/default/<component name> and it receives 500 error with message:
{"error":"Readme failure: Error: Failed converting /<group>/<subgroup>/<repository>/README to a project id. Url path must include /blob/."}

### 👟 Reproduction steps

1. Create component with annotation 
backstage.io/source-location: url:https://<gitlab host>/<group>/<subgroup>/<repository>
2. Go to component site on backstage and see Readme section- it will be Error 500
3. If you add "-/blob/main/" at the end of your url in backstage.io/source-location then Readme is loaded correctly.

### 👀 Have you spent some time to check if this bug has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

camilaibs:
Hi @mkoralewski 👋🏻 ,
In conversation with @backstage/maintainers, this is the way we expect it to work, passing it through source location annotation...

mkoralewski:
Hi @camilaibs 
Thank you for comment. I think it's very unintuitive, but if for some reason it has to be implemented this way, then maybe you can describe it in documentation here: https://backstage.io/docs/features/software-catalog/well-known-annotations#backstageiosource-location ?

camilaibs:
Hi @mkoralewski, we agree with you.
I have added the "documentation" and "help-wanted" labels to this issue.
Maybe adding a few words about this on this page would be a good start:
https://backstage.io/docs/integrations/github/locations.
Please let us know if you have a moment to help us update the docs, we would appreciate it 🙂 .

[...]

PR #29312, opened 2025-03-20
Docs: add note about correct readme path in GitLab integration

starlightknown:

added a note in GitLab integration about correct readme path when using backstage.io/source-location annotation

https://github.com/backstage/backstage/issues/28981


docs/integrations/gitlab/locations.md
```diff
@@ -21,6 +21,18 @@ integrations:
 
 :::note Note
 
+When using the [`backstage.io/source-location`](../../features/software-catalog/well-known-annotations.md#backstageiosource-location): `url:https://{gitlab host}/{group}/{subgroup}/{repository}/` annotation to create a component, you must include `/blob/main/` in the URL path. For example:
```
Parsifal-M, line 24: Heya! 👋 

Sorry for not getting around to this sooner, we use Gitlab internally and we've never had to manually set the `backstage.io/source-location`, as far as I can see it does automatically add the `/blob/main/` parts when you import something via the catalog.

I'm just curious how this came about, because I also see you mention that without this, the _**README**_ section of the component page will fail to load, what does this refer to? Just trying to puzzle it together 😅 

LMK! 👏

Parsifal-M, line 24: Heya! 👋 

I was just looking at https://github.com/backstage/backstage/issues/28981 I am not sure what the `README` part refers to, I assume maybe its a custom plugin maybe 😅 

But I think the message should be here that if you want to manually set this annotation then it must contain the `/blob/main/` 

AFAIK, If you import something from Gitlab into the catalog via the UI it automatically adds this annotation so I assume in that issue its being manually set for some reason or other.

Sorry for taking so long to look into this also! 😞

starlightknown: hello, I believe the readme section refers to readme but since the link is wrong it will fail to load

Parsifal-M: Hey! 👋 

What I am trying to say is that, I think the plugin (the part that displays the README) is a custom plugin, but correct me if I am wrong!

So I'm worried people who don't use this plugin might read it and get confused "What do you mean the README wont display? What README?" 😅, so I guess the aim of the message should be that if you are setting the annotation yourself (manually) then make sure you include the `/blob/main/` part as its needed, what do you think? Or am I way off here 😆

Issue #20561, opened 2023-10-12
🔌 Plugin: events-backend-module-kafka


Silthus:
### 🔖 Summary

It would be great to have a Kafka event module that can subscribe to topics in a Kafka cluster and forward them as backstage events. KafkaJS could be used for the client.

### 🌐 Project website (if applicable)

https://kafka.js.org/

### ✌️ Context

This would be similiar to the AWSSqS event-module.

### 👀 Have you spent some time to check if this plugin request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

camilaibs:
Sounds great 🎉, looking forward to your contribution!
Would you like us to attribute the issue to you 🙂 ?

Silthus:
That would be great, thank you!

I will put something together in the next weeks.

camilaibs:
Assigned to you @Silthus , thanks in advance 🙌🏻

ZaradarBH:
@Silthus are you still working on this? I was contemplating a similar requirement for a eventprovider that needs to consume messages from a kafka broker and it would be much cleaner if my custom backend module just used the eventsservice as a facade and then we had a module like the one your envisioning that can pickup on the presence of the kafka configuration section in app-configuration and start consuming events from topics marked as auto-ingestible or something like that. Feel free to hit me up on discord for a chat if you want to explore the subject further. :)

Silthus:
Hey @ZaradarBH 

Cool to see someone else is interested in this 😃 

Unfortunately I am working on a different company now where we don't use Kafka and I don't have time on the side to tackle this.

Maybe someone from my old team wants to chime in and pair? @Roshick ?

ZaradarBH:
@Silthus thanks for the input! I can do the work on my own if needs be. I just figured it would be fun if others chimed in, so @Roshick feel free to ping me if you want to talk :)

pjungermann:
@Silthus @ZaradarBH 
As the AWS SQS support was mentioned: Did you think about this as a source for external events or a possible alternative implementation for `EventsService`?

AWS SQS is used as a source for external events as an alternative to the HTTP endpoints. However, we might want to use something similar within an alternative `EventsService` implementation using i.e. AWS EventBridge or others.

[...]

PR #29315, opened 2025-03-20
feat(events): add a new kafka module for events backend

Jonas-Beck:

This introduces a new kafka module for the `plugin-events-backend` that can be used to receive events from a Kafka topic and publish them using the `EventService`, similar to the `events-backend-module-aws-sqs` module.

This module was requested in the following issue: https://github.com/backstage/backstage/issues/20561

We have been using this module in production for the last 4 months without any issues, as an alternative to receiving events [via HTTP Endpoint](https://github.com/backstage/backstage/blob/master/plugins/events-backend/README.md). 

I tried to follow the structure and naming conventions used in the `events-backend-module-aws-sqs` module.


plugins/events-backend-module-kafka/src/publisher/config.ts
```diff
@@ -0,0 +1,96 @@
...
+  const requestTimeout = kafkaConfig.getOptionalNumber('requestTimeout');
+  const enforceRequestTimeout = kafkaConfig.getOptionalBoolean(
+    'enforceRequestTimeout',
+  );
+
+  const ssl = kafkaConfig.getOptional('ssl') as KafkaConfig['ssl'];
+  const sasl = kafkaConfig.getOptional('sasl') as KafkaConfig['sasl'];
+  const retry = kafkaConfig.getOptional('retry') as KafkaConfig['retry'];
+
+  const kafkaConsumerConfig: KafkaConsumerConfig[] = kafkaConfig
```
pjungermann, line 58:
```suggestion
  const kafkaConsumerConfigs: KafkaConsumerConfig[] = kafkaConfig
```

plugins/events-backend-module-kafka/src/publisher/KafkaConsumingEventPublisher.ts
```diff
@@ -0,0 +1,111 @@
...
+/**
+ *
+ * This class subscribes to Kafka topics and publishes events received to the registered subscriber.
+ * The message payload will be used as the event payload and passed to the subscribers.
+ *
+ * @public
+ */
+export class KafkaConsumingEventPublisher {
+  private readonly kafkaConsumer: Consumer;
+  private readonly consumerSubscribeOptions: ConsumerSubscribeTopics;
```
pjungermann, line 32:
```suggestion
  private readonly consumerSubscribeTopics: ConsumerSubscribeTopics;
```

plugins/events-backend-module-kafka/src/publisher/LoggerServiceCreator.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { LoggerService } from '@backstage/backend-plugin-api';
+import { LogEntry, logLevel } from 'kafkajs';
+
+export const LoggerServiceCreator =
```
pjungermann, line 19:
```suggestion
export const LoggerServiceAdapter =
```

plugins/events-backend-module-kafka/src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @packageDocumentation
+ */
+
+export { KafkaConsumerClient } from './publisher/KafkaConsumerClient';
+export { KafkaConsumingEventPublisher } from './publisher/KafkaConsumingEventPublisher';
+export { eventsModuleKafkaConsumingEventPublisher as default } from './service/eventsModuleKafkaConsumingEventPublisher';
+export type {
+  KafkaConsumerConfig,
+  KafkaEventSourceConfig,
+} from './publisher/config';
```
pjungermann, line 32:
```suggestion
export { eventsModuleKafkaConsumingEventPublisher as default } from './service/eventsModuleKafkaConsumingEventPublisher';
```

I think you only need to expose the module and nothing else. This will reduce the API surface a lot, too.

pjungermann (commented):
Thank you! The change looks good to me.

I left a few minor/optional suggestions.
However, the API surface should be reduced.

I have not been able to try it myself, yet.

pjungermann:
Yes, definitely! Having this connector is perfect. Even if there was already a Kafka-based events backend implementation, it would be great to have.

I only wanted to share those thoughts with you. I figured you might be interested later down the road.

plugins/events-backend-module-kafka/src/publisher/config.ts
Jonas-Beck: I updated this in d4b22973

plugins/events-backend-module-kafka/src/publisher/KafkaConsumingEventPublisher.ts
Jonas-Beck: I updated this in 75faf7d5

plugins/events-backend-module-kafka/src/publisher/LoggerServiceCreator.ts
Jonas-Beck: I updated this in 34bdbc4f

Jonas-Beck:
Hi @pjungermann, I forgot to rebase the PR before the github-actions bot closed it as stale. Could you reopen the PR?

yarn.lock
```diff
@@ -35254,6 +35269,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"kafkajs@npm:^2.2.4":
```
freben, line 35272: what an exemplary self contained library :)

plugins/events-backend-module-kafka/src/publisher/config.ts
```diff
@@ -0,0 +1,96 @@
...
+export interface KafkaEventSourceConfig {
+  kafkaConfig: KafkaConfig;
+  kafkaConsumerConfigs: KafkaConsumerConfig[];
+}
+
+const CONFIG_PREFIX_PUBLISHER =
+  'events.modules.kafka.kafkaConsumingEventPublisher';
+
+export const readConfig = (config: Config): KafkaEventSourceConfig => {
+  const kafkaConfig = config.getConfig(CONFIG_PREFIX_PUBLISHER);
```
freben: If I am reading things correctly, this makes the config be required in practice. It's declared as optional in the config.d.ts which is good. It is also convenient if people can safely install the module and it just lies dormant if you haven't added the config yet. So I think it might make sense to use `getOptionalConfig` here and return undefined if there's nothing, and if that's the case the module just doesn't do anything. That's the pattern we have in a couple of places.

plugins/events-backend-module-kafka/src/publisher/config.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  const clientId = kafkaConfig.getString('clientId');
+  const brokers = kafkaConfig.getStringArray('brokers');
+
+  const authenticationTimeout = kafkaConfig.getOptionalNumber(
+    'authenticationTimeout',
+  );
+  const connectionTimeout = kafkaConfig.getOptionalNumber('connectionTimeout');
```
freben: for all types of timeouts / durations, we generally specify the type in config.d.ts as `string | HumanDuration` (check elsewhere in the repo for examples). Then we read it with `readDurationFromConfig` if it exists. This lets people specify both strings like `10m` and object forms like `{ hours: 2 }` in a consistent way without having to know in each instance what the required underlying unit is.

This applies here as well as in related places below. You could for example have `maxWaitTime` instead of `maxWaitTimeInMs` then.

plugins/events-backend-module-kafka/src/publisher/LoggerServiceAdapter.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { LoggerService } from '@backstage/backend-plugin-api';
+import { LogEntry, logLevel } from 'kafkajs';
+
+export const LoggerServiceAdapter =
```
freben:
```suggestion
export const loggerServiceAdapter =
```

since it's not a type

plugins/events-backend-module-kafka/src/publisher/KafkaConsumingEventPublisher.ts
```diff
@@ -0,0 +1,111 @@
...
+      kafkaTopics: config.consumerSubscribeTopics.topics.toString(),
+      backstageTopic: config.backstageTopic,
+      taskId: id,
+    });
+  }
+
+  async start(): Promise<void> {
+    try {
+      await this.kafkaConsumer.connect();
+      this.logger.info('Kafka consumer connected');
```
freben: maybe not needed? or at least should it contain the name of the thing that connected, or what it connected to?

plugins/events-backend-module-kafka/src/publisher/KafkaConsumingEventPublisher.ts
```diff
@@ -0,0 +1,111 @@
...
+        },
+      });
+    } catch (error: any) {
+      this.logger.error('Kafka consumer connection failed ', error);
+    }
+  }
+
+  async shutdown(): Promise<void> {
+    await this.kafkaConsumer.disconnect();
+    this.logger.info('Kafka consumer disconnected');
```
freben: and this one may not be needed really either

Jonas-Beck:
Thanks for the feedback. I will get started updating the PR

plugins/events-backend-module-kafka/src/publisher/KafkaConsumingEventPublisher.ts
Jonas-Beck: The KafkaConsumingEventPublisher creates a child logger that includes the class, groupId, kafkaTopics, backstageTopic, and taskId

But I looked into our current logs and saw that kafkajs also logs `Kafka consumer starting` along with some metadata, so I think it makes sense to remove the log to avoid duplicate logs with almost the same information.

Removed in 551daf9

plugins/events-backend-module-kafka/src/publisher/KafkaConsumingEventPublisher.ts
Jonas-Beck: I have removed this log statement as well - I think if `kafkajs` log the `Kafka consumer starting` message it will also have one for when the consumer is disconnected.

plugins/events-backend-module-kafka/src/publisher/config.ts
Jonas-Beck: Makes sense, I can see the use case for installing the module without adding the configuration at the same time. I’ve updated the readConfig  in 9093d77 to use `getOptionalConfig` and return undefined if nothing is provided.

plugins/events-backend-module-kafka/src/publisher/config.ts
Jonas-Beck: I have updated all timeouts / durations to be `string | HumanDuration`. The `readConfig` function was also updated to now use `readDurationFromConfig` to read the duration and convert it to milliseconds

plugins/events-backend-module-kafka/src/publisher/LoggerServiceAdapter.ts
```diff
@@ -0,0 +1,45 @@
...
+import { LoggerService } from '@backstage/backend-plugin-api';
+import { LogEntry, logLevel } from 'kafkajs';
+
+export const loggerServiceAdapter =
+  (loggerService: LoggerService) => (_level: logLevel) => {
+    return (entry: LogEntry) => {
+      const { namespace, level, log } = entry;
+      const { message, ...extra } = log;
+
+      const logMethods: Record<
```
freben, line 25: this could have been placed inside the outermost function to not have to reconstruct it over and over, but that can be done as a follow-up and is no biggie

plugins/events-backend-module-kafka/src/publisher/config.ts
```diff
@@ -0,0 +1,159 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Config, readDurationFromConfig } from '@backstage/config';
+import { ConsumerConfig, ConsumerSubscribeTopics, KafkaConfig } from 'kafkajs';
+import { Duration } from 'luxon';
+
+/**
+ * @public
```
freben, line 21: There's a lot of occurrences of `@public` where the item actually isn't public. Not really harmful and can be cleaned out as a follow-up.

plugins/events-backend-module-kafka/src/publisher/config.ts
```diff
@@ -0,0 +1,159 @@
...
+  config: Config,
+  key: string,
+): number | undefined => {
+  const humanDuration = config.has(key)
+    ? readDurationFromConfig(config, { key })
+    : undefined;
+
+  if (!humanDuration) return undefined;
+
+  return Duration.fromObject(humanDuration).as('milliseconds');
```
freben, line 57: There's a widely used `durationToMilliseconds` in `@backstage/types` if you want to get rid of the luxon dependency.

freben (approved):
Looking fantastic - just optional comments that are happily left as followups. Thank you! Merging.

Jonas-Beck:
Thank you. I will create a PR for the follow-up comments, as well as one for adding documentation after #30044 is merged.

Do you think that #20561 should be closed now that this has been merged?

PR #29319, opened 2025-03-20
feat: Improve Error Handling for Existing Branches in GitLab Actions

jboeijenga:

When creating a Merge Request using GitLab actions, if the branch already exists in the target repository, the feedback to the user was minimal. Previously, the message "_Please check that your repo does not already contain a branch named ..._" required the user to manually verify the existence of the branch.

This update improves the user experience by:

- Automatically checking if the branch exists before attempting to create it.
- Providing a direct link to the existing branch in the error message.
- Checking for any open Merge Requests for the branch and including the link if available.

This change enhances the developer workflow by reducing unnecessary manual steps and improving visibility into existing branches and MRs.

### DIFF
--- a/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabMergeRequest.ts
+++ b/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabMergeRequest.ts
@@ -414,20 +414,39 @@ which uses additional API calls in order to detect whether to 'create', 'update'
               execute_filemode: file.executable,
             }));
 
-      let createBranch: boolean;
-      if (actions.length) {
-        createBranch = true;
-      } else {
-        try {
-          await api.Branches.show(repoID, branchName);
-          createBranch = false;
-          ctx.logger.info(
-            `Using existing branch ${branchName} without modification.`,
-          );
-        } catch (e) {
-          createBranch = true;
+      let createBranch = actions.length > 0;
+
+      try {
+        const branch = await api.Branches.show(repoID, branchName);
+        if (createBranch) {
+          const mergeRequests = await api.MergeRequests.all({
+            projectId: repoID,
+            source_branch: branchName,
+          });
+
+          if (mergeRequests.length > 0) {
+            // If an open MR exists, include the MR link in the error message
+            throw new InputError(
+              `The branch creation failed because the branch already exists at: ${branch.web_url}. Additionally, there is a Merge Request for this branch: ${mergeRequests[0].web_url}`,
+            );
+          } else {
+            // If no open MR, just notify about the existing branch
+            throw new InputError(
+              `The branch creation failed because the branch already exists at: ${branch.web_url}.`,
+            );
+          }
         }
+
+        ctx.logger.info(
+          `Using existing branch ${branchName} without modification.`,
+        );
+      } catch (e) {
+        if (e instanceof InputError) {
+          throw e;
+        }
+        createBranch = true;
       }
+
       if (createBranch) {
         try {
           await api.Branches.create(repoID, branchName, String(targetBranch));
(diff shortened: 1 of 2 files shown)


plugins/scaffolder-backend-module-gitlab/src/actions/gitlabMergeRequest.ts
```diff
@@ -414,20 +414,39 @@ which uses additional API calls in order to detect whether to 'create', 'update'
...
-          createBranch = false;
-          ctx.logger.info(
-            `Using existing branch ${branchName} without modification.`,
-          );
-        } catch (e) {
-          createBranch = true;
+      let createBranch = actions.length > 0;
+
+      try {
+        const branch = await api.Branches.show(repoID, branchName);
```
acierto, line 420: Can you please wrap it in `checkpoint`?

acierto, line 420: Can you please wrap it in `checkpoint`?

plugins/scaffolder-backend-module-gitlab/src/actions/gitlabMergeRequest.ts
```diff
@@ -414,20 +414,39 @@ which uses additional API calls in order to detect whether to 'create', 'update'
...
+        if (createBranch) {
+          const mergeRequests = await api.MergeRequests.all({
```
acierto, line 422: And this invocation too

acierto (changes requested):

jboeijenga:
@acierto: thanks for the feedback! I wrapped the api calls in checkpoint. Can you check if this is done correctly?

plugins/scaffolder-backend-module-gitlab/src/actions/gitlabMergeRequest.ts
```diff
@@ -414,31 +414,60 @@ which uses additional API calls in order to detect whether to 'create', 'update'
...
-          await api.Branches.create(repoID, branchName, String(targetBranch));
-        } catch (e) {
-          throw new InputError(
-            `The branch creation failed. Please check that your repo does not already contain a branch named '${branchName}'. ${getErrorMessage(
-              e,
-            )}`,
-          );
-        }
-      }
+      const { createBranch } = await ctx.checkpoint({
```
benjdlambert, line 417: @acierto we probably don't want to wrap this in a checkpoint right? Otherwise it's going to cache the value?

Thinking it's better to read from source at this point, as the state might change right?

jboeijenga: I have reverted my latest changes.
@benjdlambert, @acierto: can you verify this MR?

benjdlambert:
@jboeijenga it looks like we're missing a signoff here too :pray:

jboeijenga:
@benjdlambert, @acierto: any change we can get this one merged?

acierto:
I'll leave it to @benjdlambert. It's a matter of fact if we want to cache the executions of API in checkpoint or not. It's kinda the same topic as here: https://github.com/backstage/backstage/pull/29283

benjdlambert:
@acierto yeah, I think it's an interesting point - I wonder what we wanna do long term.

Thinking for this case however though, it's unfortunate to cache state which is mutable by definition and external state does impact what this action should do. It would be pretty unexpected for an action to fail, and then if I change something in remote state and it's cached and still doesn't update unless I restart the task all over again.

Maybe we need someway to restart actions without the checkpoint cache too? or something like that - not sure at the moment though.

Let's ship this as is for now, and iterate :tada:

benjdlambert (approved):

PR #29329, opened 2025-03-21
Gateway plugin

vinzscam:

This PR adds a new plugin designed for organizations that have [split their backend plugins across multiple Backstage deployments](https://backstage.io/docs/backend-system/building-backends/index#split-into-multiple-backends) and implemented a custom Discovery service to resolve backend plugin URLs.

While a custom discovery service handles routing between backend plugins, it doesn't address frontend-to-backend routing without either:

- Hardcoding URLs in the frontend
- Implementing a custom reverse proxy

The Gateway Backend Plugin solves this by providing a centralized routing solution in a dedicated "gateway" Backstage deployment. It routes frontend requests to the appropriate backend plugins using the Discovery service, while prioritizing local plugins when available.

We have used this plugin internally at Spotify to simplify the process of exposing Backstage backend plugins to the frontend and we thought it might be useful to other adopters.


plugins/gateway-backend/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@backstage/plugin-gateway-backend",
+  "version": "1.0.0",
```
freben, line 3:
```suggestion
  "version": "0.0.0",
```

and add a changeset at minor level?

plugins/gateway-backend/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "pluginPackages": [
+      "@backstage/plugin-gateway-backend"
+    ]
+  },
+  "publishConfig": {
+    "access": "public",
+    "main": "dist/index.cjs.js",
+    "types": "dist/index.d.ts"
+  },
+  "private": true,
```
freben, line 16:
```suggestion
```

plugins/gateway-backend/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "lint": "backstage-cli package lint",
+    "prepack": "backstage-cli package prepack",
+    "postpack": "backstage-cli package postpack",
+    "start": "backstage-cli package start",
+    "test": "backstage-cli package test"
+  },
+  "dependencies": {
+    "@backstage/backend-plugin-api": "workspace:^",
+    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/core": "^2.0.0",
```
freben, line 40:
```suggestion
    "@opentelemetry/core": "^1.0.0",
```

this matches the rest of the stuff in the lockfile

plugins/gateway-backend/report.api.md
```diff
@@ -0,0 +1,13 @@
...
+> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
+
+```ts
+import { BackendFeature } from '@backstage/backend-plugin-api';
+
+// @public
+const gatewayPlugin: BackendFeature;
+export default gatewayPlugin;
+
+// (No @packageDocumentation comment for this package)
```
freben, line 12: it's kinda nice to add that to the index.ts

plugins/gateway-backend/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "express": "^4.17.1",
+    "http-proxy-middleware": "^3.0.3"
```
freben, line 42: did you explicitly need v3? i guess we can migrate the rest too separately some day

plugins/gateway-backend/src/router.ts
```diff
@@ -0,0 +1,63 @@
...
+import { getRPCMetadata } from '@opentelemetry/core';
+
+export function createRouter({
+  discovery,
+  instanceMeta,
+}: {
+  discovery: DiscoveryService;
+  instanceMeta: InstanceMetadataService;
+  logger: LoggerService;
+}) {
```
freben, line 30:
```suggestion
}): Handler {
```

perhaps? so you don't need the cast in the plugin file above

plugins/gateway-backend/package.json
vinzscam: I've upgraded the deps in `backstage/package.json` instead

plugins/gateway-backend/package.json
vinzscam: can we start from 1.0.0?

plugins/gateway-backend/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+A plugin for managing request routing in distributed Backstage deployments.
+
+## Overview
+
+This plugin is designed for organizations that have [split their backend plugins across multiple Backstage deployments](https://backstage.io/docs/backend-system/building-backends/index#split-into-multiple-backends) and implemented a custom Discovery service to resolve backend plugin URLs.
+
+While a custom discovery service handles routing between backend plugins, it doesn't address frontend-to-backend routing without either:
+
+- Hardcoding URLs in the frontend
```
freben, line 11: need to add "hardcoding" (lowercase) to vane's word list

plugins/gateway-backend/package.json
freben: But now you've tied your feature PR with that pretty large otel bump. If the library changed significantly and emits metrics differently in unexpected ways it might have been better to make that bump separately. Maybe it's ok but just noting.

plugins/gateway-backend/package.json
freben: I guess same here though, when making the feature PR it's usually a good idea to make the yarn.lock change as small as is humanly possible, and then doing bumps separately. It's just that, when reviewing, it's easy to get stuck up on details like this because like "what if this has a side effect on our next deployment of the catalog so it has broken / changed metrics" etc, maybe unnecessarily but we don't really know.

You don't have to change this now, just mentioning

plugins/gateway-backend/report.api.md
freben: Ping on this

plugins/gateway-backend/package.json
freben: If you explicitly want to start at 1.0 without room for letting it mature for a while at 0 - sure, it's a choice we can make. We've traditionally always started at 0.1.0 but it's a matter of taste and how secure you feel about probably not having to make breaking changes to the interface

plugins/gateway-backend/package.json
vinzscam: Fixed by #29467

aramissennyeydd (approved):
Nice, love the simplicity!

yarn.lock
```diff
@@ -6733,6 +6733,23 @@ __metadata:
...
+  version: 0.0.0-use.local
+  resolution: "@backstage/plugin-gateway-backend@workspace:plugins/gateway-backend"
+  dependencies:
+    "@backstage/backend-app-api": "workspace:^"
+    "@backstage/backend-defaults": "workspace:^"
+    "@backstage/backend-plugin-api": "workspace:^"
+    "@backstage/backend-test-utils": "workspace:^"
+    "@backstage/cli": "workspace:^"
+    "@opentelemetry/api": "npm:^1.9.0"
+    "@opentelemetry/core": "npm:^1.30.0"
```
freben, line 6746: can this even be ^1.29.0? That reduces the lockfile change even more

plugins/gateway-backend/src/router.ts
```diff
@@ -0,0 +1,63 @@
...
+  const localPluginIds = new Set(
+    instanceMeta
+      .getInstalledFeatures()
+      .filter(f => f.type === 'plugin')
+      .map(f => f.pluginId),
+  );
+
+  const proxy = createProxyMiddleware({
```
Rugvip, line 38: Wonder if it might be worth bringing in the fix from https://github.com/backstage/backstage/pull/29498 here too?

Rugvip (approved):
One minor bit, but good to go from my PoV :shipit:

yarn.lock
freben: you'll need to check out the unchanged master yarn.lock first and then rerun yarn install again, because it's already resolved to the double block of both 1.29.0 and 1.30.1 below

plugins/gateway-backend/src/router.ts
vinzscam: mmm [this test](https://github.com/backstage/backstage/pull/29498/files#diff-282f493556b02275f0ae86b16dc6b3f40b6e970673ed142fad5d23f13493930bR609) doesn't seem to cover the entire e2e flow. I will try to see if I can quickly test it properly

vinzscam: alright added https://github.com/backstage/backstage/pull/29329/commits/e6947e1790f540fc2f7e6b63a530fab2b687205a 🚀

PR #29343, opened 2025-03-21
Removes domain from being set with cookies in OAuth flows

### LINKED ISSUE #28126: 🐛 Bug Report: Cookie domain causing subdomain login issues
### 📜 Description

Our dev environment is a subdomain of our prod environment. If a prod refresh token cookie is present, the user gets into an infinite look when trying to sign into dev because it attempts to use the prod refresh token, which fails. This is due to the "domain" value in the refresh token cookie being set to the prod domain, which allows the cookie on subdomains. We would like to be able to customize the cookie domain to prevent the cookie from being used on subdomains (e.g., set it to blank), however there is no way to configure a custom [CookieConfigurer](https://backstage.io/docs/reference/plugin-auth-node.cookieconfigurer/).

### 👍 Expected behavior

Signing in on a subdomain uses that subdomain's cookie only.

Sign-in attempts to use the prod cookie on dev, resulting in an infinite loop on the sign-in page.

### 👟 Reproduction steps

1. Login on an instance of Backstage, e.g., example.com
2. Go to another Backstage instance on a subdomain, e.g., dev.example.com
3. The cookie from example.com will be used on sign-in, and if the auth servers are different, this will result in an error.

Since our dev environment is a subdomain of prod, our users are not able to switch between the two environments easily since they must delete their prod cookies to use dev. Ideally we'd be able to customize the CookieConfigurer to the values we need.

### 🖥️ Your Environment

Browser: Version 129.0.6668.58

Backstage info: 

OS:   Darwin 22.4.0 - darwin/arm64
node: v18.18.2
yarn: 4.5.0
cli:  0.27.1 (installed)
backstage:  1.31.3

sana-malik:

This PR fixes #28126, where the domain was set by default and allowed cookies to be shared across subdomains. This PR removes `domain` from the default cookie configurer, which results in the cookie only being usable on the host and not on subdomains.

- [ ] A changeset describing the change and affected packages. ([more info](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md#creating-changesets)) **TODO**

### DIFF
--- a/plugins/auth-node/src/oauth/OAuthCookieManager.ts
+++ b/plugins/auth-node/src/oauth/OAuthCookieManager.ts
@@ -47,7 +47,7 @@ const defaultCookieConfigurer: CookieConfigurer = ({
     ? pathname.slice(0, -'/handler/frame'.length)
     : `${pathname}/${providerId}`;
 
-  return { domain, path, secure, sameSite };
+  return { path, secure, sameSite };
 };
 
 /** @internal */
@@ -152,6 +152,16 @@ export class OAuthCookieManager {
     };
     const req = res.req;
     let output = res;
+
+    // clear cookie with domain due to a bug during sign-in with redirect flow
+    // and invalid cookie being present that causes infinite redirect loop
+    const { hostname: domain } = new URL(this.options.callbackUrl);
+
+    output = output.cookie(name, '', {
+      ...this.getRemoveCookieOptions(),
+      domain,
+    });
+
     if (val.length > MAX_COOKIE_SIZE_CHARACTERS) {
       const nonChunkedFormatExists = !!req.cookies[name];
       if (nonChunkedFormatExists) {
--- a/plugins/auth-node/src/types.ts
+++ b/plugins/auth-node/src/types.ts
@@ -389,7 +389,7 @@ export type CookieConfigurer = (ctx: {
   /** The origin URL of the app */
   appOrigin: string;
 }) => {
-  domain: string;
+  domain?: string;
   path: string;
   secure: boolean;
   sameSite?: 'none' | 'lax' | 'strict';


plugins/auth-node/src/oauth/OAuthCookieManager.ts
```diff
@@ -152,6 +152,36 @@ export class OAuthCookieManager {
...
+      name,
+    );
+    if (chunkedFormatExists) {
+      for (let chunkNumber = 0; ; chunkNumber++) {
+        const key = OAuthCookieManager.getCookieChunkName(name, chunkNumber);
+        const exists = !!req.cookies[key];
+        if (!exists) {
+          break;
+        }
+        output = output.cookie(key, '', {
```
sana-malik, line 172: not sure if always sending a cookie to clear the old domain is the right choice and would like some feedback here

plugins/auth-node/src/oauth/OAuthCookieManager.ts
```diff
@@ -152,6 +152,16 @@ export class OAuthCookieManager {
...
     let output = res;
+
+    // clear cookie with domain due to a bug during sign-in with redirect flow
+    // and invalid cookie being present that causes infinite redirect loop
+    const { hostname: domain } = new URL(this.options.callbackUrl);
+
+    output = output.cookie(name, '', {
+      ...this.getRemoveCookieOptions(),
+      domain,
+    });
```
Rugvip, line 163: Hmm, this is a bit tricky since we can't really assume that we're not setting a domain since a custom cookie configurer might be installed. If there's a different solution that's fairly simple that might be better. Otherwise we might want to check what we're actually getting back from `.getRemoveCookieOptions()` and potentially skip this if a domain is set.

Rugvip (commented):
Yep seems like a good direction 👍

Need a changeset and whatnot, but please mark as ready for review to get this queued up

coreydaley:
Maybe it would be worthwhile to have an option in the app-config that can enable/disable this functionality in case anyone is relying on it currently? Maybe leave this default behavior on by default and then an option to disable it? Thoughts?

sana-malik:
This is still active.

coreydaley:
@Rugvip Can you open this back up please? I think it is still being iterated on and waiting for a response to https://github.com/backstage/backstage/pull/29343#discussion_r2044844208

JessicaJHee:
Hi @sana-malik! Thanks for the great work on this! I'm interested in helping get this fix merged.

If you're no longer working on this, would you mind if I took it over? I can open a new PR with your changes and handle the remaining work from there?

plugins/auth-node/src/oauth/OAuthCookieManager.ts
sana-malik: As an update here -- I haven't been able to find a good way to address the infinite-looping issue on the frontend. The only approach that has worked is to set a query parameter after sign-in, to show that we've had a successful sign-in, and if we attempt to sign in again, we should stop the loop. This isn't great since the parameter shows up in the URL to the user.

sana-malik:
> Hi @sana-malik! Thanks for the great work on this! I'm interested in helping get this fix merged.
> 
> If you're no longer working on this, would you mind if I took it over? I can open a new PR with your changes and handle the remaining work from there?

Hi @JessicaJHee -- happy for you to move it forward! Feel free to use this as a base. The main fix is already done, the only issue is a side effect changing the cookie -- having an invalid cookie with the same name, but different domain causes it to be erroneously used and causes infinite looping at sign-in when the redirect flow is on.

Issue #29339, opened 2025-03-21
🐛 Bug Report: Catalog Processing Histogram metrics emit in seconds when buckets are for milliseconds


alecjacobs5401:
### 📜 Description

Default histogram buckets for OpenTelemetry are in **milliseconds**, but the histograms that are created for catalog processing emit metrics in **seconds** ([source](https://github.com/backstage/backstage/blob/v1.37.0/plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts#L415-L461))

This leads to metrics being improperly bucketed and results in lack of proper bucketing scope/true data being shown and data is vastly different than what is displayed from the deprecated prometheus metrics using the same data because of the different bucketing rules.

### 👍 Expected behavior

Consistent seconds buckets should be used for processing metrics that rely on seconds.

### 👎 Actual Behavior with Screenshots

Example difference in consumed metrics (these are p99 for both)

![Image](https://github.com/user-attachments/assets/44078297-841d-4086-b377-b936a3760fa3)

Baseline for opentelemetry is not representative of the actual behavior due to bucketing

### 👟 Reproduction steps

Can be reproduced in a local `yarn node` shell from any Backstage repo that is currently configured with OpenTelemetry as per https://backstage.io/docs/tutorials/setup-opentelemetry/

In one terminal, spawn a node terminal using `yarn node` in the backstage repo (ensure that the `@opentelemetry` packages are installed in the repo) and paste the following:
```js
const { NodeSDK } = require('@opentelemetry/sdk-node');
const {
  getNodeAutoInstrumentations,
} = require('@opentelemetry/auto-instrumentations-node');
const { PrometheusExporter } = require('@opentelemetry/exporter-prometheus');

// By default exports the metrics on localhost:9464/metrics
const prometheus = new PrometheusExporter();
const sdk = new NodeSDK({
  // You can add a traceExporter field here too
  metricReader: prometheus,
  instrumentations: [getNodeAutoInstrumentations()],
});

sdk.start();

const { metrics } = require('@opentelemetry/api')
const meter = metrics.getMeter('default')

const histogramDefaultBuckets = meter.createHistogram('testing.default.buckets', {
  description: 'Testing with explicit boundaries for seconds',
  unit: 'seconds',
});
const histogramSecondsBuckets = meter.createHistogram('testing.seconds.buckets', {
  description: 'Testing with explicit boundaries for seconds',
  unit: 'seconds',
  advice: {
    explicitBucketBoundaries: [
[...]

PR #29345, opened 2025-03-22
feat: otel bucket configuration

alecjacobs5401:

Resolves https://github.com/backstage/backstage/issues/29339

This PR adds in configuration support for OpenTelemetry for plugins that currently define OpenTelemetry Histogram metrics. This configuration allows for custom defined histogram buckets at the individual histogram level or configuring a default for the plugin.

If no configuration is provided, existing behavior takes place.

I'm well aware this may be over-engineered for the use case, but thought it could be nice to have it configuration based since different adopters might have different buckets/minimums we care about. It also prevents this from introducing any real breaking changes by swapping bucket levels out from under (that may be a surprise for some consumers).

I wanted to get this PR opened to iterate on feedback/ideas before I move forward on further providing screenshots/demos or updating any user facing documentation

### DIFF
--- a/.changeset/ripe-ducks-own.md
+++ b/.changeset/ripe-ducks-own.md
@@ -1,0 +1,8 @@
+---
+'@backstage/plugin-scaffolder-backend': patch
+'@backstage/backend-defaults': patch
+'example-backend-legacy': patch
+'@backstage/plugin-catalog-backend': patch
+---
+
+Support configuration for OpenTelemetry Histogram Buckets on OpenTelemetry histogram metrics
--- a/packages/backend-defaults/config.d.ts
+++ b/packages/backend-defaults/config.d.ts
@@ -697,6 +697,32 @@ export interface Config {
         paths?: string[];
       }>;
     };
+
+    /**
+     * Configuration for the task scheduler
+     */
+    scheduler?: {
+      /**
+       * Defines configuration options for OpenTelemetry metrics
+       */
+      opentelemetry?: {
+        metrics?: {
+          /**
+           * Allows for configuring the default buckets for all histogram metrics or for specific histogram metrics.
+           * Learn more about histograms at https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram.
+           *
+           * @remarks
+           *
+           * Configuration for the non 'default' keys will take precedence over 'default'. If no configuration is provided,
+           * falls back to the OpenTelemetry default buckets of [0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000].
+           */
+          histogramBuckets?: {
+            default?: Array<number>;
+            duration?: Array<number>;
+          };
+        };
+      };
+    };
   };
 
   /**
--- a/packages/backend-defaults/report-scheduler.api.md
+++ b/packages/backend-defaults/report-scheduler.api.md
@@ -5,6 +5,7 @@
 ```ts
 import { DatabaseService } from '@backstage/backend-plugin-api';
 import { LoggerService } from '@backstage/backend-plugin-api';
+import { RootConfigService } from '@backstage/backend-plugin-api';
 import { RootLifecycleService } from '@backstage/backend-plugin-api';
 import { SchedulerService } from '@backstage/backend-plugin-api';
 import { ServiceFactory } from '@backstage/backend-plugin-api';
@@ -16,6 +17,7 @@ export class DefaultSchedulerService {
     database: DatabaseService;
     logger: LoggerService;
     rootLifecycle?: RootLifecycleService;
+    config?: RootConfigService;
   }): SchedulerService;
 }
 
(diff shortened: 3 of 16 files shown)


Rugvip (changes requested):
Thank you for pushing this area forward! 👍 This is something that we know is needed and have discussed in the past, but haven't yet prioritized working on directly in the maintainer team.

We're at the stage with observability in Backstage where we want to avoid any more one-off solutions, and instead move to something like a core `MetricsService` of whatever might be needed to make that happen. This includes user configuration of the metrics collection, such as histogram buckets.

What that means for this PR is that we'd like to avoid any specific solutions within plugins for how to manage metrics, but instead lift it up to the framework level and basically provide this to plugins for free. Happy to discuss potential paths forward for that if this is something you want to help drive, as we don't have capacity to prioritize this in the maintainer team just yet.

alecjacobs5401:
@Rugvip Totally understand, as I was putting this together I was thinking "this really feels like its time for a form of MetricsService", but didn't want to boil the ocean in this PR.

What does that mean for https://github.com/backstage/backstage/issues/29339? Right now, its pretty hard to trust the metrics that are coming from histograms since all of the outputs are in seconds but the bucket configurations are for milliseconds. This works fairly well for some longer running tasks, but for others it just makes the metrics harder to reason about. Its definitely not a world breaking incompatibility, but it makes having an actionable SLO on things like processing latency difficult since real world behavior likely won't match what the metrics show.

Are there concerns from the maintainers end of just hardcoding updated buckets for the catalog processing metrics? Its _technically_ a breaking change since the change may be surprising, but happy to roll with it if the team feels its prudent!

alecjacobs5401:
bumping this

@Rugvip Wanted to follow up here - I'm not sure I have the bandwidth currently to fully implement a Metrics Service. If there is a path forward for resolving the linked issue, I'd be happy to push it forward as well

alecjacobs5401:
Bumping this to keep it alive, will resolve conflicts

Rugvip:
Sorry no news on being able to move `MetricService` forward, we don't have any capacity to move that forward on our side for the time being.

For working around #29339 I think it would be alright to configure some hardcoded buckets for now so that we ship better defaults. Only thing there is whether or not they can still be overridden via the `views` option to `new NodeSDK`, although even if that's not the case I think it might be better to opt for sensible hardcoded bucket configs. Do you happen to know how that all ends up wired together @benjdlambert?

benjdlambert:
You can give the metrics `bucket` hints when you create them, and then you can also override them with the `views` config too AFIAK.

https://github.com/open-telemetry/opentelemetry-js/pull/3876

Should allow you to do something like:

```ts
    this.test = meter.createHistogram('catalog.test', {
      advice: {
        explicitBucketBoundaries: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
      }
    })
```

and then views config to override in an otel metrics plugin:

```ts
export const sdk = new NodeSDK({
  metricReader: new PrometheusExporter({
  preventServerStart: true,
});
  views: [
      new View({
        instrumentName: 'catalog.test',
        aggregation: new ExplicitBucketHistogramAggregation([
          0.01, 0.1, 0.5, 1, 5, 10, 25, 50, 100, 500, 1000,
        ]),
     })
   ]
})
```

alecjacobs5401:
Oh! There goes my lack of knowledge about OTEL showing. Thanks for the pointers there about the views stuff, I'll dig in there and see what we can do!

alecjacobs5401:
Great news! This turned out to work exactly was I needed - not sure how I missed this before with my searching on how to override things!

I went with a more general approach if it helps anyone in the future:
```ts
export const sdk = new NodeSDK({
  metricReader: new PrometheusExporter({
    preventServerStart: true,
  }),
  views: [
    new View({
      instrumentType: InstrumentType.HISTOGRAM,
      instrumentUnit: 'seconds',
      aggregation: new ExplicitBucketHistogramAggregation([
        0, 0.01, 0.05, 0.1, 0.25, 0.5, 1, 2, 5, 10, 30, 60, 120, 300, 1000,
      ]),
    }),
  ],
});
```

acierto:
hey @alecjacobs5401 👋 

Can you please resolve conflicts?

awanlin:
Hi @alecjacobs5401, just curious on the plan forward with this PR, given that you can use the `views` feature to get what you need is it safe to assume that this PR as is won't move forward? If that's right could it be repurposed to add this to the [OpenTelemetry tutorial](https://github.com/backstage/backstage/blob/master/docs/tutorials/setup-opentelemetry.md) so that way others know about this config option?

alecjacobs5401:
@awanlin Correct! I can close this one out and then move that to a separate PR, but I'm not sure I'll have immediate bandwidth to do so soon.

awanlin:
If I submit a PR with the changes would you have time to review?

alecjacobs5401:
Now that I could do for you!

PR #29353, opened 2025-03-23
refactor: experimental checkpoint types

kurtaking:

This PR includes new type definitions for the task checkpoint functionality and aims to reduce duplication where possible. I worked on this as I've been doing research on how the [rollback BEP](https://github.com/backstage/backstage/pull/29196) fits into the existing design. 

### DIFF
--- a/plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
+++ b/plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
@@ -50,7 +50,6 @@ import {
 import { createConditionAuthorizer } from '@backstage/plugin-permission-node';
 import { actionExecutePermission } from '@backstage/plugin-scaffolder-common/alpha';
 import {
-  CheckpointContext,
   TaskContext,
   TemplateAction,
   TemplateFilter,
@@ -61,7 +60,10 @@ import { scaffolderActionRules } from '../../service/rules';
 import { createCounterMetric, createHistogramMetric } from '../../util/metrics';
 import { BackstageLoggerTransport, WinstonLogger } from './logger';
 import { convertFiltersToRecord } from '../../util/templating';
-import { CheckpointState } from '@backstage/plugin-scaffolder-node';
+import {
+  CheckpointState,
+  CheckpointContext,
+} from '@backstage/plugin-scaffolder-node/alpha';
 
 type NunjucksWorkflowRunnerOptions = {
   workingDirectory: string;
--- a/plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
+++ b/plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
@@ -50,6 +50,7 @@ import {
 import { createConditionAuthorizer } from '@backstage/plugin-permission-node';
 import { actionExecutePermission } from '@backstage/plugin-scaffolder-common/alpha';
 import {
+  CheckpointOptions,
   TaskContext,
   TemplateAction,
   TemplateFilter,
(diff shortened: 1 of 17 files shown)


plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
```diff
@@ -397,10 +389,9 @@ export class NunjucksWorkflowRunner implements WorkflowRunner {
           logger: loggerToWinstonLogger(taskLogger),
           logStream: streamLogger,
           workspacePath,
-          async checkpoint<T extends JsonValue | void>(opts: {
-            key?: string;
-            fn: () => Promise<T> | T;
-          }) {
+          async checkpoint<T extends JsonValue | void>(
+            opts: CheckpointContext<T>,
```
kurtaking, line 393: Wish this could be `ctx` here. Might need help wording types if this needs to be `CheckpointOptions` as I used that elsewhere.

plugins/scaffolder-node/src/checkpoints/types.ts
```diff
@@ -0,0 +1,77 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { JsonValue } from '@backstage/types';
+
+/**
+ * The status of a checkpoint, indicating whether it succeeded or failed.
+ *
+ * @public
```
benjdlambert, line 21: I think that these should probably be `@alpha` types exported from a `/alpha` endpoint for now.

Don't really want to add these to the public API until we're ready :pray:

benjdlambert (changes requested):
Hey @kurtaking :wave:

Sorry for the delay in getting round to reviewing this, been a busy few weeks.

One high level thing here that I think we want to do, which is to move the types to a `/alpha` export of `-node` for now.

Checkpoints and Rollbacks are still experimental, and the types should also be marked as such,.

kurtaking:
> @kurtaking re-opened this 🙏

Hey @benjdlambert 👋🏼  hopefully we can 🚢 this week. I rearranged the types to come from `alpha` exports, but am still working through errors in the api reports. I will be trying to sporadically chip away at this over the next few days.

plugins/scaffolder-node/report.api.md
```diff
@@ -474,20 +472,10 @@ export interface TaskContext {
   spec: TaskSpec;
   // (undocumented)
   taskId?: string;
+  // Warning: (ae-forgotten-export) The symbol "UpdateTaskCheckpointOptions" needs to be exported by the entry point index.d.ts
```
acierto, line 475: @kurtaking Seems this has to be fixed

plugins/scaffolder-node/src/tasks/index.ts
```diff
@@ -25,4 +25,5 @@ export type {
   TaskContext,
   TaskEventType,
   TaskStatus,
+  // UpdateTaskCheckpointOptions,
```
acierto, line 28: 👀

kurtaking:
Not stale. Super bummed I haven't had the time to get back to this.

kurtaking:
hey @benjdlambert - the API reports are unhappy because we are modifying the public interface `ActionContext` by updating the inlined types with `CheckpointContext`. It wants `CheckpointContext` to be exported from the main entry point rather than `/alpha`

```
  // Warning: (ae-forgotten-export) The symbol "UpdateTaskCheckpointOptions" needs to be exported by the entry point index.d.ts
 ```
 
I tried to relocate the type, but it then cascades warnings since the rest of the types its built on are also in the `/alpha` export. Not really sure where to go from here? 🤔 Any ideas? Lemme know if I need to clarify anything. Thanks.

benjdlambert:
@kurtaking I fixed this up and rebased it on top of a load of changes from this release too.

If you're in a public export you'll need to import it from `@backstage/plugin-scaffolder-node/alpha` instead of the relative import `../alpha/etc..` instead. It's still works, but it's an little unconventional.

I also condensed a few types, just because I don't see that much value in having them as they're better used in a union instead, what do you think?

Think this should pass now :pray:

kurtaking:
> If you're in a public export you'll need to import it from `@backstage/plugin-scaffolder-node/alpha` instead of the relative import `../alpha/etc..` instead. It's still works, but it's an little unconventional.
> 

ah, of course

> I also condensed a few types, just because I don't see that much value in having them as they're better used in a union instead, what do you think?
> 
> Think this should pass now 🙏

This looks go to me. Thank you for helping get it fixed up. Hopefully we can continue iterating from here.

kurtaking:
@acierto - pinging you in case you want to give it another look. @benjdlambert - I'm good to go. 🚀

kurtaking:
Are we still pending one more review from `@backstage/reviewers`?

PR #29357, opened 2025-03-24
Canon - Add anchor reset

cdedreuille:

If a `Text` or `Heading` component is wrapped around a link, the styles were overridden. The only way to fix that is to remove the text-decoration globally. MUI also has this rule in place so we should not see any conflicts there.

### DIFF
--- a/packages/canon/css/components.css
+++ b/packages/canon/css/components.css
@@ -584,7 +584,7 @@
   margin: .375rem 1rem;
 }
 
-a.canon-Link {
+.canon-Link {
   font-family: var(--canon-font-regular);
   color: var(--canon-fg-link);
   cursor: pointer;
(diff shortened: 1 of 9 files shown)


packages/canon/css/components.css
```diff
@@ -584,7 +584,7 @@
   margin: .375rem 1rem;
 }
 
-.canon-Link {
+a.canon-Link {
```
Rugvip, line 587: Seems unfortunate to have `a` in there, it's specific enough as is right? I'm thinking for some weird situations is someone wants to render a disabled link as a `span` or w/e. More generally it seems like a good pattern to keep tags away from component styles

Rugvip (approved):
pre-approved with a Q on the tags in the selectors

packages/canon/css/components.css
cdedreuille: Because we set the `a` tag globally, if we don't do this it will not override it.

cdedreuille: That is unfortunate indeed but I'm not sure we can avoid it unfortunately.

cdedreuille: I found a better way to set global styles to avoid having to specify the `a` tag in class names using `:where(a)`

PR #29363, opened 2025-03-24
Display entity-ref in GroupProfileCard

juliusKB4:

This adds the entity-ref to the `GroupProfileCard` so that groups can easily identify their Group ID.

GroupProfileCard

### DIFF
--- a/plugins/org/src/components/Cards/Group/GroupProfile/GroupProfileCard.tsx
+++ b/plugins/org/src/components/Cards/Group/GroupProfile/GroupProfileCard.tsx
@@ -203,19 +203,17 @@ export const GroupProfileCard = (props: {
                 secondary="Child Groups"
               />
             </ListItem>
-            {stringifyEntityRef(group) && (
-              <ListItem>
-                <ListItemIcon>
-                  <Tooltip title="Group ID">
-                    <PermIdentityIcon />
-                  </Tooltip>
-                </ListItemIcon>
-                <ListItemText
-                  primary={stringifyEntityRef(group)}
-                  secondary="Group ID"
-                />
-              </ListItem>
-            )}
+            <ListItem>
+              <ListItemIcon>
+                <Tooltip title="Entity Ref">
+                  <PermIdentityIcon />
+                </Tooltip>
+              </ListItemIcon>
+              <ListItemText
+                primary={stringifyEntityRef(group)}
+                secondary="Entity Ref"
+              />
+            </ListItem>
             {props?.showLinks && <LinksGroup links={links} />}
           </List>
         </Grid>
--- a/plugins/org/src/components/Cards/Group/GroupProfile/GroupProfileCard.tsx
+++ b/plugins/org/src/components/Cards/Group/GroupProfile/GroupProfileCard.tsx
@@ -150,6 +150,17 @@ export const GroupProfileCard = (props: {
         </Grid>
         <Grid item md={10} xl={11}>
           <List>
+            <ListItem>
+              <ListItemIcon>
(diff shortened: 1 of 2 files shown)


juliusKB4:
> Hi @juliusKB4, just a heads up that changes to these cards have been generally closed to contributions with the recommendation to fork them as needed. Will leave that to the Maintainers to make a final call on though, juts don't want you to be surprised by their decision.
> 
> I'll note that the `entityRef `is listed in the Inspect Entity dialog, here's an example: https://demo.backstage.io/catalog/default/group/boxoffice?inspect=

Thank you for the info!

plugins/org/src/components/Cards/Group/GroupProfile/GroupProfileCard.tsx
```diff
@@ -202,6 +203,19 @@ export const GroupProfileCard = (props: {
                 secondary="Child Groups"
               />
             </ListItem>
+            {stringifyEntityRef(group) && (
```
Rugvip, line 206: The check here isn't needed right? `stringifyEntityRef` will always return a truthy value

plugins/org/src/components/Cards/Group/GroupProfile/GroupProfileCard.tsx
```diff
@@ -202,6 +203,19 @@ export const GroupProfileCard = (props: {
...
+              <ListItem>
+                <ListItemIcon>
+                  <Tooltip title="Group ID">
+                    <PermIdentityIcon />
+                  </Tooltip>
+                </ListItemIcon>
+                <ListItemText
+                  primary={stringifyEntityRef(group)}
+                  secondary="Group ID"
```
Rugvip, line 215: Thinking we should use the term "entity reference", "entity ref", or "group ref" instead. This isn't referred to as an ID in other places

Rugvip (commented):
Thank you!

Was thinking this was redundant if we group doesn't have a title, but realizing there's enough space and we kinda might as well put the entire ref in here. You can already see the ref in the tooltip header as well though but you can't copy it from there.

I do think a more general solution might be useful in the long term though, but we always roll this back if we find a that solution.

plugins/org/src/components/Cards/Group/GroupProfile/GroupProfileCard.tsx
juliusKB4: you're correct, updated/removed

plugins/org/src/components/Cards/Group/GroupProfile/GroupProfileCard.tsx
juliusKB4: @Rugvip updated to "Entity Ref"

juliusKB4:
> Thank you!
> 
> Was thinking this was redundant if we group doesn't have a title, but realizing there's enough space and we kinda might as well put the entire ref in here. You can already see the ref in the tooltip header as well though but you can't copy it from there.
> 
> I do think a more general solution might be useful in the long term though, but we always roll this back if we find a that solution.

@Rugvip made the edits you suggested

vinzscam:
it feels strange to have the entityRef at the bottom, especially below the parent and child groups. Would it make sense to put  it at the top instead?

juliusKB4:
> it feels strange to have the entityRef at the bottom, especially below the parent and child groups. Would it make sense to put it at the top instead?

@vinzscam all the way at the top? above the email?

Rugvip (approved):
wfm 👍

vinzscam (approved):
🙏

PR #29369, opened 2025-03-24
deps: force transitive update of cpu-features to 0.0.10

BethGriggs:

When running `yarn install` on a fresh clone using Node.js 22, you may see a native build failure for cpu-features@0.0.9:

```console
➤ YN0008: │ cpu-features@npm:0.0.9 must be rebuilt because its dependency tree changed
➤ YN0009: │ cpu-features@npm:0.0.9 couldn't be built successfully (exit code 1)
```

Errors are:
```
In file included from ../src/binding.cc:3:
../../nan/nan.h:2548:8: error: no matching member function for call to 'SetAccessor'
 2548 |   tpl->SetAccessor(
      |   ~~~~~^~~~~~~~~~~
/Users/bgriggs/Library/Caches/node-gyp/22.14.0/include/node/v8-template.h:1055:8: note: candidate function not viable: no known conversion from 'v8::AccessControl' to 'PropertyAttribute' for 5th argument
 1055 |   void SetAccessor(
      |        ^
 1056 |       Local<Name> name, AccessorNameGetterCallback getter,
 1057 |       AccessorNameSetterCallback setter = nullptr,
 1058 |       Local<Value> data = Local<Value>(), PropertyAttribute attribute = None,
      |                                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This happens because that version of `cpu-features` isn’t compatible with Node 22’s V8 API. The package is optional, so it doesn’t actually break the `yarn install`, but the error is noisy and can be misleading.

The issue only shows up on first install (when the native module builds), but it can also be reproduced by running `yarn rebuild`.

This PR updates the `yarn.lock` to pull in cpu-features@0.0.10, which builds cleanly and resolves the build error/warning.

This was the most minimally invasive way to fix the issue without doing a full `yarn.lock` regeneration. Open to exploring a wider update if preferred, but this seemed like the safest low-impact change for now.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -26304,13 +26304,13 @@ __metadata:
   linkType: hard
 
 "cpu-features@npm:~0.0.9":
-  version: 0.0.9
-  resolution: "cpu-features@npm:0.0.9"
+  version: 0.0.10
+  resolution: "cpu-features@npm:0.0.10"
   dependencies:
     buildcheck: "npm:~0.0.6"
-    nan: "npm:^2.17.0"
+    nan: "npm:^2.19.0"
     node-gyp: "npm:latest"
-  checksum: 10/05a4ec51fff87fcde9ca91021d9a0e73a34eb190b14163f7394757ce2470a85393bc49985915bc0bc0ad7ac1ac79caee21f8626a82b535262d9b7d012f805d64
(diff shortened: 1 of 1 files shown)


BethGriggs:
Rebased, I still think this is worthwhile as the noisy errors do make it harder to debug/find real failures.

benjdlambert (approved):
Nice - let's go with this :pray:

PR #29370, opened 2025-03-25
added support for multiple group patterns instead of a single one

### LINKED ISSUE #29141: 🚀 Feature: Support list of groups in GitLab Discovery Plugin
### 🔖 Feature description

The feature will support a list of GitLab groups to lookup instead of single one.  At the moment, you need to choose between one group or GitLab entirely.

### 🎤 Context

The current implementation supports only a single group. I can create multiple providers with different groups, but they all share the same activity cache, as seen in the [code](https://github.com/backstage/backstage/blob/8aa63b60beaab4111f682ce8a21e1a1b7230ee85/plugins/catalog-backend-module-gitlab/src/GitLabDiscoveryProcessor.ts#L116C3-L116C66).

With more than 25k projects and a high rate of change, crawling our GitLab instance is not working well. The ability to pick and choose specific GitLab groups would help us gradually roll out Backstage within our organization.

I propose two options.

## Support a list of GitLab group in the configuration
The configuration could look like this:
```yaml
catalog:
  providers:
    gitlab:
      yourProviderId:
        host: gitlab-host 
        branch: main
        [...]
        group:
          - groupa
          - groupb/anothergroup
        entityFilename: catalog-info.yaml # Optional. Defaults to `catalog-info.yaml`
        [...]
```

It means we could add a for loop over [here](https://github.com/backstage/backstage/blob/8aa63b60beaab4111f682ce8a21e1a1b7230ee85/plugins/catalog-backend-module-gitlab/src/GitLabDiscoveryProcessor.ts#L111)

## Support many provider ID with their own cache

No need to support any changes in the configuation, what we would need is to return the name of the provider id over [here](https://github.com/backstage/backstage/blob/8aa63b60beaab4111f682ce8a21e1a1b7230ee85/plugins/catalog-backend-module-gitlab/src/GitLabDiscoveryProcessor.ts#L82)

vaceal:
Added support for multiple group patterns instead of a single one. If at least one group pattern is defined, it will retrieve all groups and filter based on the patterns; otherwise, it will scan all GitLab projects.

Fixed #29141


plugins/catalog-backend-module-gitlab/report.api.md
```diff
@@ -112,6 +112,7 @@ export type GitlabProviderConfig = {
   projectPattern: RegExp;
   userPattern: RegExp;
   groupPattern: RegExp;
```
pjungermann, line 114: Should we deprecate this field as it is included in the field?

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
```diff
@@ -207,10 +208,107 @@ export class GitlabDiscoveryEntityProvider implements EntityProvider {
...
+    return Array.from(uniqueProjects.values());
+  }
+
+  /**
+   * Retrieve a list of projects that match configuration.
+   *
+   * @param group - a full path of a GitLab group, can be empty
+   * @returns An array of project to be processed and the number of project scanned
+   */
+  private async GetProjectsToProcess(group: string) {
```
vinzscam, line 302:
```suggestion
  private async getProjectsToProcess(group: string) {
```

plugins/catalog-backend-module-gitlab/src/lib/types.ts
```diff
@@ -184,12 +184,20 @@ export type GitlabProviderConfig = {
    * Defaults to `[\s\S]*`, which means to not filter anything
    */
   userPattern: RegExp;
+
   /**
-   * Filters found groups based on provided patter.
+   * Filters found groups based on provided pattern.
    * Defaults to `[\s\S]*`, which means to not filter anything
+   *
+   * @deprecated Use groupPatterns that provide a list instead.
```
vinzscam, line 192: isn't groupPattern used by another provider? I guess your goal was to deprecated `group`? 🤔

vinzscam (changes requested):

plugins/catalog-backend-module-gitlab/src/lib/types.ts
vaceal: This is a very good question, the configuration is shared by many providers, not only GitLab?

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
```diff
@@ -207,10 +208,107 @@ export class GitlabDiscoveryEntityProvider implements EntityProvider {
...
+    };
+
+    const groupToProcess = new Map<string, GitLabGroup>();
+
+    if (
+      this.config.groupPatterns !== undefined &&
+      this.config.groupPatterns.length > 0
+    ) {
+      for (const pattern of this.config.groupPatterns) {
+        const groups = paginated<GitLabGroup>(
```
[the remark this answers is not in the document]
vaceal: Very good point, fixed.

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
```diff
@@ -207,10 +208,107 @@ export class GitlabDiscoveryEntityProvider implements EntityProvider {
...
+          ) {
+            groupToProcess.set(group.full_path, group);
+          }
+        }
+      }
+
+      for (const group of groupToProcess.values()) {
+        const tmpRes = await this.GetProjectsToProcess(group.full_path);
+        res.scanned += tmpRes.scanned;
+        res.matches.push(...tmpRes.matches);
```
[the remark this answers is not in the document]
vaceal: It seems there are a couple of ways to merge arrays in JavaScript. I chose the _safest_ approach, where I add each project individually in a for loop.

Would you prefer that I use **concat** instead?

vaceal:
hey @vinzscam let me know if my change are good with you!

docs/integrations/gitlab/discovery.md
```diff
@@ -152,6 +152,9 @@ catalog:
         skipForkedRepos: false # Optional. If the project is a fork, skip repository
         includeArchivedRepos: false # Optional. If project is archived, include repository
         group: example-group # Optional. Group and subgroup (if needed) to look for repositories. If not present the whole instance will be scanned
+        groupPatterns: # Optional. Filters for groups based on a list of patterns. Default, no filters.
```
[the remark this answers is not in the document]
vaceal: Fixed.  Let me know if this is enough.

vaceal:
Hey @awanlin and @vinzscam, let me know if I can resolve your comments!

freben:
Hey sorry, there's a conflict it seems. Could you rebase?

vaceal:
Rebase done, let me know if you need other modification!

plugins/catalog-backend-module-gitlab/src/lib/types.ts
freben: It's probably ok to keep it as proposed in the PR. I agree that it would be nice with just one way of doing things, but `group` is going to be a common, simple option right? Regexes are a bit hard, especially in strings where extra wonky escaping rules apply - so maybe `group` still has a reason to live. But should it be optional? Is it a bit weird to be required to specify a `group` when the `groupPatterns` is what you wanted to use?

freben (commented):
Alright, looks good, just a question. @vinzscam ok?

vaceal:
Hello, I can remove de the deprecation if you like, I just want to avoid confusion around this configuration.  However, the 2 config has a different behavior, one filter before project crawling amd the other after

nebulon42:
Is there something still open preventing this from merging? We would love to see this in an upcoming release. Thanks.

freben:
It's vacation times mostly - we'll get around to it!

plugins/catalog-backend-module-gitlab/report.api.md
```diff
@@ -112,6 +112,7 @@ export type GitlabProviderConfig = {
...
+  groupPatterns: RegExp[];
```
freben: Could prefer this to this to be a single `groupPattern: RegExp | RegExp[]`

.changeset/chubby-results-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-gitlab': minor
```
freben:
```suggestion
'@backstage/plugin-catalog-backend-module-gitlab': patch
```

let's make it a nonbreaking addition (to a 0.x version package)

plugins/catalog-backend-module-gitlab/src/providers/config.ts
```diff
@@ -42,6 +42,10 @@ function readGitlabConfig(id: string, config: Config): GitlabProviderConfig {
   const groupPattern = new RegExp(
     config.getOptionalString('groupPattern') ?? /[\s\S]*/,
   );
+  const groupPatterns: Array<RegExp> =
+    config
+      .getOptionalStringArray('groupPatterns')
```
freben: Any additions to what gets read out of config, needs to be also added to the schema in the corresponding config.d.ts file in the same PR. Burt as mentioned above, I'd prefer it to be `RegExp | RegExp[]` (and `string | string[]` in the config schema) instead.

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.test.ts
```diff
@@ -558,4 +558,125 @@ describe('GitlabDiscoveryEntityProvider - events', () => {
     expect(entityProviderConnection.applyMutation).toHaveBeenCalledTimes(0);
   });
   // EventSupportChange: stop add tests >>>
+
+  // add the following test:
```
freben: this comment seems like it should be deleted

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.test.ts
```diff
@@ -558,4 +558,125 @@ describe('GitlabDiscoveryEntityProvider - events', () => {
...
+    await provider.refresh(logger);
+
+    expect(entityProviderConnection.applyMutation).toHaveBeenCalledWith({
+      type: 'full',
+      entities: [],
+    });
+
+    expect(entityProviderConnection.applyMutation).toHaveBeenCalledTimes(1); // No entities should be applied
+  });
+});
```
freben: You didn't mean to have the closing brace etc here right? It's all wrapped in `describe` blocks typically

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
```diff
@@ -207,10 +208,110 @@ export class GitlabDiscoveryEntityProvider implements EntityProvider {
...
+      const groups = paginated<GitLabGroup>(
+        options => this.gitLabClient.listGroups(options),
+        {
+          page: 1,
+          per_page: 50,
+        },
+      );
+
+      for (const pattern of this.config.groupPatterns) {
+        for await (const group of groups) {
```
freben: i don't think you want to do the _server_ retrieval many times if there are many patterns. This `for` loop should probably be the outer one, and then the inner one should do something like `if (this.config.groupPatterns.some(...` or whatnot

plugins/catalog-backend-module-gitlab/src/lib/types.ts
vaceal: You're absolutely right!  If you're using groupPatterns, the group config should be ignored. I can make them mutually exclusive and update the documentation to clarify that group is optional when groupPatterns is set.

What do you think?

plugins/catalog-backend-module-gitlab/report.api.md
vaceal: I'm fine with this change. However, your proposal will alter the current behavior by filtering groups before retrieving all projects.

That said, I actually think it's more efficient to retrieve the groups first rather than filtering on the projects afterward , especially when you have 20k+ projects on your instance :).

Let me know if that's what you’re aiming for!

freben: Does it? I only see one usage of `groupPattern`, and that's [here](https://github.com/backstage/backstage/blob/e003e92e64c78b90a3c49ae432a7a74b27efe3dd/plugins/catalog-backend-module-gitlab/src/providers/GitlabOrgDiscoveryEntityProvider.ts#L801) which seems to be a post filtering step.

What I'm proposing is to change the code such that it probably instead will be

```ts
// coerce to always be an array for simplicity
const groupPatterns = [this.config.groupPattern].flat();
if (
  this.config.groupPattern.some(p => p.test(group.full_path)) &&
  ...
```

vaceal: Indeed! My mistake , I mixed up org discovery and repository discovery. It would make sense to only import users from filtered projects as well.

I’ll make the change! Thanks!

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
```diff
@@ -207,10 +208,119 @@ export class GitlabDiscoveryEntityProvider implements EntityProvider {
...
+
+    const groupToProcess = new Map<string, GitLabGroup>();
+    let groupFilters;
+
+    if (this.config.groupPattern !== undefined) {
+      const patterns = Array.isArray(this.config.groupPattern)
+        ? this.config.groupPattern
+        : [this.config.groupPattern];
+
+      if (patterns.length === 1 && patterns[0].source === '[\\s\\S]*') {
```
freben, line 243: Ah this hurts a bit :) Feels like something that will explode down the line because someone thought (reasonably) that `.*` was a better placeholder pattern and didn't change it here too.

Feels more sensible to have made `GitlabProviderConfig. groupPattern` to for example have the empty array be defined to mean "match all", or even change its type to like `RegEx[] | undefined` to explicitly mark that yeah it can be left out. The original design of slapping the pattern in there forcibly no matter what, was arguably a bit of an iffy decision.

I'll merge this for now as "probably works", but I would appreciate if you had the opportunity to make a followup patch that clears up this confusion.

plugins/catalog-backend-module-gitlab/src/providers/config.ts
```diff
@@ -39,9 +39,21 @@ function readGitlabConfig(id: string, config: Config): GitlabProviderConfig {
...
+
+  if ((configValue && typeof configValue === 'string') || !configValue) {
+    groupPattern = new RegExp(
+      config.getOptionalString('groupPattern') ?? /[\s\S]*/,
+    );
+  } else if (configValue && Array.isArray(configValue)) {
+    const configPattern = config.getOptionalStringArray('groupPattern') ?? [];
+    groupPattern = configPattern.map(pattern => new RegExp(pattern));
+  } else {
+    groupPattern = new RegExp(/[\s\S]*/);
```
freben, line 54: Note that this theoretically allows _any_ value including numbers, and it'll silently just swallow that and translate it to the match-all pattern. Would have been safer to just check `Array.isArray(configValue)`, and if it matches then do `config.getStringArray` (NOT optional), and otherwise, do `config.getOptionalString` and handle the outcome of that. That would have captured all cases with a bit less code. We can leave that for a follow-up patch too though

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
vaceal: Yes, I will!

PR #29383, opened 2025-03-25
factor JSON schema rendering and scaffolder usage examples out of Act…

mbenson:
…ionsPage

### DIFF
--- a/plugins/scaffolder/report-alpha.api.md
+++ b/plugins/scaffolder/report-alpha.api.md
@@ -324,12 +324,7 @@ export const scaffolderTranslationRef: TranslationRef<
     readonly 'fields.repoUrlPicker.repository.inputTitle': 'Repository';
     readonly 'actionsPage.content.emptyState.title': 'No information to display';
     readonly 'actionsPage.content.emptyState.description': 'There are no actions installed or there was an issue communicating with backend.';
-    readonly 'actionsPage.content.tableCell.name': 'Name';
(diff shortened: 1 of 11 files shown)


plugins/scaffolder/report-alpha.api.md
```diff
@@ -367,6 +362,11 @@ export const scaffolderTranslationRef: TranslationRef<
     readonly 'ongoingTask.hideLogsButtonTitle': 'Hide Logs';
     readonly 'ongoingTask.showLogsButtonTitle': 'Show Logs';
     readonly 'templateEditorForm.stepper.emptyText': 'There are no spec parameters in the template to preview.';
+    readonly 'renderSchema.undefined': 'No schema defined';
```
benjdlambert, line 365: This is technically a breaking change to the translations which I don't think we've come across before.

Wondering what the best practice here is @backstage/maintainers?

Should we call this out in a separate changeset that the keys have moved?

benjdlambert (approved):
OK - let's go with this, thinking that this is probably fine for the changeset for now :pray:

mbenson:
So, approved... Now what? I see no scaffolder SIG this week

benjdlambert:
@mbenson sorry - we're all out at KubeCon this week so things are moving a little slower than usual. I think I was just waiting for this to build before merging. We cancelled the SIG this week because we're all out also.

mbenson:
Next: https://github.com/backstage/backstage/pull/29502

PR #29403, opened 2025-03-26
[TechDocs]: fix entity catalog docs page not loading

PeaWarrior:

### DIFF
--- a/.changeset/common-parrots-wink.md
+++ b/.changeset/common-parrots-wink.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-techdocs-react': patch
+---
+
+Fix catalog entity docs page not loading due to multiple addons data attachment in the New Frontend System.
--- a/plugins/techdocs-react/src/alpha.ts
+++ b/plugins/techdocs-react/src/alpha.ts
@@ -13,8 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import React from 'react';
 import { TechDocsAddonOptions } from './types';
-import { attachComponentData } from '@backstage/core-plugin-api';
+import {
+  attachComponentData,
+  getComponentData,
+} from '@backstage/core-plugin-api';
 import { ComponentType } from 'react';
 import { getDataKeyByName, TECHDOCS_ADDONS_KEY } from './addons';
 import {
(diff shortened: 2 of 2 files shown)


.changeset/common-parrots-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-techdocs-react': patch
+---
+
+Fix catalog entity docs page not loading due to multiple addons data attachment.
```
awanlin, line 5: Might make sense to mention this is part of the New Frontend System

awanlin (changes requested):
Thanks @PeaWarrior, when I first saw this I was very curious as I had not heard of this issue. But seems this is for the New Frontend System. We might want to clarify that in the changeset. 👍

plugins/techdocs-react/src/alpha.ts
```diff
@@ -53,6 +57,19 @@ export const attachTechDocsAddonComponentData = <P>(
   techDocsAddon: ComponentType<P>,
   data: TechDocsAddonOptions,
 ) => {
-  attachComponentData(techDocsAddon, TECHDOCS_ADDONS_KEY, data);
-  attachComponentData(techDocsAddon, getDataKeyByName(data.name), true);
+  const element = React.createElement(techDocsAddon as ComponentType);
```
benjdlambert, line 60: I'm a little confused on the fix here, and apologies for my very little knowledge of how all this fits together.

Both `attachComponentData` and `getComponentData` will be removed when the legacy frontend system goes away, so is there actually a different fix here?

Should we be removing the requirement that this is actually set instead of setting it on New Frontend System methods?

benjdlambert (changes requested):
Hey :wave: apologies for maybe the lack of understanding as to why this is why it is but wondered if there's another fix?

plugins/techdocs-react/src/alpha.ts
PeaWarrior: From what I understand, all extension factories are called at startup. In techdocs case, we're using these 2 [blueprints](https://github.com/backstage/backstage/blob/8ab9af844cea5c467e28cda36e8b7205af647fb9/plugins/techdocs/src/alpha.tsx#L145-L218) which both call `attachTechDocsAddonComponentData`. Since it's being ran twice, the techdocs addons page ends up throwing an error on the entity docs page. This prevents the `attachComponentData` from being attempted twice on the same addon.

https://github.com/user-attachments/assets/5e6bd1dd-b566-4f61-9692-0b46bcc0304c

I am unaware that `attachComponentData` is being removed. That would require a bit of reworking done to techdocs addons which our team doesn't have capacity for atm.

PR #29413, opened 2025-03-28
microsite: faster docusaurus

benjdlambert:
Is this ready yet?

### DIFF
--- a/microsite/package.json
+++ b/microsite/package.json
@@ -23,10 +23,10 @@
     "node-polyfill-webpack-plugin": "^3.0.0"
   },
   "dependencies": {
-    "@docusaurus/core": "^3.1.1",
-    "@docusaurus/plugin-client-redirects": "^3.1.1",
-    "@docusaurus/preset-classic": "^3.1.1",
-    "@docusaurus/types": "^3.1.1",
+    "@docusaurus/core": "^3.7.0",
+    "@docusaurus/plugin-client-redirects": "^3.7.0",
+    "@docusaurus/preset-classic": "^3.7.0",
+    "@docusaurus/types": "^3.7.0",
     "@mdx-js/react": "^3.0.0",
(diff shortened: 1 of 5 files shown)


awanlin:
Hi @benjdlambert, I think the only thing holding this back was clarity if it would work or not, we had to revert it in the past: https://github.com/backstage/backstage/pull/27722. And @schultzp2020 has given it ago here as well but we never gave him clear guidance on moving forward sadly: https://github.com/backstage/backstage/pull/28174

Parsifal-M:
Also just FYI https://github.com/backstage/backstage/pull/29573 🙏

awanlin:
> @awanlin yeah we chatted abit about it at kubecon actually, I just wanted to get some benchmarks to see how fast it would be compared to the current build - but can't seem to get the thing to build as we get a rust panic when building with the new rspack.

That does sound like the issue we had when we turned it on the first time around. Essentially it would work away then suddenly fail with no real details as to why.

microsite/package.json
```diff
@@ -20,15 +20,16 @@
...
   "dependencies": {
-    "@docusaurus/core": "^3.1.1",
-    "@docusaurus/plugin-client-redirects": "^3.1.1",
-    "@docusaurus/preset-classic": "^3.1.1",
-    "@docusaurus/types": "^3.1.1",
+    "@docusaurus/core": "^3.7.0",
+    "@docusaurus/faster": "^3.7.0",
+    "@docusaurus/plugin-client-redirects": "^3.7.0",
+    "@docusaurus/preset-classic": "^3.7.0",
+    "@docusaurus/types": "^3.7.0",
```
slorber, line 31: it would be safer to first upgrade to Docusaurus v3.7, and only then turn Docusaurus faster on in another PR

Doing many things in the same PR makes things harder to troubleshoot

PR #29416, opened 2025-03-28
Scaffolding detail button color fix

Vity01:
image
Icon color now matches header font color. 


plugins/scaffolder-react/src/next/components/TemplateCard/CardHeader.tsx
```diff
@@ -36,6 +36,9 @@ const useStyles = makeStyles<
     display: 'flex',
     justifyContent: 'space-between',
   },
+  detailIcon: {
+    color: ({ cardFontColor }) => cardFontColor,
```
freben, line 40: Wasn't it enough to put an `inherit` color here? Or, perhaps, inside the `TemplateDetailButton` component itself instead if possible. The favourite star next to this icon doesn't have to do any trickery I think?

Vity01: I tried to find a root cause of different color, without success. It's probably taken from <Link>. I didn't find any other difference in usage this icon and Favorite star otherwise. Only this solution worked for me.

awanlin: ➕ I tried to take a swing at this a well and it wasn't a simple task. Agree it looks to be from the Link.

.changeset/social-seals-shout.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/plugin-scaffolder-react': patch
+---
+
+Detail icon - light theme - fix
+
+- Fixes the detail icon in light theme to be visible in proper color (same as favorite star).
```
awanlin, line 7:
```suggestion
Fixes the detail icon in light theme to be visible in proper color (same as favorite star).
```
Then just remove the sentence before. That keeps this nice and concise.

awanlin (changes requested):
Thanks for running with this @Vity01 🚀 

Left a comment about the changeset and can you add an image to the PR for what it looks like using the dark theme as well, ideally this would be the same for both.

benjdlambert (approved):

Issue #26854, opened 2024-09-25
🐛 Bug Report: Bitbucket Cloud integration maxes out at 900 repositories


benjidotsh:
### 📜 Description

We have been struggling with a fluctuating amount of components for a while now, with some components being suddenly unavailable, and we have narrowed it down to an issue with the Bitbucket Cloud integration.

As far as I can tell, this happens because Bitbucket Cloud's API ([specifically this one](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-other-operations/#api-workspaces-workspace-search-code-get)) doesn't go past 1000 repositories.

In page 9, the property `next` is missing. It's possible to go to page 10 by calling the endpoint manually, but as soon as you try to go page 11 it returns "Cannot fetch results past result number 1000".

Same goes for pages with a higher `pagelen`: for some reason the maximum `pagelen` is 373, so if we use a `pagelen` of 373, it works until we try to go to page 3, since it exceeds 1000 results.

I'm not sure how to resolve this since it sounds like a limitation in Bitbucket Cloud's API, so I would love to discuss possible solutions in this issue. As soon as we come to a conclusion, I'd be more than happy to submit a fix.

### 👍 Expected behavior

It should retrieve all repositories

### 👎 Actual Behavior with Screenshots

It actually maxes out at 900 repositories

### 👟 Reproduction steps

- Use [Bitbucket Cloud Locations](https://backstage.io/docs/integrations/bitbucketCloud/locations)
- Have more than 900 repositories in Bitbucket

### 📃 Provide the context for the Bug.

`{"class":"BitbucketCloudEntityProvider","level":"info","message":"Committed 900 Locations for catalog files in Bitbucket Cloud repositories","plugin":"catalog","service":"backstage","target":"bitbucketCloud-provider:default","taskId":"bitbucketCloud-provider:default:refresh","taskInstanceId":"41b49bcf-799c-4697-9980-d843c2419471"}`

### 🖥️ Your Environment

```
OS:   Darwin 24.0.0 - darwin/arm64
node: v20.15.1
yarn: 4.5.0
cli:  0.27.1 (installed)
backstage:  1.31.1

Dependencies:
  @backstage/app-defaults                                          1.5.11
  @backstage/backend-app-api                                       0.7.9, 0.9.3, 1.0.0
  @backstage/backend-common                                        0.17.0, 0.21.7, 0.23.3, 0.24.1, 0.25.0
[...]

PR #29423, opened 2025-03-28
improvement in BitbucketCloudEntityProvider and BitbucketCloudUrlRead…

wmenezes-br:

This pull request resolves issue #29300 

Goal: reduce the number of calls to the Bitbucket Cloud API to reduce the risk of hitting rate limit quotas

* BitbucketCloudEntityProvider, function findExistingLocations
The hierarchy in BitbucketCloud is Workspace > Project > Repository. The findExistingLocations function runs through the projects in a workspace, executing the same query and concatenating the result in the result variable. However, the project is just an attribute for organizing the repositories. If we execute the query only in the workspace, we get the same result, with just one call to the Bitbucket Rest API.

image

* BitbucketCloudUrlReader
The urlReader function returns the content of a file from the repository and uses the Bitbucket API to obtain this content. In the scenario where the backstage receives an event for each push to any repository, the rate limit is quickly reached. The change consists of using git commands, which have a much higher quota than the Rest API. In this scenario, the backstage host needs to have the git client installed.

### DIFF
--- a/packages/backend-defaults/report-urlReader.api.md
+++ b/packages/backend-defaults/report-urlReader.api.md
@@ -130,6 +130,7 @@ export class BitbucketCloudUrlReader implements UrlReaderService {
     integration: BitbucketCloudIntegration,
     deps: {
       treeResponseFactory: ReadTreeResponseFactory;
+      logger: LoggerService;
     },
   );
   // (undocumented)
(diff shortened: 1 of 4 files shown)


pjungermann:
the previous state was without the project-specific filter.

This change was introduced to overcome a non-documented limit in the amount of results that can be returned.

If I remember correctly, it was about 900.

While this may not affect you, we will still need to keep this.

Also, based the discussion back then, the overhead for additional requests to the Bitbucket Cloud API were minimal.
In the overall request volume, you could even say it can be neglected.

Of course, the actual number varies depending on the number of projects you have and the distribution of repositories within them, especially how full the last page per project is.

I will look for the previous PR for reference later today.

pjungermann:
This is the issue #26854 I referred to earlier, and this is the PR #26873.

wmenezes-br:
@pjungermann, I understand the issue regarding the 900 result limit. I believe there will be no impact if the function is only executed in Discovery. I will remove the PR change. 

As for UrlReader, I understand that the change is necessary because a large company will quickly reach the rate limit if, with each commit, the backstage makes a call to the rest API. The limits for git operations are 60,000 requests per hour, while the other rest API endpoints are less than 5,000.

https://support.atlassian.com/bitbucket-cloud/docs/api-request-limits/

 If you disagree with my reasoning, I'm probably missing some knowledge. In this case I would be very grateful if you could guide/recommend a configuration. 

Two or three years ago I had the same problem with Github Enterprise.

PR #29424, opened 2025-03-28
[docs] add branch naming caveat to github discovery integration

lukealbao:

- [ ] n/a Screenshots attached (for UI changes)

👋 Github, like git itself, allows [slashes in branch names](https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names). But the implementation of `GithubDiscoveryProcessor` identifies the catalog path and branch name by [splitting the Location target on slashes](https://github.com/backstage/backstage/blob/6b0cae3fc0665e9d82961fd8e2357f623380c168/plugins/catalog-backend-module-github/src/processors/GithubDiscoveryProcessor.ts#L179).

As a result, it's possible to configure the integration to look for any catalog-info file on the `backstage-poc/main` branch with
```yaml
github:
  invalid:
    organization: example
    catalogPath: /**/catalog-info.yaml
    filters:
      branch: 'backstage-poc/main'
```

And the discovery process will silently discover no files, because it's really looking for

```
branch = backstage-poc
catalogPath = /main/**/catalog-info.yaml
```

I thought a little about how slash-names could be supported, but could not come up with an easy enough, but fully generalizable, solution. So I think there should be documentation to prevent others from going down the rabbit hole I did. 

Thanks!

## Update: validate when reading config
f0c22eb give users a runtime error when using an offending branch name.

### DIFF
--- a/.changeset/clear-pigs-share.md
+++ b/.changeset/clear-pigs-share.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-github': minor
+---
+
+Explicitly rejects branch names containing a slash character
--- a/.changeset/clear-pigs-share.md
+++ b/.changeset/clear-pigs-share.md
@@ -2,4 +2,8 @@
 '@backstage/plugin-catalog-backend-module-github': minor
 ---
 
-Explicitly rejects branch names containing a slash character
+**BREAKING**: Explicitly rejects branch names containing a slash character.
+
+The module now rejects any configuration that contains slashes in branch names. The reason for this is that the ingestion will run into downstream problems if they were let through. If you had configuration with a slash in the branch name in `filters.branch`, your application may fail to start up.
+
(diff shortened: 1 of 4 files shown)


camilaibs (approved):
Thank you for contributing to us 🎉 .

.changeset/clear-pigs-share.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-github': minor
+---
+
+Explicitly rejects branch names containing a slash character
```
freben, line 5:
```suggestion
**BREAKING**: Explicitly rejects branch names containing a slash character.

The module now rejects any configuration that contains slashes in branch names. The reason for this is that the ingestion will run into downstream problems if they were let through. If you had configuration with a slash in the branch name in `filters.branch`, your application may fail to start up.

If you are affected by this, please move over to using branches that do not have slashes in them.
```

freben (approved):

Issue #7109, opened 2021-09-08
catalog model: support additional System relations


brethubbard:
## Feature Suggestion

I would like support for additional relations on the System entity.

## Possible Implementation

On the System entity there could be a `spec.consumesApis` or `spec.dependsOn` relation defined explicitly.

Alternatively, a visibility could be declared on an API entity and then we wouldn't have to declare the system then we could derive the public APIs from the sub-components that are part of the system. This would mean that all API relations would be inferred through a component.

## Context

The docs say

> With increasing complexity in software, systems form an important abstraction level to help us reason about software ecosystems.

Part of understanding those systems is understanding its boundaries. That is covered in part by an API that is part of a system but a critical part of that is surfacing the APIs that system depends on.

Fox32:
That's a discussion we already had before. But I'm having trouble finding it... Maybe I find it again 😁

There were different discussion, either having it as relations, or having a flag or level on the API that marks their level of being "public".

brethubbard:
Oh that's interesting, I spent some time trying to determine if there was guidance on this. Anything that I could find that was related was different enough that I thought a new issue made the most sense. If you can find that conversation I would love to take a look at it!

Fox32:
[This](https://github.com/backstage/backstage/pull/3957) could be an entrypoint into researching the discussion.

freben:
Hi!

Yes, interesting and complex subject! :) As you may have seen, it was always intended that systems should serve as this boundary and that it should clarify which of its constituent APIs were intended for outside consumption. It would also give system owners a clearer overview of the collected dependencies of its constituent parts.

Now it sounds like you are suggesting movement in that direction, which is great. We've been working hard in other areas for a while and not leaving a whole lot of room for thinking about this, but let's ponder it and see where it goes.

There are at least two parts to it: what the input should be, and what the output should be. On the output side, having `providesApi`/`apiProvidedBy` relations between the system and some of its APIs might make a lot of sense. So what would drive the generation of those relations? You suggest a `visibility` declaration on the API side, and I think that's a reasonable approach. It puts the responsibility and power clearly in the hands of the API owner, which may not have the ability to change the system itself as easily. I'd consider naming it `scope` perhaps, with values such as `private` (default), `internal`, and `external` (directly accessed by external clients, e.g. the general public) or something like that. Naming is hard!

[...]

PR #29426, opened 2025-03-30
RFCPR [catalog]: add new relations endpoint with transitive capabilities

Sarabadu:
### General idea 

Provide a new endpoint in the catalog to get related entities with transitive relations. As is now plugins like `org` requires a frontend recursive loop to get aggregated views in entity pages. 

### Use cases

While implementing more cards on the "group" entity page it could become a recurrent  case to create this kind of aggregations

I believe it can also be used for the cases pointed on https://github.com/backstage/backstage/issues/7109, [transitive relations](https://github.com/backstage/backstage/issues/14574) without needed to add more relation metadata 

#### System `consumesApis` 

querying with this parameters:
```yaml
systemRef: system/default:systemX
relationsToConsider: ['partOf', 'consumesApi'] # if any chid have this relations looks into the related entities

```
we will get all this entities
```mermaid
graph TD;
    s1[systemX]--| hasPart/partOf |--> component1
    s1--| hasPart/partOf |-->  component2
    component1--| consumeApi |-->  api4
    component2--| hasPart/partOf |-->  subcomponent3
    subcomponent3--| consumeApi |-->  api5
    subcomponent3--| providesApi |-->  api1

   system2--| hasPart/partOf |-->component4
   system2--| hasPart/partOf |-->component5
   component4 --| providesApi |-->api4
   component5 --| providesApi |-->api5

    style api4 fill:#36baa2,stroke:#f66,stroke-width:1px,color:#111
    style api5 fill:#36baa2,stroke:#f66,stroke-width:1px,color:#111
    style s1 fill:#36baa2,color:#111
    style component1 fill:#36baa2,color:#111
    style component2 fill:#36baa2,color:#111
    style subcomponent3 fill:#36baa2,color:#111
```

applying a filter
```yaml
systemRef: system/default:systemX
relationsToConsider: ['partOf', 'consumesApi'] # if any chid have this relations looks into the related entities
filter: 
  kind: Api
```
we get only the Apis
```mermaid
graph TD;
    s1[systemX]--| hasPart/partOf |--> component1
    s1--| hasPart/partOf |-->  component2
    component1--| consumeApi |-->  api4
    component2--| hasPart/partOf |-->  subcomponent3
    subcomponent3--| consumeApi |-->  api5
    subcomponent3--| providesApi |-->  api1

   system2--| hasPart/partOf |-->component4
   system2--| hasPart/partOf |-->component5
   component4 --| providesApi |-->api4
   component5 --| providesApi |-->api5

    style api4 fill:#36baa2,stroke:#f66,stroke-width:1px,color:#111
    style api5 fill:#36baa2,stroke:#f66,stroke-width:1px,color:#111

```

### TODO
- get some feedback 😄 
- look on naming things 
- add more test cases
- 


freben:
Hi!

One initial reaction is that we generally want to keep the server side interface small, and if complex graph traversal is needed it's probably better to build outside of the catalog itself. An efficient traversal usually can typically get its job done in very few round trips anyway. And whatever we build into a server query would risk solving only a subset of problems and quickly get a pretty complex API surface when people want more.

I took a stab some time ago [in this branch](https://github.com/backstage/backstage/compare/master...freben/traverse-catalog) to build a traversal function that takes an initial set (that can be a ref or a query or whatever), an "edge handler" so to speak, and a collector. It's DataLoader based to get efficient batch based traversal in "layers".

Any feedback on that?

Sarabadu:
Thanks @freben for the quick feedback.
I will look closely to you branch there, I will need more time to follow it. my my initial reaction is that solving it in the catalog-client make it not posible to use this logic outside backstage, right? 

I totally get your point, my first though was to make the `org` hooks exported to make it posible to reuse. But looking at the `relations` table push me to see if we can get a more in place solution without needing to do multiple roundtrips in the client side.

I think for many `XGroupEntityCard` or `XSystemEntityCard` could make use of "aggregated views" but it feels wrong to do all the roundtrips for each card. 
I saw also the rbac community plugin is trying to solve a similar problem by getting access to the relation table directly.

I would see you approach, maybe its good to start on the client side and depending on the feedback or usage we could look on optimisation.

PR #29445, opened 2025-04-02
refactor: consolidate entity route param extraction in catalog-react

tklever:

`entityRouteParams` and `useEntityRoute` both extract entity route parameters for use with an entity route reference.
However, they two methods are different in minor ways.

- `useEntityRoute` will accept an `Entity` or an entity reference, `entityRouteParams` only accepts `Entity`s
- `useEntityRoute` URI encodes route params, `entityRouteParams` does not.
- `entityRouteParams` returns a readonly `const` data structure

In practice, this means routes generated by `useEntityRoute(entity)` and
`useRouteRef(entityRouteRef)(entityRouteParams(entity))` are most likely functionally equivalent, but not strictly
equal. Additionally, you might have to choose to use one or the other based on the data available in scope (entity
references).

This change refactors the two functions to centralize and consolidate similar functionality. It widens
`entityRouteParams` to accept entity references and it allows for configurable URI encoding. The encoding is disabled
by default to preserve backwards compatibility.

### DIFF
--- a/.changeset/dull-masks-occur.md
+++ b/.changeset/dull-masks-occur.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-react': minor
+---
+
+`entityRouteParams` now also accepts entity refs, and can help with encoding the resulting parameters.
--- a/plugins/catalog-react/src/components/EntityRefLink/EntityRefLink.tsx
+++ b/plugins/catalog-react/src/components/EntityRefLink/EntityRefLink.tsx
@@ -14,16 +14,11 @@
  * limitations under the License.
  */
 
-import {
-  CompoundEntityRef,
-  DEFAULT_NAMESPACE,
-  Entity,
-  parseEntityRef,
-} from '@backstage/catalog-model';
+import { CompoundEntityRef, Entity } from '@backstage/catalog-model';
 import { Link, LinkProps } from '@backstage/core-components';
 import { useRouteRef } from '@backstage/core-plugin-api';
 import { ReactNode, forwardRef } from 'react';
-import { entityRouteRef } from '../../routes';
+import { entityRouteParams, entityRouteRef } from '../../routes';
 import { EntityDisplayName } from '../EntityDisplayName';
 
 /**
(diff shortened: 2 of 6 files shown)


plugins/catalog-react/src/routes.ts
```diff
@@ -37,16 +42,54 @@ export const entityRouteRef = getOrCreateGlobalSingleton(
...
+};
+
 /**
  * Utility function to get suitable route params for entityRoute, given an
  * @public
  */
-export function entityRouteParams(entity: Entity) {
+export function entityRouteParams(
+  entityOrRef: Entity | CompoundEntityRef | string,
+  options: EntityRouteParamsOptions = { encodeParams: false },
```
freben, line 59: make the assignment (defaults) inside the function body instead - leaving it visible in the contract actually has surprising effects

.changeset/dull-masks-occur.md
```diff
@@ -0,0 +1,19 @@
+---
+'@backstage/plugin-catalog-react': minor
+---
+
+`entityRouteParams` and `useEntityRoute` both extract entity route parameters for use with an entity route reference.
```
freben, line 5: Changesets are for end users to understand the "what" of the changes made, not so much the "why" or targeting developers of the code itself.

This changeset could be condensed down to something along the lines of

```
`entityRouteParams` now also accepts entity refs, and can help with encoding the resulting parameters.
```

plugins/catalog-react/src/routes.ts
tklever: Removed from the contract.

Additionally, If you have a link or a doc on those surprising effects, I'd love to learn more.

.changeset/dull-masks-occur.md
tklever: Updated to suggestion.

plugins/catalog-react/src/routes.ts
freben: You know what, thanks for pushing on that because this led me down a deep dark rabbit hole of the js language specification :) And I was wrong! Indeed they've pinned down that those arguments are evaluated at call time, _every time_, including complex defaults. I was so hurt by python subtly stabbing my back with the way that its defaults are evaluated at _definition time_ leading to subtle bugs unless you're careful. And I transferred that onto js. But for all its flaws, that was not one of them.

plugins/catalog-react/src/components/EntityRefLink/EntityRefLink.tsx
```diff
@@ -88,33 +83,7 @@ function useEntityRoute(
...
-
-  kind = kind.toLocaleLowerCase('en-US');
-  namespace = namespace?.toLocaleLowerCase('en-US') ?? DEFAULT_NAMESPACE;
-
-  const routeParams = {
-    kind: encodeURIComponent(kind),
-    namespace: encodeURIComponent(namespace),
-    name: encodeURIComponent(name),
-  };
+  const routeParams = entityRouteParams(entityRef, { encodeParams: true });
```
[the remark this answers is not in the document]
tklever: > Doesn't entityRoute below deal with encoding?

Sort of. If my understanding of `entityRoute` is correct, the encoding happens in the `RouteResolver`. The encoding there is selective.

https://github.com/backstage/backstage/blob/335a3f1664493c12606ef3bbd1ab93c117300f12/packages/core-app-api/src/routing/RouteResolver.ts#L229-L244

Previously, this function did its own encoding with `encodeURIComponent` on lines 114 - 116. For complete backwards compatibility, I added the `encodeParams` functionality and set it to true here.

freben (approved):
Alright, I think it retains what was there before then

PR #29450, opened 2025-04-02
[Catalog] Use a different id for the legacy catalog client ref

camilaibs:

Use a different ID for the deprecated alpha version of the catalog service, as it has a different type definition and cannot be used interchangeably with the non-alpha version.

### DIFF
--- a/packages/cli/templates/backend-plugin/src/plugin.ts.hbs
+++ b/packages/cli/templates/backend-plugin/src/plugin.ts.hbs
@@ -3,7 +3,7 @@ import {
   createBackendPlugin,
 } from '@backstage/backend-plugin-api';
 import { createRouter } from './router';
-import { catalogServiceRef } from '@backstage/plugin-catalog-node/alpha';
+import { catalogServiceRef } from '@backstage/plugin-catalog-node';
 import { createTodoListService } from './services/TodoListService';
 
 /**
@@ -17,15 +17,13 @@ export const {{pluginVar}} = createBackendPlugin({
     env.registerInit({
       deps: {
         logger: coreServices.logger,
-        auth: coreServices.auth,
         httpAuth: coreServices.httpAuth,
         httpRouter: coreServices.httpRouter,
         catalog: catalogServiceRef,
       },
-      async init({ logger, auth, httpAuth, httpRouter, catalog }) {
+      async init({ logger, httpAuth, httpRouter, catalog }) {
         const todoListService = await createTodoListService({
           logger,
-          auth,
           catalog,
         });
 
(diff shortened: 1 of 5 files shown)


freben (approved):

Rugvip (approved):
👍, let's ship this as a patch as well 🙏

packages/cli/templates/backend-plugin/src/plugin.ts.hbs
```diff
@@ -3,7 +3,7 @@ import {
   createBackendPlugin,
 } from '@backstage/backend-plugin-api';
 import { createRouter } from './router';
-import { catalogServiceRef } from '@backstage/plugin-catalog-node/alpha';
+import { catalogServiceRef } from '@backstage/plugin-catalog-node';
```
camilaibs, line 6: To fix this test error:
image

camilaibs: People using the previous version of this plugin will either have to fix the test to mock the legacy catalog service or update the plugin backend to use the new service ref, right? Should we add a note about this in the changeset file?

freben: Yeah maybe some form of note could be helpful.

Not sure how to best describe it in words! It's a bit complex. But mentioning that they may want to double check that they are mocking the right thing, if they start seeing new errors - yeah

freben: Overall it's a bit unhelpful toward people to have these two nearly identical refs. How can they know beforehand which one to mock ion a given situation? But some day we'll probably remove the alpha one and the problem goes away.

camilaibs: PTAL: https://github.com/backstage/backstage/pull/29450/files#diff-5e31e46b45caa7f38b4d318fb84bd8e9a78dc14665680d1c908a37f855f82c69

mejiatadeo (commented):
cac1c170bb9f7285577bc01b23b9b6c92bf8762e

mejiatadeo (commented):
cac1c170bb9f7285577bc01b23b9b6c92bf8762e

packages/cli/templates/backend-plugin/src/plugin.ts.hbs
mejiatadeo: cac1c170bb9f7285577bc01b23b9b6c92bf8762e

mejiatadeo (commented):
cac1c170bb9f7285577bc01b23b9b6c92bf8762e

PR #29472, opened 2025-04-03
make certificate strings optional for Dev environments

### LINKED ISSUE #28011: 🐛 Bug Report: Error: Missing required config value at 'app.https.certificate.cert' in 'app-config.yaml'
### 📜 Description

I have a proxy (load balancer) sitting in front of the app component which does TLS termination.

so `app.baseUrl` should start with `https://` despite it doing http server. I thought that if `app.https.certificate` is not set, then it automatically assumes its http. However instead its just throwing the fact that its missing this cert.

### 👍 Expected behavior

if `app.https.certificate` is not set, don't expect it, just because `baseUrl` says its https.

`baseUrl` dictates whether `app.https.certificate` should be set or not.

### 👟 Reproduction steps

set up demo but with `app.baseUrl` as a https link.

_No response_

### 🖥️ Your Environment

OS:   Linux 6.12.1 - linux/x64
node: v20.16.0
yarn: 1.22.22
cli:  0.29.3-next.1 (installed)
backstage:  1.34.0-next.1

None

jpbottinelli:

fixes #28011

### DIFF
--- a/.changeset/tangy-dryers-lick.md
+++ b/.changeset/tangy-dryers-lick.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+make certificate strings optional for Dev environments
--- a/packages/cli/src/modules/build/lib/bundler/server.ts
+++ b/packages/cli/src/modules/build/lib/bundler/server.ts
@@ -195,8 +195,10 @@ DEPRECATION WARNING: React Router Beta is deprecated and support for it will be 
           ? {
               type: 'https',
               options: {
-                cert: fullConfig.getString('app.https.certificate.cert'),
-                key: fullConfig.getString('app.https.certificate.key'),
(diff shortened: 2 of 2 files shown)


freben:
I see that the issue mentions a codespaces use case. But even if that's what we are targeting here, what's the effect of not supplying these parameters? Does the dev server just skip cert validation then, or what?

jpbottinelli:
Thanks @freben. I understand why you're asking. I was at the contribfest with @vinzscam and we looked at this in detail. I wasn't sure if we should do this if it affected production, but as it's only for dev envs, we didn't see a problem with making them optional. 
From the original issue, looks like they're running https locally, but without certificate in order to test other stuff.
I tested it locally and was able to reproduce it before doing the fix. Do you see any issues with this?

freben (approved):
Nope! Then I'm all happy. Thanks!

PR #29489, opened 2025-04-06
cli: add new 'repo start' command

Rugvip:

Figured this might be worth adding, it replaces scripts like `yarn dev` and `yarn start-backend` with a single one with a bit more stuff built-in. By default it will start `packages/app` and `packages/backend` concurrently, or just one of them if only one exists. It falls back to other `frontend` and `backend` packages, and if none are available it instead expects single `frontend-plugin` and `backend-plugin` packages. That last one is to polish the experience in community plugin workspaces.

If the target packages are plugins, they'll be started as plugins would, i.e. `dev/index`. This closes the gap on there not being a good way to run plugins in isolation across both frontend and backend. There's also a `--plugin <pluginId>` option to make it simpler, although right now all involved plugin packages need to have `backstage.pluginId` set for it to work.

Another improvement over the current `yarn dev` pattern is that we can add a few options that are forwarded to the individual packages. For example `--inspect` and `--link` are forwarded to appropriate packages, which isn't possible with `yarn dev` and forced you to run the packages individually.

Couple of examples of things you can now do in a standard @backstage/create-app:

```shell
yarn start # run packages/app and packages/backend
yarn start --inspect-brk # same as above, but backend will break on start
yarn start app # run only packages/app
yarn start app backend # also run packages/app and packages/backend
yarn start --plugin catalog # run dev entry points in local catalog plugin frontend and backend, if they exist
yarn start --config app-config.other.yaml # run app and backend, overriding config
yarn start --link ../backstage # run app and backend, linking both to a separate workspace
yarn start plugins/catalog # run only catalog plugin, i.e. select package by path rather than name
yarn start ./plugins/catalog # explicitly run by path instead of name by prefixing with ./
```

Some other things I considered but have not implemented, lmk if you think it makes sense to add any of these:

```shell
yarn start --app # run only app package, whether it's packages/app or a different frontend package
yarn start --backend # same as above, but backend
```


Parsifal-M:
Heya! 👋 

Maybe an odd question but we have something like:

```json
    "debug": "LOG_LEVEL=debug yarn dev",
```

Just thinking out loud, Is it worth adding something like:

```text
yarn start --debug
```

not sure if this would be useful 💭

Rugvip:
@Parsifal-M hmm, I might be inclined to `--log-level debug` or such instead, just to keep it a bit more open. `--debug` vs `--inspect` might also be confusing, but could be alright too. Feeling it's not an obvious enough addition at this point

packages/cli/src/modules/start/commands/package/start/startBackend.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  await waitForExit();
+}
+
+export async function startBackendPlugin(options: StartBackendOptions) {
+  const hasDevIndexEntry = await fs.pathExists(
+    resolvePath(options.targetDir ?? paths.targetDir, 'dev/index.ts'),
+  );
+  if (!hasDevIndexEntry) {
+    console.warn(
```
freben, line 49: is it really ok to just bail with a "happy" error code here?

packages/cli/src/modules/start/commands/repo/start.ts
```diff
@@ -0,0 +1,216 @@
...
+    `Starting ${packageOptions
+      .map(({ pkg }) => pkg.packageJson.name)
+      .join(', ')}`,
+  );
+
+  // Each of these block until interrupted by user
+  await Promise.all(packageOptions.map(entry => startPackage(entry.options)));
+}
+
+async function findTargetPackages(namesOrPaths: string[], pluginIds: string[]) {
```
freben, line 64: tests might be nice

packages/cli/src/modules/start/commands/repo/start.ts
```diff
@@ -0,0 +1,216 @@
...
+          type: 'string',
+          multiple: true,
+        },
+        require: {
+          type: 'string',
+        },
+      },
+    });
+    const parsedRequire =
+      typeof parsedOpts.require === 'string' ? parsedOpts.require : undefined;
```
freben, line 195: what about supporting multiple require?

freben (approved):
There's a lack of tests overall though.

packages/cli/src/modules/start/commands/package/start/startBackend.ts
Rugvip: this is existing code, did have the same thought though, thinking it's fine as is for now

packages/cli/src/modules/start/commands/repo/start.ts
Rugvip: it's not supported by package start atm

Rugvip: leaving this for a followup to add support for it both here and in the existing `package start` command

PR #29502, opened 2025-04-07
add template extensions page

mbenson:

### DIFF
--- a/plugins/scaffolder-react/report-alpha.api.md
+++ b/plugins/scaffolder-react/report-alpha.api.md
@@ -317,6 +317,7 @@ export type ScaffolderPageContextMenuProps = {
   onActionsClicked?: () => void;
   onTasksClicked?: () => void;
   onCreateClicked?: () => void;
+  onTemplatingExtensionsClicked?: () => void;
 };
 
 // @alpha (undocumented)
@@ -339,6 +340,7 @@ export const scaffolderReactTranslationRef: TranslationRef<
     readonly 'scaffolderPageContextMenu.editorLabel': 'Manage Templates';
     readonly 'scaffolderPageContextMenu.actionsLabel': 'Installed Actions';
     readonly 'scaffolderPageContextMenu.tasksLabel': 'Task List';
+    readonly 'scaffolderPageContextMenu.templatingExtensionsLabel': 'Templating Extensions';
     readonly 'stepper.backButtonText': 'Back';
     readonly 'stepper.createButtonText': 'Create';
     readonly 'stepper.reviewButtonText': 'Review';
--- a/plugins/scaffolder/report-alpha.api.md
+++ b/plugins/scaffolder/report-alpha.api.md
@@ -41,6 +41,7 @@ const _default: FrontendPlugin<
     actions: SubRouteRef<undefined>;
     listTasks: SubRouteRef<undefined>;
     edit: SubRouteRef<undefined>;
+    templatingExtensions: SubRouteRef<undefined>;
   },
   {
     registerComponent: ExternalRouteRef<undefined>;
@@ -367,6 +368,26 @@ export const scaffolderTranslationRef: TranslationRef<
     readonly 'renderSchema.tableCell.type': 'Type';
     readonly 'renderSchema.tableCell.title': 'Title';
     readonly 'renderSchema.tableCell.description': 'Description';
+    readonly 'templatingExtensions.content.values.title': 'Values';
+    readonly 'templatingExtensions.content.values.notAvailable': 'There are no global template values defined.';
+    readonly 'templatingExtensions.content.emptyState.title': 'No information to display';
+    readonly 'templatingExtensions.content.emptyState.description': 'There are no templating extensions available or there was an issue communicating with the backend.';
+    readonly 'templatingExtensions.content.filters.title': 'Filters';
+    readonly 'templatingExtensions.content.filters.schema.input': 'Input';
+    readonly 'templatingExtensions.content.filters.schema.output': 'Output';
+    readonly 'templatingExtensions.content.filters.schema.arguments': 'Arguments';
+    readonly 'templatingExtensions.content.filters.examples': 'Examples';
+    readonly 'templatingExtensions.content.filters.notAvailable': 'There are no template filters defined.';
+    readonly 'templatingExtensions.content.filters.metadataAbsent': 'Filter metadata unavailable';
+    readonly 'templatingExtensions.content.searchFieldPlaceholder': 'Search for an extension';
+    readonly 'templatingExtensions.content.functions.title': 'Functions';
+    readonly 'templatingExtensions.content.functions.schema.output': 'Output';
+    readonly 'templatingExtensions.content.functions.schema.arguments': 'Arguments';
+    readonly 'templatingExtensions.content.functions.examples': 'Examples';
+    readonly 'templatingExtensions.content.functions.notAvailable': 'There are no global template functions defined.';
+    readonly 'templatingExtensions.title': 'Templating Extensions';
+    readonly 'templatingExtensions.subtitle': 'This is the collection of available templating extensions';
+    readonly 'templatingExtensions.pageTitle': 'Templating Extensions';
     readonly 'templateTypePicker.title': 'Categories';
     readonly 'templateIntroPage.title': 'Manage Templates';
     readonly 'templateIntroPage.subtitle': 'Edit, preview, and try out templates, forms, and custom fields';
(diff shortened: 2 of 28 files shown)


plugins/scaffolder/src/components/TemplateExtensionsPage/TemplateExtensionsPage.tsx
```diff
@@ -0,0 +1,321 @@
...
+  const classes = useStyles();
+  const { t } = useTranslationRef(scaffolderTranslationRef);
+
+  const { loading, value, error } = useAsync(async () => {
+    if (api.listTemplateExtensions) {
+      return api.listTemplateExtensions();
+    }
+    // eslint-disable-next-line no-console
+    console.warn(
+      'listTemplateExtensions is not implemented in the scaffolderApi; please make sure to implement this method.',
```
camilaibs, line 139: Just to double check, the method exists but is not implemented in the default Scaffolder API implementation, right? I would recommend adding this note to the changeset file as well 🙂

plugins/scaffolder/src/plugin.tsx
```diff
@@ -126,6 +127,7 @@ export const scaffolderPlugin = createPlugin({
     editor: editorRouteRef,
     customFields: customFieldsRouteRef,
     templateForm: templateFormRouteRef,
+    templateExtensions: templateExtensionsRouteRef,
```
camilaibs, line 130: Should we add the route ref here as well?
https://github.com/backstage/backstage/blob/30471f7661cbbc8fca1d195f4b21c000ef4e45c7/plugins/scaffolder/src/alpha/plugin.tsx#L41

plugins/scaffolder/src/components/TemplateExtensionsPage/TemplateExtensionsPage.tsx
```diff
@@ -0,0 +1,321 @@
...
+    <Page themeId="home">
+      <Header
+        pageTitleOverride={t('templateExtensions.pageTitle')}
+        title={t('templateExtensions.title')}
+        subtitle={t('templateExtensions.subtitle')}
+      >
+        <ScaffolderPageContextMenu {...scaffolderPageContextMenuProps} />
+      </Header>
+      <Content>
+        <TemplateExtensionsPageContent linkPage="" />
```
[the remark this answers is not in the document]
mbenson: This is making it possible to render links properly such that if the template extensions content is rendered as part of the template editor, links will return to the actual template extensions page, so the link page is overridden to `''` on the actual page causing _its_ links to be rendered pointing back to the current page. I'm quite new to React still so I'm entirely receptive to any advice you have on a cleaner-looking way to achieve this. 😄

plugins/scaffolder/src/plugin.tsx
mbenson: If `alpha` means here what it normally does, I guess so 😉 . Will do.

plugins/scaffolder/src/components/TemplateExtensionsPage/TemplateExtensionsPage.tsx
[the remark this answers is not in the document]
mbenson: I have refactored this to trigger the behavior using a `linkLocal: boolean` property, and handle all the consequential `Link` customization in the `TemplateExtensionsPageContent` component, making the subordinate components less fiddly IMO. See how this hits you, if you would, @camilaibs 🙏

plugins/scaffolder/src/components/TemplateExtensionsPage/TemplateExtensionsPage.tsx
camilaibs: > Should we include a note to the effect that custom Scaffolder API impls will need to implement this in order to derive usefulness from this page?

I would say yes, the text in the changeset is the same one people will see in the changelog file. I think it would be frustrating to add the page and have it not work as expected without knowing that this is necessary. But it seems that the default Scaffolder API client already implements this method by default, so maybe the warning is no longer necessary?

camilaibs (approved):
Thanks for the contribution. Overall, it looks great. I'll hand over the final components review to @backstage/scaffolder-maintainers 🙂

plugins/scaffolder/src/components/TemplateExtensionsPage/TemplateExtensionsPage.tsx
benjdlambert: The reason that things are marked as optional is if adopters are providing their own instance of the `scaffolderApiRef` and ultimately `ScaffolderClient`. 

Marking members and functions as required on the implementation is a breaking change unfortunately, so that's why we mark as optional.

I think it's fine to leave as such and we'll bring everything to required in the next breaking change to the scaffolder.

plugins/scaffolder-react/report-alpha.api.md
```diff
@@ -317,6 +317,7 @@ export type ScaffolderPageContextMenuProps = {
   onActionsClicked?: () => void;
   onTasksClicked?: () => void;
   onCreateClicked?: () => void;
+  onTemplateExtensionsClicked?: () => void;
```
benjdlambert, line 320: I'm thinking that we call this `templatingExtensions` instead.

`Template Extensions` is a bit ambiguous in naming I'm feeling.

plugins/scaffolder/report-alpha.api.md
```diff
@@ -367,6 +368,26 @@ export const scaffolderTranslationRef: TranslationRef<
     readonly 'renderSchema.tableCell.type': 'Type';
     readonly 'renderSchema.tableCell.title': 'Title';
     readonly 'renderSchema.tableCell.description': 'Description';
+    readonly 'templateExtensions.content.values.title': 'Values';
```
benjdlambert, line 371:
```suggestion
    readonly 'templatingExtensions.content.values.title': 'Values';
```

plugins/scaffolder/report.api.md
```diff
@@ -618,6 +618,7 @@ export const scaffolderPlugin: BackstagePlugin<
     editor: SubRouteRef<undefined>;
     customFields: SubRouteRef<undefined>;
     templateForm: SubRouteRef<undefined>;
+    templateExtensions: SubRouteRef<undefined>;
```
benjdlambert, line 621:
```suggestion
    templatingExtensions: SubRouteRef<undefined>;
```

benjdlambert (changes requested):
Hey @mbenson :wave:

Sorry for the delay in getting round to review this, been a busy few weeks.

So I'm feeling that we should move to `templatingExtensions` everywhere instead of `templateExtensions` as it's a little cleaner and clearer that it's customizing the templating logic rather than the `kind: Template` itself.

Does this make sense to you too?

plugins/scaffolder/report.api.md
mbenson: Sure, I'm not picky about that. I can't remember offhand but I'm guessing we may already have used that terminology on the backend. If so do we want to go ahead and change those as well since they should have been alpha?

benjdlambert: Yeah if there's anything that stands out as sounds better with `templating` instead then lets go ahead and update that. `minor` is fine for making breaking changes to `/alpha` export also.

i didn't see so much earlier when I searched so maybe there's not too much.

mbenson:
@benjdlambert it's close. I knew the 1.38.0 release was supposed to be today, but it hadn't happened yet so I didn't know if we'd be able to get in under the wire the scaffolder-backend API rename from `template-extensions` to `templating-extensions` but it looks like you've been doing some release tasks while I've been reworking this, so IDK whether we want to do anything special since that endpoint will have been released and doesn't seem to have any kind of alpha qualifier around it.

Other than this, do I need to move the `TemplatingExtensionsPage` from `src/components` to `src/alpha/components`?

benjdlambert:
@mbenson soery for the late reply here, but I managed to ship a patch release today with the change to `templating-extensions` instead.

Had a chat amongst other maintainers and came to the conclusion that if we do it fast enough we can probably get away with it being a bug fix as the usage should be pretty limited.

Hope that's OK that I cherry picked the bits out of here that we would need to do ASAP.

Might need a rebase here but I'll try and get a review through of the remaining stuff tomorrow - holidays permitting.

benjdlambert (approved):
Nice! Let's get this in :tada: thanks for all the effort :Tada;

mbenson:
Thanks @benjdlambert ... I do wonder, though, whether we want the page-neutral `RenderSchema` to rely on a type defined with `TemplatingExtensionsPage`.

benjdlambert:
@mbenson yeah can probably fix that up in a followup, it's all internal types at least, but yeah probably poorly placed.

mbenson:
I guess we've got a bit of time before it releases, maybe discuss on 5/2?

PR #29514, opened 2025-04-08
docs/frontend-system: polish plugin migration docs

Rugvip:
I sat down with @secustor at KubeCon last week and watched him migrate a plugin to the new frontend system (thanks! 🙏)

This is a collection of simpler fixes, while there's still some separate stuff that needs more long-term work. Main pieces in that category are improved API reference (extension blueprint) docs and type inference or other solution to avoid the confusion around still needing to use `createApiFactory` with `ApiBlueprint`.

### DIFF
--- a/docs/frontend-system/architecture/60-migrations.md
+++ b/docs/frontend-system/architecture/60-migrations.md
@@ -1,9 +1,9 @@
 ---
 id: migrations
-title: Frontend System Migrations
-sidebar_label: Migrations
+title: Frontend System Changelog
+sidebar_label: Changelog
 # prettier-ignore
-description: Migration documentation for different versions of the frontend system core APIs.
+description: Changelog documentation for different versions of the frontend system core APIs.
 ---
 
 > **NOTE: The new frontend system is in alpha and is only supported by a small number of plugins.**
--- a/docs/frontend-system/building-plugins/03-common-extension-blueprints.md
+++ b/docs/frontend-system/building-plugins/03-common-extension-blueprints.md
@@ -10,7 +10,7 @@ description: Extension blueprints provided by the frontend system and core featu
 
 This section covers many of the [extension blueprints](../architecture/23-extension-blueprints.md) available at your disposal when building Backstage frontend plugins.
 
-## Built-in extension blueprints
+## Extension blueprints in `@backstage/frontend-plugin-api`
 
 These are the [extension blueprints](../architecture/23-extension-blueprints.md) provided by the Backstage frontend framework itself.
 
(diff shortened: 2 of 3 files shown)


docs/frontend-system/architecture/60-migrations.md
```diff
@@ -1,9 +1,9 @@
 ---
 id: migrations
-title: Frontend System Migrations
-sidebar_label: Migrations
+title: Frontend System Changelog
```
Rugvip, line 3: This is where you land if you search for frontend migration, which is not what we want

docs/frontend-system/building-plugins/03-common-extension-blueprints.md
```diff
@@ -46,18 +46,26 @@ Icon bundle extensions provide the ability to replace or provide new icons to th
...
 Translation extension provide custom translation messages for the app. They can be used both to override the default english messages to custom ones, as well as provide translations for additional languages.
 
-## Core feature extension blueprints
+## Extension blueprints in `@backstage/plugin-catalog-react/alpha`
 
-These are the [extension blueprints](../architecture/23-extension-blueprints.md) provided by the Backstage core feature plugins.
+These are the [extension blueprints](../architecture/23-extension-blueprints.md) provided by the Catalog plugin.
 
-### EntityCard - [Reference](https://github.com/backstage/backstage/blob/master/plugins/catalog-react/report-alpha.api.md)
+### EntityCard - [Example](https://github.com/backstage/backstage/blob/75e79518eafc6e6eb55585f166667418419662de/plugins/org/src/alpha.tsx#L27-L36)
```
Rugvip, line 53: Best I could find so far, we need better references here. Ideally we could completely skip this section and rely on some other auto generated documentation though.

docs/frontend-system/building-plugins/05-migrating.md
```diff
@@ -102,15 +117,16 @@ import {
 
 const fooPage = PageBlueprint.make({
   params: {
+    // This is the path that was previously defined in the app code.
+    // It's labelled as the default one because it can be changed via configuration.
     defaultPath: '/foo',
-    // you can reuse the existing routeRef
-    // by wrapping into the convertLegacyRouteRef.
+    // You can reuse the existing routeRef by wrapping into the convertLegacyRouteRef.
```
freben, line 123:
```suggestion
    // You can reuse the existing routeRef by wrapping it with convertLegacyRouteRef.
```

docs/frontend-system/building-plugins/05-migrating.md
```diff
@@ -79,6 +85,8 @@ The code above binds all the extensions to the plugin. _Important_: Make sure to
 
 Pages that were previously created using the `createRoutableExtension` extension function can be migrated to the new Frontend System using the `PageBlueprint` [extension blueprint](../architecture/23-extension-blueprints.md), exported by `@backstage/frontend-plugin-api`.
 
+In the new system plugins are responsible more information than they used to. For example, the plugin is now responsible for providing the path for the page, rather than it being part of the app code.
```
freben, line 88:
```suggestion
In the new system plugins provide more information than they used to. For example, the plugin is now responsible for providing the path for the page, rather than it being part of the app code.
```

docs/frontend-system/architecture/60-migrations.md
freben: ...but the id is still `migrations`? Alright

freben (approved):

PR #29526, opened 2025-04-09
Jira home

anuragpachauri:
### jira home Plugin
---
## Features
- View **jira home** associated with a specific component in Backstage.
---

### DIFF
--- a/microsite/data/plugins/azure-resource.yaml
+++ b/microsite/data/plugins/azure-resource.yaml
@@ -5,5 +5,5 @@ authorUrl: https://github.com/anuragpachauri
 category: Infrastructure
 description: A plugin showing Azure resource based on tag in relation to an entity in the catalog.
 documentation: https://github.com/anuragpachauri/backstageMain/tree/main/plugins/azure-resource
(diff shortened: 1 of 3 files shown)


Parsifal-M:
Hello 👋 

I have not gone over **_all_** the code in those plugins but there are a few things that jumped out to me as concerns from a quick look like:

- This is probably an oversight but I don't think you want to be making this fetch [here](https://github.com/anuragpachauri/backstageMain/blob/main/plugins/jira-home/src/components/ExampleComponent/ExampleComponent.tsx#L62)?
- You should try and use the [Fetch Api](https://backstage.io/docs/reference/core-plugin-api.fetchapi) in these plugins as its considered best practise.
- [This](https://github.com/anuragpachauri/backstageMain/blob/main/plugins/linode-resource/src/components/ExampleFetchComponent/ExampleFetchComponent.tsx) is a bit strange? I don't think this is needed either, and again should be using the [Fetch Api](https://backstage.io/docs/reference/core-plugin-api.fetchapi)

My advice would be to take a look at how some plugins are built in the [Community Plugins](https://github.com/backstage/community-plugins) repository and use them as a "scaffold"

PR #29527, opened 2025-04-09
feat migrate some code to detectPackageManager 

JPeer264:

First off all, this wouldn't have been possible without the initial work of @tylerd-canva from #27095 - so big shoutout 👏 

(ref: #20658)

Follow up to #27269. This PR is mainly focusing on moving towards using `detectPackageManager` - to keep the PR smaller (it is still huge IMO) I only touched what was really necessary (a followup PR will come that is updating the rest of the code, where `yarn` is used directly). 

Initially I wanted to only migrate `YarnLockfile`, as suggested in https://github.com/backstage/backstage/pull/27095#pullrequestreview-2379334029, which was a little impossible without either exporting the `YarnLockfile` in `@backstage/cli` or duplicating the logic within `@backstage/plugin-devtools-backend`. Since neither seemed to be a great practice I decided to keep it in. If you think exporting it and marking it as `@internal` to keep the PR smaller let me know (I am in this codebase since yesterday, so I don't know all the good practices from this repo just yet).

The changes can be grouped into:
1. making `detectPackageManager` public available via `@backstage/cli`, and using it where ever `Lockfile` was used
2. moving code from `packages/cli` to `packages/cli-node` to keep everything closer together (e.g. `getHasYarnPlugin` or `fetchPackageInfo`)

This PR **does not** include pnpm support, just starting to migrate slowly to make it easier to add (goal of mine is ofc to add this too)


JPeer264:
Regarding the failing Windows E2E tests: Could it be that these are flaky? It seems that it always hangs at the "login" page when I run them directly via `yarn test:e2e`. However, when I am running each test individually it always works. I tried it on following systems:

- MacOS (tests are always passing, even when they run in parallel)
- Windows 11 via Parallels
- Windows 11 native
- Ubuntu 22 via Windows 11 WSL2

plugins/devtools-backend/src/api/DevToolsBackendApi.ts
```diff
@@ -223,8 +223,8 @@ export class DevToolsBackendApi {
       backstageJson = JSON.parse(buffer.toString());
     }
 
-    const lockfilePath = paths.resolveTargetRoot('yarn.lock');
-    const lockfile = await Lockfile.load(lockfilePath);
+    const pacman = await detectPackageManager();
```
awanlin, line 226: With Yarn it was always safe to assume that the lock file would be available for this feature. With these changes I assume that would still be true, but can you confirm? Also, would that be true for other package managers, is it safe to assume their lock files will be included with Backstage when deployed?

awanlin (commented):
Just gave this a look in regards to the DevTools plugin changes and left one comment. I haven't look at the other changes, not really an area I have expertise in.

plugins/devtools-backend/src/api/DevToolsBackendApi.ts
JPeer264: The logic didn't change, so yes, your assumption is still true. To confirm, when using another package manager then the `yarn.lock` won't be present., so with pnpm there would be a `pnpm-lock.yaml` instead.

JPeer264:
For me this PR is not stale. If there is anything I can do, please let me know

rushrs:
is this still being worked on, I can pick this up where you left off @JPeer264

sebdanielsson:
@cdedreuille can someone from the team take a look at this?

sebdanielsson:
Nope

sebdanielsson:
It's not tho.

packages/cli-node/src/pacman/yarn/YarnLockfile.ts
```diff
@@ -0,0 +1,256 @@
...
+];
+
+export class YarnLockfile implements Lockfile {
+  private constructor(
+    private packages: Map<string, LockfileQueryEntry[]>,
+    private data: LockfileData,
+    private readonly legacy: boolean,
+  ) {}
+
+  static parse(content: string) {
```
aramissennyeydd, line 74: I think it's worth keeping the comment here since it won't come from the interface

packages/cli-node/src/pacman/yarn/YarnLockfile.ts
```diff
@@ -0,0 +1,256 @@
...
+  `version`,
+  `resolution`,
+  `dependencies`,
+  `peerDependencies`,
+  `dependenciesMeta`,
+  `peerDependenciesMeta`,
+  `binaries`,
+];
+
+export class YarnLockfile implements Lockfile {
```
aramissennyeydd, line 67: Can we add a comment and visibility here?

aramissennyeydd (commented):
👋 This is a chunky one for sure - what do we think of splitting this into (at least) 2? I know you mention review speed as a reason for keeping these together, but it's going to be really difficult to get this in as it is. It touches a _lot_ of code paths and is fairly complex.

I'm thinking this could be 
1. Rename Lockfile to YarnLockfile and update internals to use the detectPackageManager utility.
2. Update usages of Lockfile.load to the new detectPackageManager utility.
3. Update `yarn` hard coded strings to use a generic value.

JPeer264:
@rushrs @sebdanielsson does anyone have some capacity atm and can help out on this? You should have access to my repo to make changes to my branch if you want.

PR #29553, opened 2025-04-11
add a scheduler api

freben:
This amends the `DefaultSchedulerService` with a REST API rooted at `/api/<plugin>/.backstage/scheduler/v1`. It essentially establishes a more full version of https://github.com/backstage/backstage/pull/28726.

## `GET /.backstage/scheduler/v1/tasks`

Returns all registered tasks and their current statuses. Example:

```json
{
  "tasks": [
    {
      "taskId": "InternalOpenApiDocumentationProvider:refresh",
      "pluginId": "catalog",
      "scope": "global",
      "settings": {
        "version": 2,
        "cadence": "PT10S",
        "initialDelayDuration": "PT10S",
        "timeoutAfterDuration": "PT1M"
      },
      "taskState": {
        "status": "idle",
        "startsAt": "2025-04-11T20:35:13.418+02:00",
        "lastRunEndedAt": "2025-04-11T20:35:03.453+02:00"
      },
      "workerState": {
        "status": "initial-wait"
      }
    }
  ]
}
```

## `POST /.backstage/scheduler/v1/tasks/<taskId>/trigger`

Triggers the given task. Returns 200 if successful, 404 if no such task, or 409 if it's already running.

### DIFF
--- a/.changeset/dirty-grapes-vanish.md
+++ b/.changeset/dirty-grapes-vanish.md
@@ -1,0 +1,5 @@
+---
+'@backstage/backend-defaults': patch
+---
+
+The `DefaultSchedulerService` now accepts `HttpRouterService` and `PluginMetadataService` arguments. If you supply a router, the scheduler will register a REST API for listing and triggering tasks.
--- a/.changeset/dirty-grapes-vanish.md
+++ b/.changeset/dirty-grapes-vanish.md
@@ -1,5 +1,7 @@
 ---
-'@backstage/backend-defaults': patch
+'@backstage/backend-defaults': minor
 ---
 
-The `DefaultSchedulerService` now accepts `HttpRouterService` and `PluginMetadataService` arguments. If you supply a router, the scheduler will register a REST API for listing and triggering tasks. Please see [the scheduler documentation](https://backstage.io/docs/backend-system/core-services/scheduler) for more details about this API.
+**BREAKING**: The `DefaultSchedulerService` constructor options now requires `RootLifecycleService`, `HttpRouterService`, and `PluginMetadataService` fields.
+
+The scheduler will register a REST API for listing and triggering tasks. Please see [the scheduler documentation](https://backstage.io/docs/backend-system/core-services/scheduler) for more details about this API.
--- a/.changeset/dirty-grapes-vanish.md
+++ b/.changeset/dirty-grapes-vanish.md
@@ -2,4 +2,4 @@
 '@backstage/backend-defaults': patch
 ---
 
-The `DefaultSchedulerService` now accepts `HttpRouterService` and `PluginMetadataService` arguments. If you supply a router, the scheduler will register a REST API for listing and triggering tasks.
+The `DefaultSchedulerService` now accepts `HttpRouterService` and `PluginMetadataService` arguments. If you supply a router, the scheduler will register a REST API for listing and triggering tasks. Please see [the scheduler documentation](https://backstage.io/docs/backend-system/core-services/scheduler) for more details about this API.
--- a/packages/backend-defaults/report-scheduler.api.md
+++ b/packages/backend-defaults/report-scheduler.api.md
@@ -4,7 +4,9 @@
 
 ```ts
 import { DatabaseService } from '@backstage/backend-plugin-api';
+import { HttpRouterService } from '@backstage/backend-plugin-api';
 import { LoggerService } from '@backstage/backend-plugin-api';
+import { PluginMetadataService } from '@backstage/backend-plugin-api';
 import { RootLifecycleService } from '@backstage/backend-plugin-api';
 import { SchedulerService } from '@backstage/backend-plugin-api';
 import { ServiceFactory } from '@backstage/backend-plugin-api';
@@ -16,6 +18,8 @@ export class DefaultSchedulerService {
     database: DatabaseService;
     logger: LoggerService;
     rootLifecycle?: RootLifecycleService;
+    httpRouter?: HttpRouterService;
+    pluginMetadata?: PluginMetadataService;
   }): SchedulerService;
 }
 
--- a/packages/backend-defaults/report-scheduler.api.md
+++ b/packages/backend-defaults/report-scheduler.api.md
@@ -17,9 +17,9 @@ export class DefaultSchedulerService {
   static create(options: {
     database: DatabaseService;
     logger: LoggerService;
-    rootLifecycle?: RootLifecycleService;
-    httpRouter?: HttpRouterService;
-    pluginMetadata?: PluginMetadataService;
+    rootLifecycle: RootLifecycleService;
+    httpRouter: HttpRouterService;
+    pluginMetadata: PluginMetadataService;
   }): SchedulerService;
 }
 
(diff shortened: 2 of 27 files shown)


aramissennyeydd (approved):
Nice!

packages/backend-defaults/src/entrypoints/scheduler/lib/types.ts
```diff
@@ -97,3 +98,39 @@ export const taskSettingsV2Schema = z.object({
...
+    | {
+        status: 'initial-wait';
+      }
+    | {
+        status: 'idle';
+      }
+    | {
+        status: 'running';
+      }
+    | null;
```
Rugvip, line 135: Hmm, should we even bother with this one? Feels a bit sketchy in a scaled deployment

packages/backend-defaults/src/entrypoints/scheduler/lib/PluginTaskSchedulerImpl.ts
```diff
@@ -147,6 +152,47 @@ export class PluginTaskSchedulerImpl implements SchedulerService {
...
+  getRouter(): express.Router {
+    const router = Router();
+
+    router.get('/.backstage/scheduler/v1/tasks', async (_, res) => {
+      const globalState = await TaskWorker.taskStates(
+        await this.databaseFactory(),
+      );
+
+      const tasks = new Array<TaskApiTasksResponse>();
+      for (const task of this.allScheduledTasks) {
```
Rugvip, line 164: Hmm, bit torn on showing local tasks in this API at all, might be better to just stick to the global ones?

packages/backend-defaults/src/entrypoints/scheduler/database/tables.ts
```diff
@@ -20,8 +20,10 @@ export const DB_TASKS_TABLE = 'backstage_backend_tasks__tasks';
 export type DbTasksRow = {
   id: string;
   settings_json: string;
-  next_run_start_at: Date;
+  next_run_start_at?: Date | string; // This can be null when in manual trigger mode
```
Rugvip, line 23: null as in `null`?

packages/backend-defaults/src/entrypoints/scheduler/lib/PluginTaskSchedulerImpl.ts
```diff
@@ -147,6 +152,47 @@ export class PluginTaskSchedulerImpl implements SchedulerService {
...
+            this.localWorkersById.get(task.id)?.workerState() ??
+            this.globalWorkersById.get(task.id)?.workerState() ??
+            null,
+        });
+      }
+
+      res.json({ tasks });
+    });
+
+    router.post(
```
Rugvip, line 184: Thoughts on auth for this one?

.changeset/dirty-grapes-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-defaults': patch
+---
+
+The `DefaultSchedulerService` now accepts `HttpRouterService` and `PluginMetadataService` arguments. If you supply a router, the scheduler will register a REST API for listing and triggering tasks.
```
Rugvip, line 5: I think this also deserves an explanation of the new feature 😅

Rugvip (commented):
Nice! 🎉 

Couple of bits

packages/backend-defaults/src/entrypoints/scheduler/lib/types.ts
freben: The idea is, that it would add value specifically in a scaled deployment. Especially when we have something like a `coreServices.systemMetadata` or so that can be used to discover all nodes.

If you imagine a status page, you would show what tasks exist and what their statuses are. But you would also sometimes want to see which single individual machine that is executing the task at any given moment; if you have heterogenous machines or one is misbehaving or has a full disk or so, you'd see that it's that one which is causing problems.

If we were to go all out on this feature, we'd actually persist historical state changes for every task so you see how it was run over time and by what workers and with what outcome, AND perhaps with what logs for each run etc :) Or at least there should be a correlation ID added to the logs for each run such that they can be gathered after the fact. More to come on that some other day!

packages/backend-defaults/src/entrypoints/scheduler/lib/PluginTaskSchedulerImpl.ts
freben: I don't think so - wouldn't that be artificially limiting? Again given a a `coreServices.systemMetadata` you'd see how each individual host is doing and whether it's been failing some important recurring task.

packages/backend-defaults/src/entrypoints/scheduler/database/tables.ts
freben: In the database row NULL yes.

This is an interesting question overall actually; should the TS types for these rows also always have `| undefined | null` at the end when they are nullable? Because you MAY pass in undefined or null depending on circumstances when you do writes, and in reads, I guess they should come out as nulls or do they ever get coerced to undefined? Not sure. But yeah there may be a risk of accidentally doing `if (row.column !== undefined)` instead of `if (row.column)` if you don't add `null` as an explicit type option here.

.changeset/dirty-grapes-vanish.md
freben: An expansion of the "REST API for listing and triggering tasks" to be more descriptive? Sure, I can add that

packages/backend-defaults/src/entrypoints/scheduler/lib/types.ts
freben: Another approach would be to add a separate `/.backstage/scheduler/v1/workers`.

.changeset/dirty-grapes-vanish.md
freben: Added complete docs to https://backstage.io/docs/backend-system/core-services/scheduler

packages/backend-defaults/report-scheduler.api.md
```diff
@@ -16,6 +18,8 @@ export class DefaultSchedulerService {
     database: DatabaseService;
     logger: LoggerService;
     rootLifecycle?: RootLifecycleService;
+    httpRouter?: HttpRouterService;
+    pluginMetadata?: PluginMetadataService;
```
Rugvip, line 22: eh, tbh I'm kinda thinking it's worth adding these as required dependencies straight away, unless you view this as a useful feature flag to disable the behavior by overriding the factory?

packages/backend-defaults/src/entrypoints/scheduler/lib/PluginTaskSchedulerImpl.ts
Rugvip: I think we can keep as is for now seeing as the ability to DoS is kinda limited due to retriggers not affecting current runs

.changeset/dirty-grapes-vanish.md
Rugvip: Noice 👌

Rugvip (approved):
Minor thing to consider, but let's :shipit:, great stuff! 👍

packages/backend-defaults/report-scheduler.api.md
freben: Yeah I agree. Made all arguments required while I was at it, and updated the changeset wording accordingly.

PR #29561, opened 2025-04-12
chore: add official buildkite plugin

mcncl:
## Add [official] Buildkite plugin

So, not sure what the process is here; do I create a new plugin called `buildkite_official.yaml` or PR to overwrite the existing one?

Buildkite officially has a plugin, so I went with the latter but comments welcome if that's not the way.

Source: https://github.com/buildkite/backstage-plugin
NPM package: https://www.npmjs.com/package/@buildkite/plugin-buildkite


microsite/data/plugins/buildkite.yaml
```diff
@@ -1,13 +1,10 @@
 ---
 title: Buildkite
-author: roadie.io
-authorUrl: https://roadie.io/?utm_source=backstage.io&utm_medium=marketplace&utm_campaign=buildkite
```
Rugvip, line 4: @dtuite what's your preference on how to handle this one? Do you want to hand over to the new official one or keep yours around with a name that makes it clear that it's the Roadie edition?

microsite/data/plugins/buildkite.yaml
```diff
@@ -1,13 +1,10 @@
...
-iconUrl: https://roadie.io/images/logos/buildkite.png
-npmPackageName: '@roadiehq/backstage-plugin-buildkite'
-tags:
-  - ci
-  - cd
-addedDate: '2021-04-20'
+description: View Buildkite CI builds from within Backstage
+documentation: https://github.com/buildkite/backstage-plugin
+iconUrl: /img/buildkite.svg
+npmPackageName: '@buildkite/plugin-buildkite'
```
Rugvip, line 9: For separate namespaces we'd generally recommend using `backstage-plugin-` as the plugin prefix, so `@buildkite/backstage-plugin-buildkite`

Rugvip (commented):
Thank you! 👍 

Will leave it to @dtuite whether we go with replace or keeping both. If we do end up keeping both I'd prefer renaming the old one to `buildkite_roadie.yaml` or similar

microsite/data/plugins/buildkite.yaml
mcncl: @Rugvip thanks for the feedback! Would you recommend we use `npm deprecate` and then publish a version with the preferred name?

mcncl: I've made those changes; added a deprecation to the incorrectly named and republished under the correct name. Docs updated too.

microsite/data/plugins/buildkite.yaml
dtuite: I'd imagine the official one will be better than our one so I'm fine with deprecating our plugin. What's the process for that? Do we just remove it from the marketplace or mark it as deprecated or what? If people are using it it might be weird if it just disappears completely.

mcncl:
Hey @freben 👋 Just wanted to follow this one up in the interest that the GH bots don't mark it as stale 🙂

microsite/data/plugins/buildkite.yaml
Rugvip: Marking it as deprecated on NPM would be the way to go. Afaik you can either do it via the CLI or publish a new version with `deprecated: true / <msg>` in `package.json`.

Rugvip (approved):
Thank you! 👍 

Given the desire to have this plugin take over as the official one we can move forward by shipping this

PR #29562, opened 2025-04-12
add `backend.trustProxy`

freben:
This is off the back of https://github.com/backstage/backstage/pull/28708 . I feel that, whether the rate limiting lands or not, this setting is important and universal enough that it may warrant being added separately either way.

It's a complex type, and I opted to just blankly make it use `getOptional` instead of writing a thorough validation for it, expecting Express to handle it instead.


camilaibs (approved):

Rugvip (approved):
:shipit:

PR #29564, opened 2025-04-12
Canon - Add new breakpoint helpers `up()`, `down()` and current breakpoint

cdedreuille:

To help users build responsive interfaces, we are now exporting our `useBreakpoint` hook that returns 3 values:
- `up` - A function that will return true if the window is up the selected breakpoint
- `down` - A function that will return true if the window is below the selected breakpoint
- `breakpoint` - The current breakpoint


packages/canon/src/hooks/useBreakpoint.ts
```diff
@@ -25,18 +25,38 @@ export const breakpoints: { name: string; id: Breakpoint; value: number }[] = [
...
+      break;
     }
   }
 
-  return breakpoints[0].id;
+  return {
+    breakpoint,
+    up: (key: Breakpoint): boolean => {
+      const value = getBreakpointValue(key);
+      return useMediaQuery(`(min-width: ${value}px)`);
```
Rugvip, line 55: Hmm, can't call hooks from here

Rugvip (commented):
Yep, but probably needs a bit of a refactor of the hook calls

packages/canon/src/hooks/useBreakpoint.ts
cdedreuille: You're right. I just updated the functions to move the hooks up.

Here's some insights from Cursor:

> The key improvements in this refactored version:
> - All useMediaQuery calls are made at the top level of the hook
> - The results are stored in Maps that can be efficiently queried
> - The up and down functions are now pure functions that just look up pre-calculated values
> - We maintain the same API, so consumers of the hook don't need to change their code
> 
> This approach might use more memory (storing all possible breakpoint values), but it's a necessary trade-off to follow React's hook rules correctly. It's also more performant because the media queries are only set up once per hook instance rather than being recreated every time up() or down() is called.
> Remember that hooks are meant to be called at the top level of your React components or custom hooks to ensure that React can properly maintain the state between renders. Moving the hook calls outside of the returned functions ensures that this rule is followed.

packages/canon/src/hooks/useBreakpoint.ts
```diff
@@ -25,18 +25,40 @@ export const breakpoints: { name: string; id: Breakpoint; value: number }[] = [
...
+/** @public */
+export const useBreakpoint = () => {
+  // Call all media queries at the top level
   const matches = breakpoints.map(breakpoint => {
-    const match = useMediaQuery(`(min-width: ${breakpoint.value}px)`);
-    return match;
+    return useMediaQuery(`(min-width: ${breakpoint.value}px)`);
   });
 
+  // Pre-calculate all the up/down values we need
```
Rugvip, line 35: oof, there are better ways to do this :sweat_smile:, throw in a TODO imo but happy to merge as is since the API won't change. Just that the implementation could be optimized.

Rugvip (approved):

Issue #13891, opened 2022-09-28
Permission authorization happening inside of the permission policy


Joonpark13:
## Expected Behavior

We actually could see a few alternatives on expected behavior - would love consultation of intent from plugin author

## Actual Behavior

If an adopter installs the playlist plugin, features within the playlist plugin do not respect the [(presumably) intended behavior](https://github.com/kuangp/backstage/blob/3b8eb1500723dfc56ea46660b4c7a11367825011/plugins/playlist-backend/src/permissions/DefaultPlaylistPermissionPolicy.ts#L77).

## Context

[Original PR](https://github.com/backstage/backstage/pull/12870#discussion_r961573814)

So sorry to mislead you in the comment in the original PR! Upon some further thought and discussion, we've discovered some issues with the pattern established by the playlist plugin inside the [permission policy here](https://github.com/backstage/backstage/blob/26c32ef614ec9733477dd15262b25028e06ee842/packages/backend/src/plugins/permission.ts#L41). We understand that the exposed `DefaultPlaylistPermissionPolicy` is only meant to be a default version of a possible policy that an adopter could implement. However, the structure of the current policy means that authorization is happening inside the policy instead of inside the plugin. This is an anti-pattern since it places the protection of possibly sensitive data under the control of the adopter's policy. If the adopter installs the plugin but does not configure the policy, for example, you may be exposing entities/access/features that you do not intend.

Of course, the nature of what is considered sensitive is highly dependent upon the context of the plugin, and we're not sure if the plugin author, in this case, is ok with this exposure, but regardless of intent we think it would be good to do the following:

- move the authorization out of the policy and inside the plugin (this should be fairly straightforward since the plugin is already integrated with the permission framework)
- rollback to the AllowAllPolicy in the example backend
- consider refactoring the permission framework integration, with more of a focus on exposing sensible permissions instead of exposing a default policy

kuangp:
hmm ...

The original intent here was to follow the same convention as the catalog plugin as the example. In the same way that the default out of the box experience (from my understanding) does not inherently enforce any default policy (eg. any user can read/delete any entity) and it's up to the adopter to author their own policy as part of the permissions backend configuration to place the proper access controls desired. 

[...]

PR #29578, opened 2025-04-14
BEP-0012: Default Permission Decisions

Rugvip:
BEP for adding default permission decisions to the permission system. This addresses #13891.

Summary:

The Backstage Permission System currently requires explicit policy handlers to be defined for all permissions. This BEP proposes adding support for default permission decisions that can be defined when creating a permission. These default decisions can be either definitive or conditional, leveraging existing evaluation mechanisms. This enhancement enables plugins to define built-in behavior through the permission system while maintaining the flexibility for adopters to override this behavior with custom policies.

### DIFF
--- a/beps/0012-default-permission-decisions/README.md
+++ b/beps/0012-default-permission-decisions/README.md
@@ -1,0 +1,190 @@
+---
+title: Default Permission Decisions
+status: implementable
+authors:
+  - '@Rugvip'
+owners:
+  - '@Rugvip'
+project-areas:
+  - permissions
+creation-date: 2025-04-10
+---
+
+# BEP: Default Permission Decisions
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Design Details](#design-details)
+  - [Default Decisions with Permissions Disabled](#default-decisions-with-permissions-disabled)
+  - [Default Decisions with Permissions Enabled](#default-decisions-with-permissions-enabled)
+  - [API Changes](#api-changes)
+- [Release Plan](#release-plan)
+- [Alternatives](#alternatives)
+  - [Return Default Decisions from Permission Backend](#return-default-decisions-from-the-permission-backend)
+  - [Communicate Default Decisions Over Separate Channel](#communicate-default-decisions-over-separate-channel)
+
+## Summary
+
+The Backstage Permission System currently requires explicit policy handlers to be defined for all permissions. This BEP proposes adding support for default permission decisions that can be defined when creating a permission. These default decisions can be either definitive or conditional, leveraging existing evaluation mechanisms. This enhancement enables plugins to define built-in behavior through the permission system while maintaining the flexibility for adopters to override this behavior with custom policies.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


beps/0012-default-permission-decisions/README.md
```diff
@@ -0,0 +1,190 @@
...
+- Allow permission policy authors to defer to the default decision
+
+### Non-Goals
+
+- Configuration of default decisions
+- Complex business logic in default decisions
+
+## Proposal
+
+A new "default" authorize result is added to the permission system that can be returned by a policy handler to indicate that the permission should be evaluated using the default decision. The permission system supports two variants of default decisions: "default or allow" and "default or deny". These indicate that the permission should be evaluated using the default decision, and if no default decision is defined, it will fall back to "allow" or "deny" respectively.
```
vinzscam, line 54: An alternative to simplify this a bit could be to add an utility function that detects whether a permission has default decision:

```ts
  if(hasDefaultDecision(permission)) {
    return DefaultPolicyDecision;
  } 
  return DENY;
```

this way adopters could build "default or allow" or "default or deny"

beps/0012-default-permission-decisions/README.md
```diff
@@ -0,0 +1,190 @@
...
+// Conditional default decision for a resource permission
+const catalogEntityUpdatePermission = createPermission({
+  name: 'catalog.entity.update',
+  attributes: { action: 'update' },
+  resourceType: 'catalog-entity',
+  defaultDecision: placeholders => ({
+    result: AuthorizeResult.CONDITIONAL,
+    conditions: catalogConditions.isEntityOwner({
+      claims: placeholders?.ownershipEntityRefs ?? [],
+    }),
```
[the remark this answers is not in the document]
Rugvip: Ah yep you're right, we'll prolly need to switch it all over to the integration router/registry alternative for registering default decisions instead 🤔

beps/0012-default-permission-decisions/README.md
Rugvip: Hmm, kinda tricky to make that information available at that point thought right? Especially considering https://github.com/backstage/backstage/pull/29578#discussion_r2056684747

vinzscam: it depends if the permission framework can discover the information somehow

beps/0012-default-permission-decisions/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Design Details
+
+The proposal introduces default decision support in the permission creation APIs for both basic permissions and resource permissions, including support for conditional decisions that delegate evaluation to resource owners. Default decisions behave differently depending on whether permissions are enabled or not.
+
+### Default decisions with permissions disabled
+
+When permissions are disabled, the frontend will be able to use default definitive decisions, but it will not be able to evaluate default conditional permissions and will always assume `AuthorizeResult.ALLOW`. Backend plugins will still be able to evaluate default conditional permissions however, as they can be returned directly without a request to the permission plugin backend. This means that a user attempting an action on a resource where the default decision is conditional will always be able to attempt the action, but may be denied access once the backend evaluates the permission. In order to improve the user experience, the permission system would need to be enabled.
```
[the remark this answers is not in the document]
drodil: Yes and no 😅 Nothing really prevents removing the whole permissions plugin and I bet people have done so if they are not using permissions

PR #29618, opened 2025-04-15
Add `events-backend-module-google-pubsub`

freben:
Adds a module that is able to transfer messages from Google Pub/Sub subscriptions into the Backstage events system.

This is useful for example if you already dump webhooks to a topic, and would like to subscribe to that topic instead of reconfiguring your source to sending the same webhooks to Backstage directly.

Possibly of particular interest to the reviewer, is the string placeholder substitution that lets you do

```yaml
targetTopic: 'github.{{ message.attributes.x-github-event }}'
```

which of course could have security implications. It carefully checks each step with `Object.hasOwn` (which `lodash.get`, it turns out, does not) so it should be safe. I added tests that try to follow `constructor` and `__proto__`, let me know if there are more tests that you think should be added.


Rugvip (approved):
Let's :shipit: 👍

PR #29629, opened 2025-04-16
Error Handling in Gitlab Repo Push

Parsifal-M:

Hey Team! 👋 

An attempt at making the error messages a bit more clear for the end user when using 'update' or 'delete' I noticed if I use 'update' and try to add a file from the skeleton that doesn't exist in the repo it will throw:

```text
Committing the changes to adding-catalog-info-file failed. Please check that none of the files created by the template already exists. GitbeakerRequestError: Bad Request
```

Which can be a bit misleading if you're trying to update a file (or delete one). I've not used the `commitAction: 'delete'` before so I'm not sure if it would have a similar problem but I would imagine it would.

I also added some tests for the error handling parts, if this is OTT or not really useful please let me know? Also happy to change the approach but just wanted to try and keep it simple for now as I think later on we will move to not using `Gitbeaker` anymore so it will probably change? 💭 

### DIFF
--- a/.changeset/bumpy-showers-design.md
+++ b/.changeset/bumpy-showers-design.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend-module-gitlab': patch
+---
+
+If the commit action is not `create` log a more appropriate error message to the end user advising that the files they're trying to modify might not exist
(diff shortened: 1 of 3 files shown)


benjdlambert (approved):
Makes sense :tada: Thanks!

PR #29647, opened 2025-04-16
packages: remove backend-legacy package

Rugvip:

I think it's time to remove the legacy backend example, there's not very much left of it 😅

If we need to point to examples somewhere we can always pick them from history.


docs/features/software-catalog/external-integrations.md
```diff
@@ -53,9 +53,7 @@ Some defining traits of entity providers:
 
 ### Creating an Entity Provider
 
-The recommended way of instantiating the catalog backend classes is to use the
-`CatalogBuilder`, as illustrated in the
-[example backend here](https://github.com/backstage/backstage/blob/master/packages/backend-legacy/src/plugins/catalog.ts).
+The recommended way of instantiating the catalog backend classes is to use the `CatalogBuilder`.
```
freben, line 56: Ah I need to rewrite these some day

package.json
```diff
@@ -52,7 +52,6 @@
     "snyk:test:package": "yarn snyk:test --include",
     "start": "backstage-cli repo start",
     "start-backend": "echo \"Use 'yarn start example-backend' instead\"",
```
freben, line 54: I wonder how much value we're actually getting out of this one still being named example-backend instead of backend 🤔

freben (approved):
👌

package.json
Rugvip: yep had the same thought

PR #29654, opened 2025-04-17
try to get rid of https://backstage.io/docs/v1.1.0/

freben:
I think this is how?


Rugvip (approved):
Waddafakk, how could this affect it? 😅, but sure, let's try

PR #29664, opened 2025-04-18
fix(scaffolder): make EntityPicker display consistent between dropdown and selected value

tobyh-canva:
This PR fixes an inconsistency in the EntityPicker component where the dropdown options and the selected value were rendered differently.

### DIFF
--- a/.changeset/entity-picker-display-fix.md
+++ b/.changeset/entity-picker-display-fix.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-scaffolder': patch
 ---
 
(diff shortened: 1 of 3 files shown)


drodil:
There was also another try for this in #26866

zcmander:
Also, this seems to use the same trick with blur-handlers as used here: #27218 It's not a bad thing but still feels like a hack around material ui limitations.

rdnkmrr:
Whether these changes are merged or available? why this PR is closed before merging?

tobyh-canva:
@rdnkmrr Unfortunately this PR didn't quite work as expected. I'm afraid I don't have the time (or experience in React) to get it to work 😞

PR #29665, opened 2025-04-18
chore(typos): fix various code typos

GabDug:

Fix many typos, in code, comments, CLI help,  examples, JSDocs...

Some typos in docs/microsite have already been fixed in #29666.

I used the [typos](https://github.com/crate-ci/typos?tab=readme-ov-file) CLI to identify most typos.

This PR does not include the `typos.toml` config or any CI, as there are still a dozen files with false-positives, and the choice for a code typo linter may probably need a decision record.

### Bug fixes / API changes

- ~~`core-components`: `TableFiltersClassKey` had an invalid type value `heder`, while the code used `header`~~
  - ~~I've added to force the resolution of `core-components`, to avoid a type conflict between the versions.~~
  - Removed from this PR.
- `backend-dynamic-feature-service`: deprecated `getAdditionaRemoteInfo`, replaced by `getAdditionalRemoteInfo`. Deprecated one is used as a fallback.
- A few error messages / logs changed

### DIFF
--- a/packages/backend-dynamic-feature-service/src/features/__fixtures__/node_modules/@backstage/backend-plugin-api/package.json
+++ b/packages/backend-dynamic-feature-service/src/features/__fixtures__/node_modules/@backstage/backend-plugin-api/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@backstage/backend-plugin-api",
   "version": "0.0.0",
-  "description": "dummy backstage package that should be skipped by the ComonJSLoduleLoader",
+  "description": "dummy backstage package that should be skipped by the CommonJSModuleLoader",
   "main": "index.js",
   "dependencies": {}
 }
--- a/.changeset/early-colts-accept.md
+++ b/.changeset/early-colts-accept.md
@@ -1,0 +1,6 @@
+---
+'@backstage/backend-dynamic-feature-service': patch
+---
+
+Fixed various typos.
+`FrontendRemoteResolver`'s misspelled `getAdditionaRemoteInfo` has been deprecated. Use the correct spelling `getAdditionalRemoteInfo` instead.
--- a/.changeset/short-teeth-juggle.md
+++ b/.changeset/short-teeth-juggle.md
@@ -1,0 +1,49 @@
+---
+'@backstage/plugin-scaffolder-backend-module-bitbucket-server': patch
+'@backstage/plugin-scaffolder-backend-module-bitbucket-cloud': patch
+'@backstage/plugin-catalog-backend-module-backstage-openapi': patch
+'@backstage/plugin-scaffolder-backend-module-bitbucket': patch
+'@backstage/plugin-search-backend-module-elasticsearch': patch
+'@backstage/plugin-catalog-backend-module-github-org': patch
+'@backstage/plugin-scaffolder-backend-module-gerrit': patch
+'@backstage/plugin-scaffolder-backend-module-github': patch
+'@backstage/plugin-scaffolder-backend-module-gitlab': patch
+'@backstage/plugin-catalog-backend-module-msgraph': patch
+'@backstage/plugin-catalog-backend-module-gitlab': patch
+'@backstage/plugin-catalog-backend-module-logs': patch
+'@backstage/plugin-catalog-backend-module-aws': patch
+'@backstage/frontend-plugin-api': patch
+'@backstage/backend-plugin-api': patch
+'@backstage/plugin-kubernetes-backend': patch
+'@backstage/plugin-kubernetes-cluster': patch
+'@backstage/plugin-scaffolder-backend': patch
+'@backstage/backend-defaults': patch
+'@backstage/frontend-app-api': patch
+'@backstage/plugin-kubernetes-common': patch
+'@backstage/plugin-permission-common': patch
+'@backstage/backend-app-api': patch
+'@backstage/core-compat-api': patch
+'@backstage/core-components': patch
+'@backstage/plugin-kubernetes-react': patch
+'@backstage/plugin-scaffolder-react': patch
+'@backstage/plugin-techdocs-backend': patch
+'@backstage/plugin-kubernetes-node': patch
+'@backstage/plugin-scaffolder-node': patch
+'@backstage/config-loader': patch
+'@backstage/plugin-techdocs-react': patch
+'@backstage/plugin-catalog-graph': patch
+'@backstage/plugin-techdocs-node': patch
+'@backstage/plugin-auth-backend': patch
+'@backstage/repo-tools': patch
+'@backstage/plugin-auth-react': patch
+'@backstage/plugin-kubernetes': patch
+'@backstage/plugin-scaffolder': patch
+'@backstage/plugin-api-docs': patch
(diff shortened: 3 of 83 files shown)


beps/0002-dynamic-frontend-plugins/README.md
awanlin, line 1: This could be included with the others docs as it doesn't get shipped.

canon-docs/src/app/(docs)/theme/theming/page.mdx
awanlin, line 1: This could be included with the others docs as it doesn't get shipped.

docs/releases/v1.3.0.md
```diff
@@ -32,7 +32,7 @@ Several new [entity providers](https://backstage.io/docs/features/software-catal
 
 ### New plugin: Vault
 
-View secrets from [HashiCorp Vault](https://www.vaultproject.io/) alongside your components. PR [#11423](https://github.com/backstage/backstage/pull/11423) contributed by [@ivangonzalezacuna](https://github.com/ivangonzalezacuna)
+View secrets from [HashCorp Vault](https://www.vaultproject.io/) alongside your components. PR [#11423](https://github.com/backstage/backstage/pull/11423) contributed by [@ivangonzalezacuna](https://github.com/ivangonzalezacuna)
```
awanlin: This should be reverted as we are use the correct name already.

microsite/data/plugins/vault.yaml
```diff
@@ -3,7 +3,7 @@ title: Vault
 author: Spread Group
 authorUrl: https://github.com/ivangonzalezacuna
 category: Security
-description: Visualize a list of the secrets stored in your HashiCorp Vault instance.
+description: Visualize a list of the secrets stored in your HashCorp Vault instance.
```
awanlin: This should be reverted as we are use the correct name already.

packages/backend-dynamic-feature-service/src/features/__fixtures__/node_modules/@backstage/backend-plugin-api/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@backstage/backend-plugin-api",
   "version": "0.0.0",
-  "description": "dummy backstage package that should be skipped by the ComonJSLoduleLoader",
+  "description": "dummy backstage package that should be skipped by the CommonJSLoduleLoader",
```
awanlin, line 4:
```suggestion
  "description": "dummy backstage package that should be skipped by the CommonJSModuleLoader",
```

packages/core-components/report.api.md
```diff
@@ -1411,7 +1411,7 @@ export type TableFilter = {
 // Warning: (ae-missing-release-tag) "TableFiltersClassKey" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
 //
 // @public (undocumented)
-export type TableFiltersClassKey = 'root' | 'value' | 'heder' | 'filters';
+export type TableFiltersClassKey = 'root' | 'value' | 'header' | 'filters';
```
awanlin, line 1414: I'll need confirmation from @Rugvip but I don't think we can actually do this. There's some issue around how this works where we can't add new keys or changes the existing ones. Hopefully I'm wrong and just remembering things incorrectly.

packages/core-components/src/components/Table/Filters.tsx
```diff
@@ -24,7 +24,7 @@ import { SelectProps } from '../Select/Select';
 import { coreComponentsTranslationRef } from '../../translation';
 import { useTranslationRef } from '@backstage/core-plugin-api/alpha';
 
-export type TableFiltersClassKey = 'root' | 'value' | 'heder' | 'filters';
+export type TableFiltersClassKey = 'root' | 'value' | 'header' | 'filters';
```
awanlin, line 27: Same here, not sure we can do this 🤔

package.json
```diff
@@ -106,7 +106,8 @@
     "csstype@npm:^3.0.2": "3.0.9",
     "csstype@npm:^3.1.2": "3.0.9",
     "csstype@npm:^3.1.3": "3.0.9",
-    "jest-haste-map@^29.7.0": "patch:jest-haste-map@npm%3A29.7.0#./.yarn/patches/jest-haste-map-npm-29.7.0-e3be419eff.patch"
+    "jest-haste-map@^29.7.0": "patch:jest-haste-map@npm%3A29.7.0#./.yarn/patches/jest-haste-map-npm-29.7.0-e3be419eff.patch",
+    "@backstage/core-components": "workspace:^"
```
awanlin, line 110: Why do we need this?

awanlin (changes requested):
Thanks for taking the time to work through this @GabDug, this is a massive improvement! 

I know this is currently draft but I think it's the best time to make this request: can you please split this up? Ideally everything in the `/docs`, `/microsite`, and `/contrib` folders can be in a separate PR and we can help quickly get that merged in. 

I also went though all the changes and left a few comments as well. 👍 

Also, want to mention we use [Vale](https://github.com/errata-ai/vale) for some amount of spell checking and would prefer we use exiting tools before introducing new ones, can you take a look to see if it can be used to help catch these issues?

packages/core-components/report.api.md
GabDug: You're probably right! My understanding  is that if 2 versions of a plugin are installed, there is a type conflict for the MUI theme overrides. When running `tsc:full`, you get an error that can't be overridden easily as it's in third-party code.
This is why the `resolutions` in the main package.json was required, to fix this type error in the repo.

It could be annoying for consumers, so I fully understand if we rollback and add a comment instead.

GabDug:
@awalin Regarding the Vale overlap, I agree that we should be cautious before introducing any new tool.

I believe Vale is intended for documentation only: it will work best on text files, but it seems we can also configure it to check comments /  JSDocs. It seems our current configuration for Vale is very simple, but we could borrow from various configuration (Google, Redhat, write-good, proselint, ...) and precisely configure them to have a consistent style.

[typos](https://github.com/crate-ci/typos?tab=readme-ov-file) is much simpler and less configurable: it's typos only! Unlike Vale, it also works with source code, whether it's comments, file names or tokens: we can avoid another `heder`, `starredEntitie`, `getAdditionaRemoteInfo`, etc. 

I believe Vale + another tool for code could be complementary, but I'll have too dig deeper into the spellchecking and prose writing tools.

package.json
GabDug: I've removed any changes related to `TableFiltersClassKey` from this PR.

packages/core-components/src/components/Table/Filters.tsx
GabDug: I've removed any changes related to `TableFiltersClassKey` from this PR.

freben (commented):
I like it. We could perhaps consider adding a pre-commit hook but NOT a build check later down the line? Not sure what's the best balance here.

This needs a rebase, and also changesets so that the updated code (where applicable - in particular the dynamic feature service) actually gets released.

freben (approved):

awanlin (approved):
Thanks @GabDug 🚀

PR #29677, opened 2025-04-20
Issue management overhaul

Rugvip:
This is a fairly large change to the way we handle issues in the project. The goal is to guide towards more informative and actionable issues that we are able to respond to faster and more accurately. It's a complete change to our existing issue templates, and it introduces a new system for GitHub labels.

All new labels are documented in a root `LABELS.md` file. The purpose of many of the new labels is to be able to more effectively communicate the state and expectations of issues. For example, the new `needs:repro` label indicates that a bug can't be reproduces and needs reproduction instructions before it can move forward, and the new `priority:contrib-welcome` label signals that an issue might be worked on by owners, but contributions are also welcome. The new documented labels are intended to replace all of our existing labels.

Apart from the new labels, here's a rough list of the other changes:

- Most issue templates now link to the `LABELS.md` to help everyone find that file.
- `bug.yaml`: This was updated to split into description and reproduction sections, with the repro section strongly encouraging that a reproduction repository is provided.
- `feature.yaml`, `rfc.yaml`: RFCs where being misunderstood quite frequently, and the feature request form didn't quite ask for enough information. These are now merged into a single "Change Suggestion" template at `suggestion.yaml` that is essentially a lightweight BEP.
- `documentation.yaml`: A new template for reporting problems in the documentation, since bug reports are a bit overkill for that case.
- `maintenance.yaml`: A new template for tracking maintenance tasks, since change suggestions don't quite fit and it's a fairly common issue type
- `other.yaml`: A catch-all issue template that can be used for any other issue, but with the acknowledgement that issues can be immediately closed without explanation. That one links to Discord as well.
- To make it easier to update common fields a sync script and `.common.yaml` is added to the issue template folder, where common fields are sync'd by ID to the other templates.
- Most issue templates have dropdowns to let the creators select the project area and external integration for the issue, where applicable.
- The `issue.yaml` workflow has been updated to use the `advanced-issue-labeler` action to help label issues based on the choices in the issue template.


.github/ISSUE_TEMPLATE/.common.yaml
```diff
@@ -0,0 +1,92 @@
...
+        - TechDocs
+    validations:
+      required: true
+  - type: dropdown
+    id: integration
+    attributes:
+      label: '🔗 External Integration'
+      description: Is this issue related to an integration with any particular external system?
+      options:
+        - N/A
```
Parsifal-M, line 59: Do we need both N/A and Other? 🤔

Parsifal-M (commented):
Really cool and I think a good step in the right direction! Just one thought 💭

.github/ISSUE_TEMPLATE/.common.yaml
Rugvip: N/A is the default and is intended to signal that this issue isn't related to any particular integration, while Other is for when it is related to some integration, but not listed. I skipped a few rare ones, but if there's a lot of use of Other it indicates that we might need to update the list

.github/workflows/issue.yaml
```diff
@@ -20,6 +20,10 @@ jobs:
         with:
           egress-policy: audit
 
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: .github
```
freben, line 25: 👀

freben (approved):

.github/workflows/issue.yaml
Rugvip: Yep it needs the issue templates to do the labeling, I can throw in a comment. The label actions do have some magic to fetch the config if it's not available locally, but while we're at it we might as well check out all of `.github` to avoid the extra fetch.

Rugvip: Comment added in https://github.com/backstage/backstage/pull/29677/commits/d868732710f37afba382f31a8a43c856da077882

PR #29683, opened 2025-04-21
@backstage/create-app: Add default `.nvmrc` (lts/iron)

nicolasmondain:
## Pull Request

Add default `.nvmrc` (lts/iron) file to the app (`moveAppTask`) to prevent `yarn install` from failing.

**Why?**
Using node v23.11.0, `yarn install` fails on MacOS (reproductible locally using `yarn backstage-create-app`). Could save some time for developers who did not update their nvm current before running the `@backstage/create-app` command line.

```
➤ YN0009: │ isolated-vm@npm:5.0.4 couldn't be built successfully (exit code 1, logs can be found here: /private/var/folders/f9/yznbmq7s4nj44s6jz429cw040000gp/T/xfs-42e3178a/build.log)
```
```
/Users/nicolas.m/Library/Caches/node-gyp/23.7.0/include/node/v8config.h:13:2: error: "C++20 or later required."
   13 | #error "C++20 or later required."
```

> _"If you are using a version of nodejs 20.x or later, you must pass --no-node-snapshot to node"_
> [isolated-vmW](https://github.com/laverdet/isolated-vm?tab=readme-ov-file#requirements)

**notes:** if this PR makes sense to you, I would love to improve this by adding some extras, like:
- asking the developer which TLS he would prefer to use (from 16 to 20)
- checking the current version before `yarn install`


packages/create-app/src/lib/tasks.ts
```diff
@@ -227,6 +227,9 @@ export async function moveAppTask(
           `Failed to move app from ${tempDir} to ${destination}: ${error.message}`,
         );
       })
+      .then(() => {
+        fs.writeFileSync(resolvePath(destination, '.nvmrc'), 'lts/iron');
```
awanlin, line 231: Would it not make more sense to just include the file, say here: https://github.com/backstage/backstage/tree/master/packages/create-app/templates/default-app, that way you don't have to create it?

.changeset/silver-bottles-remain.md
```diff
@@ -0,0 +1,9 @@
+---
+'@backstage/create-app': patch
+---
+
+---
```
awanlin, line 5: Can you remove this extra line? 👍

awanlin (changes requested):
Thanks @nicolasmondain, this makes sense to me but the maintainers will have final say, left a few comments.

Also, regarding your notes: we've generally avoided adding more prompts to keep things streamlined, as long as they are using an active LTS version as per our [Node Policy](https://github.com/backstage/backstage/tree/master/packages/create-app/templates/default-app) they'll be fine. I've also done some work to make this a better experience but I've had other things to focus on: https://github.com/backstage/backstage/pull/27848 but would very much like to complete it.

.changeset/silver-bottles-remain.md
```diff
@@ -0,0 +1,9 @@
...
+
+## '@backstage/create-app': patch
```
awanlin, line 7: This too, please. It will get added to the existing changelog when these changes get released.

awanlin: For example: https://github.com/backstage/backstage/blob/master/packages/create-app/CHANGELOG.md

.changeset/silver-bottles-remain.md
nicolasmondain: @awanlin remove extra lines reverted since prettier needs it.

.changeset/silver-bottles-remain.md
nicolasmondain: @awanlin remove extra lines reverted since prettier needs it.

packages/create-app/src/lib/tasks.ts
freben: you can circumvent that, for example with `--force` https://git-scm.com/docs/git-add

.changeset/silver-bottles-remain.md
freben: No it doesn't :) These should for sure be removed, no other changesets have this. Check the other ones in the same folder

.changeset/silver-bottles-remain.md
awanlin: This is still not correct, sorry. This is what the entire file should look like, it would all have been generated for you when you use the `yarn changeset` command and follow the steps. The contributing guide is correct.

```
---
'@backstage/create-app': patch
---

Add default `.nvmrc` (lts/iron) file to the default-app to prevent `yarn install` from failing.
```

Here's an example from the `.changeset` folder that was mentioned: https://github.com/backstage/backstage/blob/master/.changeset/breezy-hotels-deny.md?plain=1

nicolasmondain: Ouch. No excuses.

vinzscam:
I'm not sure about the proposed fix, as nvm is not a mandatory tool and there might be other adopters who use similar tools. I wonder if we could add a Yarn hook that runs on `yarn install`, checking that we are using the proper version of Node 🤔

PR #29698, opened 2025-04-23
improve LDAP missing metadata.name error message

JessicaJHee:

Currently, the LDAP catalog provider stops after receiving this error for a user:

`"LdapOrgEntityProvider:default refresh failed, TypeError: Malformed entity envelope, TypeError: /metadata/name must NOT have fewer than 1 characters - limit: 1 Malformed entity envelope, TypeError: /metadata/name must NOT have fewer than 1 characters - limit: 1"`

This PR adds a more informative error message that clearly explains the problem and recommends a solution.

The new error message looks like:
`"Error: LDAP search at DN \"ou=users,dc=example,dc=org\" failed; caused by Error: Transform function threw an exception, InputError: User syncing failed: missing 'uid' attribute, consider applying a user filter to skip processing users with incomplete data.`

### DIFF
--- a/plugins/catalog-backend-module-ldap/src/ldap/read.ts
+++ b/plugins/catalog-backend-module-ldap/src/ldap/read.ts
@@ -34,6 +34,7 @@ import { LdapVendor } from './vendors';
 import { GroupTransformer, UserTransformer } from './types';
 import { mapStringAttr } from './util';
 import { LoggerService } from '@backstage/backend-plugin-api';
+import { InputError } from '@backstage/errors';
 
 /**
  * The default implementation of the transformation from an LDAP entry to a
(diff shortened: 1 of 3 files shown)


plugins/catalog-backend-module-ldap/src/ldap/read.ts
```diff
@@ -70,6 +71,13 @@ export async function defaultUserTransformer(
   mapStringAttr(entry, vendor, map.name, v => {
     entity.metadata.name = v;
   });
+
+  if (!entity.metadata.name) {
```
benjdlambert, line 75: @JessicaJHee can we add a test for this new behavior too? :pray:

benjdlambert (changes requested):
Thanks for this just one comment! :pray:

plugins/catalog-backend-module-ldap/src/ldap/read.ts
JessicaJHee: Done! I've also added the same check for the group transformer as well 👍

JessicaJHee: @benjdlambert do you mind taking another look at this? Thanks so much in advance :)

benjdlambert (approved):
Nice :tada:

PR #29716, opened 2025-04-24
feat(logviewer): add textwrapping support for long lines

beanview:

Added a `textWrap` prop that wraps log lines within the `LogViewer` component when the text no longer fits the width instead of creating a horizontal scroll bar
- created a story showing usage
- default the component to use `textWrap` as false so that consumers can opt into this feature when needed 

download

### DIFF
--- a/packages/core-components/src/components/LogViewer/LogLine.tsx
+++ b/packages/core-components/src/components/LogViewer/LogLine.tsx
@@ -193,7 +193,7 @@ export function LogLine({
               (highlight === highlightResultIndex
                 ? classes.textSelectedHighlight
                 : classes.textHighlight),
-            !!setRowHeight && classes.modifierTextWrap,
+            { [classes.textWrap]: !!setRowHeight },
           )}
         >
           <Linkify options={{ render: renderLink }}>{text}</Linkify>
--- a/packages/core-components/src/components/LogViewer/LogLine.tsx
+++ b/packages/core-components/src/components/LogViewer/LogLine.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useMemo } from 'react';
+import { useEffect, useMemo, useRef } from 'react';
 import { AnsiChunk, AnsiLine, ChunkModifiers } from './AnsiProcessor';
 import startCase from 'lodash/startCase';
 import classnames from 'classnames';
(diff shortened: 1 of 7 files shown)


packages/core-components/src/components/LogViewer/LogLine.tsx
```diff
@@ -159,19 +159,28 @@ export interface LogLineProps {
...
   const chunks = useMemo(
     () => calculateHighlightedChunks(line, searchText),
     [line, searchText],
   );
 
+  useEffect(() => {
+    if (lineRef.current && setRowHeight) {
+      setRowHeight(line.lineNumber, lineRef.current.offsetHeight);
+    }
+  }, [line.lineNumber, setRowHeight]);
```
freben, line 182: does this need `lineRef.current` as a dep too? Or is that a non-issue, ie is the ref always set at the point where the effect runs

packages/core-components/src/components/LogViewer/RealLogViewer.tsx
```diff
@@ -81,70 +84,103 @@ export function RealLogViewer(props: RealLogViewerProps) {
     selection.setSelection(line, event.shiftKey);
   };
 
+  function setRowHeight(index: number, size: number) {
+    if (shouldTextWrap && listInstance) {
+      (listInstance as VariableSizeList<AnsiLine[]>).resetAfterIndex(0);
+      // lineNumber is 1-based but index is 0-based
+      heights.current = { ...heights.current, [index - 1]: size };
```
freben, line 91:
```suggestion
      heights.current[index - 1] = size;
```

that spread is going to get expensive to perform over and over for every single line, for large files

packages/core-components/src/components/LogViewer/styles.ts
```diff
@@ -66,10 +66,13 @@ export const useStyles = makeStyles(
     log: {
       fontFamily: '"Monaco", monospace',
       fontSize: theme.typography.pxToRem(12),
+      lineHeight: '20px',
```
freben, line 69: should this also be a `pxToRem` thing or does that not apply here?

freben (commented):
Alright, cool!

Did you check if there was any performance difference for massive files? Sometimes the logs we look at are positively huge and it's important that runtime performance is preserved.

packages/core-components/src/components/LogViewer/styles.ts
beanview: in production code, the `FixedSizeList` is hard coded to 20px tall in the `RealLogViewer` so I set this here so that overflow lines also match that for VariableSizeList

PR #29721, opened 2025-04-24
leverage webhook secrets from github integrations too

### LINKED ISSUE #27598: 🚀 Feature: Support for multiple GitHub webhook secrets
### 🔖 Feature description

Support to configure the [`events-backend-module-github`](https://github.com/backstage/backstage/tree/master/plugins/events-backend-module-github) plugin with multiple webhook secrets would be very useful as currently only one webhook secret can be used to validate incoming webhooks.

### 🎤 Context

I have a Backstage instance configured with two private GitHub apps, each installed in a different GitHub organization. Backstage handles this quite well already, but it does not support validating the webhook callbacks from both app installations if their webhook secret differs.

AFAICT the [`integrations.github.[].apps.[].webhookSecret`](https://github.com/backstage/backstage/blob/7bedf44bf6541a0b7eaeffb481a0ed4912a6bc55/packages/integration/src/github/config.ts#L95) is entirely unused. I propose using any configured GitHub app webhook secret values in addition to `events.modules.github.webhookSecret`, making the latter optional if any apps are configured.

freben:
Fixes #27598

### DIFF
--- a/plugins/events-backend-module-github/src/http/createGithubSignatureValidator.ts
+++ b/plugins/events-backend-module-github/src/http/createGithubSignatureValidator.ts
@@ -15,6 +15,7 @@
  */
 
 import { Config } from '@backstage/config';
+import { ScmIntegrations } from '@backstage/integration';
 import {
   RequestDetails,
   RequestValidationContext,
@@ -36,10 +37,25 @@ import { verify } from '@octokit/webhooks-methods';
 export function createGithubSignatureValidator(
   config: Config,
 ): RequestValidator | undefined {
-  const secret = config.getOptionalString(
+  const webhookSecrets = new Set<string>();
+
+  const integrations = ScmIntegrations.fromConfig(config);
+  for (const integration of integrations.github.list()) {
+    for (const app of integration.config.apps ?? []) {
+      if (app.webhookSecret) {
+        webhookSecrets.add(app.webhookSecret);
+      }
+    }
+  }
+
+  const moduleSecret = config.getOptionalString(
     'events.modules.github.webhookSecret',
   );
-  if (!secret) {
+  if (moduleSecret) {
+    webhookSecrets.add(moduleSecret);
+  }
+
+  if (webhookSecrets.size === 0) {
     return undefined;
   }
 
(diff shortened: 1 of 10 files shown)


plugins/events-backend-module-github/src/http/createGithubSignatureValidator.ts
```diff
@@ -36,10 +37,25 @@ import { verify } from '@octokit/webhooks-methods';
 export function createGithubSignatureValidator(
   config: Config,
 ): RequestValidator | undefined {
-  const secret = config.getOptionalString(
+  const webhookSecrets = new Set<string>();
+
+  const integrations = ScmIntegrations.fromConfig(config);
+  for (const integration of integrations.github.list()) {
```
Rugvip, line 43: do we have info that allows us to select the correct integration for each event? would be a bit more efficient and safer

Rugvip (approved):
Nice :+1:, optional thing

plugins/events-backend-module-github/src/http/createGithubSignatureValidator.ts
freben: Turns out that it's a bit complex.

The config contains app IDs, but the only thing you can get out of webhooks are app _installation_ IDs. So I have to do the octokit dance to make requests to resolve installation IDs to app IDs to find the right secret.

freben: OK, added a full implementation for that, let me know what you think. It came at a pretty high cost but, once we make something like an octokit provider into a core integrations thing, this might be less of an issue.

jhsqsp: @freben It looks like the webhooks' headers have the App ID in `X-GitHub-Hook-Installation-Target-ID`, now. 

[Github's docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads) say this about the header:
```X-GitHub-Hook-Installation-Target-Type: The type of resource where the webhook was created.```

I confirmed it to be consistently the App ID with the various webhooks I spot checked.

Using that header should make the logic querying GH's API to match Installation ID to App ID unneeded. (That API call is currently broken because of a token mismatch, to the best of my knowledge.)

PR #29739, opened 2025-04-25
Fix: renderInTestApp act warnings

vinzscam:

I've noticed that all our tests using `renderInTestApp` from `@backstage/frontend-test-utils` are flooded with the following warnings:

```
   Warning: An update to OAuthRequestDialog inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
```    

this is caused by the `BehaviorSubject` inside `OAuthRequestManager` which causes the re-render of `OAuthRequestDialog` on mount. Since the component is globally mounted, I don't think `BehaviorSubject` is necessary. I'm happy to fix the problem internally in `OAuthRequestDialog` if there are other use cases for `BehaviorSubject` that I may have missed.

### DIFF
--- a/packages/core-app-api/src/apis/implementations/OAuthRequestApi/MockOAuthApi.ts
+++ b/packages/core-app-api/src/apis/implementations/OAuthRequestApi/MockOAuthApi.ts
@@ -17,11 +17,19 @@
 import {
   OAuthRequestApi,
   OAuthRequesterOptions,
+  PendingOAuthRequest,
 } from '@backstage/core-plugin-api';
 import { OAuthRequestManager } from './OAuthRequestManager';
 
 export default class MockOAuthApi implements OAuthRequestApi {
   private readonly real = new OAuthRequestManager();
+  private requests: PendingOAuthRequest[] = [];
+
+  constructor() {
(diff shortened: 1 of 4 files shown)


packages/core-app-api/src/apis/implementations/OAuthRequestApi/MockOAuthApi.ts
```diff
@@ -17,11 +17,19 @@
...
 } from '@backstage/core-plugin-api';
 import { OAuthRequestManager } from './OAuthRequestManager';
 
 export default class MockOAuthApi implements OAuthRequestApi {
   private readonly real = new OAuthRequestManager();
+  private requests: PendingOAuthRequest[] = [];
+
+  constructor() {
+    this.authRequest$().subscribe(requests => {
+      this.requests = requests;
```
freben, line 30: do we never need to merge with what was there before? as in, pushing the new requests onto the array

freben (approved):
Alright, I am honestly not 100% sure but it makes sense at face value!

PR #29747, opened 2025-04-27
cli: experiment with helper for loading clipanion commands

Rugvip:

Experimenting with a little helper to load and run Clipanion commands

On top of #29746

### DIFF
--- a/packages/cli/src/modules/config/alpha.ts
+++ b/packages/cli/src/modules/config/alpha.ts
@@ -17,7 +17,8 @@ import { createCliPlugin } from '../../wiring/factory';
 import yargs from 'yargs';
 import { Command } from 'commander';
 import { lazy } from '../../lib/lazy';
-import { Command as ClipanionCommand, runExit } from 'clipanion';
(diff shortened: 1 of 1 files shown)


aramissennyeydd (approved):
Love it

packages/cli/src/modules/config/alpha.ts
```diff
@@ -102,15 +103,26 @@ export default createCliPlugin({
       path: ['config', 'check'],
       description:
         'Validate that the given configuration loads and matches schema',
-      execute: async ctx => {
-        const { ValidateCommand } =
-          require('./commands/validate') as typeof import('./commands/validate');
-        ValidateCommand.usage = ClipanionCommand.Usage({
+      execute: clipanionLoader(
+        () => require('./commands/validate').ValidateCommand,
```
Rugvip, line 107: We've found that `require` is broken. It won't cause the target to be included in the output and only works if the module happens to exist in the output for other reasons. This needs to be switched to and `import` instead, which until we can fix dynamic imports of CommonJS means we need a pattern similar to the `lazy` helper.

PR #29761, opened 2025-04-28
Added "New" ribbon option to SidebarItem

SonilPro:

Added `newRibbon` option to `SidebarItem` and `SidebarSubmenuItem` for easier recognition of new features.

Adding it because I wasn't able to get the desired effect on my own Backstage instance 😅.
Let me know if it makes sense.
Also, additional feedback is welcome.

I tried to make it work on both collapsed and open sidebar as well as submenus

image

### DIFF
--- a/packages/core-components/src/layout/Sidebar/Items.tsx
+++ b/packages/core-components/src/layout/Sidebar/Items.tsx
@@ -33,6 +33,7 @@ import ArrowDropDown from '@material-ui/icons/ArrowDropDown';
 import ArrowDropUp from '@material-ui/icons/ArrowDropUp';
 import ArrowRightIcon from '@material-ui/icons/ArrowRight';
 import SearchIcon from '@material-ui/icons/Search';
+import NewReleasesIcon from '@material-ui/icons/NewReleases';
 import classnames from 'classnames';
 import type { Location } from 'history';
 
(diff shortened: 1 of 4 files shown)


benjdlambert (changes requested):
I wonder if there's a better way to implement this that doesn't involve adding the explicit ribbon component here.

I think that it might be more beneficial here to build an API where the end user is able to pass in a component which can be rendered in the sample place as the `<NewRibbon>` so that people can provide their own designs rather than being opinionated about it.

Something like:

```tsx
<SidebarSubmenuItem adornment={<NewRibbon />} />
```

Looking at [this](https://mui.com/material-ui/react-text-field/#input-adornments) pattern in MUI especially for doing such things.

SonilPro:
Yeah, I had similar thoughts about this. To be honest they only reason I couldn't add this ribbon in my own backstage instance is because the Link element didn't have relative positioning.

The adornment props seems logical. Give me a few days and I will try to whip up something

packages/core-components/src/layout/Sidebar/Items.tsx
```diff
@@ -154,6 +156,7 @@ const makeSidebarStyles = (sidebarConfig: SidebarConfig) =>
         alignItems: 'center',
         justifyContent: 'center',
         lineHeight: '0',
+        flexShrink: 0,
```
SonilPro, line 159: avoid the icon from being pushed to the left in cases where the text is too big

SonilPro:
@benjdlambert Updated the PR. I tried to keep it as simple as possible, while allowing different usecases

SonilPro:
Can we reopen this?

PR #29773, opened 2025-04-29
Minor changes of software catalog overview document

deepthi-28:
Minor changes of software catalog - overview document


freben (approved):
Sur but, this is mostly a matter of taste. It's best to only make PRs for actual changes or additions.

PR #29774, opened 2025-04-29
Close button added in My Squad - inspect entity section

deepthi-28:
Close icon added in inspect entity

image

### DIFF
--- a/.changeset/blue-planes-reply.md
+++ b/.changeset/blue-planes-reply.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-react': minor
+---
+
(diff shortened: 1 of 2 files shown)


benjdlambert:
Hey @deepthi-28 :wave:

What's the motivation behind this change? I'm not 100% convinced that this is an improvement over the current UI with the close `X` in the top right, and now there's two ways to close this modal.

It's also unfortunate to have a close button in the bottom right in the `ActionsPanel` and something which I've not really seen around Backstage or MUI in general. Do you have any examples with MUI that you can point to?

Thanks :pray:

deepthi-28:
I've added the close button at the top, as I feel it offers a more user-friendly experience. However, if this approach doesn't align with the requirement I can close the pull request.

PR #29785, opened 2025-04-29
Filter-out unrelated schemas for frontend plugin module federation remotes

davidfestal:

Filter-out unrelated schemas for frontend plugin built as a module federation remote.
We should not include the schemas of all the core backstage packages in the configuration schema of the plugin.

### DIFF
--- a/.changeset/clean-geese-heal.md
+++ b/.changeset/clean-geese-heal.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+Only keep relevant schemas when collecting and compiling the configuration schemas of a frontend plugin built as a module federation remote.
--- a/.changeset/yummy-kiwis-end.md
+++ b/.changeset/yummy-kiwis-end.md
@@ -1,0 +1,5 @@
+---
+'@backstage/config-loader': patch
+---
+
+Allow filtering the traversed dependencies when collecting the configuration schemas
--- a/packages/cli/src/modules/build/lib/buildFrontend.ts
+++ b/packages/cli/src/modules/build/lib/buildFrontend.ts
@@ -20,6 +20,7 @@ import { buildBundle, getModuleFederationOptions } from './bundler';
 import { getEnvironmentParallelism } from '../../../lib/parallel';
 import { loadCliConfig } from '../../config/lib/config';
 import { BackstagePackageJson } from '@backstage/cli-node';
+import { buildModuleFederationHostDependencyFilter } from './bundler/moduleFederation';
 
 interface BuildAppOptions {
   targetDir: string;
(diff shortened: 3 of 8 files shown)


davidfestal:
@Rugvip Do you think this could be merged before the 1.39 release ?

Rugvip (changes requested):
Discussed in the framework SIG. It's not clear what the value is at this point and I do think this is something that consumers of the schema probably need to handle instead.

Configuration schema is intended to include all transitive dependencies too, without them the schema is incomplete. I also expect this to break module federation containers that bundle existing `@backstage/*` packages, as well as any direct feature package that happens to depend on a `@backstage/*` dependency outside of its plugin. 

That said I do see that this space is a bit tricky at the moment. The hope is that the move to the new frontend system is going to enable us to completely get rid of plugin-defined frontend configuration schema.

davidfestal:
> Discussed in the framework SIG. It's not clear what the value is at this point and I do think this is something that consumers of the schema probably need to handle instead.

I'm a bit surprised about this, since in a [previous Framework SIG](https://docs.google.com/document/d/1q6wavWvX6byPlO2q7KjD7CvGzZknHWQrqk8Q9il2T3I/edit?tab=t.0#heading=h.uz9rrh50sndl) it had been mentioned as 1 of 4 points that should be tackled to move module federation out of experimental status, which is why I started to work on this one: 
> - Potential items
>     - Guide for open source adopters
>     - Limit to only relevant configuration schema
>     - Solidify design of new package roles
>     - Complete API/design review from core maintainer(s)

So maybe the main point is : what is **_"relevant configuration schema"_** ?

>
> Configuration schema is intended to include all transitive dependencies too, without them the schema is incomplete. I also expect this to break module federation containers that bundle existing `@backstage/*` packages, as well as any direct feature package that happens to depend on a `@backstage/*` dependency outside of its plugin.

As far as I can tell a module federation **_remote_** is expected to be loaded inside a module federation **_host_**.
The schema filtering logic proposed here is **_only_** applied for module federation **_remotes_**.
I agree that this (possibly too simplistic) logic targets MF remotes containing only 1 plugin package, that are loaded inside a backstage frontend application MF host, in other terms it targets the dynamic frontend plugin use-case.
In this use-case the config schemas of `@backstage` packages are mostly expected to be provided by the module federation remote host.
But I get the fact that it might not fit other module federation use-cases like the ones you mention above (splitting the backstage frontend app itself into parts through module federation).
Maybe there is a way to provide a configuration option that would allow tackling both use-cases ?

> 
> That said I do see that this space is a bit tricky at the moment. The hope is that the move to the new frontend system is going to enable us to completely get rid of plugin-defined frontend configuration schema.

davidfestal:
@Rugvip Do you think we could reopen this PR ?
It seems that the discussion on it didn't finish, and the last comment didn't get an answer...

PR #29786, opened 2025-04-29
core-{app,plugin}-api: add support for JSX in translation messages

Rugvip:

Did a quick spike to try out this way of adding support for JSX in translation interpolations. This is an alternative to the implementation in #27193.

The tl;dr of the design is that there's a new `jsx` interpolation format that expects a `ReactNode` and will cause the returned result to be a `ReactNode` as well.

There's a bit of an issue when it comes to communicating support for React nodes in any given message, explained by this part in the new docs:

> When overriding a message you must always keep the `jsx` format for any interpolated values that use it in the original message.

This is of course fairly easy to miss, and there might be an expectation that adding `jsx` to an existing message that doesn't already use it is going to magically do something. Then again that wouldn't really be a reasonable thing for someone overriding messages to do, since you don't control the provided replacements anyway. In the end I think this might actually be the best we can do, since something like the `Trans` component used in #27193 doesn't have a way to communicate support for `ReactNode`s at all.

### DIFF
--- a/.changeset/cool-bikes-push.md
+++ b/.changeset/cool-bikes-push.md
@@ -1,0 +1,5 @@
+---
+'@backstage/core-app-api': patch
+---
+
+Updated `I18nextTranslationApi` to support interpolation with the new `jsx` format.
--- a/.changeset/wicked-dingos-stand.md
+++ b/.changeset/wicked-dingos-stand.md
@@ -1,0 +1,5 @@
+---
+'@backstage/core-plugin-api': patch
+---
+
+Added a new `jsx` interpolation format to `TranslationsApi`. If any of the interpolations in the default translation message uses the `jsx` format, the translation function will always return a `ReactNode`.
--- a/docs/plugins/internationalization.md
+++ b/docs/plugins/internationalization.md
@@ -147,6 +147,43 @@ export const myPluginTranslationRef = createTranslationRef({
 });
 ```
 
+#### React Nodes
+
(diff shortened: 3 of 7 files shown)


benjdlambert:
@rugvip why do we need the `jsx` part at all?

I was thinking that if you provide a interpolation argument that is of type `React.ReactNode` then it would automatically return `React.ReactNode` for you rather than being based off the names of the translation strings?

`t('my.message, { value: <h1>t('my.message.2')</h1> })` would return `React.ReactNode` whilst `t('my.message, { value: 'hello' })` would just return a `string` keeping backwards compatibility?

Rugvip:
@benjdlambert yep the the tricky bit is handling of `undefined` and `false` I'm thinking. Should we always assume that `undefined` is a `ReactNode`, or maybe just prevent that from ever being passed?

Thinking about this some more it might actually be that we want to only consider `JSX.Element`s instead, so if you pass in a JSX element, you get one back.

Thinking about the `jsx` format and behavior of overrides some more: what's in this PR doesn't really make sense. When overriding messages you really shouldn't care about whether what a particular value is, it's completely up to the caller of the translation function.

I'll refactor to only consider `JSX.Element`s and get rid of the format bit, and see where that takes us.

Rugvip:
Updated in https://github.com/backstage/backstage/pull/29786/commits/ae4b3dca5a2b8878a669f9379e13ffa16550a23c, implementation gets a lil bit hackier but also a bit cleaner at the same time as it's able to grab the values. Types end up pretty clean but require overloading `t`. Over-all quite a lot better imo.

PR #29787, opened 2025-04-29
custom fields ui:description

mbenson:
Migrate custom fields to new schema factory function;
standardize field descriptions to prefer ui:description and present consistently,
utilizing ScaffolderField component where possible.


plugins/scaffolder/src/components/fields/EntityPicker/schema.ts
```diff
@@ -13,72 +13,74 @@
...
-import { makeFieldSchemaFromZod } from '../utils';
+import { z as zod } from 'zod';
+import { makeFieldSchema } from '@backstage/plugin-scaffolder-react';
 
 /**
  * @public
  */
-export const entityQueryFilterExpressionSchema = z.record(
-  z
+export const entityQueryFilterExpressionSchema = zod.record(
```
benjdlambert, line 22: Hmm, I wonder if we wanna deprecate this export and have it be `(z: typeof Zod) =>` for support with the new config formats. So that `zod` is injected instead of provided?

plugins/scaffolder/src/components/fields/EntityPicker/schema.ts
```diff
@@ -13,72 +13,74 @@
...
 /**
  * The input props that can be specified under `ui:options` for the
  * `EntityPicker` field extension.
  *
  * @public
  */
-export type EntityPickerUiOptions =
-  typeof EntityPickerFieldSchema.uiOptionsType;
+export type EntityPickerUiOptions = NonNullable<
+  (typeof EntityPickerFieldSchema.TProps.uiSchema)['ui:options']
```
benjdlambert, line 76: Wonder if we should have an equivalent type for this too?

Maybe a `TUiOptions` or something? Can come back to this though.

plugins/scaffolder/src/components/fields/OwnedEntityPicker/OwnedEntityPicker.tsx
```diff
@@ -59,7 +60,11 @@ export const OwnedEntityPicker = (props: OwnedEntityPickerProps) => {
             {...params}
             label={title}
             margin="dense"
-            helperText={description}
+            helperText={
```
benjdlambert, line 63: Can we just migrate this to using `ScaffolderField` instead?

plugins/scaffolder/src/components/fields/RepoBranchPicker/RepoBranchPicker.test.tsx
```diff
@@ -333,4 +335,56 @@ describe('RepoBranchPicker', () => {
       expect(getByText('abc123')).toBeInTheDocument();
     });
   });
+
+  describe('RepoBranchPicker description', () => {
+    const customDescription = 'Custom RepoBranchPicker description';
+
+    it.each([
```
benjdlambert, line 342: I'm a little torn on the use of each in tests.

It makes the tests really hard to understand without a basic description as to what we're really testing. Could we refactor these into explicit tests with a `it('description')` and what's it's testing?

benjdlambert (changes requested):
Hey :wave: thanks for this again!

Just a few things, but looking pretty good :tada:

plugins/scaffolder/src/components/fields/RepoBranchPicker/RepoBranchPicker.test.tsx
mbenson: you're the boss

plugins/scaffolder/src/components/fields/EntityPicker/schema.ts
mbenson: Fine with me; I thought about defining a new export as you suggest and deprecating/refactoring the old export to call the new one. I just wasn't sure how the export might be being used.

mbenson: I've done essentially what I stated above; I note that even though the `entityQueryFilterExpressionSchema` export is marked as `@public` it doesn't seem to actually appear in the API reports. I may push a subsequent commit to remove the `@public` annotation and test the theory that doing so won't break the build.

mbenson: Removed the original export as discussed in SIG 5/2; exported the replacement function without (misleading) TSDoc for consumption by owned/owner picker fields.

plugins/scaffolder/src/components/fields/MyGroupsPicker/schema.ts
```diff
@@ -14,42 +14,34 @@
...
-  z.object({
-    title: z.string().default('Group').describe('Group'),
-    description: z
-      .string()
-      .default('A group you are part of')
-      .describe('The group to which the entity belongs'),
-  }),
-);
+export const MyGroupsPickerFieldSchema = makeFieldSchema({
+  output: z => z.string(),
```
benjdlambert, line 25: Are we missing the `uiSchema` here?

benjdlambert (changes requested):
I think there's just one thing we're missing here :pray:

plugins/scaffolder/src/components/fields/MyGroupsPicker/schema.ts
mbenson: The schema that was there before declared a `default` in the UI schema which, while displayed in the Custom Field Explorer, was not actually picked up in the component and didn't make sense to me. Same for the `description` it declared which AIUI is already "part of" the broader RJSF UI schema. Having eliminated both of these, I omitted the `uiSchema` property since it was marked optional among `makeFieldSchema`'s `options`. Do you agree with my assessment of these individual properties?

benjdlambert (approved):
Let's go with this :tada:

PR #29793, opened 2025-04-29
[Notifications] preprocess notifications to replace slack user ids with display names

PeaWarrior:

Replaces the slack user ids in the Notifications UI with display names instead.

### DIFF
--- a/.changeset/yummy-frogs-swim.md
+++ b/.changeset/yummy-frogs-swim.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-notifications-backend-module-slack': patch
+---
+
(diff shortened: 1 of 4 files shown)


PeaWarrior:
Is there a way to add links to the notifications UI? That might be something we would want to replace as well.

PeaWarrior:
This is very specific in the case that the expected text payload being passed through notifications to the slack module would contain slack specific formatting. example: ```Hello <@U12345678>!```

After having a bit more understanding of notifications, I don't think this is the right approach and instead the expected text payload would be passing in some sort of entity refs where each specific notifications module processor would then parse and process the entity ref to their correct usage for their platform.

PR #29819, opened 2025-05-01
NFS: Fix issue with `UnregisterEntityDialog` not being able to access `routeRefs` in `DialogApi.show`

benjdlambert:
The `compatWrapper` is not used here when we use `DialogApi.show` as they're rendered outside of the original extension context.

Not sure if this is the right fix or if we should be pushing this to the `DialogApi` implementation instead.

### DIFF
--- a/.changeset/yellow-beans-eat.md
+++ b/.changeset/yellow-beans-eat.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog': patch
+---
+
+Fix for missing `routeRef` when using `core-plugin-api` in a dialog context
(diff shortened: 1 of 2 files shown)


benjdlambert:
@Rugvip for some eyes here too.

freben (commented):
It's _supposed_ to be harmless to wrap potentially unnecessarily right?

Yeah the question indeed is if this should be done by the framework, since there is bound to be more dialogs that are at risk 🤔 

Is DOUBLE wrapping a problem? If it is, I'd be wary of approving this pr since it might make it harder to add a framework level fix later. Do you know? Can we double check and test it?

Rugvip (approved):
Makes sense :+1:

PR #29821, opened 2025-05-01
[catalog]: Support disabling one form of relationships from `ldap` module

benjdlambert:
Groups have a `member` attribute and users have a `memberOf` attribute, however these often can drift out of sync, leaving weird states in the Catalog as we collate these results together and deduplicate them.

You can chose to optionally disable one side of these relationships, or even both by providing config in `app-config.yaml` under either the `GroupConfig` or `UserConfig`:

### DIFF
--- a/.changeset/yellow-rats-argue.md
+++ b/.changeset/yellow-rats-argue.md
@@ -1,0 +1,38 @@
+---
+'@backstage/plugin-catalog-backend-module-ldap': patch
+---
+
+Added the ability to configure disabling one side of the relations tree with LDAP.
+
+Groups have a `member` attribute and users have a `memberOf` attribute, however these often can drift out of sync, leaving weird states in the Catalog as we collate these results together and deduplicate them.
+
+You can chose to optionally disable one side of these relationships, or even both by providing config in `app-config.yaml` under either the `GroupConfig` or `UserConfig`:
+
+```yaml
+catalog:
+  providers:
+    ldapOrg:
+      default:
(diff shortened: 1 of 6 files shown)


.changeset/yellow-rats-argue.md
```diff
@@ -0,0 +1,38 @@
...
+          - dn: ou=access,ou=groups,ou=example,dc=example,dc=net
+            options:
+              filter: (&(objectClass=some-group-class)(!(groupType=email)))
+            map:
+              description: l
+            set:
+              metadata.customField: 'hello'
+            parsing:
+              # this defaults to false, to disable use the following:
+              skipMember: true
```
freben, line 37: Hey I had a thought. We probably don't need a new config property.

There's already a system for choosing which attribute to use for [memberOf](https://github.com/backstage/backstage/blob/64ddf3097d9b1395c5983fa191968a584b426758/plugins/catalog-backend-module-ldap/config.d.ts#L224) / [members](https://github.com/backstage/backstage/blob/64ddf3097d9b1395c5983fa191968a584b426758/plugins/catalog-backend-module-ldap/config.d.ts#L409).

Shouldn't we just do like in some other places in the code, and change the type there to `string | null`? We could honestly do that for ALL non-mandatory fields, like picture etc. But starting small.

benjdlambert: Hmm - that's a good shout - perhaps we should do that instead. I can re-use the test suite and try and get this working using that instead.

PR #29822, opened 2025-05-01
fix(GithubUrlReader): use token from options to fetch repoDetails

koalaty-code:

The GithubUrlReader is not currently using the `token` supplied via `options` when making calls to fetch commit details or the repository's default branch. We were seeing rate limit errors in Scaffolder Tasks (`fetch:template` action) due to our Backstage GH integration token running out of rate limit. However, we always pass the user's OAuth token to the `fetch:template` action so we should be using that token wherever possible.

This change updates the two API calls to use the provided token when available, otherwise falling back to the token provided via the `GithubCredentialsProvider`.

### DIFF
--- a/packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.test.ts
+++ b/packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.test.ts
@@ -492,7 +492,7 @@ describe('GithubUrlReader', () => {
     });
 
     it('should override the token when provided', async () => {
-      expect.assertions(1);
+      expect.assertions(2);
 
       const mockHeaders = {
         Authorization: 'bearer blah',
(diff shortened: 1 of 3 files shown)


koalaty-code:
The failing tests don't seem related to my tests. They are all specific to `KubernetesFetcher`. Any help would be appreciated if there is something I need to address there.

packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.test.ts
```diff
@@ -503,6 +503,21 @@ describe('GithubUrlReader', () => {
       });
 
       worker.use(
+        rest.get(
+          'https://ghe.github.com/api/v3/repos/backstage/mock/commits/main/status',
+          (req, res, ctx) => {
+            console.log('IN THE RIGHT HANDLER');
```
benjdlambert, line 509:
```suggestion

```

benjdlambert (changes requested):
Just one small thing around the console log, but otherwise looking pretty good. :tada:

packages/backend-defaults/src/entrypoints/urlReader/lib/GithubUrlReader.test.ts
koalaty-code: Oops! This has been removed.

benjdlambert (approved):
Nice thanks :tada:

PR #29823, opened 2025-05-01
EntityContextMenuItemBluePrint: Add a filter param and config option

marknotfound:

This makes it easier to determine which entities a given context menu item should appear on by implementing a `filter` param which accepts an entity and returns a boolean.


plugins/catalog-react/report-alpha.api.md
```diff
@@ -343,6 +343,7 @@ export const EntityContextMenuItemBlueprint: ExtensionBlueprint<{
 export type EntityContextMenuItemParams = {
   useProps: UseProps;
   icon: JSX_2.Element;
+  filter?: (entity: Entity) => boolean;
```
benjdlambert, line 346: Hmm, I'm a little torn on this.

I wonder if we should really be doubling down on the text query format everywhere and align it with cards like we do [here](https://github.com/backstage/backstage/pull/21480).

It makes it easier to configure this in config I guess rather than having to create your own extension override in code and put it in a module.

I'll defer to @camilaibs and @Rugvip though.

marknotfound: @benjdlambert The reason I didn't do it that way is because `buildFilterFn` which is used to create the filter function from the filter expression/function lives in the `catalog` plugin itself. Though I suppose it wouldn't be a huge lift to refactor it out into `catalog-react`. If that sounds reasonable I can go ahead and do that.

marknotfound:
@benjdlambert Please see latest changes incorporating string and predicate style filtering

plugins/catalog-react/report-alpha.api.md
freben: Right, so when you arrive at the component props it's [already been converted](https://github.com/backstage/backstage/blob/e1a471f81bddb6fe3dca8a0be8e43d94cf334484/plugins/catalog/src/alpha/entityContents.tsx#L91) into purely a callback, up in the blueprint loader. 🤔

plugins/catalog-react/src/alpha/blueprints/EntityContextMenuItemBlueprint.tsx
```diff
@@ -42,17 +51,45 @@ export type UseProps = () =>
...
   useProps: UseProps;
   icon: JSX.Element;
+  filter?: string | EntityPredicate | ((entity: Entity) => boolean);
 };
 
 /** @alpha */
 export const EntityContextMenuItemBlueprint = createExtensionBlueprint({
   kind: 'entity-context-menu-item',
   attachTo: { id: 'page:catalog/entity', input: 'contextMenuItems' },
   output: [coreExtensionData.reactElement],
```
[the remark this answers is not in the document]
marknotfound: @freben I did look into how the entity card filtering works before building this, and I do agree that filtering in the container is a better design. Ultimately I implemented it this way to avoid too much yak shaving. Let me take a look and see exactly how much refactoring would have to happen to implement it in the same style as entity cards.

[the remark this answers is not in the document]
marknotfound: Hey @freben please check out the latest commit when you have some time and let me know if it aligns with what you were thinking.

plugins/catalog/src/alpha/pages.tsx
```diff
@@ -100,7 +108,7 @@ export const catalogEntityPage = PageBlueprint.makeWithOverrides({
 
         const header = inputs.header?.get(
           EntityHeaderBlueprint.dataRefs.element,
-        ) ?? <EntityHeader contextMenuItems={menuItems} />;
+        );
```
marknotfound, line 111: I believe this change is safe to do since the alpha `<EntityLayout />` falls back to this same header and passes on its `contextMenuItems` prop.

plugins/catalog-react/report-alpha.api.md
```diff
@@ -332,17 +338,48 @@ export const EntityContextMenuItemBlueprint: ExtensionBlueprint<{
...
+      {}
+    >;
+  };
 }>;
 
 // @alpha (undocumented)
 export type EntityContextMenuItemParams = {
   useProps: UseProps;
   icon: JSX_2.Element;
+  filter?: string | EntityPredicate | ((entity: Entity) => boolean);
```
Rugvip, line 382: Let's skip `string` filters, as in `entityFilterExpressionDataRef`. They're deprecated and will be removed soon anyway. That was we don't need to move `buildFilterFn` either.

Rugvip (changes requested):
Let's drop support for `string` filters since they are going away. With that I expect this to be a much smaller change and probably good to ship.

marknotfound:
> Let's drop support for `string` filters since they are going away. With that I expect this to be a much smaller change and probably good to ship.

Thanks for the review, @Rugvip ! I believe we're all set to go, please take another look when you've got a moment.

benjdlambert:
Apologies for the misdirection here @marknotfound, glad we got to a solution in the end, even if it's the one that you started with :pray: Just wanted to make sure we had our options covered 👍

marknotfound:
> Apologies for the misdirection here @marknotfound, glad we got to a solution in the end, even if it's the one that you started with 🙏 Just wanted to make sure we had our options covered 👍

@benjdlambert no worries, your input is always appreciated! And a little `git revert` never hurt anybody!

PR #29837, opened 2025-05-04
Adding title prop to customHomePage

GersonTf:
## Add optional title to CustomizableHomepageGrid and improve example app with feature flag

This PR introduces two minor enhancements:

### Optional title prop for CustomizableHomepageGrid:

The title element was previously hardcoded as an empty string. This change makes the title prop optional, maintaining existing behavior by default. However, it allows customization by passing a value like:

`<CustomHomepageGrid config={defaultConfig} title="My Dashboard" />
`

This displays the provided title as shown in the screenshot:
image

### Enhanced example app homepage with feature flag:
Following @ahhhndre's recommendation (based on the [demo site](https://github.com/backstage/demo/blob/f8f53d1dd324d3166e63552d2f8cc0588e7b1093/packages/app/src/App.tsx#L121-L130)), the example app now uses a customizable-home-page-preview feature flag. This allows /home to toggle between the default static homepage and the customizable homepage. This setup is useful for developing and testing features without impacting the default experience.

image

This feature flag changes the home page in /home, the next screenshot shows both options:
image
image

### DIFF
--- a/.changeset/honest-states-repeat.md
+++ b/.changeset/honest-states-repeat.md
@@ -1,5 +1,4 @@
 ---
-'example-app': minor
 '@backstage/plugin-home': minor
 ---
 
--- a/.changeset/honest-states-repeat.md
+++ b/.changeset/honest-states-repeat.md
@@ -1,0 +1,6 @@
+---
+'example-app': minor
+'@backstage/plugin-home': minor
+---
+
+Added optional title prop to customHomePageGrid and improved the example app homepage to add the custom one
--- a/.changeset/honest-states-repeat.md
+++ b/.changeset/honest-states-repeat.md
@@ -1,5 +1,5 @@
 ---
-'@backstage/plugin-home': minor
+'@backstage/plugin-home': patch
 ---
 
 Added optional title prop to `customHomePageGrid`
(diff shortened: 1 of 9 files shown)


.changeset/honest-states-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-home': minor
+---
+
+Added optional title prop to customHomePageGrid and improved the example app homepage to add the custom one
```
awanlin, line 5:
```suggestion
Added optional title prop to `customHomePageGrid`
```

We don't ship the example app and as this gets rolled up in the Home plugin's changelog it won't make a lot of sense in that context.

.changeset/honest-states-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-home': minor
```
awanlin, line 2:
```suggestion
'@backstage/plugin-home': patch
```

As this plugin isn't 1.0 yet this would just be patch, it doesn't look to be breaking either.

awanlin (changes requested):
Thanks @gersontf, left a pair a comments, missed this the first time around but should be quick to update 👍

GersonTf:
> Thanks @GersonTf, left a pair a comments, missed this the first time around but should be quick to update 👍

I commited both your suggestions!

awanlin (approved):
Awesome!

freben (approved):

tikabom (approved):

PR #29844, opened 2025-05-05
[Notifications] format user entity ref mentions to be Slack compatible

PeaWarrior:

Replaces mentioned user entity refs in the notifications payload with Slack compatible mentions.

### DIFF
--- a/.changeset/true-trains-tie.md
+++ b/.changeset/true-trains-tie.md
@@ -1,0 +1,7 @@
+---
+'@backstage/plugin-notifications-backend-module-slack': patch
+---
+
+Notifications which mention user entity refs are now replaced with Slack compatible mentions.
+
+Example: `Welcome <@user:default/billy>!` -> `Welcome <@U123456890>!`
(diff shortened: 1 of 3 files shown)


benjdlambert (approved):
Think this looks good :tada:

@drodil @mareklibra also +1? :pray:

drodil (approved):
LGTM! Maybe one idea is to add cache for the resolved users but can also be done later if necessary

freben:
Agreed on the caching, especially considering that [this dataloader](https://github.com/backstage/backstage/pull/29844/files#diff-c880a6efbd83150686443397fab1c91e042c979801bdcf8af89e2dee8a51a658R330) seems to get created every time and thrown away so it's entirely ineffective.

If that dataloader was placed outside the method, and perhaps [given an ExpiryMap as its storage](https://github.com/backstage/backstage/blob/d133f38ee60f36c99049f36a6f58b8e09b8945fd/plugins/catalog/src/apis/EntityPresentationApi/DefaultEntityPresentationApi.ts#L443), you'd have the perfect cached lookup system for this.

In its current form it's a bit dangerous, especially since it does an unbounded `Promise.all` which makes it so that it can make requests with an unlimited concurrency.

freben:
Added a fix for the above here, so hopefully we can merge them independently https://github.com/backstage/backstage/pull/29856

andrewthauer:
We ended up creating a slack catalog processor that adds an annotation to each user entity that contains their slack id. This acts as a cache and is also useful else where.

PR #29856, opened 2025-05-06
Fix dataloader caching, and use the proper catalog service ref

freben:
As called out in https://github.com/backstage/backstage/pull/29308#pullrequestreview-2732140978


drodil (approved):

vinzscam (approved):
👍

PR #29880, opened 2025-05-07
Add refresh button to "Unprocessed Entities" `FailedEntities` component

KevinSnyderCodes:
Refreshing a failed entity calls the Catalog API endpoint `POST /api/catalog/refresh` with the selected `entityRef`.

Useful for refreshing entities that do not yet exist in the catalog (e.g. first time processing failed) without having to wait for the refresh period.

UI behavior is the same as the delete button, including confirmation dialog.


KevinSnyderCodes:
Let me know if this is a change you would be open to accepting, then I will work through the PR checklist.

freben:
Oh functionality wise, absolutely! So a changeset is needed too.

benjdlambert:
@KevinSnyderCodes is this still something you'd be wanting to contribute? :pray:

KevinSnyderCodes:
Still open to contributing but have not had the time to return to this. Feel free to leave this open or close it, whatever works best for you.

benjdlambert:
@KevinSnyderCodes no worries, we can keep this around :pray: just would be a shame for the work to go stale.

Keep us posted, no stress! :pray:

KevinSnyderCodes:
@benjdlambert Added changeset and attached screenshots to PR description. Let me know if there's anything else you'd like to see done. Thanks!

KevinSnyderCodes:
@awanlin Changed the refresh logic to more closely match the `AboutCard`. Please note the similarities and (perhaps more importantly) the differences outlined in the commit message: https://github.com/backstage/backstage/pull/29880/commits/6b1bd9b66fef5ad2b65ff4cb4b5790c52175374f

Let me know your thoughts. Thanks!

awanlin:
Thanks for the follow up @KevinSnyderCodes, can you add the commit message to the related comments please? I have some follow up questions but with it being in the commit message it's hard to be able to do that 👍

plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
```diff
@@ -157,7 +192,7 @@ export const FailedEntities = () => {
         severity: 'error',
       });
     }
-    setConfirmationDialogOpen(false);
+    setDeleteConfirmationDialogOpen(false);
```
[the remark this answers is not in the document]
KevinSnyderCodes: Unlike the `AboutCard`, we do not check for permissions using the `useEntityPermission()` hook. This requires the current entity to be captured in context, which does not happen with the "Unprocessed Entities" page. Additionally, since unprocessed entities are hidden behind the "DevTools", and since we are not checking for permission for the delete action either, it seems acceptable to not do any additional permissions checks.

plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
```diff
@@ -157,7 +159,32 @@ export const FailedEntities = () => {
...
+  };
+
+  const handleRefresh = async ({ entityRef }: { entityRef: string }) => {
```
KevinSnyderCodes, line 165: Another difference is that the `AboutCard` limits refreshes to locations that begin with `url:` or `file:` as they are less expensive. This makes sense for a view that is made accessible to a wider group of users, but for the "DevTools" page where users with elevated permissions will be using the refresh to debug and fix issues, we should allow refreshes for all locations, regardless of how expensive it is.

plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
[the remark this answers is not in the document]
awanlin: I think it's important to note that this plugin can and is used standalone, it's not required to be used as part of DevTool. Also, I'm a bit confused we do have permissions for delete? https://github.com/backstage/backstage/blob/176d5eb26a7df4ff37a36c47cbe0682c65bad1c2/plugins/catalog-unprocessed-entities-common/src/permissions.ts#L23 Are you saying it's not being used?

plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
awanlin: I think that restriction is valid here as well. The `AboutCard` uses the `backstage.io/managed-by-location` annotation - https://backstage.io/docs/features/software-catalog/well-known-annotations#backstageiomanaged-by-location - entities without these prefixes are likely to come from a discovery mechanism that controls its refresh. Triggering a refresh in these cases would give you the false impression something is going to happen but won't. 

It's been a long time since I dug into this area so I might be wrong, would like the Maintainers insights on this 👍

plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
```diff
@@ -27,12 +27,17 @@ import Box from '@material-ui/core/Box';
 import Typography from '@material-ui/core/Typography';
 import IconButton from '@material-ui/core/IconButton';
 import { Theme, makeStyles } from '@material-ui/core/styles';
-import { alertApiRef, useApi } from '@backstage/core-plugin-api';
+
+import { alertApiRef, errorApiRef, useApi } from '@backstage/core-plugin-api';
+import { catalogApiRef } from '@backstage/plugin-catalog-react';
+import { catalogTranslationRef } from '@backstage/plugin-catalog/alpha';
```
awanlin, line 33: You wouldn't use this here but create your own, more details here: https://backstage.io/docs/plugins/internationalization. 

I'd say skip the Internationalization for now just to keep this PR focused on the feature your adding 👍

plugins/catalog-unprocessed-entities/package.json
```diff
@@ -56,6 +56,9 @@
     "@backstage/core-plugin-api": "workspace:^",
     "@backstage/errors": "workspace:^",
     "@backstage/frontend-plugin-api": "workspace:^",
+    "@backstage/plugin-catalog": "workspace:^",
```
awanlin, line 59: You'll want to remove this, it's generally not a pattern you'd follow - importing a frontend plugins into another frontend plugin - as that's what the `*-react` and/or `*-common` packages are for. It doesn't look like you use anything from this as well. Once removed it should also help get the CI green or at least move forward.

awanlin (changes requested):
Took another quick look, will complete my review tomorrow 👍

plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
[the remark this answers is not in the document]
KevinSnyderCodes: Sorry, I didn't check for the permission in the backend code. Now I see that it's there and being used.

I went ahead and updated the `FailedEntities` component to check permissions for the refresh and delete actions, and disable the button if the user does not have permission to perform the action.

The backend already checks these permissions when the API is called, so no changes needed to be made there.

KevinSnyderCodes:
@awanlin Got some more changes for you to review. Let me know if you have any questions. Thanks!

.changeset/green-eyes-do.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-unprocessed-entities': minor
```
awanlin, line 2:
```suggestion
'@backstage/plugin-catalog-unprocessed-entities': patch
```

As this plugin isn't `1.0` yet this will be `patch` more details here: https://github.com/backstage/backstage/blob/master/REVIEWING.md#reviewing-changeset-bump-levels

plugins/catalog-unprocessed-entities/src/components/FailedEntityActions.tsx
```diff
@@ -0,0 +1,75 @@
...
+  handleDelete: ({
+    entityId,
+    entityRef,
+  }: {
+    entityId: string;
+    entityRef: string;
+  }) => void;
+}) => {
+  const { entity_id, entity_ref } = entity as UnprocessedEntity;
+  const canRefresh = usePermission({
```
awanlin, line 43: Is there a reason we don't also do this:

```ts
const canDelete = usePermission({
    permission: unprocessedEntitiesDeletePermission,
  });
```

Here as well? It's called in the parent component and then just passed into here.

plugins/catalog-unprocessed-entities/src/components/FailedEntities.tsx
awanlin: @freben thoughts on this?

freben: Just to dig into things a bit, [this](https://github.com/backstage/backstage/blob/4157d45c7ede7524dedda4af1815f8124a2f11c8/plugins/catalog-backend/src/service/DefaultRefreshService.ts#L29) is the code that ultimately gets run when you issue a refresh for a given entity ref by calling the catalog backend service. So it looks up the tree to whatever original `kind: Location` (if any) that caused the target component to come into existence.

I'm not sure the original rationale for limiting the location type was one of performance. Maybe the check rather was put in place to not fool the user into trying to refresh something that the engine actually won't be able to handle anyway (if it was some form of specialized thing that didn't come from a Location).

As long as this is what the service API offers, maybe the check makes sense to keep as-is.

awanlin:
Hi @KevinSnyderCodes, this got close due to stale, I can see there was no follow up to my last set of comments. I can re-open this if you still wanted to get this in, I do think this was pretty close to being able to approve from my perspective. What do you think?

PR #29895, opened 2025-05-09
Fix missing edges in the catalog-graph plugin

grantila:

The catalog-graph has a hook called `useEntityRelationNodesAndEdges` which creates the nodes and edges for a set of entities. The current mechanism is incorrect and non-deterministic. It has a set `visitedNodes` so that it doesn't visit nodes more than once while traversing - this part is correct. However, it also uses this set to look for edges to be created or not, which causes graphs to be incomplete and the result depending on the order of the entities.

The PR replaces the `visitedNodes` check, with a check that ensures the edge doesn't already exist before creating it, so that it creates edges to nodes that _have_ been visited, which is correct.

Some tests needed to be updated, and the change feels very reasonable IMO.

### DIFF
--- a/.changeset/some-horses-ask.md
+++ b/.changeset/some-horses-ask.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-graph': patch
+---
+
+Fix rendering of missing relations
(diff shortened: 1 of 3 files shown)


freben (approved):
Alright :)

drodil (approved):

Rugvip (approved):
Thank you! 🎉

PR #29904, opened 2025-05-12
remove explicit tag fetching

freben:
Attempt to undo https://github.com/backstage/backstage/pull/28715 because https://github.com/actions/checkout/issues/2041#issuecomment-2766377359 claims that runners now use a version of git that does the right thing. [Recent runs](https://github.com/backstage/backstage/actions/runs/14936392728/job/41964701450#step:5:21) seem to confirm that that git version is indeed in use.


Rugvip (approved):
Yaaaaaaaaaas

PR #29907, opened 2025-05-12
use the new version of the changeset action

freben:
Pulls in https://github.com/backstage/changesets-action/pull/16 with a pinned changesets CLI that hopefully fixes https://github.com/changesets/changesets/issues/1658


vinzscam (approved):
🚀

PR #29923, opened 2025-05-13
Add tests for Scaffolder

### LINKED ISSUE #28768: 🚀 Feature: Add E2E tests for Scaffolder
### 🔖 Feature description

Given how critical of a component Scaffolder is, It would be beneficial to introduce one or more E2E tests as means to offer some confidence regarding changes to Scaffolder.

### 🎤 Context

This would specially be helpful for major changes to Scaffolder, such as the ones introduced in #27771.

https://backstage.io/docs/plugins/new-backend-system/#testing

solimant:

Closes #28768

### DIFF
--- a/plugins/scaffolder-backend/src/ScaffolderPlugin.test.ts
+++ b/plugins/scaffolder-backend/src/ScaffolderPlugin.test.ts
@@ -1,0 +1,1332 @@
+/*
+ * Copyright 2025 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { IncomingMessage } from 'http';
+import request from 'supertest';
+import waitForExpect from 'wait-for-expect';
+
+import { createBackendModule } from '@backstage/backend-plugin-api';
+import { CatalogClient } from '@backstage/catalog-client';
+import {
+  mockCredentials,
+  mockServices,
+  startTestBackend,
+} from '@backstage/backend-test-utils';
+import { stringifyEntityRef } from '@backstage/catalog-model';
+import { TemplateEntityV1beta3 } from '@backstage/plugin-scaffolder-common';
+import { scaffolderAutocompleteExtensionPoint } from '@backstage/plugin-scaffolder-node/alpha';
+
+import { scaffolderPlugin } from './ScaffolderPlugin';
+
+jest.mock('@backstage/catalog-client');
+
+describe('scaffolderPlugin', () => {
+  const getEntityByRefSpy = jest.spyOn(
+    CatalogClient.prototype,
+    'getEntityByRef',
+  );
+
+  const getMockTemplate = (): TemplateEntityV1beta3 => ({
+    apiVersion: 'scaffolder.backstage.io/v1beta3',
+    kind: 'Template',
+    metadata: {
+      description: 'Create a new CRA website project',
+      name: 'create-react-app-template',
+      tags: ['experimental', 'react', 'cra'],
+      title: 'Create React App Template',
+      annotations: {
+        'backstage.io/managed-by-location': 'url:https://dev.azure.com',
+      },
+    },
+    spec: {
+      owner: 'web@example.com',
+      type: 'website',
+      steps: [
+        {
+          id: 'step-one',
+          name: 'First log',
+          action: 'debug:log',
+          input: {
+            message: 'hello',
+          },
+        },
+        {
+          id: 'step-two',
+          name: 'Second log',
+          action: 'debug:log',
+          input: {
+            message: 'world',
+          },
+          'backstage:permissions': {
+            tags: ['steps-tag'],
+          },
+        },
+      ],
+      parameters: [
+        {
+          type: 'object',
+          required: ['requiredParameter1'],
+          properties: {
+            requiredParameter1: {
+              type: 'string',
+              description: 'Required parameter 1',
+            },
+          },
+        },
+        {
+          type: 'object',
+          required: ['requiredParameter2'],
+          'backstage:permissions': {
+            tags: ['parameters-tag'],
+          },
+          properties: {
+            requiredParameter2: {
+              type: 'string',
+              description: 'Required parameter 2',
+            },
+          },
+        },
+      ],
+    },
+  });
+
+  beforeEach(() => {
+    getEntityByRefSpy.mockImplementation(async () => getMockTemplate());
+  });
+
+  afterEach(() => {
+    getEntityByRefSpy.mockReset();
+  });
+
+  it('supports fetching template parameters schema', async () => {
+    const { server } = await startTestBackend({
+      features: [scaffolderPlugin],
+    });
+
+    const { body, status } = await request(server).get(
+      '/api/scaffolder/v2/templates/default/template/create-react-app-template/parameter-schema',
+    );
+
+    expect(status).toBe(200);
+
+    expect(body).toMatchObject({
+      description: 'Create a new CRA website project',
+      steps: [
+        {
+          schema: {
+            properties: {
+              requiredParameter1: {
+                description: 'Required parameter 1',
+                type: 'string',
+              },
+            },
+            required: ['requiredParameter1'],
+            type: 'object',
+          },
+          title: 'Please enter the following information',
+        },
+        {
+          schema: {
+            'backstage:permissions': { tags: ['parameters-tag'] },
+            properties: {
+              requiredParameter2: {
+                description: 'Required parameter 2',
+                type: 'string',
+              },
+            },
+            required: ['requiredParameter2'],
+            type: 'object',
+          },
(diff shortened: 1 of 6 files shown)


plugins/scaffolder-backend/src/ScaffolderPlugin.test.ts
```diff
@@ -0,0 +1,1004 @@
...
+import { createBackendModule } from '@backstage/backend-plugin-api';
+import { startTestBackend } from '@backstage/backend-test-utils';
+import { CatalogClient } from '@backstage/catalog-client';
+import { stringifyEntityRef } from '@backstage/catalog-model';
+import { TemplateEntityV1beta3 } from '@backstage/plugin-scaffolder-common';
+import { scaffolderAutocompleteExtensionPoint } from '@backstage/plugin-scaffolder-node/alpha';
+
+import { scaffolderPlugin } from './ScaffolderPlugin';
+
+jest.mock('@backstage/catalog-client');
```
benjdlambert, line 29: You can actually use `catalogApiMock` to mock something in-memory instead :tada:

https://github.com/backstage/backstage/blob/2e26579e06197d0ba2d8e55952b40afeee6bf996/plugins/catalog-react/src/components/EntityPeekAheadPopover/EntityPeekAheadPopover.test.tsx#L27-L41

plugins/scaffolder-backend/src/ScaffolderPlugin.test.ts
```diff
@@ -0,0 +1,1004 @@
...
+
+describe('scaffolderPlugin', () => {
+  const getEntityByRefSpy = jest.spyOn(
```
Rugvip, line 32: Wanna quickly jump in and highlight this option, which is a bit simpler to maintain: https://github.com/backstage/backstage/blob/5108226057c62aa0a66dc9f8bb0b7a9ebb258a9c/packages/cli/templates/backend-plugin/src/plugin.test.ts.hbs#L53

solimant: Thanks, Patrik. I've incorporated the change in its own commit since I ended touching more files than I anticipated, and wasn't sure if we're okay with that.

solimant (commented):
Thank you @benjdlambert and @Rugvip - please have a look at the changes you've requested. Happy to address anything further. I've incorporated the changes in a separate commit since I ended touching more files than I anticipated, and wasn't sure if we're okay with that.

plugins/scaffolder-backend/src/ScaffolderPlugin.test.ts
aramissennyeydd: There's a new catalog backend service 😄 - very similar to the frontend one.

solimant: @aramissennyeydd - Yup, that's the one I've used in the updated code, but I had to touch several other files. Is that the idea?

aramissennyeydd: Yeah that looks good, @benjdlambert to check the usage across modules though.

plugins/scaffolder-backend/report.api.md
```diff
@@ -102,17 +102,16 @@ export interface CreateBuiltInActionsOptions {
   // (undocumented)
   additionalTemplateGlobals?: Record<string, TemplateGlobal_2>;
   auth?: AuthService;
-  catalogClient: CatalogApi;
+  catalogClient: CatalogService;
```
Rugvip, line 105: Ah, these are gonna be breaking changes that we'd like to avoid at this point. So we'll need to keep the `CatalogApi` in use here for the time being : /

Rugvip (commented):
Thank you! 👍

Looks good, realized we'll need to roll some of the changes back though in order to maintain API stability. The other option would be to make the changes in a non-breaking way, but not sure exactly how tricky that is at this point.

plugins/scaffolder-backend/report.api.md
solimant: Alright, will revert 👍🏼

plugins/catalog-node/src/testUtils/catalogServiceMock.ts
```diff
@@ -21,7 +21,7 @@ import {
 } from '@backstage/backend-plugin-api';
 import { InMemoryCatalogClient } from '@backstage/catalog-client/testUtils';
 import { Entity } from '@backstage/catalog-model';
-import { catalogServiceRef } from '@backstage/plugin-catalog-node';
+import { catalogServiceRef } from '@backstage/plugin-catalog-node/alpha';
```
solimant, line 24: @Rugvip - I've managed to use `catalogServiceMock` in the tests by just making this slight change without disturbing the public API. Once it's decided to transition to using the non-alpha, `catalogServiceRef`, the transition would reflect here as well. What do you think?

solimant:
@benjdlambert, @Rugvip - following up so this doesn't die out. Are we good with the changes?

packages/cli/templates/backend-plugin/src/plugin.ts.hbs
```diff
@@ -3,7 +3,7 @@ import {
   createBackendPlugin,
 } from '@backstage/backend-plugin-api';
 import { createRouter } from './router';
-import { catalogServiceRef } from '@backstage/plugin-catalog-node';
```
benjdlambert, line 6: I'm not sure we want to make this change? @freben?

I think that these changes are basically going backwards, and we want to be using the service everwhere from the root export.

plugins/catalog-node/src/testUtils/catalogServiceMock.ts
benjdlambert: Hmm, I think that we're probably gonna want to use the one from the root, I have a feeling that this is using the old API definition in the apiRef, rather than the new one, which is why this works.

There's a PR that andre put together for removing the old backend system from scaffolder which is going to be a breaking change anyways. I wonder if we can get that shipped and include the breaking change for the actions pretty quick as to unblock this work.

solimant: @benjdlambert - could you please share that PR is that so I can follow?

packages/cli/templates/backend-plugin/src/plugin.ts.hbs
benjdlambert: But these changes are not in the scaffolder package. They're part of the cli templates, we should not be changing any files that aren't in the scaffolder packages in this PR i think :pray:

plugins/scaffolder-backend/report.api.md
benjdlambert: @Rugvip how do we feel about actually shipping the breaking change at the same time as the legacy backend system removal?

benjdlambert:
@solimant apologies for the delay here, but I think after https://github.com/backstage/backstage/pull/30152 is merged, you should be able to rebase and just use the `catalogServiceMock` everywhere in tests.

Once that PR is merged we will have made most of the breaking changes required for us to get this and the open API schema stuff in for the next release :tada:

plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -857,6 +857,11 @@ export async function createRouter(
           throw new InputError('Input template is not a template');
         }
 
+        const { token } = await auth.getPluginRequestToken({
+          onBehalfOf: credentials,
+          targetPluginId: 'catalog',
+        });
+
```
solimant: @benjdlambert - I reverted this after the change introduced in #30152 because it broke the new tests, but wanted to confirm if it looks right...? I see we're doing this pattern in one other place, but also see `(credentials as any).token` in other places, so not sure which one fits where.

solimant (commented):
@benjdlambert - thank you very much, the PR is much smaller now 🙏🏼 
I have one question regarding how we're computing `backstageToken`. Could you please have a look?

plugins/scaffolder-backend/src/service/router.ts
benjdlambert: Hmm, not sure which one we should be using here to be honest.

Can you see if the previous code actually does get a token when you run `yarn start`? Otherwise happy to change this :pray:

I wonder if there's just no token setup in testing?

solimant: I've managed to overcome this part by setting the `defaultCredentials` on `mockServices.httpAuth`.

solimant: Ben, I thought again about this, and I don't think the test should be mocking `defaultCredentials` because it should come from the request `Authorization` header itself when we do:
```ts
    const { body, status } = await request(server)
      .post('/api/scaffolder/v2/dry-run')
      .auth(mockToken, { type: 'bearer' })
      // .set('Authorization', `Bearer ${mockToken}`) << Or this
      .send({
        template: mockTemplateEntity,
        values: {
          requiredParameter1: 'required-value-1',
          requiredParameter2: 'required-value-2',
        },
        directoryContents: [],
      });
```
So, let me look again into this and see if I can use the actual httpAuth service, and not the mock one.

solimant: @benjdlambert - I've adjusted `mockCredentials.user()` a little so it returns the `token` in the `BackstageCredentials` object, similar to how `createCredentialsWithUserPrincipal` does it in `packages/backend-defaults/src/entrypoints/auth/helpers.ts`. What do you think?

benjdlambert: Had a chat with @freben and we think it's probably OK to add that, although a little unfortunate.

I've just fixed up the rest of the tests, so hopefully this goes green now :pray:

benjdlambert:
@solimant also need a changeset for `@backstage/backend-test-utils` :pray:

PR #29948, opened 2025-05-14
update `refresh_state_references.id` to big int

freben:
This does put an exclusive lock on the table while changing the type and it rewrites the entire table when doing so since the data type is wider than what was there before.

However, since it's a primary column, doing it entirely without blocking is quite involved. It's interesting to note that this column is currently unused in code, and mostly there as a best practice. Also in even the largest of installations this table is a few tens of megabytes and takes a handful of seconds to update. Therefore I judged this in-place simpler solution as a reasonable balance.

### DIFF
--- a/.changeset/tough-heads-knock.md
+++ b/.changeset/tough-heads-knock.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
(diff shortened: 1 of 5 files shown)


Rugvip (approved):
👍

PR #10256, opened 2022-03-17
Add plugin metadata support

grantila:
## Plugin metadata

This PR adds plugin metadata to plugins, in a field called `info`. It allows implementors (and plugin developers) to decorate plugins with:
 * Information about the plugin itself (coming from `package.json`, or totally custom)
 * "Ownership" of a plugin

The metadata (`PluginInfo`) contains name, version, owner, links etc. It will be deduced from `package.json`, and the way to do the decoration is for a plugin to:

```ts
import packageJson from '../package.json';

// ...

export const myPlugin = createPlugin({
  id: 'my-plugin',
  apis: [ /* ... */ ],
  routes: { /* ... */ },
  info: { packageJson }, // <-- add fields here, or just packageJson content
});
```

This means we can use the package.json as the basis for plugin metadata, and plugins can get information from the `package.json` about themselves or other plugins. The idea here is to re-use the `backstage` field for `backstage.name` and `backstage.role` etc.

An implementor can provide a callback `pluginInfoDecorator` to `createApp()` which will be called for each plugin, to allow further decorations of the plugin (extend with links to slack channels or whatnot).

You could argue that adding the package.json to each plugin will grow the webpack bundles. That's true, but it shouldn't be that bad though. A 1.5kB package.json is gzipped down to <600b. Having this would allow other future plugins to inspect this at runtime and indicate package dependency mismatches etc. Could be useful for more than just the metadata explicitly used here.

### Ownership

Alongside the metadata is ownership. It's very useful for implementors to be able to decorate their own _internal_ plugins with owners; what team/group owns a certain plugin. Combined with metadata, this would allow for error boundaries, help/support buttons etc to use that information to provide the user with relevant links to contact the corresponding owner for help / bug reports etc. Implementors can choose their own ownership model (maybe a field in package.json) and use the `pluginInfoDecorator` to decorate that to the `PluginInfo` blob.

This isn't only useful for _internal_ plugins though, but also for those plugins pulled in from this very open source repo (or any other). There's likely several teams being responsible for certain parts of a backstage implementation. You might have a core team being overall responsible for the implementation, maybe a documentation team, CI/CD team etc. This is where the `pluginOwners` option to `createApp()` comes in handy;

An implementor can create a mapping (e.g. in a yaml-file) between package names (npm package names) and internal team names, and these should (preferably) be **catalog groups**, or an entity-ref. With wild-card matching, you could create a mapping like this:

```yaml
# plugin-metadata.yaml
owners:
[...]

PR #29953, opened 2025-05-15
frontend-{app,plugin}-api: add support for plugin info and manifests

Rugvip:

This picks up the work of #10256 and implements it for the new frontend system. On the consumer side there's a new `info()` method on the plugin object that resolves to a new plugin info interface. The interface contains some basic information but is intended to be extended by consuming projects to contain additional fields.

The info is primarily sourced from two placed: a `packageJson` loader that should be used by more or less all plugin instances, and an opaque `manifest` loader that is only for internal use and can point to arbitrary JSON data.

The app instance is responsible for interpreting the `packageJson` and `manifest` and converting it into plugin info objects. By default a couple of fields from `packageJson` are read, and the `manifest` is expected to be a `catalog-info.yaml` file describing the plugin component. This logic is customizable in the app though, and it's expected that projects that make use of this will often add their own fields and logic, or even custom manifest format.

In addition there's a new `useAppNode` hook to be able to access the `AppNode` of the closest `ExtensionBoundary`.


packages/frontend-app-api/src/wiring/createPluginInfoAttacher.ts
```diff
@@ -0,0 +1,249 @@
...
+  };
+}
+
+function normalizePluginInfo(info: FrontendPluginInfo) {
+  return {
+    ...info,
+    ownerEntityRefs: info.ownerEntityRefs?.map(ref =>
+      stringifyEntityRef(
+        parseEntityRef(ref, {
+          defaultKind: 'group',
```
freben, line 93:
```suggestion
          defaultKind: 'Group',
```

doesn't really matter, but felt good having set right :)

packages/frontend-app-api/src/wiring/createPluginInfoAttacher.ts
```diff
@@ -0,0 +1,249 @@
...
+  return (feature: FrontendFeature) => {
+    if (!OpaqueFrontendPlugin.isType(feature)) {
+      return feature;
+    }
+
+    const plugin = OpaqueFrontendPlugin.toInternal(feature);
+
+    return {
+      ...plugin,
+      info: once(async () => {
```
freben, line 65: Is there something to be said about error handling? These are going to be pretty "late" runtime errors if something goes wrong right? How will those errors generally be noticed by users in the end? Can for example a slight syntax error in a custom field in a catalog-info make the entire interface be unusable? Or will it generally be more localized problems for example inside the support button and similar places that use the new hook for that one single plugin page?

packages/app-next/app-config.yaml
```diff
@@ -8,6 +8,20 @@ app:
...
+      info:
+        description: 'This description was overridden in packages/app-next/app-config.yaml'
+    - match:
+        pluginId: /^catalog(-.*)?$/
+      info:
+        ownerEntityRefs: [cubic-belugas]
+    - match:
+        packageName: '@backstage/plugin-scaffolder'
+      info:
+        ownerEntityRefs: [cubic-belugas]
```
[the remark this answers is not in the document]
Rugvip: Eh, could put pretty much anything here, figured this is fine 😅. Obviously it's not aiming to be extensive

packages/frontend-app-api/src/wiring/createPluginInfoAttacher.ts
```diff
@@ -0,0 +1,247 @@
...
+  ) {
+    try {
+      const url = new URL(packageJson.repository?.url);
+      if (url.protocol === 'http:' || url.protocol === 'https:') {
+        // TODO(Rugvip): Support more variants
+        if (
+          url.hostname === 'github.com' &&
+          typeof packageJson.repository.directory === 'string'
+        ) {
+          const path = `${url.pathname}/tree/master/${packageJson.repository.directory}`;
```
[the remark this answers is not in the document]
Rugvip: Yep maybe we should drop this altogether and find a different way. I brought this over from the old PR but it's indeed pretty tricky. I'll switch it to `-` for now though to handle this case at least

packages/frontend-plugin-api/src/wiring/createFrontendPlugin.ts
```diff
@@ -38,9 +99,21 @@ export interface FrontendPlugin<
...
+   * Loads the plugin info.
+   */
+  info(): Promise<FrontendPluginInfo>;
   getExtension<TId extends keyof TExtensionMap>(id: TId): TExtensionMap[TId];
   withOverrides(options: {
     extensions: Array<ExtensionDefinition>;
+
+    /**
+     * Overrides to merge with the original plugin info. The merging is done for
+     * each top-level field. Setting a field explicitly to `undefined` will
```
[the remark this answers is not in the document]
Rugvip: Oh this comment is actually outdated. It used to support overriding the actual info, but now it's just the sources.

packages/frontend-plugin-api/src/components/AppNodeProvider.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+          ),
+        }),
+      ).toThrow('AppNodeContext v1 not available');
+    });
+    expect(error).toEqual([
+      expect.objectContaining({
+        detail: new Error('AppNodeContext v1 not available'),
+      }),
+      expect.objectContaining({
+        detail: new Error('AppNodeContext v1 not available'),
```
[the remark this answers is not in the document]
Rugvip: Yep not quite sure why it's logged twice, but it is

packages/frontend-app-api/src/wiring/createPluginInfoAttacher.ts
Rugvip: Yep I aimed to push errors as late as possible everywhere, leaving it up to projects to validate manifests if needed or just generally being very permissive with the input. The intention is to have errors pop up pretty deep related to support buttons or error boundaries, but not completely break those either.

packages/frontend-app-api/src/wiring/createPluginInfoAttacher.ts
```diff
@@ -0,0 +1,249 @@
...
+  stringifyEntityRef,
+} from '../../../catalog-model/src/entity/ref';
+
+/**
+ * A function that resolves plugin info from a plugin manifest and package.json.
+ *
+ * @public
+ */
+export type FrontendPluginInfoResolver = (ctx: {
+  packageJson?(): Promise<JsonObject>;
```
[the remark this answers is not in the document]
Rugvip: Yep I put the `{name: string}` in there to make it a tiny bit harder to make the mistake of importing the wrong file. In the end there are very few guarantees here though and they're simply just treated as opaque JSON blobs.

docs/frontend-system/architecture/10-app.md
```diff
@@ -77,3 +77,98 @@ app:
...
+
+By default the app will pick a few common fields from `package.json` files, and assume that the opaque manifests are `catalog-info.yaml` files that some information can be gathered from too. This information will then be available via the `info()` method on plugin instances, returning a structure of the `FrontendPluginInfo` type.
+
+### Extending Plugin Info
+
+The default plugin info is intended as a base to build upon. As part of setting up an app you can both customize the way that the plugin info is resolved, as well as extend the `FrontendPluginInfo` type to include more information.
+
+In order to extend the `FrontendPluginInfo` type you use [TypeScript module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation). This makes it possible to extend the `FrontendPluginInfo` inferface with additional fields, which you can then add custom resolution logic for as well as access within the app. For example, you might add a `slackChannel` field as follows:
```
Parsifal-M, line 91: Probably mean interface here?

Parsifal-M (commented):
Nice! Love the examples usage as well 😁

docs/frontend-system/architecture/10-app.md
Rugvip: which bit exactly?

freben, line 91:
```suggestion
In order to extend the `FrontendPluginInfo` type you use [TypeScript module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation). This makes it possible to extend the `FrontendPluginInfo` interface with additional fields, which you can then add custom resolution logic for as well as access within the app. For example, you might add a `slackChannel` field as follows:
```

Parsifal-M: Sorry I was on my phone on the train I should have made a suggestion instead 😅 

`This makes it possible to extend the `FrontendPluginInfo` inferface with additional fields`

Rugvip: Aha 😁

awanlin (approved):
DevTools changes look good to me, excited to see this feature moving forward! 🚀

PR #29959, opened 2025-05-15
fix(catalog-model): add support for icons containing colons

benjidotsh:
This PR adds support for icons containing colons, e.g. `kind:api`. Since some of the [default icons](https://backstage.io/docs/reference/core-app-api.appicons) contain a colon, I think it makes sense to allow them during validation as well.

### DIFF
--- a/packages/catalog-model/src/entity/policies/FieldFormatEntityPolicy.ts
+++ b/packages/catalog-model/src/entity/policies/FieldFormatEntityPolicy.ts
@@ -161,7 +161,7 @@ export class FieldFormatEntityPolicy implements EntityPolicy {
       optional(
         `links.${i}.icon`,
         links[i]?.icon,
-        KubernetesValidatorFunctions.isValidObjectName,
+        CommonValidatorFunctions.isNonEmptyString,
       );
     }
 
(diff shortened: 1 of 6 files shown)


benjidotsh:
Tests still seem to be failing, but it doesn't look related?

packages/catalog-model/src/entity/policies/FieldFormatEntityPolicy.ts
```diff
@@ -161,7 +165,7 @@ export class FieldFormatEntityPolicy implements EntityPolicy {
       optional(
         `links.${i}.icon`,
         links[i]?.icon,
-        KubernetesValidatorFunctions.isValidObjectName,
+        CommonValidatorFunctions.isValidIcon,
```
Rugvip, line 168: Let's simply use `isNonEmptyString` instead. The icon IDs don't really have a very strict format, and it's a bit overkill to create a bespoke validation function for them.

Rugvip (changes requested):
Thank you! 👍 

Makes sense to move away from `isValidObjectName`, but wanna try to keep things simple 😁

Rugvip (approved):
Thanks! 👍

PR #29552, opened 2025-04-11
Permission: improve request batching

vinzscam:
## Hey, I just made a Pull Request!

This PR optimizes the permission framework requests flow by batching requests. This fixes #23303, as the search plugin currently performs result-by-result authorization. When attempting to authorize the catalog search results, search sends: 

```json
[
  {
    "items": [
      {
        "id": "b8c3abdf-2f8a-44ff-ab52-5af9671a92bf",
        "resourceRef": "user:default/sarah.gilroy",
        "permission": {
          "type": "resource",
          "name": "catalog.entity.read",
          "attributes": {
            "action": "read"
          },
          "resourceType": "catalog-entity"
        }
      },
      {
        "id": "6a38049d-de04-46fa-84cf-9861aaebba0f",
        "resourceRef": "group:default/team-c",
        "permission": {
          "type": "resource",
          "name": "catalog.entity.read",
          "attributes": {
            "action": "read"
          },
          "resourceType": "catalog-entity"
        }
      },
      ... // x times depending on the set limit
]
```

if you have a permission policy returning a conditional result, for example containing the following conditional decision:

```js
  if (
    isPermission(request.permission, catalogEntityReadPermission)
  ) {
    return createCatalogConditionalDecision(request.permission, {
      not: catalogConditions.hasMetadata({ key: 'something' }),
    });
  }
```

the permission-backend sends to catalog the resolved value of the permission policy:

```json
  {
    "items": [
      {
        "id": "f2e0b4a1-79ee-41c4-8dcb-9187286d3e74",
        "resourceRef": "location:default/example-systems",
        "resourceType": "catalog-entity",
        "conditions": {
          "not": {
            "rule": "HAS_METADATA",
            "resourceType": "catalog-entity",
            "params": {
              "key": "something"
            }
          }
        }
      },
      {
        "id": "cdef0c2e-2ec3-4284-8204-251b54483f35",
        "resourceRef": "user:default/lucy.sheehan",
        "resourceType": "catalog-entity",
        "conditions": {
          "not": {
            "rule": "HAS_METADATA",
            "resourceType": "catalog-entity",
            "params": {
              "key": "something"
            }
          }
        }
      },
      // ... x times depending on the set limit
]
```

Note that in the request above the resolved value of the permission policy is repeated x times. So if your permission policy is more complex than the one above, the payload limit is exhausted faster.

With the new approach the first request becomes:

```json
  {
    "items": [
      {
[...]

PR #29983, opened 2025-05-16
feat(permissions): dataloader for permission client

drodil:

this introduces dataloader per user/service that can be used to batch authorize and authorizeConditional requests for improved performance.

tested this in backstage local environment and already it batched more than 80% of the authorize requests making it much more lightweight network wise.

relates to #23303 #28836 #20188 #28837 #28853 #29552


drodil:
Maybe something especially @vinzscam is interested?

plugins/permission-common/src/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  set(key: K, value: V) {
+    this.prune();
+    const result = super.set(key, value);
+    this.#timestamps.set(key, Date.now());
+    return result;
+  }
+
+  get(key: K) {
+    this.prune();
```
freben: Seems expensive. The other implementations just cut out the one entry. Did we need to prune this periodically to a greater degree here than in other usages of dataloader?

plugins/permission-common/report.api.md
```diff
@@ -183,9 +183,19 @@ export class PermissionClient implements PermissionEvaluator {
   ): Promise<QueryPermissionResponse[]>;
 }
 
+// @public
+export type PermissionClientOptions = {
+  discovery: DiscoveryApi;
+  config: Config;
+  loaderCacheTtl?: number;
+  cacheTtl?: number;
+  batchDelay?: number;
```
freben: A bit unfortunate to have these here. They're for tests right? Can you make it a dedicated `@internal` static `forTests` method or something like that instead?

plugins/permission-common/report.api.md
```diff
@@ -183,9 +183,19 @@ export class PermissionClient implements PermissionEvaluator {
...
+};
+
 // @public
 export type PermissionClientRequestOptions = {
   token?: string;
+  identifier?: string;
```
freben: A fairly generic name - I don't think it has a doc comment either right? Naming is hard.

Especially in permission-node we could actually make it a credentials object instead, making it an internal concern what the client does with it at least from the backend standpoint, right? And a similar thing on the frontend side, exposing the backend identity (the return value from the identity service), or even removing the parameter entirely and forcibly getting the identity inside the implementation instead. Compare the distinction between the catalogService api on the node side being different from the frontend catalog utility api.

plugins/permission-node/src/ServerPermissionClient.ts
```diff
@@ -110,16 +110,28 @@ export class ServerPermissionClient implements PermissionsService {
...
+        return { identifier: 'none' };
       }
 
-      return this.#auth.getPluginRequestToken({
+      const { token } = await this.#auth.getPluginRequestToken({
         onBehalfOf: options.credentials,
         targetPluginId: 'permission',
       });
+
+      let identifier: string | undefined = undefined;
```
freben: It's interesting that we've seen a couple of times now that people stringify credentials. We might want to standardise this.

plugins/permission-common/src/PermissionClient.ts
```diff
@@ -281,6 +336,77 @@ export class PermissionClient implements PermissionEvaluator {
...
+          options,
+        );
+      },
+      {
+        name: 'PermissionClient.authorizeLoader',
+        cacheMap: new ExpiryMap(this.cacheTtl),
+        maxBatchSize: MAX_BATCH_SIZE,
+        batchScheduleFn: cb => setTimeout(cb, this.batchDelay),
+        cacheKeyFn: req => {
+          return btoa(JSON.stringify(req));
```
freben: are these always cleanly json serializable?

freben (commented):
Alright i just made a quick runthrough to maybe help move a little forward, had some thoughts

plugins/permission-node/src/ServerPermissionClient.ts
drodil: Yes, it would be nice to have common `toString` method in the `BackstageCredentials`

plugins/permission-common/report.api.md
drodil: Yeah, for tests. I will refactor these out of the public API.

plugins/permission-common/src/utils.ts
drodil: We could move this to only be done in `set`, but it might give back already "old" dataloaders, which work totally fine and are not an issue. It's just balancing between memory consumption and performance; with a lot of concurrent users, the number of loaders can get quite big.

plugins/permission-common/report.api.md
drodil: Yes, the issue is that there is no common type that works for both backend and frontend (AFAIK). In the frontend, you can get the `BackstageUserIdentity` from `IdentityApi`, and in the backend, you can get `BackstageCredentials`.  There's not really a common ground between these two, and it's not possible to import the frontend type in the `PermissionService`.

plugins/permission-common/src/PermissionClient.ts
drodil: These are of type `QueryPermissionRequest` or `AuthorizePermissionRequest`, so they should be.

plugins/permission-common/report.api.md
drodil: I created a new `PermissionIdentity` type that forces the identifier to be in specific format. The stringifying is then done inside the `PermissionClient` internally.

drodil:
@freben made some refactoring based on your comments, WDYT?

freben:
I'll review this as soon as I can, but wanted to also link to https://github.com/backstage/backstage/pull/30127 which might be useful here if it existed. There are similar use cases in techdocs backend that caches information on a per-credential basis and ends up using the raw token for that - which isn't desirable because it blocks migration to using the new-backend credentials system which doesn't expose the raw token.

drodil:
Thanks @freben will check that tomorrow!

drodil:
Rebased

drodil:
@vinzscam any chance to get a review?

plugins/permission-react/src/hooks/usePermission.ts
```diff
@@ -61,6 +62,39 @@ export function usePermission(
...
 ): AsyncPermissionResult {
   const permissionApi = useApi(permissionApiRef);
+  const configApi = useApi(configApiRef);
+
+  const enableDataloaderRequests =
+    configApi.getOptionalBoolean(
+      'permission.EXPERIMENTAL_enableDataloaderRequests',
+    ) ?? false;
+
+  if (enableDataloaderRequests) {
```
vinzscam: the early return breaks the rule of hooks. We don't know where we are heading with config in the future, so it would be nice to fix this

plugins/permission-react/src/hooks/usePermission.ts
```diff
@@ -61,6 +62,39 @@ export function usePermission(
...
+      // must be asynchronously loaded, so instead we short-circuit to a deny when
+      // no resourceRef is supplied, on the assumption that the resourceRef is
+      // still loading outside the hook.
+      if (isResourcePermission(input.permission) && !input.resourceRef) {
+        return AuthorizeResult.DENY;
+      }
+
+      const { result } = await permissionApi.authorize(input);
+      return result;
+    }, [permissionApi, input]);
```
vinzscam: since `input` changes on every re-render, this will cause `permissionApi.authorize()` to be invoked multiple times with the same permissions. Will this send more requests?

plugins/permission-react/src/hooks/usePermission.ts
```diff
@@ -61,6 +62,39 @@ export function usePermission(
...
+
+    return {
+      loading: value === undefined,
```
vinzscam: ideally this should be just `loading` from `useAsync`, right?
```suggestion
      loading,
```

plugins/permission-react/src/hooks/usePermission.ts
```diff
@@ -61,6 +62,39 @@ export function usePermission(
...
+      // We could make the resourceRef parameter required to avoid this check, but
+      // it would make using this hook difficult in situations where the entity
+      // must be asynchronously loaded, so instead we short-circuit to a deny when
+      // no resourceRef is supplied, on the assumption that the resourceRef is
+      // still loading outside the hook.
+      if (isResourcePermission(input.permission) && !input.resourceRef) {
+        return AuthorizeResult.DENY;
+      }
+
+      const { result } = await permissionApi.authorize(input);
```
vinzscam: Thanks. It looks like all the issues described in the description are all related to the payload being too large. I wonder if splitting `input` into batches can help solving the issue and simplify things a bit? We are already using swr so it would be nice to not add another library to handle similar problems

vinzscam (changes requested):

plugins/permission-react/src/hooks/usePermission.ts
drodil: Yes in the frontend, but there are so many authorize requests coming also from backend. That's why I think unifying the logic to CatalogClient is good for performance and the issue this addresses. In our case for example, the 413 error is originating from the backend, not frontend. 

Of course I can modify this to still use swr in the frontend but it might come with a cost of double batching the requests.

plugins/permission-react/src/hooks/usePermission.ts
drodil: Yep but also conditional hooks are not cool so not sure there is a way around this 😅 Maybe just use swr and keep this file as original

drodil:
Not stale, @vinzscam any chance for a second look?

drodil:
Not stale

plugins/permission-common/report.api.md
```diff
@@ -183,9 +183,16 @@ export class PermissionClient implements PermissionEvaluator {
...
+// @public
+export type PermissionClientOptions = {
+  discovery: DiscoveryApi;
+  config: Config;
+};
+
 // @public
 export type PermissionClientRequestOptions = {
   token?: string;
+  permissionIdentity?: PermissionIdentity;
```
freben: After https://github.com/backstage/backstage/pull/30127 we have more deliberately well defined `toString` and `toJSON` forms of backend identities. Would it make sense to make this just a string, and calling it something like `identityKey` or so? Then typical backend callers can pass in `String(credentials)`

drodil: Done

plugins/permission-common/src/PermissionClient.ts
```diff
@@ -149,10 +188,14 @@ export class PermissionClient implements PermissionEvaluator {
       return requests.map(_ => ({ result: AuthorizeResult.ALLOW as const }));
     }
 
+    if (this.enableDataloaderRequests) {
+      const loader = this.getAuthorizeLoader(options);
+      return Promise.all(requests.map(r => loader.load(r)));
```
freben: just a thought: there's also `loadMany` that's more efficient, but the semantics of that one is a bit different since it returns actual errors as elements if something goes wrong, and maybe not worth the fuss

plugins/permission-common/src/PermissionClient.ts
```diff
@@ -281,6 +329,101 @@ export class PermissionClient implements PermissionEvaluator {
   private getAuthorizationHeader(token?: string): Record<string, string> {
     return token ? { Authorization: `Bearer ${token}` } : {};
   }
+
+  private getAuthorizeLoader(options?: PermissionClientRequestOptions) {
+    if (options?.identityKey) {
```
freben: Hm. I'd rather have expected something like `const key = options?.token || options?.identityKey || '';` or something like that and use that below. Or even ``${options?.token}${options?.identityKey}`` :shrug: Some way of always taking the inputs into account, either way. Because I thought that some callers will be passing in a token and expect the responses to vary depending on that. Right?

plugins/permission-common/src/PermissionClient.ts
```diff
@@ -281,6 +329,101 @@ export class PermissionClient implements PermissionEvaluator {
...
+      this.authorizeLoaderMap.set(options.identityKey, newLoader);
+    }
+    return newLoader;
+  }
+
+  private getAuthorizeConditionalLoader(
+    options?: PermissionClientRequestOptions,
+  ) {
+    if (options?.identityKey) {
```
freben: same here

plugins/permission-common/src/PermissionClient.ts
```diff
@@ -281,6 +329,101 @@ export class PermissionClient implements PermissionEvaluator {
...
+          options,
+        );
+      },
+      {
+        name: 'PermissionClient.authorizeConditionalLoader',
+        cacheMap: new ExpiryMap(this.cacheTtl),
+        maxBatchSize: MAX_BATCH_SIZE,
+        batchScheduleFn: cb => setTimeout(cb, this.batchDelay),
+        cacheKeyFn: (req: QueryPermissionRequest) => {
+          return btoa(JSON.stringify(req));
```
freben: We could skip the the btoa right? Did you mean to use a hash or something like that instead? Otherwise we can probably remove

plugins/permission-common/src/PermissionClient.ts
drodil: Oh, this is old way, please check the updated version. Getting rid of the identityKey all together.

plugins/permission-common/src/PermissionClient.ts
drodil: Hash could sure work better

drodil: Hm. Actually, cannot use `crypto` as this is common client also for frontend. So maybe just remove it.

drodil:
Not stale

drodil:
Not stale

freben (approved):
Alright, I'm happy about the state of this. @vinzscam ?

drodil:
bump and ping @vinzscam

drodil:
Not stale, waiting for @vinzscam

Issue #29938, opened 2025-05-13
Azure Location Integration/ ADO does not work with Legacy Azure URL org.visualstudio.com


Parms-d:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
Azure Integration
Azure Legacy Integration
Azure visualstudio.com issues
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

Azure

### 📝 Description & Context

It seems the azure integration for backstage only accepts dev.azure.com, our env is still using the legacy URL org.visualstudio.com. When trying to register a new component or and existing component I get an error of 

app-config.yaml

integrations:
  azure:
    - host: org.visualstudio.com
      credentials:
        - organizations:
            - org
          personalAccessToken: ${PAT_TOKEN}

-with org.visualstudio.com as integration (registering existing component and same issue using templates):

`{"error":{"name":"InputError","message":"NotFoundError: Unable to read url, no matching files found for https://org.visualstudio.com/Sandbox/_git/backstage_catalog?path=/catalog-info.yaml"},"request":{"method":"POST","url":"/locations?dryRun=true"},"response":{"statusCode":400}}`

`92025-05-13T16:13:16.620Z Finished step Fetch Base
102025-05-13T16:13:16.623Z Beginning step Publish
112025-05-13T16:13:16.750Z Error: Failed to find api location for area: Location id: e81700f7-xxx    at /app/node_modules/azure-devops-node-api/VsoClient.js:86:2313    at Object.dynatraceRegularInvoke [as doInvoke] (/opt/dynatrace/oneagent-paas/agent/bin/1.311.78.20250507-174450/any/nodejs/nodejsagent.js:4087:22)14    at /opt/dynatrace/oneagent-paas/agent/bin/1.311.78.20250507-174450/any/nodejs/nodejsagent.js:9190:5915    at AsyncLocalStorage.run (node:async_hooks:335:14)`

----------------
app-config.yaml

integrations:
  azure:
    - host: dev.azure.com
      credentials:
        - organizations:
            - org
          personalAccessToken: ${PAT_TOKEN}

-with dev.azure.com as integration (able to register a existing component fine changing URL to dev.azure.com, but using template fails, adding both in integration then gives the above API location error on register)

`2025-05-13T16:03:39.406Z Finished step Publish
[...]

PR #29985, opened 2025-05-16
Legacy Azure URL support ADO

fcamgz:

Include hostname endsWith visualstudio.com to enable code search for legacy Azure Url. #29938

### DIFF
--- a/.changeset/wise-teeth-study.md
+++ b/.changeset/wise-teeth-study.md
@@ -3,17 +3,3 @@
 ---
 
 visualstudio.com domains are now supported along with dev.azure.com
-
-```diff
-+ const isCloud = (host: string) => {
-+   if (host === 'dev.azure.com') {
-+     return true;
-+   }
-+
-+   if (host.endsWith('.visualstudio.com')) {
-+     return true;
-+   }
-+
-+   return false;
-+ };
-```
--- a/.changeset/wise-teeth-study.md
+++ b/.changeset/wise-teeth-study.md
@@ -1,0 +1,10 @@
+---
+'@backstage/plugin-catalog-backend-module-azure': minor
+---
+
+isCloud function now also checks if hostname endswith visualstudio.com along with dev.azure.com
+
+```diff
+- const isCloud = (host: string) => host === 'dev.azure.com';
++ const isCloud = (host: string) => host === 'dev.azure.com' || host.endsWith('visualstudio.com');
+```
--- a/.changeset/wise-teeth-study.md
+++ b/.changeset/wise-teeth-study.md
@@ -2,9 +2,19 @@
 '@backstage/plugin-catalog-backend-module-azure': minor
 ---
 
-isCloud function now also checks if hostname endswith visualstudio.com along with dev.azure.com
-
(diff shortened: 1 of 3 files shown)


awanlin:
HI @fcamgz, I'll give this a closer look in a few hours but for next time it would be really helpful to keep the checklist in place in the description as it helps both you and I make sure things are in a good state. Like the initially missing changeset and the still missing signed DCO. 👍

.changeset/wise-teeth-study.md
```diff
@@ -0,0 +1,20 @@
+---
+'@backstage/plugin-catalog-backend-module-azure': minor
```
awanlin, line 2:
```suggestion
'@backstage/plugin-catalog-backend-module-azure': patch
```

As the existing functionality still work this can be set as just a `patch`

.changeset/wise-teeth-study.md
```diff
@@ -0,0 +1,20 @@
...
++     return true;
++   }
+
++   if (host === 'visualstudio.com' || host.endsWith('.visualstudio.com')) {
++     return true;
++   }
+
++   return false;
++ };
+```
```
awanlin, line 20: Please remove this, we can always see the diff here in the commit history. 👍

plugins/catalog-backend-module-azure/src/lib/azure.ts
```diff
@@ -46,7 +46,18 @@ interface CodeSearchRequest {
   };
 }
 
-const isCloud = (host: string) => host === 'dev.azure.com';
+const isCloud = (host: string) => {
+  if (host === 'dev.azure.com') {
+    return true;
+  }
+
+  if (host === 'visualstudio.com' || host.endsWith('.visualstudio.com')) {
```
awanlin, line 54: Would this `host === 'visualstudio.com'` ever be `true`? My understanding it's always `{org}.visualstudio.com`. If that's right we should remove the first part of this check.

.changeset/wise-teeth-study.md
```diff
@@ -0,0 +1,20 @@
+---
+'@backstage/plugin-catalog-backend-module-azure': minor
+---
+
+The `isCloud` function now also checks if hostname is visualstudio.com or ends with .visualstudio.com along with dev.azure.com
```
awanlin, line 5:
```suggestion
```suggestion
visualstudio.com domains are now supported  along with dev.azure.com
```

The changeset will get rolled up in the the changelog, we generally like to keep them short and to the point. 👍 
```

fcamgz:
Thanks for the heads up @awanlin! Apologies for the missing information, I will make sure to add/update them correctly next time upon creating a PR

awanlin (changes requested):
Thanks for your contribution @fcamgz, left a few small comments.

I'm not sure this will fix all the issues noted in #29938, this would address the entity provider aspect but not the issue when running a template. I would need more details to suggest a fix for that as the stack trace in the error isn't pointing me in a direction to suggest.

freben (approved):
Let's go! The comments seem to have been addressed.

PR #29986, opened 2025-05-16
Canon - Improve nested menu + combobox

cdedreuille:

I'm improving the look and feel of UI component to align with our density updates. This change also include an addition to the `Menu` component -> `Combobox` for nested menus.

### DIFF
--- a/packages/canon/src/components/Menu/Combobox.tsx
+++ b/packages/canon/src/components/Menu/Combobox.tsx
@@ -26,7 +26,7 @@ import {
   useEffect,
 } from 'react';
 import clsx from 'clsx';
-import { ComboboxOption, ComboboxProps } from './types';
+import { MenuComboboxOption, MenuComboboxProps } from './types';
 import { Icon } from '@backstage/canon';
 
 const getListboxItemId = (listboxId: string, optionValue: string): string =>
@@ -42,7 +42,7 @@ function ComboboxItem({
   onItemSelect,
   listboxId,
 }: {
-  option: ComboboxOption;
+  option: MenuComboboxOption;
   optionIndex: number;
   value?: string[];
   activeOptionIndex: number;
@@ -85,7 +85,7 @@ function ComboboxItem({
 }
 
 /** @public */
-export const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
+export const Combobox = forwardRef<HTMLDivElement, MenuComboboxProps>(
   (props, ref) => {
     const {
       options,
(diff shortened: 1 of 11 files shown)


packages/canon/src/components/Menu/Combobox.tsx
```diff
@@ -0,0 +1,245 @@
...
+    const handleSearchChange = useCallback(
+      (e: ChangeEvent<HTMLInputElement>) => {
+        setFilterString(e.target.value);
+        setActiveOptionIndex(0);
+        e.preventDefault();
+      },
+      [],
+    );
+
+    const handleKeyDown = useCallback(
```
Rugvip, line 146: ✨

packages/canon/src/components/Menu/Combobox.tsx
```diff
@@ -0,0 +1,245 @@
...
+      <div className="canon-SubmenuComboboxItemCheckbox">
+        {isSelected && <Icon aria-hidden="true" name="check" size={12} />}
+      </div>
+      <div className="canon-SubmenuComboboxItemLabel">{option.label}</div>
+    </div>
+  );
+}
+
+/** @public */
+export const Combobox = forwardRef<HTMLDivElement, MenuComboboxProps>(
```
Rugvip, line 88: Do you have any indication on whether a combobox is on the way from Base UI? I imagine we'll need to move over to their API at some point.

This autocomplete dropdown territory is also quite complex. It tends to end up in the need to be able to load in paginated data from the server with typeahead search and other fun bits.

packages/canon/src/components/Menu/types.ts
```diff
@@ -13,7 +13,29 @@
...
+  disabled?: boolean;
+};
+
+/** @public */
+export interface MenuComboboxProps extends ComponentProps<'div'> {
+  options: MenuComboboxOption[];
+  value?: string[];
+  onValueChange?: (value: string[]) => void;
+  multiselect?: boolean;
+  closeParentOnEsc?: boolean;
```
Rugvip, line 37: I don't see this one used anywhere? Same thing with the existing definition of it actually.

Rugvip (approved):

packages/canon/src/components/Menu/Combobox.tsx
cdedreuille: Apparently it's coming in August.

packages/canon/src/components/Menu/types.ts
cdedreuille: These are used in the `Combobox.tsx` file.

Rugvip: Not that I can see O_o

PR #29989, opened 2025-05-16
Canon - Improve the way we treat custom render on Text and Heading

cdedreuille:

We are improving the way we handle custom render using BaseUI useRender() function under the hood.

You will now be able to do something like this:
```tsx
<Text render={<span />}>Hello world</Text>
```

### DIFF
--- a/packages/canon/src/components/Heading/Heading.tsx
+++ b/packages/canon/src/components/Heading/Heading.tsx
@@ -14,46 +14,41 @@
  * limitations under the License.
  */
 
-import { forwardRef } from 'react';
+import { forwardRef, useRef } from 'react';
 import clsx from 'clsx';
 import { useResponsiveValue } from '../../hooks/useResponsiveValue';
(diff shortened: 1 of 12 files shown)


packages/canon/src/components/Heading/Heading.tsx
```diff
@@ -14,46 +14,41 @@
...
 export const Heading = forwardRef<HTMLHeadingElement, HeadingProps>(
   (props, ref) => {
     const {
-      children,
       variant = 'title1',
-      as = 'h1',
+      color = 'primary',
       truncate,
       className,
+      render = <h1 />,
```
Rugvip, line 31: Element version had slightly worse perf right? Thinking that for a default like this it's best to use the format with the least overhead. No biggie though

Rugvip (approved):
Very interesting API this one. It does feel more explicit than the old `asChild` though.

PR #29992, opened 2025-05-17
docs/backend-system: update index page to be an introduction

Rugvip:
I figured this page needs an update: https://backstage.io/docs/backend-system/

It's titled as an introduction but doesn't really provide that at the moment 😅


Parsifal-M (approved):
Makes sense and looks good! Don't _think_ I missed anything 👏

freben (approved):
Najs!

PR #29996, opened 2025-05-18
Canon - Add new Tabs component

cdedreuille:

CleanShot 2025-05-18 at 21 04 35@2x

### DIFF
--- a/packages/canon/src/components/Tabs/Tabs.stories.tsx
+++ b/packages/canon/src/components/Tabs/Tabs.stories.tsx
@@ -16,7 +16,6 @@
 
 import type { Meta, StoryObj } from '@storybook/react';
 import { Tabs } from './Tabs';
-import { Text } from '../../index';
 
 const meta = {
   title: 'Components/Tabs',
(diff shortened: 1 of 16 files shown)


packages/canon/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+    children: (
+      <Tabs.Root>
+        <Tabs.List>
+          <Tabs.Tab>Tab 1</Tabs.Tab>
+          <Tabs.Tab>Tab 2</Tabs.Tab>
+          <Tabs.Tab>Tab 3 With long title</Tabs.Tab>
+        </Tabs.List>
+        <Tabs.Panel style={panelStyles}>Content for Tab 1</Tabs.Panel>
+        <Tabs.Panel style={panelStyles}>Content for Tab 2</Tabs.Panel>
+        <Tabs.Panel style={panelStyles}>Content for Tab 3</Tabs.Panel>
```
Rugvip, line 44: Huh, so just the same number of panels as tabs in order? That's actually kinda nice and simple to some extent. I assume it's possible to be more explicit or avoid including all of these in the tree too. If it's possible to define these in any other way then I think it's be good to have an example of that, otherwise nvm 😁

Rugvip (approved):
Nice! 😁 🎉

packages/canon/src/components/Tabs/Tabs.stories.tsx
cdedreuille: Yes you can add the `value` prop to be more explicit but by default it takes it as an array which is indeed pretty nice.

PR #29998, opened 2025-05-19
Fix: 29961: Do not propogate SQL query to API response

abinavsridhar-wk:

This will address a CVE in search plugin that was exposing the entire SQL query in the response when the special character '<' is a part of the search query term.

Approach: 
* Catch the error and verify the error message and the query string
* Return an empty response with 200 status code


plugins/search-backend/src/service/router.ts
```diff
@@ -201,6 +201,17 @@ export async function createRouter(
         // re-throw and let the default error handler middleware captures it and serializes it with the right response code on the standard form
         throw error;
       }
+      if (query.term.indexOf('<') !== -1) {
+        // Search queries that contain '<' are likely to be a SQL injection attack, log the error and return an empty response
```
Rugvip, line 205: We're not trying to protect against SQL injection, only information leakage via error responses.

plugins/search-backend/src/service/router.ts
```diff
@@ -201,6 +201,17 @@ export async function createRouter(
...
+        const message = (error as any)?.message || '';
+        if (
+          typeof message === 'string' &&
+          message.indexOf('syntax error in tsquery') !== -1
```
Rugvip, line 209: This check needs to be a bit more explicit, ideally just checking `error.name` or something like that. Alternatively we just return a generic error response and log the actual error instead.

Rugvip (changes requested):

plugins/search-backend/src/service/router.ts
abinavsridhar-wk: error.name is just "Error" it is a generic name for this specific type of exception. I guess the path forward is to log the message instead of returning the error message in the response. 
I will probably return a generic error response and log the actual error instead.

plugins/search-backend/src/service/router.ts
abinavsridhar-wk: Updated the comment.

abinavsridhar-wk:
@Rugvip gentle reminder on this one

coreydaley:
@abinavsridhar-wk Looks like there are a couple of tests that are failing and need to be updated to look for the new error message syntax.

abinavsridhar-wk:
@coreydaley , @Rugvip , Thank you, I was under the impression that these were fleeting failures, I have fixed the failing tests. Can I get reviews on this please?

coreydaley:
@Rugvip It looks like the tests are passing and your change requests have been addressed. Can we get another look at this one please?

abinavsridhar-wk:
Pretty please

abinavsridhar-wk:
@Rugvip gentle reminder

abinavsridhar-wk:
@Rugvip gentle reminder again

freben (approved):

Issue #21534, opened 2023-11-24
🚀 Feature: Improve sidebar submenu UX 


Bogdan-Lyashenko:
### 🔖 Feature description

User keeps losing opened submenu while trying to move from hovered item to its subitems. Especially if they are far apart - like shown on the image below, the only way to select submenu item is to strictly follow red line to not accidentally close submenu.
![image](https://github.com/backstage/backstage/assets/2098309/65668632-d121-40c6-b5ba-e014c91ebe3e)

### 🎤 Context

Current behaviour with on hover-only makes it quite hard to use if menu item and its subitems are far apart, can be improved. 

### ✌️ Possible Implementation

Possible (alternative) solutions:
1) Add configuration to opt in for 'onClick' behaviour to open submenu (and not just on mouse over by default).
2) Align submenu vertically so it always next to the right of it's parent (and not sticking to the top of the page)
3) Make desktop submenu to behave same way as mobile submenu (open down relative to the parent item and not to the right)

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

camilaibs:
Hi @tudi2d , could you share your recommendations?

tudi2d:
Tja @Bogdan-Lyashenko - raising a good point! 

First the hit box of the sidebar items could be improved. I'm thinking of this code example since I saw it the first time - it makes so much sense: https://codepen.io/hakimel/pen/joQZbq.
<img width="347" alt="Screenshot 2023-11-27 at 13 58 12" src="https://github.com/backstage/backstage/assets/8904624/0ddfae9c-aad8-493b-abda-1478abb4f320">
This is not a solution to your specific problem, but might be a addition to the felt frustration when navigating.

Also I like your second suggestion. It could make sense to introduce a different type of the submenu component similar to WordPress, where the submenu is only showing at the specific item. 
<img width="282" alt="Screenshot 2023-11-27 at 14 48 58" src="https://github.com/backstage/backstage/assets/8904624/615c6987-b2e2-4aab-a37b-ee31454ec039">
[...]

PR #30002, opened 2025-05-19
feat: add sidebar popper submenu type

amhassen97:

This adds a `subMenuType` property to the `subMenuOptions` prop on the sidebar to choose a different submenu rendering layout. The types to choose from are `default` and `popper`. The `default` type keeps the original submenu layout, whilst the `popper` type allow the submenu to render as a Material UI Popper component, anchored near the menu item containing the submenu, for easier accessibility.

msedge_Y0VHfEkRHc

### DIFF
--- a/.changeset/nine-kiwis-study.md
+++ b/.changeset/nine-kiwis-study.md
@@ -1,5 +1,5 @@
 ---
-'@backstage/core-components': minor
+'@backstage/core-components': patch
 ---
 
 This adds a `subMenuType` property to the `subMenuOptions` prop on the sidebar to choose a different submenu rendering layout. The types to choose from are `default` and `popper`. The `default` type keeps the original submenu layout, whilst the `popper` type allow the submenu to render as a Material UI Popper component, anchored near the menu item containing the submenu, for easier accessibility.
--- a/.changeset/nine-kiwis-study.md
+++ b/.changeset/nine-kiwis-study.md
@@ -1,0 +1,5 @@
+---
+'@backstage/core-components': minor
+---
+
+This adds a `subMenuType` property to the `subMenuOptions` prop on the sidebar to choose a different submenu rendering layout. The types to choose from are `default` and `popper`. The `default` type keeps the original submenu layout, whilst the `popper` type allow the submenu to render as a Material UI Popper component, anchored near the menu item containing the submenu, for easier accessibility.
--- a/packages/core-components/src/layout/Sidebar/config.ts
+++ b/packages/core-components/src/layout/Sidebar/config.ts
@@ -30,7 +30,7 @@ export type SidebarOptions = {
 export type SubmenuOptions = {
   drawerWidthClosed?: number;
   drawerWidthOpen?: number;
-  subMenuType: 'default' | 'popper';
+  subMenuType?: 'default' | 'popper';
 };
 
 /** @internal */
--- a/packages/core-components/src/layout/Sidebar/config.ts
+++ b/packages/core-components/src/layout/Sidebar/config.ts
@@ -30,6 +30,7 @@ export type SidebarOptions = {
 export type SubmenuOptions = {
   drawerWidthClosed?: number;
   drawerWidthOpen?: number;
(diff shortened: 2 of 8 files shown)


.changeset/nine-kiwis-study.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/core-components': minor
+---
+
+This adds a `subMenuType` property to the `subMenuOptions` prop on the sidebar to choose a different submenu rendering layout. The types to choose from are `default` and `popper`. The `default` type keeps the original submenu layout, whilst the `popper` type allow the submenu to render as a Material UI Popper component, anchored near the menu item containing the submenu, for easier accessibility.
```
awanlin, line 5: If you add "submenu" to this file it will address these errors: https://github.com/backstage/backstage/blob/master/.github/vale/config/vocabularies/Backstage/accept.txt

awanlin (changes requested):
Hi @amhassen97, thanks for the contribution, didn't review this too deeply but wanted to give you a heads up on the errors on the changeset and also if you rebase that should resolve the E2E failures 👍

.changeset/nine-kiwis-study.md
amhassen97: Hi @awanlin, thanks for the feedback, I've updated the vocab file and rebased the PR 👍

packages/core-components/src/layout/Sidebar/config.ts
```diff
@@ -30,6 +30,7 @@ export type SidebarOptions = {
 export type SubmenuOptions = {
   drawerWidthClosed?: number;
   drawerWidthOpen?: number;
+  subMenuType: 'default' | 'popper';
```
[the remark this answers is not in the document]
amhassen97: @freben I've fixed the optional type issue in the latest commit 😄

.changeset/nine-kiwis-study.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/core-components': minor
```
awanlin, line 2:
```suggestion
'@backstage/core-components': patch
```

Now that this isn't a breaking change based on @freben comment being addressed 👍

awanlin (commented):
Just a quick comment about the changeset again, leaving a proper follow up review for others. 👍

amhassen97:
> Hm. Is this something you just want to do for a subset of entries?
> 
> I'm not sure how to best do this, it's almost like I'm wondering if the builtin sidebar should have support for a multitude of rendering styles, or if it's better to replace the sidebar with one's own implementation if one's needs are advanced enough 🤔

Hey @freben 

My intention was to try and address the UX issues mentioned in https://github.com/backstage/backstage/issues/21534, but also provide the option to use/make it part of the default sidebar to keep in line with the core Backstage components and implementation. 

The current UX is rather frustrating - especially if you have a long list of items on the main sidebar - if an item is closer to the bottom and it has submenu items, a user has to hover and then move their mouse from the bottom of the viewport/sidebar right to the top of the viewport again to access the submenu items. Also the hover/hit area of the submenu is pretty small - so any movement not in line with the exact item to access the submenu items results in the submenu disappearing - which in the cases of sidebar items that are further down, containing submenu items, happens quite often 😅.

By using a popper, this offers something of a middle ground by placing the submenu in-line with the sidebar item a user is trying to access - making it less likely for it to disappear for items that are further down, and making the mouse navigation experience for users a lot smoother. This was one of the suggestions @tudi2d made in the https://github.com/backstage/backstage/issues/21534 thread, and how I decided on making use of the Popper itself.

amhassen97:
reopen

amhassen97:
up

freben:
Reopened for consideration but I'm still hesitant if this should be a feature in this particular way.

amhassen97:
Thanks @freben 

Do you have a suggestion or something in mind for an alternative solution with respect to improving the sidebar UX? I'd be happy to contribute if you have an alternative way in mind

PR #30028, opened 2025-05-21
feat(plugin/notifications) add generic attribute field to NotificationPayload

FrankySpot:

     
Add a JSON blob field to `NotificationPayload` that will be used to store additional metadata for notifications.

This field will be available to processors when a notification is created.

### Why

Adds a convenient and simple metadata field that can be used by processors.

Some potential use cases for this:
1. Additional formatting options for different processors, i.e.  slack mrkdwn.
2. Overrides for processors, i.e. email/slack channel overrides.


plugins/notifications-common/report.api.md
```diff
@@ -24,27 +24,32 @@ export type NewNotificationSignal = {
...
+  T extends Record<string, unknown> = Record<string, unknown>,
+> = {
   title: string;
   description?: string;
   link?: string;
   severity?: NotificationSeverity;
   topic?: string;
   scope?: string;
   icon?: string;
+  attributes?: T;
```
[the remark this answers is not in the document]
FrankySpot: That makes sense, tbh i was deep into OTEL stuff so I ended up just using attributes.

FrankySpot:
@Rugvip rebased and updated

Rugvip (commented):
As discussed I think it'd be best to add some docs to `docs/notifications` that explains how to both produce and consume the metadata, just to make sure we have a bit of thinking around that, and ofc that it's documented x)

docs/notifications/usage.md
```diff
@@ -210,3 +210,39 @@ notificationsApi.getNotification(yourId);
...
+When sending a notification:
+
+```ts
+notificationService.send({
+  recipients: { type: 'entity', entityRef: 'group/default:team-a' },
+  payload: {
+    title: 'Notification',
+    description: 'Description'
+    metadata: {
+      markdown: `
```
Rugvip: I'm thinking we'd likely want to make the metadata keys a bit more specific, something like `body.markdown` or even more specific, e.g. like catalog annotations: `backstage.io/body.markdown`.

I do think it's important that we add some form of registry for well-known metadata fields, starting with the markdown one in this example. Again, similar to catalog annotations: https://backstage.io/docs/features/software-catalog/well-known-annotations

plugins/notifications-common/report.api.md
```diff
@@ -52,6 +53,9 @@ export type NotificationPayload = {
   topic?: string;
   scope?: string;
   icon?: string;
+  metadata?: {
+    [KMetadataKey in string]?: JsonValue;
```
Rugvip, line 57: Related to the registry discussion above, `JsonValue` here is a tricky one. I do think it's what it should be, since only allowing strings is very limiting and leads to ugly workarounds. It does require some careful parsing though, and/or schemas. It could be that we for example want a JSON Schema associated with every well-known metadata key or something like that? DX there is somewhat important too with the best care being that there are TypeScript types as well.

Rugvip (commented):
Couple of thoughts for moving this forward. I think the implementation itself is good as is, and it's more in the management of metadata keys and schemas that I'm thinking we need a bit more standardization to avoid things getting too messy.

docs/notifications/usage.md
FrankySpot: Added to docs

plugins/notifications-common/report.api.md
FrankySpot: As discussed, we will offload the schema validation to the processors for now. In the example in the docs I added Zod schema parsing.

FrankySpot:
> This ended up ridiculously slim in the end 😁

Most of that initial logic was adding things into the DB for persistence, but we're not doing that now so it simplified things.

All the complexity from this will likely end up in the processor code.

plugins/notifications-common/report.api.md
```diff
@@ -52,6 +53,9 @@ export type NotificationPayload = {
   topic?: string;
   scope?: string;
   icon?: string;
+  metadata?: {
```
drodil, line 56: In the 0001-notifications BEP this was `metadata: { type: string; value: obj }[]` but I like this plain object way better as it's more customizable for the future.

Dare to update the `NotificationPayload.metadata` to the BEP as well?

drodil:
On the second look, it seems the metadata is not persisted at all 🤔 Should it? For example in case some external system fetches notifications or adopter wants to render them in some custom way? 

If the metadata is not persisted it should at least be mentioned in the docs and maybe we can add the peristance in another round if needed.

FrankySpot:
> On the second look, it seems the metadata is not persisted at all 🤔 Should it? For example in case some external system fetches notifications or adopter wants to render them in some custom way?

it was originally persisted but after speaking with @Rugvip we decided in the first round to make it processor only since we need to flesh out a bit how this would work for front-end systems, so I reverted the DB changes (realized that processor do not need metadata to be persisted to use them).

> Dare to update the NotificationPayload.metadata to the BEP as well?

Can do.

FrankySpot:
@drodil  updated the BEP typing

beps/0001-notifications-system/README.md
```diff
@@ -211,10 +211,7 @@ export type NotificationPayload = {
   topic?: string;
   scope?: string;
   icon?: string;
-  metadata?: Array<{
-    type: string;
-    value: JsonValue;
-  }>;
+  metadata?: { [KMetadataKey in string]?: JsonValue };
```
Rugvip, line 214: Ah alright, we gonna say we shouldn't update the BEP at this point, but this is a clear immediate fix

drodil (approved):

PR #30036, opened 2025-05-22
Patch 2

valentingregoire:

Updated the code examples for the Policies docs.

### DIFF
--- a/docs/permissions/frontend-integration.md
+++ b/docs/permissions/frontend-integration.md
@@ -13,16 +13,11 @@ However, there are some cases where the integrator needs to supplement the polic
 If your Backstage permission policy may return a `DENY` for users requesting the `catalogEntityCreatePermission`, it may make sense, for example, to remove access to the `/catalog-import` page entirely:
 
 ```tsx title="packages/app/src/App.tsx"
-/* highlight-add-start */
 import { RequirePermission } from '@backstage/plugin-permission-react';
-import { catalogEntityCreatePermission } from '@backstage/plugin-catalog-common';
-/* highlight-add-end */
+import { catalogEntityCreatePermission } from '@backstage/plugin-catalog-common/alpha';
 
 const routes = (
   <FlatRoutes>
-    {/* highlight-remove-next-line */}
-    <Route path="/catalog-import" element={<CatalogImportPage />} />
-    {/* highlight-add-start */}
     <Route
       path="/catalog-import"
       element={
@@ -31,7 +26,6 @@ const routes = (
         </RequirePermission>
       }
     />
-    {/* highlight-add-end */}
     {/* ... */}
   </FlatRoutes>
 );
--- a/docs/permissions/writing-a-policy.md
+++ b/docs/permissions/writing-a-policy.md
@@ -92,6 +92,13 @@ import {
   catalogEntityDeletePermission,
 } from '@backstage/plugin-catalog-common/alpha';
 /* highlight-add-end */
+import {
+  PermissionPolicy,
+  PolicyQuery,
+  /* highlight-add-next-line */
+  PolicyQueryUser,
+} from '@backstage/plugin-permission-node';
+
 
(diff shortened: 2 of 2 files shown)


docs/permissions/frontend-integration.md
```diff
@@ -13,16 +13,11 @@ However, there are some cases where the integrator needs to supplement the polic
 If your Backstage permission policy may return a `DENY` for users requesting the `catalogEntityCreatePermission`, it may make sense, for example, to remove access to the `/catalog-import` page entirely:
 
 ```tsx title="packages/app/src/App.tsx"
-/* highlight-add-start */
 import { RequirePermission } from '@backstage/plugin-permission-react';
-import { catalogEntityCreatePermission } from '@backstage/plugin-catalog-common';
-/* highlight-add-end */
+import { catalogEntityCreatePermission } from '@backstage/plugin-catalog-common/alpha';
```
awanlin, line 17: I'm no entirely sure I understand this change. The way it is now it shows the before and after of the change which can be helpful. Is there something technical I'm missing here?

valentingregoire: Yes, but the current setup (meaning, starting from a brand-new Backstage installation) will provide the exact code that is shown in the example (and some more stuff).

awanlin: Ah, I stand corrected, thanks for pointing that out! I think we need to either keep it as is with the `/alpha` correction or find a better example. I think moving forward with this as you have it when you read the text before the example it doesn't make much sense to me. Going to ask the other Docs Maintainers for their thoughts. 🤔

valentingregoire: Just to clarify, only the frontend integration example is the same as the default code. The other changes do apply correctly.

Parsifal-M: Heya! 👋 

Apologies if I am misunderstanding, but there is two ways to look at this document I suppose, from the point of "I've just spun up a new backstage app with the create-app" and the I've had backstage for a while but never used permissions.

Would it not be better to instead just highlight somewhere that if you have spun up a new backstage app you don't/won't need to make these changes?

Am I on the same page here? 😅  WDYT?

valentingregoire: That's also a good idea, then it's clear why there's no added (green) lines of code, and no removed (red) lines of code. I'll add that in.

docs/permissions/frontend-integration.md
```diff
@@ -4,6 +4,10 @@ title: Frontend Integration
 description: How to place your Backstage frontend components behind authorization
 ---
 
+:::info
+Starting with a [fresh Backstage installation](https://backstage.io/docs/getting-started/), the code example below is already implemented, so there are no code actual code changes required.
```
freben:
```suggestion
Starting with a [fresh Backstage installation](https://backstage.io/docs/getting-started/), the code example below is already implemented, so there are no actual code changes required.
```

freben (approved):
Alright. Before we merge this, could you try rebasing it on latest master to be sure that the microsite build steps actually go green?

valentingregoire:
> Alright. Before we merge this, could you try rebasing it on latest master to be sure that the microsite build steps actually go green?

@freben, I wasn't aware that this was needed, I was already wondering what was wrong. The Merge is still blocked because of the change requests. It's unclear why, but I assume just a rereview.

docs/permissions/frontend-integration.md
```diff
@@ -31,7 +30,6 @@ const routes = (
         </RequirePermission>
```
aramissennyeydd, line 31: Could we add a highlight here to indicate that this is the relevant permission section? Small thing but with the changes this code block would have no highlighting which is likely confusing.

aramissennyeydd (approved):

awanlin (approved):
🚢

PR #30038, opened 2025-05-23
OpenAPI Basic Auth

mpsanchis:

     
 Consider security of OpenAPI specs and generate client's parameters (username+password instead of token) and headers (Basic instead of Bearer) if basicAuth.
 
 ⚠️  Note: other cases not considered. The attempt is to only include BasicAuth in this PR, but this probably needs more attention to tackle other security specs. So far, it was defaulting to Bearer + Token in all cases.

- [❓] Added or updated documentation  


sxmair:
Thanks @mpsanchis . Adding additional context here to clarify the scope and implementation of this PR:

**What this PR fixes**
Previously, the generated clients always included:

- A `token` field in the RequestOptions interface
- An unconditional `Authorization` header injected as `Bearer ${token}`

This behavior occurred regardless of the actual `http` auth scheme defined in the OpenAPI spec, including cases where `basic` authentication was intended. As a result, even when `basicAuth` was declared, the generated clients expected a token and injected a Bearer token in the header.


------------------------

**What this PR changes**
This PR updates the Mustache template to explicitly support `basicAuth` by:

- Conditionally rendering `username` and `password` in `RequestOptions` only when `isBasicBasic` is true
- Falls back to rendering `token` as before, when `isBasicBasic` is false
- Generating the appropriate Authorization header based on the detected auth scheme:
  - `Authorization: Basic <base64-encoded-credentials>` when using `BasicAuth`
  - `Authorization: Bearer <token>` as the default fallback for other schemes

----------------------

**Why `authMethods` is now required**
In the previous implementation, the Authorization header was always included using a fixed logic block like:

```
...(options?.token && { Authorization: `Bearer ${options?.token}` }),
```
This was applied regardless of whether the declared `http` auth scheme was `basic` or `bearer`, the template always defaulted to injecting a `Bearer` token header.


By introducing the `{{#authMethods}}` block:

- The template now avoids incorrect generation of the `RequestOptions` interface. For example, without `authMethods`, `basicAuth` would still render a `token?: string` field, even though `username` and `password` are the intended inputs
- It enables proper iteration over defined auth schemes, which is necessary because Mustache only evaluates nested conditionals `{{#isBasicBasic}}` and `{{^isBasicBasic}}` if `authMethods `is present. Without `authMethods`, those inner conditionals never evaluate, and the Authorization header block is never generated.

This change ensures that when a `basic` scheme is explicitly declared in the OpenAPI `securitySchemes`, the generated client correctly expects `username` and `password` in `RequestOptions`, and injects the appropriate `Authorization: Basic <credentials>` header. For all other schemes, it falls back to the previous template behavior of using a `token` and injecting a `Bearer` token header, aligning the output with what the OpenAPI spec defines.

PR #30043, opened 2025-05-23
Update the URL when the "Processing Status" filters change

### LINKED ISSUE #29999: "Processing Status" filters are not shareable by URL
### 📜 Issue Labels

### 🔎 Search Terms

```plain
"processing status", "orphan"
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

N/A

### 📝 Description & Context

On the catalog page, all default filters in the side panel are added to/updated in the URL except for the ones concerning "Processing Status".
Therefore, a view to the specific entities that are orphaned or have errors cannot be shared via URL.

### 👍 Expected Behavior

"Processing Status" should be treated like the other catalog filters.

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

1. Go to the [demo deployment](https://demo.backstage.io/catalog) of Backstage
2. Go to the Component view of the UI (the above link should already bring you here)
3. Use the "Processing Status" filter "Is Orphan"
4. Notice how the URL was not updated when applying this specific filter
5. Notice how applying any other filter category updates the URL with the filter condition

letrunghieu:

Close #29999.

Update the URL when the "Processing Status" filters change.

### DIFF
--- a/plugins/catalog-react/src/components/EntityProcessingStatusPicker/EntityProcessingStatusPicker.tsx
+++ b/plugins/catalog-react/src/components/EntityProcessingStatusPicker/EntityProcessingStatusPicker.tsx
@@ -21,11 +21,12 @@ import FormControlLabel from '@material-ui/core/FormControlLabel';
 import { makeStyles } from '@material-ui/core/styles';
 import CheckBoxIcon from '@material-ui/icons/CheckBox';
 import CheckBoxOutlineBlankIcon from '@material-ui/icons/CheckBoxOutlineBlank';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { useEntityList } from '../../hooks';
 import { catalogReactTranslationRef } from '../../translation';
 import { useTranslationRef } from '@backstage/core-plugin-api/alpha';
 import { CatalogAutocomplete } from '../CatalogAutocomplete';
+import { EntityFilter } from '../../types';
 
 /** @public */
 export type CatalogReactEntityProcessingStatusPickerClassKey = 'input';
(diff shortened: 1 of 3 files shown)


benjdlambert:
Hey @letrunghieu :wave:

Thanks for this PR! Can we get some accompanying tests here to avoid regressions in the future? :pray:

letrunghieu:
> Hey @letrunghieu 👋
> 
> Thanks for this PR! Can we get some accompanying tests here to avoid regressions in the future? 🙏

Yes, sure. Let me add some tests to the PR.

letrunghieu:
I've added new test cases to cover the bug in this PR @benjdlambert

plugins/catalog-react/src/components/EntityProcessingStatusPicker/EntityProcessingStatusPicker.test.tsx
```diff
@@ -109,4 +109,23 @@ describe('<EntityProcessingStatusPicker/>', () => {
...
+          queryParameters,
+        }}
+      >
+        <EntityProcessingStatusPicker />
+      </MockEntityListContextProvider>,
+    );
+
+    expect(screen.getByText('Processing Status')).toBeInTheDocument();
+    expect(screen.getByText('Is Orphan')).toBeInTheDocument();
+    expect(screen.getByText('Has Error')).toBeInTheDocument();
```
vinzscam, line 129: mmm these two strings are always rendered even when the filters are not selected

plugins/catalog-react/src/components/EntityProcessingStatusPicker/EntityProcessingStatusPicker.tsx
```diff
@@ -42,29 +43,50 @@ const useStyles = makeStyles(
...
+    updateFilters(
+      Object.fromEntries(
+        Object.keys(advanceFilterOptions).map(key => [
+          key,
+          selectedAdvancedItems.includes(advanceFilterOptions[key].label)
+            ? advanceFilterOptions[key].createFilterFn()
+            : undefined,
+        ]),
+      ),
+    );
```
vinzscam: any reason why you are removing the filter when the value is not selected? I would to keep it, otherwise manually toggling `orphan=false` won't work

vinzscam (changes requested):
thank you 🙏

plugins/catalog-react/src/components/EntityProcessingStatusPicker/EntityProcessingStatusPicker.tsx
```diff
@@ -42,29 +43,50 @@ const useStyles = makeStyles(
...
   const classes = useStyles();
-  const { updateFilters } = useEntityList();
+  const { updateFilters, queryParameters } = useEntityList();
   const { t } = useTranslationRef(catalogReactTranslationRef);
 
   const [selectedAdvancedItems, setSelectedAdvancedItems] = useState<string[]>(
-    [],
+    Object.keys(queryParameters)
+      .filter(key => key in advanceFilterOptions)
+      .map(key => advanceFilterOptions[key].label),
```
vinzscam: this is a bit ambiguous, I guess we should check the value of the filters instead?

plugins/catalog-react/src/components/EntityProcessingStatusPicker/EntityProcessingStatusPicker.test.tsx
letrunghieu: These two strings are only rendered after we clicked the `processing-status-picker-expand` button. In this test, we don't click on the expand button yet. The test will fails without the changes in this PR.

We can comment the line of `fireEvent.click(screen.getByTestId('processing-status-picker-expand'));` in the `renders all processing status options` test case the that test will fails because these two string are not rendered before we clicking the expand button.

plugins/catalog-react/src/components/EntityProcessingStatusPicker/EntityProcessingStatusPicker.tsx
letrunghieu: Nice catch! I have updated the filter here and added the test case for this.

plugins/catalog-react/src/components/EntityProcessingStatusPicker/EntityProcessingStatusPicker.tsx
letrunghieu: The current UI only has two states: checked (the label is visible) and unchecked (the label is removed). I don't think we can filter entities with `orphan = false` with the current implementation 🤔 

We can manually set the `filters[orphan]=false` in the query string, but @vinzscam, I don't see the current UI supporting rendering that case.

letrunghieu:
Hi @vinzscam , could you take a look at this PR again, please?

PR #30070, opened 2025-05-27
scaffolder-backend: :broom: part 2

benjdlambert:
Cleans up the old backend system exports.


plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -170,11 +128,6 @@ export interface RouterOptions {
   catalogClient: CatalogApi;
   scheduler?: SchedulerService;
   actions?: TemplateAction<any, any, any>[];
-  /**
-   * @deprecated taskWorkers is deprecated in favor of concurrentTasksLimit option with a single TaskWorker
-   * @defaultValue 1
-   */
-  taskWorkers?: number;
```
freben, line 177: Is it reasonable to have a config setting for this? Feels fairly reasonable to have more than one per host, for most types of actions etc, if they are mostly I/O bound.

freben (approved):

Issue #30021, opened 2025-05-21
Auditor error logging incompatible with Winston + Elasticsearch Error parsing


KaiReichart:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
error, winston, auditor
```

### 🗃️ Project Area

Auditor

### 🔗 External Integration

Other

### 📝 Description & Context

The auditor error logging [creates logs the following format](https://github.com/backstage/backstage/blob/master/packages/backend-defaults/src/entrypoints/auditor/DefaultAuditorService.ts#L198):

```
        await this.log({
          ...options,
          ...params,
          error: String(params.error),
          meta: { ...options.meta, ...params?.meta },
          status: 'failed',
        });
```

```
{
   [...]
    error: "NotFoundError ..."
   [...]
}
```

```
[backend]: {"@timestamp":"2025-05-21T12:28:05.009Z","actor":{"actorId":"plugin:statuses","hostname":"localhost","ip":"::1","userAgent":"node-fetch/1.0 (+https://github.com/bitinn/node-fetch)"},"ecs.version":"8.10.0","error":"NotFoundError: No entity named 'REDACTEX' found, with kind 'component' in namespace 'default'","eventId":"entity-fetch","isAuditEvent":true,"log.level":"debug","message":"catalog.entity-fetch","meta":{"entityRef":"component:default/REDACTEDo","queryType":"by-name"},"plugin":"catalog","request":{"method":"GET","url":"/api/catalog/entities/by-name/component/default/REDACTED"},"severityLevel":"low","span_id":"53836d3cc2227b4f","status":"failed","trace_flags":"01","trace_id":"81f27bf03e229313f94f348d28b2d3df"}
```

This is incompatible with the error parsing logic of ElasticSearch, which expects the value behind the `error` key to be an object. This is leading to parsing errors for these log lines in ElasticSearch

```
      "{\"type\":\"document_parsing_exception\",\"reason\":\"[1:958] object mapping for [error] tried to parse field [error] as object, but found a concrete value\"}"
```

### 👍 Expected Behavior

The auditor should differentiate between error, debug and info log messages, and use the appropriate methods of the logger. Then, the logged error messages should follow the configured log format of the root logger.

### Have you read the Code of Conduct?

[...]

PR #30072, opened 2025-05-27
feat: propagate errors as objects and align Winston auditor service

schultzp2020:

fixes https://github.com/backstage/backstage/issues/30021

Enhances error handling in the auditor service factory to pass errors as objects. 
Aligns `WinstonRootAuditorService` with the default service factory's error handling.

### DIFF
--- a/packages/backend-defaults/report-auditor.api.md
+++ b/packages/backend-defaults/report-auditor.api.md
@@ -8,6 +8,7 @@ import type { AuditorServiceCreateEventOptions } from '@backstage/backend-plugin
 import type { AuditorServiceEvent } from '@backstage/backend-plugin-api';
 import type { AuditorServiceEventSeverityLevel } from '@backstage/backend-plugin-api';
 import type { AuthService } from '@backstage/backend-plugin-api';
+import { Config } from '@backstage/config';
 import type { Format } from 'logform';
 import type { HttpAuthService } from '@backstage/backend-plugin-api';
 import type { JsonObject } from '@backstage/types';
@@ -58,7 +59,7 @@ export type AuditorEventStatus =
     }
   | {
       status: 'failed';
-      error: string;
+      error: Error;
     };
 
 // @public
@@ -95,6 +96,7 @@ export class WinstonRootAuditorService {
   // (undocumented)
   forPlugin(deps: {
     auth: AuthService;
+    config: Config;
     httpAuth: HttpAuthService;
     plugin: PluginMetadataService;
   }): AuditorService;
--- a/packages/backend-defaults/src/entrypoints/auditor/WinstonRootAuditorService.test.ts
+++ b/packages/backend-defaults/src/entrypoints/auditor/WinstonRootAuditorService.test.ts
@@ -32,6 +32,7 @@ describe('WinstonRootAuditorService', () => {
       plugin: {
         getId: () => 'test-plugin',
       },
+      config: mockServices.rootConfig.mock(),
     });
     expect(childLogger).toBeInstanceOf(DefaultAuditorService);
   });
@@ -45,6 +46,7 @@ describe('WinstonRootAuditorService', () => {
       plugin: {
         getId: () => pluginId,
       },
+      config: mockServices.rootConfig.mock(),
     });
     // workaround to spy on private method
     const auditorSpy = jest.spyOn(auditor as any, 'log');
@@ -68,6 +70,7 @@ describe('WinstonRootAuditorService', () => {
       plugin: {
         getId: () => pluginId,
       },
+      config: mockServices.rootConfig.mock(),
     });
     // workaround to spy on private method
     const auditorSpy = jest.spyOn(auditor as any, 'log');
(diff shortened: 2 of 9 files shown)


packages/backend-defaults/src/entrypoints/auditor/WinstonRootAuditorService.test.ts
```diff
@@ -32,6 +32,7 @@ describe('WinstonRootAuditorService', () => {
       plugin: {
         getId: () => 'test-plugin',
       },
+      config: mockServices.rootConfig.mock(),
```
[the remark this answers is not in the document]
fearphage: @freben The noise of this repetition is obscuring what's unique about each test. There's a reason that `setup`/`beforeEach` hooks exist. However for some reason all these tests have decided to inline the setup instead of using those hooks. I don't believe the added noise is benefiting the readers/maintainers.

I'm planning to submit the test clean up after this is merged if it doesn't land here.

schultzp2020:
@freben PTAL? 🙏

packages/backend-defaults/src/entrypoints/auditor/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Validates the `backend.auditor.severityLogLevelMappings` configuration.
+ *
+ * @param config - The root Backstage {@link @backstage/config#Config} object.
+ * @returns The validated severity-to-log-level mappings.
+ * @throws error - {@link @backstage/errors#InputError} if the mapping configuration is invalid.
+ *
+ * @public
+ */
+export function validateSeverityLogLevelMappings(config: Config) {
```
freben: this one could have been named `getX` since that's what it does (validation while doing so is implied); `validateX` functions often have a void return type. But no need to change it

packages/backend-defaults/src/entrypoints/auditor/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+import { CONFIG_ROOT_KEY, severityLogLevelMappingsSchema } from './types';
+
+/**
+ * Validates the `backend.auditor.severityLogLevelMappings` configuration.
+ *
+ * @param config - The root Backstage {@link @backstage/config#Config} object.
+ * @returns The validated severity-to-log-level mappings.
+ * @throws error - {@link @backstage/errors#InputError} if the mapping configuration is invalid.
+ *
+ * @public
```
freben:
```suggestion
```

right?

packages/backend-defaults/src/entrypoints/auditor/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+    critical:
+      auditorConfig?.getOptionalString('severityLogLevelMappings.critical') ??
+      'info',
+  } as Required<z.infer<typeof severityLogLevelMappingsSchema>>;
+
+  const res = severityLogLevelMappingsSchema.safeParse(
+    severityLogLevelMappings,
+  );
+  if (!res.success) {
+    const key = res.error.issues.at(0)?.path.at(0) as string;
```
freben, line 51: This rich error handling risks throwing new errors 🤔 especially with the unknown casts. There are also no tests for this. And if I read the code right, it just assumes that the error is of a certain class of error, while others might occur as well perhaps?

I think we often just show `res.error.toString()` which turns out to be human readable.

packages/backend-defaults/src/entrypoints/auditor/utils.ts
schultzp2020: Good point, I forgot to rename it when I added a return value. 😅

packages/backend-defaults/src/entrypoints/auditor/utils.ts
schultzp2020: Unfortunately the error is not clean when using `res.error.toString()`. The current state was requested in a previous PR to clean up the error message. 

```
"Failed to instantiate service 'core.auditor' for 'test' because the factory function threw an error, InputError: [
      {
        \"received\": \"invalidloglevel\",
        \"code\": \"invalid_enum_value\",
        \"options\": [
          \"debug\",
          \"info\",
          \"warn\",
          \"error\"
        ],
        \"path\": [
          \"low\"
        ],
        \"message\": \"Invalid enum value. Expected 'debug' | 'info' | 'warn' | 'error', received 'invalidloglevel'\"
      }
    ]"
```

schultzp2020:
@freben PTAL?

packages/backend-defaults/report-auditor.api.md
```diff
@@ -95,6 +96,7 @@ export class WinstonRootAuditorService {
   // (undocumented)
   forPlugin(deps: {
     auth: AuthService;
+    config: Config;
```
freben, line 99: One minor nit here that maybe you can fix in a follow-up: The rest of the deps here are plugin scoped (and services), while this is root scoped (and not a service, which it maybe could be). I'd suggest adding a `config: RootConfigService` to `WinstonRootAuditorServiceOptions` and removing it from here. Then there may also be opportunity to just compute the level mapping once in the constructor.

freben (approved):

PR #30076, opened 2025-05-28
feat: catalog table columns support i18n

mario-mui:


mario-mui:
> @mario-mui Can you please provide a screenshot for the areas affected by the i18n changes? 🙏

sure

plugins/catalog/report.api.md
```diff
@@ -154,35 +154,7 @@ export interface CatalogSearchResultListItemProps {
 }
 
 // @public
-export const CatalogTable: {
-  (props: CatalogTableProps): JSX_2.Element;
-  columns: Readonly<{
```
benjdlambert: We can't make this change just yet, it's a breaking change. 

If we want to go down the hooks route, it's most likely what we want to do is move the hook to `-react` package, and then deprecate this export. But with the upcoming changes to NFS, that could also have an affect on how these columns are structured and used.

I'm wondering if we should just skip this refactor for now, and revisit once we have a clearer understanding of what we need to do?

benjdlambert (changes requested):
I believe that we can solve this another way, even though I kind of like the hooks approach it's a breaking change that I think we want to avoid for now.

I think that you can provide a `() => ReactElement` as the `title`, so we'd want to create probably a little `HOC` that takes a translation key and does something like;

```tsx
const TitleColumn = (props: { translationKey }) => { 
  const { t } = useTranslation();
  return t(translationsKey)
});
...

createSystemColumn(): TableColumn<CatalogTableRow> {
    return {
      title: <TitleColumn translationKey="title" />
      field: 'resolved.partOfSystemRelationTitle',
      customFilterAndSearch: (query, row) => {
        if (!row.resolved.partOfSystemRelations) {
          return false;
        }

        const systemNames = row.resolved.partOfSystemRelations.map(
          ref => ref.name,
        ); // Extract system names from entityRefs

        const searchText = systemNames.join(', ').toLocaleUpperCase('en-US');
        return searchText.includes(query.toLocaleUpperCase('en-US'));
      },
      render: ({ resolved }) => (
        <EntityRefLinks
          entityRefs={resolved.partOfSystemRelations}
          defaultKind="system"
        />
      ),
    };
  },
...
```

Or something like that?

plugins/catalog/report.api.md
mario-mui: Ok, I will try some effort to stop this breaking change.

mario-mui:
@benjdlambert please have a look

mario-mui:
ping @benjdlambert please have a look

plugins/catalog-react/src/components/EntityTable/TitleColumn.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * @alpha
+ */
+export type EntityTableColumnTitleProps = {
+  translationKey: keyof typeof entityTableColumnTitleTranslationRef.T;
+};
+
+/**
+ * @alpha
+ */
+export const EntityTableColumnTitle = ({
```
schultzp2020, line 42: Do you think it would be beneficial to expose this in the `@backstage/frontend-plugin-api` package so we can do the same non breaking changes elsewhere?

benjdlambert: Hmm, maybe, but probably doesn't have to block this PR. Ideally it would be part of the translation API's that would let you pass in the `translationRef` and a `key` from that ref.

plugins/catalog-react/src/translation.ts
```diff
@@ -126,3 +126,22 @@ export const catalogReactTranslationRef = createTranslationRef({
     },
   },
 });
+
+/** @alpha */
+export const entityTableColumnTitleTranslationRef = createTranslationRef({
```
benjdlambert: Why is this a separate ref? Should it not just be added to the above messages?

mario-mui: ok

mario-mui:
@benjdlambert please have a look

plugins/catalog-react/src/translation.ts
```diff
@@ -16,6 +16,24 @@
 
 import { createTranslationRef } from '@backstage/core-plugin-api/alpha';
 
+/**
+ * @alpha
+ */
+export const entityTableColumnTitle = {
```
benjdlambert: I think that it's going to be better just to write these inline in the ref instead. Looking at the API reports it looses the type narrowing and they're just `string`'s which could cause issue for templating later on down the line.

That and I don't see so much value in having this declared separate for now.

mario-mui: ok, But I need to define `EntityTableColumnTitleProps.translationKey` to pass `t` function type check

mario-mui:
ping @benjdlambert

mario-mui:
@awanlin someone else can review this PR?

PR #30082, opened 2025-05-29
[Catalog] GitLab Org SAAS - Ingest Users from Sub Groups

UsainBloot:
## Current Status
The current behaviour of GitLab org ingestion is that will bring in all the users within the group specified, and also the sub groups of the specified parent group, as long as members from the specified group are also in the sub group.

This feels sub-optimal, and does not work for Saas setups where everyone belongs to a single group.

## Change
When using Saas GitLab (gitlab.com), and you specify a group in the config, then all the sub groups and their users will be ingested into the Software Catalog and modelled appropriately.

> [!NOTE]  
> This change can cause more users to be ingested into Backstage without altering pre-existing gitlab config.
> 
> You can use the existing `restrictUsersToGroup` option to turn off this behaviour.

Using the following setup
```yaml
catalog:
  providers:
    gitlab:
      default:
        host: gitlab.com
        orgEnabled: true
        group: group-name-here
        relations:
          - DIRECT # Default
        schedule:
          frequency: { seconds: 30 }
          timeout: { minutes: 3 }
        restrictUsersToGroup: false # pre-existing - defaults to false
```

### DIFF
--- a/.changeset/major-heads-scream.md
+++ b/.changeset/major-heads-scream.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-gitlab': minor
+---
+
+Changing default saas discovery to include members from subgroups of the root group
--- a/.changeset/major-heads-scream.md
+++ b/.changeset/major-heads-scream.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-backend-module-gitlab': minor
 ---
 
-Changing default SAAS discovery to include members from subgroups of the root group
+Breaking: User and Group discovery will default to ingesting all users in sub groups that belong to the specified root group in config. Disable setting config `restrictUsersToGroup: true`.
--- a/.changeset/major-heads-scream.md
+++ b/.changeset/major-heads-scream.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-backend-module-gitlab': minor
 ---
 
-Breaking: User and Group discovery will default to ingesting all users in sub groups that belong to the specified root group in config. Disable setting config `restrictUsersToGroup: true`.
+**BREAKING CHANGE**: User and Group discovery will default to ingesting all users in sub groups that belong to the specified root group in config. Disable by setting `restrictUsersToGroup: true` in app-config under your module settings.
--- a/.changeset/major-heads-scream.md
+++ b/.changeset/major-heads-scream.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-backend-module-gitlab': minor
 ---
 
-Changing default saas discovery to include members from subgroups of the root group
+Changing default SAAS discovery to include members from subgroups of the root group
--- a/plugins/catalog-backend-module-gitlab/src/providers/GitlabOrgDiscoveryEntityProvider.ts
+++ b/plugins/catalog-backend-module-gitlab/src/providers/GitlabOrgDiscoveryEntityProvider.ts
@@ -359,7 +359,7 @@ export class GitlabOrgDiscoveryEntityProvider implements EntityProvider {
     }
 
     let groups;
-    let users;
+    const allUsers = [];
 
     // Self-hosted: Fetch the users either from the defined group (restrictUsersToGroup) or fetch all users from the GitLab instance
     // SaaS: Fetch the users from the defined group (restrictUsersToGroup) or fetch all users from the root group.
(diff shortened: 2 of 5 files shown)


.changeset/major-heads-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-gitlab': minor
+---
+
+Changing default saas discovery to include members from subgroups of the root group
```
freben:
```suggestion
Changing default SAAS discovery to include members from subgroups of the root group
```

plugins/catalog-backend-module-gitlab/src/providers/GitlabOrgDiscoveryEntityProvider.ts
```diff
@@ -387,35 +389,46 @@ export class GitlabOrgDiscoveryEntityProvider implements EntityProvider {
...
+        : [rootGroupSplit[0], ...descendantGroups.map(g => g.id)];
+
+      // Fetch users group and descendant groups
+      for (const group of groupPaths) {
+        logger.debug(`Fetching users for group: ${group}`);
+        allUsers.push(
+          paginated<GitLabUser>(
+            options =>
+              this.gitLabClient.listSaaSUsers(
+                String(group),
```
freben: This cast looks odd - why is it needed?

plugins/catalog-backend-module-gitlab/src/providers/GitlabOrgDiscoveryEntityProvider.ts
```diff
@@ -430,16 +443,23 @@ export class GitlabOrgDiscoveryEntityProvider implements EntityProvider {
...
     };
 
-    for await (const user of users) {
-      userRes.scanned++;
+    for (const users of allUsers) {
+      // Iterate through paginated users
+      for await (const user of users) {
+        // Skip if user already processed and do not count it as scanned
+        if (user.id in idMappedUser) {
+          continue; // Skip if user already processed
```
freben, line 451: Yeah I was gonna say earlier up in the code, do we need to dedupe this actually? Should the deduping have happened more explicitly higher up?

freben (approved):
Preapproving but some questions.

This changes the default to include subgroups, is that counterproductive or surprising to people? Is it a common enough thing to want to do, or is that commonly the WRONG thing to do? I am honestly not sure. And should it be configurable?

UsainBloot:
> Preapproving but some questions.
> 
> This changes the default to include subgroups, is that counterproductive or surprising to people? Is it a common enough thing to want to do, or is that commonly the WRONG thing to do? I am honestly not sure. And should it be configurable?

I debated this with myself for quite some time.

I think the current behaviour is weird:

If you have the following setup, and have selected "Root Team" as the org in the config
```
  Root Team
   /      \
Sub A     Sub B
```

And members are setup as the following

```
Root Team: Jack
Sub A: Jack, Freben
Sub B: Freben
```

Then the catalog will only be ingested with the following
```
Users: Jack
Groups: Root Team, Sub A (only contains Jack)
```


I appreciate changing default behaviour of user ingestion is less than optimal. I thought about adding config to enable this, but with there is a pre-existing `restrictUsersToGroup`. If I added a new config option, it would be the opposite of `restrictUsersToGroup`, which would be strange. I've used that config value so that if it is enabled, the current behaviour will be used.

plugins/catalog-backend-module-gitlab/src/providers/GitlabOrgDiscoveryEntityProvider.ts
UsainBloot: Addressed in 27a00a5, using string literal instead when mapping the ids, as they are numbers

plugins/catalog-backend-module-gitlab/src/providers/GitlabOrgDiscoveryEntityProvider.ts
UsainBloot: I added the deduping here so that when that the log for `userRes.scanned` (`Scanned ${userRes.scanned} users and processed ${userEntities.length} users`) is more representative.

I found it a bit tricky to move above as `users` is a generator that is only resolved with the `for await (const user of users)`, this is the earliest chance I have to access the user id.

Happy to make changes though, but might need some guidance

.changeset/major-heads-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-gitlab': minor
+---
+
+Breaking: User and Group discovery will default to ingesting all users in sub groups that belong to the specified root group in config. Disable setting config `restrictUsersToGroup: true`.
```
freben:
```suggestion
**BREAKING CHANGE**: User and Group discovery will default to ingesting all users in sub groups that belong to the specified root group in config. Disable by setting `restrictUsersToGroup: true` in app-config under your module settings.
```

freben (approved):

PR #30092, opened 2025-05-30
validate plugin and module ids

Sarabadu:

With the introduction of `onPluginModuleBootFailure` cofiguration is expecrted that module and plugin ids are valid config keys

Here we are adding an early validation to make it clear instead of failing on the configuration read

### DIFF
--- a/.changeset/itchy-sites-cheer.md
+++ b/.changeset/itchy-sites-cheer.md
@@ -1,0 +1,8 @@
+---
+'@backstage/backend-plugin-api': minor
+'@backstage/backend-app-api': minor
+'@backstage/config-loader': patch
+'@backstage/config': patch
+---
+
+Now plugin and module Ids needs to be a valid config keys
--- a/docs/backend-system/architecture/08-naming-patterns.md
+++ b/docs/backend-system/architecture/08-naming-patterns.md
@@ -12,10 +12,10 @@ As a rule, all names should be camel case, with the exceptions of plugin and mod
 
 ### Plugins
 
-| Description | Pattern           | Examples                              |
-| ----------- | ----------------- | ------------------------------------- |
-| export      | `<camelId>Plugin` | `catalogPlugin`, `userSettingsPlugin` |
-| ID          | `'<kebab-id>'`    | `'catalog'`, `'user-settings'`        |
+| Description | Pattern           | Examples                              | Notes                                                                 |
+| ----------- | ----------------- | ------------------------------------- | --------------------------------------------------------------------- |
+| export      | `<camelId>Plugin` | `catalogPlugin`, `userSettingsPlugin` |                                                                       |
+| ID          | `'<kebab-id>'`    | `'catalog'`, `'user-settings'`        | letters, digits, dashes, and underscores only, starting with a letter |
 
 Example:
 
@@ -28,10 +28,10 @@ export const userSettingsPlugin = createBackendPlugin({
 
 ### Modules
 
-| Description | Pattern                      | Examples                            |
-| ----------- | ---------------------------- | ----------------------------------- |
-| export      | `<pluginId>Module<ModuleId>` | `catalogModuleGithubEntityProvider` |
-| ID          | `'<module-id>'`              | `'github-entity-provider'`          |
+| Description | Pattern                      | Examples                            | Notes                                                                 |
+| ----------- | ---------------------------- | ----------------------------------- | --------------------------------------------------------------------- |
+| export      | `<pluginId>Module<ModuleId>` | `catalogModuleGithubEntityProvider` |                                                                       |
+| ID          | `'<module-id>'`              | `'github-entity-provider'`          | letters, digits, dashes, and underscores only, starting with a letter |
 
 Example:
 
--- a/packages/backend-app-api/src/wiring/BackendInitializer.test.ts
+++ b/packages/backend-app-api/src/wiring/BackendInitializer.test.ts
@@ -774,7 +774,7 @@ describe('BackendInitializer', () => {
     const init = new BackendInitializer([]);
     init.add(
       createBackendPlugin({
-        pluginId: 'test-1',
+        pluginId: 'test-t1',
         register(reg) {
           reg.registerInit({
             deps: {},
@@ -787,7 +787,7 @@ describe('BackendInitializer', () => {
     );
     init.add(
       createBackendPlugin({
-        pluginId: 'test-2',
+        pluginId: 'test-t2',
         register(reg) {
           reg.registerInit({
             deps: {},
@@ -804,10 +804,10 @@ describe('BackendInitializer', () => {
     await expect(result).rejects.toMatchObject({
       errors: [
         expect.objectContaining({
-          message: "Plugin 'test-1' startup failed; caused by Error: NOPE A",
+          message: "Plugin 'test-t1' startup failed; caused by Error: NOPE A",
         }),
         expect.objectContaining({
-          message: "Plugin 'test-2' startup failed; caused by Error: NOPE B",
+          message: "Plugin 'test-t2' startup failed; caused by Error: NOPE B",
         }),
       ],
     });
@@ -1015,6 +1015,53 @@ describe('BackendInitializer', () => {
     );
     await expect(init.start()).rejects.toThrow(
       "Service or extension point dependencies of module 'test-mod' for plugin 'test' are missing for the following ref(s): serviceRef{a}",
+    );
+  });
+  it('should reject plugins with invalid pluginId', async () => {
+    const init = new BackendInitializer(baseFactories);
+    init.add(
+      createBackendPlugin({
+        pluginId: 'test:invalid&id',
+        register(reg) {
+          reg.registerInit({
+            deps: {},
+            async init() {},
+          });
+        },
+      }),
+    );
+    await expect(init.start()).rejects.toThrow(
+      "Invalid pluginId 'test:invalid&id', must match the pattern /^[a-z][a-z0-9]*(?:[-_][a-z][a-z0-9]*)*$/i (letters, digits, dashes, and underscores only, starting with a letter)",
+    );
+  });
+
+  it('should reject modules with invalid moduleId', async () => {
+    const init = new BackendInitializer(baseFactories);
+    init.add(
+      createBackendPlugin({
+        pluginId: 'test',
+        register(reg) {
+          reg.registerInit({
+            deps: {},
+            async init() {},
+          });
+        },
+      }),
+    );
+    init.add(
+      createBackendModule({
+        pluginId: 'test',
+        moduleId: 'invalid:module&id',
+        register(reg) {
+          reg.registerInit({
+            deps: {},
+            async init() {},
+          });
+        },
+      }),
+    );
+    await expect(init.start()).rejects.toThrow(
+      "Invalid moduleId 'invalid:module&id' for plugin 'test', must match the pattern /^[a-z][a-z0-9]*(?:[-_][a-z][a-z0-9]*)*$/i (letters, digits, dashes, and underscores only, starting with a letter)",
     );
   });
 
--- a/packages/backend-plugin-api/src/wiring/createBackendModule.ts
+++ b/packages/backend-plugin-api/src/wiring/createBackendModule.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { CONFIG_KEY_PART_PATTERN } from '@backstage/config';
 import { BackendFeature } from '../types';
 import {
   BackendModuleRegistrationPoints,
@@ -55,6 +56,12 @@ export function createBackendModule(
   options: CreateBackendModuleOptions,
 ): BackendFeature {
   function getRegistrations() {
+    if (!CONFIG_KEY_PART_PATTERN.test(options.moduleId)) {
+      throw new Error(
+        `Invalid moduleId '${options.moduleId}' for plugin '${options.pluginId}', must match the pattern ${CONFIG_KEY_PART_PATTERN} (letters, digits, dashes, and underscores only, starting with a letter)`,
+      );
+    }
+
     const extensionPoints: InternalBackendPluginRegistration['extensionPoints'] =
       [];
     let init: InternalBackendModuleRegistration['init'] | undefined = undefined;
--- a/packages/backend-plugin-api/src/wiring/createBackendPlugin.ts
+++ b/packages/backend-plugin-api/src/wiring/createBackendPlugin.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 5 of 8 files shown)


freben:
Hi! Remember `yarn build:api-reports` if they change

aramissennyeydd:
Linking #30196 as I ran into a similar issue, little bit different approach.

docs/backend-system/architecture/08-naming-patterns.md
```diff
@@ -12,10 +12,10 @@ As a rule, all names should be camel case, with the exceptions of plugin and mod
...
 ### Plugins
 
-| Description | Pattern           | Examples                              |
-| ----------- | ----------------- | ------------------------------------- |
-| export      | `<camelId>Plugin` | `catalogPlugin`, `userSettingsPlugin` |
-| ID          | `'<kebab-id>'`    | `'catalog'`, `'user-settings'`        |
+| Description | Pattern           | Examples                              | Notes                                                                 |
+| ----------- | ----------------- | ------------------------------------- | --------------------------------------------------------------------- |
+| export      | `<camelId>Plugin` | `catalogPlugin`, `userSettingsPlugin` |                                                                       |
+| ID          | `'<kebab-id>'`    | `'catalog'`, `'user-settings'`        | letters, digits, dashes, and underscores only, starting with a letter |
```
Parsifal-M: Is this intentional? would this be valid `foo_bar-baz`? or would that be `foo-bar-baz`? Just wondering if we need the `notes` column or just saying use kebab-case is enough? WDYT?

Parsifal-M (commented):
Just a thought! 🤔

docs/backend-system/architecture/08-naming-patterns.md
Sarabadu: I don't have strong opinions to be honest, but using the existing regex for the config keys as we are using the module id to read some configs, sounded like a good idea for me.

We could be more restrictive with the id but I think this is already a bit breaking change for the plugin ecosystem

Sarabadu:
Sry that I didn't come back to this earlier, I will try to come back to it next week. Thanks for the comments

.changeset/itchy-sites-cheer.md
```diff
@@ -0,0 +1,8 @@
+---
+'@backstage/backend-plugin-api': minor
+'@backstage/backend-app-api': minor
+'@backstage/config-loader': patch
+'@backstage/config': patch
+---
+
+Now plugin and module Ids needs to be a valid config keys
```
vinzscam: something like that maybe?

```suggestion
The backend will now throw an error if a plugin or a module doesn't have a valid ID
```

docs/backend-system/architecture/08-naming-patterns.md
Sarabadu: Im leaving it as it now, changing to kebab-case only seems more breaking. But happy to change it if we think we should be more restrictive

Sarabadu:
sry for leaving this hanging for a while...
adressed the comments, and changed the id [here](https://github.com/backstage/backstage/pull/30092/files#diff-a51ca908f0d6a33258dd202c1a9b3b97d09f45885552e72a699ed882ac3bf939R34) to complain with the regex too.

@vinzscam can you take another look?

Sarabadu:
Not stale

Sarabadu:
not stale, just waiting for reviews

vinzscam (approved):
thanks @Sarabadu 🙏

packages/backend-plugin-api/src/wiring/createBackendModule.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { CONFIG_KEY_PART_PATTERN } from '@backstage/config';
```
Rugvip: Let's just duplicate this one instead :p

They just happen to be similar, but I think this deserves it's own pattern. The ID pattern should definitely be a subset of the config key one, but might be even more strict.

docs/backend-system/architecture/08-naming-patterns.md
Rugvip: Yep let's not allow underscores

Sarabadu:
@Rugvip I though in the massive warning, but this is already falling if the module does not have a valid config keys...

So this is mostly to get a better error message instead of the error we get now at startup

packages/backend-plugin-api/src/wiring/createBackendModule.ts
```diff
@@ -55,6 +56,17 @@ export function createBackendModule(
...
+    if (!ID_PATTERN.test(options.moduleId)) {
+      console.warn(
+        `WARNING: The moduleId '${options.moduleId}' for plugin '${options.pluginId}', will be invalid soon please must match the pattern ${ID_PATTERN} (letters, digits, and dashes only, starting with a letter)`,
+      );
+    }
+    if (!ID_PATTERN_OLD.test(options.moduleId)) {
+      throw new Error(
+        `Invalid moduleId '${options.moduleId}' for plugin '${options.pluginId}', must match the pattern ${ID_PATTERN} (letters, digits, and dashes only, starting with a letter)`,
+      );
+    }
```
Sarabadu: @Rugvip would something like this work for having a warning before imposing the new pattern? note sure if we have something else that the console.warn on this level

The ID_PATTERN_OLD is currently failing after the `onPluginModuleBootFailure` implementation

Sarabadu:
@Rugvip if we want to implement this on the frontend too, do we have any `-common` package at framework level to store the id patterns?
should we implement the same id for Plugins, RouteRef id, extensions, APIref ? I need to get myself a bit more familiar with the new frontend frameworks, so i would love to have some pointers in case I'm missing something there

packages/backend-app-api/src/wiring/BackendInitializer.test.ts
```diff
@@ -774,7 +776,7 @@ describe('BackendInitializer', () => {
     const init = new BackendInitializer([]);
     init.add(
       createBackendPlugin({
-        pluginId: 'test-1',
+        pluginId: 'test-t1',
```
freben: This didn't need to change, did it?

packages/backend-plugin-api/src/wiring/constants.ts
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * The pattern that IDs must match.
+ *
+ * @remarks
+ * ids must only contain the letters `a` through `z` and digits, in groups separated by
+ * dashes. Additionally, the very first character of each such group
```
freben: Not of each such group - only the very first one should not be a digit, because that risks messing with yaml parsing it as a number instead of a string identifier.

packages/backend-plugin-api/src/wiring/createBackendPlugin.ts
```diff
@@ -49,6 +50,17 @@ export function createBackendPlugin(
   options: CreateBackendPluginOptions,
 ): BackendFeature {
   function getRegistrations() {
+    if (!ID_PATTERN.test(options.pluginId)) {
+      console.warn(
+        `WARNING: The pluginId '${options.pluginId}' will be invalid soon please must match the pattern ${ID_PATTERN} (letters, digits, and dashes only, starting with a letter)`,
```
freben:
```suggestion
        `WARNING: The pluginId '${options.pluginId}' will be invalid soon, please change it to match the pattern ${ID_PATTERN} (letters, digits, and dashes only, starting with a letter)`,
```

packages/config/src/constants.ts
```diff
@@ -0,0 +1,27 @@
...
+ * The pattern that config keys must match.
+ *
+ * @remarks
+ * keys must only contain the letters `a` through `z` and digits, in groups separated by
+ * dashes or underscores and colon. Additionally, the very first character of each such group
+ * must be a letter, not a digit
+ *
+ * @public
+ */
+export const CONFIG_KEY_PART_PATTERN = /^[a-z][a-z0-9]*(?:[-_:][a-z0-9]+)*$/i;
```
freben: I'm torn about this one. I wonder what it is that we are protecting against here. When we check these keys, it would already be _after_ yaml parsing has happened. Then it's already too late to capture e.g. the problem of having an initial digit. So I think there's little reason to cover for that here.

So then what remains? There's nothing stopping someone having eg

```yaml
emailSettings:
  some-user@example.com:
    enabled: false
```

Then what should we do about that? Well we know that hey, you put a dot in the key, that's risky because it gets hard to read out of the config object since we use dots as separators. But you CAN and maybe they were already using just doing `get` and parsing with zod anyway and get angry with us for complaining. And arguably that's just a limitation of our config system and maybe we should be the ones implementing support for advanced syntax like

```ts
.getConfig('emailSettings["some-user@example.com"].enabled')
```

or even

```ts
.getConfig(['emailSettings', 'some-user@example.com', 'enabled'])
```

instead of bugging the user about it?

I think at the raw config level, this type of thing perhaps fits better in e.g. a `config:print` or lint command or whatnot, where we can just warn the user gently that hey you seem to have some config here that might be tough to read out.

Sarabadu: I been thinking on the same, not sure the initial reason on this restrictions. But I didn't wanted to change the current restriction we have on the config validation to not create a bigger scope here.
The change here is just to not have replicated regex on config and config loader package.

I feel a bit lost in order to push something like "Everything valid on yaml, should be backstage config valid", I think it could be great to have that though. Do you think it is something we should bring to the framework sig to collect feedback on that? or maybe create a separated PR with that proposal, WDYT?

That will also remove this initial issue too, i mean if all yaml valid is valid, the current errors should not be a problem. But respecting naming id maybe is better to be more restrictive 🤔

Rugvip: Iirc the restriction were mostly due to the ability to set config via the `APP_CONFIG_<key>` env var pattern. That's not quite as important anymore though given people mostly rely on the substitution nowadays

packages/backend-plugin-api/src/wiring/createBackendModule.ts
Rugvip: Yep, sounds good 👍 

Was it a deliberate choice to have this check live in `getRegistrations` rather than the create function itself? I could see it being good to not break loading of packages, but it's also often best to throw as early as possible.

packages/config/report.api.md
```diff
@@ -43,6 +43,9 @@ export type Config = {
   getOptionalStringArray(key: string): string[] | undefined;
 };
 
+// @public
+export const CONFIG_KEY_PART_PATTERN: RegExp;
```
Rugvip: Hmm, why the changes to the config packages? We could keep all of that as is right?

.changeset/slow-numbers-study.md
```diff
@@ -0,0 +1,8 @@
+---
+'@backstage/backend-plugin-api': minor
+'@backstage/backend-app-api': minor
+'@backstage/config-loader': patch
+'@backstage/config': patch
+---
+
+The backend will now throw an error if a plugin or a module doesn't have a valid ID
```
Rugvip: This description only applies to `@backstage/backend-plugin-api`

packages/config/report.api.md
Sarabadu: The only change was exporting this pattern to reuse it in the config-loader, but as it generates some confusion i will remove the change

Sarabadu:
coming back to this from pto, i will look to replicate the same in the frontend part too, thanks 👍🏽

packages/backend-plugin-api/src/wiring/createBackendModule.ts
Sarabadu: Yes I was afraid of breaking the framework logic by doing it outside, but you are right for this type of validation maybe make sense to throw as soon as the function is called

packages/backend-plugin-api/src/wiring/createBackendModule.ts
```diff
@@ -54,6 +55,17 @@ export interface CreateBackendModuleOptions {
...
   options: CreateBackendModuleOptions,
 ): BackendFeature {
+  if (!ID_PATTERN.test(options.moduleId)) {
+    console.warn(
+      `WARNING: The moduleId '${options.moduleId}' for plugin '${options.pluginId}', will be invalid soon, please change it to match the pattern ${ID_PATTERN} (letters, digits, and dashes only, starting with a letter)`,
+    );
+  }
+  if (!ID_PATTERN_OLD.test(options.moduleId)) {
+    throw new Error(
+      `Invalid moduleId '${options.moduleId}' for plugin '${options.pluginId}', must match the pattern ${ID_PATTERN} (letters, digits, and dashes only, starting with a letter)`,
```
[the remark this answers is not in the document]
Sarabadu: we still want to match the new pattern, if any change is needed

packages/frontend-plugin-api/src/wiring/constants.ts
```diff
@@ -0,0 +1,30 @@
...
+ * The pattern that IDs must match.
+ *
+ * @remarks
+ * ids must only contain the letters `a` through `z` and digits, in groups separated by
+ * dashes. Additionally, the very first character of the first group
+ * must be a letter, not a digit
+ *
+ * @public
+ */
+export const ID_PATTERN = /^[a-z][a-z0-9]*(?:-[a-z0-9]+)*$/i;
```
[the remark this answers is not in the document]
Sarabadu: we do not need the old pattern on the frontend as is not breaking atm

Sarabadu:
@Rugvip I've updated the branch 
for the frontend I just added the warning for the pluginId, as it is not generating issues atm,  hope it makes

.changeset/odd-lemons-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/frontend-plugin-api': minor
+---
+
+new deprecation warning will be thrown by the framework `pluginId` should have letters, digits, and dashes, starting with a letter
```
Rugvip:
```suggestion
Plugin IDs that do not match the standard format are deprecated (letters, digits, and dashes only, starting with a letter). Plugin IDs that do no match this format will be rejected in a future release.
```

.changeset/slow-numbers-study.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-plugin-api': minor
+---
+
+The backend will now throw an error if a plugin or a module doesn't have a valid ID
```
Rugvip:
```suggestion
Plugin IDs that do not match the standard format are deprecated (letters, digits, and dashes only, starting with a letter). Plugin IDs that do no match this format will be rejected in a future release.

In addition, plugin IDs that don't match the legacy pattern that also allows underscores, with be rejected.
```

Rugvip (approved):
Yep let's :shipit:, just refining the changesets a lil bit

PR #30093, opened 2025-05-30
chore: update to node 22 and 24

drodil:

Updated engines and node types to 22. Node.js 24, on the other hand, is generally faster than Node.js 22 due to improvements in the V8 engine and other core features. Specifically, Node.js 24 uses V8 version 13.6, which provides performance improvements and also includes enhanced security checks and lifecycle script hardening. 

### DIFF
--- a/.changeset/fifty-walls-argue.md
+++ b/.changeset/fifty-walls-argue.md
@@ -1,0 +1,15 @@
+---
+'@backstage/plugin-scaffolder-backend-module-rails': patch
+'@backstage/plugin-kubernetes-cluster': patch
+'@backstage/plugin-scaffolder-backend': patch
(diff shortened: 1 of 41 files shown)


awanlin (commented):
Hi @drodil, though I agree with your reasons it needs to be noted that this would break our current Node version policy as Node 24 is not an Active LTS version yet: https://backstage.io/docs/overview/versioning-policy#nodejs-releases

drodil:
@freben @awanlin yep, agreed, was meaning to make this a draft in the beginning, just to test things out and what to expect from this update. And it seems it will break a lot of tests and other stuff :D 

After all, it's not so far away in the future (2025-10-28), and time just flies.

PR #30095, opened 2025-05-30
Export function for creating `rootLogger`

KevinSnyderCodes:
Backstage docs currently recommend duplicating the `rootLoggerServiceFactory.ts` code if they want to override the service: https://backstage.io/docs/backend-system/core-services/root-logger/#configuring-the-service

However, in some cases it would be preferable to reuse the existing logic so that you don't miss any upstream changes to the default `rootLogger` implementation.

For example, if you wanted to add additional attributes to the `rootLogger`:

```ts
import { createBackend } from '@backstage/backend-defaults';
import {
  createServiceFactory,
  coreServices,
import { createRootLogger } from '@backstage/backend-defaults/rootLogger';

const backend = createBackend();

backend.add(
  createServiceFactory({
    service: coreServices.rootLogger,
    deps: {
      config: coreServices.rootConfig,
    },
    async factory({ config }) {
      const logger = await createRootLogger({ config });
      return logger.child({ someKey: 'some-value' });
    },
  }),
);
```

This PR implements the above `createRootLogger()` function and exports it from the `@backstage/backend-defaults` package. The existing logic for creating the default `rootLogger` is moved to this function, and the `rootLoggerServiceFactory` simply calls `createRootLogger()`.

---

Right now this is just a proposal. There are some implementation details I'm struggling to work through (see code comments). However, overall I think a strategy like this would be a good way to reduce the amount of code duplication required to override the default `rootLogger` implementation.

### DIFF
--- a/packages/backend-defaults/src/entrypoints/rootLogger/index.ts
+++ b/packages/backend-defaults/src/entrypoints/rootLogger/index.ts
@@ -14,5 +14,8 @@
  * limitations under the License.
  */
 
-export { rootLoggerServiceFactory } from './rootLoggerServiceFactory';
+export {
+  createRootLogger,
+  rootLoggerServiceFactory,
+} from './rootLoggerServiceFactory';
 export { WinstonLogger, type WinstonLoggerOptions } from './WinstonLogger';
--- a/packages/backend-defaults/src/entrypoints/rootLogger/rootLoggerServiceFactory.ts
+++ b/packages/backend-defaults/src/entrypoints/rootLogger/rootLoggerServiceFactory.ts
@@ -17,10 +17,31 @@
 import {
   createServiceFactory,
   coreServices,
+  RootConfigService,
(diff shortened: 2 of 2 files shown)


packages/backend-defaults/src/entrypoints/rootLogger/index.ts
```diff
@@ -14,5 +14,8 @@
  * limitations under the License.
  */
 
-export { rootLoggerServiceFactory } from './rootLoggerServiceFactory';
+export {
+  createRootLogger,
```
KevinSnyderCodes, line 18: For some reason running `yarn build` on this package does _not_ include the `createRootLogger` function in the generated `.d.ts` file. However, the export does appear in the `.cjs.js` file. So the runtime export is there, but the type definition is missing, making this function inaccessible outside the package.

I have tried multiple combinations of things to get this working without success. Any help here would be greatly appreciated!

PR #30098, opened 2025-05-31
cli: combine build and start modules

Rugvip:

I think these are too intertwined to be separate modules. Might be that we're able to break it down in the future somehow, but with the current implementation I think it's best to keep it all in one place.


freben (approved):
Sure but, doesn't make it any easier to find commands really. Is this refactor a symptom of a lack of a nice way of sharing lib code between commands? Or the need for some form of command-model/controller separation model?

Rugvip:
Yep, basically it would be nice if this kind of split was possible, but with the current implementation it really isn't. I don't think we should spend time towards enabling it right now either, there are other more important things when it comes to CLI modularization that I think are better to focus on first, like auth and actually shipping it.

PR #30102, opened 2025-06-01
cli: disable caching Jest module loader for node

Rugvip:

This fixes the `Test environment has been torn down` flaky test issue, like this one: https://github.com/backstage/backstage/actions/runs/15364676304/job/43236105338?pr=30098

Turns out it happens when you get the same dynamic ESM imports happening again on the same worker, and it could be reproduced reliably with `--runInBand`. Bit more context here: https://github.com/jestjs/jest/issues/10944#issuecomment-781523091

Still a bit interesting that caching the script of the importing (!) module breaks things in this way, but I don't think it's worth digging into too much. Fix is to simply not use the caching module loader for node packages. I do think it's still worth keeping it for frontend packages as it's a pretty decent performance boost and there's no risk of that issue there since dynamic imports are transpiled rather than left as is.

### DIFF
--- a/.changeset/slow-impalas-wear.md
+++ b/.changeset/slow-impalas-wear.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+Only use the caching Jest module loader for frontend packages in order to avoid breaking real ESM module imports.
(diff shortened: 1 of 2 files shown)


freben (approved):
Oh wow. Good catch.

Issue #29963, opened 2025-05-15
Scaffolder templates in `Top Visited` and `Recently Visited` cards show generic title — `Create a new component` — instead of specific template name


its-mitesh-kumar:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
Top Visited, Recent Visited
```

### 🗃️ Project Area

Software Templates

### 🔗 External Integration

N/A

### 📝 Description & Context

On the  Home Page, the `Top Visited` and `Recently Visited` cards display entries for pages visited under  `/create` and individual template URLs like:

`create/templates/default/react-ssr-template`
`create/templates/default/springboot-template`

**Currently:**

Links to the general `/create` page appear with the title `create`.

Links to specific software templates show up with the title `Create a new component`, regardless of which template was actually used.

**This is problematic because:**

Users cannot distinguish which specific scaffolder template each entry corresponds to.

Every visited template looks the same in these cards, even if the underlying URLs  are different.

This diminishes the value of the cards for users who frequently work with multiple templates.

We would expect the title displayed in the cards to reflect the actual template name — for example : `Spring Boot gRPC Service`, `React SSR Template`, `ArgoCD Template`, `Backend Plugin Template`, etc.

Attaching screen recording  where you can see multiple entries of `Create a new component | Custom Backstage` in Top and Recent Visited which correspond to different templates we visited . 

### 👍 Expected Behavior

Each scaffolder template visited via `/create/templates/${namespace}/${template_name}` should appear in the Top/Recently Visited cards with its `actual template title`, rather than the static text `Create a new component`.

This allows users to quickly recognize and return to the specific templates they’ve used.

### 🥾 Reproduction steps

1. Spin up a Backstage app.
3. Navigate to /create and use two different templates (e.g., ArgoCD, Backend Plugin).
5. Return to the Home page where Top Visited / Recently Visited cards are added.

Observe that:

Each template appears with a unique path (e.g., /create/templates/default/argocd-template).

But all entries show the same title: `Create a new component`.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes, and I have enough information to get started

benjidotsh:
[...]

PR #30110, opened 2025-06-03
fix(scaffolder): show specific template title in Top & Recently Visited cards

its-mitesh-kumar:
## Description 
This PR resolves an issue where the `Top Visited` and `Recently Visited` cards on the home page always display the generic title `Create a new component`, regardless of the actual template selected. We now ensure the specific template title is captured as `Template ${templateTitle}`.

## Fixes: 

Fixing [issue-29963](https://github.com/backstage/backstage/issues/29963)

## UI before changes

https://github.com/user-attachments/assets/3b9fcdac-8f28-4dab-b5ab-7c2832e36111

## UI after changes

https://github.com/user-attachments/assets/e1aa9af0-3767-465d-aaaa-c24e448821ae


plugins/scaffolder/src/translation.ts
```diff
@@ -353,6 +353,7 @@ export const scaffolderTranslationRef = createTranslationRef({
       pageContextMenu: {
         editConfigurationTitle: 'Edit Configuration',
       },
+      template: 'Template',
```
benjdlambert: I wonder if this should actually be `Template {{templateName}}` instead and do interpolation in the messages in the `i18n`?

benjdlambert (commented):
Just a small comment about the translation. Would also be good to add a test here too please :pray:

plugins/scaffolder/src/translation.ts
its-mitesh-kumar: **@benjdlambert** I've addressed your earlier comment. Regarding the test, would it be okay to skip it for now? It’s proving a bit tricky because `templateName` is being passed via `location.state` in `TemplateWizardPage`:

```const location = useLocation(); const templateTitle = location.state?.templateTitle;```

If you happen to have a reference to a similar test case or any documentation that covers this pattern, I’d really appreciate it.

File: .changeset/brown-waves-hide.md [hunk 2]
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder': patch
+---
+
+Fixes an issue where the `Top Visited` and `Recently Visited` cards displayed a generic title (`Create a new component`) instead of the specific template name. Titles are now shown as `Template ${templateTitle}`, using the template's metadata.
```
logonoff:
```suggestion
Fixed an issue where the `Top Visited` and `Recently Visited` cards displayed a generic title (`Create a new component`) instead of the specific template name. Titles are now shown as `Template ${templateTitle}`, using the template's metadata.
```

logonoff (approved):

benjdlambert:
Still wanting to get some tests for this behavior before we merge I think.

You should be able to either mock out `useLocation` in a test so that you can return your own `state` or find a different way to set some state inside the `MemoryRouter` or something that you render the component in.

its-mitesh-kumar:
> Still wanting to get some tests for this behavior before we merge I think.
> 
> You should be able to either mock out `useLocation` in a test so that you can return your own `state` or find a different way to set some state inside the `MemoryRouter` or something that you render the component in.

@benjdlambert, I have added unit tests, please have a look.

plugins/scaffolder/src/alpha/components/TemplateWizardPage/TemplateWizardPage.tsx
```diff
@@ -82,6 +82,9 @@ export const TemplateWizardPage = (props: TemplateWizardPageProps) => {
   const { templateName, namespace } = useRouteRefParams(
     selectedTemplateRouteRef,
   );
+  const location = useLocation();
+  const templateTitle = location.state?.templateTitle;
```
benjdlambert, line 86: I'm not sure we want to use `state` to pass the name here. Feels a little bit out of place to do that.

I think that the `manifest` object has a `title` prop that we should use for the title instead, which is a few lines further down.

That way I don't think we need any special `react-router-dom` things at all, we can just pick that out from the `manifest` object that's returned from the backend instead.

benjdlambert (changes requested):
Hey :wave:

Thanks for this, although I think I found a cleaner way to do this which doesn't require that much of a diff from the original code.

plugins/scaffolder/src/alpha/components/TemplateWizardPage/TemplateWizardPage.tsx
its-mitesh-kumar: @benjdlambert, Since the `manifest` is loaded asynchronously via `useTemplateParameterSchema(templateRef)`, there’s a brief delay before it’s available. During that window, a fallback or previous title is written to `localStorage`; that value is what feeds the **Recently Visited** and **Top Visited** cards.

If we rely directly on `manifest.title`, it will lead to an incorrect title saved in storage and then show up in these cards. **Hence I opted to pass the title explicitly via React Router state** to guarantee we persist the correct value.

ScreenRecording for your reference. 

https://github.com/user-attachments/assets/5def514e-345a-482b-84ef-c024949d679d

its-mitesh-kumar: @benjdlambert, once you get time please have a look.

its-mitesh-kumar: @benjdlambert this PR need your attention.

its-mitesh-kumar: /cc @vinzscam can you have a look ?

debsmita1:
@its-mitesh-kumar I tested your PR locally, and noticed that when I click on a template from the top visited card/ recently visited card and return, it no longer displays the template name. Is this expected ??

https://github.com/user-attachments/assets/bdebecef-3e9f-4264-a527-0569d229ee02

its-mitesh-kumar:
Created a new PR https://github.com/backstage/backstage/pull/30732 so closing it. 
cc @@debsmita1  @benjdlambert

PR #30113, opened 2025-06-03
`scaffolder-module-bitbucket-cloud`: migrate to using new actions

benjdlambert:
Signed-off-by: benjdlambert <ben@blam.sh>

### DIFF
--- a/plugins/scaffolder-backend-module-bitbucket-cloud/report.api.md
+++ b/plugins/scaffolder-backend-module-bitbucket-cloud/report.api.md
@@ -20,15 +20,56 @@ export const createBitbucketPipelinesRunAction: (options: {
   {
     workspace: string;
     repo_slug: string;
-    body?: object;
-    token?: string;
+    body?:
+      | {
+          target?:
+            | {
+                type?: string | undefined;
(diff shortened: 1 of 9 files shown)


plugins/scaffolder-backend-module-bitbucket-cloud/report.api.md
```diff
@@ -20,15 +20,56 @@ export const createBitbucketPipelinesRunAction: (options: {
...
+      | undefined;
+    token?: string | undefined;
   },
   {
-    buildNumber: number;
-    repoUrl: string;
-    pipelinesUrl: string;
+    buildNumber?: number | undefined;
+    repoUrl?: string | undefined;
+    pipelinesUrl?: string | undefined;
```
freben, line 70: just noting that these used to be mandatory but now aren't - i don't believe it's unknown or a problem but interesting

benjdlambert: I think that it's fine. They were required in the type parameter, but the jsonschema marks them as optional, so I think it's fine for us just to use the jsonschema as the source of truth here.

PR #30127, opened 2025-06-03
Add a standard `toString` on credentials objects

freben:
This is useful when cache-keying, or wanting to log a string form, or track credentials of actors in a database, etc. One such example is techdocs caching reads [here](https://github.com/backstage/backstage/blob/a58e48e983224686a9ef7862d6b29976dd202732/plugins/techdocs-backend/src/service/CachedEntityLoader.ts#L44), which is currently based on a token which blocks migration of this code to using credentials and the `CacheService` instead of `CacheApi`.

This also at the same time adds tests that check the JSON stringified form of credentials, to make visible the fact that these might be sent across the wire or persisted.

### DIFF
--- a/packages/backend-defaults/src/entrypoints/auth/helpers.test.ts
+++ b/packages/backend-defaults/src/entrypoints/auth/helpers.test.ts
@@ -82,33 +82,62 @@ describe('credentials', () => {
     ).not.toMatch(/my-token/);
   });
 
-  it('should have a serializable form', () => {
-    expect(
-      String(createCredentialsWithServicePrincipal('my-service')),
-    ).toMatchInlineSnapshot(
-      `"{"$$type":"@backstage/BackstageCredentials","type":"service","subject":"my-service"}"`,
+  it('should have a serializable form both as strings and as JSON', () => {
+    const simpleService = createCredentialsWithServicePrincipal('my-service');
+    expect(String(simpleService)).toMatchInlineSnapshot(
+      `"backstageCredentials{servicePrincipal{my-service}}"`,
     );
-    expect(
-      String(
-        createCredentialsWithUserPrincipal('user:default/mock', 'my-token'),
-      ),
-    ).toMatchInlineSnapshot(
-      `"{"$$type":"@backstage/BackstageCredentials","type":"user","userEntityRef":"user:default/mock"}"`,
+    expect(JSON.stringify(simpleService)).toMatchInlineSnapshot(
+      `"{"$$type":"@backstage/BackstageCredentials","version":"v1","principal":{"type":"service","subject":"my-service"}}"`,
     );
-    expect(
-      String(
-        createCredentialsWithUserPrincipal(
-          'user:default/mock',
-          'my-token',
-          undefined,
-          'my-actor',
-        ),
-      ),
-    ).toMatchInlineSnapshot(
-      `"{"$$type":"@backstage/BackstageCredentials","type":"user","userEntityRef":"user:default/mock","actor":{"type":"service","subject":"my-actor"}}"`,
+
+    const serviceWithAccessRestrictions = createCredentialsWithServicePrincipal(
+      'my-service',
+      undefined,
+      {
+        permissionNames: ['perm'],
+        permissionAttributes: {
+          action: ['read'],
+        },
+      },
     );
-    expect(String(createCredentialsWithNonePrincipal())).toMatchInlineSnapshot(
-      `"{"$$type":"@backstage/BackstageCredentials","type":"none"}"`,
+    expect(String(serviceWithAccessRestrictions)).toMatchInlineSnapshot(
+      `"backstageCredentials{servicePrincipal{my-service,accessRestrictions=cXWOJgUirHkHNZIowUi/YO5nwEwhTicC38iXi2XTYCk}}"`,
+    );
+    expect(JSON.stringify(serviceWithAccessRestrictions)).toMatchInlineSnapshot(
+      `"{"$$type":"@backstage/BackstageCredentials","version":"v1","principal":{"type":"service","subject":"my-service","accessRestrictions":{"permissionNames":["perm"],"permissionAttributes":{"action":["read"]}}}}"`,
+    );
+
+    const simpleUser = createCredentialsWithUserPrincipal(
+      'user:default/mock',
+      'my-token',
+    );
+    expect(String(simpleUser)).toMatchInlineSnapshot(
+      `"backstageCredentials{userPrincipal{user:default/mock}}"`,
+    );
+    expect(JSON.stringify(simpleUser)).toMatchInlineSnapshot(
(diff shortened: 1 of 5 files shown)


packages/backend-defaults/src/entrypoints/auth/helpers.test.ts
```diff
@@ -42,6 +42,23 @@ describe('credentials', () => {
...
+      ),
+    ).toEqual({
+      $$type: '@backstage/BackstageCredentials',
+      version: 'v1',
+      principal: {
+        type: 'user',
+        userEntityRef: 'user:default/mock',
+        actor: { type: 'service', subject: 'my-actor' },
+      },
+    });
```
freben, line 60: this is unrelated; i just noticed that this test didn't have a case for the actor input

packages/backend-defaults/src/entrypoints/auth/helpers.ts
```diff
@@ -28,23 +28,31 @@ export function createCredentialsWithServicePrincipal(
...
+      subject: sub,
+      accessRestrictions,
+    },
+  } as const;
+  Object.defineProperty(result, 'token', {
+    enumerable: false,
+    configurable: true,
+    value: token,
+  });
+  Object.defineProperty(result, 'toString', {
```
Rugvip: Thinking `.defineProperties` instead

packages/backend-defaults/src/entrypoints/auth/helpers.ts
```diff
@@ -28,23 +28,31 @@ export function createCredentialsWithServicePrincipal(
...
+    enumerable: false,
+    configurable: true,
+    value: () =>
+      JSON.stringify({
```
Rugvip: Thoughts on a slightly nicer string version like we do in some other places, like `backstageCredentials{type=...,subject=...json...}`?

Not sure making these intentionally parsable is a great idea tbh. You'd always have access to desired fields and the time or serialization anyway. Any particular use-case you have in mind with an exact example?

packages/backend-defaults/src/entrypoints/auth/helpers.ts
```diff
@@ -28,23 +28,31 @@ export function createCredentialsWithServicePrincipal(
...
+        $$type: '@backstage/BackstageCredentials',
         type: 'service',
         subject: sub,
```
Rugvip: If we're gonna use it as a key where the same `toString` result means the same level of access - as it's intended to be used in TechDocs - I'd say we need to include `accessRestrictions` here too, either in plain form or a hash.

Rugvip (commented):
Nice! 👍 

Should we explicitly call out that `toString` is intended to uniquely identify the access scope as well? The token is being used that way in the TechDocs cache right now, and it'd be important to make sure that it remains the case as we add more fields to the credentials.

packages/backend-defaults/src/entrypoints/auth/helpers.ts
freben: Oh interesting - absolutely, i had not seen that one for some reason - fixed

packages/backend-defaults/src/entrypoints/auth/helpers.ts
freben: So that's the question. Should this one only be lightly informative, or actually useful for systems.

My gut feeling was that it would much more common to need to parse and display these sensibly, than to just get something roughly representative.

Use cases:

```ts
auditor.createEvent({
  ...
  actor: credentials.toString()
})
```

and then having an audit log viewer where it displays an actual control for that, clickable to the right person etc

The scaffolder wanting to [persist and read back the initiator](https://github.com/backstage/backstage/blob/4ce48a70646e2283a2b20d5c2857f42b0db8e394/plugins/scaffolder-backend/src/service/router.ts#L503) to workers

```ts
const secrets: InternalTaskSecrets = {
  __initiatorCredentials: JSON.stringify({
    ...credentials,
    token: (credentials as any).token, // this one needs addressing separately
  }),
};
```

In https://github.com/backstage/backstage/pull/30146, routers needing to persist the current actor and then that traversing through layers and eventually being sent as Backstage events to be consumed by third parties.

In https://github.com/backstage/backstage/pull/29983/files#diff-84a97da68c3a1de6244d988ac786d2f40b59b5dd3f3470cc3fa851f93e44b375R232, inventing a custom wrapper permission type only for the reason of making a cache key out of it (interestingly though, that one is in a common library!)

... and the list goes on. Sure, there are some fringe cases where essentially just a hash of the blob would suffice, but that's likely more rare.

I do think that it's overall a clear hole in the auth offering that there's no canonical serialized form. It could be as a URN, as JSON, as ... well anything, but something that's well known with a schema and a documented definition and is parseable.

There are also places that currently extract the user ref out of credentials and persist that, but where they (1) partly do it because that's the only realistic option, and (2) that doesn't work for the other credential types so it leads to hacks and workarounds

packages/backend-defaults/src/entrypoints/auth/helpers.ts
freben: I'm mot sure it does mean that in their case though, does it? Maybe I'm missing some nuance, but at this point they are happy to use the incoming raw token string as the key and that seems to be what gets their needs served

packages/backend-defaults/src/entrypoints/auth/helpers.ts
freben: An obvious compromise here could be to make `toString` in the form you suggest, and defining `toJSON` in basically the form that I suggest (and let people stringify/parse that if they specifically want to).

freben: Made up my mind to update them all to do what you proposed, and at the same time make sure that the plain JSON stringification has a nice understandable form.

packages/backend-defaults/src/entrypoints/auth/helpers.test.ts
```diff
@@ -64,4 +81,63 @@ describe('credentials', () => {
...
     ).not.toMatch(/my-token/);
   });
+
+  it('should have a serializable form both as strings and as JSON', () => {
+    const simpleService = createCredentialsWithServicePrincipal('my-service');
+    expect(String(simpleService)).toMatchInlineSnapshot(
+      `"backstageCredentials{servicePrincipal{my-service}}"`,
+    );
+    expect(JSON.stringify(simpleService)).toMatchInlineSnapshot(
+      `"{"$$type":"@backstage/BackstageCredentials","version":"v1","principal":{"type":"service","subject":"my-service"}}"`,
```
freben, line 91: So this is a remaining question. Is there a risk that people deserialize this on the receiving end and try to treat them as an ACTUAL real credentials and pass it in somewhere else? Should we change it to a different `$$type` like for example `@backstage/SerializedBackstageCredentials` or whatever? So it's not even directly castable

benjdlambert (approved):

PR #30128, opened 2025-06-03
module-elasticsearch: include custom document id to batch upload

matheusjv11:

This PR introduces a new option, `batchKeyField`, to the Elasticsearch indexer. This feature allows users to standardize the ID of indexed documents by specifying a consistent field to be used as the document ID during batching operations.  
 

.changeset/stale-adults-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-search-backend-module-elasticsearch': patch
+---
+
+Added support for batchKeyField in the Elasticsearch indexer to allow consistent document IDs during bulk uploads, enabling easier lookups and avoiding duplication.
```
awanlin:
```suggestion
Added support for batchKeyField in the Elasticsearch indexer to allow consistent document IDs during bulk uploads.
```

Lets just simplify this, that will make Vale happy as well.

docs/features/search/search-engines.md
```diff
@@ -239,6 +239,18 @@ search:
...
+### Elasticsearch batch key field
+
+By default, during bulk uploads with the Elasticsearch indexer, each document is assigned a custom `_id` because no explicit ID is provided. However, by setting the `batchKeyField`, you can standardize the identifier used for documents within the index. This helps streamline direct searches and prevents duplicate entries when adding new documents.
+
+**Set batch key field to id**
+
+```yaml
+search:
+  elasticsearch:
+    batchKeyField: id
```
awanlin: The use of `id` has me a bit confused. What will the end result look like? Can we use an example other than `id` to make that more clear? Also add an example to show the default as well as with this feature in place?

awanlin (changes requested):
Thanks for the contribution @matheusjv11, I gave this a quick look over and left two comments. This is just outside my ES understanding so we going to need the Search Maintainers to give this a proper look.

If you rebase that should fix the CI error 👍

squid-ney:
Hi [matheusjv11](https://github.com/matheusjv11), is your team updating documents in the index often? From my understanding overriding the autogenerated id can slow indexing. But, if you are often looking up and making updates to documents, then the custom id is beneficial. Internally, we don't really have this use case so the autogenerated id has been working fine. 

I don't see any cons to opening up this configuration option. However, I agree with @awanlin that the documentation updates could be clearer [docs/features/search/search-engines.md](https://github.com/backstage/backstage/pull/30128/files/3fa68291c83e4691bf23d3ac10077f190ad34faa#diff-fe886a3fc6c17dec6bdda815f2ff3a85b7404963b1660c2fc64edc7fa898bc51). 
> By default, during bulk uploads with the Elasticsearch indexer, each document is assigned a custom `_id` because no explicit ID is provided. However, by setting the `batchKeyField`, you can standardize the identifier used for documents within the index. This helps streamline direct searches and prevents duplicate entries when adding new documents.

^To me this implies to me that search users should set a custom id which generally I don't think they need to do. Maybe you could frame this as a useful setting if the user does lots of lookups or modifications in the index. Also, I'd make sure to highlight that if the `batchKeyField` is not unique across documents, Elasticsearch will overwrite any existing document that shares the same _id.

matheusjv11:
> Hi [matheusjv11](https://github.com/matheusjv11), is your team updating documents in the index often? From my understanding overriding the autogenerated id can slow indexing. But, if you are often looking up and making updates to documents, then the custom id is beneficial. Internally, we don't really have this use case so the autogenerated id has been working fine.
> 
> I don't see any cons to opening up this configuration option. However, I agree with @awanlin that the documentation updates could be clearer [docs/features/search/search-engines.md](https://github.com/backstage/backstage/pull/30128/files/3fa68291c83e4691bf23d3ac10077f190ad34faa#diff-fe886a3fc6c17dec6bdda815f2ff3a85b7404963b1660c2fc64edc7fa898bc51).
> 
> > By default, during bulk uploads with the Elasticsearch indexer, each document is assigned a custom `_id` because no explicit ID is provided. However, by setting the `batchKeyField`, you can standardize the identifier used for documents within the index. This helps streamline direct searches and prevents duplicate entries when adding new documents.
> 
> ^To me this implies to me that search users should set a custom id which generally I don't think they need to do. Maybe you could frame this as a useful setting if the user does lots of lookups or modifications in the index. Also, I'd make sure to highlight that if the `batchKeyField` is not unique across documents, Elasticsearch will overwrite any existing document that shares the same _id.

Hey @squid-ney, thanks for the feedback! I’ve updated the description to present a more cohesive explanation. (please let me know if there’s anything else that could be improved)

As for the use case, we need the ability to update individual documents because we're also inserting data via custom webhooks. Without a stable unique ID, there’s a risk of duplicating documents when webhook-based inserts overlap with batch uploads. Additionally, with the right hooks in place, we’re aiming to significantly reduce the frequency of batch uploads, since updates will occur live.

matheusjv11:
> Thanks for the contribution @matheusjv11, I gave this a quick look over and left two comments. This is just outside my ES understanding so we going to need the Search Maintainers to give this a proper look.
> 
> If you rebase that should fix the CI error 👍

Thank you @awanlin for the feedback! When you have a moment, could you please check if the latest updates address your comments? Regarding the default example, I included it directly below the configured example to show the contrast clearly. Since batchKeyField is optional, the default behavior is simply not specifying the field in the configuration.

docs/features/search/search-engines.md
```diff
@@ -239,6 +239,26 @@ search:
 
 > You can also increase the batch size if you are using a large ES instance.
 
+### Elasticsearch batch key field
+
+By default, during bulk uploads with the Elasticsearch indexer, each document is assigned an autogenerated `_id` unless a `batchKeyField` is explicitly set. This configuration is optional and most users won’t need to customize it. However, if your use case involves frequent lookups or updates to existing documents, setting `batchKeyField` can be beneficial. It allows you to define a consistent identifier for each document, helping to streamline updates and prevent duplicate entries. Be aware that if the value provided for `batchKeyField` is not unique across documents, Elasticsearch will overwrite any existing document with the same `_id`.
```
awanlin: You can add the words that vale isn't happy with to the list here to resolve this errors: https://github.com/backstage/backstage/blob/master/.github/vale/config/vocabularies/Backstage/accept.txt

awanlin (changes requested):
There are still vale error that need to be addressed 👍

awanlin:
Just following up that all this needs is the Vale errors to be addressed, not sure if maybe you're away as it's summer time for many @matheusjv11?

webark:
We are very interested in this to allow for partial updates. Do you need any help moving this forward to completion?

matheusjv11:
Hi @awanlin sorry for the delay. I believe the vale errors are fixed now

freben (approved):

awanlin (approved):
Thanks @matheusjv11 🙏

squid-ney (approved):

PR #30142, opened 2025-06-04
`scaffolder-backend`: last of the actions migrated, and removal of the old `TemplateAction` types

benjdlambert:


plugins/scaffolder-backend-module-bitbucket-cloud/report.api.md
```diff
@@ -103,14 +102,16 @@ export function createPublishBitbucketCloudPullRequestAction(options: {
   {
     repoUrl: string;
     title: string;
-    description?: string;
-    targetBranch?: string;
     sourceBranch: string;
-    token?: string;
-    gitAuthorName?: string;
-    gitAuthorEmail?: string;
+    description?: string | undefined;
```
freben, line 106: needs a changeset

plugins/scaffolder-backend/src/scaffolder/actions/builtin/debug/wait.ts
```diff
@@ -48,27 +48,30 @@ export function createWaitAction(options?: {
...
-          seconds: {
-            title: 'Waiting period in seconds.',
-            type: 'number',
-          },
-          milliseconds: {
-            title: 'Waiting period in milliseconds.',
-            type: 'number',
-          },
-        },
+        minutes: z =>
```
freben, line 57: This one would be nice to allow supporting `ms` (eg "30s") just like `HumanDuration` parsing in the config does, some time in the future.

plugins/scaffolder-node-test-utils/report.api.md
```diff
@@ -7,12 +7,12 @@ import { ActionContext } from '@backstage/plugin-scaffolder-node';
 import { JsonObject } from '@backstage/types';
 
 // @public
-export const createMockActionContext: <
+export function createMockActionContext<
   TActionInput extends JsonObject = JsonObject,
-  TActionOutput extends JsonObject = JsonObject,
+  TActionOutput extends JsonObject = any,
```
freben, line 12: I guess here's where I wonder if it could have been `= {}` instead of any.

plugins/scaffolder-node/src/actions/createTemplateAction.test.ts
```diff
@@ -43,7 +34,9 @@ describe('createTemplateAction', () => {
         const b: string = ctx.input.repoUrl;
         expect(b).toBeDefined();
 
+        // @ts-expect-error - logStream is not available
         const stream = ctx.logStream;
+
         expect(stream).toBeDefined();
```
freben: feels like there should have been a `not` here? Or do you actually supply it anyway 🤯

plugins/scaffolder-backend-module-bitbucket-cloud/report.api.md
benjdlambert: I was just piggybacking off the previous changeset that moved all the other actions 😅 The interface hasn't changed really.

plugins/scaffolder-node/src/actions/createTemplateAction.test.ts
benjdlambert: These assertions are actually not run, they're just there to keep typescript happy. I can find a different way to keep typescript happy as it's confusing.

plugins/scaffolder-node-test-utils/report.api.md
benjdlambert: Unfortunately not, the reasons it's like this is that currently you don't provide a `typeof createGithubPushAction` or whatever as a separate type parameters. So it can't infer the `ActionOutputType` without you typing an output function that takes those params. I was thinking of doing this for now as this is non-breaking, and it means that all the new types should work properly with output, and making something like this:

```ts
const mockContext = createMockActionContext<typeof createGithubPushAction>({
  input: {
   // all this would be validated
  }
});

// also typed param
mockContext.getOutput('repoUrl')
```

benjdlambert: The actions that have changed their output types from `JsonObject` to a narrowed type breaks currently.

Issue #8219, opened 2021-11-23
[RFC] Catalog Event Stream Endpoint


Rugvip:
**Status:** Open for comments and community contribution

## Need

The Backstage catalog is designed be a central hub of information within an organization. It typically contains information about and organization's software, resources, and the structure of the organization itself. It achieves this by collecting data from external services and then presents that data using a unified data model.

While the catalog might collect quite a lot of information from these external services, it is best not to ingest **all** of this data into the catalog in too much detail. Attempting to do so will typically lead to a bloated data model that forces in data from too many domains, which in turn makes it harder to work with and evolve the catalog. It is also likely to cause reliability issues because of the increasing number of services that the catalog depends on, or that it simply assumes too much responsibility within an organization.

A great pattern to use instead is have have external services store data that is associated with entities in the catalog. A service would typically use catalog entity names as keys of its data store, and possibly synchronize its data store with the catalog. One example of a service that uses this pattern is the TechDocs backend, which does not store its content within the catalog, but instead uses entity names as keys for its internal or external documentation storage. Other examples of open source plugins that use this pattern are the search, todo, and tech-insights backends, and there is also a use-case with an external service catalogs described in #8162. We've also learned of instances of this pattern being used within adopting organizations, and it is commonly used in services related to our catalog at Spotify as well.

Implementing a service that synchronizes itself with the catalog in a quick and dirty way is quite simple. You fetch all the current entities in the catalog, and then update your database accordingly. This is however wasteful in both the catalog and consuming service ends, and if the client has a need to compute a delta of the updates there's a fair amount of complexity involved in that as well. It is also a solution where you won't be able to get anywhere close to realtime latencies of updates, with it being more realistic to have updates happen once a minute or much more rarely.

I think there is space for providing a much better primitive for building services that integrate with the catalog. One that makes it much simpler to keep up to date with the catalog, and enables services to react to updates in a much more efficient way and with lower delay.

## Proposal

I propose that we extend the catalog REST API with an `/events` endpoint. This endpoint would expose a single linearized stream of events that would function much like a Kafka stream. The consumption would based on offsets, and it would be up to each consuming service to keep track of their own offset as they consume the stream. There would be no server-side state except for the list of events, meaning any number of services and consume this API in parallel.

The following is a high level view of what this API might look like:

```json
GET /api/catalog/events

200 OK
{
  "lastEventOffset": 74,
  "events": [{
    "type": "added",
    "offset": 72,
    "entityRef": "component:default/foo",
[...]

PR #10115, opened 2022-03-10
catalog-backend: implement /events endpoint

Rugvip:
A proposal for a very lightweight implementation of #8219, to see what you think @backstage/maintainers :grin:

PR #30146, opened 2025-06-04
BEP-0013: Catalog History and Events

freben:
Implements https://github.com/backstage/backstage/issues/8219, by making the software catalog track history internally and emitting them as history events.

### DIFF
--- a/beps/0013-catalog-history/README.md
+++ b/beps/0013-catalog-history/README.md
@@ -59,6 +59,7 @@ Primary high level goals:
 
 - External systems that want to react to changes to entities in the catalog can subscribe to history events and get near-realtime updates, so that they can avoid resorting to polling.
 - End users and administrators can see the historical changes to an entity and related items in the catalog, to understand the "why" of its contents.
+- Make the history feature robust and complete enough to implement perfect application level replication of the catalog contents.
 
 Lower level summary of goals:
 
@@ -71,10 +72,6 @@ Lower level summary of goals:
 - History events should contain enough metadata to be usefully possible to correlate to things around the catalog. For example, events that are related to a given entity ID / entity ref / URL location should persist those to be able to easily request those events in a context.
 - History events should retain the actual historic shapes of the entity body itself, since this allows for more use cases of understanding what happened over time to an entity.
 - The events endpoints should be covered by permissions.
-
-Possibly desirable goals, still under discussion:
-
-- Make the history feature robust and complete enough to implement perfect application level replication of the catalog contents.
 
 ### Non-Goals
 
@@ -105,9 +102,11 @@ Add a new history events table to the catalog database. The table has an auto-in
 
 Add triggers to the `locations` and `final_entities` tables, and specifically to changes to the `final_entity` column in the latter. These triggers add history events accordingly, and send `PG_NOTIFY` signals.
 
+Add a history summary table to the catalog database. Whenever a history entry is added, this table gets updated with the latest event state. This table can be used to quickly ask for a rollup summary of everything you need to update yourself with since any given previous point in the history.
+
 Add a history subscriptions table to the catalog database. This is essentially a place to globally track the progress and filter rules of individual subscribers. It also allows for acknowledgement-based locking, so that even distributed consumers are able to easily read from a subscription all together without synchronizing among themselves.
 
-Add an internal history exporter to the catalog, which leverages the history subscriptions feature to "pump" entries from the history events table as they happen to the events backend where other parts of the Backstage ecosystem can consume them.
+Add an internal history exporter to the catalog, which leverages the history subscriptions feature to "pump" entries from the history events table as they happen to the events backend where other parts of the Backstage ecosystem can consume them. This feature can be turned on and off with a config flag.
 
 Add a REST API to make direct filtered reads of the history table. This API can have an optional "blocking" feature like the events backend does, to let consumers both just quickly read events as well as long poll for new ones as they happen.
 
@@ -139,6 +138,13 @@ The following is the proposed shape of the history events table.
 Triggers `AFTER INSERT OR DELETE OR UPDATE OF final_entity ON final_entities` and `AFTER INSERT OR DELETE OR UPDATE ON locations` are added. These lead to the automatic `INSERT` of a corresponding new history event in the events table. At the same time, a `NOTIFY` signal is sent on the Postgres engine.
 
 Add a REST API to read from the history table directly, on `/history/v1/events`. This allows the consumption of events, optionally with filtering on the available columns above so that you could for example grab the complete history of a given entity, or events on a certain organization URL. This API optionally supports blocking long-poll reads so that you can reliably both read the currently existing history and then stream in whatever happens after that point on. The underlying database mechanisms use Postgres `LISTEN`/`NOTIFY` to efficiently know when to unblock pending blocked readers.
+
+The following is the proposed shape of the history entity summary table.
+
+| column       | type    | description                                                   |
+| ------------ | ------- | ------------------------------------------------------------- |
+| `entity_ref` | string  | The relevant entity ref                                       |
+| `event_id`   | big int | A foreign reference to the last encountered relevant event ID |
 
 The following is the proposed shape of the history subscriptions table.
 
--- a/beps/0013-catalog-history/README.md
+++ b/beps/0013-catalog-history/README.md
@@ -55,6 +55,13 @@ List the specific goals of the BEP. What is it trying to achieve? How will we
 know that this has succeeded?
 -->
 
+Primary high level goals:
+
+- External systems that want to react to changes to entities in the catalog can subscribe to history events and get near-realtime updates, so that they can avoid resorting to polling.
+- End users and administrators can see the historical changes to an entity and related items in the catalog, to understand the "why" of its contents.
+
+Lower level summary of goals:
+
 - The catalog should internally track important history events, at least including create/update/delete of entities (as seen on the public API side), and create/update/delete of locations. This tracking must be reliable including complex cases like cascading deletes, and persistent.
 - The catalog should be able to push those history events to the events backend, for easy consumption by other Backstage backend plugins or translation into notifications etc.
 - It should be an open possibility to later easily write integrations that push those history events to any other event system, such as PubSub or SQS.
@@ -64,6 +71,10 @@ know that this has succeeded?
 - History events should contain enough metadata to be usefully possible to correlate to things around the catalog. For example, events that are related to a given entity ID / entity ref / URL location should persist those to be able to easily request those events in a context.
 - History events should retain the actual historic shapes of the entity body itself, since this allows for more use cases of understanding what happened over time to an entity.
 - The events endpoints should be covered by permissions.
+
+Possibly desirable goals, still under discussion:
+
+- Make the history feature robust and complete enough to implement perfect application level replication of the catalog contents.
 
 ### Non-Goals
 
--- a/beps/0013-catalog-history/README.md
+++ b/beps/0013-catalog-history/README.md
@@ -1,6 +1,6 @@
 ---
 title: Catalog History and Events
-status: provisional
+status: implementable
 authors:
   - '@freben'
 owners:
--- a/beps/0013-catalog-history/README.md
+++ b/beps/0013-catalog-history/README.md
@@ -1,0 +1,216 @@
+---
+title: Catalog History and Events
+status: provisional
+authors:
+  - '@freben'
+owners:
+  - '@backstage/catalog-maintainers'
+project-areas:
+  - catalog
+  - events
+creation-date: 2025-06-03
+---
+
+<!--
+**Note:** When your BEP is complete, all these pre-existing comments should be removed
+-->
+
+# Catalog History and Events: <!-- Your short, descriptive title -->
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Design Details](#design-details)
+- [Release Plan](#release-plan)
+- [Dependencies](#dependencies)
+- [Alternatives](#alternatives)
+
+## Summary
+
+<!--
+The summary of the BEP is a few paragraphs long and give a high-level overview of the features to be implemented. It should be possible to read *only* the summary and understand what the BEP is proposing to accomplish and what impact it has for users.
+-->
+
+This BEP proposes giving the software catalog the ability to internally track the history of changes to its state (for example, entities changing or being deleted). And then, to expose those changes both via a REST API and through the events system. This will open up for building realtime/reactive consumption patterns of the catalog, including replicating (all or parts of) its state, instead of requiring polling. It also lets operators and entity owners dig into the historical state of the catalog, understanding causes and effects over time.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals, and non-goals of
+this BEP. Describe why the change is important and the benefits to users.
+-->
+
+The [Catalog Event Stream Endpoint RFC](https://github.com/backstage/backstage/issues/8219) is one of the oldest issues still around in the project and has received a lot of feedback. It contains some discussion and motivation on this topic, but I'll list some here as well:
+
+- Consumers currently have no choice but to use periodic polling to stay up to date with the catalog. This is expensive, and introduces delays before reaching consistency. Having access to event driven mechanisms for staying up to date can mean cheaper, faster updates in external systems as an effect of catalog changes. The catalog itself already benefits massively from being event driven from version control sources; extending the same courtesy to consumers of the catalog could be equally beneficial to them while significantly offloading the catalog backend itself.
+- The internal state machine of the catalog is mostly a black box. It can be challenging to understand why things aren't progressing as expected, and operators regularly resort to digging through backend logs that may or may not contain relevant historical information about the situation. Those logs are also typically not available to entity owners. Making entity owners more self sufficient in being able to diagnose unexpected situations can offload operators and improve the end user experience.
+- While the addition of the auditor service helps track some changes on the input side of the catalog, there is no corresponding tracking of internal or end-user-visible output side changes. Arguably those can be equally important in exceptional circumstances to understand outcomes.
+
+### Goals
+
+<!--
+List the specific goals of the BEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+- The catalog should internally track important history events, at least including create/update/delete of entities (as seen on the public API side), and create/update/delete of locations. This tracking must be reliable including complex cases like cascading deletes, and persistent.
+- The catalog should be able to push those history events to the events backend, for easy consumption by other Backstage backend plugins or translation into notifications etc.
+- It should be an open possibility to later easily write integrations that push those history events to any other event system, such as PubSub or SQS.
+- An external system should be able to subscribe to history events via the REST API, to stream out events as they happen.
+- An external system should be able to read history events in a structured way (with filtering) to for example see the complete history of a given entity up until now.
+- The catalog may evict history events based on some conditions (event age, or entities being deleted, etc) to recover storage space.
+- History events should contain enough metadata to be usefully possible to correlate to things around the catalog. For example, events that are related to a given entity ID / entity ref / URL location should persist those to be able to easily request those events in a context.
+- History events should retain the actual historic shapes of the entity body itself, since this allows for more use cases of understanding what happened over time to an entity.
+- The events endpoints should be covered by permissions.
+
+### Non-Goals
+
+<!--
+What is out of scope for this BEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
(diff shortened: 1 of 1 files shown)


beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+What is out of scope for this BEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+- This BEP mainly covers the backend parts of the feature. End users of the catalog frontend plugin should be able to inspect the history events related to a given entity as a good initial use case, but that is not given as a required completion goal for this particular BEP.
+- Specific implementations for third party receiving event buses are not part of the initial plan.
+- The main target database is as always Postgres; the feature may or may not be fully functional for other engines.
+- There are many types of events and sources of events that could be conceived including custom ones in processors. But the initial scope is for create/update/remove events for entities and create/update/delete events for locations.
+- The initial BEP does not mention conveying credentials of those who take actions that lead to history events. This would be a valuable extension of the system though.
+- The BEP does not describe a way of opting out of the feature, for example through config. This could likely be solved, but at a complexity cost that the author does not think is worthwhile.
```
aramissennyeydd, line 88: Is there value in keeping this as a catalog module? Letting folks opt-in more than opt-out by installing it to existing implementation but maybe making it the default through create-app going forwards.

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+When someone reads from a busy subscription (which is waiting for an acknowledgement) or a subscription that is already at the end of current history, they can long-poll-block on that request to efficiently wait for when it's due for retrying the read. The underlying database mechanisms use Postgres `LISTEN`/`NOTIFY` to efficiently know when to unblock pending blocked readers.
+
+The events defined are named `entity_created`, `entity_updated`, `entity_deleted`, `location_created`, `location_updated`, and `location_deleted`. They always contain the event ID (as a stringified number to avoid overflow problems) and timestamp (as an ISO string with time zone), and type. In addition they may contain the entity ID, entity ref, entity body, location ID, and/or location ref, depending on whether they are null or not in the database.
```
aramissennyeydd, line 172: Would this support tracking relation updates as well? Thinking about https://github.com/backstage/backstage/pull/22130 / https://github.com/backstage/backstage/pull/22358.

aramissennyeydd (approved):
🎉 🚀

beps/0013-catalog-history/README.md
freben: I don't think so - at least not now.

Triggers on the relation table itself is not particularly feasible right now, I don't think. Just because of how they get batch deleted and re-inserted and it's not clear if that affects the end result or not (because of deduping etc).

And detecting it on the final_entities shape, well, I guess we sort of could if we parsed it out of the entity body and did some diffing but it'd have to be done in SQL and it'd be pretty hairy 🤔

beps/0013-catalog-history/README.md
freben: I worry about having a module add triggers, onto an external database model. It's a very hard connection formed between the origin tables and the triggers that act upon them so we would probably end up with a hard-to-evolve catalog. When upgrading packages I'm for example not sure that we could avoid ending up in situations where the catalog cannot run its migrations because they conflict with the ones added by the module, or vice versa.

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+| -------------- | ------------------------- | ----------------------------------------------------------------------------------------------------- |
+| `event_id`     | auto-incrementing big int | Auto generated ID                                                                                     |
+| `event_at`     | timestamp                 | Auto generated time when the event was generated                                                      |
+| `event_type`   | string                    | The distinct type of event                                                                            |
+| `entity_ref`   | string or null            | Affected entity ref, where applicable                                                                 |
+| `entity_id`    | string or null            | Affected entity UID, where applicable                                                                 |
+| `entity_json`  | long text or null         | The body of the affected entity, where applicable (e.g. its shape after updating, or before deletion) |
```
schultzp2020, line 134: What do you think about only containing the diff?

freben: Hm, we're constrained in the proposed design to whatever is possible to do in a SQL trigger. And the receivers of these events in the other end of the line, are likely to not already have a persisted previous shape of the entity.

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
+---
+title: Catalog History and Events
+status: provisional
```
Rugvip: Should go straight to implementable https://github.com/backstage/backstage/pull/27723

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Goals
+
+<!--
+List the specific goals of the BEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+- The catalog should internally track important history events, at least including create/update/delete of entities (as seen on the public API side), and create/update/delete of locations. This tracking must be reliable including complex cases like cascading deletes, and persistent.
+- The catalog should be able to push those history events to the events backend, for easy consumption by other Backstage backend plugins or translation into notifications etc.
```
Rugvip, line 67: If a REST API exists for consuming these events directly from the catalog, I think it would be better to avoid also sending the events to the events service. It'll essentially just duplicate all communication and storage, with a mix of different retention rules and consistency guarantees. I think it's likely that it will always be better to consume the events from the catalog API directly.

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+- It should be an open possibility to later easily write integrations that push those history events to any other event system, such as PubSub or SQS.
+- An external system should be able to subscribe to history events via the REST API, to stream out events as they happen.
```
Rugvip, line 69: A very important aspect of this imo is to be able to read the current state of the catalog and then start consuming updates in a safe way. That or for it to be reasonable to read from the start of the event stream and get a full view of the catalog. A potential option here could be a streaming mode that replicates #10115 in that it only returns the latest state for each entity, ignoring deletions that happened before the start of the stream.

Any thoughts on how either of those can be achieved?

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+- An external system should be able to read history events in a structured way (with filtering) to for example see the complete history of a given entity up until now.
+- The catalog may evict history events based on some conditions (event age, or entities being deleted, etc) to recover storage space.
```
Rugvip, line 71: Do you have a sense of how much storage space we're talking about? I'm kind of expecting that in practice the limiting factor is the time it takes to consume all events, rather than the storage itself. That is unless the catalog is a badly behaving one with a lot of rapid data changes.

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+- History events should contain enough metadata to be usefully possible to correlate to things around the catalog. For example, events that are related to a given entity ID / entity ref / URL location should persist those to be able to easily request those events in a context.
+- History events should retain the actual historic shapes of the entity body itself, since this allows for more use cases of understanding what happened over time to an entity.
```
Rugvip, line 73: The more I think about the overall capability the more I think that the best solution is one where you can choose whether you want to consume the full history with all updates, or just the latest update for each entity. Both are very useful for different use-cases, but the both for many of those use-cases the other variant is a bad fit.

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+- The internal state machine of the catalog is mostly a black box. It can be challenging to understand why things aren't progressing as expected, and operators regularly resort to digging through backend logs that may or may not contain relevant historical information about the situation. Those logs are also typically not available to entity owners. Making entity owners more self sufficient in being able to diagnose unexpected situations can offload operators and improve the end user experience.
+- While the addition of the auditor service helps track some changes on the input side of the catalog, there is no corresponding tracking of internal or end-user-visible output side changes. Arguably those can be equally important in exceptional circumstances to understand outcomes.
+
+### Goals
```
Rugvip, line 51: I think it would be useful to list some use-cases we're aiming for as well, rather than the capabilities added to the catalog.

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+Triggers `AFTER INSERT OR DELETE OR UPDATE OF final_entity ON final_entities` and `AFTER INSERT OR DELETE OR UPDATE ON locations` are added. These lead to the automatic `INSERT` of a corresponding new history event in the events table. At the same time, a `NOTIFY` signal is sent on the Postgres engine.
+
+Add a REST API to read from the history table directly, on `/history/v1/events`. This allows the consumption of events, optionally with filtering on the available columns above so that you could for example grab the complete history of a given entity, or events on a certain organization URL. This API optionally supports blocking long-poll reads so that you can reliably both read the currently existing history and then stream in whatever happens after that point on. The underlying database mechanisms use Postgres `LISTEN`/`NOTIFY` to efficiently know when to unblock pending blocked readers.
+
+The following is the proposed shape of the history subscriptions table.
+
+| column                       | type              | description                                                                                                                             |
```
camilaibs, line 151: Does it make sense to track the author of the event (could be a system actor like a processor or a human deleting an entity)?

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+| `entity_ref`   | string or null            | Affected entity ref, where applicable                                                                 |
+| `entity_id`    | string or null            | Affected entity UID, where applicable                                                                 |
+| `entity_json`  | long text or null         | The body of the affected entity, where applicable (e.g. its shape after updating, or before deletion) |
+| `location_id`  | string or null            | The registered location ID affected, where applicable                                                 |
+| `location_ref` | string or null            | The location affected, where applicable                                                               |
+
+Triggers `AFTER INSERT OR DELETE OR UPDATE OF final_entity ON final_entities` and `AFTER INSERT OR DELETE OR UPDATE ON locations` are added. These lead to the automatic `INSERT` of a corresponding new history event in the events table. At the same time, a `NOTIFY` signal is sent on the Postgres engine.
```
camilaibs, line 138: Does it make sense to support databases other than Postgres?

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+implementation.
+-->
+
+Define a set of semantic event types and their fields, for example `entity_deleted` which then should contain at least the ref, ID, and body of the entity that was deleted.
+
+Add a new history events table to the catalog database. The table has an auto-incrementing integer ID as the primary key which allows the ordered tracking of history.
+
+Add triggers to the `locations` and `final_entities` tables, and specifically to changes to the `final_entity` column in the latter. These triggers add history events accordingly, and send `PG_NOTIFY` signals.
+
+Add a history subscriptions table to the catalog database. This is essentially a place to globally track the progress and filter rules of individual subscribers. It also allows for acknowledgement-based locking, so that even distributed consumers are able to easily read from a subscription all together without synchronizing among themselves.
```
Rugvip, line 107: Not sure if this is desirable tbh, since we have the incrementing events ID it'd be a lot safer to use that as a pointer instead. With stateful subscriptions it's tricky to know whether the events where actually consumed properly and acknowledgements get a bit tricky too. I do think it's likely easier to coordinate the pointer on the reader side instead. Either way I'd label this as a potential future addition if we find that it's needed.

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+Add an internal history exporter to the catalog, which leverages the history subscriptions feature to "pump" entries from the history events table as they happen to the events backend where other parts of the Backstage ecosystem can consume them.
```
Rugvip: Probably quite easy to build one or more modules for this tbh, essentially just needs a single row with a pointer.

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+Add a REST API to make direct filtered reads of the history table. This API can have an optional "blocking" feature like the events backend does, to let consumers both just quickly read events as well as long poll for new ones as they happen.
```
Rugvip, line 111: 👍 

One thing I realized might be useful on top of this is to have some form of marker for when one is caught up with recent events. I'm thinking in particular if you're syncing with the catalog on startup, it's likely very useful to know when you've received the full catalog and switch over to "update" mode, knowing that you can start serving traffic at that point.

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+| `filter_entity_id`           | string or null    | Filter to only events pertaining to this entity UID (if specified at creation time)                                                     |
+
+Add a REST API to create, read, and acknowledge subscriptions, rooted on `/history/v1/subscriptions`.
+
+When subscriptions are initially created, you can choose whether to start from the start of history or from the end (now onward). `last_sent_event_id` and `last_acknowledged_event_id` are then set accordingly to `0` or the current highest available history event ID at that time.
+
+When someone reads from an idle subscription, they get sent a batch of the latest available history events (if any), along with a randomly generated acknowledgement ID. They have to send an acknowledgement request with that ID in a timely manner for the ID columns to be advanced and the subscription to return to idle. At that point any other active subscriber can be sent the next batch. Failure to acknowledge in a timely manner will lead to the batch being considered failed. Then the ID columns are _not_ advanced, which means that the same batch may be issued again to any other active reader.
```
Rugvip, line 168: > in a timely manner

👀

beps/0013-catalog-history/README.md
```diff
@@ -0,0 +1,216 @@
...
+- While the addition of the auditor service helps track some changes on the input side of the catalog, there is no corresponding tracking of internal or end-user-visible output side changes. Arguably those can be equally important in exceptional circumstances to understand outcomes.
+
+### Goals
+
+<!--
+List the specific goals of the BEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+- The catalog should internally track important history events, at least including create/update/delete of entities (as seen on the public API side), and create/update/delete of locations. This tracking must be reliable including complex cases like cascading deletes, and persistent.
```
Rugvip, line 66: What's the purpose of tracking locations?

freben: It contributes to the full picture. For a given repo, you see the locations being registered, the Location kind entities being spawned from them, the actual child component being processed and made. Then when your entity suddenly disappears unexpectedly, you check back and can see that the location was deleted and everything else flowed from that because of eager deletion. There's so much that we should track that helps an end user understand how, why, and when things happened.

beps/0013-catalog-history/README.md
freben: Yeah. There had to be some time aspect to it, to cover for consumers just disappearing mid-"transaction"

beps/0013-catalog-history/README.md
freben: Then you'd read in nonblocking mode (which never 202s) until you get no more data, and then start reading in blocking mode starting from the last event ID you saw.

beps/0013-catalog-history/README.md
freben: The subscriptions table is exactly that table! Anyone can use the subscriptions REST API to upsert a subscription with a fixed known ID and start consuming that, without needing any state of their own.

The way the implementation of the subscription API works, is that it only hands out a single batch of events to one single random consumer at a time, and then waits for their acknowledgement. That way you can autoscale consumers and not worry about concurrency problems because there's no overlapping delivery (as long as you are not extremely slow as a consumer leading to ack timeouts). This obviously trades speed for reliability. If you want to consume at maximum speed you use the non-subscription read API instead.

In the end, feeding history to the events backend system is such a common need that my prototype implementation does not even implement it in a separate module. Instead it is a builtin feature that can directly use the database layer (but still functionally identical, with a chosen subscription name etc).

beps/0013-catalog-history/README.md
freben: See [the other comment](https://github.com/backstage/backstage/pull/30146#discussion_r2269913004) - it's a way for distributed consumers to centrally agree and follow such a pointer, without having to build all of the primitives themselves to perform such agreement/control among themselves. Instead, anyone can with an absolute minimum of code just start reading from a subscription, and know that data will flow from whatever the last known good state was - in a distributed fashion among all of those who use the same subscription.

It's perfectly fine to make your own unique subscription instead of choosing a well known identifier, if that is what you want. You'll just get a uuid assigned to you and then you go on as usual.

I feel this particular feature of the API is a really powerful tool and it would be vastly more complex to consume reliably if it didn't exist. I expect the majority of services who want to consume history events and react to them, to use this exact mechanism. It's like with a pubsub subscription - you "just read from it" and it just works.

beps/0013-catalog-history/README.md
freben: My prototype implementation does support pg, mysql and sqlite actually. The latter two fall back to polling instead of notifications though.

beps/0013-catalog-history/README.md
freben: I would love to. This would require adding columns to _other_ tables too though, so that the info is there to pick up for the triggers.

This would make it slightly overlap with what the audit log service wants to do though. So it'll probably need some discussion as to whether it's worth doing.

beps/0013-catalog-history/README.md
freben: Hm. My prototype implementation has a little class that directly uses the database layer to lift over events to the events service bus. I thought it might be a convenient thing to have - and it could be made possible to enable/disable with a config flag.

Note that we already have for example `@backstage/plugin-events-backend-module-google-pubsub` which is able to pump backstage events to pubsub. If you deploy that on the same nodes as the catalog itself, you get a very easy-to-setup in-process driver for sending catalog history events to pubsub. If you don't do that, you essentially have to make a new module that uses the catalog history API, and reinvents the pushing semantics of the above mentioned pubsub module. Seemed to me that being able to leverage the already existing mechanics was useful.

beps/0013-catalog-history/README.md
freben: Hm. To properly support that we probably need to add a "rollup" table where for example the latest event ID per entity ref is stored, and have a dedicated API for safely atomically getting a snapshot of that that cannot change across pagination or so. And that would only cover entities then, not other types of event. This is tricky, how would you conceive supporting that truly safely?

Or maybe I misunderstand. Thinking of it more, getting just the latest update for each entity is basically to call the old `/entities` endpoint :) Did you mean, to get all updates that have happened since some previously known point in time or event ID, ignoring duplicate updates to identical entities? What were the query parameters you envisioned?

beps/0013-catalog-history/README.md
freben: updated

beps/0013-catalog-history/README.md
freben: Updated these a bit defensively

beps/0013-catalog-history/README.md
freben: `final_entities` is less than 14% of our current database size. The vast majority of that table is occupied by the `final_entity` column of course.

Adding this feature with _no_ entities ever changing, would add another 14% to the size due to the `entity_created` history events. And yes, fast-changing entities would be "risky" in that sense. Using compression for the database would probably be very successful here, but it's likely that only a minority are using compression.

The history janitor settings are probably very relevant here.

beps/0013-catalog-history/README.md
freben: That does sound like an extra rollup table. One that maps entity refs (if that is even good enough?) to last-seen event IDs and whatever state that corresponds to.

I wonder how the user would get access to that data though, in the initial "get me the current state of things" read. The entity bodies don't have room for a `metadata.generation` and we can't really dynamically insert one at query time without costing a lot of performance. `/entities/by-query` has an `items` field at the root of the response; maybe it could have a `generations: string[]` too with the same length as the items array, and you have to ask for it to get it?

Then you can make a new query again at a later time with `ifNewerGenerationThan=x` filter parameter to the query. But the question is how it would represent deletions. A null in the items array and a deletion marker in the `generations` array?

Maybe it's better with a dedicated history API at that point, leaving the existing ones unchanged, even if it means some level of duplication of functionality.

I like this whole discussion though. The use case is worth catering to for sure. A rollup sounds like it might be great to add. That way it's easier to treat the actual history as transient, so that only retaining it for X days isn't really as big of a deal because some of the important mirroring use cases are covered by the rollup either way.

freben: Alright - added a summary table too, this actually turned out to be very easy, just gotta make a nice delta consumption API and I think this one is golden

beps/0013-catalog-history/README.md
freben: Made the emitter configurable and default disabled

beps/0013-catalog-history/README.md
freben: Summary table added!

awanlin:
Hi @freben, should this be closed?

freben:
No - still hoping to get some time to settle this

awanlin:
Hi @freben, should we label this with `no-stale`?

PR #30153, opened 2025-06-05
remove some usages of the old alpha catalog service

freben:
There are very few left; one is in the scaffolder and that's being addressed, and one in techdocs that currently needs https://github.com/backstage/backstage/pull/30127


Rugvip (approved):
👍, but with changesets

PR #30157, opened 2025-06-05
feat: allow streaming catalog entities

drodil:

Catalog API and Catalog Service now includes a `streamEntities` method that allows for streaming entities from the catalog.
This method is designed to handle large datasets efficiently by processing entities in a stream rather than loading them
all into memory at once. This is useful when you need to fetch a large number of entities but do not want to use pagination
or fetch all entities at once.

### DIFF
--- a/.changeset/olive-moons-burn.md
+++ b/.changeset/olive-moons-burn.md
@@ -4,7 +4,7 @@
 '@backstage/plugin-catalog-node': minor
 ---
 
-Introduced new `streamEntities` and `streamEntityPages` async generator methods for the catalog.
+Introduced new `streamEntities` async generator method for the catalog.
 
 Catalog API and Catalog Service now includes a `streamEntities` method that allows for streaming entities from the catalog.
 This method is designed to handle large datasets efficiently by processing entities in a stream rather than loading them
@@ -14,22 +14,11 @@ or fetch all entities at once.
 Example usage:
 
 ```ts
-const stream = catalogClient.streamEntities({}, { token });
-for await (const entity of stream) {
-  // Handle entity
-}
-```
-
-Additionally, a `streamEntityPages` method is available that streams entities in pages, allowing for batch processing of entities.
-This method is more efficient than `streamEntities` when you can process entities in chunks.
-Example usage:
-
-```ts
-const pageStream = catalogClient.streamEntityPages(
-  { pageSize: 100 },
-  { token },
-);
+const pageStream = catalogClient.streamEntities({ pageSize: 100 }, { token });
 for await (const page of pageStream) {
   // Handle page of entities
+  for (const entity of page) {
+    console.log(entity);
+  }
 }
 ```
--- a/packages/catalog-client/src/CatalogClient.ts
+++ b/packages/catalog-client/src/CatalogClient.ts
@@ -464,21 +464,6 @@ export class CatalogClient implements CatalogApi {
   async *streamEntities(
     request?: StreamEntitiesRequest,
     options?: CatalogRequestOptions,
-  ): AsyncIterable<Entity> {
-    const pages = this.streamEntityPages(request, options);
-    for await (const page of pages) {
-      for (const entity of page) {
-        yield entity;
-      }
-    }
-  }
-
-  /**
-   * {@inheritdoc CatalogApi.streamEntityPages}
-   */
-  async *streamEntityPages(
-    request?: StreamEntitiesRequest,
-    options?: CatalogRequestOptions,
   ): AsyncIterable<Entity[]> {
     let cursor: string | undefined = undefined;
     const limit = request?.pageSize ?? DEFAULT_STREAM_ENTITIES_LIMIT;
--- a/packages/catalog-client/src/types/api.ts
+++ b/packages/catalog-client/src/types/api.ts
@@ -710,20 +710,6 @@ export interface CatalogApi {
 
   /**
    * Asynchronously streams entities from the catalog. Uses `queryEntities`
-   * to fetch entities in batches, and yields them one by one.
-   *
-   * @public
-   *
-   * @param request - Request parameters
-   * @param options - Additional options
-   */
-  streamEntities(
-    request?: StreamEntitiesRequest,
-    options?: CatalogRequestOptions,
-  ): AsyncIterable<Entity>;
-
-  /**
-   * Asynchronously streams entity pages from the catalog. Uses `queryEntities`
    * to fetch entities in batches, and yields them one page at a time.
    *
    * @public
@@ -731,7 +717,7 @@ export interface CatalogApi {
    * @param request - Request parameters
    * @param options - Additional options
    */
-  streamEntityPages(
+  streamEntities(
     request?: StreamEntitiesRequest,
     options?: CatalogRequestOptions,
   ): AsyncIterable<Entity[]>;
--- a/packages/catalog-client/report-testUtils.api.md
+++ b/packages/catalog-client/report-testUtils.api.md
@@ -71,9 +71,7 @@ export class InMemoryCatalogClient implements CatalogApi {
   // (undocumented)
   removeLocationById(_id: string): Promise<void>;
   // (undocumented)
-  streamEntities(request?: StreamEntitiesRequest): AsyncIterable<Entity>;
-  // (undocumented)
-  streamEntityPages(request?: StreamEntitiesRequest): AsyncIterable<Entity[]>;
+  streamEntities(request?: StreamEntitiesRequest): AsyncIterable<Entity[]>;
   // (undocumented)
   validateEntity(
     _entity: Entity,
--- a/packages/catalog-client/report.api.md
+++ b/packages/catalog-client/report.api.md
@@ -91,10 +91,6 @@ export interface CatalogApi {
   streamEntities(
     request?: StreamEntitiesRequest,
     options?: CatalogRequestOptions,
-  ): AsyncIterable<Entity>;
-  streamEntityPages(
-    request?: StreamEntitiesRequest,
-    options?: CatalogRequestOptions,
   ): AsyncIterable<Entity[]>;
   validateEntity(
     entity: Entity,
@@ -179,10 +175,6 @@ export class CatalogClient implements CatalogApi {
     options?: CatalogRequestOptions,
   ): Promise<void>;
   streamEntities(
-    request?: StreamEntitiesRequest,
-    options?: CatalogRequestOptions,
-  ): AsyncIterable<Entity>;
-  streamEntityPages(
     request?: StreamEntitiesRequest,
     options?: CatalogRequestOptions,
   ): AsyncIterable<Entity[]>;
--- a/packages/catalog-client/src/CatalogClient.test.ts
+++ b/packages/catalog-client/src/CatalogClient.test.ts
@@ -581,15 +581,6 @@ describe('CatalogClient', () => {
 
     it('should stream entities', async () => {
       const stream = client.streamEntities({}, { token });
-      const results: Entity[] = [];
-      for await (const entity of stream) {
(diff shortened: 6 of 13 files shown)


drodil:
Not stale, something to discuss

.changeset/olive-moons-burn.md
```diff
@@ -0,0 +1,21 @@
+---
+'@backstage/catalog-client': patch
+'@backstage/plugin-catalog-react': patch
+'@backstage/plugin-catalog-node': patch
```
vinzscam: I think these need to be minor 

```suggestion
'@backstage/catalog-client': minor
'@backstage/plugin-catalog-react': minor
'@backstage/plugin-catalog-node': minor
```

drodil:
@vinzscam I am on holiday so can't really do any changes to this PR. Any chance you could finish it up with the changeset changes? Thanks!

vinzscam:
> @vinzscam I am on holiday so can't really do any changes to this PR. Any chance you could finish it up with the changeset changes? Thanks!

We'll figure it out. Enjoy! 🌴

drodil:
@vinzscam thanks a lot! 😎

drodil:
Not stale

drodil:
@vinzscam ready for another round, rebased and fixed the changeset.

drodil:
Not stale

packages/catalog-client/src/CatalogClient.ts
```diff
@@ -49,6 +50,9 @@ import type {
   AnalyzeLocationResponse,
 } from '@backstage/plugin-catalog-common';
 
+// Number of entities to return in a single streamEntities request
+const STREAM_ENTITIES_LIMIT = 500;
```
vinzscam: @freben can we suggest any "magic number" here? I remember we ran some benchmarks internally

vinzscam (changes requested):
sorry for the long wait 🙈
It looks great, let's figure out the magic number for the limit and we can 🚢

packages/catalog-client/src/CatalogClient.ts
freben: Hm I don't think I have the numbers at hand. This seems like a reasonable compromise number!

freben:
@drodil do you have any opinion on having entity vs entity array (chunk) as the iterable element? No performance worries for large iterations? I'm not sure if hundreds of thousands of extra ticks might be spent?

packages/catalog-client/src/CatalogClient.ts
drodil: We could of course add it as optional parameter `batchSize` to the `StreamEntitiesRequest` and use this (or other magic number) as default. Now that I think of it, 500 might be quite high for default 🤔

drodil:
> @drodil do you have any opinion on having entity vs entity array (chunk) as the iterable element? No performance worries for large iterations? I'm not sure if hundreds of thousands of extra ticks might be spent?

We've been using single-entity streaming successfully without performance hiccups. But of course, it would make sense to benchmark this.

packages/catalog-client/src/CatalogClient.ts
drodil: I added the `batchSize` parameter to the request but kept the default as 500

drodil:
> @drodil do you have any opinion on having entity vs entity array (chunk) as the iterable element? No performance worries for large iterations? I'm not sure if hundreds of thousands of extra ticks might be spent?

I added support for both modes, the single being the default. Is this too fancy? :D

drodil:
Tried to do some benchmarking with this:

```ts
import { createBackendModule } from '@backstage/backend-plugin-api';
import {
  mockServices,
  startTestBackend,
  TestDatabases,
} from '@backstage/backend-test-utils';
import { CatalogApi, CatalogClient } from '@backstage/catalog-client';
import { catalogProcessingExtensionPoint } from '@backstage/plugin-catalog-node/alpha';
import { Knex } from 'knex';
import { default as catalogPlugin } from '../..';
import {
  SyntheticLoadEntitiesProcessor,
  SyntheticLoadEntitiesProvider,
  SyntheticLoadOptions,
} from './lib/catalogModuleSyntheticLoadEntities';
import { describePerformanceTest, performanceTraceEnabled } from './lib/env';
import fs from 'fs-extra';
import * as v8Profiler from 'v8-profiler-next';

v8Profiler.setGenerateType(1);

// #region Helpers

jest.setTimeout(600_000);

const databases = TestDatabases.create({
  ids: [/* 'MYSQL_8', */ 'POSTGRES_17', /* 'POSTGRES_13',*/ 'SQLITE_3'],
  disableDocker: false,
});

const traceLog: typeof console.log = performanceTraceEnabled
  ? console.log
  : () => {};

async function createBackend(
  knex: Knex,
  load: SyntheticLoadOptions,
): Promise<{
  client: CatalogApi;
  numberOfEntities: number;
  stop: () => Promise<void>;
}> {
  const numberOfEntities =
    load.baseEntitiesCount + load.baseEntitiesCount * load.childrenCount;

  traceLog(`Creating test backend with ${numberOfEntities} entities`);

  const backend = await startTestBackend({
    features: [
      catalogPlugin,
      mockServices.database.factory({ knex }),
      createBackendModule({
        pluginId: 'catalog',
        moduleId: 'synthetic-load-entities',
        register(reg) {
          reg.registerInit({
            deps: {
              catalog: catalogProcessingExtensionPoint,
            },
            async init({ catalog }) {
              catalog.addEntityProvider(
                new SyntheticLoadEntitiesProvider(load, {}),
              );
              catalog.addProcessor(new SyntheticLoadEntitiesProcessor(load));
            },
          });
        },
      }),
    ],
  });

  while (
    (await knex('final_entities')
      .count({ count: '*' })
      .whereNotNull('final_entity')
      .then(rows => Number(rows[0].count))) !== numberOfEntities
  ) {
    await new Promise(resolve => setTimeout(resolve, 200));
  }

  const client = new CatalogClient({
    discoveryApi: {
      getBaseUrl: jest
        .fn()
        .mockResolvedValue(
          `http://localhost:${backend.server.port()}/api/catalog`,
        ),
    },
  });

  return {
    client,
    numberOfEntities,
    stop: backend.stop.bind(backend),
  };
}

describePerformanceTest('streamEntities', () => {
  let knex: Knex;
  let backend: Awaited<ReturnType<typeof createBackend>>;

  describe.each(databases.eachSupportedId())(
    'stream entities, %p',
    databaseId => {
      beforeAll(async () => {
        knex = await databases.init(databaseId);
        backend = await createBackend(knex, {
          baseEntitiesCount: 2000,
          baseRelationsCount: 3,
          baseRelationsSkew: 0.3,
          childrenCount: 3,
        });
      });

      beforeEach(() => {
        const title = expect.getState().currentTestName;
        v8Profiler.startProfiling(title, true);
      });

      afterEach(() => {
        const title = expect.getState().currentTestName;
        const profile = v8Profiler.stopProfiling(title);
        profile.export((error, result: any) => {
          fs.writeFileSync(
            `${title.replaceAll(' ', '-').replaceAll('"', '')}.cpuprofile`,
            result,
          );
          profile.delete();
        });
      });

      afterAll(async () => {
        await backend.stop();
        await knex.destroy();
      });

      it('one by one', async () => {
        const stream = await backend.client.streamEntities();
        const entities = [];
        for await (const entity of stream) {
          entities.push(entity);
        }
        expect(entities).toHaveLength(backend.numberOfEntities);
      });

      it('by array', async () => {
        const stream = await backend.client.streamEntities({ mode: 'array' });
        const entities = [];
        for await (const entity of stream) {
          entities.push(...entity);
        }
        expect(entities).toHaveLength(backend.numberOfEntities);
      });
    },
  );
});
```

But it didn't really differ much in profiles. Not sure if the performance test is valid though :D

drodil:
Also made this that helps getting the profiles out of performance tests (or why not from other ones too):
https://github.com/backstage/backstage/pull/31017

packages/catalog-client/src/CatalogClient.ts
```diff
@@ -456,6 +460,34 @@ export class CatalogClient implements CatalogApi {
     return response.json() as Promise<AnalyzeLocationResponse>;
   }
 
+  /**
+   * {@inheritdoc CatalogApi.streamEntities}
+   */
+  async *streamEntities<
+    T extends StreamEntitiesRequest,
+    R = T extends { mode: 'array' } & StreamEntitiesRequest ? Entity[] : Entity,
```
vinzscam: question: Shall we stick to single mode for now? Do you have a use case in mind for `mode: 'array'`?

drodil: Good question, @freben you think there should be support for getting the entities as batches?

freben: I think the numbers kind of speak for having only single-item mode and living with that. It's not so much a use case, more a performance question. And the difference seemed to be pretty small.

I mean, one could argue the other way around that forcing people to write a double `for` loop really is a pretty minimal cost to pay for end users and sidesteps the question while enabling maximum performance potential. But I like the simpler item interface, even if people end up very often doing `Array.from()` on it.

🤔 Hm. This is all a matter of taste at this point.

I vote: Let's go for just single items, and then in the future we can still make this mode addition if we feel that it's necessary. Which maybe it won't be.

freben: Backing up a bit, sorry for the noise.

Seems that there are several hundred milliseconds of overhead for large iterations if you wanted an array shaped output in the end. Maybe this actually does make a lot of sense with the mode parameter after all.

Could we make it be `mode?: 'single' | 'pages' | undefined` instead? Only because `'array'` sounds like all of them all at once, not several pages. Or `'batches'` - just that the naming should match the `batchSize`/`pageSize` nomenclature consistently

Maybe even `streamEntities` and `streamEntityPages` as two methods is a worthy distinction to keep the interface plain and simple in all of the places it's used, no type wrangling or mimicking the way it has to handle several use cases.

drodil: Added the `streamEntityPages` method now 👍

packages/catalog-client/src/types/api.ts
```diff
@@ -465,6 +465,21 @@ export type QueryEntitiesResponse = {
   };
 };
 
+/**
+ * Stream entities request for {@link CatalogClient.streamEntities}.
+ *
+ * @public
+ */
+export type StreamEntitiesRequest = Omit<
+  QueryEntitiesRequest,
```
freben:
```suggestion
  QueryEntitiesInitialRequest,
```

i think this is more correct

packages/catalog-client/src/types/api.ts
```diff
@@ -465,6 +465,21 @@ export type QueryEntitiesResponse = {
...
+  'limit' | 'offset'
+> & {
+  /**
+   * The number of entities to fetch in each batch. Defaults to 500.
+   */
+  batchSize?: number;
```
freben:
```suggestion
   * The number of entities to fetch in each page. Defaults to 500.
   */
  pageSize?: number;
```

does that make sense? to match the `streamEntityPages` name.

The other alternative is to name the method `streamEntitiesBatched`

packages/catalog-client/src/CatalogClient.ts
vinzscam: Good catch. I would vote for having just the `streamEntities` method, using the code from the `streamEntityPages` method. I've checked around, and this seems to be a pattern others are using as well.
WDYT?

freben: I could go for that too. Very low convenience cost for the user, just one method to override, and no foot guns. Maybe it's the reasonable choice.

drodil: Made the change, please check it out

vinzscam (approved):
🚢

Issue #30156, opened 2025-06-05
Modules providing extension point gets initialized before all modules depending on it


lucasmdrs:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
"module extension", "dependency graph", "banckend initialization", "extension point"
```

### 🗃️ Project Area

Core Framework

### 🔗 External Integration

N/A

### 📝 Description & Context

When adding multiple modules that depend on the [incrementalIngestionProvidersExtensionPoint](https://github.com/backstage/backstage/blob/a58e48e983224686a9ef7862d6b29976dd202732/plugins/catalog-backend-module-incremental-ingestion/src/module/index.ts#L19C3-L19C46) to the backend, not all providers get registered before the `incremental-ingestion-entity-provider` gets initialized. 

This might be related to the [BackendInitializer dependency graph](https://github.com/backstage/backstage/blob/a58e48e983224686a9ef7862d6b29976dd202732/packages/backend-app-api/src/wiring/BackendInitializer.ts#L348), I've noticed that when only a single module that depends on the extension point is added, the order is respected regardless of the time it takes to initialize it, but when more than one module that depends on the extension point are added, the after the first module gets initialized, all the others race against the `incremental-ingestion-entity-provider` initialization itself.

### 👍 Expected Behavior

All modules that depend on a module extension point gets initialized before the module providing the extension point without a race condition.

### 📦 Reproduction Repo

https://github.com/lucasmdrs/backstage-module-init-bug

### 🥾 Reproduction steps

See the repository README.md

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Undecided

freben:
Thank you, awesome that you have a reproduction repo added as well, that really helps! I'll add some labels to begin with and let's try to find the cause of this.

freben:
Alright, I made https://github.com/backstage/backstage/pull/30159 thanks to your great report here, we'll consider making a patch out of it.

PR #30159, opened 2025-06-05
fix module initialization order bug

freben:
Fixes https://github.com/backstage/backstage/issues/30156

### DIFF
--- a/packages/backend-app-api/src/lib/DependencyGraph.ts
+++ b/packages/backend-app-api/src/lib/DependencyGraph.ts
@@ -188,11 +188,24 @@ export class DependencyGraph<T> {
     fn: (value: T) => Promise<TResult>,
   ): Promise<TResult[]> {
     const allProvided = this.#allProvided;
-    const producedSoFar = new Set<string>();
     const waiting = new Set(this.#nodes.values());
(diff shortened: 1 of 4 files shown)


packages/backend-app-api/src/lib/DependencyGraph.ts
```diff
@@ -188,12 +188,25 @@ export class DependencyGraph<T> {
...
   ): Promise<TResult[]> {
     const allProvided = this.#allProvided;
-    const producedSoFar = new Set<string>();
     const waiting = new Set(this.#nodes.values());
     const visited = new Set<Node<T>>();
     const results = new Array<TResult>();
     let inFlight = 0; // Keep track of how many callbacks are in flight, so that we know if we got stuck
 
+    // This keeps track of a counter of how many providers there are still left
+    // to be visited for each depdendency. This needs to be a counter instead of
```
aramissennyeydd:
```suggestion
    // to be visited for each dependency. This needs to be a counter instead of
```

aramissennyeydd (approved):
Nice! Interesting edge case

benjdlambert (approved):
Looks good to me, probably want a peek from @Rugvip also :pray:

PR #30166, opened 2025-06-06
Migrate typography to canon in search

birdhb:

In the Search plugin and its child components, all Typography components should be migrated to use Heading or Text components from Canon. 

### DIFF
--- a/plugins/search-react/src/components/DefaultResultListItem/DefaultResultListItem.test.tsx
+++ b/plugins/search-react/src/components/DefaultResultListItem/DefaultResultListItem.test.tsx
@@ -19,23 +19,9 @@ import { renderInTestApp } from '@backstage/test-utils';
 import FindInPageIcon from '@material-ui/icons/FindInPage';
 import { DefaultResultListItem } from './DefaultResultListItem';
 
-describe('DefaultResultListItem', () => {
-  beforeAll(() => {
-    Object.defineProperty(window, 'matchMedia', {
-      writable: true,
-      value: jest.fn().mockImplementation(query => ({
-        matches: false,
-        media: query,
-        onchange: null,
-        addListener: jest.fn(),
-        removeListener: jest.fn(),
-        addEventListener: jest.fn(),
-        removeEventListener: jest.fn(),
-        dispatchEvent: jest.fn(),
-      })),
-    });
-  });
+jest.mock('@backstage/canon/src/hooks/useMediaQuery');
 
+describe('DefaultResultListItem', () => {
   const result = {
     title: 'title',
     text: 'text',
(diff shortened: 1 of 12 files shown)


plugins/search-react/src/components/DefaultResultListItem/DefaultResultListItem.test.tsx
```diff
@@ -20,6 +20,22 @@ import FindInPageIcon from '@material-ui/icons/FindInPage';
 import { DefaultResultListItem } from './DefaultResultListItem';
 
 describe('DefaultResultListItem', () => {
+  beforeAll(() => {
```
djamaile: if you export `useMediaQuery` from canon, you can just mock the hook itself 👍

plugins/search-react/src/components/DefaultResultListItem/DefaultResultListItem.test.tsx
```diff
@@ -19,6 +19,8 @@ import { renderInTestApp } from '@backstage/test-utils';
 import FindInPageIcon from '@material-ui/icons/FindInPage';
 import { DefaultResultListItem } from './DefaultResultListItem';
 
+jest.mock('@backstage/canon/src/hooks/useMediaQuery');
```
Rugvip, line 22: This deep mock will not be available to projects outside of this repo. We have [`mockBreakpoint`](https://github.com/backstage/backstage/blob/d8cf6837ae4d2281c44d89e9b2a06716274e990c/packages/core-components/src/testUtils.ts#L81) in `@backstage/core-components`, perhaps we could add something similar to `@backstage/canon`? (👀 @cdedreuille)

plugins/search-react/src/components/SearchResultList/SearchResultList.test.tsx
```diff
@@ -54,6 +54,19 @@ describe('SearchResultList', () => {
 
   beforeEach(() => {
     jest.clearAllMocks();
+    Object.defineProperty(window, 'matchMedia', {
```
Rugvip, line 57: This is effectively the same problem as the `jest.mock`s right?

plugins/search/src/components/SearchType/SearchType.Accordion.tsx
```diff
@@ -147,9 +147,7 @@ export const SearchTypeAccordion = (props: SearchTypeAccordionProps) => {
 
   return (
     <Box>
-      <Typography variant="body2" component="h2">
-        {name}
-      </Typography>
+      <Text variant="body">{name}</Text>
```
Rugvip, line 150: The removal of `h2` here seems like it might hurt accessibility, is there an equivalent option available? Or maybe it was never necessary to begin with.

Rugvip (changes requested):
Sorry we haven't yet made the decision to start migrating anything in this project to Canon. It's up for discussion but until that decision is made this will have to be on hold.

That said I gave it a look as there might be some learnings here to help move Canon forward.

plugins/search/src/components/SearchType/SearchType.Accordion.tsx
cdedreuille: Yes you can use the render prop to render it as h2.

plugins/search-react/src/components/DefaultResultListItem/DefaultResultListItem.test.tsx
cdedreuille: Yes agreed that would be a good idea. The more we can make tests easier to build with Canon the better :)

Issue #30079, opened 2025-05-28
GitLab URL Reader Confuses Integration and OAuth Tokens, Causing 401s


elaine-mattos:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
gitlab
gitlab url reader
fetch:plain:file
```

### 🗃️ Project Area

Software Templates

### 🔗 External Integration

GitLab

### 📝 Description & Context

Hi!

After we upgraded our instance to version 1.39.0, a partner team reported that their templates using the `fetch:plain:file` action were no longer working. After digging in, I found the issue is with how the GitLab OAuth token is being passed through the URL reader.

In our setup, we use GitLab OAuth tokens to authenticate requests in our scaffolder templates. But in the current implementation of `GitlabUrlReader`, the OAuth token (`token`) is being spread into the integration config [here](https://github.com/backstage/backstage/blob/master/packages/backend-defaults/src/entrypoints/urlReader/lib/GitlabUrlReader.ts#L344-L345), which causes it to be treated as the integration token (`config.token`) instead.

Later, in `getGitLabRequestOptions` [here](https://github.com/backstage/backstage/blob/master/packages/integration/src/gitlab/core.ts#L68-L73), the OAuth token (`token`) is expected to be passed explicitly as a second argument so it can determine the right header (`Authorization: Bearer` vs `PRIVATE-TOKEN`). Since that’s not happening, it falls back to using the integration token and incorrectly sends the request with a PRIVATE-TOKEN header.

Because of that, calls that should be authenticated using a Bearer token (OAuth) are instead sent using a PRIVATE-TOKEN header, which results in 401 Unauthorized errors from GitLab.

This may be related to this change: https://github.com/backstage/backstage/pull/29707

### Proposed Fix (tested locally, not yet submitted)

I've tested a local fix that resolves the issue by doing the following:

- Updated `getGitLabFileFetchUrl` to accept the `token` as a third parameter.
- Updated `getProjectId` and `getGitLabRequestOptions` to forward the token correctly.
- Adjusted `GitlabUrlReader.getGitlabFetchUrl` to pass the token explicitly instead of merging it into the config.

This ensures the OAuth token is passed explicitly and interpreted as a **Bearer token**, not incorrectly as a `PRIVATE-TOKEN`.

[...]

PR #30170, opened 2025-06-06
Fix(Integration - GitLab): use authorization header for GitLab requests

CptnFizzbin:
Follow up to #29707, since the fix that I thought would work for us didn't *quite* work as expected. this fix updates the request headers for GitLab to send all access tokens via the Authorization header. I was able to test a similar patch in our copy of backstage and found that this should work as expected.

Potential fix for #30079 as well


CptnFizzbin:
this can probably be merged with/alongside #30136

benjdlambert (approved):
Ok let's give this a go and test it out :pray: :tada: I'm a little concerned that this could break older installations, but maybe it's fine as it's public GitLab.

CptnFizzbin:
> Ok let's give this a go and test it out :pray: :tada: I'm a little concerned that this could break older installations, but maybe it's fine as it's public GitLab.

Looks like the docs for 16.11 also mention that tokens can be passed via the Authorization header, so this should be supported for the last 2 major versions of GitLab

PR #30183, opened 2025-06-09
fix(Scaffolder): filter MultiEntityPicker options based on rendered option values

koalaty-code:

This change updates the behavior of the MultiEntityPicker to match the EntityPicker. Currently, the MultiEntityPicker filters values based on the underlying entityRef even when the EntityPresentationApi uses a different value to render the option (e.g. `spec.profile.displayName`, or `metadata.title`). This leads to confusion as the rendered option labels may not match the value typed in the input.

### DIFF
--- a/.changeset/empty-mugs-agree.md
+++ b/.changeset/empty-mugs-agree.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder': patch
+---
+
+Filter MultiEntityPicker options based on rendered option value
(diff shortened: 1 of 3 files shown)


.changeset/empty-mugs-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder': patch
```
schultzp2020, line 2: @freben do we consider this behavior a feature or a bug?

koalaty-code: Bump. I'd like to get this merged in time for the June release. I chose "bug" for this as the MultiEntityPicker did not function the same as the EntityPicker.

freben (approved):
I'm happy to frame it as a bug, since it's inconsistent with the other control. Thanks!

PR #30188, opened 2025-06-10
🧹  import fixes

freben:
Triggered by https://github.com/backstage/backstage/pull/30186


schultzp2020:
@freben Is it possible to backport this to 1.39?

freben:
Hm how come?

drodil:
There are another 100+ things to fix if you run the `madge` without types and async imports excluded. But that's probably for another PR

Edit: Added those in a comment to the #30186

schultzp2020:
> Hm how come?

We're currently blocked from adopting 1.39 because of the following error. I haven't confirmed this PR fixes the error, but I have a hunch it does. 


```
(node:79340) Warning: Accessing non-existent property 'dynamicPluginsServiceRef' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:79340) [UNDICI-EHPA] Warning: EnvHttpProxyAgent is experimental, expect them to change at any time.
Loading config from MergedConfigSource{FileConfigSource{path="/Users/bgriggs/backstage/backstage-showcase/app-config.yaml"}, FileConfigSource{path="/Users/bgriggs/backstage/backstage-showcase/app-config.local.yaml"}, EnvConfigSource{count=0}}
/Users/bgriggs/backstage/backstage-showcase/node_modules/@backstage/backend-app-api/src/wiring/ServiceRegistry.ts:179
          Object.values(factory.deps).map(d => d.id),
```

PR #30189, opened 2025-06-10
fix(docs): Correct PostgreSQL environment variable injection in Kubernetes guide

Miaoxiang-philips:

I fully agree with the idea in #28193, but there seems to be some issue - we shouldn't add `POSTGRES_HOST` and `POSTGRES_PORT` in the Postgres deployment, that doesn't make sense.

If I do it as described in the current document, I will get an error:

```
│ {"level":"error","message":"Plugin 'app' threw an error during startup, waiting for 8 other plugins to finish before shutting down the process. Failed to instantiate service 'cor │
│ e.httpRouter' for 'app' because the factory function threw an error, Error: Failed to instantiate service 'core.auth' for 'app' because the factory function threw an error, Error │
│ : Failed to connect to the database to make sure that 'backstage_plugin_app' exists, AggregateError","service":"backstage","stack":"Error: Failed to instantiate service 'core.htt │
│ pRouter' for 'app' because the factory function threw an error, Error: Failed to instantiate service 'core.auth' for 'app' because the factory function threw an error, Error: Fai │
│ led to connect to the database to make sure that 'backstage_plugin_app' exists, AggregateError\n    at /app/node_modules/@backstage/backend-app-api/dist/wiring/ServiceRegistry.cj │
...
```
Kubernetes doesn't automatically inject `POSTGRES_HOST` and `POSTGRES_PORT`. What Kubernetes actually injects are `POSTGRES_SERVICE_HOST` (where POSTGRES is the Service name). If your service name is `postgres-svc`, then it will inject `POSTGRES_SVC_SERVICE_HOST` and `POSTGRES_SERVICE_PORT`.

Here is the official documentation explanation of Kubernetes: [Environment variables](https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables)

That's why I removed that section about injected variables([L349](https://github.com/backstage/backstage/blob/master/docs/deployment/k8s.md?plain=1#L394)). Of course, I also want to discuss which injection method the Backstage Team prefers - there are actually many options:
- Automatic injection (`POSTGRES_SERVICE_HOST`)
- Specified in secrets
- Statically specified in env

I chose to statically specify in env here because I think this is a simple example where simplicity is key - it helps beginners understand quickly.


Miaoxiang-philips:
Hey @Rugvip , Are there any comments?

PR #30191, opened 2025-06-10
Allow resolveSlackId to retrieve users by email in Slack notification plugin.

JeremyVanderMotte:

- resolveSlackId now accepts also an email input.
- findSlackIdByEmail updated to use directly the email if present.


benjidotsh:
Not stale. @backstage/notifications-maintainers can you guys take a look at this? 😇

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -334,6 +334,10 @@ export class SlackNotificationProcessor implements NotificationProcessor {
     // First try to get Slack ID from annotations
     const slackId = entity.metadata?.annotations?.[ANNOTATION_SLACK_BOT_NOTIFY];
     if (slackId) {
+      // If slack id is an email, look up for the id
+      if (slackId.includes('@')) {
```
freben, line 338: Can slack IDs for users START with an at-sign too? Does the check need to be like if `slackId.trim().indexOf('@') > 0`?

JeremyVanderMotte: Slack Id for a user is formatted as : UXXXXXXXX.
The @ sign is to ping someone in a channel.

kunickiaj:
This is already supported without changes. From the [docs](https://backstage.io/docs/notifications/processors#entity-requirements):

> Entities must be annotated with the following annotation:
> 
> slack.com/bot-notify
>
> The value may be any Slack ID supported by chat.postMessage, for example a user (U12345678), channel (C12345678), group, or direct message chat.
> 
> It's also possible to use a user's email address or channel name, however IDs are recommended by Slack. Private channels/chats must use an ID.

JeremyVanderMotte:
> This is already supported without changes. From the [docs](https://backstage.io/docs/notifications/processors#entity-requirements):
> 
> > Entities must be annotated with the following annotation:
> > slack.com/bot-notify
> > The value may be any Slack ID supported by chat.postMessage, for example a user (U12345678), channel (C12345678), group, or direct message chat.
> > It's also possible to use a user's email address or channel name, however IDs are recommended by Slack. Private channels/chats must use an ID.

I indeed saw that in the documentation and tried to use it but it was not possible when I tested it. 
It is possible with the channel name but not the username that's why I did the change with the PR

kunickiaj:
Hmm email should work (`@username` should not as it's not a unique identifier).

I intentionally left resolution logic out of the processor to keep things simpler and responsibility clear. What we do in our setup is we have a catalog processor that annotates the entities with resolved IDs. The processor does lookups against a backend plugin that provides the lookup api and caching.

JeremyVanderMotte:
Well in the meantime I retry using the email address without the implementation I did in the PR and it's working. So probably something I did wrong before :) 
Guess the PR has no sense anymore ;)

Issue #29349, opened 2025-03-23
🚀 Feature: MCP server support for Backstage


iocanel:
### 🔖 Feature description

It would be amazing if Bakcstage functionality would be exposed via [Model Context Protocl](https://www.anthropic.com/news/model-context-protocol). This would allow people using tools like [goose](https://github.com/block/goose) or [claude](https://claude.ai/download) to access Backstage functionality.

In other words it would allow users of AI related tools to use prompts like:

`Create an applciation using backstage template xxxx`
`Is there a register API for accessing TODO list management`
`Can you clone component xxx`

and more.

In other words, it will enable access to backstage functionality from tools becoming increasingly popular, without even requiring to open the backstage UI.

### 🎤 Context

I wanted to familiarize myself with MCP, so I started from scratch and tried to implement it in a domain that I familiar with.

I ended up implementing an MCP server for Backstage that is accessible from the [Quarkus Backstage](https://github.com/quarkiverse/quarkus-backstage) CLI. The result was mind blowing! If you want to get a glimpse of how it looks like I wrote a blog post about it: https://iocanel.com/2025/03/creating-an-mcp-server-with-quarkus-and-backstage.

However, this feature is way to cool, to limit it to [Quarkus](https://quarkus.io) users. So, I strongly feel that it should be part of backstage itself.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

Rugvip:
Yep this is on our radar 👍

lujmo11:
I came across this which could be some inspiration https://docs.gitbook.com/

evanshortiss:
I put together a basic example of an MCP server for Backstage using the Model Context Protocol SDK for TypeScript in the repository [evanshortiss/mcp-backstage-example](https://github.com/evanshortiss/mcp-backstage-example). I also included a simple Agent that communicates with the MCP server using the IBM Bee Agent Framework. Might be useful to some folks that stumble across this issue. Right now it just supports reading entities.

aramissennyeydd:
[...]

PR #30192, opened 2025-06-10
feat: add new scaffolder module for catalog

drodil:

moving the catalog specific things from builtin actions started here as well as adding most commonly used catalog actions to the new module.


benjdlambert (changes requested):
Hey @drodil :wave:

Thanks for this, and we've been certainly been thinking the same thoughts here.

I've been holding off reviewing this as we needed to get an RFC out with a plan forward, which means we're going to implement this in a slightly different way. 

#30218

drodil:
@benjdlambert this is INTERESTING. But not sure what the timeline is? We would at least need the validate entity functionality for the scaffolder. Of course we can implement it ourselves but could be something useful for the community.

benjdlambert:
@drodil this https://github.com/backstage/backstage/issues/29349#issuecomment-2909039873 suggests as early as next week there will be movement in this space. 

We've been working on pulling everything together, so no promises, but let's see what we can get out for next week.

PR #30199, opened 2025-06-10
Create stackoverflow-teams

EstoesMoises:
Request to add the Stack Overflow for Teams plugin that I created for Stack Overflow. This is part of the public Stack Exchange repository.

This includes two new plugins and a modified version of the existing Stack Overflow collator created by Spotify - https://github.com/StackExchange/backstage-stackoverflow/tree/main/plugins

Let me know if any questions ! - Thanks


EstoesMoises:
Hey @freben I completely missed adding the .yaml extension when I did this PR over the browser. I've now fixed this.

Additionally, thanks for having a look into the plugin and noticed that about the config.d.ts, I've also fixed that for the three plugins from my repository.

Many thanks !

PR #30206, opened 2025-06-11
feat: clean up old notifications

drodil:

a new scheduled task that will delete old notifications. the default is that over 1 year old notifications will be deleted. the scheduled task is run by default every 24 hours. this can be disabled by setting the expiration period to zero in the notifications config.


.changeset/common-goats-exist.md
```diff
@@ -0,0 +1,25 @@
...
+If the period is set to `0`, the expiration feature will be disabled, and notifications will not be automatically deleted.
+
+The scheduled task can also be scheduled to run at a different interval by setting the
+`notifications.expiration.schedule` in the `app-config.yaml` file.
+
+```yaml
+notifications:
+  expiration:
+    schedule: '0 0 * * *' # Default is every 24 hours
+```
```
Rugvip: Could you add this to `docs/notifications` instead? There's a lot of useful info here that will get lost in the changelog.

plugins/notifications-backend/src/plugin.ts
```diff
@@ -80,15 +83,18 @@ export const notificationsPlugin = createBackendPlugin({
...
+
         httpRouter.use(
           await createRouter({
             auth,
             httpAuth,
             userInfo,
             logger,
             config,
-            database,
+            store,
```
Rugvip, line 97: 👍

plugins/notifications-backend/src/service/NotificationCleaner.ts
```diff
@@ -0,0 +1,80 @@
...
+    expirationMs: number,
+  ) => {
+    logger.info('Starting notification cleaner task');
+    try {
+      const result = await database.clearNotifications(expirationMs);
+      logger.info(
+        `Notification cleaner task completed successfully, removed ${result} notifications`,
+      );
+    } catch (error) {
+      logger.error(`Notification cleaner task failed: ${error}`);
```
Rugvip: @freben would you suggest to re-throw `ForwardedError` with context instead of logging here?

Rugvip (approved):
Nice! 👍 

Some smaller stuff but pre-approved

plugins/notifications-backend/src/database/NotificationsStore.ts
```diff
@@ -109,4 +113,6 @@ export interface NotificationsStore {
   }): Promise<void>;
 
   getTopics(options: TopicGetOptions): Promise<{ topics: string[] }>;
+
+  clearNotifications(olderThanMs: number): Promise<number>;
```
freben:
```suggestion
  clearNotifications(options: { maxAge: HumanDuration }): Promise<{ deletedTotal: number }>;
```

perhaps? :)

plugins/notifications-backend/config.d.ts
```diff
@@ -28,5 +29,19 @@ export interface Config {
...
      */
     throttleInterval?: HumanDuration | string;
+    /**
+     * Notification expiration config
+     */
+    expiration?: {
+      /**
+       * Schedule to run the notification cleaner, defaults to once a day at midnight.
+       */
+      schedule?: SchedulerServiceTaskScheduleDefinitionConfig;
```
freben: Is it particularly valuable to be possible to configure? Not really a problem to have here, but just checking if it's worth omitting until someone (perhaps doesn't) ask for it to keep things lean?

plugins/notifications-backend/config.d.ts
```diff
@@ -28,5 +29,19 @@ export interface Config {
...
+      /**
+       * Time to keep the notifications in the database, defaults to 365 days.
+       * Can be disabled by setting to 0.
```
freben: Not clear what this last sentence means - disable cleanup, or disable keeping them at all?

`0` isn't a valid value actually. You could add an `| false` if people are meant to actively disable the feature, but then maybe it's better to put that on `expiration` instead so you say `expiration: false` meaning infinite retention. If you instead semantically meant "infinitely short retention", you might tell people to set `"0s"` or so, you need a unit if you want it to be a duration

.changeset/common-goats-exist.md
```diff
@@ -0,0 +1,25 @@
...
+
+If the period is set to `0`, the expiration feature will be disabled, and notifications will not be automatically deleted.
+
+The scheduled task can also be scheduled to run at a different interval by setting the
+`notifications.expiration.schedule` in the `app-config.yaml` file.
+
+```yaml
+notifications:
+  expiration:
+    schedule: '0 0 * * *' # Default is every 24 hours
```
freben: does this work? doesn't it have to be `{ cron: '<cronstring>' }`?

.changeset/common-goats-exist.md
```diff
@@ -0,0 +1,25 @@
...
+There is a new scheduled task that runs every 24 hours to delete notifications older than 1 year.
+This can be configured by setting the `notifications.expiration.period` in the `app-config.yaml` file.
+
+```yaml
+notifications:
+  expiration:
+    period: 365d # Default is 365 days
+```
+
+If the period is set to `0`, the expiration feature will be disabled, and notifications will not be automatically deleted.
```
freben: see other comment, `0` is not a valid duration

.changeset/common-goats-exist.md
```diff
@@ -0,0 +1,25 @@
...
+
+Notifications are now automatically deleted after 1 year by default.
+
+There is a new scheduled task that runs every 24 hours to delete notifications older than 1 year.
+This can be configured by setting the `notifications.expiration.period` in the `app-config.yaml` file.
+
+```yaml
+notifications:
+  expiration:
+    period: 365d # Default is 365 days
```
freben: you call it period in this file, but duration in config.d.ts.

Completely alternative suggestion:

```yaml
notifications:
  retention: 1y
```

?

plugins/notifications-backend/src/service/NotificationCleaner.ts
```diff
@@ -0,0 +1,80 @@
...
+  readSchedulerServiceTaskScheduleDefinitionFromConfig,
+  SchedulerService,
+  SchedulerServiceTaskScheduleDefinition,
+} from '@backstage/backend-plugin-api';
+import { Config, readDurationFromConfig } from '@backstage/config';
+import { NotificationsStore } from '../database';
+import { durationToMilliseconds, HumanDuration } from '@backstage/types';
+
+export class NotificationCleaner {
+  static readonly initNotificationCleaner = async (
```
freben: I don't know why but this static assignment in a class feels so odd :) Couldn't this just have been free functions or a normal class with a constructor

plugins/notifications-backend/src/service/NotificationCleaner.ts
```diff
@@ -0,0 +1,80 @@
...
+          key: 'notifications.expiration.duration',
+        })
+      : { years: 1 };
+    const expirationMs = durationToMilliseconds(expiration);
+    if (expirationMs > 0) {
+      const taskRunner = scheduler.createScheduledTaskRunner(schedule);
+      await taskRunner.run({
+        id: 'notification-cleaner',
+        fn: async () => {
+          await this.clearNotifications(logger, database, expirationMs);
```
freben: yeah ESPECIALLY using `this` here is kind of mind bending, if anything `NotificationCleaner. clearNotifications` would have been less surprising but java-ish

plugins/notifications-backend/src/service/NotificationCleaner.ts
freben: I think the task logger will handle it then, yes

drodil:
Ready for another round!

plugins/notifications-backend/src/service/NotificationCleaner.ts
drodil: This was to get rid of calling async function in constructor which is a bad habit. But replaced it with a new function to initialize the taskRunner.

docs/notifications/index.md
```diff
@@ -158,6 +158,21 @@ You can customize the origin names shown in the UI by passing an object where th
...
+
+Notifications are deleted automatically after a certain period of time to prevent the database from growing indefinitely
+and to keep the user interface clean. The default retention period is set to 1 year, meaning that notifications older
+than that will be deleted automatically.
+
+The retention period can be configured by setting the `notifications.retention` in the `app-config.yaml` file.
+
+```yaml
+notifications:
+  retention: 1y
```
Parsifal-M, line 171: Asking because we're doing some task related stuff as well in this format can you do:

`1y 6m 10d`? or `3m`? or even `700d`?

Just wondering 😅

drodil: Well, you can do it couple of ways the `readDurationFromConfig` supports:

- A string in the format of '1d', '2 seconds' etc. as supported by the `ms`
  library.
- A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
- An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.

drodil: Apparently, https://github.com/vercel/ms does not support having multiple units in a string; in that case, you would need to create an object.

Parsifal-M: Awesome! Makes sense!

.changeset/common-goats-exist.md
```diff
@@ -0,0 +1,15 @@
...
+
+There is a new scheduled task that runs every 24 hours to delete notifications older than 1 year.
+This can be configured by setting the `notifications.retention` in the `app-config.yaml` file.
+
+```yaml
+notifications:
+  retention: 1y
+```
+
+If the retention is set to false, notifications will not be automatically deleted.
```
freben, line 15: haha now "retention: false" does seem to indicate "nah don't retain things at all" :D no matter, it's no big deal, let's go

plugins/notifications-backend/src/service/NotificationCleaner.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      clearNotifications: jest.fn(),
+    } as unknown as NotificationsStore;
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('initNotificationCleaner', () => {
+    it('should initialize the notification cleaner with the correct schedule', async () => {
```
freben, line 42: It would be appreciated if all tests (not just here of course) were using TestDatabases to get it tested on all major database engines. It may turn out that you'll sometimes find you need shit like this instead of client side logic 😆 

```ts
export function knexRawNowPlus(knex: Knex, duration: HumanDuration): Knex.Raw {
  const seconds = durationToMilliseconds(duration) / 1000;
  if (knex.client.config.client.includes('sqlite3')) {
    return knex.raw(`datetime('now', ?)`, [`${seconds} seconds`]);
  } else if (knex.client.config.client.includes('mysql')) {
    return knex.raw(`now() + interval ? second`, [seconds]);
  }
  return knex.raw(`now() + (? || ' seconds')::interval`, [seconds]);
}
```

plugins/notifications-backend/src/service/NotificationCleaner.ts
```diff
@@ -0,0 +1,91 @@
...
+        await this.clearNotifications(
+          this.logger,
+          this.database,
+          this.retention,
+        );
+      },
+    });
+  }
+
+  private readonly clearNotifications = async (
```
freben: also doesn't need to be an assignment right

freben (approved):

plugins/notifications-backend/src/service/NotificationCleaner.ts
drodil: True

plugins/notifications-backend/src/service/NotificationCleaner.test.ts
drodil: Yeah, should convert the whole notification tests to use test databases and fix possible issues. But I will do this in another PR to not make this one go out of scope

drodil: Hm, but the `DatabaseNotificationsStore.test.ts` is already using `describe.each(databases.eachSupportedId())` to test with all databases. Notification cleaner itself does not do any database operations so this should be fine.

.changeset/common-goats-exist.md
drodil: Yeah, just following PR review comments blindly :D

plugins/notifications-backend/src/service/NotificationCleaner.test.ts
freben: Ah my bad mixing up the files.

PR #30215, opened 2025-06-11
cleanup: package-docs fixes for docsite

aramissennyeydd:

Hide all `@internal` packages in the Typedocs output and ensure that we aren't using previously cached output between runs.

### DIFF
--- a/packages/repo-tools/src/commands/package-docs/command.ts
+++ b/packages/repo-tools/src/commands/package-docs/command.ts
@@ -132,10 +132,6 @@ export default async function packageDocs(paths: string[] = [], opts: any) {
     recursive: true,
     force: true,
   });
-  await rm(cliPaths.resolveTargetRoot('dist-types'), {
-    recursive: true,
-    force: true,
-  });
   const selectedPackageDirs = await resolvePackagePaths({
     paths,
     include: opts.include,
--- a/packages/repo-tools/src/commands/package-docs/command.ts
+++ b/packages/repo-tools/src/commands/package-docs/command.ts
@@ -17,7 +17,7 @@ import { exec } from 'child_process';
 import { promisify } from 'util';
 import { paths as cliPaths, resolvePackagePaths } from '../../lib/paths';
 import { createTemporaryTsConfig } from './utils';
-import { readFile, writeFile } from 'fs/promises';
+import { readFile, rm, writeFile } from 'fs/promises';
 import pLimit from 'p-limit';
 import { mkdirp } from 'fs-extra';
 import { PackageDocsCache } from './Cache';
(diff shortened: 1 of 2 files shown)


packages/repo-tools/src/commands/package-docs/command.ts
```diff
@@ -113,6 +113,14 @@ async function generateDocJson(pkg: string) {
...
 export default async function packageDocs(paths: string[] = [], opts: any) {
   console.warn('!!! This is an experimental command !!!');
+  await rm(cliPaths.resolveTargetRoot('type-docs'), {
+    recursive: true,
+    force: true,
+  });
+  await rm(cliPaths.resolveTargetRoot('dist-types'), {
+    recursive: true,
+    force: true,
+  });
```
freben: This seemed surprising at first glance - in a CI flow this would ruin previously run tsc commands right

packages/repo-tools/src/commands/package-docs/command.ts
```diff
@@ -128,7 +136,13 @@ export default async function packageDocs(paths: string[] = [], opts: any) {
...
     selectedPackageDirs.map(pkg =>
       limit(async () => {
-        if (EXCLUDE.includes(pkg)) {
+        const pkgJson = JSON.parse(
+          await readFile(
+            cliPaths.resolveTargetRoot(pkg, 'package.json'),
+            'utf-8',
+          ),
+        );
+        if (EXCLUDE.includes(pkg) || pkgJson.name.startsWith('@internal/')) {
```
freben, line 156: is it generally true that this prefix should never be included for any adopter?

packages/repo-tools/src/commands/package-docs/command.ts
aramissennyeydd: ah yeah, good call - updated in [0da677d](https://github.com/backstage/backstage/pull/30215/commits/0da677d864501a3b53550d92e1506315e7c806d4) to just remove all docs.json

packages/repo-tools/src/commands/package-docs/command.ts
aramissennyeydd: Not completely, what we probably want here is to check the backstage metadata and skip based on role + if the package gets inlined into other packages. I think the issue with revealing this on the OSS docs is that it's a little confusing, ie https://backstage.io/api/stable/modules/_internal_frontend.html? Is that a package I can use?

packages/repo-tools/src/commands/package-docs/command.ts
freben: Just double checking .... this is still here.

freben (approved):
Preapproving but we still do want to keep the dist-types untouched right? See comment

packages/repo-tools/src/commands/package-docs/command.ts
```diff
@@ -113,6 +114,28 @@ async function generateDocJson(pkg: string) {
...
+  }
+  console.warn('!!! Deleting existing docs output !!!');
+  await rm(cliPaths.resolveTargetRoot('type-docs'), {
+    recursive: true,
+    force: true,
+  });
+  await rm(cliPaths.resolveTargetRoot('dist-types'), {
+    recursive: true,
+    force: true,
+  });
```
aramissennyeydd:
```suggestion
```

packages/repo-tools/src/commands/package-docs/command.ts
aramissennyeydd: 😅  - actually removed in [9786eec](https://github.com/backstage/backstage/pull/30215/commits/9786eec23c5d77a22cdaa3cc0f20f487044e6eef)

Issue #30220, opened 2025-06-12
Export vendor types for testing custom transformers


ganievs:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
'Active Directory', 'ldap', 'custom transformer', 'testing'
```

### 🗃️ Project Area

Catalog

### 🔖 Need

Request to export LDAP vendor type definitions from `plugins/catalog-backend-module-ldap/src/ldap/vendors.ts` to enable proper testing of custom LDAP transformers.

### 📝 Proposal

Currently, when implementing custom LDAP transformers for the catalog-backend-module-ldap plugin, developers cannot properly type their test mocks because the vendor types (ActiveDirectoryVendor, DefaultLdapVendor, etc.) are not exported from the module.

### 🔄 Alternatives

Copy vendor types to test files

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Yes, and I have enough information to get started

vinzscam:
Thank you  🙏
let's discuss this in the PR

benjdlambert:
Closing as this was merged :tada:

PR #30221, opened 2025-06-12
export vendor types for testing custom transformers

ganievs:

Exports LDAP vendor types and instances from the module's public API to enable 
proper testing of custom LDAP transformers.

### DIFF
--- a/.changeset/empty-actors-relax.md
+++ b/.changeset/empty-actors-relax.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-ldap': patch
+---
+
+Export LDAP vendor types and instances for testing custom transformers
--- a/plugins/catalog-backend-module-ldap/report.api.md
+++ b/plugins/catalog-backend-module-ldap/report.api.md
@@ -275,4 +275,14 @@ export type VendorConfig = {
   dnAttributeName?: string;
   uuidAttributeName?: string;
 };
+
+// @public
+export const vendors: {
+  readonly activeDirectory: LdapVendor;
+  readonly aeDir: LdapVendor;
+  readonly freeIpa: LdapVendor;
+  readonly googleLdap: LdapVendor;
+  readonly lldap: LdapVendor;
+  readonly default: LdapVendor;
+};
 ```
--- a/plugins/catalog-backend-module-ldap/src/ldap/index.ts
+++ b/plugins/catalog-backend-module-ldap/src/ldap/index.ts
@@ -13,6 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import {
+  ActiveDirectoryVendor,
+  AEDirVendor,
+  DefaultLdapVendor,
+  FreeIpaVendor,
+  GoogleLdapVendor,
+  LdapVendor,
+  LLDAPVendor,
+} from './vendors';
 
 export { LdapClient } from './client';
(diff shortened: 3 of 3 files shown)


ganievs:
related to an issue #30220

.changeset/empty-actors-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-ldap': minor
```
freben:
```suggestion
'@backstage/plugin-catalog-backend-module-ldap': patch
```

since it's a nonbreaking change in a 0.x package

plugins/catalog-backend-module-ldap/report.api.md
```diff
@@ -24,6 +24,12 @@ import { SearchOptions } from 'ldapjs';
 import { UserEntity } from '@backstage/catalog-model';
 import { UserTransformer as UserTransformer_2 } from '@backstage/plugin-catalog-backend-module-ldap';
 
+// @public
+export const ActiveDirectoryVendor: LdapVendor;
```
freben: Mmmm. If we are making these part of the public interface, could we at least make it like

```ts
export const vendors = {
  activeDirectory: ... as LdapVendor,
  default: ... as LdapVendor,
} as const;
```

or so? That way they are all collected in a struct neatly together in the public interface instead of being all over the place

ganievs: Agree. It would keep the public interface cleaner

ganievs: done

plugins/catalog-backend-module-ldap/src/ldap/vendors.ts
```diff
@@ -40,6 +40,15 @@ export type LdapVendor = {
   decodeStringAttribute: (entry: SearchEntry, name: string) => string[];
 };
 
+/**
+ * Default LDAP vendor configuration for standard LDAP servers.
+ *
+ * This vendor configuration works with most standard LDAP implementations
+ * including OpenLDAP, 389 Directory Server, and other RFC-compliant servers.
+ *
+ * @public
```
freben: now you can remove all of the `@public` actually in this file since they aren't directly exported.

plugins/catalog-backend-module-ldap/src/ldap/index.ts
```diff
@@ -37,3 +46,16 @@ export {
...
+ * @public
+ */
+export const vendors = {
+  activeDirectory: ActiveDirectoryVendor,
+  aeDir: AEDirVendor,
+  freeIpa: FreeIpaVendor,
+  googleLdap: GoogleLdapVendor,
+  lldap: LLDAPVendor,
+  default: DefaultLdapVendor,
+} as const as Record<string, LdapVendor>;
```
freben:
```suggestion
export const vendors = {
  activeDirectory: ActiveDirectoryVendor as LdapVendor,
  aeDir: AEDirVendor as LdapVendor,
  freeIpa: FreeIpaVendor as LdapVendor,
  googleLdap: GoogleLdapVendor as LdapVendor,
  lldap: LLDAPVendor as LdapVendor,
  default: DefaultLdapVendor as LdapVendor,
} as const;
```

and then generate the API reports again. Right? I think we actually are distinctly interested in seeing the individual vendor keys.

PR #30227, opened 2025-06-12
feat: eslint rule to check forbidden plugin imports

drodil:

basically verify-local-dependencies.js but done during linting also in the 3rd party repositories.


packages/eslint-plugin/rules/no-mixed-plugin-imports.js
```diff
@@ -0,0 +1,160 @@
...
+    if (!pkg) {
+      return {};
+    }
+
+    const options = context.options[0] || {};
+    /** @type {string[]} */
+    const ignoreTargetPackages = options.excludedTargetPackages || [];
+    /** @type {string[]} */
+    const ignorePatterns = options.excludedFiles || [
+      '**/*.{test,spec}.[jt]s?(x)',
```
drodil: Test files are ignored by default as for example `startTestBackend` requires to import the actual plugin:

```
Lint failed in plugins/auth-backend-module-oidc-provider

✘  https://google.com/#q=%40backstage%2Fno-mixed-plugin-imports

     @backstage/plugin-auth-backend-module-oidc-provider (backend-plugin-module) uses forbidden import from @backstage/plugin-auth-backend (backend-plugin)


     src/module.test.ts:147:9
     145 |       features: [
     146 |         authModuleOidcProvider,
   > 147 |         import('@backstage/plugin-auth-backend'),
         |         ^
     148 |         mockServices.rootConfig.factory({
     149 |           data: {
     150 |             app: { baseUrl: 'http://localhost' },
```

.eslintrc.js
```diff
@@ -18,8 +18,24 @@ var path = require('path');
 
 module.exports = {
   root: true,
-  plugins: ['@spotify', 'notice', 'react', 'testing-library'],
+  plugins: ['@spotify', 'notice', 'react', 'testing-library', '@backstage'],
```
drodil, line 21: Adding the same rules to the main backstage repo, just to eat your own dog food

.eslintrc.js
```diff
@@ -18,8 +18,24 @@ var path = require('path');
...
   rules: {
+    '@backstage/no-mixed-plugin-imports': [
+      'error',
+      {
+        // TODO: Fix these either by right role or by moving things to new packages
+        excludedTargetPackages: [
+          '@backstage/test-utils',
```
drodil: These ones have actual issues in the source code (which are also in exception in the `verify-local-dependencies.js`). Some of them could be fixed, some of them maybe not

packages/eslint-plugin/rules/no-mixed-plugin-imports.js
```diff
@@ -0,0 +1,215 @@
...
+      '**/dev/index.[jt]s?(x)',
```
Rugvip: Thinking this should match all of `dev`, or even anything outside of `src`, similar to https://github.com/backstage/backstage/blob/098ef95bfee6cb41324eaa3006e9e7f83d8e014a/packages/eslint-plugin/rules/no-undeclared-imports.js#L32-L37

plugins/catalog-backend-module-logs/package.json
```diff
@@ -35,6 +35,7 @@
   "dependencies": {
     "@backstage/backend-plugin-api": "workspace:^",
     "@backstage/plugin-catalog-backend": "workspace:^",
+    "@backstage/plugin-catalog-node": "workspace:^",
```
freben: was this accidental?

plugins/catalog-backend-module-logs/package.json
```diff
@@ -35,6 +35,7 @@
   "dependencies": {
     "@backstage/backend-plugin-api": "workspace:^",
     "@backstage/plugin-catalog-backend": "workspace:^",
```
freben: oof this is unfortunate, we don't want to import backends from modules -but it's using an export that needs to be moved to the -node package to resolve it. Was that what you were intending to do?

packages/eslint-plugin/rules/no-mixed-plugin-imports.js
```diff
@@ -0,0 +1,215 @@
...
+          suggest.push({
+            messageId: 'useNodePlugin',
+            data: {
+              targetPackage: targetName,
+            },
+            /** @param {import('eslint').Rule.RuleFixer} fixer */
+            fix(fixer) {
+              const source = context.sourceCode;
+              const nodeSource = source.getText(imp.node);
+              const newImport = nodeSource.replace(/-backend'$/, "-node'");
```
freben, line 208: we'll just have to assume that one exists :)

freben (approved):
Sweet!

Rugvip (approved):
Nice! 👍 🎉 

Happy to ship, minor comment if you have time

packages/eslint-plugin/rules/no-mixed-plugin-imports.js
drodil: Yep. Actually there could be another suggestion for `common` package.

packages/eslint-plugin/rules/no-mixed-plugin-imports.js
drodil: I made a new `includePatterns` option that defaults to all files in `**/src/**`, works better than excluding things (which is also still possible).

Additionally removed the default exclude patterns and moved the errornous packages in this project to `.eslintrc.js`. Those will need some addressing later on, maybe one-by-one.

julzllanillo:
I would like to understand this change more. We are in a multiple teams-per-plugin monorepo setup and we provide core services-as-a-plugin which can be imported in other plugins also. With this explicit rule, we started to have errors in most of our plugins due to the setup. Of course, for now we can opt-out but the thing is in the changes:

> This rule is initially only set at a warning level, to give you some time to react. You may see warnings appear in your build output that were not there before - do pay attention to them, because they may expose some issues that you had not yet noticed.
From [v1.40 Changes](https://backstage.io/docs/releases/v1.40.0/#new-lint-rule-added)

Our idea is to have that open communication between plugins for seamless creation between teams who wants to contribute to one another. The way I see this change, we will need to move any _central_ plugins that is hosted in plugin environment up to the `packages` level.

What I would like to get is more understanding why this choice was made and also what can we do in our setup like this.

Thank you.

drodil:
This was made to enforce the backstage plugin architecture. So basically this means in your plugins you should not import directly from other frontend or backend plugins but instead use the `common`, `react` and `node` plugins for shared functionality. This has not been so visible previously but it's how the plugin structure should be. 

You do not need to move plugins to packages but instead split them to the plugin types mentioned above.

julzllanillo:
Hi @drodil 
We are currently in the process of migrating plugin roles, and I want to ask why this scenario is not ticking the rule?
`frontend-plugin` imports from `backend-plugin`
Am I missing something? 'Cause both of them should import from `common` plugin instead right?

drodil:
@julzllanillo it should definetely trigger the rule. But as mentioned, at the moment it's by default only a warning.

julzllanillo:
@drodil 
Yes, we are doing `package lint` in order to trigger these off to be fixed properly, and VSCode shows it in the terminal with a yellowish squiggly line. It was just weird that we are not getting warnings for this specific scenario. Maybe it was something that was missed in this update? :(

drodil:
@julzllanillo the rule definetely is there. Sounds weird.

julzllanillo:
@drodil 
I found the issue! Nevermind this. I tried replicating the scenario on other plugins and it did give us the warning. Seems we were missing the `.eslintrc.js` file on this specific plugin.

PR #30231, opened 2025-06-13
`ActionsRegistry`: Allow providing attributes to the `Actions` that are registered

benjdlambert:
Again, no changeset because we already have them.

Looking for some input here as to what are sensible defaults, open to suggestions :pray:


packages/backend-defaults/src/entrypoints/actionsRegistry/DefaultActionsRegistryService.ts
```diff
@@ -67,13 +67,19 @@ export class DefaultActionsRegistryService implements ActionsRegistryService {
         actions: Array.from(this.actions.entries()).map(([id, action]) => ({
           id,
           ...action,
+          attributes: {
+            // todo(blam): what's safe defaults?
```
drodil, line 71: IMO one of them should be true when registering an action 🤔

packages/backend-defaults/src/entrypoints/actionsRegistry/DefaultActionsRegistryService.ts
```diff
@@ -67,13 +67,19 @@ export class DefaultActionsRegistryService implements ActionsRegistryService {
...
+            destructive: action.attributes?.destructive ?? false,
```
benjdlambert: Wonder if we should set this to `true` a la `modelcontextprotocol/sdk` by default: https://github.com/modelcontextprotocol/typescript-sdk/blob/dd69efa1de8646bb6b195ff8d5f52e13739f4550/src/types.ts#L777-L812

Rugvip (approved):

PR #30233, opened 2025-06-13
`ActionsRegistry`: Implementing some `mockServices` for testing

benjdlambert:
Opening this up for some comments here and thoughts forward.

Wanted to create some form of test harness to create actions with.

```ts
const actionsRegistry = mockServices.actionsRegistry();

actionsRegistry.register({
  name: 'test',   
  title: 'Test',
  description: 'Test',
  schema: {
    input: z.object({ name: z.string() }),
    output: z.object({ name: z.string() }),
  },
  action: async ({ input }) => ({ output: { name: input.name } }),
});

const result = await actionsRegistry.invoke({
  id: 'testing:test',
  input: { name: 'test' },
});

expect(result).toEqual({ output: { name: 'test' } });
```


packages/backend-test-utils/report.api.md
```diff
@@ -56,6 +60,33 @@ export interface CreateMockDirectoryOptions {
   mockOsTmpDir?: boolean;
 }
 
+// @public
+export class MockActionsRegistry
+  implements ActionsRegistryService, ActionsService
```
freben, line 65: Kind of neat to make a single mock service instead of two, when they functionally are a pair. The question is what to name it then :) This seems ok.

Would be kinda nice though if it just automatically worked out right if you do

```ts
const a = mockServices.actionsRegistry();
const b = mockServices.actions();
```

And it just did the right thing behind the scenes.

I was pondering for a bit whether there's something that can be done using like

```ts
// at the root of mockServices.ts
let actionsState = undefined;

beforeEach(() => {
  actionsState = undefined;
});

return function getActionsState() {
  actionsState ??= { /* whatever a good empty state is */ };  
  return actionsState;
}

export function actionsRegistry(): ActionsRegistryService {
  return new MockActionsRegistry(getActionsState());
}
export function actions(): ActionsService {
  return new MockActionsRegistry(getActionsState());
}
```

just illustrative pseudocode of course. This way, ANY pair of created actionsRegistry / actions services (on the same level of the code) can get a state container with `getActionsState` in their respective constructor and be sure that they get the same one. The only thing that wouldn't work would be to have eg `actionsRegistry` in a `define` and the `actions` inside an `it` because they would get different state containers. But it's a bit of an odd state.

Another alternative is

```ts
// at the root of mockServices.ts
const actionsState = { registered: [] };

beforeEach(() => {
  actionsState.registered.length = 0;
});
```

With that, you can't register things in `describe` and expect them to be visible inside `it`. But you can circumvent that too, by "pushing" a state in `beforeEach` and restoring it to its previous shape in `afterEach`.

```ts
// at the root of mockServices.ts
const actionsState = { registered: [] };
let savedState;

beforeEach(() => {
  savedState = [...actionsState.registered];
});
afterEach(() => {
  actionsState.registered = savedState
});
```

Or something like that.

Am I just tired and that doesn't work, or is it actually workable?

benjdlambert: It probably works it's just a bit unfortunate to add jest as a dependency of it all. Maybe that's fine though and we can of course just check if it's available before calling it.

Was just thinking that these `mockService` things might not only be called in `jest`.

Maybe guarding it in `typeof beforeEach === 'function'` or something?

benjdlambert: I did really want to try and move this state to like a separate `serviceRef` or something instead of having this global `let` state, as that felt a little cleaner. But's really only useful in `startTestBackend`, and not out of band of a Backstage backend process.

packages/backend-test-utils/src/next/services/MockActionsRegistry.ts
```diff
@@ -0,0 +1,164 @@
...
+      );
+    }
+  }
+
+  register<
+    TInputSchema extends AnyZodObject,
+    TOutputSchema extends AnyZodObject,
+  >(options: ActionsRegistryActionOptions<TInputSchema, TOutputSchema>): void {
+    // hardcode testing: prefix similar to how the default actions registry does it
+    const id = `testing:${options.name}`;
```
Rugvip: I think `test` is best here right? Similar to https://github.com/backstage/backstage/blob/a9219496d5c073aaa0b8caf32ece10455cf65e61/packages/backend-test-utils/src/next/services/mockServices.ts#L321

and

https://github.com/backstage/backstage/blob/861f162b4a39117b824669d67a951ed1db142e3d/packages/backend-test-utils/src/next/wiring/ServiceFactoryTester.ts#L99

PR #30252, opened 2025-06-16
feat(scaffolder-github): Allow file deletion in pull requests

fearphage:

This adds the ability to delete files in GitHub pull requests. Files are marked for deletion based on file contents being set to a user-supplied sentinel value. The implementation of this is a bit odd/unexpected. The _why_ is explained below in the "background" section. 

To delete a file, you need to set its contents to a special "deletion marker" (read: sentinel value). I made the minimum length of this 33 characters since UUID's are 32 characters and some files only contain UUID's and that's a perfectly valid file contents. I didn't go with empty file contents since some files are intentionally left empty. So if you have a file that contains this sentinel value, it will be deleted in the pull request.

### Background - Why Implement it this way?

The implementation is bit odd, because of the way Backstage handles pull requests. Backstage currently only does something similar to a "sparse checkout" of the files that you plan to change. This means the absence of a file does not mean it should be deleted. So I needed a way to know a file that exists  on disk should be deleted. I went with "magic" file contents. **I'm open to other suggestions**, but this is the best thing I came up with at the time.

### DIFF
--- a/plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts
+++ b/plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts
@@ -27,7 +27,10 @@ import {
 } from '@backstage/plugin-scaffolder-node';
 import { Octokit } from 'octokit';
 import { CustomErrorBase, InputError } from '@backstage/errors';
-import { createPullRequest } from 'octokit-plugin-create-pull-request';
+import {
+  createPullRequest,
+  DELETE_FILE,
+} from 'octokit-plugin-create-pull-request';
 import { getOctokitOptions } from '../util';
 import { examples } from './githubPullRequest.examples';
 import {
@@ -143,6 +146,15 @@ export const createPublishGithubPullRequestAction = (
           z.string({
             description: 'The name for the branch',
           }),
+        deletionMarker: z =>
+          z
+            .string({
+              description: 'Contents of files that will be deleted',
+            })
+            .min(33, {
+              message: 'deletion marker must be at least 33 characters long',
+            })
+            .optional(),
         targetBranchName: z =>
           z
             .string({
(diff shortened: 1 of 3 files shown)


fearphage:
> I wouldn't be too opposed to passing in a paths that should be removed from a PR instead in the Action, rather than setting the content to the delete marker.

@benjdlambert Started implementing this change, but then I came to a fork in the road. There are two ways to handle the `filesToDelete` list:

1. When looping through the list of files in the "sparse checkout" and I come across one that's on the list of `filesToDelete`, I delete it.
1. I just append the list of file changes with `filesToDelete` no matter what else is going on.

I started down the path of the first implementation, but then I wrote a test where I added a file to `filesToDelete`, but that file didn't exist in the "sparse checkout" so I'm wondering how to handle that behavior. Option 2 takes that option out of my hands entirely and will blindly delete files no matter what else is going on.

Thoughts?

EDIT: I went with option 2 for simplicity and lack of ambiguity.

plugins/scaffolder-backend-module-github/src/actions/githubPullRequest.ts
```diff
@@ -343,6 +353,18 @@ export const createPublishGithubPullRequestAction = (
         ]),
       );
 
+      if (filesToDelete) {
```
benjdlambert: Instead of using `Object.assign` here, can we just update the `files` assignment to something like this:


```ts
const files = Object.fromEntries([
...directoryContents.map((file) => ...),
...filesToDelete.map((file) => ... ),
])
```

Thinking that's a bit cleaner?

fearphage: Sounds good to me. I had to make a slight tweak since the param is optional and also I added a note to make sure others think about the order of operations in the future while they're changing this code.

fearphage:
> Will you be adding this feature to other providers?

@schultzp2020 No plans to do so. I'm not familiar with the providers.

benjdlambert (approved):

PR #30256, opened 2025-06-16
:broom: just a few more circular dependencies

freben:


packages/frontend-plugin-api/src/wiring/coreExtensionData.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 import { JSX } from 'react';
-import { RouteRef } from '../routing';
+import { RouteRef } from '../routing/RouteRef';
```
Rugvip, line 18: These are a bit unfortunate, they're not really circular in reality since it's just type imports. It would be nice if we didn't see any warnings for these, but it's ofc a trickier fix.

PR #30272, opened 2025-06-16
`Actions` / `ActionsRegistry`: Move things around to export as `alpha` services

benjdlambert:
A little bit of a cleanup and moving the API's to `alpha` exports where possible.

Only little bit of trickiness here is that I had to copy `simpleMock` implementation here as the `ServiceMock` is already exported as `@public` and it didn't like that when trying to export under `/alpha`, and didn't want to export `simpleMock` really.

And I would have liked to have `ActionsRegistry` and `Actions` under `@backstage/backend-defaults/alpha/actionsRegistry` and `@backstage/backend-defaults/alpha/actions` respectively but the API reports don't like that because of the filenames it would generate. Feeling it's not a blocker right now, and that we can change it later if needed.

Also decided to export them from `@backstage/backend-defaults/alpha` anyways, as they're used ion `backend-test-utils` but we could just do internal imports if we're feeling thats a better approach.


packages/backend-plugin-api/src/alpha/refs.ts
```diff
@@ -0,0 +1,56 @@
...
+ * See {@link ActionsService}
+ * and {@link https://backstage.io/docs/backend-system/core-services/actions | the service docs}
+ * for more information.
+ *
+ * @alpha
+ */
+export const actionsServiceRef = createServiceRef<
+  import('./ActionsService').ActionsService
+>({
+  id: 'alpha.core.actions',
```
Rugvip, line 40: reflecting a bit on this I can't tell whether `alpha.core.actions`, `core.alpha.actions`, or `core.actions.alpha` is the way to go. Either way it doesn't matter for this but thinking more about future process docs, both for alpha but also something like a `v2` of a service.

packages/backend-test-utils/report-alpha.api.md
```diff
@@ -0,0 +1,85 @@
...
+  const // (undocumented)
+    factory: () => ServiceFactory<ActionsService, 'plugin', 'singleton'>;
+  const // (undocumented)
+    mock: (
+      partialImpl?: Partial<ActionsService> | undefined,
+    ) => ServiceMock<ActionsService>;
+}
+
+// @alpha
+export class MockActionsRegistry
```
Rugvip, line 47: Is it important to export this one? Thinking we could keep it internal. Once could inject a mock logger via the factory tester anyway

packages/backend-test-utils/src/alpha/services/simpleMock.ts
```diff
@@ -0,0 +1,57 @@
...
+} & {
+  [Key in keyof TService]: TService[Key] extends (
+    ...args: infer Args
+  ) => infer Return
+    ? TService[Key] & jest.MockInstance<Return, Args>
+    : TService[Key];
+};
+
+/** @internal */
+export function simpleMock<TService>(
```
Rugvip, line 34: fine to copy this, we still need to settle whether this is something we want to export for broader use or not. Copy until then is the way 😁

packages/backend-test-utils/report-alpha.api.md
benjdlambert: Let's come back to this, marking this as `@internal` caused things to break with the `api-reports`.

PR #30286, opened 2025-06-16
Canon - Update TextField to use React Aria

cdedreuille:

### DIFF
--- a/packages/canon/src/components/TextField/TextField.stories.tsx
+++ b/packages/canon/src/components/TextField/TextField.stories.tsx
@@ -16,18 +16,19 @@
 
 import type { Meta, StoryObj } from '@storybook/react';
 import { TextField } from './TextField';
-import { Flex } from '../Flex';
+import { Form } from 'react-aria-components';
 import { Icon } from '../Icon';
+import { Flex } from '../Flex';
 
 const meta = {
-  title: 'Components/TextField',
+  title: 'Forms/TextField',
   component: TextField,
   argTypes: {
-    secondaryLabel: {
-      control: 'text',
-    },
-    required: {
+    isRequired: {
       control: 'boolean',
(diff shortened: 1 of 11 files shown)


packages/canon/src/components/TextField/TextField.stories.tsx
```diff
@@ -38,137 +40,107 @@ type Story = StoryObj<typeof meta>;
...
+    validate: value => (value === 'admin' ? 'Nice try!' : null),
+  },
+};
+
+export const CustomField: Story = {
+  render: () => (
+    <>
+      <FieldLabel
+        htmlFor="custom-field"
+        id="custom-field-label"
```
Rugvip, line 135: Thinking we may want to encourage `useId`? although we do still aim to support React 17 for the time being so a bit of a dilemma.

packages/canon/src/components/TextField/TextField.tsx
```diff
@@ -14,110 +14,80 @@
...
+      'aria-label': ariaLabel,
+      'aria-labelledby': ariaLabelledBy,
       ...rest
     } = props;
 
+    useEffect(() => {
+      if (!label && !ariaLabel && !ariaLabelledBy) {
+        console.warn(
+          'TextField requires either a visible label, aria-label, or aria-labelledby for accessibility',
+        );
```
Rugvip, line 49: Would you go as far as just make this an error? 😁

Rugvip (approved):

packages/canon/src/components/TextField/TextField.tsx
cdedreuille: This was the recommended way from React Aria. We could push it to an error if that's not enough but perhaps to start with, let's keep it as a warning.

Issue #30218, opened 2025-06-12
RFC: Distributed Actions with MCP in Backstage


benjdlambert:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
`ActionsRegistry`, MCP, Distributed Actions, Scaffolder, ModelContextProtocol
```

### 🗃️ Project Area

Core Framework

### 🔖 Need

[Model Context Protocol (MCP)](https://modelcontextprotocol.io/) is a fast moving space right now, and there’s a need to enable different MCP tools that provide access to different information and data which is surfaced through plugins in Backstage.

While MCP tools are useful on their own, there is also an opportunity to align this work with the broader Backstage framework rather than having MCP be bolted on. In many cases we expect Software Template Actions to be more or less identical to their MCP tool counterparts. Consider examples like registering a new component in the Software Catalog, paging the owner of an Entity, or opening a Pull Request; these all have potential to provide value both as actions in a Software Template, and as MCP tools.

Another related area is the opportunity to simplify the installation of Scaffolder actions. Today each action needs to be installed as a module for the scaffolder, meaning that they need to be published in their own package separately from what might be the main plugin backend package. There is room for improvement here, where we instead only need to install a single package to both provide the base functionality of the plugin and actions related to the plugin.

### 📝 Proposal

We propose the ability for plugins to define actions without the need to install a module for the Scaffolder. This means that we need to provide APIs for plugins to use to define actions that would then later be able to be invoked from other plugins.

### Actions Registry
For this, we’ve started iteration on an `ActionsRegistryService` and `ActionsService`, which would be part of the Core Services that are shipped out the box in Backstage. These allow plugins to define actions with a very similar API to the Scaffolder actions, but effectively these become distributed and installed with the plugins themselves. 
The `ActionsService` is a client on top of all these actions that are defined in the `ActionsRegistry`, allowing a plugin author to list the available actions and invoke them.

### MCP Integration
We will provide an MCP server on top of the  `ActionsRegistryService` and `ActionsService` in order to map the available actions as tools for a server, and invoke them. This will be the first step in bringing MCP integration to Backstage, with a focus on simplicity and a lightweight integration.

[...]

PR #30288, opened 2025-06-17
Add initial implementation of MCP Server

benjdlambert:
Here's the MCP Server which implements the RFC #30218.

More docs to follow shortly.


.changeset/lovely-lands-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-mcp-backend': patch
```
freben:
```suggestion
'@backstage/plugin-mcp-backend': minor
```

plugins/mcp-backend/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "backstage": {
+    "role": "backend-plugin",
+    "pluginId": "mcp"
+  },
+  "publishConfig": {
+    "access": "public",
+    "main": "dist/index.cjs.js",
+    "types": "dist/index.d.ts"
+  },
+  "private": true,
```
freben:
```suggestion
```

Right?

plugins/mcp-backend/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@backstage/plugin-mcp-backend",
+  "version": "0.1.0",
```
freben:
```suggestion
  "version": "0.0.0",
```

plugins/mcp-backend/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "express": "^4.17.1",
+    "express-promise-router": "^4.1.0",
+    "zod": "^3.22.4"
+  },
+  "devDependencies": {
+    "@backstage/backend-test-utils": "workspace:^",
+    "@backstage/cli": "workspace:^",
+    "@types/express": "^4.17.6",
+    "@types/supertest": "^2.0.8",
+    "supertest": "^6.2.4"
```
freben: This one was a new yarn.lock entry

plugins/mcp-backend/src/routers/createSseRouter.ts
```diff
@@ -0,0 +1,64 @@
...
+    await server.connect(transport);
+  });
+
+  router.post('/messages', async (req, res) => {
+    const sessionId = req.query.sessionId as string;
+    const transport = transportsToSessionId.get(sessionId);
+    if (transport) {
+      await transport.handlePostMessage(req, res, req.body);
+    } else {
+      res.status(400).send('No transport found for sessionId');
```
freben:
```suggestion
      res.status(400).send(`No transport found for session '${sessionId}'`);
```

good to have

plugins/mcp-backend/src/routers/createSseRouter.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    res.on('close', () => {
+      transportsToSessionId.delete(transport.sessionId);
+    });
+
+    await server.connect(transport);
+  });
+
+  router.post('/messages', async (req, res) => {
+    const sessionId = req.query.sessionId as string;
```
freben: Check for missing/empty sessionid? Which is a bit different than the check below even though they both result in a 400

plugins/mcp-backend/src/plugin.ts
```diff
@@ -0,0 +1,72 @@
...
+          mcpService,
+          httpAuth,
+          logger,
+        });
+
+        const router = Router();
+        router.use(json());
+
+        router.use('/sse', sseRouter);
+        router.use('/', streamableRouter);
```
Rugvip: Thinking we should mount this on some path, since right now it'll all just be `/api/mcp/`, which is a bit sketchy. I remember the `signals` backend being a bit trickier to deploy due to the use of `/`. Thinking `/api/mcp/v1` or something like that at least? along with `v1/sse`

plugins/mcp-backend/src/plugin.ts
```diff
@@ -0,0 +1,72 @@
...
+  actionsServiceRef,
+} from '@backstage/backend-plugin-api/alpha';
+
+/**
+ * mcpPlugin backend plugin
+ *
+ * @public
+ */
+export const mcpPlugin = createBackendPlugin({
+  pluginId: 'mcp',
```
Rugvip: Wondering if we might want to scope this a little bit more, like `mcp-actions`?

plugins/mcp-backend/src/routers/createStreamableRouter.ts
```diff
@@ -0,0 +1,100 @@
...
+          },
+          id: null,
+        });
+      }
+    }
+  });
+
+  router.get('/', async (_, res) => {
+    // We only support POST requests, so we return a 405 error for all other methods.
+    res.writeHead(405).end(
```
freben: Ah do you need writeHead because of nuances with the streaming protocol? Does that also mean that you can't use `.json` after `writeHead`?

plugins/mcp-backend/src/services/McpService.ts
```diff
@@ -0,0 +1,87 @@
...
+        })),
+      };
+    });
+
+    server.setRequestHandler(CallToolRequestSchema, async ({ params }) => {
+      const { actions } = await this.actions.list({ credentials });
+      const action = actions.find(a => a.name === params.name);
+
+      if (!action) {
+        throw new Error(`Action ${params.name} not found`);
```
freben: we could make this an InputError

freben (approved):
Alright just some random little comments

plugins/mcp-backend/src/routers/createStreamableRouter.ts
benjdlambert: ah, I think so.

This is just heavily inspired by the `@modelcontextprotocol/sdk` example implementation for stateless.

It might change over time when if/when we add state using the `DatabaseService`.

plugins/mcp-backend/src/services/McpService.ts
benjdlambert: Changed it to `NotFoundError`

plugins/mcp-actions-backend/src/routers/createSseRouter.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (!sessionId) {
+      res.status(400).send('sessionId is required');
+      return;
+    }
+
+    const transport = transportsToSessionId.get(sessionId);
+    if (transport) {
+      await transport.handlePostMessage(req, res, req.body);
+    } else {
+      res.status(400).send(`No transport found for sessionId "${sessionId}"`);
```
Rugvip: [**NOPE**](https://github.com/backstage/backstage/blob/master/SECURITY.md#express-responses)

benjdlambert: OH YEAH. We should raise a PR to `mcp`.

plugins/mcp-actions-backend/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@backstage/plugin-mcp-actions-backend",
+  "version": "0.0.0",
+  "backstage": {
+    "role": "backend-plugin",
+    "pluginId": "mcp",
```
Rugvip: Needs an update

.changeset/lovely-lands-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-mcp-actions-backend': minor
+---
+
+Initial implementation of an `mcp` backend
```
Rugvip: Needs an update

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -0,0 +1,88 @@
...
+      {
+        name: 'backstage',
+        // TODO: this version will most likely change in the future.
+        version,
+      },
+      { capabilities: { tools: {} } },
+    );
+
+    server.setRequestHandler(ListToolsRequestSchema, async () => {
+      const { actions } = await this.actions.list({ credentials });
```
Rugvip, line 46: TODO to switch this to configuration-based instead?

vinzscam (approved):
🚢

PR #30292, opened 2025-06-17
fix: hide pagination query params if pagination mode is set to none

alexvoedi:
When entity list pagination is not enabled, there should be no `limit`, `cursor` or `offset` in the url query parameters since they are not used and have no effect.

We stumbled across the problem in our E2E tests, where we check the URL including query parameters. With pagination deactivated, the query parameters make no sense anyway.

### DIFF
--- a/.changeset/open-seas-ring.md
+++ b/.changeset/open-seas-ring.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-react': patch
 ---
 
-fix: hide pagination query params if pagination mode is set to none
+hide pagination `queryparams` if pagination mode is set to none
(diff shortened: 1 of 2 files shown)


benjdlambert:
@vinzscam do you wanna 👍  and 🚢  this?

.changeset/open-seas-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-react': patch
+---
+
+fix: hide pagination query params if pagination mode is set to none
```
benjdlambert:
```suggestion
hide pagination `queryparams` if pagination mode is set to none
```

benjdlambert (changes requested):
Hey :wave:

Thanks for this, just a quick mark about the changeset.

benjdlambert (approved):
Let's give this a go :tada:

PR #30293, opened 2025-06-17
`catalog`: Implement simple `get-catalog-entity` action

benjdlambert:
This is a simple action implementation that allows you to retrieve an entity from the Catalog.


Rugvip (approved):
👍

PR #30294, opened 2025-06-17
`mcp-actions`: Use `output` instead of `structuredOutput` for the time being

benjdlambert:
See the comments in the code for why we need to avoid this for now.


Rugvip (approved):
👍

PR #30295, opened 2025-06-17
`mcp-actions`: Include some more docs about setup

benjdlambert:


plugins/mcp-actions-backend/README.md
```diff
@@ -19,10 +19,115 @@ const backend = createBackend();
...
+});
+```
+
+### Authentication Configuration
+
+By default, the Backstage backend requires authentication for all requests.
+
+We're looking at [device authentication](https://github.com/backstage/backstage/pull/27680) to make this much easier with authentication for MCP clients, but in the meantime you can take advantage of the [static tokens](https://backstage.io/docs/auth/service-to-service-auth/#static-tokens) feature.
+
+#### External Access with Static Tokens
```
Rugvip, line 78: Perhaps worth calling out again within this section that this is intended as a temporary workaround until we can get user auth going.

PR #30313, opened 2025-06-18
docs(search): replaced plain note with styled ::note block for clarity

jaisuriyaNatarajan:
Updated the search engine documentation to replace a plain `**Note:**` markdown block with a properly styled `::note` directive for improved readability and consistency across Backstage docs.

###  Before:
image

### After:
image

### DIFF
--- a/docs/features/search/search-engines.md
+++ b/docs/features/search/search-engines.md
@@ -91,7 +91,10 @@ search:
       fragmentDelimiter: ' ... ' # Delimiter string used to concatenate fragments. Defaults to " ... ".
 ```
 
-**Note:** the highlight search term feature uses `ts_headline` which has been known to potentially impact performance. You only need this minimal config to disable it should you have issues:
+:::note Note
+The highlight search term feature uses `ts_headline` which has been known to potentially impact performance. You only need this minimal config to disable it should you have issues:
+
+:::
 
 ```yaml
 search:


docs/features/search/search-engines.md
```diff
@@ -91,7 +91,10 @@ search:
       fragmentDelimiter: ' ... ' # Delimiter string used to concatenate fragments. Defaults to " ... ".
 ```
 
-**Note:** the highlight search term feature uses `ts_headline` which has been known to potentially impact performance. You only need this minimal config to disable it should you have issues:
+:::note Note
```
fearphage, line 94: I didn't know this existed. Does this work in Backstage's techdocs or just for the generated Backstage documentation?

I wonder why they didn't go with the [GitHub flavor of callouts](https://github.com/orgs/community/discussions/16925) instead.

jaisuriyaNatarajan: Yeah, this works in Backstage TechDocs because it's using MkDocs with the Material theme, which supports that :::note syntax through the Admonition plugin.

I think they went with this instead of GitHub-style callouts since MkDocs doesn’t support those natively, and this one’s more flexible with styling and usage. Kind of a “best fit” for the setup they’re using

awanlin: That's not correct, these admonition are not supported by TechDocs, at least out of the box. Also, the MkDocs Material theme uses `!!!` for it admonitions: https://squidfunk.github.io/mkdocs-material/reference/admonitions/#usage

Historically when Backstage was initially open sourced we needed a quick way to get a marketing site, a blog, and docs up and running quickly. Using Docusaurus allowed us to do that which is why there's a break between how TechDocs works and how the docs that drive backstage.io/docs work.

PR #30314, opened 2025-06-18
 Fix: Consistent Casing for Category Filter Display

jaisuriyaNatarajan:
**Issue**
Some categories like "api-manager" were displayed in lowercase or kebab-case while others used Proper Case (e.g., API Management, App Delivery).

 **Fix**
Updated the display text to ensure all category labels follow Proper Case (e.g., API Manager), regardless of how the category key is stored (e.g., kebab-case or camelCase). This improves visual consistency in the filter list UI.

 **Checklist**
 
 A changeset describing the change and affected packages.

 UI text display normalized

 All commits have a Signed-off-by line

📸 Screenshot Before:
Api manager before

📸 Screenshot After:
Api manager after

### DIFF
--- a/microsite/data/plugins/kong-service-manager.yaml
+++ b/microsite/data/plugins/kong-service-manager.yaml
@@ -2,7 +2,7 @@
 title: VeeCode Kong Service Manager
 author: VeeCode Platform
 authorUrl: https://platform.vee.codes/
-category: api-manager
+category: API Manager
 description: The Kong Service Manager plugin provides information about the kong service, a list of the routes it has and also offers the possibility of manipulating plugins without leaving the backstage.
 documentation: https://platform.vee.codes/plugin/kong-service-manager/
 iconUrl: https://veecode-platform.github.io/support/imgs/logo_3.svg


awanlin (changes requested):
Hi @jaisuriyaNatarajan, can you update the PR description to better reflect that you simply updated the category text? 

When I read : "Normalized display text to follow Proper Case (e.g., API Manager) regardless of how the category key is stored in data (e.g., kebab-case or camelCase)." I expected there to be more to this but there wasn't which is fine just created a bit of confusion.

Also, it generally best to keep the existing checklist. 👍

jaisuriyaNatarajan:
> Hi @jaisuriyaNatarajan, can you update the PR description to better reflect that you simply updated the category text?
> 
> When I read : "Normalized display text to follow Proper Case (e.g., API Manager) regardless of how the category key is stored in data (e.g., kebab-case or camelCase)." I expected there to be more to this but there wasn't which is fine just created a bit of confusion.
> 
> Also, it generally best to keep the existing checklist. 👍

Thanks for the feedback @awanlin! 🙌
I've updated the PR description to clearly reflect that it's just a UI display text update for consistent casing — no changes to how the data is stored. Also restored the default checklist as suggested. Let me know if anything else needs tweaking!

jaisuriyaNatarajan:
can you please review it @awanlin

PR #30321, opened 2025-06-19
fix: Gitlab auth issue

jboeijenga:

Previously, when using GitLab SCM authentication in the scaffolder, tokens were not reused as expected. This was caused by the scope check failing because the `repoWrite` scopes `api` and `write_repository` were combined into a single space-separated string instead of two separate string.

As a result, users were repeatedly prompted to authenticate, even when an existing token with the correct scopes was available and not expired.

### DIFF
--- a/.changeset/tired-bats-attend.md
+++ b/.changeset/tired-bats-attend.md
@@ -1,0 +1,5 @@
+---
+'@backstage/integration-react': patch
+---
+
+Separated gitlab `write_repository` scope to pass checks in `RefreshingAuthSessionManager`
--- a/.changeset/tired-bats-attend.md
+++ b/.changeset/tired-bats-attend.md
@@ -2,4 +2,4 @@
 '@backstage/integration-react': patch
 ---
 
-Separated gitlab `write_repository` scope to pass checks in `RefreshingAuthSessionManager`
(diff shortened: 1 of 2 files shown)


.changeset/tired-bats-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/integration-react': patch
+---
+
+Seperated gitlab repoWrite scope to pass checks in RefreshingAuthSessionManager
```
freben:
```suggestion
Separated gitlab `write_repository` scope to pass checks in `RefreshingAuthSessionManager`
```

this is what you meant right?

freben (commented):
Remember to put DCO on every single commit. Squasking into a single commit probably helps. Check the comment above from the bot.

jboeijenga:
@freben, thanks for your comment.
I have updated the changeset description and squashed and signed the commits to this branch.

freben:
~~Sorry again but the delivery of a third party plugin stopped working and [we issued a fix](https://github.com/backstage/backstage/pull/30353) - you will want to rebase on latest master to make things build agian.~~ EDIT: I did the rebase to try to get you unblocked - sorry, let's see if it goes green now.

freben (approved):

PR #30324, opened 2025-06-19
Fix(footer): open Spotify link in new tab with target='_blank'

jaisuriyaNatarajan:

Updated the footer link ("Made with ❤️ at Spotify") to open in a new tab using `target="_blank"`.

### DIFF
--- a/microsite/docusaurus.config.ts
+++ b/microsite/docusaurus.config.ts
@@ -480,7 +480,7 @@ const config: Config = {
           ],
         },
       ],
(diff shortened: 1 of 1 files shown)


jaisuriyaNatarajan:
> Why?

Thanks for asking!

I noticed that the “Made with ❤️ at Spotify” link opens in the same tab, which navigates users away from the Backstage site. Since it’s an external link, opening it in a new tab (target="_blank") helps retain user context and session — especially useful when users are browsing documentation or using the microsite.

This also follows common UX practices for external links.

jaisuriyaNatarajan:
Got it — thanks for the clarification!

I understand that opening links in a new tab can be subjective and might interfere with personal browsing preferences. I’ll go ahead and close this PR and keep that in mind for future contributions.

Appreciate the feedback!

PR #30325, opened 2025-06-19
Canon - Remove support for separated css styles

cdedreuille:

To simplify theming and make maintenance easier, we're consolidating all styles for Canon into a single `styles.css` file. We've removed individual component style files and updated our build process accordingly.


canon-docs/public/styles.css
```diff
@@ -0,0 +1,2 @@
+/*! modern-normalize v3.0.1 | MIT License | https://github.com/sindresorhus/modern-normalize */
```
vinzscam, line 1: is this minified on purpose?

Rugvip: +1 to keeping styles non-minified to simplify readability in editors etc. These are less important I guess but especially `packages/canon/css/styles.css` in the published package should not be minified.

Rugvip (approved):
Pre-approved, but let's avoid the minification

canon-docs/public/styles.css
cdedreuille: This is minified for performance purpose. If anyone wants to look into it there's plenty of tools to beautify it. I don't think we should remove the magnification at this point just for clarity.

cdedreuille:
@Rugvip @vinzscam I removed the minification for now but I would love to revisit that later on.

canon-docs/public/styles.css
Rugvip: It'll be minified in the final production build anyway though when built in production mode

PR #30335, opened 2025-06-20
Canon - Bring styles into a single place

cdedreuille:

In our road to make Canon stable we need to bring more stability to our styling definition. We are styling each one of our components with a mix of CSS classes + data attributes for props. We brought all definitions into a single place and created a brand new `useStyles()` hook that will take the component name + custom responsive props to be paste down our component.

**This PR is non-breaking and doesn't bring any changes in API nor visual changes.**

In practice this is how it look like:

```tsx
const { classNames, dataAttributes } = useStyles('Button', {
  size,
  variant,
});

console.log('classNames', classNames);
console.log('dataAttributes', dataAttributes);

// This will return
classNames { root: 'canon-Button' }
dataAttributes { data-size: 'small', data-variant: 'primary' }
```


packages/canon/src/hooks/useStyles.ts
```diff
@@ -0,0 +1,92 @@
...
+  return undefined;
+}
+
+/**
+ * React hook to get class names and data attributes for a component with responsive support
+ * @param componentName - The name of the component
+ * @param props - Object with prop values (can be responsive)
+ * @returns Object with classNames and dataAttributes
+ */
+export function useStyles<T extends ComponentDefinitionName>(
```
benjdlambert, line 61: Wondering if we should think about what to call this if this becomes part of the public API, but fine for now. Feels a little confusing with the old patterns that we had in `material-ui@4`

benjdlambert (approved):

packages/canon/src/hooks/useStyles.ts
cdedreuille: Yes 100%. For now this is only meant for internal use only and it is not exported. If we do I agree we should find a better name for it.

benjdlambert:
@Rugvip for you to have a look through once you're back, but merging to get into the release today. :pray:

PR #30336, opened 2025-06-20
feat(catalog): make content header of `CatalogIndexPage` customizable

benjidotsh:
This PR makes the content header of `CatalogIndexPage` customizable, allowing for replacing the default content header with something entirely different:

Scherm­afbeelding 2025-06-20 om 16 46 56

### DIFF
--- a/docs/features/software-catalog/catalog-customization.md
+++ b/docs/features/software-catalog/catalog-customization.md
@@ -416,6 +416,29 @@ const routes = (
 
 The same method can be used to customize the _default_ filters with a different interface - for such usage, the generic argument isn't needed since the filter shape remains the same as the default.
 
+## Customize Content Header
+
+The `CatalogIndexPage` comes with a default content header, but you may want to customize it:
+
+```tsx title="packages/app/src/App.tsx"
+import { DismissableBanner } from '@backstage/core-components';
+
+<Route
+  path="/catalog"
+  element={
+    <CatalogIndexPage
+      contentHeader={
+        <DismissableBanner
+          id="repository_migration_notice"
+          variant="info"
+          message="We're currently migrating repositories to a new provider. Your entities might be temporarily missing from the catalog."
+        />
+      }
+    />
+  }
+/>;
+```
+
 ## Advanced Customization
 
 For those where none of the above fits their needs you can take the option of creating a fully custom `CatalogIndexPage`.
--- a/plugins/catalog/src/components/CatalogPage/DefaultCatalogContentHeader.tsx
+++ b/plugins/catalog/src/components/CatalogPage/DefaultCatalogContentHeader.tsx
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2025 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  ContentHeader,
+  CreateButton,
+  SupportButton,
+} from '@backstage/core-components';
+import { useRouteRef } from '@backstage/core-plugin-api';
+import { createComponentRouteRef } from '../../routes';
+import { catalogTranslationRef } from '../../alpha/translation';
+import { useTranslationRef } from '@backstage/core-plugin-api/alpha';
+import { catalogEntityCreatePermission } from '@backstage/plugin-catalog-common/alpha';
+import { usePermission } from '@backstage/plugin-permission-react';
+
+export function DefaultCatalogContentHeader() {
+  const createComponentLink = useRouteRef(createComponentRouteRef);
+  const { allowed } = usePermission({
+    permission: catalogEntityCreatePermission,
+  });
+  const { t } = useTranslationRef(catalogTranslationRef);
+
+  return (
+    <ContentHeader title="">
+      {allowed && (
+        <CreateButton
+          title={t('indexPage.createButtonTitle')}
+          to={createComponentLink && createComponentLink()}
+        />
+      )}
+      <SupportButton>{t('indexPage.supportButtonContent')}</SupportButton>
+    </ContentHeader>
+  );
+}
--- a/.changeset/long-jeans-fail.md
+++ b/.changeset/long-jeans-fail.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog': minor
+---
+
+Added support for customizing the content header of `CatalogIndexPage`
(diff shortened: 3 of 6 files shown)


benjidotsh:
As discussed with @ahhhndre on [Discord](https://discord.com/channels/687207715902193673/1389605429142884473), I took out support for the "old frontend system" and exclusively added support for the new frontend system.

benjidotsh:
not stale

plugins/catalog/src/alpha/pages.tsx
```diff
@@ -40,6 +41,10 @@ import { buildFilterFn } from './filter/FilterWrapper';
 export const catalogPage = PageBlueprint.makeWithOverrides({
   inputs: {
     filters: createExtensionInput([coreExtensionData.reactElement]),
+    contentHeader: createExtensionInput(
+      [CatalogContentHeaderBlueprint.dataRefs.element.optional()],
+      { singleton: true, optional: true },
```
freben: Not sure it has to be singleton? Is it ok if different sources provide individual admonitions that stack vertically?

plugins/catalog/src/alpha/pages.tsx
```diff
@@ -40,6 +41,10 @@ import { buildFilterFn } from './filter/FilterWrapper';
 export const catalogPage = PageBlueprint.makeWithOverrides({
   inputs: {
     filters: createExtensionInput([coreExtensionData.reactElement]),
+    contentHeader: createExtensionInput(
+      [CatalogContentHeaderBlueprint.dataRefs.element.optional()],
```
freben: See the filters above - you probably don't need to put core things in `dataRefs`; just use `coreExtensionData.reactElement` directly wherever needed instead.

plugins/catalog-react/src/alpha/blueprints/CatalogContentHeaderBlueprint.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const CatalogContentHeaderBlueprint = createExtensionBlueprint({
+  kind: 'catalog-content-header',
+  attachTo: { id: 'page:catalog', input: 'contentHeader' },
+  dataRefs: {
+    element: coreExtensionData.reactElement,
+  },
+  output: [coreExtensionData.reactElement.optional()],
+  *factory(
+    params: {
+      loader: () => Promise<JSX.Element>;
```
freben: since the data is optional, did you intend for the loader to be able to return null too, resulting in no output yielded, resulting in no header shown?

plugins/catalog/src/components/CatalogPage/DefaultCatalogContentHeader.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <ContentHeader title="">
+      {allowed && (
+        <CreateButton
+          title={t('indexPage.createButtonTitle')}
+          to={createComponentLink && createComponentLink()}
+        />
+      )}
+      <SupportButton>{t('indexPage.supportButtonContent')}</SupportButton>
+    </ContentHeader>
```
freben: Note though that if you move it out here and it's also a singleton, people would have to reimplement / copy-paste all of this into their custom implementation whenever they wanted to add something at the top of the page.

Could it be argued that this particular header content is different from "extra admonitions and other random stuff" that you just want to easily add _in addition to_ the core functionality? Eg completely different inputs.

freben (commented):
alright some early thoughts

plugins/catalog-react/src/alpha/blueprints/CatalogContentHeaderBlueprint.tsx
benjdlambert: ~~I also think that this loader definition is also incorrect.~~

~~Loader is `() => Promise<() => JSX.Element | null>` or `() => () => JSX.Element | null`~~

Nevermind, was thinking this was a Component.

docs/features/software-catalog/catalog-customization.md
```diff
@@ -698,3 +698,33 @@ filter:
       target:
         $in: [group:default/admins, group:default/viewers]
 ```
+
+### Catalog Content Header
+
+The catalog page comes with a default content header, but you may want to customize it.
+
+You can do so by creating a new extension for the `catalog` plugin and attaching it to the `contentHeader` input of the `page:catalog` page:
```
benjdlambert: Think we should just use `CatalogContentHeaderBlueprint` in these docs instead?

benjidotsh: oops

plugins/catalog-react/src/alpha/blueprints/CatalogContentHeaderBlueprint.tsx
benjidotsh: I guess this is not applicable anymore since it's now a list of items?

docs/features/software-catalog/catalog-customization.md
```diff
@@ -698,3 +698,36 @@ filter:
       target:
         $in: [group:default/admins, group:default/viewers]
 ```
+
+### Catalog Content Header
+
+The catalog page comes with a content header containing default elements, but you may want to add your own elements to it.
```
awanlin, line 702: I think adding an image here would help as it's not clear what this will all let you customize. This naming - `CatalogContentHeaderItemBlueprint` - makes me think it will replace the whole header but I'm not sure that's actually what it does.

docs/features/software-catalog/catalog-customization.md
```diff
@@ -698,3 +698,36 @@ filter:
...
+  CatalogContentHeaderItemBlueprint.make({
+    name: 'hello-banner',
+    params: {
+      loader: async () => {
+        const { DismissableBanner } = await import(
+          '@backstage/core-components'
+        );
+
+        return (
+          <DismissableBanner
```
awanlin: Does this need a `compatWrapper` or will it truly work copy and paste?

awanlin (changes requested):
Thanks for the contribution @benjidotsh, just reviewing the docs and left a pair of comments 👍

benjidotsh:
not stale, but just got back from holiday + sick leave, so got some things to catch up on first :)

benjidotsh:
not stale, but still haven't had much time

docs/features/software-catalog/catalog-customization.md
benjidotsh: Hmm, I'm not sure why it would need a `compatWrapper`? Isn't `compatWrapper` solely to maintain compatibility with old refs in the NFS between `@backstage/core-*-api` and `@backstage/frontend-*-api`?

benjidotsh:
Feel free to to re-review 😄

benjdlambert:
Hey :wave: 

With the redesign to BUI we're actually looking at rolling out new headers to plugins #31412.

See https://github.com/backstage/backstage/issues/31412#issuecomment-3401692165 for what this looks like for this exact page.

This means that all pages will get options to display these buttons and that the extension point to override them should be factored into that ticket rather than doing this in isolation for the Catalog plugin.

@benjidotsh I want to keep this around as a use case, but it's more than likely that we will implement this in a totally different way than this PR given that parallel work.

benjidotsh:
@benjdlambert does that mean this PR is still relevant for the time being or would you prefer closing this one for now?

benjdlambert:
@benjidotsh sorry missed the ping here. Thinking that we close this one for now, and can always re-open it if it's not solved by the header changes. Trying to cleanup some of the PR's that's in flight currently. :pray:

PR #30343, opened 2025-06-20
Added support for `{org}.visualstudio.com` domains used by Azure DevOps

awanlin:

Added support for `{org}.visualstudio.com` domains used by Azure DevOps. With this change you will be able to import entities into the Catalog via static configuration or the Catalog Import plugin.

### DIFF
--- a/packages/integration/src/azure/DefaultAzureDevOpsCredentialsProvider.test.ts
+++ b/packages/integration/src/azure/DefaultAzureDevOpsCredentialsProvider.test.ts
@@ -445,5 +445,26 @@ describe('DefaultAzureDevOpsCredentialProvider', () => {
         });
       });
     });
+    describe('Azure DevOps (visualstudio.com)', () => {
+      it('Should return a token when a credential with the same organization is specified', async () => {
+        const provider = buildProvider([
+          {
+            host: 'org1.visualstudio.com',
+            credentials: [
+              {
+                organizations: ['org1'],
+                personalAccessToken: 'pat',
+              },
+            ],
+          },
+        ]);
+
+        const credentials = provider.getCredentials({
+          url: 'https://org1.visualstudio.com/project1',
+        });
+
+        expect(credentials).toBeDefined();
+      });
+    });
   });
 });
--- a/packages/integration/src/azure/DefaultAzureDevOpsCredentialsProvider.ts
+++ b/packages/integration/src/azure/DefaultAzureDevOpsCredentialsProvider.ts
@@ -110,6 +110,18 @@ export class DefaultAzureDevOpsCredentialsProvider
     return undefined;
   }
 
+  private forVisualstudioOrganization(
+    url: URL,
+  ): AzureDevOpsCredentialsProvider | undefined {
+    const parts = url.host.split('.');
+    if (url.host.includes('visualstudio.com') && parts.length > 0) {
+      // url format: https://{organization}.visualstudio.com
+      return this.providers.get(`${url.host}/${parts[0]}`);
+    }
+
+    return undefined;
+  }
+
   private forHost(url: URL): AzureDevOpsCredentialsProvider | undefined {
     return this.providers.get(url.host);
   }
@@ -121,6 +133,7 @@ export class DefaultAzureDevOpsCredentialsProvider
     const provider =
       this.forAzureDevOpsOrganization(url) ??
       this.forAzureDevOpsServerOrganization(url) ??
+      this.forVisualstudioOrganization(url) ??
       this.forHost(url);
 
     if (provider === undefined) {
--- a/.changeset/calm-knives-glow.md
+++ b/.changeset/calm-knives-glow.md
@@ -1,0 +1,5 @@
+---
+'@backstage/integration': patch
+---
+
+Added support for `{org}.visualstudio.com` domains used by Azure DevOps
--- a/packages/integration/src/azure/AzureUrl.test.ts
+++ b/packages/integration/src/azure/AzureUrl.test.ts
@@ -198,4 +198,53 @@ describe('AzureUrl', () => {
       ),
     ).toThrow('Azure URL must point to a git repository');
   });
+
+  it('should work with the old visualstudio long URL', () => {
+    const url = AzureUrl.fromRepoUrl(
+      'https://my-org.visualstudio.com/my-project/_git/my-repo',
+    );
+
+    expect(url.getOwner()).toBe('my-org');
+    expect(url.getProject()).toBe('my-project');
+    expect(url.getRepo()).toBe('my-repo');
+    expect(url.getRef()).toBeUndefined();
+    expect(url.getPath()).toBeUndefined();
+  });
+
+  it('should work with the old visualstudio long URL form with a path', () => {
+    const url = AzureUrl.fromRepoUrl(
+      'https://my-org.visualstudio.com/my-project/_git/my-repo?path=%2Ftest.yaml',
+    );
+
+    expect(url.getOwner()).toBe('my-org');
+    expect(url.getProject()).toBe('my-project');
+    expect(url.getRepo()).toBe('my-repo');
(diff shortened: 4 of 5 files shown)


awanlin:
Hi @sanderaernouts, could use your help with reviewing this? This should support all the existing credential patterns used for `dev.azure.com` but I wasn't sure on how to test them all. Happy to make changes based on your suggestions!

sanderaernouts:
> Hi @sanderaernouts, could use your help with reviewing this? This should support all the existing credential patterns used for `dev.azure.com` but I wasn't sure on how to test them all. Happy to make changes based on your suggestions!

@awanlin  I was on holiday until today, but I'll have a look for sure 👍

packages/integration/src/azure/DefaultAzureDevOpsCredentialsProvider.test.ts
```diff
@@ -445,5 +445,26 @@ describe('DefaultAzureDevOpsCredentialProvider', () => {
...
       });
     });
+    describe('Azure DevOps (visualstudio.com)', () => {
+      it('Should return a token when a credential with the same organization is specified', async () => {
+        const provider = buildProvider([
+          {
+            host: 'org1.visualstudio.com',
+            credentials: [
+              {
+                organizations: ['org1'],
```
sanderaernouts: You could argue that this field is not needed since it is embedded in the URL in this case. But I'm not sure how hard that is to enforce in the configuration schema. The reason I added the `organizations` field is because `dev.azure.com` applies to all organizations but the `.visualstudio.com` URLs are organization specific.

It is a potential point for misconfiguration but, if I remember correctly, you can also leave it out and then this credential would apply to "all orgs" for `org1.visualstudio.com` (which technically is only `org1`).
```suggestion
```

packages/integration/src/azure/DefaultAzureDevOpsCredentialsProvider.ts
```diff
@@ -110,6 +110,18 @@ export class DefaultAzureDevOpsCredentialsProvider
     return undefined;
   }
 
+  private forVisualstudioOrganization(
```
sanderaernouts: Microsoft uses "Visual Studio" I believe, in which case this should be an upper-case "S"
```suggestion
  private forVisualStudioOrganization(
```

packages/integration/src/azure/DefaultAzureDevOpsCredentialsProvider.ts
```diff
@@ -110,6 +110,18 @@ export class DefaultAzureDevOpsCredentialsProvider
...
+    url: URL,
+  ): AzureDevOpsCredentialsProvider | undefined {
+    const parts = url.host.split('.');
+    if (url.host.includes('visualstudio.com') && parts.length > 0) {
+      // url format: https://{organization}.visualstudio.com
+      return this.providers.get(`${url.host}/${parts[0]}`);
```
sanderaernouts: Reading through the code, as long as you don't specify `organizations` you should be able to simply do this:
```suggestion
      return this.providers.get(`${url.host}`);
```

This could be another argument to enforce this through the config format. Right now I'm not sure if the code will work for this configuration:
```yaml
host: 'org1.visualstudio.com',
credentials: [
 {
    personalAccessToken: 'pat'
 }
]
```

sanderaernouts: Thinking about it, it would be weird to even have more than one credential for this URL format. But that would mean (re)introducing the `credential` property.

sanderaernouts (commented):
@awanlin I'm thinking about the `organizations` property on the credential (configuration) object, which does not make sense for org specific URLs and the support for multiple credentials which also does not make sense for org specific URLs since you can only have a (more) specific credential for one or more orgs.

But, since the `visualstudio.com` URLs are "legacy" (right?) and as far as I know you cannot create new Azure DevOps accounts with that URL format I'm a bit torn between two opinions. 

On the one side, it's not that bad in my opinion to have some unnatural things in the configuration schema for those "legacy" URLs.

On the other side it feels more clean to to have `credential` instead of `credentials` for these type URLs and to remove the `organizations` field in this case to prevent misconfiguration and confusion.

I guess it depends on how broad a user base you expect for this. I would suggest, either way, to also update [the documentation](https://backstage.io/docs/integrations/azure/locations) as part of this change to include the `visualstudio.com` format and it's potential limitations or caveats.

awanlin:
Hi @sanderaernouts, thanks for the review, appreciate your feedback! Seems we are missing each other as now I'm on vacation. One thing I wanted to mention about this, you can easily toggle these legacy domains in the UI and MS has no clear plans to remove them any time soon. This is why I ended up submitting this PR. I'm gine with some rough edges for these to be honest as they will be rare. 👍

sanderaernouts:
@awanlin I see that I can indeed enable the old style URL on companies Azure DevOps org through the settings. I thought this was already disabled from the UI and that you could not go back. But still I agree this should be uncommon as I expect most orgs to be on the `dev.azure.com` domain.

So some rough edges are acceptable but it would be good to document the limitations and caveats.

And last but not least, enjoy your holiday 🌴

packages/integration/src/azure/core.ts
```diff
@@ -53,3 +53,14 @@ export function getAzureDownloadUrl(url: string): string {
...
+
+/**
+ * Given a URL, return true if it contains `visualstudio` and false if it does not
+ * URLs cna be in these two formats: `dev.azure.com/{org}` or the legacy `{org}.visualstudio.com`
+ *
+ * @param origin - A URL origin string pointing to an Azure DevOps instance
+ * @public
+ */
+export function isVisualStudioDomain(origin: string): boolean {
+  return origin.split('.')[1] === 'visualstudio';
```
Rugvip: Is this always and origin and are visualstudio domains always .com, in that case it might be cleaner with `origin.endsWith('visualstudio.com')`. There's a potential issue here where submitting something like `https://foo.visualstudio.evil.hacker.site.com` could be troublesome. Thinking it would be good to make this a bit more strict in some way, just for safety.

awanlin:
Hi @sanderaernouts, just a heads up I'm picking this back up. I've addressed the CodeQL comments, I'll be following up on the comments you left previously by the end of the week. Hopefully I'm not reaching out when you are on vacation again 😉

packages/integration/src/azure/DefaultAzureDevOpsCredentialsProvider.ts
awanlin: I'd like to avoid brining that back. I think we can document around this.

packages/integration/src/azure/core.ts
awanlin: Yes, everywhere I'm using this I'm safely able to get the `origin` value. I'll change this to `endsWith`, that is cleaner and makes more sense

awanlin:
The code in its current state allowed me to manually import `catalog-info.yaml` files from these legacy domains using the Catalog Import UI. I've setup the discovery entity provider by have hit some errors in the `CachedAzureDevOpsCredentialsProvider`, I'll work through that next. Once that's working I'll be able to add documentation about support for these legacy domains as part of this PR 👍

awanlin:
Alright, this is ready for review. Static config, manual import, and discovery are all working. I've also updated the related docs where needed. Tried to make this change as small as possible 👍

sanderaernouts:
> Hi @sanderaernouts, just a heads up I'm picking this back up. I've addressed the CodeQL comments, I'll be following up on the comments you left previously by the end of the week. Hopefully I'm not reaching out when you are on vacation again 😉

Hi @awanlin,

Apologies for the delayed response. I only saw your message just now. I'm not on holiday this time, but I do have some pressing stuff going on in my private life that requires most of my attention at the moment.

The PR is already merged, but I did have a quick look at the changes you made. Especially at the Azure DevOps authentication code. LGTM 👍

PR #30357, opened 2025-06-23
Canon - New SearchField component

cdedreuille:

### DIFF
--- a/packages/canon/src/components/Heading/Heading.tsx
+++ b/packages/canon/src/components/Heading/Heading.tsx
@@ -37,13 +37,13 @@ function HeadingComponent<T extends ElementType = 'h1'>(
   const { classNames, dataAttributes } = useStyles('Heading', {
     variant,
     color,
-    truncate,
   });
 
   return (
     <Component
       ref={ref}
       className={clsx(classNames.root, className)}
+      data-truncate={truncate}
       {...dataAttributes}
       style={style}
       {...restProps}
--- a/packages/canon/src/components/Text/Text.tsx
+++ b/packages/canon/src/components/Text/Text.tsx
@@ -39,13 +39,13 @@ function TextComponent<T extends ElementType = 'p'>(
     variant,
     weight,
     color,
-    truncate,
   });
 
   return (
     <Component
       ref={ref}
       className={clsx(classNames.root, className)}
+      data-truncate={truncate}
       {...dataAttributes}
       style={style}
       {...restProps}


package.json
```diff
@@ -103,13 +103,13 @@
...
     "@types/react-dom": "^18.0.0",
     "@yarnpkg/plugin-npm@npm:^3.1.0": "patch:@yarnpkg/plugin-npm@npm%3A3.1.0#~/.yarn/patches/@yarnpkg-plugin-npm-npm-3.1.0-6533d0f5a1.patch",
+    "GendocuPublicApis": "npm:gendocu-public-apis@^1.0.0",
     "ast-types@0.14.2": "patch:ast-types@npm%3A0.14.2#./.yarn/patches/ast-types-npm-0.14.2-43c4ac4b0d.patch",
     "ast-types@^0.14.1": "patch:ast-types@npm%3A0.14.2#./.yarn/patches/ast-types-npm-0.14.2-43c4ac4b0d.patch",
     "csstype@npm:^3.0.2": "3.0.9",
     "csstype@npm:^3.1.2": "3.0.9",
     "csstype@npm:^3.1.3": "3.0.9",
-    "jest-haste-map@^29.7.0": "patch:jest-haste-map@npm%3A29.7.0#./.yarn/patches/jest-haste-map-npm-29.7.0-e3be419eff.patch",
-    "GendocuPublicApis": "npm:gendocu-public-apis@^1.0.0"
```
benjdlambert: Think you'll need to bring this back in order for the build to pass.

cdedreuille: It should all be good to go now 🙏

benjdlambert:
@Rugvip another one for you to have a look when you're back, but merging to unblock.

canon-docs/scripts/build-css.js
```diff
@@ -13,7 +13,7 @@ const bundleAndCopyFile = async (source, destination) => {
   try {
     const result = await bundle({
       filename: source,
-      minify: true,
```
benjdlambert: Is this an intended change?

cdedreuille: We did remove the minify in the package per @Rugvip's request but you're right in that case it's better to keep it minified and make it faster for production. I also think that Next.js is bundling this CSS so this would not have any effect.

cdedreuille: I brought it back.

benjdlambert: I think that we would minify it anyways in our tooling, was just checking.

Happy to do either. If we're going to keep the `minification` on though we should regenerate the styles right?

cdedreuille: Ah yes you're right. I just rebuilt it. Thank you 🙏

PR #30376, opened 2025-06-24
fix: incremental burstLength check

chriswyattcook:

The IncrementalIngestionEngine did not support burstLength for making API calls in a burst, meaning Incremental Entity Providers did burst API calls until the last cursor of data without respecting the burstLength. This PR adds a burstLength check after one burst completes to ensure the bursting stops once the burstLength has passed. Also added a new test file (IncrementalIngestionEngine.test.ts) to test this functionality.

### DIFF
--- a/.changeset/fancy-nails-call.md
+++ b/.changeset/fancy-nails-call.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-incremental-ingestion': patch
+---
+
+Fixed bug in IncrementalIngestionEngine by adding burstLength check when a burst completes
--- a/.changeset/fancy-nails-call.md
+++ b/.changeset/fancy-nails-call.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-backend-module-incremental-ingestion': patch
 ---
 
-Fixed bug in IncrementalIngestionEngine by adding burstLength check when a burst completes
+Fixed bug in `IncrementalIngestionEngine` by adding `burstLength` check when a burst completes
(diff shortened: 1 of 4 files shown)


.changeset/fancy-nails-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-incremental-ingestion': patch
+---
+
+Fixed bug in IncrementalIngestionEngine by adding burstLength check when a burst completes
```
freben:
```suggestion
Fixed bug in `IncrementalIngestionEngine` by adding `burstLength` check when a burst completes
```

plugins/catalog-backend-module-incremental-ingestion/src/engine/IncrementalIngestionEngine.ts
```diff
@@ -247,6 +249,14 @@ export class IncrementalIngestionEngine implements IterationEngine {
         });
         if (signal.aborted || next.done) {
           break;
+        } else if (
+          performance.now() - start >
+          this.burstLength.as('milliseconds')
+        ) {
+          this.options.logger.info(
+            `incremental-engine: Ingestion '${id}' burst exceeded length of ${this.burstLength.toHuman()}.`,
```
freben:
```suggestion
            `incremental-engine: Ingestion '${id}' burst ending after ${this.burstLength.toHuman()}.`,
```

freben (dismissed):
Wow this is an interesting find; it probably actually was timeout-driven before this, if even that (the scheduler did have its timeout set based on this, and triggers an abort signal based on it).

I just tweaked some strings.

freben:
Sorry for the noise - there's a test error actually. But that test probably should be rewritten because it seems to assert the wrong thing. Could you check it out?

chriswyattcook:
> But that test probably should be rewritten because it seems to assert the wrong thing

adjusted the tests to remove the logger.info assertions, and replaced with duration, number of calls

PR #30380, opened 2025-06-24
Fix a blocking behavior of the dynamic plugin scanner on MacOS/

davidfestal:

This PR fixes a platform-specific blocking bug that occurs only on MacOS, when running Backstage locally with the dynamic plugins enabled. It also fixes unit tests accordingly. 

### DIFF
--- a/packages/backend-dynamic-feature-service/src/scanner/plugin-scanner.ts
+++ b/packages/backend-dynamic-feature-service/src/scanner/plugin-scanner.ts
@@ -236,7 +236,7 @@ export class PluginScanner {
         this.rootDirectoryWatcher = chokidar
           .watch(this._rootDirectory, {
             ignoreInitial: true,
-            followSymlinks: true,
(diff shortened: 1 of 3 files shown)


packages/backend-dynamic-feature-service/src/scanner/plugin-scanner.ts
```diff
@@ -236,7 +236,7 @@ export class PluginScanner {
         this.rootDirectoryWatcher = chokidar
           .watch(this._rootDirectory, {
             ignoreInitial: true,
-            followSymlinks: true,
+            followSymlinks: false,
```
freben, line 239: What was it about this that made it actually block?

Would it be prudent to put a comment right above this field to explain why it's set to false? It seems like the type of thing I'd be likely to discover some day down the line and set to true because "why is this disabled" :) If someone has a setup where they actually want to leverage symlinks.

PR #30381, opened 2025-06-24
Added more Non technical FAQ

AmbrishRamachandiran:
Added more Non technical FAQ

### DIFF
--- a/docs/faq/product.md
+++ b/docs/faq/product.md
@@ -6,61 +6,64 @@ description: Questions related to product and design.
 
 ### Can we call Backstage something different? So that it fits our company better?
 
-Yes, Backstage is just a framework for building your own developer portal. We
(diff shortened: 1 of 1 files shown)


Parsifal-M:
I think before we add more stuff to these FAQs I'd like to possibly discuss in the Docs SIG what the intention of these FAQs are, otherwise I'm not sure where we draw the line on _**what should be in the FAQs**_ (as you could expand it almost endlessly)

awanlin:
Hi @AmbrishRamachandiran, just wanted to follow up on this after we covered it during the Docs SIG. We won't be adding more items to the FAQ with our goal to be having these properly documented in the documentation instead.

AmbrishRamachandiran:
Thanks for the clarification @awanlin  That makes sense !!

PR #30387, opened 2025-06-25
Just some more circular dep cleanup

freben:
Just a bunch of places importing from index files, or from its own package name, leading to loops


plugins/kubernetes-node/report.api.md
```diff
@@ -256,7 +255,7 @@ export class PinnipedHelper {
   constructor(logger: LoggerService);
   // (undocumented)
   tokenCredentialRequest(
-    clusterDetails: ClusterDetails_2,
```
benjdlambert, line 259: Oh my god this has been driving me wild why everything is always duplicated in the `api-reports`.

Turns out its partly this 🙃

freben: Indeed!

freben: It might make sense to make a lint rule against importing your own library. The tricky thing is, there may be some legitimate fringe uses of it across entrypoints etc.

vinzscam (approved):
✨

Issue #30328, opened 2025-06-19
Default backend provides a new MetricsService


kurtaking:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
core service metric metrics otel
```

### 🗃️ Project Area

Core Framework

### 🔖 Need

**Core and Community Plugin Authors** gain a straightforward, documented way to emit metrics, and can focus on business logic instead of reimplementing metrics plumbing. The **community** receives a standard approach to metrics.

**Backstage Operators/SRE/Platform Engineers** receive a consistent, reliable stream of metrics from all plugins and the core system, simplifying monitoring, alerting, and troubleshooting. They can enforce metrics standards, aggregate data, and build dashboards more easily. Organizations with their own observability or compliance requirements can override the core service to integrate with proprietary or custom telemetry solutions, without forking or modifying plugins.

### 📝 Proposal

Introduce a core metrics service (`MetricsService`) to the Backstage backend. The service will support root and plugin-scoped usage, enforce naming and labeling conventions, and provide an override mechanism for custom implementations.

* A new core service `metricsServiceRef` will be introduced, similar to other core services (logger, config, httpAuth).
* The service will provide methods to create counters, histograms, and observable gauges, following OpenTelemetry standards.
* Both root and plugin-scoped APIs will be supported, allowing plugins to easily register metrics with proper namespacing and labeling.
* The service will provide a default OpenTelemetry-based implementation, with an option for users to override with their own implementation.

---

```ts
export interface MetricsService {
  forPlugin(pluginId: string): PluginMetricsService;

  createCounter(...): CounterMetric;
  createObservableGauge(...): ObservableGaugeMetric;
  createHistogram(...): HistogramMetric;

  // expose meter for advanced use cases ?
  getMeter(): Meter;
}
```

```ts
// Example based on https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/database/metrics.ts
const metrics = metricsService.forPlugin('catalog-backend');

// update to use metrics service
initDatabaseMetrics(knex: Knex, metrics: MetricsService)

metrics.createObservableGauge('catalog_entities_count', {
    description: 'Total amount of entities in the catalog, grouped by kind',
    observe: async observer => {
      const results = await knex<DbSearchRow>('search')
        .where('key', '=', 'kind')
        .whereNotNull('value')
[...]

Issue #27229, opened 2024-10-17
🚀 Feature: Add Persistence Metrics for Database Performance and Availability in OpenTelemetry


ypetrych:
### 🔖 Feature description

Request to add Persistence Metrics to track database performance and availability in Backstage using OpenTelemetry. Metrics should cover query execution times, connection pool usage, and error rates, exposed via Prometheus or supported telemetry backends.

### 🎤 Context

Why It’s Needed:
- **Enhanced Observabilit**y: Provides deeper insights into database interactions, currently lacking in OpenTelemetry metrics.
- **Performance Monitoring**: Helps identify bottlenecks, optimize queries, and ensure database reliability.
- **Faster Incident Response**: Enables quicker isolation of database issues, improving resolution times during incidents.

These metrics will improve database monitoring and performance within Backstage, offering more reliable system operations.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

jhaals:
Sounds like a great thing to have!

Nilay1999:
@jhaals i can look into this if no one has worked on it yet

Rugvip:
@Nilay1999 on this one is available 👍

I think it'd be best to start with suggesting an approach to this here, what metrics etc. Going straight to PR on this one might be a bit tricky if we're not already aligned on what metrics to expose.

We should also explore possible deps to help out here, like https://www.npmjs.com/package/@opentelemetry/instrumentation-knex but for metrics rather than tracing

Nilay1999:
Yah agreed @Rugvip , Actually i was going through current implementation of Backstage's metrics.
As per my knowledge, connection pool usage state might be easy but query execution time calculation is kinda complex.

Nilay1999:
`total_used_connections: meter
      .createObservableGauge('total_used_connection', {
        description: 'Total used connection pool',
      })
      .addCallback(async gauge => {
        const usedConnection: number = knex.client.pool.numUsed();
        gauge.observe(Number(usedConnection));
      }),
    total_free_connections: meter
      .createObservableGauge('total_free_connections', {
        description: 'Total free connection pool',
      })
      .addCallback(async gauge => {
        const freeConnection: number = knex.client.pool.numFree();
[...]

PR #30391, opened 2025-06-25
BEP: MetricsService

kurtaking:

Proposal to add new metrics services to the Backstage framework. 

```ts
interface MetricsService {
  // Synchronous instrumentation
  createCounter(name: string, options?: MetricOptions): Counter;
  createUpDownCounter(name: string, options?: MetricOptions): UpDownCounter;
  createHistogram(name: string, options?: MetricOptions): Histogram;
  createGauge(name: string, options?: MetricOptions): Gauge;

  // Asynchronous instrumentation
  createObservableCounter(
    name: string,
    options?: MetricOptions,
  ): ObservableCounter;
  createObservableUpDownCounter(
    name: string,
    options?: MetricOptions,
  ): ObservableUpDownCounter;
  createObservableGauge(name: string, options?: MetricOptions): ObservableGauge;

  // Future - add additional convenience methods as we learn more about the needs of the framework
}

interface RootMetricsService extends MetricsService {
  static forRoot(): RootMetricsService
  forPlugin(pluginId: string): MetricsService
  // ...
}
```

---

### Todos

- [x] Confirm tracing is out of scope for this deliverable
- [ ] Add strong guidance and examples as to when to use or not use the metrics service
- [ ] Finalize the naming convention (still need to read through [otel semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/metrics/) and [prometheus recommendations](https://prometheus.io/docs/practices/naming/#metric-names)
- [ ] Add section on deprecating and migrating from existing metrics

---

### Related Issues 

_Note: there may be issues missed_

- https://github.com/backstage/backstage/issues/30328
- https://github.com/backstage/backstage/pull/30224
- https://github.com/backstage/backstage/issues/27229
- https://github.com/backstage/backstage/issues/22665
- https://github.com/backstage/backstage/issues/22555
- https://github.com/backstage/backstage/issues/14505
- https://github.com/backstage/backstage/issues/1519
- https://github.com/backstage/backstage/pull/29345
- https://github.com/backstage/backstage/pull/17534
- https://github.com/backstage/community-plugins/issues/2423

### One-off implementations

_Note: there may be other one-off implementations not listed - refer to search query at the bottom_

- (metrics) catalog backend [[ref](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/util/metrics.ts)]
- (metrics) scaffolder-backend [[ref](https://github.com/backstage/backstage/blob/master/plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts)]- 
- (metrics) notifications-backend [[ref](https://github.com/backstage/backstage/blob/master/plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts)]
- (metrics) events plugin todo [[comment](https://github.com/backstage/backstage/blob/c6cb6ce6ba1cdbbfad3dfe0011c7f322bcb55670/plugins/events-node/src/api/DefaultEventsService.ts#L383)]

Additional results from [search query](https://github.com/search?q=repo%3Abackstage%2Fbackstage+%22opentelemetry%22+language%3ATypeScript&type=code&l=TypeScript)


beps/0012-telemetry-service/README.md
```diff
@@ -0,0 +1,140 @@
...
+6. ...TBD
+
+## Dependencies
+
+1. The root telemetry service MUST BE initialized as EARLY as possible to prevent dependents from receiving no op meters, while still allowing for app-wide configuration.
+2. There are one-off implementations of telemetry in the wild that may conflict with the proposed service. However, this is unlikely to be a problem as the SDK should continue to pick things up.
+
+## Alternatives
+
+1. Plugin authors continue to implement their own telemetry as they see fit.
```
drodil: How about as an alternative just `analyticsService` that can be used from backend plugins like `analyticsApi` is currently used from frontend? Expanding the analytics backend would be another story. WDYT?

kurtaking: That still requires the addition of a new `AnalyticsService`, correct? Or, am I missing something that already exists?

drodil: Yes it requires that. `analyticsApi` is available in frontend but there's really nothing typed for backend analytics

drodil: Basically this one: https://github.com/backstage/backstage/blob/master/packages%2Fcore-plugin-api%2Fsrc%2Fapis%2Fdefinitions%2FAnalyticsApi.ts

I think the same way `catalogService` <-> `catalogApi` is done could work for this as well 🤔

kurtaking:
I've been spinning on how the core service provides a nice interface for plugin authors to implement telemetry in their plugins while also providing utilities at a plugin-level for modules to use? The example that comes to mind are catalog processors. 

Should the catalog backend be responsible for providing plugin-level telemetry utils, such as `trackEntityIngestion`, `trackProcessorDuration`, etc that modules can leverage for consistency, or can the backend handle the telemetry around each registered processor (or provider, etc)?

beps/0012-telemetry-service/README.md
freben: At the same time, the terms "metrics" and "tracing" are very established for backend (and align with what otel itself calls them). Could it be the case that analytics in frontend and metrics in backend are a bit different from each other in nature, and therefore it's OK that the nomenclature doesn't align?

I'll also say, maybe metrics and tracing are so different (and might be powered by different platforms even at times?) that they should be two different services? Not sure there honestly.

freben:
Oh I'd expect the surrounding plugin (the catalog) to handle that itself for generic concerns like processing.

But at the same time, maybe a module adds additional module specific metrics too, like maybe the incremental ingestion module has some of its own, separate from the host catalog.

beps/0012-telemetry-service/README.md
kurtaking: We need a telemetry version of `coreServices` 👀.  `telemetryServices.tracing|metrics|logging`. In seriousness, I think there is value in having them separate. They can grow at their own pace while still being leveraged together to solve for _telemetry_.

kurtaking:
> Oh I'd expect the surrounding plugin (the catalog) to handle that itself for generic concerns like processing.
> 
> But at the same time, maybe a module adds additional module specific metrics too, like maybe the incremental ingestion module has some of its own, separate from the host catalog.

Yeah, say I want additional metrics on my custom catalog processors. How do modules attach additional metrics while staying in the guidelines outlined by the plugin? How are conflicts handled, etc?

kurtaking:
Sharing a [branch](https://github.com/kurtaking/backstage/pull/2/files) that consolidates the existing catalog metrics into a single class. It's still a wrapper that isn't offering much value, but I'm hoping it highlights some patterns to build on.

beps/0012-telemetry-service/README.md
ScottGuymer: For me `telemetry` is a much better and more instantly understood term for backend. Encompassing metrics, tracing, etc.

When I hear analytics I instantly think front end.

freben: I'm just arguing that it's probably a bad thing that the word telemetry covers metrics and tracing - since we might want to explicitly separate the interfaces.

andrewthauer: I think having separate services for metrics & tracing makes most sense. The `LoggerService` is already its own. The configuration and concepts don't really overlap much from the perspective of a consumer of these. I don't think analytics fits and would rather this maps to OTEL as close as possible.

kurtaking: I am headed in the direction of having two additional services. And agree @andrewthauer, they will complement the existing `LoggerService` nicely. 

I would still like to solve for setting up telemetry aka, deprecate the need for these [docs](https://backstage.io/docs/tutorials/setup-opentelemetry/), but doing that after the three services are in place sounds more reasonable.

kurtaking: I am going to scope this back to only focus on metrics and then we follow up with a separate BEP for tracing. That puts the three _pillars_ (logger, tracer, metrics) of telemetry in place and then we can focus on overall improvements from there.

beps/0012-metrics-naming-conventions/README.md
```diff
@@ -0,0 +1,113 @@
+---
+title: Metrics Naming Conventions for Backstage
+status: implementable
```
awanlin: Is this actually true? I don't see a review by any of the [project owners](https://github.com/backstage/backstage/tree/master/beps#bep-status) and there's already a [draft for code](https://github.com/backstage/backstage/pull/30438), should we get the proper review first?

beps/0013-metrics-service/README.md
```diff
@@ -0,0 +1,169 @@
+---
+title: Backstage Metrics Service
+status: implementable
```
awanlin: Same here: Is this actually true? I don't see a review by any of the [project owners](https://github.com/backstage/backstage/tree/master/beps#bep-status) and there's already a [draft for code](https://github.com/backstage/backstage/pull/30438), should we get the proper review first?

awanlin (commented):
Hi @kurtaking, sorry to be a stickler for process but can we make sure this gets the proper review before we start coding? If this was reviewed in a SIG and I missed that it should be noted here and this probably should be merged in.

beps/0012-metrics-naming-conventions/README.md
awanlin: Looks like this is a bit unclear what the initial status should be so understand why you would have used this but we should get this approved before we move forward with code.

awanlin (commented):
Also, looks like `0012` and `0013` are used by existing BEP PRs.

beps/0013-metrics-service/README.md
kurtaking: Consolidate into the other [thread](https://github.com/backstage/backstage/pull/30391#discussion_r2183235325).

beps/0012-metrics-naming-conventions/README.md
kurtaking: I think this is the expected status. This is what comes with the template and there isn't a better BEP Status to put while the PR is under review. 🤷🏼‍♂️ The provision state is [deprecated](https://github.com/backstage/backstage/blob/master/beps/README.md#whats-does-it-mean-for-a-bep-to-be-in-provisional-state).

> sorry to be a stickler for process but can we make sure this gets the proper review before we start coding?

As far as the draft code goes, that's okay in my opinion. I'm experimenting and it's in a draft state for a reason. If that's somehow noisy, then I am fine with doing it in my own fork. For example, I was already experimenting on refactoring the catalog plugin [here](https://github.com/kurtaking/backstage/pull/2/files), although that is no longer relevant. 

This [line](https://github.com/backstage/backstage/blob/6ef06cc84e555de13910814a0dc6798d2805716e/beps/README.md?plain=1#L18) is in the docs 

> While the BEP PR is open the owner can open separate PRs to ship experimental features in support of the BEP to help drive the design forward, and these may be merged pending usual PR review. If the BEP is withdrawn or rejected these features should generally be removed.

All-in-all, perhaps I should not have opened the BEP in general and kept discussion to the initial issue I opened. However, I felt it was hard to separate topics into their own conversation threads.

kurtaking:
> Also, looks like `0012` and `0013` are used by existing BEP PRs.

I noticed that too. 

Should BEPs get named in the order they merge, or in the order in which they are created. My approach was going to be to check the BEPs folder on `master` to take the next number before merging. 

i.e., if this BEP was merged before those, I assumed they need to be updated, not this.

beps/0012-metrics-naming-conventions/README.md
kurtaking: Your second message didn't load in my screen before I responded.

awanlin: We're good regarding the status, probably something we should update in the BEP docs to include a provisional state. 👍 

I would expect at least some sort of initial review by area owners before draft code otherwise why not just go directly to the draft or PRFC as they have also been called. As someone wanting to review this it made it feel like a direction was already decided so why should I bother reviewing this and/or sharing my thoughts on the direction. It also spreads out where conversations are happening in the early stages. But I get that's not how the process is structured.

beps/0013-metrics-service/README.md
awanlin: Let's resolve this then 👍

awanlin:
> > Also, looks like `0012` and `0013` are used by existing BEP PRs.
> 
> I noticed that too.
> 
> Should BEPs get named in the order they merge, or in the order in which they are created. My approach was going to be to check the BEPs folder on `master` to take the next number before merging.
> 
> i.e., if this BEP was merged before those, I assumed they need to be updated, not this.

Not sure which is why I wanted to mention it.

beps/0012-metrics-naming-conventions/README.md
kurtaking: I didn't expect many (if any) people to start looking at the draft PR until I pointed them in that direction. Will keep it in mind as this is a first and learning experience for me. 

> otherwise why not just go directly to the draft or PRFC

Yeah.. that makes sense. Lines up with me _All-in-all, perhaps I should not have opened the BEP in general and kept discussion to the initial issue I opened._ comment. 

That probably would have turned into issue + pr to iterate on.

kurtaking: Created https://github.com/backstage/backstage/issues/30441

kurtaking: Pushed up the last local changes I had to draft PR. Will stop there for now.

awanlin: Sorry, my intention wasn't to be blocking things. For context I've been on vacation and I'm catching up. Observability and making life easier for Backstage Admins is a core focus for me so I have "thoughts" on the subject. So when I didn't see a review here but code in a Draft PR things felt out of sync to me.

kurtaking: No, that's okay. I don't feel like you are blocking things. It's awesome to hear this is a focus area for you and look forward to hearing your thoughts on the subject. I did not bring it up in the last framework sig, but hoped to do so in the next. Give me some time to get things in order on this PR and then I'll ping you for input.

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
+---
+title: Backstage Metrics Service
+status: implementable
+authors:
+  - '@kurtaking'
+owners:
+  - '@kurtaking'
+project-areas:
+  - core-framework
+  - observability
```
awanlin: I know there was talk of creating an observability project area when the auditor project area was created. Is this something your still interested in @schultzp2020? We created the notification project area well after it's first BEP so we can probably push that out.

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Design Details
+
+### Configuration
+
+<!--
+todo: I want to provide full configuration via the app-config, but that requires us wrapping the entire SDK and handling the `instrumentation.js`
+      instructions. I'm not sure we want to start there.
+
+Metrics configuration will continue to be defined by the adopter in their `instrumentation.js` file as outlined in the [setup-opentelemetry](https://backstage.io/docs/tutorials/setup-opentelemetry/) docs.
```
awanlin: This isn't seen in the BEP when viewing it as MD, but seems worthwhile to share with readers.

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
...
+## Proposal
+
+Following similar patterns to other core services, create a new `RootMetricsService` responsible for initializing metrics, root-level concerns, and the creation of plugin-specific `MetricsService` instances. The root service delegates to the plugin-scoped `MetricsService` to initialize a meter for each registered plugin based on the `serviceName` and optional `serviceVersion` provided in the app's `app-config.yaml` file. This is based on the recommendation in the OpenTelemetry [documentation](https://opentelemetry.io/docs/languages/js/instrumentation/#acquiring-a-meter).
+
+## Design Details
+
+### Configuration
+
+<!--
+todo: I want to provide full configuration via the app-config, but that requires us wrapping the entire SDK and handling the `instrumentation.js`
```
awanlin: Why? Would like to know your reasoning for wanting this. I have no preferences. 

From my poking around `getNodeAutoInstrumentations()` brings in a lot of things that maybe we don't need and limiting this could reduce size but I also wonder if this is out of scope.

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
...
+
+The initial alpha release will include a simple configuration for the root metrics service.
+-->
+
+```yaml
+backend:
+  metrics:
+    serviceName: backstage
+    # todo: this is optional - should we just default it to the current Backstage release and allow override?
+    serviceVersion: 0.0.1
```
awanlin: Yeah, the Backstage version seems a reasonable default but Adopters may also tag their releases and that version would be handy here.

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
...
+    return `${this.serviceName}.plugin.${this.pluginId}.${name}`;
+  }
+}
+```
+
+### Naming Conventions
+
+All Backstage metrics MUST follow this hierarchical pattern:
+
+`backstage.{scope}.{scope_name}.{category}.{metric_name}`
```
awanlin: How do we plan to handle the existing metrics, will those stay for some period with a deprecation wanting before being removed? None of them match this pattern: https://backstage.io/docs/tutorials/setup-opentelemetry#available-metrics. If this is the plan we should also have some way to note deprecated metrics in the logs as well.

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
...
+
+**Where:**
+
+- `backstage` is the root namespace for all Backstage metrics
+- `{scope}` is the system scope (`pluginId`, `core` for core services)
```
awanlin: What is considered `core`? Would be good to know some example and maybe we can spot a convention for this.

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
...
+- `{scope_name}` is the name of the scope (e.g. `my-plugin`, `catalog`, `scaffolder`, etc.)
```
awanlin: Won't we have duplication here with `{scope}.{scope_name}`? `catalog` is a `pluginId` same with `scaffolder`. You'd end up with `catalog.catalog` and `scaffolder.scaffolder`. Is there a case you have in mind for why we need both?

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
...
+##### Plugin-Scoped Metrics
+
+Pattern: `backstage.plugin.{pluginId}.{category}.{metric_name}`
+
+```yaml
+# Examples
+backstage.plugin.catalog.entities.processed.total
+backstage.plugin.scaffolder.tasks.completed.total
+backstage.plugin.techdocs.builds.active
+backstage.plugin.auth.sessions.active.total # todo: technically a core service and a backend plugin
```
awanlin: OK, think I'm following now. But you say `{scope}` is `pluginId` or `core` but here you have the word "plugin"

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  // Ensures consistency across all metrics
+  private prefixMetricName(name: string): string {
+    return `${this.serviceName}.plugin.${this.pluginId}.${name}`;
+  }
+}
+```
+
+### Naming Conventions
```
awanlin: Just sharing this as a potential source of guidance: https://prometheus.io/docs/practices/naming/#metric-names. Otel sent me here from their docs which are a bit limited in this area IMO - https://opentelemetry.io/docs/specs/semconv/general/metrics/

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
...
+- `{category}` is the logical grouping within the `{scope}`
+- `{metric_name}` is the name of the metric
+
+The metric name will be the name as provided by the plugin author in the form `{category}.{name}`.
+
+```ts
+const rootMetrics = DefaultRootMetricsService.fromConfig(config);
+const pluginMetricsService = rootMetrics.forPlugin('my-plugin');
+
+const metric = pluginMetricsService.createCounter('my_category.my_metric');
```
awanlin: Can you clarify, not sure if this intended or just rough example code, but when I use `coreServices` in my `deps` like this - https://github.com/backstage/backstage/blob/72f52b6ff964e9f6202874de718d9b8fb0cedc29/plugins/devtools-backend/src/plugin.ts#L33 - I would expect that doing something like `metricsService: coreServices.metricsService` would get me an instance that's coped to my plugin and I would just be off and running making `metricService.createCounter()` calls as needed. Is that the way this will work as well?

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
...
+1. Create a new `RootMetricsService` that initializes metrics and app-wide concerns.
+1. Create the plugin-scoped `MetricsService` that provides a metrics service for plugins.
+1. Create alpha-related documentation to add to existing core service [docs](https://backstage.io/docs/backend-system/core-services/index).
+1. Release the metrics service as alpha.
+1. Refactor catalog and scaffolder implementations to use the new (alpha) `MetricsService`.
+1. Release the metrics service
+1. Offer a migration path for existing adopters to migrate to the new metrics service.
+1. Update all documentation to reference the new metrics service.
+1. Create follow-up action items to integrate the new metrics service into the core system.
+1. Fully deprecate all existing metrics implementations like the existing Prometheus one-off implementations.
```
awanlin, line 234: Should we do this sooner, like now? At least for things in the main Backstage repo.

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,255 @@
...
+### References
+
+- [General naming considerations](https://opentelemetry.io/docs/specs/semconv/general/naming/#general-naming-considerations)
+- [Acquiring a meter](https://opentelemetry.io/docs/languages/js/instrumentation/#acquiring-a-meter)
+
+## Release Plan
+
+1. Create a new `RootMetricsService` that initializes metrics and app-wide concerns.
+1. Create the plugin-scoped `MetricsService` that provides a metrics service for plugins.
+1. Create alpha-related documentation to add to existing core service [docs](https://backstage.io/docs/backend-system/core-services/index).
```
awanlin, line 226: One thing that I think will be important is having strong guidance and examples as to when to use or not use this new service.

awanlin (commented):
Great work @kurtaking, finally had some time to review this end to end. 🚀

beps/0012-telemetry-service/README.md
kurtaking: Resolving this thread. Much of the BEP has been updated to reflect focusing on a metrics service. Feel free to review and leave additional CR 😄

beps/0012-metrics-service/README.md
kurtaking: I need a better way of tracking todos, outstanding questions, and areas where I am looking for input. Was thinking of a checklist in the PR description.

beps/0012-metrics-service/README.md
kurtaking: Would it be worth adding a section on deprecating and migrating from the existing set of metrics? I only mention a single line item on this in the release plan section.

beps/0012-metrics-service/README.md
kurtaking: When I think of `core`, I think of the _Core Backend Service APIs_ listed [here](https://backstage.io/docs/backend-system/core-services/index/).

beps/0012-metrics-service/README.md
kurtaking: I've fixed this in my latest commit.

beps/0012-metrics-service/README.md
kurtaking: I will add this to my resources section and use it as guidance. Really want to lean on the existing guidance OTEL provides around naming patterns configuration, etc.

beps/0012-metrics-service/README.md
kurtaking: > Is that the way this will work as well?

Yes, this is a rough example outlining what is happening behind the scenes. I can think of something better.

beps/0012-metrics-service/README.md
kurtaking: Yeah, it's worth marking them deprecated as one of the first tasks in this list.

kurtaking:
**edit**: can view the poc by visiting https://github.com/backstage/backstage/pull/30495

@Rugvip - check out this [branch](https://github.com/backstage/backstage/compare/master...kurtaking:backstage:instrumentation-initializer). I've deleted the `instrumentation.js` file and removed the `--require` from the start script.

The `RootMetricsService` handles starting the SDK with provided configuration from `app-config`. There is some repetitiveness, but this is a good POC to show that it's possible. 

I added a `MetricsService` to the catalog plugin [here](https://github.com/backstage/backstage/compare/master...kurtaking:backstage:instrumentation-initializer#diff-3320c1fccf36da64786791589cd4f2099717b687b4b45483a35a61e328fd7fbc).

```ts
//...
metrics.createCounter('kurt.test.total').add(1);

// terminal output
{
  descriptor: {
    name: 'backstage.plugin.catalog.kurt.test.total',
    type: 'COUNTER',
    description: '',
    unit: '',
    valueType: 1,
    advice: {}
  },
  dataPointType: 3,
  dataPoints: [
    {
      attributes: {},
      startTime: [ 1752040219, 592000000 ],
      endTime: [ 1752040230, 327000000 ],
      value: 1
    }
  ]
}
```

[Replacing the existing meter](https://github.com/backstage/backstage/commit/2b12a9e17f2c7788d33ca765f0eb2dee3e37b417) in the `DefaultProcessingDatabase`, we can see the name being incorrectly created which is the expected behavior for now.

image

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,354 @@
...
+    enabled: boolean;
+    include?: string[];
+    exclude?: string[];
+  };
+}
+```
+
+```yaml
+backend:
+  metrics:
```
kurtaking: I am thinking `metrics` and eventually `tracing` should be grouped under some sort of `observability`, `telemetry`, or `instrumentation` property as they leverage a lot of the same configuration.

beps/0012-metrics-service/README.md
kurtaking: Example code has been removed.

kurtaking:
I have a POC available that handles the existing instructions for setting up telemetry, meaning it removes the need for the user to provide a separate `instrumentation.js` file with the `--require` flag. 

https://github.com/backstage/backstage/pull/30495

This also includes some refactoring to the `catalog` plugin to demonstrate the use of a `MetricsService`. Really looking forward to feedback from anyone interested in this project area. 

Highlights

* Initialize the `RootMetricsService` as early in the [backend initialization process ](https://github.com/backstage/backstage/pull/30495/files#diff-b094cc5f3d6cc1b159e6f99c9e4ee6c4ca440aabdcf15f672de5491aa89d5371R268)
* Supported [configuration](https://github.com/backstage/backstage/pull/30495/files#diff-72c07a14759a92a787fc935af625949c30ef630a228a01eb837f04ae6c1cb8f4) as of now
* Includes auto instrumentation library from OTEL, where we disable everything but the known libaries used to power the backstage framework. For example, we enable `express` for the `HttpRouterService` and `knex` for the `DatabaseService`
* [Utility wrappers](https://github.com/backstage/backstage/pull/30495/files#diff-c890dbfc075873e3c2755136e0094d20442318c81c88a4f799332f56689c221b) around OTEL instrumentations
* [Helper function](https://github.com/backstage/backstage/pull/30495/files#diff-d3718a210b2e1819359d3c25cd16da1b5b57e486932dcd215beb8e72a982de1b) to manage namespacing and prefixing for proper metric naming conventions - could easily be overwritten via extension point.

beps/0012-metrics-service/README.md
freben: Right - since this area doesn't exist yet, I think this row should be removed for now

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,351 @@
...
+- `{scope}` is the system scope (either **plugin** or **core**)
+- `{scope_name}` is the name of the plugin or core service (e.g., `catalog`, `scaffolder`, `database`)
+- `{metric_name}` is the hierarchical metric name as provided by the plugin author (e.g., `entities.processed.total`, `tasks.completed.total`)
+
+### Scope
+
+The `scope` represents where it belongs in the Backstage ecosystem.
+
+- `plugin` - A plugin-specific metric (e.g. `backstage.plugin.catalog.entities.count`)
+- `core` - A core service metric (e.g. `backstage.core.database.connections.active`)
```
freben: Suggestion: `service` instead of `core`. Seeing a `backstage.service.scheduler.` metric makes sense to me.

What do we do with plugin scoped services vs root scoped services? Will we standardize on ALL services being under `backstage.service.foo` but plugin-scoped ones have an automatically injected `pluginId` attribute? Or is it up to the plugin scoped services themselves to add such an attribute with that exact name as a standard?

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Naming Conventions
+
+All Backstage metrics follow this hierarchical pattern:
+
+`backstage.{scope}.{scope_name}.{metric_name}`
+
+**Where:**
+
+- `backstage` is the root namespace for all Backstage metrics
```
freben, line 74: This seems like an ok choice - does make the names longer of course, but I guess it's part of best practice? The existing metrics generally start with just the plugin ID directly, like `catalog.stitched.entities.count`

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,351 @@
...
+## Design Details
+
+### References
+
+- [General naming considerations](https://opentelemetry.io/docs/specs/semconv/general/naming/#general-naming-considerations)
+- [Acquiring a meter](https://opentelemetry.io/docs/languages/js/instrumentation/#acquiring-a-meter)
+
+### Integration with OpenTelemetry Auto-Instrumentation
+
+The `RootMetricsService` will automatically enable instrumentation for known libraries leveraged by the Backstage framework. Configuration will be provided to enable or disable auto-instrumentation via inclusion or exclusion lists.
```
freben: It would be desirable that the individual _implementations_ of those services could add those instrumentations themselves. Ie that the code for it would effectively live in `backend_defaults` and the metrics service implementation itself had no idea about the details of what they were. But I realize that it is a bit tricky since init often needs to happen very very early in the startup.

But the question will probably arise pretty quickly, how an adopter that has their own private root scoped service might instrument it through this

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,351 @@
...
+}
+```
+
+#### Root Metrics Service
+
+The `RootMetricsService` is responsible for initializing the OpenTelemetry SDK and creating plugin-scoped metrics services. If the end user wants to initialize their own SDK, they are responsible for initializing the OpenTelemetry SDK with their own configuration. The `RootMetricsService` is responsible for providing metrics to other root services and creating plugin-scoped metrics services.
+
+```ts
+interface RootMetricsService {
+  forPlugin(pluginId: string): MetricsService;
```
freben, line 173: Would it also have

```ts
forService(service: ServiceRef<unknown, 'root'>): MetricsService;
forService(service: ServiceRef<unknown, 'plugin'>, pluginId: string): MetricsService; 
```

or something like that?

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,351 @@
...
+```
+
+```ts
+class DefaultRootMetricsService implements RootMetricsService {
+  private sdk: NodeSDK;
+
+  static fromConfig(config: Config): RootMetricsService {
+    const metricsConfig = config.getOptionalConfig('backend.metrics');
+
+    const sdk = new NodeSDK({
```
freben: Was it really sufficient to have this here in "normal" code? I'd expect this code to be run too late in the startup sequence. The current recommended approach is to have a bootstrap file that's run via `--include` for this, to ensure that it runs even before any regular-code `import` could ruin any global state. If we don't do that, the sdk just doesn't work fully and different parts of the system end up with different competing meter instances. What's your take on that?

freben (commented):
Alright, super happy about this. Some questions

beps/0012-metrics-service/README.md
freben: One more thought - we leverage some custom instrumentations that require actual code, like tweaking spans through instrumentation-http. Gotta have a good way of getting that kind of customization injected too, whether there's config driven approaches or not on top of that

beps/0012-metrics-service/README.md
Rugvip: Thinking that's a sensible default, but might even be useful to make that part configurable?

kurtaking: Yes, the idea being we want to define namespace to prevent collisions with other metrics in telemetry backends. [[docs](https://opentelemetry.io/docs/specs/semconv/general/naming/#recommendations-for-application-developers)]

@Rugvip - after doing more research, I'm inclined to think we don't want this root namespace to be configurable. As if we wanted to submit `backstage` as a reserved namespace to the [registry](https://opentelemetry.io/docs/specs/semconv/registry/).

> The name may be generally applicable to applications in the industry. In that case consider submitting a proposal to this specification to add a new name to the semantic conventions, and if necessary also to add a new namespace.

beps/0012-metrics-service/README.md
kurtaking: The end goal is to be config driven which would require starting the SDK in some fashion similar to the example provided. If end users wanted different, they would provide their own override where they handle the SDK themselves. 

I say we push off trying to anwer this now since we are going to revise the BEP to exclude any changes to current guidance on starting the SDK via `--require` flag.

kurtaking:
The next revision to the BEP will exclude any configuration changes. We need more time to understand the best way to start the otel sdk during the backend initialization process. I believe we can still design a core metrics service to work based on the guidance [in the docs](https://backstage.io/docs/tutorials/setup-opentelemetry) while providing some sensible defaults and abstracting away some of the repetitive plumbing. 

Just wanted to post here as the entire configuration section will be replaced with an _out of scope_ bullet point.

beps/0012-metrics-service/README.md
kurtaking: **edit**: I no longer think we should use service names as part of the metric namespace. See my next response below on newer recommendation.

---

My original intention was to separate something occuring within the framework vs. a plugin. So maybe, we don't represent services under _core_, but we still keep it and the service name comes next within the hierarchy. 

```ts
backstage.plugin.database.connections.active
backstage.core.database.connections.active
```

or maybe leverage `framework` or `root`

```ts
backstage.plugin.database.connections.active
backstage.{framework|root}.database.connections.active
```

beps/0012-metrics-service/README.md
kurtaking: Marking resolved since configuration has been marked out of scope.

beps/0012-metrics-service/README.md
kurtaking: Marking resolved since configuration has been marked out of scope.

beps/0012-metrics-service/README.md
kurtaking: I went down the _services vs plugins_ path a little too much and think that it was the incorrect approach. Here is a table I put together of what the existing metrics would look like if we were to separate `framework vs plugin` scope. The end goal is the framework provides native instrumentation across all core services. So, end users get native telemetry there for free and only need to focus on business concerns within the confinements of their plugin. 

Using the scheduler service as an example, regardless of which plugin schedules something, they should all get some level of instrumentation provided by the framework.

image

kurtaking: FYI, I don't think that works either. It's better, but we need a `plugin` and/or `service` namespace in there. I have reached out to the OTEL community for feedback because naming is hard and there is probably someone out there that enjoys this more than I do.

beps/0012-metrics-service/README.md
kurtaking: This would be awesome. It'll be out of scope for this BEP, but should definitely be an enhancement captured somewhere.

beps/0012-metrics-service/README.md
kurtaking: @freben - do we want to put it on the calling service factory to call `rootMetrics.forService()` like the rootLogger?
image

 I was hoping to get that information from context, similar to injecting into a plugin.

beps/0012-metrics-service/README.md
kurtaking: I have started [documentation](https://github.com/backstage/backstage/pull/30845) around guidance on metric naming conventions. The PR includes moving the setup otel docs you've worked on, so you might be interested in the PR. 😄 

Guidance and examples for the services themselves will come with implementation PRs. 🎉

beps/0012-metrics-service/README.md
kurtaking: Let's try something different. I've created a [PR](https://github.com/backstage/backstage/pull/30845) that focuses on providing documentation around naming conventions for metrics. This will help focus full attention on this topic and it's something that can be published without ever implementing a metrics service.

angeliski:
Hey @kurtaking loved the proposes, the name and semantics make sense to me and can bring value in the short term 

And I believe we can have strong observability guides to help the community after shipping that

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Deprecation Plan
+
+1. Deprecation warning are added to all existing metrics
+2. New metrics will run in parallel with the deprecated ones for a period of time
+3. All existing metrics are removed from the codebase in the next major version
+
+## Dependencies
+
+1. The otel sdk MUST BE initialized as EARLY as possible to prevent dependents from receiving no-op meters - we will not change the current guidance on this
```
freben:
```suggestion
1. The `otel` SDK MUST BE initialized as EARLY as possible to prevent dependents from receiving no-op meters - we will not change the current guidance on this
```

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,248 @@
...
+backstage.plugin.auth.sessions.active.total # todo: technically a core service and a backend plugin
+```
+
+### Framework-Scoped Metrics
+
+Pattern: `backstage.framework.{service}.{metric_name}`
+
+```yaml
+# Examples
+backstage.framework.database.connections.active
```
benjdlambert: We've used `core` in most of the places before. So thinking that:

`backstage.core.scheduler.tasks` might be a better fit?

benjdlambert: Maybe even just `core.scheduler`?

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+// ...
+// metric is now available as `backstage.plugin.catalog.entities.processed.total`
+```
+
+## Release Plan
+
+1. Create the new metrics-related services.
+1. Create alpha-related documentation to add to existing core service [docs](https://backstage.io/docs/backend-system/core-services/index).
+1. Release the metrics service under `@alpha`.
```
benjdlambert, line 227: 👍

beps/0012-metrics-service/README.md
kurtaking: haha 😅 I started with `core`, but `service` was recommended and maybe my intent wasn't as clear then. So, I was going to come back to the table with `framework` because I worry service is too vague. Here is @freben's [comment](https://github.com/backstage/backstage/pull/30391#discussion_r2254023241).

> Maybe even just `core.scheduler`

I think it's important to start namespacing all metrics to start with `backstage` based on the semantic convention documentation I've read through. For example, here are the recommendations for naming [system metrics](https://opentelemetry.io/docs/specs/semconv/general/naming/#system-specific-naming) and [application metrics](https://opentelemetry.io/docs/specs/semconv/general/naming/#recommendations-for-application-developers).

With that, I'm now going as far as recommending [we don't make it configurable](https://github.com/backstage/backstage/pull/30391#discussion_r2255455602).

kurtaking: We settled on `core` in the sig today.

beps/0012-metrics-service/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Deprecation Plan
+
+1. Deprecation warning are added to all existing metrics
+2. New metrics will run in parallel with the deprecated ones for a period of time
+3. All existing metrics are removed from the codebase in the next major version
+
+## Dependencies
+
+1. The `otel` SDK MUST BE initialized as EARLY as possible to prevent dependents from receiving no-op meters - we will not change the current guidance on this
```
freben, line 244: My impression from the SIG is that maybe this can be relaxed a bit. But it's not directly related to the BEP, more just a happy improvement in general if we can down the line.

kurtaking:
Fantastic, thank you all for your help in getting this across the line 💯 ❤️

beps/0012-metrics-service/README.md
Rugvip: As far as I can tell the only reserved namespace is `otel.*`: https://opentelemetry.io/docs/specs/otel/semantic-conventions/#reserved-namespace

Just thinking from a UX PoV it may good to have it be configurable to improve discoverability in internal metric services. Not a strong opinion though, and it should be something that's easy to add in afterwards if needed.

Issue #29838, opened 2025-05-05
Scaffolder - Exclude params from output dump (Security issue and size for image)


veenarm:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
upload
audit
```

### 🗃️ Project Area

Software Templates

### 🔖 Need

Currently the Scaffolder work task dumps all the meta actions params to the audit log of info.

When using a param with format `data-url` to allow file uploads  this log dump outputs the entire file as output which is excessively large, and also potential a security risk as any file can be rebuilt from the log output.

Attempted to add `ui:field: Secret` however that completely changed the input field to a text input instead of the 'file upload' input.

### 📝 Proposal

We currently use 

```yaml
ui:backstage:
  review:
    show: false
```
Which precludes the data dump showing on the review page.

Options I see could be:
1. If review is false to also preclude that from showing in the audit log.
2. Add another `ui:backstage` such as `audit: show false`

### 🔄 Alternatives

Upgrading logging level is the only way we've found at the moment, but means we lose all the benefits of audit.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Yes, but I would like some more guidance

veenarm:
Can someone have a look at this? it's a bit of a security issue as well as dumping large files to logs/tables.

I'm happy to help with a PR - just need to know how you'd like it best approached.

veenarm:
Any thoughts @Rugvip  maybe this belongs in auditor section not scaffolding?

freben:
@schultzp2020 This may be of interest to you.

Rugvip:
Thinking that it's best to keep including the params if possible, without secrets still imo. To solve this issue I think the best approach would be to traverse through the parameter object and truncate any parameter that is longer than some maximum length before logging it, probably putting something like `...<truncated>` at the end instead.

veenarm:
@Rugvip  something like...

```ts
  truncateParameters(parameters: JsonObject) {
    const truncatedParameters: JsonObject = { ...parameters };
    Object.keys(truncatedParameters).forEach(key => {
      let value = truncatedParameters[key]?.toString();
      if (value && value.length > 5) {
        truncatedParameters[key] = value
[...]

PR #30396, opened 2025-06-26
fix: Scaffolder Auditor length issue with file uploads or very long task parameters

veenarm:

Added ability to limit the length on audit fields for scaffolder parameters.
This solves issues where using dataUrl fields for file uploads storing huge amounts of data (entire file as a dataurl).

Set a default of 256 length, but added config in allow a change/override this default behaviour.

```yaml
scaffolder:
  auditor:
    maxLength: 256
```

- [X] Added or updated documentation (Added in main config file)
- [] Tests for new functionality and regression tests for bug fixes
- [] Screenshots attached (for UI changes)

I ran api docs but it just produced the exact same file (no api changes).

There are no tests currently for any `parameter` checking, i can investigate further to add more if needed?

fixes https://github.com/backstage/backstage/issues/29838

Remade from #30354 and #30382

### DIFF
--- a/plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.test.ts
+++ b/plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.test.ts
@@ -448,4 +448,70 @@ describe('createParameterTruncator', () => {
       },
     });
   });
+
+  it('should not truncate if max length is -1', async () => {
+    const params = {
+      test: 'short',
+      test2: 'thisisaverylongstring',
+      nested: {
+        test3: 'anotherlongstringhere',
+        test4: ['ok', 'toolongstring', { prop: 'thisisaverylongstring' }],
+      },
+    };
+
+    const result = createParameterTruncator(
+      mockServices.rootConfig({
+        data: {
+          scaffolder: {
+            auditor: {
+              taskParameterMaxLength: -1,
+            },
+          },
+        },
+      }),
+    )(params);
+
+    expect(result).toEqual({
+      test: 'short',
+      test2: 'thisisaverylongstring',
+      nested: {
+        test3: 'anotherlongstringhere',
+        test4: ['ok', 'toolongstring', { prop: 'thisisaverylongstring' }],
+      },
+    });
+  });
+
+  it('should throw on invalid max length', async () => {
+    expect(() =>
+      createParameterTruncator(
+        mockServices.rootConfig({
+          data: {
+            scaffolder: {
+              auditor: {
+                taskParameterMaxLength: -2,
+              },
+            },
+          },
+        }),
+      ),
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"Invalid configuration for 'scaffolder.auditor.taskParameterMaxLength', got -2. Must be a positive integer or -1 to disable truncation."`,
+    );
+
+    expect(() =>
+      createParameterTruncator(
+        mockServices.rootConfig({
+          data: {
+            scaffolder: {
+              auditor: {
+                taskParameterMaxLength: 1.5,
+              },
+            },
+          },
+        }),
+      ),
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"Invalid configuration for 'scaffolder.auditor.taskParameterMaxLength', got 1.5. Must be a positive integer or -1 to disable truncation."`,
+    );
+  });
 });


veenarm:
Re did it a third time with out committing and now its generating the reports properly 👍

veenarm:
@schultzp2020 - Our area is about to roll out a major file upload to bucket system with scaffolding and this fix is quite important else we'll be getting major dumps in logs and tables. If you'd prefer, we can disable it with `-1` by default, and only if a value entered then it sets its value?

veenarm:
@Rugvip hey, are you able to have a look as per your comments on the issue created?

schultzp2020:
@veenarm - Sorry, I was out on PTO. 

Do you think we'll need to make changes like this in other areas, like the catalog system? If so, I'm wondering if a different key placement is better. Thoughts? 🤔

veenarm:
I guess it depends on what the audit system is doing in catalog or other core areas. In general I don't believe it's logging any user input anywhere else, which could be significant in size (like dataUrls) or large user input systems such as in the scaffolder.

This is specific to the Scaffolder Parameter entries, I have no examples to work off for catalog/auditing so I'd have to defer that to your team for info.

veenarm:
@schultzp2020  - As per your comment earlier, I've added a negative 1 check for disabling the truncating.

Happy to change the key to what you think is more appropriate, I was also tossing up `auditor.scaffolder.maxLength` .

plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
```diff
@@ -182,14 +190,44 @@ export class TaskWorker {
     });
   }
 
+  protected truncateParameters(parameters: JsonObject) {
```
Rugvip: Let's make this `private`

plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
```diff
@@ -182,14 +190,44 @@ export class TaskWorker {
...
+    if (auditMaxLength === -1) {
+      this.logger?.debug(
+        `scaffolder.auditor.maxLength manually disabled via configuration, no task parameter length limit set.`,
+      );
+      return parameters;
+    }
+
+    const truncatedParameters: JsonObject = {};
+
+    for (const key in parameters) {
```
Rugvip: Afaik `parameters` can have nested objects, so this would need to be recursive

plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
```diff
@@ -182,14 +190,44 @@ export class TaskWorker {
...
+      if (Object.prototype.hasOwnProperty.call(parameters, key)) {
```
Rugvip: `Object.hasOwn` is the nicer one to use nowadays

plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
```diff
@@ -182,14 +190,44 @@ export class TaskWorker {
     });
   }
 
+  protected truncateParameters(parameters: JsonObject) {
+    const auditMaxLength =
+      this.config?.getOptionalNumber('scaffolder.auditor.maxLength') ?? 256;
```
Rugvip: I'm thinking we might be able to get away with skipping config for this one. We could perhaps set it to something fairly large, like somewhere around 4kB-128kB? That way uploaded text files won't get truncated, but images generally would.

If we do think it's best to keep the config, I think we need to update the name a bit, perhaps something like `taskParameterMaxLength`. You'd also need to add config schema.

Rugvip (commented):
Thanks! 👍 

Some thoughts to help move this forward

plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
veenarm: Done, however I've added a test, so the test has to comment ignore as its private.

plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
veenarm: done

plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
veenarm: I'm not sure how you'd do that, a 128kb file is like 130,000 characters, personally I don't think you should be able to rebuild any uploaded files/documents. Or maybe an extra way needed to make them secret as well however was advised to go down this path instead. I've changed the parameter name to match.

plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts
veenarm: Yup, have implemented a little more complexed function to loop the parameters, also added some test cases.

veenarm:
@Rugvip  - See updates, also patched/added the config.d.ts changes.

veenarm:
@Rugvip  - happy to do another vet / review, this is causing us a few issues at the moment, we use scaffolder a lot for file uploading to initiate workflow transfers, any eta?

veenarm:
@Rugvip / @freben  - ping to keep alive, this is quite huge for us at the moment as we're planning a roll out of file uploads to S3 storage systems.

We've found other issues but they're separate and can be raised as such (e.g. everything uploaded is loaded into memory).

PR #30406, opened 2025-06-27
Fix duplicate route warning for TechDocs Addons documentation in microsite

AmbrishRamachandiran:
image

This PR resolves a Docusaurus routing warning caused by two Markdown files (`addons.md` and `addons--new.md`) using the same `id: addons` in their frontmatter. This resulted in both files attempting to generate the same route (`/docs/addons`), leading to non-deterministic routing behavior.

## Changes

- Updated the `id` in `addon-new.md` to `addons--new`

### DIFF
--- a/microsite/sidebars.ts
+++ b/microsite/sidebars.ts
@@ -238,7 +238,7 @@ export default {
           'features/techdocs/techdocs-overview',
           'features/techdocs/getting-started',
           'features/techdocs/concepts',
-          'features/techdocs/addons',
+          'features/techdocs/addons--new',
           'features/techdocs/architecture',
           'features/techdocs/extensions',
           'features/techdocs/creating-and-publishing',
--- a/docs/features/techdocs/addons--new.md
+++ b/docs/features/techdocs/addons--new.md
@@ -1,5 +1,5 @@
 ---
-id: addons
+id: addons--new
 title: TechDocs Addons
 description: How to find, use, or create TechDocs Addons.
 ---


AmbrishRamachandiran:
@freben tagging you for review as no reviewer's assigned

microsite/sidebars.ts
```diff
@@ -238,7 +238,7 @@ export default {
           'features/techdocs/techdocs-overview',
           'features/techdocs/getting-started',
           'features/techdocs/concepts',
-          'features/techdocs/addons',
```
Parsifal-M, line 241: Can you spin up the doc site to test this? won't this change the TechDocs Addons sidebar entry to TechDocs Addons--new?

Wondering what it does to this section:

image

Parsifal-M:
Ah, this is because of the new frontend system docs I think, I am not sure if we have a way we want to sort these out just yet... @backstage/documentation-maintainers we don't have a plan for this yet AFAIK?

AmbrishRamachandiran:
Any updates?

Parsifal-M:
Not yet, the sig is planned for the `18th July`, lots of people are on holiday breaks 😃 and I think this is quite low priority right now.

Thanks for your patience!

AmbrishRamachandiran:
> Not yet, the sig is planned for the `18th July`, lots of people are on holiday breaks 😃 and I think this is quite low priority right now.
> 
> Thanks for your patience!

any updates?

Parsifal-M:
Heya,

I think the idea we came up with is that we mark the old one as --old and then keep the new one without anything and make a reference to the new page in the old page.

Am I more or less correct with that @awanlin ? 😅

PR #30410, opened 2025-06-29
Canon - New Header component ✨

cdedreuille:
CleanShot 2025-06-30 at 22 49 18@2x


.changeset/early-trains-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/canon': patch
+---
+
+Add new Toolbar component to Canon.
```
Rugvip: Needs and update right?

packages/canon/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { HeaderToolbar } from './HeaderToolbar';
+import { HeaderTabs } from './HeaderTabs';
+import { HeaderSubNav } from './HeaderSubNav';
+
+/**
+ * A component that renders a toolbar.
+ *
+ * @public
+ */
+export const Header = (props: HeaderProps) => {
```
Rugvip, line 26: Having a bit of a hard time wrapping my head around what the scope of this component is. It seems to go a bit beyond just the header and more towards implementing the full page with navigation in a fairly opinionated way. I have a bit of a feeling we may want to split things up a little bit. Especially curious about the purpose of the SubNav

cdedreuille: The `Header` is only the part set in the middle and doesn't include the side navs as shown on the image. I've placed it as a decorator in Storybook just to make sure it work in context.

The reason why we wanted to make this as the same component is that updating the header can't be done in one go. We have to place it in every single plugin. From a design perspective we might evolve this design and play with different layout, specially with the relation between the title and breadcrumb. If we have it all under a single component, then everyone will get the update straight away.

Maybe Header is not the right name. Any other suggestion?

.changeset/early-trains-relax.md
cdedreuille: Oops. Thank you 🙏

cdedreuille: Fixed

packages/canon/src/components/Header/Header.tsx
Rugvip: Tricky bit is that often the page header and the page content are not part of the same extension or even plugin, so it's less about naming and more about that the component may be crossing what is often a plugin boundary. I'm thinking at least the subnav might be good to leave as a separate component.

Tabs are probably fine, they tend to be owned by the page but sometimes populated by other plugins. The main bit there is to make sure we can tie them into react-router nicely

cdedreuille: After our discussion I separated the `HeaderPage` to be its own component. This will help to separate content from global header and sub navs.

packages/canon/report.api.md
```diff
@@ -916,6 +937,109 @@ export interface GridProps extends SpaceProps {
...
+  // (undocumented)
+  setTabRef: (key: string, element: HTMLDivElement | null) => void;
+  // (undocumented)
+  tab: HeaderTab;
+}
+
+// @public
+export const HeaderTabsIndicators: (
+  props: HeaderIndicatorsProps,
+) => JSX_2.Element;
```
Rugvip: Is this indented to be public?

packages/canon/report.api.md
```diff
@@ -916,6 +937,109 @@ export interface GridProps extends SpaceProps {
...
+
+// @public
+export interface HeaderTabProps {
+  // (undocumented)
+  setHoveredKey: (key: string | null) => void;
+  // (undocumented)
+  setTabRef: (key: string, element: HTMLDivElement | null) => void;
+  // (undocumented)
+  tab: HeaderTab;
+}
```
Rugvip: Shouldn't be exported right?

packages/canon/report.api.md
```diff
@@ -916,6 +937,109 @@ export interface GridProps extends SpaceProps {
...
+export interface HeaderIndicatorsProps {
+  // (undocumented)
+  hoveredKey: string | null;
+  // (undocumented)
+  prevHoveredKey: MutableRefObject<string | null>;
+  // (undocumented)
+  tabRefs: MutableRefObject<Map<string, HTMLDivElement>>;
+  // (undocumented)
+  tabsRef: MutableRefObject<HTMLDivElement | null>;
+}
```
Rugvip: If it's the case that the indicators shouldn't be public, these should be removed too

packages/canon/report.api.md
```diff
@@ -916,6 +937,109 @@ export interface GridProps extends SpaceProps {
...
+
+// @public
+export interface HeaderPageOption {
+  // (undocumented)
+  label: string;
+  // (undocumented)
+  onClick?: () => void;
+  // (undocumented)
+  value: string;
+}
```
Rugvip, line 971: Guess it could be worth considering having the `HeaderPage` reuse the `HeaderOption` and `HeaderTab` types, but no strong opinion on that really

packages/canon/report.api.md
```diff
@@ -916,6 +937,109 @@ export interface GridProps extends SpaceProps {
...
+
+// @public
+export interface HeaderProps {
+  // (undocumented)
+  breadcrumbs?: HeaderBreadcrumb[];
+  // (undocumented)
+  icon?: React.ReactNode;
+  // (undocumented)
+  name?: string;
```
Rugvip: `name` actually seems a bit strange here, I assume you're thinking about it from the PoV of the "page name"? Wonder if just `title` or `text` is better

packages/canon/report.api.md
```diff
@@ -916,6 +937,109 @@ export interface GridProps extends SpaceProps {
...
+  // (undocumented)
+  value: string;
+}
+
+// @public
+export interface HeaderPageProps {
+  // (undocumented)
+  description?: string;
+  // (undocumented)
+  name?: string;
```
Rugvip: Same as the `HeaderProps` prop

packages/canon/report.api.md
```diff
@@ -916,6 +937,109 @@ export interface GridProps extends SpaceProps {
...
+  // (undocumented)
+  onClick?: () => void;
+  // (undocumented)
+  value: string;
+}
+
+// @public
+export interface HeaderPageProps {
+  // (undocumented)
+  description?: string;
```
Rugvip, line 976: Could also be something like `secondaryText` or `subTitle` right? I haven't looked at what we use in other places with double text without obvious context though, so whatever is most consistent

packages/canon/report.api.md
cdedreuille: Not necessary but if I don't make it public, the report will fail as I'm using these props in a sub component. What would be the best approach then?

packages/canon/report.api.md
cdedreuille: We actually use `description` in other places like `TextField`

packages/canon/report.api.md
Rugvip: I don't see it used anywhere else? should be fine to simply not export afaik

Rugvip (approved):

PR #30411, opened 2025-06-30
Adding Additional options to Microsoft-Provider auth-plugin

nicholas-rees:
## Hey, I just made a Pull Request to Use Backstage with Government Cloud!

We are using the government versions of Microsoft Azure, and we need to pass in the government urls (login.microsoftonline.us and graph.microsoft.us). We have patched backstage on our end, but we figure it would be nice to get this to everyone who uses Backstage (and also simplify our build process a bit). Just so you know, when I say we patched backstage on our end, I mean that I hacked the javascript modules with sed scripts in the Dockerfile. So this isn't exactly what we have deployed--it should be loads nicer.

Really all this does is make it so that we can use the microsoft-provider backend module.  I also configured (I think) the configSchema to reflect these additional options.

I did not add any new testing for this, I'm hoping the existing test works, but I'm having tests that are failing on the master branch with node v20.19.3.

***update*** I'm checking that we are testing for "new functionality." The justification is that these variables are optional, and should not change anything, so the microsoft-provider test should have exactly the same result as before. If you would prefer a test of some sort, then please request it.

### DIFF
--- a/plugins/auth-backend-module-microsoft-provider/src/strategy.ts
+++ b/plugins/auth-backend-module-microsoft-provider/src/strategy.ts
@@ -23,14 +23,14 @@ import {
 import { VerifyFunction } from 'passport-oauth2';
 
 export class ExtendedMicrosoftStrategy extends MicrosoftStrategy {
-  private _apiEntryPoint: string;
-  private _graphApiVersion: string;
+  private apiEntryPoint: string;
+  private graphApiVersion: string;
   private shouldSkipUserProfile = false;
 
   constructor(config: MicrosoftStrategyOptions, verify: VerifyFunction) {
     super(config, verify);
-    this._apiEntryPoint = config.apiEntryPoint || 'https://graph.microsoft.com';
-    this._graphApiVersion = config.graphApiVersion || 'v1.0';
+    this.apiEntryPoint = config.apiEntryPoint || 'https://graph.microsoft.com';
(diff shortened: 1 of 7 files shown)


plugins/auth-backend-module-microsoft-provider/src/strategy.ts
```diff
@@ -16,11 +16,23 @@
...
 import { decodeJwt } from 'jose';
-import { Strategy as MicrosoftStrategy } from 'passport-microsoft';
+import {
+  Strategy as MicrosoftStrategy,
+  MicrosoftStrategyOptions,
+} from 'passport-microsoft';
+import { VerifyFunction } from 'passport-oauth2';
 
 export class ExtendedMicrosoftStrategy extends MicrosoftStrategy {
+  private _apiEntryPoint: string;
```
schultzp2020:
```suggestion
  private apiEntryPoint: string;
```

Can you remove the underscore prefix on private property names? 🙏

nicholas-rees: No problem, should be good. Thank you for the review. If there's anything else, I'll pay closer attention now that I see this is being looked at.
[See updates to property names](https://github.com/backstage/backstage/pull/30411/commits/48e8fe47e5c594d574781cf247688fad9e55e7c0)

nicholas-rees:
I would like to continue with this MR, but I have not heard about it. I am going to create a related issue to try to get more attention.

Issue #30402, opened 2025-06-26
🐛 Bug Report: a11y - Tabbing through clickable catalog nodes doesn't work


logonoff:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
Keyboard, DefaultRenderNode
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

N/A

### 📝 Description & Context

The nodes with onClick passed don't have the `tabindex={0}` prop set, so those nodes cannot be focused on when using a keyboard

### 👍 Expected Behavior

I expect the clickable nodes in the catalog relation graph to be keyboard focusable

### 🥾 Reproduction steps

1. Go to https://demo.backstage.io/catalog/default/component/playback-order
2. Tab around the catalog graph
3. Notice that the nodes cannot be tabbed on

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Undecided

MihaiTabara:
Found in today's team triaging mtg. 
@logonoff thanks for filing this! are you willing to make a contribution here and if so, how might we help?

logonoff:
> Found in today's team triaging mtg. [@logonoff](https://github.com/logonoff) thanks for filing this! are you willing to make a contribution here and if so, how might we help?

I figured it out, thanks for the traige!

PR #30417, opened 2025-06-30
fix(catalog): add keyboard support to graph nodes

logonoff:

fixes https://github.com/backstage/backstage/issues/30402

note: due to the order that dagre renders the graph, the tab order is a bit weird.. not sure how to address this, but I think that enabling tabbing on the nodes is a good start 

https://github.com/user-attachments/assets/a6569797-5efa-4c8a-a51a-46894fd91726

### DIFF
--- a/.changeset/early-wolves-brake.md
+++ b/.changeset/early-wolves-brake.md
@@ -1,0 +1,6 @@
+---
+'@backstage/plugin-catalog-graph': patch
+'@backstage/plugin-catalog-react': patch
+---
+
+Enable keyboard navigation for catalog nodes. Fixed a typo in the pinch and zoom description.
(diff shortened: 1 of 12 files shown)


logonoff:
bump

Rugvip (commented):
Sorry gotta question the value a little bit on this one as it's introducing some complexity. I'd say keyboard navigation for the graph for non-impaired users might be a tiny improvement but probably not worth the hassle. For impaired users I don't think we're anywhere close to providing a good experience for the graph right? Afaik there's no context on the layout of the nodes, their descriptions, the edges or description and directions of those. You just end up with a fairly random collection of entities in some arbitrary order, especially for anything but very slim filters. Basically it's much better to use relations cards on the entity page because that's better anchored in a specific node with context for the edges.

Issue #30399, opened 2025-06-26
🐛 Bug Report: a11y - `EntityContextMenu` with no `extraContextMenuItems` has poor keyboard navigation


logonoff:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
EntityContextMenu, EntityHeader, EntityLayout
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

N/A

### 📝 Description & Context

When the `UNSTABLE_extraContextMenuItems` prop is not provided, the first menu item does not receive keyboard focus when the menu is open. When `UNSTABLE_extraContextMenuItems` is an empty array, a divider shows up when there are no items. 

### 👍 Expected Behavior

The divider is hidden when `UNSTABLE_extraContextMenuItems` is an empty array, and the "unregister entity" menu item is focused when the menu is open when no `extraContextMenuItems` are passed

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes, and I have enough information to get started

MihaiTabara:
Found in today's triaging meeting.
@logonoff thanks for filing this issue! 

Are you open to contributions to the project and if so, how might we help you bootstrapping in it?  I see you marked the last time as having enough information. Want me to assign this to you? Thank you!

MihaiTabara:
Ongoing PR reviews, no stale.

PR #30418, opened 2025-06-30
fix(catalog): autofocus first `EntityContextMenu` item

logonoff:

fixes https://github.com/backstage/backstage/pull/30399

https://github.com/user-attachments/assets/eac69d6a-493a-4079-b025-fb0e998365f5

### DIFF
--- a/plugins/catalog/src/components/EntityContextMenu/EntityContextMenu.tsx
+++ b/plugins/catalog/src/components/EntityContextMenu/EntityContextMenu.tsx
@@ -103,22 +103,58 @@ export function EntityContextMenu(props: EntityContextMenuProps) {
     }
   }, [copyState, alertApi, t]);
 
-  const extraItems = UNSTABLE_extraContextMenuItems && [
-    ...UNSTABLE_extraContextMenuItems.map(item => (
-      <MenuItem
-        key={item.title}
-        onClick={() => {
-          onClose();
-          item.onClick();
-        }}
-      >
-        <ListItemIcon>
-          <item.Icon fontSize="small" />
-        </ListItemIcon>
-        <ListItemText primary={item.title} />
-      </MenuItem>
-    )),
-    <Divider key="the divider is here!" />,
+  const extraItems = UNSTABLE_extraContextMenuItems?.length
+    ? [
+        ...UNSTABLE_extraContextMenuItems.map(item => (
+          <MenuItem
+            key={item.title}
+            onClick={() => {
+              onClose();
+              item.onClick();
+            }}
+          >
+            <ListItemIcon>
+              <item.Icon fontSize="small" />
+            </ListItemIcon>
+            <ListItemText primary={item.title} />
+          </MenuItem>
+        )),
+        <Divider key="the divider is here!" />,
+      ]
+    : null;
+
+  const defaultMenuItems = [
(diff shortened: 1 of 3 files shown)


plugins/catalog/src/components/EntityContextMenu/EntityContextMenu.tsx
```diff
@@ -103,22 +103,55 @@ export function EntityContextMenu(props: EntityContextMenuProps) {
...
+              <item.Icon fontSize="small" />
+            </ListItemIcon>
+            <ListItemText primary={item.title} />
+          </MenuItem>
+        )),
+        <Divider key="the divider is here!" />,
+      ]
+    : undefined;
+
+  const defaultMenuItems = [
```
Rugvip, line 126: Need some `key`s for these now right? I'm assuming there's a new console error popping up now

logonoff: Good catch! I've added some `key`s to each item in this array

logonoff:
bump

plugins/catalog/src/components/EntityContextMenu/UnregisterEntity.tsx
```diff
@@ -34,7 +35,10 @@ interface UnregisterEntityProps {
   onClose: () => void;
 }
 
-export function UnregisterEntity(props: UnregisterEntityProps) {
+export const UnregisterEntity = forwardRef<
```
freben, line 39: I'll note that forwardRef [seems to be on its way out entirely](https://react.dev/reference/react/forwardRef) - if you would be able to address that (possibly in a followup), it'd be appreciated.

plugins/catalog/src/components/EntityContextMenu/UnregisterEntity.tsx
```diff
@@ -72,5 +76,5 @@ export function UnregisterEntity(props: UnregisterEntityProps) {
     );
   }
 
-  return <>{unregisterButton}</>;
-}
+  return undefined;
```
freben: As long as we support react 17, we should never return undefined from components, only `null`. This changed in 18 afaik but we can't quite rely on that yet

plugins/catalog/src/components/EntityContextMenu/EntityContextMenu.tsx
```diff
@@ -103,22 +103,58 @@ export function EntityContextMenu(props: EntityContextMenuProps) {
...
+          >
+            <ListItemIcon>
+              <item.Icon fontSize="small" />
+            </ListItemIcon>
+            <ListItemText primary={item.title} />
+          </MenuItem>
+        )),
+        <Divider key="the divider is here!" />,
+      ]
+    : undefined;
```
freben: hm i think this needs to be null too

freben (commented):
Hey sorry - a few small notes from me

plugins/catalog/src/components/EntityContextMenu/UnregisterEntity.tsx
logonoff: I can add a todo but unfortunately for the menu keyboard navigation to work we need the forwardRef

freben (approved):

PR #30421, opened 2025-07-01
Fix (techdocs-node): Fix `parseReferenceAnnotation` missing annotation error message to include annotation name

mhaley37:

The `parseReferenceAnnotation` accepts an annotation name and an error will be thrown IF that annotation is not found. However the error message _does not_ include the annotation name argument, rather it is simply:
```
No location annotation provided in entity
```

Having the error message reflect the annotation argument will give a lot more context when these are found ( which is most likely to happen while going through logs debugging an issue)

When this error message was found in the debug log, it was confusing as it looked to indicate an annotation named `location` was missing, however it was actually the `backstage.io/techdocs-ref` annotation.

### DIFF
--- a/.changeset/whole-buckets-fix.md
+++ b/.changeset/whole-buckets-fix.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-techdocs-node': patch
+---
+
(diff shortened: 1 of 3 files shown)


benjdlambert (approved):
Thanks :tada: makes sense!

PR #30426, opened 2025-07-01
`auth-backend`: Internal refactor of the existing OIDC router

benjdlambert:
Just a bit of :broom: here to cleanup the existing OIDC router.


Rugvip (approved):
:shipit:

plugins/auth-backend/src/database/UserInfoDatabase.test.ts
```diff
@@ -38,15 +39,19 @@ describe('UserInfoDatabaseHandler', () => {
 
     return {
       knex,
-      dbHandler: new UserInfoDatabaseHandler(knex),
+      dbHandler: await UserInfoDatabase.create({
```
fearphage, line 42: Should this naming be updated? It was a `dbHandler`, now it's a `database`, right? At least that's how the code reads.

plugins/auth-backend/src/service/OidcRouter.ts
```diff
@@ -0,0 +1,73 @@
...
+    router.get('/v1/token', (_req, res) => {
+      res.status(501).send('Not Implemented');
+    });
+
+    // This endpoint doesn't use the regular HttpAuthoidc, since the contract
+    // is specifically for the header to be communicated in the Authorization
+    // header, regardless of token type
+    router.get('/v1/userinfo', async (req, res) => {
+      const matches = req.headers.authorization?.match(/^Bearer[ ]+(\S+)$/i);
+      const token = matches?.[1];
```
fearphage, line 56: You could name the group instead of using numbered groups.


```suggestion
      const matches = req.headers.authorization?.match(/^Bearer[ ]+(?<token>\S+)$/i);
      const token = matches?.groups.token
```

PR #30453, opened 2025-07-04
Canon - Add tertiary variant ton Button, ButtonIcon and ButtonLink

cdedreuille:


packages/canon/src/components/Button/styles.css
```diff
@@ -77,12 +77,37 @@
...
-  &:disabled {
+  &[data-disabled='true'] {
     box-shadow: inset 0 0 0 1px var(--canon-border-disabled);
     color: var(--canon-fg-disabled);
   }
 }
 
+.canon-Button[data-variant='tertiary'] {
+  background-color: transparent;
+  color: var(--canon-fg-primary);
```
Rugvip, line 88: This seems a little bit strange, what is really the purpose of the tertiary variant?

Rugvip: Is this perhaps a completely different dedicated variant, like `text`?

cdedreuille: Our tertiary button has the same padding but no borders. We do have a hover state for it. It is quite a common pattern where you do not want to have these borders like in the toolbar.

CleanShot 2025-07-04 at 16 02 59@2x

PR #30456, opened 2025-07-05
feat(scaffolder): adjustable panels width in template editor layouts

stephenglass:

The template editor and template form playground layouts are difficult to use on small, medium, and even some large sized screens. In the template editor layout, the preview panel takes only 33% of the content width. And in the form playground layout, it's a 50% split. There are times when creating or editing templates where you may want to focus on the editor panel or the preview panel and this PR helps make this easier when you are on a small/medium/large screen.

This PR allows for the width of the editor and preview panels to be adjustable by dragging a bar in between the two panels. See video for example. The layout maintains a responsive design in very small screen sizes where the layout goes from columns to row orientation.

Template Editor:

https://github.com/user-attachments/assets/834aeca9-6f06-4607-8da5-4bb9a77fc19a

Template Form Playground:

https://github.com/user-attachments/assets/9f63abb8-dfab-4271-85dd-1726f9a63f01

### DIFF
--- a/.changeset/petite-spoons-flash.md
+++ b/.changeset/petite-spoons-flash.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder': minor
+---
+
+Add resizable panels width for the editor and preview panels in the template editor and template form playground layouts. Users can now resize these panels by dragging the divider between the two areas.
--- a/.github/vale/config/vocabularies/Backstage/accept.txt
+++ b/.github/vale/config/vocabularies/Backstage/accept.txt
@@ -543,3 +543,4 @@ zod
 Zolotusky
 zoomable
 zsh
+resizable
--- a/plugins/scaffolder/package.json
+++ b/plugins/scaffolder/package.json
@@ -97,6 +97,7 @@
     "luxon": "^3.0.0",
     "qs": "^6.9.4",
     "react-resizable": "^3.0.5",
+    "react-resizable-panels": "^3.0.4",
     "react-use": "^17.2.4",
     "react-window": "^1.8.10",
     "yaml": "^2.0.0",
--- a/plugins/scaffolder/src/alpha/components/TemplateEditorPage/TemplateEditor.tsx
+++ b/plugins/scaffolder/src/alpha/components/TemplateEditorPage/TemplateEditor.tsx
@@ -34,6 +34,7 @@ import {
   TemplateEditorLayoutFiles,
   TemplateEditorLayoutPreview,
   TemplateEditorLayoutConsole,
+  TemplateEditorPanels,
 } from './TemplateEditorLayout';
 import { TemplateEditorToolbar } from './TemplateEditorToolbar';
 import { TemplateEditorToolbarFileMenu } from './TemplateEditorToolbarFileMenu';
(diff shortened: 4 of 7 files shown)


plugins/scaffolder/src/alpha/components/TemplateEditorPage/useAdjustablePanelWidth.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { useRef, useState, useCallback } from 'react';
+
+export function useAdjustablePanelWidth(options: {
```
schultzp2020: Should we move this hook into a common lib since it may be useful for other React components?

.changeset/petite-spoons-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder': patch
```
schultzp2020: Should we consider this a fix or a new feature?

plugins/scaffolder/src/alpha/components/TemplateEditorPage/useAdjustablePanelWidth.ts
stephenglass: Is there a common lib that would make sense to move this to? The only I can think of would be `scaffolder-react` but lmk if that makes sense since the `scaffolder` plugin also has some react hooks added

.changeset/petite-spoons-flash.md
stephenglass: I personally think it's a feature since the current UX is not broken .. just difficult to use which this PR improves. Either way wouldn't it be considered a `patch` in the changeset given that it's a non-breaking/API change?

stephenglass: @schultzp2020 I see what you mean now, I updated the changeset to `minor` since the package version is > `1.0` and it's a new feature.

plugins/scaffolder/src/alpha/components/TemplateEditorPage/useAdjustablePanelWidth.ts
benjdlambert: Thinking that this is fine to keep private for now, let's leave it in the `scaffolder` plugin until it's requested to used elsewhere. Not feeling we should add something to the public API straight away yet :pray:

benjdlambert:
@stephenglass thanks for raising this PR, my first initial thoughts are is there some way we can remove some of the complexities of having this code in the repo, and maybe leaning on using a library instead to deal with some of these resizable panels. Could something like https://github.com/bvaughn/react-resizable-panels work for this use case?

Just feels a little unfortunate to bring in a chunk of complexity where we might be able to grab something off the shelf that works pretty nicely for our use case whilst keeping things relatively simple?

Happy for pushback here too, but just trying to balance the two concerns here of a nice feature and improvement over code ownership :)

stephenglass:
> @stephenglass thanks for raising this PR, my first initial thoughts are is there some way we can remove some of the complexities of having this code in the repo, and maybe leaning on using a library instead to deal with some of these resizable panels. Could something like https://github.com/bvaughn/react-resizable-panels work for this use case?
> 
> Just feels a little unfortunate to bring in a chunk of complexity where we might be able to grab something off the shelf that works pretty nicely for our use case whilst keeping things relatively simple?
> 
> Happy for pushback here too, but just trying to balance the two concerns here of a nice feature and improvement over code ownership :)

Absolutely @benjdlambert, I updated the PR to use the react-resizable-panels library to do the heavy lifting. The code changes are a lot less complex now and having less changes needed overall. Let me know what you think, cheers!

plugins/scaffolder/src/alpha/components/TemplateEditorPage/useAdjustablePanelWidth.ts
stephenglass: Marking this resolved since we opted to use a library instead of a custom hook

.changeset/petite-spoons-flash.md
stephenglass, line 1: "resizable" is a real word 😁

stephenglass:
bumping to not get stale

benjdlambert (approved):
Nice! thanks for this! let's try it out! :tada:

PR #30463, opened 2025-07-07
`auth-backend`: Move out the `UserInfoDatabase` updates away from the `TokenIssuer`

benjdlambert:
:broom:

### DIFF
--- a/plugins/auth-backend/migrations/20250707164600_user_created_at.js
+++ b/plugins/auth-backend/migrations/20250707164600_user_created_at.js
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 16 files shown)


plugins/auth-backend/migrations/20250707164600_user_created_at.js
```diff
@@ -0,0 +1,55 @@
...
+exports.up = async function up(knex) {
+  await knex.schema.alterTable('user_info', table => {
+    table.renameColumn('exp', 'updated_at');
+  });
+
+  // Sqlite doesn't support adding a column with non-constant default when table has data
+  // so we need to add it as nullable first, then set the value to the updated_at value
+  await knex.schema.alterTable('user_info', table => {
+    table.timestamp('created_at').nullable();
+  });
```
Rugvip: As discussed: moving this back to the simpler form and skipping the migration test for SQLite that tests it with data

plugins/auth-backend/src/lib/resolvers/CatalogAuthResolverContext.ts
```diff
@@ -73,11 +76,27 @@ export class CatalogAuthResolverContext implements AuthResolverContext {
...
+      sub,
+      ent,
+      ...additionalClaims,
+    };
+
+    // Store the user info in the database upon successful token
+    // issuance so that it can be retrieved later by limited user tokens
+    await this.userInfo.addUserInfo({
+      claims,
+    });
```
Rugvip, line 99: Hmm, realized it's best to do this after we issue a token. That way if the token creation fails we won't add anything to the DB. Or we could just do them in parallel too tbh for a bit of extra speed.

PR #30471, opened 2025-07-08
Fix duplicate slash in `renderInTestApp` routing

jabrks:

The `renderInTestApp` util currently registers the element you pass in with a route path of `/`
https://github.com/backstage/backstage/blob/f78a03b57baf0ff50f95a458a556cc4036065b3f/packages/frontend-test-utils/src/app/renderInTestApp.tsx#L157

This is then passed into the AppRoutes extension as an input which also appends a `/*` to the end
https://github.com/backstage/backstage/blob/2b939c8ffac2c675621215a29901368be0299b67/plugins/app/src/extensions/AppRoutes.tsx#L45

This means that the route mounted in the test is `//*` rather than `/*` as you might expect. This causes problems with route matching, as React Router removes duplicate slashes from patterns when determining matches meaning that most paths can never be matched.

In order to resolve this, I think the correct course of action would be to set an empty route path in the test util and leave the existing behaviour in the AppRoutes extension as it is, resulting in the test element being served on `/*` as expected. That is what I have implemented in this PR.

I've also written a test that fails with the old behaviour to try and prevent a regression in future.

### DIFF
--- a/.changeset/yellow-spies-rescue.md
+++ b/.changeset/yellow-spies-rescue.md
@@ -2,4 +2,4 @@
 '@backstage/frontend-test-utils': patch
 ---
 
-Update route path for index route in `renderInTestApp` util to fix duplicate slash and ensure any nested routing behaves correctly.
+Update route path for index route in `renderInTestApp` to fix duplicate slash and ensure any nested routing behaves correctly.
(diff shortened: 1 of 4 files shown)


jabrks:
More than happy to go about solving this differently, I just opted for what I thought would be easiest here. I can take a look at if there's a way to go about addressing this in `AppRoutes` instead.

jabrks:
That's now done, although I'm not sure if there's now a better place for the test to live

Issue #30449, opened 2025-07-04
Allow configuring HTTP server timeouts declaratively via backend.http in app-config


BethGriggs:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
Customise HTTP Server Options, keepAliveTimeout
```

### 🗃️ Project Area

Core Framework

### 🔖 Need

Similar to part of what was proposed in https://github.com/backstage/backstage/issues/21808.  

Although we can currently customise low-level HTTP server configuration through the configure hook in `rootHttpRouterServiceFactory`, this requires access to backend application code.

In practice, we've observed HTTP errors due to mismatches between the load balancer's timeout and some of the default HTTP server timeouts. The rationale is that type of users who need to tweak these values (according to their infrastructure) are more likely to work at the configuration level.

### 📝 Proposal

Extend `readHttpServerOptions` to support additional keys to customise the HTTP server under `backend.http` in `app-config.yaml`. This could be used to override all potential values, or only a subset of allowed values.

### 🔄 Alternatives

If there's already another way of achieving this that I have missed, happy to add to the documentation 🙏🏻 

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Yes, and I have enough information to get started

Rugvip:
Yep makes sense! 👍

Only bit to figure out imo is where to put the config. Framework SIG brought up options like `backend.serverOptions` to avoid confusion with `backend.https`. I could also see `backend.httpServer` or just `backend.server`.

BethGriggs:
This was shipped in [Backstage v1.44](https://backstage.io/docs/releases/v1.44.0#easier-server-options-through-config) (https://github.com/backstage/backstage/pull/30483)

Issue #21808, opened 2023-12-11
🚀 Feature: Allow custom Node HTTP Server configuration in new Backend System


eipc16:
### 🔖 Feature description

Old backend system allows for modification of Node HTTP Server configuration by exposing `http.Server` class through `.start()` method in the ServiceBuilder class ([reference](https://backstage.io/docs/reference/backend-common.servicebuilder)). Unfortunately, this is no longer the case in the new Backend System where the `http.Server` object is completely encapsulated in the `rootHttpRouterServiceFactory`. 

I would like to propose implementing a way to update the server configuration during application startup.

### 🎤 Context

We are currently using this "feature" in the old backend system to modify server configuration like `keepAliveTimeout` ([docs](https://nodejs.org/api/http.html#serverkeepalivetimeout)). 

We need to modify those properties, because default value for `keepAliveTimeout` (and few other properties) does not play well with our infrastructure configuration. Basically, when Idle Timeout of Load Balancer is greater than Keep Alive Timeout of Application (which is 5s for Node) then 502s happen a lot and we really don't want to change Idle Timeout on our Load Balancer.

Currently, the only way to achieve this in the new backend system is to provide a completely custom implementation of the `rootHttpRouterServiceFactory` it would be nice if we could use the existing factory.

### ✌️ Possible Implementation

Two solutions come to my mind for now, but I'm open for discussing other proposals. I'm also willing to submit a PR, but want to know which approach will be better from your perspective before I start to work.

### Option 1: Server configuration in `app-config.yaml`

Allow configuration of selected server properties in `app-config.yaml` and load them in `readHttpServerOptions`. In this approach it's harder to "break" the server and is consistent with the more declarative approach of the new Backend System.

Few selected configuration options could be exposed under `http.server` key. The options that I'm most interested in are:

- [server.keepAliveTimeout](https://nodejs.org/api/http.html#serverkeepalivetimeout)
- [server.timeout](https://nodejs.org/api/http.html#servertimeout)
- [server.headersTimeout](https://nodejs.org/api/http.html#serverheaderstimeout)

Though, other properties could be useful to other Backstage users.

### Option 2: Server configuration in `RootHttpRouterFactoryOptions`

Expose a callback in `RootHttpRouterFactoryOptions` that would allow changing server configuration before (or after?) triggering the `.start()` method. It's more prone to user errors, but also more flexible.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

[...]

PR #30483, opened 2025-07-09
feat(http): support server-level timeout and socket options via app-c…

BethGriggs:
…onfig

Adds support for configuring server-level HTTP options through the `app-config.yaml` file under the `backend.server` key. This includes support for: `headersTimeout`, `keepAliveTimeout`, `requestTimeout`, `timeout`, `maxHeadersCount`, and `maxRequestsPerSocket`.

These options are passed directly to the underlying Node.js HTTP server, when omitted, the default values are used.

Refs: https://github.com/backstage/backstage/issues/21808
Refs: https://github.com/backstage/backstage/issues/30449

----

From reviewing past issues and user requests related to HTTP server tuning, I found that the options implemented here cover the requested settings. These are also the ones that can be safely applied to the Server instance after creation, rather than needing to modify how the server is constructed.

### DIFF
--- a/packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
+++ b/packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
@@ -146,15 +146,7 @@ const rootHttpRouterServiceFactoryWithOptions = (
                     `ISO duration (e.g., 'PT30S'), or duration object (e.g., {seconds: 30}). ` +
                     `Falling back to number parsing.`,
                 );
-                // Fallback to reading as number if duration parsing fails
-                const fallbackValue = serverConfig.getOptionalNumber(key);
-                if (fallbackValue === undefined && typeof value === 'string') {
-                  logger.error(
-                    `backend.server.${key} value '${value}' could not be parsed as either ` +
-                      `a duration or a number. This setting will be ignored.`,
-                  );
-                }
-                return fallbackValue;
+                return undefined;
               }
             };
 


BethGriggs:
Is there anything I can do to help move this one forward?

BethGriggs:
Not stale, perhaps I can put this on the agenda for the next Framework SIG?

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/config.ts
```diff
@@ -99,3 +100,32 @@ function readHttpsOptions(config?: Config): HttpServerOptions['https'] {
...
+): HttpServerOptions['serverOptions'] {
+  const serverConfig = config?.getOptionalConfig('server');
+  if (!serverConfig) {
+    return undefined;
+  }
+
+  const serverOptions: HttpServerOptions['serverOptions'] = {};
+
+  const keys = [
+    'headersTimeout',
```
freben: Note that all changes to config should be reflected [in the corresponding schema](https://github.com/backstage/backstage/blob/f247fe4090d3020e2f9f3c5fc21faf3da97a541a/packages/backend-defaults/config.d.ts) as well.

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/createHttpServer.ts
```diff
@@ -84,17 +84,40 @@ async function createServer(
...
+    } else {
+      server = https.createServer(certificate, listener);
     }
-    return https.createServer(certificate, listener);
+  } else {
+    server = http.createServer(listener);
+  }
+
+  // apply custom server options
+  if (options.serverOptions) {
```
freben: Hm. The canonical place where these overall settings are changed (even if you are an adopter overriding things, before this PR), [is in `configure`](https://github.com/backstage/backstage/blob/f247fe4090d3020e2f9f3c5fc21faf3da97a541a/packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts#L113). Do you think it would make sense to change the `applyDefaults` function instead, to read config and apply all of these fields? Especially since they clearly can be set after the fact, not just during construction of the server.

That way, you would not have to change `types.ts` to contain the server options either, because `applyDefaults` would be the one reading from config.

I understand that it's a bit subjective whether something that appears in config should be applied unconditionally or not, so I'm happy to hear your thoughts on the suggestion

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/config.ts
```diff
@@ -99,3 +100,32 @@ function readHttpsOptions(config?: Config): HttpServerOptions['https'] {
...
+    'requestTimeout',
+    'keepAliveTimeout',
+    'timeout',
+    'maxHeadersCount',
+    'maxRequestsPerSocket',
+  ] as const;
+
+  for (const key of keys) {
+    const value = serverConfig.getOptionalNumber(key);
```
freben: While I appreciate that numbers might be a good familiar format for those who are used to configuring servers, the Backstage config way commonly is `HumanDuration | string` for all durations. What would you think about making the config.d.ts file contain `HumanDuration | string | number` since people might actually expect numbers to work here? Then they could also do `{ hours: 1 }` or `'30 minutes'` as is customary if they prefer.

Then your might make a little helper function that checks the type of the thing if present, and decides whether to just pass through a number or call `readDurationFromConfig` on it

BethGriggs:
Thanks for the review @freben, I plan to pick this back up now I'm back from vacation

freben:
Just noting that I'm holding off until this is out of Draft mode. Hope you're doing well being back!

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/config.ts
BethGriggs: Added to `packages/backend-defaults/config.d.ts` in the latest [commit](https://github.com/backstage/backstage/pull/30483/files#diff-3b45dbb2aa9a6a351eed560b0388e73b659ab55bc693efabecbb307b7bbe9b0a)

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/createHttpServer.ts
BethGriggs: I have tried to move logic to `applyDefaults()` in latest commit d501e8c7686500428b008d869a6297cceb464bff, I'm hoping I've interpreted the suggestion correctly 🙏🏻 

Also, applying the config unconditionally feels fine to me on the basis that anyone who has gone to the effort of supplying these additional values has good control. So, I am hopeful that it is unlikely to cause confusion or any problems that can’t be easily remediated (by just removing/adjusting the values).

packages/backend-defaults/src/entrypoints/rootHttpRouter/http/config.ts
BethGriggs: I believe I've added support for each of these type inputs, and covered a mixture in the added test cases.

BethGriggs:
Thanks for patience - I figured I'd put it in draft for a little while to give me time to read through the logic/feedback with a fresh mind.

One thing I’m still unsure about is whether the test cases are sufficient. Right now they only check that the server accepts the supplied config and starts. I figured we do not need to go as far as validating that Node.js internals apply each option, but let me know if you feel stronger validation is needed.

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
```diff
@@ -117,6 +117,83 @@ const rootHttpRouterServiceFactoryWithOptions = (
...
+                );
+                // Fallback to reading as number if duration parsing fails
+                const fallbackValue = serverConfig.getOptionalNumber(key);
+                if (fallbackValue === undefined && typeof value === 'string') {
+                  logger.error(
+                    `backend.server.${key} value '${value}' could not be parsed as either ` +
+                      `a duration or a number. This setting will be ignored.`,
+                  );
+                }
+                return fallbackValue;
```
freben:
```suggestion
                return undefined;
```


this should no longer be needed since it's handled above

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
```diff
@@ -117,6 +117,83 @@ const rootHttpRouterServiceFactoryWithOptions = (
           if (trustProxy !== undefined) {
             app.set('trust proxy', trustProxy);
           }
+
+          // Apply server-level HTTP options from config
+          const backendConfig = config.getOptionalConfig('backend');
+          const serverConfig = backendConfig?.getOptionalConfig('server');
+
+          if (serverConfig) {
```
freben, line 125: Simplification suggestion: You could skip this entire wrapping `if`, and the consts above. Then inside `readDurationValue` you could do

```ts
const fullKey = `backend.server.${key}`;
const value = config.getOptional(fullKey);
if (value === undefined || typeof value === 'number)' {
  return value;
}
```

and of course correspondingly `readDurationFromConfig(config, { key: fullKey }) `

and then also refer to `fullKey` in the logged error below instead of reconstructing it

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
BethGriggs: I believe I added this to handle people mistakenly supplying plain numbers as Strings, but it looks like the general Backstage convention is to not allow that (from [`readDurationFromConfig.ts`](https://github.com/backstage/backstage/blob/master/packages/config/src/readDurationFromConfig.ts#L148C1-L150C20)):

```
    // We explicitly disallow the only-digits form of the ms library, because
    // from a configuration perspective it's just confusing to even be able to
    // specify that
```

I'll remove it to align with the convention, thanks!

freben: We said in an earlier comment that it's OK here though - since the underlying setting is so very node specific and well known to people, and maybe they expect to be able to specify the same things as they do when doing this hard coded.

And either way, you do have the number check higher up in the code already; I believe this block was a remnant from before that was added.

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
freben: We can do this later.

packages/backend-defaults/src/entrypoints/rootHttpRouter/rootHttpRouterServiceFactory.ts
BethGriggs: I was more considering the confusion between being able to supply `timeout: 3000` but not `timeout: "3000"` as it'll be rejected by `readDurationFromConfig`.

PR #30485, opened 2025-07-09
Add signadot.yaml

davixcky:

Add signadot plugin to the catalog list

#### :heavy_check_mark: Checklist\

### DIFF
--- a/microsite/data/plugins/signadot.yaml
+++ b/microsite/data/plugins/signadot.yaml
@@ -1,0 +1,10 @@
+---
+title: Signadot
+author: signadot
+authorUrl: https://www.signadot.com/
+category: CI 
+description: Create and manage Signadot resources like Sandboxes, Clusters, etc
+documentation: https://github.com/signadot/backstage-plugin
+iconUrl: https://avatars.githubusercontent.com/u/57196635
+npmPackageName: @signadot-lib/backstage-plugin
+addedDate: '2025-07-09' 


microsite/data/plugins/signadot.yaml
```diff
@@ -0,0 +1,10 @@
+---
+title: Signadot
+author: signadot
+authorUrl: https://www.signadot.com/
+category: CI 
+description: Create and manage Signadot resources like Sandboxes, Clusters, etc
+documentation: https://github.com/signadot/backstage-plugin
```
aramissennyeydd, line 7: 👋 I get a 404 clicking this link, is this plugin open source?

davixcky: Hi @aramissennyeydd, plugin is now open source. Thanks for checking

benjdlambert:
Hmm, should the `apiKey` https://github.com/signadot/backstage-plugin/blob/main/config.d.ts#L23-L27 really be visible to the frontend? This means that anyone can access that secret which seems a little bit concerning?

davixcky:
> Hmm, should the `apiKey` https://github.com/signadot/backstage-plugin/blob/main/config.d.ts#L23-L27 really be visible to the frontend? This means that anyone can access that secret which seems a little bit concerning?

The plugin needs to fetch data from the Signadot API. Do you have a recommended approach for handling this—perhaps using a custom backend?

benjdlambert:
@davixcky yeah ideally there should be a backend plugin.

However, there is another option which is less secure than having a backend plugin, but better than leaking the secret to the frontend which would be to add the header to the proxy config like you already have setup like some other plugins do:

https://github.com/backstage/community-plugins/blob/d44eb8d2406f60ba70f2c4f9c798b61511f2609d/workspaces/newrelic/plugins/newrelic/README.md?plain=1#L16-L22

That way you don't need to expose the API key to the frontend it's automatically added to any request which goes through the proxy, but again, this is also has downsides that anyone that makes requests to the proxy become authenticated with the service.

davixcky:
> @davixcky yeah ideally there should be a backend plugin.
> 
> However, there is another option which is less secure than having a backend plugin, but better than leaking the secret to the frontend which would be to add the header to the proxy config like you already have setup like some other plugins do:
> 
> https://github.com/backstage/community-plugins/blob/d44eb8d2406f60ba70f2c4f9c798b61511f2609d/workspaces/newrelic/plugins/newrelic/README.md?plain=1#L16-L22
> 
> That way you don't need to expose the API key to the frontend it's automatically added to any request which goes through the proxy, but again, this is also has downsides that anyone that makes requests to the proxy become authenticated with the service.

Thanks, we decided to [move forward](https://github.com/signadot/backstage-plugin/pull/7) with the Proxy header. Already fixed in the latest example.

davixcky:
Hey @benjdlambert any update on this? Anything else is required from our end? Just pinging in case were missed

benjdlambert (approved):
:tada:

benjdlambert:
Apologies for the delay, it's been a busy week 😅

davixcky:
> Apologies for the delay, it's been a busy week 😅

No worries at all 😌 I think we’ve all had one of those weeks. 🎉🎉

Issue #30363, opened 2025-06-23
Consolidate Database Documentation


awanlin:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🗃️ Project Area

Documentation

### 🔗 External Integration

N/A

### 📝 Description

There are a number of documentation pages that cover Database setup, configuration, and use. It would make them much more useful if they were consolidated into a single place and then referred to as needed.

https://backstage.io/docs/tutorials/configuring-plugin-databases
https://backstage.io/docs/tutorials/switching-sqlite-postgres

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

Bumbleblo:
Hi @awanlin, I can help here! ✋ 

Just to starting at some point I'm proposing some changes on https://github.com/backstage/backstage/pull/30496. My reasoning here is that maybe we can have a new section just to address admin docs, seems that we can centralize a lot of pages this way. 

WDYT? I would appreciate feedback on that.

awanlin:
Hi @Bumbleblo, ideally most of the two docs should be added here: https://backstage.io/docs/getting-started/config/database. This is going to be the first place that a new Backstage adopter lands when getting started and honestly already has a lot of the content from these two pages.

Ayushmore1214:
@awanlin is @Bumbleblo Still working on this issue or I mind giving it a try

awanlin:
There's a PR in the works for this, I would leave this in @Bumbleblo hands for the time being, thanks!

Ayushmore1214:
No worries , @Bumbleblo If you need a hand just ping me

PR #30496, opened 2025-07-10
Database docs

Bumbleblo:

I'm proposing this solution to https://github.com/backstage/backstage/issues/30363. 

The reasoning that I'm following is to create a new section just to database administration content, plugin table organization and other things can be addressed outside this new section. 

2025-07-10-010725_2301x843_scrot

Besides creating this new section, I also changed the following files:

* Deleted [sqlite migration guide](https://backstage.io/docs/tutorials/switching-sqlite-postgres) and moved pool connection and schema settings to the client configuration doc. 


Parsifal-M:
Heya 👋 

Good stuff, I'm not sure about putting it on the top level though.

The reason for that is you could say the same about authentication, deployment... etc... then we end up back to not long ago when the side nav was just a massive list 😅

What was your reasoning for wanting it at the top level? Did you struggle to find it where it was? Or it didn't make sense to look where it was? Just wondering so we can think of a plan 💭

Bumbleblo:
> What was your reasoning for wanting it at the top level? Did you struggle to find it where it was? Or it didn't make sense to look where it was? Just wondering so we can think of a plan 💭

Hi @Parsifal-M! I was trying to think in a way to group some database stuff, as the issue suggested.

I create the top section just to have a place where an admin could read all docs related to handling databases: how to set up, rollback migrations, define schemas, etc. But the intention is just to have an index for that, could be a subsection on the Configuration, or we can just index those documents in this section.

awanlin:
Hi @Bumbleblo, ideally most of the two docs should be added here: https://backstage.io/docs/getting-started/config/database. This is going to be the first place that a new Backstage adopter lands when getting started and honestly already has a lot of the content from these two pages.

microsite/sidebars.ts
```diff
@@ -393,6 +392,10 @@ export default {
         ],
       },
     ],
+    'Database': [
+      'database/database',
+      'database/manual-knex-rollback',
```
awanlin: Please don't move this. It's a great single purpose tutorial with a clear goal and use. It does make sense to link to this from the database documentation though.

Bumbleblo: I changed my approach and therefore those files are in the same place now.

Bumbleblo:
I did some changes, the [migration guide to postgres](https://backstage.io/docs/tutorials/switching-sqlite-postgres) still is deleted, since there are a lot of overlapping between the database setup guide, I tried to add only the information that wasn't on the setup guide.

About the guide to setup plugin databases: I think that add a lot of context specific to plugin developers, seems to me that should be under the "Framework > Backend System > Bulding Plugins & Modules", WDYT?

microsite/sidebars.ts
Bumbleblo: I'll add that was resolved since it's outdated

Parsifal-M:
> About the guide to setup plugin databases: I think that add a lot of context specific to plugin developers, seems to me that should be under the "Framework > Backend System > Bulding Plugins & Modules", WDYT?

I think theres a slight misconception with these docs, I think they should also be in the [Getting Started](https://backstage.io/docs/getting-started/config/database/)

The reason being, its just an "extension" of your database setup i.e. setting up a plugin you build yourself of install or anything related to your backstage plugin(s) database is an extension of the config you've defined in the [Getting Started](https://backstage.io/docs/getting-started/config/database/) section IMO.

I feel like we could fairly easily merge the two into a single docs page maybe shuffling some things around 💭 

Edit: https://backstage.io/docs/plugins/ section _could_ also work but I'm trying to avoid having to update docs in multiple places if something changes of course 😅

Bumbleblo:
@Parsifal-M @awanlin I think that it's in a good point to receive feedback now.

Parsifal-M:
> @Parsifal-M @awanlin I think that it's in a good point to receive feedback now.

Awesome! I have some time aside tomorrow I can have a read of it!

Thanks for taking the time to put this together!

docs/getting-started/config/database.md
```diff
@@ -25,12 +25,17 @@ This guide assumes a basic understanding of working on a Linux based operating s
 
 ## 1. Install and Configure PostgreSQL
 
+In this section, you will know some options to create you own Postgres instance. Those methods best fit the use case of local
```
Parsifal-M:
```suggestion
In this section, we will go over some options to create you own Postgres instance. These methods best fit the use case of local
```

docs/getting-started/config/database.md
```diff
@@ -108,59 +153,269 @@ If you opt for the second option of replacing the entire string, take care to no
...
+    # highlight-add-end
+```
 
-## Alternatives
+## 3. Configuring Database settings per plugin
 
-You may not want to install Postgres locally, the following sections outline alternatives.
+By default, Backstage uses automatically created databases for each plugin whose
+names follow the `backstage_plugin_<pluginId>` pattern, e.g.
+`backstage_plugin_auth`. You can configure a different database name prefix for
```
Parsifal-M:
```suggestion
By default, Backstage automatically creates databases for each plugin with the following naming convention `backstage_plugin_<pluginId>`, e.g.
`backstage_plugin_auth`. You can configure a different database name prefix for
```

docs/getting-started/config/database.md
```diff
@@ -108,59 +153,269 @@ If you opt for the second option of replacing the entire string, take care to no
...
+use cases where you have multiple deployments running on a shared database
+instance or cluster.
 
-### Docker
+With infrastructure defined as code or data (Terraform, AWS CloudFormation,
```
Parsifal-M:
```suggestion
With infrastructure defined as code (Terraform, AWS CloudFormation,
```

docs/getting-started/config/database.md
```diff
@@ -108,59 +153,269 @@ If you opt for the second option of replacing the entire string, take care to no
...
+`@backstage/backend-common`. We will now cover how to use and configure
+Backstage's databases.
 
-First we need to pull down the container image, we'll use Postgres 17, check out the [Postgres Version Policy](../../overview/versioning-policy.md#postgresql-releases) to learn which versions are supported.
+### Configuration
 
-```shell
-docker pull postgres:17.0-bookworm
+You should set the base database client and connection information in your
+`app-config.yaml` (or equivalent) file. The base client and configuration is
```
Parsifal-M:
```suggestion
`app-config.yaml` file. The base client and configuration is
```

docs/getting-started/config/database.md
```diff
@@ -108,59 +153,269 @@ If you opt for the second option of replacing the entire string, take care to no
...
+**`backend.database.plugin.<pluginId>`**. As an example, `catalog` is the
+`pluginId` for the catalog plugin and any configuration defined under that block
+is specific to that plugin. We will now explore more detailed example
+configurations below.
+
+#### Minimal In-Memory Configuration
+
+In the example below, we are using `better-sqlite3` in-memory databases for all
+plugins. You may want to use this configuration for testing or other non-durable
+use cases.
```
Parsifal-M:
```suggestion
In the example below, we are using `better-sqlite3` in-memory databases for all
plugins. You may want to use this configuration for testing or other non-durable
use cases. It is not recommended to use this in a production deployment!
```

docs/getting-started/config/database.md
```diff
@@ -108,59 +153,269 @@ If you opt for the second option of replacing the entire string, take care to no
...
+    connection: 'postgresql://foo:bar@some.example-pg-instance.tld:5432'
+    plugin:
+      auth:
+        client: better-sqlite3
+        connection: ':memory:'
+```
+
+## Manual Rollback
+
+In some cases you may want to downgrade a migration applied on the database due to some problem, under the hood Backstage uses [Knex](https://knexjs.org/) to handle [migrations](https://knexjs.org/guide/migrations.html) and you can interact direct with Knex interface to **rollback a migration** applied during an upgrade. You can use the `migrate:down` command to rollback a specific migration. You can also use the `migrate:rollback` command to rollback the last batch of migrations. This is necessary because Knex will mark migrations as corrupted if you try to downgrade your Backstage instance without the rollback. Be aware to run those commands in the new version of the Backstage instance, so you can avoid the corrupted migrations for lower versions.
```
Parsifal-M:
```suggestion
In some cases you may want to rollback a migration applied on the database, under the hood Backstage uses [Knex](https://knexjs.org/) to handle [migrations](https://knexjs.org/guide/migrations.html) and you can interact direct with Knex interface to **rollback a migration** applied during an upgrade. You can use the `migrate:down` command to rollback a specific migration. You can also use the `migrate:rollback` command to rollback the last batch of migrations. This is necessary because Knex will mark migrations as corrupted if you try to downgrade your Backstage instance without the rollback. Be aware to run those commands in the new version of the Backstage instance, so you can avoid the corrupted migrations for lower versions.
```

Parsifal-M (commented):
Heya! 👋 

Sorry for the delay, I'm technically on holiday 😬 - I did a first read through and left some suggestions, I'm not 100% sure what to do about the headings, some have numbering and some do not, I'd rather we just stick to no numbering to keep it consistent 💭

WDYT?

Bumbleblo:
> I'm not 100% sure what to do about the headings, some have numbering and some do not, I'd rather we just stick to no numbering to keep it consistent 💭

I'm fine with that. Looking on other pages seems that the pattern is no number at all.

Bumbleblo:
> Sorry for the delay, I'm technically on holiday 😬

Thanks @Parsifal-M, I really appreciate the help. 

I should be able to get more reviews now that I'm integrating a little bit more with the community, so no worries. Next time I can get this on SIG to get at least something to working with if isn't any.

Parsifal-M (approved):
Heya, sorry for letting this slip, IMHO I think it looks good, but I'd also like to see if the other @backstage/documentation-maintainers agree with going this way 😄

Bumbleblo:
> Heya, sorry for letting this slip, IMHO I think it looks good, but I'd also like to see if the other @backstage/documentation-maintainers agree with going this way 😄

Hi @Parsifal-M, thanks for reviewing. I had one feedback about being too much content for a getting started page, I'm working on an alternative to create an "advanced settings" page in order to address this issue. 

Would be awesome to have that and other issues described here.

PR #30497, opened 2025-07-10
Follow same pattern for keyboard shortcuts in docs

AmbrishRamachandiran:
Follow same pattern for keyboard shortcuts in docs

image

current:
image

updated:

image


AmbrishRamachandiran:
@Rugvip Any update on this please

Parsifal-M (commented):
I think if we want to me specific, on Windows I guess it would be `Ctrl + C` but on Mac would be `control + C` ?

PR #30498, opened 2025-07-10
Reduce API calls made for single user events in GitHub Org Provider

### LINKED ISSUE #30397: GitHub Provider - Excessive API usage for large teams
### 📜 Issue Labels

### 🔎 Search Terms

```plain
github provider api rate limit
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

GitHub

### 📝 Description & Context

The GitHub provider will make excessive number of API calls when it receives an organisation event to refresh a single user, if this user is a member of a large GH team (>100 members).

When the provider processes an event for a single users be in added/changed in an org it triggers the function `[onMemberChangeInOrganization](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts#L484)` this in turn triggers the function `[getOrganizationTeamsFromUsers](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts#L559C33-L559C62)` which is intended to gather the teams that the user is a member of.

However, this function seems shared with other functions so that for each request it also enumerates all the members of all teams a user is a member of. Members are paginated by 100.

In our use case with have a team containing 6k users in GH. Most people in the org are members of this team.

So in this case, each event to update a user uses 60+ requests just to enumerate this team. We have 6k+ people which is 360k api calls to load everything.. 

### 👍 Expected Behavior

The provider should be economical with the required data and only load the required info.

In this case when refreshing a single user it does not need to load the membership of all teams.

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

_No response_

ScottGuymer:

Currently the provider will load all users group and all members of the groups when responding to an event on a single user. This can cause excessive API calls to page through group members when this information is just discarded in this instance.

This change introduces single user versions of the user group resolution code in the multi org provider that will not page through all membership when updating a single user.

Closes #30397

### DIFF
--- a/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts
+++ b/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts
@@ -801,15 +801,15 @@ export class GithubMultiOrgEntityProvider implements EntityProvider {
           headers: orgHeaders,
         });
 
-        const { teams } = await getOrganizationTeamsFromUsers(
+        const { teams } = await getOrganizationTeamsForUser(
           orgClient,
           userOrg,
-          [login],
+          login,
           this.defaultMultiOrgTeamTransformer.bind(this),
         );
 
         if (areGroupEntities(teams)) {
-          assignGroupsToUsers([user], teams);
+          assignGroupsToUser(user, teams);
         }
       }
 
--- a/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.test.ts
+++ b/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.test.ts
@@ -2306,23 +2306,7 @@ describe('GithubMultiOrgEntityProvider', () => {
             organization: {
               teams: {
                 pageInfo: { hasNextPage: false },
-                nodes: [
-                  {
-                    slug: 'team',
-                    combinedSlug: 'orgA/team',
-                    name: 'TeamA',
-                    description: 'The one and only team',
-                    avatarUrl: 'http://example.com/team.jpeg',
-                    editTeamUrl: 'https://example.com',
-                    parentTeam: {
-                      slug: 'parent',
-                    },
-                    members: {
-                      pageInfo: { hasNextPage: false },
-                      nodes: [],
-                    },
-                  },
-                ],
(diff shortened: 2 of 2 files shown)


.changeset/ripe-comics-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-github': minor
```
benjdlambert: This isn't a breaking change right? I think we can set this to `patch` seen as the package is 0.x currently.

benjdlambert (commented):
:wave: thanks for this, just a quick comment about the changeset.

.changeset/ripe-comics-sip.md
ScottGuymer: Yeah this is non breaking and wont affect any outward behaviour

```suggestion
'@backstage/plugin-catalog-backend-module-github': patch
```

plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts
```diff
@@ -556,15 +558,15 @@ export class GithubMultiOrgEntityProvider implements EntityProvider {
...
           orgClient,
           userOrg,
-          [login],
+          login,
           this.defaultMultiOrgTeamTransformer.bind(this),
         );
 
         if (isUserEntity(user) && areGroupEntities(teams)) {
-          assignGroupsToUsers([user], teams);
+          assignGroupsToUser(user, teams);
```
schultzp2020, line 569: I think you can make the same change [here](https://github.com/philips-forks/backstage/blob/scott/fix_provider/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts#L812).

ScottGuymer: Nice catch.

I had to slightly change a test to accommodate this. 

If the user is no longer part of a team it will not be returned from the graphQL endpoint, this test seemed to expect the team would be returned but the user would not be listed in the members.

benjdlambert (approved):
Nice - let's give this a go! :tada:

PR #30502, opened 2025-07-10
feat(techdocs): Allow client caching for hashed CSS files

GabDug:

Adds cache control headers for CSS files with 8-character hash suffixes (e.g., `main.a1b2c3d4.min.css`).

Sets `Cache-Control: public, max-age=31536000, immutable` to allow browsers to cache these files indefinitely since their content never changes.

Very conservative approach - only allows caching for hashed CSS files, which should save ~23kb per page.

We could add support for other files, but I haven't found other use-cases (must say I haven't looked very far, let me know if you find something for mkdocs that will suffix images with a unique hash!).

Tested in backstage app...
image

...and in production
Network tab showing request served from cache, with the proper Cache-Control header

- [X] Screenshots attached

### DIFF
--- a/.changeset/swift-ravens-argue.md
+++ b/.changeset/swift-ravens-argue.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-techdocs-node': minor
+---
+
+feat(techdocs): add client caching for hashed CSS files
--- a/plugins/techdocs-node/src/stages/publish/awsS3.ts
+++ b/plugins/techdocs-node/src/stages/publish/awsS3.ts
@@ -46,7 +46,7 @@ import {
   bulkStorageOperation,
   getCloudPathForLocalPath,
   getFileTreeRecursively,
-  getHeadersForFileExtension,
+  getHeadersForFilename,
   getStaleFiles,
   isValidContentPath,
   lowerCaseEntityTriplet,
@@ -468,8 +468,7 @@ export class AwsS3Publish implements PublisherBase {
       }
 
       // Files with different extensions (CSS, HTML) need to be served with different headers
-      const fileExtension = path.extname(filePath);
-      const responseHeaders = getHeadersForFileExtension(fileExtension);
+      const responseHeaders = getHeadersForFilename(filePath);
 
       try {
         const resp = await this.storageClient.send(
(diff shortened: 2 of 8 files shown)


GabDug:
unstale

GabDug:
unstale

GabDug:
Not sure who to ping to reopen this PR, maybe @freben? Thanks!

freben:
Sure.

GabDug:
Still looking for feedback, maybe from @PeaWarrior? Thanks!

PeaWarrior:
Hey @GabDug, thanks for looking into this!

Do we know what the hashed CSS filenames are based on? Are they derived from the actual file content, or from something else like the file path or build timestamp? (in regards to techdocs-core)

GabDug:
Hey @PeaWarrior,
As far as I known they're coming from [mkdocs-material](https://github.com/squidfunk/mkdocs-material/blob/97e275c953e5b8255ca24ca166ddd802e23d84b1/material/templates/assets/stylesheets/main.e53b48f4.min.css), and appears to change depending on the version of mkdocs-material.

I don't think techdocs-core plays a role here, but could be mistaken.

GabDug:
Hey @PeaWarrior or @squid-ney, would it be possible to reopen this PR? Thanks!

squid-ney:
Hi @GabDug , I think what @PeaWarrior 's question is getting at is that if the hashing is not content based then we could run into some issues with the indefinite caching strategy. But, it would be unusual if `mkdocs-material` didn't use a content hash in the file name because that would not work with industry expected cache-busting techniques. What do you think @GabDug ? I think it should be low-risk enough to merge specially since it only caches css files.

GabDug:
@squid-ney Mkdocs material uses [css-nano & postcss](https://github.com/squidfunk/mkdocs-material/blob/master/tools/build/transform/index.ts#L98), [with a custom `digest` function to compute the hash with the file content](https://github.com/squidfunk/mkdocs-material/blob/master/tools/build/transform/index.ts#L78), so I think we can use this content hash.

There's one (unsupported) edge case that could be impacted: if docs are not built server sides & users transform the CSS between the `techdocs build` & `techdocs publish`. I would say this was never a supported way of styling the docs.

GabDug:
Anti stale bump

squid-ney:
> There's one (unsupported) edge case that could be impacted: if docs are not built server sides & users transform the CSS between the techdocs build & techdocs publish. I would say this was never a supported way of styling the docs.

Thanks for identifying this. I think making the `max-age` value configurable would be reasonable for a `ttl` variable. We inject some styles before the generation stage and I could see a possibility of another backstage user injecting styles between `generation` and `publish`.

PR #30515, opened 2025-07-13
feat: revamped sidebar with collapsible groups, header toggle, and theme switcher

naaa760:
- Improved the Backstage sidebar UX:

- Added SidebarToggle (collapse / expand) in the header
- Added ThemeToggle (light / dark) in the footer
- Replaced fly-out sub-menus with expandable SidebarGroup dropdowns
- Introduced SidebarContext for collapsed / group-expansion state (persisted via localStorage)
- Updated exports and cleaned file headers

- heavy_check_mark: Checklist

- A changeset describing the change and affected packages
- Added or updated documentation
- Tests for new functionality and regression tests for bug fixes
- Screenshots attached (for UI changes)
- All commits have a Signed-off-by line

### DIFF
--- a/packages/core-components/src/layout/Sidebar/Controls/ThemeToggle.tsx
+++ b/packages/core-components/src/layout/Sidebar/Controls/ThemeToggle.tsx
@@ -1,0 +1,42 @@
+
+
+import { IconButton, Tooltip, makeStyles } from '@material-ui/core';
+import Brightness4Icon from '@material-ui/icons/Brightness4';
+import Brightness7Icon from '@material-ui/icons/Brightness7';
+import { useApi, appThemeApiRef } from '@backstage/core-plugin-api';
+import useObservable from 'react-use/lib/useObservable';
+
+const useStyles = makeStyles(theme => ({
+  button: {
+    color: theme.palette.navigation.color,
+    width: '40px',
+    height: '40px',
+  },
+}));
+
+export const ThemeToggle = () => {
+  const classes = useStyles();
+  const appThemeApi = useApi(appThemeApiRef);
+  const themeId = useObservable(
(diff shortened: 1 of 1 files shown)


patriciamcphee:
@naaa760, this is a good start. I have the working code from my design, so I can help implement it.  

Also, I designed it so that the sidebar is more customizable using the `packages/app/src/components/Root/Root.tsx` file (which is missing from this PR).

nehaaprasad:
> @naaa760, this is a good start. I have the working code from my design, so I can help implement it.
> 
> Also, I designed it so that the sidebar is more customizable using the `packages/app/src/components/Root/Root.tsx` file (which is missing from this PR).

Thanks for letting me know!

Parsifal-M:
Heya 👋 

Just my two cents here, without looking into this too deeply a few things to consider:

- Will this all need to be changed once the new frontend system drops? If so maybe its worth pausing further work on this?
- I would probably in all honesty before making too many changes try to get some additional feedback on the idea/implementation as its quite a focal point of the app so it will of course impact everyone who uses the framework and some of this could be down to personal choice and flavour

This could be raised in the framework sig to get some initial thoughts, or maybe create a thread in the discord to raise some awareness.

nehaaprasad:
Hey @Parsifal-M and @awanlin !! 👋

Thanks for the feedback - totally makes sense! You're absolutely right that this is a pretty big change for such a core part of the app, and I definitely don't want to get too far down the rabbit hole if the new frontend system is going to shake things up anyway.

I'll hit pause on this for now and take your advice.
- Will start a discussion in #framework-sig to get some initial thoughts from the maintainers
- Keep this PR in draft mode until we have more consensus

> Really appreciate you both taking the time to review this - better to get it right than rush it!

PR #30519, opened 2025-07-14
UI - Move Select to React Aria

cdedreuille:

We are moving our Select component to use React Aria under the hood. Even though we tried to keep this move close to the original API, we introduced a couple of breaking changes in both the props and theming.
- Add support for  a leading `icon` property.
- Fix popover width being contained within the trigger
- Better error handling 


packages/canon/src/components/Select/Select.tsx
```diff
@@ -30,103 +40,72 @@ export const Select = forwardRef<HTMLDivElement, SelectProps>((props, ref) => {
...
     options,
     placeholder = 'Select an option',
-    size = 'medium',
-    required,
-    error,
+    size = 'small',
+    icon,
+    'aria-label': ariaLabel,
+    'aria-labelledby': ariaLabelledBy,
+    isRequired,
```
Rugvip, line 47: Was this switched to match React Area better? Otherwise I prefer omitting bool prefixes and just having it be `required` like before

Rugvip (approved):
One bit, but otherwise ready to :shipit:

packages/canon/src/components/Select/Select.tsx
cdedreuille: It is to match React Aria and it is connected with all the other ones as well. We can do a naming pass before beta if you want to get things aligned.

PR #30520, opened 2025-07-14
Update app-custom-theme.md

duncan485:

The docs suggest to create the Icon file in assets/icons/CustomIcons. I noticed that the include does not match that path. This change will make sure the exampleIcon will work for anyone following the docs.


benjdlambert (approved):
Thanks :tada:

PR #30531, opened 2025-07-15
feat: add props to overridable `NotFoundErrorPage` component and update usages

### LINKED ISSUE #30450: How to implement a custom TechDocs error page in Backstage
### 📝 Acknowledgement

### Content

I’m currently running Backstage version 1.40.2, and I want to implement a custom error page for TechDocs so that when a doc site fails to load (e.g., missing mkdocs.yaml, 404, or build failure), we can show a friendly custom message instead of the default error screen.

I noticed that this capability was introduced around v1.36.0 (https://github.com/backstage/backstage/blob/master/docs/releases/v1.36.0-changelog.md) but there isn’t much documentation or an example for how to use this feature properly on the latest versions.

What I’m trying so far:

I created a simple CustomTechDocsErrorPage.tsx React component.

I tried passing it as the errorPage prop to <TechDocsReader> and <TechDocsPage> in my app.

I also saw that some people wrap TechDocs in an <ErrorBoundary> but I’m not sure if that’s the preferred approach.

It’s not clear to me if there are any other steps needed, like updating app-config.yaml or registering the error page somewhere globally.

What I’d love help with:
✅ What’s the recommended or “best practice” way to use a custom TechDocs error page in Backstage v1.40.2?
✅ Do I need to configure anything else in app-config.yaml or TechDocsReader routes?
✅ If possible, can you please share a complete step-by-step example showing where exactly to place the custom component, how to pass it, and how it ties into the TechDocs plugin?

stephenglass:

Fixes #30450 
Follow-up to PR #28625

This adds **optional** additional props `status`, `statusMessage`, `additionalInfo` to the overridable `NotFoundErrorPage` component. This was made to be a backwards compatible change. This PR fixes a few issues:

1. All error pages in Backstage framework now use the overridable error page component.
2. Usages of error page component throughout the framework are updated to use the correct status codes and messages in error pages. E.g. 404 for not found, 403 for forbidden. Currently, no matter the cause the error shows the same (a 404 error), even if something like the `RequirePermission` component was showing the error due to lack of permission which should really be a 403 forbidden.
3. TechDocs error page now uses the overridable component instead of the hard coded error page component. This means if you have a custom error page, TechDocs will now use that error page. Currently, since it's hard coded it always shows the default error page with no way to change it.

Example of custom error page using new status props:

```ts
const app = createApp({
  components: {
    NotFoundErrorPage: ({ status, statusMessage }) => (
      <>
        <div>This is my status: {status}</div>
        <div>This is my status message: {statusMessage}</div>
      </>
    )
  },
});
```


.changeset/calm-moments-sell.md
```diff
@@ -0,0 +1,26 @@
+---
+'@backstage/frontend-plugin-api': minor
+'@backstage/core-plugin-api': minor
+'@backstage/plugin-permission-react': minor
+'@backstage/plugin-app': minor
+'@backstage/app-defaults': minor
+'@backstage/core-app-api': minor
+'@backstage/plugin-scaffolder': minor
+'@backstage/plugin-techdocs': minor
+---
```
schultzp2020, line 6:
```suggestion
---
'@backstage/frontend-plugin-api': patch
'@backstage/core-plugin-api': minor
'@backstage/plugin-permission-react': patch
'@backstage/plugin-app': patch
'@backstage/app-defaults': minor
'@backstage/core-app-api': minor
'@backstage/plugin-scaffolder': minor
'@backstage/plugin-techdocs': minor
---
```

Please separate the `patch` packages into a separate changeset. We have a slightly different [versioning policy](https://github.com/backstage/backstage/blob/master/REVIEWING.md#reviewing-changeset-bump-levels) when the major version is less than one. 🙂

.changeset/calm-moments-sell.md
```diff
@@ -0,0 +1,26 @@
...
+
+Add optional props to the overridable error page component. TechDocs not found page now uses the overridable not found error page. Error pages now send correct status code and status message i.e. not found gives 404 and failed permission results in a 403 forbidden. When creating a custom error page you may now use additional props.
```
schultzp2020: Can you please separate the TechDocs changeset into a new one? 🙏

.changeset/calm-moments-sell.md
```diff
@@ -0,0 +1,26 @@
+---
+'@backstage/frontend-plugin-api': minor
+'@backstage/core-plugin-api': minor
+'@backstage/plugin-permission-react': minor
+'@backstage/plugin-app': minor
+'@backstage/app-defaults': minor
+'@backstage/core-app-api': minor
+'@backstage/plugin-scaffolder': minor
+'@backstage/plugin-techdocs': minor
```
byan1197: Hey! Similar to what @schultzp2020 was saying, this looks more like a patch for the techdocs related changes. Otherwise this looks a good change

.changeset/calm-moments-sell.md
stephenglass: Pushed a new commit resolving this 🙏

.changeset/calm-moments-sell.md
stephenglass: Pushed a new commit resolving this 🙏

stephenglass:
ci started failing due to unrelated

squid-ney (approved):

joanna-kli:
Hi, I'm currently waiting for these changes to be able to override the default "Not Found" error page in TechDocs. Do you have any idea when this PR is expected to be merged? Also, is there a planned date for the next release that would include it?

Thanks for the great work! This will be very helpful

benjdlambert:
@stephenglass sorry, there's some conflicts which have crept in here as we've been working on `componentRefs` in the new system. The types in `frontend-plugin-api` will also need updating here :pray:

stephenglass:
> @stephenglass sorry, there's some conflicts which have crept in here as we've been working on `componentRefs` in the new system. The types in `frontend-plugin-api` will also need updating here :pray:

@benjdlambert updated

Edit: need to fix api reports

stephenglass:
API reports fixed

packages/frontend-plugin-api/src/types.ts
```diff
@@ -14,16 +14,18 @@
...
 
 /** @public */
 export type ProgressProps = {};
 
 /** @public */
-export type NotFoundErrorPageProps = {
-  children?: ReactNode;
-};
+export type NotFoundErrorPageProps = PropsWithChildren<{
+  status?: string;
```
benjdlambert, line 25: This feels a bit strange to have for a component called `NotFoundErrorPage` as it should really always be 404.

I'm wondering if we should actually call this `ErrorPage` in the new frontend system instead as the swappable component. Kind of happy for this to be a bit weird for the old legacy system as that's going away and don't wanna make any breaking changes there.

/cc @freben @patriko thoughts?

stephenglass:
bumping to not get stale

stephenglass:
Not stale

plugins/techdocs/src/reader/components/TechDocsNotFound.tsx
```diff
@@ -48,7 +54,7 @@ export const TechDocsNotFound = ({ errorMessage }: Props) => {
   }
 
   return (
-    <ErrorPage
+    <NotFoundErrorPage
```
squid-ney, line 57: From a TechDocs perspective this change to use the `NotFoundErrorPage` swappable component instead of the hard coded core component looks good. If you want to separate this change into its own PR we can get that merged faster. 

Changes 1 & 2 will need some more discussion with the @backstage/maintainers .

packages/frontend-plugin-api/src/types.ts
freben: I think `NotFoundErrorPage` is distinct enough of a use case, that it could be its own thing. It kind of explicitly shouldn't have any error info or further deep details, and a very dedicated type of messaging/design.

An `ErrorPage` if it existed, sounds like it basically would be used by the wrapper that we put around pages right? If something goes catastrophically wrong in its rendering.

And then the third bit, more granular error panels that are used intentionally for explicitly caught problems.

freben: So in summary - in the new frontend system, having a dedicated NotFoundErrorPage (without any complex props) feels reasonable. And adding a separate ErrorPage with more props later on would also be OK.

stephenglass: Hi @freben  and @benjdlambert , would you be able to clarify if any changes requested for this PR to proceed?

packages/frontend-plugin-api/report.api.md
```diff
@@ -1497,9 +1498,11 @@ export const NotFoundErrorPage: {
 };
 
 // @public (undocumented)
-export type NotFoundErrorPageProps = {
-  children?: ReactNode;
-};
+export type NotFoundErrorPageProps = PropsWithChildren<{
+  status?: string;
+  statusMessage?: string;
+  additionalInfo?: string;
```
freben, line 1504: These props are my main gripe with the change.

The component is very explicitly named `NotFoundErrorPage` because it's a helpful page that people can put up when trying to access something that really does not exist. So it's a placeholder page so to speak, not meant to be richly informative. It's the app/plugin saying "there's no such thing and there's nothing to show here".

The old type [also has](https://github.com/backstage/backstage/blob/13507b42ddfdade3be798c77eb5ab8e3ffd765a1/packages/core-app-api/src/app/types.ts#L70) a `children` prop already. We could add that to the swappable component type too. Question is, will maybe that be enough? I can see the need for wanting to distinguish between "Entity not found" and "Some other thing not found" and giving more info. Then you can put that inside the page contents as children. What gets rendered is the mic drop thing, and then the children beside that. If we want to make the mic drop replaceable too, then that should not be a prop on the swappable component type (ie it should not be possible to make different in each rendered instance), but rather on the underlying react component (the operator who owns the backstage instance can replace the swappable component with something that renders the underlying react component with a different image, for ALL rendered instances).

I still think that that should be completely distinct from other errors. As in, the `NotFoundErrorPage` doesn't need to get richer and more featureful; instead, maybe there should be additional more specialised error pages if that's needed.

Just some loose examples:

* `NotFoundErrorPage`: basically no props
* `ForbiddenErrorPage`: basically no props, maybe a different wording and default image than the above
* `RequestErrorPage`: specialised for response errors, can take an actual Response object from `fetch` as a prop and unpack that to show rich info using something [like the `ResponseErrorPanel`](https://github.com/backstage/backstage/blob/13507b42ddfdade3be798c77eb5ab8e3ffd765a1/packages/core-components/src/components/ResponseErrorPanel/ResponseErrorPanel.tsx#L50)
* `GenericErrorPage`: could have several generic props

Again, just examples to illustrate. Not sure that it's a single generic error page that we need.

stephenglass: Thanks for the additional comments! I'll think about this a bit more and update or create new PR maybe this week.

PabloPikanto:
Would be nice if this could be merged

awanlin:
Hi @stephenglass, any chance you can pick this up again and resolve the remaining comments?

You'll likely want to rebase as well 👍

modethirteen:
@stephenglass @awanlin @freben This is a such a significant UX improvement than an unbranded, dead-end 404 page, especially in TechDocs, I'd happily step in and close the gap if this is stalling out.

I can't overstate how important it is to heal broken knowledge gaps in documentation. At my org, we're going to insert search results inferred from the URL path to give users a chance at navigating to what they were looking for. Candidly most of our internal developers don't really know or care much about Backstage internal technicalities, so the addition of the `ERROR 404 ... Note that techdocs.builder is not set to 'local' in your config ...` message is also just confusing and creates more questions. It's important that the 404 experience just gets the back on the path to find the right information - so consistent, useful 404 pages across the portal experience are a necessity.

Let me know Stephen if you need support or someone to apply the maintainers' feedback and push this forward.

Issue #29674, opened 2025-04-19
🚀 Feature: New DevTools tab using the upcoming Scheduler API


awanlin:
### 🔖 Feature description

Once the [Scheduler API](https://github.com/backstage/backstage/pull/29553) is added we should also add a tab to the DevTools plugin displaying this information! 🚀 

### 🎤 Context

It's been a long running ask to see all the scheduled tasks and often the [idea is suggested to add this to DevTools](https://github.com/backstage/backstage/issues/24286) as it's a good place for items like this. Now what we will have the [Scheduler API](https://github.com/backstage/backstage/pull/29553) it looks like we can add this long requested feature.

### ✌️ Possible Implementation

Leverage the [Scheduler API](https://github.com/backstage/backstage/pull/29553) to show all the scheduled tasks for a Backstage instance as a tab in the DevTools plugin.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

aramissennyeydd:
Work started in #28726 :)

GabDug:
Left a comment (on the wrong issue) in https://github.com/backstage/backstage/issues/30049#issuecomment-3035548890, but I've made an internal plugin regarding this issue, with support for task-level permissions. 

I'll try to release the source next week so it can be contributed to backstage or backstage-community, or used as an inspiration.

awanlin:
Wouldn't say you posted on the wrong issue @GabDug, the permissions are something that needs to be sorted, just felt the further comments might fit here better. 

This is for sure something that the Community has asked for and built custom internal solutions for. Getting something open source would be greatly appreciated. The approach I would take is just to add it directly to DevTools. BUT if your experience says people outside of Backstage Admins find this helpful then I might take the approach of the Unprocessed entities plugin which can run standalone as well as as a tab in DevTools.

Also, this seems like a better fit here in the main Backstage repo then as a Community Plugin IMO. 👍

Parsifal-M:
[...]

PR #30533, opened 2025-07-15
fix: Add protection for scheduler tasks REST API

### LINKED ISSUE #30049: Add permissions for SchedulerService REST API
### 📜 Issue Labels

### 🔎 Search Terms

```plain
permissions, scheduler, tasks
```

### 🗃️ Project Area

Core Framework

### 🔖 Need

1.39.0 introduced a new REST API to list and trigger scheduled tasks (https://backstage.io/docs/backend-system/core-services/scheduler/#rest-api), but it is not possible to limit who can read these tasks or trigger them, allowing anyone to consume resources by triggering tasks. 

### 📝 Proposal

Add at least one permission that can be used to control who can trigger tasks. Read permission would be nice too, while not so critical.

### 🔄 Alternatives

_No response_

Undecided

stephenglass:

Fixes #30049 
PR as a follow-up to my comment in https://github.com/backstage/backstage/issues/30049#issuecomment-3070553507

Plugs up a security gap in the scheduler tasks REST API since the endpoints are unprotected from any permission checks. That means any user with a valid access token can abuse your app by triggering potentially performance exhaustive or high rate limit tasks.

This PR makes it so the APIs can only be triggered using a valid service or static token. My thoughts are the chance where a user would need to trigger these endpoints from their own user credentials is low. And if a task scheduler UI is created (like in #29674, #30544) then a corresponding backend plugin would be created to add permission checks and proxy the task trigger API requests via service-to-service auth.

I'd be happy to discuss or make changes as needed. 🙏

### DIFF
--- a/packages/backend-defaults/src/entrypoints/scheduler/lib/PluginTaskSchedulerImpl.test.ts
+++ b/packages/backend-defaults/src/entrypoints/scheduler/lib/PluginTaskSchedulerImpl.test.ts
@@ -448,7 +448,6 @@ describe('PluginTaskManagerImpl', () => {
     it.each(databases.eachSupportedId())(
       'POST /.backstage/scheduler/v1/tasks/:id/trigger is protected, %p',
       async databaseId => {
-        console.log('database id', databaseId);
         const { manager } = await init(
           databaseId,
           mockServices.auth(),


stephenglass:
PR is ready to review 🎉

packages/backend-defaults/src/entrypoints/scheduler/lib/PluginTaskSchedulerImpl.test.ts
```diff
@@ -405,4 +417,57 @@ describe('PluginTaskManagerImpl', () => {
...
+        expect(res.body.error?.message).toMatch(
+          /must be invoked by a service/i,
+        );
+      },
+    );
+
+    it.each(databases.eachSupportedId())(
+      'POST /.backstage/scheduler/v1/tasks/:id/trigger is protected, %p',
+      async databaseId => {
+        console.log('database id', databaseId);
```
benjdlambert:
```suggestion

```

packages/backend-defaults/src/entrypoints/scheduler/lib/PluginTaskSchedulerImpl.ts
```diff
@@ -155,7 +164,20 @@ export class PluginTaskSchedulerImpl implements SchedulerService {
   getRouter(): express.Router {
     const router = Router();
 
-    router.get('/.backstage/scheduler/v1/tasks', async (_, res) => {
+    router.get('/.backstage/scheduler/v1/tasks', async (req, res) => {
+      const credentials = await this.httpAuth.credentials(req);
+      if (this.auth.isPrincipal(credentials, 'user')) {
+        if (!credentials.principal.actor) {
```
benjdlambert, line 170: Hmm not sure if this check is a bit too strict. It's almost like an admin like permissions which we don't really have persona for, maybe a permission rule or something might be useful here. 

Just wondering if there's consumers that have already built interfaces around this in the frontend to retrigger jobs using this API and a backstageToken from the UI. This would then break. They'd have to create a backend plugin to do the call.

Thoughts @Rugvip?

stephenglass: Just a thought, with these changes and in the case someone created a UI wrapper with user token auth, a simpler upgrade may be to use the proxy plugin as opposed to creating a backend plugin.

stephenglass:
bumping to not get stale

stephenglass:
Not stale

freben:
Added @vinzscam who knows permissions best.

I would agree that this fits into the (not yet nicely solved) admin/root use case, because it should definitely be handled through the permissions system but the question is how.

Pure service tokens can already do "everything" unless bounded by explicit access controls in their app-config options. So that's a given.

If we made a new permission for this action, it would by default reject users, but also services who act OBO users. Right?

And would the permission be a basic permission, or a resource permission - and if the latter, for what resource? Would we control things on a task granularity or on a plugin granularity? I'd guess the latter. So we can say for example "members of group X can manage catalog tasks, but not scaffolder tasks" or so.

vinzscam:
> If we made a new permission for this action, it would by default reject users, but also services who act OBO users. Right?

By convention, unknown permissions are usually allowed, so it won't be a breaking change.

But yeah, I 100% agree with what @freben mentioned. We need to understand and plan how we want to manage permissions for this feature (should we target plugins, task IDs, and/or other properties?).

freben:
@vinzscam Do you have any suggestions for how to best proceed?

stephenglass:
not stale

vinzscam (changes requested):
> @vinzscam Do you have any suggestions for how to best proceed?

I think we could start brainstorming about adding a new `tasks.execute` resource permission, with rules for targeting specific plugins and perhaps, task ids. This way, adopters could opt in to allow only specific tasks to be executed.

PR #30539, opened 2025-07-15
canon.backstage.io -> ui.backstage.io

Rugvip:
Already updated in the docs repo, but this makes sure we don't override that on the next build


cdedreuille (approved):
Thanks 🙏

PR #30541, opened 2025-07-15
feat(techdocs): POC livereload for techdocs-cli serve

### LINKED ISSUE #30514: Techdocs: Support autoreload when working with techdocs-cli serve
### 📜 Issue Labels

### 🔎 Search Terms

```plain
autoreload, mkdocs serve, livereload
```

### 🗃️ Project Area

TechDocs

## Problem
 
As we adopt Techdocs in our org, we add moreand more customization as Techdocs addons or theming, that require the use of `techdocs-cli`, and a custom embedded app.

We push our users to use the techdocs-cli, with our customization, so they can have a 1/1 match with what they will see on our Backstage instance (themes, diagrams...)

Even though the preview is closer to reality than `mkdocs serve`, and doesn't require a Python setup, **techdocs-cli does not support auto reload of the docs**.

This is a feature our users have been accustomed to, when using mkdocs or other doc frameworks; and that would lead to less friction when adopting Techdocs.

## Success criteria

- (MUST) Automatically reload the page when doc changes
- (MUST) As it's a QOL feature, if unavailable, it should not break `techdocs-cli serve`
- (SHOULD) Preserve scroll position
- (SHOULD) No visual flashes on reload

### 📝 Proposal

Add support for `mkdocs serve` autoreload when using `techdocs-cli serve`

## Mkdocs autoreload internal

Mkdocs [already implements autoreload](https://github.com/mkdocs/mkdocs/blob/4c7404485f988f409ccaf42fefe705222ff5965a/mkdocs/livereload/__init__.py), using a long-polling call from the frontend, that resolves whenever a browser reload is needed.

The script they inject contains an epoch of the current page built-in, required to make a call and compare with responses.

The logic hasn't changed in a few years, and works well, so we can probably safely rely on it.

## POC research

- We can't use the <script> they inject directly, as it's removed by DomPurify (and we probably want to keep it that way!)
- To use their livereload route, we would need an epoch of the build-time. 
  - Unfortunately, they don't use UNIX epoch, but python's time.monotonic, which returns a precise monotonic clock, from the system.
  - Unfortunately, the results from python.monotonic and Node's process. hrtime differ (at least on MacOS)
  - => We can't craft the request ourselves, we need the epoch that mkdocs generated.
- We can easily extract that from the HTML response, before it's sanitized, either from the embedded app techdocs storage API, or from techdocs-cli httpserver proxy.

## POC 

Note: To avoiding forking/patching too much, this was made with a constraint not to patch the main `techdocs` plugin, but only patch `techdocs-cli` and use a custom Techdocs Addon. I will create a MR for this POC in the coming days.

- Patch `techdocs-cli` httpServer:
  - To extract the livereload epoch from the HTML <script> tag, and put it in a custom-element which we can exclude from DomPurify.
  - To add a `.livereload/` route, proxying to mkdocs's `/livereload` with valid cross-origin headers.
- Add a new techdocs addon in the embedeed-app that will replicate original mkdocs livereload feature:
  - Read the live-reload epoch custom component data
  - Make a request to the `techdocs-cli` `.livereload/` route, and compare the epochs on response: reload if the epoch response is bigger.

## Further work

- The reload is currently made with `location.reload`, instead of only re-rendering the document shadow dom.
  - This is not something we can currently do from a TechdocsAddon, so API changes would be needed, or this future would need to be included in `techdocs` plugin.
- Unlike original `mkdocs-serve`, we loose the scroll position.

### 🔄 Alternatives

Instead of piggybacking on mkdocs `/livereload` route, we could also watch the documents ourselves in `techdocs-cli`, either by watching the file system changes, and/or parsing mkdocs logs.

While we wouldn't have to extract the livereload epoch from mkdocs HTML responses, we would still need to create a techdocs-cli endpoint to watch the file changes and add a reload feature in the frontend app. To avoid reloading before mkdocs has finished building, we would probably need to watch `mkdocs` logs to detect reloads.

GabDug:

Fix #30514 

### DIFF
--- a/packages/techdocs-cli-embedded-app/app-config.yaml
+++ b/packages/techdocs-cli-embedded-app/app-config.yaml
@@ -7,3 +7,8 @@ backend:
 
 techdocs:
   builder: 'external'
+  sanitizer:
+    # Allow live reload locally
+    # XXX(GabDug): move out of this config, so we don't conflict with config in custom embedded apps
+    allowedCustomElementTagNameRegExp: 'live-reload'
+    allowedCustomElementAttributeNameRegExp: 'live-reload-epoch|live-reload-request-id'
--- a/packages/techdocs-cli-embedded-app/src/App.tsx
+++ b/packages/techdocs-cli-embedded-app/src/App.tsx
@@ -36,6 +36,7 @@ import * as plugins from './plugins';
 import { configLoader } from './config';
 import { Root } from './components/Root';
 import { techDocsPage, TechDocsThemeToggle } from './components/TechDocsPage';
+import { TechDocsLiveReload } from './LiveReloadAddon';
 
 const app = createApp({
   apis,
@@ -51,6 +52,14 @@ const ThemeToggleAddon = techdocsPlugin.provide(
     name: 'ThemeToggleAddon',
     component: TechDocsThemeToggle,
     location: TechDocsAddonLocations.Header,
+  }),
+);
+
+const LiveReloadAddon = techdocsPlugin.provide(
+  createTechDocsAddonExtension({
+    name: 'LiveReloadAddon',
+    component: TechDocsLiveReload,
+    location: TechDocsAddonLocations.Content,
   }),
 );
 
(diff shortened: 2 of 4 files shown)


GabDug:
/unstale

Waiting for feedback here or on #30514

Issue #28592, opened 2025-01-22
🐛 Bug Report: Issue with Owned filter in Catalog


sydris:
### 📜 Description

We are seeing an issue with the Catalog where it is loading the owned view, as we want it to, and then reloading the page with no user intervention to show everything.

### 👍 Expected behavior

We expect the catalog to go to and stay on the owned view.

### 👎 Actual Behavior with Screenshots

The catalog appears to be loading multiple times. First to all, then the owned view (what we expect), then back to all. Also, after the page loads, hitting the "Owned" button on the side nav has no effect.
This started happening right after we updated to Backstage version 1.34.2, but didn't see anything in the release notes that would've caused it.

When hitting /catalog directly:

https://github.com/user-attachments/assets/9cf805e5-27cc-44da-83e8-fb1ce8cdb662

Hitting catalog through the sidebar: 

https://github.com/user-attachments/assets/b367c994-1262-48f3-b848-9e3de30fdeb0

### 👟 Reproduction steps

In App:
```
   <Route path="/catalog" element={<CatalogIndexPage />}>
      <CustomCatalogIndexPage initiallySelectedFilter="owned" />
    </Route>
```
Note: Also tested with below code, and issue was still occurring, so don't believe it has anything to do with the custom catalog.
```
   <Route path="/catalog" element={<CatalogIndexPage />}>
      <CatalogIndexPage initiallySelectedFilter="owned" />
    </Route>
```

In Root:
```
        <SidebarItem icon={CategoryIcon} text="Catalog">
          {' '}
          <SidebarSubmenu title="Catalog">
            <SidebarSubmenuItem
              title="Components"
              to="catalog?filters[kind]=component&filters[user]=owned"
            />
            <SidebarSubmenuItem
              title="Cloud Accounts"
              to="catalog?filters[kind]=resource&filters[type]=account&filters[user]=owned"
            />
            <SidebarSubmenuItem
              title="Templates"
              to="catalog?filters[kind]=template&filters[user]=owned"
            />
            <SidebarSubmenuItem
              title="APIs"
              to="catalog?filters[kind]=api&filters[user]=owned"
            />
            <SidebarSubmenuItem
              title="Owning Teams"
              to="catalog?filters[kind]=group"
            />
            <SidebarSubmenuItem title="Users" to="catalog?filters[kind]=user" />
            <SidebarSubmenuItem title="Documentation" to="docs" />
          </SidebarSubmenu>{' '}
        </SidebarItem>
```
[...]

PR #30545, opened 2025-07-15
Entity filter race condition

raihle:

This is a fix for the filter "rubber-banding" described in https://github.com/backstage/backstage/issues/28592

- [ ] ~~Added or updated documentation~~
- [ ] ~~Screenshots attached (for UI changes)~~

### DIFF
--- a/plugins/catalog-react/src/testUtils/MockPromise.ts
+++ b/plugins/catalog-react/src/testUtils/MockPromise.ts
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2025 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


plugins/catalog-react/src/testUtils/MockPromise.ts
```diff
@@ -0,0 +1,47 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+type ResolveFn<T> = (value: T | PromiseLike<T>) => void;
+type RejectFn = (reason?: any) => void;
+type Executor<T> = (resolve: ResolveFn<T>, reject: RejectFn) => void;
+
+export class MockPromise<T> extends Promise<T> {
```
schultzp2020: Do we want this mock in a shared library?

raihle: My usual principle is not to set something like that up until you have at least three use cases, but maybe there is already something like that in the repo?

freben: Note that we already have `createDeferred` in `@backstage/types`. You can remove these.

freben (commented):
Sorry, reopened. One immediate comment came to mind.

PR #30546, opened 2025-07-15
catalog-backend: switch flags to new default for 3.0

Rugvip:

With catalog backend shipping 3.0 in the next release we thought it was time to include a few more changed to the default configuration:

- `disableRelationsCompatibility: true`: Relations compatibility is now disabled by default, but it can be re-enabled with the new `enableRelationsCompatibility` flag. Relations compatibility makes sure there's both `target` and `targetRef` in all relation objects, which significantly reduces overall performance of the catalog. With compatibility disabled, only the `targetRef` will be present. Only bespoke external consumers of the catalog could be affected by this change, as the `targetRef` has been in use for many years at this point.
- `orphanStrategy: 'delete'`: This switches the default orphan strategy to eagerly delete entities that are no longer referenced within the catalog.
- `orphanProviderStrategy: 'delete'`: This switches the default provider installation strategy to remove and entities that have been omitted by providers that are no longer present.
- `stitichingStrategy: { mode: 'deferred' }`: Switches the stitching strategy to happen out-of-band of the main processing loop, instead being run in a background task. This makes stitching a lot more performant in large catalogs.

These are all flags that have been around for a while and are often cited as fixes to common catalog problems. We feel it's worth switching over all of these defaults to provide a better out-of-the-box experience for the catalog. These new defaults should behave well with existing catalogs as well, but there may be some edge cases that we haven't thought of, and will be highlighting this change in the release notes.

These are in addition to the `catalog.useUrlReadersSearch` switch in https://github.com/backstage/backstage/pull/29788

### DIFF
--- a/.changeset/clever-windows-judge.md
+++ b/.changeset/clever-windows-judge.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': major
+---
+
+**BREAKING**: The default `catalog.stitchingStrategy` has been switched to `{ mode: 'deferred' }`.
--- a/plugins/catalog-backend/config.d.ts
+++ b/plugins/catalog-backend/config.d.ts
@@ -172,7 +172,7 @@ export interface Config {
     orphanProviderStrategy?: 'keep' | 'delete';
 
     /**
-     * The strategy to use when stitching together the final entities.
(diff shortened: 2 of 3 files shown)


benjdlambert:
:ship:

stephenglass:
Great changes here, awesome!!

PR #30583, opened 2025-07-17
feat(sanitizer): add support for allowlisting additional URI protocols

nickhudkins:
## Allow additional protocols to be added to DOMPurify config

Past discussions about configurability of DOMPurify within TechDocs indicate that a targeted, restrictive approach to allowing DOMPurify configuration is preferred over full user-land flexibility (hooray for secure by default!). As such, this PR introduces a constrained means by which to allow additional protocols to be permitted.

#### Motivation
Question asked in Discord:
https://discord.com/channels/687207715902193673/714754240933003266/1394274725466865695

- [N/A] Screenshots attached (for UI changes)

### DIFF
--- a/plugins/techdocs/src/reader/transformers/html/transformer.ts
+++ b/plugins/techdocs/src/reader/transformers/html/transformer.ts
@@ -78,6 +78,35 @@ export const useSanitizerTransformer = (): Transformer => {
       const attributeNameCheck = config?.getOptionalString(
         'allowedCustomElementAttributeNameRegExp',
       );
+      const additionalAllowedURIProtocols =
+        config?.getOptionalStringArray('additionalAllowedURIProtocols') || [];
+
+      // Define allowed URI protocols, including any additional ones from the config.
+      // The default protocols are based on the DOMPurify defaults.
+      const allowedURIProtocols = [
+        'callto',
+        'cid',
+        'ftp',
+        'ftps',
+        'http',
+        'https',
+        'mailto',
+        'matrix',
+        'sms',
+        'tel',
+        'xmpp',
+        ...additionalAllowedURIProtocols,
+      ].filter(Boolean);
+
+      const allowedURIRegExp = new RegExp(
+        // This regex is not exposed by DOMPurify, so we need to define it ourselves.
+        // It is possible for this to drift from the default in future versions of DOMPurify.
+        // See: https://raw.githubusercontent.com/cure53/DOMPurify/master/src/regexp.ts
+        `^(?:${allowedURIProtocols.join(
+          '|',
+        )}:|[^a-z]|[a-z+.-]+(?:[^a-z+.\\-:]|$))`,
+        'i',
+      );
 
       // using outerHTML as we want to preserve the html tag attributes (lang)
       return DOMPurify.sanitize(dom.outerHTML, {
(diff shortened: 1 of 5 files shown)


awanlin (approved):
Thanks for the contribution @nickhudkins, this makes a lot of sense to me 👍 

I've also reached out to the TechDocs maintainers to looks this over and give their approval.

plugins/techdocs/src/reader/transformers/html/transformer.ts
```diff
@@ -78,6 +78,21 @@ export const useSanitizerTransformer = (): Transformer => {
...
+
+      // This regex is not exposed by DOMPurify, so we need to define it ourselves.
+      // It is possible for this to drift from the default in future versions of DOMPurify.
+      // See: https://raw.githubusercontent.com/cure53/DOMPurify/master/src/regexp.ts
+      const allowedURIRegExp =
+        additionalAllowedURIProtocols.length > 0
+          ? new RegExp(
+              `^(?:(?:(?:f|ht)tps?|${additionalAllowedURIProtocols.join(
+                '|',
+              )}|mailto|tel|callto|sms|cid|xmpp|xxx):|[^a-z]|[a-z+.-]+(?:[^a-z+.\\-:]|$))`,
```
alexlorenzi: Hey overall this is great. One small nitpick... Wondering if it would be better to pull out the default protocols into a list and merge them with the custom ones and then do the `.join('|')` where rather than have some written in the regex and some in the list. Might make it easier to see what's been included... For example, not sure what the `xxx` protocol is, but assume it's meant to be the `matrix` from the original DomPurify regex.

nickhudkins: I went back and forth on this myself. Happy to extract the list!

nickhudkins: Holy copy-pasta-autocomplete-trashfire, good looking out on the 'xxx' protocol. 

As for pulling the other protocols into a list, I'm inclined to leave any non simple strings as part of the template. 

Alternatively, I could expand them into discrete values like ftp,ftps,http,https for the first piece of the regex.

The other subtlety that perhaps should be changed is the conditional nature of how the override occurs. As it stands in this PR, we default to whatever DOMPurify if additional protocols are not provided.  This could introduce strange behavior between folks who extend the list of protocols and folks who do not, should the DOMPurify regex change upstream. 

I'd like to propose that we go ahead and *always* override, at which point, the URI pattern becomes ours to manage rather than a shared / mix source of truth.

plugins/techdocs/src/reader/transformers/html/transformer.ts
```diff
@@ -78,6 +78,35 @@ export const useSanitizerTransformer = (): Transformer => {
...
+      // The default protocols are based on the DOMPurify defaults.
+      const allowedURIProtocols = [
+        'callto',
+        'cid',
+        'ftp',
+        'ftps',
+        'http',
+        'https',
+        'mailto',
+        'matrix',
```
byan1197, line 94: ~~To keep it consistent, can we keep this matrix protocol out just to match the [defaults that DOMPurify](https://www.npmjs.com/package/dompurify/v/2.0.12#can-i-configure-dompurify) comes with?~~ Looks like I was looking at an older version of the documentation
Changes looks great!

squid-ney (approved):

PR #30591, opened 2025-07-18
UI - update styling of Tooltip

mtlewis:

Updates styling of ui Tooltip elements. A couple of example screenshots attached.

CleanShot 2025-07-19 at 21 27 29@2x

CleanShot 2025-07-19 at 21 27 17@2x


packages/ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -40,12 +43,20 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
     return (
       <AriaTooltip
         className={clsx(classNames.tooltip, className)}
+        offset={16}
```
mtlewis: Using the offset property with a fixed value here since the height of the tooltip is fixed. The gap between the tooltip arrow and the target element is handled by the css margins, which use theme variables.

packages/ui/report.api.md
```diff
@@ -591,6 +591,8 @@ export const componentDefinitions: {
     readonly classNames: {
       readonly tooltip: 'bui-Tooltip';
       readonly arrow: 'bui-TooltipArrow';
+      readonly arrowFill: 'bui-TooltipArrowFill';
```
mtlewis: Worth discussing a bit whether these classes should be hidden from the public API.

packages/ui/src/components/Tooltip/Tooltip.tsx
cdedreuille: I changed the value to use a default of `12px` that you can override if needed.

packages/ui/report.api.md
cdedreuille: You're right it might be better at this stage to remove the need to support these class names. I removed them for now and set the CSS to use `svg path:nth-child(1) ` and `svg path:nth-child(2)` instead.

cdedreuille (approved):
Thanks a lot @mtlewis 🙏 I did a few design tweaks. Let me know if that work with you and we can merge it.

PR #30592, opened 2025-07-18
BUI - New Text component

cdedreuille:

New Text component

We are consolidating the way we treat our text. We used to have two components for that: `Text` and `Heading` but these caused some confusions, specially with the `as` props as some texts wanted to be used as titles.

Our solution was to consolidate both `Text` and `Heading` into a new `Text` component with a simplified structure. We now have 4 title variants and 4 body variants that should give enough room and context on what to use.

We also aligned the `Link` component to follow the same structure with a special visual treatment for it.

Existing Text + Heading


.changeset/tender-crabs-stay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': minor
+---
+
+**Breaking** We are upgrading our `Text` component to support all font sizes making the `Heading` component redundant. The new `Text` component introduces 4 sizes for title and 4 sizes for body text. All of these work in multiple colors and font weights. We improved the `as` prop to include all possible values. The component also match with the `Link` component.
```
Rugvip: > The component also match with the `Link` component.

This I didn't get until looking at the code. Perhaps "The `Link` component has also been updated to match the new `Text` component" or such?

packages/ui/report.api.md
```diff
@@ -1769,35 +1698,48 @@ export interface TextFieldProps
...
-  style?: React.CSSProperties;
 };
 
 // @public (undocumented)
-export type TextProps<T extends ElementType = 'p'> = TextOwnProps &
+export type TextProps<T extends ElementType = 'span'> = TextOwnProps &
   Omit<ComponentPropsWithRef<T>, keyof TextOwnProps>;
 
+// @public (undocumented)
+export type TextVariants =
```
Rugvip, line 1730: Thoughts on this vs `variant: title | body` + `size: x-small | small | medium | large`?

cdedreuille: Interesting. This is a nice pattern for sure to have the same values for both `title` and `body` but we might bring a bigger size later on or another variant for uppercase small types that would break this pattern. Perhaps keeping it all on the same variant is better future proof. WDYT?

.changeset/tender-crabs-stay.md
cdedreuille: You're right. I updated it with your input.

Issue #29891, opened 2025-05-09
Move to the LDAPts library for LDAP plugin


JessicaJHee:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
"LDAP", "LDAPts", "LDAPjs", "LDAP plugin"
```

### 🗃️ Project Area

Catalog

### 🔖 Need

The [LDAPjs library](https://www.npmjs.com/package/ldapjs) has been decommissioned as of May 15, 2024, while the [LDAPts library](https://github.com/ldapts/ldapts) is actively maintained.

There have also been discussions in the past to move towards LDAPts instead [here](https://github.com/backstage/backstage/issues/6606#issuecomment-887292909).

### 📝 Proposal

We would need to update the LDAP client wrapper [here](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend-module-ldap/src/ldap/client.ts). With the LDAPts library, a custom wrapper may no longer be necessary.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Yes, and I have enough information to get started

benjdlambert:
Hey :wave:

Thanks for raising this, I actually saw this library a bit back and thought that we should track this work in an issue!

Think that this makes sense :tada:

MihaiTabara:
Still relevant - unstale.

ganievs:
Hi there! I think the following PR is ready for review.

Some key points to highlight:
- So because LDAPS doesn't expose the client host outside, I use some indirect checks in 
the `isGoogleLDAP(root)` method to evaluate if the provider is Google.

- It also has some configuration changes: `attributeValues: true` instead of `typesOnly: false`

ganievs:
Up

benjdlambert:
Closing as this has been merged

PR #30594, opened 2025-07-18
[WIP] Feature/move to ldapts

ganievs:

related to #29891


plugins/catalog-backend-module-ldap/src/ldap/client.ts
```diff
@@ -58,37 +57,26 @@ export class LdapClient {
...
+    const ldapClient = new LdapClient(client, logger);
 
-    if (!bind) {
-      return new LdapClient(client, logger);
+    if (bind) {
+      try {
+        await client.bind(bind.dn, bind.secret);
+      } catch (error) {
+        await client.unbind();
+        throw error(`LDAP bind failed for ${bind.dn}, ${error}`);
```
JessicaJHee:
```suggestion
        throw new ForwardedError(`LDAP bind failed for ${bind.dn}, ${error}`, error);
```

Fixes error I ran into: `error is not a function`

plugins/catalog-backend-module-ldap/src/ldap/client.ts
```diff
@@ -261,18 +142,73 @@ export class LdapClient {
...
+
+    if (!hasGoogleRootDSEPattern) {
+      return false;
+    }
+
+    try {
+      const schemaHasGoogleAttributes = this.checkGoogleSchema(rootDSE);
+      return schemaHasGoogleAttributes;
+    } catch (error) {
+      throw error('Schema check failed:', error);
```
JessicaJHee:
```suggestion
      throw new ForwardedError('Schema check failed', error);
```

JessicaJHee (commented):
Thank you for this contribution! 

Everything works as expected when I tested with ldaps connection, but I ran into `read ECONNRESET` error while trying to bind to ldap (on port 389). The exact setup worked with the current LDAP plugin. I'm wondering if it works on your end?

ganievs:
> Thank you for this contribution!
> 
> Everything works as expected when I tested with ldaps connection, but I ran into `read ECONNRESET` error while trying to bind to ldap (on port 389). The exact setup worked with the current LDAP plugin. I'm wondering if it works on your end?

Hello @JessicaJHee! Thanks for participation! 
I've tested on real ldap server only via ldaps (636 port). So I'll check on plain ldap too.

ganievs:
> > Thank you for this contribution!
> > Everything works as expected when I tested with ldaps connection, but I ran into `read ECONNRESET` error while trying to bind to ldap (on port 389). The exact setup worked with the current LDAP plugin. I'm wondering if it works on your end?
> 
> Hello @JessicaJHee! Thanks for participation! I've tested on real ldap server only via ldaps (636 port). So I'll check on plain ldap too.

Сhecked on plain ldap, looks like everything works.
@JessicaJHee what is your ldap vendor?

JessicaJHee:
@ganievs Thank you for confirming! I'm using Active Directory. I'll debug and see if it's just a configuration issue on my end.

plugins/catalog-backend-module-ldap/src/ldap/client.ts
```diff
@@ -58,37 +57,29 @@ export class LdapClient {
       });
     }
 
-    const client = ldap.createClient({
+    const client = new Client({
       url: target,
       tlsOptions: {
         secureContext,
         rejectUnauthorized: tls?.rejectUnauthorized,
       },
```
JessicaJHee: After some debugging, I realized that plain LDAP wasn't working on my end because of [this check](https://github.com/ldapts/ldapts/blob/f117fb0aec0a3ce62f2e10ad7690b7fa41bbae2c/src/Client.ts#L182).


That expression evaluates to true whenever `tlsOptions` is an object, even if all its properties are undefined:
```
{
  secureContext: undefined,
  rejectUnauthorized: undefined,
}
```
As a result, the client behaves as if TLS options were provided and the connection is expected to be secure. I believe in this case we should instead pass undefined when no TLS configs are present.

ganievs: Hi @JessicaJHee! I applied a workaround not to pass the tlsOptions object if both options are undefined. Could you confirm if this solved the issue?

I also created an [issue](https://github.com/ldapts/ldapts/issues/262) in the ldapts repo.

JessicaJHee: @ganievs Thanks so much for addressing this and creating the issue! I tested again, but the same issue persists :(

 `Object.keys(tlsOptions).length > 0` evaluates to true since it's checking for the length of the keys in the object and `tlsOptions` is created with keys even when all values are undefined. I think in this case we can do something like this that checks for any defined values: 

```
...(Object.values(tlsOptions).some(v => v !== undefined) ? { tlsOptions } : undefined)
```

plugins/catalog-backend-module-ldap/src/ldap/read.ts
```diff
@@ -263,16 +264,12 @@ export async function readLdapGroups(
...
-        return;
-      }
-
+    const searchResult = await client.search(dn, options);
+    for (const entry of searchResult.searchEntries) {
       const entity = await transformer(vendor, cfg, entry);
 
       if (!entity) {
-        return;
+        continue;
```
vinzscam, line 272: was this a bug?

.changeset/heavy-cars-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-ldap': major
+---
+
+Moved from ldapjs to ldapts
```
vinzscam: so that it fixes the vale issues

```suggestion
Moved from `ldapjs` dependency to `ldapts`
```

vinzscam (changes requested):
thanks @ganievs and @JessicaJHee for your effort on this 🙏

plugins/catalog-backend-module-ldap/src/ldap/read.ts
ganievs: No. Original code uses `return` in callback -> exits only that callback iteration and continues to the next entity in the stream.
The code in PR uses `continue` in loop -> skips to next loop iteration

plugins/catalog-backend-module-ldap/src/ldap/client.ts
ganievs: > @ganievs Thanks so much for addressing this and creating the issue! I tested again, but the same issue persists :(
> 
> `Object.keys(tlsOptions).length > 0` evaluates to true since it's checking for the length of the keys in the object and `tlsOptions` is created with keys even when all values are undefined. I think in this case we can do something like this that checks for any defined values:
> 
> ```
> ...(Object.values(tlsOptions).some(v => v !== undefined) ? { tlsOptions } : undefined)
> ```
Fixed with your suggestion - `Object.values(tlsOptions).some(v => v !== undefined)`. Сould you confirm if this solved the problem? Thanks!

JessicaJHee: It works perfectly now, thanks for addressing this!

.changeset/heavy-cars-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-ldap': major
```
benjdlambert:
```suggestion
'@backstage/plugin-catalog-backend-module-ldap': minor
```

This package is not yet 1.x so any breaking changes would mean that this is a `minor` change

plugins/catalog-backend-module-ldap/report.api.md
```diff
@@ -36,14 +37,14 @@ export default catalogModuleLdapOrgEntityProvider;
 export function defaultGroupTransformer(
   vendor: LdapVendor,
   config: GroupConfig,
-  entry: SearchEntry,
+  entry: Entry,
```
benjdlambert, line 40: How similar are these types? What's the breakages if any?

I think we should call out any steps that users are required to do in the changeset also, and because both these default functions and the ability to provide your own `group/userTransformer` there could be breakages.

plugins/catalog-backend-module-ldap/report.api.md
```diff
@@ -92,14 +93,9 @@ export class LdapClient {
     bind?: BindConfig,
     tls?: TLSConfig,
   ): Promise<LdapClient>;
-  getRootDSE(): Promise<SearchEntry | undefined>;
+  getRootDSE(): Promise<Entry | undefined>;
   getVendor(): Promise<LdapVendor>;
-  search(dn: string, options: SearchOptions): Promise<SearchEntry[]>;
-  searchStreaming(
```
benjdlambert, line 98: This also probably needs adding to the changeset that this has been removed, and what to migrate to instead

benjdlambert (changes requested):
Hey :wave: thanks for this!

Thinking that we can't really get away from some of the breaking changes here so want to add a lot more information to the changeset to make it easier for end users to understand what's required of them when they bump to the latest version.

plugins/catalog-backend-module-ldap/report.api.md
ganievs: The `searchStreaming` method was needed because `ldapjs` is event-based and we accumulated all entries from events into an array before processing. With ldapts, which returns all results directly, this intermediate accumulation step is no longer needed.
The method is internal (not exported in the public API) and only used within this module. The behavior of search operations hasn't changed at all.
Since this is an internal method and the behavior remains the same, no migration guide is needed for users.

plugins/catalog-backend-module-ldap/report.api.md
ganievs: Agree, will add it to the changeset

ganievs: I update the changeset. Does it enough or something needs to be changed

ganievs:
Up

plugins/catalog-backend-module-ldap/report.api.md
benjdlambert: This most definitely is in the public API, as it's on the LDAP Client. 

`import { LdapClient } from '@backstage/plugin-catalog-backend-module-ldap'` is part of the public API for this package.

I think it's fine to call out that it's been removed and users should move to just using `search` instead.

benjdlambert (changes requested):

plugins/catalog-backend-module-ldap/report.api.md
ganievs: Done

benjdlambert:
@ganievs sorry, there's a small conflict that's crept in here :pray:

ganievs:
> @ganievs sorry, there's a small conflict that's crept in here 🙏
Done 🙌

Issue #30530, opened 2025-07-15
create-github-app does not enable adequate permissions


evanshortiss:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
create-github-app admin permission cli
```

### 🗃️ Project Area

CLI Tooling

### 🔗 External Integration

GitHub

### 📝 Description & Context

The `create-github-app` command provides three permission checkboxes:

- Read access to content (required by Software Catalog to ingest data from repositories)
- Read access to members (required by Software Catalog to ingest GitHub teams)
- Read and Write to content and actions (required by Software Templates to create new repositories)

Enabling all three suggests that a software template should be able to create repositories, however this isn't the case. The default setup fails to create a repository due to insufficient permissions. The *Administration* permission is required to create repositories:

```
Administration
Repository creation, deletion, settings, teams, and collaborators
```

Additionally, the *Workflows* permission is required if the repository being created contains GitHub Workflows.

### 👍 Expected Behavior

The GitHub Application generated by the CLI with all permissions checked should be able to support the `publish:github` action in a template, going by the current descriptions.

### 🥾 Reproduction steps

. Create a GitHub org
. Run `npx @backstage/cli create-github-app $YOUR_ORG_ID` and enable all three permission levels
. Configure the GitHub Integration in your Backstage instance with the new GitHub App
. Load a template that uses `publish:github` to create a new repository in your GitHub org
. Use the template (it will fail with a permission error)

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes, but I would like some more guidance

nehaaprasad:
Hello @evanshortiss !,

- This PR fixes the GitHub App permissions issue where users couldn't create repositories via software templates even after selecting all available permissions. and added the missing administration and workflows permissions that are required for the publish:github action to work properly.

evanshortiss:
PR link for @naaa760 https://github.com/backstage/backstage/pull/30597

MihaiTabara:
> PR link for [@naaa760](https://github.com/naaa760) [#30597](https://github.com/backstage/backstage/pull/30597)

[...]

PR #30597, opened 2025-07-20
fix(cli): add missing GitHub App permissions for repository creation

naaa760:

#30530

### What I added

- Fixed the `create-github-app` command to include the missing GitHub App permissions required for repository creation and GitHub Workflows support. Previously, users who selected all three permission options would still get permission errors when using software templates with the `publish:github` action.

### Changes:
- Added **Administration permission** option for repository creation, deletion, settings, teams, and collaborators
- Added **Workflows permission** option for repositories containing GitHub Workflows  
- Updated permission descriptions for better clarity

- Now when running `npx @backstage/cli create-github-app <org>`, users will see 5 permission options instead of 3, allowing them to properly configure GitHub Apps for software template repository creation.

### DIFF
--- a/packages/cli/src/modules/create-github-app/commands/create-github-app/index.ts
+++ b/packages/cli/src/modules/create-github-app/commands/create-github-app/index.ts
@@ -43,8 +43,16 @@ export default async (org: string) => {
         checked: true,
       },
       {
-        name: 'Read and Write to content and actions (required by Software Templates to create new repositories)',
+        name: 'Read and Write to content and actions (provides content and actions write access)',
         value: 'write',
+      },
+      {
+        name: 'Administration (required for repository creation, deletion, settings, teams, and collaborators)',
+        value: 'administration',
+      },
+      {
+        name: 'Workflows (required if repositories contain GitHub Workflows)',
+        value: 'workflows',
       },
     ],
   });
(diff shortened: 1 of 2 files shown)


packages/cli/src/modules/create-github-app/commands/create-github-app/index.ts
```diff
@@ -43,9 +43,17 @@ export default async (org: string) => {
         checked: true,
       },
       {
-        name: 'Read and Write to content and actions (required by Software Templates to create new repositories)',
+        name: 'Read and Write to content and actions (provides content and actions write access)',
```
evanshortiss, line 46:
```suggestion
        name: 'Read and Write to repository contents, commits, branches, downloads, releases, and merges)',
```

packages/cli/src/modules/create-github-app/commands/create-github-app/index.ts
```diff
@@ -43,9 +43,17 @@ export default async (org: string) => {
...
         value: 'write',
       },
+      {
+        name: 'Administration (required for repository creation, deletion, settings, teams, and collaborators)',
+        value: 'administration',
+      },
+      {
+        name: 'Workflows (required if repositories contain GitHub Workflows)',
```
evanshortiss, line 54:
```suggestion
        name: 'Workflows (required if Software Templates will include GitHub Workflow files)',
```

evanshortiss:
Nice PR. I wonder does it make sense to split out into 2 new permissions, or simply roll it into the existing `Read and Write to content and actions`, and update the description to make it clear it grants admin rights

PR #30606, opened 2025-07-21
`auth-backend`: Implementing Dynamic Client Registration

benjdlambert:
This is a first pass implementation of an [`Dynamic Client Registration OAuth Spec`](https://oauth.net/2/dynamic-client-registration/)

This flow is designed to allow authentication from MCP Clients to get access to a Backstage Identity Token instead of having to use static tokens.

The client will register a client and client secret, and then the user will follow a redirect to the Backstage application to approve the authorization request which will then bind the user to that session and issue a token on behalf of the user.

There's room in here to add `refresh_tokens` and a `refresh` endpoint, but for now this is purposefully kept very small.

The frontend plugin part will also be added in a followup.


plugins/mcp-actions-backend/src/plugin.ts
```diff
@@ -66,6 +75,22 @@ export const mcpPlugin = createBackendPlugin({
...
+
+          const oidcResponse = await fetch(
+            `${authBaseUrl}/.well-known/openid-configuration`,
+          );
+
+          const oidcResponseJson = await oidcResponse.json();
+
+          res.json(oidcResponseJson);
+        });
+        rootRouter.use('/.well-known/oauth-authorization-server', authRouter);
```
benjdlambert: I think that we can most likely push this to using `WWW-Authenticate` header with the 401 to target the correct endpoint.

https://modelcontextprotocol.io/specification/draft/basic/authorization#authorization-server-location

gabemontero: hey @benjdlambert wrt cross referencing with the mcp auth spec like you posted here, in addition to your link here, I'm trying to cross reference what is in this PR with the high level summary over at https://modelcontextprotocol.io/specification/draft/basic/authorization#overview 

If I was to summarize where backstage is wrt that section (including what is in this PR) I **THINK** that 
- item 1) from that section in the MCP spec already exists in backstage, and this PR per the title, obviously tackles the dynamic client registration item from 2)
- based on what is going on in line 93 here, and what I saw in the changesets, perhaps item 4) around
```
MCP authorization servers MUST provide at least one of the following discovery mechanisms:
 - OAuth 2.0 Authorization Server Metadata ([RFC8414](https://datatracker.ietf.org/doc/html/rfc8414))
 - [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html)
``` 
   is also satisfied ... am I reading that correctly?  Or maybe it is at least partially addressed, with plans for future f/up?
- that leaves the third item, `MCP servers MUST implement OAuth 2.0 Protected Resource Metadata ([RFC9728](https://datatracker.ietf.org/doc/html/rfc9728)).`  ... as far as I can tell, that is not part of this PR.  Perhaps backstage provided that previously ?  I could not find mention of that type of detail in the docs though.  Or perhaps that is a f/up item that is on the radar for the future?


thanks

@schultzp2020 @johnmcollier @thepetk FYI

plugins/auth-backend/src/database/OidcDatabase.ts
```diff
@@ -0,0 +1,318 @@
...
+      ...session,
+      status: 'pending',
+    };
+  }
+
+  async updateAuthorizationSession(
+    session: Partial<AuthorizationSession> & { id: string },
+  ) {
+    const row = this.authorizationSessionToRow(session);
+    const updatedFields = Object.fromEntries(
```
schultzp2020, line 175: I wonder if it's worth creating a helper function with a clear name of what's happening since it's used in multiple spots. At first glance it takes a bit to understand what's happening.

plugins/auth-backend/src/service/OidcRouter.ts
```diff
@@ -15,42 +15,332 @@
...
     });
 
-    router.get('/v1/token', (_req, res) => {
-      res.status(501).send('Not Implemented');
+    // Authorization endpoint
+    // https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+    // Handles the initial authorization request from the client, validates parameters,
+    // and redirects to the Authorization Session page for user approval
+    router.get('/v1/authorize', async (req, res) => {
+      // todo(blam): maybe add zod types for validating input
```
schultzp2020: Is it possible to use the JSON Schema router here or is it too complex?

plugins/auth-backend/src/service/OidcRouter.ts
```diff
@@ -15,42 +15,332 @@
...
+        }
+
+        const redirectUrl = new URL(redirectUri as string);
+        redirectUrl.search = errorParams.toString();
+        return res.redirect(redirectUrl.toString());
+      }
+    });
+
+    // Authorization Session request details endpoint
+    // Returns Authorization Session request details for the frontned
```
schultzp2020:
```suggestion
    // Returns Authorization Session request details for the frontend
```

plugins/auth-backend/src/service/OidcRouter.ts
```diff
@@ -15,42 +15,332 @@
...
+    private readonly appUrl: string,
+    private readonly httpAuth: HttpAuthService,
+  ) {}
 
   static create(options: {
     auth: AuthService;
     tokenIssuer: TokenIssuer;
     baseUrl: string;
+    appUrl: string;
+    logger: LoggerService;
```
schultzp2020, line 45: Should we include the auditor as well?

plugins/auth-backend/report.sql.md
```diff
@@ -5,6 +5,59 @@
...
+| `code_challenge`        | `character varying`        | true     | 255        | -                 |
+| `code_challenge_method` | `character varying`        | true     | 255        | -                 |
+| `expires_at`            | `timestamp with time zone` | false    | -          | -                 |
+| `id`                    | `character varying`        | false    | 255        | -                 |
+| `nonce`                 | `character varying`        | true     | 255        | -                 |
+| `redirect_uri`          | `text`                     | false    | -          | -                 |
+| `response_type`         | `character varying`        | false    | 255        | -                 |
+| `scope`                 | `text`                     | true     | -          | -                 |
+| `state`                 | `character varying`        | true     | 255        | -                 |
+| `status`                | `text`                     | true     | -          | `'pending'::text` |
```
freben, line 22: interesting that this turned into text instead of varchar - no harm done though really

plugins/mcp-actions-backend/src/plugin.ts
```diff
@@ -66,6 +75,22 @@ export const mcpPlugin = createBackendPlugin({
         router.use('/v1', streamableRouter);
 
         httpRouter.use(router);
+
+        // todo(blam): there's probably a better way to proxy this, but it's required
+        // for mcp auth spec that it lives on the root of the mcp entrypoint server.
+        const authRouter = Router();
```
freben: This should not use express Router, but the one from `express-promise-router`

plugins/auth-backend/src/database/OidcDatabase.ts
```diff
@@ -0,0 +1,318 @@
...
+  user_entity_ref: string | null;
+  redirect_uri: string;
+  scope: string | null;
+  state: string | null;
+  response_type: string;
+  code_challenge: string | null;
+  code_challenge_method: string | null;
+  nonce: string | null;
+  status: 'pending' | 'approved' | 'rejected' | 'expired';
+  expires_at: string;
```
freben: heads up that all timestamps as returned by knex end up being `Date | string` (at least from what I've experienced, depending on the database engine, and when using `table.dateTime`) - and you'll have to normalize that when reading like https://github.com/backstage/backstage/blob/c1d4a4294d436ab8ab32034b1f8e6d58727b32c4/plugins/catalog-backend-module-history/src/database/tables.ts#L103

plugins/auth-backend/src/database/OidcDatabase.ts
```diff
@@ -0,0 +1,318 @@
...
+
+/**
+ * This class provides database operations for OpenID Connect (OIDC) authentication flows.
+ * It manages OIDC clients, authorization codes, and access tokens in the database.
+ */
+export class OidcDatabase {
+  private constructor(private readonly db: Knex) {}
+
+  static async create(options: { database: AuthDatabase }) {
+    const client = await options.database.get();
```
freben, line 114: Hm, maybe you don't want this to happen eagerly right during startup init right? The actual pool creation etc is fine to defer

plugins/auth-backend/src/database/OidcDatabase.ts
```diff
@@ -0,0 +1,318 @@
...
+    const updatedFields = Object.fromEntries(
+      Object.entries(row).filter(([_, value]) => value !== undefined),
+    );
+
+    const [updated] = await this.db<OAuthAuthorizationSessionRow>(
+      'oauth_authorization_sessions',
+    )
+      .where('id', session.id)
+      .update(updatedFields)
+      .returning('*');
```
freben, line 210: not all engines support `returning`, see https://github.com/backstage/backstage/blob/c1d4a4294d436ab8ab32034b1f8e6d58727b32c4/plugins/catalog-backend-module-history/src/database/operations/readHistorySubscription.ts#L72 (but that one could probably have included mysql too)

plugins/auth-backend/src/service/OidcRouter.ts
freben: Agreed, I would recommend doing that, it adds all of the validation needed upfront

plugins/auth-backend/src/service/OidcRouter.ts
```diff
@@ -15,42 +15,332 @@
...
+        scope,
+        state,
+        nonce,
+        code_challenge: codeChallenge,
+        code_challenge_method: codeChallengeMethod,
+      } = req.query;
+
+      if (!clientId || !redirectUri || !responseType) {
+        this.logger.error(`Failed to authorize: Missing required parameters`);
+        return res.status(400).json({
```
freben: does the standard require this error format, or could the appropriate error thrown out of `@backstage/errors` be fine too?

plugins/mcp-actions-backend/src/plugin.ts
benjdlambert: Hey :wave: sorry just getting back round to wrapping up this work.

Yes, that part is omitted for now. The reason being is that many clients don't yet support most of the latest spec, so this is really an implementation of https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization rather than the latest.

Thinking that we can add this as a followup on top of the current implementation.

I'd also really like to remove the horrible proxying on the `mcp-actions` backend plugin, but there's no way around this right now as clients like Cursor and Claude only implement the above specification AFAIK.

plugins/auth-backend/src/service/OidcRouter.ts
benjdlambert: I don't think so, this is part of the oauth spec here.

plugins/auth-backend/src/database/OidcDatabase.ts
benjdlambert: Fixed!

plugins/auth-backend/src/database/OidcDatabase.ts
benjdlambert: Fixed!

plugins/mcp-actions-backend/src/plugin.ts
benjdlambert: Fixed!

plugins/mcp-actions-backend/src/plugin.ts
gabemontero: no worries @benjdlambert and thanks for the clarification wrt which spec versions you were focused on (march vs. may)

we are in a similar boat wrt evaluating both the MCP servers and clients and what they can and cannot do currently .... also similarly, Cursor/Claude etc. have been the primary client test vehicles (aside form some the developer testing clients the MCP client impls provide), but that will be changing for us later this year, when our first real world clients beyond those IDE centric ones get to the point where they will apply their view on OAuth.

we'll then perhaps have some feedback ready on which follow ups need to be taken on next

in the interim, fwiw we were able to code up some RYO python clients that were able to exercise the full OAuth 2.0 flow pointed against a Backstage server running from this PR (see https://github.com/thepetk/backstage-dcr-poc/tree/main ) ...  aside from https://github.com/thepetk/backstage-dcr-poc/tree/main?tab=readme-ov-file#bypassing-the-browser-authorization things ran seamlessly

but given, among other things, the March vs. May distinction you made wrt the MCP spec, we would not cite what we ran into as something that needs to be addressed in this PR ... this PR certainly makes worthwhile progress along the MCP auth path as it were 

thanks

plugins/auth-backend/src/service/OidcRouter.ts
benjdlambert: Thinking we can address this in a followup :pray:

plugins/auth-backend/migrations/20250701120000_oidc_client_registration.js
```diff
@@ -0,0 +1,156 @@
...
+      .notNullable()
+      .comment('JSON array of supported response types');
+
+    table
+      .text('grant_types')
+      .notNullable()
+      .comment('JSON array of supported grant types');
+
+    table
+      .text('redirect_uris', 'longtext')
```
freben: Might as well have longtext on the other text columns too right? :) I don't think it hurts.

I guess in a way it could make sense to have either `.string` or `.text(..., 'longtext')` and nothing in between, right? Just to clariify that we KNOW for absolute certain that it's an impossibility that the former would ever exceed 255 chars but the latter might (and without the longtext it still ends up equal to `.string` on mysql).

This is honestly a bit philosophical almost, and the current settings probably won't ever be an issue in practice...

plugins/auth-backend/migrations/20250701120000_oidc_client_registration.js
```diff
@@ -0,0 +1,156 @@
...
+
+    table.string('nonce').nullable().comment('OIDC nonce parameter');
+
+    table
+      .enum('status', ['pending', 'approved', 'rejected', 'expired'])
+      .defaultTo('pending')
+      .comment('Authorization session status');
+
+    table
+      .timestamp('expires_at', { useTz: true, precision: 0 })
```
freben: Is there any expiry on `pending` things? So that they're not left available forever. Is that in `oidc_authorization_codes.expires_at`?

plugins/auth-backend/migrations/20250701120000_oidc_client_registration.js
freben, line 1: This should have an entry in the existing `migrations.test.ts` that exercises this both up and down, explicitly, for all supported engines.

plugins/auth-backend/src/database/OidcDatabase.ts
```diff
@@ -0,0 +1,379 @@
...
+          throw new Error(
+            `Failed to retrieve updated authorization session with id ${session.id}`,
+          );
+        }
+
+        return this.rowToAuthorizationSession(updated) as AuthorizationSession;
+      });
+    }
+
+    const [updated] = await this.db<OAuthAuthorizationSessionRow>(
```
freben: Might be worth checking that exactly one row came back; gives a nicer error message than the runtime js errors you'd get from inside `rowToAuthorizationSession` otherwise

plugins/auth-backend/src/database/OidcDatabase.ts
```diff
@@ -0,0 +1,379 @@
...
+          throw new Error(
+            `Failed to retrieve updated authorization code with code ${authorizationCode.code}`,
+          );
+        }
+
+        return this.rowToAuthorizationCode(updated) as AuthorizationCode;
+      });
+    }
+
+    const [updated] = await this.db<OidcAuthorizationCodeRow>(
```
freben: length check?

plugins/auth-backend/src/database/OidcDatabase.ts
```diff
@@ -0,0 +1,379 @@
...
+    };
+  }
+
+  private authorizationCodeToRow(
+    authorizationCode: Partial<AuthorizationCode>,
+  ): Partial<OidcAuthorizationCodeRow> {
+    return {
+      code: authorizationCode.code,
+      session_id: authorizationCode.sessionId,
+      expires_at: authorizationCode.expiresAt,
```
freben: no toDate?

plugins/auth-backend/src/database/OidcDatabase.ts
```diff
@@ -0,0 +1,379 @@
...
+  }
+
+  private rowToAuthorizationCode(
+    row: Partial<OidcAuthorizationCodeRow>,
+  ): Partial<AuthorizationCode> {
+    return {
+      code: row.code,
+      sessionId: row.session_id,
+      expiresAt: toDate(row.expires_at),
+      used: Boolean(row.used),
```
freben, line 394: i'm a bit torn on these methods accepting a Partial input, it makes it so easy to forget like here that you need to have a `row.used === undefined ? Boolean(row.used) : undefined` in a ton of places. Could it not be made in such a way that the input never is a Partial? would sure simplify _some_ things at least. I think you may have missed some above but not sure

plugins/auth-backend/src/service/OidcRouter.test.ts
```diff
@@ -17,135 +17,803 @@
...
+                      allow: 'unauthenticated',
+                    });
+                  },
+                });
+              },
+            }),
+          ],
+        });
+
+        httpAuth.credentials.mockResolvedValueOnce({
```
freben: did not `mockCredentials.user(...)` work in these tests?

plugins/auth-backend/src/service/OidcRouter.ts
freben: These do sound sensible for being audit events. The class isn't public so we can add it as a follow-up if needed but this seems important.

plugins/auth-backend/src/service/OidcRouter.ts
```diff
@@ -68,6 +98,318 @@ export class OidcRouter {
...
+      this.config.getOptionalBoolean(
+        'auth.experimentalDynamicClientRegistration.enabled',
+      )
+    ) {
+      // Authorization endpoint
+      // https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+      // Handles the initial authorization request from the client, validates parameters,
+      // and redirects to the Authorization Session page for user approval
+      router.get('/v1/authorize', async (req, res) => {
+        // todo(blam): maybe add zod types for validating input
```
freben, line 111: 👍 It's an auth facility after all, strictness sounds good. Same comment for all places below ofc

Do the openapi facilities don't apply for the auth backend? I guess they could. But a bit of a bigger shift to rewrite the plugin to use that.

plugins/auth-backend/src/service/OidcRouter.ts
```diff
@@ -68,6 +98,318 @@ export class OidcRouter {
...
+              'Missing required parameters: client_id, redirect_uri, response_type',
+          });
+        }
+
+        try {
+          const result = await this.oidc.createAuthorizationSession({
+            clientId: clientId as string,
+            redirectUri: redirectUri as string,
+            responseType: responseType as string,
+            scope: scope as string,
```
freben: i mean these should technically be `as string | undefined` right

plugins/auth-backend/src/service/OidcRouter.ts
```diff
@@ -68,6 +98,318 @@ export class OidcRouter {
...
+            codeChallenge: codeChallenge as string,
+            codeChallengeMethod: codeChallengeMethod as string,
+          });
+
+          // todo(blam): maybe this URL could be overridable by config if
+          // the plugin is mounted somewhere else?
+          // support slashes in baseUrl?
+          const authSessionRedirectUrl = new URL(
+            `/oauth2/authorize/${result.id}`,
+            this.appUrl,
```
freben:
```suggestion
            `./oauth2/authorize/${result.id}`,
            ensureTrailingSlash(this.appUrl),
```

or something like that to support instances that are exposed on subroutes. Or was there some spec-based reason that this absolutely HAD to be immediately under a domain?

plugins/auth-backend/src/service/OidcRouter.ts
```diff
@@ -68,6 +98,318 @@ export class OidcRouter {
...
+          }
+
+          const redirectUrl = new URL(redirectUri as string);
+          redirectUrl.search = errorParams.toString();
+          return res.redirect(redirectUrl.toString());
+        }
+      });
+
+      // Authorization Session request details endpoint
+      // Returns Authorization Session request details for the frontned
```
freben:
```suggestion
      // Returns Authorization Session request details for the frontend
```

plugins/auth-backend/src/service/OidcRouter.ts
```diff
@@ -68,6 +98,318 @@ export class OidcRouter {
...
+        const { sessionId } = req.params;
+
+        if (!sessionId) {
+          return res.status(400).json({
+            error: 'invalid_request',
+            error_description: 'Missing authorization session ID',
+          });
+        }
+
+        try {
```
freben, line 276: no need to check for user credentials here?

plugins/auth-backend/src/service/OidcRouter.ts
```diff
@@ -68,6 +98,318 @@ export class OidcRouter {
...
+          res.status(201).json({
+            client_id: client.clientId,
+            redirect_uris: client.redirectUris,
+            client_secret: client.clientSecret,
+          });
+        } catch (e) {
+          const description = isError(e) ? e.message : 'Unknown error';
+          this.logger.error(`Failed to register client: ${description}`, e);
+
+          res.status(500).json({
```
freben, line 418: all of these are making me think that it wouldn't be crazy to create a little middleware that maps `@backstage/errors` errors to this format. Things would be pretty simplified in some places if that was made

plugins/auth-backend/src/service/OidcService.ts
```diff
@@ -89,4 +108,404 @@ export class OidcService {
...
+
+    const allowedRedirectUriPatterns = this.config.getOptionalStringArray(
+      'auth.experimentalDynamicClientRegistration.allowedRedirectUriPatterns',
+    );
+
+    if (allowedRedirectUriPatterns) {
+      for (const redirectUri of opts.redirectUris ?? []) {
+        if (
+          !allowedRedirectUriPatterns.some(pattern =>
+            new RegExp(pattern).test(redirectUri),
```
freben: i wonder if full regex (you know, easy to forget to escape dots etc which could have true security implications) is overkill and minimatch / globby / whatnot would work out better?

plugins/auth-backend/src/service/OidcService.ts
```diff
@@ -89,4 +108,404 @@ export class OidcService {
...
+
+    const session = await this.oidc.getAuthorizationSession({
+      id: sessionId,
+    });
+
+    if (!session) {
+      throw new NotFoundError('Invalid authorization session');
+    }
+
+    if (DateTime.fromJSDate(session.expiresAt) < DateTime.now()) {
```
freben, line 229: Not sure this needed luxon :)

plugins/auth-backend/src/service/OidcService.ts
```diff
@@ -89,4 +108,404 @@ export class OidcService {
...
+      throw new InputError('Authorization session expired');
+    }
+
+    if (session.status !== 'pending') {
```
freben, line 233: arguably these types of invariant checks should be all the way inside `oidc.updateAuthorizationSession`, using database query semantics to ensure it's checked atomically (`WHERE status = 'pending'` etc) and handling all failed such updates under the "Authorization session not found or expired" umbrella. Same in the other methods below

plugins/auth-backend/migrations/20250701120000_oidc_client_registration.js
benjdlambert: So both the `sessions` and `codes` have expiry times. Which are checked any times sessions are manipulated, or codes exchanged. Which includes pending too.

plugins/auth-backend/migrations/20250701120000_oidc_client_registration.js
benjdlambert: Done!

plugins/auth-backend/src/database/OidcDatabase.ts
benjdlambert: Done!

plugins/auth-backend/src/service/OidcRouter.ts
benjdlambert: Sure, also updated it to save the userEntityRef to the session too for conisistency.

plugins/auth-backend/src/database/OidcDatabase.ts
benjdlambert: done!

plugins/auth-backend/src/database/OidcDatabase.ts
benjdlambert: Done!

plugins/auth-backend/src/database/OidcDatabase.ts
benjdlambert: Cleaned up what I could :pray:

plugins/auth-backend/src/service/OidcRouter.test.ts
benjdlambert: Updated for these test to use that util

PR #30614, opened 2025-07-22
Add data attributes for BUI

cdedreuille:

Backstage UI is based on CSS only for its theming. To make it work we need to add custom data attributes on the body for `light` and `dark` mode. What's new compared to our existing theming capacity in MUI is that we are differentiating the mode (`light` or `dark`) with the theme name. Users will be able to set a custom theme in both and light mode with Backstage UI.

We will introduce two new data attributes on the `body`:
- `data-theme-mode` - It could be `light` or `dark`
- `data-theme-name` - The actual name of your custom theme. By default we are using `backstage`

### DIFF
--- a/packages/theme/src/unified/UnifiedThemeProvider.tsx
+++ b/packages/theme/src/unified/UnifiedThemeProvider.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { ReactNode } from 'react';
+import { ReactNode, useEffect } from 'react';
 import CssBaseline from '@material-ui/core/CssBaseline';
 import {
   ThemeProvider,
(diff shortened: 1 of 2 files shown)


packages/theme/src/unified/UnifiedThemeProvider.tsx
```diff
@@ -70,6 +70,18 @@ export function UnifiedThemeProvider(
 
   const v4Theme = theme.getTheme('v4') as Mui4Theme;
   const v5Theme = theme.getTheme('v5') as Mui5Theme;
+  const themeMode = v4Theme ? v4Theme.palette.type : v5Theme?.palette.mode;
+  const themeName = 'backstage';
```
Rugvip, line 74: Thinking we can figure out how to make this configurable in a followup

cdedreuille: You are referring to the `themeName`? Yes we need to think about that after.

PR #30621, opened 2025-07-23
NFS: add NavContentBlueprint

Rugvip:

This adds a new simplified method of overriding the sidebar in the new frontend system. It avoids the need for using an extension override and data refs, and instead lets you implement the sidebar as a regular React component with the items passed through props. It also resolves the route for each item ahead of time to avoid the need to wrap them with `useRouteRef`.

As part of this change the `NavLogoBlueprint` is removed, as it ends up being redundant. It also simplifies the default nav implementation to avoid providing the default Backstage logo, as these are actually quite large and bloat the initial load. The default logos will be part of the `create-app` template instead.

The current implementation adds `to` and `text` to the items passed to the nav content component. It allows you to do this within the component:

```tsx
{items.map((item, index) => (
  <SidebarItem {...item} key={index} />
))}
```

The `to` I think is worth it, because wrapping for `useRouteRef` or using `RouteResolutionApi` quickly gets a bit messy. The `title` -> `text` is a bit annoying though, and we could consider updating the output of the nav item data ref instead.

### DIFF
--- a/.github/vale/config/vocabularies/Backstage/accept.txt
+++ b/.github/vale/config/vocabularies/Backstage/accept.txt
@@ -288,6 +288,7 @@ namespaced
 namespaces
 Namespaces
 namespacing
+navbar
 neuro
 newrelic
 nginx


plugins/app/src/extensions/AppNav.tsx
```diff
@@ -18,93 +18,93 @@ import {
...
+        // eslint-disable-next-line no-console
+        console.warn(
+          `NavItemBlueprint: unable to resolve route ref ${item.routeRef}`,
+        );
+        return [];
+      }
+      return [
+        {
+          to: link(),
+          text: item.title,
```
benjdlambert, line 73: Shall we just break the `dataRef` instead, and deprecate the `title` param with the intention to move that to `text` too?

Feeling that it's the safer breaking change without breaking everything all at once right?

.changeset/five-ducks-hide.md
```diff
@@ -0,0 +1,31 @@
+---
+'@backstage/frontend-plugin-api': minor
+---
+
+**BREAKING**: The `NavLogoBlueprint` has been removed and replaced by `NavContentBlueprint`, which instead replaces the entire navbar. The default navbar has also been switched to a more minimal implementation.
```
benjdlambert, line 5: Probs wanna add `navbar` to the vocab?

awanlin (commented):
Would it make sense to update the following to use this new Blueprint?

- Org plugin's `MyGroupsSidebarItem` - https://github.com/backstage/backstage/blob/03f2337f54591f43067bddb255e5eb9b1c0a8b69/plugins/org/src/components/MyGroupsSidebarItem/MyGroupsSidebarItem.tsx#L45
- Notification plugins's `NotificationsSidebarItem` - https://github.com/backstage/backstage/blob/03f2337f54591f43067bddb255e5eb9b1c0a8b69/plugins/notifications/src/components/NotificationsSideBarItem/NotificationsSideBarItem.tsx#L83 There's actually a comment about this: https://github.com/backstage/backstage/blob/03f2337f54591f43067bddb255e5eb9b1c0a8b69/plugins/notifications/src/alpha.tsx#L61

plugins/app/src/extensions/AppNav.tsx
Rugvip: We can't break the data ref as that's one of the things that help bridge across versions. I.e. we'd break existing plugins that haven't been upgraded.

Thinking we merge this as is for now and then see whether we can have the blueprint work help clean this up a bit. We shouldn't release in the current form into mainline though

Rugvip:
@awanlin this blueprint overrides the entire content of the sidebar, so it can't be used for those individual pieces. We're updating the `create-app` template for NFS to include some of those out of the box though

PR #30633, opened 2025-07-24
Initialize OTEL as part of backend startup 

kurtaking:

### DIFF
--- a/packages/backend-app-api/src/wiring/BackendInitializer.ts
+++ b/packages/backend-app-api/src/wiring/BackendInitializer.ts
@@ -44,6 +44,7 @@ import { DependencyGraph } from '../lib/DependencyGraph';
 import { ServiceRegistry } from './ServiceRegistry';
 import { createInitializationLogger } from './createInitializationLogger';
 import { unwrapFeature } from './helpers';
+import { initializeTelemetry } from './initializeTelemetry';
 
 export interface BackendRegisterInit {
   consumes: Set<ServiceOrExtensionPoint>;
@@ -153,7 +154,6 @@ function createInstanceMetadataServiceFactory(
     factory: async () => ({ getInstalledFeatures: () => installedFeatures }),
   });
 }
-
 export class BackendInitializer {
   #startPromise?: Promise<void>;
   #stopPromise?: Promise<void>;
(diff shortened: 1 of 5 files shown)


packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -254,11 +254,31 @@ export class BackendInitializer {
...
+
+    const rootLogger = await this.#serviceRegistry.get(
+      coreServices.rootLogger,
+      'root',
+    );
+
+    await initializeTelemetry({
+      rootConfig,
+      rootLogger,
+    });
```
eipc16, line 270: Not a Backstage maintainer, but I struggled with something similar in our company. I'm pretty sure that this is still to late to initialize telemetry here. Try to start the app with OpenTelemetry debug logs:

> OTEL_LOG_LEVEL=debug yarn start 2>&1 | tee output.log

You should see that in your implementation few instrumentations are missing. The few I noticed are:

- @opentelemetry/instrumentation-express (this also adds path attribute to http metrics)
- @opentelemetry/instrumentation-winston

Instrumentation for `knex` also does not seem to be working correctly, because no traces are emitted. That's because those modules were loaded before telemetry was initialized. 

Also, I'm not sure how I feel about Backstage controlling which instrumentations are enabled or not. Why not leave the instrumentation to the clients, but implement the `RootMetricsService` with `metrics` from `@opentelemetry/api`?

kurtaking: Hey, thank you so much for a review. This shines some light on areas that I will focus attention on next.

> Also, I'm not sure how I feel about Backstage controlling which instrumentations are enabled or not.

This would be configurable. Here is an example [1](https://github.com/backstage/backstage/pull/30495/files#diff-863c5aca4ae6535d8b21264b373a10212bb32836d96526036123dd353c017a95) [2](https://github.com/backstage/backstage/pull/30495/files#diff-279e2ae9ec69ece704c763a54475d6f34086844efc5bd9c4f341945e7497d2ceR127-R134) with much left to finalize. The idea is that we would ship with the instrumentation libraries that are the underlying implementations for core services with the ability to override via `app-config`.

> Why not leave the instrumentation to the clients?

I know that is the official recommendation from OTEL. Only instrument your libraries with the api, but ideally, we want to package as much into the backend initialization process with support via app-config and the ability to override the core service. 

Backstage is a unique enough case where we would like to at least experiment handling this. Think _observability by default_ with the ability to turn off or override.

> ...but implement the RootMetricsService with metrics from @opentelemetry/api?

Yes, I think that's likely the best path forward to getting the `MetricsService` out the door. Here is [example](https://github.com/backstage/backstage/pull/30495/files#diff-279e2ae9ec69ece704c763a54475d6f34086844efc5bd9c4f341945e7497d2ceR85-R90) code of a fully built root metrics service.

kurtaking:
Received quite a bit of feedback from the last framework SIG and the `tl;dr` is, we are going to hold off on trying to do anything with configuration until after we have established the metrics service. Closing for now.

PR #30634, opened 2025-07-24
`nfs`: Fixing scrolling behavior for the `info` Cards in the sidebar

benjdlambert:
Hopefully this should fix some of the behavior in the sidebar info cards.

This will allow the `info` cards to scroll separately of the main card content, so that you don't need to scroll all the way down to the bottom of the container. 

Also hides the scrollbar, which I'm not sure is the best accessibility wise or usability wise, but maybe we can revisit this later down the line.

https://github.com/user-attachments/assets/c8f0f8ce-d6c8-41d4-9fe8-6c34975b200e

### DIFF
--- a/.changeset/afraid-poets-thank.md
+++ b/.changeset/afraid-poets-thank.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog': patch
+---
+
+Fix card scrolling behaviour


plugins/catalog/src/alpha/DefaultEntityContentLayout.tsx
```diff
@@ -89,6 +98,21 @@ const useStyles = makeStyles<
       gridArea: 'summary',
       marginBottom: theme.spacing(3),
     },
+    infoArea: {
+      gridArea: 'info',
+      position: 'sticky',
+      top: theme.spacing(3),
+      maxHeight: `calc(100vh - ${theme.spacing(6)}px)`,
```
benjdlambert: Want to try and get rid of this, it's a bit magic at the minute.

Rugvip: Huh, is that really the header though? Looks more like just the top and bottom margin for the card content

plugins/catalog/src/alpha/DefaultEntityContentLayout.tsx
```diff
@@ -89,6 +97,23 @@ const useStyles = makeStyles<
       gridArea: 'summary',
       marginBottom: theme.spacing(3),
     },
+    infoArea: {
+      gridArea: 'info',
+      position: 'sticky',
```
Rugvip, line 102: Kind of expecting `absolute` at this point?

benjdlambert: Think that it needs to be sticky in order to get the behavior that we want. Absolute is just going to scroll the sidebar when scrolling the content.

plugins/catalog/src/alpha/DefaultEntityContentLayout.tsx
benjdlambert: Had another go at this, we've got to use 100vh to set the `info` card area height to give it a new scrollbar. There's not really a way round that at this point I don't think.

Rugvip (approved):
Yep, better than current state

PR #30652, opened 2025-07-25
techdocs-backend: migrate CachedEntityLoader to use BackstageCredentials

Speissi:

Related to the issue https://github.com/backstage/backstage/issues/27347 I noticed that the caching mechanism might not be working as intended since the token is used as part of the cache key and the token is regenerate every time a TechDocs path is accessed.

This PR replaces the token in the cache key with user entity ref or the subject depending on the type of credentials. This ensures that the cache key is stable enough to be reused.

An example using Redis with the cache. With the original code when opening a single TechDocs page in the browser the caching mechanism creates six different keys
image 3
If you then refresh the page instead of reusing any of the cached entities we get another set of six keys:
image 4

With the changes in the PR, opening the page results in one cache key and after refresh we are still seeing this one key as expected:
image 2

Note that for the "demo" I increased the TTL of the cached entities because I'm too old to take all the screenshots in under 5 seconds. 😁 

### Downside

Of course compared to the original implementation where the token is used as is, this is going to introduce some overhead. If the impact is too big to be worth it, then maybe the whole caching should be disabled (or for when permission is enabled)? Because currently it does not actually cache anything since every single time the key is different.

I presume this was the reason why the token-based implementation was introduced [here](https://github.com/backstage/backstage/commit/f1918143e51b3d261408794362f72df5714c188e)

### DIFF
--- a/.changeset/sad-cities-lay.md
+++ b/.changeset/sad-cities-lay.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-techdocs-backend': patch
+---
+
+Updated CachedEntityLoader to use BackstageCredentials instead of raw tokens for cache key generation. It now uses principal-based identification (user entity ref for users, subject for services) instead of token-based keys, providing more consistent caching behavior.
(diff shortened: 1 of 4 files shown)


PeaWarrior (approved):
This sounds plausible. Great find - thank you!

PeaWarrior:
Unfortunately the release was pushed out earlier this morning and the main branch is currently locked atm. Will have to merge in for the next release 😓

Speissi:
> Unfortunately the release was pushed out earlier this morning and the main branch is currently locked atm. Will have to merge in for the next release 😓

No worries and thanks for the review!

squid-ney:
Great find! Thank you for your contribution 💯

Issue #26353, opened 2024-08-29
🙏 Help wanted: Deprecate old plugin backends 


jhaals:
**TL;DR** make sure that `createRouter` and other exports are marked as deprecated. For the majority of packages there should only be one default export of the backend plugin itself. The `@backstage/backend-common` package is deprecated so usages of that package should also be avoided.

Backstage’s [new backend system](https://backstage.io/docs/backend-system/) is ready for general use; we are now asking for a full transition over to the new backend system, which involves [stop supporting the old system](https://backstage.io/docs/backend-system/building-plugins-and-modules/migrating/#remove-support-for-the-old-backend-system). By old system we mean having exports of `createRouter` and related types. There should only need to be one export like [this](https://github.com/backstage/backstage/blob/5d92499845e60e154b80481a679bfa3ec12f0dfb/plugins/notifications-backend/src/index.ts#L16) in the backend plugin’s `index.ts` file.

## How do I help?

Run `yarn backstage-repo-tools lint legacy-backend-exports` in the repository to get a full report of packages that require action.

#### Ensure that plugins contain a default export

Ensure that there is a default export of the backend plugin in `index.ts`, see [this example](https://github.com/backstage/backstage/blob/5d92499845e60e154b80481a679bfa3ec12f0dfb/plugins/notifications-backend/src/index.ts#L16).

If the plugin previously had a default export in `plugins/<plugin-id>/src/alpha.ts`, make sure that that export is deprecated and that the default export is moved to the non-alpha `index.ts` instead.

#### Deprecate `createRouter`, `RouterOptions` and similar types.

Here’s an [example](https://github.com/backstage/backstage/pull/26078) of a plugin’s `createRouter` being deprecated.

```diff
/**
+ * @deprecated Please migrate to the new backend system as this will be removed in the future.
 * @public
 * */
export async function createRouter(
  options: RouterOptions,
): Promise<express.Router> {
```

#### Remove deprecated exports

Ensure that deprecations have been out for at one mainline release before proceeding with removal of all deprecated exports. Removing exports from one release to another is not recommended 

The complete migration story for a backend plugin (including deprecation) is also mentioned in our [docs](https://backstage.io/docs/backend-system/building-plugins-and-modules/migrating#remove-support-for-the-old-backend-system).

If you take on migrating a backend plugin, feel free to add a comment in this issue to avoid duplicate work.

[...]

PR #30653, opened 2025-07-25
chore: remove deprecated @backstage/backend-common

Teo-ShaoWei:

Removed the deprecated `@backstage/backend-common`.

This resulted in a breaking change to the `createRouter` interface. Users have to now provide `auth: AuthService` and `httpAuth: HttpAuthService`.

Contributes to #26353 phase 2.

In hindsight, I realized the change has mostly been covered in #29015 already, but it should be just the portion that is ready for acceptance.

### DIFF
--- a/.changeset/curly-trains-join.md
+++ b/.changeset/curly-trains-join.md
@@ -1,0 +1,7 @@
+---
+'@backstage/plugin-kubernetes-backend': minor
+---
+
+Remove the deprecated `@backstage/backend-common`.
+
+This resulted in a breaking change to the interface of `createRouter`. Users have to now provide `auth: AuthService` and `httpAuth: HttpAuthService`.
--- a/plugins/kubernetes-backend/package.json
+++ b/plugins/kubernetes-backend/package.json
@@ -62,7 +62,6 @@
     "@aws-sdk/credential-providers": "^3.350.0",
     "@aws-sdk/signature-v4": "^3.347.0",
     "@azure/identity": "^4.0.0",
-    "@backstage/backend-common": "^0.25.0",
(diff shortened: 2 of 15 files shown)


awanlin (changes requested):
As this doesn't address the concern mentioned in the pervious PR with these changes - https://github.com/backstage/backstage/pull/29015#issuecomment-2753556148 - I don't think we should move forward with this PR in its current form. I'd like the maintainers thoughts on this, perhaps I'm misunderstanding what the blockers are I'm fairly confident that changing the auth is at the root of it.

I very much want to see usages of `@backstage/backend-common` gone, I've done a ton of that work myself, but it needs to be done correctly so we don't create other problems for Adopters 👍

Teo-ShaoWei:
Saw that @freben and @benjdlambert are back 🎉 Would like seek your thoughts on this PR please 🙏

Andy2003:
seems to be already resolved via https://github.com/backstage/backstage/pull/30864

benjdlambert:
@Teo-ShaoWei hey :wave: apologies for kind of taking over this work. I've had a plan about fixing this for a while now and wanted to get most of the breaking stuff shipped in the last next release this week.

I've got another followup PR with an alternative implementation of the extension points which is backwards compatible, but will allow the ability to get the default implementations for all the original extension points which were abstracted behind the k8s builder before.

Will drop that when I have the time to polish it up, but I think we can close this PR for now? :pray:

Teo-ShaoWei:
@benjdlambert no worries, I would always prefer if the core team takes leadership over the core code. I appreciate you taking the time to explain things to me 🙏

Yupz I've saw #30864 and it's good. I'll go ahead and close this PR then.

PR #30656, opened 2025-07-25
Fix: Show additional information about the cause in error messages from GitLab in the scaffolder-backend-module-GitLab.

PeterKoegel:
## Problem

In error messages from scaffolder-backend-module-GitLab that were reported from the GitLab API only very basic information like "Unauthorized" or "BadRequest" was contained but additional information about the error cause was lost.

## Solution

For errors of type GitbeakerRequestError the error's cause.description is added to the error message if available to provide more information to the user.

## Testing

- Errors of type GitbeakerRequestError result in the expected error message.
- All existing tests continue to pass

### DIFF
--- a/.changeset/deep-apples-attack.md
+++ b/.changeset/deep-apples-attack.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-scaffolder-backend-module-gitlab': patch
 ---
 
-Show additional information about the cause in error messages from GitLab in scaffolder-backend-module-GitLab.
+Show additional information about the cause in error messages from GitLab
--- a/.changeset/deep-apples-attack.md
+++ b/.changeset/deep-apples-attack.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend-module-gitlab': patch
+---
+
+Show additional information about the cause in error messages from GitLab in scaffolder-backend-module-GitLab.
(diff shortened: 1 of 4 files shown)


plugins/scaffolder-backend-module-gitlab/src/actions/helpers.ts
```diff
@@ -57,7 +58,13 @@ function isGitlabError(e: unknown): e is GitlabError {
   return isError(e) && 'description' in e && typeof e.description === 'string';
 }
 
+function isGitbeakerRequestError(e: unknown): e is GitbeakerRequestError {
+  return isError(e) && (e as any).name === 'GitbeakerRequestError';
```
benjdlambert:
```suggestion
  return isError(e) && e.name === 'GitbeakerRequestError';
```

Do we need the `e as any` cast? When `isError` will ensure it's an `ErrorLike`?

.changeset/deep-apples-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend-module-gitlab': patch
+---
+
+Show additional information about the cause in error messages from GitLab in scaffolder-backend-module-GitLab.
```
benjdlambert:
```suggestion
Show additional information about the cause in error messages from GitLab
```

benjdlambert (approved):
Thanks! :tada:

PR #30663, opened 2025-07-28
`mcp-actions`: Propagate errors from Actions correctly

benjdlambert:
Previously the `ResponseError` was propogated all the way to the MCP server implementation which would cause issues with 400's in the MCP server.

We're now trying to pull apart the most relevant error message from the Action execution and return it to the MCP client.

### DIFF
--- a/plugins/mcp-actions-backend/src/services/McpService.ts
+++ b/plugins/mcp-actions-backend/src/services/McpService.ts
@@ -22,7 +22,13 @@ import {
 import { JsonObject } from '@backstage/types';
 import { ActionsService } from '@backstage/backend-plugin-api/alpha';
 import { version } from '@backstage/plugin-mcp-actions-backend/package.json';
-import { NotFoundError } from '@backstage/errors';
+import {
+  isError,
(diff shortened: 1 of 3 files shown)


plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -72,25 +72,41 @@ export class McpService {
...
+          ],
+        };
+      } catch (e) {
+        return {
+          content: [
+            {
+              type: 'text',
+              text: [
+                '```json',
+                JSON.stringify(e?.body?.error?.cause ?? e, null, 2),
```
Rugvip: I'm feeling this assumes a bit too much about the error coming back from the actions client. Thinking it's best to check if it's actually a `ResponseError` and unpack it like this in that case? Otherwise serialize more directly. We may need to be a bit careful with the serialization as well, since `JSON.stringify` on a regular error instance doesn't work too well. We could use `serializeError` from `@backstage/errors`, and by that also make sure that `includeStack` isn't set.

PR #30664, opened 2025-07-28
feat: pglite connector for local development

lexhuismans-wk:
## PGlite connector for local development
This is still WIP and was an attempt to create a connector for PGlite. However, I got stuck and wanted to check if someone with more context is able to work on this, or if this idea should be abandoned. 

For local development we want to use [PGlite](https://pglite.dev/) since it is lightweight compared to e.g. spinning up a docker container. And we experience some inconsistencies with SQLite (locally) compared to Postgres (prod).

### Limitations
There are a few limitations with PGlite that have to be worked around.
- PGlite only handles _one_ connection as of now
   Leveraged [this](https://github.com/czeidler/knex-pglite/tree/main) knex extension and save and return the same connection on getClient.
- Looks like the concurrent updates from [this](https://github.com/backstage/backstage/blob/770c2c76c13c2b99c7da24e6a750531597207119/plugins/catalog-backend/migrations/20240130092632_search_index.js#L25) migration are not working
   Tried removing the concurrent part for now

### Remaining issues
Even after removing the concurrency start-up fails with
```
 PGlite is closed
```
Indicating that something is closing the connection(?). I'm not sure where and why it is being closed.

### Tested with config
```yaml
backend:
  database:
    client: pglite
```


benjdlambert:
Hey @lexhuismans-wk :wave:

Thanks for taking a look at this, we're on the same page when it comes to this that it would be nice to try and use `pglite` if possible over `sqlite`.

There's some other experiments that @Rugvip did that might unblock you a little bit, but there was also some concern about the performance and speed: https://github.com/backstage/backstage/pull/28318#issuecomment-2602036853

Maybe that work could be worth building on top of in this PR? :pray:

Rugvip:
Yep, restoring the DB contents across restarts was the main issue I ran into with pglite, and it's a fairly big blocker. Very much on board with switching out to use postgres for local development though, it would simplify things a fair bit

PR #30670, opened 2025-07-28
Export `FormField` type from Scaffolder

jabrks:

Attempting to re-export the Scaffolder plugin (e.g. to use `.withOverrides`) results in a type error due to the fact that the plugin return type references the `FormField` type which itself is not exported, therefore the TypeScript compiler is unable to determine the shape of it in the new plugin. This PR exports the `FormField` type to resolve this, however let me know if that is undesirable and another solution is preferred instead

```typescript
import { default as alphaScaffolderPlugin } from "@backstage/plugin-scaffolder/alpha";

export const scaffolderPlugin = alphaScaffolderPlugin;
             ~~~~~~~~~~~~~~~~
                    ^
Exported variable 'scaffolderPlugin' has or is using name 'FormField' from external module "node_modules/@backstage/plugin-scaffolder/dist/alpha" but cannot be named.
```

### DIFF
--- a/plugins/scaffolder/src/alpha/index.ts
+++ b/plugins/scaffolder/src/alpha/index.ts
@@ -30,4 +30,6 @@ export {
   type ScaffolderFormFieldsApi,
 } from '@backstage/plugin-scaffolder-react/alpha';
 
+export { type FormField } from '@internal/scaffolder';
+
 export { default } from './plugin';
--- a/.changeset/mighty-cycles-stay.md
+++ b/.changeset/mighty-cycles-stay.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder': patch
+---
+
+Export `FormField` type.
--- a/plugins/scaffolder/report-alpha.api.md
+++ b/plugins/scaffolder/report-alpha.api.md
@@ -281,6 +281,8 @@ export const formDecoratorsApi: ExtensionDefinition<{
 // @alpha (undocumented)
 export const formDecoratorsApiRef: ApiRef<ScaffolderFormDecoratorsApi>;
 
+export { FormField };
+
 export { formFieldsApiRef };
 
 // @alpha @deprecated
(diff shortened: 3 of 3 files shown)


plugins/scaffolder/src/alpha/index.ts
```diff
@@ -30,4 +30,6 @@ export {
   type ScaffolderFormFieldsApi,
 } from '@backstage/plugin-scaffolder-react/alpha';
 
+export { type FormField } from '@internal/scaffolder';
```
benjdlambert: Hmm, not sure about exporting this type from the `@internal/` package for `OpaqueTypes`.

I wonder if we should move this type into the `scaffolder-react` package instead?

@Rugvip thoughts?

benjdlambert (changes requested):
Thanks for this, not sure we have a precedent for exporting `internal` types, and as this is technically the public `OpaqueType` for `FormFields` it probably should be moved anyways.

The other thing is that I have a feeling that all this changes soon with https://github.com/backstage/backstage/pull/30673 too, once we reimplement the `FormField` Blueprint with those params, might not need a separate type anymore as it's all going to be captured under this type instead: https://github.com/backstage/backstage/pull/30673/files/805c298578299d6b646edb333c6fd0a3fda93836#diff-46f758acef7f2a438f301acd37aedce2532f5d954d26694f5cad5bb7fb6d22e8R75

jabrks:
@benjdlambert No worries, I did figure this might not be the right way of going about it due to that type being `internal` but I wanted to get a conversation going at least 🙂

benjdlambert (changes requested):
Let's move the `FormField` definition to `scaffolder-react/alpha` and then export it from there instead for now.

We're going to look at revisiting creation of `FormField`s with the new Blueprints work anyways, and making it easier to create and define them.

benjdlambert:
@jabrks I think that this /should/ work just the same, just moved things around a little bit for now.

Gonna merge this to get into the release today :tada:

benjdlambert (approved):

jabrks:
Nice! I was just about to push something similar, thanks @benjdlambert!

PR #30673, opened 2025-07-29
frontend-plugin-api: advanced blueprint param types

Rugvip:

This introduces a new way to define params for blueprints in the new frontend system. It's done with a new option to `createExtensionBlueprint` that is shaped in such a way that it is possible to use inferred type parameters as part of the declaration. As usual the design priority favors the use of the resulting blueprint, aiming to keep that as simple and intuitive as possible.

With this in place, we can clean up a lot of the APIs that were relying on helper functions for the creation of blueprint params. Most notable is the `ApiBlueprint`, which you currently use like this:

```ts
ApiBlueprint.make({
  name: 'error',
  params: {
    factory: createApiFactory({
      api: errorApiRef,
      deps: { alertApi: alertApiRef },
      factory: ({ alertApi }) => ...,
    })
  },
})
```

A big problem here is that as you're using the bluprint, using autocomplete to find what parameters are availble, you can easily get led in the wrong direction at this point:

```ts
ApiBlueprint.make({
  name: 'error',
  params: {
    factory: 
         // ^ autocomplete tells you to add an object with `api`, `deps`, and `factory`, but that's wrong
})
```

The new version looks like this:

```ts
ApiBlueprint.make({
  name: 'error',
  params: define => define({
    api: errorApiRef,
    deps: { alertApi: alertApiRef },
    factory: ({ alertApi }) => ...,
  }),
})
```

There's no need for a separate import of `createApiFactory` anymore, and if you be more clearly guided towards the correct use. If you try to define params as an object, you'll get this error:

```ts
ApiBlueprint.make({
  name: 'error',
  params: { }
// ^ Type '{ api: ... }' is not assignable to type '"Error: This blueprint uses advanced parameter types and requires you to pass parameters as using the following callback syntax: `<blueprint>.make({ params: define => define(<params>) })`"'.
})
```

We are rolling this out as an immediate breaking change for the `ApiBlueprint` and perhaps a few other ones. It does not affect the functionality and ability to install old plugins in an app though, it's only upon upgrading to the 1.42 release that this takes affect and the blueprint usage needs to be updated. This does also not affect any other blueprints that do not use these advanced param types, which the exception that it is possible to use the callback form to define params, because that helps with backwards compatibility if one wants to switch a blueprint back to simpler types.

Forking on a doc update for this too, but figured it's best to get up a bit earlier and maybe for the release today.

Oh, and this is how the `ApiBlueprint` is now defined:

```ts
export const ApiBlueprint = createExtensionBlueprint({
  kind: 'api',
  attachTo: { id: 'root', input: 'apis' },
  output: [factoryDataRef],
  dataRefs: { factory: factoryDataRef, },
  defineParams: <TApi, TImpl extends TApi, TDeps extends { [name in string]: unknown }>(
    params: ApiFactory<TApi, TImpl, TDeps>,
  ) => createExtensionBlueprintParams(params as AnyApiFactory),
  *factory(params) {
    yield factoryDataRef(params);
  },
});
```

### DIFF
--- a/.changeset/every-schools-find.md
+++ b/.changeset/every-schools-find.md
@@ -1,0 +1,20 @@
+---
+'@backstage/plugin-catalog-unprocessed-entities': patch
+'@backstage/frontend-app-api': patch
+'@backstage/core-compat-api': patch
+'@backstage/plugin-scaffolder-react': patch
+'@backstage/plugin-catalog-import': patch
+'@backstage/plugin-notifications': patch
+'@backstage/plugin-kubernetes': patch
+'@backstage/plugin-scaffolder': patch
+'@backstage/plugin-api-docs': patch
+'@backstage/plugin-devtools': patch
+'@backstage/plugin-techdocs': patch
+'@backstage/plugin-catalog': patch
+'@backstage/plugin-signals': patch
+'@backstage/plugin-search': patch
+'@backstage/plugin-home': patch
+'@backstage/plugin-app': patch
+---
+
+Internal update to use the new variant of `ApiBlueprint`.
--- a/.changeset/quiet-parks-cheer.md
+++ b/.changeset/quiet-parks-cheer.md
@@ -39,4 +39,4 @@ const example = ExampleBlueprint.make({
 });
 ```
 
-This `define => define(<params>)` is also known as the "callback syntax" and is is required if a blueprint is created with the new `defineParams` option. The callback syntax can also optionally be used for other blueprints too, which means that it is not a breaking change to remove the `defineParams` option, as long as the external parameter types remain compatible.
+This `define => define(<params>)` is also known as the "callback syntax" and is required if a blueprint is created with the new `defineParams` option. The callback syntax can also optionally be used for other blueprints too, which means that it is not a breaking change to remove the `defineParams` option, as long as the external parameter types remain compatible.
--- a/.changeset/quiet-parks-cheer.md
+++ b/.changeset/quiet-parks-cheer.md
@@ -1,0 +1,42 @@
+---
+'@backstage/frontend-plugin-api': minor
+---
+
+Added support for advanced parameter types in extension blueprints. The primary purpose of this is to allow extension authors to use type inference in the definition of the blueprint parameters. This often removes the need for extra imports and improves discoverability of blueprint parameters.
+
+This feature is introduced through the new `defineParams` option of `createExtensionBlueprint`, along with accompanying `createExtensionBlueprintParams` function to help implement the new format.
+
+The following is an example of how to create an extension blueprint that uses the new option:
+
+```ts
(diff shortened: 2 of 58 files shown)


.changeset/quiet-parks-cheer.md
```diff
@@ -0,0 +1,42 @@
...
+```ts
+const example = ExampleBlueprint.make({
+  params: define => define({
+    component: ...,
+    fetcher: ...,
+  }),
+});
+```
+
+This `define => define(<params>)` is also known as the "callback syntax" and is is required if a blueprint is created with the new `defineParams` option. The callback syntax can also optionally be used for other blueprints too, which means that it is not a breaking change to remove the `defineParams` option, as long as the external parameter types remain compatible.
```
benjdlambert:
```suggestion
This `define => define(<params>)` is also known as the "callback syntax" and is required if a blueprint is created with the new `defineParams` option. The callback syntax can also optionally be used for other blueprints too, which means that it is not a breaking change to remove the `defineParams` option, as long as the external parameter types remain compatible.
```

packages/frontend-plugin-api/src/wiring/createExtensionBlueprint.ts
```diff
@@ -37,13 +38,73 @@ import {
...
+ * Created with {@link createExtensionBlueprintParams}.
+ *
+ * @public
+ */
+export type ExtensionBlueprintParams<T extends object = object> = {
+  $$type: '@backstage/BlueprintParams';
+  T: T;
+};
+
+const OpaqueBlueprintParams = OpaqueType.create<{
```
benjdlambert, line 75: Should this be moved to `@internal/frontend` package with the rest of the `OpaqueTypes`?

.changeset/every-schools-find.md
```diff
@@ -0,0 +1,20 @@
...
+'@backstage/plugin-devtools': patch
+'@backstage/plugin-techdocs': patch
+'@backstage/plugin-catalog': patch
+'@backstage/plugin-signals': patch
+'@backstage/plugin-search': patch
+'@backstage/plugin-home': patch
+'@backstage/plugin-app': patch
+---
+
+Internal update to use the new variant of `ApiBlueprint`.
```
MihaiTabara, line 20: Nit: 👀 s/"internal"/"core"?

packages/frontend-plugin-api/src/wiring/createExtensionBlueprint.ts
Rugvip: No need really, since it's internal to `createExtensionBlueprint` and that all lives here.

.changeset/every-schools-find.md
Rugvip: It's referring to an internal update of the package, just code being updated without affecting the external API or behavior of the package

PR #30677, opened 2025-07-29
Initial implementation of an extension-based default analytics API

### LINKED ISSUE #28855: 🚀 Feature: Better support multiple analytics APIs in the new frontend system
### 🔖 Feature description

In the "old" frontend system, Backstage provided a `MultipleAnalyticsApi` implementation which could be used by an adopter to manually wire up several concrete analytics API implementations, sending events to all those provided.

We should bring this concept forward into the new frontend system. In the process, we should apply the goals/tenets of the new system by removing the need to hand-wire multiple implementations and instead natively support multiple implementations provided in features that were potentially dynamically loaded or discovered.

### 🎤 Context

There are a variety of legitimate and valuable use-cases for having multiple analytics API implementations installed, including:

- When evaluating multiple analytics platforms before choosing one.
- When migrating from one analytics platform to another (e.g. the great Universal Analytics to GA4 migration)
- When you just want to use different platforms for different reasons (e.g. one for APM and one for product analytics)

If there is a broader use-case for supporting API aggregations within Backstage, one possible implementation could involve introducing the concept of "aggregatable" APIs, which we could somehow resolve in the API registry.

If the use-cases for that are few and far between, then the most likely approach would be to introduce a dedicated blueprint for Analytics API implementations, which would be gathered and resolved in a new default analytics API implementation.

iamEAP:
## What / Why

Introduces an `AnalyticsBlueprint`, the results of which are used as an input to the implementation of the `analyticsApi`.

Some open questions:

- Should this live in `@backstage/plugin-app`, or should we try and get it into `@backstage/frontend-defaults` somehow?  If so, how?
- It's still possible for someone to override the `analyticsApi` in their app, and completely drop all extension-based Analytics API implementations (possibly unintentionally).  ...Should we consider registering this implementation of the `analyticsApi` at the `static` scope?  And if so, how?

Closes #28855


.changeset/mira-looking-ostrich.md
```diff
@@ -0,0 +1,18 @@
+---
+'@backstage/frontend-plugin-api': patch
+---
+
+Plugins should now use the new `AnalyticsBlueprint` to define and provide concrete analytics implementations. For example:
+
+```ts
+import { AnalyticsBlueprint } from '@backstage/frontend-plugin-api';
```
Rugvip: Naming-wise, I'm thinking this could be more specific. Perhaps `AnalyticsProviderBlueprint`, `AnalyticsImplementationBlueprint`, or `AnalyticsFactoryBlueprint`?

packages/frontend-plugin-api/src/blueprints/AnalyticsBlueprint.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+/**
+ * Creates analytics implementations.
+ *
+ * @public
+ */
+export const AnalyticsBlueprint = createExtensionBlueprint({
+  kind: 'analytics',
+  attachTo: [{ id: 'api:app/analytics', input: 'analyticsImplementations' }],
```
Rugvip:
```suggestion
  attachTo: [{ id: 'api:app/analytics', input: 'implementations' }],
```

Thinking this? Or w/e naming we arrive at. It's already within the context of the analytics API

plugins/app/src/extensions/AnalyticsApi.ts
```diff
@@ -0,0 +1,82 @@
...
+
+import {
+  analyticsApiRef,
+  AnalyticsBlueprint,
+  ApiBlueprint,
+  ApiRef,
+  createExtensionInput,
+} from '@backstage/frontend-plugin-api';
+
+export const AnalyticsApi = ApiBlueprint.makeWithOverrides({
```
Rugvip: Minor nit, but typically this'd be `analyticsApi`.

plugins/app/src/extensions/AnalyticsApi.ts
```diff
@@ -0,0 +1,82 @@
...
+  factory(originalFactory, { inputs }) {
+    // Pull out and aggregate deps from every implementation input into an
+    // object keyed by the apiRef ID to be passed to this API implementation as
+    // if they were its own deps.
+    const aggregatedDeps = inputs.analyticsImplementations
+      .flatMap<ApiRef<unknown>>(impls =>
+        Object.values(impls.get(AnalyticsBlueprint.dataRefs.factory).deps),
+      )
+      .reduce<{ [x: string]: ApiRef<unknown> }>((accum, ref) => {
+        accum[ref.id] = ref;
```
Rugvip, line 43: Oh this is a nice way to handle this 😁

Rugvip (commented):
Nice! 😁

Some minor stuff

Regarding the two question, I think the current implementation is good to ship

.changeset/mira-looking-ostrich.md
iamEAP: Makes sense! Went with `AnalyticsImplementationBlueprint`.

packages/frontend-plugin-api/src/apis/definitions/AnalyticsApi.ts
```diff
@@ -112,6 +133,8 @@ export type AnalyticsTracker = {
  * with relevant methods for instrumentation.
  *
  * @public
+ * @deprecated This type is now deprecated and will be removed in an upcoming
```
Rugvip: I'm assuming this is a bit of cleanup because we're able to do it? The analytics API can still be consumed via the ref, but we might as well define it in terms of the implementation type instead at this point.

That said we try not to touch the API types too much in `frontend-plugin-api`, aiming to it be more or less a drop-in replacement in that area. If we can expect very low usage of this type in plugin code it's fine to remove, but otherwise we might want to keep it around for now until we do a round of iteration on the API definitions.

packages/frontend-plugin-api/src/apis/definitions/AnalyticsApi.ts
```diff
@@ -124,6 +147,11 @@ export type AnalyticsApi = {
...
  * The API reference of {@link AnalyticsApi}.
  *
+ * @remarks
+ *
+ * To define a concrete Analytics Implementation, use
+ * {@link AnalyticsImplementationBlueprint} instead.
+ *
  * @public
  */
 export const analyticsApiRef: ApiRef<AnalyticsApi> = createApiRef({
```
Rugvip, line 155: Just to highlight here as well, if we're deprecating `AnalyticsApi`, we should replace these usages straight away.

Rugvip (commented):
Miss a bit in the first round, otherwise good to :shipit:

packages/frontend-plugin-api/src/apis/definitions/AnalyticsApi.ts
iamEAP: Hmmmm....  Yeah, I don't think it's really possible to completely remove the API / ref at this point (although it'd be nice to be able to in the future).  So I'm inclined to just remove the deprecation warning for now.

...I wonder if it's still worth keeping `AnalyticsImplementation` and `AnalyticsApi` as separate types?  Would conceptually allow the types to evolve independently (though in practice, that's probably unlikely).

Would it be crazy, @Rugvip, to completely delete `AnalyticsApi` in this PR, update `analyticsApiRef` to use `AnalyticsImplementation`, mark it a `BREAKING` change, and use that as the opportunity to inform consumers of the new way of providing analytics implementations?

Or better to just remove the deprecation here and we'll deal with this all later?

Rugvip: I'm leaning towards the last option. There is some usage of `AnalyticsApi` here and there, and we're trying to not add too much friction to the migration. Main thing would be to clearly nudge new implementations towards using the `AnalyticsImplementation` instead.

Issue #21923, opened 2023-12-18
🚀 Feature: Write an Admin Guide


dwelsch-esi:
### Write an Admin Guide
Write an Administrator's Guide for the Admin user role.

- Write start and stop instructions for production Backstage server
- Write instructions for installing and configuring plugins
- Write a guide to upgrading Backstage
- Write a database guide for Backstage

The goal is to capture step-by-step instructions for all tasks an Admin could reasonably be expected to complete in maintaining and running a production Backstage server (app) for an organization. It is important to provide unambiguous step-by-step instructions in each case. 

The existing documentation might contain helpful source material that you can pull into this doc change. See this catalog of the existing (at the time of the CNCF tech doc analysis) Backstage technical documentation pages:
https://github.com/cncf/techdocs/blob/main/assessments/0008-backstage/backstage-doc-survey.csv

### 🎤 Context

This issue is a recommended change resulting from an analysis of the Backstage documentation commissioned by CNCF. The analysis and supporting documents are here:
https://github.com/cncf/techdocs/tree/main/assessments/0008-backstage

### ✌️ Possible Implementation

The Admin Guide will look something like this:
- Server Administration
    - On a single server
        - Starting the Backstage server
        - Stopping the Backstage server
        - Determining the status of the Backstage server
    - On Kubernetes
        - ...
- Installing and configuring plugins
    - On a single server
    - On Kubernetes
   - ...
- Upgrading Backstage
-     - On a single server
    - On Kubernetes
   - ...
- Backstage database guide
    - Overview: What is the Backstage database and what does it contain?
    - The database server: default and alternatives. When would you use each?
    - Using a Postgres database
        - Installing
        - Maintaining
        - Upgrading
    - Using a MySQL database
        - ...
    - ...
    - Plugin databases

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

None

PR #30681, opened 2025-07-29
Add comprehensive Backstage Administrator Guide for production deploy…

Ayushmore1214:
…ment and maintenance

This PR adds a complete Admin Guide for Backstage that explains how to run and maintain the server in production. It covers how to start and stop Backstage on single servers and Kubernetes, how to install and configure plugins, how to upgrade the system safely, and how to manage the database. It includes troubleshooting tips and references to help administrators. This guide addresses issue #21923 and aims to make it easier for admins to operate Backstage smoothly.

### DIFF
--- a/docs/admin/administrator-guide.md
+++ b/docs/admin/administrator-guide.md
@@ -1,0 +1,246 @@
+---
+id: admin-guide
+title: Backstage Administrator Guide
+sidebar_label: Admin Guide
+description: Step-by-step guide for Backstage administrators responsible for running and maintaining production Backstage servers.
+---
+
+:::info
+This guide is intended for administrators managing production Backstage deployments on single-server and Kubernetes setups. It covers server operations, plugin management, upgrades, and database administration.
+:::
+
+# Table of Contents
+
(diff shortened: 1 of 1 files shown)


Parsifal-M:
Heya! 👋 

Thanks for looking into picking this up, if I am honest, I think we have a lot of the documentation in #21923 setup already, I think it would be better to see which parts we don't have covered in that issue and work on those specifically.

What would be useful as well is if we could "tag" the documentation which should be listed for "Admin" users (personas) so people know who the documentation is intended for.

I think the next steps here instead of adding a brand new page potentially duplicating information is, find the gaps from #21923 and see if we can update the docs that would match what is requested and possibly "tag" them with a persona like we have in the past (good enough approach) and see if we need new docs for any of the missing parts.

WDYT?

Ayushmore1214:
Thanks for the feedback! I’ll review the existing docs in #21923, identify gaps, and update or add targeted content as needed. I’ll also work on tagging docs for Admin users. I’ll keep you posted!

PR #30687, opened 2025-07-30
BUI - Improve useStyle() to avoid unnecessary rerenders

cdedreuille:

Before the useStyle() hook would rerender every time you change breakpoint, making it quite expensive on tables. With this update we are only using the useBreakpoint() hook if necessary, reducing considerably the number of rerenders.

### DIFF
--- a/.changeset/eager-results-do.md
+++ b/.changeset/eager-results-do.md
@@ -2,4 +2,4 @@
 '@backstage/ui': patch
 ---
 
-Improved useStyle() to only use the useBreakpoint() hook if necessary to improve performance.
+Improved `useStyle()` to only use the `useBreakpoint()` hook if necessary to improve performance.


packages/ui/src/hooks/useStyles.ts
```diff
@@ -66,27 +151,10 @@ export function useStyles<T extends ComponentDefinitionName>(
...
-      if (resolvedValue !== undefined) {
-        resolvedProps[key] = resolvedValue;
-        dataAttributes[`data-${key}`] = resolvedValue;
-      }
-    }
+  // Choose the appropriate hook based on whether props contain responsive values
+  if (hasResponsiveProps(props)) {
+    return useStylesResponsive(componentName, props);
+  } else {
+    return useStylesNonResponsive(componentName, props);
```
Rugvip: Hmm, technically this is not allowed, because you're not guaranteed that the props stay the same. I'd say the risk of that is big enough that we can't go with this approach. I wonder if there's something we can do to make `useBreakpoint` more lazy instead for example?

cdedreuille:
@Rugvip I tried to improve it a little bit to make `useBreakpoint()` lazy and improve `useStyles()` to avoid duplicated rerenders. The code is becoming a bit more complicated but from all the tests we have it looks pretty performant and safe. Let me know what you think.

.changeset/eager-results-do.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Improved useStyle() to only use the useBreakpoint() hook if necessary to improve performance.
```
Rugvip:
```suggestion
Improved `useStyle()` to only use the `useBreakpoint()` hook if necessary to improve performance.
```

Rugvip:
Ah but the tricky bit is that the new implementation in `useStyles` is breaking the rules of hooks. If a caller supplies props that don't have any responsive queries we'll end up on the `useStylesNonResponsive` path without any call to `useBreakpoint`. But if the caller then updates the props the branch will change on the rerender and we'll be calling `useStylesResponsive` instead, which leads to a call to`useBreakpoint` and in turn a bunch of other hooks, which in turn leads to a crash because we've changed the number of hooks called across renders.

I'd suggest aligning `useStylesNonResponsive` and `useStylesResponsive` so that you only need one of them, but to make sure that the extra hooks calls are as cheap as possible for the simple non-breakpoint cases.

cdedreuille:
You're right, this approach was still breaking the rules of hooks. I tried a new approach, removing the dependency of `useBreakpoint()` and create a `useLazyBreakpoint()` function instead. That way:
- `useLazyBreakpoint(hasResponsive)` is always called
- `useMemo(...)` is always called
- The number and order of hooks never changes between renders

What do you think @Rugvip ?

packages/ui/src/hooks/useBreakpoint.ts
```diff
@@ -25,40 +25,111 @@ export const breakpoints: { name: string; id: Breakpoint; value: number }[] = [
...
+  // Lazy up function - creates media query only when called
+  const up = useCallback((key: Breakpoint): boolean => {
+    const targetBreakpoint = breakpoints.find(bp => bp.id === key);
+    if (!targetBreakpoint) return false;
+
+    // For 'initial', always return true since we're always at or above 0px
+    if (key === 'initial') return true;
+
+    const query = `(min-width: ${targetBreakpoint.value}px)`;
+    return useSingleMediaQuery(query);
```
Rugvip, line 115: These calls to `useSingleMediaQuery` aren't safe, both because the function is called `up` so it's not communicated to the caller that it's a hook, but also because of the conditions that can cause an early return.

Rugvip (commented):
Yep `useLazyBreakpoint` looks alright, although `useBreakpoint` doesn't. Bit curious why one uses resize listeners and the other users media queries though, although afaik either is prolly file.

Rugvip (commented):
Had a thought on this one that I wonder whether it makes sense: Would shifting this all into a root context solve the issue too? As in compute that current breakpoint once at the root and then re-use through context instead? Might even help out with things like testing because it makes it a bit more straight-forward to inject behavior.

PR #30694, opened 2025-07-31
perf(gitlab): eliminate project ID resolution API calls to prevent ra…

kc9ddi:

perf(gitlab): eliminate project ID resolution API calls to prevent rate limiting

Replace project ID lookup API calls with direct project path usage in GitLab 
URL reader. This optimization leverages GitLab API v4's native support for 
URL-encoded project paths in endpoints, eliminating the need for preliminary 
/api/v4/projects/{path} calls that were causing rate limiting issues.

Changes:
- Update getGitLabFileFetchUrl to use project paths directly (now synchronous)
- Add extractProjectPath utility for consistent path extraction from URLs
- Modify buildProjectUrl to accept both project paths and IDs
- Optimize getGitlabArtifactFetchUrl to use project paths for artifact endpoints
- Remove resolveProjectToId function (no longer needed)
- Update all unit tests to expect project path-based URLs

This eliminates ~1 API call per file operation while maintaining full 
compatibility with both GitLab.com and self-hosted instances.

### DIFF
--- a/packages/integration/src/gitlab/core.ts
+++ b/packages/integration/src/gitlab/core.ts
@@ -28,22 +28,25 @@ import {
  *
  * Converts
  * from: https://gitlab.example.com/a/b/blob/master/c.yaml
- * to:   https://gitlab.com/api/v4/projects/projectId/repository/c.yaml?ref=master
+ * to:   https://gitlab.com/api/v4/projects/a%2Fb/repository/files/c.yaml/raw?ref=master
  * -or-
  * from: https://gitlab.com/groupA/teams/teamA/subgroupA/repoA/-/blob/branch/filepath
- * to:   https://gitlab.com/api/v4/projects/projectId/repository/files/filepath?ref=branch
+ * to:   https://gitlab.com/api/v4/projects/groupA%2Fteams%2FteamA%2FsubgroupA%2FrepoA/repository/files/filepath/raw?ref=branch
  *
  * @param url - A URL pointing to a file
  * @param config - The relevant provider config
+ * @param token - An optional auth token (not used in path extraction, kept for compatibility)
  * @public
  */
-export async function getGitLabFileFetchUrl(
+export function getGitLabFileFetchUrl(
   url: string,
   config: GitLabIntegrationConfig,
-  token?: string,
-): Promise<string> {
-  const projectID = await getProjectId(url, config, token);
-  return buildProjectUrl(url, projectID, config).toString();
+  _token?: string,
+): string {
+  // Use project path directly instead of making an API call to get project ID
+  // Note: _token parameter kept for backward compatibility but not used for path extraction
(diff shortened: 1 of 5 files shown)


packages/integration/src/gitlab/core.ts
```diff
@@ -105,6 +110,37 @@ export function buildProjectUrl(
...
+    repo = repo.replace(relativePath, '');
+  }
+
+  // Remove leading slash
+  return repo.replace(/^\//, '');
+}
+
 // Convert
 // from: https://gitlab.com/groupA/teams/teamA/subgroupA/repoA/-/blob/branch/filepath
 // to:   The project ID that corresponds to the URL
```
kc9ddi, line 146: This function (getProjectId) is no longer used in the Backstage codebase, but is exported publicly, so perhaps it is used externally. I think we should deprecate it rather than committing to continued maintenance. Does that sound correct, and if so, what is the process for doing that?

packages/integration/src/gitlab/core.ts
```diff
@@ -105,6 +110,37 @@ export function buildProjectUrl(
...
+ * from: https://gitlab.com/groupA/teams/teamA/subgroupA/repoA/-/blob/branch/filepath
+ * to:   groupA/teams/teamA/subgroupA/repoA
+ */
+export function extractProjectPath(
+  target: string,
+  config: GitLabIntegrationConfig,
+): string {
+  const url = new URL(target);
+
+  if (!url.pathname.includes('/blob/')) {
```
BhavikTanna, line 124: could this potentially be the raw file as well?  e.g https://gitlab.com/group/sub-group/-/raw/main/catalog-info.yaml in which case this would fail?

kc9ddi:
I would really like to keep this PR open, it fixes a very impactful performance issue for implementations that interact with Gitlab. @Rugvip

packages/integration/src/gitlab/core.ts
kc9ddi: I don't think so - previously, getProjectId always checked that the path contains 'blob' (https://github.com/backstage/backstage/pull/30694/files#diff-46e51f9a59e02a0645b31f2510402a31699feac5f3ca8de5117794fba93056adL118) -- I don't believe 'raw' paths were ever supported here.

freben:
It seems the branch was moved (force push?) after closure - github doesn't allow reopening it anymore.

freben:
But is it true to say that this already would be addressed in part at least by https://github.com/backstage/backstage/pull/31217 ?

kc9ddi:
I will investigate and raise a new PR if #31217 doesn't fully address this.

kc9ddi:
@freben FYI I opened https://github.com/backstage/backstage/pull/31392 to replace this PR. I believe these changes are still beneficial, even accounting for #31217.

PR #30696, opened 2025-07-31
frontend-defaults: move createPublicSignInApp to alpha

Rugvip:
🧹, matching the fact that this is still fairly experimental as we move towards stabilizing the new frontend system


benjdlambert (approved):

packages/frontend-defaults/package.json
```diff
@@ -5,9 +5,7 @@
     "role": "web-library"
   },
   "publishConfig": {
-    "access": "public",
-    "main": "dist/index.esm.js",
-    "types": "dist/index.d.ts"
```
freben: Hm we have a lot of these. Is that a general cleanup we could do?

packages/frontend-defaults/report.api.md
```diff
@@ -53,7 +53,7 @@ export interface CreateAppOptions {
   pluginInfoResolver?: FrontendPluginInfoResolver;
 }
 
-// @public
+// @public @deprecated (undocumented)
```
camilaibs, line 44: Could we delete it immediately since we warned that it is a breaking change in the changeset?

plugins/app/src/alpha/appModulePublicSignIn.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+    render(app.createRoot());
+
+    await expect(
+      screen.findByText('Sign in page'),
+    ).resolves.toBeInTheDocument();
+  });
+
+  it('should render the form redirect on sign-in', async () => {
+    const submitSpy = jest
+      .spyOn(HTMLFormElement.prototype, 'submit')
```
freben, line 59: !

packages/frontend-defaults/package.json
Rugvip: Oh I think these need to stay, removed them by mistake by going back and forth with an alpha export

packages/frontend-defaults/report.api.md
Rugvip: Ah right, either way one of the two should be updated 😅

plugins/app/src/alpha/appModulePublicSignIn.test.tsx
Rugvip: oh lawd 😅, well this is just moved

packages/frontend-defaults/report.api.md
Rugvip: Opted for removing the breaking notice and keep it around as deprecated for now. Figured there's no significant benefit to removing it straight away.

PR #30710, opened 2025-08-01
frontend-plugin-api: get rid of AnyExtensionDataRef

Rugvip:
🧹, the pattern we use nowadays is to make sure that type parameters are defaulted to the wide "any" type. With that a plain `ExtensionDataRef` can now be used directly instead of `AnyExtensionDataRef`.


.changeset/seven-crabs-stick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/frontend-plugin-api': patch
+---
+
+Added added defaults for all type parameters of `ExtensionDataRef` and deprecated `AnyExtensionDataRef`, as it is now redundant.
```
freben, line 5:
```suggestion
Added defaults for all type parameters of `ExtensionDataRef` and deprecated `AnyExtensionDataRef`, as it is now redundant.
```

freben (approved):

PR #30714, opened 2025-08-01
frontend-plugin-api: switch naming recommendation from define to defineParams

Rugvip:
Had a bit of discussion about this one. In the end we're thinking it's most clear to line up the naming of `defineParams` across the blueprint definition and invocation. They're the same function after all. This also likely makes it a bit easier to read and understand code that uses this. In the end this is only a recommendation of course, since it's just the name of a caller define function parameter.

### DIFF
--- a/.changeset/rich-seals-itch.md
+++ b/.changeset/rich-seals-itch.md
@@ -1,0 +1,5 @@
+---
+'@backstage/frontend-plugin-api': patch
+---
+
+Updated the recommended naming of the blueprint param callback from `define` to `defineParams`, making the syntax `defineParams => defineParams(...)`.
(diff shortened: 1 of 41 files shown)


camilaibs (approved):

.changeset/rich-seals-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/frontend-plugin-api': patch
+---
+
+Updated the recommended naming of the blueprint param callback from `define` to `defineParams`, making the syntax `defineParams => defineParams(...)`.
```
freben, line 5: add param to vale in this pr or another one

Rugvip: Added in https://github.com/backstage/backstage/pull/30703

PR #30718, opened 2025-08-01
Use GitLab API simple=true parameter when not filtering forks

kc9ddi:

The Gitlab API has a "simple" parameter which drastically reduces the number of fields in the response. We can use the "simple" parameter to reduce the amount of network traffic with Gitlab (thus speeding up the catalog sync), _unless_ the skipForkedRepos configuration option is in use (since information about whether the repo is a fork is omitted from the response of "simple" is used.)

This PR uses the "simple" to the Gitlab projects API, unless skipForkedRepos is true. In the case where skipForkedRepos is false (as is the default), this should yield a performance improvement when communicating with Gitlab. In the case where skipForkedRepos is true, we will continue to use the existing behavior.


kc9ddi:
Would like review of this PR

PR #30721, opened 2025-08-01
frontend-plugin-api,catalog-react: remove default* prefix from blueprint params

Rugvip:
We've gone a bit back and forth on this one, but I think in the end the `default*` naming for blueprint parameters that are also configurable does more harm than good. The intention was to signal that it was the default value of the param, but that it could also be changed through config. In reality we've seen that if anything it just leads to more confusion, and it's going to be evident what is configurable within the extension based on the configuration schema either way.

### DIFF
--- a/.changeset/cold-heads-arrive.md
+++ b/.changeset/cold-heads-arrive.md
@@ -1,0 +1,5 @@
+---
+'@backstage/frontend-plugin-api': minor
+---
+
+**BREAKING**: The `defaultPath` param of `PageBlueprint` has been renamed to `path`. This change does not affect the compatibility of extensions created with older versions of this blueprint.
--- a/packages/core-compat-api/src/convertLegacyPageExtension.tsx
+++ b/packages/core-compat-api/src/convertLegacyPageExtension.tsx
@@ -32,7 +32,11 @@ export function convertLegacyPageExtension(
   LegacyExtension: ComponentType<{}>,
   overrides?: {
     name?: string;
-    defaultPath?: string;
+    path?: string;
+    /**
+     * @deprecated Use the `path` param instead.
+     */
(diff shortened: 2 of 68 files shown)


.changeset/cold-heads-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/frontend-plugin-api': minor
+---
+
+**BREAKING**: The `defaultPath` param of `PageBlueprint` has been renamed to `path`. This change does not affect the compatibility of extensions created with older versions of this blueprint.
```
Rugvip, line 5:
```suggestion
**BREAKING**: The `defaultPath` parameter of `PageBlueprint` has been renamed to `path`. This change does not affect the compatibility of extensions created with older versions of this blueprint.
```

camilaibs (approved):

packages/core-compat-api/src/convertLegacyPageExtension.tsx
```diff
@@ -32,7 +32,11 @@ export function convertLegacyPageExtension(
   LegacyExtension: ComponentType<{}>,
   overrides?: {
     name?: string;
-    defaultPath?: string;
+    path?: string;
+    /**
+     * @deprecated Use the `path` param instead.
+     */
+    defaultPath?: [Error: `Use the 'path' override instead`];
```
freben, line 39: wait is that actually a label? :) that's a fascinating use - and it's pretty crazy it ends up in the api report too

freben (approved):

packages/core-compat-api/src/convertLegacyPageExtension.tsx
Rugvip: It's a labelled tuple element, like `[x: number, y: number]`

Issue #30716, opened 2025-08-01
cli: allow passing `cwd` to `repo start` and `package start` commands


drodil:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
cli package repo start
```

### 🗃️ Project Area

CLI Tooling

### 🔖 Need

At the moment, the `package start` and `repo start` commands work in the package/plugin folder by default. For example, having the following structure:

```
* config
  * local.yaml
* packages
  * app
    * package.json
  * backend
    * package.json
* templates
   * my-template.yaml
 ```
and in the `config/local.yaml`:

```yaml
catalog:
  locations:
    - type: file
      target: 'templates/my-template.yaml'
```

This does not work with the command `backstage-cli repo start --config config/local.yaml`, but instead you have to pass the relative directory to `packages/backend` and `packages/app` like: `backstage-cli repo start --config ../../config/local.yaml`. Additionally, the config file must define the target of the template file as `../../templates/my-template.yaml`.

### 📝 Proposal

Proposal is to add a new `--cwd` flag for the `repo start` and `package start` commands that can be used to point to the root of the repository, making it possible to use relative paths in the `--config` parameter as well as in the config file itself. 

This probably could be done using the `process.chdir` function, but I'm not sure what will break. This can be tested in this repository with the command: `backstage-cli repo start --config app-config.yaml`, which does not work even though the config points to an existing file. 

### 🔄 Alternatives

Alternatively, a new flag `--root-cwd` (or similar) could be available to make the `cwd` point to the root of the repository instead of having to pass `backstage-cli repo start --cwd ../..`

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

No, but I'm happy to collaborate on a PR with someone else

awanlin:
Hi @drodil, more out of my own curiosity and might help for context, but why do you have this folder structure for your config over using this `app-config.local.yaml` pattern?

drodil:
Hi @awanlin, actually, it's not just local config, but we have 4 different configs for different environments, and they are all stored in the config directory. This was just a simple example.

awanlin:
Sure, guess I was more interested in why you are using a folder structure over using:

- `app-config.local.yaml`
[...]

PR #30722, opened 2025-08-01
feat(config): allow specifying env specific config files

drodil:

this change allows to specify loading of environment specific config files based on `BACKSTAGE_ENVIRONMENT` environment variable.

relates to #30716

### DIFF
--- a/.changeset/better-eagles-tickle.md
+++ b/.changeset/better-eagles-tickle.md
@@ -2,4 +2,4 @@
 '@backstage/config-loader': patch
 ---
 
-Allow using `BACKSTAGE_ENVIRONMENT` for loading environment specific config files
+Allow using `BACKSTAGE_ENV` for loading environment specific config files
--- a/.changeset/better-eagles-tickle.md
+++ b/.changeset/better-eagles-tickle.md
@@ -1,0 +1,5 @@
+---
+'@backstage/config-loader': patch
+---
+
+Allow using `BACKSTAGE_ENVIRONMENT` for loading environment specific config files
--- a/docs/conf/index.md
+++ b/docs/conf/index.md
@@ -17,15 +17,15 @@ allowing for customization.
 
 Configuration is stored in YAML files where the defaults are `app-config.yaml`
 and `app-config.local.yaml` for local overrides. Additionally, it is possible
-to define environment based configuration files with `BACKSTAGE_ENVIRONMENT`
-environment variable, which will load `app-config.<BACKSTAGE_ENVIRONMENT>.yaml`.
+to define environment based configuration files with `BACKSTAGE_ENV`
+environment variable, which will load `app-config.<BACKSTAGE_ENV>.yaml`.
 
 Loading order of these files is as follows:
 
 1. `app-config.yaml`
-2. `app-config.<BACKSTAGE_ENVIRONMENT>.yaml`
+2. `app-config.<BACKSTAGE_ENV>.yaml`
 3. `app-config.local.yaml`
-4. `app-config.<BACKSTAGE_ENVIRONMENT>.local.yaml`
+4. `app-config.<BACKSTAGE_ENV>.local.yaml`
 
 Other sets of files can by loaded by passing `--config <path>` flags.
 Read more about the configuration loading order in the
--- a/docs/conf/index.md
+++ b/docs/conf/index.md
@@ -25,6 +25,7 @@ Loading order of these files is as follows:
 1. `app-config.yaml`
 2. `app-config.<BACKSTAGE_ENVIRONMENT>.yaml`
 3. `app-config.local.yaml`
+4. `app-config.<BACKSTAGE_ENVIRONMENT>.local.yaml`
 
 Other sets of files can by loaded by passing `--config <path>` flags.
 Read more about the configuration loading order in the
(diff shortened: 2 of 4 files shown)


Ayushmore1214 (commented):
Have you considered how this pattern integrates with existing overrides like app-config.local.yaml or command-line --config flags? A brief note in the docs tying these together could help users understand precedence.

drodil:
@Ayushmore1214 just like it has previously worked and it's documented already in this PR as well as the existing `Writing configuration files` documentation.

docs/conf/index.md
```diff
@@ -16,10 +16,23 @@ allowing for customization.
...
-various sources using for example `$env` and `$file` keys.
+and `app-config.local.yaml` for local overrides. Additionally, it is possible
+to define environment based configuration files with `BACKSTAGE_ENVIRONMENT`
+environment variable, which will load `app-config.<BACKSTAGE_ENVIRONMENT>.yaml`.
+
+Loading order of these files is as follows:
+
+1. `app-config.yaml`
+2. `app-config.<BACKSTAGE_ENVIRONMENT>.yaml`
+3. `app-config.local.yaml`
```
Rugvip, line 27: Hmm, I'm wondering if this should be `app-config.<BACKSTAGE_ENVIRONMENT>.local.yaml` instead, or simply ignored?

Thinking about the UX of running locally mostly, since `.local.` is very much not intended for any form of deployments.

My concern is that the config in `app-config.local.yaml` will mostly be for development, so if one runs `BACKSTAGE_ENVIRONMENT=production yarn start` locally I dunno if you really expect the `.local.` config to be picked up?

Alternatively we could double down a bit an mirror the pattern from the first two, so the full search would be this:

1. `app-config.yaml`
2. `app-config.<BACKSTAGE_ENVIRONMENT>.yaml`
3. `app-config.local.yaml`
4. `app-config.<BACKSTAGE_ENVIRONMENT>.local.yaml`

drodil: Makes sense. I will add the `app-config.<BACKSTAGE_ENVIRONMENT>.local.yaml`

.changeset/better-eagles-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/config-loader': patch
+---
+
+Allow using `BACKSTAGE_ENVIRONMENT` for loading environment specific config files
```
Rugvip: One more to discuss, should this be `BACKSTAGE_ENV` instead? Simply just to match `NODE_ENV`. I think other that than there's a mix of `*_ENV` and `*_ENVIRONMENT` in the wild.

benjdlambert: :+1: for `BACKSTAGE_ENV`

drodil: NODE_ENV also affects other things, so I would definitely go with `BACKSTAGE_ENV`

Rugvip: I was referring to the `_ENV` suffix, we definitely can't use `NODE_ENV` directly, like you say

drodil: I think `BACKSTAGE_ENV` was already used in some place 🤔

drodil: Ah, it was `BACKSTAGE_ENVIRONMENT` in the docs: https://github.com/backstage/backstage/blob/master/docs%2Fconf%2Fwriting.md

Which one should we go with? `ENV` is more like in Node but `ENVIRONMENT` is, well, more complete 😁

stephenglass:
I would suggest if possible to mention this on the release notes since this could be breaking behavior. In my case, I already inject a `BACKSTAGE_ENV` into the container which is then used to pass in the correct config files using the `--config` option in my startup script. So with this change I would need to modify my startup script, it's NBD but the extra visibility in release notes would be helpful in case someone else has done something similar to me :)

drodil:
@stephenglass just out of curiosity; is it `BACKSTAGE_ENV` or `BACKSTAGE_ENVIRONMENT` variable you are using? 😁

stephenglass:
> @stephenglass just out of curiosity; is it `BACKSTAGE_ENV` or `BACKSTAGE_ENVIRONMENT` variable you are using? 😁

`BACKSTAGE_ENV` 😄

.changeset/better-eagles-tickle.md
Rugvip: Feeling `BACKSTAGE_ENV`

Rugvip (changes requested):
Poking around a bit more at this I think `BACKSTAGE_ENV` would be better than `BACKSTAGE_ENVIRONMENT`. The `_ENV` suffix just seems a lot more common and is shorter but still very clear.

Two other contenders I spotted are `NODE_CONFIG_ENV` and `APP_ENV`. The former is used by https://github.com/node-config/node-config, so it might be a bit confusing for use to use, but might be alright too. I think `APP_ENV` is pretty good though, it's a bit more neutral. An issue with it is that it might be confusing in that people thing it's just for the frontend app, so could be good to avoid for that reason.

In the end I'm still feeling `BACKSTAGE_ENV` is the best option.

Rugvip (approved):
Let's :shipit:! 👍

PR #30727, opened 2025-08-01
feat(backend-defaults): add backend.logger configuration options

tcardonne:

Add `backend.logger` config options to configure the `RootLoggerService`.

The following parameters are available:

- `level` (string, optional): Sets the global log level. Possible values are 'debug', 'info', 'warn', or 'error'. Only messages at or above this level will be logged. This can also be set via the `LOG_LEVEL` environment variable, which takes precedence. Defaults to 'info'.

- `meta` (object, optional): Additional metadata to include with every log entry.

- `overrides` (array, optional): Allows to specify logger overrides for specific plugins or messages. Each override can match on plugin names, message patterns, or any field contained in the log, and set a custom log level for those matches.

Log level overrides are useful for controlling the volume of logs generated in Backstage.
They allow you to apply a global log level, `info` for example, while setting a stricter level, such as `warn`, for specific verbose plugins.

The reverse is also possible: you can set a global log level of `warn` while enabling a more detailed level, such as `debug`, for certain logs.

Example:

```yaml
backend:
  logger:
    meta:
      env: prod # Every log message will have `env="prod"`

    level: info # Set the global log level to info (the default)

    overrides:
      # Set the log level to 'debug' for the catalog plugin logs
      - matchers:
          plugin: catalog
        level: debug

      # Ignore 'info' incoming HTTP requests logs from the rootHttpRouter service
      - matchers:
          service: rootHttpRouter
          type: incomingRequest
        level: warn

      # Ignore logs starting with "Task worker starting", unless they're warnings or errors
      - matchers:
          message: ['/^Task worker starting/']
        level: warn
```

---

I wanted to reduce the amount of logs produced by Backstage, especially during development as our instance is a little bit too verbose. This allows to filter some `info` logs (eg, catalog or even http request logs).

Took inspiration from a [Discord thread](https://discord.com/channels/687207715902193673/1238472975783559188) where this was need was discussed too.

As usual, open to feedbacks 🙌 


.changeset/silly-horses-share.md
```diff
@@ -0,0 +1,38 @@
+---
+'@backstage/backend-defaults': patch
+---
+
+Add `backend.logger` config options to configure the `RootLoggerService`.
```
awanlin, line 5: I'd probably keep this line and then include a link to the docs to keep this short, we like to keep the changelogs focused on any change an Adopters might need to make for new features it's best to sent them to the docs.

docs/backend-system/core-services/root-config.md
```diff
@@ -36,6 +36,43 @@ createBackendPlugin({
...
+  logger:
+    meta:
+      env: prod # Every log message will have `env="prod"`
+
+    level: debug # Set the global log level to debug
+
+    overrides:
+      # Set the log level to 'warn' for the catalog plugin logs
+      - matchers:
+          plugin: catalog
```
awanlin: Some of the biggest complaints I hear from Backstage adopters when working locally is the verbose logs from things like `rootHttpRouter`. As this is a core service and doesn't have a `pluginId`, how can we use this as a matcher? Honestly think that would be one of the biggest use of this feature.

.changeset/silly-horses-share.md
tcardonne: Done 👍

docs/backend-system/core-services/root-config.md
tcardonne: Yes that was my initial goal too. Found out that those logs have `service=rootHttpRouter` and `type=incomingRequest`.

I added this usecase in the configuration sample in the docs (also it demonstrates that matchers can be combined as "and").

```yaml
      # Ignore 'info' incoming HTTP requests logs from the rootHttpRouter service
      - matchers:
          service: rootHttpRouter
          type: incomingRequest
        level: warn
```

(I also moved the docs to the correct `root-logger` service page instead of `root-config` 😅)

tcardonne:
Some notes after the discussion during the framework SIG:

The `overrides` naming seems a bit misleading to me because the global log level will take precedence anyway.

To filter logs based on their content, we use a formatter (see [Filtering 'info' Objects](https://github.com/winstonjs/winston?tab=readme-ov-file#filtering-info-objects) on Winston docs).
The formatter receive all logs (even those below the global log level, that's why I placed an early [return here](https://github.com/backstage/backstage/pull/30727/files#diff-2b604272723e7b7b09a0fd324aeda02fa5f6eb18308653cd0a164826763f3ec6R101-R105))
The global level seem to be applied after.

Ideally I think we should be able to do something like this:

```yaml
backend:
  logger:
    level: warn # Set the global log level to warn

    overrides:
      # ... but for the catalog we want debug
      - matchers:
          plugin: catalog
        level: debug
```

With the current state of the PR, it's not possible to do this (you can only filter more logs, not filter less).

However I'm not sure how to correctly implement this. Maybe setting the WinstonLogger level to "silly" (not sure but I guess the default level is 'info'?) so it logs everything and only rely on a custom formatter to do the filtering?

I'll try to dig a bit deeper on that when I have some time this week, but happy if one of you have any insights!

---

One of the think I'm also not sure about, is where to implement the formatter: is the current way ok, or does it make more sense to add this logic in the `WinstonLogger` instead?

--- 

Also happy to hear @Rugvip's point of view on the regex matcher https://github.com/backstage/backstage/pull/30727/files#diff-3bdd79033ff7320e9999049badb3a61557c9d4bfdb92c2208cbde3a4dedcd081R52-R53 😄

tcardonne:
> However I'm not sure how to correctly implement this. Maybe setting the WinstonLogger level to "silly" (not sure but I guess the default level is 'info'?) so it logs everything and only rely on a custom formatter to do the filtering?

I went ahead with this, it's easier to understand the logic of the filtering this way (use global log level if a log doesn't match any override).

---

I changed the PR to draft as I'd like to iterate a bit and implement all of this in the WinstonLogger instead.

tcardonne:
Ready for another round 👍

packages/backend-defaults/src/entrypoints/rootLogger/config.ts
```diff
@@ -0,0 +1,52 @@
...
+    ?.get<JsonObject>();
+
+  const overridesConfig = config.getOptionalConfigArray(
+    'backend.logger.overrides',
+  );
+  const overrides = overridesConfig?.map((override, i) => {
+    const overrideLevel = override.getString('level');
+    if (winstonLevels[overrideLevel] === undefined) {
+      throw new Error(
+        `Invalid config at backend.logger.overrides[${i}].level, '${overrideLevel}' is not a valid Winston npm logging level`,
```
schultzp2020: I'm wondering if we should include the valid options in the error message.

packages/backend-defaults/src/entrypoints/rootLogger/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+  if (typeof matcher !== 'string') {
+    return logField === matcher;
+  }
+
+  if (
+    matcher.startsWith('/') &&
+    matcher.endsWith('/') &&
+    typeof logField === 'string'
+  ) {
+    const regex = new RegExp(matcher.slice(1, -1));
```
Rugvip: I think it's worth refactoring this a lil bit to avoid compiling the regex on every test, it can actually get quite expensive. Typically you'd do a `createLogFieldMatcher(matcher: ...): (logField: unknown) => boolean` instead.

packages/backend-defaults/src/entrypoints/rootLogger/config.ts
Rugvip: 👍 , "must be one of ....". Also no need to mention Winston here since that just happens to be what we lean on for the implementation. Actually [our levels](https://github.com/backstage/backstage/blob/210406dfc08f2391dae3648528111c55ac930a02/packages/backend-plugin-api/src/services/definitions/LoggerService.ts#L27-L30) don't even match the [winston levels](https://www.npmjs.com/package/winston#logging-levels).

packages/backend-defaults/src/entrypoints/rootLogger/WinstonLogger.test.ts
```diff
@@ -118,4 +118,56 @@ describe('WinstonLogger', () => {
...
+      format: format.json(),
+      transports: [mockTransport],
+    });
+
+    logger.debug('debug log');
+
+    expect(mockTransport.log).not.toHaveBeenCalled();
+  });
+
+  it('should not filter logs below the default log level with an override', () => {
```
Rugvip, line 139: Nice! 👍 

Just to double check, the opposite is true as well right? As in if the override lowers the level we'll actually filter out the log. Thinking it's worth to cover with a test.

packages/backend-defaults/report-rootLogger.api.md
```diff
@@ -31,14 +33,31 @@ export class WinstonLogger implements RootLoggerService {
...
 
+// @public (undocumented)
+export type WinstonLoggerLevelOverride = {
+  matchers: WinstonLoggerLevelOverrideMatchers;
+  level: string;
+};
+
+// @public (undocumented)
+export type WinstonLoggerLevelOverrideMatchers = {
+  [key: string]: JsonValue | undefined;
```
Rugvip: We can narrow the matchers down a bit more right? I'm thinking `JsonPrimitive | JsonPrimitive []`

Rugvip (commented):
Can't wait for this actually 😁, I find myself doing a lot of `LOG_LEVEL=error` and then `console.log` during development, but being able to lean on existing logs would be much nicer. Just a couple of things

Thinking as a potential followup it could be convenient to be able to specify filters via the env too, something like `LOG_LEVEL=error,plugin=catalog:debug`

packages/backend-defaults/src/entrypoints/rootLogger/config.ts
tcardonne: Wasn't sure of it either, thanks for the suggestion 👍 

Changed to:
`Invalid config at backend.logger.overrides[0].level, 'invalid' is not a valid logging level, must be one of 'error', 'warn', 'info' or 'debug'.`

packages/backend-defaults/report-rootLogger.api.md
tcardonne: Yes 💯 Didn't know it existed 😅

packages/backend-defaults/src/entrypoints/rootLogger/WinstonLogger.test.ts
tcardonne: Yes, I added the test for that

packages/backend-defaults/src/entrypoints/rootLogger/utils.ts
tcardonne: I see, I refactored a bit, let me know what you think 👍

tcardonne:
Adressed requested changes, thanks for the feedbacks 🚀 

> Thinking as a potential followup it could be convenient to be able to specify filters via the env too, something like LOG_LEVEL=error,plugin=catalog:debug

Yes it could be useful, would need to find an easy-to-use syntax. And maybe not support all matcher kinds (eg, AND/OR combinations). Also it could duplicate what could be already be achieved with `APP_CONFIG_backend_logger` env var (not sure though but writing what's in my mind).

Rugvip (approved):
Awesome work, thank you! 😁 

Let's :shipit: 🎉

PR #30732, opened 2025-08-04
fix(scaffolder): show correct template title in Top & Recently Visited

its-mitesh-kumar:
## Description 
This PR resolves an issue where the `Top Visited` and `Recently Visited` cards on the home page always display the generic title `Create a new component`, regardless of the actual template selected. We now ensure the specific template title is captured as `Template ${templateTitle}`.

## Fixes: 

Fixing [issue-29963](https://github.com/backstage/backstage/issues/29963)

## UI before changes

https://github.com/user-attachments/assets/3b9fcdac-8f28-4dab-b5ab-7c2832e36111

## UI after changes

https://github.com/user-attachments/assets/5fc1c021-ef9a-42b8-84a3-4c33ef944660

### DIFF
--- a/.changeset/petite-pears-tickle.md
+++ b/.changeset/petite-pears-tickle.md
@@ -1,0 +1,6 @@
+---
+'@backstage/plugin-scaffolder': patch
+'@backstage/plugin-home': patch
+---
+
+Fixed an issue where the `Top Visited` and `Recently Visited` cards displayed a generic title (`Create a new component`) instead of the specific template name. Titles are now shown as `Template ${templateTitle}`.
--- a/plugins/home/src/api/VisitsApi.ts
+++ b/plugins/home/src/api/VisitsApi.ts
@@ -126,6 +126,12 @@ export interface VisitsApi {
    * @param queryParams - optional search query params.
    */
   list(queryParams?: VisitsApiQueryParams): Promise<Visit[]>;
+  /**
+   * Update the name of an existing visit without affecting hit count.
+   * @param pathname - the pathname of the visit to update
+   * @param name - the new name for the visit
+   */
+  updateName(pathname: string, name: string): Promise<void>;
 }
 
 /** @public */
--- a/plugins/scaffolder/src/alpha/components/TemplateWizardPage/TemplateWizardPage.tsx
+++ b/plugins/scaffolder/src/alpha/components/TemplateWizardPage/TemplateWizardPage.tsx
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { ComponentType, useCallback, useState } from 'react';
+import { ComponentType, useCallback, useState, useEffect } from 'react';
 import { Navigate, useNavigate } from 'react-router-dom';
 import useAsync from 'react-use/esm/useAsync';
 import {
@@ -33,6 +33,7 @@ import {
   FormProps,
   FieldExtensionOptions,
   ReviewStepProps,
+  TemplateParameterSchema,
 } from '@backstage/plugin-scaffolder-react';
 import { catalogApiRef } from '@backstage/plugin-catalog-react';
 
@@ -53,6 +54,26 @@ import { scaffolderTranslationRef } from '../../../translation';
 
 import { TemplateWizardPageContextMenu } from './TemplateWizardPageContextMenu';
 import { useFormDecorators } from '../../hooks';
+import { visitsApiRef } from '@backstage/plugin-home';
+import { useLocation } from 'react-router-dom';
+
+const useUpdateVisitName = (manifest: TemplateParameterSchema | undefined) => {
+  const { t } = useTranslationRef(scaffolderTranslationRef);
+  const visitsApi = useApi(visitsApiRef);
+  const { pathname } = useLocation();
+
+  useEffect(() => {
+    if (!manifest?.title) return;
+    visitsApi
+      .updateName(
+        pathname,
+        t('templateWizardPage.templateWithTitle', {
+          templateTitle: manifest.title,
+        }),
+      )
+      .catch(() => {});
+  }, [manifest?.title, pathname, visitsApi, t]);
+};
 
 /**
  * @alpha
@@ -92,6 +113,8 @@ export const TemplateWizardPage = (props: TemplateWizardPageProps) => {
 
   const { manifest } = useTemplateParameterSchema(templateRef);
   const decorators = useFormDecorators();
+
+  useUpdateVisitName(manifest);
 
   const { value: editUrl } = useAsync(async () => {
     const data = await catalogApi.getEntityByRef(templateRef);
@@ -138,7 +161,13 @@ export const TemplateWizardPage = (props: TemplateWizardPageProps) => {
     <AnalyticsContext attributes={{ entityRef: templateRef }}>
       <Page themeId="website">
         <Header
-          pageTitleOverride={t('templateWizardPage.pageTitle')}
+          pageTitleOverride={
+            manifest?.title
+              ? t('templateWizardPage.templateWithTitle', {
+                  templateTitle: manifest.title,
+                })
+              : t('templateWizardPage.pageTitle')
+          }
           title={t('templateWizardPage.title')}
           subtitle={t('templateWizardPage.subtitle')}
           {...props.headerOptions}
--- a/plugins/scaffolder/src/alpha/components/TemplateWizardPage/TemplateWizardPage.tsx
+++ b/plugins/scaffolder/src/alpha/components/TemplateWizardPage/TemplateWizardPage.tsx
@@ -25,6 +25,8 @@ import {
   useApi,
   useRouteRef,
   useRouteRefParams,
+  useApiHolder,
+  createApiRef,
 } from '@backstage/core-plugin-api';
 import {
   scaffolderApiRef,
@@ -54,16 +56,27 @@ import { scaffolderTranslationRef } from '../../../translation';
 
 import { TemplateWizardPageContextMenu } from './TemplateWizardPageContextMenu';
 import { useFormDecorators } from '../../hooks';
-import { visitsApiRef } from '@backstage/plugin-home';
 import { useLocation } from 'react-router-dom';
+
+const visitsApiRef = createApiRef<{
+  updateName: (pathname: string, name: string) => Promise<void>;
+}>({
+  id: 'homepage.visits',
+});
 
 const useUpdateVisitName = (manifest: TemplateParameterSchema | undefined) => {
   const { t } = useTranslationRef(scaffolderTranslationRef);
-  const visitsApi = useApi(visitsApiRef);
+  const apiHolder = useApiHolder();
   const { pathname } = useLocation();
 
   useEffect(() => {
     if (!manifest?.title) return;
+
+    const visitsApi = apiHolder.get(visitsApiRef);
+    if (!visitsApi) {
+      return;
+    }
+
     visitsApi
       .updateName(
         pathname,
@@ -72,7 +85,7 @@ const useUpdateVisitName = (manifest: TemplateParameterSchema | undefined) => {
         }),
       )
       .catch(() => {});
-  }, [manifest?.title, pathname, visitsApi, t]);
+  }, [manifest?.title, pathname, apiHolder, t]);
 };
 
 /**
--- a/plugins/scaffolder/src/translation.ts
+++ b/plugins/scaffolder/src/translation.ts
@@ -353,6 +353,7 @@ export const scaffolderTranslationRef = createTranslationRef({
       pageContextMenu: {
         editConfigurationTitle: 'Edit Configuration',
       },
+      templateWithTitle: 'Template {{templateTitle}}',
     },
     templateEditorToolbar: {
       customFieldExplorerTooltip: 'Custom Fields Explorer',
--- a/plugins/home/report.api.md
+++ b/plugins/home/report.api.md
@@ -282,6 +282,7 @@ export const VisitListener: ({
 export interface VisitsApi {
   list(queryParams?: VisitsApiQueryParams): Promise<Visit[]>;
   save(saveParams: VisitsApiSaveParams): Promise<Visit>;
+  updateName(pathname: string, name: string): Promise<void>;
 }
 
 // @public
@@ -312,6 +313,7 @@ export class VisitsStorageApi implements VisitsApi {
   static create(options: VisitsStorageApiOptions): VisitsStorageApi;
   list(queryParams?: VisitsApiQueryParams): Promise<Visit[]>;
   save(saveParams: VisitsApiSaveParams): Promise<Visit>;
+  updateName(pathname: string, name: string): Promise<void>;
 }
 
 // @public (undocumented)
--- a/plugins/home/src/api/VisitsStorageApi.ts
+++ b/plugins/home/src/api/VisitsStorageApi.ts
@@ -118,6 +118,21 @@ export class VisitsStorageApi implements VisitsApi {
     return visit;
   }
 
+  /**
+   * Updates the name of an existing visit through the visitsApi
+   */
+  async updateName(pathname: string, name: string): Promise<void> {
+    const visits: Visit[] = [...(await this.retrieveAll())];
+    const visitIndex = visits.findIndex(visit => visit.pathname === pathname);
+    if (visitIndex >= 0) {
+      visits[visitIndex] = {
+        ...visits[visitIndex],
+        name,
+      };
+      await this.persistAll(visits);
+    }
+  }
+
   private async persistAll(visits: Array<Visit>) {
     const storageKey = await this.getStorageKey();
     return this.storageApi.set<Array<Visit>>(storageKey, visits);
--- a/plugins/scaffolder/package.json
+++ b/plugins/scaffolder/package.json
@@ -69,7 +69,6 @@
     "@backstage/integration-react": "workspace:^",
     "@backstage/plugin-catalog-common": "workspace:^",
     "@backstage/plugin-catalog-react": "workspace:^",
-    "@backstage/plugin-home": "workspace:^",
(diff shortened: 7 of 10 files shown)


debsmita1:
tested this locally and it works fine

its-mitesh-kumar:
@benjdlambert please have a look.

plugins/scaffolder/src/translation.ts
```diff
@@ -353,6 +353,7 @@ export const scaffolderTranslationRef = createTranslationRef({
       pageContextMenu: {
         editConfigurationTitle: 'Edit Configuration',
       },
+      templateWithTitle: 'Template {{templateTitle}}',
```
schultzp2020: I like the changes! Although I think adding the `Template` prefix may be unnecessary because it can make the title redundant when the template includes `template` in the title as well.

its-mitesh-kumar: @schultzp2020 Thanks for the feedback! We have three options:  

1. Show `Template {{templateTitle}}`  
2. Show only `{{templateTitle}}` (no extra prefix)  
3. Conditionally show:  
   - `Template {{templateTitle}}` if the title does *not* already include "template"  
   - Just `{{templateTitle}}` if it already includes "template"  

As discussed in [#29963 (comment)](https://github.com/backstage/backstage/issues/29963#issuecomment-2897108485), we aligned on option 1.  
I’m open to reconsidering if you feel option 2 or 3 would provide a better user experience.

its-mitesh-kumar: @schultzp2020 Please let me know your final thoughts on this ?

acierto: The most flexible is `{{templateTitle}}` and then end user can decide if it has to be prefixed with Template or not.

its-mitesh-kumar:
@schultzp2020 @acierto @benjdlambert , If you have permission to reopen this PR , Please re-open it. I will be pushing the suggested change as discussed [above](https://github.com/backstage/backstage/pull/30732#discussion_r2302175392)  in few days.

plugins/scaffolder/src/translation.ts
its-mitesh-kumar: @acierto @schultzp2020 I’ve addressed your comments and updated both the PR description and the screen recording.

By default (without any additional configuration), it now displays **{{templateTitle}}**.

With the following configuration:

```yaml
scaffolder:
  prefixTemplateInTitle: false  # Displays: "My React App" (default)
  # OR
  prefixTemplateInTitle: true   # Displays: "Template My React App"
```

benjdlambert: I don't think we need any configuration here.

The idea would be that if someone wanted to change the formatting of the i18n, then they provide their own translations instead which is `Template {{templateTitle}}` instead of the default which is just `{{templateTitle}}` right?

its-mitesh-kumar:
Thanks @ciiay. 

@benjdlambert @schultzp2020 @acierto Could you please re-review it.

plugins/scaffolder/src/translation.ts
its-mitesh-kumar: Hey @benjdlambert,

Done! I've removed the configuration implementation as you suggested.

**Changes:**
- Default is now `{{templateTitle}}` (no "Template" prefix)
- Users can provide custom translations if they want `Template {{templateTitle}}` format
- Updated PR description with testing steps and screen recording

.changeset/petite-pears-tickle.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-scaffolder': patch
+'@backstage/plugin-home': patch
```
benjdlambert: I think we should break out these things into two different changesets. The `updateName` addition to `home` is a feature in itself.

plugins/home/src/api/VisitsApi.ts
```diff
@@ -144,6 +144,12 @@ export interface VisitsApi {
   enrichVisit?(
     visit: VisitInput,
   ): Promise<Record<string, any>> | Record<string, any>;
+  /**
+   * Update the name of an existing visit without affecting hit count.
+   * @param pathname - the pathname of the visit to update
+   * @param name - the new name for the visit
+   */
+  updateName(pathname: string, name: string): Promise<void>;
 }
```
iamEAP, line 153: I believe what you're trying to achieve with this new method can already be done via the `enrichVisit` method declared directly above this, or at least that was the intention behind it.

Instead of adding this new `updateName` method, would it make sense for the default implementation of this method to automatically detect when a scaffolder template is visited and enrich the `name` property of the visit based on template metadata?

its-mitesh-kumar: Thanks for the feedback, @iamEAP! I considered using `enrichVisit`, but there's a fundamental timing issue that makes it impractical here.

### The Problem

The visit gets saved **before** the template title is available:

1. User navigates to `/create/templates/default/test-template`
2. Home plugin immediately saves the visit (with generic name)
3. Scaffolder plugin fetches template schema asynchronously
4. Template title becomes available in the component

`enrichVisit` runs **during step 2** (inside the save operation), but the template title isn't available until **step 4** (after the component loads the manifest).

### Why `enrichVisit` Doesn't Work

To use `enrichVisit`, we would need to:
- Fetch the template schema again inside `enrichVisit` (duplicates the API call that scaffolder already makes)
- Or delay the visit save until the template loads (bad UX, potential missed visits)

Both options add unnecessary overhead and complexity.

### Why `updateName` is Better

The `updateName` method:
- Leverages the existing template fetch (no duplication)
- Updates the name only when it becomes available (after manifest loads)
- Keeps the logic in the scaffolder plugin (no coupling to home plugin)
- Doesn't block or slow down the initial visit save

It's essentially a "late enrichment" pattern for cases where metadata loads asynchronously after the initial page visit.

Let me know if you'd prefer a different approach!

iamEAP: Thanks @its-mitesh-kumar, that all seems sensible.

My only request then is that we keep the `updateName` method as optional in the interface (even if the default implementation provides the desired functionality), and make the call to it from scaffolder conditional on its existence.

Adding the method as required would be a breaking change for folks who've implemented their own `VisitsApi` and we want to avoid breaking things for them for the time being.

its-mitesh-kumar: Thanks @iamEAP! I've made the requested changes:

**Changes made:**
1. Made `updateName` optional in the `VisitsApi` interface (added `?`)
2. Updated the scaffolder to conditionally call it only if it exists (`if (!visitsApi?.updateName)`)
3. Updated the API report to reflect the optional method
4. Added a test case for backwards compatibility

This ensures no breaking changes for anyone who has implemented their own `VisitsApi` without the `updateName` method. The test confirms it works gracefully when the method doesn't exist.

Ready for re-review!

.changeset/petite-pears-tickle.md
its-mitesh-kumar: Updated the changeset @benjdlambert .

its-mitesh-kumar: @benjdlambert , Now we have approval from @iamEAP, too.

plugins/scaffolder/src/alpha/components/TemplateWizardPage/TemplateWizardPage.tsx
```diff
@@ -53,6 +56,37 @@ import { scaffolderTranslationRef } from '../../../translation';
 
 import { TemplateWizardPageContextMenu } from './TemplateWizardPageContextMenu';
 import { useFormDecorators } from '../../hooks';
+import { useLocation } from 'react-router-dom';
+
+const visitsApiRef = createApiRef<{
```
benjdlambert, line 61: This `apiRef` should be moved to being exported from `home-react` instead, and imported from there.

plugins/scaffolder/src/alpha/components/TemplateWizardPage/TemplateWizardPage.tsx
```diff
@@ -53,6 +56,37 @@ import { scaffolderTranslationRef } from '../../../translation';
...
+
+  useEffect(() => {
+    if (!manifest?.title) return;
+
+    const visitsApi = apiHolder.get(visitsApiRef);
+    if (!visitsApi?.updateName) {
+      return;
+    }
+
+    visitsApi
```
benjdlambert, line 80: I'm gonna push back again on the `updateName` implementation here.

I think @iamEAP called this out, but the `enrichVisit` is intended for this exact use case so we should use that.

The reason that this doesn't work currently in the repository is that there is no default implementation for the `enrichVisit`.

https://github.com/backstage/backstage/blob/8c1f36f0683f26fded26d6f218669bc1c3a9939d/plugins/home/src/api/VisitsStorageApi.ts#L141-L143

This just always returns `{}`.

I'd recommend that you move the `updateName` implementation to the `enrichVisit` in one way or another. If someone passes in another implementation of `enrichVisit` then use that, otherwise use the in memory one that you have for `updateName`.

plugins/scaffolder/src/translation.ts
christoph-jerolimov: Actually, I would recommend to prefix it maybe with the page title, "Create a new component {{templateTitle}}", but sorry this is just another opinion. Maybe the templateTitle alone is good enough as well.

plugins/scaffolder/src/alpha/components/TemplateWizardPage/TemplateWizardPage.tsx
its-mitesh-kumar: Hi @benjdlambert,

Thanks for the suggestion! I've been exploring the `enrichVisit` approach, but I'd like to explain why I believe we still need the `updateName` functionality.

### The Timing Problem

Looking at `VisitsStorageApi.ts`, `enrichVisit` is called during `save()`:

```typescript
// Line 171-172 in VisitsStorageApi.ts
const enrichedData = await this.enrichVisit(visit);
const enrichedVisit = { ...visit, ...enrichedData };
```

Here's the timeline of what happens:

```
Timeline:
────────────────────────────────────────────────────────────────────►

T1: User navigates to /create/templates/default/my-template
    │
    ├── VisitListener calls visitsApi.save()
    │   └── enrichVisit(visit) is called HERE
    │       └── Template title is UNKNOWN at this point!
    │       └── Returns {} (nothing to enrich with)
    │
    └── Visit saved with name = "create" ❌ WRONG NAME
    
T2: TemplateWizardPage renders
    │
    └── useTemplateParameterSchema() fetches manifest... (async)
    
T3: API returns manifest
    │
    └── NOW we know title = "My React App" ✓
    
    But enrichVisit already ran at T1!
    The visit is ALREADY saved with wrong name!
```

### Why enrichVisit Can't Solve This

`enrichVisit` only runs **during** `save()`. Once the visit is saved, `enrichVisit` cannot modify it. 

The template title comes from an **async API call** that completes **after** the visit is saved. 

### What We Need

To fix the already-saved visit, we must:
1. Read the visit from storage
2. Update the name
3. Write it back

This is exactly what `updateName` does. Even if we add an in-memory Map to `enrichVisit`, we still need this logic to update the stored visit.

So I think we need to keep `updateName`. 
Let me know if you think differently.

christoph-jerolimov:
Hi @its-mitesh-kumar, hi @iamEAP, @benjdlambert, @schultzp2020,

thanks all for looking into this, so far with different review results. I was asked to take a look as well and I have some additional ideas. I'm sorry... :laughing: 

### enrichVisit issue

As @its-mitesh-kumar shared earlier, the "problem" with the current `VisitApi #enrichVisit` is:

1. It was called from `VisitsStorageApi #save` function so its difficult to change (without a breaking change)
2. And, when I'm allowed to say it in some "hard words" ... "it does nothing". I mean that it allows users to call the option from the constructor, nothing else. 

The only function that benefits from `enrichVisit` currently is the `save` function and that could also directly call `enrichVisitImpl` from the `VisitsStorageApi` constructor.

I understand the idea to have a more generic function `enrichVisit` then multiple small functions like `updateName`. But currently this will not work without the following change(s). Maybe you agree and we should do this here? I think we should align on this before @its-mitesh-kumar updates this.

### Use `updateName` -or- improve `VisitApi` `enrichVisit`  (breaking change!?)

When I'm correct, the following is a "small" breaking change, also if a small one and since home page plugin is currently 0.8 its hopefully okay.

Update `VisitApi #enrichVisit(visit)` so that it's not only "used by the save function" to a function that can be used to update **_and persist_** the current visit. At the moment I hope its fine to pick up the "current visited page" from the persisted list based on the "current page path" as @its-mitesh-kumar has it already in his `updateName`.

To make it really explicit, this means:

1. call `this.transformPathname` as well
2. call `this.enrichVisitImpl()` when it is defined
5. and **new / breaking change** save the visit or enriched visit in the storageApi **without** increasing the hits count or timestamp of course. So that is the difference to save.

The `VisitApi #save` needs to be changed slightly so that is calls `enrichVisitImpl` directly instead of `VisitApi #enrichVisit`, of course.

**Why is this a breaking change?** Because a call to `enrichVisit` now updates the actual data. Might what people already expected before.

### Solve the problem "for all the plugins"

We want address this maybe in a follow-up... But the current PR adds `updateName` (or, hopefully, `enrichData`) just to the `scaffolder` plugin to solve the problem there. But there are many reasons why the page title might not be correct. And currently, for almost all cases except the software catalog it just uses the path name, or?

And even when it uses the `document.title` in `VisitListener #getVisitName` this title is mostly not correct because the new page (url change) has not updated the title because:

1. The plugin code is still loading (async, even when its just a few ms...)
2. The plugin needs to load something to render the title correctly (async)

So instead of changing the `TemplateWizardPage` my ask is, should we update the [Header](https://github.com/backstage/backstage/blob/80c8d9dab7dcce2455b7daf3630b1a4f4194942e/packages/core-components/src/layout/Header/Header.tsx#L24), [RoutedTabs](https://github.com/backstage/backstage/blob/80c8d9dab7dcce2455b7daf3630b1a4f4194942e/packages/core-components/src/components/TabbedLayout/RoutedTabs.tsx#L17), [EntityTabs](https://github.com/backstage/backstage/blob/80c8d9dab7dcce2455b7daf3630b1a4f4194942e/plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx#L17) and [TechDocsReaderPageHeader](https://github.com/backstage/backstage/blob/80c8d9dab7dcce2455b7daf3630b1a4f4194942e/plugins/techdocs/src/reader/components/TechDocsReaderPageHeader/TechDocsReaderPageHeader.tsx#L18) instead.

So that when the **actual page title** changes, we'll call automatically `enrichData` to save "a real title"?

The biggest problem here is the plugin to plugin dependencies of course.

Maybe its also possible to add a listener to a Helmet Provider? But I think this generic solution would be a much bigger step forward then just fixing this for the scaffolder > create page. Wdyt?

(Maybe its worth to create a new issue for this where we can have the analyticsApi in mind as well... :thinking: )

benjdlambert:
OK, let me wind back a bit, as I think I might have gotten a little confused as to the purpose of `enrichVisit`.

I thought that it was a public API that we could call in the scaffolder, which would update the previous visit, but it looks like that's not the case and that it's called before `save()`. It's not meant to be called using the `apiRef`.

That said however, I still think we can solve this with `enrichVisit`. Look at the example:

https://github.com/backstage/backstage/blob/35f3bbb4bec1f72d8f67b71e2533d6cd52c85e26/plugins/home/README.md#L458-L488

Can we not have a default implementation of `enrichVisit` that can take the `entityRef` and set the `title` as the value from `entityPresentationApi` using the `entityRef`?  Or does that still not work?

I will say there's also some other work going on around updating the Page Title #31478. I wonder if we should actually be looking at pushing that problem over there to solve for all plugins rather than just this one in isolation.

its-mitesh-kumar:
Hi @benjdlambert,
Your suggestion would work - we could configure `enrichVisit` to fetch the entity title via `catalogApi.getEntityByRef()`.

As I already talked about disadvantage  of using `enrichVisit` [here](https://github.com/backstage/backstage/pull/30732#discussion_r2564302850). Again , I am highlighting my concerns here :

**1. Extra API call for information we already have**

The scaffolder already fetches the template manifest via `useTemplateParameterSchema()`, which includes the title. With the `enrichVisit` approach, we'd be making a **second API call** to the catalog to get the same title.

**2. `save()` is blocked until the API resolves**

Looking at `VisitsStorageApi.ts`:
```typescript
const enrichedData = await this.enrichVisit(visit);  // ← Blocked here
```

The `save()` operation awaits `enrichVisit()`. If we add an API call inside, template visit will wait for that API to resolve before the visit can be saved.

With `updateName`:
- No extra API calls - we reuse the title from the manifest that's already being fetched
- `save()` completes immediately without blocking
- Visit is updated once the title is available

Let me know your thoughts!

CC: @iamEAP @christoph-jerolimov

christoph-jerolimov:
> OK, let me wind back a bit, as I think I might have gotten a little confused as to the purpose of `enrichVisit`.
> 
> I thought that it was a public API that we could call in the scaffolder, which would update the previous visit, but it looks like that's not the case and that it's called before `save()`. It's not meant to be called using the `apiRef`.
> 
> That said however, I still think we can solve this with `enrichVisit`. Look at the example:

:100: % agree! But the example is **not** using the enrichVisit api. It is using an option `enrichVisit` from `VisitsStorageApi`. Sounds similar, but it isn't.

I think we should change the existing "public API" on the VisitApi to do what we both/we all (??) expected: It should update  the latest visit data (= load change persist).

The option `enrichVisit` from `VisitsStorageApi` should (and will) work as before. This is what I've proposed in "improve VisitApi enrichVisit" above.

benjdlambert:
As discussed in the Framework SIG, going to pause this and try and solve this part of the #31478 and #31412 work which is ongoing.

its-mitesh-kumar:
Closing the PR refer : https://github.com/backstage/backstage/pull/30732#issuecomment-3659717348

PR #30735, opened 2025-08-04
[NFS] (frontend-plugin-api): inline `CommonAnalyticsContext` type

camilaibs:

As part of the NFS API Review, we decided to inline the `CommonAnalyticsContext` type and no longer export it. The reason for this decision is that it should be an internal type and not used by other packages.


.changeset/whole-hands-cut.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/frontend-plugin-api': minor
+---
+
+BREAKING: The "CommonAnalyticsContext" type is now internal and no longer exported from this package.
```
Rugvip:
```suggestion
**BREAKING**: The `CommonAnalyticsContext` has been removed, and inlined into `AnalyticsContextValue` instead.
```

Thinking this? Given that the type isn't actually around even internally at this point

Issue #29846, opened 2025-05-05
Feature: Add Infinispan as a Backend Cache Store Option


gustavolira:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
cache infinispan performance
```

### 🗃️ Project Area

Unknown

### 🔖 Need

Currently, Backstage offers several caching backends. Adding Infinispan provides users with another robust, distributed caching option, particularly beneficial for those already utilizing Infinispan within their infrastructure or seeking features like advanced clustering, persistence options, and cross-site replication capabilities offered by Infinispan.

### 📝 Proposal

Distributed caching is crucial for scaling Backstage backends, reducing load on underlying services (like the Catalog or external APIs), and improving overall performance. While Redis and Memcached are popular choices, Infinispan offers a compelling alternative, especially in environments leveraging JBoss/WildFly or Kubernetes/OpenShift where Infinispan operators provide seamless deployment and management.
This implementation follows the established pattern within Backstage's CacheManager (@backstage/backend-defaults) by:
Introducing a new infinispan configuration section under backend.cache.
Adding a store factory (createInfinispanStoreFactory) that utilizes the official [infinispan Node.js client](https://www.npmjs.com/package/infinispan) (Hot Rod protocol).
Supporting configuration options for connection (host, port), authentication (username, password, SASL mechanism), TLS, and default cache name.
Implementing the CacheService interface (get, set, delete) with proper handling for TTL (lifespan in Infinispan terms) and JSON serialization/deserialization.
Integrating with @backstage/backend-test-utils using testcontainers to allow integration testing against a real Infinispan instance.

Example Configuration (app-config.yaml):
```
backend:
  cache:
    store: infinispan
    # Optional: Default TTL for all cache entries
    # defaultTtl: 3600000 # 1 hour in milliseconds or use ISO 8601 format e.g., PT1H
    infinispan:
      # Required: Host and port of your Infinispan server
      host: localhost
      port: 11222 # Default Hot Rod port
      # Optional: Authentication credentials
      # username: admin
      # password: yourpassword
      # Optional: SASL mechanism (e.g., PLAIN, DIGEST-MD5, SCRAM-SHA-1)
      # saslMechanism: DIGEST-MD5
      # Optional: Specify a default cache name on the server
      # cacheName: backstage_cache
[...]

PR #29896, opened 2025-05-09
Add Infinispan as a Backend Cache Store Option

gustavolira:
## Hey, I just made a Pull Request!

# PR: feat: Infinispan Cache Store

## Hey, I just made a Pull Request!

This PR introduces Infinispan as a new cache store backend for Backstage, addressing [Feature Request #GH-29846](https://github.com/backstage/backstage/issues/29846) (assuming this is the relevant issue; please update if not).

Infinispan is a distributed, in-memory, key/value data store with optional schema. It can be used as an embedded Java library or as a language-independent service accessed remotely over a variety of protocols (e.g., Hot Rod). This integration leverages the official `infinispan-client` Node.js library to connect to an Infinispan server, providing a robust and scalable caching solution for Backstage deployments.

### Key Features Implemented:

*   **`InfinispanKeyvStore.ts`**: A Keyv-compatible store implementation that wraps the `infinispan-client` library. It translates standard Keyv operations (`get`, `set`, `delete`, `clear`) into the corresponding Infinispan client calls. It also handles connection management and error handling specific to Infinispan.
*   **`createInfinispanStore.ts`**: A factory function responsible for initializing the Infinispan client and creating an instance of `InfinispanKeyvStore`. This factory handles the configuration options for connecting to Infinispan, including server addresses, authentication, and other client-specific settings.
*   **Configuration**: Support for configuring the Infinispan cache store via `app-config.yaml`, allowing users to specify connection details, authentication mechanisms (e.g., PLAIN, DIGEST-MD5, SCRAM), and other relevant parameters.
*   **Integration with `CacheManager`**: The new Infinispan store is integrated into the existing `CacheManager` in `@backstage/backend-common`, allowing it to be selected as the cache backend through configuration. This ensures that plugins relying on the `CacheService` can seamlessly use Infinispan as their caching layer.
*   **Testing**: Comprehensive unit tests have been written for `InfinispanKeyvStore.ts` and `createInfinispanStore.ts` using Jest and appropriate mocking techniques. Integration tests have also been developed using Testcontainers to validate the store against a real Infinispan server instance, ensuring reliability and correctness under various scenarios.

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (or will do in a follow-up PR)
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
[...]

PR #30743, opened 2025-08-04
(feat) Add Infinispan as a new backend cache store

veenarm:

PR: Feat - add infinispan data grid as n available cache store.

This PR introduces Infinispan as a new cache store backend for Backstage.

resolves https://github.com/backstage/backstage/issues/29846

Infinispan is a distributed, in-memory, key/value data store with optional schema. It can be used as an embedded Java library or as a language-independent service accessed remotely over a variety of protocols (e.g., Hot Rod). This integration leverages the official infinispan-client Node.js library to connect to an Infinispan server, providing a robust and scalable caching solution for Backstage deployments.

I started from the previos PR https://github.com/backstage/backstage/pull/29896 and kind of tore it all down due to previous feedback received.

This one doesn't change the API's at all, and follows the same patters.

**Note:** Infinispan has some complicated type configs, so a single line `connectionString`  is not really doable, it also has backup site clusters etc you can configure. 

**Tests:** It turns out there was a broken test in here with a default redis option, added a minor patch to resolve it as part of this PR. 

Many tests from previous PR - However I will try to add more over the coming days as this is a Draft PR waiting for reviews first.

- [] Screenshots attached (for UI changes)

Side Note: Rebuild of #30731 due to rebasing issues.

@freben - Hi I know you had reviews of the previous PR so may be more familiar with this. Could I get a quick once over when you have some time. Deleted/Old PR - https://github.com/backstage/backstage/pull/29896

@gustavolira  - for your interest, I know you're working on Infinispan at the moment.


veenarm:
@freben  should we add area: core label? And can the issue get re opened?

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -231,23 +262,25 @@ export class CacheManager {
...
-        };
-        if (this.storeOptions?.cluster) {
-          // Create a Redis cluster
-          const cluster = createCluster(this.storeOptions?.cluster);
-          stores[pluginId] = new KeyvRedis(cluster, redisOptions);
-        } else {
-          // Create a regular Redis connection
-          stores[pluginId] = new KeyvRedis(this.connection, redisOptions);
-        }
+        if (this.storeOptions?.type === 'redis') {
```
freben: a bit surprised that this was needed - should it really be possible to end up with this explicit `createRedisStoreFactory` output function to be called, when the config type is NOT redis?

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -326,4 +361,96 @@ export class CacheManager {
         store,
       });
   }
+
+  private createInfinispanStoreFactory(): StoreFactory {
+    const stores: Record<string, InfinispanKeyvStore> = {};
+
+    return (pluginId, defaultTtl) => {
+      if (!stores[pluginId]) {
+        if (this.storeOptions?.type === 'infinispan') {
```
freben: if this comparison doesn't match, the store instance never gets made, and then the Keyv call below will explode.

Are you doing this check to make typescript happy with the correct storeOptions type? If so, is it better to either hard-cast (assuming that the config loader code did the right thing for sure and already checked it), or do like

```ts
if (this.storeOptions?.type !== 'infinispan') {
  // this should never happen
  throw new Error('Internal error: Wrong config type passed to infinispan factory')
}

File: packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts [hunk 3]
```diff
@@ -0,0 +1,225 @@
...
+          `Infinispan 'servers' configuration at ${storeConfigPath} is invalid.`,
+        );
+      }
+    } else {
+      logger?.error(
+        `Infinispan configuration at ${storeConfigPath} is missing the 'servers' definition.`,
+      );
+      throw new Error(
+        `Infinispan configuration at ${storeConfigPath} must define 'servers'.`,
+      );
```
freben, line 76: probably should not log AND throw - log if it chooses a fallback (or should silently disable itself), throw if it should block backend startup.

Also I would say you're fine doing just `get` and skip the `has` - `get` will throw an error that clearly explains what the missing key was etc.

packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    // Parse Servers & Clusters Configurations
+    if (infinispanConfig.has('servers')) {
+      const serversConfig = infinispanConfig.get('servers');
+      if (Array.isArray(serversConfig)) {
+        parsedOptions.servers = infinispanConfig.getConfigArray('servers').map(
+          serverConf =>
+            ({
+              host: serverConf.getString('host'),
+              port: serverConf.getNumber('port'),
```
freben: The config schema said this field is optional and has a default - you want `getOptionalNumber` if that's true and either give a default here or let the client deal with the default if that's better

packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts
```diff
@@ -0,0 +1,225 @@
...
+            } as InfinispanServerConfig),
+        );
+      } else if (typeof serversConfig === 'object' && serversConfig !== null) {
+        const serverConf = infinispanConfig.getConfig('servers');
+        parsedOptions.servers = {
+          host: serverConf.getString('host'),
+          port: serverConf.getNumber('port'),
```
freben: the config schema says _both_ of these are optional

packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts
```diff
@@ -0,0 +1,225 @@
...
+        } as InfinispanServerConfig;
+      } else {
+        logger?.error(
+          `Infinispan 'servers' configuration at ${storeConfigPath} must be an object or an array.`,
+        );
+        throw new Error(
+          `Infinispan 'servers' configuration at ${storeConfigPath} is invalid.`,
+        );
```
freben, line 71: same here, avoid log-and-throw

packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts
```diff
@@ -0,0 +1,225 @@
...
+    if (infinispanConfig.has('clusters')) {
+      const clustersConfig = infinispanConfig.getConfigArray('clusters');
+      const clusterOptions: InfinispanClusterConfig[] = [];
+      clustersConfig?.forEach(clusterConf => {
+        const name = clusterConf.getOptionalString('name');
+        const cluster: InfinispanClusterConfig = {
+          ...(name && { name }),
+          servers: clusterConf.getConfigArray('servers').map(serverConf => ({
+            host: serverConf.getString('host'),
+            port: serverConf.getNumber('port'),
```
freben: also optional

packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts
```diff
@@ -0,0 +1,225 @@
...
+          })),
+        };
+        clusterOptions.push(cluster);
+        behaviorOptions.clusters = clusterOptions;
```
freben: the array and foreach and push work fine, but you could also have done

```ts
behaviorOptions.clusters = infinispanConfig
  .getOptionalConfigArray('clusters')
  ?.map(clusterConf => {
     // ...
```

packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts
```diff
@@ -0,0 +1,225 @@
...
+      });
+    }
+
+    // Parse Default Options Start...
+    if (infinispanConfig.has('version')) {
+      const clientVersion = infinispanConfig.getString('version');
```
freben: I think in general things get a bit clearer if you just use optional more - like

```ts
const clientVersion = infinispanConfig.getOptionalString('version');
if (clientVersion) { // just a bare check like this is ok since empty strings aren't allowed
  // ...
```

that way you don't get to repeat (less opportunity for mistakes) the key strings. Same for a lot of places below in this file.

I think it's fair to say that we very rarely use `.has` - sometimes it's necessary, but it's pretty rare

packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts
```diff
@@ -0,0 +1,225 @@
...
+    if (mediaType === 'text/plain' || mediaType === 'application/json') {
+      behaviorOptions.mediaType = mediaType;
+    } else if (mediaType !== null && mediaType !== undefined) {
+      logger?.warn(
+        `Invalid Infinispan mediaType "${mediaType}" in config at ${storeConfigPath}.mediaType. Must be "text/plain" | "application/json". It will be ignored, and the client may use a default or fail.`,
+      );
+    }
+
+    if (infinispanConfig.has('maxRetries')) {
+      behaviorOptions.maxRetries = infinispanConfig.getNumber('maxRetries');
```
freben: since a key + undefined is normally pretty much indistinguishable from a missing key, you can often just do `behaviorOptions.maxRetries = infinispanConfig.getOptionalNumber('maxRetries');` without the `has` check.

freben (commented):
Alright this is all excellent! Just some nits really. Looks like a lot but it's mostly some ideas on how config reading could be a bit simplified. Really happy about this.

packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts
veenarm: They are, but there was nothing else in the server object, seemed like a forced if you add the servers object I assume you're entering in a host.
I'll make it commonm and use the default here rather than in the client as it may change at some point and our docs be out.

packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts
veenarm: I would say for infinispan configs you're almost always going to need some configuration settings, but we can find out what they are later.

packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts
veenarm: Still getting used to new syntax sugar :)

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
veenarm: Was needed to make typescript happy, will turn it into a negative at the top. Also has to be part of the return function for happyness.

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
veenarm: Yes, it was for typescript happyness :), Modified all three to be negatives at the top of the returns.

veenarm:
@freben  - I think I've captured all your comments, and agree, the only one that I didn't change/need was defaulting for the `clusters.servers` section. The server address and port are mandatory in the clusters section as it's implied you're going remotely.

The only part I'm still in the air about is the dataFormat defaults etc.. the 'get/set' in Keyv only take strings etc, there's nothing in those functions to tell me any params being passed (e.g. Is it JSON or Text) - thus I can't use the `JSON.parse(obj)` on the setters which is what it should use when storing as Json.

Was considering just removing media/formats and using common sense just set your cache to `text/plain` and be done with it, but for now all the config options are there. Noting that it still works if you set the value cache to json and put in text.

Hotfixed missing 'cache' name.. now the minimalist configuration is:

```yaml
backend:
  cache:
    store: infinispan
```

Most would never do this, servers and cacheName with authenticaiton are generally always going to be used though, however minimalist will work for local dev users who disable authentication and create the default 'cache'.

**Important Note**: Infinispan NodeJS Client won't create a cache for you if it doesn't exist, so expects it to be pre-created.

packages/backend-test-utils/package.json
```diff
@@ -69,6 +69,7 @@
     "cookie": "^0.7.0",
     "express": "^4.17.1",
     "fs-extra": "^11.0.0",
+    "infinispan": "^0.12.0",
```
freben: this can be reverted now i believe? (and re-run `yarn install`)

.changeset/tough-clubs-pick.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/backend-test-utils': minor
```
freben: could remove this line if we revert the package.json change

packages/backend-defaults/src/entrypoints/cache/providers/infinispan/InfinispanOptionsMapper.ts
freben: I don't think this was fixed - did you mean `serverConf.getOptionalNumber('port') ?? 11222` ?

freben (commented):
alright, almost there!

freben (approved):

PR #30750, opened 2025-08-04
`nfs(componentRefs)`: Reworking how Component Refs are defined and used

benjdlambert:
This updates the way that `componentRefs` are both defined and used, with a new concept of `SwappableComponents`.

```tsx
// defining
const Progress = createSwappableComponent({
  id: 'core.components.progress',
  // default implementation
  loader: () => () => <MuiProgress />
});


// using
<Progress />


// overriding
SwappableComponentBlueprint.make({
  params: define({
    component: Progress,
    loader: () => () => <MyCustomProgress />
  })
})
```

More docs to follow...


packages/frontend-internal/src/wiring/InternalComponentRef.ts
```diff
@@ -0,0 +1,34 @@
...
+export const OpaqueComponentRef = OpaqueType.create<{
+  public: ComponentRef;
+  versions: {
+    readonly version: 'v1';
+    readonly options: {
+      transformProps?: (props: object) => object;
+      loader?:
+        | (() => (props: object) => JSX.Element | null)
+        | (() => Promise<(props: object) => JSX.Element | null>);
+    };
```
Rugvip: Dunno if there's much value in wrapping these in `options` rather than putting them at the top tbh. Over time the options might evolve so this object might not mirror the options passed to `createComponentRef` either way

packages/frontend-plugin-api/src/components/makeComponentFromRef.tsx
```diff
@@ -0,0 +1,61 @@
...
+}): (props: ExternalComponentProps) => JSX.Element {
+  const { options } = OpaqueComponentRef.toInternal(ref);
+  const FallbackComponent = (p: JSX.IntrinsicAttributes) => (
+    <div data-testid={ref.id} {...p} />
+  );
+
+  const ComponentRefImpl = (props: ExternalComponentProps) => {
+    // todo(blam): use the component that's in the API ref instead if it's defined.
+    // otherwise use the fallback..
+    const api = useApi(componentsApiRef);
```
Rugvip: I think it'd be good to make sure they can be rendered without context too, perhaps something like https://github.com/backstage/backstage/blob/b1afb710b8cf23c17c112d4161e983419509b547/packages/frontend-plugin-api/src/analytics/useAnalytics.tsx#L23-L32

packages/frontend-plugin-api/src/components/makeComponentFromRef.tsx
```diff
@@ -0,0 +1,61 @@
...
+    if ('then' in ComponentOrPromise) {
+      const DefaultImplementation = lazy(() =>
+        ComponentOrPromise.then(c => {
+          return { default: c };
+        }),
+      );
+
+      return (
+        // todo: is this necessary? can we remove this?
+        <Suspense>
```
Rugvip: Yep I don't think we should render a `Suspense` here, but instead rely on that to exist further up the tree. It goes against the being able to render anywhere a bit though so not quite sure what to do about that.

packages/frontend-plugin-api/src/components/makeComponentFromRef.tsx
benjdlambert: Yeah, I started implementing this down this path so that you don't have to use `renderInTestApp` or whatever to get yourself an `APIContext`

packages/frontend-plugin-api/report.api.md
```diff
@@ -351,15 +351,28 @@ export { bitbucketAuthApiRef };
 export { bitbucketServerAuthApiRef };
 
 // @public (undocumented)
-export type ComponentRef<T extends {} = {}> = {
+export type ComponentRef<
```
Rugvip: Hmm, should we rename this type to `SwappableComponentRef` as well? It's not really intended to be used much

packages/frontend-plugin-api/report.api.md
```diff
@@ -351,15 +351,28 @@ export { bitbucketAuthApiRef };
...
+> = {
   id: string;
-  T: T;
+  TProps: TInnerComponentProps;
+  TExternalProps: TExternalComponentProps;
+  $$type: '@backstage/ComponentRef';
 };
 
 // @public
 export interface ComponentsApi {
```
Rugvip: `SwappableComponentsApi`? 🤔, Or even `SwappableComponentsResolutionApi` or `ComponentResolutionApi` ._.

packages/frontend-plugin-api/report.api.md
```diff
@@ -351,15 +351,28 @@ export { bitbucketAuthApiRef };
...
   // (undocumented)
-  getComponent<T extends {}>(ref: ComponentRef<T>): ComponentType<T>;
+  getComponent<
```
Rugvip: Thinking it might be best to call this one something like `getComponentLoader`?

packages/frontend-plugin-api/src/components/DefaultSwappableComponents.ts
```diff
@@ -19,25 +19,27 @@ import {
...
     id: 'core.components.notFoundErrorPage',
   });
 
-const coreErrorBoundaryFallbackComponentRef =
-  createComponentRef<CoreErrorBoundaryFallbackProps>({
-    id: 'core.components.errorBoundaryFallback',
+/**
+ * @public
+ */
+export const ErrorBoundary =
```
Rugvip: Hmm, still not sure about this naming. I feel it's too easy to confuse this with being an actually error boundary, but this is just the error display rendered by the boundary. I think I prefer simply keeping the old name

packages/frontend-plugin-api/src/components/createSwappableComponent.tsx
```diff
@@ -0,0 +1,132 @@
...
+  );
+
+  const ComponentRefImpl = (props: ExternalComponentProps) => {
+    const api = useComponentRefApi();
+    const ComponentOrPromise =
+      api?.getComponent<any>(ref)?.() ??
+      internalRef.loader?.() ??
+      FallbackComponent;
+
+    const innerProps = internalRef.transformProps?.(props) ?? props;
```
Rugvip: Hmm, could it be better to have the prop transform handled by the API implementation instead? That way the component loaders returned from the API will use the external props. The tradeoff is that it becomes impossible to completely override the API, but that might be something we want anyway?

Happy to leave that as a followup too though

packages/frontend-plugin-api/src/components/createSwappableComponent.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+    if ('then' in ComponentOrPromise) {
+      const DefaultImplementation = lazy(() =>
+        ComponentOrPromise.then(c => {
+          return { default: c };
+        }),
+      );
+
+      return (
+        <Suspense>
```
Rugvip: Imo we should figure out how to get rid of this suspense and leave it to the user code to determine where to suspend. Especially since this one doesn't have any fallback.

packages/frontend-plugin-api/src/components/createSwappableComponent.tsx
```diff
@@ -0,0 +1,132 @@
...
+    transformProps:
+      options.transformProps as (typeof OpaqueComponentRef.TInternal)['transformProps'],
+  });
+
+  const component = makeComponentFromRef({ ref });
+  Object.assign(component, { ref });
+
+  return component as {
+    ref: ComponentRef<TInnerComponentProps, TExternalComponentProps>;
+  } & ((props: object) => JSX.Element);
```
Rugvip: `| null`

packages/frontend-plugin-api/src/components/createSwappableComponent.tsx
Rugvip: Nvm, we need to be able to fall back to the default implementation in the component if the API is not available

plugins/app/src/extensions/SwappableComponentsApi.ts
```diff
@@ -15,34 +15,33 @@
...
  * Contains the shareable components installed into the app.
  */
-export const ComponentsApi = ApiBlueprint.makeWithOverrides({
-  name: 'components',
+export const SwappableComponentsApi = ApiBlueprint.makeWithOverrides({
+  name: 'swappable-components',
   inputs: {
-    components: createExtensionInput(
-      [createComponentExtension.componentDataRef],
-      { replaces: [{ id: 'app', input: 'components' }] },
```
benjdlambert, line 34: I ended up removing this, as it's not really something that's compliant now anyways. Thinking that it's not worth the effort of supporting both.

I'll update changesets to be friendly instead.

packages/frontend-plugin-api/report.api.md
```diff
@@ -922,6 +870,13 @@ export { ErrorApiErrorContext };
 
 export { errorApiRef };
 
+// @public (undocumented)
+export const ErrorDisplay: ((
+  props: CoreErrorDisplayProps,
```
freben: Why is there a `Core` on props in here, when the components themselves don't have that in their names?

packages/frontend-plugin-api/report.api.md
```diff
@@ -922,6 +870,13 @@ export { ErrorApiErrorContext };
...
+) => JSX.Element | null) & {
```
freben, line 861: Did you decide against having a `SwappableComponentType` type or so? Which basically encapsulates this slight mess.

packages/frontend-app-api/src/apis/implementations/SwappableComponentsApi/DefaultSwappableComponentsApi.tsx
```diff
@@ -0,0 +1,74 @@
...
+  SwappableComponentRef,
+  SwappableComponentsApi,
+  SwappableComponentBlueprint,
+} from '@backstage/frontend-plugin-api';
+import { OpaqueSwappableComponentRef } from '@internal/frontend';
+
+import { lazy } from 'react';
+
+/**
+ * Implementation for the {@linkComponentApi}
```
freben:
```suggestion
 * Implementation for the {@link SwappableComponentsApi}
```

packages/frontend-plugin-api/src/components/createSwappableComponent.tsx
```diff
@@ -0,0 +1,110 @@
...
+const useComponentRefApi = () => {
+  try {
+    return useApi(swappableComponentsApiRef);
+  } catch (e) {
+    return undefined;
+  }
+};
+
+/**
+ * Creates a SwappableComponent that can be used to render the component, optionally overriden by the app.
```
freben: search-and-replace `riden` -> `ridden` :) in all files

.github/vale/config/vocabularies/Backstage/accept.txt
```diff
@@ -467,6 +467,7 @@ Superfences
 superset
 supertype
 SVGs
+Swappable
```
freben:
```suggestion
swappable
```

we want it in all cases

.changeset/component-refs-breaking-frontend-plugin-api.md
Rugvip, line 1: This is a pretty long changeset, thinking it might be useful to move some of this here? https://backstage.io/docs/frontend-system/architecture/migrations

Realized that also goes for the defineParams and some of the other larger changes

.changeset/fuzzy-ducks-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/core-compat-api': patch
+---
+
+This is primarily an internal change that ensures legacy plugins continue to work with the new component system without requiring immediate migration.
```
Rugvip: Best mark this as a breaking change

PR #30762, opened 2025-08-05
docs/adrs: adr 15 for loader structure

Rugvip:
Bunch of discussion among @backstage/maintainers about this topic, aiming to solidify with an ADR

### DIFF
--- a/docs/architecture-decisions/adr015-jsx-loader-structure.md
+++ b/docs/architecture-decisions/adr015-jsx-loader-structure.md
@@ -62,6 +62,14 @@ This option is used in the same cases as the async component loader, but when th
 }
 ```
 
+Note that when consuming this loader we'll need to unconditionally wrap it with `React.lazy`. This is because you can't delay the call to `React.lazy` until rendering, because you're not allowed to call it within a render function. This means that we can't first call the loader to check whether the returned value is a promise or not, and we must instead unconditionally wrap it with `React.lazy`. Therefore the implementation of accepting one of these loaders as an option needs to look something like this:
+
+```tsx
+const LazyComponent = React.lazy(() =>
+  Promise.resolve(options.loader()).then(loaded => ({ default: loaded })),
+);
+```
+
 ## Consequences
 
 We will update all APIs for the new frontend system in the `@backstage/frontend-*` packages.
--- a/docs/architecture-decisions/adr015-jsx-loader-structure.md
+++ b/docs/architecture-decisions/adr015-jsx-loader-structure.md
@@ -1,0 +1,69 @@
+---
+id: adrs-adr015
+title: 'ADR015: Types and naming for element and component options'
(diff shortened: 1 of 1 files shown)


docs/architecture-decisions/adr015-jsx-loader-structure.md
```diff
@@ -0,0 +1,69 @@
...
+
+```tsx
+{
+  component: (props: { ... }) => JSX.Element | null
+}
+```
+
+### Async element loader
+
+This option is used when a simple asynchronous JSX element is provided. It is the preferred option when only producing a single instance and there is no need to pass properties to the component. This format simplifies the creation of closures for passing additional properties in the loader implmentation.
```
freben:
```suggestion
This option is used when a simple asynchronous JSX element is provided. It is the preferred option when only producing a single instance and there is no need to pass properties to the component. This format simplifies the creation of closures for passing additional properties in the loader implementation.
```

freben (approved):
👌

docs/architecture-decisions/adr015-jsx-loader-structure.md
```diff
@@ -62,6 +62,16 @@ This option is used in the same cases as the async component loader, but when th
 }
 ```
 
+Note that when consuming this loader we'll need to unconditionally wrap it with `React.lazy`. This is because you can't delay the call to `React.lazy` until rendering, because you're not allowed to call it within a render function. This means that we can't first call the loader to check whether the returned value is a promise or not, and we must instead unconditionally wrap it with `React.lazy`. Therefore the implementation of accepting one of these loaders as an option needs to look something like this:
+
+```tsx
+const LazyComponent = React.lazy(async () => {
+  return Promise.resolve(options.loader()).then(loaded => ({
```
freben: the `Promise.resolve` can't be needed, can it? Just makes it harder to read. And then why have the arrow function be async above even?

Maybe it's already clear enough, but could be worth explicitly saying that because of all this, this particular loader argument type can generally never be used as a component prop.

Rugvip: Yep it can be solved in other ways, and with this the `async` could indeed be removed.

Here's another option, do you think it's cleaner?

```tsx
const LazyComponent = React.lazy(async () => {
  const loaded = options.loader();
  if ('then' in loaded) {
    return loaded.then(component => ({ default: component });
  }
  return ({ default: loaded });
});
```

freben: Aha. It tries to cover both thenable and non-thenable return types of course. That makes sense. Yeah in that case I'd say leave it as-is. Or, sure, your proposal maybe saves a tick - I'll leave it up to you, maybe check `typeof loaded.then === 'function')` though. It still makes more sense to do it the original way arguably

PR #30765, opened 2025-08-05
frontend-plugin-api: avoid forwarding ConfigurableExtensionDataRef

Rugvip:
🧹, just a bit of type mapping to clean up API reports and docs


packages/frontend-plugin-api/src/wiring/createExtensionBlueprint.ts
```diff
@@ -522,7 +529,7 @@ export function createExtensionBlueprint<
   } as ExtensionBlueprint<{
     kind: TKind;
     params: TParams;
-    output: UOutput;
+    output: any;
```
freben, line 540: Could this have been `ExtensionDataRef[]` and remove the `as ExtensionDataRef[]` above?

Rugvip: It requires a bit more code than that, but I went with the verbose but more correct approach

PR #30766, opened 2025-08-05
feat(notifications): extension point to modify user resolving

drodil:

this change adds a new extension point method that can be used to pass a custom function that can modify how individual user entity references are resolved inside the notifications backend.

### DIFF
--- a/plugins/notifications-backend/src/service/router.ts
+++ b/plugins/notifications-backend/src/service/router.ts
@@ -26,6 +26,7 @@ import { v4 as uuid } from 'uuid';
 import { CatalogService } from '@backstage/plugin-catalog-node';
 import {
   NotificationProcessor,
+  NotificationRecipientResolver,
   NotificationSendOptions,
 } from '@backstage/plugin-notifications-node';
 import { InputError, NotFoundError } from '@backstage/errors';
@@ -52,6 +53,7 @@ import { getUsersForEntityRef } from './getUsersForEntityRef';
 import { Config, readDurationFromConfig } from '@backstage/config';
 import { durationToMilliseconds } from '@backstage/types';
 import pThrottle from 'p-throttle';
+import { parseEntityRef } from '@backstage/catalog-model';
 
 /** @internal */
 export interface RouterOptions {
@@ -64,6 +66,7 @@ export interface RouterOptions {
   signals?: SignalsService;
   catalog: CatalogService;
   processors?: NotificationProcessor[];
+  recipientResolver?: NotificationRecipientResolver;
 }
 
 /** @internal */
@@ -80,6 +83,7 @@ export async function createRouter(
     catalog,
     processors = [],
     signals,
+    recipientResolver,
   } = options;
 
   const WEB_NOTIFICATION_CHANNEL = 'Web';
@@ -639,6 +643,17 @@ export async function createRouter(
     opts: NotificationSendOptions,
     origin: string,
   ): Promise<Notification[]> => {
+    if (
+      users.find(u => {
+        const compound = parseEntityRef(u);
+        return compound.kind.toLocaleLowerCase('en-US') !== 'user';
+      })
+    ) {
+      throw new InputError(
+        'Invalid user entity reference provided in recipients',
+      );
+    }
+
     const { scope } = opts.payload;
     const uniqueUsers = [...new Set(users)];
     const throttled = throttle((user: string) =>
@@ -702,11 +717,16 @@ export async function createRouter(
       const entityRef = recipients.entityRef;
 
       try {
-        users = await getUsersForEntityRef(
-          entityRef,
-          recipients.excludeEntityRef ?? [],
-          { auth, catalog },
-        );
+        users = recipientResolver
+          ? await recipientResolver(
+              entityRef,
+              recipients.excludeEntityRef ?? [],
+            )
+          : await getUsersForEntityRef(
+              entityRef,
+              recipients.excludeEntityRef ?? [],
+              { auth, catalog },
+            );
       } catch (e) {
         throw new InputError('Failed to resolve notification receivers', e);
       }
--- a/plugins/notifications-node/src/extensions.ts
+++ b/plugins/notifications-node/src/extensions.ts
@@ -101,11 +101,27 @@ export interface NotificationProcessor {
 }
 
 /**
+ * NotificationRecipientResolver is a function that resolves the individual users to receive the notification
+ * based on the entity reference(s) and the excluded entity reference(s).
+ *
+ * The function should return a list of user entity references that should receive the notification.
+ *
+ * @public
+ */
+export type NotificationRecipientResolver = (
+  entityRef: string | string[] | null,
+  excludeEntityRefs: string | string[],
+) => Promise<string[]>;
+
+/**
  * @public
  */
 export interface NotificationsProcessingExtensionPoint {
   addProcessor(
     ...processors: Array<NotificationProcessor | Array<NotificationProcessor>>
+  ): void;
+  setNotificationRecipientResolver(
+    resolver: NotificationRecipientResolver,
   ): void;
 }
 
--- a/.changeset/twenty-shoes-tickle.md
+++ b/.changeset/twenty-shoes-tickle.md
@@ -1,0 +1,9 @@
+---
+'@backstage/plugin-notifications-backend': patch
+'@backstage/plugin-notifications-node': patch
+---
+
+A new extension point method was added that can be used to modify how the users receiving notifications
+are resolved. The function passed to the extension point should only return complete user entity references
+based on the notification target references and the excluded entity references. Note that the input is a list
+of entity references that can be any entity kind, not just user entities.
--- a/plugins/notifications-backend/src/plugin.ts
+++ b/plugins/notifications-backend/src/plugin.ts
@@ -22,6 +22,7 @@ import { createRouter } from './service/router';
 import { signalsServiceRef } from '@backstage/plugin-signals-node';
 import {
   NotificationProcessor,
+  NotificationRecipientResolver,
   notificationsProcessingExtensionPoint,
   NotificationsProcessingExtensionPoint,
 } from '@backstage/plugin-notifications-node';
@@ -33,6 +34,7 @@ class NotificationsProcessingExtensionPointImpl
   implements NotificationsProcessingExtensionPoint
 {
   #processors = new Array<NotificationProcessor>();
+  #recipientResolver: NotificationRecipientResolver | undefined = undefined;
 
   addProcessor(
     ...processors: Array<NotificationProcessor | Array<NotificationProcessor>>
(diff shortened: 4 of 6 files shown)


drodil:
Not stale

plugins/notifications-backend/src/service/router.ts
```diff
@@ -639,6 +643,17 @@ export async function createRouter(
     opts: NotificationSendOptions,
     origin: string,
   ): Promise<Notification[]> => {
+    if (
+      users.find(u => {
```
freben: `some`, or do find it and make it part of the error message

plugins/notifications-node/src/extensions.ts
```diff
@@ -100,13 +100,29 @@ export interface NotificationProcessor {
...
+ * NotificationRecipientResolver is a function that resolves the individual users to receive the notification
+ * based on the entity reference(s) and the excluded entity reference(s).
+ *
+ * The function should return a list of user entity references that should receive the notification.
+ *
+ * @public
+ */
+export type NotificationRecipientResolver = (
+  entityRef: string | string[] | null,
+  excludeEntityRefs: string | string[],
```
freben: This should probably be a single options argument (or perhaps having the entityRef as one argument and then an options object as the second, for extensibility. That's a pattern we use a lot. Also should `excludeEntityRefs` be optional? And should the return type also be an object with like `recipients: string[]`?

freben (commented):
What's the main use case here? What type of transform is useful to have? Expanding groups or so?

plugins/notifications-node/src/extensions.ts
drodil: The extension is following the same way `notificationsService` is using these. Changing those would of course be a breaking change. That's the reason why these are as they are :D

drodil:
> What's the main use case here? What type of transform is useful to have? Expanding groups or so?

Yes, to support the custom transformation of the receivers from groups or other entities.

plugins/notifications-node/src/extensions.ts
freben: Hm maybe I misunderstand something here, but what I meant is, something like

```ts
export type NotificationRecipientResolver = (options: {
  entityRefs: string[];
  excludeEntityRefs: string[];
  // room for more fields here in the future
}) => Promise<{ entityRefs: string[] }>
```

This is a net new interface right? So there's really no reason to NOT make it as nice and compliant as possible (and then maybe the code under the hood needs to make some light transformation to call it). The above receives normalized data (always array shaped, get an empty array instead of undefined), so that it's trivial to write a correct such resolver.

drodil: Actually we could just pass the entityRefs and handle exclusion in the notifications backend. But you are right, I will modify this a bit 👍

plugins/notifications-backend/src/service/router.ts
drodil: Removed this and ignored those instead throwing an error. Much better way for custom resolvers.

plugins/notifications-node/src/extensions.ts
drodil: I rewrote the extension point to take interface instead of a function and moved the default to `DefaultNotificationRecipientResolver` class instead. Probably more in line with other extension points as well + makes the interfaces much cleaner and expandable. There could even be another function to resolve broadcast notification receivers but that can be done later.

plugins/notifications-node/src/extensions.ts
```diff
@@ -100,13 +100,37 @@ export interface NotificationProcessor {
...
+ * NotificationRecipientResolver interface is used to resolve the individual
+ * users to receive the notification based on the entity reference(s) and
+ * the excluded entity reference(s).
+ *
+ * Using this extension point replaces the default recipient resolving
+ * functionality of the notifications backend.
+ *
+ * The `resolveNotificationRecipients` should return object with a list of user
+ * entity references that should receive the notification. In case the function
+ * returns other than user entity references, those are ignored.
```
freben, line 110: Reading this comment, I feel some of it should be on this type, some of it should be on `resolveNotificationRecipients`, and some of it should be on `setNotificationRecipientResolver`.

plugins/notifications-node/src/extensions.ts
```diff
@@ -100,13 +100,37 @@ export interface NotificationProcessor {
...
+ *
+ * @public
+ */
+export interface NotificationRecipientResolver {
+  resolveNotificationRecipients(options: {
+    entityRef: string | string[] | null;
+    excludeEntityRefs?: string | string[];
```
freben: one is nullable, one un undefineable 🤔 

haha just sayin, we don't use null much and either way, is there something to say about consistency between the two

plugins/notifications-backend/src/service/DefaultNotificationRecipientResolver.ts
```diff
@@ -0,0 +1,170 @@
...
+import { AuthService } from '@backstage/backend-plugin-api';
+import { CatalogService } from '@backstage/plugin-catalog-node';
+import { NotificationRecipientResolver } from '@backstage/plugin-notifications-node';
+
+const isUserEntityRef = (ref: string) =>
+  parseEntityRef(ref).kind.toLocaleLowerCase() === 'user';
+
+// Partitions array of entity references to two arrays; user entity refs and other entity refs
+const partitionEntityRefs = (refs: string[]): string[][] =>
+  refs.reduce(
```
freben: I have to admit that I'm a bit of a `reduce` hater, this seems like it'd be vastly simpler as a for loop and using push, instead of terrifying O(N^2) re-spreads over and over.

Essentially _any_ `...` inside a loop makes me stop and think.

plugins/notifications-backend/src/service/DefaultNotificationRecipientResolver.ts
```diff
@@ -0,0 +1,170 @@
...
+  implements NotificationRecipientResolver
+{
+  constructor(
+    private readonly auth: AuthService,
+    private readonly catalog: CatalogService,
+  ) {}
+
+  async resolveNotificationRecipients(options: {
+    entityRef: string | string[] | null;
+    excludeEntityRefs?: string | string[];
```
freben: again, as a consumer of this interface, I'd be more helped if this was just `excludeEntityRefs: string[]`; there's no real reason to subject everyone to this variety of possible types.

And the above one could just as easily have been `entityRefs: string[]` too right? It can be an empty array if there aren't any and if that has well defined semantics.

Everybody's code will be simpler for it, right?

plugins/notifications-backend/src/service/DefaultNotificationRecipientResolver.ts
```diff
@@ -0,0 +1,170 @@
...
+  }): Promise<{ userEntityRefs: string[] }> {
+    const { entityRef, excludeEntityRefs = [] } = options;
+
+    if (entityRef === null) {
+      return { userEntityRefs: [] };
+    }
+
+    const excluded = [excludeEntityRefs].flat();
+    const refsArr = [entityRef].flat();
```
freben: Yeah case in point ☝️

plugins/notifications-backend/src/service/DefaultNotificationRecipientResolver.ts
```diff
@@ -0,0 +1,170 @@
...
+          )
+          .map(r => r.targetRef);
+
+        const childGroupRefs = entity.relations
+          .filter(relation => relation.type === RELATION_PARENT_OF)
+          .map(r => r.targetRef);
+
+        let childGroupUsers: string[][] = [];
+        if (childGroupRefs.length > 0) {
+          const childGroups = await this.catalog.getEntitiesByRefs(
```
freben, line 121: Hm. Is this worth memoising per ref? Will it be common to re-fetch groups over and over again?

plugins/notifications-backend/src/service/DefaultNotificationRecipientResolver.ts
```diff
@@ -0,0 +1,170 @@
...
+            {
+              entityRefs: childGroupRefs,
+              fields,
+            },
+            { credentials: await this.auth.getOwnServiceCredentials() },
+          );
+          childGroupUsers = await Promise.all(childGroups.items.map(mapEntity));
```
freben, line 128: Just noting that a promise.all will have uncontrolled concurrency, this can start any number of simultaneous requests to the catalog, hammering it bad. Maybe this could all be DataLoader backend with a controlled page size etc or so

plugins/notifications-backend/src/service/DefaultNotificationRecipientResolver.ts
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        return [...groupUsers, ...childGroupUsers.flat(2)].filter(
```
freben: Should these be deduped?

plugins/notifications-backend/src/service/DefaultNotificationRecipientResolver.ts
```diff
@@ -0,0 +1,170 @@
...
+              isUserEntityRef(relation.targetRef),
+          )
+          .map(r => r.targetRef);
+
+        const childGroupRefs = entity.relations
+          .filter(relation => relation.type === RELATION_PARENT_OF)
+          .map(r => r.targetRef);
+
+        let childGroupUsers: string[][] = [];
+        if (childGroupRefs.length > 0) {
```
freben, line 120: Should `excludeEntityRefs` cover groups too? So that we never descend into certain groups at all?

plugins/notifications-backend/src/service/DefaultNotificationRecipientResolver.ts
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        if (isUserEntityRef(ownerRef)) {
+          if (excluded.includes(ownerRef)) {
+            return [];
+          }
+          return [ownerRef];
+        }
+
+        const owner = await this.catalog.getEntityByRef(ownerRef, {
```
freben, line 154: Same here, memoize / DataLoader?

freben (commented):
Sorry I keep pestering

plugins/notifications-backend/src/service/DefaultNotificationRecipientResolver.ts
drodil: Sooo many comments on functionality that was just moved 😢 Yes, it could probably use some improvements but I feel like it's time for another PR

plugins/notifications-backend/src/service/DefaultNotificationRecipientResolver.ts
drodil: All users are deduped in the end

drodil:
@freben improved based on some of the comments, but left the dataloader stuff out for now. This code has been here for a year now, but was just moved under the class from a floating function. Another PR for the loading improvements!

drodil:
I think the `NotificationSendOptions` should also be refactored, but that's a bit harder thing to do at this point as it's public interface: https://github.com/backstage/backstage/blob/master/plugins/notifications-node/src/service/DefaultNotificationService.ts#L28

freben (approved):
Yep that's fine! I didn't realise how much was just moved.

plugins/notifications-backend/src/service/DefaultNotificationRecipientResolver.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    const cachedEntityRefs = new Map<string, string[]>();
+
+    const mapEntity = async (entity: Entity | undefined): Promise<string[]> => {
+      if (!entity) {
+        return [];
+      }
+
+      const currentEntityRef = stringifyEntityRef(entity);
```
freben, line 89: I'll just note that this array is the same length and in the same order as the input (`filtered `), so the refs are already available while iterating (if grabbing the indices as you go)). But no worries

freben (approved):

PR #30769, opened 2025-08-05
populate missing profile fields from catalog entity

kc9ddi:

Auto-populate undefined user profile fields (picture, displayName, email) from the user's catalog entity when they are not available in the identity profile. This provides a fallback mechanism to ensure user profiles are more complete by leveraging catalog data.

This is particularly useful when logging in using the Guest auth provider, and a particular catalog user entity. This ensures that the user's profile will be populated with all appropriate fields from that catalog user entity, which improves the usability of the Guest provider for local testing.

Previously, only the user's picture was populated into the identity profile. This PR extends that behavior to also include the email and displayName properties.

Before:
image

image

After:
image

image

### DIFF
--- a/.changeset/nine-times-fold.md
+++ b/.changeset/nine-times-fold.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-user-settings': patch
+---
+
+For any undefined field in the user's profile (picture, displayName, email), populate it from the user's catalog entity
--- a/plugins/user-settings/src/components/useUserProfileInfo.ts
+++ b/plugins/user-settings/src/components/useUserProfileInfo.ts
@@ -37,15 +37,24 @@ export const useUserProfile = () => {
     const catalogProfile = (await catalogApi.getEntityByRef(
       backStageIdentity.userEntityRef,
     )) as unknown as UserEntity;
-    if (
-      identityProfile.picture === undefined &&
-      catalogProfile?.spec?.profile?.picture
-    ) {
-      identityProfile = {
-        ...identityProfile,
-        picture: catalogProfile.spec.profile.picture,
-      };
-    }
+    // Merge catalog profile fields into identity profile if missing
+    const fieldsToMerge: (keyof ProfileInfo)[] = [
+      'picture',
+      'displayName',
+      'email',
+    ];
+    identityProfile = fieldsToMerge.reduce((profile, field) => {
+      if (
+        profile[field] === undefined &&
+        catalogProfile?.spec?.profile?.[field]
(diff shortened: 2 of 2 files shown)


camilaibs (commented):
Hi @kc9ddi,
I consulted my colleagues more expert on this Backstage area, and they would have hoped that their sign in resolver would already have done that. We'd prefer to get consistent sign in response data from the start, in a single, dedicated location (the login resolver), rather than delegating it to the frontend to fill the gap. They said it might be a bit tricky to implement with the current authentication backend behavior, but it's worth exploring. What do you think?

kc9ddi:
> Hi @kc9ddi, I consulted my colleagues more expert on this Backstage area, and they would have hoped that your sign in resolver would already have done that. We'd prefer to get consistent sign in response data from the start, in a single, dedicated location (the login resolver), rather than delegating it to the frontend to fill the gap. They said it might be a bit tricky to implement with the current authentication backend behavior, but it's worth exploring. What do you think?

Hi @camilaibs , yes, I think I understand what you are suggesting. Does this PR address your concern? https://github.com/backstage/backstage/pull/30789

camilaibs:
> > Hi @kc9ddi, I consulted my colleagues more expert on this Backstage area, and they would have hoped that your sign in resolver would already have done that. We'd prefer to get consistent sign in response data from the start, in a single, dedicated location (the login resolver), rather than delegating it to the frontend to fill the gap. They said it might be a bit tricky to implement with the current authentication backend behavior, but it's worth exploring. What do you think?
> 
> Hi @camilaibs , yes, I think I understand what you are suggesting. Does this PR address your concern? #30789

I'm not sure if the guest sign-in resolver is the best place to do this. Let me include @freben and @Rugvip, they could give you a better guidance here.

kc9ddi:
> > > Hi @kc9ddi, I consulted my colleagues more expert on this Backstage area, and they would have hoped that your sign in resolver would already have done that. We'd prefer to get consistent sign in response data from the start, in a single, dedicated location (the login resolver), rather than delegating it to the frontend to fill the gap. They said it might be a bit tricky to implement with the current authentication backend behavior, but it's worth exploring. What do you think?
> > 
> > 
> > Hi @camilaibs , yes, I think I understand what you are suggesting. Does this PR address your concern? #30789
> 
> I'm not sure if the guest sign-in resolver is the best place to do this. Let me include @freben and @Rugvip, they could give you a better guidance here.

Sounds good - my understanding of this area is far from comprehensive, so I would appreciate some guidance from experts. Thanks!

PR #30772, opened 2025-08-05
rename sourcePath to routePath

freben:


Rugvip (approved):
🧹, stole it from @benjdlambert

Issue #30659, opened 2025-07-26
EventEmitter memory leak detected during catalog fetching


SonilPro:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
event emitter memory leak
```

### 🗃️ Project Area

Core Framework

### 🔗 External Integration

N/A

### 📝 Description & Context

Since the recent change of making "disableRelationsCompatibility" be true as default, a new error surfaced in our Backstage instance with 10k plus entities.
Search Collators and direct backend calls towards catalog/entities or other endpoints that return a response using [writeResponseData()](https://github.com/backstage/backstage/blob/770c2c76c13c2b99c7da24e6a750531597207119/plugins/catalog-backend/src/service/response/write.ts#L99) throw the following warning:
`MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 drain listeners added to [Gzip]. Use emitter.setMaxListeners() to increase limit`.

After doing some digging I noticed that this warning only appears when using http clients that buffer the response of the backend. Using something like curl which constantly writes out the response it receives doesn't trigger the warning. Things like postman and bruno as well as the http client backstage uses during search collating will trigger the warning, presumably because the response is being buffered.

What I believe is happening is that the server tries to send data faster than the client can receive it, which causes the internal buffer in node to fill up which in turn causes backpressure which trigger 'drain' listeners to get added.

### 👍 Expected Behavior

We can technically increase the maxListeners but I believe the error could present some underlying issues with the implementation. Ideally, node should not warn us about it.

### 📦 Reproduction Repo

https://github.com/backstage/backstage

### 🥾 Reproduction steps

1. Clone the main Backstage repo
2. Inside the [createRouter() entities endpoint](https://github.com/backstage/backstage/blob/770c2c76c13c2b99c7da24e6a750531597207119/plugins/catalog-backend/src/service/createRouter.ts#L216) for the catalog, replace everything inside the nested try catch (linked) to simulate lots of entities inside the catalog
```ts
          const entityMock = JSON.stringify({
            apiVersion: 'backstage.io/v1alpha1',
            kind: 'Component',
            metadata: {
              name: 'my-component',
              namespace: 'default',
              annotations: {
                'backstage.io/managed-by-location': 'url:https://example.com',
              },
            },
            spec: {
              type: 'service',
              owner: 'me',
[...]

PR #30786, opened 2025-08-06
handle event emitter leak when streaming responses

freben:
Closes https://github.com/backstage/backstage/issues/30659


plugins/catalog-backend/src/service/response/write.ts
```diff
@@ -83,40 +85,59 @@ export async function writeEntitiesResponse(options: {
...
+      res.off('finish', onClose);
+      resolve('closed');
+    }
+    function onDrain() {
+      drainPromise?.resolve('ok');
+      drainPromise = undefined;
+    }
+    res.on('drain', onDrain);
+    res.on('close', onClose);
+    res.on('finish', onClose);
```
Rugvip: What's the purpose of also listening to the `finish` event?

plugins/catalog-backend/src/service/response/write.ts
```diff
@@ -83,40 +85,59 @@ export async function writeEntitiesResponse(options: {
...
+  });
+
+  return async data => {
+    if (res.write(data, 'utf8')) {
```
Rugvip, line 136: What do you think about adding a guard before this that makes sure that `drainPromise` is `undefined`? Just to avoid any potential race

Rugvip (commented):
Nice! 🎉

plugins/catalog-backend/src/service/response/write.ts
freben: `close` only happens in corner cases with premature disconnection, while `finish` happens in the normal case. Since we do cleanup inside the callback, I thought it just prudent to always do the cleanup in either of those cases.

plugins/catalog-backend/src/service/response/write.ts
freben: I could instead change `drainPromise = createDeferred();` to be wrapped in `if (!drainPromise)`. That way, it'll always be safe and listening for whatever the next drain event is for any reason. Right? :)

plugins/catalog-backend/src/service/response/write.ts
Rugvip: Hmm, are you sure? I think `close` is emitted both for corner cases and on successful completion

plugins/catalog-backend/src/service/response/write.ts
freben: Nah you are right. Added as a guard instead.

plugins/catalog-backend/src/service/response/write.ts
freben: Aight, checked around a bit and it may actually be safe - removed!

PR #30789, opened 2025-08-06
Guest auth provider returns profile from catalog entity

kc9ddi:

Previously, if the Guest authentication provider was used, and was configured to use a certain catalog entity as the logged in user's entity, the logged in user's profile was not populated using fields from that user catalog entity.

Now, the user's profile is populated using fields from the user catalog entity, if those fields are populated. We read these values from the entity fields as documented here: https://backstage.io/docs/features/software-catalog/descriptor-format#specprofile-optional-1

### DIFF
--- a/.changeset/all-games-rule.md
+++ b/.changeset/all-games-rule.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-auth-backend-module-guest-provider': patch
 ---
 
-If the Guest user has a corresponding entity in the catalog, the logged in user profile is populated using the displayName, picture and email fields set in the user's catalog entity.
+If the Guest user has a corresponding entity in the catalog, the logged in user profile is populated using the `displayName`, `picture` and `email` fields set in the user's catalog entity.


.changeset/all-games-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-auth-backend-module-guest-provider': patch
+---
+
+If the Guest user has a corresponding entity in the catalog, the logged in user profile is populated using the displayName, picture and email fields set in the user's catalog entity.
```
freben:
```suggestion
If the Guest user has a corresponding entity in the catalog, the logged in user profile is populated using the `displayName`, `picture` and `email` fields set in the user's catalog entity.
```

plugins/auth-backend-module-guest-provider/src/authenticator.ts
```diff
@@ -16,23 +16,60 @@
...
 
 export const guestAuthenticator = createProxyAuthenticator({
-  defaultProfileTransform: async () => {
-    return { profile: {} };
+  defaultProfileTransform: async (
+    result: { userEntityRef?: string },
+    context,
+  ) => {
+    // Get the userEntityRef from the result (passed from authenticate function)
+    const userEntityRef = result.userEntityRef || DEFAULT_USER_ENTITY_REF;
```
freben, line 33: Hm. Some of this stuff is shared with [the sign in resolver](https://github.com/backstage/backstage/blob/03fcaf3666bfddf791fb1f6f67e33e315d649cea/plugins/auth-backend-module-guest-provider/src/resolvers.ts#L28).

Is there an argument to be made, for putting the default user entity ref in just one place, and then both this file and the resolver importing that?

Maybe there should even be

```ts
export function readResolverConfig(config: Config): {
  // NOT optional - will be the default if not set in config
  userEntityRef: string;
  // still optional
  ownershipEntityRefs?: string[];
  // a computed value based on the config flag and the env
  disabled: boolean
}
```

or something like that that reads the config and does any extra logic on top of that to return a well known source of truth for anyone who needs it?

plugins/auth-backend-module-guest-provider/src/authenticator.ts
```diff
@@ -16,23 +16,60 @@
...
+
+    try {
+      // Try to find the user in the catalog
+      const catalogUser = await context.findCatalogUser({
```
freben, line 37: It is a bit unfortunate to be forced to call the catalog twice (once in the resolver and once in here) - is there nothing we can do about that? 🤔 It's no big deal really and sign-ins don't happen all that often, but wanted to check

PR #30792, opened 2025-08-06
Adjusted the dialog API types to have more sensible defaults

freben:
I guessed the void type (that is compatible with undefined) might be the right choice here. We already have usages of `close()` without args and I think this turns out looking OK.


Rugvip (approved):
Yep! 👍

PR #30800, opened 2025-08-06
BUI - Remove CSS import from Select component

jabrks:

Backstage UI styles are written with CSS and compiled together into a single file which can be loaded once by the consuming application at the top-level. The Select component was importing its own styles in the component, which is a departure from the pattern used elsewhere.

This was also causing noise in our other tests, as attempting to load styles in JS is not something that had been accounted for in the test environment.
```
Error: Could not parse CSS stylesheet
```

This PR removes the import as the styles are still included as part of the main Backstage UI stylesheet where they belong.


packages/ui/css/styles.css
Rugvip, line 1: Afaik these changes should be removed, we had a bit of trouble with the CSS being out of sync but that was fixed in https://github.com/backstage/backstage/pull/30812

Rugvip (commented):
Yep! 👍 

Just the CSS build thing

Rugvip (approved):
:shipit:

PR #30804, opened 2025-08-06
Gitlab/throttle

JohannesWill:


plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
```diff
@@ -207,6 +219,19 @@ export class GitlabDiscoveryEntityProvider implements EntityProvider {
       );
     }
 
+    const currentTime = new Date().toISOString();
+    // Get the last refresh time from cache
+    const lastRefreshKey = this.getCacheKey('lastRefresh');
+    const lastRefresh = await this.cache.get(lastRefreshKey);
```
kc9ddi: Is the cache strictly in memory, or is it persisted in the database? If it is in-memory, we'd end up with a very large sync operation whenever Backstage restarts. Perhaps that's not a problem considering the other efforts being made here to reduce API calls and improve rate limiting handling, but wanted to ask.

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +89,62 @@ export class GitLabIntegration implements ScmIntegration {
...
+      }
+
+      if (response.status === 429) {
+        this.counter429?.add(1);
+      }
+
+      retries++;
+
+      // Calculate delay from Retry-After header or use exponential backoff
+      const retryAfter = response.headers.get('Retry-After');
```
kc9ddi: Note that the project, groups and users APIs don't respond with this header (at least on requests to Gitlab.com, perhaps it is different for self-hosted Gitlab). It might be worth adding this information as a comment, since it will cause us to fall back to the exponential backoff case.

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +89,62 @@ export class GitLabIntegration implements ScmIntegration {
...
+      const delay = retryAfter
+        ? parseInt(retryAfter, 10) * 1000
+        : Math.min(500 * Math.pow(2, retries - 1), 10000); // Exponential backoff, cap at 10 seconds
+
+      await new Promise(resolve => setTimeout(resolve, delay));
```
kc9ddi: Can we log when we need to retry a request, and why? And what timeout we are using?

packages/integration/src/gitlab/GitLabIntegration.ts
kc9ddi: (Gitlab docs: https://docs.gitlab.com/user/gitlab_com/#rate-limits-on-gitlabcom)

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
JohannesWill: Backstage can be configured to use one of three cache stores: memory, memcache / redis:
Default is in memory
https://backstage.io/docs/overview/architecture-overview/#cache

kc9ddi: It might not be worth the added complexity, but I wonder if we should persist this information to the database, instead of using a cache. If the cache gets bounced, then we might get a surge in traffic to Gitlab. If we persist this flag to the database, we'd avoid that risk.

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +89,62 @@ export class GitLabIntegration implements ScmIntegration {
...
+    }
+    this.queue.pop();
+    // This should never be reached due to the loop condition, but TypeScript needs it
+    throw new Error('Max retries exceeded');
+  }
+
+  private setupTelemetry() {
+    this.gauge = this.meter.createObservableGauge('throttler.queue.size', {
+      description: 'Anzahl der Requests in der Throttle-Queue',
```
0xts: We should use English here to ensure it's widely understandable, imo.

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +89,62 @@ export class GitLabIntegration implements ScmIntegration {
...
+    });
+
+    this.gauge.addCallback(observableResult => {
+      observableResult.observe(this.queue.length);
+    });
+
+    this.counter429 = this.meter.createCounter('throttler.429.count', {
+      description: 'Anzahl der 429 Responses',
```
0xts: We should use English here to ensure it's widely understandable, imo.

0xts (approved):

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
JohannesWill: Good idea, but sad to say, I don't know how to do that.

packages/integration/src/gitlab/GitLabIntegration.ts
JohannesWill: Need advice on how to log in here.

.changeset/fluffy-owls-act.md
```diff
@@ -0,0 +1,46 @@
+---
+'@backstage/backend-defaults': patch
+'@backstage/integration': minor
+'@backstage/plugin-catalog-backend-module-gitlab': patch
+---
+
+Add configurable throttling and retry mechanism for GitLab integrations
```
schultzp2020: Can you please shorten the changeset to one or two sentence summarization and then move the remaining section to the README? 🙏

packages/backend-defaults/src/entrypoints/urlReader/lib/GitlabUrlReader.ts
```diff
@@ -15,8 +15,6 @@
  */
 
 // NOTE(freben): Intentionally uses node-fetch because of https://github.com/backstage/backstage/issues/28190
```
schultzp2020: @freben do we still need to worry about this?

packages/integration/src/gitlab/core.ts
```diff
@@ -105,62 +109,91 @@ export function buildProjectUrl(
...
+  // Check relative path exist and replace it if it's the case.
+  if (relativePath) {
+    repo = repo.replace(relativePath, '');
   }
+
+  // Remove leading slash
+  return repo.replace(/^\//, '');
 }
+
+// // Convert
```
schultzp2020: What's your plan with this section of code?

.changeset/fluffy-owls-act.md
```diff
@@ -0,0 +1,46 @@
...
+- **GitLabIntegration**: Refactored with modular fetch strategy creation supporting plain fetch, retry logic, and throttling
+- **Configuration parsing**: Enhanced `readGitLabIntegrationConfig` with new throttling options and validation
+- **OpenTelemetry integration**: Added metrics for monitoring throttling behavior and 429 responses
+- **GitLab URL reader**: Updated to extract `signal` parameter for proper request cancellation support
+- **Comprehensive testing**: Added extensive test coverage for all fetch strategies and configuration combinations
+
+**Technical implementation:**
+
+- Uses `p-throttle` library for rate limiting with configurable requests per minute
+- Implements exponential backoff retry logic with respect for `Retry-After` headers
```
freben: spell checker errors that are legitimate words, need those words added to `.github/vale/config/vocabularies/Backstage/accept.txt` in the same PR to make the spell checker happy.

But yes none of this should probably go in the changeset

packages/backend-defaults/src/entrypoints/urlReader/lib/GitlabUrlReader.ts
freben: Yes! I do believe so. Commented below

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +81,92 @@ export class GitLabIntegration implements ScmIntegration {
...
     return replaceGitLabUrlType(url, 'edit');
   }
+
+  fetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+    return this.fetchImpl(input, init);
+  }
+
+  private createFetchStrategy(): FetchFunction {
+    let fetchFn: FetchFunction = async (url, options) => {
+      return fetch(url, { ...options, mode: 'same-origin' });
```
freben, line 79: Note https://github.com/backstage/backstage/pull/30804/files#r2302183242 - you can NOT use global fetch safely for gitlab as far as I know in node packages. We had a whole little incident related to that. You sure about this one being possible to use as imported from node? Or do you have to add the ability to inject a "base" fetch implementation from the outside to this class?

packages/backend-defaults/src/entrypoints/urlReader/lib/GitlabUrlReader.ts
```diff
@@ -109,12 +105,7 @@ export class GitlabUrlReader implements UrlReaderService {
     }
 
     if (response.ok) {
-      return ReadUrlResponseFactory.fromNodeJSReadable(response.body, {
-        etag: response.headers.get('ETag') ?? undefined,
-        lastModifiedAt: parseLastModified(
-          response.headers.get('Last-Modified'),
-        ),
-      });
+      return ReadUrlResponseFactory.fromResponse(response);
```
freben, line 113: hm if we do switch to using node-fetch again as per the other comments, will this have to be reverted?

packages/backend-defaults/src/entrypoints/urlReader/lib/GitlabUrlReader.ts
```diff
@@ -335,74 +320,48 @@ export class GitlabUrlReader implements UrlReaderService {
...
+
+      // Extract project path directly instead of making API call
       const relativePath = getGitLabIntegrationRelativePath(
         this.integration.config,
       );
+
+      let projectPath = namespaceAndProject;
+      // Check relative path exist and remove it if so
+      if (relativePath) {
+        projectPath = projectPath.replace(relativePath, '');
```
freben: Is this really safe? What if that string appears twice or in unexpected parts of the project path?

packages/integration/config.d.ts
```diff
@@ -306,6 +306,24 @@ export interface Config {
...
+       * @default []
+       * @visibility frontend
+       */
+      retryStatusCodes?: number[];
+      /**
+       * Rate limit for requests per minute. Set to -1 to disable throttling.
+       * @default -1
+       * @visibility frontend
+       */
+      limitPerMinute?: number;
```
freben: maybe a bit unclear name of the field, limit of what?

packages/integration/report.api.md
```diff
@@ -588,8 +588,8 @@ export function getGitilesAuthenticationUrl(
 export function getGitLabFileFetchUrl(
   url: string,
   config: GitLabIntegrationConfig,
-  token?: string,
-): Promise<string>;
+  _token?: string,
```
freben, line 599: is it a bit suspicious that this became an unused parameter?

packages/integration/report.api.md
```diff
@@ -588,8 +588,8 @@ export function getGitilesAuthenticationUrl(
...
+): string;
```
freben: I think this was made a promise intentionally, because then you know that you can perform remote operations _if needed_ now or in the future, without breaking the contract.

Us not having promises in our url resolution elsewhere has led to severe longstanding issues :)

packages/integration/report.api.md
```diff
@@ -770,6 +772,9 @@ export type GitLabIntegrationConfig = {
   token?: string;
   baseUrl: string;
   commitSigningKey?: string;
+  maxRetries: number;
```
freben: another option here would be to collect all related fields in one object.

```ts
retry?: {
  enabled?: boolean;
  maxAttempts?: boolean;
  maxAttemptsPerMinute?: boolean;
}
```

or something like that, and same for the config

packages/integration/src/gitlab/GitLabIntegration.test.ts
```diff
@@ -17,7 +17,21 @@
...
 import { GitLabIntegration, replaceGitLabUrlType } from './GitLabIntegration';
 
+// Mock pThrottle to make testing easier
+jest.mock('p-throttle', () => {
+  return jest.fn(() => (fn: any) => fn);
+});
+
+// Mock global fetch
+const mockFetch = jest.fn();
+global.fetch = mockFetch;
```
freben: if you give the ability to inject a "base" fetch as per other comments you won't need this.

We commonly use msw also to not have to mock fetches at all; that's the best path typically. Lots of prior art in the repo to search for

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -13,13 +13,16 @@
...
  */
-
 import { basicIntegrations, defaultScmResolveUrl } from '../helpers';
 import { ScmIntegration, ScmIntegrationsFactory } from '../types';
 import {
   GitLabIntegrationConfig,
   readGitLabIntegrationConfigs,
 } from './config';
+import pThrottle from 'p-throttle';
+import { metrics, Counter, ObservableGauge, Meter } from '@opentelemetry/api';
```
freben: Hm. I wonder if we have any prior use of otel in common packages. 🤔 maybe it's ok

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +81,92 @@ export class GitLabIntegration implements ScmIntegration {
...
+  fetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+    return this.fetchImpl(input, init);
+  }
+
+  private createFetchStrategy(): FetchFunction {
+    let fetchFn: FetchFunction = async (url, options) => {
+      return fetch(url, { ...options, mode: 'same-origin' });
+    };
+
+    // Early return for plain fetch (no enhancements needed)
```
freben: not really necessary is it? just risks introducing bugs, it seems to me

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +81,92 @@ export class GitLabIntegration implements ScmIntegration {
...
+  }
+
+  private withRetry(fetchFn: FetchFunction): FetchFunction {
+    return async (url, options) => {
+      let retries = 0;
+
+      while (retries <= this.integrationConfig.maxRetries) {
+        this.queue.push(() => {});
+
+        const response = await fetchFn(url, options);
```
freben: This seems to me that it needs a try/catch. And especially [for AbortError](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason) it should be immediately short circuited and return

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +81,92 @@ export class GitLabIntegration implements ScmIntegration {
...
+
+    return fetchFn;
+  }
+
+  private withRetry(fetchFn: FetchFunction): FetchFunction {
+    return async (url, options) => {
+      let retries = 0;
+
+      while (retries <= this.integrationConfig.maxRetries) {
+        this.queue.push(() => {});
```
freben: This seems surprising. :)

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +81,92 @@ export class GitLabIntegration implements ScmIntegration {
...
+
+        // Calculate delay from Retry-After header or use exponential backoff
+        // Note: project, groups and users APIs may not respond with Retry-After header
+        // depending on the used Gitlab instance
+        const retryAfter = response.headers.get('Retry-After');
+        const delay = retryAfter
+          ? parseInt(retryAfter, 10) * 1000
+          : Math.min(500 * Math.pow(2, retries - 1), 10000); // Exponential backoff, cap at 10 seconds
+
+        await new Promise(resolve => setTimeout(resolve, delay));
```
freben: ideally also short circuit when an abort signal (if passed in) fires, otherwise there's up to 10 seconds of delay until the caller gets notified

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +81,92 @@ export class GitLabIntegration implements ScmIntegration {
...
+      }
+      this.queue.pop();
+      // This should never be reached due to the loop condition, but TypeScript needs it
+      throw new Error('Max retries exceeded');
```
freben: You can get around this by having `let response: Response` outside of the loop and returning it here. It'd be much cleaner; then you don't need to do the check inside the loop

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +81,92 @@ export class GitLabIntegration implements ScmIntegration {
...
+
+        if (
+          !this.integrationConfig.retryStatusCodes.includes(response.status) ||
+          retries >= this.integrationConfig.maxRetries
+        ) {
+          this.queue.pop();
+          return response;
+        }
+
+        if (response.status === 429) {
```
freben: shouldn't this be checked always, above the previous `if`?

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -37,7 +40,23 @@ export class GitLabIntegration implements ScmIntegration {
...
+  private readonly queue: Array<() => void> = [];
+  private readonly meter: Meter;
+  private gauge?: ObservableGauge;
+  private counter429?: Counter;
+
+  private readonly fetchImpl: FetchFunction;
+
+  constructor(private readonly integrationConfig: GitLabIntegrationConfig) {
+    this.meter = metrics.getMeter(
+      `gitlab-integration-${this.integrationConfig.host}`,
```
freben: shouldn't the host be a label instead? hard to summarize over these

freben (commented):
getting these things right is actually pretty tricky

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
fb33: Hi, if you use Redis as CacheManager then you'll have some persistence capacity. See https://redis.io/docs/latest/operate/oss_and_stack/management/persistence/

packages/integration/src/gitlab/GitLabIntegration.ts
fb33: It might be more readable if the queue were a queue of request, and we pushed url.

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
```diff
@@ -216,6 +241,7 @@ export class GitlabDiscoveryEntityProvider implements EntityProvider {
         ...(!this.config.includeArchivedRepos && { archived: false }),
         ...(this.config.membership && { membership: true }),
         ...(this.config.topics && { topics: this.config.topics }),
+        ...(lastRefreshTime && { last_activity_after: lastRefreshTime }),
```
JohannesWill: I just figured out, that last_activity_at is not updated in all scenarios when a repo is modified.
So I will remove this delta update from the PR and only keep the Throtteling

freben:
Just double checking - there was still work to do on this PR, right?

packages/integration/report.api.md
JohannesWill: It's not needed anymore, kept only for compatibility. Should be removed?

packages/integration/src/gitlab/GitLabIntegration.ts
JohannesWill: This was only for otlp => removed

packages/integration/src/gitlab/GitLabIntegration.ts
JohannesWill: Adding mode: 'same-origin' to the options will mitigate the problem.

I deployed the patch in our system. It has been running for several weeks with this change without any errors.

packages/integration/src/gitlab/GitLabIntegration.ts
JohannesWill: removed otlp

packages/integration/src/gitlab/GitLabIntegration.ts
JohannesWill: removed otlp

freben:
Remember `yarn build:api-reports` when making API changes!

JohannesWill:
> Remember `yarn build:api-reports` when making API changes!

@freben 
I don't understand what's happening here. It says that plugin-scaffolder-alpha/report-alpha.api.md has changed. I haven't made any changes in this area. The newly generated plugin-scaffolder-alpha/report-alpha.api.md only contains reorders. I don't understand where this comes from and how it interacts with my changes.

packages/integration/src/gitlab/GitLabIntegration.ts
JohannesWill: * An AbortError should immediately short-circuit the whole operation and bubble up so the caller can handle cancellation. Wrapping and swallowing it here would change that semantics

* The withRetry wrapper is explicitly designed to retry based on configured HTTP status codes. Retry of other errors (e.g. network exceptions like TypeError) was not in scope

packages/backend-defaults/src/entrypoints/urlReader/lib/GitlabUrlReader.ts
JohannesWill: see https://github.com/backstage/backstage/pull/30804#discussion_r2410742494

freben:
scheduling this for review in the coming days

packages/integration/report.api.md
freben: Yeah I think so. Since this is a public API it's kind of confusing to be able to pass in a token and it doesn't get used.

You could break out a separate changeset file for this package, and there call out this particular breaking change. That's of potential interest to users since this package is a pretty widely shared and used one.

packages/integration/src/gitlab/core.ts
```diff
@@ -105,62 +109,33 @@ export function buildProjectUrl(
...
-// Convert
-// from: https://gitlab.com/groupA/teams/teamA/subgroupA/repoA/-/blob/branch/filepath
-// to:   The project ID that corresponds to the URL
-export async function getProjectId(
+/**
+ * Extracts the project path from a GitLab URL
+ * from: https://gitlab.com/groupA/teams/teamA/subgroupA/repoA/-/blob/branch/filepath
+ * to:   groupA/teams/teamA/subgroupA/repoA
+ */
+export function extractProjectPath(
```
freben, line 117: Is it the case that _all_ versions of gitlab can handle paths instead of IDs? I don't recall the history here, but it's been the case in the past that people updated integration code and then it for example worked fine in on-prem but not in a cloud version, or worked on newer installations but not old ones. I just wonder if this `getProjectId` was here for good reason. Do you know?

@jamieklassen @OpenSourceZombie you did some of this - feedback?

plugins/catalog-backend-module-gitlab/src/lib/client.ts
```diff
@@ -378,6 +397,69 @@ export class GitLabClient {
     return true;
   }
 
+  async getCommitsTouchingFile(
```
freben: I don't think this is being used at all?

packages/integration/src/gitlab/core.ts
JohannesWill: Maybe not all, but all active maintained.
https://archives.docs.gitlab.com/16.11/ee/api/projects.html#get-single-project

However, it was already included in version v12.10 (release 2020-07-06)

packages/integration/src/gitlab/GitLabIntegration.ts
```diff
@@ -62,6 +69,97 @@ export class GitLabIntegration implements ScmIntegration {
...
+      let attempt = 0;
+      for (;;) {
+        response = await fetchFn(url, options);
+        // If response is not retryable, return immediately
+        if (!retryStatusCodes.includes(response.status)) {
+          break;
+        }
+
+        // If this was the last allowed attempt, return response
+        if (attempt++ >= maxRetries) {
```
schultzp2020, line 124: @freben should we add the logger here so we can log the failed attempts? Unfortunately it looks like the integrations do not have access to it.

PR #30813, opened 2025-08-07
Arrange most `create[Specialized]App` options into an `advanced` field

freben:
This helps highlight that it's a collection of rarely-used/needed options.


packages/frontend-app-api/report.api.md
```diff
@@ -42,7 +42,7 @@ export type CreateSpecializedAppOptions = {
   extensionFactoryMiddleware?:
     | ExtensionFactoryMiddleware
     | ExtensionFactoryMiddleware[];
-  flags?: {
+  advanced?: {
```
Rugvip, line 41: Idea is to move for example `extensionFactoryMiddleware` too, thinking it's worth doing that all at once?

freben: Ah excellent - yeah of course, updated

PR #30816, opened 2025-08-07
[NFS] frontend-plugin-api: make app spec plugin required

camilaibs:

As part of the NFS API Review effort, the `AppNodeSpec.plugin` property is now required to simplify the spec validation.

### DIFF
--- a/.changeset/thirty-eagles-run.md
+++ b/.changeset/thirty-eagles-run.md
@@ -2,4 +2,4 @@
 '@backstage/frontend-app-api': minor
 ---
 
-**BREAKING:** The `AppNodeSpec.plugin` property is now required.
+The `AppNodeSpec.plugin` property is now required.
--- a/.changeset/thirty-eagles-run.md
+++ b/.changeset/thirty-eagles-run.md
@@ -1,0 +1,5 @@
+---
+'@backstage/frontend-app-api': minor
+---
+
+**BREAKING:** The `AppNodeSpec.plugin` property is now required.
--- a/packages/frontend-app-api/src/tree/resolveAppNodeSpecs.ts
+++ b/packages/frontend-app-api/src/tree/resolveAppNodeSpecs.ts
@@ -14,7 +14,11 @@
  * limitations under the License.
  */
 
-import { Extension, FrontendFeature } from '@backstage/frontend-plugin-api';
+import {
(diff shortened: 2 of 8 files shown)


packages/frontend-app-api/src/tree/resolveAppNodeSpecs.ts
```diff
@@ -26,6 +30,8 @@ import {
 // eslint-disable-next-line @backstage/no-relative-monorepo-imports
 import { toInternalExtension } from '../../../frontend-plugin-api/src/wiring/resolveExtensionDefinition';
 
+export const rootPlugin = createFrontendPlugin({ pluginId: 'root' });
```
Rugvip: This is the tricky bit, is it fine that this shows up in tests and as the plugin of the root extension?

We also considered detecting and using the `app` plugin if one is available, and otherwise fall back to a placeholder `app` plugin.

Kinda related, I'm thinking it's probably best to create this instance within `resolveAppNodeSpecs` rather than here, just to avoid the global state.

Rugvip: Thinking a bit more about this I think my favorite solution for this is to default it to the `app` plugin but fall back to a new `app` plugin if it doesn't exist. I'm feeling that's gonna have the least impact on the overall structure, rather than introducing a new `root` plugin

camilaibs: I just applied this change. The only thing that seems inconsistent to me is that the built-in extension isn't returned when we try to get it by calling `getExtension` from the `appPlugin`. But as we discussed, it might be better to remove this method from the node spec anyway.

.changeset/thirty-eagles-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/frontend-app-api': minor
+---
+
+**BREAKING:** The `AppNodeSpec.plugin` property is now required.
```
Rugvip: This isn't actually a breaking change, since the app node spec is only consumed. It ofc is breaking from the app-api's point of view, but we view them as one

camilaibs: Fixed 🙌🏻

PR #30819, opened 2025-08-07
frontend-plugin-api: add TsDoc for some of the most important API surface

Rugvip:
📝

### DIFF
--- a/packages/frontend-plugin-api/src/wiring/createExtension.ts
+++ b/packages/frontend-plugin-api/src/wiring/createExtension.ts
@@ -308,16 +308,16 @@ export type ExtensionDefinition<
  *
  * @example
  *
- *```ts
- *const myExtension = createExtension({
- *  name: 'example',
- *  attachTo: { id: 'app', input: 'root' },
- *  output: [coreExtensionData.reactElement],
- *  factory() {
- *    return [coreExtensionData.reactElement(<h1>Hello, world!</h1>)];
- *  },
- *});
- *```
+ * ```ts
+ * const myExtension = createExtension({
+ *   name: 'example',
+ *   attachTo: { id: 'app', input: 'root' },
+ *   output: [coreExtensionData.reactElement],
+ *   factory() {
+ *     return [coreExtensionData.reactElement(<h1>Hello, world!</h1>)];
+ *   },
+ * });
+ * ```
  *
  * @public
  */
--- a/packages/frontend-plugin-api/src/wiring/createExtensionBlueprint.ts
+++ b/packages/frontend-plugin-api/src/wiring/createExtensionBlueprint.ts
@@ -419,7 +419,7 @@ function unwrapParams<TParams extends object>(
  * blueprint, removing a lot of the boilerplate and complexity that is otherwise
  * needed to create an extension.
  *
- * Each blueprint has its own `kind` that helps identity and group the
+ * Each blueprint has its own `kind` that helps identify and group the
  * extensions that have been created with it. For example the
  * {@link PageBlueprint} has the kind `'page'`, and extensions created with it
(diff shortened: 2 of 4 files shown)


packages/frontend-plugin-api/src/wiring/createExtension.ts
```diff
@@ -281,7 +286,41 @@ export type ExtensionDefinition<
...
+ *
+ * Extension definitions that are created with this function can be installed in
+ * a Backstage app via a {@link FrontendPlugin} or {@link FrontendModule}.
+ *
+ * For more details on how extensions work, see the
+ * {@link https://backstage.io/docs/frontend-system/architecture/extensions | documentation for extensions}.
+ *
+ * @example
+ *
+ *```ts
```
freben: does this work without the extra space of indentation? same in some other files below

packages/frontend-plugin-api/src/wiring/createExtensionBlueprint.ts
```diff
@@ -403,9 +404,52 @@ function unwrapParams<TParams extends object>(
...
+ * {@link https://backstage.io/docs/frontend-system/architecture/extension-blueprints | documentation for extension blueprints}
+ * in the frontend system documentation.
+ *
+ * Extension blueprints make it much easier for users to create new extensions
+ * for your plugin. Rather than letting them use {@link createExtension}
+ * directly, you can define a set of parameters and default factory for your
+ * blueprint, removing a lot of the boilerplate and complexity that is otherwise
+ * needed to create an extension.
+ *
+ * Each blueprint has its own `kind` that helps identity and group the
```
freben:
```suggestion
 * Each blueprint has its own `kind` that helps identify and group the
```

packages/frontend-plugin-api/src/wiring/createFrontendModule.ts
```diff
@@ -45,7 +45,45 @@ export interface InternalFrontendModule extends FrontendModule {
...
 
-/** @public */
+/**
+ * Creates a new module instance that can be installed in a Backstage app.
+ *
+ * @remarks
+ *
+ * Modules are used to add or override extensions for an existing plugin. If a
+ * module provides an extension with the same ID as one provided by the plugin,
+ * the extension provided by the module will always take presedence.
```
freben:
```suggestion
 * the extension provided by the module will always take precedence.
```

packages/frontend-plugin-api/src/wiring/createFrontendModule.ts
```diff
@@ -45,7 +45,45 @@ export interface InternalFrontendModule extends FrontendModule {
   readonly featureFlags: FeatureFlagConfig[];
 }
 
-/** @public */
+/**
+ * Creates a new module instance that can be installed in a Backstage app.
```
freben: Not sure about the word "instance" here and in the corresponding plugin function. Seems to me that it should be removed.

packages/frontend-plugin-api/src/wiring/createFrontendPlugin.ts
```diff
@@ -135,7 +135,36 @@ export interface PluginOptions<
   info?: FrontendPluginInfoOptions;
 }
 
-/** @public */
+/**
+ * Creates a new plugin instance that can be installed in a Backstage app.
```
freben: instance again

packages/frontend-plugin-api/src/wiring/createFrontendModule.ts
Rugvip: Yep can do

packages/frontend-plugin-api/src/wiring/createExtension.ts
Rugvip: seems like it, but let's add it

PR #30834, opened 2025-08-08
feat: scaffolder logs create time use local time zone

mario-mui:

### DIFF
--- a/plugins/scaffolder-common/src/ScaffolderClient.ts
+++ b/plugins/scaffolder-common/src/ScaffolderClient.ts
@@ -41,6 +41,7 @@ import {
   ScaffolderTask,
 } from './api';
 import { DefaultApiClient, TaskStatus, TypedResponse } from './schema/openapi';
+import { DateTime } from 'luxon';
 
 /**
  * An API to interact with the scaffolder backend.
@@ -328,6 +329,16 @@ export class ScaffolderClient implements ScaffolderApi {
           }
 
           const logs = (await response.json()) as LogEvent[];
+
+          logs.forEach(log => {
+            if (log.createdAt) {
+              log.createdAt = DateTime.fromISO(log.createdAt, {
+                zone: 'utc',
+              })
+                .setZone(DateTime.local().zoneName)
+                .toISO() as string;
+            }
+          });
 
           for (const event of logs) {
             after = Number(event.id);
--- a/.changeset/floppy-ravens-push.md
+++ b/.changeset/floppy-ravens-push.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-common': patch
+---
+
+scaffolder logs create time use local time zone
--- a/plugins/scaffolder-common/package.json
+++ b/plugins/scaffolder-common/package.json
@@ -69,6 +69,7 @@
     "@types/json-schema": "^7.0.9",
     "cross-fetch": "^4.0.0",
     "json-schema": "^0.4.0",
+    "luxon": "^3.0.0",
     "uri-template": "^2.0.0",
     "zen-observable": "^0.10.0"
   },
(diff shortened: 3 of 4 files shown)


mario-mui:
@benjdlambert ping

plugins/scaffolder-common/src/ScaffolderClient.ts
```diff
@@ -329,6 +330,16 @@ export class ScaffolderClient implements ScaffolderApi {
 
           const logs = (await response.json()) as LogEvent[];
 
+          logs.forEach(log => {
```
acierto, line 333: It can be done without an extra loop, and `createdAt` is not optional field in LogEvent.

```
const toLocalTimeZone = (isoTime: string) =>  DateTime.fromISO(isoTime, {
                zone: 'utc',
              })
                .setZone(DateTime.local().zoneName)
                .toISO() as string;

for (const event of logs) {
   event.createdAt  = toLocalTimeZone(event.createdAt);
   ....
}
```

acierto (changes requested):

freben:
Ben is out for two weeks on PTO.

As a general observation, would it be best if this was left untouched, and instead the rendering of log lines used a new component `<Timestamp defaultTimezone="local" value={row.createdAt} format="short" />` or whatever like that and then that thing can have a user selectable zone in the UI somewhere that changes the presentation everywhere?

A lot of the time when looking at logs, especially ones from a bit back in time, I distinctly want them in UTC because it so happens that almost everything else in related systems already displays UTC by default.

drodil:
> As a general observation, would it be best if this was left untouched, and instead the rendering of log lines used a new component `<Timestamp defaultTimezone="local" value={row.createdAt} format="short" />` or whatever like that and then that thing can have a user selectable zone in the UI somewhere that changes the presentation everywhere?
> 
> A lot of the time when looking at logs, especially ones from a bit back in time, I distinctly want them in UTC because it so happens that almost everything else in related systems already displays UTC by default.

This is indeed needed component and export from `core-components`. Additionally, the setting for the `Settings` page to modify this would be a great help + showing the original time in a tooltip.

drodil:
Something like:

```tsx
import { DateTime } from 'luxon';
import { storageApiRef, useApi } from '@backstage/core-plugin-api';
import useObservable from 'react-use/esm/useObservable';
import { useMemo } from 'react';

export type TimestampProps = {
  value: string | number | Date;
  defaultTimezone?: string;
  format?: Intl.DateTimeFormatOptions;
  locale?: string;
};

export const Timestamp = (props: TimestampProps) => {
  const {
    value,
    defaultTimezone = 'setting',
    format,
    locale
  } = props;
  // TODO: Settings should use the same storage bucket for saving these values
  const storageApi = useApi(storageApiRef).forBucket('settings');
  const tzSnapshot = useObservable(
    storageApi.observe$<string>('timezone'),
    storageApi.snapshot('timezone'),
  );
  // TODO: This should also affect translations i18n in the future
  const localeSnapshot = useObservable(
    storageApi.observe$<string>('locale'),
    storageApi.snapshot('locale'),
  );

  const tz: string | undefined = useMemo(() => {
    if (defaultTimezone && defaultTimezone !== 'setting') {
      return defaultTimezone;
    }
    if (tzSnapshot.presence === 'absent') {
      return undefined;
    }
    return tzSnapshot.value as string;
  }, [tzSnapshot, defaultTimezone]);

  const loc: string | undefined = useMemo(() => {
    if (locale) {
      return locale;
    }
    if (localeSnapshot.presence === 'absent') {
      return undefined;
    }
    return localeSnapshot.value as string;
  }, [localeSnapshot, locale]);

  const date = new Date(value);
  if (isNaN(date.getTime())) {
    // TODO: Translate this as well
    return <>Invalid Date</>;
  }

  const dt = DateTime.fromJSDate(date);
  dt.setZone(tz);
  return dt.toLocaleString(format, { locale: loc });
};
```

Of course, the locale should also affect the used translations in the app, AND there should be a settings card to modify these in the storage.

PR #30836, opened 2025-08-08
feat: added some docs around swappable

benjdlambert:


docs/frontend-system/building-plugins/06-swappable-components.md
```diff
@@ -0,0 +1,98 @@
...
+
+```tsx title="in @internal/plugin-example-react"
+import { createSwappableComponent } from '@backstage/frontend-plugin-api';
+
+export const ExampleSwappableComponent = createSwappableComponent({
+  name: 'example',
+
+  // This is a loader for loading the default implementation of the component when there's no overriden
+  // implementation created with `SwappableComponentBlueprint`.
+  // It can be sync like below, but is can also be async like `loader: () => import('./DefaultImplementation').then(m => m.default)`.
```
Rugvip:
```suggestion
  // It can be sync like below, but is can also be async like `loader: () => import('./DefaultImplementation').then(m => m.DefaultImplementation)`.
```

Would typically be what we recommend

docs/frontend-system/building-plugins/06-swappable-components.md
```diff
@@ -0,0 +1,98 @@
...
+  loader: () => (props: { name: string }) =>
+    <div>Your name is {props.name}</div>,
+
+  // This is an optional function that can be used to transform the props of the external component.
+  // If it's not provided, the props of the external component will be passed through unchanged.
```
Rugvip, line 30: I think it's good to explain that you typically won't want to define this when first creating the component, but it's instead something that can help keep the external API stable as it evolves

docs/frontend-system/building-plugins/06-swappable-components.md
```diff
@@ -0,0 +1,98 @@
...
+
+<ExampleSwappableComponent firstName="John" lastName="Doe" />;
+```
+
+## Overriding a Swappable Component
+
+In order to override a Swappable Component, you need to create a `SwappableComponentBlueprint` and install it with the `app` plugin.
+
+```tsx title="in packages/app/src/App.tsx"
+import { ExampleSwappableComponent } from '@internal/plugin-example-react';
```
Rugvip: Perhaps a real example here instead, like overriding the `Progress` component

docs/frontend-system/building-plugins/06-swappable-components.md
```diff
@@ -0,0 +1,98 @@
...
+
+...
```
Rugvip: dunno what this is a placeholder for, can probably remove right? 😁

docs/frontend-system/building-plugins/06-swappable-components.md
```diff
@@ -0,0 +1,98 @@
...
+
+            // or sync:
+            // loader: () => MyNewImplementation
+          })
+        })
+      ]
+    }),
+  ...
+  ]
+})
```
Rugvip: Bunch of imports missing for these, and I think we've realized they help people a lot

Rugvip (approved):
Nice! 👍 

Pre-approved, but a couple of things to iron out

docs/frontend-system/building-plugins/06-swappable-components.md
```diff
@@ -0,0 +1,129 @@
...
+                loader: () => MyCustomProgress,
+              }),
+          }),
+      ],
+    }),
+    // OR: Add another extension but with the same component ID:
+    appPlugin.withOverrides({
+      extensions: [
+        SwappableComponentBlueprint.make({
+          name: 'core.components.progress',
```
Rugvip, line 84: Looking at this I realize the `.components.` is a bit redundant. Maybe we could simplify it all the way down to `core-progress`? Both for the name and the component ID. That'd fit alright for plugin libraries too I think, e.g. `scaffolder-template-card`.

docs/frontend-system/building-plugins/06-swappable-components.md
```diff
@@ -0,0 +1,129 @@
...
+          params: defineParams =>
+            defineParams({
+              component: Progress,
+              loader: () => MyCustomProgress,
+            }),
+        }),
+      ],
+    }),
```
Rugvip, line 106: I'm feeling this is the better one to recommend tbh, because if you're overriding directly via the `appPlugin` instance it's safer to use the `.getExtension` approach. You can't do that if you want to ship it as a module though, which makes this example useful as well. So keep the first example, skip the 2nd, and keep this.

Rugvip (approved):
Nice! 👍 

If we update the IDs + names I think that's best handled in a followup

docs/frontend-system/building-plugins/06-swappable-components.md
benjdlambert: Cool - makes sense!

docs/frontend-system/building-plugins/06-swappable-components.md
benjdlambert: Yep - nice!

docs/frontend-system/building-plugins/06-swappable-components.md
```diff
@@ -0,0 +1,129 @@
...
+  ],
+});
+```
+
+## Default Swappable Components
+
+Currently there is only three different built-in Swappable Components that you can replace the implementations of, and these live in `@backstage/frontend-plugin-api`. They are as follows:
```
freben, line 113:
```suggestion
Currently there are only three different built-in Swappable Components that you can replace the implementations of, and these live in `@backstage/frontend-plugin-api`. They are as follows:
```

docs/frontend-system/building-plugins/06-swappable-components.md
```diff
@@ -0,0 +1,129 @@
...
+```tsx title="in @internal/plugin-example"
+import { ExampleSwappableComponent } from '@internal/plugin-example-react';
+
+<ExampleSwappableComponent firstName="John" lastName="Doe" />;
+```
+
+## Overriding a Swappable Component
+
+In order to override a Swappable Component, you need to create a `SwappableComponentBlueprint` and install it with the `app` plugin.
+There's two different ways to add extensions to the `app` plugin, both are documented below in an example of overriding the `Progress` Swappable Component.
```
freben, line 52:
```suggestion
There are two different ways to add extensions to the `app` plugin, both are documented below in an example of overriding the `Progress` Swappable Component.
```

PR #30839, opened 2025-08-08
Backwards compatibility of `componentsApi` and `SwappableComponents`

benjdlambert:
Added a backwards compatible `componentsApi` implementation is also provided the app plugin which uses `SwappableComponentsApi` as the implementation. 

Will be removed in the future, but here for plugins using the old `componentRefs`


Rugvip (approved):
:shipit:

plugins/app/src/extensions/LegacyComponentsApi.ts
```diff
@@ -0,0 +1,58 @@
...
+  ErrorDisplay,
+  NotFoundErrorPage,
+  Progress,
+} from '@backstage/frontend-plugin-api';
+import { ComponentType } from 'react';
+
+/**
+ * This is the old component API that has been replaced by the new SwappableComponentsApi.
+ *
+ * This backwards compatibility implementation exists to avoid breaking older plugins.$
```
freben:
```suggestion
 * This backwards compatibility implementation exists to avoid breaking older plugins.
```

PR #30840, opened 2025-08-08
nfs: only allow swappable extensions to be installed in the app plugin

benjdlambert:
We don't want random plugins providing these extensions, they should be only provided from the app plugin using an override or a module.

Will add additional docs to this and maybe a link to those docs too.


Rugvip (approved):
Yep! 👍 :shipit:

Issue #30851, opened 2025-08-09
proxy-backend get err rewrite rule when backend baseurl set subpath


amintong:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
proxy backend rewrite rule subpath
```

### 🗃️ Project Area

Core Framework

### 🔗 External Integration

N/A

### **Problem**  
When proxying a request to `/api/proxy/wedeployhook/api/get_controller` using `backend-proxy`, the upstream server receives an incorrect path:  
`/wedeployhook/api/api/proxy/wedeployhook/api/get_controller`  
instead of the expected `/wedeployhook/api/get_controller`.

### **Configuration**
```
backend
  baseUrl as "http://amintong.dev.com/webackstage"
proxy:
  endpoints:
    /wedeployhook/api: http://amintong.dev.com/wedeployhook/api

```
### **Log**
in proxy log get a rewrit rule "^/webackstage/api/proxy/wedeployhook/api/?" ~> "/"  which is not correct.
```
[backend]: 2025-08-09 18:12:55 proxy info [HPM] Proxy created: /wedeployhook/api  -> http://amintong.dev.com/wedeployhook/api 
[backend]: 2025-08-09 18:12:55 proxy info [HPM] Proxy rewrite rule created: "^/webackstage/api/proxy/wedeployhook/api/?" ~> "/" 
```

* pathprefix 
https://github.com/backstage/backstage/blob/88480e4ef22f5f0158fd0b82dbc516b9c166f53a/plugins/proxy-backend/src/service/router.ts#L308C29-L308C65
```
  const externalUrl = await options.discovery.getExternalBaseUrl('proxy');
  const { pathname: pathPrefix } = new URL(externalUrl);

```
* HPM config
https://github.com/backstage/backstage/blob/88480e4ef22f5f0158fd0b82dbc516b9c166f53a/plugins/proxy-backend/src/service/router.ts#L145C1-L147C7
```
    fullConfig.pathRewrite = {
      [`^${pathPrefix}${routeWithSlash}?`]: '/',
    };
```

### 👍 Expected Behavior

http-proxy-middler rewrite rule don't have subpath webackstage
```
[backend]: 2025-08-09 18:12:55 proxy info [HPM] Proxy created: /wedeployhook/api  -> http://amintong.dev.com/wedeployhook/api 
[...]

PR #30853, opened 2025-08-10
fix(proxy-backend): correct rewrite rule to avoid extra subpath 

amintong:
#30851

#### Summary

This PR fixes an issue where the proxy-backend rewrite rule incorrectly included extra subpaths (such as `/webackstage`) in the proxied request path. The rewrite rule now uses `getBaseUrl` instead of `getExternalBaseUrl` to determine the correct path prefix, ensuring that requests are rewritten as expected (e.g., `/api/proxy/...`).

#### Details

- Changed the path prefix logic in `createRouter` to use `getBaseUrl`, so the rewrite rule does not contain unwanted subpaths.
- Added a changeset for a patch release.
- After this fix, upstream services will receive the correct request path, and proxy logs will show the expected rewrite rule.

---

Let me know if you need a more detailed breakdown or want to add test/usage notes!


plugins/proxy-backend/src/service/router.ts
```diff
@@ -292,8 +292,8 @@ export async function createRouter(
     logger: options.logger,
   };
 
-  const externalUrl = await options.discovery.getExternalBaseUrl('proxy');
-  const { pathname: pathPrefix } = new URL(externalUrl);
+  const baseUrl = await options.discovery.getBaseUrl('proxy');
```
Rugvip: This will use the internal URL instead of the external one, which isn't necessarily correct either.

Given that we [hardcode `'/api/{pluginId}'`](https://github.com/backstage/backstage/blob/ed818545d4df4f8a69de48833535b9de1d1d16ef/packages/backend-defaults/src/entrypoints/httpRouter/httpRouterServiceFactory.ts#L67) nowadays, I think it'd be best to just do that here too, as in simply set the `pathPrefix` to `'/api/proxy'` instead.

.changeset/late-squids-feel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-proxy-backend': minor
```
Rugvip:
```suggestion
'@backstage/plugin-proxy-backend': patch
```

Switching to a patch since this is a non-breaking fix

PR #30857, opened 2025-08-11
frontend-defaults: rename to loadingElement and switch to Progress component

Rugvip:

I think we used the "Loading..." text initially when this lived in `frontend-app-api` and we couldn't have a dependency on `core-components`. It fine to add that now though, there already is one through the `plugin-app` dependency.

I also considered removing the loading component altogether and let the app define it with a `Suspense` if desired, but the `Progress` is a lot nicer as it helps prevent flicker. There's also a bit of a foot-gun where the `Progress` component from `frontend-plugin-api` doesn't work outside the app, since the default implementation is an empty div and we're shipping the proper one through the `app-plugin`.

The rename from `loadingComponent` to `loadingElement` is because it isn't actually a component 😅. Technically `loadingNode` might be more correct since it's a `ReactNode`, but I don't think that's as clear as calling it an element tbh.

### DIFF
--- a/packages/frontend-defaults/src/createApp.tsx
+++ b/packages/frontend-defaults/src/createApp.tsx
@@ -22,6 +22,7 @@ import {
   FrontendFeature,
   FrontendFeatureLoader,
 } from '@backstage/frontend-plugin-api';
+import { Progress } from '@backstage/core-components';
 // eslint-disable-next-line @backstage/no-relative-monorepo-imports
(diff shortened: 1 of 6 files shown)


packages/frontend-defaults/src/createApp.tsx
```diff
@@ -143,9 +143,10 @@ export function createApp(options?: CreateAppOptions): {
     return { default: () => rootEl };
   }
 
+  const LazyApp = lazy(appLoader);
```
freben, line 146: 👍

freben (approved):

PR #30859, opened 2025-08-11
fix(catalog): processor order not to affect hash generation

drodil:

sort the arrays such as relations, deferred entities, refresh keys and parents coming from processors before updating the entity hash. when the order of the processors is random, the order of the emitted relations can be also random making unnecessary stitching occur for the entity as the hash has changed.

this change most probably means that existing hashes for entities do not match on the first processing round but makes the upcoming processing rounds much more stable thus making the catalog processing more performant.

### DIFF
--- a/.changeset/moody-clowns-hear.md
+++ b/.changeset/moody-clowns-hear.md
@@ -1,0 +1,13 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Make the processing hash calculation not care about the order of the processors.
+
+This change does not affect the behavior of the catalog, but it will make the processing
+hash calculation more robust against changes in the order of processors. This should lead to
+more stable processing hashes, which in turn should lead to fewer unnecessary reprocessing
+of entities.
+
+After deploying this fix, you may see a period of increased processing and stitching, but
+this should stabilize over time as the processing hashes become more consistent.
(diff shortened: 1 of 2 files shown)


drodil:
We ran this patch overnight on our instance, and as we assumed, it did re-stitch around 50% of the entities. But after that, the changed entity amount has been closer to zero, even though there have been new deployments of the instance. Previously, the number of changed entities fluctuated significantly, and a considerable amount of unnecessary re-stitching was performed.

.changeset/moody-clowns-hear.md
```diff
@@ -0,0 +1,10 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Make the processing hash calculation not care about the order of the processors.
+
+This change does not affect the behavior of the catalog, but it will make the processing
+hash calculation more robust against changes in the order of processors. This should leave to
```
freben:
```suggestion
hash calculation more robust against changes in the order of processors. This should lead to
```

.changeset/moody-clowns-hear.md
freben, line 1: Could you add a line saying something like "After deploying this fix, you may see a period of increased processing and stitching, but this settles on a lower stable level once complete."? Just to prepare people for a spike in load.

freben (commented):
Nice! Just a little changeset tweak.

PR #30860, opened 2025-08-11
[K8s plugin] Support google service account creds json

soapraj:

- adds googleServiceAccountCredentials option in config
- use service account creds when available or fallback to Application Default Credentials or GOOGLE_APPLICATION_CREDENTIALS


plugins/kubernetes-backend/src/service/KubernetesBuilder.ts
```diff
@@ -474,7 +474,7 @@ export class KubernetesBuilder {
       aws: new AwsIamStrategy({ config: this.env.config }),
       azure: new AzureIdentityStrategy(this.env.logger),
       google: new GoogleStrategy(),
-      googleServiceAccount: new GoogleServiceAccountStrategy(),
+      googleServiceAccount: new GoogleServiceAccountStrategy(this.env.config),
```
Rugvip: Could you update this to be `{ config: ... }` instead? like `AwsIamStrategy`

Just to make it easier to make future additions

soapraj: done!

plugins/kubernetes-backend/src/auth/GoogleServiceAccountStrategy.ts
```diff
@@ -20,19 +20,69 @@ import {
...
+ * If no credentials are provided in config, falls back to GOOGLE_APPLICATION_CREDENTIALS or ADC.
  *
  * @public
  */
 export class GoogleServiceAccountStrategy implements AuthenticationStrategy {
+  private readonly credentials?: string;
+
+  constructor(opts: { config: Config }) {
+    this.credentials = opts.config.getOptionalString(
+      'kubernetes.googleServiceAccountCredentials',
```
Rugvip, line 55: Was thinking for a second this should by on the individual cluster configuration, but the strategy is fairly global after all. Still, it could be signaled through auth meta, but I think we can leave that for a potential future addition.

plugins/kubernetes-backend/src/auth/GoogleServiceAccountStrategy.ts
```diff
@@ -20,19 +20,69 @@ import {
...
+
+        client = new container.v1.ClusterManagerClient({
+          credentials: credentialsObject,
+        });
+      } catch (error) {
+        throw new Error(
+          `Failed to parse Google Service Account credentials from config: ${
+            error instanceof Error ? error.message : 'Invalid JSON'
+          }`,
+        );
```
Rugvip, line 74: Typically we'd `new ForwardedError(..., error)` here instead. Gonna ship as is, so just if you wanna follow up with a cleanup after

Rugvip (approved):
Alright, let's :shipit:

PR #30864, opened 2025-08-12
`kubernetes`: breaking change to remove support for legacy backend system

benjdlambert:
This PR has a few breaking changes.

- Removing `backend-common` dependency
- Removing support for the old legacy backend system and calling `createRouter` and `KubernetesBuilder` directly.
- Removing re-exported types from `kubernetes-node`


plugins/kubernetes-backend/package.json
```diff
@@ -62,7 +47,6 @@
     "@aws-sdk/credential-providers": "^3.350.0",
     "@aws-sdk/signature-v4": "^3.347.0",
     "@azure/identity": "^4.0.0",
-    "@backstage/backend-common": "^0.25.0",
```
Rugvip, line 65: patrik-smol

yarn.lock
```diff
@@ -47745,16 +47314,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"tough-cookie@npm:~2.5.0":
```
Rugvip, line 47748: aw yiss

Rugvip (approved):
:shipit: :shipit: :shipit: :shipit: :shipit: 

Bit more context: we're planning to follow this up with some additions to the extension points that make it a bit easier to build on top of the default implementations when overriding the various locators etc.

Issue #30613, opened 2025-07-22
Scaffolder action not accepting valid input


Teo-ShaoWei:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
scaffolder action input
```

### 🗃️ Project Area

Software Templates

### 🔗 External Integration

N/A

### 📝 Description & Context

Scaffolder action created with `createTemplateAction` doesn't accept input that feels valid intuitively.

For illustration, let us have a scaffolder action that makes use of `default`:
```ts
export function randomAction() {
  return createTemplateAction({
    id: 'something:random',
    description: 'Just for demo',
    schema: {
      input: {
        x: z => z.number().default(0),
      },
    },
    async handler(ctx) {
      ctx.logger.info('The value of x is $x');
    },
  });
}
```

### 👍 Expected Behavior

I expect the following to be a valid test construct:
```ts
import { createMockActionContext } from '@backstage/plugin-scaffolder-node-test-utils';
import * as winston from 'winston';

const action = randomAction();
const mockContext = createMockActionContext({
  logger: mockServices.logger.mock() as unknown as winston.Logger,
});

test('test will pass no problem', async () => {
  await action.handler({
    ...mockContext,
    input: {},
  });
  expect(mockContext.logger.info).toHaveBeenCalledWith(
    'The value of x is 0',
  );
});
```
and in fact Jest will accept it as valid. However `tsc` will flag `input` as having an invalid type, due to not providing `x`.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes, but I would like some more guidance

Teo-ShaoWei:
I guess the issue is because in the following code:
https://github.com/backstage/backstage/blob/dcf2c75cfc09d05af5b39d1678b7fe5798fefc25/plugins/scaffolder-node/src/actions/createTemplateAction.ts#L85

we make use of `z.infer` instead of `z.input` ([Inferring types | Zod docs](https://zod.dev/basics?id=inferring-types)) (I am using Zod 4 docs but it's available for Zod 3 too).

[...]

PR #30865, opened 2025-08-12
fix(scaffolder-node): properly handle zod schemas with .default() and .transform()

Eteryz:
Change action flow to parse inputs with zod schema first, fixing default values handling.

Fixes issues #30613


Eteryz:
@schultzp2020 
Good afternoon!
Please take a look at the pull request.
Previously it was in the branch "Eteryz:scaffolder/fix-30613". I have now moved this to the "Eteryz:scaffolder/fix-30613-zod" branch.
You asked to fix the tests. 
Now all the tests are fixed. Due to the lack of type checking, many tests were written poorly.

packages/backend-defaults/src/entrypoints/rootConfig/createConfigSecretEnumerator.test.ts
```diff
@@ -29,7 +30,7 @@ describe('createConfigSecretEnumerator', () => {
       mockServices.rootConfig({
         data: {
           backend: { auth: { keys: [{ secret: 'my-secret-password' }] } },
-        },
+        } as JsonObject,
```
Teo-ShaoWei, line 33: Is there a reason why we would use type assertion here? I would prefer to leave it out of codebases otherwise.

Unlike the other common languages, TypeScript uses [structural subtyping](https://www.typescriptlang.org/docs/handbook/type-compatibility.html). The fact that not passing in `JsonObject` doesn't result in tsc errors shows that it recognizes the objects as `JsonObject`s.

In TypeScript, [type assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions) is used to override TypeScript's derived types in the event it can't figure out the correct type, usually, due to usage of `any` or other sloppy typing habits that resulted in `tsc` complaining.

So, as we can see from the codebase, `JsonObject` is usually used to declare types, but is seldom used to assert types. Worse is that asserted types in TypeScript could hide typing errors that would be uncovered if we rely on it to thread the typings throughout the codebase.

plugins/scaffolder-backend-module-bitbucket-cloud/src/actions/bitbucketCloudPipelinesRun.ts
```diff
@@ -86,6 +86,7 @@ export const createBitbucketPipelinesRunAction = (options: {
           },
         );
       } catch (e) {
+        console.log(e);
```
Teo-ShaoWei, line 89: Might be an unintended `console.log`?

plugins/scaffolder-backend-module-cookiecutter/src/actions/fetch/cookiecutter.test.ts
```diff
@@ -25,6 +25,7 @@ import { createMockActionContext } from '@backstage/plugin-scaffolder-node-test-
 import { Writable } from 'stream';
 import { UrlReaderService } from '@backstage/backend-plugin-api';
 import { ContainerRunner } from './ContainerRunner';
+import { ZodError } from 'zod';
```
Teo-ShaoWei, line 28: Seems like the codebase is trying to move `zod` inwards from the backend plugin modules into the plugins. Not sure if there's a need to check for a `ZodError` explicitly, but instead check for its message signature?

plugins/scaffolder-backend/src/scaffolder/actions/builtin/filesystem/rename.test.ts
```diff
@@ -62,57 +64,47 @@ describe('fs:rename', () => {
...
-        input: { files: null } as any,
-      }),
-    ).rejects.toThrow(/files must be an Array/);
+  it('should throw ZodError when files is not an array', async () => {
+    const testCases = [
+      { input: undefined },
+      { input: {} },
+      { input: '' },
+      { input: null },
+    ];
```
Teo-ShaoWei, line 73: Using `Promise.all` within a Jest test isn't a common pattern. There's much nicety in going with Jest's `test.each` template literal I think. As in the 2nd form introduced in this [article](https://vanslaars.io/articles/jest-testing-with-test-each/).

There's some usage of the first form in the codebase, but I find the 2nd form to be very descriptive of what the test want, since we want to keep them DAMP 👍

plugins/scaffolder-backend/src/scaffolder/actions/builtin/filesystem/rename.test.ts
```diff
@@ -62,57 +64,47 @@ describe('fs:rename', () => {
...
+
+    await Promise.all(
+      testCases.map(({ input }) =>
+        expect(
+          Promise.resolve().then(() =>
+            action.handler({
+              ...mockContext,
+              input: { files: input } as any,
```
Teo-ShaoWei, line 81: There's no need to write unit tests for types that Zod will reject. Zod itself is the (runtime) test. In the code after Zod's `parse`, we know that any types that don't satisfy it would not go through. Anything that might go through wrongly, should be written as a unit test for that subsequent code.

This is the same logic as once we are in TypeScript, we no longer write a whole class of type-checking unit tests that were the norm for JavaScript. Not only does they not help prevent regression, they are more ugly to write in TypeScript than JavaScript.

plugins/scaffolder-node/src/actions/createTemplateAction.test.ts
```diff
@@ -52,18 +55,18 @@ describe('createTemplateAction', () => {
     const action = createTemplateAction({
       id: 'test',
       schema: {
-        input: z =>
-          z.union([
-            z.object({
-              repoUrl: z.string(),
+        input: (zImpl: typeof z) =>
```
Teo-ShaoWei, line 58: Writing the type here is not necessary as TypeScript can figure out the underlying type. I believe it was also the intention of using the function form of `z => schema`. If every subsequent usage at the plugin module needs `zImpl: typeof z`, it would have been better to keep it as `schema` in the first place.

plugins/scaffolder-node/src/actions/createTemplateAction.ts
```diff
@@ -150,16 +150,40 @@ export function createTemplateAction<
     TOutputSchema
   >,
 ): TemplateAction<TActionInput, TActionOutput, 'v2'> {
-  const { inputSchema, outputSchema } = parseSchemas(
+  const { inputSchema, outputSchema, inputZodSchema } = parseSchemas(
     action as TemplateActionOptions<any, any, any>,
   );
 
+  const wrapperHandler = (ctx: ActionContext<TActionInput, TActionOutput>) => {
```
Teo-ShaoWei, line 157: Hmm... The function isn't correct in isolation... It can't be that `ctx.input` and the final `schema.input` have the same type `TActionInput` 😅 This is apparent from the overuse of type assertions within this function.

Teo-ShaoWei (commented):
Interesting attempt at the problem! Glad to see someone trying to poke at it 💪

However, even if the solution perform as intended, I doubt it is clean. Using `parseSchemas` to apply `wrapperHandler` looks like a hack against TypeScript and Zod.

There are also quite a few decisions on the code that feels anti-TypeScript. I've made comments at the respective places.

I would stand by my original observation that `z.input` suffices. Since the Zod validation is doing the exact correct thing during runtime, it's just the typing that need a second look.

benjdlambert:
@Eteryz @Teo-ShaoWei would either of you like to revive this work? Feels like a good improvement that and fixing behaviour which I would hate to see go to waste :pray:

PR #30872, opened 2025-08-13
feat(catalog): allow using negation in entity filter query

drodil:

this feature adds support to make negated queries using the catalog client by suffixing the filter key with `!`. this also works for entity pickers and other components using the catalog client to fetch entities.

### DIFF
--- a/.changeset/eleven-queens-accept.md
+++ b/.changeset/eleven-queens-accept.md
@@ -1,0 +1,24 @@
+---
+'@backstage/catalog-client': patch
+'@backstage/plugin-catalog-backend': patch
+---
+
+Allow using negated filter in filter query.
+
+This allows users to filter entities that do not match a specific column in the search table by
+adding a `!` suffix to the column name in the filter query.
+
+For example, you can use the following filter to find all entities that do not have a specific name:
+
+```ts
+const response = await catalogClient.getEntities(
+  {
+    filter: {
+      'metadata.name!': 'excluded-name',
+    },
+  },
+  { token },
+);
+```
+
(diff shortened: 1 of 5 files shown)


freben:
Hey! While I understand the effort here, I think we're rather hoping to eventually support [basically this](https://github.com/backstage/backstage/pull/28951) instead - probably through a POST and actual JSON formed query. The current query language is so quirky and limited, that it may not be worth pursuing further and instead be relegated to "the simple basic query form you can use for quick experiments and curling etc" and then the actual full nested query language is what people resort to for heavier needs.

drodil:
Yeah, I agree that makes a lot of sense and is the right way. But as we all know, eventually means quite far away in the future, and this would also require a major release of the catalog plugins unless the new way of querying is totally isolated.  This is not the first time people ask for the negation, latest in this thread https://discord.com/channels/687207715902193673/1404915081795469332, so just wanted to give it a quick try at how hard it would be to add to the current filtering (and it really wasn't). 

What comes to this PR, it's up to you really :) It doesn't break the current functionality and is quite a small addition, but would allow people to use negation before the new interface(s) are available.

freben:
I think you could be the one to make it happen, if you have some time! :) Would you like to?

I'd be fine with leaving the current API as-is, and amend it with a POST endpoint that supports basically the same things except in a json body, and also a `query` parameter. You'll find that it's easily translateable to the internal `EntityFilter` type inside the catalog, although maybe you'll throw an error for the `$contains` or translate that to an allOf or whatever else is necessary.

drodil:
Yeah, have to think about this, time is a bit limited also in this end and this is quite a big change!

PR #30875, opened 2025-08-13
Fix feature discovery crash with packages using exports fields

hudsonb:

When using packages that employ modern Node.js "exports" fields without exporting ./package.json (such as OpenTelemetry packages), the feature discovery service crashes with ERR_PACKAGE_PATH_NOT_EXPORTED errors. This happens because these packages restrict which files can be accessed via require(), and many don't expose their package.json files.

Wraps the package.json resolution with error handling to gracefully skip packages that don't export their package.json. Since these packages can't be Backstage plugins anyway (we need access to the backstage.role field).

### DIFF
--- a/packages/backend-defaults/src/PackageDiscoveryService.ts
+++ b/packages/backend-defaults/src/PackageDiscoveryService.ts
@@ -23,6 +23,7 @@ import {
   RootLoggerService,
 } from '@backstage/backend-plugin-api';
 import { BackstagePackageJson } from '@backstage/cli-node';
+import { isError } from '@backstage/errors';
 
 const DETECTED_PACKAGE_ROLES = [
   'node-library',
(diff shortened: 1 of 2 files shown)


packages/backend-defaults/src/PackageDiscoveryService.ts
```diff
@@ -125,9 +125,22 @@ export class PackageDiscoveryService {
...
+      let depPkg: BackstagePackageJson;
+      try {
+        const packageJsonPath = require.resolve(`${name}/package.json`, {
+          paths: [packageDir],
+        });
+        depPkg = require(packageJsonPath) as BackstagePackageJson;
+      } catch (error) {
+        // Handle packages with "exports" field that don't export ./package.json
+        if (
+          error instanceof Error &&
```
benjdlambert: I think you can use `isError` or `assertError` which will rethrow from `@backstage/errors` here to return `ErrorLike`.

`isError(error) && error.code === 'ERR_PACKAGE_PATH_NOT_EXPORTED'`

Should let you remove the cast and `error instanceof Error` as generally we like to avoid that pattern.

hudsonb: Updated!

Rugvip (approved):
Nice! 

Let's :shipit:

PR #30879, opened 2025-08-13
Merge Storybook

cdedreuille:

### DIFF
--- a/packages/ui/.storybook/preview.tsx
+++ b/packages/ui/.storybook/preview.tsx
@@ -6,7 +6,7 @@ import { useGlobals } from '@storybook/preview-api';
 import '../src/css/styles.css';
 
 // Custom mocked styles
-import '../src/css/storybook.css';
+import './storybook.css';
 
 // Custom themes
 import './themes/spotify.css';
--- a/packages/ui/.storybook/storybook.css
+++ b/packages/ui/.storybook/storybook.css
@@ -1,0 +1,38 @@
+:root {
+  --sb-panel-radius: 0;
+  --sb-panel-top: 0;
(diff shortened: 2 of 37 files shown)


storybook/package.json
Rugvip, line 1: Was gonna suggest we keep the separated storybook install, but the deps have already crept in via BUI and they're causing much less trouble than in the past so 👍 to getting rid of this and simplifying the setup.

.gitignore
```diff
@@ -184,3 +184,6 @@ knip.json
 type-docs
 docs.json
 tsconfig.typedoc.tmp.json
+
+# Storybook
+storybook-static/
```
Rugvip: Could we call this one `dist-storybook`? we're using that pattern for other output dirs

Rugvip (commented):
Awesome! 😁 

> Storybook is using Vite which will open the door for more testing capacities with `vite-test`

This would need more discussion though :p

.gitignore
cdedreuille: Sure. I'll do the update now.

storybook/package.json
cdedreuille: Yes and it will get even better in Storybook 9. they worked a LOT on this
https://storybook.js.org/blog/storybook-bloat-fixed/

storybook-9-dep-graph

.gitignore
cdedreuille: Done. It is now `dist-storybook`

cdedreuille:
@Rugvip Yes 100% for `vite-test`. It's not an actual proposition but I was just mentioning that it could be something we could consider in the future.

PR #30880, opened 2025-08-13
feat(plugin-org): add support for scrolling to OwnershipCard component

benjidotsh:
This PR makes the `OwnershipCard` component scrollable by adding a `scrollHeight` property that is set to `292px` by default, just enough to fit 2 rows.

This also removes the default value for `entityLimit` since setting this will disable the scrolling behaviour. I'm not 100% sure if this is something we can or want to do, so feel free to give feadback.

Scherm­afbeelding 2025-08-13 om 15 47 35 (2)

### DIFF
--- a/plugins/org/report.api.md
+++ b/plugins/org/report.api.md
@@ -40,7 +40,7 @@ export const EntityOwnershipCard: (props: {
   relationsType?: EntityRelationAggregation;
   relationAggregation?: EntityRelationAggregation;
   entityLimit?: number;
-  scrollHeight?: string;
+  maxScrollHeight?: string;
 }) => JSX_2.Element;
 
 // @public (undocumented)
@@ -102,7 +102,7 @@ export const OwnershipCard: (props: {
   relationsType?: EntityRelationAggregation;
   relationAggregation?: EntityRelationAggregation;
   entityLimit?: number;
-  scrollHeight?: string;
+  maxScrollHeight?: string;
 }) => JSX_2.Element;
 
 // @public (undocumented)


plugins/org/src/components/Cards/OwnershipCard/OwnershipCard.tsx
```diff
@@ -88,16 +94,20 @@ export const OwnershipCard = (props: {
...
   relationAggregation?: EntityRelationAggregation;
   entityLimit?: number;
+  scrollHeight?: string;
 }) => {
   const {
     variant,
     entityFilterKind,
     hideRelationsToggle,
-    entityLimit = 6,
+    entityLimit,
```
camilaibs: Changing the default entity limit is a visual breaking change. We'd be better off finding a solution that doesn't change this.

plugins/org/src/components/Cards/OwnershipCard/OwnershipCard.tsx
```diff
@@ -88,16 +94,20 @@ export const OwnershipCard = (props: {
...
+    scrollHeight: propScrollHeight = '292px',
```
camilaibs: Hmm, I think to maintain compatibility with the current default style we shouldn't have a default value defined here.

plugins/org/src/components/Cards/OwnershipCard/OwnershipCard.tsx
```diff
@@ -88,16 +94,20 @@ export const OwnershipCard = (props: {
...
   } = props;
   const relationAggregation = props.relationAggregation ?? props.relationsType;
   const relationsToggle =
     hideRelationsToggle === undefined ? false : hideRelationsToggle;
+  const scrollHeight =
+    entityLimit || variant === 'fullHeight' ? 'none' : propScrollHeight;
```
camilaibs: What happens if we skip the entity limit check here?

camilaibs (commented):
Thanks for the contribution 🙌🏻 , left initial comments, please take a look.

plugins/org/src/components/Cards/OwnershipCard/OwnershipCard.tsx
benjidotsh: Then it will also become scrollable if `entityLimit` is set, which we probably don't want

benjidotsh: Actually, if we're going to revert the default behaviour anyways, I don't think there is much of a choice but skipping the check here. 😅

freben:
Note that this currently requires updating the API reports (see the build output)

benjidotsh:
@freben not sure how I missed that 🙃

freben (approved):
Alright!

Issue #30121, opened 2025-06-03
plugin-catalog-graph “Simplified” view not working after upgrade to 0.4.19


Prebiusta:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
Plugin-catalog graph 0.4.19
Simplified
Unidirectional
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

N/A

### 📝 Description & Context

Issue is rather simple, after changing some logic in v0.4.19 of plugin-catalog-graph,
“Simplified” (unidirectional) view is no longer working and has no effect on the graphs.

### 👍 Expected Behavior

The “Simplified” toggle in Catalog Graph should be working

### 🥾 Reproduction steps

1. Make sure to use @backstage/plugin-catalog-graph: 0.4.19
1. Click on Catalog Graph
2. Toggle “Simplified” button
3. Observe no change

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Undecided

MihaiTabara:
Found in team's triaging this morning for. 

Thanks for filing this. While we're hoping for some contributions from the community, can you please help with reproducting it by providing more information about your Backstage version, config file perhaps, etc?

Prebiusta:
App created with `npx @backstage/create-app@latest` faces exactly the same issue. I have updated the description with more detailed steps.

Ijaz1-Mohammed:
yes this issue is still there. I reverted to 4.18

vinzscam:
it looks like this was introduced by #29895

chicobento:
I can confirm the issue. After upgrading to 1.40.0 made our graphs unreadable. 

![Image](https://github.com/user-attachments/assets/1a5dadcf-f476-4a41-a22d-1b202bca8449)

subhanshu-shukla-ril:
Same case with me too, Any solution would be highly appreciated

subhanshu-shukla-ril:
```
            if (!unidirectional || !visitedNodes.has(rel.targetRef)) {
              if (mergeRelations) {
                const pair = relationPairs.find(
                  ([l, r]) => l === rel.type || r === rel.type,
                ) ?? [rel.type];
                const [left] = pair;
            
[...]

PR #30889, opened 2025-08-14
Add support for customizing the catalog graph relations using an API

grantila:
## Better support custom relations in the catalog graph plugin

The catalog graph plugin currently only understands the built-in relations with [this caveat](https://github.com/backstage/backstage/blob/aedd4daa661c57b6067d872f67884af461a710b8/plugins/catalog-graph/src/components/EntityRelationsGraph/relations.ts#L41-L44). It will however by default show any other relation an entity has.

This has drawbacks:
 * The UI won't autocomplete for other "unknown" relations
 * The UI will show relations that might not be suitable to show
 * Configuring this is needed everywhere the graph is shown (its own page, the entity pages, etc)

This MR adds an API to the catalog graph plugin, which can be customized. It allows an implementor to specify exactly what relations and relation pairs to support, and what of them to show/select by default. The plugin will use this configuration wherever the graph is shown (although overridable by its already existing props).

The default implementation of the API includes the same known relations as today, so should be no different than before.

### DIFF
--- a/plugins/catalog-graph/src/components/EntityRelationsGraph/EntityRelationsGraph.tsx
+++ b/plugins/catalog-graph/src/components/EntityRelationsGraph/EntityRelationsGraph.tsx
@@ -33,7 +33,6 @@ import { DefaultRenderNode } from './DefaultRenderNode';
 import { RelationPairs } from '../../lib/types';
 import { Direction, EntityEdge, EntityNode } from '../../lib/types';
 import { useEntityRelationNodesAndEdges } from './useEntityRelationNodesAndEdges';
-import { GraphTransformationDebugger } from '../../lib/graph-transformations';
 
 /** @public */
 export type EntityRelationsGraphClassKey = 'progress' | 'container' | 'graph';
@@ -92,7 +91,6 @@ export type EntityRelationsGraphProps = {
   renderLabel?: DependencyGraphTypes.RenderLabelFunction<EntityEdge>;
   curve?: 'curveStepBefore' | 'curveMonotoneX';
   showArrowHeads?: boolean;
-  onPostTransformation?: GraphTransformationDebugger;
 };
 
 /**
@@ -118,7 +116,6 @@ export const EntityRelationsGraph = (props: EntityRelationsGraphProps) => {
     renderLabel,
     curve,
     showArrowHeads,
-    onPostTransformation,
   } = props;
 
   const theme = useTheme();
@@ -142,7 +139,6 @@ export const EntityRelationsGraph = (props: EntityRelationsGraphProps) => {
     entityFilter,
     onNodeClick,
     relationPairs,
-    onPostTransformation,
   });
 
   useEffect(() => {
--- a/plugins/catalog-graph/report.api.md
+++ b/plugins/catalog-graph/report.api.md
@@ -181,15 +181,7 @@ export type EntityRelationsGraphProps = {
   renderLabel?: DependencyGraphTypes.RenderLabelFunction<EntityEdge>;
   curve?: 'curveStepBefore' | 'curveMonotoneX';
   showArrowHeads?: boolean;
-  onPostTransformation?: GraphTransformationDebugger;
 };
-
-// @public
-export type GraphTransformationDebugger = (
-  transformation: string | undefined,
-  transformationContext: TransformationContext,
-  clonedContext: TransformationContext,
-) => void;
 
 // @public
 export type RelationPairs = [string, string][];
--- a/plugins/catalog-graph/src/components/EntityRelationsGraph/useEntityRelationNodesAndEdges.ts
+++ b/plugins/catalog-graph/src/components/EntityRelationsGraph/useEntityRelationNodesAndEdges.ts
@@ -24,8 +24,6 @@ import { buildGraph } from '../../lib/graph';
 import {
   BuiltInTransformations,
   builtInTransformations,
-  cloneTransformationContext,
-  GraphTransformationDebugger,
   GraphTransformer,
   TransformationContext,
 } from '../../lib/graph-transformations';
@@ -43,7 +41,6 @@ export function useEntityRelationNodesAndEdges({
   entityFilter,
   onNodeClick,
   relationPairs: incomingRelationPairs,
-  onPostTransformation,
 }: {
   rootEntityRefs: string[];
   maxDepth?: number;
@@ -54,7 +51,6 @@ export function useEntityRelationNodesAndEdges({
   entityFilter?: (entity: Entity) => boolean;
   onNodeClick?: (value: EntityNode, event: MouseEvent<unknown>) => void;
   relationPairs?: RelationPairs;
-  onPostTransformation?: GraphTransformationDebugger;
 }): {
   loading: boolean;
   nodes?: EntityNode[];
@@ -129,20 +125,6 @@ export function useEntityRelationNodesAndEdges({
         maxDepth,
       };
 
-      const debugTransformation = (
-        transformation: BuiltInTransformations | GraphTransformer | undefined,
-      ) => {
-        onPostTransformation?.(
-          typeof transformation === 'function'
-            ? transformation.name
-            : transformation,
-          transformationContext,
-          cloneTransformationContext(transformationContext),
-        );
-      };
-
-      debugTransformation(undefined);
-
       const runTransformation = (
         transformation: BuiltInTransformations | GraphTransformer,
       ) => {
@@ -151,8 +133,6 @@ export function useEntityRelationNodesAndEdges({
         } else {
           builtInTransformations[transformation](transformationContext);
         }
-
-        debugTransformation(transformation);
       };
 
       runTransformation('reduce-edges');
--- a/plugins/catalog-graph/src/index.ts
+++ b/plugins/catalog-graph/src/index.ts
@@ -35,7 +35,4 @@ export type {
   EntityNode,
 } from './lib/types';
 export { Direction } from './lib/types';
-export type {
-  TransformationContext,
-  GraphTransformationDebugger,
-} from './lib/graph-transformations';
+export type { TransformationContext } from './lib/graph-transformations';
--- a/plugins/catalog-graph/src/lib/graph-transformations/index.ts
+++ b/plugins/catalog-graph/src/lib/graph-transformations/index.ts
@@ -14,12 +14,7 @@
  * limitations under the License.
  */
 
-export type {
-  GraphTransformer,
-  TransformationContext,
-  GraphTransformationDebugger,
-} from './types';
-export { cloneTransformationContext } from './types';
+export type { GraphTransformer, TransformationContext } from './types';
 
 import { reduceEdges } from './reduce-edges';
 import { setDistances } from './set-distance';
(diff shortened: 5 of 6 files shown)


grantila:
Rebased on master but still fails on api reports for `@backstage/plugin-scaffolder-backend-module-rails` for some reason.

.changeset/dull-eagles-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-graph': patch
```
schultzp2020: I do think these changes may warrant a minor version bump since exports have different definitions now. Overall I like the changes though!

grantila: Thanks, yeah that's fair, changed.

vinzscam (commented):
Thanks @grantila, it looks great! 👍

Do you think this could be the right place for addressing https://github.com/backstage/backstage/issues/30121?

grantila:
Yeah this PR would be a great start. I'm actively working on solving #30121 (which currently is based on the logic in this branch) and will have a fix ready shortly (this week most likely).

freben:
I rudely pushed a fix.

Not sure why this would differ, maybe we're on very very slightly different versions of something that affected the api reports. They are a little finicky. But this is what I got after just grabbing your branch and running `yarn` and `yarn build:api-reports`.

grantila:
Thanks @freben, the opposite of rude 😄

If we get this in, I'll try as quickly as possible to fix the "simplify" (internally called "unidirectional", although not being the exact same thing 😏) which isn't working anymore. 🙏

grantila:
@freben & @vinzscam I now have a fix for #30121. Before realizing the issue of 30121, I started refactoring the logic heavily, and implemented a `catalog-graph-backend` to massively speed up the load/rendering, but I have extracted the pure UI logic/fixes (to not introduce a new required backend just for a bugfix). Will provide a followup PR for the backend later, so it can introduce a breaking change (requiring the backend plugin).

However, it's a pretty big refactoring of the graph logic (lots of changes), and I have compared it side-by-side heavily with a branch before #29895 when the bug was introduced, with various graphs of various sizes with combinations of `mergeRelations`/`unidirectional`. It's based on this branch, so I could either commit it here, or wait for this and have a separate PR (if that's easier to grasp).
What's your input on this, to have it solved as quickly as possible?

vinzscam:
> @freben & @vinzscam I now have a fix for #30121. Before realizing the issue of 30121, I started refactoring the logic heavily, and implemented a `catalog-graph-backend` to massively speed up the load/rendering, but I have extracted the pure UI logic/fixes (to not introduce a new required backend just for a bugfix). Will provide a followup PR for the backend later, so it can introduce a breaking change (requiring the backend plugin).
> 
> However, it's a pretty big refactoring of the graph logic (lots of changes), and I have compared it side-by-side heavily with a branch before #29895 when the bug was introduced, with various graphs of various sizes with combinations of `mergeRelations`/`unidirectional`. It's based on this branch, so I could either commit it here, or wait for this and have a separate PR (if that's easier to grasp).
> What's your input on this, to have it solved as quickly as possible?

This is amazing, thank you 🙏
I think it makes sense to have the fix in this PR, together with all the other great improvements so that we can try to merge it ASAP

grantila:
After fixing missing edges in #29895, the "simplified" didn't work anymore (#30121). The graph logic was quite hard to grasp, and before knowing about 30121, I started refactoring it, upon which this fix was made.

The graph logic is extracted into `/lib/graph` (for constructing the graph from a set of entities) and `/lib/graph-transformations` where each transformation is standalone and more easy understand.

The [useEntityRelationNodesAndEdges.ts](https://github.com/backstage/backstage/pull/30889/commits/f60e9fc1f8ac30474ecf7cbaa305560b4d1dea91#diff-65b94eef8151c7948cc461c41801aefcc6b07134db6066257ac82a3fb347bf71) is now smaller and quite clear what it does.

With moving towards a generic interface for transforming the graph, we could eventually introduce and allow for custom transformers to be supplied as a prop to the graph, so implementors can transform the graph to whatever need they have. Defaulting to the built-in ones will be necessary for exact backwards compatibility though.

grantila:
Pushed api-report changes. It seems to be flaky, re-ordering keys in the `report*.api.md` files every now and then 🤔

grantila:
I have now gone through even more manual testing of before-after tests of graphs, especially with multi-root layout (which isn't that common), to make it as close to the original layout as possible.

Since the fixes in this PR moves from the old all-in-one-loop logic into separate "transformations" (which I will make a separate PR for - allowing to be customizable), it's sometimes useful to be able to debug what each transformation does (and potentially mutate the result), a new prop `onPostTransformation` is added, which allows custom callbacks to be called after each transformation is done. The callback gets the transformation name, the current mutable graph and a cloned graph which can be logged to the console (as it won't ever change).

I can't see any difference between the result of several graphs after this PR, from what the original graph implementation resulted in.

EDIT: ^ Meaning, I think this PR is good to go

plugins/catalog-graph/src/components/EntityRelationsGraph/useEntityRelationGraph.ts
```diff
@@ -98,19 +100,21 @@ export function useEntityRelationGraph({
...
-  );
+  const filteredEntities = useMemo(() => {
+    if (loading) {
+      return {};
+    }
+
+    return entityFilter
+      ? pickBy(entities, (value, _key) => entityFilter(value))
+      : entities;
+  }, [loading, entities, entityFilter]);
```
vinzscam, line 117: not sure if it could make sense to ignore the loading state, so that we avoid some flickering if the data is reloaded. Feel free to ignore if this introduces side effects 

```suggestion
  const filteredEntities = useMemo(() => {
    return entityFilter
      ? pickBy(entities, (value, _key) => entityFilter(value))
      : entities;
  }, [entities, entityFilter]);
```

.changeset/dull-eagles-sin.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-catalog-graph': minor
+---
+
+Support custom relations by using an API to define known relations and which to show by default
+Fixes "simplified" bug (#30121) which caused graphs not to be simplified
```
vinzscam, line 6: ✨

vinzscam (approved):
🚀

plugins/catalog-graph/src/components/EntityRelationsGraph/useEntityRelationGraph.ts
grantila: Ironically, it'll flicker less now.
The upstream `entities` will contain a bunch of entities. When the user changes the query, a new set of entities will be fetched (and `loading` will be true), but the old ones will remain in the `entities`, so leaving the UI with a progress spinner on top flickers less than "emptying" it with this `if (loading)` ☺️

plugins/catalog-graph/src/components/EntityRelationsGraph/EntityRelationsGraph.tsx
```diff
@@ -91,6 +92,7 @@ export type EntityRelationsGraphProps = {
   renderLabel?: DependencyGraphTypes.RenderLabelFunction<EntityEdge>;
   curve?: 'curveStepBefore' | 'curveMonotoneX';
   showArrowHeads?: boolean;
+  onPostTransformation?: GraphTransformationDebugger;
```
freben: Not entirely sure about having this in the public API. Could have been hidden honestly. It's exposing the workings of an internal thing, so updating the internals might entail needing to break this needlessly. I'd be fine with having this as a private thing inside the implementation class that you'd have to go a bit `as any` on.

plugins/catalog-graph/src/components/CatalogGraphPage/SelectedRelationsFilter.tsx
```diff
@@ -41,25 +41,31 @@ const useStyles = makeStyles(
...
+  );
   const { t } = useTranslationRef(catalogGraphTranslationRef);
 
   const handleChange = useCallback(
     (_: unknown, v: string[]) => {
-      onChange(relations.every(r => v.includes(r)) ? undefined : v);
+      onChange(v);
     },
-    [relations, onChange],
+    [onChange],
```
freben, line 68: did this useCallback become unnecessary and can be removed?

freben (commented):
Alright, this is nice.

There's a ton of spread->`Set`->`Array.from` and `array.some` and stuff like that. I wonder if there's ever going to be reports of "i tried opening the graph page for my entity with 2000 relations and the tab crashed".

plugins/catalog-graph/src/components/EntityRelationsGraph/EntityRelationsGraph.tsx
grantila: The reasoning for this is surely not obvious - I should have explained it earlier, but here it is;

I'm working on a follow-up PR which (among other things - like replacing the recursive catalog calls into 1 call to a new `catalog-graph-backend` to dramatically improve performance of larger graphs) adds two more props:

```
  transformations?: (GraphTransformer | BuiltInTransformations)[];
  noDefaultTransformations?: boolean;
```

so that you can build and implement custom transformations that modify the graph - removes edges / nodes, adds edges / nodes, filters out certain edges to certain nodes, or whatnot.

With the possibility of letting the user create custom transformations, it's quite useful to be able to debug how the transformations alter the state of the graph. So, the `onPostTransformation` shouldn't be used in production to mutate the graph - a custom transformer should be used instead, but it's useful to be able to console log the state after each transformation to understand what's going on.

plugins/catalog-graph/src/components/CatalogGraphPage/SelectedRelationsFilter.tsx
grantila: Almost. It forwards the second argument of `AutoComplete.onChange`, as the first is an `Event` that's not useful for upstream `onChange`.

grantila:
Good points @freben, there is still one O(n * n) which has been there for years, I'll fix that in a followup. I chose to keep that snippet of code in this PR to not introduce more changes than necessary (ironically, given the PR size 😄)

I have been quite pedantic about complexity to ensure that every transformation and all logic added is not above O(n * log n) so I'm quite sure there's not going to be any memory or CPU issue with this code. I haven't tried _huge_ graphs (which are quite hard to view anyway), but they will still be O(n * log n) in the _added_ code, so 2000 relations would be 22000 iterations which is fast. In my manual testing with timing logic, I haven't seen anything above 5 ms of graph logic (which could be anything - like the browser doing other things - kernel context switching etc).

plugins/catalog-graph/src/components/EntityRelationsGraph/EntityRelationsGraph.tsx
grantila: ^ I could hold off this prop for a follow-up PR, and we can discuss wether it makes sense there instead. Would that be better?

freben: Yepyep.

In general, debugging stuff shouldn't be part of public interfaces.

grantila:
After feedback; removed the `onPostTransformation` prop and including debug related logic. cc @freben

PR #30894, opened 2025-08-14
Migrate packages/app to msw 2

cdedreuille:
No changeset since it only affects dev deps and a test

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -96,7 +96,7 @@
     "@types/zen-observable": "^0.8.0",
     "axios": "^1.11.0",
     "cross-env": "^7.0.0",
-    "msw": "^1.0.0"
+    "msw": "^2.10.5"
   },
   "bundled": true
 }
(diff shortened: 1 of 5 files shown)


packages/app/package.json
```diff
@@ -96,7 +96,7 @@
     "@types/zen-observable": "^0.8.0",
     "axios": "^1.11.0",
     "cross-env": "^7.0.0",
-    "msw": "^1.0.0"
+    "msw": "^2.10.5"
```
freben, line 99:
```suggestion
    "msw": "^2.0.0"
```

Let's just use the same exact range globally in the whole repo; keeps our yarn.lock and customers' yarn.locks clean and easy to maintain

freben (commented):
Nice!

freben:
For TextEncoder problems, that's just an annoying jest quirk. You can [do this](https://github.com/backstage/backstage/blob/adef00907996afc79526df04567732cf38d0fc66/plugins/api-docs/src/setupTests.ts#L19) in the corresponding `setupTests.ts` file in the package root.

PR #30902, opened 2025-08-15
Patch release of #30901

benjdlambert:
This release fixes an issue where the logger would log both redacted and redacted secrets if used as input into `fetch:template`.


Rugvip (approved):
:shipit:

PR #30903, opened 2025-08-15
backend-test-utils: add mock implementation for scheduler service

Rugvip:

Ran into some trouble testing plugins with scheduled tasks. With the default implementation we need to wait 5s for the task to be picked up, and the same applies to manually triggered tasks.

This switches out the use of the default implementation with a new scheduler service mock. I tried to find some sensible defaults to avoid breaking existing tests, but we'll see what CI says 😁

In addition to the existing `triggerTask` method the mock also provides a `triggerAllTasks` method that can help avoid the need to know the exact task IDs or wait for a larger number of tasks. ~Might also add a `waitForRunningTasks` if this seems like something that's possible to ship.~ Added docs and `waitForTask`, also considered making `triggerAllTasks`
private because it's not super useful, but it's used for the startup hook and mirrors those option, so figured we might as well make it available.

### DIFF
--- a/docs/backend-system/core-services/scheduler.md
+++ b/docs/backend-system/core-services/scheduler.md
@@ -134,7 +134,7 @@ Responds with
 
 The `@backstage/backend-test-utils` package provides `mockServices.scheduler`, which provides a mocked implementation of the scheduler service that can be used in tests. This mocked implementation is used by default in `startTestBackend`, and it will immediately run any registered tasks on startup as long as they're not configured to run manually or with an initial delay.
 
-A dedicated instance can be used for more control during testing, with the mock implementation providing additional utilities to trigger and wait for tasks to complete:
+A dedicated instance can be used for more control during testing:
 
 ```ts
 it('should trigger a task', async () => {
@@ -144,16 +144,9 @@ it('should trigger a task', async () => {
     features: [scheduler.factory()],
   });
 
-  // Start waiting for some task to complete
-  const waitForTask = scheduler.waitForTask('some-task-id');
+  await scheduler.triggerTask('some-task-id');
 
-  // Call an endpoit that triggers a task
-  const res = await request(server).post(
-    '/api/my-plugin/route-that-triggers-a-task',
-  );
-  expect(res.status).toBe(200);
-
-  // Wait for the task to complete
-  await waitForTask;
+  // Next verify that the plugin state is updated accordingly
+  // e.g. by calling the API or verifying database state
 });
 ```


packages/backend-test-utils/report.api.md
```diff
@@ -366,9 +366,15 @@ export namespace mockServices {
       ) => ServiceMock<RootLoggerService>;
   }
   // (undocumented)
+  export function scheduler(): SchedulerService;
```
freben, line 369: doesn't the return type here need to be `MockSchedulerService` or at least `SchedulerService & { factory: ... }` to support the `features: [scheduler.factory()]` mentioned in the docs?

packages/backend-test-utils/src/services/mockServices.ts
```diff
@@ -496,8 +497,15 @@ export namespace mockServices {
...
+  export function scheduler(): SchedulerService {
+    return new MockSchedulerService();
+  }
   export namespace scheduler {
-    export const factory = () => schedulerServiceFactory;
+    export const factory = (options?: {
+      skipTaskRunOnStartup?: boolean;
+      includeManualTasksOnStartup?: boolean;
+      includeInitialDelayedTasksOnStartup?: boolean;
+    }) => new MockSchedulerService().factory(options);
```
freben, line 508: It is slightly surprising that these options only exist on the factory, right? I thought they would be constructor arguments (and on the `scheduler()` above too) and the instance follows that behavior no matter if you make a factory for it or not.

The scheduler could have a `start` method if needed.

packages/backend-test-utils/src/services/MockSchedulerService.ts
```diff
@@ -0,0 +1,167 @@
...
+  ): Promise<void> {
+    this.#tasks.set(task.id, {
+      ...task,
+      descriptor: {
+        id: task.id,
+        scope: task.scope ?? 'global',
+        settings: { version: 1 },
+      },
+    });
+    return Promise.resolve();
```
freben: why tho

packages/backend-test-utils/src/services/MockSchedulerService.ts
```diff
@@ -0,0 +1,167 @@
...
+  ): SchedulerServiceTaskRunner {
+    return {
+      run: async task => {
+        await this.scheduleTask({ ...task, ...schedule });
+      },
+    };
+  }
+
+  getScheduledTasks(): Promise<SchedulerServiceTaskDescriptor[]> {
+    return Promise.resolve(
```
freben: did you want to avoid making the method async?

packages/backend-test-utils/src/services/MockSchedulerService.ts
```diff
@@ -0,0 +1,167 @@
...
+    const task = this.#tasks.get(id);
+    if (!task) {
+      throw new Error(`Task ${id} not found`);
+    }
+    if (this.#runningTasks.has(id)) {
+      return;
+    }
+    this.#runningTasks.add(id);
+    try {
+      await task.fn(new AbortController().signal);
```
freben: i wonder if it's actually worth subscribing to the lifecycle shutdown hook and pass in a signal that actually triggers, to get cleaner test shutdowns in some cases

packages/backend-test-utils/src/services/mockServices.ts
Rugvip: `triggerAllTasks` is the `start` method, and that one has all the options. I did have the options on the instance for a lil bit but it's less flexible than being able to forward new options to each trigger.

packages/backend-test-utils/report.api.md
Rugvip: Yep, totally

packages/backend-test-utils/src/services/MockSchedulerService.ts
Rugvip: Just AI slop that I didn't bother changing :p

packages/backend-test-utils/src/services/MockSchedulerService.ts
Rugvip: Yep, threw that in there

packages/backend-test-utils/report.api.md
Rugvip: I reverted this actually, and made `triggerAllTasks` internal. I figured it's quite a lot cleaner if we can just have the regular scheduler service be the public mock API too, just with the additional options.

packages/backend-test-utils/src/services/MockSchedulerService.ts
```diff
@@ -0,0 +1,179 @@
...
+} from '@backstage/backend-plugin-api';
+import { createDeferred, DeferredPromise } from '@backstage/types';
+
+export class MockSchedulerService implements SchedulerService {
+  readonly #tasks = new Map<
+    string,
+    SchedulerServiceTaskInvocationDefinition &
+      SchedulerServiceTaskScheduleDefinition & {
+        descriptor: SchedulerServiceTaskDescriptor;
+        abortControllers: AbortController[];
```
freben: i was thinking more like a `readonly #shutdownController;` that gets made at startup and then `this.#shutdownController.signal` is passed in everywhere 🤷

packages/backend-test-utils/src/services/MockSchedulerService.ts
```diff
@@ -0,0 +1,179 @@
...
+      }
+      if (scope === 'all' || scope === task.scope) {
+        selectedTaskIds.push(task.id);
+      }
+    }
+
+    await Promise.all(selectedTaskIds.map(id => this.triggerTask(id)));
+  }
+
+  async #shutdownAllTasks() {
```
freben, line 154: should it wait for `deferredTaskCompletions` or is that just a waste? it would hold back shutdown properly so it might be cleaner

packages/backend-test-utils/src/services/MockSchedulerService.ts
```diff
@@ -0,0 +1,179 @@
...
+   * @param id - The task ID to wait for
+   * @returns A promise that resolves when the task is completed
+   */
+  async waitForTask(id: string): Promise<void> {
+    const existing = this.#deferredTaskCompletions.get(id);
+    if (existing) {
+      return existing;
+    }
+    const defferred = createDeferred<void>();
+    this.#deferredTaskCompletions.set(id, defferred);
```
freben:
```suggestion
    const deferred = createDeferred<void>();
    this.#deferredTaskCompletions.set(id, deferred);
```

packages/backend-test-utils/src/services/MockSchedulerService.ts
```diff
@@ -0,0 +1,179 @@
...
+    this.#runningTasks.add(id);
+    try {
+      const abortController = new AbortController();
+      task.abortControllers.push(abortController);
+      await task.fn(abortController.signal);
+      this.#deferredTaskCompletions.get(id)?.resolve();
+    } catch (error) {
+      this.#deferredTaskCompletions.get(id)?.reject(error);
+    } finally {
+      this.#runningTasks.delete(id);
```
freben, line 114: this should probably `this.#deferredTaskCompletions.delete(id)`, shouldn't it

packages/backend-test-utils/src/services/MockSchedulerService.ts
```diff
@@ -0,0 +1,179 @@
...
+    for (const task of this.#tasks.values()) {
+      if (task.initialDelay && !includeInitialDelayedTasks) {
+        continue;
+      }
+      if ('trigger' in task.frequency && task.frequency.trigger === 'manual') {
+        if (includeManualTasks) {
+          selectedTaskIds.push(task.id);
+        }
+        continue;
+      }
```
freben:
```suggestion
      if ('trigger' in task.frequency && task.frequency.trigger === 'manual' && !includeManualTasks) {
        continue;
      }
```

this thing was hard to read

packages/backend-test-utils/src/services/MockSchedulerService.ts
```diff
@@ -0,0 +1,179 @@
...
+      for (const abortController of task.abortControllers) {
+        abortController.abort();
+      }
+    }
+  }
+
+  /**
+   * Wait for the task with the given ID to complete.
+   *
+   * If the task has not yet been scheduler or started, this will wait for it to be scheduled, started, and completed
```
freben:
```suggestion
   * If the task has not yet been scheduled or started, this will wait for it to be scheduled, started, and completed
```

packages/backend-test-utils/src/services/MockSchedulerService.ts
Rugvip: Yep that's simpler

packages/backend-test-utils/src/services/MockSchedulerService.ts
Rugvip: Thinking it's mostly just a waste tbh, and maybe risks hanging the runner longer than needed.

packages/backend-test-utils/src/services/MockSchedulerService.ts
Rugvip: ~That's captured in the `try`/`catch` above~, nvm, you want the cleanup, sure 😅, it doesn't add a whole lot of value

freben: Except the ability to trigger more than once and waiting for it

PR #30905, opened 2025-08-15
Home plugin - Visit enrichment, transform, save, display context provider

sswane:

Added a few ways to customize the Recently Visited & Top Visited lists for the Home plugin. First, it allows 3 optional functions to be passed into the VisitsStorageApi:

```typescript
export const apis: AnyApiFactory[] = [
  createApiFactory({
    api: visitsApiRef,
    deps: {
      storageApi: storageApiRef,
      identityApi: identityApiRef,
      catalogApi: catalogApiRef,
    },
    factory: ({ storageApi, identityApi, catalogApi }) =>
      VisitsStorageApi.create({
        storageApi,
        identityApi,
        transformPathname,
        canSave,
        enrichVisit: createEnrichVisit(catalogApi),
      }),
  }),
];
```

Then it adds a new `VisitDisplayProvider` with optional functions to update the chip colors and labels, which can be wrapped around the `HomePageRecentlyVisited` and `HomePageTopVisited` components. 

`<VisitDisplayProvider getChipColor={getChipColor} getLabel={getChipLabel}>`

See the new section of the README of the Home plugin which explains why users may use the functions. 

Example screenshot without customizations (everything works the same as before):
image

Example screenshot with customizations:
image

### DIFF
--- a/plugins/home/report-alpha.api.md
+++ b/plugins/home/report-alpha.api.md
@@ -105,6 +105,7 @@ export default _default;
 export const homeTranslationRef: TranslationRef<
   'home',
   {
+    readonly 'starredEntities.noStarredEntitiesMessage': 'Click the star beside an entity name to add it to this list!';
     readonly 'addWidgetDialog.title': 'Add new widget to dashboard';
     readonly 'customHomepageButtons.clearAll': 'Clear all';
     readonly 'customHomepageButtons.edit': 'Edit';
@@ -123,7 +124,6 @@ export const homeTranslationRef: TranslationRef<
     readonly 'quickStart.title': 'Onboarding';
     readonly 'quickStart.description': 'Get started with Backstage';
     readonly 'quickStart.learnMoreLinkTitle': 'Learn more';
-    readonly 'starredEntities.noStarredEntitiesMessage': 'Click the star beside an entity name to add it to this list!';
     readonly 'visitedByType.action.viewMore': 'View more';
     readonly 'visitedByType.action.viewLess': 'View less';
     readonly 'featuredDocsCard.empty.title': 'No documents to show';


sswane:
This is not stale due to the PR author. Author has been waiting for review since submission.

sswane:
@benjdlambert how can we get a review of this? it has been open for 3 weeks without comment. we have a patch working in our project but would like the customizability to come from the open source plugin itself if possible.

plugins/home/src/components/VisitList/ItemCategory.tsx
```diff
@@ -27,53 +26,17 @@ const useStyles = makeStyles(theme => ({
...
-  const foundIndex = entityKinds.indexOf(
-    entity.kind.toLocaleLowerCase('en-US'),
-  );
-  return foundIndex === -1 ? defaultColor : getColorByIndex(foundIndex + 1);
-};
 
 export const ItemCategory = ({ visit }: { visit: Visit }) => {
   const classes = useStyles();
-  const entity = maybeEntity(visit);
+  const { getChipColor, getLabel } = useVisitDisplay();
```
iamEAP, line 32: I didn't see anywhere in this PR where the `<VisitDisplayContext>` is provided by default. Would this result in the default home throwing errors?

(Apologies if I'm missing broader context around the Home plugin--it's been awhile since I've looked through the code)

iamEAP (commented):
Hey @sswane! Thanks for the patience, and thanks especially for the contribution--I've seen the need for this kind of thing, and making it flexible in this way is a great idea.

While I don't think it's strictly necessary for this to support the New Frontend System straight out the gate, I do think it's at least worth keeping in mind how it could be supported, and offering the functionality in such a way where it could easily be ported over when the time comes.

To that end:

- In terms of the data ingestion side of things, do you think it makes sense to expand on the API surface of `VisitsStorageApi`, rather than introducing a loose collection of functional types?  Wondering if adding optional `VisitsApi.canSave()` and `VisitsApi.enrichVisit()` methods could help keep the write-side of this limited to just the utility API, which can more easily be overridden in the NFS.

I'm a little less certain how we might handle this on the read-side of things; perhaps we'd introduce a new blueprint for visit list items, and you could supply your own display logic in a completely custom component?  Perhaps something we can figure out later, and it's fine for with this context with custom callbacks.  Although see my comment inline.

Let me know what you think!

plugins/home/src/components/VisitList/ItemCategory.tsx
sswane: Inside the new `Context.tsx` file (where `useVisitDisplay()` is called from), it defaults to all the previous functionality like using the `colorVariants` from @backstage/theme for the chip colors and entity kind or 'other' for the labels.

sswane:
@iamEAP thanks for the review!! I made the updates to move the 3 optional functions from the VisitListener to the VisitsStorageApi. The home plugin still functions as before if the user does not provide those functions or does not add the VisitDisplayProvider. The first screenshot in the PR description shows the changes without customizations (no extra functions or provider).

zjperschall:
Not stale

iamEAP (approved):
Thanks y'all! And thanks again for the patience. Changes look good to me.

PR #30906, opened 2025-08-16
frontend-app-api: add lazy external route ref resolution

Rugvip:

There's a gap in the new frontend system where modules and plugin overrides don't allow you to properly use external route refs, since they currently need to be installed via a plugin instance. This change simply enables the `defaultTarget` of external route refs to be evaluated at any time, allowing them to function as intended. The downside is of course that the ref won't be made available for configuration, but that isn't particularly important given that these are modules anyway. There is still a bit of a gap where published modules don't have a way to provide configurable external route refs, but I think we can wait and see if that's actually an issue in practice.


packages/frontend-app-api/src/routing/RouteResolver.ts
```diff
@@ -52,54 +55,50 @@ export function joinPaths(...paths: string[]): string {
...
+
+  if (isExternalRouteRef(ref)) {
+    let resolvedRoute = routeBindings.get(ref);
     if (!resolvedRoute) {
-      return [undefined, ''];
+      const internal = toInternalExternalRouteRef(ref);
+      const defaultTarget = internal.getDefaultTarget();
+      if (defaultTarget) {
+        resolvedRoute = routeRefsById.get(defaultTarget);
+      }
```
Rugvip, line 75: This snippet is the new bit, the rest is just restructuring to simplify the logic

PR #30914, opened 2025-08-18
fix(catalog): sort default spec fields for stable hash

drodil:

sort fields of known spec relation fields so that the default processing engine hash calculation is more stable leading to less re-stitching of entities

### DIFF
--- a/plugins/catalog-backend/src/processors/BuiltinKindsEntityProcessor.test.ts
+++ b/plugins/catalog-backend/src/processors/BuiltinKindsEntityProcessor.test.ts
@@ -26,6 +26,49 @@ import {
 import { BuiltinKindsEntityProcessor } from './BuiltinKindsEntityProcessor';
 
 describe('BuiltinKindsEntityProcessor', () => {
+  describe('preProcessEntity', () => {
+    const processor = new BuiltinKindsEntityProcessor();
+    afterEach(() => jest.resetAllMocks());
+
+    it('should order relation fields correctly', async () => {
+      const entity: ComponentEntity = {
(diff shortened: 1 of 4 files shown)


plugins/catalog-backend/src/processors/BuiltinKindsEntityProcessor.test.ts
```diff
@@ -26,6 +26,49 @@ import {
...
+        kind: 'Component',
+        metadata: { name: 'n' },
+        spec: {
+          type: 'service',
+          owner: 'o',
+          subcomponentOf: 's',
+          lifecycle: 'l',
+          providesApis: ['b', 'a'],
+          consumesApis: ['c', 'x'],
+          dependsOn: ['Resource:r', 'Component:d'],
```
freben: i know it doesn't really matter for the test, but maybe make these lowercase just for the sake of good form

plugins/catalog-backend/src/processors/BuiltinKindsEntityProcessor.ts
```diff
@@ -83,6 +84,27 @@ export class BuiltinKindsEntityProcessor implements CatalogProcessor {
     return false;
   }
 
+  async preProcessEntity(entity: Entity): Promise<Entity> {
+    function sortField(field: string) {
+      const value = get(entity, field);
+      if (!value || !Array.isArray(value)) {
```
freben: i wouldn't mind being extra mega-explicit here and checking `value.every(v => typeof v === 'string')` too while we're at it.

Maybe invert the condition and do the work inside the `if` too?

PR #30915, opened 2025-08-18
`kubernetes-backend`: Added the ability to get access to default implementations in extension points

benjdlambert:
Following on from the deprecation and removal of the `KubernetesBuilder` from the public API, the existing extension point pattern was not enough if you only wanted to customize a certain element of the builder.

The real underlying issue is that you can't get access to any of the default implementations of any of the supporting services, so you would have to create your own everything if just wanting to customize one part so you can use the extension points to override all the services.

This is an alternative implementation as opposed to moving them to actual service refs instead, which may be a followup improvement if this pattern does not suffice.

This change is non-breaking, but allows the integrator to pass in a callback to the extension point, which will be given default implementations for all of the underlying services and dependencies.

### DIFF
--- a/docs/features/kubernetes/installation.md
+++ b/docs/features/kubernetes/installation.md
@@ -123,7 +123,8 @@ export const kubernetesModuleCustomClusterDiscovery = createBackendModule({
         clusterSupplier: kubernetesClusterSupplierExtensionPoint,
         serviceLocator: kubernetesServiceLocatorExtensionPoint,
       },
-      async init({ clusterSupplier }) {
+      async init({ clusterSupplier, serviceLocator }) {
+        // simple replace of the internal dependency
         clusterSupplier.addClusterSupplier(
           CustomClustersSupplier.create(Duration.fromObject({ minutes: 60 })),
         );
@@ -136,7 +137,7 @@ export const kubernetesModuleCustomClusterDiscovery = createBackendModule({
             const defaultImplementation = await getDefault();
 
             // build your own with the clusterSupplier dependency:
-            return MyNewServiceLocator({ clusterSupplier });
+            return new MyNewServiceLocator({ clusterSupplier });
           },
         );
       },


plugins/kubernetes-node/report.api.md
```diff
@@ -95,7 +97,15 @@ export interface KubernetesClustersSupplier {
...
   // (undocumented)
-  addClusterSupplier(clusterSupplier: KubernetesClustersSupplier_2): void;
+  addClusterSupplier(
+    clusterSupplier:
+      | KubernetesClustersSupplier_2
+      | (({
+          getDefault,
+        }: {
+          getDefault: () => KubernetesClustersSupplier_2;
+        }) => KubernetesClustersSupplier_2),
```
Rugvip:
```suggestion
      | (({
          getDefault,
        }: {
          getDefault: () => Promise<KubernetesClustersSupplier_2>;
        }) => Promise<KubernetesClustersSupplier_2>),
```

Thinking this is safer?

plugins/kubernetes-node/report.api.md
```diff
@@ -95,7 +97,15 @@ export interface KubernetesClustersSupplier {
 // @public
 export interface KubernetesClusterSupplierExtensionPoint {
   // (undocumented)
-  addClusterSupplier(clusterSupplier: KubernetesClustersSupplier_2): void;
+  addClusterSupplier(
+    clusterSupplier:
+      | KubernetesClustersSupplier_2
+      | (({
+          getDefault,
+        }: {
```
freben: Remove all these eager unpackings of arguments to make the report more easy to read.

Also wouldn't mind `addClusterSupplier(supplier: KubernetesClustersSupplier | KubernetesClustersSupplierFactory)` and have a named type for them, especially once the number of fields given becomes extensive

plugins/kubernetes-backend/src/service/KubernetesInitializer.ts
```diff
@@ -0,0 +1,242 @@
...
+import { DispatchStrategy } from '../auth/DispatchStrategy';
+import { CatalogService } from '@backstage/plugin-catalog-node';
+import { buildDefaultServiceLocator } from '../service-locator/buildDefaultServiceLocator';
+import {
+  ALL_OBJECTS,
+  DEFAULT_OBJECTS,
+  KubernetesFanOutHandler,
+} from './KubernetesFanOutHandler';
+
+export class KubernetesInitializer {
```
Rugvip, line 48: image

benjdlambert: hahaha, I really wanted to move away from the name `KubernetesBuilder` to indicate a move away from that pattern being exposed. Can you tell? :eyes: 😆

PR #30922, opened 2025-08-18
fix: remove domain attribute to set OAuth cookies as host-only

JessicaJHee:

This PR unsets the domain attribute in the default cookie configurer to make OAuth cookies host-only to prevent duplicate-cookie sign-in issues.

Related to https://github.com/backstage/backstage/issues/28126
- PR is a continuation of https://github.com/backstage/backstage/pull/29343

### DIFF
--- a/.changeset/wet-kiwis-strive.md
+++ b/.changeset/wet-kiwis-strive.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-auth-node': patch
+---
+
+Changes OAuth cookies from domain-scoped to host-only by avoid setting the domain attribute in the default cookie configurer.
--- a/plugins/auth-node/src/oauth/OAuthCookieManager.ts
+++ b/plugins/auth-node/src/oauth/OAuthCookieManager.ts
@@ -47,7 +47,7 @@ const defaultCookieConfigurer: CookieConfigurer = ({
     ? pathname.slice(0, -'/handler/frame'.length)
     : `${pathname}/${providerId}`;
 
-  return { domain, path, secure, sameSite };
+  return { path, secure, sameSite };
 };
 
 /** @internal */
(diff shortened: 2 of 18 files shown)


.changeset/wet-kiwis-strive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-auth-node': patch
+---
+
+Unsets domain attribute of default cookie configurer to set OAuth cookies as host-only
```
freben: These words need to be added to `.github/vale/config/vocabularies/Backstage/accept.txt`

plugins/auth-node/src/oauth/OAuthCookieManager.ts
```diff
@@ -184,6 +184,18 @@ export class OAuthCookieManager {
...
+    if (
+      this.cookieConfigurer === defaultCookieConfigurer &&
+      !this.getRemoveCookieOptions().domain
+    ) {
+      const { hostname: domain } = new URL(this.options.callbackUrl);
+      output = output.cookie(name, '', {
+        ...this.getRemoveCookieOptions(),
+        domain,
+      });
+    }
```
Rugvip: Notes from the framework SIG on what aspects are important to test in various browsers:

Given a state where they user has an existing cookie, what happens?
- Best: The cookie is re-used but switched out for one without domain on sign-in
- Good: The existing cookie is ignored and the new cookie is picked up correctly
- Bad: The existing cookie takes precedence and we're unable to write any new cookies

JessicaJHee: Did more testing and found that:

When the user has an existing cookie like `github-refresh-token` with the domain set to `.backstage.local` (not host-only), it will land in the best scenario you described.

i.e. The old cookie prefixed by `.` is deleted and replaced with a new cookie with `backstage.local` as the domain, and the new one is issued without the user needing to sign in again

freben (approved):
@JessicaJHee Am I reading it right that your testing pointed to this probably being safe? Happy to approve on my end; we just made a mainline release so that gives this some time to brew

JessicaJHee:
@freben Yes that's right! I tested with Chrome and safari and it's working as expected :) 

> we just made a mainline release so that gives this some time to brew

That's good to know 👍

PR #30930, opened 2025-08-19
microsite: add datacontract plugin

remunda:
Hey Backstage devs :)

I've created a plugin that allows to ingest Data Contract YAML files via API Entities (type=datacontract)

see docs for details https://www.npmjs.com/package/@remunda/backstage-plugin-datacontract

Thank you!

### DIFF
--- a/microsite/data/plugins/datacontract.yaml
+++ b/microsite/data/plugins/datacontract.yaml
@@ -1,0 +1,10 @@
+---
+title: Data Contract
+author: Jan Remunda
+authorUrl: https://www.remunda.cz/#backstage
+category: Discovery
+description: Ingest and visualize Data Contracts to your catalog via API Entities.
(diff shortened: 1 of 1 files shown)


freben (approved):
Hey, a few notes, but approving

* The package exports a module AND a plugin. The plugin is unnecessary and can be removed. If you DO want to make one package expose multiple things you can [use a feature loader](https://backstage.io/docs/reference/backend-plugin-api.createbackendfeatureloader/) as the default export.
* I realise that validation is useful, but at the same time it may not be the right place to do so in a processor. Sometimes right, sometimes wrong. It's a bit obscure and frustrating for devs to only be informed of model errors deep inside some processing loop - it may be worth surfacing errors in some different way, preferably all the way up at the editing phase if posslble.

remunda:
Hi,

thanks a lot for the feedback! 
I'll try to change the validation so issues will be visible in the UI. The plugin export isn't necessary.

PR #30936, opened 2025-08-20
[NFS] (create-app): redirect from home to catalog

camilaibs:

Creates a plugin that redirects from the Home page to the Catalog index page to avoid seeing a not found page error when starting the app.

### DIFF
--- a/packages/create-app/templates/next-app/packages/app/src/plugins/home/index.tsx
+++ b/packages/create-app/templates/next-app/packages/app/src/plugins/home/index.tsx
@@ -1,0 +1,8 @@
+
+import { createFrontendPlugin } from '@backstage/frontend-plugin-api';
+import { HomePage } from './HomePage';
+
+export const homePlugin = createFrontendPlugin({
+  pluginId: 'home',
+  extensions: [HomePage],
+});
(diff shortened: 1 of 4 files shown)


packages/create-app/templates/next-app/packages/app/src/plugins/home/index.tsx
```diff
@@ -0,0 +1,8 @@
+
+import { createFrontendPlugin } from '@backstage/frontend-plugin-api';
+import { HomePage } from './HomePage';
+
+export const homePlugin = createFrontendPlugin({
```
benjdlambert: I wonder if this would be better as a module instead?

```ts
createFrontendModule({
  pluginId: 'app',
  extensions: [HomeRedirect]
})
```

I think it's just a little confusing to have a `home` plugin already when one exists but it just doesn't work yet.

camilaibs: I tried to use module before, and it doesn't work. Unless I'm doing something wrong, I think it only overrides a page if a plugin with the plugin id actually exist

benjdlambert (approved):

freben (approved):

PR #30940, opened 2025-08-20
Patch release of #30936, #30938

benjdlambert:
This release fixes two small issues uncovered with the new frontend system.

- Fixes support for the deprecated `default*` props and packages published with an older `catalog-react` API that happen to get the latest version.

- Fixes missing `/` route when using `npx @backstage/create-app --next`

### DIFF
--- a/packages/create-app/CHANGELOG.md
+++ b/packages/create-app/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @backstage/create-app
 
+## 0.7.3
+
+### Patch Changes
+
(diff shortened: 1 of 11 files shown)


packages/create-app/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @backstage/create-app
 
+## 0.7.3
+
+### Patch Changes
+
+- 0f0dec3: Creates a plugin that redirects from the Home page to the Catalog index page to avoid seeing a not found page error when starting the app.
```
freben, line 7: this message isn't really accurate now, but it is what it is

benjdlambert: can edit after

PR #30953, opened 2025-08-21
`k8s`: fix issue with `objectsToFetch` defaulting

benjdlambert:
Previously when this was marked as `undefined` it would get defaulted in to `DEFAULT_OBJECTS`.

I incorrectly marked the defaults as an empty array, so it never got defaulted correctly, but we need to default them to something pretty early.

This marks them as the `DEFAULT_OBJECTS` a little ealier.

### DIFF
--- a/.changeset/ripe-plants-pump.md
+++ b/.changeset/ripe-plants-pump.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-kubernetes-backend': patch
+---
+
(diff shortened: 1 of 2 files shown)


freben (approved):
Might be worth adding a test for, at least after patching

benjdlambert:
Yeah there's a bunch of tests that are missing from the previous builder and this plugin. Will create some in a followup.

PR #30954, opened 2025-08-21
`nfs`: fix `compatWrapper` behavior in NFS to not render `ForwardsCompatWrappers`

benjdlambert:
This causes some of the hanging when migrating your apps. There's a recursion that happens with the polyfill we have for `componentRefs` and `SwappableComponent` which should not be happening.

This should hopefully fix the crashing of the tabs when working through the migration guide.

### DIFF
--- a/.changeset/sixty-places-push.md
+++ b/.changeset/sixty-places-push.md
@@ -1,0 +1,5 @@
+---
+'@backstage/core-compat-api': patch
+---
+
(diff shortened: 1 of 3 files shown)


camilaibs (approved):

jboeijenga:
Pulled these changes locally and can confirm this fixes the "hanging browser" issue while migrating my application.

PR #30969, opened 2025-08-22
feat(catalog): processor processor

### LINKED ISSUE #30963: Catalog processor priority
### 📜 Issue Labels

### 🔎 Search Terms

```plain
catalog processor order
```

### 🗃️ Project Area

Catalog

### 🔖 Need

As stated in docs (by PR by me), order of the processors is random between catalog modules. This can have really random effects if not carefully planned. For example if some other processor adds an annotation other needs, there's no way to guarantee the order of these processors. Of course you can manage some of these with pre/post process options, but that only takes you some way.

### 📝 Proposal

My proposal is to add a new optional `getProcessorPriority` method to the catalog processor interface. By default, the priority could be for example 10. When all processors are finally registered, they would be sorted based on the priority (lowerst first). This allows more control for processor order even over module limits. 

Additionally it would make sense to have a possibility to override the priority via config, for example for built-in or 3rd party processors. 

### 🔄 Alternatives

Creating a graph out of processors based on their dependencies. Implementation in #30970 

drodil:

This change enables the ordering of catalog processors by their priority, allowing for more control over the catalog processing sequence. The default priority is set to 20, and processors can be assigned a custom priority to influence their execution order. Lower number indicates higher priority. The priority can be set by implementing the `getPriority` method in the processor class or by adding a `catalog.processors.<processorName>.priority` configuration in the `app-config.yaml` file. The configuration takes precedence over the method.

closes #30963


drodil:
Not stale

plugins/catalog-backend/config.d.ts
```diff
@@ -218,5 +218,20 @@ export interface Config {
      * housing catalog-info files.
      */
     processingInterval?: HumanDuration | false;
+
+    /**
+     * Catalog processor configuration. Key is the processor name for example
+     * `BuiltinEntityKindProcessor`.
+     */
+    processors?: {
```
vinzscam: as an alternative we could order the processors as a list? 🤔

drodil: Yes, but it's a bit different than the current config for existing providers where the key is the provider name. I think having consistent stucture is important in config schema.

drodil: Oh, did you mean that just a list of processor names would be the priority? If so, I think that prevents any additional configuration processors may have in the future 🤔

plugins/catalog-backend/src/service/CatalogBuilder.ts
```diff
@@ -722,6 +722,24 @@ export class CatalogBuilder {
 
     this.checkMissingExternalProcessors(processors);
 
+    // Lastly sort the processors by priority. Config can override the
+    // priority of a processor to allow control of 3rd party processors.
+    processors.sort((a, b) => {
+      const aPriority =
+        config.getOptionalNumber(
```
freben: Alright - should we rebase this on top of latest master now that https://github.com/backstage/backstage/pull/31034 is merged, and break out this prio reading into a helper function that does try/catch to be safe in the same way?

plugins/catalog-node/src/api/processor.ts
```diff
@@ -100,6 +100,14 @@ export type CatalogProcessor = {
...
     cache: CatalogProcessorCache,
   ): Promise<Entity>;
+
+  /**
+   * Returns a processor priority, which is used to determine the order in which
+   * processors are run. The default priority is 20, and lower value means
+   * that the processor runs earlier.
+   * @returns A number representing the priority of the processor.
+   */
+  getPriority?(): number;
```
freben, line 110: I was wondering for a bit whether we could make this a config-only thing but I agree that it makes sense to have as code as well to make it easy for people to just add individual new things without being forced to configure as well

plugins/catalog-backend/config.d.ts
vinzscam: yep you're right 👍

plugins/catalog-backend/src/service/CatalogBuilder.ts
drodil: Yep. And add the try-catch, just a sec.

plugins/catalog-node/src/api/processor.ts
drodil: Yep, I think the "default" should be possible to be defined by the processor creator, but still keep the option to go around it with config.

plugins/catalog-backend/config.d.ts
```diff
@@ -223,6 +223,7 @@ export interface Config {
     processingInterval?: HumanDuration | false;
 
     /**
+<<<<<<< HEAD
```
freben:
```suggestion
```

drodil: fcckk :D

PR #30974, opened 2025-08-22
feat(scaffolder-backend): add defaultEnvironment config to scaffolder

### LINKED ISSUE #9461: Provide some default values + secrets for the Scaffolder context
Following on from the closed #6632, it would be great if we could define some default values that can be used in the template manifest (`template.yaml`) and also in `fetch:template`.

## Feature Suggestion

The aforementioned issue was closed because the solution of having an action that could use `ctx.output('config', configApi.get('scaffolder.environment'));` was enough to satisfy the request.

The template author could use the output of one action to template values into the next using `${{ ctx.outputs.myStep.blah }}` to template that into a succeeding action.

The problem that we have with that solution is there is no first class support for secrets, and how to have some default secrets context that can be used in the templating manifest file or in the actions under `ctx.secrets`.

## Possible Implementation

So first off, I'm contemplating removing the `ctx.secrets` way of accessing secrets in actions, and thinking that it would be better if these things were passed in through `inputs` to the action and are just templated in rather than the action assuming that secrets exist. It's going to make action writing a little less coupled to the context and things that might or might not exist there. Can probably address this issue separately though.

In order to supply some default parameters, and default secrets, we could go two ways. One I think is simpler, and one uses some more of the more known capabilities of config that backstage integrators and plugin developers might be aware of.

### Open Config approach

We could have the option to provide a default environment that would get injected into the nunjucks templating in `app-config.yaml`

```yaml
# app-config.yaml 
...
scaffolder:
  defaultEnvironment:
    parameters:
      region: eu-west-1
    secrets:
      AWS_ACCESS_KEY: ${AWS_ACCESS_KEY}
...
```

The above is a pretty simple config, and pretty clear distinction between secrets and parameters that could provide some defaults at the templating level in `template.yaml`. 

`${{ parameters.region }}` would be resolved to `eu-west-1` unless the `parameters` in the `template.yaml` had an option to override that. And the same for `${{ secrets.AWS_ACCESS_KEY }}`, unless this was overridden when creating the task by using the `SecretsContext` or/and a `CustomFieldExtension`.

### `config.d.ts` approach

It would also be possible for developers to just define a `defaultEnvironment`, and the keys under that object would be subject to `secret` or `backend` visibility part of a `config.d.ts`.

## Usage

The idea here, is that in the `betav3` versions of templates, you could set up your `defaultEnvironment` either way, and be able to do things like this:

```yaml
apiVersion: scaffolder.backstage.io/v1beta3
kind: Template
metadata:
  name: v1beta3-demo
  title: Test Action template
  description: scaffolder v1beta3 template demo
spec:
  owner: backstage/techdocs-core
  type: service

  parameters:
    ...

    - title: Choose a location
      properties:
        region:
          title: Region
          type: string
    ...

  steps:
    ...

    - id: template
      name: Create Repo
      action: fetch:template
      input:
        region:  ${{ parameters.region }}
    - id: push
      name: Push to S3
      action: publish:s3
      input:
        region:  ${{ parameters.region }}
        accessToken: ${{ secrets.AWS_ACCESS_TOKEN }}
    ...

### PR DESCRIPTION
**Pull request by angeliski**

Just a way to closes #9461 

### CODE REVIEW

File: plugins/scaffolder-backend/sample-templates/notifications-demo/template.yaml [hunk 1]
```diff
@@ -67,7 +67,7 @@ spec:
         recipients: ${{ parameters.recipients }}
         entityRefs: ${{ parameters.entityRefs }}
         title: ${{ parameters.title }}
-        description: ${{ parameters.description }}
+        info: ${{ parameters.description }}
```
angeliski: Catch this when testing the template, description isn't in the action 😅

plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
```diff
@@ -495,8 +499,16 @@ export class NunjucksWorkflowRunner implements WorkflowRunner {
       const taskTrack = await this.tracker.taskStart(task);
       await fs.ensureDir(workspacePath);
 
+      const defaultEnvironment = this.options.config
+        ? resolveDefaultEnvironment(this.options.config)
+        : { parameters: {}, secrets: {} };
+
+      // optional to avoid break changes
+      task.enrichSecrets?.(defaultEnvironment.secrets);
```
angeliski: I choose to use the current secrets in the task to avoid the necessity to handle those "new secrets" in every place

plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
```diff
@@ -495,8 +499,16 @@ export class NunjucksWorkflowRunner implements WorkflowRunner {
...
+
       const context: TemplateContext = {
         parameters: task.spec.parameters,
+        environment: defaultEnvironment.parameters,
```
angeliski: I used the suggestion `environment` from the initial issue, but the config `defaultEnvironment.parameters` is a little weird. I will try to think about that

plugins/scaffolder-node/src/tasks/types.ts
```diff
@@ -169,6 +169,8 @@ export interface TaskContext {
   getWorkspaceName(): Promise<string>;
 
   getInitiatorCredentials(): Promise<BackstageCredentials>;
+
+  enrichSecrets?(secrets: TaskSecrets): void;
```
angeliski: Used optional here to avoid break changes for now

plugins/scaffolder-backend/src/scaffolder/tasks/StorageTaskBroker.ts
```diff
@@ -108,8 +109,11 @@ export class TaskManager implements TaskContext {
     return this.signal;
   }
 
-  get secrets() {
-    return this.task.secrets;
+  get secrets(): TaskSecrets | undefined {
+    return {
```
acierto: I wonder if this small change can break something for someone, as it will never be undefined anymore, but rather empty object.

angeliski: Maybe 🤔 It's a small indirect behavior change, so yeah, something can go wrong. I can use some if's here to return undefined, but I prefer to avoid that (and in the future remove the undefined behavior). WDYT?

angeliski:
Hey @benjdlambert could you give some feedback here, please?

plugins/scaffolder-node/report.api.md
```diff
@@ -431,6 +431,8 @@ export interface TaskContext {
   // (undocumented)
   emitLog(message: string, logMetadata?: JsonObject): Promise<void>;
   // (undocumented)
+  enrichSecrets?(secrets: TaskSecrets): void;
```
benjdlambert: Can we skip this?

plugins/scaffolder-backend/src/scaffolder/tasks/StorageTaskBroker.ts
```diff
@@ -216,6 +220,10 @@ export class TaskManager implements TaskContext {
     }
     return this.auth.getNoneCredentials();
   }
+
+  enrichSecrets(secrets: TaskSecrets): void {
```
benjdlambert: Is there a way we can avoid this entirely, and make the secrets part of the original `this.task.secrets`

Or we just move the decoration out higher up into the `NunjucksWorkflowRunner` instead to that it just gets added there instead?

benjdlambert (changes requested):
Hey :wave: sorry for the late review here, had some holidays and didn't see this before heading out!

Thanks for taking a stab at this! Really nice!

I guess I have a higher level question that if we're planning on merging secrets so `${{ secrets.x }}` can come from config, is it still desired that we have `${{ environment.parameters.x }}`?

Or would it be more expected that it all just get's merged onto `${{ parameters.x }}` instead?

Could lead to some weirdness in the templates and collisions, but maybe that's fine?

plugins/scaffolder-backend/src/scaffolder/tasks/StorageTaskBroker.ts
angeliski: We have some options to consider:

Option 1: Use the creation point in the router ([here](https://github.com/backstage/backstage/blob/master/plugins/scaffolder-backend/src/service/router.ts#L588)) to enrich the secrets and achieve the same result.

Option 2: Use the broker ([here](https://github.com/backstage/backstage/blob/master/plugins/scaffolder-backend/src/scaffolder/tasks/StorageTaskBroker.ts#L351)) to enrich the secrets and achieve the same result.

I chose to avoid Option 1 because it creates a frozen snapshot of the environment when the task is created. If you update your environment and try to run the task again, nothing changes. While this provides more deterministic execution, I prefer getting fresh values from the environment at execution time.

Option 2 provides fresh values, but it has a significant drawback for anyone with a custom TaskBroker implementation, as it introduces an internal behavior expectation from the core (which I'd prefer to avoid).

Note: I just noticed that WorkflowRunner is also an interface 😅 So maybe put that here is bad too

That's why I chose the enrichSecrets approach.

I'm open to changing this to Option 1, or if you have other suggestions I could explore?

Perhaps we could add a decorator around the context in the [TaskWorker](https://github.com/backstage/backstage/blob/master/plugins/scaffolder-backend/src/scaffolder/tasks/TaskWorker.ts#L172) to encapsulate this behavior?

angeliski:
> I guess I have a higher level question that if we're planning on merging secrets so ${{ secrets.x }} can come from config, is it still desired that we have ${{ environment.parameters.x }}?
>
> Or would it be more expected that it all just get's merged onto ${{ parameters.x }} instead?
>
> Could lead to some weirdness in the templates and collisions, but maybe that's fine?

I like the approach to separate those values, maybe provide ${{ environment.secrets.x }} it's a better outcome? That will need more code to refactor and ensure secrets are not exposed, but the user can expect a more clear result 

I will try to work toward that to see how much effort is needed

angeliski:
Hey @benjdlambert and @acierto I tried a different approach, please give me some feedback about that :)

acierto:
> Hey @benjdlambert and @acierto I tried a different approach, please give me some feedback about that :)

Hey @angeliski, regarding option 1, it won't be frozen snapshot—every time you run the task, a new one is created. Reused values will only apply if you enable the experimental feature to make your task idempotent. However, this flow is better suited for scenarios where you deploy new versions while your application is running tasks, allowing them to automatically restart.

benjdlambert:
@angeliski honestly, the `TaskBroker` and `NunjucksWorkflowRunner` and all those interfaces are going away pretty soon and they should already be deprecated.

Thanks for having another stab at the change, but I think I'm fine with just making this change pretty simple for a start and having the `WorkflowRunner` just adding `environment: generateEnvironment()` every time rather than changing a lot of interfaces for now.

Let's start simple, and extrapolate from there later down the line if it's needed.

angeliski:
Hey @benjdlambert 
Thanks for the feedback

Just to ensure I got want you expect. Did you want to see something like that?
```ts
     await this.options.runners.workflowRunner.generateEnvironment();
      const { output } = await this.options.runners.workflowRunner.execute(
        task,
      );
```

I can put a new method in the `WorkflowRunner` interface and the `TaskWorker` (or other code) is responsible to call that. 
Is that the behavior you expect? Or something different?

benjdlambert:
Nearly, but not quite. I'm thinking we just move this inside the `execute` method inside the `NunjucksWorkflowRunner` now.

Also, a little update is that we deleted a lot of public API from the scaffolder now, so it's going to make this change easier to roll out either way. If you run `yarn build:api-reports` now you should have a lot less in terms of changes and what's exported.

angeliski:
Ok! I will refactor and get some feedback :)

angeliski:
Done @benjdlambert 

I expose the method in the `WorkflowRunner` because it makes sense to me that behavior be more explicit than implicit, but I can remove if necessary

plugins/scaffolder-backend/README.md
```diff
@@ -64,6 +64,67 @@ you will not have any templates available to use. These need to be [added to the
 To get up and running and try out some templates quickly, you can or copy the
 catalog locations from the [create-app template](https://github.com/backstage/backstage/blob/master/packages/create-app/templates/default-app/app-config.yaml.hbs).
 
+## Configuration
+
+### Default Environment
```
awanlin, line 69: Can we make sure this gets added to the docs over here please? https://backstage.io/docs/features/software-templates/configuration

awanlin (changes requested):
Hi @angeliski, thanks for the contributions, just reviewing from a docs perspective and left one comment related to that 👍

angeliski:
Sorry @awanlin and @benjdlambert I missed the last comment

Updated and ready to review! :)

.changeset/fruity-rivers-arrive.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-scaffolder-backend': minor
+'@backstage/plugin-scaffolder-node': patch
```
benjdlambert: Don't think we need `-node` package now as there's no changes in that package anymore :pray:

plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
```diff
@@ -103,10 +108,12 @@ const createStepLogger = ({
   task,
   step,
   rootLogger,
+  secretsForRedaction,
 }: {
   task: TaskContext;
   step: TaskStep;
   rootLogger: LoggerService;
+  secretsForRedaction?: string[];
```
benjdlambert: `additionalRedactions` instead maybe?

Otherwise we should just call this `redactions` and pass `{...task.secrets, ...environment.secrets }` instead?

benjdlambert (changes requested):
@angeliski sorry there's a conflict that's crept in here too.

benjdlambert:
Ping @angeliski :pray:

angeliski:
Sorry! I missed the notification, I will check the requests :)

angeliski:
Ready to go @benjdlambert 🙏🏾

Issue #30867, opened 2025-08-12
Redirect when navigating to a TechDocs URL for an entity that uses the `backstage.io/techdocs-entity` annotation


Frueber:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
TechDocs, backstage.io/techdocs-entity, URL, Redirect, area:techdocs, TechDocsReaderPage, TechDocsReaderPageProvider
```

### 🗃️ Project Area

TechDocs

### 🔖 Need

The `backstage.io/techdocs-entity` annotation allows you to link to another entity that declares TechDocs, and to share those TechDocs. Currently, if you attempt to go to a TechDocs formatted URL based on the entity using the `backstage.io/techdocs-entity` annotation then you get a 404 (mic drop ). Instead, it would be better to be routed to the TechDocs of the entity that's referenced with the `backstage.io/techdocs-entity` annotation.

This helps with:
1. Manually entered TechDocs URLs that lead to entities with a `backstage.io/techdocs-entity` annotation.
2. Prior bookmarked URLs of entities that switch to using the `backstage.io/techdocs-entity` annotation rather than declaring their own TechDocs for generation.
3. Clarifying the URLs to be used (with redirection) and the TechDocs page that is being used for the entity and the primary entity with the TechDocs is respected.

### 📝 Proposal

If you navigate to a TechDocs URL in the format `docs/{namespace}/{kind}/{name}` for an entity that has the `backstage.io/techdocs-entity` annotation (instead of the `backstage.io/techdocs-ref` annotation), then Backstage will redirect to the TechDocs page of the entity referenced in the value of that annotation.

*Suggestion*
It may be a better UI experience if the page shows as loading then either direct TechDocs load or a redirect occurs. Avoiding a 404 (mic drops) appearing prior to redirect is ideal.

#### Other-Option 1  

Instead of redirecting, the TechDocs page of the referenced entity, per the `backstage.io/techdocs-entity` annotation, would be shown directly under the URL used.

This is not recommended due to the increase in longer lasting URL usage with less clarity on the current TechDocs' entity (source). With this option it would be harder to discern where TechDocs are defined and where ownership lies.

#### Other-Option 2  

We could create an add-on but there are concerns with the order of rendering and page flashes, and the increased confusion when trying to get a somewhat simple feature into the mix.  

[...]

PR #30984, opened 2025-08-25
Adding handling which checks if the current entity (the catalog entit…

Frueber:
## Redirect when navigating to a TechDocs URL for an entity that uses the backstage.io/techdocs-entity annotation

Adding handling which checks if the current entity (the catalog entity being loaded) has an annotation for an external entity's TechDocs. If it does then we will redirect there rather than allowing a 404 (mic drop). This helps keep older URLs routing to the updated locations.

[Backstage Issue #30867](https://github.com/backstage/backstage/issues/30867)

tech-docs-external-entity-redirect

Steps to Test

1. Run the application with `yarn start` (see [Getting Started](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md#get-started)).
2. Navigate to [http://localhost:3000/docs/default/component/techdocs-entity-documented-component](http://localhost:3000/docs/default/component/techdocs-entity-documented-component). This URL is constructed as a TechDocs URL but is for an entity which currently points to another entity for TechDocs.
3. You should have been redirected to [http://localhost:3000/docs/default/Component/documented-component/inner-component-docs](http://localhost:3000/docs/default/Component/documented-component/inner-component-docs) per the `backstage.io/techdocs-entity` and `backstage.io/techdocs-entity-path` annotations on the newly added component.

### DIFF
--- a/docs/features/techdocs/FAQ.md
+++ b/docs/features/techdocs/FAQ.md
@@ -46,6 +46,10 @@ annotation should still be present in entity descriptor file (e.g.
 `catalog-info.yaml`) for Backstage to know that TechDocs is enabled for the
 entity.
 
+#### What happens when you navigate to a TechDocs URL for an entity uses the `backstage.io/techdocs-entity` annotation?
+
+If you navigate to a TechDocs URL in the format `docs/{namespace}/{kind}/{name}` for an entity that has the `backstage.io/techdocs-entity` annotation (instead of the `backstage.io/techdocs-ref` annotation), then Backstage will redirect to the TechDocs page of the entity referenced in the value of that annotation.
+
 #### Is it possible for users to suggest changes or provide feedback on a TechDocs page?
 
 This is supported for TechDocs sites whose source code is hosted in either
@@ -57,7 +61,3 @@ your `mkdocs.yml` files per
 If the host name of your source code hosting URL does not include `github` or
 `gitlab`, an `integrations` entry in your `app-config.yaml` pointed at your
 source code provider is also needed (only the `host` key is necessary).
-
-#### What happens when you navigate to a TechDocs URL for an entity uses the `backstage.io/techdocs-entity` annotation?
-
-If you navigate to a TechDocs URL in the format `docs/{namespace}/{kind}/{name}` for an entity that has the `backstage.io/techdocs-entity` annotation (instead of the `backstage.io/techdocs-ref` annotation), then Backstage will redirect to the TechDocs page of the entity referenced in the value of that annotation.


plugins/techdocs-backend/examples/documented-component/docs/inner-component-docs/index.md
```diff
@@ -0,0 +1,5 @@
+# Inner Component Docs
+
+This is a basic example of documentation within a larger suite of TechDocs that can be referenced by whatever entities necessary. It is intended as a showcase of the `backstage.io/techdocs-entity-path` annotation for linking to a "subpage" in TechDocs that are declared by another entity.
```
freben, line 3: You'll need to add the word "subpage" to `.github/vale/config/vocabularies/Backstage/accept.txt` to make the spell checker happy

plugins/techdocs/src/reader/components/TechDocsReaderPage/TechDocsReaderPage.test.tsx
```diff
@@ -94,6 +96,10 @@ const entityPresentationApiMock: jest.Mocked<
   }),
 };
 
+const catalogApiMock = {
+  getEntityByRef: jest.fn().mockResolvedValue(mockEntityMetadata),
+};
```
freben: You don't have to change this, but note that you can `import { catalogApiMock } from '@backstage/plugin-catalog-react/testUtils'`. That gives you a lot of power and type safety. You can make mocks,

```ts
const catalog = catalogApiMock.mock({
  getEntityByRef: async () => mockEntityMetadata,
})
```

and the return type still is a proper jest.fn - but much more importantly you can do

```ts
const catalog = catalogApiMock({
  entities: [mockEntityMetadata],
})
```

and that ends up being an implementation that has actual functioning methods. That way your test keeps working even if you start using something other than `getEntityByRef` in a refactor

plugins/techdocs-backend/examples/documented-component/docs/inner-component-docs/index.md
Frueber: Thank you for the callout -- updated 👍.

plugins/techdocs/src/reader/components/TechDocsReaderPage/TechDocsReaderPage.test.tsx
Frueber: Ohh, good tip. Makes sense to me. I've updated to using the `catalogApiMock` export though I'm not certain if the usage I went with is inline with your thinking.
Let me know if you think it's alright to continue, @freben 🙇‍♂️.

Frueber:
Hey @camilaibs , I may be mistaken but I believe you were added due to an addition to the spell checker word list.
Let me know if anything needs to be changed 🙇‍♂️, thank you!

docs/features/techdocs/FAQ.md
```diff
@@ -57,3 +57,7 @@ your `mkdocs.yml` files per
 If the host name of your source code hosting URL does not include `github` or
 `gitlab`, an `integrations` entry in your `app-config.yaml` pointed at your
 source code provider is also needed (only the `host` key is necessary).
+
+#### What happens when you navigate to a TechDocs URL for an entity uses the `backstage.io/techdocs-entity` annotation?
```
awanlin: Can we move this to be right under the question about the `backstage.io/techdocs-ref` annotation, please?

awanlin (changes requested):
Thanks for the contribution @Frueber, just looking over the docs changes as one of the Docs Maintainers and have one adjustment. 👍

docs/features/techdocs/FAQ.md
Frueber: I've updated as suggested. Thank you, @awanlin !

PR #30971, opened 2025-08-22
deprecate explore collator

freben:
This has now been moved. Do not merge before the actual `@backstage-community/plugin-search-backend-module-explore` package has been shipped to NPM.

PR #30985, opened 2025-08-26
Patch release of #30982, #30978, #30971

vinzscam:
This release fixes an issue in `@backstage/cli` that caused conflicts with detected modules in workspaces containing multiple apps. It also fixes the `new-frontend-plugin` template, which was incorrectly passing `id` instead of `pluginId`
 to `createFrontendPlugin`.

Additionally, the `@backstage/plugin-search-backend-module-explore` module has been marked as deprecated in favor of `@backstage-community/plugin-search-backend-module-explore`.


docs/features/search/collators.md
freben, line 1: 👍

PR #30987, opened 2025-08-26
Add user-settings-node package

grantila:

This PR adds a package `@backstage/plugin-user-settings-node` which exposes a backend service so that backend plugins can query and manage user settings.

This is useful for plugins that can perform backend actions based on user settings. Either settings only applicable for the backend, or keys (and types) shared with the frontend using a `-common` package. The API calls require a token, which can be a service token on-behalf-of a user token.

The PR also fixes a missing forwarding of the signal api in the frontend, when using `forBucket()`.


grantila:
Updated the backend service API to look as similar as possible to the frontend API. It reads the bucket namespace from a config now (`userSettings.namespace`), both the backend and the frontend. The backend service has `get()`, `set()` and `remove()` functions, as well as `forBucket()`, same as in the frontend.

plugins/user-settings-node/src/service.ts
```diff
@@ -0,0 +1,164 @@
...
+  defaultFactory: async service =>
+    createServiceFactory({
+      service,
+      deps: {
+        config: coreServices.rootConfig,
+        discoveryApi: coreServices.discovery,
+      },
+      async factory({ config, discoveryApi }) {
+        const namespace =
+          config.getOptionalString('userSettings.namespace') ?? 'default';
```
vinzscam, line 160: don't forget to define `userSettings.namespace` in a `config.d.ts` file

vinzscam:
looks great @grantila 🙏

grantila:
Thanks for the input @freben!

I sense some hesitation for this, and I get that adding a package with a somewhat fringe use case into the `/backstage` repo will mean more maintenance with very little value. There might be other (better) ways - let me explain the reasoning, and what could be an option.

For complex pages with several user settings, having to wait for each of them to finish fetching before things can be rendered is unideal, especially if the backend runs on HTTP/1.1 and you're far away from the backend (in terms of latency) and you run out of the usual 6 connections per origin. Letting a backend forward (just like you say, with on-behalf-of credentials) a set of user settings all-in-one solves this. That's the primary reason in this case.

Perhaps an option would be to add support for fetching multiple settings in one go, by introducing a new route (along side the [current one](https://github.com/backstage/backstage/blob/master/plugins/user-settings/src/apis/StorageApi/UserSettingsStorage.ts#L232)) that can read several settings and return them all? Combined with a small cache with a timeout of a few seconds in the `UserSettingsStorage`, a page can pre-fetch all settings, then allow child components to `get()` or `observe()`individual settings that would resolve to the cache initially. Would this be a better option? It would be a perfectly fine alternative for this use case.

freben:
Not really hesitant to having a node client that has the familiar get/put/delete - just hesitant to the token based interface and the namespacing.

Adding a multiget endpoint sounds cool too, either way, whether it's called from the frontend or the backend.

The default frontend client could be `DataLoader` backed internally and do batching calls transparently without anyone ever noticing. Well, except the inevitable delay window of a few milliseconds to roll up batches.

grantila:
Sounds like a good idea. DataLoader would be useful, but a small TTL-based cache of the result would be useful too, at least in my use case. They're complementary features.
The use case would be to pre-fetch all user settings for a page as it loads, then allowing sub components to also run `get()` or `observe()` as they're eventually rendered and re-use the cached value, instead of making further (unnecessary) calls.

There are probably more reasons for a brief cache; It's not uncommon that components in pages are rendered one after another with a short delay, due to data changing, and they might read the same user setting. So reusing a value within say 2 seconds makes sense IMO, to not fetch user settings over and over again.

I'll work on this and see how it ends up, perhaps we can close this PR in favor of multiget.

In any case, changing from token to credentials in this API would certainly be fine. The uglier issue though is that we need to share "namespace" between backend and frontend. It's used ([as bucket name](https://github.com/backstage/backstage/blob/master/plugins/user-settings/src/apis/StorageApi/UserSettingsStorage.ts#L230-L232)) for every key in the frontend, and it's [configurable there](https://github.com/backstage/backstage/blob/master/plugins/user-settings/src/apis/StorageApi/UserSettingsStorage.ts#L74). So that backend would need to know what namespace the frontend uses - preferably by a configuration option - to be able to read its values. So if this use case can be solved by multiget, it'd be far smaller and cleaner.

plugins/user-settings-node/src/service.ts
```diff
@@ -0,0 +1,164 @@
...
+   * Create a sub instance of UserSettingsService with a certain bucket.
+   */
+  forBucket(bucketName: string): UserSettingsService;
+
+  /**
+   * Get a user setting given a user token and key.
+   *
+   * Will throw NotFoundError or ResponseError upon failures.
+   */
+  get(userToken: string, key: string): Promise<UserSetting>;
```
vinzscam, line 45: we could make the api more evolvable 

```suggestion
  get(options: { userToken: string, key: string }): Promise<UserSetting>;
```

grantila:
For my use case, https://github.com/backstage/backstage/pull/31106 would suffice, if this node package isn't useful to othes

PR #30990, opened 2025-08-26
Added tooltip to advancesetting user option toggle

AmbrishRamachandiran:
**Current:**

image

**Changes:**

image

image

### DIFF
--- a/packages/app/src/components/advancedSettings/AdvancedSettings.tsx
+++ b/packages/app/src/components/advancedSettings/AdvancedSettings.tsx
@@ -22,6 +22,7 @@ import ListItem from '@material-ui/core/ListItem';
 import ListItemText from '@material-ui/core/ListItemText';
 import ListItemSecondaryAction from '@material-ui/core/ListItemSecondaryAction';
 import Switch from '@material-ui/core/Switch';
+import Tooltip from '@material-ui/core/Tooltip';
 import useLocalStorage from 'react-use/esm/useLocalStorage';
 
 export function AdvancedSettings() {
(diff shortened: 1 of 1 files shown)


AmbrishRamachandiran:
@freben  any thoughts? TIA

packages/app/src/components/advancedSettings/AdvancedSettings.tsx
```diff
@@ -39,18 +40,28 @@ export function AdvancedSettings() {
       <Grid item xs={12} md={6}>
         <InfoCard title="Advanced settings" variant="gridItem">
           <List>
-            <ListItem>
+            <ListItem
+              divider
+              button
+              onClick={() => setValue(value === 'on' ? 'off' : 'on')}
```
freben, line 46: There's already a `toggleValue` that the switch itself uses, it doesn't feel right to have two distinct ways to to achieve the same thing. You should probably rewrite `toggleValue` to be based on the previous value instead of `currentTarget`; then it can be used directly. And probably do `setValue(previous => ` to get access to the previous value without capturing it from the outside at all

packages/app/src/components/advancedSettings/AdvancedSettings.tsx
```diff
@@ -39,18 +40,28 @@ export function AdvancedSettings() {
...
-                />
+                <Tooltip
+                  placement="top"
+                  arrow
+                  title={value === 'on' ? 'Disable' : 'Enable'}
+                >
+                  <Switch
+                    color="primary"
+                    checked={value === 'on'}
+                    onChange={toggleValue}
```
freben, line 61: Is it safe to keep this one or should it be deleted, when the wrapping element has an onClick too? Do you need to stop the bubbling of events in `toggleValue`?

freben (commented):
This is just in the example app to be clear - doesn't get shipped

PR #30991, opened 2025-08-26
techdocs: Add transformer tests for sanitizing javascript: hrefs 

MegaManSec:

`techdocs` now checks whether a dangerous URL, using the `javascript` schema, is going to be written used for the `href` attribute of an `a` tag. In the event this does happen, the link will just be a text. This change brings `techdocs` in line with how `techdocs-node` works.

### DIFF
--- a/plugins/techdocs/src/reader/transformers/html/transformer.sanitizer.test.tsx
+++ b/plugins/techdocs/src/reader/transformers/html/transformer.sanitizer.test.tsx
@@ -143,4 +143,35 @@ describe('Transformers > Html > Sanitizer Custom Elements', () => {
     expect(elements).toHaveLength(1);
     expect(elements[0].hasAttribute('dominant-baseline')).toBe(true);
   });
+
+  it('removes javascript: hrefs while preserving link text', async () => {
+    const { result } = renderHook(() => useSanitizerTransformer(), { wrapper });
+    const dirtyDom = document.createElement('html');
+    const dirtyHTML = `
+      <body>
+        <!-- eslint-disable-next-line no-script-url -->
+        <a id="s1" href="javascript:alert(1)">JS 1</a>
+        <a id="s2" href=" javascript:alert(2)">JS 2 (leading space)</a>
+        <a id="s3" href="\n\tjavascript:alert(3)">JS 3 (whitespace)</a>
+        <!-- eslint-disable-next-line no-script-url -->
+        <a id="s4" href="JaVaScRiPt:alert(4)">JS 4 (mixed case)</a>
+        <a id="s5" href="javascript&#x3A;alert(5)">JS 5 (entity-encoded colon)</a>
+      </body>`;
+    dirtyDom.innerHTML = dirtyHTML;
+    const clearDom = await result.current(dirtyDom); // calling html transformer
+    const elements = Array.from(
+      clearDom.querySelectorAll<HTMLAnchorElement>('body > a'),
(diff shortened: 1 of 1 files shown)


GabDug:
Hey @MegaManSec! I'm trying to understand where it would be needed, my understanding was that the `sanitizerTransformer` with dompurify should do this kind of stuff? Thanks!

MegaManSec:
Howdy.

In theory, dompurify should handle it. But dompurify is [outdated](https://github.com/backstage/backstage/blob/6bf7c028a80b7fdc14e116f34f592e414aaab2b5/yarn.lock#L28366-L28378) (note: the vulnerability, CVE-2025-26791, would not affect the functionality changed here). dompurify is a protection, not a solution, I think.

MegaManSec:
Actually, I'm wrong: that CVE does specifically affect backstage.

https://github.com/backstage/backstage/blob/6bf7c028a80b7fdc14e116f34f592e414aaab2b5/plugins/techdocs/src/reader/transformers/html/transformer.ts#L17-L125 uses uses `ALLOWED_URI_REGEXP` and `CUSTOM_ELEMENT_HANDLING`, which that CVE affects specifically.

MegaManSec:
wdyt, @freben ?

GabDug:
While backstage's locked version for dompurify is outdated, latest versions are in range and can be installed by administrators in the meantime (e.g. as part of regular maintenance through `yarn npm audit`). 

dompurify 2.3.5 works properly with Backstage, 2.3.6 support needs a fix for techdocs redirects (relevant PR: https://github.com/backstage/backstage/pull/30723)

I think we could update the required version of dompurify to lessen the burden on administrators, and instead of adding custom processing, we could keep your tests but apply them to the `sanitizerTransformer`, to check dompurify's behaviours? 

Wdyt @MegaManSec?

MegaManSec:
Generally speaking, I think that's fine. However note, the "custom processing" is used throughout backstage, and this was the only place I could ńind where it was relevant such that it is not used (https://github.com/search?q=repo%3Abackstage%2Fbackstage+react-dom-bindings%2Fsrc%2Fshared%2FsanitizeURL&type=code)

plugins/techdocs/src/reader/transformers/rewriteDocLinks.ts
```diff
@@ -33,6 +38,12 @@ export const rewriteDocLinks = (): Transformer => {
             }
 
             try {
+              if (scriptProtocolPattern.test(elemAttribute)) {
+                throw new TypeError(
```
freben: Is a hard throw the right action here?

MegaManSec: afaict, yes: `// Non-parseable links should be re-written as plain text.` which would make sense for rewritten links: it can show a non-href'd `javascript:..` which i think is fine

MegaManSec:
Hey @PeaWarrior

Where is this performed in `sanitzerTransformer`? I do not see such a transformation. Or do you mean the `allowedURIRegExp` parameter to DOMPurify?

MegaManSec:
... in any case, done:)

PeaWarrior:
> Hey @PeaWarrior
> 
> Where is this performed in `sanitzerTransformer`? I do not see such a transformation. Or do you mean the `allowedURIRegExp` parameter to DOMPurify?

Yep, exactly!

PR #30993, opened 2025-08-26
Add detail in error messages when yarn plugin can't detect backstage version

rmartine-ias:

I was confused for ~15 minutes today when updating to using the yarn plugin. It was failing in docker but not locally, and I didn't know why. It turned out to be because I forgot to copy the backstage.json into the docker image. This was confusing, because the error seemed to indicate I was failing the semver checks. This change propagates error detail down the line, so people will see the actual cause. (i.e. missing file, no version field, semver wrong)

### DIFF
--- a/packages/yarn-plugin/package.json
+++ b/packages/yarn-plugin/package.json
@@ -31,6 +31,7 @@
   },
   "dependencies": {
     "@backstage/cli-common": "workspace:^",
+    "@backstage/errors": "workspace:^",
     "@backstage/release-manifests": "workspace:^",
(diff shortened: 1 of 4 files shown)


.changeset/two-moons-sneeze.md
freben, line 1: You can actually remove this file since the package is private (it gets shipped differently)

rmartine-ias: Done, removed it. Left it in because I saw a changelog file.

PR #31002, opened 2025-08-27
feat: Update backend/Dockerfile - `node:20` --> `node:22`

mathieu-benoit:

When using `create-app` CLI, the goal is to have the `Dockerfile` for `backend` using the `node:22` (instead of `node:20`) as the base image by default.

Related to this in the docs too: https://github.com/backstage/backstage/pull/31001.

### DIFF
--- a/packages/create-app/templates/default-app/packages/backend/Dockerfile
+++ b/packages/create-app/templates/default-app/packages/backend/Dockerfile
@@ -12,7 +12,7 @@
 # Alternatively, there is also a multi-stage Dockerfile documented here:
 # https://backstage.io/docs/deployment/docker#multi-stage-build
 
-FROM node:20-bookworm-slim
+FROM node:22-bookworm-slim
 
 # Set Python interpreter for `node-gyp` to use
 ENV PYTHON=/usr/bin/python3


freben:
Same question as the other PR - anything in particular that prompted you to make the change?

mathieu-benoit:
> Same question as the other PR - anything in particular that prompted you to make the change?

@freben, no particular issue. Just wanted to align with what's in the demo: https://github.com/backstage/demo/blob/master/Dockerfile too.

planeiii-te:
I just got bitten by this and was coming to make the same PR. [`Set` operations like intersection](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection) weren't added until Node v22.

freben:
@planeiii-te But was this a problem in a newly made create-app (as in, core features using those)?

To be clear, anyone is able to bump this in their own installation. The core offering is not eager to go to the newest though.

The same thing goes for everything, for example typescript. We don't eagerly move that to latest either.

mathieu-benoit:
> The core offering is not eager to go to the newest though.

@freben, interesting to learn and understand, why?

Couple of thoughts:
- `22` is the current `Active LTS` while `20` is `Maintence LTS` ([source](https://nodejs.org/en/about/previous-releases#looking-for-the-latest-release-of-a-version-branch))
- Backstage app support officially `20 || 22` [in the code](https://github.com/backstage/backstage/blob/master/package.json#L170)
- There was a [tentative to bump from `20` to `24`](https://github.com/backstage/backstage/pull/30093), it's maybe a good idea to default to `22` first to avoid a too big jump?

planeiii-te:
I misread the PR. This is the value that I wanted to upgrade:

https://github.com/backstage/backstage/blob/2926f773c2d73ca79cbca9b10e36fb178b28ccbf/packages/backend/Dockerfile#L12

I think that should also be covered in this PR.

> The same thing goes for everything, for example typescript. We don't eagerly move that to latest either.

@freben Why not? Why are you reluctant to upgrade?

freben:
No worries, I was mostly curious if there was a particular rush. Some prefer to stay back, some prefer to be on the edge.

We spoke about it a bit among the maintainers and this is fine. We should also update our versioning policy so there's a clear rule for when this create-app upgrade should happen going forward.

PR #31004, opened 2025-08-28
fix(catalog): msgraph to use only first parent for groups

drodil:

the order of the groups from msgraph is random, thus the parent group for groups having multiple parents can be random. this leads to situation where the grpup parent is changing over the msgraph provider runs leading to unnecessary stitching.

other possibility for the fix would be to change the model to allow multiple parents for groups (like mentioned in the TODO) but that could have some weird consequences in apps.


drodil:
bump, @freben any chance to get this for next release?

PR #31006, opened 2025-08-28
fix: deduplicate discovered backend features

stanislav-c:

This pull request makes a small improvement to the `PackageDiscoveryService` by ensuring that the returned `features` array contains only unique values. This allows using `discoveryFeatureLoader` with backend plugins that still export `alpha` modules. Without deduplication such plugins result in errors similar to the following:

```
2025-08-28T08:59:07.759Z backstage info Detected: @backstage/plugin-events-backend 
2025-08-28T08:59:07.768Z backstage info Detected: @backstage/plugin-events-backend 
2025-08-28T08:59:07.773Z rootHttpRouter info Listening on :7007 
Backend failed to start up Error: ExtensionPoint with ID 'events' is already registered
```

### DIFF
--- a/packages/backend-defaults/src/PackageDiscoveryService.ts
+++ b/packages/backend-defaults/src/PackageDiscoveryService.ts
@@ -175,6 +175,6 @@ export class PackageDiscoveryService {
       }
     }
 
-    return { features };
+    return { features: Array.from(new Set(features)) };
   }
 }
--- a/.changeset/afraid-streets-pay.md
+++ b/.changeset/afraid-streets-pay.md
@@ -1,0 +1,5 @@
+---
+'@backstage/backend-defaults': patch
+---
+
+Deduplicate discovered features discovered with discoveryFeatureLoader


freben:
Is this always safe to do? No risks involved with it?

stanislav-c:
> Is this always safe to do? No risks involved with it?

We have been running a fork of this for a few weeks now. Have not seen any issues.

I guess there is always a possibility that two different packages would export features with the same ID, but that would result in an error as before, since their ES modules would be different. Deduplication approach in this PR only handles exact ES module duplicates.

packages/backend-defaults/src/PackageDiscoveryService.ts
```diff
@@ -175,6 +175,6 @@ export class PackageDiscoveryService {
       }
     }
 
-    return { features };
+    return { features: Array.from(new Set(features)) };
```
camilaibs, line 178: Would it also work if we instead changed the type of the "features" variable to `Set<BackendFeature>`?

camilaibs: I also think it would be useful to have a warning saying that a duplicate was found and ignored, to somehow notify that this happened.

camilaibs (approved):
Thank you for the fix @stanislav-c 🙏🏻 
I would appreciate @Rugvip 's review as well

packages/backend-defaults/src/PackageDiscoveryService.ts
Rugvip: I think this is fine as is tbh, it's a nice and simple way to dedupe with strict equality comparisons. Don't think we really need a warning either, it wouldn't be particularly actionable most of the time and is very low risk to do this deduplication.

Rugvip (approved):
Thank you! 👍 

Makes sense, can't think of any cases where this could potentially lead to actual errors or swipe them under the rug. Let's :shipit:

PR #31007, opened 2025-08-28
Only run provider orphan cleanup if the engine is started

freben:
If you set `processingInterval: false`, you don't expect ANY aspect of the processing and provider loops to run. Especially if you have a read-write split; it doesn't make sense to have your read nodes try to perform deletions on their probably readonly restricted replica.

I would argue that connecting the providers should be moved in here too. But that could be thought of as a breaking change - what if people actually do metrics tracking or perform some other related tasks in these providers (such as reacting to webhooks)? So I chose to postpone that for now.

### DIFF
--- a/plugins/catalog-backend/src/service/CatalogBuilder.ts
+++ b/plugins/catalog-backend/src/service/CatalogBuilder.ts
@@ -622,18 +622,21 @@ export class CatalogBuilder {
       enableRelationsCompatibility,
     });
 
-    if (config.getOptionalString('catalog.orphanProviderStrategy') !== 'keep') {
-      await evictEntitiesFromOrphanedProviders({
-        db: providerDatabase,
-        providers: entityProviders,
-        logger,
-      });
-    }
     await connectEntityProviders(providerDatabase, entityProviders);
 
     return {
       processingEngine: {
         async start() {
+          if (
+            config.getOptionalString('catalog.orphanProviderStrategy') !==
+            'keep'
+          ) {
+            await evictEntitiesFromOrphanedProviders({
+              db: providerDatabase,
+              providers: entityProviders,
+              logger,
+            });
+          }
           await processingEngine.start();
           await stitcher.start();
         },
--- a/.changeset/whole-dingos-lay.md
+++ b/.changeset/whole-dingos-lay.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Only run provider orphan cleanup if the engine is started in the first place


plugins/catalog-backend/src/service/CatalogBuilder.ts
```diff
@@ -622,18 +622,21 @@ export class CatalogBuilder {
...
     });
 
-    if (config.getOptionalString('catalog.orphanProviderStrategy') !== 'keep') {
-      await evictEntitiesFromOrphanedProviders({
-        db: providerDatabase,
-        providers: entityProviders,
-        logger,
-      });
-    }
     await connectEntityProviders(providerDatabase, entityProviders);
```
Rugvip, line 625: Hold up, should it be the case for connecting entity providers too?

freben: Yeah see the PR comment. I absolutely think that connecting providers should be optional in SOME way. It'd be breaking if people for example use providers to react to webhook events through the connection but don't want to also run processing on the same node. So I left it out for now, and maybe we should make it a more explicit config option for fine grained control?

freben: @drodil @Rugvip also connecting the dots to https://github.com/backstage/backstage/pull/30969

Maybe we could have something like

```yaml
catalog:
  providers: false
```

and

```yaml
catalog:
  providers:
    - name: MyProvider
      // all options below optional and just examples
      enabled: false
      required: true // fail startup if not registered
      onInitError: continue
      // other options ...
      // this is a nice place for future expansion into having per-provider
      // actual options support that gets fed into a factory (or as a new field
      // on the connection during connect()), like for external access handlers etc
```

or something along those lines - and the same for processors - to enable ordering and enablement etc one by one or as a whole.

These could of course also be object shaped with an order field, as proposed in https://github.com/backstage/backstage/pull/30969, to better enable config merging - which of course has benefits and drawbacks too :)

drodil: Just small note that most of the providers use config where the provider name is the key of the config object, for example:

```ts
 catalog?: {
    providers?: {
      bitbucketServer?: {}
    }
}
```

freben: Right, so allowing `false` there would actually make their config loading attempt throw an exception even if they used `getOptionalConfig` right? That's a good callout

drodil: Yep. Started the work a bit with #31034 but it's still missing some of the features mentioned here. But it's easier to start small :)

vinzscam (approved):

PR #31010, opened 2025-08-28
feat(search): add actions to query search engine

drodil:

Added action for search backend to query search engine using the actions registry

the authorized search engine initialization had to be moved to plugin from router as it should also be used in the query action if permissions are enabled.

### DIFF
--- a/plugins/search-backend/src/actions/createQueryAction.ts
+++ b/plugins/search-backend/src/actions/createQueryAction.ts
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2025 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { z as zv3 } from 'zod/v3';
+import { JsonObject, JsonValue } from '@backstage/types';
+import { ActionsRegistryService } from '@backstage/backend-plugin-api/alpha';
+import { LoggerService } from '@backstage/backend-plugin-api';
+import { SearchEngine } from '@backstage/plugin-search-backend-node';
+import { SearchIndexService } from '@backstage/plugin-search-backend-node/alpha';
+import { filterResultSet, toSearchResults } from '../utils/search_result_utils';
+
+const jsonObjectSchema: zv3.ZodSchema<JsonObject> = zv3.lazy(() => {
+  const jsonValueSchema: zv3.ZodSchema<JsonValue> = zv3.lazy(() =>
+    zv3.union([
+      zv3.string(),
+      zv3.number(),
+      zv3.boolean(),
+      zv3.null(),
+      zv3.array(jsonValueSchema),
+      jsonObjectSchema,
+    ]),
+  );
+  return zv3.record(jsonValueSchema);
+});
+
+export const createQueryAction = ({
+  engine,
+  searchIndexService,
+  actionsRegistry,
+  logger,
+}: {
+  engine: SearchEngine;
+  searchIndexService: SearchIndexService;
+  actionsRegistry: ActionsRegistryService;
+  logger: LoggerService;
+}) => {
+  const allTypes = Object.keys(searchIndexService.getDocumentTypes());
+  const quotedTypes = allTypes.map(t => JSON.stringify(t)).join(', ');
+  const typesDescription =
+    allTypes.length > 0
+      ? `The supported document types are: ${quotedTypes}.`
+      : '';
+  actionsRegistry.register({
+    name: 'query',
+    title: 'Query Search Engine',
+    description: `
+This allows you to query the search engine for documents.
+You can search across all document types, or restrict the query to specific types.
+${typesDescription}
+Pagination is supported via the \`pageLimit\` and \`pageCursor\` parameters and is enabled by default with limit of 10.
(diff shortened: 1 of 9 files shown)


drodil:
Not stale

drodil:
Not stale, waiting for a review

plugins/search-backend/src/actions/createQueryAction.ts
```diff
@@ -0,0 +1,102 @@
...
+    `,
+    attributes: {
+      readOnly: true,
+    },
+    schema: {
+      input: z =>
+        z.object({
+          term: z.string().describe('The search term to query for'),
+          types: z
+            .array(z.string())
```
drodil: It would be nice to have support for defining the `input` in an async function, get the allowed types on demand from the `SearchIndexService`, and add them as an `enum` here. This would render the `getDocumentTypesAction` obsolete and improve the validation.

niallthomson:
I've been playing with an implementation of this here https://github.com/awslabs/backstage-plugins-for-aws/tree/main/plugins/genai/backend/src/tools

I found that it can return pretty verbose results due to the way the search API returns the full document for results. I ended up implementing "highlighting" myself (which I see I might have been able to inherit) and relied on separate tools to read the full content (entity, TechDocs page) if the model chose to.

I think it might also be a good idea to set a default page size.

plugins/search-backend/src/actions/createQueryAction.ts
```diff
@@ -0,0 +1,118 @@
...
+  engine,
+  searchIndexService,
+  actionsRegistry,
+}: {
+  engine: SearchEngine;
+  searchIndexService: SearchIndexService;
+  actionsRegistry: ActionsRegistryService;
+}) => {
+  const [firstType, ...otherTypes] = Object.keys(
+    searchIndexService.getDocumentTypes(),
```
drodil: This might actually not work as the searchIndexService might not be initialized 🤔

drodil: Actually it works with lifecyecle hook, great. Modifying this!

vinzscam:
cc @backstage/search-maintainers

thomvaill:
Hey! Thanks a lot for your PR! We am very interested in this search action.
It would enable us to make our Backstage instance like our "internal [context7](https://context7.com/)".
Can't wait to try it out.

However this makes me think: in the case of catalog entities, if the agent needs more information about it, it can use the `get-catalog-entity` action. But in the case of TechDocs, there is no similar action available.
This search action does return a `text` field, but if I understand correctly, it only contains the paragraph that matched, not the whole document? (which is the intended behavior imo)
The agent could use its web fetch tool to visit the URL directly but there would be authentication issues and would get HTML instead of raw text.
This is why I created this related issue to discuss creating another action to retrieve TechDocs content: https://github.com/backstage/backstage/issues/33120
I think the two actions would be complementary.
What do you think?

awanlin:
Hi @drodil, can you rebase and add this to the list of well-known actions here, please? https://github.com/backstage/backstage/blob/master/docs/ai/well-known-actions.md

awanlin (changes requested):
Thanks for this as always @drodil, can you rebase? I'd like to help move this forward. 👍

thenmozhi-loganathan:
> Hey! Thanks a lot for your PR! We am very interested in this search action. It would enable us to make our Backstage instance like our "internal [context7](https://context7.com/)". Can't wait to try it out.
> 
> However this makes me think: in the case of catalog entities, if the agent needs more information about it, it can use the `get-catalog-entity` action. But in the case of TechDocs, there is no similar action available. This search action does return a `text` field, but if I understand correctly, it only contains the paragraph that matched, not the whole document? (which is the intended behavior imo) The agent could use its web fetch tool to visit the URL directly but there would be authentication issues and would get HTML instead of raw text. This is why I created this related issue to discuss creating another action to retrieve TechDocs content: #33120 I think the two actions would be complementary. What do you think?

Hey! Thanks for the thoughtful feedback and for creating issue #33120!

You're absolutely right that the search action and a content retrieval action would be complementary. I've opened a PR that adds TechDocs MCP actions including the ability to retrieve TechDocs content: https://github.com/backstage/backstage/pull/33412

This should address the use case you described - agents can now search TechDocs and then fetch the full document content when needed, similar to how get-catalog-entity works for catalog entities.

Would love to hear your thoughts on the implementation!

drodil:
@awanlin rebased and added the docs!

plugins/search-backend/src/actions/createQueryAction.ts
```diff
@@ -0,0 +1,148 @@
...
+          hasMoreResults: z
+            .boolean()
+            .describe('Whether there are more results'),
+        }),
+    },
+    action: async ({ input, credentials }) => {
+      const resp = await engine.query(input, { credentials });
+      const safeResults = resp.results.map(result => ({
+        ...result,
+        document: {
```
alexlorenzi: Hey @drodil, nice work on this PR! I noticed that the Copilot review flagged both `authorization` stripping and unsafe location protocol filtering — it looks like you addressed the `authorization` part, but the protocol filtering may have been missed.

The existing `/query` REST endpoint in `router.ts` applies `filterResultSet()` to reject documents with unsafe location protocols (anything other than `http:` or `https:`, like `javascript:` or `data:` URIs). The new action handler doesn't apply the same filtering.

It might be worth extracting the `filterResultSet` / `allowedLocationProtocols` logic into a shared utility and applying it here too, so both the REST and actions paths have the same safety guarantees.

What do you think?

drodil: Yeah, maybe. I just wonder, shouldn't the filtering logic be actually in the collator part where the documents are ingested? That would probably be a big refactor. 

What separates the rest endpoint from action is that results from rest are rendered thus making it vulnerable while action output is only used by LLM or scaffolder task. Anyways, I can add it but it will take some time as I am on holiday at the moment 👍

drodil: Done and dusted.

awanlin (approved):

awanlin:
Hi @drodil, I've asked the TechDocs maintainers to give this another look. It feels very close to being approved and merged IMO!

alexlorenzi (approved):

PR #31011, opened 2025-08-28
Improve MCP backend error handling

grantila:

When using the MCP backend, an action that throws an error will turn into a "500 Internal Server Error".

For known Backstage errors e.g. `NotFoundError`, the error will now be converted into a textual response with the `isError` flag set to true. May of the errors that can be thrown aren't "internal server errors", but conflicts, not found, input error, etc - rather something like 400's.

This logic mimics the default backend [error handler](https://github.com/backstage/backstage/blob/923491eece9bd571aceb9d44f119db2fdd98c782/packages/backend-common/src/middleware/errorHandler.ts#L115)

### DIFF
--- a/plugins/mcp-actions-backend/src/services/handleErrors.ts
+++ b/plugins/mcp-actions-backend/src/services/handleErrors.ts
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2025 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


plugins/mcp-actions-backend/src/services/handleErrors.ts
```diff
@@ -0,0 +1,90 @@
...
+export async function handleErrors(
+  fn: () => RequestResultType | Promise<RequestResultType>,
+): Promise<RequestResultType> {
+  try {
+    return await fn();
+  } catch (err) {
+    // This will rethrow if the error is not a known Backstage error
+    const description = describeError(err);
+    return {
+      content: [{ type: 'text', text: description }],
```
freben, line 86: Was this array shape meant to be able to support actually BOTH the cause and the wrapper, where relevant? Is that at all interesting? Maybe a future extension

freben (approved):

plugins/mcp-actions-backend/src/services/handleErrors.ts
grantila: I think the array is for being able to return more things in the same response, like attachments, images, etc. Not very important for errors though 😏

PR #31021, opened 2025-08-29
cli: update backend plugin template with new service pattern

Rugvip:

Been using this pattern a bit for services and grown quite fond of it for any plugin-internal things. Using service refs for internal services makes plugin tests much nicer, and it reduces the amount of manual DI wiring you have to do by a lot. I think the existing pattern with closures can be quite confusing as well, and you still have to declare methods twice.

This new pattern lets you declare just the single implementation class and along with an accompanying service ref with a default factory. It removes the manual wiring in the plugin init, keeping it all in one file. The only real downside is that you have to use `typeof myServiceRef.T` when referring to the service type, at least if you want to be able to mock it cleanly.


packages/cli/templates/backend-plugin/src/plugin.test.ts.hbs
```diff
@@ -82,4 +89,41 @@ describe('plugin', () => {
       createdAt: expect.any(String),
     });
   });
+
+  it('should forward errors from the TodoListService', async () => {
```
Rugvip, line 93: Added this test to show how to mock services in a plugin integration test.

vinzscam (approved):
snyggt 👍

PR #31022, opened 2025-08-29
root: use pprettier

Rugvip:

This switches the main prettier runs from plain prettier to instead use `@mixer/parallel-prettier`, just a change in this repo for now.

Before:

```
Checking formatting...
All matched files use Prettier code style!
yarn prettier:check  75.92s user 3.98s system 137% cpu 58.263 total
```

Verifying the number of files checked from before:

```
yarn prettier:check --loglevel debug 2>&1 | grep "resolve config from" | wc -l
    7446
```

After:

```
✔ Checked 7446 files
yarn prettier:check  2.45s user 0.56s system 25% cpu 11.576 total
```

The config is a little bit more awkward because `@mixer/parallel-prettier` uses separate logic for enumerating files to check, which means that none of the defaults from prettier's own selection applies. This means we both need to add some extra ignores to `.prettierignore` and also have slightly more precise glob on the commandline or we'll spend a lot of time just enumerating and ignoring files in `node_modules` etc.


Rugvip:
In practice in CI it went from roughly 2:20 to 1:20. Not quite sure why the difference isn't bigger as we should still be able to parrallelize over 4 cores. Probably worth keeping the existing setup to avoid the extra complexity, and I'll explore using a cache instead.

PR #31023, opened 2025-08-29
github/workflows: add prettier cache

Rugvip:

Replaces #31022

Moving the prettier to the test step to be able to use the same cache, expecting impact to be minimal because it should now be much faster, but will verify that assumption.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,9 +102,6 @@ jobs:
       - name: type checking and declarations
         run: yarn tsc:full
 
-      - name: prettier
-        run: yarn prettier:check
-
         # We need to generate the API references as well, so that we can verify the doc links
       - name: check api reports and generate API reference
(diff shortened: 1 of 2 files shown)


Rugvip:
Yep, took 16s to run prettier this time around :shipit:

Ayushmore1214:
@Rugvip Just for learning purpose , this PR makes the prettier command execute faster right ?

Rugvip:
@Ayushmore1214 yep! you can for example compare the duration of the prettier step in an old version such as [this one](https://github.com/backstage/backstage/actions/runs/17322581862/job/49181187411?pr=31021) to [this PR](https://github.com/backstage/backstage/actions/runs/17323146335/job/49181553412?pr=31023).

Ayushmore1214:
Thanks , the difference seems to be huge from 2m 9s to 16s , that's crazy 😉

PR #31030, opened 2025-08-31
Create apicurio-registry.yaml metdata file for plugin directory

simonnepomuk:
Add apicurio-registry plugin to directory


microsite/data/plugins/apicurio-registry.yaml
```diff
@@ -0,0 +1,10 @@
+  ---
```
freben:
```suggestion
---
```

freben (approved):
Looks good!

PR #31035, opened 2025-09-01
Add RenderEdge"

MrOpperman:


freben:
There's a missing changeset, DCO signature is missing, there's an empty file, there are unrelated changes etc - this needs a careful eye and the builds going green before review

plugins/catalog-graph/src/components/EntityRelationsGraph/useEntityStore.ts
```diff
@@ -64,7 +64,7 @@ export function useEntityStore(): {
       return;
     }
 
-    const { items } = await catalogClient.getEntitiesByRefs({ entityRefs });
+    const { items } = await catalogClient.getEntities();
```
vinzscam: was this change intended?

MrOpperman: Sort of but not sure if it belongs here, unfortunately running this plugin locally doesn't work due to the the previous line. I can create a separate change for that or leave it in here. 

edit: decided to remove it.

MrOpperman:
@freben @vinzscam concerns have been addressed

MrOpperman:
@vinzscam here is a more elaborate example, similar to what we plan to use it for. 

Screenshot 2025-09-11 at 14 04 46

plugins/catalog-graph/src/components/EntityRelationsGraph/DefaultRenderEdge.tsx
freben, line 1: This file is empty. Did you mean to put the default implementation in here?

plugins/catalog-backend-module-gitea/package.json
freben, line 1: Unrelated change

packages/core-components/src/components/DependencyGraph/types.ts
```diff
@@ -89,6 +90,39 @@ export namespace DependencyGraphTypes {
...
+    id: dagre.Edge;
+  };
+
+  /**
+   * Custom React component for graph {@link DependencyGraphTypes.DependencyEdge}
+   *
+   * @public
+   */
+  export type RenderEdgeFunction<T = {}> = (props: {
+    edge: T & {
```
freben, line 102: Not sure what happened here - this was all meant to be part of `RenderEdgeProps` wasn't it? The other callbacks don't have this extra complexity in them.

freben (commented):
Do read the diff manually carefully

packages/core-components/src/components/DependencyGraph/types.ts
MrOpperman: It was, and I've updated it. Apologies for the oversight here.

.changeset/flat-peas-run.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/core-components': minor
+'@backstage/plugin-catalog-graph': minor
```
benjdlambert:
```suggestion
'@backstage/core-components': patch
'@backstage/plugin-catalog-graph': patch
```

Both of these packages are 0.x so you can mark them as patch changes which essentially is new feature for these packages.

Minor bumps are breaking changes for 0.x packages.

.changeset/flat-peas-run.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/core-components': patch
+'@backstage/plugin-catalog-graph': patch
+---
+
+Added `renderEdge` prop to <DependencyGraph /> component in @backstage/core-components to allow custom rendering of graph edges.
```
freben:
```suggestion
Added `renderEdge` prop to `<DependencyGraph />` component in `@backstage/core-components` to allow custom rendering of graph edges.
```

freben (approved):

packages/core-components/src/components/DependencyGraph/types.ts
```diff
@@ -93,6 +94,38 @@ export namespace DependencyGraphTypes {
...
+      width?: number;
+      height?: number;
+      weight?: number;
+      minlen?: number;
+      showArrowHeads?: boolean;
+      from?: string;
+      to?: string;
+      relations?: string[];
+    };
+    id: dagre.Edge;
```
vinzscam: how complex is this type? I think it would be good to hide the external dependency from api report

vinzscam (changes requested):

packages/core-components/src/components/DependencyGraph/types.ts
MrOpperman: It's not complex.   

```
export interface Edge {
    v: string;
    w: string;
    name?: string | undefined;
  }
```

I'll replace it.

Issue #31036, opened 2025-09-01
Actions registry permissions


drodil:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
actions registry, mcp, permissions
```

### 🗃️ Project Area

Core Framework

### 🔖 Need

To limit the actions users can run using the actions registry / MCP server, there should be integration with the permission framework.

### 📝 Proposal

Add action-specific resource permissions to the actions registry service. In the list action of the router, the user identity should be checked, and the available actions filtered according to the permission framework responses. The read part is easy, but there should also be some way to handle `idempotent` and `destructive` actions through the permission framework - probably even based on the input of the action.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Undecided

freben:
Yes!

vinzscam:
The credentials are passed to each action so the actions can perform authorization there. Perhaps this is more powerful than performing authorization based on the action name?

drodil:
@vinzscam the issue with that is that the authorization is done only when calling the action while this would filter out actions before that. Especially with the mcp server, I feel it's bad user experience to see an action in a list and only after using it, it fails. WDYT?

PR #31039, opened 2025-09-01
feat(actions): add read permission for actions registry

drodil:

this allows controlling the read access for actions registered in actions registry. the action id is passed as resource reference. tbd later if more permissions are needed to have more fine-grained control which actions are available through permissions.

relates to #31036

### DIFF
--- a/packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/actionsRegistryServiceFactory.test.ts
+++ b/packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/actionsRegistryServiceFactory.test.ts
@@ -24,6 +24,7 @@ import request from 'supertest';
 import { actionsRegistryServiceFactory } from './actionsRegistryServiceFactory';
 import { InputError } from '@backstage/errors';
 import { actionsRegistryServiceRef } from '@backstage/backend-plugin-api/alpha';
+import { AuthorizeResult } from '@backstage/plugin-permission-common';
 
 describe('actionsRegistryServiceFactory', () => {
   const defaultServices = [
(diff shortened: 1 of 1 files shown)


drodil:
Not stale

drodil:
Not stale

packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.ts
```diff
@@ -44,29 +47,55 @@ export class DefaultActionsRegistryService implements ActionsRegistryService {
...
   createRouter(): Router {
     const router = PromiseRouter();
     router.use(json());
 
-    router.get('/.backstage/actions/v1/actions', (_, res) => {
+    router.get('/.backstage/actions/v1/actions', async (req, res) => {
+      const credentials = await this.httpAuth.credentials(req);
+
+      const ids = Array.from(this.actions.keys());
+      const results = await this.permissions.authorize(
```
vinzscam, line 83: I'm trying to brainstorm how we could get this to work 🤔 Perhaps one way could be using an `authorizeConditional()`?
Then we need to figure out who is going to expose the permissions AND the rules (which right now can only be bound to plugins). This is tricky 🤔 

But perhaps as the credentials are passed to each action, the actions themselves can perform authorization?

vinzscam (changes requested):

packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.ts
drodil: Hmm, true. What about making it possible to pass a optional permission to check along the registration of the action? Or a callback for checking permissions that would return either allow or deny 🤔 Have to think about this a bit

drodil: I was thinking, should we add this to the `ActionsRegistryActionOptions`:

```ts
  authorize?: (ctx: {
  }) => Promise<DefinitivePolicyDecision>;
```
And filter the actions based on that, if filtering just by name is not sufficient? Or should both be possible?

drodil: Here's the optional way to do this #31356

PR #31045, opened 2025-09-02
fix(repo-tools): resolve knip-reports failure with spaces in workspace paths

schultzp2020:


packages/repo-tools/src/commands/util.ts
```diff
@@ -24,14 +24,33 @@
...
+/**
+ * Quotes an argument if it contains spaces or other characters that need escaping
+ */
+function quoteArgument(arg: string): string {
+  if (!arg) return arg;
+
+  // If the argument contains spaces, wrap it in double quotes
+  if (arg.includes(' ')) {
+    // Escape any existing double quotes in the argument
+    const escapedArg = arg.replace(/"/g, '\\"');
```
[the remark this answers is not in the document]
freben: I wonder if this should be heeded :)

packages/repo-tools/src/commands/util.ts
```diff
@@ -24,14 +24,33 @@ import pLimit from 'p-limit';
...
 const limiter = pLimit(os.cpus().length);
 
+/**
+ * Quotes an argument if it contains spaces or other characters that need escaping
+ */
+function quoteArgument(arg: string): string {
+  if (!arg) return arg;
+
+  // If the argument contains spaces, wrap it in double quotes
+  if (arg.includes(' ')) {
```
freben: Silly but, do we need to check if it already has quotes around it so we don't double them? I'm honestly not sure

packages/repo-tools/src/commands/knip-reports/knip-extractor.ts
```diff
@@ -106,13 +106,17 @@ async function handlePackage({
 
   let report = await run(
     `${knipDir}/knip.js`,
-    `-W ${packageDir}`, // Run the desired workspace
-    '--config knip.json',
+    '-W', // Run the desired workspace
+    packageDir,
```
Rugvip, line 110: perhaps enough to just quote this otherwise instead?

packages/repo-tools/src/commands/util.ts
```diff
@@ -24,14 +24,33 @@ import pLimit from 'p-limit';
...
+          // Quote arguments that contain spaces
+          const quotedArgs = [path, ...command].map(quoteArgument);
+
           execFile(
             'node',
-            [path, ...command],
+            quotedArgs,
             {
               cwd,
               shell: true,
```
Rugvip: Not sure why we're using shell here, might be for windows? But I think removing it might fix the issue too.

schultzp2020: I'll look into that. I found this issue on a Windows machine.

schultzp2020:
@freben @Rugvip What do you think of this solution using `spawn`? The `execFile` function was having quite a few issues even when I removed the `shell` option.

.changeset/fix-knip-reports-spaces-path.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/repo-tools': patch
+---
+
+Fixed knip-reports command failing when workspace path contains spaces and process termination issues by replacing execFile with spawn and removing shell option.
```
freben:
```suggestion
Fixed knip-reports command failing when workspace path contains spaces and process termination issues by replacing `execFile` with `spawn` and removing `shell` option.
```

packages/repo-tools/src/commands/util.ts
```diff
@@ -29,26 +29,47 @@ export function createBinRunner(cwd: string, path: string) {
...
+          child.stdout?.on('data', data => {
+            stdout += data.toString();
+          });
+
+          child.stderr?.on('data', data => {
+            stderr += data.toString();
+          });
+
+          child.on('error', err => {
+            console.log('err', err);
```
freben: Should these console.log be removed? The `err` seems like debugging code?

packages/repo-tools/src/commands/util.ts
```diff
@@ -29,26 +29,47 @@ export function createBinRunner(cwd: string, path: string) {
...
+              console.log('err', { code, signal, cmd: cmdString });
+              reject(
+                new Error(
+                  `Process was killed with signal ${signal}\n${stderr}`,
+                ),
+              );
+            } else if (code !== 0) {
+              console.log('err', { code, signal, cmd: cmdString });
+              reject(new Error(`Process exited with code ${code}\n${stderr}`));
+            } else if (stderr.trim() && !stdout.trim()) {
```
freben: Why do we check stdout here?

freben (commented):
Seems alright. Just some questions

packages/repo-tools/src/commands/util.ts
schultzp2020: Good point, it's probably not needed.

schultzp2020:
@freben Changes applied, PTAL?

PR #31049, opened 2025-09-03
notifications-node: feature toggle with system metadata PoC

Rugvip:

PoC for #28129

### DIFF
--- a/plugins/notifications-node/src/service/DefaultNotificationService.ts
+++ b/plugins/notifications-node/src/service/DefaultNotificationService.ts
@@ -15,13 +15,53 @@
  */
 
 import { NotificationService } from './NotificationService';
-import { AuthService, DiscoveryService } from '@backstage/backend-plugin-api';
+import {
+  AuthService,
+  BackstageInstance,
+  DiscoveryService,
+} from '@backstage/backend-plugin-api';
+import { SystemMetadataService } from '@backstage/backend-plugin-api';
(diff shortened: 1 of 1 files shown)


freben:
EDIT: Oh sorry just a POC :) 

Original message:

Does this have any benefit over trying a discovery query and like a liveness check? Is it meant to cover for individual flaky nodes spuriously leading to failed detection?

A `Promise.all` like this on a live ecosystem feels iffy either way, just one succeeding should be fine. Or some majority heuristic perhaps?

plugins/notifications-node/src/service/DefaultNotificationService.ts
```diff
@@ -15,13 +15,53 @@
...
+      instances = newInstances;
+    },
+  });
+  return {
+    async hasPlugin(pluginId: string) {
+      const instanceMetas: { items: BackendFeatureMeta[] }[] =
+        await Promise.all(
+          instances.map(async instance => {
+            const instanceMeta = await fetch(
+              `${instance.internalUrl}/.backstage/instanceMetadata/v1/features/installed`,
```
aramissennyeydd, line 40: My thinking here was you would use the current `systemMetadata` API to fetch this - something like
```
const systemMetadata = await fetch(`localhost:7007/.backstage/systemMetadata/v1/features/installed`)
const plugins = (await systemMetadata.json()).items;
return !!plugins[pluginId];
```
it's a little clunky though - if we surfaced this on the service itself would definitely make things easier.

PR #31051, opened 2025-09-03
remove the last remnants of old system structure

### LINKED ISSUE #26353: 🙏 Help wanted: Deprecate old plugin backends

freben:
Closes #26353

This was basically all done already. After this, `yarn backstage-repo-tools lint legacy-backend-exports` reports no errors.


plugins/app-backend/src/index.ts
```diff
@@ -21,4 +21,3 @@
  */
 
 export { appPlugin as default } from './service/appPlugin';
-export * from './service/router';
```
freben, line 24: This was marked as `@internal` but actually exported - double checking with @Rugvip whether this is actually safe to remove?

Rugvip: Yep, prolly just not caught precicely because it was marked as `@internal` and therefore doesn't show up in public types / API reports

PR #31053, opened 2025-09-03
feat: promote instance metadata

aramissennyeydd:

From conversation in #28129, pulling this out to simplify review there.

This promotes the `instanceMetadata` service to a coreService - renaming it to `rootInstanceMetadata`. It also changes the API to be more plugin focused. Modules are still available as children of a given plugin.


packages/backend-plugin-api/src/services/definitions/InstanceMetadataService.ts
```diff
@@ -14,19 +14,14 @@
...
-  | {
-      type: 'plugin';
-      pluginId: string;
-    }
-  | {
-      type: 'module';
-      pluginId: string;
-      moduleId: string;
-    };
+export interface BackendPlugin {
```
freben: Since backend-plugin-api is a very large flat name space, we've sometimes pretty aggressively given types and helpers names that are long and include the service name. See for example `SchedulerServiceTaskFunction` and `readSchedulerServiceTaskScheduleDefinitionFromConfig`.

Maybe this type could be named `InstanceMetadataServicePluginInfo` or something like that. `BackendPlugin` is probably unfortunate to put in this big mash of things.

aramissennyeydd: good call, updated in [0a24d7f](https://github.com/backstage/backstage/pull/31053/commits/0a24d7f49f7b62679f454ce1c585cd7af66f0538)

aramissennyeydd:
Have been sick, plan to get back to this later this week - I also spoke with @Rugvip last week and I think it's best to ship this without the HTTP routes since that opens up some new attack vectors.

aramissennyeydd:
@freben / @Rugvip Finally got build passing, this should be ready for another review 🙏!

.changeset/open-items-open.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-app-api': minor
+---
+
+Updates API for `instanceMetadata` service to return a list of plugins not features. Also adds an HTTP endpoint that returns information about the current instance.
```
aramissennyeydd:
```suggestion
Updates API for `instanceMetadata` service to return a list of plugins not features.
```

packages/backend-plugin-api/src/services/definitions/InstanceMetadataService.ts
```diff
@@ -14,19 +14,15 @@
...
+  modules: {
+    moduleId: string;
+  }[];
+}
 
-/** @alpha */
+/** @public */
 export interface InstanceMetadataService {
-  getInstalledFeatures: () => BackendFeatureMeta[];
+  getInstalledPlugins: () => readonly InstanceMetadataServicePluginInfo[];
```
freben: Just double checking for the maintainers - are we safe with leaving this as a non-promise? Forever certain that we never ever want to do anything async in here? :)

packages/backend-plugin-api/src/services/definitions/coreServices.ts
```diff
@@ -277,4 +277,15 @@ export namespace coreServices {
...
+
+  /**
+   * Information about the current Backstage instance.
+   *
+   * @public
+   */
+  export const instanceMetadata = createServiceRef<
+    import('./InstanceMetadataService').InstanceMetadataService
+  >({
+    id: 'core.instanceMetadata',
```
freben: Another one for the maintainers? Do we want to have this plugin scoped? Not sure what an instance metadata might do that's plugin specific, but maybe this has been discussed already?

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -104,53 +102,51 @@ const instanceRegistry = new (class InstanceRegistry {
...
-    service: instanceMetadataServiceRef,
-    deps: {},
-    factory: async () => ({ getInstalledFeatures: () => installedFeatures }),
+    service: coreServices.instanceMetadata,
+    deps: {
+      logger: coreServices.rootLogger,
+    },
+    factory: async ({ logger }) => {
+      const instanceMetadata = {
+        getInstalledPlugins: () => Object.values(installedPlugins),
```
freben: I'd go as far as constructing the array and then recursively Object.freeze'ing it, but maybe that's paranoid :D

freben (approved):
Alright, looks good to me. Let's see what the others think about the questions before merging.

This does deserve a small documentation page under Framework > Backend System > Core Services on the microsite though. Would appreciate if you got one in before shipping.

packages/backend-app-api/src/wiring/BackendInitializer.ts
aramissennyeydd: not paranoid at all, updated in [797d244](https://github.com/backstage/backstage/pull/31053/commits/797d2444a6ec31f4ef870315d499851c0501f8a9)

packages/backend-plugin-api/src/services/definitions/InstanceMetadataService.ts
aramissennyeydd: updated in [797d244](https://github.com/backstage/backstage/pull/31053/commits/797d2444a6ec31f4ef870315d499851c0501f8a9), no real downside to just making it async for now.

packages/backend-app-api/src/wiring/helpers.ts
```diff
@@ -34,3 +34,29 @@ export function unwrapFeature(
 
   return feature;
 }
+
+/** @internal */
+export type DeepReadonly<T> = {
+  readonly [K in keyof T]: T[K] extends object ? DeepReadonly<T[K]> : T[K];
+};
```
aramissennyeydd, line 41: let me know if i should dedupe this type?

freben:
@aramissennyeydd Sorry I had to rebase this one.

aramissennyeydd:
@freben Is there anybody who should take a look at this as well before we :shipit: ?

packages/backend-plugin-api/report.api.md
```diff
@@ -416,6 +421,24 @@ export interface HttpRouterServiceAuthPolicy {
   path: string;
 }
 
+// @public (undocumented)
+export interface InstanceMetadataService {
+  // (undocumented)
+  getInstalledPlugins: () => Promise<
+    ReadonlyArray<InstanceMetadataServicePluginInfo>
+  >;
```
Rugvip: Hmm, don't think we need this to be a promise? It does keep consuming code simpler and we're pretty guaranteed for this to be a fixed list.

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -104,53 +102,59 @@ const instanceRegistry = new (class InstanceRegistry {
...
-                },
-              );
-            }
-            // Ignore unknown feature types.
-            return undefined;
-          })
-          .filter(Boolean) as BackendFeatureMeta[];
+  const installedPlugins: {
+    [pluginId: string]: InstanceMetadataServicePluginInfo;
+  } = {};
```
Rugvip: Could you make this a `Map`? Just a bit simpler and more secure in general. No need to worry about prototype pollution and such.

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -104,53 +102,59 @@ const instanceRegistry = new (class InstanceRegistry {
...
+        getInstalledPlugins: () => Promise.resolve(readonlyInstalledPlugins),
+      };
+
+      const plugins = await instanceMetadata.getInstalledPlugins();
+
+      logger.info(
+        `Installed plugins on this instance: ${plugins
+          .map(p => p.pluginId)
+          .join(', ')}`,
+      );
```
Rugvip: Hmm, I think this is already logged at startup, think we could skip

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -104,53 +102,59 @@ const instanceRegistry = new (class InstanceRegistry {
...
+              modules: [],
+            };
+          }
+          (
+            installedPlugins[feature.pluginId].modules as Array<{
+              moduleId: string;
+            }>
+          ).push({
+            moduleId: feature.moduleId,
+          });
```
Rugvip: To double check, are modules without a plugin already trimmed away at this point? We wouldn't want to list a plugin here that only has a module installed without the plugin itself.

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -36,14 +36,12 @@ import type {
...
-  instanceMetadataServiceRef,
-  BackendFeatureMeta,
-} from '@backstage/backend-plugin-api/alpha';
 import { DependencyGraph } from '../lib/DependencyGraph';
 import { ServiceRegistry } from './ServiceRegistry';
 import { createInitializationLogger } from './createInitializationLogger';
-import { unwrapFeature } from './helpers';
+import { deepFreeze, unwrapFeature } from './helpers';
+// eslint-disable-next-line @backstage/no-relative-monorepo-imports
+import type { InstanceMetadataServicePluginInfo } from '../../../backend-plugin-api/src/services/definitions/InstanceMetadataService';
```
Rugvip: This thing is public right? No need for the internal import

packages/backend-plugin-api/src/services/definitions/coreServices.ts
Rugvip: Making it plugin scoped would allow us to add plugin-specific utilities like `getOtherInstalledPlugins()` and whatnot. I don't think the value is quite there compared to the ability to be able to use this in other root services like the root HTTP router. It does raise the question of whether we should call this `rootInstanceMetadata` though, which I think we should, and update the ID.

packages/backend-app-api/src/wiring/helpers.ts
Rugvip: nah :p

packages/backend-plugin-api/report.api.md
freben: I suggested this one. Do we REALLY know that we never ever need this to be a promise, is the question :)

packages/backend-app-api/src/wiring/BackendInitializer.ts
aramissennyeydd: How does this work if there's no plugin to be a module of?

aramissennyeydd:
@Rugvip I think I've addressed the main concerns - still a few open questions above that I've left unresolved, let me know what you think.

packages/backend-app-api/src/wiring/BackendInitializer.ts
Rugvip: In that case the module would not be loaded, and I don't think we should list it in the metadata.

Dunno if we mention this anywhere in docs/tests, but we only start up plugins with actual plugins present: https://github.com/backstage/backstage/blob/dd69cf60c4b211b8df9cbec7cdd8f57d15d0aa16/packages/backend-app-api/src/wiring/BackendInitializer.ts#L342

aramissennyeydd: @Rugvip Ah - I'll update this to 2 loops which should circumvent this. AFAIK the current service has this issue today.

aramissennyeydd: updated in [769e2b7](https://github.com/backstage/backstage/pull/31053/commits/769e2b7880d91aae2310d1147425a19164d61ff1)

aramissennyeydd:
@Rugvip let me know if there's anything I can do to help get this over the line 🙇

PR #31055, opened 2025-09-03
feat(devtools): support custom tabs using extensions

secustor:

app-next: 
image

This pull request introduces a new shared web library package, `@backstage/plugin-devtools-react`, to support extension-based routing for the DevTools plugin. It refactors the DevTools plugin to accept contributed routes from other plugins, such as the new "Unprocessed Entities" tab from the Catalog Unprocessed Entities plugin. The changes improve modularity, extensibility, and developer experience by allowing plugins to easily add their own DevTools tabs using a standardized blueprint.

**DevTools Extension System and Routing**

* Added the new package `@backstage/plugin-devtools-react`, which provides extension blueprints (`DevToolsRouteBlueprint`) and data references for contributing routes to the DevTools page. 
* Refactored the DevTools plugin to accept contributed routes via an extension input, and updated the DevTools page to render these routes. 

**Integration of Unprocessed Entities Tab**

* Implemented a DevTools route extension (`unprocessedEntitiesDevToolsRoute`) in the Catalog Unprocessed Entities plugin, using the new blueprint system, and exported it for use in the app.
* Registered the new route in the app, making "Unprocessed Entities" available as a tab in DevTools. 

**Documentation and Release Notes**

* Added documentation, API reports, and release notes for the new package and the extension mechanism. 
These changes make it much easier for other plugins to add their own DevTools tabs, improving extensibility and developer experience in Backstage.


plugins/devtools-common/package.json
```diff
@@ -8,7 +8,8 @@
     "pluginPackages": [
       "@backstage/plugin-devtools",
       "@backstage/plugin-devtools-backend",
-      "@backstage/plugin-devtools-common"
+      "@backstage/plugin-devtools-common",
+      "@backstage/plugin-devtools-react"
```
awanlin, line 12: Why do we need this change? There's no other updates to the common package? Also, common packages are intended to work in both the frontend and backend adding a react package here would break that.

plugins/devtools-backend/package.json
```diff
@@ -7,7 +7,8 @@
     "pluginPackages": [
       "@backstage/plugin-devtools",
       "@backstage/plugin-devtools-backend",
-      "@backstage/plugin-devtools-common"
+      "@backstage/plugin-devtools-common",
+      "@backstage/plugin-devtools-react"
```
awanlin, line 11: Pretty sure this is a mistake, there's no other backend changes and brining in a react package here seems very odd!

plugins/catalog-unprocessed-entities/src/alpha/devToolsRoute.ts
```diff
@@ -0,0 +1,34 @@
...
+  params: {
+    path: 'unprocessed-entities',
+    title: 'Unprocessed Entities',
+    loader: () =>
+      import('../components/UnprocessedEntities').then(
+        ({ UnprocessedEntitiesContent }) =>
+          createElement(UnprocessedEntitiesContent),
+      ),
+  },
+});
```
awanlin, line 37: Out of all the changes this is the one I'm the least sure about. Is this the convention we want to use? Like is is expected that this would potentially automatically get added to DevTools when it gets installed by users? I know a lot of people do that but it's opt in, going this route would change that. Generally we've gone the opt-in path for most things in Backstage. This isn't a hard no but more would like to discuss it and see what the maintained think.

awanlin (changes requested):
Thanks for the contribution @secustor, saw this in my feed so gave it a very quick look, I'll try and give a full review on Friday.

Also wanted to get the conversation going on the one item that jumped out at me that I'd like to hear more thoughts on 👍

plugins/devtools-common/package.json
secustor: These are automatic changes by [`yarn fix`](https://github.com/backstage/backstage/blob/4f83da1f7d2b2924e5c6a491c673291d9af8121a/package.json#L39) especially the `--publish` flag triggers this. 

My understanding is that this is used in internal tooling like `@backstage/cli` or the [Metadata Service](https://backstage.io/docs/backend-system/core-services/plugin-metadata/) 

See for example `catalog-common`

https://github.com/backstage/backstage/blob/4f83da1f7d2b2924e5c6a491c673291d9af8121a/plugins/catalog-common/package.json#L8-L14

plugins/devtools-backend/package.json
secustor: See comment above

plugins/devtools-common/package.json
freben: These are a system of just cross referencing packages. It's a builtin cli thing that lets it bundle up a plugin and all of its constituent parts just based on metadata. See for example 

https://github.com/backstage/backstage/blob/4f83da1f7d2b2924e5c6a491c673291d9af8121a/plugins/catalog-common/package.json#L8

plugins/catalog-unprocessed-entities/src/alpha/devToolsRoute.ts
secustor: Disclaimer my understanding of the new frontend system is not thorough so maybe I misinterpret things here....

Looking at what I have seen from the new frontend system, I understood most should work out of the box. That is if 
```
app: 
  packages: all
```
is used. 

In other cases it would still be opt in like the usage examples in the [Readmes](https://github.com/backstage/backstage/pull/31055/files#diff-0a370ffd2e8388a74e89fb9a947ff9bdf7311d95bd07cab8f24314e441608f70) and the adaption of the [`app-next`](https://github.com/backstage/backstage/pull/31055/files#diff-9f1faf3645c1286627892aa7c28e4a89ce08134719b54cebfd8c492f9c618e7dR118) try to show.

freben: For the new frontend system I think it's fair to say that most things are enabled by default. You install a plugin, and that plugin "jacks itself into" a multitude of places. You expect stuff to just work with no or little config. For example you may get entity cards and an entity tab and a sidebar item and... etc from a single plugin. I bet there are cases where one might feel that it gets a bit of a stretch, but I'm not sure this would be one of them. It's rather a question of - how should we think going forward about the dev tools plugin? SHOULD it be the one stop shop where all admin stuff should go? Would a backend scheduler service task manager thing go here too? If so, maybe it's perfectly fine and the "native" `/catalog/unprocessed-entities` could even be removed (or made into a redirect)? This gets a bit theoretical

plugins/devtools-common/package.json
awanlin: And this is why I should not look at code on my phone. I saw this as being added to the `dependencies` section but that is clear not where it's being added. This is totally fine, sorry about this!

plugins/devtools-backend/package.json
awanlin: I saw this as being added to the `dependencies` section but that is clear not where it's being added. This is totally fine, sorry about this!

plugins/catalog-unprocessed-entities/src/alpha/devToolsRoute.ts
awanlin: Just for context on my thinking: when I logged the issue for this feature I had it in my mind this would be like how you might use a transformer for an entity provider in the New Backend System where the Adopter needs to create a little module to do that.

If this is the way the New Frontend System is intended to work this way then I'm cool with that, thanks for the clarity on this @freben 👍

secustor: > It's rather a question of - how should we think going forward about the dev tools plugin? SHOULD it be the one stop shop where all admin stuff should go?

This is at least how I have interpreted it so far 🤔 

Assuming it shouldn't be the one stop shop. I'm not sure what the difference between the current `devtools` and a central `admin` plugin would be. Outside maybe of a redesign from Content to Cards that is.

That being said my intent with that PR hasn't been to introduce an new standard or pattern, but rather simply reflect the current possibilities which we used on the old frontend system. 

---

If the current `devtools` plugin is chosen, a possible option could be to add a filter which is tied to the permissions a user has. 

Something like this: 
```yaml
app: 
  extensions:
    - name: devtools-content:unprocessed-entities
      config:
        filter: 
          permissions: <a permission id>
```

plugins/catalog-unprocessed-entities/README.md
```diff
@@ -52,16 +52,22 @@ Import `catalogUnprocessedEntitiesPlugin` in your `App.tsx` and add it to your a
...
     catalogUnprocessedEntitiesPlugin,
     // ...
   ],
 });
+
+// Optionally add unprocessed entities route to devtools
+const appFeature = createFrontendModule({
+  pluginId: 'app',
+  extensions: [unprocessedEntitiesDevToolsRoute],
+});
```
awanlin: Should this not come first before the `createApp` call?

plugins/catalog-unprocessed-entities/src/alpha/devToolsRoute.ts
awanlin: For now let's leave permissions as a problem to solve elsewhere as I think that an overall open question related to the New Frontend system. 👍

.changeset/slimy-ghosts-rescue.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/plugin-catalog-unprocessed-entities': patch
+'@backstage/plugin-devtools-react': patch
```
awanlin:
```suggestion
'@backstage/plugin-devtools-react': minor
```

plugins/devtools-react/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@backstage/plugin-devtools-react",
+  "version": "0.1.0",
```
awanlin:
```suggestion
  "version": "0.0.0",
```

Usually like to start like this.

plugins/devtools/README.md
```diff
@@ -160,6 +160,33 @@ You can also add tabs to show content from other plugins that fit well with the
 
 #### Catalog Unprocessed Entities Tab
 
+##### New Frontend System
+
+Create an extension and/or load a 3rd party extension to add additional tabs.
+
+```tsx
+import { DevToolsRouteBlueprint } from '@backstage/plugin-devtools-react';
```
awanlin: We should probably add a line to do the yarn add for `@backstage/plugin-devtools-react` as you likely won't have that already installed.

plugins/devtools-react/README.md
```diff
@@ -0,0 +1,5 @@
+# @backstage/plugin-devtools-react
+
+Welcome to the web library package for the devtools plugin!
```
awanlin:
```suggestion
Welcome to the web library package for the `devtools` plugin!
```

That will make Vale happy.

.changeset/slimy-ghosts-rescue.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/plugin-catalog-unprocessed-entities': patch
+'@backstage/plugin-devtools-react': minor
+'@backstage/plugin-devtools': patch
```
freben, line 4: you need to add the -backend and -common packages too to the changeset, so that their updated package.json files get published

plugins/catalog-unprocessed-entities/src/alpha/devToolsRoute.ts
```diff
@@ -0,0 +1,36 @@
...
+import { createElement } from 'react';
+import { DevToolsRouteBlueprint } from '@backstage/plugin-devtools-react';
+
+/**
+ * DevTools route extension for catalog unprocessed entities
+ * This demonstrates how to create extensions at the app level using the new DevTools extension system
+ *
+ * @alpha
+ */
+export const unprocessedEntitiesDevToolsRoute = DevToolsRouteBlueprint.make({
```
freben:
```suggestion
export const unprocessedEntitiesDevToolsContent = DevToolsContentBlueprint.make({
```

"Content" is the name we give to pages (generally in tabs) added to other containers. See [for example `EntityContentBlueprint`](https://github.com/backstage/backstage/blob/master/plugins/catalog-react/src/alpha/blueprints/EntityContentBlueprint.ts#L39)

plugins/catalog-unprocessed-entities/src/alpha/devToolsRoute.ts
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createElement } from 'react';
+import { DevToolsRouteBlueprint } from '@backstage/plugin-devtools-react';
+
+/**
+ * DevTools route extension for catalog unprocessed entities
+ * This demonstrates how to create extensions at the app level using the new DevTools extension system
```
freben:
```suggestion
 * DevTools content for catalog unprocessed entities.
```

I don't think this will turn out to be particularly temporary, so could be worth having a long lived comment for it.

plugins/catalog-unprocessed-entities/src/alpha/devToolsRoute.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @alpha
+ */
+export const unprocessedEntitiesDevToolsRoute = DevToolsRouteBlueprint.make({
+  params: {
+    path: 'unprocessed-entities',
+    title: 'Unprocessed Entities',
+    loader: () =>
+      import('../components/UnprocessedEntities').then(
+        ({ UnprocessedEntitiesContent }) =>
+          createElement(UnprocessedEntitiesContent),
```
freben:
```suggestion
          compatWrapper(<m.UnprocessedEntitiesContent />),
```

avoid createElement; usages of that will be removed anyway in react 19 migrations soon. Also use the compat wrapper to be sure that the page renders well even if it uses non-NFS features.

plugins/devtools-react/src/devToolsRouteBlueprint.tsx
```diff
@@ -0,0 +1,67 @@
...
+ *   params: {
+ *     path: 'my-feature',
+ *     title: 'My Feature',
+ *     loader: () => import('./MyContent').then(m => ({ default: m.MyContent }))
+ *   }
+ * });
+ * ```
+ * @public
+ */
+export const DevToolsRouteBlueprint = createExtensionBlueprint({
```
freben:
```suggestion
export const DevToolsContentBlueprint = createExtensionBlueprint({
```

plugins/devtools-react/src/devToolsRouteBlueprint.tsx
```diff
@@ -0,0 +1,67 @@
...
+  ExtensionBoundary,
+} from '@backstage/frontend-plugin-api';
+import { devToolsRouteDataRef } from './devToolsRouteDataRef';
+import { JSX } from 'react';
+
+/**
+ * Parameters for creating a DevTools route extension
+ * @public
+ */
+export interface DevToolsRouteBlueprintParams {
```
freben:
```suggestion
export interface DevToolsContentBlueprintParams {
```

plugins/devtools-react/src/devToolsRouteBlueprint.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import {
+  createExtensionBlueprint,
+  ExtensionBoundary,
+} from '@backstage/frontend-plugin-api';
+import { devToolsRouteDataRef } from './devToolsRouteDataRef';
+import { JSX } from 'react';
+
+/**
+ * Parameters for creating a DevTools route extension
```
freben, line 26:
```suggestion
 * Parameters for creating a DevTools content page extension
```

plugins/devtools-react/src/devToolsRouteBlueprint.tsx
```diff
@@ -0,0 +1,67 @@
...
+  title: string;
+  loader: () => Promise<JSX.Element>;
+}
+
+/**
+ * Extension blueprint for creating DevTools routes
+ *
+ * @example
+ * ```tsx
+ * const myDevToolsRoute = DevToolsRouteBlueprint.make({
```
freben:
```suggestion
 * const myDevToolsRoute = DevToolsContentBlueprint.make({
```

plugins/devtools-react/src/devToolsRouteBlueprint.tsx
```diff
@@ -0,0 +1,67 @@
...
+ * @public
+ */
+export interface DevToolsRouteBlueprintParams {
+  path: string;
+  title: string;
+  loader: () => Promise<JSX.Element>;
+}
+
+/**
+ * Extension blueprint for creating DevTools routes
```
freben:
```suggestion
 * Extension blueprint for creating DevTools content pages
```

plugins/devtools-react/src/devToolsRouteBlueprint.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { JSX } from 'react';
+
+/**
+ * Parameters for creating a DevTools route extension
+ * @public
+ */
+export interface DevToolsRouteBlueprintParams {
+  path: string;
+  title: string;
+  loader: () => Promise<JSX.Element>;
```
freben, line 32: You may want to give this a `routeRef` field too, [like the entity content does](https://github.com/backstage/backstage/blob/39a15fdf4dd1cc53da5dfa1563a69048e56d2d8d/plugins/catalog-react/src/alpha/blueprints/EntityContentBlueprint.ts#L84). Note how it emits the given route if any. That way, it will be possible to route to the content tab with links

plugins/devtools-react/src/devToolsRouteBlueprint.tsx
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Extension blueprint for creating DevTools routes
+ *
+ * @example
+ * ```tsx
+ * const myDevToolsRoute = DevToolsRouteBlueprint.make({
+ *   params: {
+ *     path: 'my-feature',
+ *     title: 'My Feature',
+ *     loader: () => import('./MyContent').then(m => ({ default: m.MyContent }))
```
freben: this should be returning JSX though

plugins/devtools-react/src/devToolsRouteBlueprint.tsx
```diff
@@ -0,0 +1,67 @@
...
+ *   }
+ * });
+ * ```
+ * @public
+ */
+export const DevToolsRouteBlueprint = createExtensionBlueprint({
+  kind: 'devtools-route',
+  attachTo: { id: 'page:devtools', input: 'routes' },
```
freben:
```suggestion
  attachTo: { id: 'page:devtools', input: 'contents' },
```

plugins/devtools/src/alpha/plugin.tsx
```diff
@@ -46,14 +48,27 @@ export const devToolsApi = ApiBlueprint.make({
...
-  params: {
-    path: '/devtools',
-    routeRef: convertLegacyRouteRef(rootRouteRef),
-    loader: () =>
-      import('../components/DevToolsPage').then(m =>
-        compatWrapper(<m.DevToolsPage />),
-      ),
+export const devToolsPage = PageBlueprint.makeWithOverrides({
+  inputs: {
+    routes: createExtensionInput([devToolsRouteDataRef], {
```
freben:
```suggestion
    contents: createExtensionInput([
      coreExtensionData.reactElement,
      coreExtensionData.routePath,
      coreExtensionData.routeRef.optional(),
      // ....
    ], {
```

see [what the entity page does](https://github.com/backstage/backstage/blob/39a15fdf4dd1cc53da5dfa1563a69048e56d2d8d/plugins/catalog/src/alpha/pages.tsx#L79).

The reason that you need to leave these as dedicated outputs, is that this is how the routing system is able to detect that they exist and should be mapped to the routing table.

plugins/devtools/src/components/DefaultDevToolsPage/DefaultDevToolsPage.tsx
```diff
@@ -37,5 +38,14 @@ export const DefaultDevToolsPage = () => (
         <ConfigContent />
```
freben, line 40: Note that these two pre-existing ones should be made into extensions themselves as well. We can do it as a standalone effort perhaps, but note that for example the [entity overview page itself](https://github.com/backstage/backstage/blob/39a15fdf4dd1cc53da5dfa1563a69048e56d2d8d/plugins/catalog/src/alpha/entityContents.tsx#L33) is a content.

Making them extensions makes it possible for users to rearrange, enable/disable, group and reconfigure them just like the other ones.

freben (commented):
Hey, thanks!

plugins/devtools-react/src/devToolsRouteDataRef.ts
```diff
@@ -0,0 +1,37 @@
...
+import { JSX } from 'react';
+
+/**
+ * Represents a DevTools route that can be contributed by extensions
+ * @public
+ */
+export interface DevToolsRouteData {
+  path: string;
+  title: string;
+  children: JSX.Element;
```
Rugvip: Hmm, I think we should document this more clearly actually, but looking at this it'd be much better to rely on existing `coreExtensionData`. That'll allow things like extension middleware to wrap the React element, but more importantly the `coreExtensionData.routePath` is required to make it possible to route to sub pages.

Essentially, if a piece of data already exists in `coreExtensionData`, **never** make it part of custom extension data unless it's in a substantially altered form.

plugins/devtools/src/components/DefaultDevToolsPage/DefaultDevToolsPage.tsx
secustor: Do you want to merge this before or after this PR?

plugins/devtools/src/alpha/plugin.tsx
secustor: I seeeee, now it also make sense why my plugin is not auto discovered

secustor:
I have tried to implement the requested changes, but now it seems that it can not resolve the route anymore. :/ 

image

Any ideas what I'm missing here?  The error seems to originating from that path is not propagated

.changeset/slimy-ghosts-rescue.md
freben, line 1: Ping for needing to add more packages here

plugins/devtools-react/src/devToolsContentBlueprint.tsx
```diff
@@ -0,0 +1,86 @@
...
+  loader: () => Promise<JSX.Element>;
+  routeRef?: RouteRef;
+}
+
+/**
+ * Extension blueprint for creating DevTools routes
+ *
+ * @example
+ * ```tsx
+ * const myDevToolsRoute = DevToolsRouteBlueprint.make({
```
freben:
```suggestion
 * const myDevToolsContent = DevToolsContentBlueprint.make({
```

plugins/devtools-react/src/devToolsContentBlueprint.tsx
```diff
@@ -0,0 +1,86 @@
...
+ */
+export interface DevToolsRouteBlueprintParams {
+  path: string;
+  title: string;
+  loader: () => Promise<JSX.Element>;
+  routeRef?: RouteRef;
+}
+
+/**
+ * Extension blueprint for creating DevTools routes
```
freben:
```suggestion
 * Extension blueprint for creating DevTools content pages (appearing as tabs)
```

plugins/devtools-react/src/devToolsContentBlueprint.tsx
```diff
@@ -0,0 +1,86 @@
...
+  kind: 'devtools-content',
+  attachTo: { id: 'page:devtools', input: 'contents' },
+  output: [
+    coreExtensionData.reactElement,
+    coreExtensionData.routePath,
+    coreExtensionData.routeRef.optional(),
+    contentTitleDataRef,
+  ],
+  dataRefs: {
+    title: contentTitleDataRef,
```
freben: There's a `coreExtensionData.title`; you may not have to create this custom ref at all

plugins/devtools-react/src/devToolsContentBlueprint.tsx
```diff
@@ -0,0 +1,86 @@
...
+  RouteRef,
+} from '@backstage/frontend-plugin-api';
+import { JSX } from 'react';
+import { contentTitleDataRef } from './extensionData';
+
+/**
+ * Parameters for creating a DevTools route extension
+ * @public
+ */
+export interface DevToolsRouteBlueprintParams {
```
freben:
```suggestion
export interface DevToolsContentBlueprintParams {
```

plugins/devtools-react/src/devToolsContentBlueprint.tsx
```diff
@@ -0,0 +1,86 @@
...
+  dataRefs: {
+    title: contentTitleDataRef,
+  },
+  config: {
+    schema: {
+      path: z => z.string().optional(),
+      title: z => z.string().optional(),
+    },
+  },
+  *factory(params: DevToolsRouteBlueprintParams, { node, config }) {
```
freben:
```suggestion
  *factory(params: DevToolsContentBlueprintParams, { node, config }) {
```

plugins/devtools/README.md
```diff
@@ -160,6 +160,37 @@ You can also add tabs to show content from other plugins that fit well with the
...
+##### New Frontend System
+
+Create an extension and/or load a 3rd party extension to add additional tabs.
+
+```shell
+yarn --cwd plugins/<your-plugin> add @backstage/plugin-devtools-react
+```
+
+```tsx
+import { DevToolsRouteBlueprint } from '@backstage/plugin-devtools-react';
```
freben:
```suggestion
import { DevToolsContentBlueprint } from '@backstage/plugin-devtools-react';
```

plugins/devtools/README.md
```diff
@@ -160,6 +160,37 @@ You can also add tabs to show content from other plugins that fit well with the
...
+
+export const unprocessedEntitiesDevToolsRoute = DevToolsRouteBlueprint.make({
```
freben:
```suggestion
export const unprocessedEntitiesDevToolsContent = DevToolsContentBlueprint.make({
```

plugins/devtools/README.md
```diff
@@ -160,6 +160,37 @@ You can also add tabs to show content from other plugins that fit well with the
...
+      import('../components/UnprocessedEntities').then(
+        ({ UnprocessedEntitiesContent }) =>
+          createElement(UnprocessedEntitiesContent),
+      ),
+  },
+});
+
+const appFeature = createFrontendModule({
+  pluginId: 'app',
+  extensions: [unprocessedEntitiesDevToolsRoute],
```
freben:
```suggestion
  extensions: [unprocessedEntitiesDevToolsContent],
```

plugins/devtools-react/src/devToolsContentBlueprint.tsx
secustor: There hasn't been when I have created the PR 😅

Will change it

plugins/devtools/src/components/DefaultDevToolsPage/DefaultDevToolsPage.tsx
```diff
@@ -37,5 +38,14 @@ export const DefaultDevToolsPage = () => (
...
       </RequirePermission>
     </DevToolsLayout.Route>
+    {extensions?.map((extension, index) => (
+      <DevToolsLayout.Route
+        key={`extension-${extension.path}-${index}`}
```
freben: just the index is probably fine here

plugins/devtools/src/components/DevToolsPage/DevToolsPage.tsx
```diff
@@ -17,8 +17,15 @@
 import { useOutlet } from 'react-router-dom';
 import { DefaultDevToolsPage } from '../DefaultDevToolsPage';
 
-export const DevToolsPage = () => {
+/**
+  @public
+ */
+export interface DevToolsPageProps {
+  extensions?: any[];
```
freben: We very rarely, if ever, go with `any` in public interfaces.

Also at the react level like here, you probably don't use the word "extension" at all. Here you've left the NFS world and have more concrete types.

See for example what the `EntityContentLayoutBlueprint` does; it expects a component that accepts a number of `cards`, that have an `element` for the rendering and then any additional props relevant to the parent-child relationship next to it.

So maybe you would say `contents: DevToolsPageContent[]` where that type has whatever is necessary to render tabs. Like title, path, element.

plugins/devtools/src/components/DefaultDevToolsPage/DefaultDevToolsPage.tsx
freben: I think my main question is, did you check whether the routing worked? I was curious what a mixture of NFS routing and this explicit DevToolsLayout does

.changeset/slimy-ghosts-rescue.md
```diff
@@ -0,0 +1,9 @@
...
+'@backstage/plugin-devtools-backend': patch
+'@backstage/plugin-devtools-common': patch
+---
+
+Add support for adding `unprocessed-entities` and other tabs to `devtools`
```
awanlin:
```suggestion
Add support for adding `unprocessed-entities` and other tabs to `devtools` when using the New Frontend system
```

awanlin (changes requested):
Thanks again for this contributions, sorry this is taking some time to get approved and merged. I left a small comment about the changeset and there looks to be some open comments from @freben still. I'd very much like to have this in place for the December release so will be checking in on this regularly. 🙏

plugins/devtools/src/components/DefaultDevToolsPage/DefaultDevToolsPage.tsx
secustor: Yes, I test locally using `app-next`.

plugins/devtools-react/src/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Web library for the devtools plugin.
+ *
+ * @packageDocumentation
+ */
+export {
+  type DevToolsContentBlueprintParams,
+  DevToolsContentBlueprint,
+} from './devToolsContentBlueprint.tsx';
```
freben:
```suggestion
} from './devToolsContentBlueprint';
```

plugins/catalog-unprocessed-entities/README.md
```diff
@@ -52,11 +52,17 @@ Import `catalogUnprocessedEntitiesPlugin` in your `App.tsx` and add it to your a
 
 ```typescript
 import catalogUnprocessedEntitiesPlugin from '@backstage/plugin-catalog-unprocessed-entities';
+import { unprocessedEntitiesDevToolsContent } from '@backstage/plugin-catalog-unprocessed-entities/alpha';
 
-// ...
+// Optionally add unprocessed entities route to devtools
+const appFeature = createFrontendModule({
+  pluginId: 'app',
```
freben: So here's a weird thing.

Usually extensions aren't exported standalone. They are more commonly just put into the relevant plugin in the first place, and then they appear in the plugin's extensions list.

The extension in question here, was created without a name. It just has a kind. Its full extension ID will be derived, then, from the plugin that it associates with. If you had put the extension on the `catalog-unprocessed-entities` plugin, its full ID would end up being `devtools-content:catalog-unprocessed-entities` which is the desired outcome. But here you are suggesting putting it out as an `app` module, and I believe that that will make its ID be `devtools-content:app`.

I suspect that the reason you did it this way, was to make it an optional install. But if so, I think we should try to find a better way of achieving that.

Option 1 could be to actually put this extension [in the extensions list](https://github.com/backstage/backstage/blob/59e27508f1a664bc69edec301fa8f4c5dd201d83/plugins/catalog-unprocessed-entities/src/alpha/plugin.tsx#L75) of the unprocessed-entities plugin, but marking it as `disabled: true` by default.

Option 2 could be to update these installation instructions to instead suggest

```ts
import devToolsOriginal from '@backstage/plugin-catalog-unprocessed-entities/alpha';

const devTools = devToolsOriginal.withOverrides({
  extensions: [
    unprocessedEntitiesDevToolsContent,
  ],
});

// add this overridden version to the app
```

as an alternative to `createFrontendModule`, but that feels a bit cumbersome too.

Option 3 would be to simply change the plugin ID on this row to be `catalog-unprocessed-entities` instead of `app`. But it loses all type safety and is a bit wonky in the first place (you shouldn't normally have to install extensions one by one; they should be exposed by something else).

secustor: IIRC `catalog-unprocessed-entities` nor `dev-tools` has worked an the beginning and I concluded that it is probably a scoping issue as both sides need to be connected. As it started to work when changed to `app`, I thought I'm on the correct path. 

I have tried out option 1 and it seems to be the best balanced and most idiomatic option as the user can overwrite via config file rather override in typescript. 
So I moved forward and implemented that, tell me what you think!

PR #31058, opened 2025-09-04
Fixed duplicate route warning for TechDocs Addons documentation in Do…

AdityaK60:
image

This PR resolves a Docusaurus routing warning caused by two markdown files _(addons.md and addons--new.md)_ using the same `id: addons` in their frontmatter. This resulted in both files attempting to generate the same route (/docs/addons), leading to non-deterministic routing behavior as shown in the waring.

Changes

- Updated the id in addon.md to addons--old
- Added route in addons--new.md to navigate to addons.md

### DIFF
--- a/docs/features/techdocs/addons.md
+++ b/docs/features/techdocs/addons.md
@@ -1,5 +1,5 @@
 ---
-id: addons--old
+id: addons
 title: TechDocs Addons
 description: How to find, use, or create TechDocs Addons.
 ---
--- a/docs/features/techdocs/addons.md
+++ b/docs/features/techdocs/addons.md
@@ -1,5 +1,5 @@
 ---
-id: addons
+id: addons--old
 title: TechDocs Addons
 description: How to find, use, or create TechDocs Addons.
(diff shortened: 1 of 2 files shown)


AdityaK60:
@Rugvip tagging you for review as no reviewer's assigned

AdityaK60:
Commenting to avoid stale. Please review the changes.

AdityaK60:
Please review 🙏

docs/features/techdocs/addons.md
```diff
@@ -1,5 +1,5 @@
 ---
-id: addons
+id: addons--old
```
freben: Sorry, I think I've changed my mind. This one should be called just `addons`, and the one above should be called `addons--new`. That's both more consistent, and more in line with what create-app actually promotes at this point.

AdityaK60: made the change. Updated the id as suggested

AdityaK60:
@freben Please review

PR #31069, opened 2025-09-05
The frontend package, when executed within a Continuous Integration (CI) environment, will not have its file watcher functionality enabled.

acierto:

I had the use-case where I have to run frontend package on the CI and it is not possible right now due too many files are watched. Logically I think there is no need to watch them even if it is a dev mode, that would make sense to do it only on developer laptop.

### DIFF
--- a/.changeset/sixty-hornets-cheat.md
+++ b/.changeset/sixty-hornets-cheat.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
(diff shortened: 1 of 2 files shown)


freben:
Adding @Rugvip to this one.

I like the idea, question is mostly, is it OK to use the CI flag for it or should there be some other mechanism for disablement. I'll let him decide.

acierto:
Hey @Rugvip 👋 , does it make sense for you?

PR #31074, opened 2025-09-05
Disable Ajv's allErrors option

MegaManSec:

The problem is that Ajv is configured with allErrors: true, which forces the validator to keep checking every possible error instead of stopping at the first one. This makes validation much slower and more resource-intensive, allowing malicious inputs to trigger excessive CPU and memory usage, potentially causing denial of service. See https://hackerone.com/reports/903521 for more information.

### DIFF
--- a/.changeset/heavy-breads-kneel.md
+++ b/.changeset/heavy-breads-kneel.md
@@ -1,0 +1,7 @@
+---
+'@backstage/backend-openapi-utils': minor
+'@backstage/catalog-model': minor
+'@backstage/config-loader': minor
+---
+
+Disable Ajv's allErrors option in production.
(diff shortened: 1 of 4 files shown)


freben:
As an end user, it also means that you will fix one small mistake in your file, resubmit and wait, find another problem, resubmit and wait...

This setting was intentional I believe. There's a level of trust placed in users that are employed and have access to registering entities.

[The threat model](https://backstage.io/docs/overview/threat-model/) has some more details.

MegaManSec:
Right. Thank you for the link.

freben:
Feel free to close if you agree.

MegaManSec:
If there should be no unauthenticated endpoint, such that the threat model does not assume malicious (annoying) users, then I'd agree to close.

PR #31082, opened 2025-09-08
frontend-app-api: new error handling for app startup

Rugvip:

This changes the way errors are reported in apps in the new frontend system. Rather than throwing on the first encountered error and then having that be logged to the console, errors will instead be collected an returned under an `errors` property from `createSpecializedApp`.

The default `createApp` has been updated to handle these errors by showing the following error page:

image

For lower level usage the function to create the error page is exported as `maybeCreateErrorPage` from `@backstage/frontend-defaults`.

This is a starting point for this functionality that aims to cover the most critical app startup errors, especially those related to extension tree resolution and instantiation. There are still some cases that aren't moved over to this system yet, but figured it's better to aim to ship this first and then increase coverage in followups.


packages/frontend-defaults/src/maybeCreateErrorPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FrontendPluginInfo } from '@backstage/frontend-plugin-api';
+import { JSX, useEffect, useState } from 'react';
+// eslint-disable-next-line @backstage/no-relative-monorepo-imports
+// eslint-disable-next-line @backstage/no-relative-monorepo-imports
```
freben:
```suggestion
```

packages/frontend-app-api/src/wiring/createErrorCollector.ts
```diff
@@ -0,0 +1,140 @@
...
+    context: { node: AppNode };
+  };
+};
+
+/**
+ * @public
+ */
+export type AppError =
+  keyof AppErrorTypes extends infer ICode extends keyof AppErrorTypes
+    ? ICode extends any
```
freben, line 86: image

packages/frontend-app-api/src/wiring/createErrorCollector.ts
```diff
@@ -0,0 +1,140 @@
...
+    : never;
+
+/** @internal */
+export interface ErrorCollector<TContext extends {} = {}> {
+  report<TCode extends keyof AppErrorTypes>(
+    report: Omit<
+      AppErrorTypes[TCode]['context'],
+      keyof TContext
+    > extends infer IContext extends {}
+      ? [{}] extends [IContext]
```
freben: image

packages/frontend-defaults/report.api.md
```diff
@@ -45,6 +47,16 @@ export function discoverAvailableFeatures(config: Config): {
   features: (FrontendFeature | FrontendFeatureLoader)[];
 };
 
+// @public
+export function maybeCreateErrorPage(
```
freben, line 51: I guess this one won't ever be pluggable.

packages/frontend-app-api/report.api.md
```diff
@@ -4,16 +4,113 @@
...
+import { FrontendPlugin } from '@backstage/frontend-plugin-api';
 import { FrontendPluginInfo } from '@backstage/frontend-plugin-api';
 import { JsonObject } from '@backstage/types';
 import { RouteRef } from '@backstage/frontend-plugin-api';
 import { SubRouteRef } from '@backstage/frontend-plugin-api';
 
+// @public (undocumented)
+export type AppError =
+  keyof AppErrorTypes extends infer ICode extends keyof AppErrorTypes
+    ? ICode extends any
```
freben, line 22: With all this fancy typing, is there any risk that someone upgrades app-api but has duplicate backend-defaults versions and then stuff won't compile with some "never" errors because the set of possible app errors changed? Maybe it's just not an issue

freben (approved):
Alright, some thoughts but happy to get in

packages/frontend-app-api/src/wiring/createErrorCollector.ts
Rugvip: Kinda blame TypeScript for hiding useful type system features only behind sepecific syntax constructs :p

Without the `extends any` we'd end up with this:

```ts
{
  code: CODE_A | CODE_B | ...
  ...
}
```

instead of what we get now:

```ts
| {
  code: CODE_A
   ...
}
| {
  code: CODE_B
   ...
}
| ...
```

packages/frontend-app-api/src/wiring/createErrorCollector.ts
Rugvip: This one's actually not needed anymore, a remnant from when this was checking against `never` :p

packages/frontend-defaults/report.api.md
Rugvip: Nah, just aiming to export all the default, but if one wants a custom error page it's fairly trivial to construct.

packages/frontend-app-api/report.api.md
Rugvip: Yep but I think that's fine tbh. If someone's using `createSpecializedApp` it'd uncover a version incomapatiblity. It won't error on the internal usage in `createApp` though, since that's not part of the published type declarations.

Maybe there's some case I haven't thought about, but perhaps we can add in an `AnyAppError` or such if we discover it's needed in the future.

PR #31084, opened 2025-09-08
Tool-tip text correction for the Theme selection in settings page

gyan9427:
Before it was showing an extra "theme" string:
image

After the change:
image


freben (commented):
Hey! Thanks. You'll want to run `yarn build:api-reports` and put those results in the PR too, to make the build happy.

PR #31085, opened 2025-09-08
Wait for storage before loading the custom homepage layout

### LINKED ISSUE #31072: Customizable home page race condition when loading custom page
### 📜 Issue Labels

### 🔎 Search Terms

```plain
customizable home page race condition default layout config flicker
```

### 🗃️ Project Area

Home

### 🔗 External Integration

N/A

### 📝 Description & Context

The CustomHomepageGrid component has a race condition during initial load that causes a brief flicker of the default layout before displaying the user's customized homepage layout. This happens because the component immediately renders default widgets while storage is being loaded asynchronously, then switches to the custom layout once storage resolves. The duration which the default layout is shown depends on what you are using as storage. If it is local storage you might not even see it but if it is a remote database it might be slow enough to manifest itself as a flicker.

In the useHomeStorage hook, when the component first renders, homeSnapshot is either undefined or has presence: 'unknown'. The useMemo logic doesn't explicitly handle these states and falls through to return defaultWidgets, then later updates when storage resolves.

### 👍 Expected Behavior

The CustomHomepageGrid should either:
1. Show a loading state while storage is being checked (homeSnapshot is undefined or presence === 'unknown')
2. Wait until storage presence is definitively known ('absent' or 'present') before rendering any widgets
3. Provide a prop to control this behavior (e.g., waitForStorage={true})

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

1. Set up a Backstage app with CustomHomepageGrid and some default configuration
2. Customize the homepage layout using the grid's edit functionality to create a different layout than the default
3. Refresh the page
5. Observe: Brief flash of the default layout before the customized layout appears

Speissi:

Added a loading check that prevents rendering the custom layout until storage is fully loaded.

The `useHomeStorage` hook uses `useObservable` to watch storage changes asynchronously. Without this check, the component would render with default widgets before the actual stored layout loads, causing a potential visual "glitch" if the storage is not fast enough. While not very serious functionally could cause bad user experience.

Closes #31072 

  - This file does not seem to have any tests, if wanted I'm more than happy to create some (in addition to just testing this particular case)


drodil (approved):
LGTM!

plugins/home/src/components/CustomHomepage/CustomHomepageGrid.tsx
```diff
@@ -322,6 +325,11 @@ export const CustomHomepageGrid = (props: CustomHomepageGridProps) => {
     );
   };
 
+  // Don't render anything while storage is loading
+  if (isStorageLoading) {
+    return null;
```
freben: Or `<Progress />` perhaps?

Speissi: Excellent suggestion. Replaced `null` with `<Progress />`. I hope no one has such a slow storage, that they get to enjoy the bar for very long. 😁

PR #31097, opened 2025-09-09
fix(module-federation): disallow imported fallback modules in mf

callebstrom:
### Changes
- Ensure that a plugin in a module federation container cannot pollute the host app with a direct dependency on shared dependencies. 

### DIFF
--- a/packages/cli/src/modules/build/lib/bundler/config.ts
+++ b/packages/cli/src/modules/build/lib/bundler/config.ts
@@ -264,24 +264,30 @@ export async function createConfig(
             singleton: true,
             requiredVersion: '*',
             eager: !isRemote,
+            import: false,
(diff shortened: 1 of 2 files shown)


packages/cli/src/modules/build/lib/bundler/config.ts
```diff
@@ -264,22 +264,26 @@ export async function createConfig(
...
+            import: false,
           },
           'react-router-dom': {
             singleton: true,
             requiredVersion: '*',
             eager: !isRemote,
+            import: false,
           },
           // MUI v4
           '@material-ui/core/styles': {
```
Rugvip, line 291: How come we don't add it for all of theme? These are pretty critical too

PR #31102, opened 2025-09-10
Updated Note format

AdityaK60:
Updated Note format for https://backstage.io/docs/features/software-catalog/api/catalog .


AdityaK60:
Yes it seems to render correctly.

Before:
image

After:
image

PR #31103, opened 2025-09-10
Catalog: add optional spec.type to allow rules

vinzscam:

This PR adds support for specifying an optional `spec.type` to allow's catalog rules, to allow the ingestion of entities of a specified kinds and types.

For example:

```diff
  catalog:
    rules:
      - allow:
          - Component
          - API
          ...
+     - allow:
+         - kind: API
+           spec.type: openapi
        locations:
          - type: url
            pattern: https://github.com/org/*\/blob/master/*.yaml
```

This will prevent entities of kind `API` with a type other than `openapi` from being ingested into the catalog.


packages/app/src/components/Root/Root.tsx
```diff
@@ -79,7 +79,7 @@ const SidebarLogo = () => {
 
 export const Root = ({ children }: PropsWithChildren<{}>) => (
   <SidebarPage>
-    <Sidebar>
+    <Sidebar sidebarOptions={{ drawerWidthOpen: 256 }}>
```
freben: Unrelated? :)

.changeset/rotten-hairs-attack.md
```diff
@@ -0,0 +1,25 @@
+---
+'@backstage/plugin-catalog-backend': minor
+---
+
+Add support for specifying an entity `spec.type` in `catalog.rules.allow` rules within the catalog configuration.
```
freben:
```suggestion
Add support for specifying an entity `spec.type` in `catalog.rules` and `catalog.locations.rules` within the catalog configuration.
```

Right? Isn't it important in both

plugins/catalog-backend/config.d.ts
```diff
@@ -38,8 +38,11 @@ export interface Config {
        * Allow entities of these particular kinds.
        *
        * E.g. ["Component", "API", "Template", "Location"]
+       *
+       * You can also specify the type of the entity by using an object with `kind` and optional `type` properties.
+       * E.g. [{ kind: "Component", type: "service" }]
```
freben:
```suggestion
       * You can also specify the type of the entity by using an object with `kind` and optional `spec.type` properties.
       * E.g. [{ kind: "Component", 'spec.type': "service" }]
```

I just realized a thing. Didn't we just add validation for config keys? Does this really work or could the config system throw an exception if you try to put dots in key names?

plugins/catalog-backend/src/ingestion/CatalogRules.ts
```diff
@@ -26,16 +26,19 @@ import { minimatch } from 'minimatch';
  * An undefined list of matchers means match all, an empty list of matchers means match none.
  */
 export type CatalogRule = {
-  allow: Array<{
-    kind: string;
-  }>;
+  allow: CatalogRuleAllow[];
```
freben, line 30: The cool thing is, now these rules are actually of type [EntityPredicate](https://backstage.io/docs/next/features/software-catalog/catalog-customization/#entity-predicate-queries)? As in, we can expand support for any set of filters in the future if we want. And that one has already been implemented in [`entityPredicateToFilterFunction`](https://github.com/backstage/backstage/blob/3e574055b1ce15f8ba5697a6ba69a5647e212242/plugins/catalog-react/src/alpha/predicates/entityPredicateToFilterFunction.ts#L25). Buuuut that one is in the -react package for some reason. I wish it were in the client package instead 🤔 What do you think about actually copying that implementation temporarily and then we can move it around later? Or, honestly, you could just make a file system import across module boundaries temporarily, that's fine too.

plugins/catalog-backend/src/ingestion/CatalogRules.ts
```diff
@@ -78,6 +81,9 @@ export class DefaultCatalogRulesEnforcer implements CatalogRulesEnforcer {
    * catalog:
    *   rules:
    *   - allow: [Component, API]
+   *   - allow:
+   *     - kind: Resource
+   *       type: database
```
freben:
```suggestion
   *       'spec.type': database
```

plugins/catalog-backend/src/ingestion/CatalogRules.ts
```diff
@@ -105,7 +111,14 @@ export class DefaultCatalogRulesEnforcer implements CatalogRulesEnforcer {
...
         .getConfigArray('catalog.rules')
         .map(ruleConf => ({
-          allow: ruleConf.getStringArray('allow').map(kind => ({ kind })),
+          allow: ruleConf
+            .get<Array<string | CatalogRuleAllow>>('allow')
+            .map(kindOrObject => {
+              if (typeof kindOrObject === 'string') {
+                return { kind: kindOrObject };
+              }
+              return kindOrObject;
```
freben: Maybe worth making a little zod type, or a full `readEntityPredicateFromConfig` function just like we already have `readDurationFromConfig` and `readSchedulerServiceTaskScheduleDefinitionFromConfig`

packages/app/src/components/Root/Root.tsx
vinzscam: 🙈

.changeset/rotten-hairs-attack.md
vinzscam: right now it is only implemented in `catalog.rules`

plugins/catalog-backend/config.d.ts
vinzscam: yep, it works

plugins/catalog-backend/src/ingestion/CatalogRules.ts
vinzscam: I think for now this is enough. But it would be good to plan implementing the full entity filters after mainline release

plugins/catalog-backend/src/ingestion/CatalogRules.ts
```diff
@@ -44,6 +48,16 @@ export type CatalogRulesEnforcer = {
...
 };
 
+const allowRuleParser = z.array(
+  z
+    .object({
+      kind: z.string(),
+      'spec.type': z.string().optional(),
+    })
+    .or(z.string())
+    .transform(val => (typeof val === 'string' ? { kind: val } : val)),
```
freben, line 58: 👍

plugins/catalog-backend/src/ingestion/CatalogRules.ts
```diff
@@ -199,13 +216,17 @@ export class DefaultCatalogRulesEnforcer implements CatalogRulesEnforcer {
...
       return true;
     }
 
     for (const matcher of matchers) {
-      if (entity?.kind?.toLowerCase() !== matcher.kind.toLowerCase()) {
+      if (entity.kind?.toLowerCase() !== matcher.kind.toLowerCase()) {
+        continue;
+      }
+
+      if (matcher['spec.type'] && matcher['spec.type'] !== entity.spec?.type) {
```
freben: Maybe case insensitive here too? Since that matches what a database query would do (and what the kind matcher already does)

PR #31106, opened 2025-09-10
Batch consecutive get calls to user-settings

grantila:

This PR adds a multi-get endpoint to user settings. The frontend plugin will now use DataLoader to batch several `get()` requests (within 10ms) into one API call and cache them for a short period (2 seconds).

This improves performance for frontend plugins/pages which have several user settings needed to be fetched, especially when the backend is not very close to the client (in terms of latency) and exposed as HTTP/1.1 where each call consumes a socket, which is usually limited to 6 per origin in browsers. It also reduces the amount of consecutive `get()` calls that can happen when several components read the same value when they render.


plugins/user-settings-backend/src/service/router.ts
```diff
@@ -14,12 +14,17 @@
  * limitations under the License.
  */
 
-import { InputError } from '@backstage/errors';
+import { InputError, serializeError } from '@backstage/errors';
+import { map } from 'already';
```
vinzscam: I would say to use `p-limit` for this as it's already used in the repo

plugins/user-settings/src/apis/StorageApi/UserSettingsStorage.test.ts
```diff
@@ -25,6 +25,9 @@ import {
   mockApis,
   registerMswTestHooks,
 } from '@backstage/test-utils';
+import { NotFoundError } from '@backstage/errors';
+import { parseDataLoaderKey } from '@backstage/plugin-user-settings-common';
+import { defer } from 'already';
```
vinzscam:
```suggestion
import { createDeferred } from '@backstage/types';
```

grantila:
After feedback; Now using `p-limit` for concurrency limiting, and `createDeferred`. Also added unit tests for the new backend endpoint `/multi`.

grantila:
> I'm curious how many settings are we talking about that need to be fetched?

Let's say a half dozen, which in-and-of-themselves consume the max number of sockets a browser keeps concurrent over http/1.1. If you're far away from the server (physically / in terms of latency), you'll have to wait for them, before your can fetch other data (or the other way around), causing unnecessary lag.

The primary use-case for this that I have in mind is the [Manage plugin](https://github.com/backstage/community-plugins/tree/main/workspaces/manage), which has a few settings itself, and allow other modules to integrate with it, potentially adding up to even more settings.

plugins/user-settings-backend/src/service/router.ts
```diff
@@ -40,6 +45,61 @@ export async function createRouter(options: {
...
+          return parseDataLoaderKey(item);
+        }),
+      );
+    } else {
+      throw new InputError('Expected query param "items" to be an array');
+    }
+
+    const limit = pLimit(10);
+
+    const userSettings = await Promise.all(
```
freben: There should be a `userSettingsStore.multiGet` instead, that accepts an array of `keys` and returns `Array<JsonValue | null>` where null means it was not present, in the same order as the keys were given. This is how multiget works in the catalog too. It would probably be implemented in terms of `.andWhereIn('key`, batch)` or so for some reasonable batch size (100?) before it hits database query builder limits, and some code after that which ensures that they are returned in the same order as the input. Note that the caller may be silly and ask for the keys `'a', 'b', 'a'`, and that should work too.

Then there's no per-key error state returned by the service, but instead it only throws on truly "hard" errors. Which includes the inability to JSON parse a value, which arguably is a catastrophic service error level thing to happen.

plugins/user-settings-backend/src/service/router.ts
```diff
@@ -40,6 +45,61 @@ export async function createRouter(options: {
     return credentials.principal.userEntityRef;
   };
 
+  // get multiple values
+  router.get('/multi', async (req, res) => {
```
freben: Maybe rather `/buckets/:bucket/multiget` or so.

URL limits eventually become a problem. The catalog batch fetch is a POST with `{ entityRefs: string[] }` for example, just to be safe.

Note also that the response should ideally be an object as well, e.g. `{ items: Array<JsonValue | null> }` or something like that, in case we want to add more stuff there such as etags or whatever.

grantila:
Rebased and resolved the feedback. Very good points @freben!

 * Changed the endpoint to `/multiget` from `/multi` and turned it into POST to not get into query size issues. The query handles multiple buckets _and_ keys, so it can't effectively be prefixed `/buckets/:bucket`.
 * No longer internally calling the single `get` on `UserSettingsStore` iteratively, but instead a new `multiget()` is implemented, doing `whereIn` for fewer/faster db calls. It splits by buckets (with a concurrency limit of 10), then splits each set of keys into a maximum of 100 elements per `whereIn`, performing them sequentially chunk by chunk.
 * Responds similarly to the catalog with `{ items: ({value: JsonValue} | null)[]; }` matching the input items requested.

grantila:
👋 What do you think @freben, would this be good to go?

freben:
Sorry I've been pondering this one and being slow to get back to you. But it's not forgotten. Main thought is - is the complexity and necessary delay worth it :) But I'll get back to it

grantila:
Sure. Regarding the delay; it could make sense to lower the window of 10ms to something like 2ms, to not take into account deeper component's usages (where there might be quite a bit of compute going on in between), but only when multiple keys are being requested more or less _at once_, such as in #32100.

grantila:
For reference, in https://github.com/backstage/community-plugins/pull/6468, the plugin now loads a bunch of user settings (currently 4, but extensions can add more) at page load to have the settings ready for sub components. These reflect tab order, view options, table sizes, etc. It would benefit from this PR in situations where the backend runs on HTTP/1.1 to have the page load faster.

plugins/user-settings-backend/package.json
```diff
@@ -55,9 +55,11 @@
     "@backstage/plugin-signals-node": "workspace:^",
     "@backstage/plugin-user-settings-common": "workspace:^",
     "@backstage/types": "workspace:^",
-    "express": "^4.22.0",
+    "express": "^4.17.1",
     "express-promise-router": "^4.1.0",
-    "knex": "^3.0.0"
+    "knex": "^3.0.0",
+    "p-limit": "^3.1.0",
+    "zod": "^3.22.4"
```
freben:
```suggestion
    "zod": "^3.25.76"
```

sorry, there was an issue with zod versions < 3.25 and this is the new one we've settled on

plugins/user-settings-backend/src/database/DatabaseUserSettingsStore.ts
```diff
@@ -91,6 +92,77 @@ export class DatabaseUserSettingsStore implements UserSettingsStore {
...
+    for (const item of options.items) {
+      let keys = bucketMap.get(item.bucket);
+      if (!keys) {
+        keys = new Set();
+        bucketMap.set(item.bucket, keys);
+      }
+      keys.add(item.key);
+    }
+
+    const dbLimit = pLimit(10);
```
freben: Move this out to the root level of the file, to make it apply across all callers (to avoid pool exhaustion issues)

grantila:
Good points! Updated accordingly.

Very sorry for pulling in other maintainers due to a merge that failed 🙏 - rebased on master instead to get `yarn.lock` consistency. No other changes but the requested changes from @freben

PR #31107, opened 2025-09-10
feat(dev): allow running example app with docker

drodil:

this PR allows backstage contributors to run the example app against actual postgres, opensearch and redis services using docker and docker compose file.

no changeset as this doesn't affect any published packages.

follow-up for #30722

### DIFF
--- a/app-config.docker.yaml
+++ b/app-config.docker.yaml
@@ -1,0 +1,23 @@
+# To run local backstage instance with docker services,
+# run:
+# 1. `docker compose -f docker-compose.deps.yml up --wait`
+# 2. `BACKSTAGE_ENV=docker yarn dev`
+backend:
+  database:
+    client: pg
+    connection:
+      host: localhost
+      port: 5432
+      user: postgres
+      password: postgres
+  cache:
+    store: redis
+    connection: redis://localhost:6379
+
+search:
+  elasticsearch:
+    provider: opensearch
+    node: 'http://localhost:9200'
+    auth:
+      username: admin
+      password: admin
--- a/docker-compose.deps.yml
+++ b/docker-compose.deps.yml
@@ -1,0 +1,39 @@
+#!/usr/bin/env docker compose -f
+name: backstage
+services:
+  psql:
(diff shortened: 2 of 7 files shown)


drodil:
@Rugvip WDYT? First real use case for #30722 :)

drodil:
Not stale, would be great addition ;)

drodil:
Not stale, come ooon! 👀

docker-compose.deps.yml
Rugvip, line 1: Weird quirk but I really like having this: could you make this file executable and add `#!/usr/bin/env docker compose -f` to the top? That way it can be run directly with `./docker-compose.yml`. Less important for this one but quite convenient for non-default ones.

Speaking of that, maybe this one should actually be something like `docker-compose.deps.yml`? Thinking it might make sense to have a compose at some point that also includes the main app.

app-config.docker.yaml
```diff
@@ -0,0 +1,24 @@
+# To run local backstage instance with docker services,
+# run:
+# 1. `docker-compose up --wait`
+# 2. `export BACKSTAGE_ENV=docker`
+# 3. `yarn dev`
```
Rugvip: Hmm, maybe `BACKSTAGE_ENV=docker yarn dev` instead? Feels like exporting is a bit of a trap where you forget that you set it in the past.

package.json
```diff
@@ -53,6 +53,7 @@
     "snyk:test:package": "yarn snyk:test --include",
     "start": "backstage-cli repo start",
     "start-backend": "echo \"Use 'yarn start example-backend' instead\"",
+    "start:docker": "docker compose up --wait && BACKSTAGE_ENV=docker yarn start",
```
Rugvip: hold up, doesn't `up` block?

I'm thinking this should perhaps only be the `yarn start`?

drodil: No it doesn't, works like a charm 👌

drodil:
@Rugvip no worries, I know the feeling 😃 Stale bot is just too eager to close everything down 😅

PR #31117, opened 2025-09-11
Add the ability to show icons for the tabs on the entity page (new frontend)

Andy2003:
With this PR the ability is added, to configure icons for the tabs of the entity page in the new frontend system.

The Icons can be enabled via config

```yaml
app:
  extensions:
    # Entity page
    - page:catalog/entity:
        config:
          # Enable tab- and group-icons
          showIcons: true
          # Optionally override default groups and their icons
          groups:
            - overview:
                title: Overview
                icon: dashboard
            - documentation:
                title: Docs
                icon: description
    - entity-content:techdocs:
        config:
          group: documentation
          icon: techdocs
```

image

### DIFF
--- a/docs/features/software-catalog/catalog-customization.md
+++ b/docs/features/software-catalog/catalog-customization.md
@@ -725,7 +725,7 @@ app:
 
 Notes:
 
-- Icons for groups and tabs are resolved via the app's IconsApi. When using a string icon id (for example `"dashboard"`), ensure that the corresponding icon bundles are enabled/installed in your app (see the [IconBundleBlueprint documentation](../../reference/frontend-plugin-api.iconbundleblueprint.md)).
+- Icons for groups and tabs are resolved via the app's IconsApi. When using a string icon id (for example `"dashboard"`), ensure that the corresponding icon bundles are enabled/installed in your app (see the [IconBundleBlueprint documentation](https://backstage.io/api/stable/variables/_backstage_plugin-app-react.IconBundleBlueprint.html)).
 - Group icons are only rendered if `showNavItemIcons` is set to `true`.
 
 ### Overriding or disabling a tab's group (per extension)
--- a/plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
+++ b/plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
@@ -102,7 +102,7 @@ export function EntityTabs(props: EntityTabsProps) {
           id: path,
           path: to,
           label: title,
-          icon: icon,
+          icon,
         };
       }),
     [routes],


docs/frontend-system/building-plugins/01-index.md
```diff
@@ -198,6 +198,15 @@ const exampleEntityContent = EntityContentBlueprint.make({
...
     title: 'Example',
+    // Optional: associate this content with a group on the entity page tabs
+    // Use a known default group id like "overview", "quality", "documentation",
+    // or provide a custom string. You can also override or disable this via app-config (see note below).
+    // group: 'overview',
+    // Optional: set a tab icon. When using a string, the icon is resolved via the IconsApi.
+    // Ensure your app has icon bundles enabled/installed so the icon id is available.
+    // Note: icons are shown in the entity page tab groups only if `showIcons` is enabled in the
+    //       catalog entity page config (page:catalog/entity) via app-config.
+    // icon: 'dashboard',
```
awanlin: I'm not sure that this makes sense to be included here, I believe the intention is to give just a basic example here and then use the Common Extension Blueprints page where you can see/find all the exhaustive options.

docs/frontend-system/building-plugins/03-common-extension-blueprints.md
```diff
@@ -57,6 +57,31 @@ Avoid using `convertLegacyEntityCardExtension` from `@backstage/core-compat-api`
...
+          showIcons: true
+          # Optionally override default groups and their icons
+          groups:
+            - overview:
+                title: Overview
+                icon: dashboard
+            - documentation:
+                title: Docs
+                icon: description
+```
```
awanlin, line 99: Should some of this either be added or linked to this: https://backstage.io/docs/features/software-catalog/catalog-customization#new-frontend-system.

awanlin: Also, if we are going to start adding details like this, which I do think is a good idea, we should make sure to include all the props that are available.

docs/frontend-system/building-plugins/01-index.md
Andy2003: removed

docs/frontend-system/building-plugins/03-common-extension-blueprints.md
Andy2003: I updated the doc to reflect the options that are configurable

docs/features/software-catalog/catalog-customization.md
```diff
@@ -698,3 +698,63 @@ filter:
...
+```yaml
+app:
+  extensions:
+    # ...
+    # Example entity content extension instance id
+    - entity-content:example/my-content:
+        config:
+          # Move this tab to a custom group you defined above
+          group: custom
+          # Show an icon for this entity content page but only if
```
freben: Missing some text

docs/features/software-catalog/catalog-customization.md
```diff
@@ -698,3 +698,63 @@ filter:
...
+                title: Quality
+                icon: verified
+            - documentation:
+                title: Docs
+                icon: description
+```
+
+Notes:
+
+- Icons for groups and tabs are resolved via the app's IconsApi. When using a string icon id (for example `"dashboard"`), ensure that the corresponding icon bundles are enabled/installed in your app (see the Icons blueprint in the frontend system documentation).
```
freben: Is it possible to link to that documentation?

docs/frontend-system/building-plugins/03-common-extension-blueprints.md
```diff
@@ -57,6 +57,31 @@ Avoid using `convertLegacyEntityCardExtension` from `@backstage/core-compat-api`
 
 Creates entity content to be displayed on the entity pages of the catalog plugin. Exported as `EntityContentBlueprint`.
 
+Supports optional params such as `group` and `icon`to:
```
freben:
```suggestion
Supports optional params such as `group` and `icon` to:
```

packages/app-next/src/App.tsx
```diff
@@ -45,6 +45,8 @@ import { convertLegacyPageExtension } from '@backstage/core-compat-api';
 import { convertLegacyEntityContentExtension } from '@backstage/plugin-catalog-react/alpha';
 import { pluginInfoResolver } from './pluginInfoResolver';
 import { appModuleNav } from './modules/appModuleNav';
+import { default as catalogPlugin } from '@backstage/plugin-catalog/alpha';
```
freben:
```suggestion
import catalogPlugin from '@backstage/plugin-catalog/alpha';
```

.changeset/light-meals-cover.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-catalog-react': minor
+'@backstage/plugin-catalog': minor
```
freben, line 6: We ned to add changesets for api-docs, kubernetes, and techdocs too I think - their capabilities technically changed

plugins/catalog/src/alpha/pages.tsx
```diff
@@ -189,13 +207,18 @@ export const catalogEntityPage = PageBlueprint.makeWithOverrides({
                 header={header}
                 contextMenuItems={filteredMenuItems}
               >
-                {Object.values(groups).flatMap(({ title, items }) =>
+                {Object.values(groups).flatMap(({ title, icon, items }) =>
                   items.map(output => (
                     <EntityLayout.Route
-                      group={title}
+                      group={{ title, icon }}
```
freben: Hm this feels wrong.

I had a look through the code for the entity tabs controls (especially all the way down in `EntityTabsList`). What I expected, is that the `group` value should be nothing but _a unique identifier_, not related to rendering at all. For example, if your config is

```yaml
            - foo:
                title: Overview
                icon: dashboard
```

then the `group` value should be `foo`, and the `title` value should be `Overview`. But it seems to me, that `EntityTabsList` erroneously renders the `group` of the item, instead of the `title` (the latter is effectively just thrown away!).

Rather than making this situation even worse, could you look into whether you could actually fix the tab control to render the right thing? And of course to try it all out visually, to see if it still groups and renders correctly rather than leading to unexpected breakages because code relied on that faulty behavior in some places.

Andy2003: I updated the whole group resolution logic

plugins/catalog-react/src/alpha/blueprints/extensionData.tsx
```diff
@@ -33,18 +41,50 @@ export const entityFilterExpressionDataRef =
     id: 'catalog.entity-filter-expression',
   });
 
+/** @alpha */
+export type GroupDefinitions = Record<
```
freben:
```suggestion
export type EntityContentGroupDefinitions = Record<
```

It's so specific and narrowly used, that I thought it could have a specific name. Unless you think it'll be used in more places?

plugins/catalog/src/alpha/components/EntityTabs/EntityTabsGroup.tsx
```diff
@@ -138,22 +142,41 @@ const styles = (theme: Theme) =>
...
 
+function resolveIcon(
+  icon: string | ReactElement | undefined,
+  iconsApi: IconsApi,
+  showIcons: boolean,
+) {
+  if (!showIcons) {
+    return undefined;
+  }
+  const itemIcon = icon;
```
freben: you don't need this extra variable do you? the type checker should be fine with narrowing to just `string` inside the `if`, directly for the `icon` parameter

plugins/catalog/src/alpha/components/EntityTabs/EntityTabsGroup.tsx
```diff
@@ -138,22 +142,41 @@ const styles = (theme: Theme) =>
...
+  if (typeof itemIcon === 'string') {
+    const Icon = iconsApi.getIcon(itemIcon);
+    if (Icon) {
+      return <Icon />;
+    }
```
freben, line 171: Should this one have an `else { return undefined; }`? I thought this function was meant to just return `JSX.Element | undefined`, not `string`, which would happen on the line below otherwise

freben (commented):
Alright, still probably want to play around with it to see how it works since a fair bit changed, but looking promising

Andy2003:
> * `showIcons` is fairly generic - should it be `showNavItemIcons` or so throughout?

I renamed it.

> * I note that the styling of the tabs changes quite a lot with the button update; that probably wasn't intentional and might interact with people's custom styles?

The style change did not come with my changes. Even on master if you run `yarn start:next` you will get this:

image

@freben Should I change the styling to match the old ui?

Andy2003:
@freben can you please restart the Mircosite Job, it was just canceled after a timeout

Andy2003:
@freben Is there any blocker on this PR or can it be merged? We wait for this changes to be released...

plugins/catalog/src/alpha/components/EntityTabs/EntityTabsGroup.tsx
```diff
@@ -191,50 +215,44 @@ const Tab = forwardRef(function Tab(props: EntityTabsGroupProps, ref: any) {
...
-      [classes.labelIcon!]: label && icon,
-      [classes.fullWidth!]: fullWidth,
-      [classes.wrapped!]: wrapped,
-    },
-  ];
-
   if (items.length === 1) {
     return (
-      <ButtonBase
+      <Button
```
freben, line 227: Did this change noticeably affect styling?

freben:
Sorry I saw that the API reports were conflicting so I rebased and pushed (without changing anything else) to be able to review better without bothering you further.

cdedreuille:
Thanks a lot for this work on it. Before giving more design feedback on this feature I would like to share some light on some of the improvements we are looking to bring in the next few months on both the plugin header and navigation.
- We have a new header design that we are looking to bring. In short the idea is for this header to be implemented automatically and adapt to each plugin automatically. [RFC](https://github.com/backstage/backstage/issues/31412)
- This new header will also include a new structure for navigation to be automatically generated and adaptable. The idea here is to be able to have the plugin navigation at the top of the plugin or as sub navigation on the left sidebar for example.

This is just to say that a lot of new improvements are coming into this part very soon.

From a design perspective I'm not really sure that allowing tabs to have their own icon actually make sense right now. Today each plugin has their own icon as a powerful brand item and I fear that adding more icons for each tabs will dilute the visual language. We would love to push even more the main icon for each plugin and keep navigation as simple as possible.

If we allow users to move plugin navigation to the sidebar, having icons everywhere could be quite overwhelming.

Andy2003:
@cdedreuille This is how we use it today (and also before the NFS):

image

So we need a way to customize these tabs, with our icons.

If a user do not want to use these icons he can disable them in the app-config via `showIcons: false` (as explained in the PR description)

cdedreuille:
Thanks a lot for sharing. I'm always amazed to see how teams customise their Backstage instance. I would be happy to understand how you customised the experience and see if there's any way to achieve your goal without adding this new functionality for everyone, specially as we are very close to a migration in that area.

Andy2003:
We have a custom `EntityLayoutBlueprint` that is rendering the header, second level navigation and the entity content.
The loader of this extension is getting the `routes` with icons:

```typescript
EntityLayoutBlueprint.makeWithOverrides({
    name: 'custom-layout',
    factory: originalFactory => {
      return originalFactory({
        order: 10,
        loader: async () => {
          return ({ routes }) => {
            const { route, element, selectedSubPath } =
              useSelectedSubRoute(routes);

            const subRouteState = useMemo(
              () => ({
                subRoutes: routes,
                selectedSubRoute: route,
                selectedSubPath,
              }),
              [selectedSubPath, routes, route],
            );

            const subNavigation = (
              <SubRouteNavigation
                subRoutes={routes} // routes with icons
                selectedSubRoute={route}
                selectedSubPath={selectedSubPath}
              />
            );
            return (
              <SubRouteContext.Provider value={subRouteState}>
                <CustomEntityLayout subNavigation={subNavigation}>
                  {element}
                </CustomEntityLayout>
              </SubRouteContext.Provider>
            );
          };
        },
      });
    },
  })
```

Andy2003:
Any updates on this PR? Will it be merged?

alexvoedi:
This is still a desired feature.

cdedreuille:
Sorry for the long delay here. As long as the icons are not shown by default and this is an option, then I'm personally fine with it. @freben how does that sound for you?

awanlin (approved):

.changeset/light-meals-cover.md
```diff
@@ -0,0 +1,9 @@
+---
+'@backstage/plugin-catalog-react': minor
+'@backstage/plugin-kubernetes': minor
+'@backstage/plugin-api-docs': minor
+'@backstage/plugin-techdocs': minor
```
Rugvip:
```suggestion
'@backstage/plugin-kubernetes': patch
'@backstage/plugin-api-docs': patch
'@backstage/plugin-techdocs': patch
```

Bringing these down to patch to avoid breaking bumps for the 0.x packages and in general the changes in these are slim

Rugvip (approved):
Awesome stuff, thank you! 🎉 :shipit:

docs/features/software-catalog/catalog-customization.md
```diff
@@ -696,3 +696,63 @@ filter:
...
+                title: Quality
+                icon: verified
+            - documentation:
+                title: Docs
+                icon: description
+```
+
+Notes:
+
+- Icons for groups and tabs are resolved via the app's IconsApi. When using a string icon id (for example `"dashboard"`), ensure that the corresponding icon bundles are enabled/installed in your app (see the [IconBundleBlueprint documentation](../../reference/frontend-plugin-api.iconbundleblueprint.md)).
```
Rugvip: ah, this link needs an update, we've moved the reference docs

Rugvip (approved):
Let's :shipit: 🚀

PR #31121, opened 2025-09-11
fix(cli): identify custom patterns that extend the default pattern

pjungermann:
If a custom pattern is used, `backstage.json` will not be updated. However, if the custom pattern extends the default pattern, `backstage.json` will be updated, too.

Example:

- `@backstage/*` (default)
- `@{backstage,backstage-community}/*`
- `@{extra1,backstage,extra2}/*`

Fixes: #28839

### DIFF
--- a/packages/cli/src/modules/migrate/commands/versions/bump.ts
+++ b/packages/cli/src/modules/migrate/commands/versions/bump.ts
@@ -18,6 +18,7 @@ maybeBootstrapProxy();
 
 import fs from 'fs-extra';
 import chalk from 'chalk';
+import { minimatch } from 'minimatch';
 import semver from 'semver';
 import { OptionValues } from 'commander';
 import yaml from 'yaml';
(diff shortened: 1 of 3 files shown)


packages/cli/src/modules/migrate/commands/versions/bump.ts
```diff
@@ -78,6 +78,18 @@ type PkgVersionInfo = {
...
 };
 
+function extendsDefaultPattern(pattern: string): boolean {
+  // Matches
+  // - '@backstage/*'
+  // - '@{backstage}/*'
+  // - '@{backstage,...}/*'
+  // - '@{...,backstage}/*'
+  // - '@{...,backstage,...}/*'
+  return /@(?:backstage|\{(?:[^,}]*,)?backstage(?:,[^,}]*)*})\/\*/.test(
```
freben: Can we invert this and basically do (pseudocode) `pattern.test('@backstage/')`? :) Or are you worried that patterns might be used that require a specific shape of the name to the right of the slash? If not, that would feel like the more robust and expected test.

pjungermann: I've inverted it. Thanks for the suggestion!

freben (approved):

PR #31122, opened 2025-09-11
feat(yarn-plugin): custom manifest location

### LINKED ISSUE #31101: Backstage yarn plugin to support offline mode
### 📜 Issue Labels

### 🔎 Search Terms

```plain
yarn plugin tooling
```

### 🗃️ Project Area

CLI Tooling

### 🔖 Need

We are using the awesome Backstage yarn plugin and are building Backstage in an environment where internet access is limited. This causes errors during `yarn install` as the yarn plugin cannot fetch the correct package versions from the internet. After digging in the plugin itself, it seems it tries to connect to two different addresses:

- https://versions.backstage.io/v1/releases/1.42.2/manifest.json
- https://raw.githubusercontent.com/backstage/versions/main/v1/releases/1.42.2/manifest.json

Opening the internet access is one option, but it might be hard or impossible due to regulatory requirements. Thus, it would be great if the plugin could support offline mode.

### 📝 Proposal

The proposal is to allow downloading the manifest file from the internet to the local repository and passing the file name, for example, in `BACKSTAGE_YARN_PLUGIN_MANIFEST` environment variable to the plugin. The plugin would then use that file for determining the package versions instead of fetching the information from the internet.

### 🔄 Alternatives

_No response_

Undecided

drodil:

added two environment variables that can be used to control the location of backstage manifest file used by the yarn plugin. this helps using the plugin in restricted environments without direct or access at all to the internet.

closes #31101

### DIFF
--- a/packages/cli/src/modules/migrate/commands/versions/bump.test.ts
+++ b/packages/cli/src/modules/migrate/commands/versions/bump.test.ts
@@ -1061,18 +1061,19 @@ describe('environment variables', () => {
 
   beforeEach(() => {
     delete process.env.BACKSTAGE_MANIFEST_FILE;
-    delete process.env.BACKSTAGE_MANIFEST_BASE_URL;
+    process.env.BACKSTAGE_VERSIONS_BASE_URL = 'https://custom.example.com';
   });
 
   afterEach(() => {
     jest.resetAllMocks();
-    delete process.env.BACKSTAGE_MANIFEST_FILE;
-    delete process.env.BACKSTAGE_MANIFEST_BASE_URL;
   });
 
-  it('should use custom base URL when BACKSTAGE_MANIFEST_BASE_URL is set', async () => {
-    process.env.BACKSTAGE_MANIFEST_BASE_URL = 'https://custom.example.com';
+  afterAll(() => {
+    delete process.env.BACKSTAGE_MANIFEST_FILE;
+    delete process.env.BACKSTAGE_VERSIONS_BASE_URL;
+  });
 
+  it('should use custom base URL when BACKSTAGE_VERSIONS_BASE_URL is set', async () => {
     mockDir.setContent({
       'yarn.lock': lockfileMock,
       'package.json': JSON.stringify({
@@ -1230,9 +1231,7 @@ describe('environment variables', () => {
     });
   });
 
-  it('should use custom base URL for yarn plugin when BACKSTAGE_MANIFEST_BASE_URL is set with yarn plugin', async () => {
-    process.env.BACKSTAGE_MANIFEST_BASE_URL = 'https://custom.example.com';
-
+  it('should use custom base URL for yarn plugin when BACKSTAGE_VERSIONS_BASE_URL is set with yarn plugin', async () => {
     mockDir.setContent({
       '.yarnrc.yml': yarnRcMock,
       'yarn.lock': lockfileMock,
@@ -1331,8 +1330,6 @@ describe('environment variables', () => {
   });
 
   it('should handle network errors when using custom base URL', async () => {
-    process.env.BACKSTAGE_MANIFEST_BASE_URL = 'https://custom.example.com';
-
     mockDir.setContent({
       'yarn.lock': lockfileMock,
       'package.json': JSON.stringify({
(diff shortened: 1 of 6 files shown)


packages/cli/src/modules/migrate/commands/versions/bump.test.ts
```diff
@@ -1055,3 +1054,313 @@ describe('createVersionFinder', () => {
     );
   });
 });
+
+describe('environment variables', () => {
+  const worker = setupServer();
+  registerMswTestHooks(worker);
+
+  beforeEach(() => {
```
fearphage, line 1062: If the goal is to stop the local environment from affecting the tests, this should be a `beforeAll`. Otherwise you're effectively running the following for each test.

```js
    // afterEach
    delete process.env.BACKSTAGE_MANIFEST_FILE;
    delete process.env.BACKSTAGE_VERSIONS_BASE_URL;
    // beforeEach
    delete process.env.BACKSTAGE_MANIFEST_FILE;
    delete process.env.BACKSTAGE_VERSIONS_BASE_URL;
``

File: packages/cli/src/modules/migrate/commands/versions/bump.test.ts [hunk 2]
```diff
@@ -1055,3 +1054,313 @@ describe('createVersionFinder', () => {
...
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+    delete process.env.BACKSTAGE_MANIFEST_FILE;
+    delete process.env.BACKSTAGE_VERSIONS_BASE_URL;
+  });
+
+  it('should use custom base URL when BACKSTAGE_VERSIONS_BASE_URL is set', async () => {
+    process.env.BACKSTAGE_VERSIONS_BASE_URL = 'https://custom.example.com';
```
fearphage: All of these tests look very similar, but it's hard to tell what's unique about each test without a lot of scrolling and flipping back and forth. Could you pull some of the setup into a shared `beforeEach` or maybe a function that makes the unique bits obvious?

packages/yarn-plugin/src/util/getPackageVersion.test.ts
```diff
@@ -0,0 +1,460 @@
...
+const mockGetCurrentBackstageVersion =
+  getCurrentBackstageVersion as jest.MockedFunction<
+    typeof getCurrentBackstageVersion
+  >;
+
+describe('getPackageVersion', () => {
+  const mockConfiguration = {} as Configuration;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
```
fearphage: This should be in the `afterEach`.

drodil:
Not stale

drodil:
@madsteer @mtlewis WDYT? This would really help our case where we don't have direct internet access to the `versions.backstage.io` nor `github.com`. Would still like to use the yarn plugin, as the old way is a mess.

madsteer:
> @madsteer @mtlewis WDYT? This would really help our case where we don't have direct internet access to the `versions.backstage.io` nor `github.com`. Would still like to use the yarn plugin, as the old way is a mess.

I'm probably not qualified to comment on this.  I'm not part of the support team for backstage etc.

I do have a "higher level" question about this.  can you not just add [npmRegistries:](https://yarnpkg.com/configuration/yarnrc#npmRegistries) and [npmRegistryServer:](https://yarnpkg.com/configuration/yarnrc#npmRegistryServer) config to your `.yarnrc.yml` file to proxy out?  I took don't have direct access to the internet for when I work on this, but my org uses JFrog Artifactory to "proxy" out for javascript dependencies etc. and so we just put config in the `.yarnrc.yml` file to tell yarn where to go to get dependencies.  I'm not following how this is needed versus just doing that?

drodil:
@madsteer thanks for the inputs. The issue is not downloading the packages from private registry, but with the yarn plugin reading backstage version number from `backstage.json`, and then using that to fetch the correct versions from `versions.backstage.io` (or github as a fallback) for all `@backstage/*` packages.

As we do not have internet access from our build infra, it cannot resolve the versions to be used. This PR would allow us to set up a proxy or cache server to serve the `manifest.json` file, or even just copy the manifest file to the repository and use it from local filesystem.

PR #31126, opened 2025-09-12
`scaffolder`: Allow executing `ActionsRegistry` actions from the Scaffolder

benjdlambert:
Not 100% happy with this implementation, but I think it's the best way forward for now.

Really wanna clean this up once the old types can be removed and we can evolve the interfaces easier.

This allows the `Actions` defined in the `ActionsRegistry` to be both browsable in the `Actions` list page, as well as being able to execute them in a template run.

### DIFF
--- a/plugins/scaffolder-backend/src/service/router.test.ts
+++ b/plugins/scaffolder-backend/src/service/router.test.ts
@@ -280,7 +280,7 @@ describe('scaffolder router', () => {
       expect(response.body.length).toBe(2);
     });
 
-    it('should include actiosn from the remote actions registry', async () => {
+    it('should include actions from the remote actions registry', async () => {
       const mockActionsRegistry = actionsRegistryServiceMock();
       mockActionsRegistry.register({
         name: 'my-demo-action',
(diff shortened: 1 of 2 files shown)


plugins/scaffolder-backend/src/service/router.test.ts
```diff
@@ -275,6 +279,49 @@ describe('scaffolder router', () => {
       expect(response.body[0].id).toBeDefined();
       expect(response.body.length).toBe(2);
     });
+
+    it('should include actiosn from the remote actions registry', async () => {
```
freben:
```suggestion
    it('should include actions from the remote actions registry', async () => {
```

plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -301,7 +306,47 @@ export async function createRouter(
...
+        action.attributes?.readOnly === true &&
+        action.attributes?.destructive === false,
+      handler: async ctx => {
+        const { output } = await actionsRegistry.invoke({
+          id: action.id,
+          input: ctx.input,
+          credentials: await ctx.getInitiatorCredentials(),
+        });
+
+        if (isObject(output)) {
```
freben: Is `isPlainObject` too harsh? Maybe it is. Just heads up that [`isObject` is pretty lenient](https://lodash.com/docs/4.17.15#isObject), but maybe that's fine

benjdlambert: Yeah I changed it, requires a cast now though in the `Object.entries` but it's fine,

Issue #31114, opened 2025-09-11
backstage-cli version:bump: package moves with yarn plugin don't replace `backstage:^` version


pjungermann:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
- yarn plugin
- version:bump
- version bump
- "version bump"
```

### 🗃️ Project Area

CLI Tooling

### 🔗 External Integration

N/A

### 📝 Description & Context

```
Checking for moved packages to the @backstage-community namespace...
Found a moved package @backstage/plugin-search-backend-module-explore@backstage:^ -> @backstage-community/plugin-search-backend-module-explore in backend (dependencies)
Updated 1 files in backend to use the new package names
[94m➤[39m YN0000: · [1mYarn 4.9.4[22m
[94m➤[39m [90mYN0000[39m: ┌ Resolution step
[91m➤[39m ]8;;https://yarnpkg.com/advanced/error-codes#yn0001---exceptionYN0001]8;;: │ Error: Package @backstage-community/plugin-search-backend-module-explore not found in manifest for Backstage v1.42.5. This means the specified package is not included in this Backstage release. This may imply the package has been replaced with an alternative - please review the documentation for the package. If you need to continue using this package, it will be necessary to switch to manually managing its version.
    at f (/Users/patrick.jungermann/workspaces/bonial/sdlc/backstage/backstage/.yarn/plugins/@yarnpkg/plugin-backstage.cjs:6:4053)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async $ (/Users/patrick.jungermann/workspaces/bonial/sdlc/backstage/backstage/.yarn/plugins/@yarnpkg/plugin-backstage.cjs:6:5486)
    at async t.reduceHook (/Users/patrick.jungermann/workspaces/bonial/sdlc/backstage/backstage/.yarn/releases/yarn-4.9.4.cjs:148:3961)
    at async t.preparePackage (/Users/patrick.jungermann/workspaces/bonial/sdlc/backstage/backstage/.yarn/releases/yarn-4.9.4.cjs:209:5961)
    at async X (/Users/patrick.jungermann/workspaces/bonial/sdlc/backstage/backstage/.yarn/releases/yarn-4.9.4.cjs:209:7587)
    at async Promise.allSettled (index 2)
[...]

PR #31146, opened 2025-09-15
fix(cli): replace `backstage:^` with explicit version for moved packages

pjungermann:
Example:

```
@backstage/some-package@backstage:^ -> @backstage-community/some-package@^1.0.1
```

Fixes: #31114

### DIFF
--- a/packages/cli/src/modules/migrate/commands/versions/migrate.test.ts
+++ b/packages/cli/src/modules/migrate/commands/versions/migrate.test.ts
@@ -18,6 +18,7 @@ import {
   createMockDirectory,
 } from '@backstage/backend-test-utils';
 import * as runObj from '@backstage/cli-common';
+import * as versioning from '../../../../lib/versioning';
 import migrate from './migrate';
 import { withLogCollector } from '@backstage/test-utils';
 import fs from 'fs-extra';
@@ -52,6 +53,12 @@ jest.mock('@backstage/cli-common', () => {
   };
 });
 
+jest.mock('../../../../lib/versioning', () => {
+  return {
+    fetchPackageInfo: jest.fn(),
+  };
+});
+
 function expectLogsToMatch(receivedLogs: String[], expected: String[]): void {
   expect(receivedLogs.filter(Boolean).sort()).toEqual(expected.sort());
 }
@@ -81,7 +88,7 @@ describe('versions:migrate', () => {
         '@backstage': {
           custom: {
             'package.json': JSON.stringify({
-              name: '@backstage-extra/custom',
+              name: '@backstage/custom',
               version: '1.0.1',
               backstage: {
                 moved: '@backstage-community/custom',
@@ -90,7 +97,7 @@ describe('versions:migrate', () => {
           },
           'custom-two': {
             'package.json': JSON.stringify({
-              name: '@backstage-extra/custom-two',
+              name: '@backstage/custom-two',
               version: '1.0.0',
               backstage: {
                 moved: '@backstage-community/custom-two',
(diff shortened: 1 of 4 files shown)


freben:
Sorry for the build error, that's being fixed in https://github.com/backstage/backstage/pull/31147

packages/cli/src/modules/migrate/commands/versions/migrate.ts
```diff
@@ -94,20 +96,32 @@ export async function migrateMovedPackages(options?: {
...
-          delete depsObj[depName];
+        if (depName.startsWith('@backstage/') && depVersion === 'backstage:^') {
+          const info = await fetchPackageInfo(movedPackageName);
+          depsObj[movedPackageName] = `^${info['dist-tags'].latest}`;
+        } else if (
+          movedPackageName.startsWith('@backstage/') &&
+          hasYarnPlugin
+        ) {
+          depsObj[movedPackageName] = 'backstage:^';
+        } else {
```
Rugvip: Feels like the order of these two conditions should be swapped? 🤔

Rugvip (commented):
Nice, thank you! 😁

packages/cli/src/modules/migrate/commands/versions/migrate.ts
pjungermann: Do you mean the following?
```
        } else if (
          hasYarnPlugin &&
          movedPackageName.startsWith('@backstage/')
        ) {
```

Rugvip: Yep exactly, probably this in total:

```ts
if (movedPackageName.startsWith('@backstage/') && hasYarnPlugin) {
} else if (depVersion === 'backstage:^') {
} else {
}
```

pjungermann: I think `movedPackageName.startsWith('@backstage/') && hasYarnPlugin` is the least likely case, while `depName.startsWith('@backstage/') && depVersion === 'backstage:^'` is the most likely case among these -- besides the else-case.

The check for the package name might be unnecessary, true.

```ts
if (depVersion === 'backstage:^') {
} else if (hasYarnPlugin && movedPackageName.startsWith('@backstage/')) {
} else {
}
```

packages/cli/src/modules/migrate/lib/utils.ts
```diff
@@ -14,10 +14,53 @@
...
-import ora from 'ora';
+import fs from 'fs-extra';
 import chalk from 'chalk';
+import ora from 'ora';
+import yaml from 'yaml';
+import z from 'zod';
+import { paths } from '../../../lib/paths';
 import { run } from '../../../lib/run';
 
+const yarnRcSchema = z.object({
```
freben, line 25: Super silly corner case: Parsing a completely empty file, or a file of only whitespace, returns `null`. Same if someone starts adding document markers like `---` at the top; that's a `null` document preceding the actual content.

Not sure if worth caring about, but I would believe that at least a file of only whitespace is valid. I do realize this is just copied code too. Feel free to ignore.

packages/cli/src/modules/migrate/commands/versions/migrate.ts
```diff
@@ -94,20 +96,32 @@ export async function migrateMovedPackages(options?: {
...
+        didPackageChange = true;
+        didAnythingChange = true;
 
-          didPackageChange = true;
-          didAnythingChange = true;
+        delete depsObj[depName];
 
-          depsObj[movedPackageName] = depsObj[depName];
-          delete depsObj[depName];
+        if (depVersion === 'backstage:^') {
```
freben: I'd go as far as saying `.startsWith('backstage:')`, but no biggie

pjungermann: I will adjust this

packages/cli/src/modules/migrate/lib/utils.ts
pjungermann: for context: this was moved as it is from `bump.ts`

Currently, the cli would fail with `Unexpected content in .yarnrc.yml` I think.

For now, I would leave it as it is.

vinzscam (approved):
good catch 🙏

pjungermann:
trying to fix the pending DCO check with a "reupload". Not sure if there is another way.

packages/cli/src/modules/migrate/commands/versions/migrate.ts
```diff
@@ -94,20 +96,32 @@ export async function migrateMovedPackages(options?: {
...
+          const info = await fetchPackageInfo(movedPackageName);
+          depsObj[movedPackageName] = `^${info['dist-tags'].latest}`;
+        } else if (
+          hasYarnPlugin &&
+          movedPackageName.startsWith('@backstage/')
+        ) {
+          depsObj[movedPackageName] = 'backstage:^';
+        } else {
+          depsObj[movedPackageName] = depVersion;
         }
```
Rugvip: I'm still thinking the logic here is wrong, or am I misunderstanding the state?

Feeling it would make sense to first check if we're using the yarn plugin and that the target package is a `@backstage/*` package, then we can use `'backstage:^'`. If that is not the case we know for sure we can't use the Backstage protocol, so if the current version range uses the protocol, then look up the latest version, and otherwise keep using the existing range.

packages/cli/src/modules/migrate/commands/versions/migrate.test.ts
```diff
@@ -169,6 +176,129 @@ describe('versions:migrate', () => {
...
+        };
+      });
+
+    const { warn, log: logs } = await withLogCollector(async () => {
+      await migrate({});
+    });
+
+    expectLogsToMatch(logs, [
+      'Checking for moved packages to the @backstage-community namespace...',
+      'Found a moved package @backstage/custom@backstage:^ -> @backstage-community/custom in a (dependencies)',
```
Rugvip, line 262: Could you add a test for `@backstage/custom` -> `@backstage/custom-two` as well? That'd test what I'm talking about below

packages/cli/src/modules/migrate/commands/versions/migrate.ts
vinzscam: cc @pjungermann

pjungermann: @rugvip means you would ignore the "moved package" part and only focus on the package itself, right?

freben:
Ping @pjungermann

packages/cli/src/modules/migrate/commands/versions/migrate.test.ts
pjungermann: I will add one later today

packages/cli/src/modules/migrate/commands/versions/migrate.ts
pjungermann: I wonder if we want to continue using the existing range even for the last case. 🤔

pjungermann: I've swapped the order of conditions checked. However, this had no impact on the test/functionality. It still seemed good to apply.

packages/cli/src/modules/migrate/commands/versions/migrate.test.ts
```diff
@@ -345,4 +475,84 @@ describe('versions:migrate', () => {
...
+      expect.any(Object),
+    );
+
+    const pkgJson = await fs.readJson(
+      mockDir.resolve('packages/a/package.json'),
+    );
+    expect(pkgJson).toEqual({
+      name: 'a',
+      dependencies: {
+        '@backstage/custom-two': '^2.0.0',
```
pjungermann, line 545: @Rugvip This is where I wondered whether we shouldn't fetch the version from custom-two for these cases, too, instead of continuing to use the same version range. 🤔 E.g., here we could have `^3.0.0` instead. Of course, you might not want that for changes like `^2.0.0` to `^2.1.2`, etc. to reduce unnecessary changes to these version ranges.

benjdlambert:
@Rugvip do you wanna have another check of this? :pray:

packages/cli/src/modules/migrate/commands/versions/migrate.test.ts
Rugvip: Yep, happy with fetching the new version ofc 👍. My concern was with the use of the Yarn plugin and keeping it around. If the original dep here uses `backstage:^` we end up resetting that to a fixed version right? while we'd instead rather want it to stick to the `backstage:^` range.

Rugvip:
Still think https://github.com/backstage/backstage/pull/31146#discussion_r2495674649 needs to be addressed. I think moves from `@backstage/*` to `@backstage/*` don't really work as intended, removing `backstage:` ranges.

vinzscam:
@pjungermann would you mind rebasing this? (sorry I don't have permissions)

PR #31151, opened 2025-09-15
Restrict Chromatic to only build BUI

cdedreuille:

As we merged all our Storybook into a single version, the amount of Chromatic Snapshots went through the roof. This PR restrict our visual regression testing to only work on Backstage UI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -187,3 +187,4 @@ tsconfig.typedoc.tmp.json
 
 # Storybook
 dist-storybook/
+dist-storybook-chromatic/
--- a/.github/workflows/verify_storybook.yml
+++ b/.github/workflows/verify_storybook.yml
@@ -49,8 +49,8 @@ jobs:
         with:
           cache-prefix: ${{ runner.os }}-v${{ matrix.node-version }}
 
-      - name: Build Storybook
-        run: yarn build-storybook
+      - name: Build Storybook for Chromatic
+        run: yarn build-storybook:chromatic
 
       - name: Deploy Storybook to Chromatic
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -186,4 +186,5 @@ docs.json
 tsconfig.typedoc.tmp.json
 
 # Storybook
-dist-storybook/
\ No newline at end of file
+dist-storybook/
+dist-storybook-chromatic/
```
Rugvip: Thinking we don't really need different outputs for these, but just stick to `dist-storybook` for all?

cdedreuille: We do have a different build for https://backstage.io/storybook

This one has all stories. I'm not sure how much people are using it but this would be a regression to remove them don't you think?

Rugvip: Just talking about the output directory. There doesn't seem to be a need to be able generate both of these outputs simultaneously?

cdedreuille: Not sure I understand. For Chromatic we need to build Storybook first with only Backstage UI. Chromatic will then build on top of that build. If I won't have two `dist` folders then how can we do these two tasks:
- Publish Storybook with all stories
- Publish & run visual testing only on Backstage UI with Chromatic

Rugvip: But do we ever need to do those things at once? I'm just looking to keep config/workspaces clean by not having too many different output directories. From what I can see it would be fine to output both build variants to the same directory because we never need to have both available at once.

cdedreuille: I finally understood - We will use `dist-storybook` for both.

PR #31154, opened 2025-09-15
frontend-plugin-api: add coreExtensionData.title

Rugvip:

Been seeing the need for this in a couple of places. Previously for the evolution of nav items, and most recently https://github.com/backstage/backstage/pull/31055#discussion_r2350206387. Thinking it's worth adding.

### DIFF
--- a/.changeset/nasty-moose-rescue.md
+++ b/.changeset/nasty-moose-rescue.md
@@ -2,4 +2,4 @@
 '@backstage/frontend-plugin-api': patch
 ---
 
-Added `coreExtensionData.title`, especially useful for creating extensible layout with tabbed pages, but availble for use for other cases too.
+Added `coreExtensionData.title`, especially useful for creating extensible layout with tabbed pages, but available for use for other cases too.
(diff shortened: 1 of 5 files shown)


freben (approved):
No love for `icon` too? :) Maybe as `string | Element`

This is fine to have as a plain string but I wonder if there's going to be asks pretty soon for Element as well. Maybe that ends up with some convention of using `title` as "the text-only version" because it's usable in all contexts and then a more render focused one by its side.

I note you don't inject usage of this new data type anywhere?

The compat/migration aspect is interesting. If you were to add it to an existing extension, it might have to do `myCustomTitleRef.optional(), coreExtensionData.title.optional()`. At that point it's really tempting to invent a `.deprecated()` on data refs.

Rugvip:
Yep we could do more. Didn't intend to ship this one before the release today, so we could always add a few more.

Only tricky part of icon is the type to me, whether we stick with out `IconComponent` or go for something different. Feel we need to make sure it'll work well with BUI too.

Strongly prefer just `string` for the title. We could have `string | Element` or just `ReactNode` in some other places, but this one is indeed intended to help create both things like tab titles and breadcrumbs, but also for example setting the browser page title.

Rugvip:
Shipping, given that a need for this is showing up in https://github.com/backstage/backstage/pull/31225

PR #31164, opened 2025-09-16
chore(deps): update octokit dependencies

JessicaJHee:

Bumps Octokit packages for maintenance and to resolve CVEs

`@octokit/request-error`
- CVE-2025-25289

`@octokit/plugin-paginate-rest`
- CVE-2025-25288

`@octokit/endpoint`
- CVE-2025-25285

`@octokit/request`
- CVE-2025-25290

### DIFF
--- a/plugins/scaffolder-backend-module-github/report.api.md
+++ b/plugins/scaffolder-backend-module-github/report.api.md
@@ -8,7 +8,7 @@ import { CatalogService } from '@backstage/plugin-catalog-node';
 import { Config } from '@backstage/config';
 import { createPullRequest } from 'octokit-plugin-create-pull-request';
 import { GithubCredentialsProvider } from '@backstage/integration';
-import { Octokit } from '@octokit/rest';
+import { Octokit } from '@octokit/core';
 import { OctokitOptions } from '@octokit/core/dist-types/types';
 import { ScmIntegrationRegistry } from '@backstage/integration';
 import { ScmIntegrations } from '@backstage/integration';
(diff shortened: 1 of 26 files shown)


freben:
Correct me if I'm wrong but, isn't the `octokit` overall package a "richer" version of the more low level @opctokit/x` ones? In essence, adding semantics / plugins for retries etc.

JessicaJHee:
@freben Thanks for taking a look at this! You're totally right, the `octokit` package would be the ideal choice with the richer features we want.

The problem I'm facing is that the version of `octokit` [v4.0.0,](https://github.com/octokit/octokit.js/releases/tag/v4.0.0) which includes necessary CVE fixes) is ESM. When I upgrade to them, there are build errors requiring updating `moduleResolution` to `'node16', 'nodenext', or 'bundler’` and tests fail.

As a workaround, I'm trying to pin the individual `@octokit/*` packages to their latest CJS compatible versions that still include the security patches. I'm still a newbie to dependency management - do you think this is a reasonable approach for now or is there a better strategy you'd recommend for handling this?

freben:
Does it work to do `const octokit = require('octokit') as typeof import('octokit');` dynamically instead?

JessicaJHee:
I did try with a few more approaches and learned that:
- all ESM modules need to be dynamically imported with something like `const { createAppAuth } = await import('@octokit/auth-app');`
- to make Jest tests work with ESM module, need to use `unstable_mockModule ` and include `--experimental-vm-modules` flag in `NODE_OPTIONS` of the test script in `package.json` (refer to this [PR](https://github.com/backstage/community-plugins/pull/3269))

Due to a shift in priorities, unfortunately I'm unable to continue working on this PR. Closing it for now, but I'm happy for someone else to take it over.

PR #31166, opened 2025-09-16
refactor: convert constructor parameter properties for erasableSyntaxOnly compatibility

schultzp2020:

Note, I did use AI to update the constructors quickly. I've tried to separate my changes logically into separate commits. 

For context, [here](https://www.typescriptlang.org/tsconfig/#erasableSyntaxOnly) is what the flag does.

> Node.js supports running TypeScript files directly as of v23.6; however, only TypeScript-specific syntax that does not have runtime semantics are supported under this mode. In other words, it must be possible to easily erase any TypeScript-specific syntax from a file, leaving behind a valid JavaScript file.

### DIFF
--- a/packages/config/report.api.md
+++ b/packages/config/report.api.md
@@ -48,7 +48,7 @@ export class ConfigReader implements Config {
   constructor(
     data: JsonObject_2 | undefined,
     context?: string,
-    fallback?: ConfigReader | undefined,
+    fallback?: ConfigReader,
     prefix?: string,
   );
   static fromConfigs(configs: AppConfig[]): ConfigReader;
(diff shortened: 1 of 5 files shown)


packages/config/report.api.md
```diff
@@ -48,7 +48,7 @@ export class ConfigReader implements Config {
   constructor(
     data: JsonObject_2 | undefined,
     context?: string,
-    fallback?: ConfigReader | undefined,
```
schultzp2020, line 51: If we want to keep `undefined` in the API, we'll need to explicitly declare it. I have decided against explicitly declaring `undefined` because I don't believe it was originally intended to function as such. 

Here's the difference, it's explained more in detail [here](https://tkdodo.eu/blog/optional-vs-undefined):

```
declare function canFetch1(networkMode?: NetworkMode): boolean
declare function canFetch2(networkMode: NetworkMode | undefined): boolean
```

- The first one means: This parameter is optional.
- The second one means: This parameter is required, but can be undefined.

Rugvip:
Release is out :shipit:

Issue #31165, opened 2025-09-16
catalog-graph crops graphs in 1.43


grantila:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
catalog graph
```

### 🗃️ Project Area

Unknown

### 🔗 External Integration

N/A

### 📝 Description & Context

When running 1.43, graphs get cropped and cannot be zoomed out to show the full graph anymore. The bug is most likely instroduced in https://github.com/backstage/backstage/pull/31005.

<img width="1379" height="833" alt="Image" src="https://github.com/user-attachments/assets/406da2a3-3c81-4bda-9eef-ce43fc523b70" />

Seems the svg doesn't get resized to the parent (it's clear when using the browser developer tools to see its size), and the button's "float: right" doesn't work either - it pushes down the graph by 48px. `position: absolute; right: 0;` works though. The svg pushed down and not resized to the parent:

<img width="1380" height="704" alt="Image" src="https://github.com/user-attachments/assets/21011c2f-90c6-447a-a562-361275eed6da" />

### 👍 Expected Behavior

Same as before #31005 (when not in fullscreen)

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Undecided

freben:
Mentioning @drodil on here.

drodil:
Hm, weird, didn't see any of this during the fullscreen development. What browser are you using? Maybe it only happens on specific window sizes 🤔

drodil:
Got this reproduced, but it seems it depends on the window size. Any chance to get the fix to the next patch release, if you have more coming for 1.43, @freben ?

roysuranjan:
We recently upgraded Backstage to the latest version and have encountered a significant issue—many of our catalogs are highly nested, with some having over 100 dependencies. This bug has had a considerable impacted. I truly appreciate your ongoing efforts to address it. In the meantime, is there any recommended workaround or temporary patch we could apply to mitigate the issue until a permanent fix is available? Thank you again for your support!

grantila:
[...]

PR #31168, opened 2025-09-16
Fixed DependencyGraph size issue

grantila:

Since #31005 reported in #31165, the catalog graph didn't size the graph correctly.

When debugging this, I also saw in the developer tools that the inner `<svg>` element got updated repeatedly in an indefinite loop, caused by `setEdge` being a function that's not memoized, and the inner `<Edge>` has a hook dependency on it, causing the graph to re-render. The side effect of this was that the sizing measurements were recalculated, so when resizing the window, the graph was resized too, out of pure "luck".

This PR:

 * Fixes the indefinite re-rendering by instead using `useMeasure` to trigger the component to re-render, and implicitly resize, no more hogging browser cpu for no reason.
 * Fixes the "Fullscreen" button to not consume height from the inner svg (pushing it down). It's now properly floating.
 * Fixes inner svg size when `fit` is "contained", and using that prop from the `EntityRelationsGraph`.

### DIFF
--- a/.changeset/public-sites-admire.md
+++ b/.changeset/public-sites-admire.md
@@ -3,4 +3,4 @@
 '@backstage/plugin-catalog-graph': patch
 ---
 
-Fixed DependencyGraph svg size not adapting to the container size
+Fixed DependencyGraph `svg` size not adapting to the container size
--- a/.changeset/public-sites-admire.md
+++ b/.changeset/public-sites-admire.md
@@ -1,0 +1,6 @@
+---
+'@backstage/core-components': patch
+'@backstage/plugin-catalog-graph': patch
+---
+
+Fixed DependencyGraph svg size not adapting to the container size
--- a/packages/core-components/src/components/DependencyGraph/DependencyGraph.tsx
+++ b/packages/core-components/src/components/DependencyGraph/DependencyGraph.tsx
@@ -22,6 +22,8 @@ import {
   useRef,
   useState,
 } from 'react';
+import useMeasure from 'react-use/esm/useMeasure';
+import { once } from 'lodash';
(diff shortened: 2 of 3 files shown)


packages/core-components/src/components/DependencyGraph/DependencyGraph.tsx
```diff
@@ -431,28 +444,43 @@ export function DependencyGraph<NodeData, EdgeData>(
...
+      ref={containerRef}
+      className={combineClasses(
+        styles.root,
+        fit === 'contain' && styles.fixedHeight,
+      )}
+    >
       <FullScreen
         handle={fullScreenHandle}
-        className={fullScreenHandle.active ? styles.fullscreen : styles.root}
+        className={combineClasses(
```
drodil: Maybe use `classnames` utility like in other components instead of introducing own function?

drodil (approved):
Looks good, just one minor idea for improvement

.changeset/public-sites-admire.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/core-components': patch
+'@backstage/plugin-catalog-graph': patch
+---
+
+Fixed DependencyGraph svg size not adapting to the container size
```
freben:
```suggestion
Fixed DependencyGraph `svg` size not adapting to the container size
```

packages/core-components/src/components/DependencyGraph/DependencyGraph.tsx
freben: yeah agreed

grantila: Alright, solved.

Issue #30989, opened 2025-08-26
<br /> tags not handled properly by MarkdownContent


lexhuismans-wk:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
MarkdownContent, html, <br> <br /> </ br>
```

### 🗃️ Project Area

Design System

### 🔗 External Integration

N/A

### 📝 Description & Context

When using `import { MarkdownContent } from '@backstage/core-components'` it is displaying html tags in plain text instead of rendering them properly. Resulting in e.g. `<br />` tags in the markdown content to be shown as plain text.

### 👍 Expected Behavior

Since it should render according to Github flavoured Markdown `<br />` should show up as new lines in the Backstage theme.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Undecided

Ayushmore1214:
@vinzscam Can i work on this

gauravagrawal01:
Similar use case had been registered in past, however, didn't reach out to any conclusion though - https://github.com/backstage/backstage/pull/10973

gauravagrawal01:
My guess is backstage natively doesn't support rendering html inside react markdown yet, as for rendering it inside react markdown needed additional option to be pass on which is `rehypePlugins` (https://github.com/remarkjs/react-markdown?tab=readme-ov-file#fields-2)

Ayushmore1214:
@ahhhndre any thoughts on this ?

freben:
This i sopen for contributions if someone is keen on tackling it.

gauravagrawal01:
> This i sopen for contributions if someone is keen on tackling it.

I can explore...

PR #31169, opened 2025-09-17
feat(core-components): Allow HTML rendering in MarkdownContent

SarthakMeshram:
This change addresses issue #30989, where HTML tags like `<br />` were not being rendered by the `MarkdownContent` component and were instead displayed as plain text.

To fix this, I've introduced a new optional prop, `allowUnsafeHtml`. When this prop is set to `true`, the component will use the `rehype-raw` plugin to safely parse and render HTML within the markdown. This is an opt-in feature to maintain security by default and avoid breaking changes.

---

### ## Checklist Instructions

Here's how to handle each item on the checklist.

#### **:heavy_check_mark: Checklist**

* **[ ] A changeset describing the change and affected packages.**
    * **What to do:** This is required. In your terminal, at the root of the project, run the command `yarn changeset`.
    * **Follow the prompts:**
        1.  Select `@backstage/core-components` (using the spacebar, then Enter).
        2.  Choose `patch` when asked for the version bump.
        3.  For the description, enter something like: `The MarkdownContent component now supports an 'allowUnsafeHtml' prop to render raw HTML.`
    * Once you're done, a new markdown file will be created in the `.changeset` directory. Add and commit this file.

* **[ ] Added or updated documentation**
    * **What to do:** You can check this box. The code comments (TSDoc) you added for the new `allowUnsafeHtml` prop in `MarkdownContent.tsx` count as documentation.

* **[ ] Tests for new functionality and regression tests for bug fixes**
    * **What to do:** For a contribution of this size, manual testing is often acceptable. You can leave this unchecked and mention in the PR description that you verified the fix manually using a test page.

* **[ ] All your commits have a `Signed-off-by` line in the message.**
    * **What to do:** This is required. Your current commit is missing this. You need to amend your last commit to add it.
    * Run this command in your terminal:
        ```bash
        git commit --amend --signoff
        ```
    * Your text editor will open. Just save and close it.
    * Because you've amended the commit, you'll need to "force push" the change to your branch:
        ```bash
        git push --force
        ```

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,6 +98,8 @@
     "rejectFrontendNetworkRequests": true
   },
   "resolutions": {
+    "vfile": "^5.0.0",
+    "vfile-message": "^4.0.0",
     "@changesets/assemble-release-plan@^6.0.0": "patch:@changesets/assemble-release-plan@npm%3A6.0.0#./.yarn/patches/@changesets-assemble-release-plan-npm-6.0.0-f7b3005037.patch",
     "@material-ui/pickers@^3.2.10": "patch:@material-ui/pickers@npm%3A3.3.11#./.yarn/patches/@material-ui-pickers-npm-3.3.11-1c8f68ea20.patch",
     "@material-ui/pickers@^3.3.10": "patch:@material-ui/pickers@npm%3A3.3.11#./.yarn/patches/@material-ui-pickers-npm-3.3.11-1c8f68ea20.patch",
--- a/packages/core-components/src/components/MarkdownContent/MarkdownContent.tsx
+++ b/packages/core-components/src/components/MarkdownContent/MarkdownContent.tsx
@@ -20,6 +20,7 @@ import gfm from 'remark-gfm';
 import { Children, createElement } from 'react';
 import { CodeSnippet } from '../CodeSnippet';
 import { HeadingProps } from 'react-markdown/lib/ast-to-react';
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -98,6 +98,8 @@
     "rejectFrontendNetworkRequests": true
   },
   "resolutions": {
+    "vfile": "^5.0.0",
```
benjdlambert, line 101: What are these for?

tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "outDir": "dist-types",
     "rootDir": ".",
     "useUnknownInCatchVariables": false,
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "esModuleInterop": true
```
benjdlambert, line 16: Why is this needed

packages/core-components/src/components/MarkdownContent/MarkdownContent.tsx
```diff
@@ -20,6 +20,7 @@ import gfm from 'remark-gfm';
 import { Children, createElement } from 'react';
 import { CodeSnippet } from '../CodeSnippet';
 import { HeadingProps } from 'react-markdown/lib/ast-to-react';
+import rehypeRaw from 'rehype-raw'; // 1. IMPORT rehype-raw
```
benjdlambert, line 23: Remove the comments, they're really not relevant to the file as a whole. Just this PR.

PR #31179, opened 2025-09-17
POC: Introduce Biome to replace ESLint and Prettier

schultzp2020:

For the POC I have decided to only modify the cli directories to keep it narrow in scope. This will allow us to slowly roll out Biome to the entire repository if that is the direction we want to go down. 

1. The first commit is introducing Biome
2. The second commit fixes the errors Biome detected when linting. I denoted changes with `// NOTE: Is this correct?`.
3. The third commit is misc changes Biome made without issue.


packages/cli-node/src/monorepo/Lockfile.ts
```diff
@@ -178,9 +178,9 @@ export class Lockfile {
    */
   diff(otherLockfile: Lockfile): LockfileDiff {
     const diff = {
-      added: new Array<{ name: string; range: string }>(),
-      changed: new Array<{ name: string; range: string }>(),
-      removed: new Array<{ name: string; range: string }>(),
+      added: [] as { name: string; range: string }[],
+      changed: [] as { name: string; range: string }[],
+      removed: [] as { name: string; range: string }[],
```
Rugvip, line 183: What's up with this change? 🤔

packages/cli-node/src/roles/PackageRoles.ts
```diff
@@ -135,7 +135,7 @@ export class PackageRoles {
         );
       }
 
-      return this.getRoleInfo(role).role;
+      return PackageRoles.getRoleInfo(role).role;
```
Rugvip, line 138: This seems weirdly opinionated? 🤔

Rugvip (commented):
Thanks! 👍 

I actually experimented with the formatter the other day but wasn't too happy with how it didn't seem to have the same kind of print width handling as prettier.

We do need feature parity with our existing lint rules before moving forward, do you have any idea is that area? In particular the monorepo import rules that makes sure that packages declare their own dependencies correctly, and don't import from outside the package, etc.

PR #31193, opened 2025-09-18
feat: Allow configurable pagelength for Bitbucket Cloud provider

robertoj921:

The BitbucketCloud provider performs entity scans using the [Workspace Code Search](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-other-operations/#api-workspaces-workspace-search-code-get) API.  Our organization had been experiencing regular fluctuations in the number of Locations being generated after each cycle of the provider.  After considerable testing, we determined that a bug in the BitbucketCloud api's pagination can cause duplicate results to be returned.
Though the current default page length is set at 100 results, we had some Bitbucket projects exceeding that number, requiring pagination, and resulting in the duplicate results.  This change allows that parameter to be configurable from the app-config so that we can set that number explicitly to avoid pagination.

This bug has already been [reported](https://jira.atlassian.com/browse/BCLOUD-23644) to Atlassian.


tomislacker:
We thought it was about the number of repositories in a _Project_ and were about ready to tell our devel teams that they should avoid having more than 80-100 repositories in a _Project_ since lower numbers seemed okay. Then I saw this PR show up...

Thanks @robertoj921 ! Hopefully we can get this merged because we definitely want it! I pulled it down locally and can no longer reproduce the problem by changing this parameter to 130 myself.

Atlassian should still fix their garbage endpoint but at least this seems to get me some relief.

pjungermann:
I also got into some discussions with the support. However, they were a bit reluctant to see this as a bug.. 

@robertoj921 What kind of value would you like to use with pagelen?

> so that we can set that number explicitly to avoid pagination.

Sounds like you want to use higher values than 100. However, this is already the max.

> pagelen | Current number of objects on the existing page. Globally, the minimum length is 10 and the maximum is 100. Some APIs may specify a different default.

The default for the workspace code search is 10.

robertoj921:
> Sounds like you want to use higher values than 100. However, this is already the max.

The documentation does state that global pagination has a max of 100 but we are currently using 200 as our pagelength with this fork and I can confirm that it works just fine.  This can be easily confirmed if you have a project with > 100 repos with which you can test.

pjungermann:
The intent behind using 100 is to use the max and reduce unnecessary requests.

We could change the global setting to a higher value.

However, the official docs don't mention a higher limit for that endpoint.

Besides the generic statement that it is possible but, in that case, would be mentioned at the endpoint

Relying on undocumented behavior, maybe even unintentional behavior might cause problems if they decide to change it. 🤔

Did you explore what the maximum is that can be used?

robertoj921:
> Did you explore what the maximum is that can be used?

I just ran a series of tests.  It looks like the `pagelen` limit is 1,000.  Anything over that returns a 400 Bad Request.

> Relying on undocumented behavior, maybe even unintentional behavior might cause problems if they decide to change it.

I agree completely.  I wouldn't recommend simply increasing the global setting for exactly that reason, which is why this PR is to parameterize it so that we users can assume that risk.

.changeset/tame-hairs-smash.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-catalog-backend-module-bitbucket-cloud': patch
+'@backstage/plugin-bitbucket-cloud-common': patch
+---
+
+Allow for passing a pagelen parameter to configure the pagelength property of the BitbucketCloudEntityProvider searchCode pagination to resolve [bug](https://jira.atlassian.com/browse/BCLOUD-23644) pertaining to duplicate results being returned.
```
freben:
```suggestion
Allow for passing a `pagelen` parameter to configure the `pagelength` property of the `BitbucketCloudEntityProvider` `searchCode` pagination to resolve [bug](https://jira.atlassian.com/browse/BCLOUD-23644) pertaining to duplicate results being returned.
```

freben (approved):
Okay, but, is there really nothing we can do client-side to just deduplicate what we get back? That would be the more robust fix, no?

robertoj921:
> Okay, but, is there really nothing we can do client-side to just deduplicate what we get back? That would be the more robust fix, no?

I'm sorry, I should have provided more details on the actual problem.  The problem isn't so much that duplicates are returned, it's that duplicates are returned **IN PLACE OF** valid results.  So, let's say for example, we have a project with 150 repos, each with a `catalog-info.yaml` file.  The BitbucketCloud provider will very consistently return 150 results... but repo 13 may appear twice while repo 134 won't appear at all.  If we eliminate the pagination, so that all results appear on a single page, this problem disappears.

freben:
Oof. Well, even if that gets fixed remotely or we do some other fix locally down the line, this parameter can still be valuable for performance tradeoff tweaking etc, so I don't mind merging it.

PR #31195, opened 2025-09-19
Feat: Mcp-actions - output type support

s1008096:
## Feat: Mcp-actions - output type support

At the moment the return type is always text. As a result of this when an image is returned the MCP client tends to parse this data with the LLM and tends to slow down to a crawl.

### DIFF
--- a/.changeset/funny-areas-follow.md
+++ b/.changeset/funny-areas-follow.md
@@ -1,0 +1,95 @@
+---
+'@backstage/plugin-mcp-actions-backend': minor
+---
+
+## Summary
+
+Adds support for MCP action outputs of type `image`. When an action returns `{ type: 'image', data: <base64> }`, the service now emits an image content block instead of coercing the entire output into a fenced JSON text blob. This avoids unnecessary LLM parsing overhead and enables compatible clients to render images directly.
+
+Previously every tool/action result was wrapped as a text payload:
+
+````json
+{
+  "content": [
+    {
+      "type": "text",
+      "text": "```json\n{...}\n```"
+    }
+  ]
+}
+````
+
+This caused:
+
+- Image generation tools to be treated as text, slowing responses.
+- Failures or degraded behavior when large base64 data was reinterpreted downstream.
+
+Now image outputs pass through structurally:
+
+```json
+{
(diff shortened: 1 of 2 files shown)


grantila:
Support for this can be useful. But an output might naturally contain JSON with a property "type" being "image". So automagically detecting this is unideal.

It might make more sense however, to allow multiple responses from the same action, so that you can return text (markdown) _combined_ with one or more images. For example, [the return type](https://github.com/backstage/backstage/blob/master/packages/backend-plugin-api/src/alpha/ActionsRegistryService.ts#L55) could have an additional property "additional" as `additional: AdditionalResponseType[];` which can be defined to allow certain content types. That way we maintain perfect schema compatibility, while allowing more types of data to be returned in the same response.

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -81,16 +93,26 @@ export class McpService {
...
             {
-              type: 'text',
-              text: ['```json', JSON.stringify(output, null, 2), '```'].join(
-                '\n',
-              ),
+              type: isImageType ? 'image' : 'text',
+              ...(isImageType
+                ? {
+                    data: output.data,
+                    mimeType: 'image/png',
```
drodil, line 107: Shouldn't the `mimeType` come from the action instead of forcing it to `image/png` for every output?

.changeset/funny-areas-follow.md
```diff
@@ -0,0 +1,95 @@
+---
+'@backstage/plugin-mcp-actions-backend': minor
```
schultzp2020, line 2:
```suggestion
'@backstage/plugin-mcp-actions-backend': patch
```

Can you change this to a patch? We generally make non breaking features a patch when a package is still on it's 0th version. 🙂

.changeset/funny-areas-follow.md
```diff
@@ -0,0 +1,95 @@
...
+---
+
+## Summary
```
schultzp2020, line 5: I think it may be better to add this to the README and keep the changeset limited to 1-2 sentences.

PR #31197, opened 2025-09-19
Feat(slack-notification-backend): Adds username as optional config for the Slack notifications

henrikedegrd:

Adds an optional config-value "username" in order for Slack notifications to be sent with a different username. This can be handy if using one App for multiple things.

If no config-value is present, the default name of the App is used as before.

### DIFF
--- a/plugins/notifications-backend-module-slack/src/lib/util.ts
+++ b/plugins/notifications-backend-module-slack/src/lib/util.ts
@@ -23,12 +23,14 @@ import { ChatPostMessageArguments, KnownBlock } from '@slack/web-api';
 export function toChatPostMessageArgs(options: {
   channel: string;
   payload: NotificationPayload;
+  username?: string;
 }): ChatPostMessageArguments {
-  const { channel, payload } = options;
+  const { channel, payload, username } = options;
 
   const args: ChatPostMessageArguments = {
     channel,
     text: payload.title,
+    ...(username && { username }),
     attachments: [
       {
         color: getColor(payload.severity),
--- a/plugins/notifications-backend-module-slack/src/lib/util.ts
+++ b/plugins/notifications-backend-module-slack/src/lib/util.ts
@@ -30,7 +30,7 @@ export function toChatPostMessageArgs(options: {
   const args: ChatPostMessageArguments = {
     channel,
     text: payload.title,
-    ...(username && { username }),
+    username,
     attachments: [
       {
         color: getColor(payload.severity),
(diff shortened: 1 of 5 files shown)


benjdlambert:
@backstage/notifications-maintainers @drodil wanna take a look at this? :pray:

plugins/notifications-backend-module-slack/src/lib/util.ts
```diff
@@ -23,12 +23,14 @@ import { ChatPostMessageArguments, KnownBlock } from '@slack/web-api';
...
   payload: NotificationPayload;
+  username?: string;
 }): ChatPostMessageArguments {
-  const { channel, payload } = options;
+  const { channel, payload, username } = options;
 
   const args: ChatPostMessageArguments = {
     channel,
     text: payload.title,
+    ...(username && { username }),
```
drodil: Wouldn't this just work with:

```ts
ChatPostMessageArguments = { 
  channel,
  payload: formattedPayload,
  username,
};
```

Same for other `username` uses

henrikedegrd: It will work, but it will also send "undefined" as username to slack in the case one have not configured a username. And the Slack API says that username is optional, but I have not tried how it handles undefined.

So in order not to break anything with the Slack-integration, I would keep it as is. (Since username is completely absent from the call if not configured in this case.)

And with a little chat with CoPilot, it also came up with the solution that it would be best to keep it as is.

drodil: Ok, maybe you can leave it here, but then it's unnecessary in https://github.com/backstage/backstage/pull/31197/files#diff-c880a6efbd83150686443397fab1c91e042c979801bdcf8af89e2dee8a51a658R274 and https://github.com/backstage/backstage/pull/31197/files#diff-c880a6efbd83150686443397fab1c91e042c979801bdcf8af89e2dee8a51a658R215 -> those could be just `username: this.username` and let this function handle the undefined like it does. WDYT?

drodil: On the side note, I would expect the Slack library to just stringify this as JSON, which removes undefined values from the request, but I'm not sure.

henrikedegrd: I dug a bit deeper, and you are correct, the slack library removes the undefined values.

So I will fix that and modify the tests for this.

drodil (approved):
LGTM

PR #31203, opened 2025-09-19
Fix light mode colour token in ItemCardHeader

jabrks:

The text colour in the ItemCardHeader component is always white, which makes the content unreadable that when you're rendering on a white background in light mode. This switches to a semantic token instead that will dynamically adjust to whichever theme is currently set.

### DIFF
--- a/.changeset/fast-queens-guess.md
+++ b/.changeset/fast-queens-guess.md
@@ -1,0 +1,5 @@
+---
+'@backstage/core-components': patch
+---
+
+Swap base token for semantic token in ItemCardHeader to ensure readability in light mode.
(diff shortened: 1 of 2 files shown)


martina-equinix:
Hi @jabrks , @camilaibs , I was upgrading to the latest version when I noticed this change made the cards in the Docs custom page look a bit weird:

With the light theme:
 
Screenshot 2025-12-04 at 12 48 02 pm

With the dark theme:


Screenshot 2025-12-04 at 12 56 15 pm



The card title used to be white for both which, for our theme at least, made it readable in both options.

I'm wondering if `theme.palette.text.primary` is the right choice for this 🤔 I also noticed that the cards in the `/create` page have the white titles for both themes as well. And it's the same behaviour in https://demo.backstage.io/create.

I wonder if this should be reverted or if there is a way for me to override it in the theme?

jabrks:
@martina-equinix Thanks for raising, I'll take a look at a fix for this today

PR #31207, opened 2025-09-19
fix(scaffolder): distributed actions to be available

### LINKED ISSUE #31187: Actions registry actions not visible in scaffolder
### 📜 Issue Labels

### 🔎 Search Terms

```plain
scaffolder actions registry
```

### 🗃️ Project Area

Software Templates

### 🔗 External Integration

N/A

### 📝 Description & Context

Actions registered to `actionsRegistryService` are not visible in available actions list in scaffolder (/create/actions list)

### 👍 Expected Behavior

Actions should be visible in the list

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

1. Go to /create/actions
2. Search for get-entities action registered by the catalog backend

Reproducable also on demo site

Undecided

drodil:

depending on the plugin startup order, some distributed actions might not be available when the scaffolder backend starts. this was mentioned in the comment that is now removed. while this might change in the future, this would patch the current way the distributed actions are fetched.

basically every time a task is started and when listing of the actions is requested, the distributed actions will get a refresh. while it's still possible that the distributed action suddenly disappears between the start of the task and when the action is fetched from the actionregistry, it's very slight chance of this happening.

closes #31187

image


plugins/scaffolder-backend/report.api.md
```diff
@@ -610,12 +624,12 @@ export class TaskWorker {
   stop(): Promise<void>;
 }
 
-// @public @deprecated
-export class TemplateActionRegistry {
+// @public @deprecated (undocumented)
+export interface TemplateActionRegistry {
```
benjdlambert, line 628: This change we can't make as it's breaking.

This is what I was trying to avoid before in relation to the comment.

If we can rework the change to leave this interface intact and create a new one that maybe isn't even exported, but is still used in the the underlying default implementations I think that makes more sense.

Basically a distributed actions class which takes the existing built-in modules in the constructor, and then provides async API's for listing would be fine I think.

I wonder if it makes sense to just remove the `.get` method entirely, and have the `list` return a `Map<string, TemplateAction>` and the consumption pattern just becomes `list().get('asd')` instead.

That way we don't have to deal with caching or refreshing the distributed ones.

drodil: Yeah, I understand. It's a bit pity it's a public interface (even though it cannot be used by anyone really, as there is no extension point to override this). The issue with `list()`, especially when chained with `get()`, is that it would always get the distributed actions. This basically means `[number of plugins providing actions] x [scaffolder template steps]` amount of requests for every template run, which can be a huge overhead. I think there must be some way to cache/refresh the distributed actions anyway.

benjdlambert: Yep +1 to the unfortunateness. It really should have been removed part of the other major change, but we needed to deprecate it first before removing, and was in a bit of a rush to remove the old interfaces.

It's really only going to use `pluginSources` in config, so it's going to make requests to each of those plugins. Which means that it scales individually of the amount of backend plugins you actually have.

Thinking that we can skip the caching for now, and wait for it to be come an issue really. The requests are pretty lightweight as they're basically just hitting an in-memory cache of the plugin list anyways.

Wdyt?

drodil: Sounds good, but there's still an issue with the `TaskWorker` and `NunjucksWorkflowRunner` interfaces being public, and creating a new class for distributed actions means we have to change those APIs anyways. I think there's no way around this without breaking some of the public APIs.

drodil: Made a new PR #31213 with minimum changes to existing APIs, WDYT?

drodil:
A new try in a new PR

PR #31213, opened 2025-09-22
fix(scaffolder): distributed actions to be available

### LINKED ISSUE #31187: Actions registry actions not visible in scaffolder

drodil:

depending on the plugin startup order, some distributed actions might not be available when the scaffolder backend starts.
this was mentioned in the comment that is now removed. while this might change in the future, this would patch the current way the distributed actions are fetched.

closes #31187


plugins/scaffolder-backend/src/scaffolder/actions/DefaultDistributedActionRegistry.ts
```diff
@@ -0,0 +1,93 @@
...
+    const builtinActions = this.registry.list();
+    for (const action of builtinActions) {
+      if (ret.has(action.id)) {
+        throw new Error(`Duplicate action id '${action.id}' found`);
+      }
+      ret.set(action.id, action);
+    }
+
+    const { actions } = await this.actionsRegistry.list({
+      credentials: await this.auth.getOwnServiceCredentials(),
```
drodil: I wonder should this actually be the user's credentials at some point if the permissions for actions are added?

plugins/scaffolder-backend/report.api.md
```diff
@@ -395,6 +396,14 @@ export type DatabaseTaskStoreOptions = {
   events?: EventsService;
 };
 
+// @public
+export interface DistributedActionRegistry {
```
benjdlambert: Can we just inline this type for now? Rather not export something new if possible, as it's just going to go away and perhaps broken once we remove the rest.

plugins/scaffolder-backend/src/scaffolder/actions/DefaultDistributedActionRegistry.ts
```diff
@@ -0,0 +1,99 @@
...
+        schema: {
+          input: action.schema.input as Schema,
+          output: action.schema.output as Schema,
+        },
+      }),
+    );
+
+    for (const action of distributedActions) {
+      if (ret.has(action.id)) {
+        throw new Error(`Duplicate action id '${action.id}' found`);
```
benjdlambert: Not sure we want to throw an error here, otherwise wouldn't this break the entire scaffolder if someone was to register a plugin with the same ID in a another plugin?

drodil: But distributed action ids are `${pluginId}:${actionId}` so it should be plugin scoped, right?

plugins/scaffolder-backend/report.api.md
drodil: Yeah we could just use the Default one

drodil: The issue is that `TaskWorker` is a public interface, and if we keep only the `DefaultDistributedActionRegistry`, it will fail to build the API reports, as it doesn't know anything about that, as it's not exported.

benjdlambert: Just duplicate the type in the TaskWorker definition.

```tsx
type CreateWorkerOptions = {
  distributedActions?: { list(options?: {} } // inline
}
```

I just don't want another type in the public API for this that's a separate export.

plugins/scaffolder-backend/src/scaffolder/actions/DefaultDistributedActionRegistry.ts
benjdlambert: It's not impossible that they would conflict though, and it's unfortunate if I created a `bitbucket:pull-request` action locally that then crashes the startup of the scaffolder when someone installs a bitbucket plugin.

I think that the behavior should be:

- Register all locally installed modules
- Register all remote distributed actions
- On conflict of remote actions, log a warning that the distributed action will not be called as the internal module overrides it.

This way this also gives end users the ability to override distributed actions with their own local version of it for the scaffolder perhaps by using the same name.

drodil: Is there a check that same plugin cannot register two actions with same id? Otherwise I think this should be fine like this.

benjdlambert: Yes there is, but I don't think that's the only requirement: 

https://github.com/backstage/backstage/blob/36417fbe266884c6b9d1e0a9bcdad34ed0394905/packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.ts#L160-L162

I can do `createBackendModule` that uses the scaffolder extension point to register an action with the `id: catalog:blah`.

The namespace is only controlled at the distributed actions end in the `ActionsRegistry`.

The local modules are freeform, so I think the logic has to be as above.

drodil: Fix'd

benjdlambert: Sorry @drodil.

This PR has motivated me to just go and do this: https://github.com/backstage/backstage/pull/31235

I'm getting tired of things being so hard to evolve because we can't break interfaces that are public.

Let's wait for that to be merged, and then just go back to the original design making `TemplateActionsRegistry` a promise based flow instead.

Should cleanup a lot of this code.

benjdlambert: @drodil do you use the `TaskBroker` or `TaskContext` types and the extension points at all?

drodil: @benjdlambert no we don't. Only `scaffolderActionsExtensionPoint`, `scaffolderTemplatingExtensionPoint`, and `scaffolderAutocompleteExtensionPoint`

drodil:
I think this PR can be closed and refactor now that the old interfaces are gone

Issue #30147, opened 2025-06-04
Prefer use of project.id over path_with_namespace when calling GitLab repository file API


Pivouane:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
GitLab API, path_with_namespace, project.id, GitlabDiscoveryEntityProvider, repository/files, Backstage catalog, file lookup, GitLab integration
```

### 🗃️ Project Area

Catalog

### 🔖 Need

Using project.id avoids edge cases caused by path encoding or project structure changes.

This benefits:

    Users working with nested GitLab groups or non-standard characters in project names

    Probably Backstage maintainers, by reducing the likelihood of obscure bugs or support issues

    API stability in the face of future GitLab changes

It also simplifies reasoning about the GitLab API usage, since IDs are immutable, while paths are not.

### 📝 Proposal

In the GitlabDiscoveryEntityProvider, the current implementation calls client.hasFile(...) using project.path_with_namespace to identify the GitLab project:

```ts
const hasFile = await client.hasFile(
  project.path_with_namespace ?? '',
  project_branch,
  this.config.catalogFile,
);
```

GitLab’s official API documentation for GET /projects/:id/repository/files/:file_path states that the :id path parameter can be the project ID or URL-encoded path of the project. However, using the ID can avoid encoding issues and ambiguity.

Update the implementation to use project.id instead of path_with_namespace when calling client.hasFile(...), for greater reliability.

```ts
const hasFile = await client.hasFile(
  project.id,
  project_branch,
  this.config.catalogFile,
);

```
This change simplifies the code and improves robustness, particularly in environments with:

    deeply nested group structures,

    special characters in project paths,

    renamed or moved projects.

### 🔄 Alternatives

# Continue using path_with_namespace

This is the current behavior. It works in many common scenarios but has known edge cases.

# Hybrid approach (support both id and path_with_namespace)

The code could be modified to support both project.id and project.path_with_namespace, possibly falling back from one to the other or allowing the user to configure a preference.

Example logic:

const identifier = preferId ? project.id : project.path_with_namespace;
const hasFile = await client.hasFile(
  identifier,
  project_branch,
  this.config.catalogFile,
);

However maintaining support for both approaches may increase complexity without a clear benefit, unless there is a strong backward compatibility need.

### Have you read the Code of Conduct?

[...]

PR #31217, opened 2025-09-22
GitlabDiscoverEntityProvider: update shouldProcessProject to use project.id for hasFile check

clairep94:

Closes https://github.com/backstage/backstage/issues/30147
Update `shouldProcessProject` to use `project.id` for the `hasFile` check

### DIFF
--- a/plugins/catalog-backend-module-gitlab/src/__testUtils__/handlers.ts
+++ b/plugins/catalog-backend-module-gitlab/src/__testUtils__/handlers.ts
@@ -268,8 +268,7 @@ const httpProjectFindByIdDynamic = all_projects_response.map(project => {
 });
 
 /**
- * Checks for both project id and namespaced path, as these can both be used for the :id segment in Gitlab API:
- * https://docs.gitlab.com/api/repository_files/#get-file-from-repository
+ * See https://docs.gitlab.com/api/repository_files/#get-file-from-repository
  */
 const httpProjectCatalogDynamic = all_projects_response.flatMap(project => {
   return rest.head(


.changeset/twelve-oranges-grin.md
```diff
@@ -0,0 +1,20 @@
+---
+'@backstage/plugin-catalog-backend-module-gitlab': patch
+---
+
+Update GitlabDiscoveryEntityProvider to use `project.id` rather than `project.path_with_namespace` for `hasFile` check in `shouldProcessProject`
+
+Solves [#30147](https://github.com/backstage/backstage/issues/30147):
+
+> Use of project.id avoids edgecases caused by path encoding or project structure changes
```
freben:
```suggestion
> Use of project.id avoids edge cases caused by path encoding or project structure changes
```

Note the spell checker complaining about unknown words (you can see it in the Files Changed tab only) - for technical identifiers they can be enclosed in backticks to silence the checker, or if they really should be allowed words they can be added to `.github/vale/config/vocabularies/Backstage/accept.txt` as part of the PR. In this case I opted to just add a space, but it's a matter of taste.

.changeset/twelve-oranges-grin.md
```diff
@@ -0,0 +1,20 @@
...
+> It also simplifies reasoning about the GitLab API usage, since IDs are immutable, while paths are not.
+
+```diff
+const hasFile = await client.hasFile(
+-    project.path_with_namespace,
++    project.id.toString(),
+    project_branch,
+    this.config.catalogFile,
+);
+```
```
freben: If you like, you can remove the deeper technical info from the changeset since its intended audience is a casual reader that just wants brief actionable summary.

The actual change of "the how" is already covered in the code itself below, and the issue with sufficient details about "the why" is already linked.

Changesets are great for informing a reader about things that THEY need to do though where applicable, for example if there is a breaking change of behaviour or config so that they need to take action with something while upgrading to this version of the package.

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
```diff
@@ -590,7 +590,7 @@ export class GitlabDiscoveryEntityProvider implements EntityProvider {
       this.config.fallbackBranch;
 
     const hasFile = await client.hasFile(
-      project.path_with_namespace ?? '',
+      project.id.toString(),
```
freben: DId we conclude if the toString was needed or not?

I also see another usage of this function [here](https://github.com/backstage/backstage/blob/498332ae5348bcabdbb63f0e4dccf565466e82fa/plugins/catalog-backend-module-gitlab/src/GitLabDiscoveryProcessor.ts#L141). Should that be updated to use the ID as well?

.changeset/twelve-oranges-grin.md
clairep94: done!

.changeset/twelve-oranges-grin.md
clairep94: done!

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
clairep94: nice catch, updated!

[it seems `id` will be a number](https://docs.gitlab.com/api/repository_files/#get-file-from-repository), but I opted to update `hasFile` `projectIdentifier` param to the below, bc it's getting plugged into a string interpol later anyway, what do you think?

```
sync hasFile(
    projectIdentifier: string | number,
    branch: string,
    filePath: string,
  ): Promise<boolean> {
    const endpoint: string = `/projects/${encodeURIComponent(
      projectIdentifier,
    )}/repository/files/${encodeURIComponent(filePath)}`;
    const request = new URL(`${this.config.apiBaseUrl}${endpoint}`);
    request.searchParams.append('ref', branch);
	
```

.changeset/twelve-oranges-grin.md
```diff
@@ -0,0 +1,11 @@
...
+'@backstage/plugin-catalog-backend-module-gitlab': patch
+---
+
+Update GitlabDiscoveryEntityProvider to use `project.id` rather than `project.path_with_namespace` for `hasFile` check in `shouldProcessProject`
+
+Solves [#30147](https://github.com/backstage/backstage/issues/30147):
+
+> Use of project.id avoids edge cases caused by path encoding or project structure changes
+> [...]
+> It also simplifies reasoning about the GitLab API usage, since IDs are immutable, while paths are not.
```
vinzscam:
```suggestion
Fixed an issue in `GitlabDiscoveryEntityProvider` where entity fetching could fail for projects with special characters or that had been renamed or moved.
```

freben: Yep, that is a good summary!

plugins/catalog-backend-module-gitlab/src/__testUtils__/handlers.ts
```diff
@@ -266,24 +266,34 @@ const httpProjectFindByIdDynamic = all_projects_response.map(project => {
...
-        return res(ctx.status(200));
-      }
-      return res(ctx.status(404, 'Not Found'));
-    },
+/**
+ * Checks for both project id and namespaced path, as these can both be used for the :id segment in Gitlab API:
+ * https://docs.gitlab.com/api/repository_files/#get-file-from-repository
+ */
+const httpProjectCatalogDynamic = all_projects_response.flatMap(project => {
+  const possibleIdSegments: string[] = [
```
freben: Do we need to check both? If we have an ID, can that ever NOT be perfectly right? Maybe I misunderstand. but it seems to me that they should be effectively equivalent and both requests should give identical results always below.

plugins/catalog-backend-module-gitlab/src/__testUtils__/handlers.ts
```diff
@@ -266,13 +266,14 @@ const httpProjectFindByIdDynamic = all_projects_response.map(project => {
...
 });
-const httpProjectCatalogDynamic = all_projects_response.map(project => {
-  const path: string = project.path_with_namespace
-    ? project.path_with_namespace!.replace(/\//g, '%2F')
-    : `${project.path_with_namespace}%2F${project.name}`;
 
+/**
+ * Checks for both project id and namespaced path, as these can both be used for the :id segment in Gitlab API:
+ * https://docs.gitlab.com/api/repository_files/#get-file-from-repository
+ */
```
freben, line 272:
```suggestion
/**
 * See https://docs.gitlab.com/api/repository_files/#get-file-from-repository
 */
```

freben (approved):
Nice! The comment is no longer true, so we can cut it down, but then let's merge.

PR #31231, opened 2025-09-23
fix(catalog): config for priority and enabled

drodil:

Moved catalog processor and provider disabling and priorities under own config objects.

This is due to issue with some existing providers, such as GitHub, using array syntax for the provider configuration.

The new config format is not backwards compatible, so users will need to update their config files. The new format
is as follows:

```yaml
catalog:
  providerOptions:
    providerA:
      disabled: false
    providerB:
      disabled: true
  processorOptions:
    processorA:
      disabled: false
      priority: 10
    processorB:
      disabled: true
```

### DIFF
--- a/.changeset/fast-heads-brake.md
+++ b/.changeset/fast-heads-brake.md
@@ -1,0 +1,25 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Moved catalog processor and provider disabling and priorities under own config objects.
+
+This is due to issue with some existing providers, such as GitHub, using array syntax for the provider configuration.
+
+The new config format is not backwards compatible, so users will need to update their config files. The new format
+is as follows:
+
+```yaml
+catalog:
+  providerOptions:
+    providerA:
+      disabled: false
+    providerB:
+      disabled: true
+  processorOptions:
+    processorA:
+      disabled: false
+      priority: 10
+    processorB:
+      disabled: true
+```
--- a/docs/features/software-catalog/configuration.md
+++ b/docs/features/software-catalog/configuration.md
@@ -90,6 +90,35 @@ also write a _custom processor_ to convert between the existing system and
 Backstage's descriptor format. This is documented in
 [External Integrations](external-integrations.md).
 
+### Processor configuration
+
+You can configure processors under the `catalog.processorOptions` key. You can define
+options for each processor by its name. With the processor options you can disable
+a processor or set its priority.
+
+```yaml
+catalog:
+  processorOptions:
+    processorName:
+      disabled: false # Defaults to false
+      priority: 100
+```
+
+The priority is a number defining the order in which processors are executed. The lower the number,
+the higher the priority. The default priority is `20`.
+
+## Provider configuration
+
+It's possible to also disable entity providers using the config. You can configure providers
+under the `catalog.providerOptions` key. The key is the provider name.
+
+```yaml
+catalog:
+  providerOptions:
+    providerName:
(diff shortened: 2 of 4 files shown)


.changeset/fast-heads-brake.md
```diff
@@ -0,0 +1,25 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Moved catalog processor and provider disabling and priorities under own config objects.
+
+This is due to issue with some existing providers, such as GitHub, using array syntax for the provider configuration.
+
+The new config format is not backwards compatible, so users will need to update their config files. The new format
+is as follows:
```
Rugvip, line 10: The new config was part of the most recent release right? Fine to treat this as a non-breaking patch fix as you are, as long as we do a patch release for it.

.changeset/fast-heads-brake.md
```diff
@@ -0,0 +1,25 @@
...
+
+```yaml
+catalog:
+  enable:
+    processors:
+      processorA: true
+      processorB: false
```
Rugvip: Hmm, I find that `<key>: false` is a bit of a smell, usually you'd want `<key>: true`, with `<key>: false` being the same as the default.

.changeset/fast-heads-brake.md
```diff
@@ -0,0 +1,25 @@
...
+Moved catalog processor and provider disabling and priorities under own config objects.
+
+This is due to issue with some existing providers, such as GitHub, using array syntax for the provider configuration.
+
+The new config format is not backwards compatible, so users will need to update their config files. The new format
+is as follows:
+
+```yaml
+catalog:
+  enable:
```
Rugvip: Thoughts on `providerOptions` as a top-level key for this instead? or something similar. Same for processors ofc

freben: So:

```yaml
catalog:
  providerOptions:
    providerA:
      disabled: true
  processorOptions:
    processorA:
      disabled: true
      order: 10
```

.changeset/fast-heads-brake.md
drodil: Yes it was in 1.43.0 so probably not very used.

.changeset/fast-heads-brake.md
drodil: Changed to `disabled` flag instead `enabled`

.changeset/fast-heads-brake.md
drodil: Yep, done, except that `order` is still `priority` as there is the `getPriority` method in the `CatalogProcessor` interface not to mix things up.

freben: Right, that's just my mistake.

Rugvip:
Not waiting for the full microsite build 😁

benjdlambert:
@drodil we found a bit of an issue with the fact that some of the provider names are not valid config keys:

https://github.com/backstage/backstage/blob/eb80a1ad5ae071277c1f284c7f616a6804f328b5/plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts#L196

leads to this:

https://github.com/backstage/backstage/blob/9e76ddb8f9149edd3a73b998bad38bb44f200d56/packages/config/src/reader.ts#L21

https://github.com/backstage/backstage/blob/9e76ddb8f9149edd3a73b998bad38bb44f200d56/packages/config/src/reader.ts#L451-L454

Throwing some errors, working on a different fix for now.

PR #31235, opened 2025-09-23
`scaffolder`: Removing old deprecated types and cleanup

benjdlambert:
Breaking change here, but hopefully won't be affecting anyone as these types that remain are of very little value to consumers as they can't be used for anything with the scaffolder plugin.

This is part of a bigger re-architecting of the scaffolder backend that we're planning over the coming months and into 2026.

The next part to be removed is the `/alpha` export for the `TaskBroker` interface, but leaving that around for now.

### DIFF
--- a/.changeset/legal-eagles-jog.md
+++ b/.changeset/legal-eagles-jog.md
@@ -1,0 +1,18 @@
+---
+'@backstage/plugin-scaffolder-backend-module-confluence-to-markdown': patch
+'@backstage/plugin-scaffolder-backend-module-bitbucket-server': patch
+'@backstage/plugin-scaffolder-backend-module-bitbucket-cloud': patch
+'@backstage/plugin-scaffolder-backend-module-notifications': patch
+'@backstage/plugin-scaffolder-backend-module-cookiecutter': patch
+'@backstage/plugin-scaffolder-backend-module-bitbucket': patch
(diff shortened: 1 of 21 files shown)


drodil:
Holy moly, that's a lot of changes 🫣😅

.changeset/ready-poems-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-node': minor
```
Rugvip, line 2: Let's move/copy all extension points to the main export as well, they're currently all in `/alpha`

freben: Oh good catch!

benjdlambert: Hmm, there's probably really only one I would wanna move at this point though, which would be the `templateActions` one.

The other ones are:

- `templateExtensions` for filters and globals, maybe could move this?
- `TaskBroker` this is going away so no point
- `autocomplete` maybe this is also fine to move?
- `workspaceProvider` experimental feature so fine to leave there until things are settled?

benjdlambert: OK, decided just to move the `templateActions` one for now, as I'm not 100% confident in the API's of the rest, or that they're alpha features.

punkle:
@benjdlambert could you link me to info about the re-architecture?

PR #31238, opened 2025-09-23
Add PasswordField to BUI

birdhb:

- Add `PasswordField` to BUI
- Create stylesheet and types
- Add `PasswordField` stories

[Watch Loom Here](https://www.loom.com/share/fae918cd24774721a6624f0930cdd7cb?sid=8e84f0dd-67eb-4883-b919-a2abee8a467b)


packages/ui/src/utils/componentDefinitions.ts
```diff
@@ -286,6 +286,20 @@ export const componentDefinitions = {
...
+      disabled: [true, false] as const,
+    },
+  },
+  PasswordField: {
+    classNames: {
+      root: 'bui-PasswordField',
+      inputWrapper: 'bui-InputWrapper',
+      input: 'bui-Input',
+      inputIcon: 'bui-InputIcon',
+      inputAction: 'bui-InputAction',
```
cdedreuille: These class names should be unique. You don't need to repeat them. The best way would be to use the class names from `TextField` directly. We do the same for `SearchField` I believe.

packages/ui/src/components/TextField/types.ts
```diff
@@ -23,6 +23,22 @@ import type { FieldLabelProps } from '../FieldLabel/types';
...
+   */
+  type?:
+    | 'text'
+    | 'email'
+    | 'tel'
+    | 'url'
+    /** @deprecated Use `SearchField` instead */
+    | 'search'
+    /** @deprecated Use `PasswordField` instead */
+    | 'password';
```
cdedreuille: I would just remove them for now instead of deprecating them

cdedreuille (changes requested):
Thanks a lot @birdhb 🙏 Here are a couple of feedbacks:
- Since we are removing some types, this is considered a breaking change so I would move the changeset to a `minor`
- The show/hide button is touching the top and bottom part. I'm not sure we can actually use our `ButtonIcon` for that. I would suggest to create a custom button for that.

CleanShot 2025-09-25 at 12 43 18@2x

- Since the show/hide button is floating, it is not picked up by external Chrome extensions like 1Password. I think we need to revisit that to make sure the 1Password button fits nicely next to the button. Chakra is doing that as well for inspiration - https://www.chakra-ui.com/docs/components/password-input

CleanShot 2025-09-25 at 12 43 28@2x

- It would be great to add these aria labels on the button to make it fully accessible - https://stackoverflow.com/questions/61270823/accessibility-show-hide-password-button-in-a-password-input

.changeset/eighty-cows-care.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': minor
+---
+
+**Breaking** We are adding a new component `PasswordField`. We will be removing support for types `password` and `search` on `TextField`.
```
Rugvip:
```suggestion
**BREAKING**: Added a new `PasswordField` component. As part of this change, the `password` and `search` types have been removed from `TextField`.
```

Generally we keep changesets in third-person + some minor style tweak

Rugvip (approved):
Awesome! 😁 

Just a changeset nit

PR #31239, opened 2025-09-23
Revert "Patch release of #31231"

benjdlambert:
Reverts backstage/backstage#31237


drodil:
See #31243. Another option would be to remove the config validation completely as @freben suggested. And `CatalogBuilder` really needs some unit tests.

drodil:
Another option: #31250

benjdlambert:
Just gonna force push the patch branch back to the previous release instead

PR #31240, opened 2025-09-23
Revert "fix(catalog): config for priority and enabled"

benjdlambert:
Reverts backstage/backstage#31231

See https://github.com/backstage/backstage/pull/31231#issuecomment-3324906666


freben:
Double checking with @drodil maybe making a fix already?

drodil:
@freben not sure what would be the proper fix for this. Only thing that comes to my mind at this hour is to replace all `:` characters in the processor/provider name with `.` but could it fail in some other way then 🤔

drodil:
Made #31243 to help with this

PR #31247, opened 2025-09-23
chore(notifications): force more strict api contract

drodil:

the old contract with entity recipients has been too loose and can cause unwanted effects. this change introduces a new way to pass notification recipients and the old way should be removed after migration period.

tests and docs have been updated to use the new recipient structure but there might be some users out there who will need to update their calls to notification service before removal.

### DIFF
--- a/plugins/notifications-backend-module-email/src/processor/NotificationsEmailProcessor.test.ts
+++ b/plugins/notifications-backend-module-email/src/processor/NotificationsEmailProcessor.test.ts
@@ -246,6 +246,39 @@ describe('NotificationsEmailProcessor', () => {
     });
   });
 
+  it('should send user email with old recipients', async () => {
+    (createTransport as jest.Mock).mockReturnValue(mockTransport);
+    const processor = new NotificationsEmailProcessor(
+      logger,
+      mockServices.rootConfig({ data: DEFAULT_SENDMAIL_CONFIG }),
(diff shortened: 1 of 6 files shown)


.changeset/legal-crews-admire.md
```diff
@@ -0,0 +1,32 @@
+---
+'@backstage/plugin-notifications-backend': minor
+'@backstage/plugin-notifications-node': minor
```
freben, line 3:
```suggestion
'@backstage/plugin-notifications-backend': patch
'@backstage/plugin-notifications-node': patch
```

this isn't really breaking yet is it?

plugins/notifications-node/src/service/DefaultNotificationService.ts
```diff
@@ -27,15 +27,33 @@ export type NotificationServiceOptions = {
...
   | {
+      type: 'entities';
+      /**
+       * Entity references to send the notifications to
+       */
+      entityRefs: string[];
+      /**
+       * Optional entity reference(s) to filter out of the resolved recipients.
+       * Usually the currently logged-in user for preventing sending notification
+       * of user action to him/herself.
```
freben, line 38:
```suggestion
       * of a user action to themselves.
```

freben (commented):
Again, did this really make the interface any more robust?

PR #31253, opened 2025-09-24
feat(scaffolder): add missing form fields for the nfs

drodil:

Adds missing scaffolder fields to NFS

### DIFF
--- a/plugins/scaffolder/src/components/fields/EntityNamePicker/validation.ts
+++ b/plugins/scaffolder/src/components/fields/EntityNamePicker/validation.ts
@@ -17,9 +17,6 @@
 import { FieldValidation } from '@rjsf/utils';
 import { KubernetesValidatorFunctions } from '@backstage/catalog-model';
 
-/**
- * @public
- */
 export const entityNamePickerValidation = (
   value: string,
   validation: FieldValidation,
--- a/plugins/scaffolder/src/components/fields/EntityNamePicker/validation.ts
+++ b/plugins/scaffolder/src/components/fields/EntityNamePicker/validation.ts
@@ -17,6 +17,9 @@
 import { FieldValidation } from '@rjsf/utils';
 import { KubernetesValidatorFunctions } from '@backstage/catalog-model';
 
+/**
+ * @public
+ */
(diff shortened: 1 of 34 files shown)


plugins/scaffolder/src/components/fields/EntityNamePicker/validation.ts
```diff
@@ -17,6 +17,9 @@
 import { FieldValidation } from '@rjsf/utils';
 import { KubernetesValidatorFunctions } from '@backstage/catalog-model';
 
+/**
+ * @public
```
benjdlambert: I don't think that these need to be public right? Same as the schemas?

drodil: Well, the existing ones were so followed the pattern. Maybe some clean up necessary here as well?

drodil:
Interestingly, `yarn build:api-reports plugins/scaffolder` gives a different api report than building all with `yarn build:api-reports` 🤔

freben:
@drodil yeah that's a known quirk unfortunately - the full reports are what always should take precedence when this happens.

plugins/scaffolder/src/components/fields/EntityNamePicker/validation.ts
drodil: Cleaned them all off; not completely sure if there was a reason they were public in the first place?

benjdlambert: Sorry, to clarify. I don't think we should add anything /new/ to the exports for the public API just to add the field extensions. I think we need to keep the existing ones in there for now, but shouldn't need to export anything else just to export the `FormFieldExtensions` from `/alpha`.

drodil: No idea why the `allowedHosts` in the report just changed the place 🌵

benjdlambert (approved):

benjdlambert:
Perfect! thank you! :pray:

drodil:
Can we get a ship here?

Issue #31118, opened 2025-09-11
NFS: Need documented guidance on standard for testing migrated plugins


hopehadfield:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🗃️ Project Area

Documentation

### 🔗 External Integration

N/A

### 📝 Description

When migrating plugins to Backstage’s new frontend system, there is currently no clear, documented direction for how those plugins should be tested in development.

Right now, it seems like the standard is to introduce an app-next package to support testing, though I'm not sure how that applies to workspaces that don’t currently have an app package.

It's my understanding that there are efforts to avoid adding full Backstage apps to every workspace, and instead use the `dev/index` pattern exclusively for many plugins. If this is still the expected practice, then an issue presents itself in testing the legacy frontend once modifying `dev/index` to test the new frontend instead.

That being said, what is the proposed method of testing plugins that currently use the `dev/index` pattern for testing?

Not sure if this guidance belongs in the official documentation for the [New Frontend System](https://backstage.io/docs/frontend-system/) or in the community-plugins guide for [local development of plugins](https://github.com/backstage/community-plugins/blob/main/CONTRIBUTING.md#local-development-of-plugins), but clarifying this would give contributors a more clear path to enable smooth plugin migrations.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes, but I would like some more guidance

benjdlambert:
@camilaibs I know you've been looking at this? 

Do you want to whip up some docs around what you found is a good dev app experience for NFS? 

I wonder if we can add docs about having support of both that switches depending on an ENV variable or flag or something?

camilaibs:
> [@camilaibs](https://github.com/camilaibs) I know you've been looking at this?
> 
> Do you want to whip up some docs around what you found is a good dev app experience for NFS?
> 
> I wonder if we can add docs about having support of both that switches depending on an ENV variable or flag or something?

[...]

PR #31255, opened 2025-09-24
[CLI] Package Start Entrypoint Option

camilaibs:

Fixes https://github.com/backstage/backstage/issues/31118

Added a new `--entrypoint` option to the `package start` command, which allows you to specify a custom entry directory/file for development applications. This is particularly useful when maintaining separate dev apps for different versions of your plugin (e.g., stable and alpha).

**Example usage:**

Consider the following plugin dev folder structure:

```
dev/
  index.tsx
  alpha/
    index.ts
```

- The default `yarn package start` command uses the `dev/` folder as the entry point and executes `dev/index.tsx` file;
- Running `yarn package start --entrypoint dev/alpha` will instead use `dev/alpha/` as the entry point and execute `dev/alpha/index.ts` file.

### DIFF
--- a/packages/cli/src/modules/build/commands/package/start/startPackage.ts
+++ b/packages/cli/src/modules/build/commands/package/start/startPackage.ts
@@ -20,6 +20,7 @@ import { startFrontend } from './startFrontend';
 
 export async function startPackage(options: {
   role: PackageRole;
+  entryDir: string;
   targetDir: string;
   configPaths: string[];
   checksEnabled: boolean;
@@ -45,8 +46,8 @@ export async function startPackage(options: {
     case 'frontend-plugin':
     case 'frontend-plugin-module':
       return startFrontend({
-        entry: 'dev/index',
         ...options,
+        entry: `${options.entryDir ?? 'dev'}/index`,
       });
     case 'frontend-dynamic-container' as PackageRole: // experimental
       return startFrontend({
--- a/packages/cli/src/modules/build/commands/package/start/startPackage.ts
+++ b/packages/cli/src/modules/build/commands/package/start/startPackage.ts
@@ -17,21 +17,19 @@
 import { PackageRole } from '@backstage/cli-node';
 import { startBackend, startBackendPlugin } from './startBackend';
 import { startFrontend } from './startFrontend';
-import { statSync, readdirSync } from 'fs';
-import { resolve, join, parse } from 'path';
+import { parse, resolve, join } from 'path';
+import { glob } from 'glob';
 
 export function resolveEntryPath(
   entrypoint: string = 'dev',
   targetDir: string,
 ): string {
   const { dir: entryDir, name: entryName } = parse(entrypoint);
-  const entryPath = join(entryDir, entryName);
-  const parentPath = resolve(targetDir, entryDir);
-  const isFile = readdirSync(parentPath).some(base => {
-    const path = resolve(parentPath, base);
-    return statSync(path).isFile();
-  });
-  return isFile ? entryPath : join(entryPath, 'index');
+  const [entryFile] = glob.sync(`${resolve(targetDir, entryDir, entryName)}.*`);
+  if (entryFile) {
+    return join(entryDir, entryName);
+  }
+  return join(entryDir, entryName, 'index');
 }
 
 export async function startPackage(options: {
--- a/packages/cli/src/modules/build/commands/package/start/startPackage.ts
+++ b/packages/cli/src/modules/build/commands/package/start/startPackage.ts
@@ -17,10 +17,26 @@
 import { PackageRole } from '@backstage/cli-node';
 import { startBackend, startBackendPlugin } from './startBackend';
 import { startFrontend } from './startFrontend';
+import { statSync, readdirSync } from 'fs';
+import { resolve, join, parse } from 'path';
+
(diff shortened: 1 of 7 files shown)


benjdlambert:
I wonder if it's better to encourage `/alpha.ts` and just have `--entrypoint /dev/alpha` instead of `dir`. That way you can either have `/alpha/index.ts` or `/alpha.ts` as options. 

Unless you think there's reasons why a directory would always be best? :pray:

freben:
I like `--entrypoint` too, but where you can point to basically anything and it "just works". The code detects whether it's a file or a folder, and looks for relevant files under the dir if that's what it is

* `--entrypoint dev` -> finds and uses `dev/index.ts` or `dev/index.tsx`
* `--entrypoint dev/index` -> finds and uses `dev/index.ts` or `dev/index.tsx`
* `--entrypoint dev/index.tsx` -> also works of course

So when someone makes a `dev-new-frontend-system` folder next to `dev` and they run `--entrypoint dev-new-frontend-system` it delights them that it works and does the expected thing :)

camilaibs:
> I wonder if it's better to encourage `/alpha.ts` and just have `--entrypoint /dev/alpha` instead of `dir`. That way you can either have `/alpha/index.ts` or `/alpha.ts` as options.
> 
> Unless you think there's reasons why a directory would always be best? 🙏

@freben / @benjdlambert Updated to `--entrypoint` 🎉 , PTAL.

~I believe the folder helps us group and isolate files that are only used by the alpha application, such as the app config and API mocks in this example:~

~image~

~The entry point you suggested would also allow this. The only thing that confuses me is that for the `--entrypoint` flag I would expect a path to a file instead of a folder path:~

- ~`--entrypoint dev/alpha.ts` instead of `dev/alpha`~
- ~`--entrypoint dev/alpha/index.ts` instead of `dev/alpha`~

packages/cli/src/modules/build/index.ts
```diff
@@ -137,6 +137,10 @@ export const buildPlugin = createCliPlugin({
             '--link <path>',
             'Link an external workspace for module resolution',
           )
+          .option(
+            '--entrypoint <path>',
+            'Entry directory path (uses index file) or entry file path (without extension). Defaults to "dev"',
```
freben:
```suggestion
            'The entrypoint to start from, relative to the package root. Can point to either a file (without extension) or a directory (in which case the index file in that directory is used). Defaults to "dev"',
```

just a suggested tweak to wording.

The "without extension" is again something that just stands out to me. Why would it not work to specify an extension? :) Especially if the command happens to be constructed by a shell script or something, would that need to take the effort of removing the extension then?

packages/cli/src/modules/build/commands/package/start/startPackage.ts
```diff
@@ -17,9 +17,26 @@
...
+import { statSync, readdirSync } from 'fs';
+import { resolve, join, parse } from 'path';
+
+export function resolveEntryPath(
+  entrypoint: string = 'dev',
+  targetDir: string,
+): string {
+  const { dir: entryDir, name: entryName } = parse(entrypoint);
+  const entryPath = join(entryDir, entryName);
+  const parentPath = resolve(targetDir, entryDir);
```
freben: Hm. Could all of the code above be using `paths.resolveTarget` instead [like for example here](https://github.com/backstage/backstage/blob/80f7b646a570b6537fabb6157fd8b37ea16b3081/packages/cli/src/modules/maintenance/commands/package/clean.ts#L21)? Then when you get the result of that, you can start doing stat and stuff on that target.

packages/cli/src/modules/build/commands/package/start/startPackage.ts
```diff
@@ -17,9 +17,26 @@
...
+  const isFile = readdirSync(parentPath).some(base => {
```
freben: Hm I think you could start with stat on the exact full path and if that's a directory, perform one set of logic and otherwise perform another set of logic 🤔 Right? Haha I just feel uneasy with readdir on what could be NOT a directory to begin with, even if it works in practice. Also if you flip the logic you don't need a parent dir variable at all, I think.

packages/cli/src/modules/build/index.ts
camilaibs: If I understand correctly, there are a set of supported extensions, and they're added [here](https://github.com/backstage/backstage/blob/1f88dc369e7de107c8ba662d3eab81871839b4dc/packages/cli/src/modules/build/lib/bundler/paths.ts#L30). I don't have enough context to explain why they are added later, so I tried not to change the implementation too much.
In theory the extension can be passed in the flag, but at the beginning of the `resolveEntryPath` function I'll ignore it.

packages/cli/src/modules/build/commands/package/start/startPackage.ts
camilaibs: I simplified the logic, please tell me what you think 🙂

packages/cli/src/modules/build/commands/package/start/startPackage.ts
camilaibs: Maybe it is not necessary after simplifying the logic, tell me what you think.

benjdlambert (approved):

Issue #30521, opened 2025-07-14
publish:gitlab action fails with timeout after successful push on GitLab Enterprise


Aluush:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
publish:gitlab timeout
GitLab Enterprise push success but action fails"
simple-get/index.js request timed out
backstage template publish gitlab enterprise
```

### 🗃️ Project Area

Software Templates

### 🔗 External Integration

GitLab

### 📝 Description & Context

I'm encountering an issue while generating a template using the publish:gitlab action.

When using our GitLab Enterprise instance, the repository is created successfully, and the code is pushed, but the action still fails with the following error:

<img width="2084" height="994" alt="Image" src="https://github.com/user-attachments/assets/a812a72b-8ffa-45f6-8fc0-7bf69daed14e" />

Interestingly, everything works fine when using gitlab.com, so the problem seems specific to the Enterprise GitLab instance.
template.yaml:

```bash
apiVersion: scaffolder.backstage.io/v1beta3
# https://backstage.io/docs/features/software-catalog/descriptor-format#kind-template
kind: Template
metadata:
  name: Java-template
  title: Java Template
  description: An example template for the scaffolder that creates a simple Java service
  annotations:
    backstage.io/techdocs-ref: dir:.
spec:
  owner: user:aliatattou
  type: service

  # These parameters are used to generate the input form in the frontend, and are
  # used to gather input data for the execution of the template.
  parameters:
    - title: Fill in some steps
      required:
        - name
      properties:
        name:
          title: Name
          type: string
          description: Unique name of the component
          ui:autofocus: true
          ui:options:
            rows: 5
    - title: Choose a location
      required:
        - repoUrl
      properties:
        repoUrl:
          title: Repository Location
          type: string
          ui:field: RepoUrlPicker
          ui:options:
            allowedHosts:
              - gitlab.my.company.com
            requestUserCredentials:
              secretsKey: USER_OAUTH_TOKEN
  # These steps are executed in the scaffolder backend, using data that we gathered
  # via the parameters above.
  steps:
    # Each step executes an action, in this case one templates files into the working directory.
    - id: fetch-base
      name: Fetch Base
      action: fetch:cookiecutter
      input:
        url: ./content
        values:
          project_name: ${{ parameters.name }}
[...]

PR #31258, opened 2025-09-24
feat(scaffolder-gitlab): add hybrid git implementation with useCustomGit option

allamand:
This PR introduces an optional custom git implementation for the GitLab scaffolder to address HTTP timeout issues while maintaining backward compatibility.

Trying to fix : #30521

### Changes
- **New custom git implementation**: Added `customGitPush` function using native git commands to bypass HTTP client timeout issues
- **Opt-in parameter**: Added `useCustomGit` parameter (defaults to `false` for backward compatibility)
- **Enhanced error handling**: Improved repository creation error messages with detailed HTTP status codes and response context
- **Better authentication**: Uses 'oauth2' as username for GitLab token authentication following standard practices
- **Improved error reporting**: Enhanced `printGitlabError` function with more detailed error information

### Benefits
- Resolves timeout issues for large repositories or slow network connections
- Maintains existing behavior by default
- Provides clear error messages for troubleshooting
- Allows users to opt-in when experiencing issues

### Backward Compatibility
The original `initRepoAndPush` remains the default behavior, ensuring existing users are unaffected while providing an escape hatch for timeout scenarios.

### DIFF
--- a/plugins/scaffolder-backend-module-gitlab/src/actions/gitlab.examples.test.ts
+++ b/plugins/scaffolder-backend-module-gitlab/src/actions/gitlab.examples.test.ts
@@ -86,6 +86,12 @@ describe('publish:gitlab', () => {
 
   beforeEach(() => {
     jest.resetAllMocks();
+    // Reconfigure the mock after reset
+    (
+      initRepoAndPush as jest.MockedFunction<typeof initRepoAndPush>
+    ).mockResolvedValue({
+      commitHash: '220f19cc36b551763d157f1b5e4a4b446165dbd6',
+    });
   });
 
   it('should call initRepoAndPush with the correct values', async () => {
--- a/plugins/scaffolder-backend-module-gitlab/src/actions/gitlab.test.ts
+++ b/plugins/scaffolder-backend-module-gitlab/src/actions/gitlab.test.ts
@@ -15,14 +15,13 @@
  */
 
 jest.mock('@backstage/plugin-scaffolder-node', () => {
+  const mockResult = {
+    commitHash: '220f19cc36b551763d157f1b5e4a4b446165dbd6',
+  };
   return {
     ...jest.requireActual('@backstage/plugin-scaffolder-node'),
-    initRepoAndPush: jest.fn().mockResolvedValue({
-      commitHash: '220f19cc36b551763d157f1b5e4a4b446165dbd6',
-    }),
-    commitAndPushRepo: jest.fn().mockResolvedValue({
-      commitHash: '220f19cc36b551763d157f1b5e4a4b446165dbd6',
-    }),
+    initRepoAndPush: jest.fn().mockResolvedValue(mockResult),
+    commitAndPushRepo: jest.fn().mockResolvedValue(mockResult),
   };
 });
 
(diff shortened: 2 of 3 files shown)


allamand:
Hi, Can I have point of view on this ? for us it was the only way to avoid thoe timeouts, and this fix worked perfectly in our use case. wonder how to progress onthis.

benjdlambert:
hey @allamand :wave:

Sorry it's taken a while to get round to this - just wanna ask a few clarifying questions!

What's the cause of the timeouts here? Is there some HTTP proxy or something that's not being respected by `isomorphic-git` over the `git` binary that might be already configured? Are they consistent? Does increasing the timeout not help here?

Overall I would really like to avoid shelling out to tools in the environment if we can avoid it, or maybe moving this to a separate action / module entirely might be a better option?

allamand:
Hi @benjdlambert ,

I tried to understand why there was a timeout, but was never able to find out why. Backstage and gitlab I'm using are deployed into same Eks cluster so no proxy, nothing strange, but I was never able to make it work.

For now this setup is working for my use case that was the most important part, but I would definitely prefer to now work from mainline. 

I agree this PR may be not the more elegant solution and I can test any suggestion you may have.

elaine-mattos:
Hi @allamand ,

This is an interesting issue. My organization has been using a similar setup for about two years (a GitLab instance running on an EKS cluster and Backstage on another EKS cluster, in separate AWS accounts). We also rely on templates that use the `publish:gitlab action` and have not encountered issues beyond the OAuth token considerations you address in this PR.

I am also not a fan of passing shell commands inside code, it's great for debugging but I don't really think it's safe  for prod. I guess it would also mean that the container image running Backstage would have to have git installed ... :(

I'm not sure it helps but if you can't find the root cause right now, wouldn't it be interesting to create a custom scaffolder action, using the `isomorphic-git`library? We had other use cases and created one scaffolder action for each git operation: add, clear, clone, push, createBranch, etc.


ps.: have you tried to set the `settings.visibility: internal` in your template? I am asking because the default is `private` and depending on your integration token's scopes, you could have issues...
This is a complete example:
```yarn
    - id: publish
      name: Publish Gitlab repository
      action: publish:gitlab
      input:
        gitCommitMessage: "feat:  initial commit [ci skip]"
        gitAuthorName: Developer Portal
        gitAuthorEmail: developerportal@company.com
        settings:
          visibility: internal
        defaultBranch: main
        repoUrl: githost.com?owner=${{ steps['get-user'].output.entity.metadata.name }}&repo=yourepo-${{ parameters.namespace }}
        token: ${{ secrets.USER_OAUTH_TOKEN }}
        branches:
          - name: main
            protect: true
```

allamand:
Hi @elaine-mattos thanks do this, I'll look into it to see if I can use that and improve the workaround

benjdlambert:
@allamand did @elaine-mattos suggestion help here? :pray:

allamand:
> @allamand did @elaine-mattos suggestion help here? 🙏

HI @benjdlambert I didn't had the time to validate this, i'll log an action item so that I can test is quickly. thanks

PR #31259, opened 2025-09-24
break out provider/processor handling into a separate file

freben:
This carefully does the sensitive config read where we need to be defensive about unsafe characters. It also breaks out the functions for this into a separate file so we can add tests.

Piggybacks on top of the changeset in https://github.com/backstage/backstage/pull/31231, so that's intentionally left out

### DIFF
--- a/plugins/catalog-backend/src/service/util.ts
+++ b/plugins/catalog-backend/src/service/util.ts
@@ -194,7 +194,7 @@ export function filterAndSortProcessors(
     }
   }
 
-  function getProcessorDisabled(processor: CatalogProcessor): boolean {
+  function isProcessorDisabled(processor: CatalogProcessor): boolean {
     return (
       getProcessorOptions(processor)?.getOptionalBoolean('disabled') === true
     );
@@ -216,7 +216,7 @@ export function filterAndSortProcessors(
   }
 
   return processors
-    .filter(p => !getProcessorDisabled(p))
+    .filter(p => !isProcessorDisabled(p))
     .sort((a, b) => getProcessorPriority(a) - getProcessorPriority(b));
 }
 
@@ -240,11 +240,11 @@ export function filterProviders(
     }
   }
 
-  function getProviderDisabled(provider: EntityProvider): boolean {
+  function isProviderDisabled(provider: EntityProvider): boolean {
     return (
       getProviderOptions(provider)?.getOptionalBoolean('disabled') === true
     );
   }
 
-  return providers.filter(p => !getProviderDisabled(p));
+  return providers.filter(p => !isProviderDisabled(p));
 }


plugins/catalog-backend/src/service/util.ts
```diff
@@ -166,3 +171,80 @@ export function expandLegacyCompoundRelationsInEntity(entity: Entity): Entity {
...
+      return root?.getOptionalConfig(processor.getProcessorName());
+    } catch {
+      // We silence errors specifically here, to cover for cases where the
+      // processor name contains special characters which makes the config
+      // reader throw an error.
+      return undefined;
+    }
+  }
+
+  function getProcessorDisabled(processor: CatalogProcessor): boolean {
```
drodil: Maybe `isProcessorDisabled`?

plugins/catalog-backend/src/service/util.ts
```diff
@@ -166,3 +171,80 @@ export function expandLegacyCompoundRelationsInEntity(entity: Entity): Entity {
...
+  config: Config,
+): CatalogProcessor[] {
+  function getProcessorOptions(
+    processor: CatalogProcessor,
+  ): Config | undefined {
+    const root = config.getOptionalConfig('catalog.processorOptions');
+    try {
+      return root?.getOptionalConfig(processor.getProcessorName());
+    } catch {
+      // We silence errors specifically here, to cover for cases where the
```
drodil, line 190: I still think it would be beneficial to log these, just for future fixes with the config loading 🤔

plugins/catalog-backend/src/service/util.ts
```diff
@@ -166,3 +171,80 @@ export function expandLegacyCompoundRelationsInEntity(entity: Entity): Entity {
...
+
+  function getProcessorDisabled(processor: CatalogProcessor): boolean {
+    return (
+      getProcessorOptions(processor)?.getOptionalBoolean('disabled') === true
+    );
+  }
+
+  function getProcessorPriority(processor: CatalogProcessor): number {
+    let priority =
+      getProcessorOptions(processor)?.getOptionalNumber('priority');
```
drodil, line 205: Could we get an error from this in any case?

plugins/catalog-backend/src/service/util.ts
```diff
@@ -166,3 +171,80 @@ export function expandLegacyCompoundRelationsInEntity(entity: Entity): Entity {
...
+      return root?.getOptionalConfig(provider.getProviderName());
+    } catch {
+      // We silence errors specifically here, to cover for cases where the
+      // provider name contains special characters which makes the config
+      // reader throw an error.
+      return undefined;
+    }
+  }
+
+  function getProviderDisabled(provider: EntityProvider): boolean {
```
drodil: `isProviderDisabled`?

plugins/catalog-backend/src/service/util.ts
```diff
@@ -166,3 +171,80 @@ export function expandLegacyCompoundRelationsInEntity(entity: Entity): Entity {
...
+ * through the `catalog.providerOptions` config.
+ */
+export function filterProviders(
+  providers: EntityProvider[],
+  config: Config,
+): EntityProvider[] {
+  function getProviderOptions(provider: EntityProvider): Config | undefined {
+    const root = config.getOptionalConfig('catalog.providerOptions');
+    try {
+      return root?.getOptionalConfig(provider.getProviderName());
```
drodil, line 234: Can we already use 
```ts
const [_, config] ?= root?.getOptionalConfig(provider.getProviderName());
return config;
```
?

plugins/catalog-backend/src/service/util.ts
freben: This intentionally isn't inside the try/catch; I want to propagate all _other_ errors than the unintended one, from the config system.

plugins/catalog-backend/src/service/util.ts
freben: I ... don't like how that reads :D

plugins/catalog-backend/src/service/util.ts
freben: I thought about it, but it kind of turns out unhelpful for the users who see those logs.

Remember, we don't log when they have already entered something in config that we can't parse - we log for exclusively internal reasons, whether the config is in place or not, just by the very act of doing `getOptional` or `has`. So we would warn about them not being able in a future hypothetical scenario to enter a certain config value. It's not actionable. And there's currently no way around it as far as I know.

plugins/catalog-backend/src/service/util.ts
freben: updated!

plugins/catalog-backend/src/service/util.ts
freben: updated!

plugins/catalog-backend/src/service/util.ts
drodil: I on the other land love this and the `??=` :D But I think it requires some newer ECMAScript support to use these

drodil (approved):

PR #31270, opened 2025-09-25
feat: allow custom ReactElement as title in EntityTable

tklever:

Expanded the EntityTable `title` prop definition to allow `ReactElement`s to be passed as the `title` prop. This allows for complete customization of the title display in EntityTable components.

### DIFF
--- a/.changeset/young-doodles-repeat.md
+++ b/.changeset/young-doodles-repeat.md
@@ -3,4 +3,4 @@
 '@backstage/plugin-catalog-react': minor
 ---
 
-Allow ReactElement type for `title` property of `EntityTable`
+Allow `ReactElement` type for `title` property of `EntityTable`


packages/core-components/src/components/Table/Table.tsx
```diff
@@ -481,16 +482,20 @@ export function Table<T extends object = {}>(props: TableProps<T>) {
...
+          isValidElement(title) ? (
+            title
+          ) : (
+            <>
+              <Typography variant="h5" component="h2">
+                {title}
               </Typography>
-            )}
-          </>
+              {subtitle && (
```
vinzscam, line 492: I think we should keep `subtitle` otherwise it would never be rendered

.changeset/young-doodles-repeat.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/core-components': minor
```
freben:
```suggestion
'@backstage/core-components': patch
```

because it's nonbreaking on a 0.x version

.changeset/young-doodles-repeat.md
```diff
@@ -0,0 +1,6 @@
...
+'@backstage/plugin-catalog-react': minor
+---
+
+Allow ReactElement type for `title` property of `EntityTable`
```
freben:
```suggestion
Allow `ReactElement` type for `title` property of `EntityTable`
```

packages/core-components/src/components/Table/Table.tsx
```diff
@@ -481,16 +482,20 @@ export function Table<T extends object = {}>(props: TableProps<T>) {
...
         title={
-          <>
-            <Typography variant="h5" component="h2">
-              {title}
-            </Typography>
-            {subtitle && (
-              <Typography color="textSecondary" variant="body1">
-                {subtitle}
+          isValidElement(title) ? (
+            title
```
freben, line 486: Would it be better to keep the wrapping Typography? It serves as a good set of defaults. What was your reason for wanting this feature in the first place? Did you for example want to add some icon or so in the text? For the majority of such use cases, it feels like the user mostly wants to change the _contents_ of the title, not the styling. If we keep the Typography, the user will benefit from staying consistent with our future tweaks to these default styles

packages/core-components/src/components/Table/Table.tsx
tklever: `subtitle` is rendered under the following conditions:
- `title` is a `string`
AND
- `subtitle` is truthy

`subtitle` is NOT rendered under the following conditions:
- `title` is a `ReactElement`
OR
- `subtitle` is falsy

is this ok?

packages/core-components/src/components/Table/Table.test.tsx
```diff
@@ -209,6 +209,24 @@ describe('<Table />', () => {
     expect(rendered.getByText('subtitle')).toBeInTheDocument();
```
tklever, line 209: Here's the assertion for when subtitle renders in the unit tests..

packages/core-components/src/components/Table/Table.tsx
freben: Why would it be predicated on the title being a string?

tklever: > Why would it be predicated on the title being a string?

title is currently typed as a string, so this preserves existing functionality. If we allow a `ReactElement` to be passed, the user is assuming full control of the header, they can configure the subtitle (or sub-sub-title, or super-title, etc.) in the `ReactElement` they pass.

If the title is a string, the assumption is we're using the default recommended rendering behavior. If it's a ReactElement, we're assuming you've made a choice to manage the render yourself completely.

packages/core-components/src/components/Table/Table.tsx
tklever: Obviously, it's your call as a maintainer, but I'd argue against leaving the wrapping Typography. 

I think we already have some nice sensible defaults in this component, and I assume sending through a string `title` and optionally a `subtitle` work for the majority of users. If those defaults DON'T work for some users, this allows them to assume full control, take on the maintenance responsibility and send through a `ReactElement` where they're free to do anything, but know, you're taking on the responsibility of things like Typography, etc.

This set up leaves us with:
- Sensible Default - string title
- Full Control - ReactElement title

If we leave the wrapping Typography, we're left with 
- Sensible Default - string title
- Slightly Less Sensible Default - only wrapping Typography
- No full control option

I think invariably, we'll eventually circle back to full control for someone wanting to customize the header completely, and we end up with a compromised middle option that isn't really the recommended default, and doesn't really satisfy those that want complete control.

>  If we keep the Typography, the user will benefit from staying consistent with our future tweaks to these default styles

I think we're ok here. If we pass a string and use the defaults, the styling will stay consistent. If someone wants full control, they take on the burden of maintenance, no different than forking or providing alternative versions of UI components like `AboutCard` or `AboutContent`. When you make your own, it comes with cost, that's a choice that we make elsewhere as well. 

> What was your reason for wanting this feature in the first place?

Specific use case was adding table tools to the header that would alter a fetched data set, similar to the Direct<->Aggregate toggle for `EntityRelationAggregation`, but more generally getting access to use the whitespace on the right side of the table header and throw things into that space.

tklever:
Rebased against `master` and corrected DCO sign off issues.

packages/core-components/src/components/Table/Table.tsx
freben, line 492: Sorry I'll be bullish on this one. There's no need to remove the subtitle rendering just because you for example wanted a `<Fragment data-testid="for my test">` around the title.

If they don't want a subtitle, they don't supply a subtitle. It's easy for the end user to accommodate. The end result as rendered will be the same as in your suggestion, but better because of the added flexibility.

tklever: I don't think you're being bullish, but I do think I'm doing a poor job of communicating. That's on me, not you.

Let me try one last time before I resolve to say "I failed to make the case, closed".

You'll excuse me for being overly descriptive of things you probably already know, I just want to make sure I summarize everything as best I can

- The `Table` component wraps `MTable` from `@material-table/core`.
- `MTable` it typed with a `title` prop that accepts a string or an element `title?: string | React.ReactElement<any>;`
- `Table` is typed with a `title` prop that accepts a string or an element `title?: string | React.ReactElement<any>;`

When a `title` prop is passed,`Table` wraps the `title` prop in a `<Typography variant="h5" component="h2" />` before forwarding it on to `MTable` (additionally appending an optional subtitle).

I don't currently have a way to "jailbreak" out of that `Typography` element. If I pass a ReactElement as the `title` prop to `Table`, be it a `<span />` or a complex series of flexboxes, it's going to be wrapped with `<Typography />`.

I see the `Typography` wrapper as obviously useful default formatting when the user is passing a `string` title, but maybe a little over-prescriptive (and easily replicable) when a user passes a `ReactElement`. If you're passing a `ReactElement`, and you want the `Typography` wrapper, add it yourself. Additionally, if you wanted a different `Typography` variant or element, you now have that option, in addition to any other element structure you dream up

The consequence of this choice is the "default formatting" now only works when the title is a `string`. Which, correctly means `subTitle` is ignored when you pass a `ReactElement` instead of two strings.  

Accepting this PR means the user can eliminate or replace the `Typography` wrapper around the `title` if they choose, with the trade off of needing to append their own `subTitle` should they wish.

Rejecting this PR means the user can not eliminate or replace the `Typography` wrapper around the `title` if they choose, that formatting choice is static and immovable (without forking off a whole new Table component)

If we're in agreement on the choice, and you still choose to reject, hopefully I at least articulated myself well. :-)

tklever: Would a solution of optionally appending the subtitle, if it's passed, to the supplied `ReactElement` `title` prop be acceptable? 

That behavior seems like it might cause some interesting outcomes, but if that works, I could add that to this.

PR #31275, opened 2025-09-25
[NFS] Providing the default `searchFilters` and `resultTypes` as extensions of the search plugin

Andy2003:
This commit refactors the Backstage search plugin architecture to make search filters and result types extensible through a plugin system.
 Instead of hardcoding filters like entity kind, lifecycle, and TechDocs filters directly in the search page component, they are now defined as separate extensions that are registered with the plugin.

1. The `SearchFilterBlueprint` is enhanced with a new `typeFilter` function that determines when a filter should be visible based on the selected result types, plus a configuration schema that allows specifying which result types a filter applies to.
2. The search page now dynamically renders only the relevant filters based on the current search context.
3. Default result types (Software Catalog and Documentation) and common filters (entity kind, lifecycle, and TechDocs entity filters) are provided as built-in extensions but can now be customized or replaced.

This change makes the search functionality more modular and allows other plugins to extend search capabilities without modifying the core search plugin code.

### DIFF
--- a/.changeset/sweet-ads-teach.md
+++ b/.changeset/sweet-ads-teach.md
@@ -1,0 +1,6 @@
+---
+'@backstage/plugin-search-react': minor
+'@backstage/plugin-search': minor
+---
+
+Providing the default `searchFilters` and `resultTypes` as extensions of the search plugin
--- a/plugins/search-react/report-alpha.api.md
+++ b/plugins/search-react/report-alpha.api.md
@@ -24,17 +24,23 @@ export const SearchFilterBlueprint: ExtensionBlueprint<{
   output: ExtensionDataRef<
     {
       component: SearchFilterExtensionComponent;
+      typeFilter?: (types: string[]) => boolean;
     },
     'search.filters.filter',
     {}
   >;
   inputs: {};
-  config: {};
-  configInput: {};
+  config: {
+    types: string[] | undefined;
+  };
+  configInput: {
+    types?: string[] | undefined;
+  };
   dataRefs: {
     searchFilters: ConfigurableExtensionDataRef<
       {
         component: SearchFilterExtensionComponent;
+        typeFilter?: (types: string[]) => boolean;
       },
       'search.filters.filter',
       {}
@@ -46,6 +52,7 @@ export const SearchFilterBlueprint: ExtensionBlueprint<{
 export interface SearchFilterBlueprintParams {
   // (undocumented)
   component: SearchFilterExtensionComponent;
+  typeFilter?: (types: string[]) => boolean;
 }
 
 // @alpha (undocumented)
--- a/plugins/search-react/src/alpha/blueprints/SearchFilterBlueprint.test.tsx
+++ b/plugins/search-react/src/alpha/blueprints/SearchFilterBlueprint.test.tsx
@@ -42,7 +42,23 @@ describe('SearchFilterBlueprint', () => {
           "id": "page:search",
           "input": "searchFilters",
         },
-        "configSchema": undefined,
+        "configSchema": {
+          "parse": [Function],
+          "schema": {
+            "$schema": "http://json-schema.org/draft-07/schema#",
+            "additionalProperties": false,
+            "properties": {
+              "types": {
+                "description": "A list of result types where this search filter should be shown for",
+                "items": {
+                  "type": "string",
+                },
+                "type": "array",
+              },
+            },
+            "type": "object",
+          },
+        },
         "disabled": false,
         "factory": [Function],
         "inputs": {},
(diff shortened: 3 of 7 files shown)


Andy2003:
@freben / @benjdlambert I tagged you, since there was nobody auto-assigned to this PR

Andy2003:
up

freben:
Hey, thanks for your PR. I just wanted to say that we actually did get our eyes on this one.

One thing I'm personally asking myself is, what is the right level of granularity of the app tree? As in, how many types of thing should be extensions in the first place, and which ones should be modeled in other ways (eg using utility apis, or as props given to something else, etc).

So we'll have to talk a bit in the team. And in the mean time, the below is just a little bit of a brain dump to communicate what I'm personally thinking about in regards to the PR. Don't take it as actionable feedback; it's just me thinking out loud to set the stage a bit and see if you have any preferences that might be good feedback for us.

So.

### Parent filtering, or child bailout?

Making it possible for filters to have a selector for which types to appear on, is a reasonable desire. It COULD theoretically have been implemented entirely inside the React layer though, if one so desired. We could make it so that all filters are always available, but some of them choose to render themselves as a `null` under some circumstances and contribute nothing to the filterset (instead of the surrounding component making that decision). [The `CatalogFilterBlueprint`](https://github.com/backstage/backstage/blob/28d4e2218d0ae1312d148e9b39688f2c0d050687/plugins/catalog-react/src/alpha/blueprints/CatalogFilterBlueprint.ts#L27) does not have this filtering ability for example, while [some filters just bail out](https://github.com/backstage/backstage/blob/28d4e2218d0ae1312d148e9b39688f2c0d050687/plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx#L175). I'm not saying that this is the better approach - and of course the card and page blueprints [do take your approach](https://github.com/backstage/backstage/blob/28d4e2218d0ae1312d148e9b39688f2c0d050687/plugins/catalog-react/src/alpha/blueprints/EntityCardBlueprint.ts#L48)!

Is there a point when more/deeper NFS modelling just adds abstractions that don't help?

### What's a good filter?

Assuming that we go for parent filtering:

In the runtime world, returning a [filter function output](https://github.com/backstage/backstage/blob/28d4e2218d0ae1312d148e9b39688f2c0d050687/plugins/catalog-react/src/alpha/blueprints/EntityCardBlueprint.ts#L43) is fine. It'll always be contextual what the input for that function is, so we'd have different such filter functions for each context.

But on the config side of things, the design directly affects end users and we need to have even stricter stability guarantee so that we don't need to "break" people's config files in future releases.

So for the config side of filtering, do we want to go for some form of consistency? The entities go for [the new predicate expressions](https://github.com/backstage/backstage/blob/28d4e2218d0ae1312d148e9b39688f2c0d050687/plugins/catalog-react/src/alpha/blueprints/EntityCardBlueprint.ts#L55), as [defined here](https://backstage.io/docs/features/software-catalog/catalog-customization#entity-predicate-queries). The reason that this format was made was, whatever bespoke format we came up with, users always needed something different or more complex. That's why the [string filter form](https://github.com/backstage/backstage/blob/28d4e2218d0ae1312d148e9b39688f2c0d050687/plugins/catalog-react/src/alpha/blueprints/EntityCardBlueprint.ts#L44) probably is taking a back seat and might get deprecated - unless we invent an entire string form "mini programming language" that is just as expressive as the predicates - and should we?

What I'm saying is, should the config format for these filters, if we want it, be more like

```yaml
  filter:
    'result.type': { $in: ['a', 'b'] }
```

and then we settle on a well defined such context object shape that happens to have a `{result: {type:string}}` in it?

### "Theming" using extensions

I'm torn on the `search-filter-result-type` kind. I see that you're leaning into the NFS paradigm and that's cool, but I wonder if it's the right thing to do or if they should be modeled differently. I see that it's neat that people can contribute more result types both from third party plugins and their private extensions. But how does this interact with translations, where do those strings live and what format do they have?

And also, could this have been one single `SwappableComponent` instead? And we make the framework in such a way that we lean into "special outputs" like the routing ones. So any plugin can emit a "search plugin filter result type output" or several, and those just get sleuthed up throughout the app tree for use by that swappable component.

```ts
...({
  output: [searchExtensionData.resultType],
  *factory() {
    yield searchExtensionData.resultType({ ... })
```

... or even consider mechanisms similar to the extension points in the backend system, where there's an opportunity for "early registration" of things before the framework completes its startup phase?

```ts
...({
  deps: {
    searchResultTypes: searchResultTypeRegistryRef, 
  },
  *factory({...}, { deps }) {
    deps.searchResultTypes.addResultType({ ... });
  }
```

### Foreign registrations

In the PR, the receiving plugin defines a lot of things that technically belong to other plugins. The catalog plugin should arguably be registering catalog search results, even if it's within the set of "core" plugins. How do we want to think about these foreign registrations going forward?

### Summary

So those are some thoughts that came to mind. Sorry to present this much insecurity over such a reasonably sized PR. All I'm saying is, I think the team should talk through some of these things before giving clearer direction here :) I hope this came across as more informative and transparent, than confusing and frustrating. If you have any preferences or thoughts as an adopter of this system, do let me hear them, and I'll communicate that to the team.

Andy2003:
any updates on this?

alexvoedi:
are there any updates for this?

Andy2003:
Up

alexvoedi:
up

alexvoedi:
up

Issue #30066, opened 2025-05-27
Support long-lasting sessions issued by auth-backend


Rugvip:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
"auth-backend", "long-lasting sessions", "offline access"
```

### 🗃️ Project Area

Auth

### 🔖 Need

There are a couple new types of clients that we want to add support for in Backstage. Both [CLI clients that want to authenticate end users towards local Backstage deployments](https://github.com/backstage/backstage/issues/15820), as well as [MCP clients that want to do the same](https://github.com/backstage/backstage/issues/29349). A good starting point for this is adding support for device-auth through https://github.com/backstage/backstage/pull/27680, but that only takes us as far as being able to hand over a short-lived session to the client. Ideally we'd also be able to issue credentials that lasts for a longer (configurable) amount of time. That way you can go through the log-in flow once to authenticate an CLI or MCP server, and use that session for an extended period of time without needing to re-auth.

### 📝 Proposal

We implement a persistent session store in `auth-backend` that is able to issue long-lived refresh tokens for users. The sessions would need to contain the original claims that were issued for the users, and it might therefore be best to simply connect this to the existing user info store that already exists in the auth-backend.

The session store would be tied to long-lived tokens that expire after a configurable amount of time. The tokens themselves are refresh tokens and can not be used as access tokens. In order to make use of the token the client must exchange the token for the usual short-lived access token that is then used for subsequent requests.

In addition to the above, we may also want to consider extending the session store to also support storing information for auth providers. This would allow us to replace the existing pattern of re-using refresh tokens from OAuth providers and instead store them using a single Backstage session instead. This in turn can also allow for offline access delegation and fixing some refresh flows that invalidate the existing token.

### 🔄 Alternatives

Instead of issuing refresh tokens that are tied to a session store entry, we could instead issue regular access tokens with a longer expiration time instead. This has two problems: It's more insecure as it does not allow for any current of future method of revoking tokens without complex architectural additions. The other issue is that we'd need to store and publish the keys that tokens were signed with for a lot longer, potentially leading to excessively long lists that eventually end up hurting performance.

[...]

PR #31278, opened 2025-09-25
feat(auth): allow configuring DCR token expiration

drodil:

this adds a new config value for exprimental dynamic client registration feature that allows configuring the token expiration.

added also missing config values to the config schema for this feature.

### DIFF
--- a/plugins/auth-backend/config.d.ts
+++ b/plugins/auth-backend/config.d.ts
@@ -95,6 +95,7 @@ export interface Config {
 
     /**
      * The backstage token expiration.
+     * Defaults to 1 hour (3600s). Maximum allowed is 24 hours.
      */
     backstageTokenExpiration?: HumanDuration | string;
 
@@ -102,5 +103,28 @@ export interface Config {
      * Additional app origins to allow for authenticating
      */
     experimentalExtraAllowedOrigins?: string[];
+
+    /**
+     * Configuration for dynamic client registration
+     */
+    experimentalDynamicClientRegistration?: {
+      /**
+       * Whether to enable dynamic client registration
+       * Defaults to false
+       */
+      enabled?: boolean;
+
+      /**
+       * A list of allowed URI patterns to use for redirect URIs during
+       * dynamic client registration. Defaults to '[*]' which allows any redirect URI.
+       */
+      allowedRedirectUriPatterns?: string[];
+
+      /**
+       * The expiration time for the client registration access tokens.
+       * Defaults to 1 hour (3600s). Maximum allowed is 24 hours.
+       */
+      tokenExpiration?: HumanDuration | string;
+    };
   };
 }
--- a/plugins/auth-backend/src/service/OidcService.ts
+++ b/plugins/auth-backend/src/service/OidcService.ts
@@ -17,8 +17,8 @@ import { AuthService, RootConfigService } from '@backstage/backend-plugin-api';
 import { TokenIssuer } from '../identity/types';
 import { UserInfoDatabase } from '../database/UserInfoDatabase';
 import {
-  InputError,
   AuthenticationError,
+  InputError,
   NotFoundError,
 } from '@backstage/errors';
 import { decodeJwt } from 'jose';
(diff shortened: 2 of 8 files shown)


plugins/auth-backend/src/service/OidcService.ts
```diff
@@ -403,7 +404,7 @@ export class OidcService {
     return {
       accessToken: token,
       tokenType: 'Bearer',
-      expiresIn: 3600,
+      expiresIn: expiresIn ?? 3600,
```
benjdlambert: So there's a slight issue here.

As you might have noticed above, the original `readTokenExpiration` call is used to set how long the tokens that are issued from the `TokenIssuer` are actually valid for as they have also `exp` claims inside the token.

The issue with only changing this value and increasing from the default is that the token would not be valid unless you updated both.

I guess we have a few options here, we either create a new `TokenIssuer` for this which feels a little bit unfortunate, we refactor the `TokenIssuer` interface to take an `expiry` when calling `issueToken` or we just use the original value from config in this expiry block, and don't treat them as two different config values at all?

Wdyt?

Also gonna bring in @Rugvip for some other opinions here too.

benjdlambert (changes requested):

plugins/auth-backend/src/service/OidcService.ts
Rugvip: Wouldn't the existing config work for this? 🤔 https://github.com/backstage/backstage/blob/e99fadfbad9501f7937941f0972a60693caa82d4/plugins/auth-backend/config.d.ts#L99

Rugvip: If we want it to be separate, then I guess we'd need to wire up a new option for `issueToken` like you said

drodil: I think using just the existing would be the easiest way forward?

benjdlambert: Using the existing config seems fine for me, was just trying to understand if we did want to keep this entirely separate. If we wanted DCR to last 2 weeks for example, but normal expiration to still be 1hr then that wouldn't really be possible.

drodil: Yeah, that might be the issue. I think DCR should support more long-living tokens

drodil: Decided to refactor this by creating a new `TokenIssuer` for the OIDC and supporting token expiration up to 14 days for DCR tokens. I bet this would be the right thing in the long run, WDYT?

drodil:
image
interesting failures 🤔

plugins/auth-backend/config.d.ts
```diff
@@ -102,5 +102,28 @@ export interface Config {
...
+
+      /**
+       * A list of allowed URI patterns to use for redirect URIs during
+       * dynamic client registration. Defaults to '[*]' which allows any redirect URI.
+       */
+      allowedRedirectUriPatterns?: string[];
+
+      /**
+       * The expiration time for the client registration access tokens.
+       * Defaults to 1 hour (3600s). Maximum allowed is 14 days.
```
Rugvip: Hmm, I think this is a bit too long tbh. Best stick to our current limit of 8h. These are cryptographically signed with no ability to revoke old tokens. Having a large expiration time also has the downside of forcing us to keep all the certificates that have been used to sign tokens around, which I think with this kind of expiration actually might impact performance.

In the end, the path forward for supporting long-term sessions is #30066.

drodil: Good point, I will make the maximum 8 hours!

drodil: Except that the current maximum is actually 24 hours `const TOKEN_EXP_MAX_S = 86400;`

drodil: Changed to 24 hours instead, which makes more sense than 8 hours actually :)

Rugvip: Ah yep, remembered wrong, 24h is good 👍

benjdlambert (approved):
nice! let's give this a go :tada:

PR #31280, opened 2025-09-25
Log before provider-orphaning eviction happens

freben:
To make it more clear why it happens.

Also add more tests that use the real database.


Rugvip (approved):
Yep! 👍

PR #31281, opened 2025-09-25
fix(bui): remove default selection for tabs that have href defined

ssjoblad:

**Current behavior:**
If tab selection logic returns `undefined`, the first tab is auto-selected. This causes mismatches when the current page isn’t part of the tab set (the first tab gets highlighted even though its content doesn’t match).

**Proposed fix:**
If all tabs have href, return null → prevents auto-select, avoids mismatched tab states.
Otherwise, keep the current fallback (auto-select first tab).

### DIFF
--- a/.changeset/thin-hoops-bathe.md
+++ b/.changeset/thin-hoops-bathe.md
@@ -1,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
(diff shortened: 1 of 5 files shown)


cdedreuille (approved):
LGTM!

PR #31301, opened 2025-09-29
cli: remove jest script module cache

Rugvip:

The script cache seems to be broken by Jest 30. Trying it out in CI to see what the performance difference looks like nowadays.

### DIFF
--- a/.changeset/full-chefs-roll.md
+++ b/.changeset/full-chefs-roll.md
@@ -2,4 +2,4 @@
 '@backstage/cli': patch
 ---
 
-Removed the script transform cache from the default Jest configuration. The script cache provided a moderate performance boost, but it is incomatible with Jest 30.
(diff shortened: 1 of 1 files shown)


benjdlambert:
guess this is a non starter? 😏

Rugvip:
@benjdlambert not particularly, but I think it adds a minute or two to full test runs. The impact is lower for cached runs because what this cache aims to do to begin with is to cache across projects/packages.

.changeset/full-chefs-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+Removed the script transform cache from the default Jest configuration. The script cache provided a moderate performance boost, but it is incomatible with Jest 30.
```
freben:
```suggestion
Removed the script transform cache from the default Jest configuration. The script cache provided a moderate performance boost, but it is incompatible with Jest 30.
```

Issue #31063, opened 2025-09-04
Add support for YAML merging


valentingregoire:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
`YAML`, `merge`, `parseAllDocuments`, `options`
```

### 🗃️ Project Area

Catalog

### 🔖 Need

Adding support for merging YAML documents with the `<<:` operator is a huge benefit for reusing pieces in YAML files. Especially when using reusable placeholders like `$yaml`. This makes templates more flexible as now anchors can't be unpacked.

### 📝 Proposal

Original code: https://github.com/backstage/backstage/blob/4f83da1f7d2b2924e5c6a491c673291d9af8121a/plugins/catalog-backend/src/processors/PlaceholderProcessor.ts#L143

Add an optional extra parameter for the options one can provide to the `parseAllDocuments` method. For now, let's just hard-code this parameter in:

```diff js
- documents = yaml.parseAllDocuments(content).filter(d => d);
+ documents = yaml.parseAllDocuments(content, { merge: true }).filter(d => d);
```

One could also foresee an extra parameter to be provided to the `yamlPlaceholderResolver` method where this snippet above resides. This way, one could potentially shape the `YAML` processing to their needs. More about this in the official [docs](https://eemeli.org/yaml/#options).

### 🔄 Alternatives

Writing a custom placeholder resolver, but that is far-fetched probably.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

No, but I'm happy to collaborate on a PR with someone else

freben:
That does sound like a worthwhile default to add. Opening up for contribs

Dharma-09:
Hi, I’d like to pick this up.

valentingregoire:
> That does sound like a worthwhile default to add. Opening up for contribs

I have little Backstage knowledge and how contributing works. In this case, would that be just adding the `parse=true` parameter hard-coded, or would it have to be configurable and adapt the unit tests therefore?

freben:
See above - a pull request seems to already have been submitted!

PR #31303, opened 2025-09-29
Enable YAML merge keys in yamlPlaceholderResolver

Dharma-09:

This PR updates the `yamlPlaceholderResolver` in `PlaceholderProcessor.ts` to support YAML merge keys (`<<`) and anchors by enabling the `{ merge: true }` option in `yaml.parseAllDocuments`.

Issue: https://github.com/backstage/backstage/issues/31063

### 💡 Suggestion

I’d like to make the YAML merge support **configurable** rather than always enabled.  
By default, it stays `true` for backward compatibility, but users can choose to disable it if their use case requires stricter parsing.  

Proposed approach:
```ts
const parseOptions = { merge: params?.options?.enableYamlMerge ?? true };
documents = yaml.parseAllDocuments(content, parseOptions).filter(d => d);
```

### DIFF
--- a/.changeset/twelve-spoons-feel.md
+++ b/.changeset/twelve-spoons-feel.md
@@ -1,6 +1,6 @@
 ---
-'@backstage/plugin-catalog-backend': minar
-'@backstage/plugin-catalog-node': minar
+'@backstage/plugin-catalog-backend': minor
+'@backstage/plugin-catalog-node': minor
 ---
 
 Enable YAML merge keys in yamlPlaceholderResolver


valentingregoire:
The current implementation is not using the configurable approach, but has it hard-coded to true. To have backwards compatibility, I think we need to set it configurable and `false` by default.

plugins/catalog-backend/src/processors/PlaceholderProcessor.ts
```diff
@@ -140,7 +140,7 @@ export async function yamlPlaceholderResolver(
 
   let documents: yaml.Document.Parsed[];
   try {
-    documents = yaml.parseAllDocuments(content).filter(d => d);
+    documents = yaml.parseAllDocuments(content, { merge: true }).filter(d => d);
```
valentingregoire: This is hard-coded, which is different than the proposed solution. This is not backwards compatible I believe.

freben: I'd say it's backward compatible in the sense that it has _more_ capabilities than the code had before the change. It won't start breaking on files that used to work - right?

Dharma-09:
> Did you intentionally only add this in the placeholder resolver, and not [in the actual main](https://github.com/backstage/backstage/blob/6fc3b9f8fb1df313fa90ee7760a1d6d823c3abae/plugins/catalog-node/src/processing/parse.ts#L47) entity reader/parser?

Yes, merge key support is intentional in the resolver for now ,but I’d like to make it configurable so it’s not always forced on everyone.

plugins/catalog-backend/src/processors/PlaceholderProcessor.ts
Dharma-09: You’re both right ,the current implementation has `merge: true` hard-coded, which enables merge key support globally. While it technically adds functionality and doesn’t break existing YAMLs (as @freben mentioned).

freben:
Seems that the build isn't passing - API reports need to be updated.

freben (approved):

PR #31320, opened 2025-10-01
fix: trim mspraph query params

karpfslp:
…he app-config.yaml adds a line-break at the end of the string

The folded style definition in the `app-config.yaml` automatically adds a line-break to the end of the string.
In the Backstage version 1.43.0 the `encode` was set to `true` in the `qs.stringify()` in the _msgraph_.

That is why we have to `trim()` the query params to cut the unnecessary line breaks.
Without trimming the discovery does not work.

More context:

The folded style definition is forced in the [documentation](https://backstage.io/docs/integrations/azure/org/#installation):

```
        filter: >
            securityEnabled eq false
            and mailEnabled eq true
            and groupTypes/any(c:c+eq+'Unified')
``` 

But from v1.43.0 only the double-quoted style definition works:

```
        filter: "securityEnabled eq false \
            and mailEnabled eq true \
            and groupTypes/any(c:c+eq+'Unified')"
``` 

### DIFF
--- a/.changeset/some-signs-sort.md
+++ b/.changeset/some-signs-sort.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-msgraph': patch
+---
+
+Trim `mspraph` query params because the folded-styled definition in the `app-config.yaml` adds a line-break to the end of the string.
--- a/plugins/catalog-backend-module-msgraph/src/microsoftGraph/client.test.ts
+++ b/plugins/catalog-backend-module-msgraph/src/microsoftGraph/client.test.ts
@@ -358,6 +358,28 @@ describe('MicrosoftGraphClient', () => {
 
     expect(organization).toEqual({ displayName: 'Example' });
   });
+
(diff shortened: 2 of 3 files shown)


stephenglass:
Hey, thanks for raising this PR. I can't seem to reproduce this on my environment using the latest 1.43.3 version :/ maybe I'm missing something or some other difference. Also, another workaround fix is to use the strip indicator for the folding block like this:

```yaml
filter: >-
    securityEnabled eq false
    and mailEnabled eq true
    and groupTypes/any(c:c+eq+'Unified')
```

karpfslp:
Hmm. This wasn't the main problem. Under investigation ...

karpfslp:
The inner formatting also caused similar issue:

```
filter: >-
    mail in (
        'me@examlpe.com',
        'you@example.com'
    )
``` 

Therefore the double-quoted style will be the best in long run because it protects me against the inner formatting in the yaml:

```
filter: "mail in ( \
    'me@examlpe.com', \
    'you@example.com' \
    )"
``` 

This PR can be closed without merging.

Issue #30952, opened 2025-08-21
RFC: Use storageApi in AppLanguageSelector and appThemeSelector


karthikjeeyar:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
user setting backend, storageApi, ApplanguageSelector, persisting user language in database, appThemeSelector
```

### 🗃️ Project Area

Unknown

### 🔖 Need

Currently, the [AppLanguageSelector](https://github.com/backstage/backstage/blob/master/packages/core-app-api/src/apis/implementations/AppLanguageApi/AppLanguageSelector.ts#L64-L68) and [AppThemeSelector](https://github.com/backstage/backstage/blob/ed818545d4df4f8a69de48833535b9de1d1d16ef/packages/core-app-api/src/apis/implementations/AppThemeApi/AppThemeSelector.ts)  persists the selected value using localStorage directly. This makes the persistence mechanism limited to browser storage. Moreover these APIs are registered as `static` Apis, so there is no way to override them completely.

The current approach works for single-device usage but does not support syncing preferences across multiple browsers or devices.As organizations adopt Backstage widely, developers expect their personalizations (preferred theme, language, etc.) to follow them seamlessly across devices and sessions.

Without cross-device persistence:

- Switching browsers or devices resets preferences.
- Teams cannot recommend defaults effectively.
- User experience becomes inconsistent.

### 📝 Proposal

**PROPOSAL:**

 **Update AppLanguageSelector and AppThemeSelector**

   - Replace direct localStorage access with storageApi.
   - If no storageApi is configured, fallback to webStorage (localStorage-backed) implementation.
   - Apps can provide their own storageApi instance (e.g., backed by DB) for cross-device sync.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Undecided

freben:
Agreed on both of those. Well, language in particular. There's something to be said for wanting different themes on small- and large-screen devices though if the themes change item density / paddings etc. But maybe that's not used quite as much as just light/dark variations?

karthikjeeyar:
Good point, I agree that light/dark tends to be the more common use case. maybe we can keep light/dark as the baseline and later explore density changes if we see demand?

[...]

PR #31325, opened 2025-10-02
feat(core-app-api): use storageApi in app language and theme settings

karthikjeeyar:
## Use storageApi in appLanguageSelector and appThemeSelector

**RFC:** https://github.com/backstage/backstage/issues/30952

Replaces direct localStorage usage with StorageApi integration for theme and language settings, enabling flexible storage backends and cross-device synchronization using signalApi.

**Key changes:**

- Removed direct `window.localStorage` usage in favor of `StorageApi`
- Maintains cross-tab synchronization through StorageApi events
- Implemented signal-based cross-device synchronization via SignalApi

**Behavior:**

- **Default**: Uses StorageApi (typically WebStorage) for cross-tab synchronization. Other browser tabs sync automatically, but other devices need to refresh the page to get the preference.
- **With** `UserSettingsStorage`: Persists to backend database and enables cross-device synchronization
- **Backward Compatible**: Maintains same user experience with improved architecture

## video

https://github.com/user-attachments/assets/d217e66c-d72c-4e5e-915b-1603cfa5968c

__Notes__ : `UserSettingsStorage` is used in the above screen recording.

```
  createApiFactory({
    api: storageApiRef,
    deps: {
      discoveryApi: discoveryApiRef,
      errorApi: errorApiRef,
      fetchApi: fetchApiRef,
      identityApi: identityApiRef,
      signalApi: signalApiRef,
    },
    factory: deps => UserSettingsStorage.create(deps),
  }),
```


.changeset/brown-plants-hunt.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/core-app-api': minor
+'@backstage/plugin-app': minor
+---
+
+Replace localStorage with StorageApi for app theme and language settings
```
freben:
```suggestion
Replace `localStorage` with `StorageApi` for app theme and language settings
```

.changeset/brown-plants-hunt.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/core-app-api': minor
+'@backstage/plugin-app': minor
```
freben:
```suggestion
'@backstage/plugin-app': patch
```

nonbreaking in 0.x package = patch

packages/core-app-api/src/apis/implementations/AppThemeApi/AppThemeSelector.ts
```diff
@@ -14,11 +14,19 @@
...
-import { Observable } from '@backstage/types';
+import {
+  AppThemeApi,
+  AppTheme,
+  StorageApi,
+  ErrorApi,
+} from '@backstage/core-plugin-api';
+import { Observable, Subscription } from '@backstage/types';
 import { BehaviorSubject } from '../../../lib/subjects';
+import { SignalApi, SignalSubscriber } from '@backstage/plugin-signals-react';
```
freben: Sorry but this implies a core library importing a plugin, which is backwards. I see what you are aiming for and it's really cool but as long as the implementation is inside `core-app-api` or `frontend-app-api`  we can't do this.

One option is to put this "fancy" implementation instead in `plugins/app/src/apis` where it will be usable by all new-frontend-system installations.

karthikjeeyar: Thank you @freben for the review and inputs.  I have a followup question on this,  If I put the signals based implementation in the `plugins/app/src/apis`, is it okay to also include the signals ApiBlueprint to `plugins/app/src/defaultApis`? I'm asking because I am not sure if that is okay to make signals as a default API in new frontend system installation.

karthikjeeyar: I have moved signal based implementation to`plugins/app/src/apis` and added `signalApi` to defaultApis, let me know if this is okay or needs some changes.

Edit:  When adding `signalApi` to defaultApis I am again running into same issue where `signals` plugin is imported into `plugins/app` which is not permitted, so reverted the above changes.

I have created two new files `plugins/app/src/apis/DefaultAppThemeSelector` and `plugins/app/src/apis/DefaultAppLanguageSelector` with signal implementation but not able to wire them into AppThemeApi and AppLanguageApi because of the signalApi dependancy. @freben I would appreciate your inputs here to move forward. Thanks in advance.

plugins/app/src/extensions/AppThemeApi.tsx
```diff
@@ -28,6 +28,7 @@ import {
 } from '@backstage/frontend-plugin-api';
 // eslint-disable-next-line @backstage/no-relative-monorepo-imports
 import { AppThemeSelector } from '../../../../packages/core-app-api/src/apis/implementations';
+import { errorApiRef, storageApiRef } from '@backstage/core-plugin-api';
```
freben: Import these from `frontend-plugin-api` instead (which is the new-frontend-system library, as opposed to the old `core-plugin-api`)

packages/core-app-api/src/apis/implementations/AppLanguageApi/AppLanguageSelector.ts
```diff
@@ -53,49 +58,215 @@ export class AppLanguageSelector implements AppLanguageApi {
...
+      initialLanguage,
+      options?.storageApi,
+      options?.errorApi,
+    );
   }
 
-  static createWithStorage(options?: AppLanguageSelectorOptions) {
-    const selector = AppLanguageSelector.create(options);
+  static createWithStorage(
+    options: Omit<AppLanguageSelectorOptions, 'storageApi' | 'errorApi'>,
```
freben: This just gets a bit wonky. Keep it as a single object argument and add them as required here instead.
`options: AppLanguageSelectorOptions & { storageApi: StorageApi; errorApi: ErrorApi }` for example.

packages/core-app-api/src/apis/implementations/AppThemeApi/AppThemeSelector.test.ts
```diff
@@ -29,57 +31,101 @@ describe('AppThemeSelector', () => {
...
+    selector.setStorage(mockStorageApi, mockErrorApi);
+
     selector.setActiveThemeId('x');
     expect(subFn).toHaveBeenLastCalledWith('x');
     expect(selector.getActiveThemeId()).toBe('x');
 
     selector.setActiveThemeId(undefined);
     expect(subFn).toHaveBeenLastCalledWith(undefined);
-    expect(selector.getActiveThemeId()).toBe(undefined);
+    expect(selector.getActiveThemeId()).toBe('null');
```
freben: That seems bad. Why would you return the string `'null'`?

packages/core-app-api/src/apis/implementations/AppLanguageApi/AppLanguageSelector.ts
```diff
@@ -53,49 +58,215 @@ export class AppLanguageSelector implements AppLanguageApi {
...
+    return AppLanguageSelector.create({
+      ...options,
+      storageApi,
+      errorApi,
+    });
+  }
+  readonly #localStorageKey = `/${BUCKET_NAME}/${STORAGE_KEY}`;
+  readonly #languages: string[];
+  #language!: string;
+  #subject!: BehaviorSubject<{ language: string }>;
```
freben: what's up with the exclamation marks?

packages/core-app-api/src/apis/implementations/AppLanguageApi/AppLanguageSelector.ts
```diff
@@ -53,49 +58,215 @@ export class AppLanguageSelector implements AppLanguageApi {
...
+  #errorApi?: ErrorApi;
+  #storageSubscription?: Subscription;
+  #persistSubscription?: Subscription;
+  #isUpdatingFromStorage = false;
+  #lastStoredValue: any = undefined;
+  #isInitialLoad = true;
+  #lastKnownValue?: string;
+  #storageSetupRequired = false;
+  #localCacheSubscription?: StorageApi;
+  #isUpdatingFromCache = false;
```
freben: I feel this is WAY too much complexity and boolean state variables and stuff. This will not be fun to debug.

For starters, the contract for observable things is to always first call `.snapshot` to get an immediate state. That may or may not result in a known value. Then you call `.observe$` to start streaming in updates. The `getLanguage` + `language$` pair should be implementable in terms of those.

This makes you immune to the difference of storage methods that DO have sync access to state, and those that do not (and only send an async value a bit later when storage is ready).

Then you should be able to get rid of all usages of `instanceof` in these implementations.

You should not need to have a `setStorage` method, I feel. It just adds to the complexity to be able to do so after the constructor completes.

If you need to, to maintain the sync nature of the interfaces, maybe you can make `createWithStorage` async instead, so that it does all of the necessary messy async startup work (like transferring old local storage values to the store if needed etc) upfront.

packages/core-app-api/src/apis/implementations/AppThemeApi/AppThemeSelector.ts
karthikjeeyar: I had removed the signal based implementation to keep it simple.

packages/core-app-api/src/apis/implementations/AppThemeApi/AppThemeSelector.test.ts
karthikjeeyar: fixed it.

packages/core-app-api/src/apis/implementations/AppLanguageApi/AppLanguageSelector.ts
karthikjeeyar: removed non-null assertion operator and refactored.

packages/core-app-api/src/apis/implementations/AppLanguageApi/AppLanguageSelector.ts
karthikjeeyar: Updated both the appLanguageSelector and appThemeSelector Apis.

packages/core-app-api/src/app/AppManager.tsx
```diff
@@ -323,6 +329,14 @@ DEPRECATION WARNING: React Router Beta is deprecated and support for it will be
...
 
         const featureFlagsApi = this.getApiHolder().get(featureFlagsApiRef)!;
+        const storageApi = this.getApiHolder().get(storageApiRef)!;
+        const errorApi = this.getApiHolder().get(errorApiRef)!;
+        this.appLanguageApi = AppLanguageSelector.createWithStorage({
+          defaultLanguage: this.defaultLanguage,
+          availableLanguages: this.availableLanguages,
+          storageApi,
+          errorApi,
+        });
```
freben: Why is it being reconstructed here?

packages/core-app-api/src/app/AppManager.tsx
```diff
@@ -189,11 +191,15 @@ export class AppManager implements BackstageApp {
...
-    this.appLanguageApi = AppLanguageSelector.createWithStorage({
+    this.appLanguageApi = AppLanguageSelector.create({
       defaultLanguage: options.__experimentalTranslations?.defaultLanguage,
       availableLanguages:
         options.__experimentalTranslations?.availableLanguages,
     });
+    this.defaultLanguage =
+      options.__experimentalTranslations?.defaultLanguage ?? 'en';
+    this.availableLanguages = options.__experimentalTranslations
+      ?.availableLanguages ?? ['en'];
```
freben: If we don't reconstruct it again in the feature flag code, we can revert most of the changes here

packages/core-app-api/src/apis/implementations/AppLanguageApi/AppLanguageSelector.ts
```diff
@@ -53,77 +70,372 @@ export class AppLanguageSelector implements AppLanguageApi {
...
+    }
+  }
+
+  #setCachedValue(value: string | null): void {
+    if (typeof window === 'undefined' || !window.localStorage) {
+      return;
+    }
+    try {
+      if (value === null) {
+        window.localStorage.setItem(CACHE_KEY, JSON.stringify(null));
```
freben: This `if` currently achieves nothing; the branches do the same thing.

packages/core-app-api/src/apis/implementations/AppThemeApi/AppThemeSelector.ts
```diff
@@ -27,55 +45,348 @@ const STORAGE_KEY = 'theme';
...
   activeThemeId$(): Observable<string | undefined> {
-    return this.subject;
+    if (!this.storage) {
+      return this.#createEmptyObservable();
+    }
+
+    const bucket = this.storage.forBucket(BUCKET_NAME);
+    const initialValue = this.#getInitialThemeValue(bucket);
+
+    return new ObservableImpl<string | undefined>(subscriber => {
```
freben: You don't have to make a new complete observable for every request (along with the extra storage logic etc).

I'd have expected a single observable to be made in the constructor, where also subscriptions to external sources (local storage, the storage api etc) gets made once and once only. This is basically what the old code already did, right? It was in the form of a `BehaviorSubject` which comes with additional semantics; it publishes on subscription.

Overall it seems to me that this class has become unnecessarily complex. Was it not possible to keep its old _exact shape_ and just adding a few extra rows to `createWithStorage`?

Same goes for the other class basically.

packages/core-app-api/src/app/AppManager.tsx
karthikjeeyar: If I remember correctly, Api's were not available at the first time. Hence I had reconstructed it later once the APIs are available to use. Let me take a look again and see if I can avoid this.

karthikjeeyar:
Thanks @freben  for the detailed review!

 I’ve addressed the concerns as follows:
- **Kept the old shape**: both selectors still use a single BehaviorSubject and do not create new observables per call. The core class remains minimal. most logic lives in createWithStorage.
- **Last‑known value + async refresh**: added a small "last known value" cache in localStorage that’s only used when the storage snapshot is async `(presence === 'unknown')`. This prevents flicker on reload and syncs across tabs. (This aligns with your [comment](https://github.com/backstage/backstage/issues/30952#issuecomment-3265752859) on the issue)

-  **Options - Single argument**: createWithStorage now takes a single options object. storageApi stays optional to support early boot where APIs aren’t ready.


**For the reconstruction comment:**


I still need a lightweight instance during early boot because storageApiRef isn’t available until config/APIs resolve. If I pull it earlier, we hit “Tried to access config API before config was loaded” error. Do you have a recommended pattern to avoid the second createWithStorage while keeping API creation synchronous? I’m happy to adopt a suggested approach.

Issue #31302, opened 2025-09-29
GitlabDiscoveryEntityProvider does not check for "orgEnabled" variable


Alexandru-Pita:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
catalog
gitlab
integration:gitlab
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

GitLab

### 📝 Description & Context

There seems to be a logic mismatch between [GitlabDiscoveryEntityProvider](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts) and [GitlabOrgDiscoveryEntityProvider](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend-module-gitlab/src/providers/GitlabOrgDiscoveryEntityProvider.ts).

Right now `GitlabOrgDiscoveryEntityProvider` will look for `orgEnabled` variable in `app-config.yaml`. Based on this value it will / will not add a provider that will fetch users and groups. This can be seen in this block of code:
```
if (!providerConfig.orgEnabled) {
  options.logger.info(`Org not enabled for ${providerConfig.id}.`);
  return;
}
```

The same logic does not apply to `GitlabDiscoveryEntityProvider`.

Let's say that I have this `app-config.yaml` configuration:
```yaml
catalog:
  orphanStrategy: delete
  providers:
    gitlab:
      gitlabDiscovery:
        host: hosted.giltab.com 
        branch: main
        fallbackBranch:
        skipForkedRepos: true
        includeArchivedRepos: false
        entityFilename: catalog-info.yaml
        projectPattern: '[\s\S]*'
        excludeRepos: []
        schedule:
          frequency: { hours: 6 }
          timeout: { hours: 2 }
          initialDelay: { hours: 1 }
      gitlabOrgDiscovery:
        host: hosted.giltab.com
        orgEnabled: true
        groupPattern: '[\s\S]*'
        userPattern: '[\s\S]*'
        schedule:
          frequency: { hours: 2 }
          timeout: { minutes: 30 }
          initialDelay: { minutes: 1 }
```

The result would be that 3 providers will be added in the end, instead of 2, and here is why:
[...]

PR #31332, opened 2025-10-05
GitlabDiscoveryEntityProvider now checks for "orgEnabled" config variable

Alexandru-Pita:

This PR will have `GitlabDiscoveryEntityProvider` now check for `orgEnabled` config variable.

I opened this PR as a solution for the issue I posted [here](https://github.com/backstage/backstage/issues/31302).

I know that this PR has a **major change** which will affect all current backstage users that use **GitLab ingestion**.

### DIFF
--- a/.changeset/giant-corners-thank.md
+++ b/.changeset/giant-corners-thank.md
@@ -1,0 +1,7 @@
+---
+'@backstage/plugin-catalog-backend-module-gitlab': minor
+---
+
+**BREAKING** The `GitlabDiscoveryEntityProvider` and `GitlabOrgDiscoveryEntityProvider` can no longer be used together with the same `provider id`. This is a **major change** which will affect database entries as well.
+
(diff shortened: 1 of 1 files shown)


freben:
As mentioned in the issue, I think it might be worth

* Adding a new `gitlabOrg` config section (including in `config.d.ts`) specifically for org data ingestion.
  * This will be the new "blessed" place for such ingestion
  * and of course those do not need an `orgEnabled` flag at all
  * and only has the settings that apply to org data.
* Unbreak support for the `orgEnabled` flag in the old provider structure if that's appropriate (i'll let you decide)
  * If encountered, log a warning that suggests to please move over to use `gitlabOrg` instead
* Change to a `patch` level changeset since it's no longer breaking

This gives people ample time to move over. And of course the docs etc should promote the new way right away.

Alexandru-Pita:
Thank you for the comments! I will try to implement the suggestions in a few days

PR #31344, opened 2025-10-07
feat(scaffolder): add swappable components for nfs

### LINKED ISSUE #31251: New frontend: can't override template card in scaffolder-plugin
### 📜 Issue Labels

### 🔎 Search Terms

```plain
template card templatecard scaffolder override
```

### 🗃️ Project Area

Software Templates

### 🔖 Need

Before starting migration to new frontend system, we have been using a custom TemplateCard component in scaffolder page. 

```typescript
<ScaffolderPage components={{ TemplateCardComponent: TemplateCard }} />
```

I didn't find a way to override TemplateCard in the scaffolderPlugin when using the new frontend version of the plugin, however it is possible to override the whole scaffolder page, which is not optimal when we want to override the card only.

Image

I believe we are not the only team using a custom TemplateCard component.

### 📝 Proposal

Implement a possibility to override the TemplateCard. Something like:

`scaffolderPlugin.getExtension("component:template-card").withOverrides(...)`

### 🔄 Alternatives

_No response_

Undecided

drodil:

this adds necessary swappable components to scaffolder nfs to be able to override the default components.

and on top of that, scaffolder page also supports basic config for header, default preview template, and context menu.

closes #31251

### DIFF
--- a/plugins/scaffolder-react/src/components/types.ts
+++ b/plugins/scaffolder-react/src/components/types.ts
@@ -56,5 +56,6 @@ export type ReviewStepProps = {
     uiSchema: UiSchema;
     mergedSchema: JsonObject;
     schema: JsonObject;
+    title: string;
   }[];
 };
--- a/plugins/scaffolder/src/alpha/extensions.tsx
+++ b/plugins/scaffolder/src/alpha/extensions.tsx
@@ -33,6 +33,14 @@ import { FormFieldBlueprint } from '@backstage/plugin-scaffolder-react/alpha';
 import { scmIntegrationsApiRef } from '@backstage/integration-react';
 import { scaffolderApiRef } from '@backstage/plugin-scaffolder-react';
 import { ScaffolderClient } from '../api';
+import {
+  SwappableReviewStepComponent,
+  SwappableTaskPageComponent,
+  SwappableTemplateCardComponent,
+  SwappableTemplateListPageComponent,
+  SwappableTemplateOutputsComponent,
+  SwappableTemplateWizardPageComponent,
+} from './swappableComponents';
 
 export const scaffolderPage = PageBlueprint.makeWithOverrides({
   inputs: {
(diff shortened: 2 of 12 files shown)


plugins/app/src/apis/SwappableComponentsApi/DefaultSwappableComponentsApi.tsx
```diff
@@ -71,4 +71,14 @@ export class DefaultSwappableComponentsApi implements SwappableComponentsApi {
...
   }
+
+  componentIsLoadable(ref: SwappableComponentRef<any>): boolean {
+    if (!this.#components.has(ref.id)) {
+      return false;
+    }
+    const OverrideComponent = this.#components.get(ref.id);
+    const { defaultComponent: DefaultComponent } =
+      OpaqueSwappableComponentRef.toInternal(ref);
+    return Boolean(OverrideComponent || DefaultComponent);
```
drodil: Just noticed this might not work because there is the `FallbackComponent` defined, @Rugvip any pointers how to make this https://github.com/backstage/backstage/pull/31344/files#diff-5fb0067db499739a8def2b948fb25eed16fcfa5c430a77d18221bd2da6adc8b8R94 happen?

plugins/scaffolder/src/alpha/extensions.tsx
```diff
@@ -26,31 +26,83 @@ import {
...
       loader: () =>
         import('../components/Router/Router').then(m =>
           compatWrapper(
-            <m.InternalRouter formFieldLoaders={formFieldLoaders} />,
+            <m.InternalRouter
+              {...config}
+              formFieldLoaders={formFieldLoaders}
+              components={{
+                TemplateCardComponent: SwappableTemplateCardComponent,
+                ReviewStepComponent: reviewStepComponentExists
```
benjdlambert: I'm not sure that this is strictly necessary.

We should just have the default implementation baked into the `swappableComponent` and just always pass that through. That way it will render the swapped one, or the default one.

drodil: The issue is that the default component is inlined in the `Stepper` and requires quite a bit of refactoring. I was hoping there was support for optional swappable components but apparently not; thus making the refactoring of the stepper the only option. I will get back to this tomorrow!

plugins/scaffolder-react/src/components/types.ts
```diff
@@ -56,5 +56,6 @@ export type ReviewStepProps = {
     uiSchema: UiSchema;
     mergedSchema: JsonObject;
     schema: JsonObject;
+    title: string;
```
drodil, line 59: Unfortunately, this has to be added, or the title in `ParsedTemplateSchema` must be changed to allow undefined. Otherwise, the separation of the default review step is not possible because of TypeScript errors.

benjdlambert:
Hey @drodil!

Thanks for this PR!

I'm a little torn on this PR, whilst this is the solution in the old system, I think really we should be looking at how we actually break out the scaffolder routes into different `PageExtensions`, and then you can override the `PageExtension` with your own implementation instead.

Feels like that really should be the right solution? But I still think we're learning here.

There's some work going on right now with the plugin header, which I think is going to help us understand how best to lay out these `PageExtensions` and I think that's going to push this solution in a particular direction. Are we happy to settle that work first?

https://github.com/backstage/backstage/issues/31412

Issue #31214, opened 2025-09-22
InvalidPart error during Techdocs S3 multipart uploads


vidhanshah:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
Upload S3 
Techdocs
Upload files
Aws S3
Multipart
Bulk files
```

### 🗃️ Project Area

TechDocs

### 🔗 External Integration

N/A

### 📝 Description & Context

AWS S3 publisher for Techdocs is experiencing intermittent failures during file uploads, particularly when uploading larger files or when multiple files are being uploaded concurrently. The error suggests issues with S3 upload handling

**Current behavior**:

- Upload process fails intermittently with InvalidPart errors

-  Error message: Unable to upload file(s) to AWS S3. InvalidPart: One or more of the specified parts could not be found. The part may not have been uploaded, or specified entry tag may not match the part's entity tag.

- The error occurs during the bulkStorageOperation phase in the publish method.
- Particularly affects uploads with multiple large files.

### 👍 Expected Behavior

- All documentation files should upload successfully to S3 without multipart upload errors.
- The publish process should complete successfully for entities with multiple files.
- Large files should be handled properly via multipart uploads.
- Concurrent uploads should work reliably without part conflicts.

### 🥾 Reproduction steps

1. Setup Techdocs with AWS S3 publisher:
     techdocs:
         publisher:
             type: awsS3
             awsS3:
                  bucketName: your-techdocs-bucket
                  region: us-east-1
                  bucketRootPath: techdocs

2. Create documentation with large assets:

- Add multiple large images.
- Ensure total documentation size is > 10MB

3. Trigger documentation generation and publishing
4. Observe the failure:

-  Monitor backend logs during publish.
-  Error occurs during upload phase with InvalidPart message

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes, but I would like some more guidance

benjdlambert:
@backstage/techdocs-maintainers

vidhanshah:
Hi, I wanted to kindly follow up on this issue. I was wondering if there are any updates or if additional information is needed from my side to help address the problem.

Looking forward to your response. Thank You!!

vidhanshah:
[...]

PR #31352, opened 2025-10-07
Modifying large size files upload to S3

vidhanshah:
Improved large size file issue to AWS S3. We have now modified the uploading functionality with the multipart upload logic.

Open Issue: https://github.com/backstage/backstage/issues/31214

### DIFF
--- a/plugins/techdocs-node/src/stages/publish/awsS3.test.ts
+++ b/plugins/techdocs-node/src/stages/publish/awsS3.test.ts
@@ -31,7 +31,7 @@ import {
   AwsCredentialProviderOptions,
   DefaultAwsCredentialsManager,
 } from '@backstage/integration-aws-node';
-import { mockClient, AwsClientStub } from 'aws-sdk-client-mock';
+import { mockClient } from 'aws-sdk-client-mock';
 import express from 'express';
 import request from 'supertest';
 import path from 'path';
@@ -44,9 +44,10 @@ import {
 } from '@backstage/backend-test-utils';
 
 const env = process.env;
-let s3Mock: AwsClientStub<S3Client>;
+let s3Mock: any;
 
-const mockDir = createMockDirectory();
+// Create a new MockDirectory for each test to avoid Windows file locking issues
+let mockDir: ReturnType<typeof createMockDirectory>;
 
 function getMockCredentialProvider(): Promise<AwsCredentialProvider> {
   return Promise.resolve({
@@ -155,7 +156,7 @@ describe('AwsS3Publish', () => {
     build_timestamp: 612741599,
   };
 
-  const directory = getEntityRootDir(entity);
+  let directory: string;
 
   const files = {
     'index.html': '',
@@ -176,7 +177,7 @@ describe('AwsS3Publish', () => {
     },
   };
 
-  beforeEach(() => {
+  beforeEach(async () => {
     process.env = { ...env };
     process.env.AWS_REGION = 'us-west-2';
 
@@ -185,20 +186,26 @@ describe('AwsS3Publish', () => {
       getMockCredentialProvider(),
     );
 
+    // Create a fresh mockdirectory for each test to avoid windows file locking
+    mockDir = createMockDirectory();
+    // Calculate directory path with the new mockDir instance
+    directory = getEntityRootDir(entity);
+
+    // Set up the test files
     mockDir.setContent({
       [directory]: files,
     });
 
-    s3Mock = mockClient(S3Client);
+    s3Mock = mockClient(S3Client as any);
 
-    s3Mock.on(HeadObjectCommand).callsFake(input => {
+    s3Mock.on(HeadObjectCommand).callsFake((input: any) => {
       if (!fs.pathExistsSync(mockDir.resolve(input.Key))) {
         throw new Error('File does not exist');
       }
       return {};
     });
 
-    s3Mock.on(GetObjectCommand).callsFake(input => {
+    s3Mock.on(GetObjectCommand).callsFake((input: any) => {
       if (fs.pathExistsSync(mockDir.resolve(input.Key))) {
         return {
           Body: Readable.from(fs.readFileSync(mockDir.resolve(input.Key))),
@@ -208,14 +215,14 @@ describe('AwsS3Publish', () => {
       throw new Error(`The file ${input.Key} does not exist!`);
     });
 
-    s3Mock.on(HeadBucketCommand).callsFake(input => {
+    s3Mock.on(HeadBucketCommand).callsFake((input: any) => {
       if (input.Bucket === 'errorBucket') {
         throw new Error('Bucket does not exist');
       }
       return {};
     });
 
-    s3Mock.on(ListObjectsV2Command).callsFake(input => {
+    s3Mock.on(ListObjectsV2Command).callsFake((input: any) => {
       if (
         input.Bucket === 'delete_stale_files_success' ||
         input.Bucket === 'delete_stale_files_error'
@@ -227,7 +234,7 @@ describe('AwsS3Publish', () => {
       return {};
     });
 
-    s3Mock.on(DeleteObjectCommand).callsFake(input => {
+    s3Mock.on(DeleteObjectCommand).callsFake((input: any) => {
       if (input.Bucket === 'delete_stale_files_error') {
         throw new Error('Message');
       }
@@ -235,7 +242,7 @@ describe('AwsS3Publish', () => {
     });
 
     s3Mock.on(UploadPartCommand).rejects();
-    s3Mock.on(PutObjectCommand).callsFake(input => {
+    s3Mock.on(PutObjectCommand).callsFake((input: any) => {
       mockDir.addContent({ [input.Key]: input.Body });
     });
   });
@@ -320,7 +327,7 @@ describe('AwsS3Publish', () => {
           `default/component/backstage/assets/main.css`,
         ]),
       });
-    });
+    }, 30000);
 
     it('should publish a directory as well when legacy casing is used', async () => {
       const publisher = await createPublisherFromConfig({
@@ -333,7 +340,7 @@ describe('AwsS3Publish', () => {
           `default/Component/backstage/assets/main.css`,
         ]),
       });
-    });
+    }, 30000);
 
     it('should publish a directory when root path is specified', async () => {
       const publisher = await createPublisherFromConfig({
@@ -346,7 +353,7 @@ describe('AwsS3Publish', () => {
           `backstage-data/techdocs/default/component/backstage/assets/main.css`,
         ]),
       });
-    });
+    }, 30000);
 
     it('should publish a directory when root path is specified and legacy casing is used', async () => {
       const publisher = await createPublisherFromConfig({
@@ -360,7 +367,7 @@ describe('AwsS3Publish', () => {
           `backstage-data/techdocs/default/Component/backstage/assets/main.css`,
         ]),
       });
-    });
+    }, 30000);
 
     it('should publish a directory when sse is specified', async () => {
       const publisher = await createPublisherFromConfig({
@@ -373,7 +380,7 @@ describe('AwsS3Publish', () => {
           'default/component/backstage/assets/main.css',
         ]),
       });
-    });
+    }, 30000);
 
     it('should fail to publish a directory', async () => {
       const wrongPathToGeneratedDirectory = mockDir.resolve(
@@ -408,7 +415,7 @@ describe('AwsS3Publish', () => {
       expect(loggerInfoSpy).toHaveBeenLastCalledWith(
         `Successfully deleted stale files for Entity ${entity.metadata.name}. Total number of files: 1`,
       );
-    });
+    }, 30000);
 
     it('should log error when the stale files deletion fails', async () => {
       const bucketName = 'delete_stale_files_error';
@@ -419,7 +426,7 @@ describe('AwsS3Publish', () => {
       expect(loggerErrorSpy).toHaveBeenLastCalledWith(
         'Unable to delete file(s) from AWS S3. Error: Message',
       );
-    });
+    }, 30000);
   });
 
   describe('hasDocsBeenGenerated', () => {
@@ -427,7 +434,7 @@ describe('AwsS3Publish', () => {
       const publisher = await createPublisherFromConfig();
       await publisher.publish({ entity, directory });
       expect(await publisher.hasDocsBeenGenerated(entity)).toBe(true);
-    });
+    }, 30000);
 
     it('should return true if docs has been generated even if the legacy case is enabled', async () => {
       const publisher = await createPublisherFromConfig({
@@ -435,7 +442,7 @@ describe('AwsS3Publish', () => {
       });
       await publisher.publish({ entity, directory });
       expect(await publisher.hasDocsBeenGenerated(entity)).toBe(true);
-    });
+    }, 30000);
 
     it('should return true if docs has been generated if root path is specified', async () => {
       const publisher = await createPublisherFromConfig({
@@ -443,7 +450,7 @@ describe('AwsS3Publish', () => {
       });
       await publisher.publish({ entity, directory });
       expect(await publisher.hasDocsBeenGenerated(entity)).toBe(true);
-    });
+    }, 30000);
 
     it('should return true if docs has been generated if root path is specified and legacy casing is used', async () => {
       const publisher = await createPublisherFromConfig({
@@ -452,7 +459,7 @@ describe('AwsS3Publish', () => {
       });
       await publisher.publish({ entity, directory });
       expect(await publisher.hasDocsBeenGenerated(entity)).toBe(true);
-    });
+    }, 30000);
 
     it('should return false if docs has not been generated', async () => {
       const publisher = await createPublisherFromConfig();
@@ -475,7 +482,7 @@ describe('AwsS3Publish', () => {
       expect(await publisher.fetchTechDocsMetadata(entityName)).toStrictEqual(
         techdocsMetadata,
       );
-    });
+    }, 30000);
 
     it('should return tech docs metadata even if the legacy case is enabled', async () => {
       const publisher = await createPublisherFromConfig({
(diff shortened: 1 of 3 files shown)


plugins/techdocs-node/src/stages/publish/awsS3.test.ts
```diff
@@ -185,20 +186,26 @@ describe('AwsS3Publish', () => {
...
+    // Calculate directory path with the new mockDir instance
+    directory = getEntityRootDir(entity);
+
+    // Set up the test files
     mockDir.setContent({
       [directory]: files,
     });
 
-    s3Mock = mockClient(S3Client);
+    s3Mock = mockClient(S3Client as any);
```
vinzscam:
```suggestion
    s3Mock = mockClient(S3Client);
```

plugins/techdocs-node/src/stages/publish/awsS3.test.ts
```diff
@@ -185,20 +186,26 @@ describe('AwsS3Publish', () => {
...
 
-    s3Mock.on(HeadObjectCommand).callsFake(input => {
+    s3Mock.on(HeadObjectCommand).callsFake((input: any) => {
```
vinzscam: same below

```suggestion
    s3Mock.on(HeadObjectCommand).callsFake((input) => {
```

plugins/techdocs-node/src/stages/publish/awsS3.test.ts
```diff
@@ -320,7 +327,7 @@ describe('AwsS3Publish', () => {
           `default/component/backstage/assets/main.css`,
         ]),
       });
-    });
+    }, 30000);
```
vinzscam: was this on purpose? 🤔 
```suggestion
    });
```

vidhanshah: I have Added 30000 as timeout to certain tests which helps to prevent test failures in lower environments or when AWS operations take longer than the default timeout.

The timeout value of 30000 milliseconds (30 seconds) is added to specific tests that involve AWS S3 operations because: Even though we're using mocks, these tests simulate real AWS S3 operations which typically involve:

Publishing documents to S3
Fetching documents from S3
Handling large files
Dealing with multiple file operations in sequence

plugins/techdocs-node/src/stages/publish/awsS3.test.ts
vidhanshah: Done

plugins/techdocs-node/src/stages/publish/awsS3.test.ts
vidhanshah: Done and removed from rest places as well

vidhanshah:
> Thank you. How can we test this new logic?

I had one document file which was over 10MB and we simply deployed in the dev/pilot env that way our doc file got uploaded to S3.

Before this multipart upload logic: we were getting the error message as: Unable to upload the file to the AWS S3. 

The PR's fix, helps to upload the folder or doc file upto 100MB.

plugins/techdocs-node/src/stages/publish/awsS3.ts
freben: Maybe this should be a `const MAX_SINGLE_UPLOAD_BYTES` at the top of this file? You use the same constant a bit further down

freben: I don't think I understand this. We already know that the file is bigger than the limit, why is there a `max` here? Also, are you saying that this forcibly makes _ten thousand_ chunks no matter how large the file is?

Did you mean so just say `partSize: MAX_SINGLE_UPLOAD_BYTES` instead of doing any math here?

freben: Does this really make a difference? What's stopping the code from just using `Upload` all the time? Seems to me that the code would be _significantly_ simplified if the only change in this file basically was to add 

```
partSize: MAX_SINGLE_UPLOAD_BYTES,
queueSize: 3,
leavePartsOnError: false,
```

to the existing upload. It would have made this a three-line PR would it not? OK maybe plus some timeout changes too.

freben: would feel more readable to make this `for (let attempt = 1; attempt < maxAttempts; ++attempt)`

Also I am not sure about the structure of this thing. Could it be shaped more like

```ts
for (...) {
  try {
    return await operation();
  } catch (e) {
    if (attempt > limit || !shouldRetryOperation(e)) {
      throw e;
    }
    log(...)
    await sleep(...);
  }
}
```

Then `shouldRetryOperation` doesn't need to know about retry counts etc, and there doesn't need to be any last call at the bottom

freben: Not sure about the `attempts === 0` check - you are running it one final time after the loop anyway?

freben: Oof this feels like it should have been a dedicated parameter to `retryOperation` instead

freben: Because of the above `Final attempt without retry wrapper` - this check really isn't needed is it

freben: should this have three `o`?

freben: Is this really transient?

freben: same here, does this really need to check the number of remaining attempts?

freben: same here, remaining check not needed?

vidhanshah (commented):
Thank you @freben for your PR comments, but i am not able to see your reference to the comments in the exact code lines.

I have although tried to fix the code changes as expected!!

vidhanshah:
> This REALLY feels like it should have been split apart, into only the multipart change + timeouts, and then maybe making retries as a completely separate PR because it significantly increases the complexity and makes it tougher to get through review

The changes to the AWS S3 publishing mechanism represent a cohesive set of improvements that work together to enhance reliability and performance. While the multipart upload, timeout configurations, and retry mechanisms form an integrated solution to common S3 upload challenges. Here's why they belong together:

**1. Technical Integration**
- The multipart upload feature naturally requires robust retry handling for partial uploads
- Timeout configurations directly influence when retry mechanisms are triggered
- The retry logic is specifically tuned for multipart upload scenarios

**2. Real-world Usage Pattern**
The implementation mirrors real-world S3 upload patterns where:
- Large files require multipart uploads (using `MAX_SINGLE_UPLOAD_BYTES` constant)
- Network interruptions during long uploads need retry mechanisms
- Timeout configurations prevent hanging operations
- All three components work together to ensure successful uploads

Parsifal-M:
I think chatgpt didn't quite understand the ask, it's not the fact that they don't belong together it's the fact that you could raise them separately in different PRs to smooth the review process.

@freben isn't saying they don't belong together as part of the bigger picture, just that it would be easier to introduce them in smaller changes to make reviewing easier 👍

edit: to be clear I don't have much of an issue with copy pasting from LLMs just trying to clarify 😅 and I think in this case it missed the point 😬

plugins/techdocs-node/src/stages/publish/awsS3.ts
freben: You still should remove this constant in favor of `MAX_SINGLE_UPLOAD_BYTES`

plugins/techdocs-node/src/stages/publish/awsS3.ts
```diff
@@ -175,10 +184,22 @@ export class AwsS3Publish implements PublisherBase {
...
+      maxAttempts: maxAttempts || 5,
+      retryMode: 'adaptive',
       ...(httpsProxy && {
         requestHandler: new NodeHttpHandler({
           httpsAgent: new HttpsProxyAgent({ proxy: httpsProxy }),
+          // Enhanced connection setting for large file uploads
+          connectionTimeout: 60000,
+          socketTimeout: 120000,
+        }),
+      }),
```
freben: this could be simplified to

```ts
requestHandler: new NodeHttpHandler({
  ...(httpsProxy && { httpsAgent: new HttpsProxyAgent({ proxy: httpsProxy }) }),
  connectionTimeout: 60000,
  socketTimeout: 120000,
}),
```

to not have to repeat yourself

plugins/techdocs-node/src/stages/publish/awsS3.ts
```diff
@@ -250,6 +272,86 @@ export class AwsS3Publish implements PublisherBase {
...
   }
+  /**
+   * Custom retry wrapper for S3 operations with detailed error handling.
+   */
+  private async retryOperation<TOutput>(
+    operation: () => Promise<TOutput>,
+    operationName: string,
+    maxAttempts: number = 3,
+    shouldRetry: (error: S3ServiceException) => boolean = this
+      .defaultShouldRetry,
```
freben:
```suggestion
    shouldRetry: (error: S3ServiceException) => boolean = this
      .defaultShouldRetry.bind(this),
```

to not get surprising errors some time down the line when it actually accesses `this`

plugins/techdocs-node/src/stages/publish/awsS3.ts
```diff
@@ -250,6 +272,86 @@ export class AwsS3Publish implements PublisherBase {
...
+          attempt,
+          maxAttempts,
+          error: e.message,
+          errorCode: e.name,
+          httpStatusCode: e.$metadata?.httpStatusCode,
+        });
+
+        // Enhanced exponential backoff with jitter
+        const baseDelay = operationName.startsWith('Upload-') ? 2000 : 1000;
+        const backoffDelay = Math.min(baseDelay * Math.pow(2, attempt), 30000);
```
freben: this always starts at `baseDelay * 2` instead of `baseDelay`, doesn't it

plugins/techdocs-node/src/stages/publish/awsS3.ts
```diff
@@ -250,6 +272,86 @@ export class AwsS3Publish implements PublisherBase {
...
+        'RequestTimeoutException',
+        'PriorRequestNotComplete',
+      ];
+      return retriable4xxErrors.includes(errorCode);
+    }
+
+    // Check against known transient errors
+    return transientErrors.some(
+      retriableError =>
+        errorCode.includes(retriableError) ||
```
freben: should this be just an equality test instead?

plugins/techdocs-node/src/stages/publish/awsS3.ts
```diff
@@ -320,30 +438,99 @@ export class AwsS3Publish implements PublisherBase {
...
+              this.maxAttempts,
+            );
+            return result;
+          } catch (error) {
+            const s3Error = error as any;
+            const errorName = s3Error?.name || 'Unknown';
+
+            // Check if this is a multipart upload failure that we can handle
+            if (
+              fileSizeInBytes >= MAX_SINGLE_UPLOAD_BYTES &&
```
freben: would probably a better structure to do this check just once outside it all

```ts
if (large upload) {
  try {
    do multipart upload
    return;
  } catch (e) {
    could not do large upload, log and fall through
  }
}

try {
  do single-part upload
  return;
} catch (e) {
  more single-part specific error handling; throw
}
```

plugins/techdocs-node/src/stages/publish/awsS3.test.ts
freben: You can also just do a single `jest.setTimeout(30_000)` at the top of the file to capture them all automatically, that's what i always do

plugins/techdocs-node/src/stages/publish/awsS3.ts
vidhanshah: Done

plugins/techdocs-node/src/stages/publish/awsS3.ts
vidhanshah: Handled this one!!

plugins/techdocs-node/src/stages/publish/awsS3.ts
vidhanshah: Handled this one!!

plugins/techdocs-node/src/stages/publish/awsS3.ts
vidhanshah: Handled this one!!

plugins/techdocs-node/src/stages/publish/awsS3.ts
vidhanshah: Handled this one!!

vidhanshah:
> I'm approving this on my end but suggesting some optional improvements. Final word on to @backstage/techdocs-maintainers

@freben, I have made changes as well as per your new PR comments, if possible can you again re-approve from your side and then it will be left with @backstage/techdocs-maintainers

plugins/techdocs-node/src/stages/publish/awsS3.test.ts
```diff
@@ -44,9 +45,10 @@ import {
 } from '@backstage/backend-test-utils';
 
 const env = process.env;
-let s3Mock: AwsClientStub<S3Client>;
+let s3Mock: any;
```
johnphilip283: Any reason why we had to move back to `any`? We'd like to keep the typing here as strict as possible if possible

plugins/techdocs-node/src/stages/publish/awsS3.test.ts
```diff
@@ -320,7 +514,7 @@ describe('AwsS3Publish', () => {
           `default/component/backstage/assets/main.css`,
         ]),
       });
-    });
+    }, 30000);
```
johnphilip283: were there any code changes that caused these tests to consistently need their timeouts bumped? this is quite a lot of timeout bumps for sensitive parts of the techdocs pipeline so I'm quite hesitant on these changes

plugins/techdocs-node/src/stages/publish/awsS3.ts
```diff
@@ -175,11 +184,16 @@ export class AwsS3Publish implements PublisherBase {
       ...(region && { region }),
       ...(endpoint && { endpoint }),
       ...(forcePathStyle && { forcePathStyle }),
-      ...(maxAttempts && { maxAttempts }),
-      ...(httpsProxy && {
-        requestHandler: new NodeHttpHandler({
+      // Enhanced retry configuration for better reliability
+      maxAttempts: maxAttempts || 5,
```
johnphilip283, line 188: Above when setting `maxAttempts` the comment says the value is defaulted to 3 - could you either update that comment or update this default?

plugins/techdocs-node/src/stages/publish/awsS3.ts
```diff
@@ -320,30 +434,93 @@ export class AwsS3Publish implements PublisherBase {
       // e.g. ['index.html', 'sub-page/index.html', 'assets/images/favicon.png']
       absoluteFilesToUpload = await getFileTreeRecursively(directory);
 
+      let uploadCounter = 0;
+
       await bulkStorageOperation(
         async absoluteFilePath => {
+          uploadCounter++;
```
johnphilip283: I don't see this `uploadCounter` variable used anywhere - can we remove?

plugins/techdocs-node/src/stages/publish/awsS3.ts
```diff
@@ -320,30 +434,93 @@ export class AwsS3Publish implements PublisherBase {
...
             Bucket: this.bucketName,
-            Key: getCloudPathForLocalPath(
-              entity,
-              relativeFilePath,
-              useLegacyPathCasing,
-              bucketRootPath,
-            ),
-            Body: fileStream,
+            Key: s3Key,
+            Body: absoluteFilePath,
```
johnphilip283: Why did we need to change from using the file stream to just using the absolute path?

plugins/techdocs-node/src/stages/publish/awsS3.test.ts
```diff
@@ -185,20 +189,26 @@ describe('AwsS3Publish', () => {
...
+
+    // Set up the test files
     mockDir.setContent({
       [directory]: files,
     });
 
     s3Mock = mockClient(S3Client);
 
-    s3Mock.on(HeadObjectCommand).callsFake(input => {
+    s3Mock.on(HeadObjectCommand).callsFake((input: { Key: string }) => {
```
johnphilip283: do we still need the type annotations now that we've strongly typed `s3Mock` again?

plugins/techdocs-node/src/stages/publish/awsS3.test.ts
```diff
@@ -292,6 +304,190 @@ describe('AwsS3Publish', () => {
     });
   });
 
+  describe('retry mechanism', () => {
+    it('should retry with custom retry strategy', async () => {
+      const publisher = (await createPublisherFromConfig()) as AwsS3Publish;
```
johnphilip283: do we still need to cast this explicitly to AwsS3Publish? I'd prefer to avoid it as much as possible, along with the casts to any in the following tests.

plugins/techdocs-node/src/stages/publish/awsS3.test.ts
vidhanshah: Mock patterns benefit from explicit callback. It is making it consistent.
I can remove them if you prefer, but I'd recommend keeping them for test maintainability.

vidhanshah (commented):
@johnphilip283, Made the asked changes. Can you review it now?

plugins/techdocs-node/src/stages/publish/awsS3.test.ts
```diff
@@ -292,6 +302,190 @@ describe('AwsS3Publish', () => {
...
+        .rejectsOnce(
+          new S3ServiceException({
+            name: 'RequestTimeout',
+            $fault: 'client',
+            $metadata: { httpStatusCode: 408 },
+          }),
+        )
+        .resolvesOnce({ Contents: [] });
+
+      await (publisher as any).retryOperation(
```
johnphilip283: can we remove all instances of (publisher as any) and (s3Mock.send as any) in the tests now that these objects are more strictly typed?

vidhanshah: @johnphilip283, I have removed all the instances of as any so that the objects are more strictly typed!!

vidhanshah:
@backstage/techdocs-maintainers, @johnphilip283 please review this PR!!

freben:
@vinzscam to unblock?

vidhanshah:
I think after @freben approves it, we should be good to merge this one? Right?

PR #31353, opened 2025-10-07
NFS: fixed Not found page when a page extension is mounted on /

vinzscam:

This PR fixes an issue in where the `NotFound` page did not render correctly when a page extension was mounted at `/`.
As a side effect, the `renderInTestApp` utility function showed some inconsistency regarding the new routing system when `extensions` are passed as options. For this reason this PR introduce a new `renderTestApp` utility function which replaces the use of `renderInTestApp` when the `extensions` option is needed.

### DIFF
--- a/packages/frontend-plugin-api/src/blueprints/SwappableComponentBlueprint.test.tsx
+++ b/packages/frontend-plugin-api/src/blueprints/SwappableComponentBlueprint.test.tsx
@@ -60,7 +60,7 @@ describe('SwappableComponentBlueprint', () => {
       ],
     });
 
-    expect(await screen.findByText('test!')).toBeInTheDocument();
+    await expect(screen.findByText('test!')).resolves.toBeInTheDocument();
   });
 
   it('should render a component ref without a default implementation', async () => {
@@ -80,7 +80,9 @@ describe('SwappableComponentBlueprint', () => {
       ],
     });
 
-    expect(await screen.findByTestId('test.component')).toBeInTheDocument();
+    await expect(
+      screen.findByTestId('test.component'),
+    ).resolves.toBeInTheDocument();
   });
 
   it('should render a component ref with an async loader implementation', async () => {
@@ -103,7 +105,7 @@ describe('SwappableComponentBlueprint', () => {
       ],
     });
 
-    expect(await screen.findByText('test!')).toBeInTheDocument();
+    await expect(screen.findByText('test!')).resolves.toBeInTheDocument();
   });
 
   it('should render a component ref with an async loader implementation and prop transform', async () => {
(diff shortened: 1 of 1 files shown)


.changeset/real-seals-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/frontend-test-utils': patch
+---
+
+Add a new utility `renderTestApp` to `@backstage/frontend-test-utils` that simplifies rendering extensions within the Backstage application context for testing purposes. This utility replaces the use of `renderInTestApp` when the `extensions` option is needed.
```
freben: Is it worth being a bit more noisy in this message? Maybe even make it `minor` and say that the behaviour of `renderInTestApp` has changed and if your tests started failing you may want to try the other one or may need to adjust your test?

I was wondering if the way the code changed could lead to a small risk that tests that did NOT use `extensions` could start failing? Since whatever the `layout` and `routes` extensions contributed to the rendered output now is gone. Is there a risk that tests that did use routes but forgot to put in a `MemoryRouter` now need to fix that?

packages/frontend-plugin-api/src/blueprints/SwappableComponentBlueprint.test.tsx
```diff
@@ -48,44 +48,39 @@ describe('SwappableComponentBlueprint', () => {
...
+              path: '/',
               loader: async () => <TestComponent hello="test!" />,
             }),
         }),
       ],
-      initialRouteEntries: ['/test'],
     });
 
-    await waitFor(() => expect(screen.getByText('test!')).toBeInTheDocument());
+    expect(await screen.findByText('test!')).toBeInTheDocument();
```
freben:
```suggestion
    await expect(screen.findByText('test!')).resolves.toBeInTheDocument();
```

just as a general pattern (here and below) is recommended since it gives better behavior and error messages

packages/frontend-test-utils/src/app/renderTestApp.tsx
```diff
@@ -0,0 +1,115 @@
...
+    features,
+    config: ConfigReader.fromConfigs([
+      {
+        context: 'render-config',
+        data: options?.config ?? DEFAULT_MOCK_CONFIG,
+      },
+    ]),
+  });
+
+  return render(
```
freben, line 112: I'm almost tempted to suggest removing the return, which _forces_ people to use `screen` instead ;) But that's probably a bit harsh, and there's probably some complexities with `rerender` or similar to solve then. So let's not

Rugvip: Yep rerender, and sometimes it's nice to have access to the base element to verify via DOM APIs and whatnot

.changeset/real-seals-dress.md
Rugvip: You mean make this a breaking change by removing the `extensions` option completely? Honestly it's probably worth it, because currently we're kinda breaking `extensions` partially anyway. I'd say just remove it + `minor` and `**BREAKING**`

vinzscam: I like it 👍

PR #31356, opened 2025-10-08
feat(actions): add authorize callback for actions

drodil:

this allows passing optional authorize callback during action registration that can be used to limit visibility of the actions based on permissions or other conditions.

relates to #31036 and is optional implementation for #31039

### DIFF
--- a/packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.test.ts
+++ b/packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.test.ts
@@ -21,11 +21,36 @@ import express from 'express';
 import { DefaultActionsRegistryService } from './DefaultActionsRegistryService';
 import { mockServices } from '@backstage/backend-test-utils';
 import { MiddlewareFactory } from '../../../entrypoints/rootHttpRouter';
+import { ConfigReader } from '@backstage/config';
+import { catalogLocationAnalyzePermission } from '@backstage/plugin-catalog-common/alpha';
 
 describe('DefaultActionsRegistryService', () => {
   let service: DefaultActionsRegistryService;
   const mockLogger = mockServices.logger.mock();
   const mockHttpAuth = mockServices.httpAuth.mock();
+  const permissions = mockServices.permissions.mock({
+    authorize: jest.fn().mockResolvedValue([{ result: AuthorizeResult.DENY }]),
+  });
+  const config = new ConfigReader({
+    backend: {
+      actions: {
+        actionConfig: {
+          'test-plugin:configured-action': {
+            permissions: [
+              {
+                name: 'catalog.location.analyze',
+              },
+            ],
+          },
+          'test-plugin:skip-permission-action': {
+            permissions: {
+              disabled: true,
+            },
+          },
+        },
+      },
+    },
+  });
   const mockAuth = mockServices.auth.mock();
   let mockMetadata: jest.Mocked<PluginMetadataService>;
   let app: express.Application;
@@ -40,6 +65,8 @@ describe('DefaultActionsRegistryService', () => {
       httpAuth: mockHttpAuth,
       auth: mockAuth,
       metadata: mockMetadata,
+      config,
+      permissions,
     });
 
     const middleware = MiddlewareFactory.create({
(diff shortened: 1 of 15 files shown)


packages/backend-plugin-api/src/alpha/ActionsRegistryService.ts
```diff
@@ -47,6 +57,9 @@ export type ActionsRegistryActionOptions<
     idempotent?: boolean;
     readOnly?: boolean;
   };
+  authorize?: (
+    context: ActionsRegistryAuthorizeContext<TInputSchema>,
+  ) => Promise<DefinitivePolicyDecision>;
```
Rugvip, line 62: Yep thinking this is pretty good although it very much raises a question about configurability. As in what parts are hardcoded as part of the action vs configurable by adopters? To some extent I think the most important aspect is to be able to forward permissions so that they're available upfront. For example if you don't have catalog read permission you should not see any catalog read actions, etc.

Could you perhaps add an example somewhere of what that looks like in practice? For example in one of the tests.

packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.ts
```diff
@@ -64,27 +66,65 @@ export class DefaultActionsRegistryService implements ActionsRegistryService {
...
+      try {
+        const decision = await action.authorize({ credentials, input });
+        return decision.result === AuthorizeResult.ALLOW;
+      } catch (err) {
+        this.logger.error(`Failed to authorize action ${actionId}`, err);
+        return false;
+      }
+    };
+
+    router.get('/.backstage/actions/v1/actions', async (req, res) => {
```
Rugvip, line 187: Don't think this will always be the way that actions are accessed. I wonder if it might be best to have the permission be returned in the metadata instead? For example just having an `authorized: boolean` field in the response per action, leaving it to the caller to decide how to represent actions that the user doesn't have access to.

packages/backend-plugin-api/src/alpha/ActionsRegistryService.ts
drodil: Agree, there is not much configurability for adopters. Are there any examples where permissions would be handled via config? I think just having an array of action IDs and required permissions is not enough 🤔 

But I added a test to show this case

packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.ts
drodil: True. Added the `authorized` field to the response and filter based on that in mcp and scaffolder.

packages/backend-plugin-api/src/alpha/ActionsRegistryService.ts
drodil: Not sure having permissions in config is the best way? Or should there be extension point instead?

Jdubrick:
After the SIG call today I took a look at this PR and believe it addresses the auth questions myself and @johnmcollier had during the call. The only follow up I have is what (if any) plan is there to add this authorize callback to existing Backstage actions, such as those located in https://github.com/backstage/backstage/tree/master/plugins/catalog-backend/src/actions? 

Is it expected that existing Backstage supported actions don't implement this proposed callback and always authorize, and if a downstream consumer wants to implement it they need to add a wrapper around? I know that question raises in my head as @johnmcollier have work that overlaps with some of your plugins and plan on contributing to them but we do need auth capabilities too.

With that being said is there a timeline on this PR / any outstanding contributions that need to be made to push it to the finish line? @Rugvip

packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.test.ts
```diff
@@ -0,0 +1,438 @@
...
+      config: mockServices.rootConfig(),
+    });
+
+    app = express();
+    app.use(service.createRouter());
+    app.use(middleware.error());
+  });
+
+  describe('authorize callback functionality', () => {
+    it('should be abole to authorize using permissions', async () => {
```
schultzp2020, line 83:
```suggestion
    it('should be able to authorize using permissions', async () => {
```

.changeset/tricky-deer-serve.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/backend-plugin-api': patch
```
schultzp2020, line 2: Don't forget to add or update the changeset!

packages/backend-plugin-api/src/alpha/ActionsRegistryService.ts
iamEAP: Mmm-- @Rugvip are you thinking instead of an `authorize` _method_, just a `permissions` array that can be evaluated where `authorize()` is being called now?

PR #31357, opened 2025-10-08
Exclude suspend users gh enterprise

### LINKED ISSUE #31336: Option to exclude suspended users from the catalog in GitHub Enterprise instances
### 📜 Issue Labels

### 🔎 Search Terms

```plain
"Github enterprise", "exclude suspended users", "suspendedAt"
```

### 🗃️ Project Area

Catalog

### 🔖 Need

Currently, it's not possible to exclude from the catalog users that have been suspended in GitHub Enterprise instances, since the "suspendedAt" property is not part of the [default query](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend-module-github/src/lib/github.ts#L147) or the [User Entity](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend-module-github/src/lib/defaultTransformers.ts#L65) emitted by the default transformer.

[GitHub recommends](https://docs.github.com/en/enterprise-server@3.17/admin/managing-accounts-and-repositories/managing-users-in-your-enterprise/suspending-and-unsuspending-users) that users are suspended and not deleted in Enterprise instances. This is also the default behavior of some identity providers’ integrations, like Okta in our case.

Image

### 📝 Proposal

Introduce a new setting in the GitHub catalog provider allowing the users to control this behavior from their `app-config.yaml`.

This setting, when set to true, would add the "suspendedAt" property to the relevant GraphQL queries. Then, we could either:

- Return `undefined` from the default transformer if there's a "suspendedAt" value, excluding the user from the catalog.

or

- Add a new annotation to these users, allowing them to be excluded as needed in a custom transform.

Currently, no use case comes to mind for the latter.

#### Notes

The "suspendedAt" property does not exist in the `user` object of none Enterprise GitHub versions(cloud). This should not be a problem, since we can use the new config to include or not this property in the query. I already did a small proof of concept that runs against an Enterprise and Cloud instances and it works fine.

Enterprise docs: https://docs.github.com/en/enterprise-server@3.17/graphql/reference/objects#user
Cloud docs: https://docs.github.com/en/graphql/reference/objects#user 

### 🔄 Alternatives

_No response_

VDVsx:

Introduce new option in the GitHub catalog plugin to exclude suspended users.

This PR Introduces an optional setting to exclude suspended users from GitHub Enterprise instances. When it’s set to true, suspended users won’t be emitted by the default transform. If a custom transformer is used, it should check if the property `suspendedAt` from the `GithubUser` is defined, in order to exclude such users. 
The same logic was not introduced in the `GithubMultiOrgReaderProcessor.ts`, since the usage there is marked as deprecated.

This configuration option should be only used against GitHub Enterprise instances. The property does not exist in the github.com GraphQL schema. Adding it will cause a schema validation error, and the syncing of users will fail.

Closes #31336

### DIFF
--- a/plugins/catalog-backend-module-github/src/lib/defaultTransformers.ts
+++ b/plugins/catalog-backend-module-github/src/lib/defaultTransformers.ts
@@ -62,6 +62,9 @@ export const defaultUserTransformer = async (
   item: GithubUser,
   _ctx: TransformerContext,
 ): Promise<UserEntity | undefined> => {
+  if (item.suspendedAt) {
+    return undefined;
+  }
   const entity: UserEntity = {
     apiVersion: 'backstage.io/v1alpha1',
     kind: 'User',
--- a/plugins/catalog-backend-module-github/src/lib/github.test.ts
+++ b/plugins/catalog-backend-module-github/src/lib/github.test.ts
@@ -166,6 +166,54 @@ describe('github', () => {
         getOrganizationUsers(graphql, 'a', 'token'),
       ).resolves.toEqual(output);
     });
+
+    it('reads members excluding suspended users', async () => {
+      const input: QueryResponse = {
+        organization: {
+          membersWithRole: {
+            pageInfo: { hasNextPage: false },
+            nodes: [
+              {
+                login: 'a',
+                name: 'b',
+                bio: 'c',
+                email: 'd',
+                avatarUrl: 'e',
+                suspendedAt: '2025-01-01',
+              },
+              {
+                login: 'a',
+                name: 'b',
+                bio: 'c',
+                email: 'd',
+                avatarUrl: 'e',
+                suspendedAt: undefined,
+              },
+            ],
+          },
+        },
+      };
+
+      const output = {
+        users: [
+          expect.objectContaining({
+            metadata: expect.objectContaining({ name: 'a', description: 'c' }),
+            spec: {
+              profile: { displayName: 'b', email: 'd', picture: 'e' },
+              memberOf: [],
+            },
+          }),
+        ],
+      };
+
+      server.use(
+        graphqlMsw.query('users', () => HttpResponse.json({ data: input })),
+      );
+
+      await expect(
+        getOrganizationUsers(graphql, 'a', 'token', true),
+      ).resolves.toEqual(output);
+    });
   });
 
   describe('getOrganizationUsers using custom UserTransformer', () => {
@@ -226,7 +274,13 @@ describe('github', () => {
       );
 
       await expect(
-        getOrganizationUsers(graphql, 'a', 'token', customUserTransformer),
+        getOrganizationUsers(
+          graphql,
+          'a',
+          'token',
+          false,
+          customUserTransformer,
+        ),
       ).resolves.toEqual(output);
     });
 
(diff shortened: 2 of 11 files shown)


plugins/catalog-backend-module-github/src/lib/defaultTransformers.ts
```diff
@@ -62,6 +62,9 @@ export const defaultUserTransformer = async (
   item: GithubUser,
   _ctx: TransformerContext,
 ): Promise<UserEntity | undefined> => {
+  if (item.suspendedAt) {
```
benjdlambert: I'm not sure we want to add this here? 

This is the default transformer, and will always hide `suspendedAt` members right?

benjdlambert: I think rather than taking this approach, we should be filtering the list that gets passed to these transformers if the `suspendedAt` is set and `excludeSuspendedUsers` is `true`?

benjdlambert (changes requested):
Hey :wave: thanks for the PR!

Just a high level comment on the direction here.

plugins/catalog-backend-module-github/src/lib/defaultTransformers.ts
VDVsx: Hi, 

Thanks for the feedback.
I was also unsure if this is a good place. My thinking was that this could allow for more customization, since one could supply their own transformer and handle these users in different ways after having a way to detect them. If we filter before the transformer, the consumer won’t be able to access them. But we don't really have a use case for it, so I think it’s totally fine to filter this out for now.

plugins/catalog-backend-module-github/src/lib/github.test.ts
```diff
@@ -385,7 +385,7 @@ describe('github', () => {
           graphql,
           'a',
           'token',
-          true,
+          false,
```
VDVsx: Since now the filtering is done outside the original transformer, reused this test case to check that the filtering is not applied if the parameter is set to false.

plugins/catalog-backend-module-github/src/lib/defaultTransformers.ts
VDVsx: @benjdlambert Changed the code as per your suggestion. Please have a look once you have a minute and let me know if this is a better approach: https://github.com/backstage/backstage/pull/31357/commits/b137f300cd3194dd40e66bfb58870f4659735010

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -214,7 +230,7 @@ export async function getOrganizationUsers(
     query,
     org,
     r => r.organization?.membersWithRole,
```
benjdlambert: I think we should do the filtering here.

```suggestion
    r => r.organization?.membersWithRole.filter(u => excludeSuspendedUsers ? !u.suspendedAt : true),
```

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -214,7 +230,7 @@ export async function getOrganizationUsers(
...
-    userTransformer,
+    suspendedUserFilteringTransformer,
```
benjdlambert: Then we just keep this the same as before.

benjdlambert (changes requested):
Hey :wave: sorry for the delay here, been a lot of things going on.

Just a small tweak to how the filtering is implemented I think.

plugins/catalog-backend-module-github/src/lib/github.ts
VDVsx: Hi @benjdlambert,
Thanks for the suggestions. Unfortunately, the response is not a flat list of users here. It’s an object of type `Connection<GithubUser>`. The users’ objects are inside the connection nodes returned from the GraphQL API.

So I think we would need something like this instead:

```ts
r => {
      const connection = r.organization?.membersWithRole;
      if (!connection) return undefined;

      return {
        ...connection,
        nodes: connection.nodes.filter(u =>
          excludeSuspendedUsers ? !u.suspendedAt : true,
        ),
      };
    },
```

Do you prefer this instead of the transformer approach, or have any other suggestions?

Thanks for all the comments so far.

benjdlambert: AH! Ok, that API is a bit odd.

I wonder if it makes more sense to refactor the `queryWithPaging` signature into something like this then instead:

```ts
const users = await queryWithPaging({ 
  client, 
  query, 
  org, 
  selector: r => r.organization?.membersWithRole, 
  transformer: userTransformer,
  filter: u => excludeSuspendedUsers ? !u.suspendedAt : true,
  variables: {
    org,
    email: tokenType === 'token',
    organizationMembersPageSize: pageSizes.organizationMembers,
  },
});
```

There's a new filter options there which is going to use the type of the `node`, and it's also optional.

It's also a nice cleanup because I'm not a fan of functions with a load of arguments and we don't know what they are at a glance. So it cleans that up a little bit.

What do you think?

VDVsx: Yeah, that sounds like a good compromise. It’s much clearer for sure, and we are not hijacking the transformer anymore.
Honestly, I didn't particularly like my initial approach of modifying the transformer, but a similar approach is already used in the file (see `materialisedTeams`), so I decided to follow that to limit the changes and increase the likelihood of getting this in. 😅 

I'll test this a little bit to make sure all is still working as expect and then push the commit.

Thanks again for the suggestions.

VDVsx: The discussed changes are in this commit: https://github.com/backstage/backstage/pull/31357/commits/268b0f99549f637b665943c1e8731426f7b942f5

benjdlambert:
Thanks! :tada:

PR #31365, opened 2025-10-08
Remove MUICssBaseline

cdedreuille:

As we are transitioning from MUI to BUI the default CSS baseline from MUI was conflicting with our new configuration in BUI. Since we are using the same `normalize.css` under the hood. This baseline can be safely removed. The only part that was not set is the body background color.

In Backstage UI we are going to move all our CSS including our reset and core elements inside layers. Since MUI doesn't use layers, those styles were used over BUI styles, creating a lot of conflicts.

### DIFF
--- a/.changeset/warm-dancers-call.md
+++ b/.changeset/warm-dancers-call.md
@@ -2,4 +2,10 @@
 '@backstage/theme': minor
 ---
 
-**BREAKING**: Removed noCssBaseline prop in UnifiedThemeProvider. If your Backstage instance looks broken after this update, there's a high chance that you forgot to add our new Backstage UI global CSS. To do that, please add @backstage/ui/css/styles.css at the root of your application.
+**BREAKING**: Removed the built-in `CssBaseline` from `UnifiedThemeProvider`. If your Backstage instance looks broken after this update, you likely forgot to add our new Backstage UI global CSS. To do that, please import `@backstage/ui/css/styles.css` in `packages/app/src/index.tsx`:
+
+```tsx
+import '@backstage/ui/css/styles.css';
+```
+
+This change also removes the `noCssBaseline` prop, which became redundant.


.changeset/warm-dancers-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/theme': minor
+---
+
+**BREAKING**: Removed noCssBaseline prop in UnifiedThemeProvider.
```
Rugvip: I assume this is replaced with any resets in the BUI CSS? and that there may be issues if that's not present in the app?

If that's the case I'm thinking we should call out clearly that people need to add the BUI CSS to their projects, if that's not done already. It's another good kinda weak way to ensure that everyone is ready for BUI components.

cdedreuille: Good call. I'll update the changeset.

cdedreuille: This is now updated @Rugvip

.changeset/warm-dancers-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/theme': minor
+---
+
+**BREAKING**: Removed noCssBaseline prop in UnifiedThemeProvider. If your Backstage instance looks broken after this update, there's a high chance that you forgot to add our new Backstage UI global CSS. To do that, please add @backstage/ui/css/styles.css at the root of your application.
```
Rugvip:
`````suggestion
**BREAKING**: Removed the built-in `CssBaseline` from `UnifiedThemeProvider`. If your Backstage instance looks broken after this update, you likely forgot to add our new Backstage UI global CSS. To do that, please import `@backstage/ui/css/styles.css` in `packages/app/src/index.tsx`:

```tsx
import '@backstage/ui/css/styles.css';
```

This change also removes the `noCssBaseline` prop, which became redundant.
`````

The change to highlight is the fact that we removed `CssBaseline`. Also providing a code block for easy copy-paste.

cdedreuille: Updated!

PR #31371, opened 2025-10-09
BUI - New Dialog component

ssjoblad:

Adding new `Dialog` component to Backstage UI which includes a header, body, footer and closer. 

scrnli_YdcNLYf5Yo01yF


packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,146 @@
...
+            : undefined,
+        }}
+        ref={ref}
+        {...props}
+      >
+        <ScrollArea.Viewport>{children}</ScrollArea.Viewport>
+        <ScrollArea.Scrollbar orientation="vertical">
+          <ScrollArea.Thumb />
+        </ScrollArea.Scrollbar>
+      </ScrollArea.Root>
```
Rugvip: Perhaps worth implementing this differently given that we plan to remove the scroll area? Could leave for later too

packages/ui/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,206 @@
...
+      <DialogTrigger {...args}>
+        <Button variant="secondary">Open Dialog</Button>
+        <Dialog>
+          <DialogHeader>Example Dialog</DialogHeader>
+          <DialogBody>
+            <Text>This is a basic dialog example.</Text>
+          </DialogBody>
+          <DialogFooter>
+            <DialogClose>Close</DialogClose>
+            <DialogClose variant="primary">Save</DialogClose>
```
Rugvip: Hmm, how are you intended to tell these apart?

packages/ui/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,146 @@
...
+  return (
+    <div ref={ref} className={clsx(classNames.footer, className)} {...props}>
+      {children}
+    </div>
+  );
+});
+DialogFooter.displayName = 'DialogFooter';
+
+/** @public */
+export const DialogClose = forwardRef<
```
Rugvip: Header body and footer make sense to me, but I'm not feeling the close button necessarily provides that much value? Wanna make sure we don't bloat the API unnecessarily

cdedreuille: It is quite necessary as this is a special button that will do actions as well as close the dialog safely.

packages/ui/src/components/Dialog/Dialog.stories.tsx
cdedreuille: You're right. I removed `DialogClose`.

packages/ui/src/components/Dialog/Dialog.tsx
cdedreuille: Actually you're right. I removed `DialogClose` as it was not directly clear and added more complexity. Now you can use `Button` directly and if you want to close the dialog you can just add the `slot="close"` to the button to make it work.

packages/ui/src/components/Dialog/Dialog.tsx
cdedreuille: 100%. I removed the scroll area in this PR.

packages/ui/css/styles.css
```diff
@@ -10838,6 +10838,129 @@
...
+  display: flex;
+}
+
+.bui-DialogBody {
+  padding: var(--bui-space-3);
+  flex: 1;
+  overflow-y: auto;
+}
+
+@keyframes fade-in {
```
Rugvip, line 10920: Separate discussion, but should we namespace these as well @cdedreuille?

PR #31380, opened 2025-10-10
feat(catalog): add query entities action to registry

drodil:

adds new query-catalog-entities action to the actions registry

### DIFF
--- a/plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
+++ b/plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2025 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { ActionsRegistryService } from '@backstage/backend-plugin-api/alpha';
+import { CatalogService } from '@backstage/plugin-catalog-node';
+import { QueryEntitiesRequest } from '@backstage/catalog-client';
+
+export const createQueryCatalogEntitiesAction = ({
+  catalog,
+  actionsRegistry,
+}: {
+  catalog: CatalogService;
+  actionsRegistry: ActionsRegistryService;
+}) => {
+  actionsRegistry.register({
+    name: 'query-catalog-entities',
+    title: 'Query Catalog Entities',
+    attributes: {
+      destructive: false,
+      readOnly: true,
+      idempotent: true,
+    },
+    description: `
+This allows you to query multiple entities from the software catalog using filters.
+You can filter by entity properties like kind, namespace, metadata fields, and more.
+Using limit is recommended to avoid fetching too many entities at once. You can use the cursor returned in the response to fetch next set of entities.
+The action returns a list of entities matching the specified criteria.
+Fields are using dot notation, e.g. "kind", "metadata.name", "spec.type" in the filter, fields, orderFields and fullTextFilter.fields parameters.
+    `,
+    schema: {
+      input: z =>
+        z.object({
+          filter: z
+            .record(z.string())
+            .optional()
+            .describe(
+              'Filter criteria for querying entities. Keys are field paths and values are the expected values.',
+            ),
+          fields: z
+            .array(z.string())
+            .optional()
+            .describe(
+              'Specific fields to include in the response. If not provided, all fields are returned.',
+            ),
+          limit: z
+            .number()
+            .int()
+            .positive()
+            .optional()
+            .describe('Maximum number of entities to return.'),
+          offset: z
+            .number()
+            .int()
+            .min(0)
+            .optional()
+            .describe('Number of entities to skip before returning results.'),
+          orderFields: z
+            .object({
+              field: z.string().describe('Field to order by'),
+              order: z
+                .enum(['asc', 'desc'])
+                .optional()
+                .default('asc')
+                .describe('Sort order'),
+            })
+            .optional()
+            .describe('Ordering criteria for the results.'),
+          fullTextFilter: z
+            .object({
+              term: z.string().describe('Full text search term'),
+              fields: z
+                .array(z.string())
+                .optional()
+                .describe('Fields to search within'),
+            })
+            .optional()
+            .describe('Full text search criteria'),
+          cursor: z
+            .string()
+            .optional()
+            .describe(
+              'Cursor for pagination. This can be used only after first request with response containing a cursor',
+            ),
+        }),
+      output: z =>
+        z.object({
+          items: z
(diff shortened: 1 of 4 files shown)


plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,144 @@
...
+      output: z =>
+        z.object({
+          items: z
+            .array(z.object({}).passthrough())
+            .describe('List of entities'),
+          totalItems: z.number().describe('Total number of entities'),
+          pageInfo: z
+            .object({
+              nextCursor: z.string().optional().describe('Next page cursor'),
+              prevCursor: z
```
freben: Probably doesn't make sense to be able to paginate backwards at all in this context.

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,144 @@
...
+        }),
+      output: z =>
+        z.object({
+          items: z
+            .array(z.object({}).passthrough())
+            .describe('List of entities'),
+          totalItems: z.number().describe('Total number of entities'),
+          pageInfo: z
+            .object({
+              nextCursor: z.string().optional().describe('Next page cursor'),
```
freben: I find this interesting. Will a tool be smart enough to know that it needs to iterate through the collection if it wants a full result set? Will it be smart enough to COMPOSE things in such a way that it paginates and does the downstream treatment of pages one by one too? Have you experimented with this?

The reason I'm asking is, of course, whether pagination should be exposed at this level at all. Or is it better to model it more as

* always returns a full result set
* leverages the catalog's internal ability to stream out results without getting bogged down doing so
* has an (optional!) `limit` parameter that's conservatively set to for example 1000 that is prominently explained and can be safely overridden to anything, including "infinity" (like a -1 or so)
* has a `itemsTruncated: boolean` or so in the response with a clear description explaining that there were more items but the `limit` parameter needs to be set higher or to -1 to get all the data

This is meant as an open question rather than directions. I'm truly curious what the "right" design is for an endpoint like this that is meant to be usable by MCP etc as well. Do you have any thoughts or references to other similar endpoints out in the wild?

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
drodil: Yes, will remove.

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
drodil: I've used some MCP servers with tools that have pagination, and it works pretty well actually (with copilot). The tool description of course has to tell that pagination is recommended. What comes to the streaming, I don't see it helps a lot at this point as the response from actions registry does not support streaming so the whole patch of entities is returned there anyway. 

This same conversation applies to #31010

niallthomson: I generally like to see pagination in MCP server tools as otherwise the context can get blown out pretty easily. My experience has been that models process pagination tools appropriately.

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,144 @@
...
+            .optional()
+            .describe('Maximum number of entities to return.'),
+          offset: z
+            .number()
+            .int()
+            .min(0)
+            .optional()
+            .describe('Number of entities to skip before returning results.'),
+          orderFields: z
+            .object({
```
freben, line 70: The API allows this to be an array. Do we want to disallow that here? Otherwise we could allow both, or always make it an array (happy for your input on what's best there)

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,144 @@
...
+You can filter by entity properties like kind, namespace, metadata fields, and more.
+Using limit is recommended to avoid fetching too many entities at once. You can use the cursor returned in the response to fetch next set of entities.
+The action returns a list of entities matching the specified criteria.
+Fields are using dot notation, e.g. "kind", "metadata.name", "spec.type" in the filter, fields, orderFields and fullTextFilter.fields parameters.
+    `,
+    schema: {
+      input: z =>
+        z.object({
+          filter: z
+            .record(z.string())
```
freben, line 46: No way of querying for existence? (omitting value)

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,144 @@
...
+
+      const response = await catalog.queryEntities(queryRequest, {
+        credentials,
+      });
+
+      return {
+        output: {
+          items: response.items,
+          totalItems: response.totalItems,
+          hasMoreEntities: response.totalItems > response.items.length,
```
freben, line 138: Hm. Does this really make sense? Isn't it rather that if there's no cursor we're done? I'm not entirely sure that we'll always know for sure that we get exactly `limit` items back, and also, isn't totalItems potentially vastly larger than the page size anyway (ie it's the number of total hits after going through all pages)?

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,144 @@
...
+                .optional()
+                .describe('Fields to search within'),
+            })
+            .optional()
+            .describe('Full text search criteria'),
+          cursor: z
+            .string()
+            .optional()
+            .describe(
+              'Cursor for pagination. This can be used only after first request with response containing a cursor',
```
johnmcollier, line 94: How do we enforce that the cursor can only be used after the first request with a cursor? Or is the description sufficient here to prevent (most) models from erroneously trying to set it?

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,144 @@
...
+            ),
+        }),
+      output: z =>
+        z.object({
+          items: z
+            .array(z.object({}).passthrough())
```
johnmcollier, line 100: Not sure if it'll necessarily work when the filter returns only a subset of fields, but if you want, you can also specify the entity type with `z.custom<Entity>()` instead of `z.object`

johnmcollier:
Hey @drodil Just curious, have you tested on a Backstage instance with both this tool and the tool from https://github.com/backstage/backstage/pull/31010 configured? Did you notice any instances where the model tried to call the wrong tool? (e.g. trying to query entities and it calls the search tool instead)

My team worked on a simliar set of MCP tools as you (https://github.com/redhat-developer/rhdh-plugins/tree/main/workspaces/mcp-integrations), and we had issues with our techdoc search tool being called when we just wanted to query certain entities. Better / more detailed descriptions helped a lot in our case, but we also found it came down to the quality of model used.

drodil:
@johnmcollier yeah, I've tested both of these and indeed it's more up to the user's prompt to call the right tool. I haven't got wrong tool calls really, just the model not calling any tool at all and just making stuff up. Don't know if making the tool description more accurate and informal would help the model to pick the right one.

freben:
Still some changes to make on this one right?

drodil:
Not stale, trying to find time to finish this

johnmcollier:
@drodil I'm happy to help / lend a hand if there's anything you need help with

PR #31382, opened 2025-10-10
fix: fix template parsing issue preventing proper function of RJSF allOf / oneOf / anyOf with nested if / else fields

### LINKED ISSUE #30090: Conditional fields using allOf and if/then don't work in Backstage templates
### 📜 Issue Labels

### 🔎 Search Terms

```plain
conditional fields allOf anyOf oneOf
```

### 🗃️ Project Area

Software Templates

### 🔗 External Integration

N/A

### 📝 Description & Context

I'm trying to use `allOf` with `if/then` logic to define conditional fields in a Software Template's parameters block.

This schema works perfectly in the [RJSF playground](https://rjsf-team.github.io/react-jsonschema-form/) — all conditionally-rendered fields appear correctly based on the selected value of conditionalField.

However, when used in Backstage, only the top-level field (conditionalField) is rendered. None of the fields defined inside then.properties blocks under allOf show up.

In my Backstage app, even though the template is valid, the conditional fields never appear. It seems like the Backstage Scaffolder either doesn't evaluate `allOf` at the root of a `parameters[]` object, or fails to pass it through properly to RJSF.

If I switch to using dependencies, some conditionals work — but using dependencies isn't ideal since it lacks flexibility and is harder to nest. From what I can tell, this may be a limitation of how Backstage builds and passes schemas into `@rjsf/core`.

See screenshots attached:

The expected behavior (RJSF playground)

Image

One shows the broken behavior (Backstage template form)

Image

### 👍 Expected Behavior

When selecting "a" or "b" from conditionalField, the appropriate child fields (visibleIfA, visibleIfB, etc.) should dynamically appear.

**Actual Behavior in Backstage**
Only the conditionalField is rendered. None of the conditional fields (visibleIfA, visibleIfB, etc.) appear regardless of the selected value.

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

_No response_

iainvdw:
A fellow coworker once opened an issue about RJSF's `allOf` / `oneOf` / `anyOf` properties not working when they're defined in the root of a Scaffolder template step. I've done some digging since we need those features to work soon, I've actually managed to pinpoint the cause of the problem, which is located in the Scaffolder's template parsing when loading the schema in the Scaffolder's `Stepper` component.

The cause of the issue is in the Scaffolder's frontend plugin, `plugin-scaffolder-react` where it loads a template from the catalog and parses it. The parsing is done in the `useTemplateSchema` hook where it strips a title for every one of the steps, since the step is also shown in the `Stepper` component at the top of the page. 

However, when stripping the title, it is explicitly set to `undefined` which seemingly causes an issue with one of RJSF's internal validators. This invalidates (certain parts of) the schema for the template and causes `allOf` / `oneOf` / `anyOf` properties with nested `if` / `then` / `else` properties defined at the root of a step to not render properly anymore when one of the fields it depends on would be updated in the form.

See this gist for the source template yaml file, the JSON manifest equivalent of the schema and a test component:
https://gist.github.com/iainvdw/38aa5f69b762a3b15b8ba13ab64250cd

Fixes #30090.

- [x] Tests for new functionality and regression tests for bug fixes.

### DIFF
--- a/.changeset/grumpy-planes-bet.md
+++ b/.changeset/grumpy-planes-bet.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-react': patch
+---
+
+Fixed a bug in the Scaffolder's template parsing in the `useTemplateSchema` hook by removing the title instead of setting it to `undefined`


schultzp2020 (approved):
LGTM! 🛶

PR #31386, opened 2025-10-10
Improve 'OIDC From Scratch' documentation

chriskilding-relx:

Improve the OIDC From Scratch docs, so that they make it more obvious that:

- you should re-badge Backstage's generic OIDC provider with your OIDC implementation's vendor name and branding (so that users recognise it)
- you can use the default `oidc` provider rather than defining boilerplate code for your own resolver etc.

### DIFF
--- a/docs/auth/oidc.md
+++ b/docs/auth/oidc.md
@@ -1,7 +1,7 @@
 ---
 id: oidc
 title: OIDC provider from scratch
-description: This section shows how to use an OIDC provider from scratch, same steps apply for custom providers.
+description: This section shows how to enable and use the Backstage OIDC provider.
 ---
 
 :::info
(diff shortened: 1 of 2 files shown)


chriskilding-relx:
The one part I have not updated yet is using a custom resolver (as I haven't tried this in my own setup yet - I've been using the new built-in `oidc` provider to reduce boilerplate).

Need to update that part of the docs before they can be considered for merging.

chriskilding-relx:
This part of the docs also relates to https://github.com/backstage/backstage/pull/31310 which I'm currently working on

docs/auth/oidc.md
```diff
@@ -11,78 +11,67 @@ system, you may want to read [its own article](https://github.com/backstage/back
 instead, and [consider migrating](../backend-system/building-backends/08-migrating.md)!
 :::
 
-This section shows how to use an OIDC provider from scratch, same steps apply for custom
-providers. Please note these steps are for using a provider, not how to implement one,
-and Backstage recommends creating custom providers specific to the IDP, so we'll use a
-`azureOIDC` provider throughout this example, feel free to change any of those refs
-to your provider name.
+This section shows how to use an OIDC provider from scratch. The same steps apply for custom providers.
```
freben: Do you have a suggestion for a better wording than "use from scratch"? I think the original formulation was a bit unfortunate. "make from scratch" I could understand, but that's called out as NOT being the purpose below.

Maybe "the full process of ..." or "...details the steps needed to..." or something like that?

docs/auth/oidc.md
```diff
@@ -11,78 +11,67 @@ system, you may want to read [its own article](https://github.com/backstage/back
...
-To add providers not enabled by default like OIDC, we need to follow some steps, we
-assume you already have a sign-in page to which we'll add the provider so users can
-sign in through the provider. In simple steps here's how you enable the provider:
+OIDC is a protocol which has numerous implementations. It's likely that many of your users won't know what the OIDC **protocol** is, but they will recognise your OIDC **implementation**. Backstage supplies a generic `oidc` authorization strategy. You should re-badge this with the name and branding of your OIDC implementation, so that your users will recognise it on the Backstage sign-in page.
+
+For example, if your organization uses [Keycloak](https://www.keycloak.org), you would re-badge the OIDC provider as `Keycloak` and tell users to `Sign In using Keycloak`.
+
+## Steps
+
+The Backstage OIDC provider is not enabled by default. You need to manually enable the OIDC provider, and tell it which OIDC server you want to use.
```
freben:
```suggestion
The Backstage OIDC provider is not enabled by default. You need to manually enable the provider, and tell it which OIDC server you want to use.
```

just to reduce repetition a little

docs/auth/oidc.md
```diff
@@ -11,78 +11,67 @@ system, you may want to read [its own article](https://github.com/backstage/back
...
-Please note a few things, the ID can be anything you want as long as it doesn't conflict
-with other refs, backstage recommends to use a custom name that references your custom
-provider, for example we are using OIDC protocol with Azure, so we could use something
-like `auth.azure.oidc` as well.
+The `id` of the API ref can be anything you want, as long as it doesn't conflict with other refs. Backstage recommends to use a custom name that references your custom provider.
 
-Also we're exporting this reference, as well as the `typings`, we need to
-be able to import this reference anywhere in the app, and the `typings` will tell typescript
+:::note TypeScript Note
+As we're exporting this API reference, as well as the `typings`, we need to
```
freben:
```suggestion
As we're exporting this API reference, as well as the TypeScript types, we need to
```

docs/auth/oidc.md
```diff
@@ -11,78 +11,67 @@ system, you may want to read [its own article](https://github.com/backstage/back
...
+be able to import this reference anywhere in the app. The `typings` will tell TypeScript
```
freben:
```suggestion
be able to import this reference anywhere in the app. The types will tell TypeScript
```

docs/auth/oidc.md
```diff
@@ -125,27 +117,14 @@ export const apis: AnyApiFactory[] = [
...
-
-The Auth Provider is responsible for authenticating with the 3rd party service, and give
-us back the credentials, here's where you pick which protocol to use, be it Auth0, OAuth2,
-OIDC, SAML or any other that your 3rd party IDP provider supports.
-
 ### The Resolver
 
-Resolvers exist to map user identity from the 3rd party (in this case an azure IDP
-provider) to the backstage user identity.
+Resolvers exist to map user identity from the 3rd party (in this case Keycloak) to the Backstage user identity.
```
freben:
```suggestion
Resolvers exist to map the user identity from the 3rd party (in this case Keycloak) to the Backstage user identity.
```

docs/auth/oidc.md
```diff
@@ -215,76 +194,87 @@ backend.add(myAuthProviderModule);
...
 - `clientId`: Grab from the Overview page.
 - `clientSecret`: Can only be seen when creating the secret, if you lose it you'll need a
   new secret.
 - `metadataUrl`: In Overview > Endpoints tab, grab OpenID Connect metadata document URL.
+
+The OIDC provider **also** requires the `auth.session.secret` to be set.
+
+#### Optional Parameters
+
+These parameters have implicit default values. Don't override them unless you know what you're doing.
```
freben, line 233: I don't think this is true for all of these. In particular the first point in the list I think is actually mandatory.

docs/auth/oidc.md
```diff
@@ -215,76 +194,87 @@ backend.add(myAuthProviderModule);
...
+  {
+    id: 'keycloak-auth-provider',
+    title: 'Keycloak',
+    message: 'Sign In using Keycloak',
+    apiRef: keycloakAuthApiRef,
+  },
+];
+```
+
+Otherwise, please follow the [Sign In Configuration][3] docs to import and use the Keycloak API reference we defined earlier.
```
freben: What does the "Otherwise" signify here?

docs/auth/oidc.md
chriskilding-relx: Assuming you mean the URLs, we're using OIDC auto-discovery in our installation, so we simply define the metadata URL, and Backstage obtains the authorisation URL and token URL automatically from the JSON response.

docs/auth/oidc.md
chriskilding-relx: Agreed, I've changed it to simply say

> This section shows how to enable and use the Backstage OIDC provider.

docs/auth/oidc.md
chriskilding-relx: Tweaked this section to assume that the reader is using the standard SignInPage component (which, realistically, the majority probably are).

If they're doing some other custom sign-in page, well, they know they're off the standard path anyway so we don't need to provide an 'otherwise' in the docs.

awanlin (approved):
Thanks @chriskilding-relx, really appreciate these improvements 🚀

PR #31392, opened 2025-10-12
Improve GitLab API utilization to improve performance, reduce rate limiting

kc9ddi:

perf(gitlab): eliminate project ID resolution API calls to prevent rate limiting

Replace project ID lookup API calls with direct project path usage in GitLab
URL reader. This optimization leverages GitLab API v4's native support for
URL-encoded project paths in endpoints, eliminating the need for preliminary
/api/v4/projects/{path} calls that were causing rate limiting issues.

Changes:

Update getGitLabFileFetchUrl to use project paths directly (now synchronous)
Add extractProjectPath utility for consistent path extraction from URLs
Modify buildProjectUrl to accept both project paths and IDs
Optimize getGitlabArtifactFetchUrl to use project paths for artifact endpoints
Remove resolveProjectToId function (no longer needed)
Update all unit tests to expect project path-based URLs
This eliminates ~1 API call per file operation while maintaining full
compatibility with both GitLab.com and self-hosted instances.

### DIFF
--- a/.changeset/wide-showers-cry.md
+++ b/.changeset/wide-showers-cry.md
@@ -1,0 +1,11 @@
+---
+'@backstage/integration': minor
+'@backstage/backend-defaults': patch
+---
+
+Updates calls to Gitlab APIs to use a namespaced path in place of a project ID. This removes the need to call
+a separate Gitlab API to determine the project ID for a project, resulting in significantly fewer API calls
+to Gitlab, and faster catalog ingestion when syncing with Gitlab.
+
+The signature of the `getGitLabFileFetchUrl` has changed. It is no longer asynchronous, and an unnecessary
+parameter has been removed.


GabDug:
We had the same issue, currently testing this patch on our test environment. Nothing to complain so far, and we observed a x40 reduction of `/api/:version/projects/:id` calls to Gitlab. Our Gitlab guy is happy!

Thanks for the PR @kc9ddi, hope this can be merged. LMK if I can help!

kc9ddi:
Looking for review from @Rugvip - this PR would solve significant gitlab integration issues at my organization, and it sounds like others in the community are finding value here as well.

packages/integration/src/gitlab/core.ts
```diff
@@ -28,22 +27,23 @@ import {
...
- * to:   https://gitlab.com/api/v4/projects/projectId/repository/files/filepath?ref=branch
+ * to:   https://gitlab.com/api/v4/projects/groupA%2Fteams%2FteamA%2FsubgroupA%2FrepoA/repository/files/filepath/raw?ref=branch
  *
  * @param url - A URL pointing to a file
  * @param config - The relevant provider config
+ * @param _token - Unused, kept for backward compatibility
  * @public
  */
-export async function getGitLabFileFetchUrl(
+export function getGitLabFileFetchUrl(
```
Rugvip, line 39: Could you keep this one async? Removing it would be a breaking change. Alternatively mark this as a breaking change in the changeset with a `minor` bump and call out the change, and at that point you can remove the token parameter too. Actually it would be best to have a separate changeset for this change at that point.

Rugvip (commented):
Thank you! 🎉 

Big improvement indeed, just some API housekeeping

packages/integration/src/gitlab/core.ts
kc9ddi: Thank you for the feedback! I cleaned up the API and updated the changeset documentation accordingly.

kc9ddi:
Waiting on final review from @Rugvip

packages/integration/src/gitlab/core.ts
freben: It seems you did not move this back to async (returning a promise).

elaine-mattos:
Hi @kc9ddi ,

thanks a bunch! Looking forward this one!

PR #31448, opened 2025-10-16
BUI - Fix @layer order

cdedreuille:
### The Problem Explained
The CSS is being loaded from two different sources:
1. Global CSS (packages/ui/css/styles.css):
    - Contains @layer tokens { }, @layer base { }, and @layer utilities { }
    - Had the layer order declaration at the top

2. CSS Modules (*.module.css):
    - Imported individually by each React component
    - Each contained @layer components { } without the layer order declaration
    - These files were being loaded before or separately from the global CSS

### Why Layer Order Was Wrong
In CSS, when multiple stylesheets define layers, the first stylesheet to declare the layer order establishes it for the entire page. However:
- If CSS Module files loaded before styles.css, they would create the components layer first
- Without an explicit order declaration, layers are ordered by first appearance
- This caused components to appear before tokens and base in the cascade


packages/ui/src/components/Card/Card.module.css
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@layer tokens, base, components, utilities;
```
Rugvip, line 17: Is this only for BUI CSS, or would it need to be used for custom styles in plugins as well?

I'm assuming not since you generally wouldn't use layers there.

packages/ui/css/styles.css
```diff
@@ -1,3 +1,5 @@
+@layer tokens, base, components, utilities;
```
Rugvip, line 1: Random unrelated though, should these be namespaced? `bui-tokens`, etc.

packages/ui/src/components/Card/Card.module.css
cdedreuille: It's only for BUI CSS. The end user will not use `@layer` as this will always be higher priority.

packages/ui/css/styles.css
cdedreuille: We could. Not sure if that bring any benefit though as this is not something meant to be used by the end user. If other libraries are installed in some plugins (which is a bad idea) and they also use similar names, this would at least guarantee that they are placed in the right order.

Rugvip (approved):
:shipit:

PR #31452, opened 2025-10-16
microsite - Added `@backstage/cli` to allow running Prettier

awanlin:

Added `@backstage/cli` to allow running Prettier and also added `postinstall` script. I noticed if you start with a fresh clone, where you never ran `yarn install` at the root you could not run prettier for the microsite.

### DIFF
--- a/microsite/package.json
+++ b/microsite/package.json
@@ -8,6 +8,7 @@
     "deploy": "docusaurus deploy",
     "docusaurus": "docusaurus",
     "generate-openapi-docs": "yarn docusaurus clean-api-docs all && yarn docusaurus gen-api-docs all",
+    "postinstall": "cd ../ && yarn install",
     "prettier:check": "prettier --check .",
     "prettier:fix": "prettier --write .",
     "publish-gh-pages": "docusaurus-publish",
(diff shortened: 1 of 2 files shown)


Parsifal-M (approved):
Makes sense to me 😄

freben:
Wow that's some diff. Is it really for the right version of the cli? And is it really the right action to run yarn install in the _parent_ directory for it? I presume it's just to get the git hooks installed right? Then, maybe not a whole full install (which would likely waste lots of time), bur rather just the targeted hook you wanted? Or lift that installation hook over to this package.json, and can it suffice with just a dev dep on prettier itself? Not sure how far we want to take this

Parsifal-M:
> Wow that's some diff. Is it really for the right version of the cli? And is it really the right action to run yarn install in the _parent_ directory for it? I presume it's just to get the git hooks installed right? Then, maybe not a whole full install (which would likely waste lots of time), bur rather just the targeted hook you wanted? Or lift that installation hook over to this package.json, and can it suffice with just a dev dep on prettier itself? Not sure how far we want to take this

TIL 😅 - I think if we could "lift the hook" would probably be good enough but I have no clue how to do that, maybe @awanlin  does though 🙏

awanlin:
Hi @freben, that was the right CLI version at the time I created the PR. I wanted to use `workspace:^` but that gave me some issues. If you have any ideas how I can make that work happy to do so.

The `postinstall` command I actually just pulled from what we have been doing in the Community Plugins repo. I'm also happy to handle this in some other way I'm just not sure how exactly. The issue that caused me to add this was the git hook failing as I had not run `yarn install` at the root.

PR #31454, opened 2025-10-17
Version Policy Update - Node 22 to 24

awanlin:

Following our [Version Policy for Node](https://backstage.io/docs/overview/versioning-policy#nodejs-releases) this updates to add support for Node 22 to 24 in CI. The following changes were also included:

- Updated the `@types/node` version where used.
- Updated to v6 of `isolated-vm` in `@backstage/plugin-scaffolder-backend`
- Switched ECMAScript version to ES2023 in `@backstage/cli`
- Bump `@microsoft/api-documenter` and `@microsoft/api-extractor` to latest versions.

Node 24 will become an Active LTS version this month (October 2025): https://nodejs.org/en/about/previous-releases


Rugvip:
Looks like we need to bump to isolate-vm v6, since v5 doesn't seem to support node v24

.changeset/stupid-cases-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+Switched ECMAScript version to ES2024, lining up with the switch to Node.js v22 + v24.
```
Rugvip: I threw this in as well, Node.js v22+ allows us to move up to ES2024: https://node.green/#ES2024

packages/cli/src/tests/transforms/transforms.test.ts
```diff
@@ -36,18 +36,18 @@ const exportValues = {
...
       dyn: exportValues.all,
     },
-  },
-  module: {
+  }),
+  module: expect.objectContaining({
     ...exportValues.all,
     dyn: exportValues.all,
-  },
+  }),
```
Rugvip, line 50: Turns out v24 was just returning some extra exports, but that doesn't break any of our regular usage patterns

Rugvip (approved):
Hopefully good to go now 🤞

freben:
Whoa, `Error: Unexpectedly missing tree sha entry for path packages/app-defaults"` sounds like git corruption

Rugvip:
It's due to the removal of `shell: true` from all the spawn invocations. Turns out that one's a bit tricky but I think it's worth doing before we ship this. It's due to the following new warning:

```
DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
```

Don't think there are any places where we use this in a way that creates a vulnerability, but wanna get rid of it either way. @awanlin if you want to drop https://github.com/backstage/backstage/pull/31454/commits/416062acafe4c6007498c0d30cfd9843712e3088 and force push to clean up the branch that helps, or we can push a revert.

I'll look at removing the `shell: true` options separately

Rugvip:
Actually I pushed a revert so that we get the branch in what should be mergable eventually. Leaving it to you if you want to clean up the branch by removing both commits @awanlin

.changeset/stupid-cases-fold.md
awanlin: For clarification we decided to go with ES2023 right? Not ES2024?

Rugvip:
Fix for ^ on the way in #31916

freben:
The above is now merged. Time to rebase?

freben:
@awanlin Sorry I rebased and pushed this this morning to see if we can get it through - hope that's OK

PR #31459, opened 2025-10-17
cli: add template for catalog provider

Rugvip:

This adds a new template to help create new catalog entity provider modules. Idea is also to settle on a bit of a pattern for how to structure simple full-sync providers.

### DIFF
--- a/packages/cli/src/modules/new/lib/execution/writeTemplateContents.ts
+++ b/packages/cli/src/modules/new/lib/execution/writeTemplateContents.ts
@@ -22,6 +22,7 @@ import { PortableTemplate, PortableTemplateInput } from '../types';
 import { ForwardedError, InputError } from '@backstage/errors';
 import { isMonoRepo as getIsMonoRepo } from '@backstage/cli-node';
 import { PortableTemplater } from './PortableTemplater';
+import { isChildPath } from '@backstage/cli-common';
 
 export async function writeTemplateContents(
   template: PortableTemplate,
@@ -63,7 +64,12 @@ export async function writeTemplateContents(
     }
 
     for (const file of template.files) {
-      const destPath = resolvePath(targetDir, file.path);
+      const destPath = resolvePath(targetDir, templater.template(file.path));
+      if (!isChildPath(targetDir, destPath)) {
+        throw new Error(
+          `Path ${destPath} is outside of target directory ${targetDir}`,
+        );
+      }
       await fs.ensureDir(dirname(destPath));
 
       let content =
--- a/packages/cli/templates/catalog-provider-module/.eslintrc.js.hbs
+++ b/packages/cli/templates/catalog-provider-module/.eslintrc.js.hbs
@@ -1,0 +1,1 @@
+module.exports = require('@backstage/cli/config/eslint-factory')(__dirname);
--- a/packages/cli/templates/catalog-provider-module/config.d.ts.hbs
+++ b/packages/cli/templates/catalog-provider-module/config.d.ts.hbs
@@ -1,0 +1,34 @@
+import { SchedulerServiceTaskScheduleDefinitionConfig } from '@backstage/backend-plugin-api';
+
+export interface Config {
+  catalog?: {
+    providers?: {
+      /**
+       * {{providerClass}} configuration.
+       */
+      {{providerVar}}?:
+        | {
+            /**
+             * The target that this provider should consume.
+             */
+            target: string;
+            /**
+             * Overrides the schedule at which this provider runs.
+             */
+            schedule?: SchedulerServiceTaskScheduleDefinitionConfig;
+          }
+        | {
+            [name: string]: {
+              /**
+               * The target that this provider should consume.
+               */
+              target: string;
+              /**
+               * Overrides the schedule at which this provider runs.
+               */
+              schedule?: SchedulerServiceTaskScheduleDefinitionConfig;
+            };
+          };
+    };
+  };
+}
--- a/packages/cli/templates/catalog-provider-module/package.json.hbs
+++ b/packages/cli/templates/catalog-provider-module/package.json.hbs
@@ -1,0 +1,36 @@
+{
+  "name": "{{packageName}}",
+  "description": "The {{fullModuleId}} module for @backstage/plugin-catalog-backend",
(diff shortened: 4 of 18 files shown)


packages/cli/templates/catalog-provider-module/package.json.hbs
```diff
@@ -0,0 +1,36 @@
...
+  "dependencies": {
+    "@backstage/backend-plugin-api": "{{versionQuery '@backstage/backend-plugin-api'}}",
+    "@backstage/plugin-catalog-node": "{{versionQuery '@backstage/plugin-catalog-node'}}"
+  },
+  "devDependencies": {
+    "@backstage/cli": "{{versionQuery '@backstage/cli'}}",
+    "@backstage/backend-test-utils": "{{versionQuery '@backstage/backend-test-utils'}}"
+  },
+  "files": [
+    "dist"
```
awanlin:
```suggestion
    "dist",
    "config.d.ts"
```

packages/cli/templates/catalog-provider-module/package.json.hbs
```diff
@@ -0,0 +1,36 @@
...
+  ]
```
awanlin:
```suggestion
  ],
  "configSchema": "config.d.ts"
```

packages/cli/src/modules/new/lib/execution/writeTemplateContents.ts
```diff
@@ -63,7 +63,7 @@ export async function writeTemplateContents(
     }
 
     for (const file of template.files) {
-      const destPath = resolvePath(targetDir, file.path);
+      const destPath = resolvePath(targetDir, templater.template(file.path));
```
freben, line 67: Does this need a dose of safe child path stuff now?

packages/cli/templates/catalog-provider-module/config.d.ts.hbs
```diff
@@ -0,0 +1,34 @@
...
+            /**
+             * The target that this provider should consume.
+             */
+            target: string;
+            /**
+             * Overrides the schedule at which this provider runs.
+             */
+            schedule?: SchedulerServiceTaskScheduleDefinitionConfig;
+          }
+        | {
```
freben, line 20: Shall we keep this as the default config structure? It's a bit messy and you need to have very custom detection logic (which gets a bit unreliable if someone happened to use a known config key as the id). I realise config merging is desirable so it may be undesirable to use arrays. So then the question is, should we always use the extra level of nesting or not. Maybe it's 99% common to only need the one element... Eh I don't know, but thought I'd bring it up.

packages/cli/templates/catalog-provider-module/.eslintrc.js.hbs
freben, line 1: doesn't need to be an hbs if you don't want to

packages/cli/templates/catalog-provider-module/src/module.ts.hbs
```diff
@@ -0,0 +1,29 @@
+import {
+  coreServices,
+  createBackendModule,
+} from "@backstage/backend-plugin-api";
+import { catalogProcessingExtensionPoint } from '@backstage/plugin-catalog-node/alpha';
+import { {{providerClass}} } from "./provider/{{providerClass}}";
```
freben: mixing double and single quotes, but maybe that's not particularly relevant if we always lint / prettier in practice

packages/cli/templates/catalog-provider-module/src/provider/readProviderConfigs.ts.hbs
```diff
@@ -0,0 +1,78 @@
+import {
+  readSchedulerServiceTaskScheduleDefinitionFromConfig,
+  SchedulerServiceTaskScheduleDefinition,
+} from '@backstage/backend-plugin-api';
+import { Config } from '@backstage/config';
+
+const DEFAULT_PROVIDER_ID = 'default';
+const DEFAULT_SCHEDULE = {
```
freben:
```suggestion
const DEFAULT_SCHEDULE: SchedulerServiceTaskScheduleDefinition = {
```

to help the user a bit

packages/cli/templates/catalog-provider-module/src/provider/readProviderConfigs.ts.hbs
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @param config - The root of the provider config hierarchy
+ *
+ * @public
+ */
+export function readProviderConfigs(
+  config: Config,
+): {{providerClass}}ProviderConfig[] {
+  const providersConfig = config.getOptionalConfig(
+    'catalog.providers.testProvider',
```
freben:
```suggestion
    'catalog.providers.{{ providerVar }}',
```

packages/cli/templates/catalog-provider-module/config.d.ts.hbs
Parsifal-M: We generally follow this pattern for all our custom providers we don't use `target` (we just word it differently) but structurally we use this format, not sure if that helps

packages/cli/src/modules/new/lib/execution/writeTemplateContents.ts
Rugvip: Ah yep 👍  not a huge issue right now but could be if we make remote templates a thing

packages/cli/templates/catalog-provider-module/config.d.ts.hbs
Rugvip: Yeah had the same though @freben, especially potentially keeping it simple with just a single layer of config. Especially with a custom provider I think it's likely a lot more common to just need a single one. At the same time this is a very well established pattern at this point, which makes me think that it's good to provide a clear pattern for it. Kinda feeling that this is simply where we're at atm, and any change is likely not worthwhile unless we're making broader config changes.

packages/cli/templates/catalog-provider-module/.eslintrc.js.hbs
Rugvip: It does iirc, because it'll be picked up by ESLint otherwise

PR #31460, opened 2025-10-17
adjust the log line wording of task worker starting

freben:
We keep getting pinged about this. The log line makes it sound like "the actual task is starting right now" (potentially on many machines at the same time like during a rollout), and also there's NO log lines when the task DOES start.

It can be argued that it's not of all that great importance to log this at all. But I chose in the end to keep the line but word it in a "milder" way that does not quite imply actual running.

### DIFF
--- a/packages/backend-defaults/src/entrypoints/scheduler/lib/LocalTaskWorker.ts
+++ b/packages/backend-defaults/src/entrypoints/scheduler/lib/LocalTaskWorker.ts
@@ -52,9 +52,7 @@ export class LocalTaskWorker {
 
   start(settings: TaskSettingsV2, options: { signal: AbortSignal }) {
     this.logger.info(
-      `Received registration for scheduler task: ${
-        this.taskId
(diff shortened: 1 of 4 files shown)


packages/backend-defaults/src/entrypoints/scheduler/lib/LocalTaskWorker.ts
```diff
@@ -52,7 +52,9 @@ export class LocalTaskWorker {
 
   start(settings: TaskSettingsV2, options: { signal: AbortSignal }) {
     this.logger.info(
-      `Task worker starting: ${this.taskId}, ${JSON.stringify(settings)}`,
+      `Received registration for scheduler task: ${
```
Rugvip: Perhaps just `Registered scheduled task ...`?

freben: Went for that

PR #31469, opened 2025-10-18
fix(ui): Set color-scheme, notably to have better default scrollbar theming

GabDug:

This PR attempts to properly set the `color-scheme` property, depending on the user theme.

This property allows to have the native scrollbars (& other UA defaults?) themed accordingly.

I've also added the `<meta name="color-scheme" content="light dark">` to the index.html, which will indicates we support both modes. This will help browser show the right (loading) colors, as soon as the HTML is here, and before the JS/CSS is downloaded or parsed, by using the user's system scheme.

> The order would indicate we support both but prefer the light mode, but browsers seem to prefer the user's system scheme instead, which is fine!

### Demo

Before: notice the scrollbars are not properly themed for Safari & Chrome. They don't react to the theme changes nor follow the system's scheme (in this case, dark)

https://github.com/user-attachments/assets/875a8ccd-6673-431e-8579-90f073393341

After: the scrollbar colors follow the system scheme, and are updated to match the currently loaded theme.

https://github.com/user-attachments/assets/e3c49570-5c3a-4ccd-aea7-c69e387f64b9

Tested on latest Firefox / Chrome / Safari.

### Other possible implementations

I've had first tested setting the color scheme meta tag dynamically in `useApplyThemeAttributes.ts`, but this CSS approach is simpler, and also work in docs-ui. The meta tag and CSS property seem to be handled equivalently anyway.

### Potential improvements

- Instead of hardcoding `<meta name="color-scheme" content="light dark">` to the index.html, provide it automatically at build.
  - Use case: set only dark or only rights if there are only themes for a given mode.
  - This would be a better "auto-magic" default when custom themes only support one scheme.
  - Probably a bit over-engineered, but might be "cleaner" once Webpack is out of the picture.
- Inject a simple inline script to set the content-scheme early, to read user storage instead of only browser default
 - Same as above 

I've haven't added the changesets yet, please let me know which packages it would be relevant for.

Let me know what you think!
Cheers

### DIFF
--- a/.changeset/wide-papers-run.md
+++ b/.changeset/wide-papers-run.md
@@ -1,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Set the color-scheme property depending on theme


GabDug:
@cdedreuille would be interested in your feedback about this non-urgent PR. Thanks!

cdedreuille:
I just noticed that Chromatic is not working for forked branches. I just did a fix for that - https://github.com/backstage/backstage/pull/31470. We'll need to get that sorted before we can merge this PR.

benjdlambert:
The description and title of this PR don’t match the changes anymore. Can we update them?

GabDug:
@benjdlambert Wow sorry about that, I pushed the content of https://github.com/backstage/backstage/pull/31483/files here by mistake.

I've fixed that and rebased.

benjdlambert:
@GabDug sorry, one more thing this needs a changeset! :pray:

benjdlambert (approved):
Lovely! Thanks :tada:

PR #31476, opened 2025-10-20
feat(mcp-actions): support _meta field for actions

drodil:

actions can now use `metadata` field to add additional metadata for mcp server use. this field is optional so it's not breaking change. see https://modelcontextprotocol.io/specification/2025-06-18/basic/index#meta for details about the metadata.


Rugvip (commented):
Looks good but wanna make sure there's actually a need for this as I think we can expect the MCP spec to keep moving with potential for changes in this area too.

Do you have a direct use for this and/or any use-case you'd be able to share? Not immediately seeing the value of the field in the docs or some examples I found online.

Issue #31412, opened 2025-10-14
💬 RFC: New Header for Backstage plugins


cdedreuille:
### 🔖 Need

We are currently using a `Header` component from `@backstage/core-components` on most plugins to set the structure of the page but this component is not reflecting the new Backstage UI design philosophy. We believe in a more condensed experience that would allow most of the content on the page to be accessible without taking too much space on each page.

Since the `Header` is what set the tone of the page, we decided to prioritise this part of the experience and have a replacement available for all plugins.

We do not have an exact solution yet on what will happen to `@backstage/core-components` but we are slowly transitioning to have all components available in `@backstage/ui`.

The main question today that remain is how should we roll out this new header since this is affecting all plugins.

### 🎉 Proposal

We built a new header structure in Backstage UI composed of two parts:
- `Header` - This part is to be set above any route as a global pattern.
- `HeaderPage` - This is to be set inside of a route to get data from this route like a title or breadcrumb.

We do not have full parity in feature with existing features and the structure is fundamentally different. We are expecting each plugin to take some time to adopt this new pattern. Backstage monorepo has 37 instances of the previous `Header`.

Our main proposition would be to replace all headers in all plugins in the monorepo in one go, potentially for the release in November 10th. Even though we took the time to match as close as possible all use cases, we do expect to have some minor updates to do on our components along the way to match specific requirements.

### 🎨 New Header Anatomy

- `Header` documentation - [https://ui.backstage.io/components/header](https://ui.backstage.io/components/header-page)
- `HeaderPage` documentation - [https://ui.backstage.io/components/header-page](https://ui.backstage.io/components/header)

<img width="2368" height="1320" alt="Image" src="https://github.com/user-attachments/assets/a235e340-3a10-41df-b6ba-3585947f219c" />

### 〽️ Alternatives

As an alternative, we could consider shipping the new component gradually in key components first like `Catalog`, `TechDocs,` `Search` to allow for feedback and move to a full migration gradually. This present some risks though as this would create quite a fragmented experience for everyone with plugins that would look quite different.

We could also find a way to create an opt-in approach (like a feature flag) that would allow users to set this new feature. This would add quite a lot of complexity but could be a good addition to allow to ship this new header gradually while allowing us to ship the final version all in one. We are not sure of its implementation though. Any ideas are welcome 🙏

### ❌ Risks

[...]

PR #31478, opened 2025-10-20
Set browser title in BUI Header component.

Naycon:

The BUI Header component now automatically sets the browser tab title.

Changes made:
- Header component now uses `react-helmet` to set browser title
- `title` prop and config `app.title` is used as the title
- Added `react-helmet` and `@backstage/core-plugin-api` dependencies to `@backstage/ui package`
- Moved definition of the default title "Your plugin" to prevent undefined values in browser title
- Updated documentation in docs-ui to explain the browser title behavior

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -38,11 +38,13 @@
     "test": "backstage-cli package test"
   },
   "dependencies": {
+    "@backstage/core-plugin-api": "workspace:^",
     "@base-ui-components/react": "1.0.0-alpha.7",
     "@remixicon/react": "^4.6.0",
     "@tanstack/react-table": "^8.21.3",
     "clsx": "^2.1.1",
-    "react-aria-components": "^1.13.0"
+    "react-aria-components": "^1.13.0",
+    "react-helmet": "6.1.0"
   },
   "devDependencies": {
     "@backstage/cli": "workspace:^",
--- a/packages/ui/src/components/Header/Header.tsx
+++ b/packages/ui/src/components/Header/Header.tsx
@@ -21,6 +21,8 @@ import { useStyles } from '../../hooks/useStyles';
 import { type NavigateOptions } from 'react-router-dom';
 import styles from './Header.module.css';
 import clsx from 'clsx';
+import { Helmet } from 'react-helmet';
+import { configApiRef, useApi } from '@backstage/core-plugin-api';
 
 declare module 'react-aria-components' {
   interface RouterConfig {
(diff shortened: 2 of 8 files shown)


packages/ui/src/components/Header/Header.tsx
```diff
@@ -35,13 +37,24 @@ declare module 'react-aria-components' {
...
+
+  const configApi = useApi(configApiRef);
+  const appTitle = configApi.getOptionalString('app.title') || 'Backstage';
+  const pageTitle = `${title} | ${appTitle}`;
 
   const hasTabs = tabs && tabs.length > 0;
 
   return (
     <>
+      <Helmet title={pageTitle} />
```
UsainBloot: Should we keep the same functionality as the core header with the titleTemplate and defaultTitle props?

https://github.com/backstage/backstage/blob/8bfe82fbc60341b15023dee0c98c6899a12f72a8/packages/core-components/src/layout/Header/Header.tsx#L225

Naycon: I was lookin into this, and it's actually causing bugs currently. The `%s` can in some instances result in "double" titles, if `Helmet` is used independently to set the title. E.g. you set the title to `My Plugin` using React Helmet in your plugin. Then you navigate to another plugin which uses the `titleTemplate` like the one you posted. This then results in the title becoming `My Second Plugin | My Plugin | My Backstage Instance`.

I think the real solution is to expose some better way of working with titles that gives the right amount of control. But let me know if you see situations where `%s` is necessary to support, or if you think setting titles directly using Helmet is an antipattern we shouldn't support.

Naycon: Found a way to work around this potentially (updated the PR). If we let `Header` set the template and default titles, `HeaderPage` components can set the title and the end result becomes `My Plugin Page | My Plugin | My Company`. If using helmet directly inside the plugin, the result would be similar, `My Helmet Title | My Plugin | My Company`. If no `HeaderPage` component or helmet is used, the title will simply be `My Plugin | My Company`. 

This will somewhat mimic the behavior of the old `Header` component from `core-components`, and I think it should work as intended.

packages/ui/package.json
```diff
@@ -38,11 +38,13 @@
     "test": "backstage-cli package test"
   },
   "dependencies": {
+    "@backstage/core-plugin-api": "workspace:^",
```
Rugvip, line 41: Let's stick to NFS if possible, i.e. `@backstage/frontend-plugin-api`.

Rugvip (commented):
Hmm, I think it would be best to settle #31412 before moving this forward tbh. I'm leaning towards the best place to implement this being within the frontend system's implementation of navigable pages that we might create as part of that change.

freben:
This is still of interest I presume!

PR #31479, opened 2025-10-20
feat: Expand `EntityFilter.getCatalogFilters` to return `EntityFilterQuery`

tklever:

This change allows `EntityFilter`s to utilize more complete functionality of the `CatalogApi`, specifically the `queryEntities` and `getEntities` methods which support an `EntityFilterQuery` as the `filter` parameter in the request.

Allowing `EntityFilter`s to return a complete `EntityFilterQuery` allows them to produce more complex filter logic.

The primary example is allowing an `EntityFilter` to produce an array of filter records that would produce an 'OR' query, filtering on A OR B criteria.

To complete this change, some of the helpers and utilities used by Entity filtering functionality (ex: `useEntityListProvider`) needed to be updated to account for the possibility of an array of filter records.

### DIFF
--- a/.changeset/bitter-suns-build.md
+++ b/.changeset/bitter-suns-build.md
@@ -1,0 +1,46 @@
+---
+'@backstage/plugin-catalog-react': major
+'@backstage/catalog-client': minor
+---
+
+**BREAKING** Expand `EntityFilter.getCatalogFilters` to return `EntityFilterQuery`
+
+```diff
+import { Entity } from '@backstage/catalog-model';
++ import { EntityFilterQuery } from '@backstage/catalog-client';
+
+/** @public */
+export type EntityFilter = {
+-  getCatalogFilters?: () => Record<
+-    string,
+-    string | symbol | (string | symbol)[]
+-  >;
++  getCatalogFilters?: () => EntityFilterQuery;
+};
+```
+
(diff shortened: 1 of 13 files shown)


aramissennyeydd:
👋 Can you share a little bit more about the problem you're looking to solve here? Curious if it can be solved by just providing 2 filters instead of a single one.

tklever:
> 👋 Can you share a little bit more about the problem you're looking to solve here?

Absolutely!

Short answer is being able to provide filters that use an "OR" condition. Selecting an option that gets expanded into "A" OR "B". A selection that gets expanded to "Annotation A OR Annotation B" or "Relationship A OR Relationship B"

The `CatalogApi` and `EntityFilterQuery` already support OR queries, this just adds the needed plumbing to get from the catalog filters to the catalog API (and maintain some existing functionality in between those two things).

My specific specific use case is selecting a user entity that may be related to other entities via multiple distinct relationships. Kind of a "show me where this person is related to an entity in any of a handful of ways" 

>  Curious if it can be solved by just providing 2 filters instead of a single one.

I believe two filters acts as an AND, not an OR. 

Other workarounds include adding some kind of aggregate calculated single relationship (or any other annotation) at ingestion, but since the Catalog already supports OR filter queries, I thought I propose the functionality.

PR #31482, opened 2025-10-21
Offline access + CLI auth

Rugvip:

This implements offline access, aka refresh tokens, for the main Backstage user sessions. Closing #30066.

This is useful for both MCP and CLI auth, and any other potential auth flow.

It also adds a new `auth` module and set of commands to `@backstage/cli`, currently focused around just login/logout and printing access tokens.

Very draft for now.

### DIFF
--- a/plugins/auth-backend/src/service/OfflineAccessService.ts
+++ b/plugins/auth-backend/src/service/OfflineAccessService.ts
@@ -95,9 +95,10 @@ export class OfflineAccessService {
    */
   async refreshAccessToken(options: {
     refreshToken: string;
+    clientId: string;
     tokenIssuer: TokenIssuer;
   }): Promise<{ accessToken: string; refreshToken: string }> {
-    const { refreshToken, tokenIssuer } = options;
+    const { refreshToken, clientId, tokenIssuer } = options;
 
     // Extract session ID from token
     let sessionId: string;
@@ -118,6 +119,12 @@ export class OfflineAccessService {
     if (this.#offlineSessionDb.isSessionExpired(session)) {
       await this.#offlineSessionDb.deleteSession(sessionId);
       throw new AuthenticationError('Refresh token expired');
+    }
+
+    // Verify client_id matches the session's client_id
+    // Refresh tokens should always have an oidcClientId set
+    if (session.oidcClientId === null || session.oidcClientId !== clientId) {
+      throw new AuthenticationError('Invalid client_id for refresh token');
     }
 
     // Verify token hash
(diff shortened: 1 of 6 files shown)


packages/cli/src/modules/auth/commands/login.ts
```diff
@@ -0,0 +1,364 @@
...
+  }
+  if (platform === 'win32') {
+    spawn('cmd', ['/c', 'start', '""', url], {
+      detached: true,
+      stdio: 'ignore',
+    }).unref();
+    return;
+  }
+  // linux and others
+  spawn('xdg-open', [url], { detached: true, stdio: 'ignore' }).unref();
```
fearphage: Since these all ultimately call the same function, why not make that more obvious/clear:

```suggestion
  const command = ({ darwin: 'open', win32: 'cmd' }[platform]) ?? 'xdg-open';
  const parameters = platform === 'win32' ? ['/c', 'start', '""', url] : [url];
  
  spawn(command, parameters, { detached: true, stdio: 'ignore' }).unref();
```

Rugvip: Don't have a strong opinion, this is all generated 😅. Gotta say I think the original is a bit more readable though, even if it's longer. It's immediately clear that args go with each command, while the more condensed version requires a tiny bit of thinking to figure out. Gonna shorten it up a bit though

packages/cli/src/modules/auth/commands/login.ts
```diff
@@ -0,0 +1,376 @@
...
+    id_token?: string;
+    refresh_token?: string;
+  }>(`${authBaseUrl}/v1/token`, {
+    method: 'POST',
+    body: {
+      grant_type: 'authorization_code',
+      code,
+      redirect_uri: redirectUri,
+      code_verifier: verifier,
+    },
```
schultzp2020, line 310:
```suggestion
  }>(`${authBaseUrl}/v1/token`, {
    method: 'POST',
    body: {
      grant_type: 'authorization_code',
      code,
      redirect_uri: redirectUri,
      code_verifier: verifier,
      client_id: clientId,
      client_secret: clientSecret,
    },
```

Is `client_id` and `client_secret` needed here?

packages/cli/src/modules/auth/lib/auth.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    const response = await httpJson<unknown>(
+      `${instance.baseUrl}/api/auth/v1/token`,
+      {
+        method: 'POST',
+        body: {
+          grant_type: 'refresh_token',
+          refresh_token: refreshToken,
+        },
```
schultzp2020, line 64:
```suggestion
      `${instance.baseUrl}/api/auth/v1/token`,
      {
        method: 'POST',
        body: {
          grant_type: 'refresh_token',
          refresh_token: refreshToken,
          client_id: clientId,
          client_secret: clientSecret,
        },
```

Is `client_id` and `client_secret` needed here?

benjdlambert:
Given the latest changes to the MCP Spec, and the fact that it looks like DCR is likely being phased out in favor of [OAuth Client ID Metadata Document](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document) should we maybe revisit the idea to build this on top of DCR?

Rugvip:
@benjdlambert ah yep happy to explore moving this on top of CIMD instead

benjdlambert:
All of this has been shipped with #32920 as the last piece

PR #31485, opened 2025-10-21
Use theme ID context instead default 'backstage' value

gaelgoth:

In my Backstage instance, we’re developing a new theme so that we have 2 `light` and `black` themes. I noticed that       `data-theme-name` attribute is always set to `backstage`, which limits theme-specific styling. For example:

```css
[data-theme-name='spotify'][data-theme-mode='light'] { // It doesn't work because of data-theme-name='backstage'
  --sb-sidebar-border: 1px solid var(--bui-border);
  --sb-sidebar-border-right: 1px solid var(--bui-border);
}
```

To resolve this, I added the `themeId` from AppThemeIdContext to correctly set the data-theme-name attribute based on the selected theme. It falls back to `backstage` if missing

### DIFF
--- a/.changeset/slick-books-sleep.md
+++ b/.changeset/slick-books-sleep.md
@@ -2,4 +2,4 @@
 '@backstage/theme': patch
 ---
 
-Added `themeName` props to `UnifiedThemeProvider` to allow setting Backstage UI `data-theme-name`CSS attribute dynamically based on the active theme.
+Added a `themeName` prop to `UnifiedThemeProvider`, enabling Backstage UI `data-theme-name` CSS attribute to be set based on active theme.


gaelgoth:
I ran `yarn build:api-reports packages/core-app-api packages/theme`, but CI is still not happy because of scaffolder report 🤔

benjdlambert:
Pulling in @backstage/design-system-maintainers too here for a review :pray:

packages/theme/src/unified/UnifiedThemeProvider.tsx
```diff
@@ -71,10 +77,10 @@ export function UnifiedThemeProvider(
   const v4Theme = theme.getTheme('v4') as Mui4Theme;
   const v5Theme = theme.getTheme('v5') as Mui5Theme;
 
-  useApplyThemeAttributes(
-    v4Theme ? v4Theme.palette.type : v5Theme?.palette.mode,
-    'backstage',
-  );
+  const themeMode = v4Theme ? v4Theme.palette.type : v5Theme?.palette.mode;
+  const themeId = useContext(AppThemeIdContext) ?? 'backstage';
```
Rugvip: Could you make this a prop for the `UnifiedThemeProvider` instead? I think we can just call it `themeName` and default it to `backstage`. We could also add `themeMode` while at it which optionally overrides the mode read from the theme.

gaelgoth: Fine for me, as long as it allows us to choose which theme name is set  for `[data-theme-name='spotify']` 😄 

I removed the logic that get the theme context and replaced it with `themeName` prop

gaelgoth:
@Rugvip Ready for second round. I've added an `optional` prop that falls back with current default `backstage` theme name. 

I'm not sure why the CI isn't happy with 2E2 tests

.changeset/slick-books-sleep.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/core-app-api': patch
```
Rugvip:
```suggestion
```

Since this no longer changes that package

benjdlambert (approved):
Nice thanks! :tada:

PR #31493, opened 2025-10-22
refactor(ui): replace Collapsible with Accordion component

Naycon:

Replaces the Base UI Collapsible component with a new Accordion component built on React Aria's Disclosure primitives.

image

image

Key changes:
- Removed Collapsible component and all related files
- Added new Accordion component with AccordionTrigger, AccordionPanel, and AccordionGroup
- Introduced opinionated styling with built-in trigger component featuring animated chevron icon
- Added support for title/subtitle props and custom trigger content via children
- Implemented AccordionGroup with single/multiple expansion modes
- Updated all documentation from collapsible.mdx to accordion.mdx
- Added comprehensive migration guide in changeset
- Updated API reports and exports

This is a breaking change. Users must migrate from Collapsible to either:
1. Accordion (opinionated styled component) - recommended for most use cases
2. React Aria Disclosure directly - for full customization

The changeset provides detailed migration examples for both paths.


Naycon:
~The `DisclosureButton` might need a design pass before we move forward with this. It's currently a raw RA button, and should likely be a BUI button instead. I'm not sure if we want it to be a secondary button with an icon (like Collapsible used to be), or if we'd like to align more with RA and use a text button + icon instead.~

Not accurate anymore - branch now contains a fully styled Accordion.

docs-ui/src/utils/changelog.ts
```diff
@@ -12,7 +12,7 @@ export type Component =
   | 'textfield'
   | 'datatable'
   | 'select'
-  | 'collapsible'
```
cdedreuille, line 15: I believe we still need this here as otherwise we will loose the data in our changelog.

docs-ui/src/utils/changelog.ts
```diff
@@ -327,9 +327,9 @@ export const changelog: ChangelogProps[] = [
     type: 'new',
   },
   {
-    components: ['collapsible'],
+    components: ['accordion'],
     version: '0.3.0',
-    description: 'Add `Collapsible` component',
+    description: 'Add `Accordion` component',
```
cdedreuille: I don't think we should change old changelog. We will add new changelog to add the new `Accordion` and remove `Collapsible`

packages/ui/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,133 @@
...
+      )}
+      {...cleanedProps}
+    >
+      {children ? (
+        children
+      ) : (
+        <Flex gap="2" align="center">
+          <Text as="span" variant="body-large" weight="bold">
+            {title}
+          </Text>
```
cdedreuille: I think we should use the `Heading` component and add a style to it. Otherwise people will not be able to style it if there's no class name. This will still be connected to our token so we are safe.

Also there might be A11y benefits of using `Heading`

packages/ui/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,133 @@
...
+          {subtitle && (
+            <Text as="span" variant="body-small" color="secondary">
+              {subtitle}
+            </Text>
```
cdedreuille: Same here, I don't think we should use our `Text` component for theming perspective.

cdedreuille (changes requested):

docs-ui/src/utils/changelog.ts
Naycon: Oh, sorry! This was me relying on some AI help to get the do iterations on the changelog. It was a bit over eager. Will revert!

docs-ui/src/utils/changelog.ts
Naycon: Fixed.

docs-ui/src/utils/changelog.ts
Naycon: Fixed.

packages/ui/src/components/Accordion/Accordion.tsx
Naycon: Done.

packages/ui/src/components/Accordion/Accordion.tsx
Naycon: Changed to `span` inside a `Heading` instead.

PR #31497, opened 2025-10-22
fix(cache): improve support for Valkey

benjidotsh:
This PR improves support for using Valkey in the Cache Service, fixing a few issues and bugs along the way:

- Most of the configuration options are not available in Valkey or have a different name and/or implementation (e.g. `namespace` and `keyPrefixSeparator` vs `keyPrefix`)
- `createCluster` is not available in `@keyv/valkey`, causing Backstage to crash on startup

### DIFF
--- a/.changeset/sour-lines-taste.md
+++ b/.changeset/sour-lines-taste.md
@@ -9,13 +9,14 @@ These changes are **required** to `app-config.yaml`:
 ```diff
 backend:
   cache:
-    valkey:
-      client:
--       namespace: 'my-app'
--       keyPrefixSeparator: ':'
-+       keyPrefix: 'my-app:'
--       clearBatchSize: 1000
--       useUnlink: false
+    store: valkey
+    connection: ...
+    client:
+-     namespace: 'my-app'
+-     keyPrefixSeparator: ':'
++     keyPrefix: 'my-app:'
+-     clearBatchSize: 1000
+-     useUnlink: false
 ```
 
 In comparison to Redis, Valkey requires the full `keyPrefix` including the separator to be specified instead of separate `namespace` and `keyPrefixSeparator` options. Also, Valkey does not support the `clearBatchSize` and `useUnlink` options.


packages/backend-defaults/src/entrypoints/cache/CacheManager.test.ts
```diff
@@ -462,6 +469,31 @@ describe('CacheManager store options', () => {
       expect(result).toBe('my-app:testPlugin');
     });
 
+    it('returns pluginId when store options have no keyPrefix (valkey)', () => {
```
fearphage: Instead of copying and pasting the test which allows for them to get out of sync, could you parameterize the tests instead? This also makes it obvious that the tests are identical.

```js
it.each([
  { type: 'redis', suffix: 'namespace', client: { /* client here */ }],
  { type: 'valkey', suffix: 'keyPrefix', client: { /* client here */ }],
])('returns pluginid when store options have no $suffix ($type)', ({ client, type }) =>{
  const storeOptions = {
    client,
    type,
  };
  // ... the rest of the test is identical
});
```

This also applies to the next test you added.

fearphage (changes requested):

packages/backend-defaults/src/entrypoints/cache/CacheManager.test.ts
benjidotsh: Good suggestion, thanks!

Although I'm not sure what other test you would combine with `uses keyPrefix for valkey`... They're either testing the combination of the namespace and the separator or testing whether the default separator is being used.

benjidotsh:
FWIW: I noticed another big issue with support for Valkey, but I'm not sure on how to proceed forward, so I created an issue to discuss: #31521

packages/backend-defaults/src/entrypoints/cache/CacheManager.test.ts
```diff
@@ -364,16 +367,16 @@ describe('CacheManager store options', () => {
...
         // eslint-disable-next-line jest/no-conditional-expect
-        expect(KeyvValkey).toHaveBeenCalledWith('redis://localhost:6379', {
-          namespace,
-          keyPrefixSeparator: separator,
-        });
+        expect(KeyvValkey).toHaveBeenCalledWith(
+          'redis://localhost:6379',
+          client,
+        );
       }
```
fearphage, line 380: You can remove the conditional assertion by dynamically deciding which plugin to target with the assertions. This also makes it obvious that the assertion is identical.

```js
const plugin = store === 'redis' ? KeyvRedis : KeyvValkey;
expect(plugin).toHaveBeenCalledWith(
  'redis://localhost:6379',
  client,
);
```

packages/backend-defaults/src/entrypoints/cache/CacheManager.test.ts
fearphage: I understand. The body of the following test is identical to several tests before it is why it seemed like they could be combined.

packages/backend-defaults/src/entrypoints/cache/CacheManager.test.ts
benjidotsh: While I agree your solution is more elegant, the original implementation is not something I introduced, so I'm not sure if this change should be part of this PR.

packages/backend-defaults/src/entrypoints/cache/CacheManager.test.ts
benjidotsh: Yeah, you definitely have a point, but I think combining them would only add confusion as they are testing different things. 😅

packages/backend-defaults/src/entrypoints/cache/CacheManager.test.ts
fearphage: But you're actively changing the code now. That's the only reason I commented on it. You feel it is okay to change the code, but not to improve the clarity of it?

It's up to you honestly. This is not my project. My approval is largely symbolic. If I remember, I'll fix it after this is merged next time I'm in the code. 

FYI while I've never been a boy scout, I do subscribe to their motto:

> Leave things better than you found them.

benjidotsh: I try to keep my pull requests focused as much as possible to make it easier for review, especially if it's just something related to code style.

Let's wait for a maintainer to chime in. If a maintainer agrees with you, I'm happy to make the change.

fearphage: > I try to keep my pull requests focused as much as possible to make it easier for review

We are 100% in agreement there. This is the way to do it for sure. It makes things easier for maintainers to accept incoming code.

> especially if it's just something related to code style.

This is not about style/preference. As you see in the code, they are having to disable an ESLint rule to make these conditional assertions. *Conditional assertions are bad.* That's why the rule is recommended and enabled by default. The rule definition outlines the [downsides of conditional assertions](https://github.com/jest-community/eslint-plugin-jest/blob/HEAD/docs/rules/no-conditional-expect.md). ESLint is attempting to improve the quality of the code, but it was overridden. That's why this felt like an opportunity to correct a previous error. 

> I'm happy to make the change.

For reference, I am **_not_** trying to convince you to make this change. I'm explaining to you that tests should not have conditional assertions in them as a rule. The code will get fixed at some point either way.

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -140,37 +141,43 @@ export class CacheManager {
...
-
-    return undefined;
   }
 
   /**
    * Parse Redis-specific options from configuration.
    */
   private static parseRedisOptions(
-    store: string,
+    store: 'redis',
```
freben: At this point let's just remove this argument. Same in the corresponding valkey method.

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -213,6 +220,52 @@ export class CacheManager {
...
+    valkeyOptions.client = {
+      keyPrefix: valkeyConfig.getOptionalString('client.keyPrefix'),
+    };
+
+    if (valkeyConfig.has('cluster')) {
+      const clusterConfig = valkeyConfig.getConfig('cluster');
+
+      if (!clusterConfig.has('rootNodes')) {
+        logger?.warn(
+          `Redis cluster config has no 'rootNodes' key, defaulting to non-clustered mode`,
```
freben: Should this say "Valkey"?

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -317,7 +380,7 @@ export class CacheManager {
 
   private createValkeyStoreFactory(): StoreFactory {
     const KeyvValkey = require('@keyv/valkey').default;
-    const { createCluster } = require('@keyv/valkey');
+    const { createCluster } = require('@keyv/redis');
```
freben, line 373: This doesn't seem right.

packages/backend-defaults/src/entrypoints/cache/types.ts
```diff
@@ -17,25 +17,38 @@
...
 
+/**
+ * Options for Valkey cache store.
+ *
+ * @public
+ */
+export type ValkeyCacheStoreOptions = {
+  type: 'valkey';
+  client?: KeyvValkeyOptions;
+  cluster?: RedisClusterOptions;
```
freben, line 41: Is this the literal `RedisClusterOptions`, or is there a `ValkeyClusterOptions` in `@keyv/valkey`?

.changeset/sour-lines-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-defaults': patch
```
freben:
```suggestion
'@backstage/backend-defaults': minor
```

This is technically breaking. The changeset should have a bit more text, saying that if you were using valkey as the cache provider, if you used config option X or Y you now have to instead set Z ... etc. Essentially a mini-migration-instruction on a few lines.

docs/backend-system/core-services/cache.md
```diff
@@ -27,6 +27,9 @@ backend:
         # Other Redis-specific options...
         clearBatchSize: 1000
         useUnlink: false
+    valkey:
+      # Optional: Global namespace prefix for all cache keys (including separator used between namespace and plugin ID)
+      keyPrefix: 'my-app:'
```
freben, line 62: This config block in the docs gets a bit confusing doesn't it - people tend to copy paste and it's not entirely clear that this is only applicable when you had the store set to valkey

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
benjidotsh: It is 😬 Valkey is compatible with a Redis cluster ([source](https://keyv.org/docs/storage-adapters/valkey/)), but has no method of its own to create a cluster

packages/backend-defaults/src/entrypoints/cache/types.ts
benjidotsh: See my comment above, there is no such thing 😞

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
freben: But - stopping using the package seems like a big deal. It's almost to the point of preferring not supporting clusters until they add that properly to the correct keyv package.

benjidotsh: I mean, as far as clustering goes, it does work like this and doesn't crash the backend like it does at the moment, but as I mentioned in #31521, there's a lot more wrong with the current implementation of Valkey. :/

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
freben: Ping on this one

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
freben: Ugh :/ Could we at least put in a comment here saying that yes we want to use `@keyv/valkey` but for the moment it doesn't have working cluster support so we will have to revisit?

And then maybe even remove it as a dependency if it's not being used right now?

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
benjidotsh: Hehe, that's a tough one. Technically it's a Redis cluster, as I mentioned in [my other comment](https://github.com/backstage/backstage/pull/31497#discussion_r2486803104), but I guess that might indeed get a little confusing. I'll rename it to Valkey.

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
benjidotsh: We are using `@keyv/valkey` at the moment on the line above, but I'll add a comment to the `createCluster` part.

Or do we just want to drop `@keyv/valkey` altogether, keeping in mind the issues it currently has? (referring to #31521 again)

docs/backend-system/core-services/cache.md
benjidotsh: I'll add a separate section for both Redis and Valkey, let me know if that looks all right

benjidotsh: I have added separate sections for each store, but I noticed we have the same information of the [Architecture overview](https://backstage.io/docs/overview/architecture-overview#cache)... Is this the right place?

Also, I think there is still a lot to improve as, for example, a lot of Redis options are still missing, but I think it's a great start

.changeset/sour-lines-taste.md
benjidotsh: If you're now using Valkey, your instance shouldn't start _at all_, but I guess you're technically right 😅 I'll elaborate a little

.changeset/sour-lines-taste.md
```diff
@@ -0,0 +1,21 @@
...
+```diff
+backend:
+  cache:
+    valkey:
+      client:
+-       namespace: 'my-app'
+-       keyPrefixSeparator: ':'
++       keyPrefix: 'my-app:'
+-       clearBatchSize: 1000
+-       useUnlink: false
```
freben:
```suggestion
backend:
  cache:
    store: valkey
    connection: ...
    client:
-     namespace: 'my-app'
-     keyPrefixSeparator: ':'
+     keyPrefix: 'my-app:'
-     clearBatchSize: 1000
-     useUnlink: false
```

PR #31505, opened 2025-10-23
feat(catalog): allow processors to emit status items

drodil:

this change allows catalog processors to emit status items to be shown in the NFS catalog page.

### DIFF
--- a/plugins/catalog-backend/src/processing/ProcessorOutputCollector.ts
+++ b/plugins/catalog-backend/src/processing/ProcessorOutputCollector.ts
@@ -18,6 +18,7 @@ import {
   ANNOTATION_LOCATION,
   ANNOTATION_ORIGIN_LOCATION,
   Entity,
+  EntityStatusItem,
   stringifyEntityRef,
   stringifyLocationRef,
 } from '@backstage/catalog-model';
@@ -45,6 +46,7 @@ export class ProcessorOutputCollector {
   private readonly relations = new Array<EntityRelationSpec>();
   private readonly deferredEntities = new Array<DeferredEntity>();
   private readonly refreshKeys = new Array<RefreshKeyData>();
+  private readonly statusItems = new Array<EntityStatusItem>();
   private done = false;
 
   constructor(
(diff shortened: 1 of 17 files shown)


plugins/catalog-node/src/api/processingResult.ts
```diff
@@ -113,4 +118,24 @@ export const processingResult = Object.freeze({
...
     return { type: 'refresh', key };
   },
+
+  /**
+   * Associates a status item with the processing state of the current entity.
+   */
+  status(
+    message: string,
+    level: EntityStatusLevel,
+    type?: string,
```
freben, line 128: I'd rather always force a type, and then we'd document that anything matching `*backstage.io/` is reserved, or something like that.

plugins/catalog-node/src/api/processingResult.ts
```diff
@@ -56,6 +59,8 @@ export const processingResult = Object.freeze({
 
   /**
    * Associates a general Error with the processing state of the current entity.
+   *
+   * TODO: Should this be deprecated in favor of `status`?
```
freben, line 63: I guess so yes.

plugins/catalog-node/src/api/processingResult.ts
```diff
@@ -113,4 +118,24 @@ export const processingResult = Object.freeze({
   refresh(key: string): CatalogProcessorResult {
     return { type: 'refresh', key };
   },
+
+  /**
+   * Associates a status item with the processing state of the current entity.
+   */
+  status(
+    message: string,
```
freben, line 126: This should definitely be a single `options` argument object; the catalog has some of these really old interfaces with a ton of arguments stacked - for example the processor interface - and it always becomes annoying.

plugins/catalog-backend/src/processing/ProcessorOutputCollector.ts
```diff
@@ -156,6 +156,8 @@ export class ProcessorOutputCollector {
       this.errors.push(i.error);
     } else if (i.type === 'refresh') {
       this.refreshKeys.push({ key: i.key });
+    } else if (i.type === 'status') {
+      this.statusItems.push(i.status);
```
freben, line 160: Here it becomes a bit confusing with the distinction of `this.errors` and `this.statusItems` - Is this perhaps a good place to start unifying the two into one array?

PR #31507, opened 2025-10-23
Fix SearchField styling

cdedreuille:

With our move to CSS modules, composed styles from two different components led to unexpected behaviours. To make it clear, we are separating the styles to avoid importing `TextField` styles into `SearchField`

### DIFF
--- a/.changeset/itchy-bars-smell.md
+++ b/.changeset/itchy-bars-smell.md
@@ -2,4 +2,4 @@
 '@backstage/ui': minor
 ---
 
-Fixing styles on SearchField in Backstage UI after migration to CSS modules. `SearchField` has now its own set of class names. We previously used class names from `TextField` but this approach was creating some confusion so going forward in your theme you'll be able to theme `TextField` and `SearchField` separatly.
+Fixing styles on SearchField in Backstage UI after migration to CSS modules. `SearchField` has now its own set of class names. We previously used class names from `TextField` but this approach was creating some confusion so going forward in your theme you'll be able to theme `TextField` and `SearchField` separately.


packages/ui/report.api.md
```diff
@@ -624,7 +624,10 @@ export const componentDefinitions: {
   readonly SearchField: {
     readonly classNames: {
       readonly root: 'bui-SearchField';
-      readonly clear: 'bui-InputClear';
+      readonly clear: 'bui-SearchFieldClear';
```
Naycon, line 627: Doesn't this mean that this might potentially be a breaking change? Not sure how many rely on this very button already, but if you were e.g. using `bui-InputClear` to style it, you'd now need to use `bui-SearchFieldClear` instead to achieve the same style.

cdedreuille: You're right. I just changed that.

.changeset/itchy-bars-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': minor
+---
+
+Fixing styles on SearchField in Backstage UI after migration to CSS modules. `SearchField` has now its own set of class names. We previously used class names from `TextField` but this approach was creating some confusion so going forward in your theme you'll be able to theme `TextField` and `SearchField` separatly.
```
Rugvip:
```suggestion
Fixing styles on SearchField in Backstage UI after migration to CSS modules. `SearchField` has now its own set of class names. We previously used class names from `TextField` but this approach was creating some confusion so going forward in your theme you'll be able to theme `TextField` and `SearchField` separately.
```

Rugvip: Vale has spoken

cdedreuille: Fixed.

PR #31515, opened 2025-10-24
frontend-plugin-api: add support for relative attachments

Rugvip:

This adds support for relative attachment points in the new frontend system. It's a bit of a gap we detected where it's not currently possible to provide a hierarchy of blueprints that attach to each other in a nice way. This makes it possible to essentially have one blueprint attach to another without needing to explicitly declare the attachment point explicitly when using the child blueprint. The `kind` is currently required in order to constraint the design a bit, as this is primarily intended to be used with blueprints.

The current design only allows this to be done easily once per plugin, which I think might be enough. If we see that relative attachment points are used a lot with the need to create multiple hierarchies from the same blueprints within a single plugin, then I think we could consider adding the ability to be able to define partial relative attachment points when using a blueprint, i.e. `attachTo: { relative: { name: 'other-parent' } }`.

The entire implementation lives in the plugin API, with the resolution happening at the same time as we resolve the full extension IDs in the plugin.

### DIFF
--- a/.changeset/plugin-relative-extensions.md
+++ b/.changeset/plugin-relative-extensions.md
@@ -1,0 +1,17 @@
+---
+'@backstage/frontend-plugin-api': patch
+---
+
+Added support for plugin-relative `attachTo` declarations for extension definitions. This allows for the creation of extension and extension blueprints that attach to other extensions of a particular `kind` in the same plugin, rather than needing to provide the exact extension ID. This is particularly useful when wanting to provide extension blueprints with a built-in hierarchy where the extensions created from one blueprint attach to extensions created from the other blueprint, for example:
+
+```ts
+// kind: 'tabbed-page'
+const parentPage = TabbedPageBlueprint.make({
+  params: {....}
+})
+// attachTo: { kind: 'tabbed-page', input: 'tabs' }
+const child1 = TabContentBlueprint.make({
+  name: 'tab1',
+  params: {....}
+})
+```
--- a/packages/frontend-plugin-api/src/wiring/resolveExtensionDefinition.ts
+++ b/packages/frontend-plugin-api/src/wiring/resolveExtensionDefinition.ts
@@ -16,7 +16,7 @@
 
 import { ApiHolder, AppNode } from '../apis';
 import {
-  ExtensionAttachToSpec,
+  ExtensionDefinitionAttachTo,
   ExtensionDefinition,
   ExtensionDefinitionParameters,
   ResolvedExtensionInputs,
@@ -25,6 +25,17 @@ import { PortableSchema } from '../schema';
 import { ExtensionInput } from './createExtensionInput';
 import { ExtensionDataRef, ExtensionDataValue } from './createExtensionDataRef';
 import { OpaqueExtensionDefinition } from '@internal/frontend';
+
+/** @public */
+export type ExtensionAttachTo =
+  | { id: string; input: string }
+  | Array<{ id: string; input: string }>;
+
+/**
+ * @deprecated Use {@link ExtensionAttachTo} instead.
+ * @public
+ */
+export type ExtensionAttachToSpec = ExtensionAttachTo;
 
 /** @public */
 export interface Extension<TConfig, TConfigInput = TConfig> {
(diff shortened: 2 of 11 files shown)


packages/frontend-plugin-api/src/wiring/resolveExtensionDefinition.ts
```diff
@@ -126,6 +137,49 @@ export type ResolveExtensionId<
     : never
   : never;
 
+function resolveExtensionId(
+  kind?: string,
+  namespace?: string,
+  name?: string,
+): string {
+  const namePart =
+    name && namespace ? `${namespace}/${name}` : namespace || name;
```
freben, line 138: good thing we know all operator precedence by heart and never get confused by stuff like this right

(i know prettier helpfully does this :) )

.changeset/plugin-relative-extensions.md
```diff
@@ -0,0 +1,17 @@
+---
+'@backstage/frontend-plugin-api': patch
+---
+
+Added support for plugin-relative `attachTo` declarations for extension definitions. This allows for the creation of extension and extension blueprints that attach to other extensions of a particular `kind` in the same plugin, rather than needing to provide the exact extension ID. This is in particular useful when wanting to provide extension blueprints with a built-in heirarchy where the extensions created from one blueprint attach to extensions created from the other blueprint, for example:
```
freben:
```suggestion
Added support for plugin-relative `attachTo` declarations for extension definitions. This allows for the creation of extension and extension blueprints that attach to other extensions of a particular `kind` in the same plugin, rather than needing to provide the exact extension ID. This is particularly useful when wanting to provide extension blueprints with a built-in hierarchy where the extensions created from one blueprint attach to extensions created from the other blueprint, for example:
```

freben (commented):
Hm alright.

The one thing I miss I think is, isn't it common to make multiple such parents?

```ts
const fooPage = TabbedPageBlueprint.make({
  name: 'foo',
});
const fooPage = TabbedPageBlueprint.make({
  name: 'bar',
});

const fooTab = TabContentBlueprint.make({
  name: 'tab1',
  attachTo: { name: 'foo' }. // <-- kinda want this
})
```

In this case the kind and input are really what's well known (blueprint users aren't interested in repeating them).

It's almost like `createExtensionBlueprint` should accept an `attachTo` that's seen as a "set of defaults" basically. Whether you supply it as an ID or as a separate kind namespace name triplet, it holds on to them separately and then `ExtensionBlueprint.make` lets you override one or more of them. If you give an `id` to `make` then obviously you are fully overriding it. But if you give only `name`, then the others are kept at their defaults.

And if the original blueprint didn't specify a namespace at all, and no namespace was given to `make` either, then the end result is that it's a relative mount.

```ts
const TabbedPageBlueprint = createExtensionBlueprint({
  kind: 'tabbed-page',
  attachTo: { id: 'app:...'  }
});

const TabContentBlueprint = createExtensionBlueprint({
  attachTo: { kind: 'tabbed-page', input: 'tabs' }
});

const myTab = ExtensionBlueprint.make({
  attachTo: {} // <-- either left out entirely if your plugin made only one tab page, or `name: 'foo'`
});
```

packages/frontend-plugin-api/src/wiring/resolveExtensionDefinition.ts
Rugvip: yup ._.

Rugvip:
@freben yep kinda the stuff I was referring to in the PR body. I've explored it some more but it would require some fairly tricky logic. Tricky in the sense that it's hard to make it intuitive. For example, if the original is an array and the override is relative, what should happen? Or if the original is by `id`, but the override is `relative`?

I'm kinda landing at leaving this as is, i.e. no support for partial `attachTo` overrides like `attachTo: { relative: { name: 'other-page' } }`. Instead, I'm thinking we revive the idea of adding code references for `attachTo`, so that you can do the following instead: `attachTo: otherPage.inputs.tabs`. That should all be possible to implement on top of relative attachment definitions, as it turned out that was kinda the missing piece.

freben (approved):

PR #31522, opened 2025-10-24
feat: add generateProjectsWithAuthSetup function for Playwright e2e tests

joseerobless:

Add new function to generate Playwright projects with authentication setup dependency. This allows setup files to run first and share authentication state between tests.

### DIFF
--- a/packages/e2e-test-utils/src/playwright/generateProjects.ts
+++ b/packages/e2e-test-utils/src/playwright/generateProjects.ts
@@ -19,6 +19,9 @@ import { resolve as resolvePath } from 'path';
 import { PlaywrightTestConfig } from '@playwright/test';
 import { getPackagesSync } from '@manypkg/get-packages';
 import type { BackstagePackage } from '@backstage/cli-node';
+import { resolvePackagePath } from '@backstage/backend-plugin-api';
+
+const DEFAULT_STORAGE_STATE_PATH = '.auth/user.json';
 
 /**
  * Generates a list of playwright projects by scanning the monorepo for packages with an `e2e-tests/` folder.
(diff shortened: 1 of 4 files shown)


joseerobless:
@Rugvip could you please check this PR out?

packages/e2e-test-utils/src/playwright/generateProjects.ts
```diff
@@ -40,3 +40,33 @@ export function generateProjects(): PlaywrightTestConfig['projects'] {
...
+
+/**
+ * Generates a list of playwright projects by scanning the monorepo for packages with an `e2e-tests/` folder.
+ * Creates a setup project that runs first, followed by test projects that depend on it and use shared authentication state.
+ *
+ * @param authStateStoragePath - Path to the file where Playwright will store/read authentication state
+ * @returns Array of Playwright project configurations with setup dependencies
+ * @public
+ */
+export function generateProjectsWithAuthSetup(
```
aramissennyeydd, line 88: 👋 What's the goal with this addition? You should be able to generate this in your own project using the existing primitives

joseerobless: Auth is a near-universal pattern in Backstage portals. This helper handles the common case without forcing users to learn Playwright's project composition and setup mechanics, and helps keep the config file readable. The alternative would be to destructure and spread generateProjects() within the file, which gets convoluted:

```typescript
import { defineConfig } from '@playwright/test';
import { generateProjects } from '@backstage/e2e-test-utils/playwright';

export default defineConfig({
  // ... other config
  projects: [
    { name: 'setup', testMatch: /.*\.setup\.ts/ },
    ...generateProjects().map(project => ({
      ...project,
      use: {
        ...project.use,
        storageState: '.auth/state.json',
      },
      dependencies: ['setup'],
    })),
  ],
});
```

With the helper, it's just one readable line. We can make the pattern more customizable if you think that'd make sense too, I can add an optional testMatch parameter so they can still adjust when needed. Let me know what you think @aramissennyeydd! Apologies for the late reply

joseerobless: This matches Backstage's current pattern for projects where it's just a one-liner:

```typescript
projects: generateProjects(), // Find all packages with e2e-test folders
```

aramissennyeydd: @joseerobless No worries - my concern with this approach is that this method now has an implicit API with an adopter's setup.ts file having to write the `authStateStoragePath` and opens the door for more complexity as adopters would want to override various Playwright properties here. 

If we were able to provide a solution that had auth automatically configured for adopters (a `setup.ts` file out of the box + this change), I think that could be very useful.

joseerobless: I got the changes in, please let me know if there are any additional changes you'd like to see @aramissennyeydd & @Rugvip!

PR #31529, opened 2025-10-27
Fix dependency graph viewbox

SonilPro:

The relation graph is a bit too small now
The issue was introduced here https://github.com/backstage/backstage/pull/31299

Before:
image

After:
image

### DIFF
--- a/.changeset/real-trains-visit.md
+++ b/.changeset/real-trains-visit.md
@@ -3,4 +3,5 @@
 '@backstage/plugin-catalog-graph': patch
 ---
 
-Reverted back viewbox to old values. Also, added fullScreen prop to EntityCatalogGraphCard
+- Revert `viewbox` back to old values.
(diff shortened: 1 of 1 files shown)


SonilPro:
@drodil Hey, could you maybe shine some light on why you changed this originally?

drodil:
@SonilPro the change added the fullscreen control to the dependency graph. It unfortunately introduced some annoying bugs which I didn't catch at the time

SonilPro:
@drodil yeah, I saw your pr description. I am just curious about the viewbox change. In my instance of Backstage it appears way too small because of the change. Would it be possible to revert it back. From my testing everything seems to be working okay with the old values, but maybe I am missing something.

drodil:
I think it can be restored as long as it's checked properly not to cause side effects (the ancestry view behind "inspect entity" was the issue last time)

PR #31531, opened 2025-10-28
fix(notifications): show default settings before first notification

drodil:

this fixes default notification configuration not showing in the notification settings before user has received first notification from the specific origin/topic

### DIFF
--- a/plugins/notifications-backend/src/service/router.ts
+++ b/plugins/notifications-backend/src/service/router.ts
@@ -207,6 +207,27 @@ export async function createRouter(
     const settings = await store.getNotificationSettings({ user });
     const channels = getNotificationChannels();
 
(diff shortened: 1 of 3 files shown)


plugins/notifications-backend/src/service/router.ts
```diff
@@ -186,14 +186,31 @@ export async function createRouter(
       c => c.id.toLowerCase() === channelId.toLowerCase(),
     );
 
+    // Merge existing origins/topics with configured settings
+    const allOrigins = origins;
+    const allTopics = topics;
```
freben: Should you really be modifying the function parameter here?

drodil: True, it should be handled in the `getNotificationSettings` instead, will move

PR #31532, opened 2025-10-28
fix(notifications): ui improvements

drodil:

- long descriptions behind show more/less button to prevent very small scrollable area
- notifications count showing as `undefined` during initial loading
- removed unnecessary `Filter` heading to mimick catalog filters
- select all count is now more separated from the checkbox

image

### DIFF
--- a/plugins/notifications/src/components/NotificationsPage/NotificationsPage.tsx
+++ b/plugins/notifications/src/components/NotificationsPage/NotificationsPage.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useState, useMemo, useEffect } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import throttle from 'lodash/throttle';
 import {
   Content,
@@ -156,13 +156,17 @@ export const NotificationsPage = (props?: NotificationsPageProps) => {
   const isUnread = !!value?.[1]?.unread;
   const allTopics = value?.[2]?.topics;
 
-  let tableTitle = `All notifications (${totalCount})`;
+  let tableTitle = `All notifications `;
   if (saved) {
-    tableTitle = `Saved notifications (${totalCount})`;
+    tableTitle = `Saved notifications`;
   } else if (unreadOnly === true) {
-    tableTitle = `Unread notifications (${totalCount})`;
+    tableTitle = `Unread notifications`;
   } else if (unreadOnly === false) {
-    tableTitle = `Read notifications (${totalCount})`;
+    tableTitle = `Read notifications`;
+  }
+
+  if (totalCount) {
+    tableTitle += ` (${totalCount})`;
   }
 
   return (
--- a/plugins/notifications/src/components/NotificationsTable/NotificationDescription.tsx
+++ b/plugins/notifications/src/components/NotificationsTable/NotificationDescription.tsx
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2025 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


plugins/notifications/src/components/NotificationsTable/NotificationDescription.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const MAX_LENGTH = 100;
+
+export const NotificationDescription = (props: { description: string }) => {
+  const { description } = props;
+  const [shown, setShown] = useState(false);
+  const isLong = description.length > MAX_LENGTH;
+
+  if (!isLong) {
+    return <Typography variant="body2">description</Typography>;
```
hudsonb:
```suggestion
    return <Typography variant="body2">{description}</Typography>;
```

drodil: Good catch!

plugins/notifications/src/components/NotificationsTable/NotificationDescription.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Typography from '@material-ui/core/Typography';
+import Button from '@material-ui/core/Button';
+import { useState } from 'react';
+
+const MAX_LENGTH = 100;
+
+export const NotificationDescription = (props: { description: string }) => {
```
aramissennyeydd, line 22: Do we need/want to pull the `classes.description` theme values into here? Might be breaking if you already overrode them.

plugins/notifications/src/components/NotificationsPage/NotificationsPage.tsx
```diff
@@ -156,13 +156,17 @@ export const NotificationsPage = (props?: NotificationsPageProps) => {
...
+    tableTitle = `Saved notifications`;
   } else if (unreadOnly === true) {
-    tableTitle = `Unread notifications (${totalCount})`;
+    tableTitle = `Unread notifications`;
   } else if (unreadOnly === false) {
-    tableTitle = `Read notifications (${totalCount})`;
+    tableTitle = `Read notifications`;
+  }
+
+  if (totalCount) {
```
aramissennyeydd, line 168:
```suggestion
  if (totalCount !== undefined) {
```
Maybe? 0 would also show no count

drodil: Yes, don't want to so zero as there is content "no notifications"

plugins/notifications/src/components/NotificationsTable/NotificationDescription.tsx
drodil: Hm, don't think these are overridable 🤔 anyways, it will change completely once BUI is introduced

drodil: Hm, don't think these are overridable 🤔 anyways, it will change completely once BUI is introduced

plugins/notifications/src/components/NotificationsTable/NotificationDescription.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const isLong = description.length > MAX_LENGTH;
+
+  if (!isLong) {
+    return <Typography variant="body2">{description}</Typography>;
+  }
+
+  if (shown) {
+    return (
+      <Typography variant="body2">
+        {description}{' '}
```
freben, line 34: Should probably rather be a wrapper element that contains a Typography and then a Button as a sibling to that - with spacing / gap as needed. That would give better alignment control etc over the current solution that mixes things in the flow and uses a textual space for separation. But that can be done some other time.

PR #31537, opened 2025-10-29
fix: only add redirects for the stable build

aramissennyeydd:

Following up on the docs SIG - this allows us to handle redirects that only exist in `/next` and aren't yet stable. This is a bit hacky but the idea is that we don't actually need any redirects from our "next" build. In most cases, they're unstable and likely to change at any point. So we just don't generate them. To add a redirect without needing to wait for a release, just add it to the second array - it will get picked up when a new release is cut as normal.

### DIFF
--- a/microsite/docusaurus.config.ts
+++ b/microsite/docusaurus.config.ts
@@ -280,7 +280,7 @@ const config: Config = {
             ? [
                 {
                   from: '/docs/not-a-real-path',
-                  to: '/docs/auth/test',
+                  to: '/docs/auth/not-a-real-path',
                 },
               ]
             : []),


aramissennyeydd:
@Parsifal-M We didn't have any redirects for /next since this pattern really only shows up in CI. We should also no longer have any colliding redirects since we're only generating the redirects once :)

Parsifal-M (approved):
🚀🙏

.github/workflows/deploy_microsite.yml
```diff
@@ -307,7 +307,7 @@ jobs:
           path: docs
 
       - name: build microsite
-        run: yarn build
+        run: VERSION=next yarn build
```
Rugvip: Won't this always remove the redirects, we need to conditionally keep them for the stable build right?

microsite/docusaurus.config.ts
```diff
@@ -30,31 +30,6 @@ backstageTheme.plain.backgroundColor = '#232323';
...
-          // Bit of a hack to make sure all remaining redirects are written, since the write uses Promise.all
-          await new Promise(resolve => setTimeout(resolve, 1000));
-        } else {
-          throw error;
-        }
-      }
-    },
-  };
-};
-
```
Rugvip, line 57: Well this is a nice cleanup 😁

.github/workflows/deploy_microsite.yml
aramissennyeydd: It shouldn't? Confusingly this is the next build and the 
```
- name: generate stable docs
        run: yarn docusaurus docs:version stable
        working-directory: microsite
```
step above handles the stable build which should have the redirects

aramissennyeydd: actually - running the above command locally outputs no redirects. looking into this some more

microsite/docusaurus.config.ts
```diff
@@ -273,6 +276,14 @@ const config: Config = {
             from: '/docs/getting-started/app-custom-theme',
             to: '/docs/conf/user-interface',
           },
+          ...(semver.gt(backstageVersion, '1.40.0')
+            ? [
+                {
+                  from: '/docs/not-a-real-path',
+                  to: '/docs/auth/test',
```
aramissennyeydd:
```suggestion
                  to: '/docs/auth/not-a-real-path',
```

PR #31556, opened 2025-11-01
feat(app): enable per-plugin theming and routing context in NFS

stephenglass:

Adds a `data-plugin="<pluginId>"` attribute around routed app content, wired through a new route context. This is done automatically when using the `PageBlueprint` extension in the new frontend system (NFS). This enables theme authors and consumers of 3rd party plugins to target styles in specific plugins via MUI/BUI theme CSS without custom plugin overrides.

The app layout now wraps each plugin's rendered tree with a `data-plugin` attribute, updating on navigation via `PluginRouteContext`. The router integration `PluginAwareRoutes` pushes matched plugin Ids from route handles into that context.

Example of how theme authors can use this new functionality in BUI/MUI themes:
```theme.css
[data-plugin='catalog'] [class^='MuiToolbar-root'] {
  background: darkseagreen;
}

[data-plugin='api-docs'] [class^='MuiToolbar-root'] {
  background: lightblue;
}
```

image

image

### DIFF
--- a/.changeset/violet-cases-admire.md
+++ b/.changeset/violet-cases-admire.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-app': patch
 ---
 
-Expose the active plugin ID from the layout so apps can apply per-plugin theming. The app layout now wraps routed content with a `data-plugin` attribute whose value is updated on navigation.Plugin consumers can target `[data-plugin="<id>"]` selectors to style individual plugin pages without custom overrides.
+Expose the active plugin ID from the layout so apps can apply per-plugin theming. The app layout now wraps routed content with a `data-plugin` attribute whose value is updated on navigation. Plugin consumers can target `[data-plugin="<id>"]` selectors to style individual plugin pages without custom overrides.


stephenglass:
Not stale

stephenglass:
Not stale

vinzscam (changes requested):
Poking @backstage/design-system-maintainers for this.
I'm not sure this is the best way for doing this, especially when using internal class names that might change in the future (as `MuiToolbar-root` in the example)

stephenglass:
@vinzscam 
MUI - you can use the plugin data selector in theme js as well. PR description is a simplified example. 

BUI - last time I checked BUI all theme overrides will be done in css files and targeting css selectors

maxiforklift:
👋 Chiming in with support for this proposal!

Specifically on our Backstage instance, we want to set a max width & larger padding for the scaffolder wizard form - without targeting every MuiCard, or forking the whole scaffolder plugin.

cdedreuille:
Thanks for this update. From a capability perspective I personally think this update could help a lot for custom capabilities and as you pointed out it will work for BUI + MUI so on my side I'm totally fine with it @vinzscam.

PR #31561, opened 2025-11-02
Add TargetBoard Plugin to Directory - Create targetboard.yaml

elladziv:

I've added our new plugin for embedding TargetBoard dashboards and metric-cards into Backstage. I've added a targetboard.yaml file to the microsite/data/plugins folder.
Signed-off-by: Elad Ziv [elad@fatlady.co.il](mailto:elad@fatlady.co.il)

### DIFF
--- a/microsite/data/plugins/targetboard.yaml
+++ b/microsite/data/plugins/targetboard.yaml
@@ -1,10 +1,10 @@
 ---
 title: TargetBoard
-author: TargetBoard.ai
+author: TargetBoard
 authorUrl: https://www.targetboard.ai
 category: Metrics
 description: Embed TargetBoard dashboards and metric cards inside Backstage to visualize real-time engineering performance, KPIs, and delivery metrics.
-documentation: https://github.com/targetboard/backstage-plugin#readme
+documentation: https://github.com/targetboard/backstage-plugin/blob/main/README.md
 iconUrl: https://app.targetboard.ai/assets/TargetBoard-Backstage-logo.svg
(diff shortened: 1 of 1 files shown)


freben (approved):
Alright.

Some quick finds if you don't mind:

* The `embedId` [is set wrong](https://github.com/targetboard/backstage-plugin/blob/83148d352dece7e94439c8f3a3ae7780798d4d74/src/components/PluginPageComponent/PluginPageComponent.tsx#L14) so they will have the same value
* You can use a flex gap to get spacing between the cards instead of the nbsp
* The start/stop/on handlers [are redefined](https://github.com/targetboard/backstage-plugin/blob/83148d352dece7e94439c8f3a3ae7780798d4d74/src/components/TargetBoardEmbedCard/TargetBoardEmbedCard.tsx#L33) each render so the run-once `useEffect` will likely not be able to unregister the right callbacks. You may want to move all of the code inside the `useEffect` to that it gets defined just once. Just mentioning this since it might lead to resource leaks
* Is it ok to put the API keys directly in the dom and have them visible in the browser requests? If you like, you can consider a future version of the plugin that leverages the proxy plugin to inject auth behind users' backs or so. But that has its own set of quirks since it's an iframe to begin with, so up to you :)

elladziv:
Hi, @freben Thanks!

Will update my plugin accordingly.
The PluginPageComponent there is just a code example. It's not supposed to work.
As for the listeners, I'll use useCallback instead.

Thanks for all the help.

PR #31566, opened 2025-11-03
feat(ui): migrate Avatar from Base UI to custom implementation

Naycon:

Removed Base UI dependency from Avatar component and reimplemented with native HTML elements. Updated size scale with new x-small and x-large options.

### Breaking changes:

  - Base UI-specific props (render, etc.) are no longer supported
  - Component now uses native div/img elements instead of Base UI primitives
  - Size scale updated: large changed from 3rem to 2.5rem
  - Added x-small (1.25rem) and x-large (3rem) sizes

#### Migration:

```diff
# Remove Base UI-specific props
- <Avatar src="..." name="..." render={...} />
+ <Avatar src="..." name="..." />

# Update large to x-large for same visual size
- <Avatar size="large" />
+ <Avatar size="x-large" />
```

### New features:

- Added `purpose` prop with `informative` (default) and `decoration` options
- Informative avatars announce name to screen readers
- Decorative avatars hidden from screen readers (use when name appears adjacent)
- Five size options: x-small, small, medium, large, x-large

### Documentation updates:

  - Updated size examples to show all five sizes
  - Added Purpose story and documentation
  - Updated prop definitions and usage examples
  - Updated changeset with migration guide for size changes


packages/ui/src/components/Avatar/Avatar.tsx
```diff
@@ -14,48 +14,72 @@
...
+    .toLocaleUpperCase('en-US')
+    .slice(0, 2);
 
   return (
-    <AvatarPrimitive.Root
+    <div
       ref={ref}
+      role="img"
+      aria-label={purpose === 'informative' ? name : undefined}
+      aria-hidden={purpose === 'decoration' ? true : undefined}
```
Rugvip, line 62: This feels a little bit clunky somehow, feels enough for callers to just set `aria-hidden`, or do we need to make sure the label is unset for that case as well? Or `role=presentation`? although not sure about that route

packages/ui/src/components/Avatar/types.ts
```diff
@@ -14,12 +14,29 @@
...
+   * Name of the person - used for generating initials and accessibility labels
+   */
   name: string;
-  size?: 'small' | 'medium' | 'large';
+
+  /**
+   * Size of the avatar
+   * @defaultValue 'medium'
+   */
+  size?: 'x-small' | 'small' | 'medium' | 'large' | 'x-large';
```
Rugvip, line 33: Are the more extreme ones here actually useful in practice? as opposed to falling back to exact values instead

packages/ui/src/components/Avatar/Avatar.tsx
Naycon: Sorry, had turned on automerge (I think I'll avoid that going forward) so this got merged before I could respond.

Yea, I can understand the sentiment of this feeling a little clunky. I had two intentions with this:
1) Make sure the markup is always correct from an a11y perspective (hidden + label + img alt)
2) Surface a11y as a primary concern for the component rather than something secondary you manage through another API.

The reasoning behind 2) is that many developers are unaware of or haven't incorporated looking out for a11y concerns as part of their daily work. My hope was that with this approach, we can help users take a conscious decision and discover that there are more aspects to consider than what shows on screen.

I'm happy to iterate on this, both here and in other BUI components going forward, to find the right level of surfacing concerns, highlighting decisions to take, and to make it convenient to cater for all users, a11y or not.

packages/ui/src/components/Avatar/types.ts
Naycon: They all came from @cdedreuille. The reason for `x-small` was because there's a need for it in the CellProfile component, where we want to replace Base UI with this component. For the `large` - `x-large` swap, I think this is for consistency. Excluding the x-small size, there's now a linear scale of sizes that I assume should cover what we (Charles + team) foresee as the need going forward. 

Reg. exact values, I think we want to promote consistency, so for that reason it would be nice to end up in a spot where available values cover most of the necessary situations. If someone _really_ want to have their own custom size, they can still do that via class + css ofc. However, if we can avoid this as much as possible while not having too many size, then I think we're in a better position to create a consistent experience across Backstage and plugins.

PR #31597, opened 2025-11-03
app-visualizer: fix tree rendering and add to create-app

Rugvip:

Started out as a change to add `app-visualizer` to `create-app`, but realized the layout is a bit broken so included a change for that as well:

image

### DIFF
--- a/.changeset/eighty-results-prove.md
+++ b/.changeset/eighty-results-prove.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-app-visualizer': patch
+---
+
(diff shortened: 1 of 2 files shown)


freben (approved):
Amaze. Will this make it appear in the sidebar on every new app then? So people will need to hide it later down the line as they expose the app to end users, until we have a better place to put it or have some form of admin/dev roles to bind it to?

Rugvip:
@freben yep it will. I figured it's very useful as you're starting up and indeed, if you wanna hide it down the line that can be done through config. Would also be nice to move it eventually too, ye

PR #31638, opened 2025-11-05
[catalog-unprocessed-entities] Move types and clients from frontend to common plugin

ivangonzalezacuna:

Moved some types, as well as the API and client definitions to the common package.

This allows not only frontend plugins, but also backend ones, to use this client if needed. Until now, and since the types and all definitions were stored in the frontend plugin, the only way to implement such feature was to copy the code in the project. It would make much more sense to make this exportable from the plugin itself and let people use this plugin if needed.

I made this change a patch, since all types have been initially deprecated and a mention to the new location has been added.

I have tested the new setup in the frontend plugin, as well as in a backend plugin (which needs sending a token like the catalogClient), and both worked perfectly fine.

### DIFF
--- a/plugins/catalog-unprocessed-entities/src/api/index.ts
+++ b/plugins/catalog-unprocessed-entities/src/api/index.ts
@@ -13,13 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import { createApiRef } from '@backstage/core-plugin-api';
 import {
-  DiscoveryApi,
-  createApiRef,
-  FetchApi,
-} from '@backstage/core-plugin-api';
-import { ResponseError } from '@backstage/errors';
-import { UnprocessedEntity } from '../types';
+  CatalogUnprocessedEntitiesApiResponse as CommonCatalogUnprocessedEntitiesApiResponse,
+  CatalogUnprocessedEntitiesApi as CommonCatalogUnprocessedEntitiesApi,
+  CatalogUnprocessedEntitiesClient as CommonCatalogUnprocessedEntitiesClient,
+} from '@backstage/plugin-catalog-unprocessed-entities-common';
 
 /**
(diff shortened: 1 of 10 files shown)


plugins/catalog-unprocessed-entities-common/src/api/types/discovery.ts
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This is a copy of the DiscoveryApi, to avoid importing core-plugin-api.
+ */
+export type DiscoveryApi = {
+  getBaseUrl(pluginId: string): Promise<string>;
```
benjdlambert: I think it's probably fine just to delete this type alias and just inline it instead like you already have in other cases.

https://github.com/backstage/backstage/pull/31638/files#diff-9ad86f6bb88c6a9d31e56106e49e385ff77d4efc8966bc4843cf91f00db363a8R77-R78

Same goes for `fetchApi` it's not really worth having these in the public types.

ivangonzalezacuna: Should be ready now. I've also rebased with master :smile:

plugins/catalog-unprocessed-entities/src/api/index.ts
aramissennyeydd, line 1: Can we make these re-exports of the -common values instead of duplicated here? Reduces confusion for future work on this plugin.

ivangonzalezacuna: Sorry that it took so long. But I could finally give this another look. I've done exactly that :)

ivangonzalezacuna:
I've resolved all the issues caused by the upgrade. Now it can be merged again if approved

PR #31649, opened 2025-11-05
feat(ui): add searchable and multiselect support to Select component

Naycon:

Enhanced the Select component with search filtering and multi-selection capabilities through new searchable, selectionMode, and searchPlaceholder props.

image

### Breaking changes

- The SelectProps interface now accepts a generic type parameter for selection mode

### Implementation details

- Created SelectTrigger, SelectContent, and SelectListBox components for modular composition
- Integrated react-aria's Autocomplete and SearchField for search functionality
- Added support for multiple selection mode
- Added "No results found" empty state when search returns no matches
- Improved CSS organization and updated visual styling for better consistency
- Exported Option type as public API for type safety
- Updated API reports and added comprehensive Storybook stories
- Added documentation with examples for searchable and multiple selection modes

### Migration guide

If using SelectProps type directly, update the type parameter:

```diff
- SelectProps
+ SelectProps<'single' | 'multiple'>
```

Component usage remains backward compatible - existing Select implementations require no changes.


Naycon:
The 28 updated Chromatic story diffs are because this PR also fixes a bug I didn't realize we had - the Select did not respect the placeholder you gave it and used a default one always. This PR fixed that by passing the placeholder on to the React Aria Select properly.

benjdlambert:
@Naycon @cdedreuille a question here on usability. 

What's the recommended approach for async items here, are we expecting that you pass one option which is `Loading...` which is `disabled`? Or heading down a `loading` prop approach instead where the component deals with it's own loading state internally?

Naycon:
> @Naycon @cdedreuille a question here on usability.
> 
> What's the recommended approach for async items here, are we expecting that you pass one option which is `Loading...` which is `disabled`? Or heading down a `loading` prop approach instead where the component deals with it's own loading state internally?

@benjdlambert Great question! I briefly thought about this but decided to keep it out of this PR. React Aria has [good support for async lists](https://react-spectrum.adobe.com/react-aria/Select.html#asynchronous-loading), but we need to think about what the API for it should look like. I'm not sure if we'd direct people to use the async hook directly, or if we'd expose that in some other way. And depending on what we decide there (and as you were asking), we do need to think about how you would indicate loading is in progress (initial vs loading more items, etc.). It may also mean we need to modify the way options are passed to the Select.

For now, as a workaround, I think what you could do is to manage the loading yourself and do as you said and pass a disabled single option with an appropriate label. I'm not sure it's great from an a11y perspective, but I hope it would at least get the message across.

PR #31655, opened 2025-11-05
feat: add scheduled tasks UI to devtools plugin

williamwu-mongodb:

Screenshot 2025-11-05 at 10 27 52 AM

scheduled tasks devtools

I added a UI feature which utilizes the REST api for the scheduler service - https://backstage.io/docs/backend-system/core-services/scheduler/#rest-api - and allows the user to have access to this in the devtools plugin to view, debug, ad-hoc trigger scheduled tasks.

### DIFF
--- a/plugins/devtools-common/src/permissions.ts
+++ b/plugins/devtools-common/src/permissions.ts
@@ -61,7 +61,7 @@ export const devToolsTaskSchedulerReadPermission = createPermission({
  */
 export const devToolsTaskSchedulerCreatePermission = createPermission({
   name: 'devtools.scheduler.trigger',
-  attributes: { action: 'create' },
+  attributes: { action: 'update' },
 });
 
 /**


plugins/devtools-common/src/types.ts
```diff
@@ -14,7 +14,41 @@
  * limitations under the License.
  */
 
-import { JsonValue } from '@backstage/types';
+import { JsonObject, JsonValue } from '@backstage/types';
+
+/** @public */
+export type TaskApiTasksResponse = {
```
awanlin: Is this type directly from the Scheduler side? I wonder if we should just export it from there?

plugins/devtools-backend/src/api/DevToolsBackendApi.ts
```diff
@@ -34,17 +36,107 @@ import { Lockfile } from '../util/Lockfile';
...
     this.config = config;
   }
 
+  public async triggerScheduledTask(
+    plugin: string,
+    taskId: string,
+    token: string,
+  ): Promise<TriggerScheduledTask> {
+    const backendUrl = this.config.getString('backend.baseUrl');
+    const triggerTaskUrl = `${backendUrl}/api/${plugin}/${DevToolsBackendApi.SCHEDULED_TASKS_PATH}/${taskId}/trigger`;
```
aramissennyeydd: Let's use the discovery API here as well, I see you have it already on the client.

plugins/devtools-backend/src/service/router.ts
```diff
@@ -113,5 +116,68 @@ export async function createRouter(
...
+      const triggerResponse = await devToolsBackendApi.triggerScheduledTask(
+        plugin,
+        taskId,
+        token,
+      );
+
+      if (triggerResponse.error) {
+        response.status(500).json(triggerResponse);
+        return;
+      }
```
aramissennyeydd: Can we just throw from our API client instead? We support all of the major status codes as Errors.

plugins/devtools-common/src/permissions.ts
aramissennyeydd, line 1: Hmm, I think we'd want to use https://github.com/backstage/backstage/pull/30533 when that goes in instead of creating our own rules.

plugins/devtools-backend/src/service/router.ts
```diff
@@ -113,5 +116,68 @@ export async function createRouter(
...
+
+      response.status(200);
+    },
+  );
+
+  router.get('/scheduled-tasks/:plugin', async (req, response) => {
+    const credentials = await httpAuth.credentials(req);
+    const decision = (
+      await permissions.authorize(
+        [{ permission: devToolsExternalDependenciesReadPermission }],
```
awanlin: Copy and paste bug? This probably should be a different permission.

plugins/devtools-common/src/permissions.ts
```diff
@@ -48,6 +48,22 @@ export const devToolsExternalDependenciesReadPermission = createPermission({
...
+export const devToolsTaskSchedulerReadPermission = createPermission({
+  name: 'devtools.task-scheduler',
+  attributes: { action: 'read' },
+});
+
+/**
+ * @public
+ */
+export const devToolsTaskSchedulerCreatePermission = createPermission({
+  name: 'devtools.external-dependencies',
```
awanlin: Copy and paste typo.

plugins/devtools/src/components/Content/ScheduledTasksContent/ScheduledTasksContent.tsx
```diff
@@ -0,0 +1,257 @@
...
+    </Typography>
+  </Box>
+);
+
+/** @public */
+export const ScheduledTasksContent = () => {
+  const classes = useStyles();
+  const configApi = useApi(configApiRef);
+  const alertApi = useApi(alertApiRef);
+  const plugins = configApi.getStringArray('devTools.scheduledTasks.plugins');
```
awanlin: I'm torn on this one, having to manually add these seems a bit tedious but I'm also not aware of any API that will give you a list of all the plugins that use the scheduler. I wonder if the middle ground for now it to allow people to type in a value if it's not in the list? 🤔

plugins/devtools/README.md
```diff
@@ -456,3 +456,37 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
...
+Scheduled tasks can be viewed and triggered in the DevTools plugin under the `Scheduled Tasks` tab. You first must add the list of plugins for scheduled tasks to your config:
+
+```yaml
+devTools:
+  scheduledTasks:
+    plugins:
+      - catalog
+```
+
+Then add a new `DevToolsLayout.Route` to the end of your `DevToolsLayout` in your `CustomDevToolsPage.tsx` like this:
```
awanlin: I think this should be included out-of-the-box as part of the default DevTools tab so this won't be needed. Can you also make the changes so it is part of the default tabs?

plugins/devtools/README.md
```diff
@@ -456,3 +456,37 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     apt-get update && \
     apt-get install -y  ... iputils-ping
 ```
+
+### Scheduled Tasks Configuration
```
awanlin, line 466: Can you add a screen shot and description for this to the Features section? It would be the last items just before the Optional Feature section.

plugins/devtools/README.md
awanlin: There will also be changes to the main customization section to include this new tab there.

plugins/devtools-backend/src/api/DevToolsBackendApi.ts
awanlin, line 1: We likely don't need any of the backend plugin changes as we can call the Scheduler REST API from the frontend plugin. Is there anything that you can think of that would require it @williamwu-mongodb

williamwu-mongodb: i removed the backend plugin changes, we are going to just call the scheduler API directly from the devtools client

plugins/devtools/src/components/Content/ScheduledTasksContent/ScheduledTasksContent.tsx
williamwu-mongodb: so i added an autocomplete functionality, where the user can type the value if its not listed in the config, however I think having a list of default plugins from the config is still the way to go until we have an endpoint that returns a list of plugins that use the scheduler

williamwu-mongodb:
@awanlin - this is ready for another look!

plugins/devtools-common/src/types.ts
williamwu-mongodb: @awanlin unfortunately since `TaskApiTasksResponse` lives in `backend-defaults` package, i am unable to import it into the devtools frontend plugin since the linter blocks me from doing so. Please advise, thanks!

https://github.com/backstage/backstage/actions/runs/19864969067/job/56925245451?pr=31655

awanlin: @freben, any suggestions on this?

plugins/devtools-common/src/permissions.ts
aramissennyeydd: @williamwu-mongodb Do we still need these new perms?

awanlin: As it's unclear the path forward in #30533 I'm willing to take on the potential breaking change this might introduce. I'd really like to ship this feature before the end of the year.

williamwu-mongodb: @aramissennyeydd my thoughts on this is that the permissions can be used on the FE side - so users can control whether or not they would like to exclude this from the default devtools page (per @awanlin's feedback, i did include this tab by default in the default devtools page)

aramissennyeydd: ah sorry missed that this was for the frontend 😅

plugins/devtools-common/src/types.ts
awanlin: I've followed up with the Maintainers about this just now, will follow up when I have an answer but for know if you don't mind reverting to how you had it duplicated in the common package that would be good 👍

awanlin: Maintainers agree duplication is fine for now but let's add a comment on the duplicate type:

`// NOTE: this needs to be kept in sync with ...`

Pointing to its source location 👍

plugins/devtools/README.md
```diff
@@ -34,6 +34,19 @@ Lists the configuration being used by your current running Backstage instance.
 
 ![Example of Config tab](./docs/devtools-config-tab.png)
 
+### Scheduled Tasks
+
+Scheduled tasks can be viewed and triggered under the `Scheduled Tasks` tab. You first must add the list of plugins for scheduled tasks to your config:
+
+```yaml
```
awanlin: For consistency can we move this down to the configuration section towards the bottom and then you can link to that from here.

packages/app/src/components/devtools/CustomDevToolsPage.tsx
```diff
@@ -43,6 +44,9 @@ const DevToolsPage = () => {
       >
         <UnprocessedEntitiesContent />
       </DevToolsLayout.Route>
+      <DevToolsLayout.Route path="scheduled-tasks" title="Scheduled Tasks">
+        <ScheduledTasksContent />
+      </DevToolsLayout.Route>
```
awanlin: Can we move this up just so it's closer to the out of the box experience? So Info, Config, Scheduled Tasks, other tabs.

plugins/devtools-backend/src/plugin.ts
awanlin, line 1: Think this got missed when you reverted the backend changes, don't think it's needed anymore.

plugins/devtools-backend/report.api.md
awanlin, line 1: This likely needs to be regenerated, there's no more backend changes at least there shouldn't be.

plugins/devtools/docs/devtools-scheduled-tasks-tab.png
awanlin, line 1: Can we update this with one row open just showing that you can also see the details?

plugins/devtools/src/components/Content/ScheduledTasksContent/ScheduledTasksContent.tsx
```diff
@@ -0,0 +1,279 @@
...
+      render: (rowData: TaskApiTasksResponse) =>
+        rowData.taskState?.status === 'idle' && rowData.taskState.startsAt
+          ? new Date(rowData.taskState.startsAt).toLocaleString()
+          : 'N/A',
+    },
+    {
+      title: 'Actions',
+      render: (rowData: TaskApiTasksResponse) => (
+        <RequirePermission permission={devToolsTaskSchedulerCreatePermission}>
+          <Tooltip title="Refresh">
```
awanlin:
```suggestion
          <Tooltip title="Run Task">
```

I'm open to other wording just Refresh doesn't feel right. I initially thought Refresh was going to refresh the data in the row but it actually runs the task.

plugins/devtools/src/components/Content/ScheduledTasksContent/ScheduledTasksContent.tsx
```diff
@@ -0,0 +1,279 @@
...
+            variant="outlined"
+            onKeyDown={handleKeyDown}
+            onBlur={handleCommitChange}
+          />
+        )}
+      />
+
+      {loading && <Progress />}
+
+      {error && <Alert severity="error">{error}</Alert>}
```
awanlin: I wonder if we can handle this a bit nicer? When I use `techdocs` I get this:

Image

Having an empty state like component with some helpful details like "The plugin id provided doesn't have any scheduled task or may have a typo, please correct"

.changeset/short-lizards-find.md
```diff
@@ -0,0 +1,8 @@
+---
+'@backstage/backend-defaults': minor
+'@backstage/plugin-devtools-backend': minor
+'@backstage/plugin-devtools-common': minor
+'@backstage/plugin-devtools': minor
+---
+
+added scheduled tasks UI feature for the devtools plugin
```
awanlin:
```suggestion
Added scheduled tasks UI feature for the DevTools plugin
```

awanlin: That should resolve the error.

awanlin (changes requested):
Thanks for the updated @williamwu-mongodb, left some more comments but we are getting very close. I'd love to get this in for the `next release` on Tuesday so I'll actively check on this for updates 👍

williamwu-mongodb:
@awanlin this is ready for another review!

.changeset/short-lizards-find.md
```diff
@@ -0,0 +1,8 @@
+---
+'@backstage/backend-defaults': minor
+'@backstage/plugin-devtools-backend': minor
```
awanlin: These two lines need to be removed as there is no changes made to them.

.changeset/short-lizards-find.md
```diff
@@ -0,0 +1,8 @@
+---
+'@backstage/backend-defaults': minor
```
awanlin:
```suggestion

```

.changeset/short-lizards-find.md
awanlin:
```suggestion

```

awanlin: Pushed my suggestions so this is resolve and we can move forward with approving and getting it merged.

plugins/devtools/config.d.ts
```diff
@@ -13,8 +13,18 @@
...
-
-/* eslint-disable-next-line no-restricted-syntax */
-export const paths = findPaths(__dirname);
+export interface Config {
+  devTools: {
+    /**
+     * Scheduled tasks configuration
+     * @visibility frontend
+     */
+    scheduledTasks: {
```
freben:
```suggestion
  devTools?: {
    /**
     * Scheduled tasks configuration
     * @visibility frontend
     */
    scheduledTasks?: {
```

plugins/devtools-common/report.api.md
```diff
@@ -83,4 +90,57 @@ export type PackageDependency = {
...
 };
+
+// @public (undocumented)
+export type ScheduledTasks = {
+  scheduledTasks?: TaskApiTasksResponse[];
+  error?: string;
+};
+
+// @public
+export interface TaskApiTasksResponse {
```
freben: You're now making "foreign" types (that aren't technically owned by the devtools plugin) as part of the public non-alpha contract. Arguably this should eventually find a better place to live, maybe in `@backstage/backend-plugin-api` or a new `@backstage/backend-client`. Should it be in `/alpha` from the get-go just as an unblocking measure for now? Then you're slightly more free to just delete them once they have found a proper home.

plugins/devtools-common/report.api.md
```diff
@@ -44,6 +45,12 @@ export const devToolsInfoReadPermission: BasicPermission;
 // @public
 export const devToolsPermissions: BasicPermission[];
 
+// @public (undocumented)
+export const devToolsTaskSchedulerCreatePermission: BasicPermission;
+
+// @public (undocumented)
+export const devToolsTaskSchedulerReadPermission: BasicPermission;
```
freben: Hm these permissions should perhaps also go on the actual service later, right? One more thing that maybe should be alpha with a plan of moving it away later

plugins/devtools-common/src/permissions.ts
```diff
@@ -48,6 +48,22 @@ export const devToolsExternalDependenciesReadPermission = createPermission({
...
+export const devToolsTaskSchedulerReadPermission = createPermission({
+  name: 'devtools.task-scheduler',
+  attributes: { action: 'read' },
+});
+
+/**
+ * @public
+ */
+export const devToolsTaskSchedulerCreatePermission = createPermission({
+  name: 'devtools.task-scheduler',
```
freben: these need to have unique names i think 🤔 right @vinzscam

.changeset/short-lizards-find.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-devtools-common': minor
+'@backstage/plugin-devtools': minor
```
freben:
```suggestion
'@backstage/plugin-devtools-common': patch
'@backstage/plugin-devtools': patch
```

plugins/devtools/src/components/Content/ScheduledTasksContent/ScheduledTasksContent.tsx
```diff
@@ -0,0 +1,285 @@
...
+            onKeyDown={handleKeyDown}
+            onBlur={handleCommitChange}
+          />
+        )}
+      />
+
+      {loading && <Progress />}
+
+      {error && (
+        <Alert severity="warning">
```
freben: This seems like a good fit for `<ErrorPanel title="Could not fetch scheduler state" error={error}` or something like that, with more detailed info in its children.

plugins/devtools/src/components/Content/ScheduledTasksContent/ScheduledTasksContent.tsx
```diff
@@ -0,0 +1,285 @@
...
+    {
+      title: 'Actions',
+      render: (rowData: TaskApiTasksResponse) => (
+        <RequirePermission permission={devToolsTaskSchedulerCreatePermission}>
+          <Tooltip title="Run Task">
+            <IconButton
+              aria-label="Trigger"
+              onClick={() => {
+                triggerTask(selectedPlugin, rowData.taskId);
+                if (isTriggering) {
```
freben: this code doesn't do what you expect - it's trying to render inside a callback handler

plugins/devtools/src/components/Content/ScheduledTasksContent/ScheduledTasksContent.tsx
```diff
@@ -0,0 +1,285 @@
...
+          : 'N/A',
+    },
+    {
+      title: 'Actions',
+      render: (rowData: TaskApiTasksResponse) => (
+        <RequirePermission permission={devToolsTaskSchedulerCreatePermission}>
+          <Tooltip title="Run Task">
+            <IconButton
+              aria-label="Trigger"
+              onClick={() => {
```
freben, line 206: you want to disable clicking on this while refreshing, right?

plugins/devtools-common/report.api.md
williamwu-mongodb: yes it shoudl be in `/alpha` so its unblocking

plugins/devtools-common/src/permissions.ts
williamwu-mongodb: if i understand permissions correctly, since the attribute is unique, the permission name doesn't have to be?

williamwu-mongodb:
@freben this is ready for another review!

plugins/devtools-common/src/permissions.ts
```diff
@@ -48,6 +48,22 @@ export const devToolsExternalDependenciesReadPermission = createPermission({
   attributes: { action: 'read' },
 });
 
+/**
+ * @alpha
+ */
+export const devToolsTaskSchedulerReadPermission = createPermission({
+  name: 'devtools.task-scheduler',
```
awanlin:
```suggestion
  name: 'devtools.scheduler.read',
```

plugins/devtools-common/src/permissions.ts
```diff
@@ -48,6 +48,22 @@ export const devToolsExternalDependenciesReadPermission = createPermission({
...
+  attributes: { action: 'read' },
+});
+
+/**
+ * @alpha
+ */
+export const devToolsTaskSchedulerCreatePermission = createPermission({
+  name: 'devtools.task-scheduler',
```
awanlin:
```suggestion
  name: 'devtools.scheduler.trigger',
```

plugins/devtools-common/src/permissions.ts
```diff
@@ -48,6 +48,22 @@ export const devToolsExternalDependenciesReadPermission = createPermission({
...
+  attributes: { action: 'create' },
```
awanlin:
```suggestion
  attributes: { action: 'update' },
```

plugins/devtools/src/components/Content/ScheduledTasksContent/ScheduledTasksContent.tsx
```diff
@@ -0,0 +1,307 @@
...
+      title: 'Actions',
+      render: (rowData: TaskApiTasksResponse) => (
+        <RequirePermission permission={devToolsTaskSchedulerCreatePermission}>
+          <Tooltip title="Run Task">
+            <IconButton
+              aria-label="Trigger"
+              disabled={isTriggering}
+              onClick={() => {
+                triggerTask(selectedPlugin, rowData.taskId);
+                if (triggerError) {
```
freben, line 208: This still doesn't do the right thing - `triggerTask` just gets deferred in the background and then this `if` check immediately compares against old state from the previous render - before the network call to perform the trigger has even been sent.

Normally you could do something like

```ts
try {
  await triggerTask(...); // note waiting for the outcome
  alertApi.post(/* ok */);
catch (e) {
  alertApi.post(/* error */);
}
```

but in your hook you don't propagate the error. So that won't work here. Either you move the alertApi posts into the hook, or restructure how the hook works. Moving into the hook is probably easiest.

plugins/devtools/src/hooks/useScheduledTasks.ts
```diff
@@ -0,0 +1,44 @@
...
+      scheduledTasks: undefined,
+      loading: false,
+      error: asyncError.message,
+    };
+  }
+
+  return {
+    scheduledTasks: value?.scheduledTasks,
+    loading,
+    error: value?.error,
```
freben, line 42: `value?.error` doesn't really happen, does it? 🤔 You have already thrown an error inside `getScheduledTasksByPlugin` if it wasn't successful, and that's caught by the `if` above

PR #31676, opened 2025-11-06
BUI - Color improvements on surfaces

cdedreuille:

This PR introduce a new way for components to render on different surfaces. BUI has 3 main surfaces and previously some components would not render well if placed in other surfaces. The idea we are shaping here is to allow for certain component to have a new `on` prop to adapt to their background.

To make it easier for anyone building components, these components will automatically get the correct `on` prop based on the component they are on. As an example, if you place a `<Button />` in a `<Box bg='surface-1' />`, your button will automatically get its `on` prop to `'surface-1`

### New color system

🌘 Dark Backstage UI

### Migration:
```diff
- --bui-bg-tint              No replacement
- --bui-bg-tint-hover        No replacement
- --bui-bg-tint-pressed      No replacement
- --bui-bg-tint-disabled     No replacement
- --bui-bg                   To be replaced by --bui-bg-surface-0

+ --bui-bg-surface-0

+ --bui-bg-on-surface-0
+ --bui-bg-hover-on-surface-0
+ --bui-bg-pressed-on-surface-0
+ --bui-bg-disabled-on-surface-0

+ --bui-bg-on-surface-1
+ --bui-bg-hover-on-surface-1
+ --bui-bg-pressed-on-surface-1
+ --bui-bg-disabled-on-surface-1

+ --bui-bg-on-surface-2
+ --bui-bg-hover-on-surface-2
+ --bui-bg-pressed-on-surface-2
+ --bui-bg-disabled-on-surface-2
```

### Components impacted
- `Box` - Added the new `bg` prop as well as the `SurfaceProvider`.
- `Flex` - Added the new `bg` prop as well as the `SurfaceProvider`.
- `Grid` - Added the new `bg` prop as well as the `SurfaceProvider`.
- `Button` - Added the new `on` prop.


.changeset/tame-eggs-fetch.md
```diff
@@ -0,0 +1,16 @@
...
+'@backstage/ui': minor
+---
+
+This adds a new automated way for components to understand on what surface they are in Backstage UI. `Box`, `Flex`, and `Grid` will now have `bg` prop and a provider that will connect with any components inside to adapt their styling.
+
+Here's an example:
+
+```tsx
+// If a Button is inside of a Box that has a set background value, then the Button will automatically adapt its style.
+<Box bg="surface-1">
```
Naycon, line 11: Should we consider calling the prop `surface` instead of `bg`? E.g. `surface="1"` or `surface={1}`. Just to have fewer concepts and terms to keep track of. It would also somewhat future-proof this if we want to affect more than just the background of a surface.

packages/ui/src/components/Box/Box.tsx
```diff
@@ -40,9 +39,16 @@ export const Box = forwardRef<HTMLDivElement, BoxProps>((props, ref) => {
         className,
       ),
       style,
+      ...dataAttributes,
       ...rest,
     },
-    children,
+    bg ? (
+      <SurfaceProvider surface={bg as unknown as SurfaceLevel}>
```
Naycon: Hmm, is there any way we could type the `useStyles` hook better to avoid this? Would be neat if the hook would output a union of the possible values when the input is responsive. 

(Btw, maybe I should ask - I assume the hook guarantees that the cleaned bg prop is one of the SurfaceLevel values, it's just not typing it correctly, right?)

packages/ui/src/hooks/useStyles.ts
```diff
@@ -171,12 +171,15 @@ export function useStyles<
     }
   }
 
-  // Create cleaned props by excluding only utility props
-  // All other props (including data attributes, style, children, etc.) remain
-  const utilityPropsSet = new Set<string>(utilityPropNames);
+  // Create cleaned props by excluding utility props and data attribute props
```
Naycon, line 174: Is this hook part of the public API of BUI? If so, this should be documented as a breaking change.

packages/ui/src/index.ts
Naycon, line 1: Do we need to export the surface provider and hook to let people create their own BUI components? Say I want to create a surface-aware graph component for my plugin, then I'd need the hook, right?

packages/ui/src/components/Avatar/Avatar.tsx
```diff
@@ -47,7 +47,7 @@ export const Avatar = forwardRef<HTMLDivElement, AvatarProps>((props, ref) => {
     };
   }, [src]);
 
-  const initialsCount = ['x-small', 'small'].includes(cleanedProps.size)
+  const initialsCount = ['x-small', 'small'].includes(props.size || 'medium')
```
Naycon, line 50: Won't this break if I pass a responsive value to `size`?

.changeset/tame-eggs-fetch.md
cdedreuille: After discussion we decided to go with this new structure and include these surfaces values:
`"0" | "1" | "2" | "danger" | "warning" | "success"`

```tsx
<Box surface="1">
  <Button onSurface="1">Hello World</Button>
</Box>
```

packages/ui/src/hooks/useStyles.ts
cdedreuille: No it is not public

packages/ui/src/index.ts
cdedreuille: Perhaps in the future but for now I'm kind of inclined to do this gradually and see how it goes first before exporting too much.

.changeset/tame-eggs-fetch.md
Rugvip: In that example, do you need to set `onSurface` of the button, or would it be defaulted? I.e. would this be equivalent?

```tsx
<Box surface="1">
  <Button>Hello World</Button>
</Box>
```

cdedreuille: Yes exactly. This will be set automatically.

PR #31732, opened 2025-11-12
fix(core-components): Add i18n support for LogViewer search control

christoph-jerolimov:

Super small change to make the LogViewer "search" placeholder (used in techdocs plugins) "translatable":

Before:

image

With this PR and the right translation:

image

### DIFF
--- a/packages/core-components/src/components/LogViewer/LogViewerControls.tsx
+++ b/packages/core-components/src/components/LogViewer/LogViewerControls.tsx
@@ -15,17 +15,20 @@
  */
 
 import { KeyboardEvent } from 'react';
+import { useTranslationRef } from '@backstage/core-plugin-api/alpha';
 import IconButton from '@material-ui/core/IconButton';
 import TextField from '@material-ui/core/TextField';
 import Typography from '@material-ui/core/Typography';
 import ChevronLeftIcon from '@material-ui/icons/ChevronLeft';
 import ChevronRightIcon from '@material-ui/icons/ChevronRight';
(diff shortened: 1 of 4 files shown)


packages/core-components/src/components/LogViewer/LogViewerControls.tsx
```diff
@@ -57,7 +60,7 @@ export function LogViewerControls(props: LogViewerControlsProps) {
       <TextField
         size="small"
         variant="standard"
-        placeholder="Search"
+        placeholder={t('logViewer.controls.search')}
```
freben: I don't think we use the `controls` structure elsewhere. Could this perhaps be `logViewer.searchButton.placeholder` instead?

christoph-jerolimov:
Thanks @freben, I've changed the translation key

packages/core-components/src/components/LogViewer/LogViewerControls.tsx
```diff
@@ -57,7 +60,7 @@ export function LogViewerControls(props: LogViewerControlsProps) {
       <TextField
         size="small"
         variant="standard"
-        placeholder="Search"
+        placeholder={t('logViewer.searchButton.placeholder')}
```
freben: Sorry. I was not thinking clearly. This isn't the button, it's the field. So something like `logViewer.searchField.placeholder` instead?

christoph-jerolimov: yeah in other places we sometimes uses the component name. But `logViewer.searchField.placeholder` is also fine :+1:

Issue #30986, opened 2025-08-26
Separate frontend runtime config injection no longer works


patrickwyler:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
"nginx", "separate frontend", "frontend", "inject config"
```

### 🗃️ Project Area

CLI Tooling

### 🔗 External Integration

N/A

### 📝 Description & Context

The method of injecting configurations during runtime with environment variables into a separated Nginx frontend no longer works. I think it has to do with the transition of moving the configurations from the static.js files into the index.html file. The inject-config.sh script is just changing the static.js files but not the index.html file.

I changed the app build command of the [Dockerfile ](https://github.com/backstage/backstage/blob/master/contrib/docker/frontend-with-nginx/Dockerfile.dockerbuild#L45) to: 
`yarn workspace app build --config ../../app-config.yaml --config ../../app-config.production.yaml`

We recently upgraded Backstage from 1.41.1 to 1.42.3.

- [frontend-with-nginx](https://github.com/backstage/backstage/tree/master/contrib/docker/frontend-with-nginx)
- [inject-config.sh]( https://github.com/backstage/backstage/blob/master/contrib/docker/frontend-with-nginx/docker/inject-config.sh)
- [injectConfig.ts](https://github.com/backstage/backstage/blob/ea69e4671c0c806f5c2cf45e6b37392ee5977194/plugins/app-backend/src/lib/config/injectConfig.ts)
- https://github.com/backstage/backstage/issues/24507

### 👍 Expected Behavior

Env vars like APP_CONFIG_app_baseUrl=https://backstage.io should be injected into the Html configuration.

### 🥾 Reproduction steps

1. Build separated frontend docker image
2. Startup with some env vars with this docker-compose.yaml
```
version: '3.8'
services:
  frontend:
    image: <image>
    environment:
      - APP_CONFIG_app_baseUrl=https://backstage.io
      - APP_CONFIG_app_title=Backstage
      - APP_CONFIG_backend_baseUrl=https://backstage.io
    ports:
      - '3000:3000'
```
3. The index.html contains the app-config.yaml default values like http://localhost:3000 for app.baseUrl instead of the env var value (for example APP_CONFIG_app_baseUrl=https://backstage.io)
4. Result index.html

```
[...]

PR #31738, opened 2025-11-12
feat: config:print using saved schema

christopherkruse-EpicGames:

This adds an option to the `config:print` command in backstage-cli to supply a previously generated configuration schema (as generated by `config:schema`).

This should help teams who make use of a detached frontend to be able to pass a per-environment configuration to override the bundled config value in `index.html`. 


aramissennyeydd:
👋 Can you share a little bit more about your intended use case here?

christopherkruse-EpicGames:
> 👋 Can you share a little bit more about your intended use case here?

Sure! We currently use a dedicated frontend image for our installation of Backstage, and we're looking at ways to make the builds more generic across environments rather than build a per-environment image. Since rspack is injecting the config at build time into `index.html`, we're hoping to have a step in the container launch that reads the config from a k8s ConfigMap and replaces the config `<script>` tag in `index.html`. 

In order to keep the frontend container image lightweight, I want to be able to do that config generation through the backstage-cli tool, but I don't want to have to copy the entirety of my backend codebase over in order to provide the config schema used to print out the frontend config. Instead, this change should allow us to run the `config:print` command with a pregenerated schema, and allow us to just depend on the backstage-cli tool alone.

The plan is as follows:

- During the build, generate a schema JSON through `backstage-cli config:schema`
- Copy that schema definition to the frontend container
- on deploy, the container will start with a process that runs `backstage-cli config:print --schemaPath /path/to/schema.json --frontend --format json` to capture the frontend config values for that environment, and then inject it into `index.html` replacing the existing `<script type="backstage.io/config">` tag contents.
- _hope that_ the environment behaves correctly 😄

aramissennyeydd:
@christopherkruse-EpicGames Interesting, I've tried to do this as well in the past. It's definitely a little finicky. Are you going with the split frontend/backend deployment to not serve assets from your Backstage server? One thing we do internally that's pretty simple is use a post-build script for our app that uploads the assets to the CDN and updates the URLs in the `index.html` to point to the new CDN assets. That allows us to use the app-backend which does this logic already while still getting a scaleable asset hosting strategy. Curious if that's more what you're looking for?

aramissennyeydd:
Also linking https://github.com/backstage/backstage/issues/30986 which has some more context on the direction the project is thinking for the split backend/frontend deployment.

christopherkruse-EpicGames:
> Are you going with the split frontend/backend deployment to not serve assets from your Backstage server? 

Yes, that's one of the intents for our split. The split runtime was set up in this environment before my time, but I believe they were separated in order to avoid frontend request load impacting the resource availability of the backend (eg: if those calls weren't cached in some way). 

> One thing we do internally that's pretty simple is use a post-build script for our app that uploads the assets to the CDN and updates the URLs in the `index.html` to point to the new CDN assets. That allows us to use the app-backend which does this logic already while still getting a scaleable asset hosting strategy. Curious if that's more what you're looking for?

To paraphrase, in order to make sure I understand the actual mechanism you're describing: After the backend is deployed, you have an action that requests the static assets through the backend, and then transfers them into your CDN? Does that include `index.html` itself, or is that still served through the app-backend plugin? 

Conceptually, that could work (our frontend is a docker image containing the prebuilt frontend assets behind nginx; all we would need to read from the backend is the updated `index.html`), but it's more of an architectural change from where we're at currently. Having the ability to read and easily generate the frontend config and load it into the pregenerated assets seemed like a smaller change than what you're suggesting, but I can look on our end to see what other avenues we have.

---

Tangentially related, I am curious about other use cases for the `config:schema` generation, though - I don't see anything that makes use of stored config schemas that way.

aramissennyeydd:
> To paraphrase, in order to make sure I understand the actual mechanism you're describing: After the backend is deployed, you have an action that requests the static assets through the backend, and then transfers them into your CDN? Does that include index.html itself, or is that still served through the app-backend plugin?

Yeah basically, the index.html is served through the app-backend plugin, everything else is served through the CDN (or your split frontend deployment). 

> all we would need to read from the backend is the updated index.html), but it's more of an architectural change from where we're at currently. Having the ability to read and easily generate the frontend config and load it into the pregenerated assets seemed like a smaller change than what you're suggesting, but I can look on our end to see what other avenues we have.

I hear you, I'm hesitant to add more complexity here though, especially as we move more and more away from supporting split frontend/backend. https://github.com/mathieu-benoit/deploy-backstage-with-score/pull/210/files may also be interesting here.

> Tangentially related, I am curious about other use cases for the config:schema generation

Me too, I'm not entirely sure it has much adoption. This is the first new use case I've heard for it in a long while.

freben (commented):
This needs `yarn build:api-reports` to capture the added option. Adding @Rugvip for visibility

PR #31741, opened 2025-11-12
Add `version` to mock user credentials

KevinSnyderCodes:
Absense of `version` causes `UserInfoService.getUserInfo(credentials)` to throw error when using mock user.

Example code from documentation: https://backstage.io/docs/backend-system/core-services/user-info

```ts
router.get('/some-request', async (req, res) => {
  const credentials = await httpAuth.credentials(req, { allow: ['user'] });
  const info = await userInfo.getUserInfo(credentials); // Error: Invalid credential version undefined
});
```

### DIFF
--- a/packages/backend-test-utils/src/services/mockCredentials.ts
+++ b/packages/backend-test-utils/src/services/mockCredentials.ts
@@ -122,6 +122,7 @@ export namespace mockCredentials {
     validateUserEntityRef(userEntityRef);
     const result = {
       $$type: '@backstage/BackstageCredentials',
+      version: 'v1',
       principal: {
         type: 'user',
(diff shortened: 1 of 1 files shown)


KevinSnyderCodes:
Disregard. Correct solution here is to use the mock service `userInfo` when using other mock service such as `auth` and `httpAuth`.

Example (from `dev/index.ts` of plugin):

```ts
import { createBackend } from '@backstage/backend-defaults';
import { mockServices } from '@backstage/backend-test-utils';

const backend = createBackend();

backend.add(mockServices.auth.factory());
backend.add(mockServices.httpAuth.factory());
backend.add(mockServices.userInfo.factory());
```

PR #31745, opened 2025-11-13
introduce generic catalog scm events handling

freben:
Screenshot 2026-02-11 at 16 04 40

## Goals

Builtin SCM event reactions in the catalog, speeding up feedback loops and reducing the need for manual catalog management.

* Put a system in place for translating any SCM system's webhook events into a common format on a shared bus
* The catalog has the builtin ability to mark entities for reprocessing as an effect of such events (irrespective of how the entity entered the catalog)
* The Location provider deletes/moves registered catalog-info file URLs as an effect of such events (including corner cases like changing the default branch name)

## Non-goals

* This does not implement discovery as an effect of eg file / repo creation, despite the event info being available for it. It's a very compelling follow-up though.
* This does not yet change other existing providers to leverage the new events

### `@backstage/plugin-catalog-node/alpha`:

Adds `catalogScmEventsServiceRef`, as an SCM event publish/subscribe bus. The event type is a generic, simplified set of descriptions along the lines of of `location.moved` and `repository.deleted`, to try to fit all possible SCM systems.

The idea is that various events backend modules publish events to this, and then various entity providers and similar can react to those events.

It was _not_ built to carry these "translated" events across the events bus again, but rather opted to do this in-memory on the local machine. This simplifies the architecture a bit and lets us reason more easily about overlapping events and blocking promises.

### `@backstage/plugin-catalog-backend-module-github`:

Adds `GithubScmEventsBridge` that translates the regular `github` webhook event topic from the events backend into publishing generic `catalogScmEventsServiceRef` events. It also does so with a much richer and smarter implementation than before; it tries to fetch and analyze complex pushes to track detailed file changes even in large and involved merges.

### `@backstage/plugin-catalog-backend`:

The builtin `DefaultLocationStore` now reacts to SCM events, deleting and moving registrations as needed when events arrive.

There's also a new `GenericScmEventRefreshProvider` that immediately schedules ANY entity affected by SCM events for immediate re-processing no matter what SCM system the events came from.

### DIFF
--- a/plugins/catalog-node/src/scmEvents/catalogScmEventsServiceRef.ts
+++ b/plugins/catalog-node/src/scmEvents/catalogScmEventsServiceRef.ts
@@ -29,7 +29,7 @@ import { DefaultCatalogScmEventsService } from './DefaultCatalogScmEventsService
  * @remarks
  *
  * The default implementation of this service acts in-memory, which requires the
- * produceers and consumer (the catalog backend) to be deployed together.
+ * producers and consumer (the catalog backend) to be deployed together.
  */
 export const catalogScmEventsServiceRef =
   createServiceRef<CatalogScmEventsService>({
--- a/plugins/catalog-backend-module-github/src/events/GithubScmEventsBridge.ts
+++ b/plugins/catalog-backend-module-github/src/events/GithubScmEventsBridge.ts
@@ -71,7 +71,7 @@ export class GithubScmEventsBridge {
 
     if (this.#shuttingDown) {
       this.#logger.warn(
-        `Skipping GitHub webhook event of type "${params.topic}" because the bridge is shutting down`,
+        `Skipping GitHub webhook event of type "${eventType}" on topic "${params.topic}" because the bridge is shutting down`,
       );
       return;
     }
@@ -93,15 +93,15 @@ export class GithubScmEventsBridge {
           await this.#catalogScmEvents.publish(output.events);
         } else if (output.result === 'ignored') {
           this.#logger.debug(
-            `Skipping GitHub webhook event of type "${params.topic}" because it is ignored: ${output.reason}`,
+            `Skipping GitHub webhook event of type "${eventType}" on topic "${params.topic}" because it is ignored: ${output.reason}`,
           );
         } else if (output.result === 'aborted') {
           this.#logger.warn(
-            `Skipping GitHub webhook event of type "${params.topic}" because it is aborted: ${output.reason}`,
+            `Skipping GitHub webhook event of type "${eventType}" on topic "${params.topic}" because it is aborted: ${output.reason}`,
           );
         } else if (output.result === 'unsupported-event') {
           this.#logger.debug(
-            `Skipping GitHub webhook event of type "${params.topic}" because it is unsupported: ${output.event}`,
+            `Skipping GitHub webhook event of type "${eventType}" on topic "${params.topic}" because it is unsupported: ${output.event}`,
           );
         }
       } catch (error) {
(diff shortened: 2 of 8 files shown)


plugins/catalog-node/src/scmEvents/catalogScmEventsServiceRef.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @alpha
+ * @remarks
+ *
+ * The default implementation of this service acts in-memory, which requires the
+ * produceers and consumer (the catalog backend) to be deployed together.
+ */
+export const catalogScmEventsServiceRef =
+  createServiceRef<CatalogScmEventsService>({
+    id: 'catalog-scm-events',
```
Rugvip: I think `catalog.scm-events.alpha`

plugins/catalog-node/src/scmEvents/catalogScmEventsServiceRef.ts
```diff
@@ -0,0 +1,46 @@
...
+    scope: 'root',
```
Rugvip: I'm thinking it might be better to have this be plugin scoped and then use the root context instead? Could always be handy to be able to access plugin services and meta

plugins/catalog-backend/src/providers/GenericScmEventRefreshProvider.ts
```diff
@@ -0,0 +1,162 @@
...
+    this.#scmEvents = scmEvents;
+  }
+
+  getProviderName(): string {
+    return 'GenericScmEventRefreshProvider';
+  }
+
+  async connect(_connection: EntityProviderConnection): Promise<void> {
+    this.#scmEvents.subscribe({ onEvents: this.#onScmEvents.bind(this) });
+  }
```
Rugvip, line 68: Hmm, why is this a provider? 🤔

.changeset/early-cases-laugh.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': minor
+---
+
+Implemented `catalogScmEventsServiceRef` event handling in the builtin entity providers.
```
Rugvip: Mention that this is in alpha in these?

.changeset/thirty-dancers-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-github': patch
+---
+
+Implemented translation of webhook events into `catalogScmEventsServiceRef` events.
```
Rugvip, line 5: Thinking it might be good to feature flag the usage of this in this module?

Rugvip (changes requested):
Nice! 👍 

Couple of things but pretty happy to ship this as an experiment in alpha

.changeset/early-cases-laugh.md
freben: Updated the text

plugins/catalog-backend/src/providers/GenericScmEventRefreshProvider.ts
freben: See the comment above:

```
 * It's implemented in the form of an entity provider even though its actually
 * does not behave like one, mostly in order to consistently start treating
 * events on connect time when the engine is known to be ready.
```

plugins/catalog-node/src/scmEvents/catalogScmEventsServiceRef.ts
freben: Yeah that's a great callout - fixed

plugins/catalog-node/src/scmEvents/catalogScmEventsServiceRef.ts
freben: Fixed!

.changeset/thirty-dancers-clean.md
freben: Yeah maybe it's good to have a "kill switch" at the very least, on a per-provider basis if you are not interested in that particular provider to leverage the feature. Left that as a future enhancement (there's only one source for now), and instead added config options for the receiving end (in the catalog)

plugins/catalog-backend/src/providers/GenericScmEventRefreshProvider.ts
Rugvip: ah yep makes sense

PR #31763, opened 2025-11-14
fix(catalog): fix refresh service ancestry to follow the tree

drodil:

if location is referenced by another location, that will not get updated properly

### DIFF
--- a/plugins/catalog-backend/src/service/DefaultRefreshService.ts
+++ b/plugins/catalog-backend/src/service/DefaultRefreshService.ts
@@ -29,20 +29,17 @@ export class DefaultRefreshService implements RefreshService {
       const { entityRefs } = await this.database.listAncestors(tx, {
         entityRef: options.entityRef,
       });
-      const locationAncestor = entityRefs.find(ref =>
+
+      const locationAncestors = entityRefs.filter(ref =>
         ref.startsWith('location:'),
       );
 
       // TODO: Refreshes are currently scheduled(as soon as possible) for execution and will therefore happen in the future.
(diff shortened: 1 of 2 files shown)


plugins/catalog-backend/src/service/DefaultRefreshService.ts
```diff
@@ -26,23 +26,24 @@ export class DefaultRefreshService implements RefreshService {
...
-        ref.startsWith('location:'),
-      );
+      const refreshEntityRef = async (entityRef: string) => {
+        const { entityRefs } = await this.database.listAncestors(tx, {
+          entityRef,
+        });
+
+        const locationAncestor = entityRefs.find(ref =>
+          ref.startsWith('location:'),
+        );
```
aramissennyeydd:
```suggestion
        const locationAncestor = entityRefs.filter(ref =>
          ref.startsWith('location:'),
        );
```
I think we can just do something like this since we're fetching the full ancestry list?

drodil: Hmm, does it actually return the tree of ancestors? I thought it just returned one level up 🤔

aramissennyeydd: It should be the full tree, https://github.com/drodil/backstage/blob/master/plugins/catalog-backend/src/database/DefaultCatalogDatabase.ts#L72-L96

drodil: Oh nice, so it could work with filter and then just loop.

PR #31776, opened 2025-11-16
Implement filtering in Google PubSub events module

freben:
To cut down on unnecessary event bus traffic, this change allows you to specify things like

```ts
filter:
  $any:
    - 'message.attributes.x-github-event': 'push'
    - 'message.attributes.x-github-event': 'repository'
      'message.data.action': { $in: ['created', 'deleted'] }
```

on your pubsub consumer / publisher definitions.

This follows the exact same filter predicate syntax as the one [used by the catalog](https://backstage.io/docs/features/software-catalog/catalog-customization#entity-predicate-queries).

This PR looks bigger than it kind of is. Almost all of the added files are pure copies from `catalog-react`, since there's no central facility for this yet. And when the `valueAtPath` helper was taken from there, it made sense to use that in the existing pattern resolver code leading to small tweaks there.

The remaining perhaps seemingly unexpected changes are there because the data passed to these filters/mappers was ill defined and partially broken. So I defined proper context types to clarify what actually gets passed in.

### DIFF
--- a/plugins/events-backend-module-google-pubsub/config.d.ts
+++ b/plugins/events-backend-module-google-pubsub/config.d.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { FilterPredicate } from '@backstage/filter-predicates';
+
 export interface Config {
   events?: {
     modules?: {
@@ -96,7 +98,7 @@ export interface Config {
                *       'message.data.action': { $in: ['created', 'deleted'] }
                * ```
                */
-              filter?: object;
+              filter?: FilterPredicate;
 
               /**
                * Pub/Sub message attributes are by default copied to the event
@@ -190,7 +192,7 @@ export interface Config {
                *       'event.eventPayload.action': { $in: ['created', 'deleted'] }
                * ```
                */
-              filter?: object;
+              filter?: FilterPredicate;
 
               /**
                * Event metadata fields are by default copied to the Pub/Sub
--- a/plugins/events-backend-module-google-pubsub/src/GooglePubSubConsumingEventPublisher/GooglePubSubConsumingEventPublisher.ts
+++ b/plugins/events-backend-module-google-pubsub/src/GooglePubSubConsumingEventPublisher/GooglePubSubConsumingEventPublisher.ts
@@ -19,7 +19,9 @@ import {
   RootConfigService,
   RootLifecycleService,
 } from '@backstage/backend-plugin-api';
+import { ForwardedError } from '@backstage/errors';
 import { EventParams, EventsService } from '@backstage/plugin-events-node';
+import { JsonValue } from '@backstage/types';
 import { Message, PubSub, Subscription } from '@google-cloud/pubsub';
 import { Counter, metrics } from '@opentelemetry/api';
 import { readSubscriptionTasksFromConfig } from './config';
@@ -197,7 +199,13 @@ export class GooglePubSubConsumingEventPublisher {
     message: Message,
     task: SubscriptionTask,
   ): EventParams | undefined {
-    const eventPayload = JSON.parse(message.data.toString());
+    let eventPayload: JsonValue;
+    try {
+      eventPayload = JSON.parse(message.data.toString());
+    } catch (error) {
+      throw new ForwardedError('Payload was not valid JSON', error);
+    }
+
     const attributes = message.attributes;
 
     const context: MessageContext = {


plugins/events-backend-module-google-pubsub/src/util/types.ts
freben: just copied from catalog-react (but the names changed to not have Entity in them)

plugins/events-backend-module-google-pubsub/src/util/filterPredicateToFilterFunction.ts
freben: just copied from catalog-react

plugins/events-backend-module-google-pubsub/src/util/createFilterPredicateSchema.test.ts
freben: just copied from catalog-react

plugins/events-backend-module-google-pubsub/src/util/createPatternResolver.test.ts
```diff
@@ -40,11 +40,6 @@ describe('createPatternResolver', () => {
         a: { b: ['first', 'second'] },
       }),
     ).toEqual('-second-');
-    expect(
-      createPatternResolver('-{{ a.b.0 }}-')({
```
freben, line 44: When reusing the new `getJsonValueAtPath`, this syntax is no longer supported. But I feel that that's pretty much a good thing - it wasn't really intuitive or needed

plugins/events-backend-module-google-pubsub/src/util/createPatternResolver.ts
freben, line 1: This looks a bit messy. But the point is, when the `getJsonValueAtPath` function landed here from the new filter predicates package, it was very compelling to use it. The old implementation was less smart than it, because it couldn't properly handle complex `a.b.c` chains such as for annotations. Changing the regex to hold those chains together instead of looking at the pieces one by one, and letting `valueAtPath` deal with the outcome, led to a better end result.

plugins/events-backend-module-google-pubsub/src/GooglePubSubConsumingEventPublisher/GooglePubSubConsumingEventPublisher.ts
```diff
@@ -197,14 +197,30 @@ export class GooglePubSubConsumingEventPublisher {
     message: Message,
     task: SubscriptionTask,
   ): EventParams | undefined {
-    const topic = task.mapToTopic(message);
+    const eventPayload = JSON.parse(message.data.toString());
```
Rugvip: Might be worth wrapping up parsing failures in more context? thinking `ForwardedError`

plugins/events-backend-module-google-pubsub/config.d.ts
```diff
@@ -67,6 +67,37 @@ export interface Config {
...
+               *
+               * ```yaml
+               * filter:
+               *   $any:
+               *     - 'message.attributes.x-github-event': 'push'
+               *     - 'message.attributes.x-github-event': 'repository'
+               *       'message.data.action': { $in: ['created', 'deleted'] }
+               * ```
+               */
+              filter?: object;
```
Rugvip: Hmm, this isn't great, think we might as well use the filter predicate type here?

plugins/events-backend-module-google-pubsub/src/EventConsumingGooglePubSubPublisher/config.ts
```diff
@@ -60,12 +65,26 @@ function readSourceTopics(config: Config): string[] {
   return [config.getString('sourceTopic')];
 }
 
+/**
+ * Handles the `filter` configuration field.
+ */
+function readFilter(config: Config): (context: EventContext) => boolean {
+  const predicate = readOptionalFilterPredicateFromConfig(config, {
+    key: 'filter',
+  });
```
Rugvip, line 74: Hmm, an error parsing this is just gonna say " at filter", which isn't very helpful. Don't have a great suggestion for how to fix that though, other than switching to fully using the config reading methods in `filter-predicates`. Even so there's invalid input like unknown matchers that we don't have any clean errors for. Think we'll have to leave as is, but it's something more general to improve I guess 🤷

freben: yeah let's make this easier in a future config system rewrite

plugins/events-backend-module-google-pubsub/config.d.ts
freben: Added! I was thinking it doesn't get handled properly anyway at this point, but you are quite right, it's valuable anyway

plugins/events-backend-module-google-pubsub/src/GooglePubSubConsumingEventPublisher/GooglePubSubConsumingEventPublisher.ts
freben: Ye, good call, added

PR #31793, opened 2025-11-17
`MetricsService`: alpha release

kurtaking:

Introduces MetricsService as a new `@alpha` core service. See related [BEP](https://github.com/backstage/backstage/blob/master/beps/0012-metrics-service/README.md).

As it stands, the service is a very thin wrapper around the `@opentelemetry.api`. This PR includes the bare minimum required to migrate the existing catalog metrics 1:1 to the new service without introducing breaking changes. It does not include features like auto namespacing and validation. Adding this with a fallback in an isolated PR would be much cleaner. 

> Expect the interface to include breaking changes while in alpha release.

### Status

This feature is marked as `@alpha` and is subject to change. It requires OpenTelemetry to be set up in the Backstage instance (see the Setup OpenTelemetry tutorial). 

### Features

* Supports all OpenTelemetry metric types: Counter, UpDownCounter, Histogram, Gauge, and their observable variants
* Full type safe support with proper typing for all metric instruments

### Related

* Follows BEP-0012: Metrics Service


packages/backend-plugin-api/src/alpha/refs.ts
```diff
@@ -54,3 +54,25 @@ export const actionsRegistryServiceRef = createServiceRef<
...
+
+/**
+ * Service for managing metrics for root scoped services.
+ *
+ * @alpha
+ */
+export const rootMetricsServiceRef = createServiceRef<
+  import('./MetricsService').RootMetricsService
+>({
+  id: 'alpha.core.rootMetrics',
```
freben:
```suggestion
  id: 'alpha.core.rootMetrics',
  scope: 'root',
```

:)

packages/backend-defaults/src/alpha/entrypoints/metrics/DefaultMetricsService.ts
```diff
@@ -0,0 +1,72 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  MetricsService,
+  ObservableMetric,
```
freben: I presume you wanted our own `ObservableMetric` as a local type (rather than mirroring `@opentelemetry/api`) because it's rare to want to have several callbacks?

packages/backend-plugin-api/package.json
```diff
@@ -60,6 +60,7 @@
     "@backstage/plugin-permission-common": "workspace:^",
     "@backstage/plugin-permission-node": "workspace:^",
     "@backstage/types": "workspace:^",
+    "@opentelemetry/api": "^1.9.0",
```
freben: Do you have any particular opinion on whether something like this should be a peer dep or not? Ping @Rugvip too

packages/backend-plugin-api/src/alpha/MetricsService.ts
```diff
@@ -0,0 +1,133 @@
...
+ */
+export interface MetricsService {
+  /**
+   * Creates a new counter metric.
+   *
+   * @param name - The name of the metric.
+   * @param opts - The options for the metric.
+   * @returns The counter metric.
+   */
+  createCounter(name: string, opts?: MetricOptions): Counter;
```
freben: The original interface goes through the trouble of doing `<TAttributes extends Attributes = Attributes>` all over the place (including the types that correspond to the `ObservableMetric` above). Any particular opinions on that? I'm not saying it's a bad opinionated choice tio skip it, just curious about how intentional it was and the reasoning

packages/backend-plugin-api/src/alpha/MetricsService.ts
```diff
@@ -0,0 +1,133 @@
...
+ * @alpha
+ */
+export interface RootMetricsService extends MetricsService {
+  /**
+   * Creates a new metrics service for a plugin.
+   *
+   * @param pluginId - The ID of the plugin.
+   * @returns The metrics service for the plugin.
+   */
+  forPlugin(pluginId: string): MetricsService;
```
freben: I don't think this should be part of the public interface. It's fine to have a `forPlugin` method on default implementation classes, absolutely. That's what we do for some other interfaces - but it shouldn't be part of the forced contract of the root thing, to provide plugin things.

For someone who overrides service factories in their own installation, they have the opportunity to do this separation "outside" of the contracts so to speak; ie to do `new MyCustomRootMetricsImpl()` in the `loader` method of a feature loader and then using that one instance to be passed into both the root and plugin level metrics service factories it makes.

If instead `MetricsService` itself had a `.withAttributes(...): MetricsService` and/or `.withMetricNamePrefix(...): MetricsService` method [similar to what the logger does](https://github.com/backstage/backstage/blob/2b3fb40d068d1a23c38805113d20f1da0ffd49d5/packages/backend-defaults/src/entrypoints/logger/loggerServiceFactory.ts#L38) then maybe that could have been a useful abstraction. Either way - this was just me thinking out loud, and it would have been on the plugin interface anyhow, and I think we can skip it.

In summary: this could probably be `export type RootMetricsService = MetricsService`.

.changeset/rare-adults-attack.md
```diff
@@ -0,0 +1,8 @@
+---
+'@backstage/backend-plugin-api': patch
+'@backstage/backend-defaults': patch
+---
+
+Adds an alpha MetricsService to provide a unified interface for metrics instrumentation. It is designed to be used by plugins to create metrics instruments with well-defined namespaces and attributes, promoting consistency and reusability across the Backstage ecosystem.
```
freben:
```suggestion
Adds an alpha `MetricsService` to provide a unified interface for metrics instrumentation. It is designed to be used by plugins to create metrics instruments with well-defined namespaces and attributes, promoting consistency and reusability across the Backstage ecosystem.
```

docs/backend-system/core-services/metrics.md
```diff
@@ -0,0 +1,146 @@
...
+
+All metrics in the Backstage ecosystem are based on [OpenTelemetry Semantic Conventions](https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/). Following these conventions ensures consistency, discoverability, and compatibility with monitoring tools.
+
+:::tip
+We strive to stay aligned with the OpenTelemetry best practices. When in doubt, refer to the OpenTelemetry documentation. If you notice any inconsistencies, please open an issue or submit a PR.
+:::
+
+### Building Blocks
+
+All Backstage metrics follow a hierarchical naming pattern that provides clear context about where the metric originates and what it measures. We recommend following this pattern for all metrics. Notice that multi-word scope and metric names follow snake_case.
```
freben:
```suggestion
All Backstage metrics follow a hierarchical naming pattern that provides clear context about where the metric originates and what it measures. We recommend following this pattern for all metrics. Notice that multi-word scope and metric names follow `snake_case`.
```

docs/backend-system/core-services/metrics.md
```diff
@@ -0,0 +1,146 @@
...
+```md
+backstage.core.database.connections.active
+backstage.core.database.operations.duration
+backstage.core.scheduler.tasks.total
+backstage.core.httpRouter.middleware.duration
+```
+
+### Metric Name Structure
+
+The `{metric_name}` component should be hierarchical using dot notation to create logical groupings. For multi-word metric names, use snake_case.
```
freben:
```suggestion
The `{metric_name}` component should be hierarchical using dot notation to create logical groupings. For multi-word metric names, use `snake_case`.
```

docs/backend-system/core-services/metrics.md
```diff
@@ -0,0 +1,146 @@
...
+#### Plugin Scope
+
+**Pattern:** `backstage.plugin.{pluginId}.{metric_name}`
+
+Plugin-scoped metrics represent functionality specific to individual plugins. Each plugin operates as an independent microservice with its own metrics namespace. All plugin metrics start with `backstage.plugin.` followed by the plugin ID and the metric name.
+
+**Examples:**
+
+```md
+backstage.plugin.catalog.entities.operations.total
```
freben: Examples here and below do NOT use snake_case for the trailing part. Should they?

docs/backend-system/core-services/metrics.md
```diff
@@ -0,0 +1,146 @@
...
+backstage.plugin.scaffolder.tasks.total
+backstage.plugin.techdocs.builds.duration
+```
+
+We currently do not support module-scoped metrics since most modules should receive metrics from the plugin (extension point) they extend.
```
freben: I think this is mostly an effect of there not even being a module-scoping concept in the backend system overall. Feel free to skip expressly calling this out if you like

packages/backend-defaults/src/alpha/entrypoints/metrics/DefaultRootMetricsService.ts
```diff
@@ -0,0 +1,104 @@
...
+  static forRoot(opts?: {
+    rootLogger?: RootLoggerService;
+  }): RootMetricsService {
+    return new DefaultRootMetricsService({
+      rootLogger: opts?.rootLogger,
+    });
+  }
+
+  forPlugin(pluginId: string): MetricsService {
+    this.rootLogger?.info('Creating plugin-scoped metrics service', {
```
freben: Probably not all that useful to log

packages/backend-defaults/src/alpha/entrypoints/metrics/DefaultRootMetricsService.ts
```diff
@@ -0,0 +1,104 @@
...
+      pluginId,
+    });
+
+    const namespace = `${this.rootNamespace}.plugin.${pluginId}`;
+    const meter = this.globalMeterProvider.getMeter(namespace);
```
freben: Is it safe to do this here, or should we defer the getMeter all the way into the service when like the first metric is made? Maybe I worry too much :D I just want to be entirely sure that at that point, any replacement of the global meter (by the adopter) really truly is complete

packages/backend-defaults/src/alpha/entrypoints/metrics/InstrumentFactory.ts
```diff
@@ -0,0 +1,162 @@
...
+    this.meter = opts.meter;
+    this.namespace = namespaceSchema.parse(opts.namespace);
+  }
+
+  private validateMetricName(name: string): string {
+    return metricNameSchema.parse(name);
+  }
+
+  private prefixName(name: string): string {
+    return `${this.namespace}.${name}`.toLowerCase();
```
freben: worth making this a `.toLocaleLowerCase('en-US')` to not run into oddness in e.g. Turkish locales

packages/backend-plugin-api/package.json
kurtaking: Yes, I believe you are correct in us needing to make this a peer dependency. They have [this guide](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md#instrumentations-implementation-guide) that I am following. Specifically [this section](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md#opentelemetry-api).

docs/backend-system/core-services/metrics.md
kurtaking: The docs look correct, so perhaps I need to better clarify. For example `entities.operations.total` and `my_thing.operations.total` are the valid options. We don't want `entities_operations_total`.

docs/backend-system/core-services/metrics.md
kurtaking: removed.

packages/backend-plugin-api/src/alpha/MetricsService.ts
kurtaking: Do we need `RootMetricsService` at that point?

packages/backend-plugin-api/src/alpha/MetricsService.ts
kurtaking: We should support this. Giving end users a way to specify a set of attributes they want on every metric the instrument emits is useful. 

**edit:** this is to help ensure type safety throughout all calls to the this instrument.

packages/backend-defaults/src/alpha/entrypoints/metrics/DefaultRootMetricsService.ts
kurtaking: The service leverages whatever meter provider the adopter sets in their `instrumentation.js` file. What you are referring to shouldn't be until we handle initialization for the end user.

packages/backend-plugin-api/src/alpha/MetricsService.ts
freben: If root scoped services (and the backend itself) should be able to emit metrics, yeah

packages/backend-defaults/src/alpha/entrypoints/metrics/DefaultMetricsService.ts
kurtaking: Yes and the idea of providing a simpler interface, but I am starting to go back on this decision. I think we use the type coming from otel directly and iterate on it as needed. Doing that will get this out in `alpha` much faster.

packages/backend-defaults/src/alpha/entrypoints/metrics/DefaultRootMetricsService.ts
kurtaking: By the way, I've been keeping up with the otel project as best as I can and the concepts of dynamic config is coming 🎉 so we will eventually have full control. 

Following - 

* https://github.com/open-telemetry/opentelemetry-js/issues/5626#event-20067418707
* https://github.com/orgs/open-telemetry/projects/157/views/1
* https://github.com/open-telemetry/opentelemetry-js/pull/4858

packages/backend-plugin-api/src/alpha/MetricsService.ts
kurtaking: What about the interface specificially, i.e., only relying on `MetricsService`?

`DefaultRootMetricsService implements MetricsService`

packages/backend-plugin-api/src/alpha/MetricsService.ts
kurtaking: @freben - alternatively, our `MetricOptions` could include an optional `attributes` prop. wdyt?

packages/backend-defaults/src/alpha/entrypoints/metrics/DefaultMetricsService.ts
kurtaking: Reverted back to matching otel 1:1. Let's revisit as we have examples from leveraging the service.

packages/backend-plugin-api/src/alpha/MetricsService.ts
kurtaking: Let's get the first alpha out matching OTEL 1:1 and revisit. Wdyt?

packages/backend-plugin-api/src/alpha/MetricsService.ts
kurtaking: I've cleaned this up to no longer need `forPlugin()` and rather have the default impl handle directly.

packages/backend-defaults/src/alpha/entrypoints/metrics/metricsServiceFactory.ts
```diff
@@ -14,14 +14,28 @@
...
+ *
+ * @alpha
+ */
+export const metricsServiceFactory = createServiceFactory({
+  service: metricsServiceRef,
+  deps: {
+    pluginMetadata: coreServices.pluginMetadata,
+  },
+  factory: () => {
+    const namespace = 'default';
```
kurtaking: Rather than attempting to make everything configurable in a single PR, I elected to provide an implementation that lets us migrate existing metrics to use the new service without introducing any breaking changes. The follow-up would be to implement the `backstage.{plugin_id}` namespacing will migrating existing metrics to their new counterparts. This would introduce a breaking change so we could decide to include a config flag to disable this during alpha phase.

freben:
Hey @kurtaking - the build seems to be furious. I tried to rebase to try it out, but that turned out to be a bit tough. Do you think you could squash to a single commit and then rebase this, to see if it makes the build happy?

freben:
Sorry scratch that, I made it work - let's see where it goes

kurtaking:
> Sorry scratch that, I made it work - let's see where it goes

Thanks for the help. Let me know where you'd like for me to pick up.

kurtaking:
> > Sorry scratch that, I made it work - let's see where it goes
> 
> Thanks for the help. Let me know where you'd like for me to pick up.

@freben - would you rather have a [smaller PR](https://github.com/backstage/backstage/compare/master...kurtaking:backstage:alpha-metrics-service-impl-only) that only include the service to start or continue here with catalog refactor included?

packages/backend-defaults/src/alpha/entrypoints/metrics/metricsServiceFactory.ts
aramissennyeydd: Would it be possible to double "publish" metrics? One for the plugin namespaced version and one for default? That way the breaking change isn't a big-bang style change and orgs that are using this service have some level of historical metrics before migrating over.

(thinking outloud, this need not be in this initial PR)

kurtaking: Yes, we will need to do something along those lines. And can be something in a follow-up 👍🏼

packages/backend-plugin-api/package.json
```diff
@@ -68,6 +68,9 @@
     "luxon": "^3.0.0",
     "zod": "^3.22.4"
   },
+  "peerDependencies": {
```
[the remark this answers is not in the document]
kurtaking: Anyone have a recommendation here?

kurtaking:
I propose adding the alpha MetricsService as part of adding two new MCP-related metrics in https://github.com/backstage/backstage/pull/32358. It's a more complete example of how it will be leveraged. 

The followup here would then be to migrate the catalog metrics with support for opting out. 

```yaml
# this would change `getMeter(catalogPluginId) -> getMeter('default') to retain current state
catalog:
  metrics:
    name: default
```

packages/backend-plugin-api/src/alpha/refs.ts
```diff
@@ -56,3 +56,14 @@ export const rootSystemMetadataServiceRef = createServiceRef<
...
+
+/**
+ * Service for managing metrics.
+ *
+ * @alpha
+ */
+export const metricsServiceRef = createServiceRef<
+  import('./MetricsService').MetricsService
+>({
+  id: 'alpha.core.metrics',
```
Rugvip, line 68:
```suggestion
  id: 'core.metrics.alpha',
```

Other way around for this one

benjdlambert:
@kurtaking hope you don't mind here but I created a separate branch with the feedback from the framework SIG and a rebase and squash of this PR. Happy to merge them back to this PR if you'd like but I didn't wanna just push over what you have here as that felt a bit rude 😅 

That other PR is basically on top of all the great work in this PR :tada:

benjdlambert:
This work has been merged in #32977! Thanks @kurtaking + others! :tada:

PR #31799, opened 2025-11-17
fix(auth): user_created_at migration non constant defaults SQLiteError

### LINKED ISSUE #31562: Is migration to Postgres mandatory from sqlite?
### 📝 Acknowledgement

### Content

Hi all, 

We have a backstage setup running at 1.35.0 with yarn v1. I'm working on upgrading it to latest. The setup is using sqllite in local and Postgrest on higher envs such as stage and production. 

 But when upgrading it from 1.38.0 to 1.39.0 or going directly to latest version and running backend gives me this error: 

```
Plugin 'auth' threw an error during startup, waiting for 1 other plugins to finish before shutting down the process. Plugin 'auth' startup failed; caused by SqliteError: alter table `user_info` add column `created_at` datetime not null default CURRENT_TIMESTAMP - Cannot add a column with non-constant default type="initialization" cause={"code":"SQLITE_ERROR"} name="SqliteError" stack="SqliteError: Plugin 'auth' startup failed; caused by SqliteError: alter table `user_info` add column `created_at` datetime not null default CURRENT_TIMESTAMP - Cannot add a column with non-constant default
    at <anonymous> ([...]/node_modules/@backstage/backend-app-api/src/wiring/BackendInitializer.ts:437:21)
    at async <anonymous> ([...]/node_modules/@backstage/backend-app-api/src/wiring/BackendInitializer.ts:436:13)
    at async Promise.allSettled (index 3)
    at async BackendInitializer.#doStart ([...]/node_modules/@backstage/backend-app-api/src/wiring/BackendInitializer.ts:355:21)
    at async BackendInitializer.start ([...]/node_modules/@backstage/backend-app-api/src/wiring/BackendInitializer.ts:245:5)
    at async BackstageBackend.start ([...]/node_modules/@backstage/backend-app-api/src/wiring/BackstageBackend.ts:38:5)"
```

One solution I found was deleting the database from previous versions in my local and then running the backend, it works. But this is not feasible in other environments such as production. 

Another solution might be to move from sqllite to Postgres in local. 

So had a couple of question: 
- Is it mandatory to move to Postgres?
- How can I proceed with this error?

jtbry:

This follows guidance from & resolves #31562. The `user_created_at` migration now creates the `created_at` column as nullable, then assigns the `now()` value and finally alters back to non-nullable with a default value.

### DIFF
--- a/.changeset/mighty-flowers-travel.md
+++ b/.changeset/mighty-flowers-travel.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-auth-backend': minor
 ---
 
-Fix user_created_at migration causing SQLiteError regarding use of non-constants for defaults
+Fix `user_created_at` migration causing `SQLiteError` regarding use of non-constants for defaults


.changeset/mighty-flowers-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-auth-backend': minor
+---
+
+Fix user_created_at migration causing SQLiteError regarding use of non-constants for defaults
```
freben:
```suggestion
Fix `user_created_at` migration causing `SQLiteError` regarding use of non-constants for defaults
```

freben (approved):

PR #31804, opened 2025-11-17
core-plugin-api: add forwards compatibility for route refs

Rugvip:

This makes it possible to use route references from the old frontend system directly in the new one. In means there's no longer a need to use `convertLegacyRouteRef` or `convertLegacyRouteRefs` to re-use route refs in implementations for the new system. As long as a plugin intends to support the old system, it can keep defining its route refs using the old system.

### DIFF
--- a/plugins/catalog/src/alpha/pages.tsx
+++ b/plugins/catalog/src/alpha/pages.tsx
@@ -108,7 +108,13 @@ export const catalogEntityPage = PageBlueprint.makeWithOverrides({
   factory(originalFactory, { config, inputs }) {
     return originalFactory({
       path: '/catalog/:namespace/:kind/:name',
-      routeRef: convertLegacyRouteRef(entityRouteRef),
+      // NOTE: The `convertLegacyRouteRef` call here ensures that this route ref
+      // is mutated to support the new frontend system. Removing this conversion
+      // is a potentially breaking change since this is a singleton and the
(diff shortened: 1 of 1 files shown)


benjdlambert (approved):
Lovely stuff! :tada:

packages/frontend-app-api/src/routing/collectRouteIds.ts
```diff
@@ -62,12 +58,12 @@ export function collectRouteIds(
         continue;
       }
 
-      if (isRouteRef(ref)) {
-        const internalRef = toInternalRouteRef(ref);
+      if (OpaqueRouteRef.isType(ref)) {
```
freben, line 61: Is there no risk that we get an input that's an ancient ref created by an old core-plugin-api? Can the `isType` checks fail?

Rugvip: Should be fine, since you can't include the old route refs in an NFS plugin unless they're of this new format. Now that you say it I do wanna double check if this could somehow affect the `entityRouteRef` though, given that it's a singleton.

Rugvip: Yep it's fine, as long as we keep converting `entityRouteRef`. Added a note about this behavior to make sure we don't break it accidentally: https://github.com/backstage/backstage/pull/31804/commits/debf9ea34f0874d5decb9b828189f842032d5fff

PR #31805, opened 2025-11-17
Notification slack processor: make rate limit configurable

kaidubauskas-dd:

This makes the rate limit for the slack notification processor configurable. Default if not set remains at 10

### DIFF
--- a/docs/notifications/processors.md
+++ b/docs/notifications/processors.md
@@ -150,7 +150,7 @@ notifications:
           - C12345678
         username: 'Backstage Bot' # Optional, defaults to the name of the Slack App.
         concurrencyLimit: 20 # Optional, number of messages allowed per interval. Defaults to 10.
-        throttleInterval: 1m # Optional, ISO 8601 duration (or ms value). Defaults to 1 minute.
+        throttleInterval: 1m # Optional, Accepts ISO-8601 duration, ms-style ("1m", "30s"), or HumanDuration ({ minutes: 2  }). Defaults to 1 minute
 ```
 
 Multiple instances can be added in the `slack` array, allowing you to have multiple configurations if you need to send
--- a/plugins/notifications-backend-module-slack/config.d.ts
+++ b/plugins/notifications-backend-module-slack/config.d.ts
@@ -31,11 +31,11 @@ export interface Config {
          */
         broadcastChannels?: string[];
         /**
-         * Concurrency limit for Slack notifications, defaults to 10
+         * Concurrency limit for Slack notifications per backend instance of the notifications plugin, defaults to 10.
          */
         concurrencyLimit?: number;
         /**
-         * Throttle duration between Slack notifications, defaults to 1 minute
+         * Throttle duration between Slack notifications per backend instance of the notifications plugin, defaults to 1 minute.
          */
         throttleInterval?: HumanDuration | string;
       }>;


plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -68,10 +68,12 @@ export class SlackNotificationProcessor implements NotificationProcessor {
       const slack = options.slack ?? new WebClient(token);
       const broadcastChannels = c.getOptionalStringArray('broadcastChannels');
       const username = c.getOptionalString('username');
+      const rateLimit = c.getOptionalNumber('rateLimit');
```
aramissennyeydd: Can we use the same config values as we have defined in the email processor here 🙏? `throttleInterval` and `concurrencyLimit` https://github.com/backstage/backstage/blob/master/plugins/notifications-backend-module-email/src/processor/NotificationsEmailProcessor.ts#L94-L102. `rateLimit` is kind of a mix of both, would make more sense to let limit and interval be configurable instead.

benjdlambert:
@backstage/notifications-maintainers do you wanna take a peek here? :pray:

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
kaidubauskas-dd: Updated

yarn.lock
aramissennyeydd, line 1: These might have snuck in from other changes? You might need to rebase again

.changeset/pretty-breads-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-notifications-backend-module-slack': patch
+---
+
+The throttle limit and interval is now a config variable
```
aramissennyeydd:
```suggestion
Slack notification handler throttling can now be configured with the `concurrencyLimit` and `throttleInterval` options.
```

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -68,10 +70,18 @@ export class SlackNotificationProcessor implements NotificationProcessor {
       const slack = options.slack ?? new WebClient(token);
       const broadcastChannels = c.getOptionalStringArray('broadcastChannels');
       const username = c.getOptionalString('username');
+      const concurrencyLimit = c.getOptionalNumber('concurrencyLimit') ?? 10;
+      const throttleInterval = c.has('throttleInterval')
```
freben, line 74: Note that all config updates need a corresponding config.d.ts file update at the same time.

drodil (approved):
LGTM!

plugins/notifications-backend-module-slack/config.d.ts
```diff
@@ -28,6 +30,14 @@ export interface Config {
          * Names, or Slack Channel IDs. Any valid identifier that chat.postMessage can accept.
          */
         broadcastChannels?: string[];
+        /**
+         * Concurrency limit for Slack notifications, defaults to 10
+         */
+        concurrencyLimit?: number;
```
benjdlambert, line 36: Might be worth a callout that these are per backend instance of the notifications plugin, they're not shared across hosts as it's not distributed rate limiting.

docs/notifications/processors.md
```diff
@@ -149,6 +149,8 @@ notifications:
         broadcastChannels: # Optional, if you wish to support broadcast notifications.
           - C12345678
         username: 'Backstage Bot' # Optional, defaults to the name of the Slack App.
+        concurrencyLimit: 20 # Optional, number of messages allowed per interval. Defaults to 10.
+        throttleInterval: 1m # Optional, ISO 8601 duration (or ms value). Defaults to 1 minute.
```
freben: Maybe the wording of this text is ambiguous. `ms` values as in strings that the `ms` library supports parsing is ok. But millisecond values (purely a number with just digits) isn't. :)

So we support

* ISO6801
* `HumanDuration` form such as `{ minutes: 2, seconds: 30 }` and
* Whatever the `ms` library can parse.

freben (commented):
Yep this is looking good. Just the doc thing that Ben mentioned above and maybe a little tweaking to the wording I mention here, then we can 🚢

docs/notifications/processors.md
kaidubauskas-dd: Agreed. Switched to ms-style instead of ms value and added HumanDuration is a valid field

PR #31807, opened 2025-11-18
Do not perform a query when search term is empty or blank

### LINKED ISSUE #31695: Search trigger empty searches when load
### 📜 Issue Labels

### 🔎 Search Terms

```plain
search empty query
```

### 🗃️ Project Area

Search

### 🔗 External Integration

N/A

### 📝 Description & Context

When the searchBar, searchPage or searchModal is loaded new empty query is triggered to the backend, which is undesired.

Image

### 👍 Expected Behavior

When searchBar, searchModal or searchPage is loaded, no searches are performed until the user types something in the search bar.

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

Open the developer tools in your browser and go to https://demo.backstage.io/home. Observe that when opening the home, the Search Modal or the Search page, a query to `https://demo.backstage.io/api/search/query?term=` is performed.

Undecided

xoanmi:

Avoid empty queries on the search plugin

close #31695

### DIFF
--- a/.changeset/some-groups-send.md
+++ b/.changeset/some-groups-send.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-search': minor
+---
+
+Do not perform a query when search term is empty or blank
--- a/plugins/search/src/apis.test.ts
+++ b/plugins/search/src/apis.test.ts
@@ -18,8 +18,13 @@ import { MockFetchApi } from '@backstage/test-utils';
 import { SearchClient } from './apis';
 
 describe('apis', () => {
-  const query = {
+  const emptyQuery = {
     term: '',
+    filters: {},
+    types: [],
+  };
+  const query = {
+    term: 'abc',
     filters: {},
     types: [],
   };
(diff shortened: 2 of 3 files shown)


xoanmi:
> The search api can be used by other plugins which might require to perform a search query with an empty term in some contexts. Perhaps we should address this in the home page and not globally?

I’m curious about the use case for sending a request with an empty term. From my perspective, returning the first 25 entities without any context might not provide much value to the end user.

Could you clarify the scenarios where this behavior is actually helpful? If it’s needed for specific plugins, maybe we can handle it at the plugin level, so we keep the experience meaningful for most users.

vinzscam:
> > The search api can be used by other plugins which might require to perform a search query with an empty term in some contexts. Perhaps we should address this in the home page and not globally?
> 
> I’m curious about the use case for sending a request with an empty term. From my perspective, returning the first 25 entities without any context might not provide much value to the end user.
> 
> Could you clarify the scenarios where this behavior is actually helpful? If it’s needed for specific plugins, maybe we can handle it at the plugin level, so we keep the experience meaningful for most users.

A valid use case is if a plugin wants to scan the entire search content paginating through different pages

freben:
Agreed. Can this be a change in the hook/context instead of the client? Such that it skips calling the api on mount if there's no term, for example.

If you make this change "closer to the user", you'll also be closer to capturing their intent. All the way down in the general purpose client, you don't know what the intent of the calling code was.

freben:
Something like this https://github.com/backstage/backstage/pull/32080

xoanmi:
Achieved on https://github.com/backstage/backstage/pull/32080

PR #31810, opened 2025-11-18
`DCR`: Change internal `state` column to `text`

### LINKED ISSUE #31806: 🐛 Bug Report: `oauth_authorization_sessions.state` column VARCHAR(255) limit breaks Dynamic Client Registration with long state parameters
### 📜 Issue Labels

### 🔎 Search Terms

```plain
oauth_authorization_sessions, state column, varchar 255, value too long, dynamic client registration, auth backend, DCR state parameter
```

### 🗃️ Project Area

Auth

### 🔗 External Integration

N/A

### 📝 Description & Context

When using Dynamic Client Registration (DCR) with the auth backend, OAuth authorization flows fail with the following database error:

```
insert into "oauth_authorization_sessions" (...) values (...) - value too long for type character varying(255)
```

This occurs because the `state` parameter can legitimately exceed 255 characters in real-world OAuth flows, but the database schema limits it to `VARCHAR(255)`.

**Root Cause:**

In the migration file [`20250909120000_oidc_client_registration.js`](https://github.com/backstage/backstage/blob/master/plugins/auth-backend/migrations/20250909120000_oidc_client_registration.js#L91), the `state` column is defined inconsistently with other similar columns:

```javascript
// ✅ Unlimited
table.text('redirect_uri', 'longtext').notNullable()
table.text('scope').nullable()

// ❌ Limited to VARCHAR(255) by Knex default
table.string('state').nullable()  // Line 91
```

In Knex, `.string()` defaults to `VARCHAR(255)`, while `.text()` creates an unlimited text column.

**Context:**

- OAuth 2.0 [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1) does not specify a maximum length for the `state` parameter
- Real-world OAuth clients (e.g., API gateways, enterprise SSO proxies, security middleware) can generate state parameters exceeding 255 characters that include encrypted session data, routing information, or security tokens
- The same table already uses `.text()` for `redirect_uri` and `scope`, establishing precedent for unlimited-length OAuth parameters

### 👍 Expected Behavior

The `state` column should use `.text()` like other OAuth parameters (`redirect_uri`, `scope`) to accommodate state parameters of any length, allowing Dynamic Client Registration to work with all compliant OAuth clients.

### 📦 Reproduction Repo

Not applicable - this is a schema design issue in the core auth backend.

### 🥾 Reproduction steps

1. Configure Backstage with Dynamic Client Registration enabled:
   ```yaml
   auth:
     experimentalDynamicClientRegistration:
       enabled: true
       allowedRedirectUriPatterns:
         - https://example.com/callback
   ```

2. Use an OAuth client that generates a `state` parameter longer than 255 characters (e.g., an enterprise API gateway with encrypted session state)

3. Initiate OAuth authorization flow: 
   ```
   GET /api/auth/v1/authorize?client_id=...&state=<LONG_STATE_PARAMETER>&...
   ```

4. Observe database error in Backstage logs:
   ```
   error: insert into "oauth_authorization_sessions" (...) - value too long for type character varying(255)
   ```

**Example state parameter that triggers this** (280 characters - typical for encrypted session state):
```
eyJzZXNzaW9uX2lkIjoiYWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXoxMjM0NTYiLCJyb3V0aW5nIjp7InNlcnZpY2UiOiJiYWNrc3RhZ2UtYXV0aCIsImVudmlyb25tZW50IjoicHJvZCJ9LCJjc3JmX3Rva2VuIjoiOTg3NjU0MzIxMGFiY2RlZmdoaWprbG1ub3BxcnN0dXZ3eHl6IiwidGltZXN0YW1wIjoxNzA5NTY3ODkwLCJtZXRhZGF0YSI6eyJ1c2VyX2NvbnRleHQiOiJlbmNyeXB0ZWQtZGF0YS1oZXJlIn19
```
(_This is base64-encoded JSON with encrypted session metadata - common pattern in production OAuth flows_)

benjdlambert:
Fixes #31806

### DIFF
--- a/plugins/auth-backend/migrations/20251118120000_oauth_state_text.js
+++ b/plugins/auth-backend/migrations/20251118120000_oauth_state_text.js
@@ -29,6 +29,12 @@ exports.up = async function up(knex) {
  * @param {import('knex').Knex} knex
  */
 exports.down = async function down(knex) {
+  // Delete sessions with state > 255 chars since they won't fit in varchar(255)
+  // These sessions would be unusable with truncated state anyway
+  await knex('oauth_authorization_sessions')
+    .whereRaw('LENGTH(state) > 255')
+    .delete();
+
   await knex.schema.alterTable('oauth_authorization_sessions', table => {
     table.string('state').nullable().alter();
   });


plugins/auth-backend/migrations/20251118120000_oauth_state_text.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+// @ts-check
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function up(knex) {
+  await knex.schema.alterTable('oauth_authorization_sessions', table => {
+    table.text('state').nullable().alter();
```
freben:
```suggestion
    table.text('state', 'longtext').nullable().alter();
```

plugins/auth-backend/src/migrations.test.ts
```diff
@@ -304,4 +304,59 @@ describe('migrations', () => {
...
+
+      // First create a client for the foreign key constraint
+      await knex
+        .insert({
+          client_id: 'test-client-id',
+          client_secret: 'test-client-secret',
+          client_name: 'Test Client',
+          response_types: JSON.stringify(['code']),
+          grant_types: JSON.stringify(['authorization_code']),
+          redirect_uris: JSON.stringify(['https://example.com/callback']),
```
freben, line 323: I'd like for this to insert actual state, and then to check that it's still there and the same after the migration

plugins/auth-backend/migrations/20251118120000_oauth_state_text.js
benjdlambert: Is this just for MySQL support right?

plugins/auth-backend/src/migrations.test.ts
benjdlambert: Done!

plugins/auth-backend/migrations/20251118120000_oauth_state_text.js
freben: I think so.

Issue #31775, opened 2025-11-16
Gateway Backend Plugin Creates Self-DDOS Loops Under Multiple Failure Scenarios


kunickiaj:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
gateway
```

### 🗃️ Project Area

Core Framework

### 🔗 External Integration

N/A

### 📝 Description & Context

The `@backstage/plugin-gateway-backend` has a vulnerability that can cause cascading self-DDOS failures through infinite proxy loops. I've identified **two distinct failure modes**, both resulting in the backend proxying requests to itself:

**Failure Mode 1: Plugin Load Order Violation**

- When the gateway plugin initializes before other backend plugins, it captures an incomplete snapshot of `localPluginIds`
- Subsequent requests to those plugins are treated as "remote" and proxied back to the same service
- Creates an infinite loop: Request → Gateway → Proxy to self → Gateway → ...

**Failure Mode 2: Invalid Plugin IDs**

- Malformed requests like `/api/entities/by-uid/...` (missing the `catalog` plugin prefix) are captured by the gateway's catch-all route `/api/:pluginId`
- The gateway extracts `pluginId = "entities"` which doesn't exist in `localPluginIds`
- Discovery resolves to `http://localhost:7007/api/entities` → proxies to self → infinite loop

Both scenarios result in the backend consuming its own connection pool, competing with legitimate external traffic, and causing cascading failures.

`plugins/gateway-backend/src/router.ts:46-54`

```typescript
if (localPluginIds.has(req.params.pluginId)) {
  next(); // Local plugin
  return;
}
// ❌ No validation that pluginId is valid - blindly proxies unknown plugins
proxy(req, res, next);
```

```typescript
async getBaseUrl(pluginId: string): Promise<string> {
  const resolver =
    this.#internalResolvers.get(pluginId) ??
    this.#internalResolvers.get('*') ??       // ❌ Wildcard still resolves to localhost
    this.#internalFallbackResolver;           // ❌ Falls back to localhost
  return await resolver(pluginId);
}
```

The wildcard `*` configuration and fallback resolver both construct URLs like `http://localhost:7007/api/{{pluginId}}`, which creates a self-loop for invalid plugin IDs.

### 👍 Expected Behavior

1. **Load Order**: Gateway should either:
[...]

PR #31812, opened 2025-11-18
gateway-backend: add loop detection

Rugvip:

Fixes part of #31775. Opted for a hops detection because we can't reliably use `getBaseUrl` to check whether we're proxying to ourselves. Kept the count a bit over 1 because one could concievably chain multiple gateway plugins, but kept the count low because let's keep things sensible 😅 

### DIFF
--- a/plugins/gateway-backend/src/router.ts
+++ b/plugins/gateway-backend/src/router.ts
@@ -23,9 +23,13 @@ import { createProxyMiddleware } from 'http-proxy-middleware';
 import { context } from '@opentelemetry/api';
 import { getRPCMetadata } from '@opentelemetry/core';
 
+const MAX_HOPS = 3;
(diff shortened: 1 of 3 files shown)


plugins/gateway-backend/src/router.ts
```diff
@@ -62,6 +72,20 @@ export async function createRouter({
       return;
     }
 
+    const currentHops = parseInt(req.headers[HOPS_HEADER] as string, 10) || 0;
+    if (currentHops >= MAX_HOPS) {
+      logger.warn(
+        `Proxy loop detected: request exceeded maximum hop count (${currentHops})`,
```
freben: might make sense to include request path

Rugvip: 👍, threw in the plugin ID a least, the requests will be logged already

PR #31818, opened 2025-11-18
feat: enable react router feature flags for v7

schultzp2020:
Signed-off-by: Paul Schultz <pschultz@pobox.com>

Prepare for React Router v7 migration by updating to v6.30.2 across all NFS packages (`app-next`, `plugins/app`, and all `frontend-*` packages). Enable v7 future flags (`v7_relativeSplatPath` and `v7_startTransition`) in `AppRoot`. Convert routes from splat paths to parent/child structure with `Outlet` components in `AppRoutes`.

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -79,8 +79,8 @@
     "history": "^5.0.0",
     "react": "^18.0.2",
     "react-dom": "^18.0.2",
-    "react-router": "^6.3.0",
-    "react-router-dom": "^6.3.0",
+    "react-router": "^6.30.2",
+    "react-router-dom": "^6.30.2",
     "react-use": "^17.2.4",
     "zen-observable": "^0.10.0"
   },
--- a/plugins/app/src/extensions/AppRoot.tsx
+++ b/plugins/app/src/extensions/AppRoot.tsx
@@ -198,7 +198,17 @@ export interface AppRouterProps {
 function DefaultRouter(props: PropsWithChildren<{}>) {
   const configApi = useApi(configApiRef);
   const basePath = getBasePath(configApi);
-  return <BrowserRouter basename={basePath}>{props.children}</BrowserRouter>;
+  return (
+    <BrowserRouter
+      basename={basePath}
+      future={{
+        v7_relativeSplatPath: true,
+        v7_startTransition: true,
+      }}
+    >
+      {props.children}
+    </BrowserRouter>
+  );
 }
 
 /**
--- a/plugins/app/src/extensions/AppRoutes.tsx
+++ b/plugins/app/src/extensions/AppRoutes.tsx
@@ -20,7 +20,7 @@ import {
   createExtensionInput,
   NotFoundErrorPage,
 } from '@backstage/frontend-plugin-api';
-import { useRoutes } from 'react-router-dom';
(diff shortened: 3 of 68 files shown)


Rugvip (commented):
Keen to try this out! 👍 

Could you update the peer dep verification just to get this in a mergable state? 🙏

schultzp2020:
@Rugvip fixed 🙂

plugins/app/src/extensions/AppRoutes.tsx
```diff
@@ -38,14 +38,42 @@ export const AppRoutes = createExtension({
       const element = useRoutes([
         ...inputs.routes.map(route => {
           const routePath = route.get(coreExtensionData.routePath);
+          const routeElement = route.get(coreExtensionData.reactElement);
 
-          return {
-            path:
-              routePath === '/'
-                ? routePath
-                : `${routePath.replace(/\/$/, '')}/*`,
```
Rugvip: Hmm, given the need for this update, I suspect this must need an update as well or relative routing within entity page content will break? https://github.com/backstage/backstage/blob/2e26579e06197d0ba2d8e55952b40afeee6bf996/plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx#L38

Rugvip:
Looks good overall but feels like this'll need some tests added to make sure that the desired behavior is maintained and we don't break existing code. Probably including some tests that verifies behavior both with and without the new flag?

packages/app/package.json
```diff
@@ -80,7 +80,7 @@
     "react": "^18.0.2",
     "react-dom": "^18.0.2",
     "react-router": "^6.3.0",
```
Rugvip: Should be updated as well right?

Although I wonder about bumping this in general though, given these changes are only for NFS so far. Should we ship this for the old frontend system too?

schultzp2020: Hmm, I wonder why yarn didn't bump this when running `yarn up`. 🤔 

I do believe we should ship the version bump to the old frontend system for continuity.

schultzp2020:
@Rugvip The PR has been updated. PTAL 🙏

plugins/app/src/extensions/AppRoot.tsx
```diff
@@ -198,7 +198,17 @@ export interface AppRouterProps {
 function DefaultRouter(props: PropsWithChildren<{}>) {
   const configApi = useApi(configApiRef);
   const basePath = getBasePath(configApi);
-  return <BrowserRouter basename={basePath}>{props.children}</BrowserRouter>;
+  return (
+    <BrowserRouter
```
Rugvip, line 202: Think we need to update this as well right?

https://github.com/backstage/backstage/blob/d7dd5bdc805f23edaf27abbf2ab8e6f1cbe0611b/packages/frontend-test-utils/src/app/renderInTestApp.tsx#L190

https://github.com/backstage/backstage/blob/d7dd5bdc805f23edaf27abbf2ab8e6f1cbe0611b/packages/frontend-test-utils/src/app/renderTestApp.tsx#L87

schultzp2020: Fixed!

awanlin:
Here we go! 🚀

PR #31829, opened 2025-11-19
feat(nfs): created a router agnostic framework

schultzp2020:
Signed-off-by: Paul Schultz <pschultz@pobox.com>

This PR decouples the frontend system from a direct dependency on `react-router-dom` by introducing a router-agnostic `RouterApi` in `@backstage/frontend-plugin-api`. This abstraction defines a standard interface for routing operations—such as matching routes, generating paths, and navigation—allowing the core framework to operate independently of the underlying routing library.

The default implementation, `ReactRouter6RouterApi`, has been moved to a new package `@backstage/frontend-module-react-router-v6` and is provided via the app plugin (`@backstage/plugin-app`) as an `ApiBlueprint`. This follows the standard FrontendFeature pattern, meaning custom router implementations can be provided by supplying an alternative API factory for `routerApiRef`.

### Exported APIs

**Hooks** - Import from `@backstage/frontend-plugin-api`:
- `useLocation`, `useNavigate`, `useParams`, `useSearchParams`, `useResolvedPath`, `useHref`, `useOutlet`, `useRoutes`

**Utility functions** - Import from `@backstage/frontend-plugin-api`:
- `matchRoutes`, `generatePath`, `resolvePath`

> **Note:** These are intentionally named to match their `react-router-dom` counterparts so that migration is a simple import swap. They require component context due to internal `useApi()` lookup, which is documented in the `@remarks` JSDoc.

**Components** - Import from `@backstage/frontend-plugin-api`:
- `Link`, `NavLink`, `Outlet`, `Routes`, `Route`, `Navigate`

The `Routes` and `Route` components work around react-router's internal child type checking by using `useRoutes` internally, allowing them to be imported and used like standard React components.

### Testing

`MockMemoryRouterApi` and `MockBrowserRouterApi` are available in `@backstage/frontend-test-utils`. `MockMemoryRouterApi` uses a `MemoryRouter` internally and is suitable for most tests. `MockBrowserRouterApi` uses a `BrowserRouter` for tests that require real browser history APIs.

Convenience providers `TestMemoryRouterProvider` and `TestBrowserRouterProvider` are also available for wrapping test renders.

### React Router v7 Compatibility

A development-time warning has been added to detect multi-segment splat paths (e.g., `dashboard/*`) that have different relative link behavior in react-router v7. See: https://reactrouter.com/upgrading/v6#v7_relativesplatpath

### Usage

```tsx
import { 
  Routes, 
  Route, 
  Link, 
  useLocation, 
  useNavigate 
} from '@backstage/frontend-plugin-api';

function MyPage() {
  const location = useLocation();
  
  return (
    <Routes>
      <Route path="/" element={<Home />} />
      <Route path="/about" element={<About />} />
    </Routes>
  );
}
```

**Note:** Since the underlying implementation provided by default is still React Router v6, existing behavior remains unchanged. The primary purpose of this change is to pave the way for future migrations; once the ecosystem fully adopts the new frontend system and routing framework, we can upgrade to newer routers (such as React Router v7 or TanStack Router) without forcing widespread refactoring on adopters.


packages/frontend-app-api/report.api.md
```diff
@@ -158,6 +161,10 @@ export type CreateSpecializedAppOptions = {
       | ExtensionFactoryMiddleware
       | ExtensionFactoryMiddleware[];
     pluginInfoResolver?: FrontendPluginInfoResolver;
+    router?: {
+      matchRoutes: RoutingContextType['matchRoutes'];
+      generatePath: RoutingContextType['generatePath'];
+    };
```
Rugvip: Do we need to expose this as something configurable? I'm thinking it's best to keep this internal to make it easier to make changes

packages/frontend-app-api/report.api.md
```diff
@@ -174,4 +181,18 @@ export type FrontendPluginInfoResolver = (ctx: {
 }) => Promise<{
   info: FrontendPluginInfo;
 }>;
+
+// @public
+export const ReactRouter6Provider: ({
+  children,
+}: {
+  children: ReactNode;
+}) => JSX_2.Element;
```
Rugvip: Same as above, although I could potentially see the need for something like a `<RoutingProvider router={...}>`

packages/frontend-app-api/src/routing/ReactRouter6Provider.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Adapts the React Router v6 context to the Backstage RoutingContext.
+ *
+ * @public
+ */
+export const ReactRouter6Provider = ({ children }: { children: ReactNode }) => {
+  const location = useLocation();
+  const navigate = useNavigate();
+  const params = useParams();
```
Rugvip: Do any of these cause rerender on anything but a location change? Best be quite careful about these and maybe consume them more lazily or not as hooks

packages/frontend-app-api/src/routing/ReactRouter6Provider.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const newRoute: RouteObject = { ...route };
+    if (newRoute.element) {
+      newRoute.element = <ParamsProvider>{newRoute.element}</ParamsProvider>;
+    }
+    if (newRoute.children) {
+      newRoute.children = collectRoutes(newRoute.children);
+    }
+    return newRoute;
+  });
+};
```
Rugvip: Feeling there's an assumption here that all packages move at once. I'd expect this to cause trouble if you for example have the catalog on an old version that doesn't use this new context yet, but some piece of content on the entity page does.

packages/frontend-plugin-api/src/blueprints/RouterBlueprint.test.tsx
```diff
@@ -59,9 +61,9 @@ describe('RouterBlueprint', () => {
     const extension = RouterBlueprint.make({
       params: {
         component: ({ children }) => (
-          <MemoryRouter>
+          <TestRouterProvider>
```
Rugvip: I do like test / main provider, doesn't need to be more complicated than that

packages/frontend-plugin-api/src/routing/RoutingContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+  Link: React.ComponentType<{
+    to: string;
+    children?: ReactNode;
+    className?: string;
+  }>;
+  Outlet: React.ComponentType<{}>;
+}
+
+/** @public */
+export const RoutingContext = createContext<RoutingContextType | undefined>(
```
Rugvip: Bit more down the line for this PR perhaps, but we'd never want to expose a React context directly, but instead use a global singleton and versioned context via the `@backstage/version-bridge` utilities

packages/frontend-plugin-api/report.api.md
```diff
@@ -1954,6 +2034,9 @@ export function useRouteRefParams<Params extends AnyRouteRefParams>(
   _routeRef: RouteRef<Params> | SubRouteRef<Params>,
 ): Params;
 
+// @public (undocumented)
+export function useRouting(): RoutingContextType;
```
Rugvip: I'm thinking the public API should most likely be the what react-router already provides, more or less? `Route`, `Routes`, `Outlet`, `useLocation`, `useNavigation`, etc. Feels like that'd be a bit of a requirement to eventually move over. With that the context type would be an internal thing. A potential alternate approach to context could be to just go with a `RoutingApi` as well, just like how we switched from the route resolution context in the old frontend system to `RouteResolutionApi` in NFS. That has already turned out to be quite useful in a couple of places.

It's a little bit scary thinking about our ability to push and verify this change though, since we can't exactly deprecate the RR methods directly. We'd need to aim for the ability to manage the full migration with lint rules and codemods I'd think.

Rugvip (commented):
Thank you for getting this ball rolling! 👍

packages/frontend-app-api/report.api.md
schultzp2020: We should expose it because without these properties, users who want to switch to React Router v7 or TanStack Router would need to reimplement createSpecializedApp entirely.

packages/frontend-app-api/report.api.md
schultzp2020: It's exported so it can be used in `ForwardsCompatProvider.tsx` and `TestRouterProvider.tsx`.

packages/frontend-plugin-api/src/blueprints/RouterBlueprint.test.tsx
schultzp2020: I'm a little confused 😅, does this comment mean you like the implementation or would you rather users need to specify the router and the provider separately?

schultzp2020:
@Rugvip PTAL?

packages/frontend-plugin-api/src/blueprints/RouterBlueprint.test.tsx
Rugvip: yep I mean the current API is good

packages/frontend-app-api/report.api.md
Rugvip: Point is to abstract this away though so that we can switch out the implementation in app-api without impact to consumers. Thinking at most we'd have a flag for it in these options?

Rugvip:
@schultzp2020 replied to some things, but I think the main discussion to focus on settling before moving forward is https://github.com/backstage/backstage/pull/31829#discussion_r2561635715

packages/frontend-app-api/report.api.md
schultzp2020: I've update the abstraction, please let me know if I am heading in the right direction.

schultzp2020:
> @schultzp2020 replied to some things, but I think the main discussion to focus on settling before moving forward is [#31829 (comment)](https://github.com/backstage/backstage/pull/31829#discussion_r2561635715)

@Rugvip Done! Please let me know what you think. I really like the API approach you suggested. 🙂

docs/frontend-system/utility-apis/05-router-api.md
```diff
@@ -0,0 +1,274 @@
+---
```
schultzp2020, line 1: Is this a good location for the docs?

schultzp2020:
Any idea why I'm having issues with the api report for unrelated changes?

docs/frontend-system/utility-apis/01-index.md
```diff
@@ -35,6 +35,12 @@ Most utility APIs are usable directly without any configuration. But they are pr
 
 These cases are all described in [the main article](./04-configuring.md).
 
+## Router API
+
+> For details, [see the main article](./05-router-api.md).
+
+The Router API abstracts routing operations in Backstage, providing hooks like `useLocation` and `useNavigate`, and components like `Link`, `Routes`, and `Route`. This abstraction allows the underlying router implementation to be swapped without affecting plugin code, enabling future migrations to newer routers like React Router v7 or TanStack Router.
```
Parsifal-M: Do we want to specifically name **React Router v7** or **TanStack Router**? Or is making it easier to migrate to newer routers enough?

Parsifal-M (approved):
Approving from docs side of things, makes sense to me, just a small comment but not blocking 😄 👍

aramissennyeydd:
How will this impact third-party packages that have a peer dependency on react-router or react-router-dom? In a yarn world that matters less because of hoisting, but in pnpm that could have hard-to-debug implications. ie does this package supply the peer dependency for resolution purposes?

schultzp2020:
> How will this impact third-party packages that have a peer dependency on react-router or react-router-dom? In a yarn world that matters less because of hoisting, but in pnpm that could have hard-to-debug implications. ie does this package supply the peer dependency for resolution purposes?

From my understanding, Backstage only officially supports Yarn right now, so this shouldn't be an immediate concern. However, the architectural approach here should work well if/when pnpm support is added.

In pnpm strict mode, these plugins would still need to provide and resolve their peer dependency from the app's direct `react-router-dom` dependency. Once all of the plugins have migrated, they can remove this dependency.

packages/frontend-plugin-api/src/routing/index.ts
```diff
@@ -14,6 +14,45 @@
...
+  useNavigate,
+  useSearchParams,
+  useResolvedPath,
+  useHref,
+  useOutlet,
+  useRoutes,
+  resolvePath,
+  matchRoutes,
+  generatePath,
+} from './hooks';
```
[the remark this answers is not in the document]
schultzp2020: These are intentionally named to match their react-router-dom counterparts (`matchRoutes`, `generatePath`, `resolvePath`) so that migration is a simple import swap. They require component context due to `useApi()` lookup, which is documented in the `@remarks` JSDoc. Renaming to `use*` would break API parity and add friction to the migration path.

packages/ui/src/components/PluginHeader/PluginHeader.tsx
```diff
@@ -18,7 +18,8 @@ import type { PluginHeaderProps } from './types';
 import { Tabs, TabList, Tab } from '../Tabs';
 import { useDefinition } from '../../hooks/useDefinition';
 import { PluginHeaderDefinition } from './definition';
-import { type NavigateOptions } from 'react-router-dom';
+// eslint-disable-next-line @backstage/no-relative-monorepo-imports
+import { type NavigateOptions } from '../../../../frontend-plugin-api/src/routing/routerTypes';
```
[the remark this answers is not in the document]
schultzp2020: `@backstage/ui` intentionally does not declare a dependency on `@backstage/frontend-plugin-api` — the UI primitives layer is kept decoupled from the frontend system layer. The relative monorepo import with the `eslint-disable` comment is a deliberate exception to `@backstage/no-relative-monorepo-imports`, following the same pattern used by `core-compat-api` (which imports from `core-app-api` and `core-plugin-api` the same way). The `backstage-cli` build tooling handles resolution during compilation.

packages/ui/src/components/InternalLinkProvider/InternalLinkProvider.tsx
```diff
@@ -23,7 +23,11 @@ import {
   useState,
 } from 'react';
 import { RouterProvider } from 'react-aria-components';
-import { useNavigate, useHref } from 'react-router-dom';
+// eslint-disable-next-line @backstage/no-relative-monorepo-imports
+import {
+  useNavigate,
+  useHref,
+} from '../../../../frontend-plugin-api/src/routing/hooks';
```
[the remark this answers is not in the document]
schultzp2020: Same rationale as the `PluginHeader.tsx` comment — `@backstage/ui` intentionally does not depend on `@backstage/frontend-plugin-api` to keep the dependency hierarchy clean. The relative monorepo import with `eslint-disable` is a deliberate pattern used elsewhere in the monorepo (e.g., `core-compat-api`). The build tooling resolves these during compilation.

packages/ui/src/components/Tabs/Tabs.tsx
```diff
@@ -34,7 +34,8 @@ import type {
   TabsContextValue,
   TabProps,
 } from './types';
-import { useLocation } from 'react-router-dom';
+// eslint-disable-next-line @backstage/no-relative-monorepo-imports
+import { useLocation } from '../../../../frontend-plugin-api/src/routing/hooks';
```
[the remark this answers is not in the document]
schultzp2020: Same as the other `@backstage/ui` import comments — intentional relative monorepo import to avoid `@backstage/ui` depending on `@backstage/frontend-plugin-api`. The eslint-disable comment signals this is a deliberate exception to the `@backstage/no-relative-monorepo-imports` rule.

packages/frontend-plugin-api/src/apis/definitions/RouterApi.ts
```diff
@@ -0,0 +1,136 @@
...
+  Location,
+  To,
+  Path,
+  NavigateFunction,
+  RouteObject,
+  RouteMatch,
+  LinkProps,
+  NavLinkProps,
+  NavigateProps,
+} from '../../routing';
```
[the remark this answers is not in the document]
schultzp2020: Already using `import type` — these are type-only imports that are erased at compile time and do not cause runtime circular dependencies.

docs/frontend-system/utility-apis/01-index.md
```diff
@@ -43,6 +43,12 @@ When testing frontend components and extensions, you often need to provide mock
 
 These are described in detail in [the main article](./05-testing.md).
 
+## Router API
+
+> For details, [see the main article](./06-router-api.md).
```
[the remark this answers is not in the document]
schultzp2020: The link target is correct — the file is `05-router-api.md`, which matches `./05-router-api.md`.

packages/frontend-test-utils/src/apis/RouterApi/MockBrowserRouterApi.tsx
```diff
@@ -0,0 +1,180 @@
...
+  private basePath: string;
+
+  constructor(options?: MockBrowserRouterApiOptions) {
+    this.basePath = options?.basePath ?? '';
+  }
+
+  Router: ComponentType<{ children: ReactNode; basePath: string }> = ({
+    children,
+  }) => <BrowserRouter basename={this.basePath}>{children}</BrowserRouter>;
+
```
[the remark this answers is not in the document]
schultzp2020: Already addressed — `MockBrowserRouterApi.Router` now accepts the `basePath` prop and passes it to `BrowserRouter` via `basename={basePath}`.

packages/ui/src/components/PluginHeader/PluginHeader.tsx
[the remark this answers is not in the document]
schultzp2020: This is intentional. Adding `@backstage/frontend-plugin-api` as a dependency of `@backstage/ui` creates a circular dependency through Yarn workspace resolution — Yarn doesn't differentiate between dev and normal dependencies when building the dependency graph, so the cycle `@backstage/ui` → `@backstage/frontend-plugin-api` → (devDep) `@backstage/frontend-test-utils` → (devDep) `@backstage/ui` causes `yarn install` to fail.

packages/ui/src/components/InternalLinkProvider/InternalLinkProvider.tsx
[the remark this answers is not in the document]
schultzp2020: Same rationale as the PluginHeader comment — `@backstage/ui` intentionally does not declare a dependency on `@backstage/frontend-plugin-api` because it creates a circular dependency through Yarn workspace resolution (Yarn treats dev and normal deps identically when resolving the graph).

packages/ui/src/components/Tabs/Tabs.tsx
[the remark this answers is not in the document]
schultzp2020: Same as the other `@backstage/ui` comments — intentional relative monorepo import to avoid a circular dependency in Yarn workspace resolution. See the reply on the PluginHeader comment for the full explanation.

schultzp2020:
replaced by https://github.com/backstage/backstage/pull/33900

PR #31832, opened 2025-11-20
add new patch release process

Rugvip:

Had an idea for improving our patch release process a bit so it scales better across people. The idea is to be able to queue up a couple of patch fixes and then release them all at once using a generated PR. A very lightweight version of the changesets flow, mostly vibes.

Also included an initial patch for https://github.com/backstage/backstage/pull/31817 to try it out

### DIFF
--- a/scripts/patch-release-for-pr.js
+++ b/scripts/patch-release-for-pr.js
@@ -314,7 +314,13 @@ async function main(args) {
   // Generate PR body
   let body;
   if (descriptions) {
-    const descriptionList = descriptions.map(desc => `- ${desc}`).join('\n');
+    const descriptionList = descriptions
+      .map((desc, index) => {
+        const prNumber = prNumbers[index];
+        const prLink = `https://github.com/${owner}/${repo}/pull/${prNumber}`;
+        return `- ${desc} ([#${prNumber}](${prLink}))`;
+      })
+      .join('\n');
     body = `This patch release includes the following fixes:\n\n${descriptionList}`;
   } else {
     body = 'This release fixes an issue where';


scripts/patch-release-for-pr.js
```diff
@@ -81,15 +82,120 @@ async function findCurrentReleaseVersion() {
...
+  if (useDirectory) {
+    // Read patches from directory
+    const patchesFromDir = await readPatchesFromDirectory();
+
+    if (patchesFromDir) {
+      // Use patches from directory
+      prNumbers = patchesFromDir.map(p => p.prNumber);
+      descriptions = patchesFromDir.map(p => p.description);
+      console.log(`Reading patches from .patches/ directory`);
+      console.log(`PR number(s): ${prNumbers.join(', ')}`);
```
freben, line 115: could also log a list of full URLs (maybe one per line?) for easy clickability. Same below

scripts/patch-release-for-pr.js
```diff
@@ -109,8 +225,24 @@ async function main(args) {
...
+  // Check if branch already exists (for CI workflows using fixed branch name)
+  try {
+    await run(
+      'git',
+      'show-ref',
+      '--verify',
+      '--quiet',
+      `refs/heads/${branchName}`,
+    );
+    await run('git', 'checkout', branchName);
```
freben, line 242: it SHOULD never be a problem ... but it's tempting to suggest a pull here

scripts/patch-release-for-pr.js
```diff
@@ -165,18 +304,35 @@ async function main(args) {
...
+  if (process.env.PATCH_RELEASE_BRANCH) {
+    await run('git', 'push', 'origin', '-u', '--force-with-lease', branchName);
+  } else {
+    await run('git', 'push', 'origin', '-u', branchName);
+  }
+
+  // Generate PR body
+  let body;
+  if (descriptions) {
+    const descriptionList = descriptions.map(desc => `- ${desc}`).join('\n');
```
freben: Nice with pr links here too

scripts/patch-release-for-pr.js
Rugvip: There's a `git fetch` at the top that covers it

scripts/patch-release-for-pr.js
Rugvip: Hmm, we don't include them in patch releases right now and I kept them out of here intentionally too just to keep the notes as simple as possible. But meh I don't think it hurts, will add.

scripts/patch-release-for-pr.js
Rugvip: it's just informative, hidden in the CI logs

PR #31849, opened 2025-11-22
reverse relationship between core-plugin-api and frontend-plugin-api

Rugvip:

In order to be able to add more compatability for NFS in the old system, we need to inverse the relationship between the packages so that core-plugin-api depends on frontend-plugin-api, instead of the other way around. This is of course also needed in general to be able to promote NFS to the default and eventually deprecate/remove the old.

The goal of this change is for all API contracts to remain intact, and just move things around between the two packages.

### DIFF
--- a/docs/api/utility-apis.md
+++ b/docs/api/utility-apis.md
@@ -18,23 +18,23 @@ plugins to communicate during their entire life cycle.
 
 ## Consuming APIs
 
-Each Utility API is tied to an [`ApiRef`](../reference/core-plugin-api.apiref.md)
+Each Utility API is tied to an [`ApiRef`](../reference/frontend-plugin-api.apiref.md)
 instance, which is a global singleton object without any additional state or
 functionality, its only purpose is to reference Utility APIs.
-[`ApiRef`](../reference/core-plugin-api.apiref.md)s are created using
-[`createApiRef`](../reference/core-plugin-api.createapiref.md), which is exported
+[`ApiRef`](../reference/frontend-plugin-api.apiref.md)s are created using
+[`createApiRef`](../reference/frontend-plugin-api.createapiref.md), which is exported
 by [`@backstage/core-plugin-api`](../reference/core-plugin-api.md). There are also
 many predefined Utility APIs in
 [`@backstage/core-plugin-api`](../reference/core-plugin-api.md), and they're all
 exported with a name of the pattern `*ApiRef`, for example
-[`errorApiRef`](../reference/core-plugin-api.errorapiref.md).
+[`errorApiRef`](../reference/frontend-plugin-api.errorapiref.md).
 
 To access one of the Utility APIs inside a React component, use the
-[`useApi`](../reference/core-plugin-api.useapi.md) hook exported by
+[`useApi`](../reference/frontend-plugin-api.useapi.md) hook exported by
 [`@backstage/core-plugin-api`](../reference/core-plugin-api.md), or the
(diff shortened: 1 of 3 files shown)


packages/frontend-plugin-api/src/components/ExtensionBoundary.test.tsx
```diff
@@ -93,7 +90,7 @@ describe('ExtensionBoundary', () => {
   it('should wrap children with analytics context', async () => {
     const action = 'render';
     const subject = 'analytics';
-    const analyticsApiMock = mockApis.analytics();
+    const analyticsApiMock = { captureEvent: jest.fn() };
```
freben, line 93: Did frontend-test-utils not work here?

Interestingly, it DOES happen to have [specifically a MockAnalyticsApi](https://github.com/backstage/backstage/blob/2b3fb40d068d1a23c38805113d20f1da0ffd49d5/packages/frontend-test-utils/src/apis/AnalyticsApi/MockAnalyticsApi.ts#L25) as a separate thing.

packages/frontend-plugin-api/package.json
```diff
@@ -16,8 +14,19 @@
...
+  },
   "main": "src/index.ts",
   "types": "src/index.ts",
+  "typesVersions": {
+    "*": {
+      "package.json": [
+        "package.json"
+      ]
+    }
+  },
```
freben, line 29: Not a necessary change right? But doesn't hurt either. Did you initially have some alpha stuff here while working on it?

packages/frontend-plugin-api/src/apis/system/index.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 The Backstage Authors
+ * Copyright 2020 The Backstage Authors
```
freben, line 2: Just noting there's a couple of these but it really doesn't matter.

Rugvip: Yeah it's because it's moving in the `core-plugin-api` ones. Arguably correct and ye let's not worry about it 😁

packages/frontend-plugin-api/package.json
Rugvip: Yep, added alpha things but realized they weren't alpha anymore so rolled that back. This is still a more convenient spot to be in imo.

packages/frontend-plugin-api/src/components/ExtensionBoundary.test.tsx
Rugvip: Ah yep didn't try it because it didn't have `mockApis`. Old one doesn't work because of the unique implementation, the context is different. Feeling this is alright to keep as is

freben: Oh it does, [it re-exports mockApis](https://github.com/backstage/backstage/blob/2b3fb40d068d1a23c38805113d20f1da0ffd49d5/packages/frontend-test-utils/report.api.md?plain=1#L97).

Maybe we want to invert that too later, but it does have them.,

Oh I see what you mean, it does have it but still of the wrong shape anyway. Gotcha. Not blocking on that either way, we can follow up some other time

Rugvip: Ah, that needs a fix, the fact that those `mockApis` aren't compatible with NFS that is

PR #31854, opened 2025-11-23
core-plugin-api: add forwards compatibility with frontend-plugin-api

Rugvip:

This adds forwards compatibility with the new frontend system for `useApp` and `useRouteRef` in `@backstage/core-plugin-api`. Together with the changes in #31804, we now have full compatibility with NFS from the old. This in turn removes the need for using `compatWrapper` when using the old frontend system within the new one, but it's still required if you want to go the other way around.

On top of #31849, since `core-plugin-api` needs to be able to import from `frontend-plugin-api` for these conversions to be possible.

### DIFF
--- a/packages/core-plugin-api/src/app/useApp.tsx
+++ b/packages/core-plugin-api/src/app/useApp.tsx
@@ -25,6 +25,7 @@ import {
   Progress,
   createFrontendPlugin,
   FrontendPlugin,
+  ApiHolder,
 } from '@backstage/frontend-plugin-api';
 import {
   AppComponents,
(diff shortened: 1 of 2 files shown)


plugins/api-docs/report-alpha.api.md
```diff
@@ -13,12 +13,13 @@ import { EntityPredicate } from '@backstage/plugin-catalog-react/alpha';
 import { ExtensionBlueprintParams } from '@backstage/frontend-plugin-api';
 import { ExtensionDataRef } from '@backstage/frontend-plugin-api';
 import { ExtensionInput } from '@backstage/frontend-plugin-api';
-import { ExternalRouteRef } from '@backstage/frontend-plugin-api';
+import { ExternalRouteRef } from '@backstage/core-plugin-api';
```
freben, line 16: Right, so because the plugin still is old-system compatible you can now see visibly on the feature that the ref is old, which wasn't the case before. Is that feature field/type not consumed directly by any piece of code that could now get confused by that fact?

I gess the alternative would be to force-convert the refs still behind the scenes.

Put more broadly - is it ok that the shape of the api report exported things change?

plugins/catalog/report-alpha.api.md
freben, line 1: Catalog seems to still have the core-compat-api package in package.json

freben (approved):

plugins/api-docs/report-alpha.api.md
Rugvip: yep the report change is expected. Since there's no longer a need to convert we get the actual ref passed though on the plugin object, but since the refs from `core-plugin-api` are now a superset of the ones from `frontend-plugin-api`, that's all fine. It just makes it a bit harder to evolve the new refs, but tbh we shouldn't be doing any of that before NFS is fully rolled out anyway 😁

plugins/catalog/report-alpha.api.md
Rugvip: Yeah it's needed for the `entityRouteRef` conversion: https://github.com/backstage/backstage/blob/debf9ea34f0874d5decb9b828189f842032d5fff/plugins/catalog/src/alpha/pages.tsx#L111-L117

freben: ohh right

PR #31855, opened 2025-11-23
chore: support Entra authentication for Azure Database for PostgreSQL

arjan07:

This PR adds passwordless Entra ID authentication support for Azure Database for PostgreSQL by setting `database.connection.type` to `azure`

This implementation is based on the earlier work by @mike-schenk in #28532. I’ve rebased the feature on current main, incorporated their initial design, and addressed the remaining comments. It also addresses a much requested feature discussed in #22535.

### DIFF
--- a/packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
+++ b/packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
@@ -33,7 +33,7 @@ import { Client } from 'pg';
 import { Connector } from '../types';
 import { mergeDatabaseConfig } from './mergeDatabaseConfig';
 import format from 'pg-format';
-import { AccessToken, TokenCredential } from '@azure/identity';
+import { TokenCredential } from '@azure/identity';
 
 // Limits the number of concurrent DDL operations to 1
 const ddlLimiter = limiterFactory(1);
@@ -127,7 +127,7 @@ export type AzureTokenCredentialConfig = {
    * - A standard ISO formatted duration string, e.g. 'P2DT6H' or 'PT1M'.
    * - An object with individual units (in plural) as keys, e.g. `{ days: 2, hours: 6 }`.
    */
-  tokenRenewalOffsetTime?: string | HumanDuration;
+  tokenRenewableOffsetTime?: string | HumanDuration;
   /**
    * The client ID of a user-assigned managed identity.
    * If not provided, the system-assigned managed identity is used.
@@ -146,9 +146,9 @@ export async function buildAzurePgConfig(config: Config): Promise<Knex.Config> {
 
   const tokenConfig = config.getOptionalConfig('connection.tokenCredential');
 
-  const tokenRenewalOffsetMs = durationToMilliseconds(
-    tokenConfig?.has('tokenRenewalOffsetTime')
-      ? readDurationFromConfig(tokenConfig, { key: 'tokenRenewalOffsetTime' })
+  const tokenRenewableOffsetTime = durationToMilliseconds(
+    tokenConfig?.has('tokenRenewableOffsetTime')
+      ? readDurationFromConfig(tokenConfig, { key: 'tokenRenewableOffsetTime' })
       : { minutes: 5 },
   );
 
@@ -181,16 +181,22 @@ export async function buildAzurePgConfig(config: Config): Promise<Knex.Config> {
   ]) as Partial<Knex.StaticConnectionConfig>;
 
   async function getConnectionConfig() {
-    const token = (await credential.getToken(
+    const token = await credential.getToken(
       'https://ossrdbms-aad.database.windows.net/.default',
-    )) as AccessToken; // This cast is safe since we know all of our TokenCredential implementations return this type.
+    );
+
+    if (!token) {
+      throw new Error(
+        'Failed to acquire Azure access token for database authentication',
+      );
+    }
 
     const connectionConfig = {
       ...sanitizedConnection,
       password: token.token,
       expirationChecker: () =>
-        /* return true if the token is within the renewal offset time */
-        token.expiresOnTimestamp - tokenRenewalOffsetMs <= Date.now(),
+        /* return true if the token is within the renewable offset time */
+        token.expiresOnTimestamp - tokenRenewableOffsetTime <= Date.now(),
     };
 
     return connectionConfig;
--- a/packages/backend-defaults/src/entrypoints/database/connectors/postgres.test.ts
+++ b/packages/backend-defaults/src/entrypoints/database/connectors/postgres.test.ts
@@ -333,7 +333,7 @@ describe('postgres', () => {
           connection: {
             type: 'azure',
             tokenCredential: {
-              tokenRenewalOffsetTime: '1 minute',
+              tokenRenewableOffsetTime: '1 minute',
             },
             user: 'some-user@domain.com',
             database: 'other_db',
@@ -356,7 +356,7 @@ describe('postgres', () => {
       expirationResult = await connectionResult.expirationChecker();
       expect(expirationResult).toBe(true);
 
-      // Check the default tokenRenewalOffsetTime of 5 minutes
+      // Check the default tokenRenewableOffsetTime of 5 minutes
       configResult = await buildPgDatabaseConfig(
         new ConfigReader({
           client: 'pg',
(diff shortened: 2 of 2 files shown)


packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -82,15 +91,113 @@ export async function buildPgDatabaseConfig(
...
+  /**
+   * The client ID of a user-assigned managed identity.
+   * If not provided, the system-assigned managed identity is used.
+   */
+  clientId?: string;
+  clientSecret?: string;
+  tenantId?: string;
+};
+
+export async function buildAzurePgConfig(config: any, sanitizedConfig: any) {
```
benjdlambert: I wonder if we should move the `sanitizedConfig` building to these individual methods as they know what extra config shouldn't be there rather than just doing this for all up front.

Also wanna suggest that we pass in `ConfigService` instead of having untyped parameters here as `any` and doing the `getOptionalString` etc in these methods where they parse out the original config.

benjdlambert: We already build the plugin config once, can we reuse that? Bit of a smell to be creating it many times from a `JSONObject`?

arjan07: Hi, thank you for reviewing. I've tried my best to address the comments you've left. Please let me know if there's anything else that needs to be addressed. Thank you!

arjan07:
Hi @benjdlambert Please could I get a re-review on this PR. Thanks

freben:
We'll review this asap again, but until then - noting that it needs a rebase since parts of the postgres code has been improved

vinzscam:
This is great @arjan07!

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -76,26 +85,146 @@ export async function buildPgDatabaseConfig(
...
+export async function buildAzurePgConfig(
+  configReader: ConfigReader,
+): Promise<Knex.Config> {
+  const {
+    DefaultAzureCredential,
+    ManagedIdentityCredential,
+    ClientSecretCredential,
+  } = require('@azure/identity');
+
+  let tokenRenewalOffsetMs = 300_000;
```
freben: (optional) A trick to make this only use human readable durations even in code if you like:

```ts
const renewalKey = 'connection.tokenCredential.tokenRenewalOffsetTime';
const tokenRenewalOffsetMs = durationToMilliseconds(
  configReader.has(renewalKey)
    ? readDurationFromConfig(configReader, { key: renewalKey }) :
    : { minutes: 5 }
);
```

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -76,26 +85,146 @@ export async function buildPgDatabaseConfig(
...
+   * The client ID of a user-assigned managed identity.
+   * If not provided, the system-assigned managed identity is used.
+   */
+  clientId?: string;
+  clientSecret?: string;
+  tenantId?: string;
+};
+
+export async function buildAzurePgConfig(
+  configReader: ConfigReader,
```
freben: this can be relaxed to be `config: Config` instead; the `ConfigReader` is an implementation class (among many) and `Config` is the actual interface

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -76,26 +85,146 @@ export async function buildPgDatabaseConfig(
...
+    case 'cloudsql':
+      return buildCloudSqlConfig(mergedConfigReader);
+    default:
+      throw new Error(`Unknown connection type: ${config.connection.type}`);
+  }
+}
+
+/* Note: the following type definition is intentionally duplicated in
+ * /packages/backend-defaults/config.d.ts so the clientSecret property
+ * can be annotated with "@visibility secret" there.
```
freben, line 118: Also - unfortunately - the `config.d.ts` format doesn't really support imported types properly in it anyway at this point. It's something that we've wanted to improve but haven't gotten around to yet. You'll see a fair bit of copy pasting of types in those files for that reason. Even saying `HumanDuration` in it is technically iffy but we still do it, so you don't have to change that.

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -76,26 +85,146 @@ export async function buildPgDatabaseConfig(
...
+  let tokenRenewalOffsetMs = 300_000;
+  if (configReader.has('connection.tokenCredential.tokenRenewalOffsetTime')) {
+    tokenRenewalOffsetMs = durationToMilliseconds(
+      readDurationFromConfig(configReader, {
+        key: 'connection.tokenCredential.tokenRenewalOffsetTime',
+      }),
+    );
+  }
+
+  const tokenConfig = (configReader.getOptional('connection.tokenCredential') ??
```
freben: I would recommend that you do

```ts
  const tokenConfig = configReader.getOptionalConfig('connection.tokenCredential');
  const clientId = tokenConfig?.getOptionalString('clientId');
```

for two reasons - first, it requires that the parent is object shaped if it exists (`getOptional` does no type validation), and second, you also ensure that the desired types are respected all the way out to the leaves. And if something is misconfigured you'll get errors with the full context of precisely what key in what file on what row was wrong and why.

Generally, staying in the Config type land as long as possible helps both developers and end users.

arjan07:
Hi! Thank you for reviewing @freben! I've addressed your comments, please let me know if there's anything else. Thank you 🙂

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -74,26 +83,136 @@ export async function buildPgDatabaseConfig(
...
+  const {
+    DefaultAzureCredential,
+    ManagedIdentityCredential,
+    ClientSecretCredential,
+  } = require('@azure/identity');
+
+  const tokenConfig = config.getOptionalConfig('connection.tokenCredential');
+
+  const tokenRenewalOffsetMs = durationToMilliseconds(
+    tokenConfig?.has('tokenRenewalOffsetTime')
```
benjdlambert: https://github.com/backstage/backstage/pull/31855/changes#diff-3b45dbb2aa9a6a351eed560b0388e73b659ab55bc693efabecbb307b7bbe9b0aR612

Config defines another key name, let's pick one :)

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -74,26 +83,136 @@ export async function buildPgDatabaseConfig(
...
+
+  const normalized = normalizeConnection(rawConfig.connection as any);
+  const sanitizedConnection = omit(normalized, [
+    'type',
+    'instance',
+    'tokenCredential',
+  ]) as Partial<Knex.StaticConnectionConfig>;
+
+  async function getConnectionConfig() {
+    const token = (await credential.getToken(
```
benjdlambert: If auth fails, could this return `null` and we'd end up running into a weird error instead? Should we add a `null` check here?

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
arjan07: Good catch. I've renamed the key to use the one from `config.d.ts`

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
arjan07: I've removed the casting, and added a check to see if the token is null. If it is null, it will now throw an error - similar behaviour to other areas in here. Also added an additional test for this new logic.

arjan07:
Hi, would it be possible to get a re-review. Thank you!

benjdlambert (approved):
Nice! this is looking good! :tada:

packages/backend-defaults/config.d.ts
```diff
@@ -592,6 +592,32 @@ export interface Config {
        */
       connection:
         | string
+        | {
```
karpfslp, line 595: This causes an error in `yarn backstage-cli config:check --strict --config app-config.yaml` because this type is equal to the last type: `[key: string]: unknown;` and always this type wins as first one.

karpfslp: I'm trying to refine it in this [PR](https://github.com/backstage/backstage/pull/32942).

PR #31508, opened 2025-10-23
Make Breadcrumbs Clickable in Backstage Documentation to Improve Navigation and Accessibility #31079

gyan9427:
Open  Issue (#31079)

This PR converts the previously commented, object-style sidebar structure in [sidebars.ts](vscode-file://vscode-app/c:/Users/gyandev/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) into the Docusaurus-compatible category object format (using type, label, link, and items). The new structure mirrors the same content (Overview, Getting Started, Core Features, Integrations, Plugins, Configuration, Framework, Tutorials, FAQ, Contribute, References, and dynamic releases). It preserves the existing dynamic sidebar arrays (for example searchSidebar, catalogSidebar, scaffolderSidebar, and releases) and keeps generated-index links where appropriate.

PR #31863, opened 2025-11-25
refactor(microsite): extract sidebar loader helper and make breadcrumbs navigable

gyan9427:
**Summary**

Extracted the sidebar loader into helper.ts to import and use it. This centralizes try/catch logic for loading optional custom API sidebars and ensures the category link logic falls back to the correct doc or generated-index, which fixes navigation so API category breadcrumbs are navigable even when a custom sidebar is missing.

**What changed**

- Added helper.ts

- Updated sidebars to import and use the helper and still use releases from microsite/releases

- No structural changes to the sidebar items themselves — only the extraction of the loader and stable fallback behavior were made.

**Motivation**

- Centralizing the require + try/catch in a helper:
     - Removes duplicated patterns and makes behavior explicit and testable.
     - Guarantees the fallback of an empty array on load failure, so the category link logic in sidebars.ts 
       always receives a consistent value.
- This stabilizes the Docusaurus category link selection so breadcrumb links for API categories correctly point to either the generated index page (when items exist) or the 404 doc fallback — making breadcrumb navigation reliable.

### DIFF
--- a/microsite/sidebars.ts
+++ b/microsite/sidebars.ts
@@ -54,7 +54,6 @@ export default {
       'overview/threat-model',
       'overview/versioning-policy',
     ]),
-    ,
     makeGeneratedIndexCategory('Getting Started', 'getting-started', [
       'getting-started/index',
       makeGeneratedIndexCategory(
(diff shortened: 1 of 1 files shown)


gyan9427:
@Parsifal-M I have done the changes as per requirements to make the breadcrumb navigable

microsite/helper.ts
aramissennyeydd, line 1: Let's keep this in the other file.

microsite/sidebars.ts
```diff
@@ -19,663 +12,894 @@ const scaffolderSidebar = tryToLoadCustomSidebar(
...
+      link: { type: 'generated-index', title: 'Overview', slug: '/overview' },
+      items: [
+        'overview/what-is-backstage',
+        'overview/technical-overview',
+        'overview/architecture-overview',
+        'overview/roadmap',
+        'overview/threat-model',
+        'overview/versioning-policy',
+      ],
+    },
```
aramissennyeydd: Can we move most of this repetitive code to a helper? Similar idea to https://github.com/backstage/backstage/pull/31508#issuecomment-3437213868

microsite/helper.ts
gyan9427: you want this to be in sidebar.ts itself?

aramissennyeydd: yup!

PR #31907, opened 2025-11-27
feat(ui): Add Table row selection support with visual states and documentation

Naycon:

This PR adds proper row selection support to the Table component, including visual state styling, checkbox rendering fixes, and comprehensive documentation.

### Changes

**Fixes:**
- Fixed checkbox rendering to only appear for multi-select toggle mode (previously showed for any toggle mode)
- Fixed Column component to properly merge the `className` prop
- Added fixed 40px width for selection column to prevent expansion in ResizableTableContainer

**Features:**
- Added row state styling for hover, selected, pressed, and disabled states
- Switched from React Aria headless Checkbox to styled Backstage UI Checkbox
- Added 12 Storybook stories covering all selection mode/behavior combinations
- Added interactive playground stories for selection mode and behavior

**Documentation:**
- Documented row selection in docs-ui with three sections: Selection mode, Selection behavior, and With row actions
- Added interactive examples that let readers toggle between options

### Screenshots

**Single select**

image

**Select multiple**

image

#### Changeset illustration

image

### DIFF
--- a/packages/ui/src/components/Table/Table.stories.tsx
+++ b/packages/ui/src/components/Table/Table.stories.tsx
@@ -16,6 +16,7 @@
 
 import { useState } from 'react';
 import type { Meta, StoryFn, StoryObj } from '@storybook/react-vite';
+import { type Selection } from 'react-aria-components';
 import {
   Table,
   TableHeader,
@@ -26,6 +27,8 @@ import {
   CellProfile as CellProfileBUI,
   useTable,
 } from '.';
+import { RadioGroup, Radio } from '../RadioGroup';
+import { Flex } from '../Flex';
 import { MemoryRouter } from 'react-router-dom';
 import { data as data1Raw } from './mocked-data1';
 import { data as data2 } from './mocked-data2';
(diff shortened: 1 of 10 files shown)


packages/ui/src/components/Table/Table.stories.tsx
```diff
@@ -367,3 +371,510 @@ export const CellProfile: Story = {
...
+    );
+  },
+};
+
+export const SelectionMultiToggle: Story = {
+  render: () => {
+    const [selectedKeys, setSelectedKeys] = useState<Selection>(new Set([]));
+
+    return (
+      <ResizableTableContainer>
```
cdedreuille: Why do we need `ResizableTableContainer` ? Is this just for the example or part of the recommendation?

cdedreuille (approved):
Very small comment but overall this looks really good. Excited to get that update in.

packages/ui/src/components/Table/Table.stories.tsx
Naycon: Good catch! It's not needed, I was just testing that the resizing didn't apply when wrapped in a `ResizableTableContainer`, as it's currently the only way to make the table responsive. There's a bunch of props set on `Column` to make sure the checkbox column doesn't expand and become too wide.

Removing this here though, it's not needed.

Issue #31684, opened 2025-11-07
MCP: Errors thrown in MCP tools / actions get swallowed


johnmcollier:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
"mcp", "error"
```

### 🗃️ Project Area

Core Framework

### 🔖 Need

My team and I have written some MCP tools for Backstage (https://github.com/redhat-ai-dev/community-plugins/tree/main/workspaces/mcp-integrations), and we had a couple tools where we wanted a thrown error shown / displayed back to the user (e.g. input validation we couldn't enforce with just the zod schema, or some error was encountered in a function we called)

However, we found that any errors that were thrown were swallowed by the MCP server, and the server returned a generic 500 error:

```
{"error":"MCP error -32603: Request failed with 500 Internal Server Error"}
```

But, if I look in the Backstage logs, I can see the full error message

```
[backend]: 2025-11-07T22:00:43.313Z software-catalog-mcp-tool error Request failed with status 500 Failed execution of action "software-catalog-mcp-tool:fetch-catalog-entities"; caused by Error: This is a test error message
```

<img width="362" height="293" alt="Image" src="https://github.com/user-attachments/assets/191947d5-9777-4413-8b4f-02f2aacc76ef" />

### 📝 Proposal

Instead of swallowing the error, it'd be nice if we could include the full error message in the response back to the user. If this would be an acceptable change, I'd be happy to take this on

### 🔄 Alternatives

To work around this, we just added an error message field to the output zod schema, and we catch any thrown errors and set the error field to the error that was caught. Most models were pretty good about interpreting the errorMsg being set that the request failed for some reason.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Yes, but I would like some more guidance

johnmcollier:
In `handleError.ts`, we only return the error message when it's a known Backstage error:

https://github.com/backstage/backstage/blob/fd502b39821d984f37f3c5630ad980596d257e62/plugins/mcp-actions-backend/src/services/handleErrors.ts#L49-L67

Could we change it so that `describeError` still returns back the original error message if the error isn't a known Backstage error?

camilaibs:
[...]

PR #31911, opened 2025-11-27
chore(mcp): Handle all Backstage errors, both built-in and custom

johnmcollier:

Follow up PR to https://github.com/backstage/backstage/pull/31847

This PR updates `handleErrors.ts` in `mcp-actions-backend` to handle any custom Backstage error, rather than just the built-in errors in `@backstage/errors`. This will be useful for any MCP tools that interact with plugins / custom integrations that have their own error set outside of the standard errors provided.

I've also added a couple basic test cases to cover the logic in `handleErrors.ts`

### DIFF
--- a/.changeset/public-animals-find.md
+++ b/.changeset/public-animals-find.md
@@ -2,4 +2,4 @@
 '@backstage/errors': minor
 ---
 
-Adds new function for validating if a given error extends CustomErrorBase, when instanceof can't be used
+Adds new function for validating if a given error extends `CustomErrorBase`, when `instanceof` can't be used


johnmcollier:
Example of custom error being thrown in an MCP action: 

Screenshot 2025-11-27 at 1 29 15 PM

plugins/mcp-actions-backend/src/services/handleErrors.ts
```diff
@@ -59,7 +49,9 @@ function describeError(err: unknown): string {
 
     const { name, message } = extractCause(serialized);
 
-    if (knownErrors.has(name)) {
+    // All Backstage errors except ResponseError extends CustomErrorBase, so we need
+    // to explicitly check both CustomErrorBase and ResponseError
+    if (err instanceof CustomErrorBase || err instanceof ResponseError) {
```
benjdlambert: The `instanceof` check here is a little bit fragile for things that aren't global singletons. The `CustomErrorBase` and `ResponseErrors` are unfortunately not global singletons as they come from the potentially different versions of `@backstage/errors` package, which is why we normally use the name.

I think we're going to need a different way to work out if this is an error that's come from either of these two.

@Rugvip @freben do you have any alternative ways here to check that it's a Backstage error, or if this is even the right path to be heading down? I'm a little torn tbh.

johnmcollier: @benjdlambert Hmm, good point, and a nuance I hadn't considered.

Perhaps one option could be adding back the known list of Backstage errors to check by name, and also adding a sort of unique symbol on the CustomErrorBase class that we can reliably check for? 

It feels a little clunky, and that would mean custom errors in packages that used older versions of `@backstage/errors` wouldn't be supported here, but it would allow custom errors in packages that use the version of `@backstage/errors` with the symbol to be supported, while also preserving the name checking that works regardless of version, without breaking anything that's supported currently.

vinzscam: also usually if you make sure that your error contains `status` or `statusCode` property then this is enough for the error to be handled correctly

freben: Adding something like `$$type: @backstage/CustomError` or so (similar to what we do for other things in the frontend and backend systems) could be useful - but I'm not sure that I like the semantics of it too much.

The big question for me is what it is that's so special about our errors that we need to have an entire separate path for them. Why is the code going through this effort at all, to distinguish between them? I don't think that's what anybody should want to do - the base class is just there to make it easier to make "oneliner" custom errors without having to think about how to properly capture stack frames etc. It's NOT to end up with a result of some certain shape or behavior. After construction they should be mostly like "just any error".

The code in question just generates `name: message`. Otherwise it throws. Why does it ever throw? Why doesn't it always just do `stringifyError`? What's the value of throwing?

johnmcollier: @freben Initially I proposed doing that, but there was concern raised [here](https://github.com/backstage/backstage/issues/31684#issuecomment-3541854024) about prompt injection / credentials leaking if any error was allowed

johnmcollier: @benjdlambert @freben any suggestions on how to proceed here? I'm happy to go with @freben's suggestion and just output any `stringifyError`'d error to simplify things, but if there's still concerns about prompt injection (as alluded to previously) , it may not be the best course of action.

(although with ForwardedError being supported, I wonder if the current code in `handleErrors.ts` would even prevent that)

freben: Sorry we haven't given any guidance yet, we talked a bit about it this morning and still not entirely sure what's best.

* Adding a `$$type` to the base class is probably OK here. The only thing is, it's the first time ever then that we put a `$$type` on a BASE thing - all of the other usages are outputs of the framework builder functions that aren't really made for overriding. And then you can't use that pattern for your subclass, should you want to :)
* I pondered for a bit whether we could leverage the same mechanics as the `packages/version-bridge`, where it's ensured that a global singleton base class gets used _even if_ you happen to have mixtures of package versions. But if we introduce that, we'll still be in a transition period for a bit where old packages didn't yet have that feature, so we'd need to keep both the old and new worlds for a bit.

So ... we've been pushing this for a bit trying to figure out what's best. Other ideas welcome too of course

freben: A variant of the `$$type` is to make a very specific hidden symbol that's a non-enumerable property set by the constructor - probably coupled with an `isCustomError` checker. That may still have the drawback of not working for all versions, but can still be a good middle ground

johnmcollier: > A variant of the $$type is to make a very specific hidden symbol that's a non-enumerable property set by the constructor - probably coupled with an isCustomError checker. That may still have the drawback of not working for all versions, but can still be a good middle ground

@freben I like that approach. Thoughts on this [new commit](https://github.com/backstage/backstage/pull/31911/commits/b1355b9e9a5671866ca9d5bcd66ebec9780d69a3)? 

It adds in the symbol (`@backstage/errors#CustomErrorBase`), and the `isCustomError` check, and reduces the changes in `handleErrors.ts` to just add in the call for `isCustomError`, which keeps the existing logic backwards compatible with all versions of `@backstage/errors` (though of course the new `isCustomError` check would only return true for custom errors using the new version)

TBH, I'm not sure how we'd add support for all `CustomErrorBase` errors and work for _all_ versions of errors, given the limitations available. But, if something does come up (or if there's a better alternative I'm not considering), I'm happy to use it instead.

plugins/mcp-actions-backend/src/services/handleErrors.ts
```diff
@@ -59,7 +60,9 @@ function describeError(err: unknown): string {
 
     const { name, message } = extractCause(serialized);
 
-    if (knownErrors.has(name)) {
+    // All Backstage errors except ResponseError extends CustomErrorBase, so we need
+    // to explicitly check both CustomErrorBase and ResponseError
+    if (knownErrors.has(name) || isCustomError(err)) {
```
vinzscam, line 65: could relaxing these rules work? 🤔 
```suggestion
    if (name && message) {
```

johnmcollier: Good question. So that would presumably work for any error with a message, but it was raised [here](https://github.com/backstage/backstage/issues/31684#issuecomment-3541854024) that they didn't want to surface _all_ error message because of concerns about potential prompt injection of PII leakage.

.changeset/public-animals-find.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/errors': minor
+---
+
+Adds new function for validating if a given error extends CustomErrorBase, when instanceof can't be used
```
freben:
```suggestion
Adds new function for validating if a given error extends `CustomErrorBase`, when `instanceof` can't be used
```

packages/errors/src/errors/CustomErrorBase.ts
```diff
@@ -63,6 +65,24 @@ export class CustomErrorBase extends Error {
...
     this.cause = isError(cause) ? cause : undefined;
   }
 }
+
+/**
+ * Returns true if the given error type extends CustomErrorBase
+ *
+ * @public
+ */
+export function isCustomError(err: unknown): err is CustomErrorBase {
```
Rugvip, line 84: Reiterating my thoughts from the framework SIG: I don't perceive there to by any legitimate reason to want this check. The `CustomErrorBase` is really just a utility to help create subclasses of `Error`, because it's a bit tricky to do that directly (`extends Error` doesn't behave as expected). For all intents and purposes I find this to essentially be equivalent of `instanceof Error`, but it happens to only be caring about our particular error instances.

I don't think the way the error is constructed to have any impact on application logic, and as such I don't think this is something we want to add.

plugins/mcp-actions-backend/src/services/handleErrors.ts
Rugvip, line 41: Bit unrelated, but I could see a variant of this being useful in `@backstage/errors`. Something that unwraps `ResponseError` and `ForwardedError` or potentially more generally walks up an error `.cause` stack

plugins/mcp-actions-backend/src/services/handleErrors.ts
Rugvip: Yep, there's the example of a SQL error that may contain too much information, but worst case there's an error where the message contains input from another actor.

I don't think filtering based on known Backstage errors really solves this problem though. I think the assumption that errors based on `@backstage/errors` are less likely to contain these vulnerabilities could be true, but I don't think that's something that is guaranteed to remain true over time. Backstage plugins are able to implement vulnerable error handling just as easily as any other piece of code.

Bit stuck with where to actually go with this from here though. One potential path is to simply move responsibility of this to action implementation, saying that all actions should be careful in how they throw errors in order to avoid injection attacks. There's some other extreme as well where we introduce something like a `UserFacingError` or wrapper to that effect, that must be used for any error that should be forwarded to the user, and is otherwise replaced with a generic error.

There's also an angle where user-facing error responses have to be intentional, such as returning an `{ error: ... }` response rather than throwing. It's very likely that a lot of `try { ... } catch (error) { return { error } }` shows up as a result of that though...

In the end I have to say that I think this is a tricky problem that needs to be solved at the source: Any code that throws an error with uncontrolled user input should be considered at least a moderate vulnerability. Such code must simply cease to exist, because it will always open up for vulnerabilities in consuming systems in a world where LLMs are prevalent. We're protecting against errors thrown by actions here in the MCP layer, but that action could also be consumed via other avenues that exposes errors to an agent.

I think my preferred way forward is to close this PR and potentially relax the `knownErrors.has(name)` check, and instead add guidance in https://backstage.io/docs/backend-system/core-services/actions-registry/ to ensure controlled throwing of errors.

Rugvip (commented):
Alright, summarized my thoughts following the discussion in the framework SIG. Will also post in https://github.com/backstage/backstage/issues/31684#issuecomment-3541854024 and reopen, as I think it's perhaps best to keep the discussion there.

As a bit of a side note: does this actually work? 🤔 I'd expect pretty much all action errors to just be `ResponseError`s with a forwarded but deserialized cause, and as such the root error would always pass the `isCustomError`, while the cause would never pass it due to it having been serialized and lacking the symbol.

plugins/mcp-actions-backend/src/services/handleErrors.ts
johnmcollier: > I think my preferred way forward is to close this PR and potentially relax the knownErrors.has(name) check, and instead add guidance in https://backstage.io/docs/backend-system/core-services/actions-registry/ to ensure controlled throwing of errors.

I'm fine with moving the responsibility to the action implementation. In terms of relaxing the `knownErrors` check, do you think what @vinzscam proposed above would be sufficient?

Issue #31921, opened 2025-12-01
Notifications Email Module Fails to use SES options after SES v2 update


coltMcKissick:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
notifications,ses,sourceARN,fromARN
```

### 🗃️ Project Area

Notifications

### 🔗 External Integration

AWS

### 📝 Description & Context

Updating the notifications email module to use version 2 of the AWS SES SDK (https://github.com/backstage/backstage/pull/31458) changed the required inputs necessary to use a SES identity across AWS accounts. Previously the `sourceArn` and `fromArn` attributes were passed as options to the `SendRawEmailCommand`, but in version 2 the command as well as the available options changed.

In version 2, the [available inputs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-sesv2/Interface/SendEmailCommandInput/) for the send email command do not include `sourceArn` and `fromArn`, resulting in the current `sesConfig.sourceArn` and `sesConfig.fromArn` to have no effect.

I have verified the issue within our environment and plan to open a PR with a proposed fix.

### 👍 Expected Behavior

I expect to be able to provide configuration to the notification email module to allow for cross account SES identities.

### 🥾 Reproduction steps

1. set `sesConfig.sourceArn` and `sesConfig.fromArn` to an identity outside of the current AWS account.
2. create a notification
3. observe the logs for error messages about failing to send an email, should see that the identity referenced in the error message is using an ARN from within the current account rather than the ARN defined in the configuration.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes, and I have enough information to get started

vinzscam:
Thank you 🙏 Closed by #31922

PR #31922, opened 2025-12-01
fix: Update email processor ses options to match v2 of ses sdk

coltMcKissick:

Related issue: https://github.com/backstage/backstage/issues/31921

Updates the configuration options and getSesOptions function to reflect input changes for the new version of `SendEmailCommand`

### DIFF
--- a/plugins/notifications-backend-module-email/src/processor/NotificationsEmailProcessor.ts
+++ b/plugins/notifications-backend-module-email/src/processor/NotificationsEmailProcessor.ts
@@ -54,6 +54,7 @@ export class NotificationsEmailProcessor implements NotificationProcessor {
   private readonly sender: string;
   private readonly replyTo?: string;
   private readonly sesConfig?: Config;
+  private readonly sesOptions?: Partial<SendEmailCommandInput>;
   private readonly cacheTtl: number;
   private readonly concurrencyLimit: number;
   private readonly throttleInterval: number;
@@ -92,6 +93,7 @@ export class NotificationsEmailProcessor implements NotificationProcessor {
     this.sender = emailProcessorConfig.getString('sender');
     this.replyTo = emailProcessorConfig.getOptionalString('replyTo');
     this.sesConfig = emailProcessorConfig.getOptionalConfig('sesConfig');
+    this.sesOptions = this.getSesOptions();
     this.concurrencyLimit =
       emailProcessorConfig.getOptionalNumber('concurrencyLimit') ?? 2;
     this.throttleInterval = emailProcessorConfig.has('throttleInterval')
@@ -308,9 +310,7 @@ export class NotificationsEmailProcessor implements NotificationProcessor {
     return contentParts.join('\n\n');
   }
 
-  private async getSesOptions(): Promise<
-    Partial<SendEmailCommandInput> | undefined
-  > {
+  private getSesOptions(): Partial<SendEmailCommandInput> | undefined {
     if (!this.sesConfig) {
       return undefined;
     }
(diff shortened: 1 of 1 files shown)


.changeset/fancy-wasps-check.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-notifications-backend-module-email': minor
+---
+
+Changes ses configuration keys to match new configuration options in SES SDK V2
```
freben:
```suggestion
Changes SES configuration keys to match new configuration options in SES SDK V2
```

plugins/notifications-backend-module-email/config.d.ts
```diff
@@ -136,14 +136,10 @@ export interface Config {
...
         sesConfig?: {
-          /**
-           * ARN of the identity to use as the source of the email
-           */
-          sourceArn?: string;
           /**
            * ARN of the identity to use for the "From"/sender address of the email
            */
-          fromArn?: string;
+          fromEmailAddressIdentityArn?: string;
```
freben: This is a pretty hard change - is there no way that you can just leave the old fields in place and mark them as `@deprecated`, and still obey them if they are set? I would like to give users a little bit of time to adapt, instead of suddenly breaking on update.

freben: If you do that, you can lower the changeset to `patch` level, but do write a short set of instructions in there about  how people are meant to update their configs. Because that text ends up in the changeset and will be read by those who are upgrading - it's essential information to them.

.changeset/fancy-wasps-check.md
coltMcKissick: Hey I made a new changeset based on the other comment, let me know if it can be better.

plugins/notifications-backend-module-email/config.d.ts
coltMcKissick: Agree that is a better approach, updated the config and code to handle `fromArn` if it is set. I did not keep `sourceArn` as it seems to not have an equivalent in the v2 ses send email command options.

Let me know if I could still approach it better, made a new changeset as well.

freben: My general take on things is that it's usually not particularly useful to try to literally match app-config to underlying libraries. The app-config is the interface, and should stay the interface, unless there's strong reason to change it, even if the underlying libraries shift.

With that in mind, was `fromArn` perfectly serviceable as-is? Yes the library clarified the name but there's no strong reason for us to do that too right? If the semantics are effectively the same anyway.

If the other parameter was entirely no longer supported, then yeah we can remove that. We could even hard-remove it from the config schema, but still read it in the backend code and log a warning that hey you have this field set but we're ignoring it.

Does that make sense?

freben: Then the instructions would also change accordingly, to say that config parameter so-and-so is no longer supported and you should remove it from your config files. But no mention of the other one since it effectively continues functioning the same from the users' point of view.

coltMcKissick: Yep makes sense and made some changes. Now fromArn maps to the new v2 option,sourceArn is removed from config schema, getSesOptions logs warning if sourceArn is still set. Updated changeset and readme to reflect that, but let me know if I am still missing something.

coltMcKissick:
> hey this is looking great and I was just about to merge - when I realised one little thing.
> 
> `getSesOptions` is called on every send. That's going to be a ton of logging for config that likely isn't changing. We should probably log once on startup or so! Maybe in the constructor after getting `this.sesConfig`?
> 
> Sorry, you're doing great and I appreciate it, but you know how it is when things are built incrementally, it happens :)

No problem, pushed up a change that adds a new sesOptions var to the class and populates it by calling getSesOptions in the constructor after sesConfig attempts to be set. Also noticed that getSesOptions didn't need to be async, so removed that.

drodil (approved):
Good catch! We don't have the custom arn in use so it has not affected our email sending. When I updated the plugin to use v2, I missed this one.

PR #31923, opened 2025-12-01
feat(core-components): adding adding internationalization (i18n) for `app.title`

its-mitesh-kumar:
## Description 
This PR adds internationalization (i18n) support for the app title across the entire Backstage application. All components that display or use `app.title` now support translations, allowing users to see the app title in their preferred language.

 ### Translation priority:
 1. Translated value for 'app.title' in the current language
 2. Fallback to app.title from config
 3. Default translation ('Backstage')

## UI after changes ( with [example translation](https://github.com/its-mitesh-kumar/backstage/pull/2) )

https://github.com/user-attachments/assets/63c2baef-e02e-4567-886c-a8ce5be27e32

### DIFF
--- a/.changeset/salty-lamps-move.md
+++ b/.changeset/salty-lamps-move.md
@@ -1,6 +1,13 @@
 ---
 '@backstage/core-components': patch
 '@backstage/core-app-api': patch
+'@backstage/plugin-auth': patch
+'@backstage/plugin-catalog-import': patch
+'@backstage/plugin-catalog-react': patch
+'@backstage/plugin-home': patch
+'@backstage/plugin-search-react': patch
+'@backstage/plugin-techdocs-module-addons-contrib': patch
+'@backstage/plugin-techdocs': patch
 ---
 
 Add internationalization support for app title with the `useAppTitle()` hook.
@@ -8,4 +15,4 @@ All components that display the app title now support translations via the fixed
 
 **Translation priority:** translation override → config value → default ('Backstage').
 
-**Updated components:**- Header, SignInPage, OAuthRequestDialog, SearchBar, ImportInfoCard, ConsentPage, TechDocsReaderPageHeader, CompanyLogo, UnregisterEntityDialog, and more.
+**Updated components:** Header, SignInPage, OAuthRequestDialog, SearchBar, ImportInfoCard, ConsentPage, TechDocsReaderPageHeader, CompanyLogo, UnregisterEntityDialog, and more.
--- a/.changeset/salty-lamps-move.md
+++ b/.changeset/salty-lamps-move.md
@@ -14,5 +14,3 @@ Add internationalization support for app title with the `useAppTitle()` hook.
 All components that display the app title now support translations via the fixed translation key `app.title`.
 
 **Translation priority:** translation override → config value → default ('Backstage').
-
-**Updated components:** Header, SignInPage, OAuthRequestDialog, SearchBar, ImportInfoCard, ConsentPage, TechDocsReaderPageHeader, CompanyLogo, UnregisterEntityDialog, and more.
--- a/.changeset/salty-lamps-move.md
+++ b/.changeset/salty-lamps-move.md
@@ -1,0 +1,11 @@
+---
+'@backstage/core-components': patch
+'@backstage/core-app-api': patch
+---
+
+Add internationalization support for app title with the `useAppTitle()` hook.
+All components that display the app title now support translations via the fixed translation key `app.title`.
+
+**Translation priority:** translation override → config value → default ('Backstage').
+
+**Updated components:**- Header, SignInPage, OAuthRequestDialog, SearchBar, ImportInfoCard, ConsentPage, TechDocsReaderPageHeader, CompanyLogo, UnregisterEntityDialog, and more.
--- a/packages/core-components/src/hooks/useAppTitle.test.tsx
+++ b/packages/core-components/src/hooks/useAppTitle.test.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 The Backstage Authors
+ * Copyright 2025 The Backstage Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/packages/core-components/src/hooks/useAppTitle.test.tsx
+++ b/packages/core-components/src/hooks/useAppTitle.test.tsx
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ReactNode } from 'react';
+import { TestApiProvider } from '@backstage/test-utils';
+import { renderHook, waitFor } from '@testing-library/react';
+import { ConfigReader } from '@backstage/core-app-api';
+import { configApiRef } from '@backstage/core-plugin-api';
+import {
+  translationApiRef,
+  TranslationApi,
+} from '@backstage/core-plugin-api/alpha';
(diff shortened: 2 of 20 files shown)


packages/core-components/src/hooks/useAppTitle.ts
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright 2024 The Backstage Authors
```
logonoff:
```suggestion
 * Copyright 2025 The Backstage Authors
```

packages/core-components/src/hooks/useAppTitle.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Hook that returns the app title with translation support.
+ *
+ * This hook automatically uses translated values for the app title based on the
+ * current language. It uses the fixed translation key 'app.title'.
+ *
+ * Translation priority:
+ * 1. Translated value for 'app.title' in the current language
+ * 2. Fallback to app.title from config
```
logonoff, line 29: Just wondering if "fallback" implies there is a desire to deprecate `app.title`. There is some documentation surrounding that config (e.g., https://backstage.io/docs/conf/writing) so we should update it if needed

packages/core-components/src/hooks/useAppTitle.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+export function useAppTitle(): string {
+  const configApi = useApi(configApiRef);
+  const { t } = useTranslationRef(coreComponentsTranslationRef);
+
+  // Default value defined in coreComponentsTranslationRef
+  const defaultTitle = 'Backstage';
+
+  // Try to get translation for the current language
+  const translatedTitle = t('app.title', {});
```
logonoff:
```suggestion
  const translatedTitle = t('app.title');
```

packages/core-components/src/hooks/useAppTitle.ts
its-mitesh-kumar: @logonoff No deprecation intended! The `app.title` config option remains the primary and recommended way to set your app's title.
The term "fallback" here is just describing the resolution priority order when i18n is enabled:

1. If a user has explicitly provided a translation for app.title in their language, use that
2. Otherwise, use the value from app.title config (this is still the standard approach)
3. If neither is set, default to "Backstage"

Translations are an optional enhancement for teams that need multi-language support — they don't replace the existing config-based approach.

ciiay (commented):
Hey @its-mitesh-kumar , the logic looks good to me. Just want to make sure whether we want to translate the user defined app title 🤔 I thought for terms especially user input terms it's common to keep it as it is.

@awanlin any preferences here?

its-mitesh-kumar:
> Hey @its-mitesh-kumar , the logic looks good to me. Just want to make sure whether we want to translate the user defined app title 🤔 I thought for terms especially user input terms it's common to keep it as it is.
> 
> @awanlin any preferences here?


@ciiay Thanks for the review! To clarify, we're not forcing users to translate their app.title - we're simply extending the functionality to give users the option to localize it if they want.

ciiay (approved):
/lgtm

Thanks for the PR 🤝

its-mitesh-kumar:
@Rugvip @schultzp2020 Please review the PR.

its-mitesh-kumar:
@Rugvip @schultzp2020 @benjdlambert Please review the PR.

packages/core-components/src/hooks/useAppTitle.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @public
+ * @example
+ * ```tsx
+ * const MyComponent = () => {
+ *   const appTitle = useAppTitle();
+ *   return <h1>{appTitle}</h1>;
+ * };
+ * ```
+ */
+export function useAppTitle(): string {
```
schultzp2020, line 48: Create you create some tests for the hook? i.e. 
- Translation override takes precedence
- Falls back to config value when no translation override
- Falls back to default 'Backstage' when neither is set

packages/core-components/src/hooks/useAppTitle.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  // Try to get translation for the current language
+  const translatedTitle = t('app.title');
+
+  // Get the fallback title from config
+  const configTitle = configApi.getOptionalString('app.title');
+
+  // Priority: translation override → config value → default
+  // If translation is different from default, it means an override exists
+  if (translatedTitle !== defaultTitle) {
```
schultzp2020, line 63: If a user explicitly translates their app title to "Backstage" (the same as default), the logic will incorrectly skip the translation and use the config value instead. Do we want to document this?

.changeset/salty-lamps-move.md
```diff
@@ -0,0 +1,18 @@
...
+'@backstage/plugin-techdocs-module-addons-contrib': patch
+'@backstage/plugin-techdocs': patch
+---
+
+Add internationalization support for app title with the `useAppTitle()` hook.
+All components that display the app title now support translations via the fixed translation key `app.title`.
+
+**Translation priority:** translation override → config value → default ('Backstage').
+
+**Updated components:** Header, SignInPage, OAuthRequestDialog, SearchBar, ImportInfoCard, ConsentPage, TechDocsReaderPageHeader, CompanyLogo, UnregisterEntityDialog, and more.
```
schultzp2020:
```suggestion
```

Since the functionality doesn't change for the user, we can probably remove this section.

packages/core-components/src/hooks/useAppTitle.ts
its-mitesh-kumar: @schultzp2020 Thanks for the review, I have added the tests for it. PTAL.

packages/core-components/src/hooks/useAppTitle.ts
its-mitesh-kumar: @schultzp2020 Good catch. I have added description for it.

.changeset/salty-lamps-move.md
its-mitesh-kumar: @schultzp2020 Yes it make sense. I have removed it.

its-mitesh-kumar:
[schultzp2020](https://github.com/schultzp2020) Thanks for reviewing the PR. I have addressed all your comments, please have a look.

its-mitesh-kumar:
@Rugvip @schultzp2020 Please check if we can merge this PR ?

packages/core-components/src/hooks/useAppTitle.ts
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2024 The Backstage Authors
```
schultzp2020:
```suggestion
 * Copyright 2025 The Backstage Authors
```

This file was authored in 2025.

schultzp2020 (commented):
Thanks for the contribution, @its-mitesh-kumar!

### Default title coupling

The hook detects translation overrides by comparing against a hardcoded `'Backstage'` string that duplicates the default in `coreComponentsTranslationRef`. If the default message in the translation ref ever changes, this constant would silently drift. Consider exporting a `DEFAULT_APP_TITLE` constant from `translation.ts` and using it in both the translation ref messages and the hook, so they stay in sync automatically.

### Minor fixes

It looks like a couple of @logonoff's earlier suggestions (copyright year, unnecessary `t()` argument) were accidentally overwritten during a rebase or push. I've re-added them as inline suggestions below.

Everything else looks great.

packages/core-components/src/hooks/useAppTitle.test.tsx
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2024 The Backstage Authors
```
schultzp2020:
```suggestion
 * Copyright 2025 The Backstage Authors
```

Same copyright year fix as the implementation file.

packages/core-components/src/hooks/useAppTitle.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+export function useAppTitle(): string {
+  const configApi = useApi(configApiRef);
+  const { t } = useTranslationRef(coreComponentsTranslationRef);
+
+  // Default value defined in coreComponentsTranslationRef
+  const defaultTitle = 'Backstage';
+
+  // Try to get translation for the current language
+  const translatedTitle = t('app.title', {});
```
schultzp2020:
```suggestion
  const translatedTitle = t('app.title');
```

The empty object argument is unnecessary since the `app.title` message has no interpolation parameters.

its-mitesh-kumar:
@Rugvip Please reopen the PR.

its-mitesh-kumar:
@Rugvip Please reopen the PR.

PR #31929, opened 2025-12-02
Add auto commitAction to pushRepo action

joaopedromgoulart:

Previously, the gitlabPushRepo action in the GitLab scaffolder module only supported create, update, or delete for the commitAction input. This limitation prevented complex pushes, such as simultaneously creating new files and updating existing ones. To resolve this, a new auto option has been introduced (similar to the existing functionality in the Merge Request action). When set to auto, the action dynamically determines the appropriate operation for each file—creating it if it doesn't exist, or updating it if it does.

### DIFF
--- a/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabRepoPush.ts
+++ b/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabRepoPush.ts
@@ -146,36 +146,33 @@ export const createGitlabRepoPushAction = (options: {
         }
       }
 
-      const actions: CommitAction[] = (
-        (
-          await (async () => {
-            const results = [];
-            for (const file of fileContents) {
-              const action = await getFileAction(
-                { file, targetPath },
-                { repoID, branch: branchName },
-                api,
-                ctx.logger,
-                remoteFiles,
-                ctx.input.commitAction,
-              );
-              results.push({ file, action });
-            }
-            return results;
-          })()
-        ).filter(o => o.action !== 'skip') as {
-          file: SerializedFile;
-          action: CommitAction['action'];
-        }[]
-      ).map(({ file, action }) => ({
-        action,
-        filePath: targetPath
-          ? path.posix.join(targetPath, file.path)
-          : file.path,
-        encoding: 'base64',
-        content: file.content.toString('base64'),
-        execute_filemode: file.executable,
-      }));
+      const fileActionMap: {
+        file: SerializedFile;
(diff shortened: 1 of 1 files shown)


plugins/scaffolder-backend-module-gitlab/src/actions/gitlabRepoPush.ts
```diff
@@ -126,8 +129,44 @@ export const createGitlabRepoPushAction = (options: {
...
+              e,
+            )}`,
+          );
+        }
+      }
+
+      const actions: CommitAction[] = (
+        (
+          await Promise.all(
+            fileContents.map(async file => {
```
benjdlambert: Im a little concerned that this could throw a lot of requests of to gitlab in parallel. Should we move this to do something like:

```ts

for (const file of fileContents) {
   const action = await getFileAction(...)
}
```

Instead so that it least it does one at a time?

Or use something like `p-limit` to make sure that we only send of a few in parallel rather than for every file in the workspace?

joaopedromgoulart: Sure, you have a point! There is no need to verify all files in parallel, I've changed the code to check them sequentially.

.changeset/all-socks-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend-module-gitlab': minor
+---
+
+In the gitlabRepoPush action, add 'auto' possibility for commitAction input.
```
freben:
```suggestion
In the `gitlabRepoPush` action, add 'auto' possibility for `commitAction` input.
```

plugins/scaffolder-backend-module-gitlab/src/actions/gitlabRepoPush.ts
```diff
@@ -126,8 +129,46 @@ export const createGitlabRepoPushAction = (options: {
...
+            `Could not retrieve the list of files for ${repoID} (branch: ${branchName}) : ${getErrorMessage(
+              e,
+            )}`,
+          );
+        }
+      }
+
+      const actions: CommitAction[] = (
+        (
+          await (async () => {
```
freben: Why is this wrapped in a function?

joaopedromgoulart: It was to execute all the async calls inside the for loop, i've changed the strategy, breaking this code block in two more readable ones.

PR #31933, opened 2025-12-02
chore: upgrade jsdom to v27 and Jest to v30

Naycon:

Upgrade jsdom to v27+ to fix `Could not parse CSS stylesheet` warnings/errors 
when using packages with CSS modules and `@layer` declarations.

Jest upgraded to v30.2.0 which requires test assertion updates due to 
breaking changes in `expect.objectContaining()` behavior with arrays.

**Note:** `jest` is now a peer dependency. The CLI auto-detects the Jest 
version and uses the appropriate environment:
- **Jest 29**: Requires `jest-environment-jsdom` (no migration needed, but CSS parsing warnings remain)
- **Jest 30**: Requires `@jest/environment-jsdom-abstract` and `jsdom` (fixes CSS parsing warnings, requires migration)

See the [migration guide](https://backstage.io/docs/tutorials/jest30-migration) for details.

### Changes

**Peer dependency changes:**
- `jest` is now a peer dependency with support for v29 or v30
- CLI auto-detects Jest version and uses the appropriate environment
- Added custom JSDOM environment for Jest 30 with Web API global fixes (fetch, streams, Error, etc.)
- Added helpful error messages when required packages are missing
- Cross-fetch polyfill is only injected for Jest 29, as with Jest 30+ our patched Jest environment is used
- Network request blocker made MSW-compatible by checking if fetch was wrapped before blocking

**Production code changes for jsdom 27 testability:**
- `AppIdentityProxy`: extract `navigateToUrl` method for spying
- `LiveReloadAddon`: export `utils.reloadPage` for spying  
- `collect.ts`: export `internal.resolvePackagePath` for mocking

**MockFetchApi fix:**
- Evaluate `global.fetch` at call time instead of construction time, allowing MSW to patch fetch after MockFetchApi is constructed

**Test adaptations for jsdom 27:**
- Use RGB values instead of named colors in CSS assertions
- Update error format expectations (hyphenated type names, SyntaxError instead of FetchError for JSON parse errors)
- Simplify URL error assertions for cross-version compatibility
- Fix accessible name whitespace handling for external links
- Use `history.replaceState` for location mocking (non-configurable)
- Use `fireEvent.blur` for contentEditable elements
- Move async assertions inside `waitFor` for race conditions
- Remove `Blob.prototype.text` polyfill (now native in jsdom 27)
- Remove test case using credentials in `plugin://` URLs

**Test adaptations for Jest 30:**
- Replace `expect.objectContaining([...])` with direct array equality
- Replace `expect.objectContaining({ length: N })` with 
  `expect.any(Array)` + separate `toHaveLength()` assertions
- Use child process for native Node.js module resolution in
  `collect.test.ts` to work around Jest 30's resolver behavior
- Update snapshot headers for new Jest format

**MSW v2 migration:**
- Migrate `config-loader`, `events-node` and `kubernetes-react` tests from MSW v1 to v2
- Replace `rest` handlers with `http` handlers
- Update response format to use `HttpResponse.json()`
- Change `toStrictEqual` to `toEqual` for response assertions
- Required to fix compatibility issues with Jest 30 and jsdom v27

**Flaky test fix:**
- Fix race condition in catalog context menu test that failed in CI
  but passed locally, by adding explicit waits for entity load and 
  menu open states

Also removes the `jest-haste-map` patch which is no longer needed.

### DIFF
--- a/.changeset/eight-things-attack.md
+++ b/.changeset/eight-things-attack.md
@@ -4,7 +4,7 @@
 
 **BREAKING**: `jest` is now a peer dependency. If you run tests using Backstage CLI, you must add Jest and its environment dependencies as `devDependencies` in your project.
 
-You can choose to install either Jest 29 or Jest 30:
+You can choose to install either Jest 29 or Jest 30. The built-in Jest version before this change was Jest 29, however, we recommend that you switch to Jest 30. Upgrading will solve the `Could not parse CSS stylesheet` errors, allow you to use MSW v2 in web packages, and ensure that you remain compatible with future versions of the Backstage CLI. Support for Jest 29 is temporary, with the purpose of allowing you to upgrade at your own pace, but it will eventually be removed.
 
 - **Jest 29**: Install `jest@^29` and `jest-environment-jsdom@^29`. No migration needed, but you may see `Could not parse CSS stylesheet` warnings/errors when testing components from `@backstage/ui` or other packages using CSS `@layer` declarations.
 - **Jest 30**: Install `jest@^30`, `@jest/environment-jsdom-abstract@^30`, and `jsdom@^27`. Fixes the stylesheet parsing warnings/errors, but requires migration steps.
--- a/docs/tutorials/jest30-migration.md
+++ b/docs/tutorials/jest30-migration.md
@@ -6,7 +6,7 @@ description: A guide to migrating your project to Jest 30 and JSDOM 27
 
 Starting with a recent version of `@backstage/cli`, `jest` is a peer dependency. If you run tests using Backstage CLI, you must add Jest and its environment dependencies as `devDependencies` in your project.
 
-You can choose to install either Jest 29 or Jest 30:
+You can choose to install either Jest 29 or Jest 30. The built-in Jest version before this change was Jest 29, however, we recommend that you switch to Jest 30. Upgrading will solve the `Could not parse CSS stylesheet` errors, allow you to use MSW v2 in web packages, and ensure that you remain compatible with future versions of the Backstage CLI. Support for Jest 29 is temporary, with the purpose of allowing you to upgrade at your own pace, but it will eventually be removed.
 
 - **Jest 29**: Install `jest@^29` and `jest-environment-jsdom@^29`. No migration needed, but you may see `Could not parse CSS stylesheet` warnings/errors when testing components from `@backstage/ui` or other packages using CSS `@layer` declarations.
 - **Jest 30**: Install `jest@^30`, `@jest/environment-jsdom-abstract@^30`, and `jsdom@^27`. Fixes the stylesheet parsing warnings/errors, but requires the migration steps below.


freben:
Is there no way that we can make this environment "pluggable" (maybe even at a per package level) so that people don't have to bump all of them at once?

Naycon:
I tested this running a dist version of `@backstage/ui`, and it seems to have fixed the `@layer` warning we've seen in tests there. However, it did uncover a bug in a dependency on JSDOM when having `@media` following `&` selectors in `@layer` blocks. I made [a PR to fix that here](https://github.com/acemir/CSSOM/pull/66), but I'm not sure if we want to include this somehow (patch?) in Backstage until the fix is available in a release?

Naycon:
The fix to `@acemir/cssom` has now landed and is included in the updated `yarn.lock` in the PR. Also added a brief note about the error we saw with the previous version to the migration guide. Hopefully that means this is ready to go now. 🤞

packages/cli/package.json
```diff
@@ -201,6 +199,8 @@
     "esbuild-loader": "^4.0.0",
     "eslint-webpack-plugin": "^4.2.0",
     "fork-ts-checker-webpack-plugin": "^9.0.0",
+    "jest": "^29.0.0 || ^30.0.0",
+    "jest-environment-jsdom": "^29.0.0 || ^30.0.0",
```
Rugvip: Was thinking we could put a `*` here instead but it's probably good to have some form of indication of which versions we might support given the built-in config. Since we're doing this decoupling we may need some form of version policy down the line as well, but I think we can skip that for now.

packages/config-loader/src/schema/collect.test.ts
```diff
@@ -37,6 +32,36 @@ const mockSchema = {
 describe('collectConfigSchemas', () => {
   const mockDir = createMockDirectory();
 
+  // Jest 30's module resolver doesn't find nested node_modules in mock directories.
+  // Use a child process for native Node.js resolution instead.
+  const originalResolvePackagePath = internal.resolvePackagePath;
```
Rugvip, line 37: There's actually a test util for this 😁  https://github.com/backstage/backstage/blob/d871c5627acc58e5b2bd7f9b7c4465eb40665bc0/packages/backend-plugin-api/src/testUtils.ts#L49

packages/config-loader/package.json
```diff
@@ -57,7 +57,7 @@
     "@backstage/cli": "workspace:^",
     "@types/json-schema-merge-allof": "^0.6.0",
     "@types/minimist": "^1.2.5",
-    "msw": "^1.0.0",
+    "msw": "^2.0.0",
```
Rugvip, line 60: Hmm, is this needed, what were the compatibility issues? We've had issues with msw v2 and are holding off on migration for now. tl;dr is that it doesn't work well in web packages, https://mswjs.io/docs/migrations/1.x-to-2.x/#frequent-issues. This might put us in a spot where everyone needs to use v1 for web and v2 for node tests ._.

That's unless that issues happens to be fixed by this bump 😁, I'll check whether that's the case

Naycon: Yea, this was needed to fix failing tests. The root cause analysis, made with some AI help I'll admit:

> **Root cause**: MSW 1.x's `@mswjs/interceptors` patches `readyState` on request objects, which conflicts with Jest 30's property setter wrapping and JSDOM v26's new property descriptors, causing infinite recursion.

packages/config-loader/src/schema/collect.test.ts
Naycon: Nice! Would it be kosher to use that here though? As in, having a dependency on `backend-plugin-api` from `config-loader`?

Naycon: Should I move it to `backend-test-utils` maybe?

Naycon: Ah no, `backend-test-utils` won't work, since we need stuff from `paths` in `backend-plugin-api` and that package already depend on `backend-test-utils`.

Naycon: > There's actually a test util for this 😁

Looking at this more carefully, are these two really doing the same thing? My interpretation is that `collect.test` is testing that the logic inside `collect` is working correctly when resolving package paths, which (I assume) typically happens in a node environment. What the new test setup is doing is making sure resolution happens via native `node`, as with Jest 30 there seems to be a custom resolver used. I've assumed Jest < 30 used native `node` `resolve`.

`overridePackagePathResolution` on the other hand let's you mock what's returned from a call to resolve. We could use that here, but wouldn't we then need to take on keeping the mocked behavior in sync with actual behavior and risk bugs? As in, we would not test that we can handle what node actually returns but rather a mock that we believe reflects what it returns.

Happy to swap if you think this is an acceptable tradeoff, the commit I made just tried to keep things in sync with what they were before the Jest upgrade.

docs/tutorials/jest30-migration.md
```diff
@@ -0,0 +1,89 @@
+---
+id: jest30-migration
+title: Migrating to Jest 30
+description: A guide to migrating your project to Jest 30 and jsdom 27
```
Rugvip: jsdom in regular prose would be JSDOM

docs/tutorials/jest30-migration.md
```diff
@@ -0,0 +1,89 @@
...
+---
+
+Starting with a recent version of `@backstage/cli`, `jest` and `jest-environment-jsdom` are peer dependencies. If you run tests using Backstage CLI, you must add these packages as devDependencies in your project.
```
Rugvip: Wrap in a `codeBlock` to avoid this issue

docs/tutorials/jest30-migration.md
```diff
@@ -0,0 +1,89 @@
...
+  "resolutions": {
+    "jsdom": "^27.1.0",
+    "@types/jsdom": "^27.0.0"
+  }
+}
+```
+
+### Jest 30
+
+**Asymmetric matchers with arrays**: `expect.objectContaining()` no longer works with arrays.
```
Rugvip, line 23: Could reword this or add to .github/vale/config/vocabularies/Backstage/accept.txt

packages/config-loader/src/schema/collect.test.ts
Rugvip: Ah derp, didn't realize this wasn't the one from `backstage-plugin-api` and the new util just happened to have the same name 😅. Yeah can keep as is

packages/config-loader/package.json
Rugvip: Hmm, but are we actually putting users in a situation where they have to use MSW 2 for backend and MSW 1 for frontend? Feeling that'd need a little bit more digging to see if we could avoid that, because it's a bit 💩 :p

.changeset/eight-things-attack.md
```diff
@@ -0,0 +1,12 @@
+---
+'@backstage/cli': minor
+---
+
+**BREAKING**: `jest` and `jest-environment-jsdom` are now peer dependencies. If you run tests using Backstage CLI, you must add these packages as devDependencies in your project.
+
+You can choose to install either Jest 29 or Jest 30:
+
+- **Jest 29**: Install `jest@^29` and `jest-environment-jsdom@^29`. No migration needed, but you may see `Could not parse CSS stylesheet` warnings/errors when testing components from `@backstage/ui` or other packages using CSS `@layer` declarations.
+- **Jest 30**: Install `jest@^30` and `jest-environment-jsdom@^30`. Fixes the stylesheet parsing warnings/errors, but requires migration steps. You must also add jsdom to your `resolutions` to force version `^27.1.0`.
```
Rugvip: Hmm, this is kinda problematic, I don't think we should force all users to add a resolution unless it's some critical urgent fix. It looks like the way forward here may be to switch to `@jest/environment-jsdom-abstract`, which is currently on 27? see https://github.com/jestjs/jest/pull/15834. That or `@jest/environment-jsdom` just needs a bump to 27 too.

Naycon: Yea, this didn't feel too great, I agree. Can look into swapping to something based on `environment-jsdom-abstract`.

packages/config-loader/package.json
Naycon: This is a bit of a misunderstanding. There are still many backend packages using msw1, so upgrading to msw2 is no general requirement. Only two needed updating to solve CI only test failures.

docs/tutorials/jest30-migration.md
Naycon: Fixed.

docs/tutorials/jest30-migration.md
Naycon: Fixed.

docs/tutorials/jest30-migration.md
Naycon: Fixed.

packages/cli/package.json
Naycon: I've updated this with new versions now to better reflect what we need with the new solution with a custom jest env for jest 30+.

.changeset/eight-things-attack.md
Naycon: New approach based on `environment-jsdom-abstract` and jest version detection is now pushed. It required fixing some more tests and test setup-related things, but now specifying the peer dependencies is enough to get the right version and behavior.

Naycon:
@Rugvip I've updated the approach now, there's no more need for resolutions. Instead peer dependency + jest version detection takes care of setting things up correctly, and jsdom is a required dependency when using jest 30+. Mind taking a look again?

I backported the test running changes and mock api updates to a clean master checkout (as in running on jest 29 + jest-environment-jsdom) to see that everything worked there as well, and all tests passed locally.

.changeset/eight-things-attack.md
```diff
@@ -0,0 +1,12 @@
+---
+'@backstage/cli': minor
+---
+
+**BREAKING**: `jest` is now a peer dependency. If you run tests using Backstage CLI, you must add Jest and its environment dependencies as devDependencies in your project.
```
Naycon: Missed this one, fixing.

packages/cli/config/jestRejectNetworkRequests.js
```diff
@@ -22,7 +22,7 @@ const errorMessage = 'Network requests are not allowed in tests';
 const origHttpAgent = http.globalAgent;
 const origHttpsAgent = https.globalAgent;
 const origFetch = global.fetch;
-const origXMLHttpRequest = global.fetch;
+const origXMLHttpRequest = global.XMLHttpRequest;
```
Naycon, line 25: This seemed like a bug, so I included a fix here as well. Can revert if this was intentional for some reason (and maybe add a comment as to why we do this).

packages/core-app-api/src/apis/implementations/FetchApi/PluginProtocolResolverFetchMiddleware.test.ts
```diff
@@ -60,12 +60,6 @@ describe('PluginProtocolResolverFetchMiddleware', () => {
       'https://real.com:8080/base',
       'https://real.com:8080/base?c=d&e=f#g',
     ],
-    [
-      'plugin://username:password@x?c=d&e=f#g',
```
Naycon, line 64: This url format (with credentials) is not accepted anymore by browsers and node, so I removed it. The test won't pass with this in since we're running native node instead of node-fetch/cross-fetch.

packages/core-app-api/src/apis/implementations/FetchApi/PluginProtocolResolverFetchMiddleware.test.ts
```diff
@@ -106,14 +100,14 @@ describe('PluginProtocolResolverFetchMiddleware', () => {
...
+    const inputRequest = new Request('plugin://a', {
+      method: 'POST',
+      body: '123',
+    });
+
+    await outer(inputRequest);
 
     expect(inner.mock.calls[1][0]).toBe('https://elsewhere.com');
-    expect(inner.mock.calls[1][1].body).toEqual(Buffer.from('123', 'utf8'));
+    expect(inner.mock.calls[1][1]).toBe(inputRequest);
```
Naycon, line 111: Unsure if this is an okay fix or not. The issue is, the `inner.mock.calls[1][1].body` is now a `ReadableStream`, and I also think the internal `new Request(input ...)` in the `apply` method consumes this stream.

packages/frontend-dynamic-feature-loader/src/loader.test.tsx
```diff
@@ -655,7 +655,7 @@ describe('dynamicFrontendFeaturesLoader', () => {
 
     const errorCalls = mocks.console.error.mock.calls.flatMap(e => e[0]);
     expect(errorCalls).toEqual([
-      `Failed fetching module federation configuration of dynamic frontend plugins: FetchError: invalid json response body at http://localhost:7007/.backstage/dynamic-features/remotes reason: Unexpected end of JSON input`,
+      `Failed fetching module federation configuration of dynamic frontend plugins: SyntaxError: Unexpected end of JSON input`,
```
Naycon, line 658: This is due to the swap from JSDOM Error to node Error.

.changeset/eight-things-attack.md
```diff
@@ -0,0 +1,12 @@
+---
+'@backstage/cli': minor
+---
+
+**BREAKING**: `jest` is now a peer dependency. If you run tests using Backstage CLI, you must add Jest and its environment dependencies as `devDependencies` in your project.
+
+You can choose to install either Jest 29 or Jest 30:
```
Rugvip:
```suggestion
You can choose to install either Jest 29 or Jest 30. The built-in Jest version before this change was Jest 29, however, we recommend that you switch to Jest 30. Upgrading will solve the `Could not parse CSS stylesheet` errors, allow you to use MSW v2 in web packages, and ensure that you remain compatible with future versions of the Backstage CLI. Support for Jest 29 is temporary, with the purpose of allowing you to upgrade at your own pace, but it will eventually be removed.
```

I think it's worth nudging users more clearly towards Jest 30

docs/tutorials/jest30-migration.md
```diff
@@ -0,0 +1,76 @@
+---
+id: jest30-migration
+title: Migrating to Jest 30
+description: A guide to migrating your project to Jest 30 and JSDOM 27
+---
+
+Starting with a recent version of `@backstage/cli`, `jest` is a peer dependency. If you run tests using Backstage CLI, you must add Jest and its environment dependencies as `devDependencies` in your project.
+
+You can choose to install either Jest 29 or Jest 30:
```
Rugvip:
```suggestion
You can choose to install either Jest 29 or Jest 30. The built-in Jest version before this change was Jest 29, however, we recommend that you switch to Jest 30. Upgrading will solve the `Could not parse CSS stylesheet` errors, allow you to use MSW v2 in web packages, and ensure that you remain compatible with future versions of the Backstage CLI. Support for Jest 29 is temporary, with the purpose of allowing you to upgrade at your own pace, but it will eventually be removed.
```

Same as the changeset

packages/cli/config/jestRejectNetworkRequests.js
Rugvip: Yup 😁, let's keep it

packages/cli/config/jest-environment-jsdom/index.js
```diff
@@ -0,0 +1,61 @@
...
+const jsdom = require('jsdom');
+
+/**
+ * A custom JSDOM environment that extends the abstract base and applies
+ * fixes for Web API globals that are missing or incorrectly implemented
+ * in JSDOM.
+ *
+ * Based on https://github.com/mswjs/jest-fixed-jsdom
+ */
+class FixedJSDOMEnvironment extends JSDOMEnvironment {
```
Rugvip, line 27: Just tested MSW v2 in a frontend plugin, and this seems to have fixed that too, so we're unblocked from that upgrade as well! 🎉 (CC: @freben, @cdedreuille)

freben: NICE

plugins/catalog/src/alpha/pages.test.tsx
```diff
@@ -721,18 +721,27 @@ describe('Entity page', () => {
...
-        if (!disabled) {
-          await userEvent.click(screen.getByText('Test Title'));
-        }
 
-        expect(onClickMock).toHaveBeenCalledTimes(disabled ? 0 : 1);
-      });
+      // Wait for entity to load first
+      await waitFor(() =>
+        expect(screen.getByText(/artist-lookup/)).toBeInTheDocument(),
+      );
```
freben, line 728: Just as a general note - prefer

```ts
await expect(screen.findBy...).resolves.toBeInTheDocument()
```

over using `waitFor` - that even has a dedicated lint rule but we haven't enabled it.

packages/test-utils/src/testUtils/apis/FetchApi/MockFetchApi.ts
```diff
@@ -127,7 +127,9 @@ function baseImplementation(
 ): typeof crossFetch {
   const implementation = options?.baseImplementation;
   if (!implementation) {
-    return crossFetch;
+    // Return a wrapper that evaluates global.fetch at call time, not construction time.
+    // This allows MSW to patch global.fetch after MockFetchApi is constructed.
+    return (input, init) => global.fetch(input, init);
```
freben, line 132: 👍

packages/cli/src/modules/test/commands/package/test.ts
```diff
@@ -83,5 +83,23 @@ export default async (_opts: OptionValues, cmd: Command) => {
...
+  // forgot/didn't know they had to add jest themselves.
+  try {
+    require.resolve('jest');
+  } catch {
+    console.error(
+      [
+        'No Jest installation found in this project.',
+        '',
+        'To support opt-in migration to Jest v30, the Backstage CLI now expects Jest to be installed as a devDependency.',
+        'See the migration guide in the changelog for version options and their consequences.',
```
freben, line 98: This message hard codes the 30 version so to speak, and refers to a changelog instead of a fixed article URL. That's fine as a temporary message while we move things over but we may want to adjust this text to something more permanent once things stabilise

Naycon: Yea, fair point. I was envisioning this as something temporary that gets removed once Jest 29 support is dropped, as then there might not be any need for keeping jest, etc. as peer dependencies anymore. But that all depends on what they way forward looks like at that point ofc.

packages/cli/config/jest-environment-jsdom/index.js
cdedreuille: DREAM 💚

PR #31934, opened 2025-12-02
feat(scaffolder): added dry run functionality to template form previewer

rahdi:

- `TemplateFormPreviewer` is now wrapped in `DryRunProvider`, so the preview panel shares the same context as the rest of the editor.
- `TemplateFormPreviewerContent` introduces `handleDryRun`, which builds a full template spec, discovers local directories referenced by `fetch:template` steps, creates placeholder files, and runs `dryRun.execute`.
- `TemplateEditorForm` now accepts `onDryRun`, enabling template parameter forms to run dry runs with the provided values and surface success or error notifications via `alertApi`.

### DIFF
--- a/plugins/scaffolder/src/alpha/components/TemplateEditorPage/TemplateFormPreviewer.tsx
+++ b/plugins/scaffolder/src/alpha/components/TemplateEditorPage/TemplateFormPreviewer.tsx
@@ -40,6 +40,7 @@ import {
   TemplateEditorLayoutToolbar,
   TemplateEditorLayoutFiles,
   TemplateEditorLayoutPreview,
+  TemplateEditorLayoutConsole,
   TemplateEditorPanels,
 } from './TemplateEditorLayout';
 import { TemplateEditorToolbar } from './TemplateEditorToolbar';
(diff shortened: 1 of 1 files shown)


plugins/scaffolder/src/alpha/components/TemplateEditorPage/TemplateFormPreviewer.tsx
```diff
@@ -194,6 +200,113 @@ export const TemplateFormPreviewer = ({
...
+  name: template-preview
+  title: Template Preview
+spec:
+  type: service
+${templateYaml
+  .split('\n')
+  .map(line => `  ${line}`)
+  .join('\n')}`;
+
+        const urlPattern = /url:\s*['".]?\.\/([\w\-/.]+)['"']?/g;
```
benjdlambert, line 270: Hmm, so I think that this is required right because there's no real workspace when you don't load the templates from disk?

I wonder if there's a better way to do this, or just drop this entirely and assume that if you're trying to dry run templates then you'll need to re-write these URL's or something to relative paths from the original template.

krzysztofjanota:
As agreed we will wait with this PR till new version of Scaffolder will be introduced and then we shall see if this PR is still valid or not.

PR #31935, opened 2025-12-02
[NFS] set the default search type to all instead of software-catalog

binary-baum:
The default value for the Result Type in the search page is currently set to `Software Catalog`. As there are usually multiple result types present, the selection of `All` is more suitable to improve user experience (avaoiding the extra click to `All`). This PR removes the setting, so the code defaults to the `All` option.


alexvoedi:
push

plugins/search/src/alpha.tsx
```diff
@@ -152,7 +152,6 @@ export const searchPage = PageBlueprint.makeWithOverrides({
                     <Grid item xs={3}>
                       <SearchType.Accordion
                         name="Result Type"
-                        defaultValue="software-catalog"
```
awanlin, line 155: Perhaps the better approach here would be to make this configurable as that would allow adopters to adjust this as they see fit.

.changeset/three-rooms-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-search': minor
+---
+
+Set the default search type to `All` instead of `Software Catalog` in the search page.
```
awanlin: This PR only changes the default for the New Frontend System, we should probably mention that here. But I also think that we should consider changing this for the current frontend system as well.

awanlin (changes requested):
Thanks for the contribution @binary-baum, left a pair of comments. You'll also want to rebase to get the CI running and passing fully 👍

PR #31936, opened 2025-12-02
techdocs-addons-test-utils: major bump to remove explicit screen

freben:
In order to properly support moving to new versions of testing-library, this breaking change removes assertions from the `TechDocsAddonTester`. Please use `screen` instead; see the changeset for more details.


alexlorenzi:
👍 - Thanks
Also, strong changeset naming `.changeset/old-parks-smell.md‎`

PR #31946, opened 2025-12-02
cli: switch tsconfig to default to bundler resolution

Rugvip:

Seems the ecosystem has caught up and this was actually quite smooth now 😁

The main issue is that the `exports` field is now being picked up, so packages with incompatible setups would break. These mostly seem to have been fixed now, although switching to `text-extensions` was a way to get around bumping `textextensions` to the latest version because that was a bit tricky.

Adopters will also have the ability to use resolutions and patches if really needed.


.changeset/honest-bears-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-test-utils': patch
+---
+
+Switched `testextensions` dependency for `text-extensions`.
```
freben:
```suggestion
Switched `textextensions` dependency for `text-extensions`.
```

packages/core-app-api/report.api.md
```diff
@@ -20,6 +20,7 @@ import { AuthProviderInfo } from '@backstage/core-plugin-api';
 import { AuthRequestOptions } from '@backstage/core-plugin-api';
 import { BackstageIdentityApi } from '@backstage/core-plugin-api';
 import { BackstageIdentityResponse } from '@backstage/core-plugin-api';
+import { BackstageIdentityResponse as BackstageIdentityResponse_2 } from '@backstage/frontend-plugin-api';
```
freben, line 23: Hm. Is this really a good thing or should we be worried? Even if they happen to be duck-type-equal, it does seem a bit concering.

plugins/catalog-react/report.api.md
```diff
@@ -3,7 +3,8 @@
 > Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
 
 ```ts
-import { ApiRef } from '@backstage/core-plugin-api';
+import { ApiRef } from '@backstage/frontend-plugin-api';
+import { ApiRef as ApiRef_2 } from '@backstage/core-plugin-api';
```
freben, line 7: ...and even mixtures of the two?

packages/core-app-api/report.api.md
Rugvip: Not too worried. These are aliased in `core-plugin-api` so the update is arguably more correct. It's not a matter of duck-typed similarity, these are the exact same types just at varying resolution depth.

PR #31992, opened 2025-12-03
Make `.withOverrides` have a more simplified return type

freben:
Got a bit triggered by the diff in the API report for [this](https://github.com/backstage/backstage/pull/31938/files).


freben:
Also - this does have redundant changesets because the external types sort of did change. Let me know if I should remove before merging.

Issue #30617, opened 2025-07-22
Inefficient calls to GitLab API results in throttling


kc9ddi:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
catalog gitlab rate limit throttle project api
```

### 🗃️ Project Area

Core Framework

### 🔗 External Integration

GitLab

### 📝 Description & Context

Our organization has a few hundred GitLab repositories, many of which contain catalog-info.yaml manifests which we ingest via the GitLab plugin. Furthermore, we have one repository that hosts several hundred individual manifests. (We use GitLab.com, not self-hosted or dedicated GitLab). We're seeing that the catalog providers are failing to ingest all entities in our catalog, because GitLab APIs are returning HTTP-429 responses, to indicate that our use of the API is being rate-limited. 

The majority of issues that I'm seeing are in the API call to retrieve the GitLab Project ID for a given project (ie, [here](https://github.com/backstage/backstage/blob/4e4a5c7d35c74ce313f74bdcc557c75438ac9990/packages/backend-defaults/src/entrypoints/urlReader/lib/GitlabUrlReader.ts#L390) and [here](https://github.com/backstage/backstage/blob/8c48896e65fc6ad926ba2f102cd90017dfff150c/packages/integration/src/gitlab/core.ts#L153)).

The mapping between a project and its ID number should change infrequently. Perhaps this is a case where a simple cache can be used to reduce the number of calls made to this API. This would address both the rate limiting problem, as well as speed up the overall execution of the ingestion, since fewer API calls should be required.

### 👍 Expected Behavior

I would expect the GitLab integration to successfully process our catalog entities without being throttled by rate limits from the GitLab API.

Specifically, I _think_ that in this case, we can reduce the number of API calls that need to be made (and thus not need to deal with retry or rescheduling logic with regard to the API calls.)

### 🥾 Reproduction steps

1. Use the GitLab integration to ingest entities from GitLab.com for all projects under a given group. In our case, we have several hundred repositories which contain a small number of manifest files, as well as one repository that contains several hundred individual manifest files.

```
yarn backstage-cli info
OS:   Darwin 24.5.0 - darwin/arm64
node: v20.19.2
yarn: 4.4.1
cli:  0.32.1 (installed)
backstage:  1.38.1

Dependencies:
  @backstage/app-defaults                                          1.6.4
[...]

Issue #29428, opened 2025-03-31
🚀 Feature: Add monorepo support to GitlabDiscoveryEntityProvider


faeshaas:
### 🔖 Feature description

Add the possibility to recursively scan repo's for catalog-info.yaml files.

### 🎤 Context

Currently it only discovers catalog-info.yaml files in the root of a repository. In case of a monorepo multiple yamls can exist in multiple folders and levels. These yamls are currently not discovered by the GitlabDiscoveryEntityProvider.

### ✌️ Possible Implementation

Add a configuration option to enable recursive scanning of repo's so it will also work for monorepo. E.g. use the 'recursive' option of the '[list repository tree](https://docs.gitlab.com/api/repositories/#list-repository-tree)' endpoint of Gitlab.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

camilaibs:
Hello @pjungermann and @elaine-mattos :wave,
Would you have any recommendations for this 🙂 ?

pjungermann:
The API you have referenced above has a max page size of 100.

pjungermann:
While looking at the code, I've noticed

https://github.com/backstage/backstage/blob/master/plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts#L215

It seems that the current implementation uses a max page size of 50.

However, this could be changed to 100.
- https://docs.gitlab.com/api/rest/#offset-based-pagination (used here)
- https://docs.gitlab.com/api/rest/#keyset-based-pagination

Maybe, someone with GitLab knowledge and access can verify this and provide a PR.

pjungermann:
The advanced search seems to have some features that could be used for that purpose:
https://docs.gitlab.com/user/search/advanced_search/#code-search

However, it seems like it's not available for all, optional. Also, I couldn't try it and verify whether the expected search would work.

Maybe you can tap into that as an optional alternative to Location entities in a root catalog files, or all entities defined in that file.

elaine-mattos:
Hey @faeshaas,

this is a great idea, but as far as I understand, it’s not an easy one to implement. There are quite a few factors to consider, for example:

[...]

PR #31993, opened 2025-12-03
enhance gitlab discovery to use group search API

McFoggy:

as commented in #30617, in order to reduce the pressure on the gitlab API (and avoid some 429) when dealing with big groups with many projects, this PR introduces discoverability of entities by using the gitlab search API on the group level.

It uses the the [group search scope:blob API](https://docs.gitlab.com/api/search/#scope-blobs-1) to find the entity files.

From the gitlab catalog provider configuration:
- a new property is added: `useSearch`

Once set to `true`, the discovery will fire a gitlab search (paginated of course) to find matching entity files.

When using search `group` and `entityFilename` becomes mandatory otherwise no search is performed.

### DIFF
--- a/plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
+++ b/plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
@@ -245,7 +245,7 @@ export class GitlabDiscoveryEntityProvider implements EntityProvider {
       options => this.gitLabClient.listFiles(options),
       {
         group: this.config.group,
-        search: `file:${this.config.catalogFile}`,
+        search: `filename:${this.config.catalogFile}`,
         page: 1,
         per_page: 50,
       },


McFoggy:
I'd be happy to add a changeset, but how do I generate one?

I cannot execute the script `yarn changeset` as described in the doc. What did I miss?

```
$ yarn changeset
Usage Error: Couldn't find a script named "changeset".
```

image

McFoggy:
got `yarn changeset` to work after cleaning my node_modules and re installing...

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
```diff
@@ -221,6 +229,56 @@ export class GitlabDiscoveryEntityProvider implements EntityProvider {
...
+    const locations: LocationSpec[] = [];
+    let foundProjects = 0;
+
+    this.logger.info(`Using gitlab search API to lookup projects`);
+
+    const foundFiles = paginated(
+      options => this.gitLabClient.listFiles(options),
+      {
+        group: this.config.group,
+        search: `file:${this.config.catalogFile}`,
```
elaine-mattos: hey, I believe you should use `filename:` instead of `file:` for the [group-level search API](https://docs.gitlab.com/api/search/#scope-blobs-1)

```suggestion
        search: `filename:${this.config.catalogFile}`,
```

elaine-mattos:
Hey @McFoggy I really like this change! :)

Using the GitLab group search API seems inherently more performant than enumerating all projects individually. 

I tested the search API against my test instance and I see that it can also find deeper nested files like these:
```json
  {
    "basename": "generated/content/.backstage/catalog-info",
    "data": "apiVersion: backstage.io/v1alpha1\nkind: Component\nmetadata:\n",
    "path": "generated/content/.backstage/catalog-info.yaml",
    "filename": "generated/content/.backstage/catalog-info.yaml",
    "id": null,
    "ref": "afebf01d6e0042c00f2244e4f611e8ccca37d93e",
    "startline": 1,
    "project_id": 76407
  }
```
so, I also think it has the additional benefit of [supporting monorepos](https://github.com/backstage/backstage/issues/29428), since it can now find catalog-info.yaml files anywhere in the repository, not just at the root. 

My only concern is, as you mentioned in the issue,  that the group search API is only available on GitLab Premium/Ultimate tiers, so it may not be usable for all installations at the moment.

plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts
McFoggy: Hi Elaine,

`file:` was the one to use when I started that modification.

Now the docs say `filename:` and in fatc when you call the api both `file:` and `filename:` work.
I'll update the PR to use `filename:`.

McFoggy: switched to `filename:`

McFoggy:
> ... it can now find catalog-info.yaml files anywhere in the repository, not just at the root

Indeed.

> My only concern is, as you mentioned in the issue, that the group search API is only available on GitLab Premium/Ultimate tiers, so it may not be usable for all installations at the moment.

Ok but what can we do if gitlab offers this search only for 'Premium' & 'Ultimate'?

At least for the companies having such license levels, it will already be an enhancement.
It is documented this way in the PR right now.

If tomorrow, Gitlab decides to change this it will be available to a broader audience.

McFoggy:
and sorry @elaine-mattos for my late answers.

Happy new year to you and the community !

elaine-mattos:
> > ... it can now find catalog-info.yaml files anywhere in the repository, not just at the root
> 
> Indeed.
> 
> > My only concern is, as you mentioned in the issue, that the group search API is only available on GitLab Premium/Ultimate tiers, so it may not be usable for all installations at the moment.
> 
> Ok but what can we do if gitlab offers this search only for 'Premium' & 'Ultimate'?
> 
> At least for the companies having such license levels, it will already be an enhancement. It is documented this way in the PR right now.
> 
> If tomorrow, Gitlab decides to change this it will be available to a broader audience.

Yeah, you're right! Sorry, I have just seen you documented it in the docs. That's perfect and I'll add this as a patch in our installation soon! =)

McFoggy:
@elaine-mattos do I have anything more to provide to ease the integration/review?

elaine-mattos:
> @elaine-mattos do I have anything more to provide to ease the integration/review?

Hi @vinzscam , hi @benjdlambert , this looks good to me! :)

plethargy:
Hi there, I recently pulled this in to test and I've gotten an error when executing, I believe there's a small typo in this MR causing the bug. 

The search parameters generated by the plugin are as follows: `search?group=my-group&search=filename%3Acatalog-info.yaml&page=1&per_page=50&scope=blob`

and the above gives me a 400 Bad Request response. I tested it in Postman and got the following response:

```
{
    "error": "scope does not have a valid value"
}
```

Which brings me to what I believe is the issue, looking at the [gitlab documentation](https://docs.gitlab.com/api/search/#scope-blobs-2), the correct value for this scope should be `blobs` and not `blob`. 

Unfortunately scope isn't an input parameter to the plugin, therefore I cannot override the value (not to my knowledge at least) which means all requests fail to this error.

elaine-mattos:
Hi!

@plethargy , thanks! When I tested this this morning, I ran into the same error and fixed it locally, but didn’t have time to post it here then.
I’ve submitted a small fix here: https://github.com/backstage/backstage/pull/32769.
@McFoggy , hope you don’t mind that I submitted the PR ... I happened to hit the issue as well and already had the code open. ^.^

PR #31998, opened 2025-12-03
`feat(integrations)`: Add `googleGcs` to `ScmIntegrations` and implement `readTree` for the `UrlReader`

benjdlambert:
Arguably, GCS is not SCM, however there's other things in there that aren't already, such as S3 and Azure Blob Storage.

Probably wanna rethink a little bit what that should be called in the future 😅

### DIFF
--- a/packages/backend-defaults/src/entrypoints/urlReader/lib/GoogleGcsUrlReader.ts
+++ b/packages/backend-defaults/src/entrypoints/urlReader/lib/GoogleGcsUrlReader.ts
@@ -150,17 +150,13 @@ export class GoogleGcsUrlReader implements UrlReaderService {
       prefix: key,
     });
 
-    const responses = [];
-    for (const file of files) {
-      const buffer = await getRawBody(file.createReadStream());
-      responses.push({
-        data: Readable.from(buffer),
-        path: relative(key, file.name),
-        lastModifiedAt: file.metadata.updated
-          ? new Date(file.metadata.updated as string)
-          : undefined,
-      });
-    }
+    const responses = files.map(file => ({
+      data: file.createReadStream(),
+      path: relative(key, file.name),
+      lastModifiedAt: file.metadata.updated
+        ? new Date(file.metadata.updated as string)
+        : undefined,
+    }));
 
     return this.deps.treeResponseFactory.fromReadableArray(responses);
   }
--- a/.changeset/hungry-mugs-fall.md
+++ b/.changeset/hungry-mugs-fall.md
@@ -2,4 +2,4 @@
 '@backstage/integration': patch
 ---
 
-Implementing `ScmIntegraion` for `GoogleGcs`
+Implementing `ScmIntegration` for `GoogleGcs`
--- a/packages/integration/src/googleGcs/config.test.ts
+++ b/packages/integration/src/googleGcs/config.test.ts
@@ -45,25 +45,4 @@ describe('readGoogleGcsIntegrationConfig', () => {
       host: 'storage.cloud.google.com',
     });
   });
-
-  it('includes host in output', () => {
-    const output = readGoogleGcsIntegrationConfig(
-      buildConfig({
-        privateKey: 'fake-key',
-        clientEmail: 'someone@example.com',
-      }),
-    );
-    expect(output).toEqual({
-      host: 'storage.cloud.google.com',
-      privateKey: 'fake-key',
-      clientEmail: 'someone@example.com',
-    });
-  });
-
-  it('returns host when config is empty', () => {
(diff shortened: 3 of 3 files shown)


packages/backend-defaults/src/entrypoints/urlReader/lib/GoogleGcsUrlReader.ts
```diff
@@ -121,8 +133,26 @@ export class GoogleGcsUrlReader implements UrlReaderService {
     return ReadUrlResponseFactory.fromReadable(stream);
   }
 
-  async readTree(): Promise<UrlReaderServiceReadTreeResponse> {
-    throw new Error('GcsUrlReader does not implement readTree');
+  async readTree(
+    url: string,
+    _options?: UrlReaderServiceReadTreeOptions,
+  ): Promise<UrlReaderServiceReadTreeResponse> {
+    const { bucket, key } = parseURL(url);
```
freben, line 140: should we check and throw if you [try to do this on a glob](https://github.com/backstage/backstage/blob/f7aef07e509251ce3145d7f502a8e537aa0060c4/packages/backend-defaults/src/entrypoints/urlReader/lib/GoogleGcsUrlReader.ts#L135)?

packages/backend-defaults/src/entrypoints/urlReader/lib/GoogleGcsUrlReader.ts
```diff
@@ -121,8 +133,26 @@ export class GoogleGcsUrlReader implements UrlReaderService {
...
+
+    const [files] = await this.storage.bucket(bucket).getFiles({
+      autoPaginate: true,
+      prefix: key,
+    });
+
+    const responses = files.map(file => ({
+      data: file.createReadStream(),
```
freben: Not super fond of creating ALL of the read streams instantly. Is this safe at scale? Do they actually open sockets / read stuff only when actively consumed? Maybe we need to form this response a bit more manually?

packages/integration/src/googleGcs/config.ts
```diff
@@ -16,12 +16,22 @@
...
 /**
  * The configuration parameters for a single Google Cloud Storage provider.
  *
  * @public
  */
 export type GoogleGcsIntegrationConfig = {
+  /**
+   * The host of the target that this matches on.
+   */
+  host: string;
```
freben, line 34: Should we add this to config.d.ts too, as optional? Or is it never ever different (even for like government special cases or something like that)

benjdlambert: not sure it's worth it, I'm pretty sure that `storage.google.com` is the public endpoint for all GCS buckets. Think we can always add this later, but feels like it's not necessary right now.

packages/backend-defaults/src/entrypoints/urlReader/lib/GoogleGcsUrlReader.ts
benjdlambert: dobne

packages/backend-defaults/src/entrypoints/urlReader/lib/GoogleGcsUrlReader.ts
```diff
@@ -121,8 +133,36 @@ export class GoogleGcsUrlReader implements UrlReaderService {
...
+    }
+
+    const [files] = await this.storage.bucket(bucket).getFiles({
+      autoPaginate: true,
+      prefix: key,
+    });
+
+    const responses = [];
+    for (const file of files) {
+      const buffer = await getRawBody(file.createReadStream());
```
Rugvip: Is this file stream opened for each file in the bucket? Is it possible to defer it until the `content()` is called on the response item?

Not a blocker but nice optimization if possible

Rugvip (approved):

packages/backend-defaults/src/entrypoints/urlReader/lib/GoogleGcsUrlReader.ts
benjdlambert: Thinking about this a bit more honestly the original `file.createReadStream` should be enough if it's a proper nodejs stream, as they don't actually generate data until consumed.

Lemme dig around and test the behaviour.

benjdlambert: https://github.com/googleapis/nodejs-storage/blob/v7.18.0/src/file.ts#L1695 

Looks like the `makeRequest` is only when the stream is consumed, so I think the original code is better and wont be consumed until `.content` is called anyways.

PR #31999, opened 2025-12-03
added patch for postgres18 test databases

freben:
Patch for https://github.com/backstage/backstage/pull/31974 since it is being disruptive for users.


freben:
Replaced by https://github.com/backstage/backstage/pull/32011

Issue #31925, opened 2025-12-01
Introduce secure signInResolver based on GitHub user ID


JessicaJHee:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
GitHub sign-in resolver, GitHub catalog provider, GitHub user ID
```

### 🗃️ Project Area

Auth

### 🔖 Need

Relying on mutable fields (like email or username) for user resolution introduces security risks and leads to issues in entity matching when users change their user profiles. Using the immutable GitHub User ID ensures a reliable mapping between the authenticated user and their matching catalog entity. This benefits users by improving security and making the sign-in resolution flow more robust.

### 📝 Proposal

- Extend the GitHub catalog provider to ingest the GitHub User ID (`id` with GraphQL, `nodeId` with REST) as an annotation on the User entity. For example:
```
kind: User
metadata:
  annotations:
   'github.com/user-id': <user-id>
```

- Introduce a GitHub-specific sign-in resolver that matches the authenticated user's ID against the new `github.com/user-id` annotation in the catalog. I've started a [PR here](https://github.com/backstage/backstage/pull/32000) with suggested changes.

If this change is encouraged by the community, it'd be nice to do the same with GitLab

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Yes, and I have enough information to get started

vinzscam:
Thank you 🙏
Closed by #32000

PR #32000, opened 2025-12-03
feat(catalog): add github user ID in user entity

JessicaJHee:

Currently, GitHub resolvers resolve based on username or email which are both changeable by the user. Node IDs are stable/immutable identifiers that persist across user profile changes, making auth more secure and robust.

Addresses #31925 

  - Added `github.com/user-id` annotation to store GitHub's user ID (immutable) in user entities
  - Created `userIdMatchingUserEntityAnnotation` sign-in resolver that matches users by the new ID
  - Updated event handlers to include `node_id`

The ingested user entity looks like this:

```
kind: User
metadata:
  annotations:
    github.com/user-login: JessicaJHee
    github.com/user-id: MDQ6VXNlcjczOTY4NDgw
```

Note: the `node_id` from REST API is equivalent to the `id` field in GraphQL. See GitHub docs [here](https://docs.github.com/en/graphql/guides/using-global-node-ids). 

### DIFF
--- a/plugins/auth-backend-module-github-provider/report.api.md
+++ b/plugins/auth-backend-module-github-provider/report.api.md
@@ -17,13 +17,25 @@ export default authModuleGithubProvider;
 // @public (undocumented)
 export const githubAuthenticator: OAuthAuthenticator<
   PassportOAuthAuthenticatorHelper,
-  PassportProfile
+  GithubProfile
 >;
+
+// @public (undocumented)
+export type GithubProfile = PassportProfile & {
+  nodeId?: string;
+};
 
 // @public
 export namespace githubSignInResolvers {
   const usernameMatchingUserEntityName: SignInResolverFactory<
-    OAuthAuthenticatorResult<PassportProfile>,
+    OAuthAuthenticatorResult<GithubProfile>,
+    | {
+        dangerouslyAllowSignInWithoutUserInCatalog?: boolean | undefined;
+      }
+    | undefined
+  >;
+  const userIdMatchingUserEntityAnnotation: SignInResolverFactory<
+    OAuthAuthenticatorResult<GithubProfile>,
     | {
         dangerouslyAllowSignInWithoutUserInCatalog?: boolean | undefined;
       }
--- a/plugins/catalog-backend-module-github/src/lib/annotation.ts
+++ b/plugins/catalog-backend-module-github/src/lib/annotation.ts
@@ -35,3 +35,13 @@ export const ANNOTATION_GITHUB_USER_LOGIN = 'github.com/user-login';
  * @public
  */
 export const ANNOTATION_GITHUB_TEAM_SLUG = 'github.com/team-slug';
+
+/**
+ * The value of this annotation is the unique identifier for a user on
+ * [GitHub](https://github.com) (either the public one, or a private GitHub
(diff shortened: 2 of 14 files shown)


JessicaJHee:
@freben Yes, from [these docs](https://docs.github.com/en/graphql/guides/using-global-node-ids) they're meant to be the same and are the established way to reference the same user across REST and GraphQL APIs.

"You can find the global node ID of many objects from within the REST API and use these IDs in your GraphQL operations". 

[This](https://docs.github.com/en/graphql/guides/using-global-node-ids#putting-global-node-ids-to-use) would be the flow:

> 1. Call a REST endpoint that returns an object's node_id.
> 2. Find the object's type in GraphQL.
> 3. Use the ID and type to do a direct node lookup in GraphQL.

plugins/catalog-backend-module-github/src/lib/annotation.ts
```diff
@@ -35,3 +35,13 @@ export const ANNOTATION_GITHUB_USER_LOGIN = 'github.com/user-login';
...
+
+/**
+ * The value of this annotation is the unique identifier for a user on
+ * [GitHub](https://github.com) (either the public one, or a private GitHub
+ * Enterprise installation) that is related to this entity. This is a stable
+ * identifier that persists even if the user changes their login name.
+ *
+ * @public
+ */
+export const ANNOTATION_GITHUB_USER_ID = 'github.com/user-id';
```
freben, line 47: I don't think this made it all the way to the public API - needs index.ts

plugins/auth-backend-module-github-provider/report.api.md
```diff
@@ -20,6 +20,11 @@ export const githubAuthenticator: OAuthAuthenticator<
   PassportProfile
```
freben: Should `GithubProfile` be replacing `PassportProfile` here (and perhaps in more places)?

plugins/auth-backend-module-github-provider/report.api.md
```diff
@@ -20,6 +20,11 @@ export const githubAuthenticator: OAuthAuthenticator<
...
 >;
 
+// @public (undocumented)
+export type GitHubProfile = PassportProfile & {
```
freben:
```suggestion
export type GithubProfile = PassportProfile & {
```

Sorry I think we went for single-word capitalization of github in identifiers

freben:
Oh! And remember changesets.

plugins/catalog-backend-module-github/src/lib/annotation.ts
JessicaJHee: I added the annotations to `index.ts`, hopefully that addresses this 👍

JessicaJHee:
@freben thank you so much for the review! I've addressed your feedback, PTAL again :)

plugins/auth-backend-module-github-provider/report.api.md
freben: Was it possible to get the type on the authenticator itself too?

JessicaJHee: Ahh shoot totally missed that, should be good now!

freben (approved):

PR #32010, opened 2025-12-04
Techdocs: use AWS S3 integration

vinzscam:

This PR adds support for TechDocs to use [AWS S3 Integration](https://backstage.io/docs/integrations/aws-s3/locations) via `integrations.awsS3` config for adopters who already have a location integration. The AWS S3 integration will be used as last fallback before using the default credential chain, so the priority will be the following:

```
1. `aws.accounts`
2. `techdocs.publisher.awsS3.credentials`
3. `integrations.awsS3`
4. Default credential chain
```

### DIFF
--- a/plugins/techdocs-node/src/stages/publish/awsS3.ts
+++ b/plugins/techdocs-node/src/stages/publish/awsS3.ts
@@ -61,6 +61,7 @@ import {
   TechDocsMetadata,
 } from './types';
 import { LoggerService } from '@backstage/backend-plugin-api';
+import { AwsS3Integration, ScmIntegrations } from '@backstage/integration';
 
 const streamToBuffer = (stream: Readable): Promise<Buffer> => {
   return new Promise((resolve, reject) => {
@@ -142,16 +143,16 @@ export class AwsS3Publish implements PublisherBase {
 
     const credsManager = DefaultAwsCredentialsManager.fromConfig(config);
 
-    const awsS3IntegrationConfig =
-      config.getOptionalConfigArray('integrations.awsS3');
+    const scmIntegrations = ScmIntegrations.fromConfig(config);
+    const awsS3Integrations = scmIntegrations.awsS3.list();
 
     const sdkCredentialProvider = await AwsS3Publish.buildCredentials(
       credsManager,
       logger,
+      awsS3Integrations,
       accountId,
       credentialsConfig,
       region,
-      awsS3IntegrationConfig,
     );
 
     // AWS endpoint is an optional config. If missing, the default endpoint is built from
@@ -220,10 +221,10 @@ export class AwsS3Publish implements PublisherBase {
   private static async buildCredentials(
     credsManager: AwsCredentialsManager,
     logger: LoggerService,
+    awsS3Integrations: AwsS3Integration[],
     accountId?: string,
     credentialsConfig?: Config,
     region?: string,
-    awsS3IntegrationConfig?: Config[],
   ): Promise<AwsCredentialIdentityProvider> {
     // Pull credentials for the specified account ID from the 'aws' config section
     if (accountId) {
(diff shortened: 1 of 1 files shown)


.changeset/whole-dodos-matter.md
```diff
@@ -0,0 +1,12 @@
+---
+'@backstage/plugin-techdocs-node': minor
+---
+
+It's now possible to use the credentials from the `integrations.awsS3` config to authenticate with AWS S3. The new priority is:
+
+1. `aws.accounts`
+2. `techdocs.publisher.awsS3.credentials`
+3. `integrations.awsS3`
```
freben, line 9: In a way, this is a risky breaking change. Is there a risk that some user out there was happily using the default credential chain for techdocs publishing but explicit credentials for catalog ingestion, that now start to see failures?

plugins/techdocs-node/src/stages/publish/awsS3.ts
```diff
@@ -148,12 +139,19 @@ export class AwsS3Publish implements PublisherBase {
     const credentialsConfig = config.getOptionalConfig(
       'techdocs.publisher.awsS3.credentials',
     );
+
     const credsManager = DefaultAwsCredentialsManager.fromConfig(config);
+
+    const awsS3IntegrationConfig =
+      config.getOptionalConfigArray('integrations.awsS3');
```
freben: You should not try to touch the integrations config directly. [Use `ScmIntegrations.fromConfig` instead](https://github.com/backstage/backstage/blob/f7aef07e509251ce3145d7f502a8e537aa0060c4/packages/integration/src/ScmIntegrations.ts#L65) and then its `.awsS3` API.

Ideally there should be an integrationsServiceRef but that's for the future.

vinzscam: Yep, good call!

.changeset/whole-dodos-matter.md
vinzscam: So just to clarify, the techdocs "publishing" config is for reading techdocs config. So in general this is a risk that exists and perhaps is worth calling out as breaking in the release notes, but the change goes in the direction to reuse existing credentials, unless other credentials are specified at the plugin level.

freben: Aight

plugins/techdocs-node/src/stages/publish/awsS3.ts
```diff
@@ -148,12 +139,19 @@ export class AwsS3Publish implements PublisherBase {
     const credentialsConfig = config.getOptionalConfig(
       'techdocs.publisher.awsS3.credentials',
     );
+
     const credsManager = DefaultAwsCredentialsManager.fromConfig(config);
+
+    const awsS3IntegrationConfig =
```
freben: Still peeking inside the integrations config - remember to make an integrations object instead that does all of the config parsing

freben (approved):
Alright, approving. Only thing I feel at this point is that the logic of `getExplicitCredentials` is a little hard to follow. It has many paths. Not many `else` that inform you that "oh no you are now falling through to the fallback case". But it's pretty OK.

.changeset/whole-dodos-matter.md
```diff
@@ -0,0 +1,20 @@
+---
+'@backstage/plugin-techdocs-node': minor
+---
+
+**BREAKING:** It's now possible to use the credentials from the `integrations.awsS3` config to authenticate with AWS S3. The new priority is:
+
+1. `aws.accounts`
+2. `techdocs.publisher.awsS3.credentials`
+3. `integrations.awsS3`
```
johnphilip283, line 9: Hm, I would almost expect the credential order to be:

1. `techdocs.publisher.awsS3.credentials`
2. `aws.accounts`
3. `integrations.awsS3`

because I'd expect the most specific configuration to take precedence - is this for a specific reason or just intuition?

vinzscam: I've kept the logic as it was. I think `aws.accounts` is at the top even before this PR

PR #32011, opened 2025-12-04
Patch release of #31974

freben:
This release fixes an issue where `TestDatabases` sometimes did not run properly for PostgreSQL 18


freben:
I don't think it's necessary. It's a "black box" transient throwaway image. It just needs to store data _somewhere_ - and if we didn't have the tempfs parameter to have to align with, we would never have noticed this either.

PR #32014, opened 2025-12-04
Use a versioned context for `useEntityList`

freben:
This does the same thing as `useEntity`, to ensure that a consistent context will be presented even if there's a mixture of `@backstage/plugin-catalog-react` versions - which in particular can happen when using nightlies or -next versions.

### DIFF
--- a/plugins/catalog-react/report.api.md
+++ b/plugins/catalog-react/report.api.md
@@ -322,7 +322,7 @@ export const EntityLifecyclePicker: (props: {
   initialFilter?: string[];
 }) => JSX_2.Element;
 
-// @public
+// @public @deprecated
 export const EntityListContext: Context<
   EntityListContextProps<any> | undefined
 >;
--- a/plugins/catalog-react/src/hooks/useEntityListProvider.tsx
+++ b/plugins/catalog-react/src/hooks/useEntityListProvider.tsx
@@ -17,6 +17,11 @@
 import { QueryEntitiesResponse } from '@backstage/catalog-client';
 import { Entity } from '@backstage/catalog-model';
 import { useApi } from '@backstage/core-plugin-api';
+import {
+  createVersionedContext,
+  createVersionedValueMap,
+  useVersionedContext,
+} from '@backstage/version-bridge';
 import { compact, isEqual } from 'lodash';
 import qs from 'qs';
 import {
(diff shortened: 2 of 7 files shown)


plugins/catalog-react/src/hooks/useEntityListProvider.tsx
```diff
@@ -500,8 +516,22 @@ export const EntityListProvider = <EntityFilters extends DefaultEntityFilters>(
...
   EntityFilters extends DefaultEntityFilters = DefaultEntityFilters,
 >(): EntityListContextProps<EntityFilters> {
-  const context = useContext(EntityListContext);
-  if (!context)
-    throw new Error('useEntityList must be used within EntityListProvider');
-  return context;
+  const versionedHolder = useVersionedContext<{
+    1: EntityListContextProps<any>;
+  }>('entity-list-context');
+  const oldContext = useContext(EntityListContext);
```
Rugvip: Don't think this is needed really, because if the producer isn't providing the new versions context we'll have the duplication problem on the old context anyway. i.e. it should be safe to just throw away the old context and use the new one only straight away

freben: The old `EntityListContext` (not just `EntityListProvider`) is exported directly. I will replace that export with the new versioned one, forgot to do that - but does that change the picture at all or is it still the same situation?

benjdlambert (approved):

plugins/catalog-react/src/hooks/useEntityListProvider.tsx
Rugvip: Hmm, nah don't export the new one at all. Should never export contexts directly, not even the versioned one, just the hook to consume them. Makes sense to keep the compatibility code and export of the old context though, but let's also try to get rid of that export asap

plugins/catalog-react/report.api.md
```diff
@@ -322,9 +323,12 @@ export const EntityLifecyclePicker: (props: {
...
 
-// @public
+// @public @deprecated
 export const EntityListContext: Context<
-  EntityListContextProps<any> | undefined
+  | VersionedValue<{
+      1: EntityListContextProps<any>;
+    }>
+  | undefined
 >;
```
Rugvip, line 328: Very much a breaking change this

plugins/catalog-react/src/hooks/useEntityListProvider.tsx
freben: @Rugvip Are you saying to revert back to still exporting (and supporting) the old context and just marking it deprecated - making this NOT a breaking change again?

freben: Reverted!

plugins/catalog-react/report.api.md
freben: Reverted!

Issue #31927, opened 2025-12-01
Flawed cookie removal in OAuthCookieManager causes authentication loops with chunked tokens


allegro-dbargowski:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
chunked token
refresh token
cookie removal
```

### 🗃️ Project Area

Auth

### 🔗 External Integration

Other

## Description

A flaw in the `OAuthCookieManager`'s cookie removal logic is causing authentication loops for users with large refresh tokens. The issue occurs because the `OAuthCookieManager` fails to properly clear old, chunked refresh token cookies after a change in the cookie domain handling, resulting in two sets of chunked cookies being present in the browser. This issue became noticable for us after updating `@backstage/plugin-auth-node` dependency from version `0.6.6` to version `0.6.8`.

## Problem Behavior

Users who had a valid session with a large (chunked) refresh token prior to the introduction of commit `3aff9e1213aa00f9de1a13955ad99ac6f0e59157` are forced into a login loop. Any attempt to navigate the application results in them being redirected to the login page. Manually clearing browser cookies for the application domain resolves the issue, but several users at Allegro are affected (therefore putting a strain on support team of backstage-based company developer portal as we're getting bug reports).

## Technical Analysis

The root cause of this issue lies in the `setCookie` method of the `OAuthCookieManager` class. Commit `3aff9e1213aa00f9de1a13955ad99ac6f0e59157` introduced a change to remove the leading dot from the cookie's domain attribute. To handle the transition, a piece of code was added to explicitly remove the old cookie that was set with the prefixed domain.

```typescript
// If using the default cookieConfigurer, delete old cookie with domain set to the callbackUrl's domain (legacy behavior)
if (this.cookieConfigurer === defaultCookieConfigurer) {
  const { hostname: domain } = new URL(this.options.callbackUrl);
  output = output.cookie(name, '', {
    ...this.getRemoveCookieOptions(),
    domain: `.${domain}`,
  });
}
```

This cleanup logic, however, only targets the main cookie name (e.g., `provider-refresh-token`) and does not account for chunked cookies (e.g., `provider-refresh-token-0`, `provider-refresh-token-1`, etc.).

[...]

PR #32033, opened 2025-12-05
fix(auth): update cookie deletion logic for chunked cookies

JessicaJHee:

Addresses #31927 

When migrating away from the old cookie format (where the domain attribute was explicitly set), the cleanup now correctly checks for and deletes any chunked cookies (`<name>-0`, `<name>-1`, ...) if they exist, in addition to the main one.
  - Moved deletion logic earlier in the `setCookie()` flow
  - Added unit tests to cover migration cases

_Aside:_ I also found another potential issue: when the new cookie has fewer chunks than the old, the excess chunks are not cleaned up. I’m not sure how common this use case is, but I can propose this fix in another PR if necessary.

### DIFF
--- a/.changeset/sour-bats-press.md
+++ b/.changeset/sour-bats-press.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-auth-node': patch
+---
+
+fix flawed cookie removal logic with chunked tokens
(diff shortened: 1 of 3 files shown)


freben (approved):
Thanks!

PR #32047, opened 2025-12-06
frontend-plugin-api: new error boundary API option + boundary for app root elements

Rugvip:

The app root elements weren't wrapped in extension boundaries because the error display didn't make sense for them. This adds a new option to the `ExtensionBoundary` that switches the error rendering from using `ErrorDisplay` to instead posting to the `ErrorApi`.  It then uses that new option to wrap elements passed to `AppRootElementBlueprint` in an `ExtensionBoundary`, making sure they can't crash the entire app.

Bit of an edge-case was a new race condition where the `AlertApi` forwarding to the `AlertDisplay` didn't work for alerts that were posted before the `AlertDisplay` got the chance to render. This fixes that by adding replay functionality to the `AlertApi` for new subscribers.

### DIFF
--- a/packages/frontend-plugin-api/src/blueprints/AppRootElementBlueprint.tsx
+++ b/packages/frontend-plugin-api/src/blueprints/AppRootElementBlueprint.tsx
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { ExtensionBoundary } from '@backstage/frontend-plugin-api';
 import { coreExtensionData, createExtensionBlueprint } from '../wiring';
 
 /**
@@ -26,7 +27,11 @@ export const AppRootElementBlueprint = createExtensionBlueprint({
   kind: 'app-root-element',
   attachTo: { id: 'app/root', input: 'elements' },
   output: [coreExtensionData.reactElement],
-  *factory(params: { element: JSX.Element }) {
-    yield coreExtensionData.reactElement(params.element);
+  *factory(params: { element: JSX.Element }, { node }) {
+    yield coreExtensionData.reactElement(
+      <ExtensionBoundary node={node} errorPresentation="error-api">
+        {params.element}
+      </ExtensionBoundary>,
+    );
   },
 });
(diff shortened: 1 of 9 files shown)


packages/frontend-plugin-api/src/components/ExtensionBoundary.tsx
```diff
@@ -70,18 +82,30 @@ export function ExtensionBoundary(props: ExtensionBoundaryProps) {
   // Skipping "routeRef" attribute in the new system, the extension "id" should provide more insight
   const attributes = {
     extensionId: node.spec.id,
-    pluginId: node.spec.plugin?.id ?? 'app',
+    pluginId: plugin.id ?? 'app',
```
freben, line 85: All other parts of our code have defensive `if` statements handling a missing plugin field, I guess those can be removed now too then.

packages/frontend-plugin-api/src/blueprints/AppRootElementBlueprint.tsx
```diff
@@ -26,7 +27,11 @@ export const AppRootElementBlueprint = createExtensionBlueprint({
   kind: 'app-root-element',
   attachTo: { id: 'app/root', input: 'elements' },
   output: [coreExtensionData.reactElement],
-  *factory(params: { element: JSX.Element }) {
-    yield coreExtensionData.reactElement(params.element);
+  *factory(params: { element: JSX.Element }, { node }) {
+    yield coreExtensionData.reactElement(
+      <ExtensionBoundary node={node} errorPresentation="error-api">
```
freben, line 32: Hm all the way up here, what will the error api do? Will this get logged, since there may be no renderable code that actually knows how to show the sent error messages?

It's almost like the error api should have code that triggers logging if there are no listeners at all in the next couple of milliseconds or so.

Rugvip: It'll be forwarded to the error and then alert API and shown as an error alert in the app

packages/frontend-plugin-api/src/components/ExtensionBoundary.tsx
Rugvip: yep should be safe to remove by now afaik

PR #32050, opened 2025-12-06
feat(ui)!: redesign Table component with new useTable hook API

cdedreuille:

Redesigns the Table component to provide a better developer experience with a new `useTable` hook supporting three pagination modes: complete (all data loaded upfront), offset (server-side), and cursor (server-side).

### Breaking changes

- `Table` component (React Aria wrapper) is renamed to `TableRoot`
- New high-level `Table` component handles data display, pagination, sorting, and selection
- `useTable` hook completely redesigned with new API

### New features

- Unified `useTable` hook with mode discriminator for all pagination patterns (complete, offset, cursor)
- Custom page caching for server-side pagination with bidirectional navigation and request cancellation
- Debounced query changes to reduce backend load
- Stale data preservation with visual indicator during reloads
- Row selection with toggle/replace behaviors
- Per-row disable control via `getIsDisabled`

### Migration guide

1. Update imports and use the new `useTable` hook:

```diff
-import { Table, useTable } from '@backstage/ui';
-const { data, paginationProps } = useTable({ data: items, pagination: {...} });
+import { Table, useTable, type ColumnConfig } from '@backstage/ui';
+const { tableProps } = useTable({
+  mode: 'complete',
+  getData: () => items,
+});
```

2. Define columns and render with the new Table API:

```diff
-<Table aria-label="My table">
-  <TableHeader>...</TableHeader>
-  <TableBody items={data}>...</TableBody>
-</Table>
-<TablePagination {...paginationProps} />
+const columns: ColumnConfig<Item>[] = [
+  { id: 'name', label: 'Name', isRowHeader: true, cell: item => <CellText title={item.name} /> },
+  { id: 'type', label: 'Type', cell: item => <CellText title={item.type} /> },
+];
+
+<Table columnConfig={columns} {...tableProps} />
```

### DIFF
--- a/packages/ui/report.api.md
+++ b/packages/ui/report.api.md
@@ -679,7 +679,7 @@ export interface FieldLabelProps
 // @public (undocumented)
 export interface FilterState<TFilter> {
   // (undocumented)
-  onFilterChange: (filter: TFilter) => void;
+  onChange: (value: TFilter) => void;
   // (undocumented)
   value: TFilter | undefined;
 }
@@ -1258,7 +1258,7 @@ export interface SearchFieldProps
 // @public (undocumented)
 export interface SearchState {
   // (undocumented)
-  onSearchChange: (value: string) => void;
+  onChange: (value: string) => void;
   // (undocumented)
   value: string;
 }
--- a/packages/ui/src/components/Table/hooks/types.ts
+++ b/packages/ui/src/components/Table/hooks/types.ts
@@ -20,13 +20,13 @@ import type { SortDescriptor, TableItem, TableProps } from '../types';
 /** @public */
 export interface FilterState<TFilter> {
   value: TFilter | undefined;
-  onFilterChange: (filter: TFilter) => void;
+  onChange: (value: TFilter) => void;
 }
 
 /** @public */
 export interface SearchState {
   value: string;
-  onSearchChange: (value: string) => void;
+  onChange: (value: string) => void;
 }
 
 /** @public */
--- a/packages/ui/src/components/Table/hooks/useTable.ts
+++ b/packages/ui/src/components/Table/hooks/useTable.ts
@@ -98,9 +98,16 @@ export function useTable<T extends TableItem, TFilter = unknown>(
 ): UseTableResult<T, TFilter> {
   const query = useQueryState<TFilter>(options);
 
+  const initialModeRef = useRef(options.mode);
+  if (initialModeRef.current !== options.mode) {
+    throw new Error(
+      `useTable mode cannot change from '${initialModeRef.current}' to '${options.mode}'. ` +
+        `The mode must remain stable for the lifetime of the component.`,
+    );
+  }
+
   let pagination: PaginationResult<T> & { reload: () => void };
 
-  // Conditional hooks - mode is stable for lifetime of component
   if (options.mode === 'complete') {
     pagination = useCompletePagination(options, query);
   } else if (options.mode === 'offset') {
@@ -125,7 +132,7 @@ export function useTable<T extends TableItem, TFilter = unknown>(
   return {
     tableProps,
     reload: pagination.reload,
-    filter: { value: query.filter, onFilterChange: query.setFilter },
-    search: { value: query.search, onSearchChange: query.setSearch },
+    filter: { value: query.filter, onChange: query.setFilter },
+    search: { value: query.search, onChange: query.setSearch },
   };
 }
--- a/packages/ui/src/components/Table/stories/Table.dev.stories.tsx
+++ b/packages/ui/src/components/Table/stories/Table.dev.stories.tsx
@@ -30,12 +30,12 @@ import {
   type ColumnConfig,
 } from '..';
 import { Button } from '../../Button';
-import { TextField } from '../../TextField';
 import { Select } from '../../Select';
 import { Flex } from '../../Flex';
 import { data as data1 } from './mocked-data1';
 import { data as data4 } from './mocked-data4';
 import { selectionData, selectionColumns, tableStoriesMeta } from './utils';
+import { SearchField } from '../../SearchField';
 
 const meta = {
   title: 'Backstage UI/Table/dev',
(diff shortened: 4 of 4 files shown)


packages/ui/src/components/Table/hooks/useTable.ts
```diff
@@ -13,154 +13,119 @@
...
-    paginationProps,
+/** @public */
+export function useTable<T extends TableItem, TFilter = unknown>(
+  options: UseTableOptions<T, TFilter>,
+): UseTableResult<T, TFilter> {
+  const query = useQueryState<TFilter>(options);
+
+  let pagination: PaginationResult<T> & { reload: () => void };
+
+  // Conditional hooks - mode is stable for lifetime of component
```
Rugvip: Thinking it would be best to add a check for this, throwing if the mode changes

packages/ui/report.api.md
```diff
@@ -1535,48 +1781,69 @@ export const useBreakpoint: () => {
...
 
 // @public (undocumented)
-export interface UseTableResult<T = any> {
-  data?: T[];
-  pagination: UseTablePagination<T>;
-  paginationProps: TablePaginationProps;
+export type UseTableOptions<T extends TableItem, TFilter = unknown> =
+  | UseTableCompleteOptions<T, TFilter>
+  | UseTableOffsetOptions<T, TFilter>
+  | UseTableCursorOptions<T, TFilter>;
```
Rugvip, line 1832: Worth checking whether overloads of `useTable` provides better TS completions compared to this union

packages/ui/src/components/Table/hooks/types.ts
```diff
@@ -15,73 +15,144 @@
...
 import type { TablePaginationProps } from '../../TablePagination/types';
+import type { SortDescriptor, TableItem, TableProps } from '../types';
 
 /** @public */
-export interface UseTablePaginationConfig {
-  /** Total number of rows in the dataset - only needed when data is not provided at the top level */
-  rowCount?: number;
+export interface FilterState<TFilter> {
+  value: TFilter | undefined;
+  onFilterChange: (filter: TFilter) => void;
```
Rugvip: Thinking this should either be `onValueChange` or `value` should be renamed to `filter`? same with search

Btw, are these named like this because they're intended to be passed somewhere? Otherwise I'd expect something like `setFilter` or `setValue` to have it from the producer PoV

packages/ui/src/components/Table/hooks/types.ts
```diff
@@ -15,73 +15,144 @@
...
+  signal: AbortSignal;
 }
 
 /** @public */
-export interface UseTableConfig<T = any> {
-  /** Full dataset - when provided, rowCount is calculated automatically and sliced data is returned */
-  data?: T[];
-  /** Pagination configuration */
-  pagination?: UseTablePaginationConfig;
+export interface OffsetResponse<T> {
```
Rugvip, line 76: Think I'd expect at least `pageSize` in this one too, since it's often up to the backend to decide what page sizes are actually allowed, and what the default is. As in you can provide some `pageSize`, but it's just gonna be a hint.

packages/ui/report.api.md
```diff
@@ -1535,48 +1781,69 @@ export const useBreakpoint: () => {
...
+  filter: FilterState<TFilter>;
+  // (undocumented)
+  reload: () => void;
+  // (undocumented)
+  search: SearchState;
+  // (undocumented)
+  tableProps: Omit<
+    TableProps<T>,
+    'columnConfig' | 'rowConfig' | 'selection' | 'emptyState'
+  >;
```
Rugvip, line 1846: We could consider making this some form of opaque type instead, just to provide a bit more flexibility in how we evolve the APIs. Basically some opaque object that contains all of this information, but isn't transparent to the caller both of `useTable` and the `Table` component. Biggest downside of that would be that you can no longer transform the props before passing them to `Table`. You could still override by providing other props to `Table`, but not make changes based on existing values.

packages/ui/report.api.md
Naycon: The union gives full completion on the options (e.g. you only see `sortFn` if you set mode to `complete`), so I think we're covered. Or did you mean something else?

packages/ui/src/components/Table/hooks/types.ts
Naycon: Sure, can rename `onFilterChange` to `onValueChange` or `onChange`. `filter.filter` or `search.search` looked silly, and the intention is to pass these on to whatever you use for search/filtering (e.g. `SearchField` via `value` and `onChange`). They were not necessarily meant to be spreadable out of the box though, but maybe it's nice to be able to do `<SearchField ... {..search}>` without having to remap those. 

I'll change to `onChange` for now, let me know if you'd prefer something else.

packages/ui/src/components/Table/hooks/types.ts
Naycon: Hmm, I see what you mean. But I'd also like to argue that it's the responsibility of the dev to make sure to only use options their backend can support. I'm reading your suggestion like we should dynamically adapt based on what the server responded with, but maybe that's not what you meant? I think it could get a little messy, and possible confusing, if you feed one desired page size, have a set of options to choose from, and then what's showing on page is something different.

I realized as I wrote that, that it should be possible to change what pagination size options the table offers though. I'll add that.

packages/ui/report.api.md
Naycon: I'd say another big downside is that we're locking down `useTable` + `Table` as a combo that has to be used together. I was thinking that being able to use `Table` for convenience but opt to roll your own complete custom logic for what we cover with `useTable` could be an USP here. E.g. perhaps you're using some state solution and you still want to use, but you don't want to manually compose the parts needed to create a good looking and accessible table. I was also thinking perhaps the opposite is true, `useTable` does exactly what you want, but you're in need of customizing the Table so you create your own and pick what you need from `useTable` and elsewhere.

The intention when designing this was to allow that, but we can of course change our minds there and only expose this as a combo to be used together. I'll willingly admit I'm not sure what the right level of openness/closeness is here.

packages/ui/src/components/Table/hooks/types.ts
Rugvip: Yep let's try that, I do agree it gets a bit strange to have it change anyway so it's a good way to make sure it's correct in the first place

packages/ui/report.api.md
Rugvip: Ah yep, true. Yeah let's not, it's a change we could consider in the future if it turns out to be a big problem

PR #32052, opened 2025-12-07
refactor: updated GitHub repo creation to make differences clear

fearphage:

It was hard to tell what was the difference between the two repo creation paths. This makes it very clear.

### Changes

* Removed code duplication between repo creation paths.
* Sorted the keys of the input params alphabetically.
* Renamed test file to `helpers.test.ts` (made helpers plural) to mirror the file it is testing. 
* Switch some property values to shorthand `{ foo: foo }` -> `{ foo }`
* ~~Inlined some single-use variables.~~

### DIFF
--- a/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
+++ b/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
@@ -88,45 +88,35 @@ export async function createGithubRepoWithCollaboratorsAndTopics(
     await validateAccessTeam(client, access);
   }
 
-  const [reposMethod, extraParams]: [
-    'createInOrg' | 'createForAuthenticatedUser',
-    any,
-  ] =
+  const baseRepoParams = {
+    allow_auto_merge: allowAutoMerge,
+    allow_merge_commit: allowMergeCommit,
+    allow_rebase_merge: allowRebaseMerge,
+    allow_squash_merge: allowSquashMerge,
+    allow_update_branch: allowUpdateBranch,
+    auto_init: autoInit,
+    delete_branch_on_merge: deleteBranchOnMerge,
+    description,
+    has_issues: hasIssues,
+    has_projects: hasProjects,
+    has_wiki: hasWiki,
+    homepage,
+    name: repo,
+    private: repoVisibility === 'private',
+    squash_merge_commit_message: squashMergeCommitMessage,
+    squash_merge_commit_title: squashMergeCommitTitle,
+  };
+  const repoCreationPromise =
     user.data.type === 'Organization'
-      ? [
-          'createInOrg',
-          {
-            // Custom properties only available on org repos
-            custom_properties: customProperties,
-            org: owner,
-            // @ts-ignore https://github.com/octokit/types.ts/issues/522
-            visibility: repoVisibility,
-          },
-        ]
-      : ['createForAuthenticatedUser', null];
-  const repoCreationPromise = client.rest.repos[reposMethod](
-    Object.assign(
-      {
-        allow_auto_merge: allowAutoMerge,
-        allow_merge_commit: allowMergeCommit,
-        allow_rebase_merge: allowRebaseMerge,
-        allow_squash_merge: allowSquashMerge,
-        allow_update_branch: allowUpdateBranch,
-        auto_init: autoInit,
-        delete_branch_on_merge: deleteBranchOnMerge,
-        description,
-        has_issues: hasIssues,
-        has_projects: hasProjects,
-        has_wiki: hasWiki,
-        homepage,
-        name: repo,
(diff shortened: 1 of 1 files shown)


plugins/scaffolder-backend-module-github/src/actions/helpers.ts
```diff
@@ -89,49 +88,45 @@ export async function createGithubRepoWithCollaboratorsAndTopics(
     await validateAccessTeam(client, access);
   }
 
-  const repoCreationPromise =
+  const [reposMethod, extraParams]: [
+    'createInOrg' | 'createForAuthenticatedUser',
+    any,
+  ] =
```
aramissennyeydd: Can we move this to something like

```
const createParams = {...}

const isOrganization = user.data.type === 'Organization'

if(isOrganization){
   createParams.custom_properties = ...
}

const repoCreationPromise = isOrganization ? client.rest.repos.createInOrg(createParams) : client.rest.repos.createForAuthenticatedUser(createParams);
```

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
```diff
@@ -336,8 +319,16 @@ export async function initRepoPushAndProtect(
...
+    commitMessage:
+      getGitCommitMessage(gitCommitMessage, config) || 'initial commit',
+    gitAuthorInfo: {
+      email:
+        gitAuthorEmail ||
+        config.getOptionalString('scaffolder.defaultAuthor.email'),
+      name:
+        gitAuthorName ||
+        config.getOptionalString('scaffolder.defaultAuthor.name'),
+    },
```
aramissennyeydd: This is stylistic right? IMO let's keep it as it was.

fearphage: @aramissennyeydd I don't believe so. Creating variables gives the implication of reuse or at a minimum renaming (read: adding clarity).  However there is no reuse here. The single-use variables don't bring any additional clarity/understanding to this code block is the reason that I removed them. From what I can tell, they seem to be created solely to use the object-property shorthand and never referenced again. 

I don't mind making the change, but can you elaborate on the benefits/goals of these intermediate variables and the additional level of indirection?

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
fearphage: @aramissennyeydd Do you have a problem with the single call to `client.rest.repos[ternary statement here to pick method](createParams)`? I think this communicates that everything is shared/identical besides the method name. It makes the unique portion obvious which is the goal of the PR.

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
aramissennyeydd: My concern here is that it's a change from one local maxima to another, as-is it's a bit more Javascript-y but I can see the other side of this as well. In these situations, I generally err on the side of keeping it as it is to get ahead of someone else coming along and trying to change it back 🙂 .

If you're still interested in helping refactor this method - the 40+ line method parameter definition + use of ordered params instead of named params is asking for trouble 😁.

fearphage: Reverted.

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
fearphage: @aramissennyeydd The problem with doing this piecemeal set of changes is convincing _typescript_ of the shape of the `createParams` variable. I was not able to overcome typescript checking after several iterations so I just reverted the changes you requested below.

What are you trying to accomplish/communicate with your proposed change above? Maybe I can capture it another way? Unless you know how to convince typescript that the required properties exist and are not `undefined` when using your suggestion above.

aramissennyeydd: @fearphage Couple of ideas,
1. Move the `repoCreationPromise` into the `if` block - if you place the `createForAuthenticatedUser` call in an `else` block, typescript should be smart enough to figure out the types. You may need to do a new object with `...createParams` to make the types happy there.
2. I'm assuming you tried using a union type here - you could also try setting it as `Partial<Params<createInOrg[0]>>` and then cast it as `as Params<createInOrg[0]>` in the if block. I _think_ typescript is smart enough to check that.

Mostly trying to get Typescript to have correct type checking - with your params using `any` here, the types essentially get erased I think.

fearphage: @aramissennyeydd Figured it out. Let me know how this looks.

PR #32065, opened 2025-12-08
fix: $contains may have string value in an entity filter.

karpfslp:
…: { 'metadata.tags': { $contains: 'java' } }

`$contains` may have string value in an entity filter.
For example:
`{ 'metadata.tags': { $contains: 'java' } }`

Typescript type and Zod parser were not the same.
I repaired the Entity Predicate types and added types to the Zod parser Jest test to make them equal.

---

Right now, only this works:

```ts
EntityContentLayoutBlueprint.make({
  params: {
    filter: {
      'metadata.tags': {
        $contains: 'java' as unknown as EntityPredicateExpression,
      },
    },
```

### DIFF
--- a/plugins/catalog-react/src/alpha/predicates/createEntityPredicateSchema.ts
+++ b/plugins/catalog-react/src/alpha/predicates/createEntityPredicateSchema.ts
@@ -36,7 +36,7 @@ export function createEntityPredicateSchema(z: typeof zImpl) {
   const expressionSchema = z.lazy(() =>
     z.union([
       z.record(z.string().regex(/^(?!\$).*$/), valuePredicateSchema),
-      z.record(z.string().regex(/(?!\$)+/), z.never()),
+      z.record(z.string().regex(/^\$/), z.never()),
     ]),
   ) as ZodType<EntityPredicateExpression>;
 
--- a/plugins/catalog-react/src/alpha/blueprints/EntityCardBlueprint.test.tsx
+++ b/plugins/catalog-react/src/alpha/blueprints/EntityCardBlueprint.test.tsx
@@ -120,7 +120,7 @@ describe('EntityCardBlueprint', () => {
                               "not": {},
                             },
                             "propertyNames": {
-                              "pattern": "(?!\\$)+",
+                              "pattern": "^\\$",
                             },
                             "type": "object",
                           },
--- a/plugins/catalog-react/src/alpha/blueprints/EntityContentBlueprint.test.tsx
+++ b/plugins/catalog-react/src/alpha/blueprints/EntityContentBlueprint.test.tsx
@@ -122,7 +122,7 @@ describe('EntityContentBlueprint', () => {
                               "not": {},
                             },
                             "propertyNames": {
-                              "pattern": "(?!\\$)+",
+                              "pattern": "^\\$",
                             },
                             "type": "object",
                           },
--- a/plugins/catalog-react/src/alpha/blueprints/EntityContextMenuItemBlueprint.test.tsx
+++ b/plugins/catalog-react/src/alpha/blueprints/EntityContextMenuItemBlueprint.test.tsx
@@ -133,7 +133,7 @@ describe('EntityContextMenuItemBlueprint', () => {
                           "not": {},
                         },
                         "propertyNames": {
-                          "pattern": "(?!\\$)+",
+                          "pattern": "^\\$",
                         },
                         "type": "object",
                       },


plugins/catalog-react/src/alpha/predicates/types.ts
```diff
@@ -18,6 +18,7 @@
 export type EntityPredicate =
   | EntityPredicateExpression
   | EntityPredicatePrimitive
+  | EntityPredicatePrimitive[]
```
freben: I am not sure about this one. Help me think here - if we actually allowed this, what are the semantics of having an array in this position? It seems to me that we should rather remove it consistently (keep the type, update the runtime).

For example, what are the semantics of matching the value `[1, 2]` against the predicate `[2]`? What does that even mean?

What does it mean to match the value `[1, 2, 3]` against the predicate `{ $contains: [1, 2] }`? Should it be full or partial? Is it `all` or `any`? If it's partial, then why not use the existing primitives like `{ $all: [{ $contains: 1 }, { $contains: 2 }] }` so it's clear which one it is?

freben: Alright, we spoke a bit. In summary,

* `EntityPredicate` should not be possible to have array type
* `$contains: X` should not allow `X` to be an array either

Do you think you could update the types and code to achieve that?

karpfslp: Yes, I do. I can. In this case the Zod validator have to be updated in `createEntityPredicateSchema.ts` as well. I am on it...

karpfslp: You can check this PR again.

I made Zod and Typescript types in sync with each other.

I had to put a `never[]` always empty array type to here according to the `createEntityPredicateSchema.test.ts`.

`$contains` may have only `EntityPredicateExpression` and `string` values. This was my original plan as well before I saw the related schema test file.

plugins/catalog-react/src/alpha/predicates/types.ts
```diff
@@ -18,6 +18,7 @@
 export type EntityPredicate =
   | EntityPredicateExpression
   | EntityPredicatePrimitive
+  | never[]
```
freben: remove

plugins/catalog-react/src/alpha/predicates/types.ts
```diff
@@ -32,9 +33,10 @@ export type EntityPredicateExpression = {
 /** @alpha */
 export type EntityPredicateValue =
   | EntityPredicatePrimitive
+  | EntityPredicatePrimitive[]
```
freben: remove

plugins/catalog-react/src/alpha/predicates/types.ts
```diff
@@ -32,9 +33,10 @@ export type EntityPredicateExpression = {
...
   | { $exists: boolean }
   | { $in: EntityPredicatePrimitive[] }
-  | { $contains: EntityPredicateExpression };
+  | { $contains: EntityPredicateExpression | string };
```
freben:
```suggestion
  | { $contains: EntityPredicate };
```

plugins/catalog-react/src/alpha/predicates/types.ts
karpfslp: Done

plugins/catalog-react/src/alpha/predicates/types.ts
karpfslp: Done

plugins/catalog-react/src/alpha/predicates/types.ts
karpfslp: Done

karpfslp: The Zod validator and its test/test cases were updated/repaired as well.

plugins/catalog-react/src/alpha/predicates/createEntityPredicateSchema.ts
```diff
@@ -14,33 +14,44 @@
...
+    z.boolean(),
+  ]) as ZodType<EntityPredicatePrimitive>;
 
   // eslint-disable-next-line prefer-const
   let valuePredicateSchema: ZodType<EntityPredicateValue>;
 
+  const expressionSchema = z.lazy(() =>
+    z.union([
+      z.record(z.string().regex(/^(?!\$).*$/), valuePredicateSchema),
+      z.record(z.string().regex(/(?!\$)+/), z.never()),
```
freben:
```suggestion
      z.record(z.string().regex(/^\$/), z.never()),
```

you meant something like this, right? The given regex does not match strings that start with `$`

karpfslp: You are right. Repaired.

plugins/catalog-react/src/alpha/predicates/types.ts
karpfslp: Please, double check you suggestion. You wrote that the $all, $any etc. constraints were not allowed in $contains. I would go with this. What do you think?

```ts
  | { $contains: EntityPredicateExpression | EntityPredicatePrimitive };
```

freben: I didn't mean quite that - I meant that in _that specific example_ it was comparing the array elements one by one, and trying to match each in an impossible way (it would have to be equal to two things at once).

If you have an array

```yaml
arr:
  - one: 1
    is: valid
  - two: 2
    is: valid
```

it's fine to match

```yaml
arr:
  $contains:
    $all:
      - one: 1
      - is: valid
```

to make a contrived example. It looks for a single element, for which all of the following are true: ... etc

freben (approved):

plugins/catalog-react/src/alpha/predicates/types.ts
karpfslp: Oh, I see, thx your explanation.

Issue #31096, opened 2025-09-09
🚀 Feature: Support publish events on external systems


xoanmi:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
"events external"
```

### 🗃️ Project Area

Events System

### 🔖 Need

Support for external event systems like Kafka, RabbitMQ, AWS EventBridge, AWS SNS, and others.

### 📝 Proposal

Currently, the Events System relies on an internal event bus. Supporting external systems would be beneficial to enhance scalability and make it easier for external services to consume the events generated by backstage.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

No, but I'm happy to collaborate on a PR with someone else

Jonas-Beck:
We are currently using an internal version of the Kafka events module plugin that supports defining Backstage topics that should be published to Kafka topics, similar to the Google Pub/Sub plugin.

I'm planning to create a PR that would add this to the Kafka events module

camilaibs:
Hi @Jonas-Beck, your contribution is greatly appreciated 🙏🏻. Would you like me to assign you the task 🙂?

Jonas-Beck:
Hi @camilaibs, yes, you can assign this to me. I plan on getting the PR created in our next sprint.

Jonas-Beck:
Initial support for publishing events to Kafka has been added in the 1.46.0 Backstage release in v0.2.0 of the Kafka events module. The implementation was added in #32078.

@camilaibs should we just keep this issue open for potential RabbitMQ, AWS EventBridge, AWS SNS etc. contributions? They are not something I have used before, so not something I think I can contribute right now.

aramissennyeydd:
@Jonas-Beck Thanks for the contribution - yup, let's keep this open for the other modules!

Issue #31020, opened 2025-08-29
events-backend-module-kafka: Support multiple instances


Roshick:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
"kafka", "events", "multiple", "configuration"
```

### 🗃️ Project Area

Events System

### 🔖 Need

As of right now, the plugin only supports the configuration of a single Kafka consumer configuration. In our company, we have multiple Kafka and EventHub instances which we want to integrate.

### 📝 Proposal

Support multiple instances of the kafkaConsumingEventPublisher (https://github.com/backstage/backstage/blob/master/plugins/events-backend-module-kafka/config.d.ts#L29).

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Undecided

freben:
Sure - but that would entail copy-pasting the config and making it into `T | Array<T>` basically (and handling that when reading the config of course), it sounds like. Cumbersome, but doable.

Jonas-Beck:
Would it make sense to update the config to be `[name]: T` instead of `T | Array<T>`? This would allow us to include the name in logs and avoid some duplicate config in `config.d.ts` while also keeping the config reading logic simpler.

The only issue is that it would be a breaking change, but it would be a pretty simple upgrade and the plugin is still in a version 0.1.5 after all

I have a local version of the plugin that I'm testing internally where I added support for publishing Backstage events to Kafka topics and also changed the config to support multiple instances using `[name]: T`. I was planning on creating a PR with these changes

Jonas-Beck:
Support for multiple instances has been added in the 1.46.0 Backstage release in v0.2.0 of the kafka events module. The implementation was added in #32078

PR #32078, opened 2025-12-09
feat(events): add support for publishing to kafka and multiple instances

Jonas-Beck:

This PR updates the events Kafka module to support publishing internal events to Kafka topics, as requested in https://github.com/backstage/backstage/issues/31096.

This PR also updates the `kafkaConsumingEventPublisher` configuration to support defining multiple Kafka instances, as requested in https://github.com/backstage/backstage/issues/31020. This was implemented by updating the config to accept a `[name: string]: T` structure instead of a single config object. This will be a breaking change to the config but will make it much easier for users to debug issues with multiple configs, as you can now identify each instance by name.

I have done some refactoring to the internals of the module with the introduction of `kafkaPublishingEventConsumer` so it follows the structure of the Google Pub/Sub module. I think this will make it easier to maintain and test going forward.

This is something we have been using internally for a few weeks now.

Currently, there are still some features missing from the publishing part that we didn't have a use case for internally yet, and I had some open questions related to the best way to implement them. These features include Kafka key and headers support. I was thinking it would be best to include the current implementation as is and add key and header support in a future version

### DIFF
--- a/.changeset/angry-views-win.md
+++ b/.changeset/angry-views-win.md
@@ -2,22 +2,4 @@
 '@backstage/plugin-events-backend-module-kafka': minor
 ---
 
-**BREAKING**: Updated `kafkaConsumingEventPublisher` configuration to support multiple named instances
-
-The Kafka configuration now requires named instances instead of a single configuration object for `kafkaConsumingEventPublisher`, this allows for multiple Kafka configurations.
-
-These changes are **required** to your `app-config.yaml`:
-
-```diff
-events:
-  modules:
-    kafka:
-      kafkaConsumingEventPublisher:
--       clientId: your-client-id
--       brokers: [...]
--       topics: [...]
-+       default:  # Or any name like 'prod', 'dev', etc.
-+         clientId: your-client-id
-+         brokers: [...]
-+         topics: [...]
-```
+Added support for multiple named instances in `kafkaConsumingEventPublisher` configuration. The previous single configuration format is still supported for backward compatibility.
--- a/plugins/events-backend-module-kafka/config.d.ts
+++ b/plugins/events-backend-module-kafka/config.d.ts
@@ -25,184 +25,359 @@ export interface Config {
         /**
          * Configuration for KafkaConsumingEventPublisher
          *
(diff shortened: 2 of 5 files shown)


.changeset/angry-views-win.md
```diff
@@ -0,0 +1,23 @@
+---
+'@backstage/plugin-events-backend-module-kafka': minor
+---
+
+**BREAKING**: Updated `kafkaConsumingEventPublisher` configuration to support multiple named instances
```
freben: It pains me a little to make this hard-breaking. A heart-breaking hard-breaking, if you will.

Some modules declare their configs to support both single and multiple entries - like effectively `T | { [key: string]: T }` but expanded into its concrete form. Lots of repetition though - especially for this complex of a config.

Maybe we could make it breaking in the sense that the config schema looks forced as the new format, but the code has a "defensive" backwards compatibility path that looks for `...kafkaConsumingEventPublisher.clientId` and if it finds that, logs a warning but continues reading that single one (probably giving it a fallback ID, like "default").

plugins/events-backend-module-kafka/src/KafkaPublishingEventConsumer/module.test.ts
```diff
@@ -0,0 +1,122 @@
...
+                      ],
+                    },
+                  },
+                },
+              },
+            },
+          },
+        }),
+      ],
+    });
```
freben, line 80: It's LISP all over again

Jonas-Beck:
Hi @freben 

Thanks for the quick feedback - I will get started configuring a fallback value for the config reader, so we can avoid the breaking change

.changeset/angry-views-win.md
Jonas-Beck: Makes sense. I have updated the PR so that it will still support the old format by checking if the `clientId` is present and then using `default` as the fallback ID. I also added an additional test case for the old format and updated the changeset to reflect the changes.

Would this implementation work? I was initially a little bit unsure how to best implement it, which is why I guess I tried to take the easy way out and went for the hard-breaking change initially.

PR #32080, opened 2025-12-09
skip the very first search on opening the app

freben:
Fixes https://github.com/backstage/backstage/issues/31695

We could take this further, and never send any requests at all when there's no input - not just on first mount. Let me know if that's preferable to you. I opted for the smallest possible effect for now.

### DIFF
--- a/.changeset/many-falcons-joke.md
+++ b/.changeset/many-falcons-joke.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-search-react': patch
+---
+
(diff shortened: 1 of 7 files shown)


xoanmi (approved):

alexlorenzi:
Heck Yeah!
I think only stopping the first request is good for now. This has been such a long standing issue, that I wouldn't be surprised if we uncover someone/thing relies of this empty search (I can't think of anything, but who knows out in the wild)

PR #32099, opened 2025-12-11
`integrations`: Add flag for using an app for `publicAccess`

benjdlambert:
The default integrations setup right now is that if an app cannot generate a token for and organization it will throw.

This means that in order to read things instance wide, if that's `GHE` or public GitHub or others, the app needs to be installed in every location that it plans to read information from.

This is a non-starter in most organizations, so this allows the fact that once an app has an installation, the token that it generates will be valid for read-only public repositories.

This can be enabled using the `publicAccess` flag in the `apps` config.

### DIFF
--- a/packages/integration/report.api.md
+++ b/packages/integration/report.api.md
@@ -665,6 +665,7 @@ export type GithubAppConfig = {
   clientId: string;
   clientSecret: string;
   allowedInstallationOwners?: string[];
+  publicAccess?: boolean;
 };
 
 // @public


packages/integration/src/github/SingleInstanceGithubCredentialsProvider.ts
```diff
@@ -170,6 +172,30 @@ class GithubAppManager {
...
 
+  async getPublicInstallationToken(): Promise<{ accessToken: string }> {
+    const [installation] = await this.getInstallations();
+
+    if (!installation) {
+      throw new Error(`No installation found for public app`);
+    }
+
+    return this.cache.getOrCreateToken(
+      `public:${installation.id}`,
```
benjdlambert, line 183: Thought that this made sense rather than it overwriting the cached token for the org / repo. 

That way if the app does get installed later then it should use a write token, even if this token is still not expired.

freben (approved):

PR #32100, opened 2025-12-11
Add support for persisted feature flags

grantila:

_This is a follow-up and re-implentation of https://github.com/backstage/community-plugins/pull/6362. Initial rationale can be found there._

This PR adds support for feature flags to be _persisted_, i.e. stored using the StorageApi. An implementer can choose certain (or all) feature flags to be persisted in the backend, such as functionality, features, etc. Some feature flags may make more sense to store per-browser, leaving this choice to the implementer.

The API surface has been changed (with backwards compatibility) to be async by nature. An initial call to query all StorageApi-based feature flags on start-up is made, to allow for the old (synchronous) `isActive(flag)` to still work. The new API also deprecates the `save()` method, which will ultimately hide the `FeatureFlagState` type (a storage type of `0` or `1`, which is a technical detail that shouldn't be needed to be exposed).

A hook `useFeatureFlag(flag)` is introduced to simplify usage in components. It listenes to observable changes.

Both the _local_ (in-browser) feature flags and StorageApi based feature flags now emit browser-level events (using [BroadcastChannel](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API)) when updated, so flipping a flag will live-update any `useFeatureFlag` hooks in the same tab and other browser tabs. If the SignalApi is enabled for the StorageApi, it will naturally also affect other browser sessions too.

Given the comment from @Rugvip in https://github.com/backstage/community-plugins/pull/6362, this PR could act as an initial support for persistency using the StorageApi, while future work to incorporate external sources could be added later, once the async behavior in this PR is in place.

This PR would greatly benefit from #31106, but not require it.

### DIFF
--- a/.changeset/chubby-gifts-fry.md
+++ b/.changeset/chubby-gifts-fry.md
@@ -1,0 +1,5 @@
+---
+'@backstage/test-utils': patch
+---
+
+Add `BroadcastChannel` to `globalThis` to allow for frontend logic to use it
--- a/.changeset/five-years-talk.md
+++ b/.changeset/five-years-talk.md
@@ -1,0 +1,10 @@
+---
+'@backstage/app-defaults': patch
+'@backstage/frontend-plugin-api': patch
+'@backstage/frontend-app-api': patch
+'@backstage/core-plugin-api': patch
+'@backstage/core-app-api': patch
+'@backstage/plugin-user-settings': patch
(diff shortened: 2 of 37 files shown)


schultzp2020:
@grantila can you please rebase? 🙏

PR #32105, opened 2025-12-11
feat(scaffolder): add RepoOwnerPicker

benjidotsh:
This PR adds a `RepoOwnerPicker` field extension to the scaffolder that allows you to select a repository owner, independently from the existing `RepoUrlPicker.

Scherm­afbeelding 2025-12-11 om 17 08 29

I'm not sure if this is something we actually want to add, but we needed it, so I figured: why not contribute it back? 😄 

In our use case, this would be useful for selecting a GitHub organization to assign a GitHub Copilot license from. Since we're using GitHub Enterprise Cloud, every area has their own organization.

For now only GitHub is supported, but I don't mind adding support for Bitbucket. I'm just not sure what would be the Bitbucket equivalent for owners or GitHub organizations... Workspaces I guess? Maybe projects?

### DIFF
--- a/.changeset/odd-mirrors-attack.md
+++ b/.changeset/odd-mirrors-attack.md
@@ -1,5 +1,5 @@
 ---
-'@backstage/plugin-scaffolder-backend-module-github': minor
+'@backstage/plugin-scaffolder-backend-module-github': patch
 ---
 
 Added autocompletion resource `owners` for retrieving GitHub repository owners.


docs/features/software-templates/ui-options-examples.md
```diff
@@ -468,3 +468,61 @@ repoUrl:
...
+
+If defined will request user credentials to auth against the given SCM platform.
+
+```yaml
+repoUrl:
+  title: Repository Owner
+  type: string
+  ui:field: RepoOwnerPicker
+  ui:options:
+    requestUserCredentials:
```
benjdlambert, line 517: Hmm, I think that I wanna deprecate this approach soon and rely on `formDecorators` to grab secrets and have a secret schema for each job.

That said, don't think we're gonna get the time to do that any time soon, so I'm happy to keep this for now as backwards compatibility with `RepoURLPicker`.

docs/features/software-templates/ui-options-examples.md
```diff
@@ -468,3 +468,61 @@ repoUrl:
...
+```yaml
+repoUrl:
+  title: Repository Owner
+  type: string
+  ui:field: RepoOwnerPicker
+  ui:options:
+    host: github.com
+```
+
+This value will be overriden if a different host is specified in a `RepoUrlPicker` field extension.
```
benjdlambert:
```suggestion
This value will be overridden if a different host is specified in a `RepoOwnerPicker` field extension.
```

plugins/scaffolder-backend-module-github/src/autocomplete/autocomplete.ts
```diff
@@ -64,6 +64,15 @@ export function createHandleAutocompleteRequest(options: {
 
         return { results };
       }
+      case 'owners': {
+        const branches = await client.paginate(
```
benjdlambert:
```suggestion
        const orgs = await client.paginate(
```

benjdlambert (changes requested):
Thanks for this!

We actually have something internally like this, so think it makes sense! 

Just a few high level comments first! :pray:

plugins/scaffolder-backend-module-github/src/autocomplete/autocomplete.ts
benjidotsh: 🙈

docs/features/software-templates/ui-options-examples.md
benjidotsh: `overridden` is definitely correct, but so is `RepoUrlPicker` in this case - if there is a `RepoUrlPicker` and it has a set host, this one will take precedence over the host defined in `RepoOwnerPicker`

plugins/scaffolder/src/components/fields/RepoOwnerPicker/RepoOwnerPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+      // set the secret using the key provided in the ui:options for use
+      // in the templating the manifest with ${{ secrets[secretsKey] }}
+      setSecrets({ [requestUserCredentials.secretsKey]: token });
+    },
+    500,
+    [host, uiSchema],
+  );
+
+  useEffect(() => {
+    if (repoUrl) {
```
benjdlambert: Hmm, if we remove this entire logic is that a problem?

This makes assumptions about property names in the form and introduces a little bit of magic coupling which we don't have precedent for, and I'm not 100% convinced that this is a good idea.

Should we just do it so that it defaults to `github.com` unless you provide a `ui:option` host and just have it so that you have to provide a host override if you're using a different DNS?

I think that duplication in this sense is better than introducing this coupling?

docs/features/software-templates/ui-options-examples.md
benjdlambert: Ah sorry I misunderstood the behavior, see the most recent comment: https://github.com/backstage/backstage/pull/32105#discussion_r2622081873

PR #32109, opened 2025-12-11
'add square shape'

SonilPro:

- Added additional square shape. Taken from the official [Backstage figma](https://www.figma.com/@backstage)
image

### DIFF
--- a/.changeset/grumpy-birds-teach.md
+++ b/.changeset/grumpy-birds-teach.md
@@ -2,4 +2,4 @@
 '@backstage/theme': patch
 ---
 
-add square shape + improve typescript types
+add square shape


freben:
Maybe a screenshot!

SonilPro:
@freben You are fast, just edited the description 😄

packages/theme/src/base/pageTheme.ts
```diff
@@ -32,18 +32,19 @@ import { PageTheme } from './types';
  * 4. Wrap the output in `url("")`
  * 5. Give it a name and paste it into the `shapes` object below.
  */
-export const shapes: Record<string, string> = {
```
SonilPro: We can revert this change, but I added it because records break the typescript intellisense, which means you can't see all the keys when using the objects

SonilPro: Reverted this back. Noticed it breaks some other code, and presumably other integrators code

awanlin:
FYI - Just noting that this will be going away as we move to BUI and its header, more details here: https://github.com/backstage/backstage/issues/31412 and here: https://ui.backstage.io/components/header

SonilPro:
@awanlin aww man, I liked the colors and shapes in the old system. It really allowed for easier differentiation between entity types :disappointed: 
It also fit quite well with the "Create" page. It would match the colors. Have you guys thought about the entity cards throughout other parts of Backstage?


PR is up to you @freben then. I am fine with merge or delete.

awanlin:
My comment was more about awareness, I don't have an opinion either way regarding the PR itself, the title drew me in as I wanted to know what the change was about at which point I figured I should share the details that I did. 👍

SonilPro:
@awanlin I meant the design approach to the new BUI Header. Not sure if you are working on it. I can leave a comment on the issue directly.

awanlin:
I was only replying to this in my last message:

> PR is up to you @freben then. I am fine with merge or delete.

PR #32121, opened 2025-12-13
Plugins - Added plugin directory validation

awanlin:

This adds a script that includes a `zod` schema to validate plugin files that drive the plugin directory. This makes it easier for us to validate the files are correct and for people when submitting new plugins to make sure all the data is correct in a clear way. We run the schema validation in the CI now as well.

The script also adds a check to make sure all files end in `.yaml`, this has often been missed in the past.

Example output when there are errors, this shows a file missing the `.yaml extension, a plugin with a single error and a plugin with multiple errors:

Screenshot 2026-01-27 at 7 10 35 PM

### DIFF
--- a/scripts/verify-plugin-directory.js
+++ b/scripts/verify-plugin-directory.js
@@ -41,15 +41,17 @@ async function main() {
 
   let hasErrors = false;
   for (const pluginDirectoryFile of pluginDirectoryFiles) {
-    if (!pluginDirectoryFile.toLocaleLowerCase().includes('.yaml')) {
-      hasErrors = true;
-      console.log(`${pluginDirectoryFile} is missing the '.yaml' extension`);
-    }
-
     const pluginDirectoryFilePath = join(
       pluginDirectoryPath,
       pluginDirectoryFile,
     );
+
+    if (!pluginDirectoryFile.toLocaleLowerCase().includes('.yaml')) {
+      hasErrors = true;
+      console.log(
+        `${pluginDirectoryFilePath}: The '.yaml' extension is missing`,
+      );
+    }
 
     const pluginDataYaml = yaml.load(
       fs.readFileSync(pluginDirectoryFilePath, { encoding: 'utf-8' }),
@@ -60,12 +62,12 @@ async function main() {
     } catch (error) {
       if (error instanceof z.ZodError) {
         console.log(
-          `YAML data validation failed for ${pluginDirectoryFile}:`,
+          `${pluginDirectoryFilePath}: YAML data validation failed`,
           error.errors,
         );
       } else {
         console.log(
-          `An unexpected error occurred for ${pluginDirectoryFile}:`,
+          `${pluginDirectoryFilePath}: An unexpected error occurred`,
           error,
         );
       }
@@ -78,7 +80,7 @@ async function main() {
   }
 }
 
-main(process.argv.slice(2)).catch(error => {
+main().catch(error => {
   console.error(error.stack || error);
   process.exit(1);
 });


microsite/data/plugin-schema.json
```diff
@@ -0,0 +1,58 @@
...
+                "documentation": {
+                    "type": "string",
+                    "format": "uri"
+                },
+                "iconUrl": {
+                    "type": "string"
+                },
+                "npmPackageName": {
+                    "type": "string",
+                    "pattern": "/^(@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/"
```
awanlin: @aramissennyeydd, any ideas why it doesn't like this pattern? I've tested the regex and it works but always seems to fail using `yarn ajv validate -s microsite/data/plugin-schema.json -d "microsite/data/plugins/*.yaml" -c ajv-formats` 🤔

freben: When you have a `-` inside a `[]` it's intended as a range - if you want to match on a _literal_ dash, you need it at the beginning or end of the group, like `[a-z0-9~-]` instead of `[a-z0-9-~]`

freben: Also do you want a `...$/i` (note the i at the end) to make it case insensitive? Not sure if that's possible to publish in practice or not.

Alternatively use `\w` as a synonym for letters and digits, giving you `[\w~-]`

awanlin: Tried all those and still had the same issue, gonna remove the pattern for now.

docs/plugins/add-to-directory.md
```diff
@@ -26,11 +23,22 @@ npmPackageName: # Your npm package name E.g. '@backstage/plugin-<etc>' quotes ar
...
 ```
 
+:::tip
+
+You can validate your YAML file is correct by running the following from the root of the repo:
+
+1. First run `yarn install`
+2. Then run `yarn ajv validate -s microsite/data/plugin-schema.json -d "microsite/data/plugins/<plugin-name>.yaml" -c ajv-formats` where `<plugin-name>` is the name of your plugin
+
+If there are any errors they will be listed and you will need to correct them. We run this came check as part of the CI.
```
aramissennyeydd:
```suggestion
If there are any errors they will be listed and you will need to correct them. We run this same check as part of the CI.
```

docs/plugins/add-to-directory.md
```diff
@@ -26,11 +23,22 @@ npmPackageName: # Your npm package name E.g. '@backstage/plugin-<etc>' quotes ar
...
+
+:::
+
 ## Submission Tips
 
 Here are a few tips to help speed up the review process when you submit your plugin:
 
-- For any icon that you use make sure you have the proper rights to use it.
+- For any icon that you use make sure you have the proper rights to use it. If you don't have an icon then feel free to use `iconUrl: '/img/logo-gradient-on-dark.svg'`
```
aramissennyeydd: Should we make this the default value so people don't have to specify it? In JSON schema, I think we'd just mark it as optional.

aramissennyeydd (approved):
Nice!

docs/plugins/add-to-directory.md
awanlin: I just removed it from the required array, that should do the trick I think.

.github/workflows/ci.yml
```diff
@@ -121,6 +121,20 @@ jobs:
       - name: verify doc links
         run: node scripts/verify-links.js
 
+      - name: verify plugin directory files
+        run: |
+          if find ./microsite/data/plugins -maxdepth 1 -type f ! \( -name "*.yaml" \) | grep -q .; then
```
freben: would be nicer to assign it to a var like

```shell
WRONG_EXTENSION="$(find ...)"
if ! [ -z "$WRONG_EXTENSION" ]; then
```

and print its contents.

No need to print anything on success, I'd say

microsite/data/plugins/argo-cd.yaml
```diff
@@ -7,7 +7,4 @@ description: View Argo CD status for your projects in Backstage.
 documentation: https://roadie.io/backstage/plugins/argo-cd/?utm_source=backstage.io&utm_medium=marketplace&utm_campaign=argo-cd
 iconUrl: https://roadie.io/images/logos/argo.png
 npmPackageName: '@roadiehq/backstage-plugin-argo-cd'
-tags:
-  - cd
-  - ci
```
freben, line 12: oh so we actually don't support tags?

.github/workflows/ci.yml
```diff
@@ -121,6 +121,20 @@ jobs:
...
+              find ./microsite/data/plugins -maxdepth 1 -type f ! \( -name "*.yaml" \)
+              echo "Error: Non-YAML files found"
+              exit 1
+          else
+              echo "Success: All files in are YAML files (based on extension)."
+              exit 0
+          fi
+
+      - name: verify plugin directory schema
+        run: yarn ajv validate -s microsite/data/plugin-schema.json -d "microsite/data/plugins/*.yaml" -c ajv-formats
```
freben: Maybe these two new checks could have been a good fit for a standalone node script in the scripts directory. Then you just run that and it checks all files with no additional parameters given at all (not just your new file), really simplifying the instructions below. It could be using zod then too.

```yaml
      ...
      - name: verify doc links
        run: node scripts/verify-links.js

      - name: verify plugins directory
        run: node scripts/verify-plugins-directory.js
      ...
```

.github/workflows/ci.yml
awanlin: Moved to a script so not relevant anymore.

microsite/data/plugins/argo-cd.yaml
awanlin: Well we don't actually show them in the Plugin Directory at all and honestly trying to keep this consistent would be really hard. We have challenges with the sprawl of categories right now. It made sense to just remove this.

.github/workflows/ci.yml
awanlin: Yes, agree. Move to script using `zod` and streamlined this to calling the new script.

awanlin:
Hi @freben and @aramissennyeydd, can you give this another look? I moved everything to a script using `zod` to streamline this. Feels a lot better IMO!

scripts/verify-plugin-directory.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }
+  if (hasErrors) {
+    throw new Error(
+      'There was an error verifying the Plugin Directory data files, review the output for guidance on what corrections are needed',
+    );
+  }
+}
+
+main(process.argv.slice(2)).catch(error => {
```
freben: Not really using the argv though.

scripts/verify-plugin-directory.js
```diff
@@ -0,0 +1,84 @@
...
+    const pluginDataYaml = yaml.load(
+      fs.readFileSync(pluginDirectoryFilePath, { encoding: 'utf-8' }),
+    );
+
+    try {
+      configSchema.parse(pluginDataYaml);
+    } catch (error) {
+      if (error instanceof z.ZodError) {
+        console.log(
+          `YAML data validation failed for ${pluginDirectoryFile}:`,
```
freben: Could perhaps flip this around (same in the others): `${pluginDirectoryFilePath}: Rest of message here` - this makes the full path clickable in the terminal of vscode and the context is the first thing you see. The one at the top that says it lacks the yaml extension could do the same too (and therefore also forming the full path first thing you do at the top)

scripts/verify-plugin-directory.js
awanlin: Removed, thanks.

scripts/verify-plugin-directory.js
awanlin: Adjusted, good call!

PR #32123, opened 2025-12-14
BUI - Fixes startCollapsed prop in SearchField

cdedreuille:

The `startCollapsed` prop on the SearchField component was not working correctly due to multiple issues:

1. **Incorrect state initialization**: The collapsed state wasn't being properly initialized from the processed `startCollapsed` prop value
2. **Inverted CSS logic**: The CSS had backwards logic where `data-collapsed='true'` made the field expanded (200px) and `data-collapsed='false'` made it collapsed (2rem)
3. **Missing focus behavior**: Clicking on a collapsed field didn't properly focus the input, preventing the expand/collapse cycle from working
4. **Layout context issues**: The field would expand to full width in non-flex-row layouts (flex column, regular divs) even when collapsed

### Changes
- Fixed state initialization to use the processed `startCollapsed` value directly instead of using raw props
- Corrected the expand/collapse logic to match the fixed CSS (no longer inverted)
- Added `useRef` to track the input element
- Modified `handleContainerClick` to programmatically focus the input when expanding from collapsed state
- This ensures the blur event fires correctly when clicking away, allowing the field to collapse back

### DIFF
--- a/packages/ui/src/components/SearchField/SearchField.tsx
+++ b/packages/ui/src/components/SearchField/SearchField.tsx
@@ -84,10 +84,6 @@ export const SearchField = forwardRef<HTMLDivElement, SearchFieldProps>(
       inputRef.current?.focus();
     };
 
-    const handleChange = (value: string) => {
-      props.onChange?.(value);
-    };
-
     const hasInputRef = !!inputRef.current;
     const hasValue = !!inputRef.current?.value;
 
@@ -102,10 +98,9 @@ export const SearchField = forwardRef<HTMLDivElement, SearchFieldProps>(
         aria-label={ariaLabel}
         aria-labelledby={ariaLabelledBy}
         data-collapsed={isCollapsed}
-        onFocusChange={handleFocusChange}
-        onChange={handleChange}
         style={style}
         {...rest}
+        onFocusChange={handleFocusChange}
         ref={ref}
       >
         <FieldLabel


.changeset/hot-colts-float.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Fixed SearchField `startCollapsed` prop not working correctly in Backstage UI. The field now properly starts in a collapsed state, expands when clicked and focused, and collapses back when unfocused with no input. Also fixed CSS logic to work correctly in all layout contexts (flex row, flex column, and regular containers).
```
Naycon, line 5: Make sure you add an `Affected components:` line to the changesets to make the automatic changelog parsing work.
```suggestion
Fixed SearchField `startCollapsed` prop not working correctly in Backstage UI. The field now properly starts in a collapsed state, expands when clicked and focused, and collapses back when unfocused with no input. Also fixed CSS logic to work correctly in all layout contexts (flex row, flex column, and regular containers).

Affected components: SearchField
```

packages/ui/src/components/SearchField/SearchField.tsx
```diff
@@ -79,13 +80,26 @@ export const SearchField = forwardRef<HTMLDivElement, SearchFieldProps>(
       props.onFocusChange?.(isFocused);
       if (shouldCollapse) {
         if (isFocused) {
-          setIsCollapsed(true);
```
Naycon, line 82: Nit, and not something you caused in your PR, but this handler feels a bit misnamed. We do this down below:
```
  onFocusChange={handleClick}
```
Maybe rename to `handleFocusChange`?

packages/ui/src/components/SearchField/SearchField.tsx
```diff
@@ -79,13 +80,26 @@ export const SearchField = forwardRef<HTMLDivElement, SearchFieldProps>(
...
+        } else {
+          // When blurring, collapse the field
+          setIsCollapsed(true);
         }
       }
     };
 
+    const handleContainerClick = () => {
+      // If the field is collapsed (small), expand it and focus the input
+      if (startCollapsed && isCollapsed) {
```
Naycon: Why should this logic only apply when you set `startCollapsed` to true? Don't you want this behavior when you started with the search field expanded?

Naycon: I feel like I'm misunderstanding what this does. 🤔

cdedreuille: You're right. This was actually not useful. I removed it.

packages/ui/src/components/SearchField/SearchField.tsx
cdedreuille: Thanks. I changed that to `handleFocusChange`

.changeset/hot-colts-float.md
cdedreuille: Fixed

Rugvip (approved):
:shipit:

PR #32130, opened 2025-12-15
Adds the Resource kind to the useGetEntities fallback options so that…

Parsifal-M:
… Resources are also returned by default on the group ownership card

Tested this locally with the Backstage repo and seems to work, I'm still not super familiar with how the new frontend system is wired up so please LMK if I have gone a stray.

This stems from a user asking me in our internal Backstage why they couldn't see their owned `Resources` and although I know you can provide a custom filter to the `EntityOwnershipCard`, we don't provide one and I would expect, if you return `Components`, `API` and `Systems` you should probably also return `Resources`

Thanks!

LMK if this should be a `minor` rather than `patch` if we're happy with the changes! 👍 

### DIFF
--- a/plugins/org/report-alpha.api.md
+++ b/plugins/org/report-alpha.api.md
@@ -110,14 +110,14 @@ const _default: OverridableFrontendPlugin<
       config: {
         initialRelationAggregation: 'direct' | 'aggregated' | undefined;
         showAggregateMembersToggle: boolean | undefined;
-        entityFilterKind: string[] | undefined;
+        ownedKinds: string[] | undefined;
         filter: EntityPredicate | undefined;
         type: 'content' | 'summary' | 'info' | undefined;
       };
       configInput: {
         showAggregateMembersToggle?: boolean | undefined;
-        entityFilterKind?: string[] | undefined;
         initialRelationAggregation?: 'direct' | 'aggregated' | undefined;
+        ownedKinds?: string[] | undefined;
         filter?: EntityPredicate | undefined;
         type?: 'content' | 'summary' | 'info' | undefined;
       };


aramissennyeydd:
Nice! A little worried about folks being able to opt-out in cases where teams own a huge number of resources. The old frontend should be fine, but for the new frontend can we plumb this through https://github.com/backstage/backstage/blob/51a4cd3b7621ba71152bb07a9fb5974ce90900cb/plugins/org/src/alpha.tsx#L64-L68 as well 🙏?

Parsifal-M:
One thing I am wondering here is that, and because I don't have much knowledge yet on the plumbing of the NFS, is this filter applied somewhere else via config or was it perhaps missed? 🤔 

Don't want to be reinventing the wheel 😆

aramissennyeydd (approved):
Nice!

plugins/org/src/alpha.tsx
```diff
@@ -65,6 +65,7 @@ const EntityOwnershipCard = EntityCardBlueprint.makeWithOverrides({
       initialRelationAggregation: z =>
         z.enum(['direct', 'aggregated']).optional(),
       showAggregateMembersToggle: z => z.boolean().optional(),
+      entityFilterKind: z => z.array(z.string()).optional(),
```
freben: Bikeshedding a little - what do you feel about the name `ownedKinds` instead?

plugins/org/src/components/Cards/OwnershipCard/OwnershipCard.test.tsx
```diff
@@ -44,6 +44,26 @@ const items = [
...
+      name: 'my-resource',
+    },
+    spec: {
+      type: 'database',
+    },
+    relations: [
+      {
+        type: 'ownedBy',
+        targetRef: 'group:default/my-team',
+        target: {
```
freben: This field normally isn't emitted by the service anymore.

plugins/org/src/alpha.tsx
Parsifal-M: Yeah that makes sense to me rolls off the tongue a bit nicer 😅

PR #32134, opened 2025-12-16
`ActionsRegistry`: Enable filtering of Actions from sources

benjdlambert:
This adds filtering to the `ActionsRegistry` config to filter out different actions from the defined `pluginSources`.

Not 100% sure about the attribute filtering, but maybe :shrug: Can also omit this if we're not feeling it's worth it.

### DIFF
--- a/packages/backend-defaults/src/alpha/entrypoints/actions/DefaultActionsService.ts
+++ b/packages/backend-defaults/src/alpha/entrypoints/actions/DefaultActionsService.ts
@@ -170,25 +170,21 @@ export class DefaultActionsService implements ActionsService {
     );
 
     return actions.filter(action => {
-      // If include rules exist, must match at least one
-      if (includeRules.length > 0) {
-        const included = includeRules.some(rule =>
-          this.matchesRule(action, rule),
-        );
-        if (!included) {
-          return false;
-        }
-      }
-
-      // Must not match any exclude rule
       const excluded = excludeRules.some(rule =>
         this.matchesRule(action, rule),
       );
+
       if (excluded) {
         return false;
       }
 
-      return true;
+      // If no include rules, include by default
+      if (includeRules.length === 0) {
+        return true;
+      }
+
+      // Must match at least one include rule
+      return includeRules.some(rule => this.matchesRule(action, rule));
     });
   }
 
@@ -198,10 +194,6 @@ export class DefaultActionsService implements ActionsService {
       Record<'destructive' | 'readOnly' | 'idempotent', boolean>
     >;
   }> {
-    if (!configArray) {
-      return [];
-    }
-
     return configArray.map(ruleConfig => {
       const idPattern = ruleConfig.getOptionalString('id');
       const attributesConfig = ruleConfig.getOptionalConfig('attributes');
--- a/packages/backend-defaults/src/alpha/entrypoints/actions/actionsServiceFactory.test.ts
+++ b/packages/backend-defaults/src/alpha/entrypoints/actions/actionsServiceFactory.test.ts
@@ -299,6 +299,74 @@ describe('actionsServiceFactory', () => {
         expect(actions.map(a => a.id)).toEqual(['my-plugin:get-entity']);
       });
 
+      it('should always apply exclude rules even when action matches include', async () => {
+        // This tests that exclude is checked FIRST and always wins,
+        // regardless of whether the action would match an include rule
+        const multipleActions: ActionsServiceAction[] = [
+          {
+            ...mockActionsDefinition,
+            id: 'my-plugin:safe-action',
+            name: 'safe-action',
+            attributes: {
+              destructive: false,
+              readOnly: true,
+              idempotent: true,
+            },
+          },
+          {
+            ...mockActionsDefinition,
+            id: 'my-plugin:dangerous-action',
+            name: 'dangerous-action',
+            attributes: {
+              destructive: true,
+              readOnly: false,
+              idempotent: false,
+            },
+          },
+        ];
+
+        server.use(
+          rest.get(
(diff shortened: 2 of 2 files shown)


Rugvip (commented):
Hmm, yep, but wondering if we also need the same for consumers of the actions registry. For example filtering out certain actions from the MCP server? It does work naturally with split deployments, but not with co-location.

A potential option could be to have a filter structure defines as part of the service, but not directly in config. Kinda similar to how the scheduler service reads schedule configurations. Then you'd be passing that filter structure to the actions registry when listing actions.

benjdlambert:
Hmm yeah I thought about this from the `mcp-actions` backend perspective, and was initially just going to introduce this config there, but feels like really that is just going to lead to plugins implementing filtering on top of the `ActionsRegistry` everywhere and it's going to get fragmented and just thought to do this as a simple first pass.

You're right thought that this is really only going to work for split deployments that want different configuration per plugin.

Should we look at adding this ontop of this PR or as a replacement for this effort? What does that look like?

```yaml
scaffolder:
  actionsRegistry:
    filter:
      ...

mcpActions:
  actionsRegistry:
    filter: 
     ...
```

Or should it be keyed by `pluginId` in the `backend.actions` config?

```yaml
backend:
  actions:
    # global filters for all consumers
    filter:
      ...
    pluginSources:
      - catalog
      - scaffolder
    pluginConfig:
      # filter config per plugin that consumes the registry
      scaffolder:
        filter:
          ...
      mcp-actions:
        filter:
          ... 
```

It's a shame that we don't have the `pluginId` in the `ActionsService` like we do the `ActionsRegistryService` otherwise we can just push that the client that reads from the `Registry` as a nice fix for all things inside the plugin. Maybe we should do that?

Rugvip:
Hmm, yep kinda happy to start with just a simple filtering at the actions registry level like this is suggesting, as long as the format allows us to also use it to potentially filter in other places further down the line. I don't think per-plugin filtering in the registry itself is the way to go tbh, doesn't feel like quite the right way to slice it. It a somewhat extreme example you might have a single plugin consuming the registry for two different usages, for example the scaffolder might use one set of actions for templates and a different set for something else.

I like @freben suggested structure btw, feels a bit more clear when it comes to how the filters are applied, and a bit more powerful too I think

freben:
Or, come to think of it - the predicate format with some defined input shape.

```
$not: { attributes.readOnly: false }
```

benjdlambert:
@freben @Rugvip updated. Decided to skip the predicate stuff for now until there's a standalone package for these filter utils rather than duplicating and didn't wanna bloat this PR with that work.

.changeset/action-filtering-feature.md
```diff
@@ -0,0 +1,33 @@
...
+  actions:
+    pluginSources:
+      - catalog
+      - scaffolder
+    filter:
+      include:
+        - id: 'catalog:*'
+          attributes:
+            destructive: false
+        - id: 'scaffolder:*'
```
Rugvip, line 22: Just to double check, you can also do something like this right?

```yaml
- attributes:
    destructive: false
```

packages/backend-defaults/src/alpha/entrypoints/actions/DefaultActionsService.ts
```diff
@@ -148,4 +150,114 @@ export class DefaultActionsService implements ActionsService {
...
+    );
+
+    return actions.filter(action => {
+      // If include rules exist, must match at least one
+      if (includeRules.length > 0) {
+        const included = includeRules.some(rule =>
+          this.matchesRule(action, rule),
+        );
+        if (!included) {
+          return false;
```
freben: I'm not sure about this logic. The way it's currently written it entirely throws away the exclude rules in some cases. And exclude should always win over include no matter what.

Should it instead be doing

- If any exclude rule matches, return false
- If no include rules exist, return true
- If any include rule matches, return true
- Return false

packages/backend-defaults/src/alpha/entrypoints/actions/DefaultActionsService.ts
```diff
@@ -148,4 +150,114 @@ export class DefaultActionsService implements ActionsService {
...
+    });
+  }
+
+  private parseFilterRules(configArray: Array<Config>): Array<{
+    idMatcher?: Minimatch;
+    attributes?: Partial<
+      Record<'destructive' | 'readOnly' | 'idempotent', boolean>
+    >;
+  }> {
+    if (!configArray) {
```
freben: not possible according to the type at least

packages/backend-defaults/config.d.ts
```diff
@@ -154,6 +154,104 @@ export interface Config {
...
+             * If specified, only match actions where destructive matches this value.
+             * Actions default to destructive: true if not explicitly set.
+             */
+            destructive?: boolean;
+
+            /**
+             * If specified, only match actions where readOnly matches this value.
+             * Actions default to readOnly: false if not explicitly set.
+             */
+            readOnly?: boolean;
```
freben, line 200: Is it readOnly with uppercase o in the specification? If so, that's a bit unfortunate, would have preferred to use `readonly` throughout consistently (that's sort of a standard in identifier cases)

benjdlambert: It's `readOnlyHint` in the MCP Spec.

benjdlambert: but it's `readOnly` in the `ActionsRegistry` implementation

.changeset/action-filtering-feature.md
benjdlambert: yes it does, there's an example on the exclude below without ID and just attributes.

PR #32136, opened 2025-12-16
chore: upgrade MSW from v1 to v2

Naycon:
Migrates the test infrastructure from MSW (Mock Service Worker) v1 to v2 across 46 packages.

## Summary

- **Scope:** 149 files changed across 46 packages
- **Impact:** Test infrastructure only - no production code changes

### Key Changes

**Syntax transformations:**
- `rest.get/post` → `http.get/post` + `HttpResponse`
- Handler signatures `(req, res, ctx)` → `({ request })`
- Response patterns `res(ctx.json(data))` → `HttpResponse.json(data)`

**Infrastructure updates:**
- Updated CLI plugin templates for new plugins

### Known Semantic Changes

A few tests were adapted to MSW v2's architectural constraints:

1. **Generic network errors:** MSW v2's `HttpResponse.error()` produces "Network error" instead of custom messages. Tests updated to match on behavior, not message strings.

2. **XMLHttpRequest:** MSW v2 cannot intercept jsdom's XHR in Node.js. One XHR test removed from jestRejectNetworkRequests.

3. **x509 cert tests:** Adapted to verify cert configuration and error handling rather than specific PEM error messages.

None of these affect production behavior - production code does not parse error message strings.

---

#### ✅ Checklist

- [x] All commits have a `Signed-off-by` line in the message


packages/test-utils/src/testUtils/msw/registerMswTestHooks.ts
```diff
@@ -20,11 +20,11 @@
  * @param worker - service worker
  */
 export function registerMswTestHooks(worker: {
-  listen: (t: any) => void;
+  listen: (options?: any) => void;
   close: () => void;
   resetHandlers: () => void;
 }) {
-  beforeAll(() => worker.listen({ onUnhandledRequest: 'error' }));
```
Rugvip, line 27: Hmm, is it possible to keep this behavior intact somehow? Wouldn't want to break existing tests that rely on this.

Thinking worst case we can perhaps detect whether it's a v1 or v2 worker and pass the option only when needed?

Btw, what's the new behavior for this in v2?

freben: Indeed, this one is meant to be polymorphic across 1 and 2. Its external types are "loose" by choice to achieve this. Can we keep the code as is and just ts ignore the flag if it's no longer supported in the v2 types? Whatever way is best to try to retain this and have the same outcome across both. I'd basically be up for an "as any" even if needed :)

Naycon: Oh this was an oversight on my part. There's no change in v2 here. Tested locally and it works fine so I'll just revert this change. I assumed this change came from the migration docs since there's a section on `onUnhandledRequest` there, but it's not at all related. Sorry about that!

Naycon: Reverted now.

alexbit-codemod:
awesome work @Naycon! did you use the [official MSW codemod](https://mswjs.io/docs/migrations/1.x-to-2.x/#codemods) for this migration? if so, how was your experience? would love to hear any feedback on how we could improve the codemod or the overall migration experience. 🍻

freben:
Full transparency we're [hitting a bit of a roadblock](https://github.com/mswjs/msw/issues/2591#issuecomment-3507851703) in regards to msw2 and need to see what we can do about that.

Naycon:
> awesome work @Naycon! did you use the [official MSW codemod](https://mswjs.io/docs/migrations/1.x-to-2.x/#codemods) for this migration? if so, how was your experience? would love to hear any feedback on how we could improve the codemod or the overall migration experience. 🍻

@alexbit-codemod Thanks! No, did not use any codemods. Claude Code pointed to the migration guide took care of the easy cases and helped investigate the more difficult ones. :)

alexbit-codemod:
@freben gotcha. if you ever decide to migrate away from jest, please let me know. would be happy to explore ways we can be helpful.

@Naycon wow! interesting! thanks for sharing.

PR #32140, opened 2025-12-16
BUI - Improve camel case props for useStyles

cdedreuille:

This solve the use case of having camel case props transforming into data attributes.

### DIFF
--- a/packages/ui/src/hooks/useStyles.ts
+++ b/packages/ui/src/hooks/useStyles.ts
@@ -86,17 +86,17 @@ export function useStyles<
       ? componentDefinition.utilityProps
       : []) || [];
 
-  // Extract data attribute names from component definition
-  const dataAttributeNames =
-    'dataAttributes' in componentDefinition
-      ? Object.keys(componentDefinition.dataAttributes || {})
-      : [];
-
   // Extract existing style from props
   const incomingStyle = props.style || {};
 
   // Generate data attributes from component definition
   const dataAttributes: Record<string, string> = useMemo(() => {
+    // Extract data attribute names from component definition
+    const dataAttributeNames =
+      'dataAttributes' in componentDefinition
+        ? Object.keys(componentDefinition.dataAttributes || {})
+        : [];
+
     const result: Record<string, string> = {};
     for (const key of dataAttributeNames) {
       const value = props[key];
@@ -114,7 +114,7 @@ export function useStyles<
       }
     }
     return result;
-  }, [dataAttributeNames, props, breakpoint]);
+  }, [componentDefinition, props, breakpoint]);
 
   // Generate utility classes and custom styles from component's allowed utility props
   const utilityClassList: string[] = [];


.changeset/clever-buses-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Fixes kamel case props on useStyles hook in Backstage UI as data attributes can't use kamel case.
```
freben:
```suggestion
Fixes camel case props on `useStyles` hook in Backstage UI as data attributes can't use camel case.
```

cdedreuille: I thought this was called kamel case, no?

cdedreuille: Ah you're right - Camel 🐪

packages/ui/src/hooks/useStyles.ts
```diff
@@ -17,6 +17,13 @@ import { useBreakpoint, breakpoints } from './useBreakpoint';
 import type { ComponentDefinition } from '../types';
 import { utilityClassMap } from '../utils/utilityClassMap';
 
+function toKebabCase(str: string): string {
+  return str
+    .replace(/([a-z0-9])([A-Z])/g, '$1-$2')
+    .replace(/_/g, '-')
+    .toLocaleLowerCase('en-US');
+}
+
```
Rugvip, line 27: Hmm, this is gonna be called a lot. I'm feeling it's better to have this fixed at the source instead to make sure we keep things performant. i.e. update https://github.com/backstage/backstage/blob/a67670d240c6c5599a61dd9d825a8941cdb69da0/packages/ui/src/components/SearchField/definition.ts#L32 etc

cdedreuille: Not sure I understand the issue you are seeing here. The prop is unchanged but we need props to be transformed to work as data attributes.

Naycon: I think @Rugvip means this is going to be called multiple times per render. The issue with fixing at the source, which I assume means changing to `'start-collapsed': [true, false] as const`, is that we want to be able to still do `props[key]` where `key` was `startCollapsed` before. Or am I misunderstanding what you spotted and what you suggested as a way forward Patrik?

I wonder if we can simply memoize the for-loop below if we're worried about performance? `const dataAttributes: Record<string, string> = useMemo(() => ...` or similar.

Naycon: Have to watch out for `dataAttributeNames` though, it will be recreated on each render so it needs to be computed inside the memo function.

cdedreuille: Thanks both. I updated the code to reflect the changes.

Rugvip: Yep exactly, would like to avoid the `toKebabCase` being called for each passed data attribute prop on render. Maybe the definition could be updated to something like this?

```ts
startCollapsed: {
  value: [true, false] as const,
  name: 'start-collapsed',
}
```

Although it is a bit clunky. A global cache would work to given it's a fixed transform, although even that feels a bit unnecessary. I don't think the memo does much because `props` will change on every render.

In the end I don't think this is that bad:

```ts
export const SearchFieldDefinition = {
  classNames: {
    root: 'bui-SearchField',
    clear: 'bui-SearchFieldClear',
    inputWrapper: 'bui-SearchFieldInputWrapper',
    input: 'bui-SearchFieldInput',
    inputIcon: 'bui-SearchFieldInputIcon',
  },
  dataAttributes: {
    startCollapsed: {
      values: [true, false] as const,
      name: 'start-collapsed',
    },
    size: {
      values: ['small', 'medium'] as const,
      name: 'size',
    },
  },
} as const satisfies ComponentDefinition;
```

PR #32141, opened 2025-12-16
Add dark contrast theme

elstarkov:

Added the opportunity to choose a dark contrast theme as default to improve accessibility.

Skärmavbild 2025-12-16 kl  16 35 53

### DIFF
--- a/docs/conf/user-interface/index.md
+++ b/docs/conf/user-interface/index.md
@@ -83,7 +83,7 @@ const app = createApp({
 | `Provider` | This is needed to set the legacy theme with MUI only. This will be become redundant later on when we fully replace with BUI but for now you need to have it for MUI to work. BUI is based on CSS and don't rely on any global providers. |
 
 :::note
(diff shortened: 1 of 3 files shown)


clairep94:
Hey nice! This is a great idea. 

Do we know if the current `dark` theme is not high-contrast enough? 
In which case, it may be better to update the `dark` theme to ensure that it is sufficiently high contrast for accessibility

I would defer to @cdedreuille on the colour choices.

I also wonder if an alternative could be to provide an optional package with your new theme that folks could choose to opt-in to, given design is quite subjective?

clairep94:
Also small nit if this is high-contrast enough:

Could we update accent colour to be the same colour as the backstage logo? I believe the hexcode is `#7df3e1`

Screenshot 2025-12-17 at 10 34 10

Screenshot 2025-12-17 at 11 23 54

cdedreuille:
I agree with @clairep94 that ideally we should fix our default dark theme instead of adding a new theme at this stage.

PR #32146, opened 2025-12-16
feat(auth): add audit log capability

beanview:

add innate audit log support into the auth-backend

logs success/failure for four pathways (login start, login completion, token refresh, logout)
logs the providerId used


freben (commented):
Hm. Is this less useful when it doesn't provide metadata about the actual user that for example signed in?

beanview:
@freben hmm that's a good point, let me try moving this logic to a different plugin to capture that

plugins/auth-node/src/audit.ts
```diff
@@ -0,0 +1,60 @@
...
+/** @internal */
+export type AuditEventMeta = {
+  providerId?: string;
+  actionType?: 'login' | 'token-refresh' | 'logout';
+  userEntityRef?: string;
+  email?: string;
+};
+
+/** @internal */
+export async function emitAuditEvent(
```
naman-spotify: Is this separate function needed? Wondering as we haven't used in other audit events elsewhere

plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
```diff
@@ -198,11 +206,23 @@ export function createOAuthRouteHandlers<TProfile>(
...
         const signInResult =
           signInResolver &&
           (await signInResolver({ profile, result }, resolverContext));
+        userEntityRef = signInResult?.identity?.userEntityRef;
+
+        await emitAuditEvent(auditor, {
+          eventId: 'auth-login',
+          request: req,
+          severityLevel: 'medium',
+          meta: { providerId, userEntityRef, email: profile?.email },
```
naman-spotify: We should already be getting the user entity ref from the request credentials

plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
```diff
@@ -198,11 +206,23 @@ export function createOAuthRouteHandlers<TProfile>(
...
 
         const signInResult =
           signInResolver &&
           (await signInResolver({ profile, result }, resolverContext));
+        userEntityRef = signInResult?.identity?.userEntityRef;
+
+        await emitAuditEvent(auditor, {
+          eventId: 'auth-login',
+          request: req,
+          severityLevel: 'medium',
```
naman-spotify: Auth login seems like it'll be a pretty common event if I'm understanding correctly, so maybe this should be low severity?

plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
```diff
@@ -291,6 +319,13 @@ export function createOAuthRouteHandlers<TProfile>(
       // remove persisted scopes
       await scopeManager.clear(req);
 
+      await emitAuditEvent(auditor, {
+        eventId: 'auth-logout',
+        request: req,
+        severityLevel: 'low',
+        meta: { providerId, actionType: 'logout' },
```
naman-spotify: The action type seems redundant here

plugins/auth-node/src/audit.ts
beanview: it's not, but it just condensed alot of repeated code i had across so I made it as a helper

plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
```diff
@@ -354,10 +389,29 @@ export function createOAuthRouteHandlers<TProfile>(
...
           response.backstageIdentity =
             prepareBackstageIdentityResponse(identity);
+
+          await emitAuditEvent(auditor, {
+            eventId: 'auth-login',
+            request: req,
+            severityLevel: 'low',
+            meta: {
+              providerId,
+              actionType: 'token-refresh',
```
naman-spotify: I'm wondering if this one is overkill to be auditing, maybe worth asking Seve on it. If we do keep it, maybe we should have a corresponding `actionType` in the `auth-login` event above so that we can differentiate the two

plugins/auth-node/src/proxy/createProxyRouteHandlers.ts
```diff
@@ -56,25 +69,48 @@ export function createProxyAuthRouteHandlers<TResult>(
...
-        providerInfo,
-        backstageIdentity: prepareBackstageIdentityResponse(identity),
-      };
+        await emitAuditEvent(auditor, {
+          eventId: 'auth-login',
+          request: req,
+          severityLevel: 'low',
+          meta: {
+            providerId,
+            actionType: 'token-refresh',
```
naman-spotify: Same note here on whether this one is overkill

plugins/auth-node/src/audit.ts
```diff
@@ -0,0 +1,60 @@
...
+      ...meta,
+      outcome: error ? 'failure' : 'success',
+    },
+  });
+
+  if (error) {
+    await auditEvent.fail({ error });
+  } else {
+    await auditEvent.success();
+  }
```
aramissennyeydd: Should foundational failures with the AuditService be blocking for the entire request?

beanview: def not, will remove awaits

plugins/auth-node/src/audit.ts
```diff
@@ -0,0 +1,63 @@
...
+  const { eventId, request, severityLevel, meta, error } = options;
+
+  (async () => {
+    const auditEvent = await auditor.createEvent({
+      eventId,
+      request,
+      severityLevel,
+      meta: {
+        ...meta,
+        outcome: error ? 'failure' : 'success',
```
freben: This [is already emitted](https://github.com/backstage/backstage/blob/e6248392fc27c02afe19ff0a974fa62ccf2ac4e8/packages/backend-defaults/src/entrypoints/auditor/DefaultAuditorService.ts#L191) as part of the outcome.

plugins/auth-node/src/audit.ts
```diff
@@ -0,0 +1,63 @@
...
+      },
+    });
+
+    if (error) {
+      await auditEvent.fail({ error });
```
freben: This removes the "timing aspect" of the audit logging. Is that OK? Or should this helper function be rewritten to accept a callback `fn: () => void | Promise<void>` and then we await that inside a try/catch here, after which we know whether to mark `fail` or `success`?

Or change the function to a `createAuditEvent` helper and then force the callers to deal with properly finishing them off. But given that we don't need the `outcome` as per the above comment - do we even need the helper at all?

I think it's already an expectation on audit system users that the need to do a two-part thing for events (when an attempt at a thing starts, and when it completes successfully or with an error)

beanview: hmmm that's true. I wrote it because i was writing the same two parts across multiple areas but I think it'd be simpler without the helper. I'll remove and update the actual files to create the auditEvent first and then emit after

plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
```diff
@@ -198,7 +203,18 @@ export function createOAuthRouteHandlers<TProfile>(
...
           authenticatorCtx,
         );
-        const { profile } = await profileTransform(result, resolverContext);
+        const { profile: resolvedProfile } = await profileTransform(
+          result,
+          resolverContext,
+        );
+        profile = resolvedProfile;
+
+        const auditEvent = await auditor?.createEvent({
```
freben, line 212: Any particular reason to make this one all the way down here? Would it not make sense to do this at the very top outside the `try`, once, and then handling that single event in all branches?

Note that `.success` and `.fail` accept an optional `meta` argument if you want to amend specific information such as the email.

It is probably sensible to have `grantedScopes` and the entity ref from `signInResult` be part of the meta too.

plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
```diff
@@ -291,6 +319,14 @@ export function createOAuthRouteHandlers<TProfile>(
       // remove persisted scopes
       await scopeManager.clear(req);
 
+      const auditEvent = await auditor?.createEvent({
```
freben, line 322: same here, create it as early as possible, wrap the rest of the code in a try/catch. We are highly interested in auditing failures that happen in here - especially if they happen to be part of an attacker trying stuff out.

freben (changes requested):
Hey! Sorry for the delay, some ideas here.

PR #32148, opened 2025-12-17
feat: schema-first audit events

aramissennyeydd:

Bit of an experiment to try and simplify how audit events are defined and keep the existing code clean instead of muddying it with either extra middleware at the route level or try/catch blocks in the application code itself.


packages/backend-openapi-utils/src/middlewares/auditor.ts
```diff
@@ -0,0 +1,220 @@
...
+      })
+      .catch(() => {
+        // If audit event creation fails, don't block the request
+      });
+
+    next(err);
+  };
+
+  return { success, error };
+}
```
schultzp2020:
```suggestion
}

```
EOF

packages/backend-openapi-utils/src/middlewares/auditor.ts
```diff
@@ -0,0 +1,220 @@
...
+type AuditorExtension = {
+  eventId: string;
+  severityLevel?: 'low' | 'medium' | 'high' | 'critical';
+  meta?: JsonObject & {
+    captureFromRequest?: {
+      body?: string[];
+      params?: string[];
+      query?: string[];
+    };
+    captureFromResponse?: {
```
schultzp2020: Is `captureFromResponse` being used anywhere right now? 🤔

packages/backend-openapi-utils/src/middlewares/auditor.ts
```diff
@@ -0,0 +1,220 @@
...
+          .createEvent({
+            eventId: auditorConfig.eventId,
+            severityLevel: auditorConfig.severityLevel,
+            meta: captureMetadata(),
+            request: req,
+          })
+          .then(auditorEvent => {
+            // Only mark as success for 2xx responses
+            // For errors, the handler should catch and call auditorEvent.fail()
+            if (res.statusCode >= 200 && res.statusCode < 300) {
```
schultzp2020: I believe this if statement is redundant because you check the status code on line 169.

packages/backend-openapi-utils/src/middlewares/auditor.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  const error = (err: Error, req: Request, res: Response, next: NextFunction) => {
+    const result = baseHandler(req, res);
+    if (!result) {
+      next(err);
+      return;
+    }
+    const { captureMetadata, auditorConfig } = result;
+    // Create audit event - capture metadata after route matching
+    auditor
```
schultzp2020: Do we need to await this function call?

packages/backend-openapi-utils/src/middlewares/auditor.ts
```diff
@@ -0,0 +1,220 @@
...
+
+function convertPathToRegex(path: string): RegExp {
+  // Convert OpenAPI path like /users/{id} to regex
+  const regexStr = path
+    .replace(/\{[^}]+\}/g, '([^/]+)') // Replace {param} with capture group
+    .replace(/\//g, '\\/'); // Escape slashes
+  return new RegExp(`^${regexStr}$`);
+}
+
+function extractValueFromObject(obj: any, path: string): any {
```
schultzp2020: Should we switch to using the `unknown` type?

packages/backend-openapi-utils/src/middlewares/auditor.ts
```diff
@@ -0,0 +1,220 @@
...
+    if (current && typeof current === 'object' && part in current) {
+      current = current[part];
+    } else {
+      return undefined;
+    }
+  }
+  return current;
+}
+
+export function auditorMiddlewareFactory(
```
schultzp2020: What are your thoughts on creating a `then()` wrapper around an async function? Personally I find reading async/await easier than `then()` callbacks.

aramissennyeydd: Yeah, I can give it a go - was struggling to get express-promise-router working in tests, let me try again.

packages/backend-openapi-utils/src/middlewares/auditor.ts
aramissennyeydd: No, this was a claude hallucination - I think it makes sense to add if we were to integrate with for example permissions

aramissennyeydd: added in https://github.com/backstage/backstage/pull/32148/commits/98ff9b825282934af759c3af0f376942ac30885f

packages/backend-openapi-utils/src/middlewares/auditor.ts
aramissennyeydd: Updated how this is called in https://github.com/backstage/backstage/pull/32148/commits/98ff9b825282934af759c3af0f376942ac30885f to have better async guarantees

packages/backend-openapi-utils/src/middlewares/auditor.ts
aramissennyeydd: also updated in https://github.com/backstage/backstage/pull/32148/commits/98ff9b825282934af759c3af0f376942ac30885f

packages/backend-openapi-utils/src/middlewares/auditor.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    // Intercept response body if we need to capture from it
+    const captureFromResponse = auditorConfig.meta?.captureFromResponse;
+    if (captureFromResponse?.body) {
+      const originalJson = res.json.bind(res);
+      const originalSend = res.send.bind(res);
+
+      const locals = res.locals as Response['locals'] &
+        WithCapturedResponseBody;
+      res.json = function overridenJson(body: any) {
```
schultzp2020:
```suggestion
      res.json = function overriddenJson(body: any) {
```

packages/backend-openapi-utils/src/middlewares/auditor.ts
```diff
@@ -0,0 +1,299 @@
...
+        await auditorEvent.fail({
+          error: new Error(
+            `Response returned with status code ${res.statusCode}`,
+          ),
+          meta: captureResponseMetadata(),
+        });
+      }
+      delete req[AUDITOR_SYMBOL];
+    } catch (err) {
+      next(new ForwardedError('Auditor middleware failed', err));
```
schultzp2020: After `await responseFinished`, the HTTP response has already been sent. If `auditorEvent.success()` or `auditorEvent.fail()` throws here, would calling `next(new ForwardedError(...))` have any effect? If not, should we log the error instead?

packages/backend-openapi-utils/src/middlewares/auditor.ts
```diff
@@ -0,0 +1,314 @@
...
+            `Response returned with status code ${res.statusCode}`,
+          ),
+          meta: captureResponseMetadata(),
+        });
+      }
+      delete req[AUDITOR_SYMBOL];
+    } catch (err) {
+      logger.error(`Failed to finalize auditor event`, err);
+      // Don't fail the request if auditing fails after the response is sent.
+      next();
```
schultzp2020:
```suggestion
```
I believe calling `next()` is a no-op, so it's safe to return.

packages/backend-openapi-utils/report.api.md
```diff
@@ -921,4 +945,22 @@ type UnknownIfNever<P> = [P] extends [never] ? unknown : P;
 
 // @public
 type ValueOf<T> = T[keyof T];
+
+// @public (undocumented)
+export interface WithAuditorEvent {
+  // (undocumented)
+  [AUDITOR_SYMBOL]?: Awaited<ReturnType<AuditorService['createEvent']>>;
```
schultzp2020: Do we want to export `AUDITOR_SYMBOL`?

packages/backend-openapi-utils/src/middlewares/auditor.ts
```diff
@@ -0,0 +1,314 @@
...
+      params?: string[];
+      query?: string[];
+    };
+    captureFromResponse?: {
+      body?: string[];
+    };
+  };
+};
+
+function extractValueFromObject(obj: unknown, path: string): any {
```
schultzp2020: Should we be more strict and remove the remaining `any` types?

packages/backend-openapi-utils/report.api.md
aramissennyeydd: is it safe to call `.success` or `.error` twice on AuditorService?

plugins/catalog-backend/src/schema/openapi/generated/router.ts
```diff
@@ -794,6 +794,16 @@ export const spec = {
         operationId: 'RefreshEntity',
         tags: ['Entity'],
         description: 'Refresh the entity related to entityRef.',
+        'x-backstage-auditor': {
```
Rugvip, line 856: For headers the `X-` prefix is discouraged nowadays, I think the same can be applied here unless there's a requirement to have the prefix. The `backstage-` prefix should do plenty to guarantee uniqueness anyway.

plugins/catalog-backend/src/schema/openapi/generated/router.ts
```diff
@@ -1448,6 +1552,15 @@ export const spec = {
         operationId: 'getLocationByEntity',
         tags: ['Locations'],
         description: 'Get a location for entity.',
+        'x-backstage-auditor': {
+          eventId: 'location-fetch',
+          meta: {
+            queryType: 'by-entity',
+            captureFromRequest: {
+              params: ['kind', 'namespace', 'name'],
```
Rugvip: This one is a bit interesting and I guess vaguely a breaking change, moving from a string to a compound ref, that is. I think it's fine though.

packages/backend-openapi-utils/report.api.md
Rugvip: My reaction was the opposite, do we really need to export `WithAuditorEvent`? 😅

packages/backend-openapi-utils/src/middlewares/auditor.ts
Rugvip: 👍, always nice to use `JsonValue` etc. if that's accurate

plugins/catalog-backend/src/schema/openapi.yaml
```diff
@@ -1117,6 +1163,14 @@ paths:
       tags:
         - Locations
       description: Create a location for a given target.
+      x-backstage-auditor:
+        eventId: location-mutate
+        severityLevel: medium
+        meta:
+          actionType: create
+          captureFromRequest:
```
Rugvip: Perhaps a bit cleaner to move this one up a level and call it something like `captureMetaFromRequest`?

Rugvip (commented):
I like! 👍 

Pretty nice cleanup of plugin routers, and it's also always possible to fall back to manual events if needed. Also not too much new API surface either given that we can mostly rely on the existing auditor API

plugins/catalog-backend/src/schema/openapi/generated/router.ts
aramissennyeydd: annoyingly enough, this is part of the spec itself, https://swagger.io/specification/ -
```
Field Pattern | Type | Description
-- | -- | --
^x- | Any | Allows extensions to the OpenAPI Schema. The field name MUST begin with x-, for example, x-internal-id. Field names beginning x-oai- and x-oas- are reserved for uses defined by the OpenAPI Initiative. The value can be any valid JSON value (null, a primitive, an array, or an object.)

Specification Extensions
While the OpenAPI Specification tries to accommodate most use cases, additional data can be added to extend the specification at certain points.

The extensions properties are implemented as patterned fields that are always prefixed by x-.

Field Pattern	Type	Description
^x-	Any	Allows extensions to the OpenAPI Schema. The field name MUST begin with x-, for example, x-internal-id. Field names beginning x-oai- and x-oas- are reserved for uses defined by the [OpenAPI Initiative](https://www.openapis.org/). The value can be any valid JSON value (null, a primitive, an array, or an object.)
```

plugins/catalog-backend/src/schema/openapi/generated/router.ts
aramissennyeydd: We can add functionality to transform the triplicate to a string, but not a_huge_ fan of adding much more dynamic logic here.

plugins/catalog-backend/src/schema/openapi/generated/router.ts
Rugvip: yep suspected as much :<

packages/backend-openapi-utils/report.api.md
schultzp2020: > is it safe to call `.success` or `.error` twice on AuditorService?

@aramissennyeydd yes, it should be.

schultzp2020: > My reaction was the opposite, do we really need to export `WithAuditorEvent`? 😅

After thinking about it, I agree.

freben: But calling them twice will also log twice, right?

plugins/catalog-backend/src/schema/openapi/generated/router.ts
schultzp2020: @Rugvip there are a few other breaking changes lingering around as well. For instance, dry runs are always considered medium whereas before it was considered low because the severity is no longer dynamic. I think the trade off is probably worth it for simplicity.

schultzp2020: Since this is already a breaking change, maybe we should rename `*-fetch` to `*-read` and `*-mutate` to `*-write` since that is the standard convention. 😅

plugins/catalog-backend/src/schema/openapi/generated/router.ts
```diff
@@ -1264,6 +1332,17 @@ export const spec = {
         operationId: 'CreateLocation',
         tags: ['Locations'],
         description: 'Create a location for a given target.',
+        'x-backstage-auditor': {
+          eventId: 'location-mutate',
+          severityLevel: 'medium',
+          meta: {
+            actionType: 'create',
```
freben: Could these be something more generic like "variant" or "type"? Instead of different forms of it in different routes.

packages/backend-openapi-utils/report.api.md
schultzp2020: > But calling them twice will also log twice, right?

That is correct.

plugins/catalog-backend/src/schema/openapi/generated/router.ts
aramissennyeydd: @schultzp2020 wdyt? These should just be copy-pasted from what we had before. Maybe also related to https://github.com/backstage/backstage/pull/32148#discussion_r2673005445 - do we even need this property or should this be communicated in the event id?

schultzp2020: > @schultzp2020 wdyt? These should just be copy-pasted from what we had before. Maybe also related to [#32148 (comment)](https://github.com/backstage/backstage/pull/32148#discussion_r2673005445) - do we even need this property or should this be communicated in the event id?

I considered using a single generic field like `variant` or `type`, but I think keeping `queryType` and `actionType` separate provides useful semantic value in audit logs. When reviewing events, you can immediately tell if an operation was a read (`queryType: by-uid`) vs a write (`actionType: delete`) without needing to cross reference the `eventId`. That said, I'm not strongly opposed to consolidating to a single field if consistency is preferred - the tradeoff is just losing the implicit read/write signal.

aramissennyeydd: Could we add HTTP verbs and route here instead? With the ability to override, thinking something like
```
GET -> query
POST -> create
DELETE -> delete
PATCH -> modify
```

aramissennyeydd: I added the actual express route + the above map to the logs, example:
```
2026-01-09T21:40:15.853Z catalog error catalog.entity-fetch isAuditEvent=true eventId="entity-fetch" severityLevel="low" actor={"actorId":"user:development/guest","ip":"::1","hostname":"localhost","userAgent":""} request={"url":"/api/catalog/entities/by-refs","method":"POST"} meta={"route":"/entities/by-refs","variant":"fetch"} status="initiated" trace_id="123" span_id="123" trace_flags="01"
```

plugins/catalog-backend/src/schema/openapi/generated/router.ts
```diff
@@ -1264,6 +1314,14 @@ export const spec = {
         operationId: 'CreateLocation',
         tags: ['Locations'],
         description: 'Create a location for a given target.',
+        'x-backstage-auditor': {
+          eventId: 'location-mutate',
+          severityLevel: 'medium',
+          captureMetaFromRequest: {
+            body: ['type', 'target'],
+            query: ['dryRun'],
```
freben: Maybe a bit of a wild idea, but what if we could do

```yaml
attributes:
  type: {{ body.type }}
  target: {{ body.target }}
  dryRun: {{ query.dryRun }}
```

https://github.com/backstage/backstage/blob/52d6767865825de8c50c6d598a807a56ec12771e/plugins/events-backend-module-google-pubsub/src/util/createPatternResolver.ts#L25

https://github.com/backstage/backstage/blob/52d6767865825de8c50c6d598a807a56ec12771e/plugins/events-backend-module-google-pubsub/config.d.ts#L65

And then you give the resolver a well-known context shape. Mappers tend to be so bespoke and weird and different in different places, otherwise.

Started trying to leverage this type of thing in other places too, but it's not really "embraced" yet so take it with a pinch of salt.

packages/backend-openapi-utils/src/middlewares/auditor.ts
```diff
@@ -0,0 +1,328 @@
...
+};
+
+function extractValueFromObject(
+  obj: JsonValue,
+  path: string,
+): JsonValue | undefined {
+  const parts = path.split('.');
+  let current: JsonValue | undefined = obj;
+  for (const part of parts) {
+    if (current && typeof current === 'object' && part in current) {
```
freben: Should there be an `Object.hasOwn(part)` in here instead, to not accidentally allow traversing along constructor chains and stuff?

plugins/catalog-backend/src/schema/openapi/generated/router.ts
aramissennyeydd: I like it, can we do something like with this too?
```
entityRef: '{{ path.kind }}/{{ path.namespace }}:{{ path.name }}'
```

edit: looking at it for more than 2 seconds - looks like yes!

aramissennyeydd: There's a small restriction that we can't have anything with a key of `id`. JSON Schema 4.0 assumes that it is actually the ID field, https://github.com/python-jsonschema/jsonschema/blob/main/json/tests/draft4/optional/id.json. OpenAPI 3.1 updated to use a newer version of JSON schema, which would resolve this.

Here the GET /location/:id and DELETE /location/:id needed to have their meta entries updated to `locationId` instead of `id`.

packages/backend-openapi-utils/src/util/createPatternResolver.ts
```diff
@@ -0,0 +1,118 @@
...
+): (context: TContext) => string {
+  // This split results in an array where even elements are static strings
+  // between placeholders, and odd elements are the contents inside
+  // placeholders.
+  //
+  // For example, the pattern:
+  //   "{{ foo }}-{{bar}}{{baz}}."
+  // will result in:
+  //   ['', 'foo', '-', 'bar', '', 'baz', '.']
+  const patternParts = pattern.split(/{{\s*([\w\[\]'"_.-]*)\s*}}/g);
```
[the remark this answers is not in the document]
aramissennyeydd: user controlled input defined in the schema

plugins/catalog-backend/src/schema/openapi.yaml
```diff
@@ -936,6 +953,10 @@ paths:
       tags:
         - Entity
       description: Get an entity by an entity ref.
+      x-backstage-auditor:
+        eventId: entity-fetch
+        meta:
+          entityRef: '{{ request.params.kind }}/{{ request.params.namespace }}:{{ request.params.name }}'
```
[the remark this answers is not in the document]
freben: There's also the matter of them needing to be lowercased in the `en-US` locale. The way it is made in here is risky.

PR #32159, opened 2025-12-17
microsite: pin swc

freben:
Due to [a recently introduced bug in swc](https://github.com/swc-project/swc/issues/11375), the transpiled code ends up having errors and cannot be built successfully.


drodil:
Good find, didn't come up in my small test run. Hope swc gets patched soonish to get rid of the pin 🤞

PR #32171, opened 2025-12-18
feat: schema-first permissions definitions

aramissennyeydd:

Related to #32148. This allows basic permissions to be defined on the schema instead of in code. My thinking here is that these basic perms are really just "can I call this endpoint" and should be represented + handled without writing any code 😁.


plugins/permission-node/src/integration/createPermissionIntegrationRouter.ts
```diff
@@ -564,5 +565,8 @@ export function createPermissionIntegrationRouter<
         getRuleByName: store.getRuleMapper(resourceRef.resourceType),
       } as PermissionRuleset<TResource, TQuery, TResourceType>;
     },
+    listPermissions(): Permission[] {
+      return store.getSerializedMetadata().permissions ?? [];
+    },
```
aramissennyeydd, line 570: make this return an object + async

PR #32200, opened 2025-12-22
BUI - Update surfaces tokens

cdedreuille:

This renamed CSS variable `--bui-bg` to `--bui-bg-surface-0` for consistency. The base background CSS variable has been renamed to follow the same naming pattern as other surface levels. This also adds a new surface `--bui-bg-surface-3`:

- `--bui-bg` → `--bui-bg-surface-0`
- `--bui-bg-surface-1` (unchanged)
- `--bui-bg-surface-2` (unchanged)
- `--bui-bg-surface-3` (New)

**Migration:**

Search and replace all instances of `--bui-bg` with `--bui-bg-surface-0` in your custom CSS and theme files. You can use this regex to find affected instances:

```regex
--bui-bg(?!-)
```

**Example:**

```diff
- background: var(--bui-bg);
+ background: var(--bui-bg-surface-0);
```

### DIFF
--- a/docs/conf/user-interface/index.md
+++ b/docs/conf/user-interface/index.md
@@ -164,6 +164,7 @@ These colors are used for the background of your application. We are mostly usin
 | `--bui-bg-surface-0`      | The background color of your Backstage instance. |
 | `--bui-bg-surface-1`      | Use for any panels or elevated surfaces.         |
 | `--bui-bg-surface-2`      | Use for any panels or elevated surfaces.         |
+| `--bui-bg-surface-3`      | Use for any panels or elevated surfaces.         |
 | `--bui-bg-solid`          | Used for solid background colors.                |
 | `--bui-bg-solid-hover`    | Used for solid background colors when hovered.   |
 | `--bui-bg-solid-pressed`  | Used for solid background colors when pressed.   |


cdedreuille:
@Rugvip we have multiple types of background colors. Surfaces are just one part of it. We also have status backgrounds like `danger`, `warning` or even component backgrounds like `solid`, ... Semantically we wanted to differentiate all type of backgrounds with foreground colors.

That said, we could explore separating surfaces from background:
- `--bui-surface-0`, `--bui-surface-1`, `--bui-surface-2`, `--bui-surface-3`
- `--bui-bg-danger`, `--bui-bg-warning`, ...

But I'm not sure we gain so much by doing that.

In a following PR we will introduce the new `bg` prop on components like `Box`, `Flex` or `Grid` with values like `surface-1`, `surface-2` or `danger`, ... For that reason it does make sense that the token follow the same structure for now.

Rugvip:
Alright, will wait for the followup PR, and let's keep `bg` in there. Still curious if it would be possible to do something like `bg="surface"` as the prop and automatically pick the correct one based on nesting depth. Not certain it's a good idea, but I'm thinking it might be tricky to make reusable mid-sized components that can be used at different depth otherwise?

packages/ui/src/css/tokens.css
```diff
@@ -140,9 +141,10 @@
     --bui-gray-8: #b4b4b4;
 
     /* Background Colors */
-    --bui-bg: #333333;
-    --bui-bg-surface-1: #424242;
+    --bui-bg-surface-0: var(--bui-black);
```
benjdlambert, line 144: hmm, is this really what we want?

Image

cdedreuille: This is fixed in this PR - https://github.com/backstage/backstage/pull/32203

PR #32202, opened 2025-12-22
BUI - Component surfaces

cdedreuille:

This PR introduce a new way for components to render on different surfaces. BUI has 4 main surfaces and previously some components would not render well if placed in other surfaces. The idea we are shaping here is to allow for certain components to adapt to their background using the `surface` and `onSurface` props.

To make it easier for anyone building components, these components will automatically get the correct `onSurface` by default based on the component they are on. As an example, if you place a `<Button variant="secondary" />` in a `<Box surface="1" />`, your button will automatically get its `onSurface` to `1`

Backstage Dark

### Migration:

We deprecated the tint color token in favour for the new neutral set of tokens on all surfaces.

```diff
- --bui-bg-tint              No replacement
- --bui-bg-tint-hover        No replacement
- --bui-bg-tint-pressed      No replacement
- --bui-bg-tint-disabled     No replacement

+ --bui-bg-neutral-on-surface-0
+ --bui-bg-neutral-on-surface-0-hover
+ --bui-bg-neutral-on-surface-0-pressed
+ --bui-bg-neutral-on-surface-0-disabled

+ --bui-bg-neutral-on-surface-1
+ --bui-bg-neutral-on-surface-1-hover
+ --bui-bg-neutral-on-surface-1-pressed
+ --bui-bg-neutral-on-surface-1-disabled

+ --bui-bg-neutral-on-surface-2
+ --bui-bg-neutral-on-surface-2-hover
+ --bui-bg-neutral-on-surface-2-pressed
+ --bui-bg-neutral-on-surface-2-disabled

+ --bui-bg-neutral-on-surface-3
+ --bui-bg-neutral-on-surface-3-hover
+ --bui-bg-neutral-on-surface-3-pressed
+ --bui-bg-neutral-on-surface-3-disabled
```

### Components impacted
- `Box` - Added the new `surface` prop as well as the `SurfaceProvider`.
- `Flex` - Added the new `surface` prop as well as the `SurfaceProvider`.
- `Grid` - Added the new `surface` prop as well as the `SurfaceProvider`.
- `Button` - Added the new `onSurface` prop.

### DIFF
--- a/packages/ui/report.api.md
+++ b/packages/ui/report.api.md
@@ -1294,26 +1294,6 @@ export type Surface =
   | 'auto';
 
 // @public (undocumented)
-export interface SurfaceContextValue {
-  // (undocumented)
-  surface: Responsive<Surface> | undefined;
-}
-
-// @public
-export const SurfaceProvider: ({
-  surface,
-  children,
-}: SurfaceProviderProps) => JSX_2.Element;
-
-// @public (undocumented)
-export interface SurfaceProviderProps {
-  // (undocumented)
-  children: ReactNode;
-  // (undocumented)
-  surface: Responsive<Surface>;
-}
-
-// @public (undocumented)
 export const Switch: ForwardRefExoticComponent<
   SwitchProps & RefAttributes<HTMLLabelElement>
 >;
@@ -1618,15 +1598,6 @@ export const useBreakpoint: () => {
   up: (key: Breakpoint) => boolean;
   down: (key: Breakpoint) => boolean;
 };
-
-// @public
-export const useSurface: (options?: UseSurfaceOptions) => SurfaceContextValue;
-
-// @public (undocumented)
-export interface UseSurfaceOptions {
-  onSurface?: Responsive<Surface>;
-  surface?: Responsive<Surface>;
-}
 
 // @public
 export function useTable<T = any>(
--- a/packages/ui/src/components/Button/Button.tsx
+++ b/packages/ui/src/components/Button/Button.tsx
@@ -54,7 +54,7 @@ export const Button = forwardRef(
         ref={ref}
         isPending={loading}
         {...dataAttributes}
-        {...(surface ? { 'data-on-surface': surface } : {})}
+        {...(typeof surface === 'string' ? { 'data-on-surface': surface } : {})}
         {...rest}
       >
         {({ isPending }) => (
--- a/packages/ui/src/hooks/useSurface.tsx
+++ b/packages/ui/src/hooks/useSurface.tsx
@@ -82,7 +82,7 @@ function incrementSurface(surface: Surface | undefined): Surface {
  * @param contextSurface - The surface from context
  * @param requestedSurface - The requested surface value (may be 'auto')
  * @returns The resolved surface value
- * @public
+ * @internal
  */
 export function resolveSurfaceForProvider(
   contextSurface: Responsive<Surface> | undefined,
@@ -156,7 +156,7 @@ function resolveSurfaceForConsumer(
  * Note: The surface value should already be resolved before passing to this provider.
  * Container components should use useSurface with the surface parameter.
  *
- * @public
+ * @internal
  */
 export const SurfaceProvider = ({
   surface,
@@ -180,7 +180,7 @@ export const SurfaceProvider = ({
  * - `onSurface`: Leaf behavior - 'auto' inherits from parent
  *
  * @param options - Optional configuration for surface resolution
- * @public
+ * @internal
  */
 export const useSurface = (
   options?: UseSurfaceOptions,
--- a/.changeset/modern-drinks-battle.md
+++ b/.changeset/modern-drinks-battle.md
@@ -9,6 +9,6 @@
 There's no direct replacement for the old tint tokens but you can use the new neutral set of color tokens on surface 0 or 1 as a replacement.
 
 - `--bui-bg-tint` can be replaced by `--bui-bg-neutral-on-surface-0`
-- --bui-bg-tint-hover can be replaced by `--bui-bg-neutral-on-surface-0-hover`
-- --bui-bg-tint-pressed can be replaced by `--bui-bg-neutral-on-surface-0-pressed`
-- --bui-bg-tint-disabled can be replaced by `--bui-bg-neutral-on-surface-0-disabled`
+- `--bui-bg-tint-hover` can be replaced by `--bui-bg-neutral-on-surface-0-hover`
+- `--bui-bg-tint-pressed` can be replaced by `--bui-bg-neutral-on-surface-0-pressed`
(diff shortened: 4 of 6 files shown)


packages/ui/src/components/Button/Button.tsx
```diff
@@ -35,15 +36,27 @@ export const Button = forwardRef(
...
+      iconStart,
+      iconEnd,
+      loading,
+      onSurface,
+      ...rest
+    } = cleanedProps;
+
+    const { surface } = useSurface();
+
+    const surfaceToUse = onSurface || surface;
```
Rugvip: Not entirely sure, but it may be better to structure this as `const { surface } = useSurface({ onSurface })`. The context is an API boundary across versions, and putting all the logic in the same place may make it easier to make changes to the API and behavior. It would mean that the context is no longer passing plain data around but instead a callback that returns that data instead.

packages/ui/src/hooks/useSurface.tsx
```diff
@@ -0,0 +1,60 @@
...
+  surface: Responsive<Surface> | undefined;
+}
+
+/** @public */
+export interface SurfaceProviderProps {
+  surface: Responsive<Surface>;
+  children: ReactNode;
+}
+
+const SurfaceContext = createContext<SurfaceContextValue>({
```
Rugvip: 👍 for not exporting this context. Since it may be used across plugin boundaries it also needs to be a versioned singleton context created with `@backstage/version-bridge`. Here's an example of that: https://github.com/backstage/backstage/blob/2e26579e06197d0ba2d8e55952b40afeee6bf996/plugins/scaffolder-react/src/secrets/SecretsContext.tsx#L29-L98

packages/ui/src/types.ts
```diff
@@ -173,3 +173,9 @@ export interface ComponentDefinition {
   dataAttributes?: DataAttributesMap;
   utilityProps?: string[];
 }
+
+/**
+ * Surface type
+ * @public
+ */
+export type Surface = '0' | '1' | '2' | '3' | 'danger' | 'warning' | 'success';
```
Rugvip: What's your thought on adding `true` here as well? which would increment the existing level by one. On one hand it feels risky since it'd be quite easy to hit the limit. On the other hand I think there's a bit of a risk of a global "z-index" problem with this approach when it comes to nesting content from different plugins. For example, are catalog entity page cards surface 1, 2, 3, something else? What if you wanted to render the same card on the home page, does it need a different level? We could also solve that more at the framework/plugin glue level where we make sure that this does not become a cross-plugin API surface. That'd need large redesigns in other areas though.

cdedreuille: Thanks @Rugvip. You're right, we should make it work with auto increment. I just pushed an update adding `auto` to the list of types. This mean that if you place a `Box` into different containers, it will inherit from its parent surface.

```tsx
<Box surface="0> 
  <Box surface="auto />  {/* Box gets surface="1" - Increment of 1 */}
</Box>
<Box surface="1> 
  <Box surface="auto />  {/* Box gets surface="2" - Increment of 1 */}
</Box>
```

packages/ui/src/components/Button/Button.tsx
cdedreuille: You're right. I improved it to pass `surface` prop for `Box`, `Flex`, `Grid` and `onSurface` for `Button`

packages/ui/src/hooks/useSurface.tsx
cdedreuille: Thanks. I moved to `@backstage/version-bridge` now.

packages/ui/report.api.md
```diff
@@ -1535,6 +1619,15 @@ export const useBreakpoint: () => {
...
 
+// @public
+export const useSurface: (options?: UseSurfaceOptions) => SurfaceContextValue;
+
+// @public (undocumented)
+export interface UseSurfaceOptions {
+  onSurface?: Responsive<Surface>;
+  surface?: Responsive<Surface>;
+}
+
```
Rugvip: Do you see an immediate need to make this one external? Thinking it could be nice to be able to keep it internal for now and evolve the API. I could definitely see this being useful in custom internal components though.

For a public API I feel this may need a bit more direction for consumers. Some potential alternatives:

```ts
export type UseSurfaceOptions = {
  onSurface: Responsive<Surface>;
} | {
  surface: Responsive<Surface>;
}
```

```ts
export type UseSurfaceOptions = {
  surface: Responsive<Surface>;
  type: 'producer' | 'consumer' // or 'bg' / 'fg', 'surface' / 'item', w/e
}
```

```ts
export type UseSurfaceOptions = {
  surface: Responsive<Surface>;
  increment?: boolean
}
```

packages/ui/src/components/Button/Button.tsx
```diff
@@ -35,15 +36,25 @@ export const Button = forwardRef(
...
+
+    const { surface } = useSurface({ onSurface });
 
     return (
       <RAButton
         className={clsx(classNames.root, styles[classNames.root], className)}
         ref={ref}
         isPending={loading}
         {...dataAttributes}
+        {...(surface ? { 'data-on-surface': surface } : {})}
```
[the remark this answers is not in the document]
Rugvip: Seems like this one might be true?

packages/ui/src/hooks/useSurface.tsx
```diff
@@ -0,0 +1,204 @@
...
+    return requestedSurface;
+  }
+
+  // For containers, 'auto' means increment to create a new elevated context
+  if (requestedSurface === 'auto') {
+    return incrementSurface(contextSurface);
+  }
+
+  return requestedSurface;
+}
```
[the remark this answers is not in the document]
Rugvip: yep ☝️, mark as `@internal` or just nothing

packages/ui/report.api.md
cdedreuille: No we can keep it 100% internal for now. This was not intended to be used externally.

packages/ui/src/hooks/useSurface.tsx
```diff
@@ -0,0 +1,204 @@
...
+  // If requestedSurface is a responsive object (breakpoint-based), return as-is
+  if (typeof requestedSurface === 'object') {
+    return requestedSurface;
+  }
+
+  // If contextSurface is a responsive object, we can't auto-increment from it
+  // Return the requested surface as-is or default to '0' for auto
+  if (typeof contextSurface === 'object') {
+    if (requestedSurface === 'auto') {
+      return '0'; // fallback to root when context is responsive
```
Rugvip, line 104: Hmm, this one seems a bit tricky. I wonder if we should even just not allow surface to be set responsively?

Rugvip (approved):
Nice, good to go, although one optional thought

packages/ui/src/hooks/useSurface.tsx
cdedreuille: Yes maybe we don't need the responsiveness.

Issue #31089, opened 2025-09-08
SearchFilter.Autocomplete displays selected value instead of its label


maiasanders:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
SearchFilter, Autocomplete
```

### 🗃️ Project Area

Search

### 🔗 External Integration

N/A

### 📝 Description & Context

When the selected option is a FilterLabelWithValue, it is displayed as the value rather than the label (e.g. when filtering by owner, it would display the entityRef rather than the display name when those are the value and label respectively).

### 👍 Expected Behavior

I expect the label to displayed like it is for the selected option in SearchFilter.Select

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Undecided

camilaibs:
Hi @maiasanders, would you be willing to open a pull request with the expected behavior you suggested? Your contribution would be greatly appreciated 🙂

maiasanders:
@camilaibs Yes, I would be willing to do that!

camilaibs:
Awesome, I've assigned the issue to you. Thank you in advance for your contribution! 🎉

maiasanders:
I have a PR up to fix this bug, but it is still waiting for approval

PR #32204, opened 2025-12-23
SearchFilter.Autocomplete renders label and has proper behavior with multiple

maiasanders:

This fixes backstage/backstage#31089. SearchFilter.Autocomplete options that have separate value and label properties are now rendered as the label when selected. Also fixes a bug that caused input to reset to an empty string when attempting to type after making a selection when multiple is set to true. 

For testing purposes, SearchContextFilterSpy was updated to work with the ensured FilterValueWithLabel shape. I also added an additional component to the SearchFilter storybook and updated the existing async components to use the new swapi.info as swapi.dev is no longer available due to an expired certificate.

A Storybook component that makes use of the object format (instead of strings) was added for testing
image

image

### DIFF
--- a/.changeset/weak-parts-hang.md
+++ b/.changeset/weak-parts-hang.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-search-react': patch
+---
+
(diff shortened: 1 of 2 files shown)


maiasanders:
Can this be reopened

PR #32211, opened 2025-12-24
feat(openapi): add spec for permissions router

aramissennyeydd:

Adding a simple OpenAPI spec for the permissions router. This is intended to replace the existing zod validation for request/response formats. 

Open questions:
- [ ] We throw an InputError with the ajv output - if we use the OpenAPI validation, the error format will be different. Is this an important API contract? (can also break across AJV versions).
- [ ] `resourceRef` being undefined for basic permissions is a no-op for a JSON API since `JSON.stringify({test: undefined})` is `{}`, we can omit from the schema completely for better validation. Is this a breaking change?


plugins/permission-backend/package.json
```diff
@@ -45,18 +45,21 @@
     "build": "backstage-cli package build",
     "clean": "backstage-cli package clean",
     "lint": "backstage-cli package lint",
+    "generate": "backstage-repo-tools package schema openapi generate --server",
```
freben, line 48: Will this be generating the client parts in `permission-common` too later?

plugins/permission-backend/src/schema/openapi/generated/models/PermissionCriteria.model.ts
```diff
@@ -0,0 +1,32 @@
...
+import { PermissionCriteriaOneOf2 } from '../models/PermissionCriteriaOneOf2.model';
+
+/**
+ * @public
+ */
+export type PermissionCriteria =
+  | PermissionCondition
+  | PermissionCriteriaOneOf
+  | PermissionCriteriaOneOf1
+  | PermissionCriteriaOneOf2;
```
freben: Do you think something could be done to these names so they reflect allOf / not ?

plugins/permission-backend/src/schema/openapi.yaml
```diff
@@ -0,0 +1,309 @@
...
+      type: object
+      properties:
+        rule:
+          type: string
+          description: The name of the permission rule
+        resourceType:
+          type: string
+          description: The type of resource this condition applies to
+        params:
+          type: object
```
freben, line 189: Oh I thought this was an array of strings 🤔 Or is that something else I'm thinking about?

plugins/permission-backend/src/schema/openapi.yaml
```diff
@@ -0,0 +1,309 @@
...
+          description: The permission name
+        attributes:
+          $ref: '#/components/schemas/PermissionAttributes'
+        resourceType:
+          type: string
+          description: The type of resource this permission applies to
+      required:
+        - type
+        - name
+        - attributes
```
freben, line 116: is it really mandatory?

freben (commented):
Alright, looks good in general! @vinzscam @Rugvip to have a peek too?

plugins/permission-backend/src/schema/openapi.yaml
aramissennyeydd: From the TS definition, https://github.com/backstage/backstage/blob/f0a117ec5c43b0f0bf7d0e3ae17353d0bafea827/plugins/permission-common/src/types/permission.ts#L35-L40, it is and the old zod schema was checking for it too,
```
const attributesSchema: z.ZodSchema<PermissionAttributes> = z.object({
  action: z
    .union([
      z.literal('create'),
      z.literal('read'),
      z.literal('update'),
      z.literal('delete'),
    ])
    .optional(),
});
```

plugins/permission-backend/src/schema/openapi.yaml
aramissennyeydd: This type comes from https://github.com/backstage/backstage/blob/f0a117ec5c43b0f0bf7d0e3ae17353d0bafea827/plugins/permission-common/src/types/api.ts#L119-L126 - TParams extends Record<string, undefined | JsonPrimitive | JsonPrimitive[]>. Maybe you're thinking of the array being per key?

plugins/permission-backend/src/schema/openapi/generated/models/PermissionCriteria.model.ts
aramissennyeydd: Yup, just need to extract each `oneOf` block to a ref - if they're defined inline they will use array numbering. Will update 🫡

aramissennyeydd: updated in [`a035c25` (#32211)](https://github.com/backstage/backstage/pull/32211/commits/a035c25c546d17593c2c57e68d8b56fb50e088a2)

plugins/permission-backend/package.json
aramissennyeydd: Yup! Should be straightforward once we're happy with the OpenAPI schema

freben (approved):
Alright! @vinzscam is back next week, let's give him a chance to take a look too

vinzscam (changes requested):
Hey, thanks for the PR.
I've got a few questions/concerns I wanted to chat about before we move forward:
The main thing I'm wondering: what problem are we solving here?
This is an internal endpoint (only Backstage calls it). So the usual OpenAPI benefits (generated clients, Swagger docs, external API discoverability) don't really apply to us. 

We're swapping out a zod validation system that works fine for a new one. That is a risk we're taking on without a clear upside and there is an error format change that could introduce side effects.
I'm not against OpenAPI in general as it makes total sense for some APIs, but for an internal endpoint like this, I'm struggling to see what we gain vs what we risk (and the risk here is huge, as if this new validation system has a bug it might effect the entire Backstage).

If we really do need an OpenAPI spec, is it possible to keep the existing zod schemas?

aramissennyeydd:
@vinzscam We chatted about this a bit in the maintainer sync - the main things moving to OpenAPI that we'd get here are
1. Standard error format for validation errors. The current zod errors here are hard to read and not easy to parse (double encoded JSON).
2. Ability to use generated clients instead of hand rolling our own. Guaranteed to have compatibility across client + server. I think this is still useful even if this is just internal.

I'm happy to add more test cases to alleviate concerns about rolling this out if that would help?

> If we really do need an OpenAPI spec, is it possible to keep the existing zod schemas?

We could, but they should be unused at that point as the validation would be handled by the OpenAPI schema + request middleware.

aramissennyeydd:
@vinzscam Gentle bump here 🙏

PR #32219, opened 2025-12-29
[BUI] Added support for custom page size options

grantila:

This PR adds support for custom page size options to the `useTable` hook, forwarded to the `TablePagination` component.


vinzscam (approved):

grantila:
@cdedreuille I think this would need your review 🙏

cdedreuille:
We are working on a complete revamp of the `useTable` hook to improve support for cursor pagination. I might be wrong but I believe the change you requested will be included in our new version.

https://github.com/backstage/backstage/pull/32050

grantila:
I don't think it is, it's still a [hard-coded list](https://github.com/backstage/backstage/blob/master/packages/ui/src/components/TablePagination/TablePagination.tsx#L66-L73).

I really like the new Table component interface, wrapping everything needed. There are a few things I would need to be able to use it though:
 * Being able to set styles, e.g. `{ minWidth: '100%', tableLayout: 'auto' }` to the `TableRoot` for full-width and adaptive column widths.
 * Being able to hook into page size change events; I'm saving it in user settings, hence need it manually handled. This was possible with the hooks in the `pagination` prop of `useTable` before.

If these are interesting to have, I could open up an issue about them, just wanted to let you know. Otherwise I'm happy to keep using the lower level `TableRoot` and corresponding components.

Naycon:
> I don't think it is, it's still a [hard-coded list](https://github.com/backstage/backstage/blob/master/packages/ui/src/components/TablePagination/TablePagination.tsx#L66-L73).

Right you are, this is an oversight. We should allow you to configure this. Thanks for raising the issue!

> * Being able to set styles, e.g. `{ minWidth: '100%', tableLayout: 'auto' }` to the `TableRoot` for full-width and adaptive column widths.

Yea, we should likely support this as well as passing a class name in. Question is if you'd expect those to end up on the wrapper that's around table root + pagination controls (a div currently), or on the table root. I'm leaning towards the former perhaps, and defaulting the table to span the full width of the container. Any thoughts on that?

Additionally, I can add that we're looking into soon adding some support for the [React Aria column resizing](https://react-aria.adobe.com/Table#column-resizing), to allow you to specify columns widths in the column config. This also improves the responsive behavior of the table I think. 

> * Being able to hook into page size change events; I'm saving it in user settings, hence need it manually handled. This was possible with the hooks in the `pagination` prop of `useTable` before.

Good catch, this is also an oversight. The pagination controls supports this, it's just not exposed via the hook or Table component. We should fix that as well and allow you to listen to both size and page changes.

Feel free to adopt your PR to the new system if you like, but as I mentioned above I'm also making some more table updates, so I can try to cover these for you if you prefer!

grantila:
Sorry for the late reply, to not duplicate the work please see if you can cover this 🙏  I'll work around this in the mean time.

Naycon:
> Sorry for the late reply, to not duplicate the work please see if you can cover this 🙏 I'll work around this in the mean time.

@grantila this has been fixed in #32321 and should've shipped already. I hope we've covered all that you need, so give it a try if you can and let us know if we've missed something. The Table docs on ui.backstage.io should now also be better at listing what props are available and with some guides on how to do some of the more common things.

PR #32232, opened 2025-12-31
feat(bui): Introduce ToggleButton/ToggleButtonGroup

### LINKED ISSUE #32215: BUI: Add ToggleButton/ToggleButtonGroup
### 📜 Issue Labels

### 🔎 Search Terms

```plain
bui, toggle, switch
```

### 🗃️ Project Area

Design System

### 🔖 Need

`ToggleButtonGroup`s & `ToggleButton`s are not used widely in Backstage: less than 5 usages in backstage + backstage-community.

However, we use them quite a lot in our internal plugins, mostly for UI controls. I believe that when used appropriately, they some benefits over switches/checkbox/selects:

- They are the same height as buttons (can be put on the same row as action buttons)
- They are compact 
- Options are always shown
- They work well with icons only, text only or a mix of both

### 📝 Proposal

The components are already available in [react-aria-components](https://react-aria.adobe.com/ToggleButton), so they can be implemented in `@backstage/ui`.

The `react-aria-components` seem to provide enough options to be a proper replacement of the MUI v4 (lab) and MUI v5 ToggleButtons components.

### 🔄 Alternatives

_No response_

GabDug:

Close #32215.

- Add 2 new components `ToggleButton` & `ToggleButtonGroup`, based on react-aria-components
- Add docs-ui & storybooks examples
- Added support for surfaces from #32202, but lightly tested as Backstage theme does not use them at the moment.

### Regarding CSS Modules Composition

The 2 components work together, but:
- The ToggleButton has its own border-radius (when used in standalone)
- When used inside a ToggleButtonGroup, the group control the buttons border-radiuses

This is my first time working with CSS Modules, and I haven't found other BUI components that are as tied, yet can be used in standalone (e.g. Tag can't be used outside of TagGroup).

As there is no current components with this pattern nor precise style guides, I've chosen to split them as 2 different components, and use [`:global`](https://github.com/css-modules/css-modules/blob/master/docs/composition.md#exceptions) for some styling overrides.

It also helps for discoverability, as the 2 components have their own docs-ui sidebar item / storybook item.

Let me know if you think it would be better in the same CSS Module, I can definitively merge the 2 components!

### Screenshots

_Please launch `docs-ui` or storybook locally to see all examples & theme-support!_

ToggleButtonGroup demo (dark)
image

ToggleButton demo (light)
image

### DIFF
--- a/packages/ui/src/components/ToggleButton/ToggleButton.stories.tsx
+++ b/packages/ui/src/components/ToggleButton/ToggleButton.stories.tsx
@@ -16,13 +16,11 @@
 
 import preview from '../../../../../.storybook/preview';
 import { ToggleButton } from './ToggleButton';
-import { ToggleButtonGroup } from '../ToggleButtonGroup/ToggleButtonGroup';
 import { Flex } from '../Flex';
 import { Text } from '../Text';
 import { useState } from 'react';
 import {
   RiCheckLine,
-  RiHeartLine,
   RiStarFill,
   RiStarLine,
   RiCloudLine,
@@ -150,20 +148,6 @@ export const IconsAndText = meta.story({
   ),
 });
 
-export const MixedIcons = meta.story({
-  render: () => (
-    <ToggleButtonGroup selectionMode="multiple" defaultSelectedKeys={['cloud']}>
-      <ToggleButton id="cloud" iconStart={<RiCloudLine />}>
-        Cloud
-      </ToggleButton>
-      <ToggleButton id="star" iconStart={<RiStarLine />} />
-      <ToggleButton id="arrow" iconEnd={<RiArrowRightSLine />}>
-        Next
-      </ToggleButton>
-    </ToggleButtonGroup>
-  ),
-});
-
 export const Disabled = meta.story({
   render: () => (
     <Flex align="center">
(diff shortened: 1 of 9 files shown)


packages/ui/src/components/ToggleButton/types.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Properties for {@link ToggleButton}
+ *
+ * @public
+ */
+export interface ToggleButtonProps extends AriaToggleButtonProps {
+  size?: 'small' | 'medium' | Partial<Record<Breakpoint, 'small' | 'medium'>>;
+  variant?:
+    | 'primary'
+    | 'secondary'
+    | Partial<Record<Breakpoint, 'primary' | 'secondary'>>;
```
cdedreuille: I'm not sure we need for now multiple variants. This component can actually toggle between our primary and secondary state. I would remove this prop for now.

packages/ui/src/components/ToggleButton/ToggleButton.stories.tsx
```diff
@@ -0,0 +1,226 @@
...
+      options: ['small', 'medium'],
+    },
+    variant: {
+      control: 'select',
+      options: ['primary', 'secondary'],
+    },
+  },
+});
+
+export const Default = meta.story({
```
cdedreuille, line 41: Could you please add a story to see how this component can look like on multiple surfaces the same way we did on `Button`?

packages/ui/src/components/ToggleButton/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  iconStart?: ReactElement;
+  iconEnd?: ReactElement;
+  children?: ReactNode;
```
cdedreuille: `AriaToggleButtonProps` should already have `children`. No need to add it once again here.

packages/ui/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+      cleanedProps;
+
+    const { surface } = useSurface({ onSurface });
+
+    return (
+      <AriaToggleButton
+        className={clsx(classNames.root, styles[classNames.root], className)}
+        ref={ref}
+        {...dataAttributes}
+        {...(typeof surface === 'string' ? { 'data-on-surface': surface } : {})}
```
cdedreuille, line 47: This does not appear for some reason on the component meaning that the styles are not properly applied.

packages/ui/src/hooks/useStyles.ts
```diff
@@ -17,16 +17,29 @@ import { useBreakpoint, breakpoints } from './useBreakpoint';
 import type { ComponentDefinition } from '../types';
 import { utilityClassMap } from '../utils/utilityClassMap';
 
+type DataAttributeValue<V> = V extends string ? V : string;
```
cdedreuille: I would try to avoid changing `useStyles` as part of this PR as it is used in a lot of places. Any change to this file would have to be done in a separate PR. What was your intention?

.changeset/rude-trains-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': minor
+---
+
+feat: Introduce new `ToggleButton` & `ToggleButtonGroup` components
```
cdedreuille: Can you please specify that this is for Backstage UI?

packages/ui/src/components/ToggleButton/ToggleButton.module.css
```diff
@@ -0,0 +1,148 @@
...
+    color: var(--bui-fg-primary);
+    box-shadow: none;
+
+    &[data-hovered]:not([data-disabled]) {
+      background-color: var(--bui-bg-neutral-on-surface-0-hover);
+    }
+
+    &[data-pressed]:not([data-disabled]),
+    &[data-selected]:not([data-disabled]) {
+      background-color: var(--bui-bg-neutral-on-surface-0-pressed);
```
cdedreuille: Pleaase use `--bui-bg-neutral-on-surface-0-disabled` instead for disabled states.

packages/ui/src/components/ToggleButton/types.ts
GabDug: The `children` attribute from `AriaToggleButtonProps` can also be a function to alter the children based on component state.

I took the same approach as `DialogProps` and used a subset of the RA type for this first iteration, for a simpler approach.

Would you prefer I document this, or try to properly support the render ffunctions?

cdedreuille: In that instance, one of the reason for simplifying the prop would be to control the UI but in this instance we could imagine a use case were the person would like to change the text based on the disable state. This would stay true to our design. I think it would be great to keep the original intent.

packages/ui/src/components/ToggleButton/ToggleButton.tsx
GabDug: Tested again and `data-on-surface` does appear for me (in docs-ui / storybook) but only when on a surface, which is consistent with the Button behaviour.

packages/ui/src/hooks/useStyles.ts
GabDug: The intention was to preserve the literal string union type for `orientation`, to avoid type casts.

I've removed the responsive orientation support for now: consumers can support it by themselves if needed.

packages/ui/src/components/ToggleButton/types.ts
GabDug: I've removed the variants for now. It was indeed getting messy with orientation, variants and sizes.

I still wish to be able to create ToggleButton(Groups) that visually match all Button variants, but I'll try to introduce variants in an other PR. 
In this PR, I'll also try to improve the experience by allowing the ToggleButtonGroup to accept variant/size and pass them to each ToggleButton via a context. Setting it on each ToggleButton is a little bit repetitive (and could easily lead to some "mixed" styling).

packages/ui/src/components/ToggleButton/types.ts
cdedreuille: That's the last remaining part @GabDug. After that we can merge the PR.

GabDug: @cdedreuille Pushed 6921b2c824ce47ea4f5a3115617ffe42ab5c809c to support that, with a few examples.

cdedreuille: This commit is not part of the branch for some reasons. This is the message I see:

> This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.

GabDug: Hum weird, it must be a safety feature when working with forks. It's the correct SHA, but I can't link it to the "view in PR context" view.

`https://github.com/backstage/backstage/pull/32232/changes/6921b2c824ce47ea4f5a3115617ffe42ab5c809c`
or https://github.com/Believe-SA/backstage/commit/6921b2c824ce47ea4f5a3115617ffe42ab5c809c

cdedreuille: All resolved now. Thanks 🙏

clairep94:
Hi @GabDug it looks like in your screenshot that the `light` mode toggle button group look like they have gaps between them, while the `dark` mode does not. Is this accurate or is this just an older screenshot?

vinzscam (approved):

GabDug:
> Hi @GabDug it looks like in your screenshot that the `light` mode toggle button group look like they have gaps between them, while the `dark` mode does not. Is this accurate or is this just an older screenshot?

Hey @clairep94, the first screenshot is of `ToggleButtonGroup`, while the second one is of individual `ToggleButton`s.
There should be no layout changes between light and dark themes.

You're right to point that these are indeed older screenshots.

PR #32249, opened 2026-01-06
fix(ui): Hide browser-specific password field controls in PasswordField

deepthi-28:
Hides browser-native password reveal buttons and autofill icons to prevent
UI conflicts with Backstage's custom visibility toggle button. This ensures
a consistent user experience across different browsers (Chrome, Edge, Safari).

- Hide Edge/Chrome password reveal button using ::-ms-reveal and ::-ms-clear
- Hide Chrome/Safari password manager icons using ::-webkit-* pseudo-elements
- Prevents duplicate visibility toggles and visual clutter

image

- [x ] A changeset describing the change and affected packages. ([more info](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md#creating-changesets))
- [ x] Screenshots attached (for UI changes)

### DIFF
--- a/packages/ui/src/components/PasswordField/PasswordField.module.css
+++ b/packages/ui/src/components/PasswordField/PasswordField.module.css
@@ -101,7 +101,6 @@
       -webkit-appearance: none;
     }
 
-    /* Hide Edge/Chrome password reveal button to avoid conflicts with custom toggle */
     &::-ms-reveal,
     &::-ms-clear {
       display: none;


deepthi-28:
@cdedreuille Please review this PR.

cdedreuille:
On Safari I can see that in your branch the autofill disappear. Was it intentional? Ideally we should keep this functionality as this is not conflicting with the hide/show icon and quite useful.

CleanShot 2026-01-12 at 11 38 01@2x

packages/ui/src/components/PasswordField/PasswordField.module.css
```diff
@@ -111,6 +111,12 @@
       -webkit-appearance: none;
     }
 
+    /* Hide Edge/Chrome password reveal button to avoid conflicts with custom toggle */
+    &::-ms-reveal,
+    &::-ms-clear {
```
Naycon: Sorry, I'm not an Edge user - are these still supported by Edge? [This](https://udn.realityripple.com/docs/Archive/Web/CSS/::-ms-reveal) and [this](https://udn.realityripple.com/docs/Archive/Web/CSS/::-ms-clear) archived MDN posts seem to suggest support was dropped after version 79 of Edge:

> Historically supported in Internet Explorer and Edge before version 79.

Given Edge is now on version 140+, it might not make sense to add support for such an old version if this is indeed not supported anymore. Can you confirm these are actually in use still today?

.changeset/bitter-laws-count.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': minor
+---
+
+Hide browser-specific password field controls in PasswordField
```
Naycon, line 5:
```suggestion
Hide Edge-specific reveal/clear password field controls in PasswordField.

Affected components: PasswordField
```

packages/ui/src/components/PasswordField/PasswordField.module.css
```diff
@@ -111,6 +111,12 @@
       -webkit-appearance: none;
     }
 
+    /* Hide Edge/Chrome password reveal button to avoid conflicts with custom toggle */
```
Naycon: Let's remove this comment, I think the code is speaks for itself and this also does nothing for Chrome afaik.

deepthi-28:
@Naycon I plan to remove the CSS changes that hide browser-specific password icons, since the double icon issue only occurs in Microsoft Edge. I can implement a code solution to handle Edge separately. Would it be acceptable to implement password control logic specifically for Edge?

Naycon:
> @Naycon I plan to remove the CSS changes that hide browser-specific password icons, since the double icon issue only occurs in Microsoft Edge. I can implement a code solution to handle Edge separately. Would it be acceptable to implement password control logic specifically for Edge?

Sorry for the slow feedback! I'm not sure I understand what issue we're trying to fix here. Have you/could you raise an issue that explains what it is that is the problem that you're trying to solve, and reference back to this PR? This PR is right now only doing fixes for Edge, using selectors that, if I read MDN correctly, are not in use anymore in Edge since many versions back. I might be wrong there, so raising an issue and explaining a bit more your situation would help me understand this proposed solution better.

PR #32268, opened 2026-01-08
fix(api-breaking-changes): add better comment identifier

aramissennyeydd:

Noticed in #32148 that we're not updating the comment if it has info in it, instead we're just creating a new comment. This is due to the string `API changes` not being found in the comment _with_ info in it, only the default comment without info. This PR fixes that by using an HTML comment with a more unique ID instead of relying on a string in the comment itself.


aramissennyeydd:
The changes to `.github/workflows/api-breaking-changes-comment.yml` mean that I can't test the full flow without merging this 🤦.

PR #32273, opened 2026-01-09
feat: add customization options to NavItemBlueprint params and NavContentBlueprint

AdrianSoovary8234:

The way the Nav in the New Frontend System works is kind of limited and it is still hard to come around using `convertLegacyAppRoot` to create a working app root with usable Sidebar. To push full adoption of the NFS for the sidebar I tried out the following changes which work fine for me.
I made a change to the NavItemBlueprint and the NavContentBluepriunt so that additional params can be passed. 

I discovered that the NavItemBlueprint lacked basic ways for adjustment which I added: 
- A Custom component can be passed which will be rendered instead of the standard SidebarItem
- nav items can be hidden (important if you wanna disable sidebar items of core or community plugins) 
- a position argument can be passed, so there is a possibility to have an influence on ordering of the items in the navigation, the argument is optional so gaps between the positioned elements will be filled with unpositioned elements. if there are two elements with the same position the one which was imported first will be shown first
- which previously could only be influenced by changing the import order of the plugins in the app 
- a boolean can be passed so that after a the element a Divider is rendered

Changes to the NavContentBlueprint: 
- extended the params so that a custom logo and Search component can be passed, previously an empty Sidebar was shown

image

### DIFF
--- a/plugins/app/src/extensions/AppNav/DefaultNavContent.test.tsx
+++ b/plugins/app/src/extensions/AppNav/DefaultNavContent.test.tsx
@@ -80,8 +80,8 @@ describe('DefaultNavContent', () => {
     expect(screen.getByText('Search')).toBeInTheDocument();
     expect(screen.queryByText('Delta')).not.toBeInTheDocument();
 
-    screen.getByText('Charlie');
     screen.getByText('Alpha');
+    screen.getByText('Charlie');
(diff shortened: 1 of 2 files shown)


Jan-HenningBacker2342:
Would improve the user experience for our users, too. Currently these options are missing in the new frontend

Issue #23002, opened 2024-02-15
🚀 Feature: Add `--pattern` to `backstage-cli info` command


awanlin:
### 🔖 Feature description

It would be pretty handy to help diagnose issues with Backstage for those who create plugins outside the `@backstage` org to be able to include a `--pattern` to the `backstage-cli info` command. This would show all the current results from this command as well as any packages that match the pattern.

### 🎤 Context

For plugin authors like @roadie or @spotify or @janus-idp being able to do something like:

```
yarn backstage-cli info --pattern "@spotify/backstage*"
```

And the results would include the matched pattern as well as the Backstage details would make troubleshooting much easier.

### ✌️ Possible Implementation

There's some prior art for this in the `yarn backstage-cli version:bump` command as it supports a similar flag.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

Rugvip:
Makes sense to have something for this! 👍 

I'm thinking it could make sense to include all packages with a `"backstage"` field in their `package.json` by default? Just for safety it probably makes sense to exclude packages in the monorepo itself, perhaps unless an `--include-workspace` flag is provided or smth like that?

awanlin:
I'd still like to see this feature and the possible solution that @Rugvip suggested would work fine. I'd work on this myself but I don't have the time. Happy to collaborate with someone! 🚀

awanlin:
Oh, this will come in more handy now given that the `community-repo` is live and the NPM organization has changed for them. They won't get listed by this command any more.

awanlin:
@Rugvip to a run at this and all we have is the lock file which won't have the package role:

https://github.com/backstage/backstage/blob/d57ebbca65fdfbd66b91bd3c8357d75ae1ecbee7/packages/cli/src/commands/info.ts#L49

Not sure the best way forward, digging into the Yarn packages is a bit of a rabbit hole. Any tips?

Rugvip:
[...]

PR #32275, opened 2026-01-09
feat: add --include and --output-file to backstage-cli-info

meganide:

This PR improves the `backstage-cli info` command with new output options and better package categorization.

Adresses #23002 .

---

### Changes

**New flags:**
- `--include <glob>` - Include additional packages matching glob patterns (e.g., `@spotify/*`, `@internal/backstage-*`)
- `--format <text|json>` - Output format, defaults to `text`. Use `json` for machine-readable output

**Output improvements:**
- Dependencies are now grouped into two sections: **Dependencies** (installed packages) and **Local** (workspace packages)
- Local workspace packages now display their actual version from `package.json` instead of showing `0.0.0-use.local`
- Packages with a `backstage` field in their `package.json` are automatically included in the output

### Examples

```bash
# Include additional package scopes
backstage-cli info --include "@spotify/*"

# Output as JSON
backstage-cli info --format json

# Export JSON to file
backstage-cli info --format json > backstage-info.json

# Export text to file
backstage-cli info --format text > backstage-info.txt

# Combine options
backstage-cli info --include "@internal/*" --format json > info.json
```

---

image

### DIFF
--- a/packages/cli/src/modules/info/commands/info.ts
+++ b/packages/cli/src/modules/info/commands/info.ts
@@ -147,14 +147,14 @@ export default async (options: InfoOptions) => {
       }
     }
 
-    // Helper to get version string for a package
-    const getVersions = (dep: string): string => {
+    // Helper to get versions for a package as an array of objects
+    const getVersions = (dep: string): Array<{ version: string }> => {
       const entries = lockfile.get(dep);
       if (!entries) {
-        return 'unknown';
+        return [{ version: 'unknown' }];
       }
       const versions = [...new Set(entries.map(i => i.version))];
-      return versions.join(', ');
+      return versions.map(v => ({ version: v }));
     };
 
     const sortedInstalled = [...installedDeps].sort();
@@ -170,7 +170,7 @@ export default async (options: InfoOptions) => {
         local: Object.fromEntries(
           sortedLocal.map(dep => [
             dep,
-            workspacePackages.get(dep) ?? 'unknown',
+            [{ version: workspacePackages.get(dep) ?? 'unknown' }],
           ]),
         ),
       };
@@ -192,7 +192,9 @@ export default async (options: InfoOptions) => {
     if (sortedInstalled.length > 0) {
       const maxLength = Math.max(...sortedInstalled.map(d => d.length));
       for (const dep of sortedInstalled) {
-        const versions = getVersions(dep);
+        const versions = getVersions(dep)
+          .map(v => v.version)
+          .join(', ');
         console.log(`  ${dep.padEnd(maxLength)} ${versions}`);
       }
     } else {


packages/cli/src/modules/info/index.ts
```diff
@@ -24,8 +24,24 @@ export default createCliPlugin({
...
+        const argv = await yargs()
+          .options({
+            include: {
+              type: 'string',
+              array: true,
+              default: [],
+              description:
+                'Glob patterns for additional packages to include (e.g., @spotify/backstage*)',
+            },
+            'output-file': {
```
freben: Is it better to have people do --format json > file.json and potentially skip the output file option? The current proposal mixes two things (format and target). Other commands like `config:print` already have a format option.

meganide: Great suggestion, added

packages/cli/src/modules/info/commands/info.ts
```diff
@@ -33,27 +65,149 @@ export default async () => {
...
+    }
+
+    // Helper to get version string for a package
+    const getVersions = (dep: string): string => {
+      const entries = lockfile.get(dep);
+      if (!entries) {
+        return 'unknown';
+      }
+      const versions = [...new Set(entries.map(i => i.version))];
+      return versions.join(', ');
```
freben: In the JSON version, should this always be an array (which often has a single element)? If you just return `versions` here and then do the `.join` in the text format, it all gets a lot easier to parse.

The JSON format could even be

```json
  "dependencies": {
    "@backstage-community/plugin-explore-common": {
      "versions": ["0.5.0", "0.6.1"]
    }
  },
```

or possibly

```json
  "dependencies": {
    "@backstage-community/plugin-explore-common": [
      { "version": "0.5.0" },
      { "version": "0.6.1" },
    ]
  },
```

to make room for more data in each, in the future.

freben (commented):
Alright, looking good, just a question about the JSON format since it's going to want to be a potentially stable API surface.

packages/cli/src/modules/info/commands/info.ts
meganide: Good catch, I opted for option 2.

docs/tooling/cli/03-commands.md
```diff
@@ -432,8 +432,79 @@ Usage: backstage-cli create-github-app <github-org>
...
+    "yarn": "3.6.0",
+    "cli": { "version": "0.27.0", "local": false },
+    "backstage": "1.20.0"
+  },
+  "dependencies": {
+    "@backstage/core-plugin-api": "1.8.0",
+    "@backstage/plugin-catalog": "1.15.0"
+  },
+  "local": {
+    "@mycompany/backstage-plugin-custom": "0.1.0"
```
freben, line 507: I wanted to merge to get this into the release, but you could follow up with a PR to update the docs to the new format.

Issue #32062, opened 2025-12-08
RFC: Modeling of MCP Servers in the Catalog


benjdlambert:
### 🔎 Search Terms

```plain
mcp servers, rfc, catalog model
```

### 🗃️ Project Area

Catalog

### 🔖 Need

As AI driven systems and MCP (Model Context Protocol) tools continue to grow across different organizations, we are seeing a clear need to track these resources in the Software Catalog. Today, various teams and community members model AI agents, MCP servers, and related components in different ways. While each implementation solves the problem locally, the lack of a shared standard leads to fragmentation across the ecosystem.

Without alignment, plugin authors must support multiple representations or choose one over another, which creates inconsistent tooling and makes adoption more difficult. This fragmentation affects several groups. Internally, it forces teams to reinvent modeling patterns and makes it harder to build shared infrastructure around AI and MCP systems. In the OSS ecosystem, contributors face incompatible formats that complicate plugin development and reduce interoperability across the community. This RFC is intended to start the conversation around how MCP resources should be represented in the Software Catalog so that plugin authors and adopters can align on a single common standard. The goal is an ecosystem where tools interoperate cleanly and users do not have to navigate incompatible formats

### 📝 Proposal

Below highlights the three different paths for modelling of MCP Servers in the Catalog model, the last of which is the proposed direction.

### Option 1: Reuse existing `kind`s

In this approach, we leverage existing kinds already supported by the Software Catalog and represent MCP Servers within those structures. This keeps the model simple and allows existing processors, plugins, and tooling to work without modification. As long as the metadata shape remains familiar, this path requires minimal additional complexity.
This option aligns with how other specifications, such as OpenAPI, are represented today. It reduces the learning curve for contributors and helps maintain consistency for adopters across different environments.
For example:
```yaml
apiVersion: backstage.io/v1alpha1
kind: API
metadata:
  name: backstage-mcp-actions
  description: Exposes tools related to the Backstage Ecosystem such as the Software Catalog
  title: Backstage MCP Server
  tags:
    - software-catalog
    - backstage
spec:
  type: mcp-server
  lifecycle: experimental
  visibility: private
  owner: backstage
  definition: |
    name: com.internal/backstage-mcp-actions
    description: "Exposes tools related to the Backstage Ecosystem such as the Software Catalog"
    version: "0.1.0"
    remotes:
      - type: streamable-http
        url: "http://internal.backstage.company:7007/api/mcp-actions/v1"
```

[...]

PR #32279, opened 2026-01-10
feat(actions): add prompts and resources support to MCP integration

westafari:
## Summary

This PR adds prompts and resources support to the Actions Registry and MCP integration.

## Why This Change

Backstage should present the same unified context layer to AI clients as it does to humans. Currently, the MCP actions integration only exposes tools (callable actions) without context. AI clients can call actions but don't understand what's available in the catalog or how to use plugins effectively.

Dynamic registration of prompts and other MCP-facing resources enables this unified context layer. Prompts provide natural language guidance about plugin capabilities, while resources allow AI clients to browse and read data (like catalog entities) the same way humans can through the UI.

This attempts to address the limitation noted in RFC #30218. With the MCP SDK (v1.25.2) now offering more stable APIs for prompts and resources, it seemed like a good time to explore this.

## What Changed

- Actions Registry and Actions Service now support registering and listing prompts and resources
- MCP Actions Backend exposes prompts and resources via the MCP protocol
- Catalog backend includes example implementations (explore-catalog prompt, catalog entities resource)

## Test Plan

Tested with an MCP client:
- Listed available resources successfully
- Read catalog entities by kind and viewed summary stats
- Used prompts for contextual guidance
- Verified existing actions continue working alongside prompts/resources

Example: AI can now ask "What's in the catalog?", read `catalog://summary`, and respond with actual entity counts.

## Related Work

- Builds on #32134 (action filtering)
- Addresses RFC #30218
- Complements RFC #32062 (MCP server catalog modeling)

---

- [ ] Screenshots attached (for UI changes) - N/A
- [x] All your commits have a `Signed-off-by` line in the message

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -51120,7 +51120,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"zod-to-json-schema@npm:^3.25.0, zod-to-json-schema@npm:^3.25.1":
+"zod-to-json-schema@npm:^3.20.4, zod-to-json-schema@npm:^3.25.0, zod-to-json-schema@npm:^3.25.1":
   version: 3.25.1
   resolution: "zod-to-json-schema@npm:3.25.1"
   peerDependencies:


.changeset/prompt-resources-support.md
awanlin, line 1: Not sure if you looked at any of the existing [changelogs](https://github.com/backstage/backstage/blob/master/packages/backend-defaults/CHANGELOG.md) or the [guides on good/bad changelogs](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md#writing-changesets) but this is way more than is needed. I'd remove everything from line 11 on. Anything important here looks to be in the `README` changes.

awanlin (changes requested):
Thanks @westafari, just did a quick scan of the docs changes and the changeset caught my eye.

.changeset/prompt-resources-support.md
westafari: thx @awanlin trimmed it https://github.com/backstage/backstage/pull/32279/commits/67d5a3e0bebefbae312284b460b127bcad2f9fbe

packages/backend-defaults/src/alpha/entrypoints/actions/DefaultActionsService.ts
```diff
@@ -120,6 +122,109 @@ export class DefaultActionsService implements ActionsService {
     return { output };
   }
 
+  async listPrompts({ credentials }: { credentials: BackstageCredentials }) {
```
benjdlambert: I don't think that prompts and resources should live on the `ActionsRegistry`.

These are purely MCP concerns, and are not something that's going to be a reusable across different consumers.

I think that a better way to implement this would be through an `extensionPoint` on the `mcp-actions-backend` plugin or something that you can register these prompts and resources with.

Another solution would be to have a separate service for these concepts instead similar to the `ActionsService` but named something else to collect these things from plugins.

There's also another parallel of prompts and how they should be modelled, could they actually live in the Catalog working off the back of #32062? Not sure currently, do you have any opinions here?

westafari: Hi @benjdlambert thanks for taking a look at this! I was coming at it from the perspective that since `ActionsRegistry` is a centralized plugin registry/capability discovery mechanism and `mcp-actions-backend` is a consumer it should be able to discover those capabilities there (prompts, resources). I can look at the `extensionPoint` idea but I think the thing that immediately comes to mind is why register some capabilities in `ActionsRegistry` and others via the extension? Now we have 2 places where we are registering tools, etc, which seemed counter to the whole idea behind the actions registry (at least my interpretation of it 😄 ) 

As far as modeling MCPs in the catalog we have taken the approach of treating them like APIs - tools exposed by the MCP are the API definition visible in Backstage. This is a little brutish but it does enable Backstage to answer the "hey, do I have any MCPs in my catalog and what do they do?" question for humans and AI clients.

benjdlambert: I would say that our intention behind the `ActionsRegistry` is that a plugin can define an action which can be executed in a lot of different contexts not just the `mcp-actions-backend` plugin which exposes these as tools in an MCP Server. 

The `ActionsRegistry` I think, really should have zero coupling between any MCP specific concerns, and that's what the `mcp-actions-backend` is for.

I don't mind maybe another separate service to hold `Resources` and `Prompts` and whatever might come along eventually, or as mentioned an extension point, they seem fine. I'm not sure that we should add `Tools` to that new thing though, I think that they should be registered with the `ActionsRegistry`, but for these more specific Model Context Protocol specific things I think it makes sense?

benjdlambert: Also would like some thoughts from other @backstage/maintainers here too?

westafari: @benjdlambert refactored to use an `extensionPoint`

Issue #32253, opened 2026-01-06
Catalog - More advanced filtering - excludes, not, in, etc.


awanlin:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
exclude, not, in, contains, filter, catalog api, catalog filter
```

### 🗃️ Project Area

Catalog

### 🔖 Need

There's often cases where you'd like to do a more advanced filtering of data within the Catalog. The first case that comes to mind is being able to do excludes or not but being able to do in, contains, etc. would be super helpful as well.

### 📝 Proposal

We introduced support for [entity predicates for filtering in the new frontend system](https://github.com/backstage/backstage/pull/28951), this is super powerful and well designed. Using this same approach for the backend would fill the needs of this feature request. 

Here's a simple example:

```yaml
filter:
  kind: component
  spec.type:
    $in: [service, website]
```

And here's a more advanced one:

```yaml
filter:
  $all:
  - kind: component
  - $any:
    - spec.type: service
    - spec.type: website
      metadata.annotations.acme.org/discovery-name: { $exists: true }
  - $not:
      spec.lifecycle: production
  - relations:
      $contains:
        type: ownedBy
        target: group:default/admins
```

Though these examples are YAML from the linked PR the same concept could be applied when doing this via code.

### 🔄 Alternatives

We could look at other options like adding in just excludes or not as that's seems to be the most frequently requested filter. Though this was not a path forward the Catalog Maintainers really want to head down as noted in this PR comment: https://github.com/backstage/backstage/pull/30872#issuecomment-3182462114. They point towards the solution proposed above.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

No, I don't have time to work on this right now

awanlin:
FYI @drodil, logged this as it came up in conversation with another Adopter and thought you might want to track it.

freben:
Yes! And switching to a POST (or, if one wants to be fancy, the new/upcoming QUERY verb) to not have to worry about url length issues.

Most of the predicate syntax features should be OK to implement. However `$contains` may be tricky or need some special treatment.

ishaanxgupta:
Hi @freben @awanlin would love to work on this issue, can you please assign me this?

[...]

PR #32282, opened 2026-01-11
feat: ✨ add predicate-based entity filtering

Nilay1999:

### DIFF
--- a/plugins/catalog-backend/src/service/AuthorizedEntitiesCatalog.ts
+++ b/plugins/catalog-backend/src/service/AuthorizedEntitiesCatalog.ts
@@ -22,6 +22,7 @@ import {
 import { Entity, stringifyEntityRef } from '@backstage/catalog-model';
 import { AuthorizeResult } from '@backstage/plugin-permission-common';
 import { ConditionTransformer } from '@backstage/plugin-permission-node';
+import { FilterPredicate } from '@backstage/filter-predicates';
 import {
   Cursor,
   EntitiesBatchRequest,
@@ -42,6 +43,7 @@ import {
   BackstageCredentials,
   PermissionsService,
 } from '@backstage/backend-plugin-api';
+import { convertEntityFilterToPredicate } from './request/convertEntityFilterToPredicate';
 
 export class AuthorizedEntitiesCatalog implements EntitiesCatalog {
   private readonly entitiesCatalog: EntitiesCatalog;
@@ -141,34 +143,31 @@ export class AuthorizedEntitiesCatalog implements EntitiesCatalog {
     }
 
     if (authorizeDecision.result === AuthorizeResult.CONDITIONAL) {
-      const permissionFilter: EntityFilter = this.transformConditions(
-        authorizeDecision.conditions,
+      const permissionFilterPredicate = convertEntityFilterToPredicate(
+        this.transformConditions(authorizeDecision.conditions),
       );
 
-      let permissionedRequest: QueryEntitiesRequest;
-      let requestFilter: EntityFilter | undefined;
+      const source = isQueryEntitiesCursorRequest(request)
+        ? request.cursor
+        : request;
 
-      if (isQueryEntitiesCursorRequest(request)) {
-        requestFilter = request.cursor.filter;
+      const { filter, query } = source;
+      const combinedQuery = this.mergePredicates(
+        permissionFilterPredicate,
+        filter ? convertEntityFilterToPredicate(filter) : undefined,
+        query,
+      );
 
-        permissionedRequest = {
-          ...request,
-          cursor: {
-            ...request.cursor,
-            filter: request.cursor.filter
-              ? { allOf: [permissionFilter, request.cursor.filter] }
-              : permissionFilter,
-          },
-        };
-      } else {
-        permissionedRequest = {
-          ...request,
-          filter: request.filter
-            ? { allOf: [permissionFilter, request.filter] }
-            : permissionFilter,
-        };
-        requestFilter = request.filter;
-      }
+      const permissionedRequest = isQueryEntitiesCursorRequest(request)
+        ? {
+            ...request,
+            cursor: {
+              ...request.cursor,
+              filter: undefined,
+              query: combinedQuery,
+            },
+          }
+        : { ...request, filter: undefined, query: combinedQuery };
 
       const response = await this.entitiesCatalog.queryEntities(
         permissionedRequest,
@@ -176,12 +175,14 @@ export class AuthorizedEntitiesCatalog implements EntitiesCatalog {
 
       const prevCursor: Cursor | undefined = response.pageInfo.prevCursor && {
         ...response.pageInfo.prevCursor,
-        filter: requestFilter,
+        filter,
+        query,
       };
 
       const nextCursor: Cursor | undefined = response.pageInfo.nextCursor && {
         ...response.pageInfo.nextCursor,
-        filter: requestFilter,
+        filter,
+        query,
       };
 
       return {
@@ -334,5 +335,12 @@ export class AuthorizedEntitiesCatalog implements EntitiesCatalog {
           : this.findParents(parentRef, allAncestryItems, newSeenEntityRefs),
       ),
     ];
+  }
+
+  private mergePredicates(
+    ...predicates: (FilterPredicate | undefined)[]
+  ): FilterPredicate {
+    const defined = predicates.filter(Boolean) as FilterPredicate[];
+    return defined.length === 1 ? defined[0] : { $all: defined };
   }
 }
--- a/plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
+++ b/plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
@@ -129,6 +129,14 @@ function applyPredicateInStrategy(
 
   // Handle $all (AND)
   if (isAllPredicate(filter)) {
+    // Empty $all array is vacuous truth - matches all entities
+    if (filter.$all.length === 0) {
+      if (negate) {
+        return targetQuery.andWhere(knex.raw('1 = 0'));
+      }
+      return targetQuery;
+    }
+
     return targetQuery[negate ? 'andWhereNot' : 'andWhere'](
       function allFilter() {
         for (const subFilter of filter.$all) {
@@ -148,6 +156,14 @@ function applyPredicateInStrategy(
 
   // Handle $any (OR)
   if (isAnyPredicate(filter)) {
+    // Empty $any array matches nothing
+    if (filter.$any.length === 0) {
+      if (negate) {
+        return targetQuery;
+      }
+      return targetQuery.andWhere(knex.raw('1 = 0'));
+    }
+
     return targetQuery[negate ? 'andWhereNot' : 'andWhere'](
       function anyFilter() {
         for (const subFilter of filter.$any) {
--- a/plugins/catalog-backend/src/service/request/convertEntityFilterToPredicate.test.ts
+++ b/plugins/catalog-backend/src/service/request/convertEntityFilterToPredicate.test.ts
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { convertEntityFilterToPredicate } from './convertEntityFilterToPredicate';
+import { EntityFilter } from '@backstage/plugin-catalog-node';
+
+describe('convertEntityFilterToPredicate', () => {
+  it('converts simple key-value filter', () => {
+    const filter: EntityFilter = {
+      key: 'kind',
+      values: ['component'],
+    };
+
+    expect(convertEntityFilterToPredicate(filter)).toEqual({
+      kind: 'component',
+    });
+  });
+
+  it('converts key with multiple values to $in', () => {
+    const filter: EntityFilter = {
+      key: 'spec.type',
+      values: ['service', 'website'],
+    };
+
+    expect(convertEntityFilterToPredicate(filter)).toEqual({
+      'spec.type': { $in: ['service', 'website'] },
+    });
+  });
+
+  it('converts key without values to $exists', () => {
+    const filter: EntityFilter = {
+      key: 'metadata.tags',
+    };
+
+    expect(convertEntityFilterToPredicate(filter)).toEqual({
+      'metadata.tags': { $exists: true },
+    });
+  });
+
+  it('converts allOf to $all', () => {
+    const filter: EntityFilter = {
+      allOf: [
+        { key: 'kind', values: ['component'] },
+        { key: 'spec.type', values: ['service'] },
+      ],
+    };
+
+    expect(convertEntityFilterToPredicate(filter)).toEqual({
+      $all: [{ kind: 'component' }, { 'spec.type': 'service' }],
+    });
+  });
+
+  it('converts anyOf to $any', () => {
+    const filter: EntityFilter = {
+      anyOf: [
+        { key: 'spec.owner', values: ['team-a'] },
+        { key: 'spec.owner', values: ['team-b'] },
+      ],
+    };
+
+    expect(convertEntityFilterToPredicate(filter)).toEqual({
+      $any: [{ 'spec.owner': 'team-a' }, { 'spec.owner': 'team-b' }],
+    });
+  });
+
+  it('converts not to $not', () => {
+    const filter: EntityFilter = {
+      not: { key: 'spec.lifecycle', values: ['experimental'] },
+    };
+
+    expect(convertEntityFilterToPredicate(filter)).toEqual({
+      $not: { 'spec.lifecycle': 'experimental' },
+    });
+  });
+
+  it('converts nested complex filters', () => {
+    const filter: EntityFilter = {
+      allOf: [
+        { key: 'kind', values: ['component'] },
+        {
+          anyOf: [
+            { key: 'spec.type', values: ['service'] },
+            { key: 'spec.type', values: ['website'] },
+          ],
+        },
+        {
+          not: { key: 'spec.lifecycle', values: ['experimental'] },
+        },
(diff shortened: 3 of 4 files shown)


Nilay1999:
PR is related to suggested changes for advanced filter by @awanlin 
https://github.com/backstage/backstage/issues/32253

plugins/catalog-backend/src/schema/openapi/generated/models/EntityPredicate.model.ts
```diff
@@ -0,0 +1,36 @@
...
+import { EntityPredicateValue } from '../models/EntityPredicateValue.model';
+
+/**
+ * A predicate-based filter supporting logical operators. - $all: All conditions must match (AND) - $any: At least one condition must match (OR) - $not: Negates the condition - $exists: Check if field exists - $in: Match any value in array
+ * @public
+ */
+export type EntityPredicate =
+  | EntityPredicateOneOf
+  | EntityPredicateOneOf1
+  | EntityPredicateOneOf2
```
freben: see slightly related discussion [here](https://github.com/backstage/backstage/pull/32211#discussion_r2668649744) about breaking apart into named specific types

plugins/catalog-backend/src/schema/openapi/generated/apis/Api.server.ts
```diff
@@ -198,6 +213,8 @@ export type EndpointMap = {
 
   '#get|/entities': GetEntities;
 
+  '#post|/entities/by-predicates': GetEntitiesByPredicates;
```
freben: I don't think it needs a dedicated endpoint name. It could be a `POST .../by-query` as a "competing" (alternative) `filter`. Just one more possible way of doing filtering into an existing framework, essentially. 

The body of that POST could be basically exactly what the existing by-query accepts but as a JSON object, except it doesn't have a `filter` field but instead a `query` field.

The other question then is what to do about callers. What should we do about the `CatalogApi` / `CatalogService` types? For every place [in this file](https://github.com/backstage/backstage/blob/52d6767865825de8c50c6d598a807a56ec12771e/packages/catalog-client/src/types/api.ts) where there is a `filter?: EntityFilter` argument, there could be a `query?: EntityPredicateQuery` next to it. We could start with `QueryEntitiesInitialRequest` of course. Then, that will mean that the cursor returned from that endpoint also needs to optionally encode a query inside it.

The client bits must know when to call the GET endpoint and when to call the POST endpoint, based on what the inputs were.

freben (commented):
Thanks! Just some quick thoughts to begin with

plugins/catalog-backend/src/schema/openapi/generated/apis/Api.server.ts
Nilay1999: I was also considering making it a POST by-query endpoint
Based on your suggestion, I’ll refactor it and check the compatibility with Catalog APIs.

Nilay1999: Since we’re supporting cursor-based pagination on the POST endpoint too, we’ll likely need a way (for example, a flag) to identify which endpoint a particular cursor was generated from.

also, i'm assuming that we will only allow query params - limit, offset, after in POST endpoint.

Nilay1999: @freben @vinzscam
The CatalogClient has been updated to support advanced search filtering. Please review and let me know your thoughts.

plugins/catalog-node/src/api/common.ts
kurtaking, line 1: Should these go into `catalog-common` package as part of this change? It would skip one round of `@deprecations`. 

**edit:** noticed slight differences in `EntityPredicateValue` right now between `catalog-react` and here.

nojaf:
Would this PR also cover `$contains` operator?
And what about relationship-based predicates?

Nilay1999:
> Would this PR also cover `$contains` operator? And what about relationship-based predicates?

Nope, it doesn't have `$contains` support right now

plugins/catalog-node/src/api/common.ts
Nilay1999: Actually backend api doesn't have support of `$contains` filter, so that's why i have kept types in different location.

nojaf:
Okay, any pointers on what we need to be done for that?
Would be happy to help out if I can be of use!

Nilay1999:
> Okay, any pointers on what we need to be done for that? Would be happy to help out if I can be of use!

You can check this thread here : https://github.com/backstage/backstage/issues/32253#issuecomment-3733400589
@freben has commented on `$contains` filter

freben:
@Nilay1999 FYI I just made https://github.com/backstage/backstage/pull/32631 which puts these predicate types out in the open, in a separate package. Hoping to get it merged fairly soon, then we can adapt this PR to that

benjdlambert:
@freben we should be able to use that package now part of this PR right?

packages/catalog-client/src/CatalogClient.ts
```diff
@@ -264,6 +292,72 @@ export class CatalogClient implements CatalogApi {
...
+
+    const result = await this.requestRequired(response);
+
+    return {
+      items: result.items,
+      totalItems: result.items.length,
+      pageInfo: {
+        nextCursor: result.pageInfo?.nextCursor,
+      },
+    };
```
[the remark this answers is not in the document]
Nilay1999: totalItems requires a full count over the filtered result set (COUNT(*) on a derived filtered subquery). On catalogs with tens of thousands of entities this is noticeably more expensive than a simple paginated fetch, which is why the new predicate POST /entities/by-query endpoint deliberately omits totalItems.

Nilay1999:
@freben I have replaced all predicate types with @backstage/filter-predicates types. Currently going through co-pilot's comments. Can you also take a look ?

plugins/catalog-backend/src/schema/openapi.yaml
```diff
@@ -1074,6 +1154,98 @@ paths:
...
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/Entity'
+                    description: The list of entities matching the predicate filter.
+                  pageInfo:
+                    type: object
+                    properties:
+                      nextCursor:
+                        type: string
+                        description: The cursor for the next batch of entities.
```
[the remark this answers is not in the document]
Nilay1999: TotalItems calculation is not included right now due to heavy SQL operations

packages/catalog-client/src/types/index.ts
```diff
@@ -40,4 +40,13 @@ export type {
...
 } from './api';
+export type {
+  EntityPredicate,
+  EntityPredicateAll,
+  EntityPredicateAny,
+  EntityPredicateNot,
+  EntityPredicateValue,
+  EntityPredicateExists,
+  EntityPredicateIn,
+} from './predicate';
```
freben: Remove - these should no longer be exported; the only public facing API should be out of the new filter-predicates package

plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
```diff
@@ -213,6 +217,156 @@ export class DefaultEntitiesCatalog implements EntitiesCatalog {
     };
   }
 
+  async queryEntitiesByPredicate(
```
freben: Hm my hope was that we would not need to add a separate method for this. Instead, inside `queryEntities`, [where it calls](https://github.com/backstage/backstage/blob/a92426920f8e39c21afa466b28ccf8b0608b1e19/plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts#L308) `applyEntityFilterToQuery` should be the only place that needs to change. So it'll be

```ts
        if (cursor.filter || cursor.query) {
          applyEntityFilterToQuery({
            filter: cursor.filter,
            query: cursor.query,
            ...
```

and then that helper knows what to do about them (and asserts that only one of them is given). Its implementation would probably leave the old `applyInStrategy` untouched and instead make a parallel implementation for the query next to it.

plugins/catalog-backend/src/service/createRouter.ts
```diff
@@ -254,6 +256,56 @@ export async function createRouter(
           throw err;
         }
       })
+      .post('/entities/by-query', async (req, res) => {
+        const auditorEvent = await auditor.createEvent({
+          eventId: 'entity-fetch',
+          request: req,
+          meta: {
+            queryType: 'by-query-predicate',
```
freben: can probably treat this as the same one as the old call, it's just "semantics" so to speak whether it's the get or the post form

plugins/catalog-backend/src/service/createRouter.test.ts
```diff
@@ -61,6 +61,7 @@ describe('createRouter readonly disabled', () => {
   beforeEach(async () => {
     entitiesCatalog = {
       entities: jest.fn(),
+      queryEntitiesByPredicate: jest.fn(),
```
freben: as mentioned in the other comment, let's try to get rid of all of these

plugins/catalog-backend/src/service/AuthorizedEntitiesCatalog.ts
```diff
@@ -196,6 +197,36 @@ export class AuthorizedEntitiesCatalog implements EntitiesCatalog {
     return this.entitiesCatalog.queryEntities(request);
   }
 
+  async queryEntitiesByPredicate(
+    request?: EntityPredicateRequest,
+  ): Promise<EntitiesResponse> {
+    if (!request) {
+      return {
+        entities: { type: 'object', entities: [] },
+        pageInfo: { hasNextPage: false },
```
freben: I'll just mention here that it's a big decision to exclude the total count. We should probably not do that in this PR, since from the client caller's point of view it'll be unclear why sometimes it's there and sometimes it's not. I would prefer if we in the future added a `count: false` parameter to opt out when you don't care about it.

If we do as said in other comments and unify on the same old query methods, then the count will of course come back automatically.

plugins/catalog-backend/src/service/AuthorizedEntitiesCatalog.ts
```diff
@@ -196,6 +197,36 @@ export class AuthorizedEntitiesCatalog implements EntitiesCatalog {
...
+    if (authorizeDecision.result === AuthorizeResult.DENY) {
+      return {
+        entities: { type: 'object', entities: [] },
+        pageInfo: { hasNextPage: false },
+      };
+    }
+
+    // Note: For CONDITIONAL results, we pass through to the underlying catalog
+    // since EntityPredicate filters are separate from EntityFilter permission conditions.
+    // The permission filter would need to be applied separately if needed.
```
freben: No, we can't do that. The code will have to form a query expression based on the inputs + the conditional rules. It's quite important to obey the permissions framework. And yes it'll have to be prepared to translate both the old format and the new format. Yes that will be a bit of code :)

But honestly, is it perhaps the case that we should upfront - all the way out in the service handlers - ALWAYS translate instantly to the query format if the old filter format was given? And then the whole database engine behind that, only speaks the query format.

freben (changes requested):

plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
Nilay1999: okay so in that case, we only process only one of them right ? Because if POST api is calling this method, we would process cursor.query only, and for GET api, filters will be processed

Nilay1999: Like this

```typescript

export function applyEntityFilterToQuery(options: {
  filter?: EntityFilter;
  query?: FilterPredicate;
  targetQuery: Knex.QueryBuilder;
  onEntityIdField: string;
  knex: Knex;
  strategy?: 'in' | 'join';
}): Knex.QueryBuilder {
  const { filter, targetQuery, onEntityIdField, knex, query } = options;

  if (query) {
    return applyPredicateInStrategy(
      query,
      targetQuery,
      onEntityIdField,
      knex,
      false,
    );
  }
  if (!filter) {
    throw new InputError('Either filter or query must be provided');
  }
  return applyInStrategy(filter, targetQuery, onEntityIdField, knex, false);
}

```

Nilay1999: Just to confirm the intended behavior here:

Priority handling: If a catalogClient were to pass both filterPredicate (in options) and filter params, the filterPredicate would take priority and the filter params would be ignored, correct?
POST endpoint restrictions: For the POST endpoint (when using filterPredicate), we're excluding filter, fields, and fullTextFilter options, allowing only:

query (the filterPredicate itself)
orderFields (for sorting)
cursor (for pagination)
limit (for pagination)

freben: Yeah you would be allowed to pass in none, or one, but not both. Right?

We COULD make a translation thing such that you can pass in both and there's an implicit AND between them but that just makes things more muddy I'd say.

I don't think the post endpoint would be less featureful at all. I would say that the POST endpoint should be a superset of the GET endpoint, such that in the future for example the catalog page in the frontend is likely to just use the POST one for all of its needs.

The old GET one is inherently limited and a bit silly, and should perhaps be confined to the historical baggage heap, to be used by those that just want to curl a bit, or get everything without filtering and is fine with some limitations.

Nilay1999: Okay, Got it. then i will refactor the POST endpoint's SQL a bit, in order to support all the feature of GET as well as predicate based filter (If both provided)

freben:
FYI we just merged https://github.com/backstage/backstage/pull/32846 that does the same for location queries. Note how it chose to back out of the modelling of the exact predicate shapes, instead relying on local zod only. It also applies permissions checks fully etc.

In the future we may revisit the predicate schema modelling question, but maybe it'd be best if that happened after there existed a way of referring to centrally defined types, so that all packages that want to use predicates don't have to copy paste the rather complex definitions.

plugins/catalog-backend/src/service/AuthorizedEntitiesCatalog.ts
Nilay1999: I’ve refactored the queryEntities method so that it now supports both query and filter, with the constraint that they remain mutually exclusive.

Additionally, I’ve put together an initial (rough) implementation of a permission converter to transform the existing EntityFilter structure into the new FilterPredicate format.

I don’t have deep expertise in the permissions model yet, but this is my current understanding and an attempt to align with the expected behavior 😅 

```typescript
function isEntitiesSearchFilter(
  filter: EntityFilter,
): filter is { key: string; values?: string[] } {
  return 'key' in filter;
}

/**
 * Type guard to check if a filter is an allOf filter
 */
function isAllOfFilter(
  filter: EntityFilter,
): filter is { allOf: EntityFilter[] } {
  return 'allOf' in filter;
}

/**
 * Type guard to check if a filter is an anyOf filter
 */
function isAnyOfFilter(
  filter: EntityFilter,
): filter is { anyOf: EntityFilter[] } {
  return 'anyOf' in filter;
}

/**
 * Type guard to check if a filter is a not filter
 */
function isNotFilter(filter: EntityFilter): filter is { not: EntityFilter } {
  return 'not' in filter;
}

/**
 * Converts an EntityFilter to a FilterPredicate.
 *
 * EntityFilter format (old):
 * - { allOf: EntityFilter[] }
 * - { anyOf: EntityFilter[] }
 * - { not: EntityFilter }
 * - { key: string, values?: string[] }
 *
 * FilterPredicate format (new):
 * - { $all: FilterPredicate[] }
 * - { $any: FilterPredicate[] }
 * - { $not: FilterPredicate }
 * - { [field]: value | { $in: value[] } | { $exists: boolean } }
 *
 * @param filter - The EntityFilter to convert
 * @returns The equivalent FilterPredicate
 * @public
 */
export function convertEntityFilterToPredicate(
  filter: EntityFilter,
): FilterPredicate {
  // Handle allOf -> $all
  if (isAllOfFilter(filter)) {
    return {
      $all: filter.allOf.map(f => convertEntityFilterToPredicate(f)),
    };
  }

  // Handle anyOf -> $any
  if (isAnyOfFilter(filter)) {
    return {
      $any: filter.anyOf.map(f => convertEntityFilterToPredicate(f)),
    };
  }

  // Handle not -> $not
  if (isNotFilter(filter)) {
    return {
      $not: convertEntityFilterToPredicate(filter.not),
    };
  }

  // Handle EntitiesSearchFilter -> field expression
  if (isEntitiesSearchFilter(filter)) {
    const { key, values } = filter;

    if (!values || values.length === 0) {
      return { [key]: { $exists: true } } as FilterPredicate;
    }

    if (values.length === 1) {
      return { [key]: values[0] } as FilterPredicate;
    }

    return { [key]: { $in: values } } as FilterPredicate;
  }

  throw new Error(`Unsupported EntityFilter format: ${JSON.stringify(filter)}`);
}
```

Nilay1999: I guess POST /by-query endpoint needs to handle permission in below manner

User A has access to see only `component` & `service` Entities.
If input is given like, entity count will 0
```json
{
    "query": {
        "$all": [
            { "kind": "component" },
            { "kind": "api" }
        ]
    }
}
```
but it will return `component` entities if query given like this as user has access to component

```json
{
    "query": {
        "$any": [
            { "kind": "component" },
            { "kind": "api" }
        ]
    }
}
```

NOTE: `AuthorizedEntitiesCatalog` will convert every filter inputs to query format in above approach.

freben: The query applies to each individual entity. So the first one can never match, because a single entity would have to be both a component and an api.

Nilay1999: yeah mb, first query would be 

```json
{
  "query": {
    "$any": [
      { "kind": "api" },
      { "kind": "user" }
    ]
  }
}
```

User A tries to fetch `api` or `user` entities which are not allowed in permission.

Nilay1999:
I’ve currently implemented the filter-to-query conversion in `AuthorizedEntitiesCatalog.ts`. However, I’m not entirely confident this is the right place for it — we could also handle this conversion at the service layer (e.g., in `plugins/catalog-backend/src/service/request/parseQueryEntitiesParams.ts`).

freben:
Hold your horses for a bit - I'm working actively to fix my own comments on the other one. We've retained your commits though to ensure you get credit for the work

benjdlambert:
@Nilay1999 gonna close this PR as your commits got moved into the other branch and then we all jumped on getting it through! Thanks for your effort and work here! Much appreciated! :tada:

Nilay1999:
@benjdlambert @freben

Thanks so much! Really appreciate the collaboration and guidance. Glad the feature made it through 😄 – looking forward to contributing more!  🚀

freben:
Now we need to add it to other places too! Like the facets endpoint! QUERY ALL THE THINGS

Nilay1999:
Sounds great! I'll definitely look into it. 🚀

PR #32292, opened 2026-01-11
Migrate techdocs-cli-embedded-app to NFS

GabDug:

Migrate the Techdocs CLI embedded app to the NFS.

No obvious regressions:
- Automatically logged-in
- Same mock APIs loaded
- Same addons loaded:
  - Theme switcher addon 
  - Live Reload addon 
- Reader Page: still shows "Live preview environment"
- Nav: Still shows "Docs Preview" & Logo

I've added `withSearch` & `withHeaders` configs to the techdocs alpha blueprints, to avoid redefining a custom layout in the techdocs-cli-embedded-app, please let me know if ou see a better approach!

image

### How to test?

1. `yarn --cwd packages/techdocs-cli-embedded-app build `
2. `cd packages/techdocs-cli/src/example-docs && techdocs-cli serve`

- ~~[ ] Added or updated documentation~~
- [X] Update tests

### DIFF
--- a/.changeset/fuzzy-suits-sit.md
+++ b/.changeset/fuzzy-suits-sit.md
@@ -2,7 +2,7 @@
 '@backstage/plugin-techdocs': minor
 ---
 
-Add two config values to the `page:techdocs/reader` extension that configure default layout, `withSearch` and `withHeader`. Default are unchanged to `true`.
+Add two config values to the `page:techdocs/reader` extension that configure default layout, `withoutSearch` and `withoutHeader`. Default are unchanged to `false`.
 
 E.g. to disable the search and header on the Techdocs Reader Page:
 
@@ -11,6 +11,6 @@ app:
   extensions:
     - page:techdocs/reader:
         config:
-          withSearch: false
-          withHeader: false
+          withoutSearch: true
+          withoutHeader: true
 ```
--- a/plugins/techdocs/src/alpha/index.tsx
+++ b/plugins/techdocs/src/alpha/index.tsx
@@ -158,18 +158,12 @@ const techDocsReaderPage = PageBlueprint.makeWithOverrides({
   },
   config: {
     schema: {
-      withSearch: z => z.boolean().default(true),
-      withHeader: z => z.boolean().default(true),
+      withoutSearch: z => z.boolean().default(false),
+      withoutHeader: z => z.boolean().default(false),
     },
   },
-  factory(originalFactory, { inputs, config }) {
+  factory(originalFactory, { apis, inputs, config }) {
     const addonsApi = apis.get(techdocsAddonsApiRef);
-    const addons = inputs.addons.map(output => {
-      const options = output.get(AddonBlueprint.dataRefs.addon);
-      const Addon = options.component;
-      attachTechDocsAddonComponentData(Addon, options);
-      return <Addon key={options.name} />;
-    });
 
     return originalFactory({
       path: '/docs/:namespace/:kind/:name',
@@ -195,8 +189,8 @@ const techDocsReaderPage = PageBlueprint.makeWithOverrides({
         return import('../Router').then(({ TechDocsReaderRouter }) => (
           <TechDocsReaderRouter>
             <TechDocsReaderLayout
-              withSearch={config.withSearch}
-              withHeader={config.withHeader}
+              withSearch={!config.withoutSearch}
+              withHeader={!config.withoutHeader}
             />
             <TechDocsAddons>{addons}</TechDocsAddons>
           </TechDocsReaderRouter>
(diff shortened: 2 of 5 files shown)


GabDug:
> Hi @GabDug, thought the New Frontend System is Adoption Ready we still need to make these changes opt-in for now. I did a quick look and that doesn't seem to be the case here, I may be wrong though but want to point that out.

Hey @awalin, as the techdocs-cli-embedded-app is not published, but only consumed by the Techdocs CLI I'm not sure it makes much sense as an opt-in.

In case adopters ship a custom embedded-app in a custom Techdocs CLI, there's no breaking change: they most likely copied a previous version, and are free to keep their own version based on the previous frontend system, or can take this one as an example.

awanlin:
Hi @GabDug, the issue is it breaks expectations and the path we've gone down, as best as possible, to make New Frontend System (NFS) something people opt-in to. This feels like something we would do when NFS ships as the default in `create-app` which we aren't near. 

If there is zero difference between the current version and NFS why do we need to ship this now? What additional value will you get as an end users? That would help me better understand the need and help convince me to ship this change 👍

GabDug:
Hi @awalin,

I was under the impression that new or internal things could adopt the NFS, without the same opt-in considerations needed for pre-existing plugins and public interfaces, but I fully understand your point. While I don't think this brings any code breaking change, as the embedded app has no public API, you're right to point out that this could break expectations. 

> why do we need to ship this now

This is a maintenance task more than a feature, so there's absolutely 0 need to ship it now: it can be revisited later, once NFS is deemed ready for broader adoption. We've fully adopted NFS for a few months now, including in our techdocs-cli-embedded-app, so I just thought it would be nice to share, to remove a ~~legacy~~ non-NFS app. 

> What additional value will you get as an end users?

No direct value appart from providing a NFS supported techdocs-cli-embeeded-app. 
Using the NFS for the embedded app _could_ be an enabler, for instance to provide a dedicated interface to build a customized embedded-app, accepting user-provided theme/addons, without copying the whole app. This is out of scope or this PR and still hypothetical, so I'll close this PR for now.

awanlin:
Hi @GabDug, thanks for the follow up. This helps me understand your reasoning much better. My intention was not to get you to close this PR but have some further discussion on the topic. The reality is that we do need to make this change eventually. Perhaps doing this now does make sense and helps avoid TechDocs from falling behind when this change is needed. I'd like to reopen this PR and get feedback from the Core Maintainers and the TechDocs team to decide on next steps for it. How does that sound to you?

GabDug:
@awanlin No issue to re-open and get more feedback! As you mentioned, we'll have to do it eventually, but it's not a pressing issue so I'll let the maintainers be the judge of the timeline.

My apologies for the initial PR that lacked context, I thought of this as yet another maintenance task and did not take into account the current state of the NFS.

awanlin:
Thanks! I've reached out to the Core Maintainers and TechDocs maintainers about this 👍

.changeset/fuzzy-suits-sit.md
```diff
@@ -0,0 +1,16 @@
+---
+'@backstage/plugin-techdocs': minor
+---
+
+Add 2 config elements to extension "page:techdocs/reader" to configure default layout `withSearch` and `withHeader`. Default are unchanged to `true`.
```
Rugvip:
```suggestion
Add two config values to the `page:techdocs/reader` extension that configure default layout, `withSearch` and `withHeader`. Default are unchanged to `true`.
```

.changeset/fuzzy-suits-sit.md
```diff
@@ -0,0 +1,16 @@
...
+
+E.g. to disable the search and header on the Techdocs Reader Page:
+
+```yaml
+app:
+  extensions:
+    - page:techdocs/reader:
+        config:
+          withSearch: false
+          withHeader: false
```
Rugvip: Generally I think it's good to keep boolean flags defaulted to `false`. In this case that means switching this around to be for example `disableSearch: true` instead. Or `withoutSearch`, `removeSearch`, `noSearch`, etc 😅

plugins/techdocs/src/alpha/index.tsx
```diff
@@ -156,8 +156,20 @@ const techDocsReaderPage = PageBlueprint.makeWithOverrides({
...
     addons: createExtensionInput([AddonBlueprint.dataRefs.addon]),
   },
-  factory(originalFactory, { apis, inputs }) {
+  config: {
+    schema: {
+      withSearch: z => z.boolean().default(true),
+      withHeader: z => z.boolean().default(true),
+    },
+  },
+  factory(originalFactory, { inputs, config }) {
```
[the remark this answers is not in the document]
Rugvip: Yep, this ☝️

GabDug:
Properly merged changes from #3252, flipped the config order and quickly manually re-tested.

Issue #30607, opened 2025-07-21
Migrate to Zod 4


Teo-ShaoWei:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
zod zod-to-json-schema zod-validation-error
```

### 🗃️ Project Area

Unknown

### 🛠️ Task

We are to upgrade Zod 3 (i.e. `zod@^3.22.4`) to Zod 4. As a result, we should upgrade `zod-validation-error@^3.4.0` to at least `zod-validation-error@^3.5.0` for [Zod 4 support](https://github.com/causaly/zod-validation-error/releases/tag/v3.5.0).

We will want to replace all usage of `zod-to-json-schema` with [native Zod 4 JSON schema support](https://zod.dev/json-schema).

Need be, we can ignore the following package versions that are in use:
- `zod@^3.23.8`, which is a nested dependency of the MCP/OpenAI external dependencies only.
- `zod-validation-error@^3.0.3` as that is a nested dependency of `knip` only.

The following NPM packages are affected:
- `@backstage/backend-defaults`
- `@backstage/backend-plugin-api`
- `@backstage/backend-test-utils`
- `@backstage/cli`
- `@backstage/cli-node`
- `@backstage/core-app-api`
- `@backstage/core-compat-api`
- `@backstage/core-components`
- `@backstage/frontend-app-api`
- `@internal/frontend`
- `@backstage/frontend-plugin-api`
- `@backstage/frontend-test-utils`
- `@backstage/repo-tools`
- `@internal/scaffolder`
- `@backstage/plugin-auth-backend-module-atlassian-provider`
- `@backstage/plugin-auth-backend-module-aws-alb-provider`
- `@backstage/plugin-auth-backend-module-azure-easyauth-provider`
- `@backstage/plugin-auth-backend-module-bitbucket-provider`
- `@backstage/plugin-auth-backend-module-bitbucket-server-provider`
- `@backstage/plugin-auth-backend-module-cloudflare-access-provider`
- `@backstage/plugin-auth-backend-module-gcp-iap-provider`
- `@backstage/plugin-auth-backend-module-github-provider`
- `@backstage/plugin-auth-backend-module-gitlab-provider`
- `@backstage/plugin-auth-backend-module-google-provider`
- `@backstage/plugin-auth-backend-module-microsoft-provider`
[...]

PR #32296, opened 2026-01-12
chore(deps): Update all packages to Zod v3 || v4

GabDug:

Part of #30607. Follow-up of #32316 & #32306 (upgrade of `zod-to-json-schema` and `zod-validation-error`.

This PR aims to unlock v4 usage (opt-in), by allowing both v3/4 installs which are backward and (partially) forward compatible.

The scope of this PR is not to use Zod 4 just yet, just to prepare the dependencies and imports to allow a smooth, gradual migration. 

By doing a one shot upgrade, we should prevent having both v3 and v4 installed for consumer. 

### Changes

- Updated all imports of `zod` to `zod/v3`
- Changed all `zod` dependencies to accept `^3.25.76 || ^4.0.0`
  - Not a breaking change, since regardless of installed version, we still load from `/v3` 

### Reference 

See [zod docs for more details about v4](https://zod.dev/v4), [migration guide](https://zod.dev/v4/changelog) and [migrations tips for libraries,](https://zod.dev/library-authors).
Also see issue #30607 and PRs #32316 & #32306

<details><summary>Next steps? Out of this PR scope</summary>
<p>

- Migrate plugin per plugin to `/v4`
- Special care will be needed for the `scaffolder-actions`
  - See @Teo-ShaoWei [remarks](https://github.com/backstage/backstage/issues/30607#issuecomment-3113101297)
  - As we forward our version of zod, I don't think a transparent migration will be possible. An API change is probably needed.
  - Other alternatives could be looked into, like zod-compatible [Standard Schema](https://zod.dev/library-authors?id=do-i-need-to-depend-on-zod).
  - I will investigate a little bit more and check w/ the Scaffolder SIG.
 
</p>
</details> 

- [ ] ~~Added or updated documentation~~
- [ ] ~~Tests for new functionality and regression tests for bug fixes~~
- [ ] ~~Screenshots attached (for UI changes)~~

### DIFF
--- a/packages/core-app-api/src/lib/AuthSessionManager/AuthSessionStore.ts
+++ b/packages/core-app-api/src/lib/AuthSessionManager/AuthSessionStore.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { ZodSchema } from 'zod/v3';
+import type { ZodSchema } from 'zod/v3';
 import {
   MutableSessionManager,
   SessionScopesFunc,
--- a/packages/core-components/src/layout/ProxiedSignInPage/types.test.ts
+++ b/packages/core-components/src/layout/ProxiedSignInPage/types.test.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { TypeOf } from 'zod/v3';
+import type { TypeOf } from 'zod/v3';
 import { ProxiedSession, proxiedSessionSchema } from './types';
 
 describe('types', () => {
--- a/plugins/auth-node/report.api.md
+++ b/plugins/auth-node/report.api.md
@@ -16,8 +16,8 @@ import { Profile } from 'passport';
 import { Request as Request_2 } from 'express';
 import { Response as Response_2 } from 'express';
 import { Strategy } from 'passport';
-import { ZodSchema } from 'zod/v3';
-import { ZodTypeDef } from 'zod/v3';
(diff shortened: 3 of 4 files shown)


jonkoops:
@GabDug In regard to providing the version ranges that provide both v3 and v4, why is this required? Would it not be possible to start requiring v4 alone? Or would this cause some kind of unexpected side effects on consumers?

GabDug:
@jonkoops It's not required, i's more for convenience to avoid having both installed if not needed.

Forcing `v4` may have some side-effects on consumers, as some packages still rely on v3 on the main export (not `zod/v3`), which can cause some nasty type issues.

Imo, we should bump the dependency range to 4 only when we actively require it.

freben:
Oh and double check that you have DCO signed all commits. And it seems to need a rebase :)

freben:
@GabDug This went a bit too far doesn't it - it seems to have added changeset entries for, for example, the confluence module, but I don't think there are any changes for that one. See the "Unexpected Changesets" comment at the top

GabDug:
@freben Sorry about that, should not have committed in a rush!

freben:
Merged via https://github.com/backstage/backstage/pull/33399 - I made sure to attribute you in all of the commits so you get the credit for it! But needed a bit of a rebase in the middle of some wild merging of things before the release. Thanks again!

PR #32306, opened 2026-01-13
Upgrade `zod-validation-error` to version 4

jonkoops:
Upgrades `zod-validation-error` to version 4, and updates all imports to remain compatible with version 3 of Zod. See #30607 for more details.

### DIFF
--- a/packages/cli/src/modules/new/lib/preparation/loadPortableTemplate.ts
+++ b/packages/cli/src/modules/new/lib/preparation/loadPortableTemplate.ts
@@ -28,7 +28,7 @@ import {
 } from '../types';
 import { PortableTemplate } from '../types';
 import { ForwardedError } from '@backstage/errors';
-import { fromZodError } from 'zod-validation-error';
+import { fromZodError } from 'zod-validation-error/v3';
 
 const templateDefinitionSchema = z
   .object({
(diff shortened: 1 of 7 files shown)


packages/cli/src/modules/new/lib/preparation/loadPortableTemplate.ts
```diff
@@ -28,7 +28,7 @@ import {
 } from '../types';
 import { PortableTemplate } from '../types';
 import { ForwardedError } from '@backstage/errors';
-import { fromZodError } from 'zod-validation-error';
+import { fromZodError } from 'zod-validation-error/v3';
```
freben, line 31: Interesting. Will this be a challenge in the future, that we need to manually detect whether it's a zod 3 or 4 object being used and calling the right version of `zod-validation-error` on it?

jonkoops: Yes, we'll likely end up with some temporary glue code to support Zod 3 and 4 at the same time as a migration path where Zod schemas are user provided. See the ongoing discussion under #30607 and the WIP implementation under #32296 by @GabDug.

freben (approved):

PR #32308, opened 2026-01-13
refactor(ui): implement useDefinition hook for component styling

Naycon:

Introduces a new `useDefinition` hook and `defineComponent` factory for
consistent component prop handling and styling in the UI package.

### Changes

- Add `defineComponent` factory for type-safe component definitions
- Add `useDefinition` hook that handles:
  - Responsive prop resolution
  - Data attribute generation
  - CSS class composition
  - Surface context (leaf/container)
- Add `LeafSurfaceProps` and `ContainerSurfaceProps` shared types
- Add `SurfacePropsConstraint` for compile-time validation of surface props
- Migrate Box, Button, ButtonIcon, and ButtonLink to use the new system
- Each button variant (Button, ButtonIcon, ButtonLink) is now fully
  independent with its own CSS, types, and definition


.changeset/polite-symbols-act.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Added `useDefinition` hook and `defineComponent` factory for building custom components with consistent styling and responsive prop support.
```
Rugvip: Always best to keep these from the point of view of a consumer of the package. These are internal things still right? Reading this I would expect a `useDefinition` hook to be exported.

Naycon: Ah, very valid point. Will rewrite! Or does it make sense to skip all together?

Naycon: Updated the changeset, we had some css targeting changes that needed to be added anyway.

.changeset/polite-symbols-act.md
```diff
@@ -0,0 +1,15 @@
+---
+'@backstage/ui': patch
```
Rugvip: minor then ye? 😁

Naycon: 🤦

PR #32313, opened 2026-01-14
BUI - Add new Popover component

cdedreuille:

CleanShot 2026-01-14 at 11 11 52@2x


packages/ui/css/styles.css
```diff
@@ -100,6 +100,7 @@
     --bui-ring: #1f5493;
     --bui-scrollbar: #a0a0a03b;
     --bui-scrollbar-thumb: #a0a0a0;
+    --bui-shadow-overlay: 0 10px 15px -3px #0000001a, 0 4px 6px -4px #0000001a;
```
Rugvip: Hmm, is shadow overlay really what this is? 🤔 Would've expected that to be on top of what this is applied to in a way, or a backdrop. Is `--bui-shadow` too generic? 😅 Some other alternatives:

- `--bui-box-shadow`
- `--bui-shadow-elevation`

cdedreuille: You're right. `--bui-shadow` is better. We can think about a scale of shadows later on but for now this should work.

PR #32324, opened 2026-01-14
frontend-plugin-api: deprecate AppRootWrapperBlueprint, replace with PluginWrapperBlueprint

Rugvip:

Draft for now while awaiting feedback, please read on 😁 

Looking at the usage of `AppRootWrapperBlueprint` in the wild, I think we made a mistake adding it in the first place. The intention was to make it easy to to migrate some of the global app wrappers like `CustomWrapper` in this example:

```tsx
export default app.createRoot(
  <>
    <AlertDisplay />
    <OAuthRequestDialog />
    <SignalsDisplay />
    <AppRouter>
      <CustomWrapper>
        <VisitListener />
        <Root>{routes}</Root>
      </CustomWrapper>
    </AppRouter>
  </>,
);
```

The problem is plugins end up using it to wrap the app in a context for their own usage. That in itself is something we'd like to enable, but not in that way. It's of course generally good if plugins stick to using Utility APIs, but there are situations where wrapping with a React component is a lot simpler and familiar, especially in the case of using things like query client providers and the like.

The problem with wrapping the entire app is that it breaks plugin isolation. It makes it much simpler for a single plugin to break the entire app, it pretty much guarantees that the plugin provides a significant amount of content that is eagerly loaded at startup rather than behind lazy loading, and it also risks polluting contexts for other plugins. For example in the case of a query client provider, if that's from a common library it may end up affecting other plugins that use the same library.

First off I suggest we deprecate and fairly quickly remove `AppRootWrapperBlueprint`, and later on also the associated inputs on the `app/root` extension. I would like to close the gap of having something that enables wrapping plugins with a context though, so that's where the `PluginWrapperBlueprint` comes in.

### The new `PluginWrapperBlueprint`

Usage looks like this:

```tsx
PluginWrapperBlueprint.make({
  params: defineParams =>
    defineParams({
      loader: () => import('./components/PluginWrapper'),
    }),
});

// in ./components/PluginWrapper
export function component(props: { children: ReactNode }) {
  return <SomeCustomProvider>{children}</SomeCustomProvider>
}
```

This will wrap every element of the plugin in the `PluginWrapper` component, which in practice is done by the `ExtensionBoundary` that is meant to wrap all plugin elements.

The tricky bit here is that each individual boundary will do it's own wrapping, meaning that you'll end up with duplicate elements of the provided wrapper `component`. This means you can't maintain a single state across your plugin, which sometimes may be desirable.

This is where this all gets a bit complex and I'm wondering if it's worth the tradeoff. In order to have a shared value across all instances of the provider, you can supply a `useWrapperValue` hook to the blueprint params. In practice it lets you do something like this:

```tsx
PluginWrapperBlueprint.make(/* same as above */);

// in ./components/PluginWrapper
const CounterContext = createContext<{
  count: number;
  increment: () => void;
} | null>(null);

export const component = CounterContext.Provider;

export function useWrapperValue() {
  const [value, setValue] = useState(0);
  return {
    count: value,
    increment: () => setValue(val => val + 1),
  };
}
```

You don't have to use `createContext` ofc, it's just simple with the existing API. The output of `useWrapperValue` will be forwarded as a `value` prop to the wrapper component, which matches the context provider API.

So looking for feedback on this general idea, and potentially naming suggestions to make things more clear. I'm thinking broadly that we could do one those these things:

1. Ship what's currently in this PR, with whatever minor changes are needed
2. Deprecate `AppRootWrapperBlueprint` but don't provide any replacement for the plugin context use-case
3. Deprecate old wrapper and ship the new `PluginWrapperBlueprint`, but only support the `component` option and not `useWrapperValue`. Rely on external stores if values need to be shared across provider elements.
4. Something completely different I haven't thought of

### DIFF
--- a/packages/frontend-plugin-api/report.api.md
+++ b/packages/frontend-plugin-api/report.api.md
@@ -271,7 +271,7 @@ export const AppRootElementBlueprint: ExtensionBlueprint_2<{
   dataRefs: never;
 }>;
 
-// @public
+// @public @deprecated
 export const AppRootWrapperBlueprint: ExtensionBlueprint_2<{
   kind: 'app-root-wrapper';
   params: {
@@ -1867,6 +1867,55 @@ export interface PluginOptions<
   // (undocumented)
   routes?: TRoutes;
 }
+
+// @public
+export type PluginWrapperApi = {
+  getRootWrapper(): ComponentType<{
+    children: ReactNode;
+  }>;
+  getPluginWrapper(pluginId: string):
+    | ComponentType<{
+        children: ReactNode;
+      }>
+    | undefined;
+};
+
+// @public
+export const pluginWrapperApiRef: ApiRef<PluginWrapperApi>;
+
+// @public
+export const PluginWrapperBlueprint: ExtensionBlueprint_2<{
+  kind: 'plugin-wrapper';
+  params: <TValue = never>(params: {
+    loader: () => Promise<PluginWrapperDefinition<TValue>>;
+  }) => ExtensionBlueprintParams_2<{
+    loader: () => Promise<PluginWrapperDefinition>;
+  }>;
+  output: ExtensionDataRef_2<
+    () => Promise<PluginWrapperDefinition>,
+    'core.plugin-wrapper.loader',
+    {}
+  >;
+  inputs: {};
+  config: {};
+  configInput: {};
+  dataRefs: {
+    wrapper: ConfigurableExtensionDataRef_2<
+      () => Promise<PluginWrapperDefinition>,
+      'core.plugin-wrapper.loader',
+      {}
+    >;
+  };
+}>;
+
+// @public
+export type PluginWrapperDefinition<TValue = unknown | never> = {
+  useWrapperValue?: () => TValue;
+  component: ComponentType<{
+    children: ReactNode;
+    value: TValue;
+  }>;
+};
 
 // @public (undocumented)
 export type PortableSchema<TOutput, TInput = TOutput> = {
--- a/packages/frontend-plugin-api/src/apis/definitions/PluginWrapperApi.ts
+++ b/packages/frontend-plugin-api/src/apis/definitions/PluginWrapperApi.ts
@@ -16,10 +16,9 @@
 
 import { ComponentType, ReactNode } from 'react';
 import { ApiRef, createApiRef } from '../system';
-import type { ExtensionBoundary } from '../../components';
 
 /**
- * The Plugin Wrapper API allows plugins to wrap their extensions with providers. This API is only intended for internal use by the Backstage frontend system. To provide contexts to plugin components, use {@link ExtensionBoundary} instead.
+ * The Plugin Wrapper API allows plugins to wrap their extensions with providers. This API is only intended for internal use by the Backstage frontend system. To provide contexts to plugin components, use `ExtensionBoundary` instead.
  *
  * @public
  */
@@ -30,7 +29,7 @@ export type PluginWrapperApi = {
   getRootWrapper(): ComponentType<{ children: ReactNode }>;
 
   /**
-   * Returns a wrapper component for a specific plugin, or undefined if no wrappers exist. Do not use this API directly, instead use {@link ExtensionBoundary} to wrap your plugin components if needed.
+   * Returns a wrapper component for a specific plugin, or undefined if no wrappers exist. Do not use this API directly, instead use `ExtensionBoundary` to wrap your plugin components if needed.
    */
   getPluginWrapper(
     pluginId: string,
--- a/packages/frontend-plugin-api/src/blueprints/PluginWrapperBlueprint.tsx
+++ b/packages/frontend-plugin-api/src/blueprints/PluginWrapperBlueprint.tsx
@@ -39,7 +39,7 @@ export type PluginWrapperDefinition<TValue = unknown | never> = {
    * @returns
    */
   useWrapperValue?: () => TValue;
-  component: ComponentType<{ children: ReactNode; value: NoInfer<TValue> }>;
+  component: ComponentType<{ children: ReactNode; value: TValue }>;
 };
 
 const wrapperDataRef = createExtensionDataRef<


benjdlambert:
Is it worth us encouraging people to use this to the best of our ability, but otherwise providing a very arbitrary context for plugins and extensions to use that are maybe rendered outside of these plugin wrappers?

Thinking that a sidebar extension item or something might want to do something like:

```ts
const { setValue, useValue }  = useGlobalState(); // this would automatically return the plugin scope context
return (
  <div onClick={() => setValue('open', true); />
)
```

Pretty bad example, but I wonder if this really simple context might help for most use cases that are outside of the plugin boundary wrapping that's a singleton and not duplicated?

Maybe it's not worth the hassle or the thought, just a quick drive by comment.

Rugvip:
@benjdlambert how would we scope the hook to the plugin if it's outside the boundary? 🤔 feels like a bit of 🐥/🥚 

Global plugin store is interesting, but we don't really need to provide anything for that I feel there are already plenty of options out there. This is mostly aiming towards situations where plugins for one reason or another want to use React contexts.

I don't think there's anything that shouldn't be rendered in an extension boundary btw, especially now that we can forward errors via alerts, including sidebar items if those could be provided as elements/hooks.

freben:
I'm definitely in favour of having a `PluginWrapperBlueprint`. But also a bit torn on entirely removing (rather than for example strongly discouraging) `AppRootWrapperBlueprint`.

I wonder if it'll be an issue to some, that the plugin wrapper in effect only is usable if your context component has a perfectly injectable state. As in, that it has the ability to `<MyContext state={state}>` and that it experiences _absolutely no side effects_ from being remounted over and over in different parts of the app tree - including having several such contexts mounted at the same time next to each other (or even nested inside each other?) using the same state instance. The contexts must then effectively not have any internal state whatsoever (using react state hooks), but has to have everything inside that injected state. Isn't that likely to commonly be problematic? Maybe I worry more than necessary 🤔

Of course we can tell people to override the entire app root if their hands are bound and they really REALLY need to put something there - so maybe that's still a good-enough fallback?

Parsifal-M:
If we strongly discourage using `AppRootWrapperBlueprint` in which scenarios is this okay to do? 

Will people always just take "the path of least resistance"?

jabrks:
What if `ExtensionBoundary` itself used context to only render the wrapper in the very first instance, so any nesting just results in a no-op? Would that remove the need for the separate `useWrapperValue` API?

e.g. 
```typescript
const WrapperContext = createContext(false);

export function ExtensionBoundary(props: ExtensionBoundaryProps) {
  const isNested = useContext(WrapperContext);

  // ...

  if (PluginWrapper && !isNested) {
    content = (
      <WrapperContext.Provider value={true}>
        <PluginWrapper>{content}</PluginWrapper>
      </WrapperContext.Provider>
    );
  }

  // return ...
}
```

freben:
@jabrks That could even be a feature of the blueprint. A param of `nesting: 'allow' | 'ignore' | 'error'` or something like that. That still leaves the issue of multiple sibling contexts (eg if there are two cards side by side).

Rugvip:
Yep the nesting isn't as much of an issue as the disconnect between siblings, for example two cards on the entity page from the same plugin wanting to share some form of query client. 

Another potential option here to add to the list is to keep the existing app root wrappers, but make them lazy such that they don't load and mount until some element from a plugin is rendered. It's a bit of a half solution to the original problem though. 

Honestly at this point I'm leaning towards shipping the plugin wrapper but without the shared state hook, leaving state sharing to be done via singletons defined globally or in the blueprint factory.

Rugvip:
> But also a bit torn on entirely removing (rather than for example strongly discouraging) `AppRootWrapperBlueprint`

What I'm aiming for is really for the ability to have peace of mind as the operator of a Backstage app, knowing that plugins will never fully break the entire app. That's been hugely important for us internally at least. And yep, as long as it exists it will be used 😅

freben:
> Another potential option here to add to the list is to keep the existing app root wrappers, but make them lazy such that they don't load and mount until some element from a plugin is rendered.

This would likely lead to the entire page getting rerendered right?

freben:
Is there any way that we could render each `AppRootWrapper` with a surrounding error component that only "shallowly" captures problems with it and renders the children without the wrapper if that happens? Deferring issues down to the actual plugins who try to consume it and fail to do so.

I guess even if that worked, it'd only capture a subset of the possible issues 🤔 But something that technically achieves like a try/catch.

Of course that also doesn't address any performance concerns etc of using app wrappers in the first place.

Rugvip:
Realized I left out some important context that'd be part of migration notes for this change. This is how you'd be able to still be able to install custom providers in the app, using a module to override the layout:

```tsx
const appRootWrapperModule = createFrontendModule({
  pluginId: 'app',
  extensions: [
    appPlugin.getExtension('app/layout').override({
      *factory(originalFactory) {
        const parent = originalFactory();
        yield* parent;
        const element = parent.get(coreExtensionData.reactElement);
        yield coreExtensionData.reactElement(
          <CustomProvider>{element}</CustomProvider>,
        );
      },
    }),
  ],
});
```

It does require you to get a bit in the weeds. I think that's alright for new apps as this isn't something you'd really want to be doing if it can be avoided, which it often can. I do think it's a bit problematic for apps that are migrating to NFS though, since the old system lacked protection against this kind of pattern.

That did make me think though, maybe we don't need to completely remove `AppRootWrapperBlueprint`, but instead limits its use to only the `app` plugin? In practice it means that you can easily install a wrapper via an `app` module, but not in other plugins. I'll open up a competing PR that implements that along with the simplified `PluginWrapperBlueprint`, so we can compare.

Rugvip:
> This would likely lead to the entire page getting rerendered right?

It's possible it could be avoided, but not sure. Not too keen on that path either way though 😅 

> I guess even if that worked, it'd only capture a subset of the possible issues 🤔 But something that technically achieves like a try/catch.

Yep, you could never guard against the wrapper failing to render its children afaik, which is why I didn't explore that path.

Rugvip:
Alright, implemented the proposal in the end of https://github.com/backstage/backstage/pull/32324#issuecomment-3753853349 in https://github.com/backstage/backstage/pull/32332. That's my preferred way forward at this point.

PR #32326, opened 2026-01-14
feat(scaffolder-backend): simplify abort handling in debug:wait action

mtt-artis:

Cleanup the abort handling logic by using the { once: true } listener option
instead of manual removeEventListener, and remove the unused AbortController.

### DIFF
--- a/plugins/scaffolder-backend/src/scaffolder/actions/builtin/debug/wait.ts
+++ b/plugins/scaffolder-backend/src/scaffolder/actions/builtin/debug/wait.ts
@@ -84,13 +84,14 @@ export function createWaitAction(options?: {
         );
       }
 
-      await new Promise(resolve => {
+      await new Promise<void>(resolve => {
         const timeoutHandle = setTimeout(abort, delayTime.toMillis());
-        ctx.signal?.addEventListener('abort', abort, { once: true });
+        ctx.signal?.addEventListener('abort', abort);
 
         function abort() {
+          ctx.signal?.removeEventListener('abort', abort);
           clearTimeout(timeoutHandle);
-          resolve('finished');
+          resolve();
         }
       });
     },
--- a/.changeset/cuddly-colts-dress.md
+++ b/.changeset/cuddly-colts-dress.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-scaffolder-backend': patch
 ---
 
-Simplify abort handling in debug:wait action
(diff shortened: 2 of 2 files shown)


plugins/scaffolder-backend/src/scaffolder/actions/builtin/debug/wait.ts
```diff
@@ -85,14 +85,11 @@ export function createWaitAction(options?: {
...
         const timeoutHandle = setTimeout(abort, delayTime.toMillis());
-        ctx.signal?.addEventListener('abort', abort);
+        ctx.signal?.addEventListener('abort', abort, { once: true });
 
         function abort() {
-          ctx.signal?.removeEventListener('abort', abort);
-          clearTimeout(timeoutHandle!);
-          controller.abort();
+          clearTimeout(timeoutHandle);
           resolve('finished');
```
freben: Unrelated, but you can change to `new Promise<void>` to be able to just do `resolve()` here without a dummy value.

plugins/scaffolder-backend/src/scaffolder/actions/builtin/debug/wait.ts
```diff
@@ -85,14 +85,11 @@ export function createWaitAction(options?: {
       }
 
       await new Promise(resolve => {
-        const controller = new AbortController();
         const timeoutHandle = setTimeout(abort, delayTime.toMillis());
-        ctx.signal?.addEventListener('abort', abort);
+        ctx.signal?.addEventListener('abort', abort, { once: true });
```
freben: Note that this still "leaks". The `once` setting only makes it so that the listener is removed IF it triggers. The only way that it triggers is if the user selects to cancel the task run or if the service is shutting down or whatnot. So normally, this will never get unregistered. I recommend reinstating the `removeEventListener` below

plugins/scaffolder-backend/src/scaffolder/actions/builtin/debug/wait.ts
```diff
@@ -85,14 +85,11 @@ export function createWaitAction(options?: {
       }
 
       await new Promise(resolve => {
-        const controller = new AbortController();
```
freben, line 88: Removing this is 👍 , since it was doing absolutely nothing

mtt-artis:
You’re right. Thanks for the feedback.

Issue #31220, opened 2025-09-22
Enable/disable NFS extensions dynamically based on permissions etc


freben:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
nfs enable disable permissions feature flags
```

### 🗃️ Project Area

Core Framework

### 🔖 Need

Users can already enable and disable frontend extensions in a fixed way, either by giving their ID a `false` value in config or by making an extension override in code that has the `disabled` flag set to `true`. But there are also cases where you need this enablement to be more dynamic, and for example base it on feature flags, permissions, or other circumstances.

This issue discusses potential new ways in the framework for expressing such dynamic enablement.

Some initial questions:

* Should this be possible to do only in code, or should it be possible to express (in a somewhat rich way) in config too?
* Should it be somewhat hard coded to e.g. only permissions and possibly feature flags, or should it be a more extensible general concept?
* Are there any overlaps with this solution, and more "local" filtering as used e.g. by entity cards/content pages? They are of course somewhat different by design (disable app node entirely, vs "be active in this context or not"), but should a solution try to consider this whole picture somehow e.g. in terms of language or rules structures used?
* Do users need to have full combinatorial freedom, using composable logic like "show this if at least one of these two AND this third one are true"?
* Will it be okay to make these decisions _once_ at app startup, or do they need to be re-evaluated at every potential usage of the extension?

### 📝 Proposal

Instead of trying to target the factory and its outputs (see Alternatives below), it may often be more appropriate to be able to disable the entire app node as if it never were there. Essentially, that outcome is the same as if you would have explicitly disabled it in your `app.extensions` config.

This proposal will initially focus on _code based_ enablement, that is, not on app-config based control of this concern.

We could do so by expanding on the more binary nature of the existing `disabled` concept. The following example passes in a callback that controls enablement, but just returns the default.

```ts
userSettingsPlugin.withOverrides({
  extensions: [
    userSettingsPlugin.getExtension('page:user-settings').override({
      // Here it latches onto the existing "disabled" field, but since its naming
      // is a bit unfortunate in this context it may be better to add a
      // separate field named e.g. "if" or "when" or "enable" or so
      disabled: async (originalDecision, { apiHolder }) => {
        // This implementation just returns whatever the default decision was,
        // e.g. based on what app-config etc had decided on. You can add
[...]

PR #32330, opened 2026-01-15
feat: support conditional rendering NFS

aramissennyeydd:

Based on conversation in https://github.com/backstage/backstage/issues/31220.

This allows users to define custom conditions for whether or not to render an element,

```
PageBlueprint.make({
    params: {
      enabled: async (originalDecision, { apiHolder }) => {
		    // execute logic here.
			return originalDecision && myCondition;
      },
    },
  });
```

I also added helpers for common tasks like the below,

```
PageBlueprint.make({
    params: {
      enabled: allOf(
        createPermissionCondition(catalogReadPermission),
        createFeatureFlagCondition('advanced-ui')
      ),
    },
  });
```


packages/frontend-plugin-api/src/blueprints/ConditionalRender.tsx
```diff
@@ -0,0 +1,76 @@
...
+    error,
+    value: conditionMet,
+  } = useAsync(async () => {
+    const originalDecision = async () => true; // Always true for now, reserved for future config-based conditions
+    return condition(originalDecision, { apiHolder });
+  }, [condition, apiHolder]);
+
+  if (error) {
+    // eslint-disable-next-line no-console
+    console.error('Error evaluating extension condition:', error);
```
Parsifal-M: 🧹

packages/frontend-app-api/src/wiring/EnabledConditionsGate.tsx
```diff
@@ -0,0 +1,75 @@
...
+      .then(newRootEl => {
+        if (!cancelled) {
+          setRootElement(newRootEl);
+        }
+      })
+      .catch(err => {
+        console.error(
+          '[EnabledConditionsGate] Failed to complete app initialization:',
+          err,
+        );
```
aramissennyeydd, line 69: How should this error be captured? not sure if error collector is the right context here

packages/frontend-plugin-api/src/blueprints/types.ts
```diff
@@ -0,0 +1,67 @@
...
+ *   // Then add additional check
+ *   const featureFlags = apiHolder.get(featureFlagsApiRef);
+ *   return featureFlags?.isActive('beta-feature') === true;
+ * }
+ * ```
+ */
+export type ExtensionConditionFunc = (
+  originalDecision: () => Promise<boolean>,
+  context: { apiHolder: ApiHolder },
+) => Promise<boolean>;
```
Rugvip, line 67: Initial reaction here for me to think about later: It's probably safest to make this an opaque type at some level in order to maximize our ability to evolve this.

packages/frontend-plugin-api/src/wiring/createExtension.ts
```diff
@@ -176,6 +195,7 @@ export type CreateExtensionOptions<
   attachTo: ExtensionDefinitionAttachTo<UParentInputs> &
     VerifyExtensionAttachTo<UOutput, UParentInputs>;
   disabled?: boolean;
+  enabled?: import('../blueprints/types').ExtensionConditionFunc;
```
Rugvip, line 198: Thought this was an API report for a second 😁 

Generally try to stick to using `import` statements at the top rather than expressions. I don't particularly have a strong reason to prefer it but it's done like that basically everywhere in the codebase so best stick to it. If needed use `import type ... from ...`

.changeset/busy-symbols-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-permission-react': minor
+---
+
+Added a new `createPermissionCondition` helper for conditional rendering with the new NFS `enabled` method.
```
Rugvip, line 5: I wonder if we could find some good work to use for these to avoid excessive namespacing or lack thereof. A bit like the new predicates. Some potential ones:

- Gate
- Switch
- Control
- Activator
- When
- Guard
- Constraint
- Toggle
- Fence

.changeset/nine-gifts-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/frontend-plugin-api': minor
+---
+
+Added support for conditionally rendering blueprints. You can now provide an `enabled` function with various criteria. Utility methods `allOf`, `anyOf` and `createFeatureFlagCondition` are provided for common use cases.
```
Rugvip, line 5: Extensions really, not only blueprints, although they are included ofc

packages/app-next/src/examples/pagesPlugin.tsx
Rugvip, line 1: 👍

packages/app-next/src/examples/pagesPlugin.tsx
```diff
@@ -0,0 +1,462 @@
...
+import {
+  createFrontendPlugin,
+  createRouteRef,
+  createExternalRouteRef,
+  useRouteRef,
+  PageBlueprint,
+  FrontendPluginInfo,
+  useAppNode,
+  allOf,
+  anyOf,
```
Rugvip, line 27: Continuing the naming discussion from the changeset, I think we need a good way to namespace there. If we where to for example pick "Guard", there could for example be `Guards.allOf`. Although to some extent I think it might be cleanest if these weren't needed as separate exports at all, thinking either of these:

structural:

```ts
enabled: {
  $allOf: [
    createFeatureFlagCondition('advanced-features'),
    async () => {
      const config = ctx.apis.get(configApiRef);
      return config?.getOptionalBoolean('app.features.advancedMode') === true;
    },
  ]
}
```

provided:

```ts
enabled: ctx => ctx.allOf([
  createFeatureFlagCondition('advanced-features'),
  async () => {
    const config = ctx.apis.get(configApiRef);
    return config?.getOptionalBoolean('app.features.advancedMode') === true;
  },
]),
```

packages/app-next/src/examples/pagesPlugin.tsx
```diff
@@ -0,0 +1,462 @@
...
+              true in config.
+            </p>
+            {indexLink && <Link to={indexLink()}>Go back</Link>}
+          </div>
+        );
+      };
+      return <Component />;
+    },
+  },
+  enabled: async (_, { apiHolder }) => {
```
Rugvip, line 260: Minor nit, but this is `apis` in NFS factories, so should stick to that here

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
```diff
@@ -390,7 +401,81 @@ export function createSpecializedApp(options?: CreateSpecializedAppOptions): {
...
+          current = current.edges.attachedTo?.node;
+        }
+      }
+
+      for (const node of nodesToClear) {
+        (node as any).instance = undefined;
+      }
+
+      // Re-instantiate with newly enabled extensions
+      instantiateAppNodeTree(
```
Rugvip, line 454: Hmm, immediate reaction here is that we should aim to only instantiate and even resolve the tree once. I guess I'm not really seeing the benefit of making this a two-step process rather than initializing everything upfront. Don't know exactly what the impact would be of making `createSpecializedApp` async, but probably worth considering. Afaik the main impact would be for tests, but they'd need to start supporting this anyway too.

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
```diff
@@ -390,7 +401,81 @@ export function createSpecializedApp(options?: CreateSpecializedAppOptions): {
   routeResolutionApi.initialize(routeInfo, routeRefsById.routes);
   appTreeApi.initialize(routeInfo);
 
-  return { apis, tree, errors: collector.collectErrors() };
+  return {
+    apis,
+    tree,
+    // Proxy the errors so that any errors collected after completeInitialization are also included.
```
Rugvip, line 407: Related to the discussion below, seems more straightforward to make sure we do all work upfront to make sure we don't mutate the response. If we do want the two-step thing, I'm thinking `completeInitialization` should instead also return updated errors.

packages/frontend-app-api/src/wiring/EnabledConditionsGate.tsx
```diff
@@ -0,0 +1,78 @@
...
+      );
+      return initialEl;
+    },
+  );
+
+  useEffect(() => {
+    let cancelled = false;
+
+    // Evaluate enabled conditions and re-instantiate in background
+    completeInitialization()
```
Rugvip, line 58: I think this might as well happen before we enter React? We already have a couple of async actions in `createApp` that are wrapped in an async app loader that I assume should be able to house this too.

packages/frontend-plugin-api/src/blueprints/helpers.ts
```diff
@@ -0,0 +1,114 @@
...
+ * ```
+ */
+export function createFeatureFlagCondition(
+  flagName: string,
+): ExtensionConditionFunc {
+  return async (_, { apiHolder }) => {
+    const featureFlags = apiHolder.get(featureFlagsApiRef);
+    return !!featureFlags?.isActive(flagName);
+  };
+}
```
Rugvip, line 114: To some extent I feel we might want to structure this to just be very simple to do without additional wrapping? For example:

```ts
PageBlueprint.make({
  enabled: ({ apis }) => apis.get(featureFlagsApiRef)?.isActive('the-flag')
})
```

And maybe that's actually enough? If the structures around this is very simple we don't even really need utilities like `anyOf` or `allOf` at all really, because one would just express that in the callback.

On the other hand there's potentially a lot of value in making these things declarative and inspectable, but then I think it's worth taking that all the way. With that I mean some form of app-level condition evaluation logic that you install that knows how to evaluate pure data declarations like

```ts
{
  $$type: '@backstage/ExtensionCondition',
  evaluator: 'feature-flag',
  params: { name: 'the-flag' }
}
```

that gets constructed with a helper like this. I'm not sure I see too much value in landing between those extremes?

packages/frontend-plugin-api/report.api.md
```diff
@@ -930,6 +947,8 @@ export interface Extension<TConfig, TConfigInput = TConfig> {
   // (undocumented)
   readonly disabled: boolean;
   // (undocumented)
+  readonly enabled?: ExtensionConditionFunc;
```
Rugvip, line 950: Doesn't really need to be present on the public `Extension` type right? I think we could leave this as an internal field only

packages/frontend-plugin-api/src/blueprints/types.ts
```diff
@@ -0,0 +1,67 @@
...
+ *   if (!(await originalDecision())) return false;
+ *
+ *   // Then add additional check
+ *   const featureFlags = apiHolder.get(featureFlagsApiRef);
+ *   return featureFlags?.isActive('beta-feature') === true;
+ * }
+ * ```
+ */
+export type ExtensionConditionFunc = (
+  originalDecision: () => Promise<boolean>,
```
Rugvip, line 65: Hmm, just realized I didn't see this used anywhere. I'm kinda thinking that a disabled extension is always disabled, and can't be enabled with a flag.

packages/frontend-plugin-api/src/blueprints/types.ts
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @public
+ * @param originalDecision - A function that checks the `disabled` field. Returns false if disabled: true, otherwise true.
+ * @param context - Context object containing the apiHolder for accessing Backstage APIs
+ * @returns A promise that resolves to true if the extension should be enabled, false otherwise
+ *
+ * @remarks
+ * Since this function is evaluated during extension tree construction, only APIs that are available
+ * early in the app lifecycle can be safely used. Feature flags and config are typically available,
+ * but permission APIs may not be fully initialized yet.
```
Rugvip, line 31: I think we want to have a rule similar to how this works for APIs, i.e. when initializing the React half of the extension tree everything is available, but when initializing APIs either none are just a few fixed ones are enabled.

Rugvip (changes requested):
Nice! 🎉 

The examples in the PR body immediately had me confused 😅

Should be this I assume?

```tsx
PageBlueprint.make({
    enabled: async (originalDecision, { apiHolder }) => {
	    // execute logic here.
		return originalDecision && myCondition;
    },
    params: { ... },
  });
```

packages/frontend-plugin-api/src/blueprints/helpers.ts
Rugvip: Entity predicate version, as discussed:

```ts
PageBlueprint.make({
  enabled: {
    featureFlags: {
      $contains: 'the-flag'
    }
  }
})
```

Permissions:

```ts
PageBlueprint.make({
  enabled: {
    permissions: {
      $contains: 'catalog.entity-read'
    }
  },
})
```

PR #32332, opened 2026-01-15
frontend-plugin-api: rework AppRootWrapperBlueprint

Rugvip:

Competing PR for #32324. Lots of context there so will keep it lightweight here. This still adds the `PluginWrapperBlueprint` and `PluginWrapperApi`, but in the simplified form where there is no built-in central state sharing solution across wrapper elements. I was at first thinking that I also wanted to move the `PluginWrapperApi` to somehow be internal in the `app` plugin, but realized it's actually a pretty neat API to have access to as it enables global wrapping of plugins in general, bit like the extension factory middleware. Of course a more advanced use-case, but though it was worth exposing still.

Other than that this implements what I proposed in https://github.com/backstage/backstage/pull/32324#issuecomment-3753853349. It moves the `AppRootWrapperBlueprint` to instead be exposed by a new `@backstage/plugin-app-react` package and is only possible to use from the `app` plugin, i.e. modules. This restricts it from use in plugins, while still makes it easier to add wrapper in the app compared to the snippet in https://github.com/backstage/backstage/pull/32324#issuecomment-3753853349.

The old `AppRootWrapperBlueprint` from `@backstage/frontend-plugin-api` stays around for now, but is marked as deprecated and triggers a deprecation warning.

### DIFF
--- a/packages/frontend-plugin-api/src/alpha.ts
+++ b/packages/frontend-plugin-api/src/alpha.ts
@@ -14,12 +14,6 @@
  * limitations under the License.
  */
 
-/**
- * Alpha exports for plugin wrapper functionality.
- *
- * @alpha
- */
-
 export { PluginWrapperBlueprint } from './blueprints/PluginWrapperBlueprint';
 export {
   type PluginWrapperApi,
--- a/plugins/app/src/apis/PluginWrapperApi/DefaultPluginWrapperApi.tsx
+++ b/plugins/app/src/apis/PluginWrapperApi/DefaultPluginWrapperApi.tsx
@@ -15,7 +15,7 @@
  */
 
 import { PluginWrapperApi } from '@backstage/frontend-plugin-api/alpha';
-import { ComponentType, ReactNode, useEffect, useState } from 'react';
+import { ComponentType, ReactNode, useEffect, useMemo, useState } from 'react';
 
 type WrapperInput = {
   loader: () => Promise<{ component: ComponentType<{ children: ReactNode }> }>;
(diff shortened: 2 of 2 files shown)


packages/frontend-plugin-api/src/alpha.ts
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Alpha exports for plugin wrapper functionality.
+ *
+ * @alpha
```
freben: Oh does this work? As opposed to `@packageDocumentation`

packages/frontend-plugin-api/report-alpha.api.md
```diff
@@ -0,0 +1,68 @@
...
+          children: ReactNode;
+        }>;
+      }>,
+      'core.plugin-wrapper.loader',
+      {}
+    >;
+  };
+}>;
+
+// (No @packageDocumentation comment for this package)
```
freben, line 67: No doesn't seem like it

plugins/app/src/apis/PluginWrapperApi/DefaultPluginWrapperApi.tsx
```diff
@@ -0,0 +1,107 @@
...
+        useEffect(() => {
+          Promise.all(loaders.map(loader => loader()))
+            .then(results => {
+              setLoadedWrappers(results.map(r => r.component));
+            })
+            .catch(setError);
+        }, []);
+
+        if (error) {
+          throw error;
```
freben, line 86: Directly in the render path?

plugins/app/src/apis/PluginWrapperApi/DefaultPluginWrapperApi.tsx
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        if (!loadedWrappers) {
+          return null;
+        }
+
+        let content = props.children;
+
+        for (const Wrapper of loadedWrappers) {
+          content = <Wrapper>{content}</Wrapper>;
```
freben: We could have constructed this composed component once, in the `useEffect`

freben (approved):

plugins/app/src/apis/PluginWrapperApi/DefaultPluginWrapperApi.tsx
Rugvip: Not quite, because it depends on the children. I threw in a `useMemo` though to make it a bit more stable.

plugins/app/src/apis/PluginWrapperApi/DefaultPluginWrapperApi.tsx
Rugvip: Yep! thrown straight up to the boundary above it

PR #7315, opened 2021-09-24
catalog-backend: refactor deferral of entites + fix dissapearing ref bug

Rugvip:
The catalog currently has a bug where adding new locations to the catalog will cause the root references to all the old ones to disappear. The impact of this has been mitigated by the fact that we don't properly surface errors and orphan status from locations yet (#6973).

This fixes #6826, and also a more recent issue where refreshes would fail for most entities in the catalog.

PR #32333, opened 2026-01-15
fix(catalog): avoid deleting unrelated refresh state references

atharrva01:
## Summary

This PR fixes a bug in the catalog processing logic where refresh state references for **unrelated parents or providers** could be deleted during normal entity refresh cycles.
The issue caused **silent catalog corruption** and could eventually lead to **valid entities being incorrectly orphaned and deleted** in multi-provider or multi-parent setups.

---

### What was broken

* When an entity provider or parent reprocessed shared child entities, the catalog removed **all incoming references to those children**, not just the references belonging to the current source.
* As a result, entities with multiple parents or providers gradually lost valid ownership references.

### Why this is critical

* `refresh_state_references` is the **source of truth** for ownership.
* Orphan detection relies on this table to decide when an entity should be deleted.
* Losing references silently caused:

  * Legitimate entities to be marked as orphans
  * Entities disappearing from the catalog without logs or errors
* The longer a Backstage instance ran, the more corrupted the catalog state became.

This primarily affects **production deployments** with:

* Multiple entity providers
* Overlapping catalogs
* Location → Component hierarchies

---

## Root Cause

The cleanup logic used an `OR`-scoped DELETE when updating refresh state references.

```ts
await tx('refresh_state_references')
  .where({ source_entity_ref: options.sourceEntityRef })
  .orWhereIn('target_entity_ref', stateReferences)
  .delete();
```

This translated to SQL that deleted:

* All references **from the current source** (intended)
* **All references to the target entities from any source** (unintended)

This behavior implicitly assumed that each entity had a single parent/provider, which is **not true** and is explicitly supported elsewhere in the catalog.

---

## Steps to Reproduce (Realistic)

1. Configure two providers that discover the same entity (e.g. `component:default/shared-service`)
2. Both providers create entries in `refresh_state_references`
3. One provider refreshes and re-emits the entity
4. The refresh deletes **all existing references to that entity**
5. Only the current provider’s reference is reinserted
6. If the remaining provider later stops emitting the entity, orphan cleanup deletes it

The only visible symptom is **missing entities**.

---

## Fix

The fix scopes deletion strictly to references **originating from the current source entity** by removing the `orWhereIn` clause entirely.

```ts
await tx('refresh_state_references')
  .where({ source_entity_ref: options.sourceEntityRef })
  .delete();
```

This ensures:

* Only stale references from the current source are removed
* References from other parents/providers are preserved
* Correct behavior for multi-parent and multi-provider catalogs

Duplicate prevention is already handled by existing constraints and reinsertion logic, so no additional cleanup is required.

---

## Tests

* Existing tests relied on the unintended deletion behavior and were updated to reflect the correct multi-source semantics
* All relevant catalog backend database tests pass

---

## Result After Fix

* Entities with multiple parents retain all valid references
* Orphan detection works as designed
* Silent catalog corruption is eliminated
* No behavior change for single-parent entities

---

### DIFF
--- a/plugins/catalog-backend/src/tests/integration.test.ts
+++ b/plugins/catalog-backend/src/tests/integration.test.ts
@@ -1096,16 +1096,22 @@ describe('Catalog Backend Integration', () => {
     await expect(harness.process()).resolves.toEqual({});
 
     // Expect to find A and B' in the catalog
-    await expect(harness.getRefreshStateReferences()).resolves.toEqual([
-      {
-        sourceKey: 'test',
-        targetEntityRef: 'component:default/a',
-      },
-      {
-        sourceEntityRef: 'component:default/a',
-        targetEntityRef: 'component:default/b',
-      },
-    ]);
+
+    await expect(harness.getRefreshStateReferences()).resolves.toEqual(
+      // The refresh state cleanup no longer removes unrelated references
+      // belonging to other sources or parents.
+      expect.arrayContaining([
+        {
+          sourceKey: 'test',
+          targetEntityRef: 'component:default/a',
+        },
+        {
+          sourceEntityRef: 'component:default/a',
+          targetEntityRef: 'component:default/b',
+        },
+      ]),
+    );
+
     await expect(harness.getRefreshState()).resolves.toEqual({
       'component:default/a': expect.objectContaining({
         locationKey: null,


atharrva01:
hi, @awanlin @vinzscam 
This PR fixes an issue where refresh state cleanup could remove references belonging to other parents or providers, leading to silent catalog corruption and premature orphaning. The fix scopes deletion to the originating source and updates tests to reflect correct multi-source behavior.

atharrva01:
Thanks for the clarification! I’ll go through the contributing guide and follow the review process going forward.

plugins/catalog-backend/src/database/DefaultProcessingDatabase.ts
```diff
@@ -385,12 +385,9 @@ export class DefaultProcessingDatabase implements ProcessingDatabase {
       }
     }
 
-    // Lastly, replace refresh state references for the originating entity and any successfully added entities
+    // Replace refresh state references for the originating entity
     await tx<DbRefreshStateReferencesRow>('refresh_state_references')
-      // Remove all existing references from the originating entity
       .where({ source_entity_ref: options.sourceEntityRef })
-      // And remove any existing references to entities that we're inserting new references for
-      .orWhereIn('target_entity_ref', stateReferences)
```
freben, line 393: This takes things back to the problematic state that https://github.com/backstage/backstage/pull/7315 supposedly fixed - could you try to make sense of what the actual perfect fix would be here?

PR #32339, opened 2026-01-15
Bui docs migration

cdedreuille:

This PR fixes the BUI docs deployment. The problem appeared after migration our stories from CSF 3 to Next. What I'm proposing here is to decouple the docs examples from actual stories as a way to simplify the structure and colocate code examples with the actual page.

This PR also introduce new pages for `Popover`, `ToggleButton`, `ToggleButtonGroup` and fixes quite a few examples.

Code examples on the `Menu` page are still broken but this will be changed in another PR.


docs-ui/src/app/components/accordion/props-definition.ts
Naycon, line 1: A thought for another PR - it'd be awesome to automatically generate these from the actual components. I think this is one source of the documentation being out of sync with reality - it's easy to miss updating this correctly.

Naycon (approved):
Nice, definitely an improvement!

docs-ui/src/app/components/accordion/props-definition.ts
cdedreuille: I tried that but sometimes we do not want all of the code to be present. For example for `ButtonLink`. This one need to be wrapped into a `MemoryRouter` since it is using `react-router-dom`. We do not want this to show. So I went back and kept it manual. They are colocated so it should be fine 🙏

Rugvip (approved):
bit of a scary setup, but alright

docs-ui/src/app/components/accordion/props-definition.ts
Naycon: But this is the props, there's no `MemoryRouter` involved in those. You're thinking about the story snippets I assume? That's not what I meant here. :)

Naycon:
> bit of a scary setup, but alright

@Rugvip anything in particular that you're concerned about? Or just the general thing that there's no guarantee the docs actually match the code?

Rugvip:
@Naycon sorry chatted with Charles separately, no remaining concerns! But yep that was basically it

PR #32344, opened 2026-01-15
fix(core-app-api): fix selector memory leak

Sanchit2662:

This PR fixes a frontend memory leak caused by duplicate `AppThemeSelector` instances and missing cleanup logic for storage event listeners and observable subscriptions.

Previously, the app created multiple selector instances that registered permanent `window` event listeners and RxJS subscriptions without any teardown mechanism. Over time—especially during development with HMR or repeated app mounts—this resulted in accumulating listeners, duplicated work, and increasing memory usage.

This change ensures a single shared selector instance is used and introduces an explicit cleanup (`dispose`) mechanism for both theme and language selectors.

---

### 1. Deduplicate selector instances (`AppManager.tsx`)

A single shared selector instance is now created and reused across the app:

```ts
private appThemeApi = AppThemeSelector.createWithStorage(this.themes);
```

Both the Provider and API factory now reference this shared instance instead of creating new ones.

---

### 2. Add explicit cleanup to selectors

Both selectors now track their side effects and expose a `dispose()` method.

Example from `AppThemeSelector`:

```ts
dispose() {
  this.#storageSubscription?.unsubscribe();
  window.removeEventListener('storage', this.#storageListener);
}
```

The same cleanup pattern is applied to `AppLanguageSelector`.

---

## Impact

* Prevents unbounded memory growth in development and long-running apps
* Improves performance by eliminating duplicate event handling
* Makes selector lifecycle explicit and safer for tests and HMR

---

* [x] A changeset describing the change and affected packages
* [x] All commits have a `Signed-off-by` line in the message

### DIFF
--- a/.changeset/backend-process-listener-cleanup.md
+++ b/.changeset/backend-process-listener-cleanup.md
@@ -1,0 +1,5 @@
+---
+'@backstage/backend-app-api': patch
+---
+
+Fixed memory leak by properly cleaning up process event listeners on backend shutdown.
--- a/packages/core-app-api/report.api.md
+++ b/packages/core-app-api/report.api.md
@@ -276,6 +276,7 @@ export class AppThemeSelector implements AppThemeApi {
   activeThemeId$(): Observable<string | undefined>;
   // (undocumented)
   static createWithStorage(themes: AppTheme[]): AppThemeSelector;
+  dispose(): void;
   // (undocumented)
   getActiveThemeId(): string | undefined;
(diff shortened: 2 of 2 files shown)


Sanchit2662:
Hi @vinzscam , 
this PR fixes a memory leak caused by duplicate theme/language selector instances and missing cleanup of storage listeners.
Whenever you have time, I’d appreciate a review — thanks!

PR #32347, opened 2026-01-15
Fixes Menu trigger

cdedreuille:

Bypassing the default popover setup created unexpected behaviours and potentially had an impact on screen readers. To avoid that, this reverts back to the default React Aria behaviour where `isNonModal` is undefined.

We previously tried to set the menu as non-modal to avoid layout shifts. React Aria recommends to keep setting it up as modal as this will guarantee full accessibility.

### DIFF
--- a/.changeset/fine-lines-warn.md
+++ b/.changeset/fine-lines-warn.md
@@ -1,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Fixed Menu component trigger button not toggling correctly. Removed custom click-outside handler that was interfering with React Aria's built-in state management, allowing the menu to properly open and close when clicking the trigger button.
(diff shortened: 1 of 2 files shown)


.changeset/fine-lines-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Fixed Menu component trigger button not toggling correctly. Removed custom click-outside handler that was interfering with React Aria's built-in state management, allowing the menu to properly open and close when clicking the trigger button.
```
Naycon, line 5: Maybe add this so the changelog script can pick this up?

```suggestion
Fixed Menu component trigger button not toggling correctly. Removed custom click-outside handler that was interfering with React Aria's built-in state management, allowing the menu to properly open and close when clicking the trigger button.

Affected components: Menu
```

Naycon (approved):

PR #32352, opened 2026-01-16
feat: improve signals event subscriptions

drodil:

This lowers the number of unnecessary message sending between nodes if there are no subscriptions to specific channel in the specific node. Previously all nodes subscribed to global `signals` topic and received messages for all channels even they did not have any client interested in those.

Ultimately it would be nice to have `unsubscribe` possibility in the `EventsService` to limit also unnecessary subscriptions in that end.

### DIFF
--- a/.changeset/loud-rules-greet.md
+++ b/.changeset/loud-rules-greet.md
@@ -1,0 +1,10 @@
+---
+'@backstage/plugin-signals-backend': patch
+'@backstage/plugin-signals-node': patch
+---
+
+Improve signals event subscriptions by utilizing the channel.
+
+This lowers the number of unnecessary message sending between nodes if there are no subscriptions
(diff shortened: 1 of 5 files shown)


drodil:
Might be not needed optimization as the signals are not very used and usually not running in very many nodes at the same time. But maybe something to consider anyways if this ever is a bottleneck. Additionally, it raises the question of support for `EventsService` unsubscription.

Rugvip (changes requested):
Hmm, after pondering this one I think it's best to keep as is tbh. I'm worried this reduces reliability of delivery while not really gaining that much in terms of reduced load. Introducing the extra topics I think could easily outweigh the benefits of avoiding event forwarding. Would somewhat expect that for most topics they're either going to be very low traffic, and need to be broadcast to all instances anyway.

PR #32358, opened 2026-01-16
[mcp-actions]: add server operation and session duration metrics with alpha MetricsService

kurtaking:

Introduces a new alpha `MetricsService` to the Backstage backend, establishing a standardized way for plugins to emit and collect metrics using the OpenTelemetry API. 

The MCP Actions backend plugin now leverages it to track two new metrics:

- mcp.server.operation.duration
- mcp.server.session.duration

These are based off the recommendations outlined in https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#metrics.

Example results in local prometheus

image

image


plugins/mcp-actions-backend/src/services/McpService.test.ts
```diff
@@ -23,6 +23,30 @@ import {
   CallToolResultSchema,
   ListToolsResultSchema,
 } from '@modelcontextprotocol/sdk/types.js';
+import { MetricsService } from '@backstage/backend-plugin-api/alpha';
+
+// Simple mock MetricsService for testing
+function createMockMetricsService(): MetricsService {
```
benjdlambert: Think we should add this to `mockServices` instead like other built in things. Check how the `ActionsRegistry` does it

plugins/mcp-actions-backend/src/metrics.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+type McpActionsLookupAttributes = Pick<McpActionsBaseAttributes, 'client'> & {
+  result: 'found' | 'not_found';
+};
+
+export function createMcpMetrics(metrics: MetricsService) {
+  const actionsExecutionDuration =
+    metrics.createHistogram<McpActionsExecutionAttributes>(
+      'backstage.mcp-actions.actions.execution.duration',
```
benjdlambert: Feeling like `backstage.mcp-actions` should be prepended by the wrapper. 

That most metrics should already have this done for them in the plugin scoped service instead?

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -20,43 +20,56 @@ import {
...
 
     server.setRequestHandler(ListToolsRequestSchema, async () => {
-      // TODO: switch this to be configuration based later
       const { actions } = await this.actions.list({ credentials });
-
       return {
         tools: actions.map(action => ({
           inputSchema: action.schema.input,
-          // todo(blam): this is unfortunately not supported by most clients yet.
-          // When this is provided you need to provide structuredContent instead.
```
benjdlambert, line 58: But why?

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -20,43 +20,56 @@ import {
...
+    credentials,
+    client,
+  }: {
+    credentials: BackstageCredentials;
+    client: 'sse' | 'streamable';
+  }) {
     const server = new McpServer(
       {
         name: 'backstage',
-        // TODO: this version will most likely change in the future.
```
benjdlambert, line 44: But why?

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -72,32 +85,72 @@ export class McpService {
...
+
+        try {
+          const { output } = await this.actions.invoke({
+            id: action.id,
+            input: params.arguments as JsonObject,
+            credentials,
+          });
+
+          const response = {
+            content: [
```
benjdlambert, line 134: I think we can actually solve the comment about only certain types not supporting `structuredContent` by providing both.

That's been a cleanup on my mind for a while, but let's either do that in this PR or put the comments back?

plugins/mcp-actions-backend/src/metrics.ts
kurtaking: 100%, that is the long term vision, but I was trying to keep introducing the metrics service as simple as possible. 

Including namespacing means we cannot migrate catalog and scaffolder metrics without causing breaking changes. I think it would be cleaner if we follow up with that feature as part of deprecating existing metrics. 

So, the plan would be 

1. These new mcp metrics get the alpha metrics service merged while following the naming pattern we know we want to follow moving forward.
2. Two PRs to migrate catalog and scaffolder metrics 1:1. 
...
n. Add feature enhancement, remove `backstage.{pluginId}` and do once single deprecation of old naming pattern

plugins/mcp-actions-backend/src/services/McpService.ts
kurtaking: I put the comment back. I'd rather that be something separate.

plugins/mcp-actions-backend/src/services/McpService.ts
kurtaking: Restored the comments.

plugins/mcp-actions-backend/src/services/McpService.ts
kurtaking: Restored the comments.

plugins/mcp-actions-backend/src/metrics.ts
kurtaking: @benjdlambert - forcing `backstage.plugin.mcp-actions.*` when OTel semantic conventions specifically define `mcp.*` as the standard namespace defeats the whole purpose of following industry standards. This requires revising the BEP which I am looking into now.

plugins/mcp-actions-backend/src/services/McpService.test.ts
kurtaking: I've refactored to match how you are doing it for other alpha services.

plugins/mcp-actions-backend/src/metrics.ts
kurtaking: The service now handles getting a meter based on the pluginId, providing discoverability at the plugin level. That way, the metric name can follow the standards outlined by OTel.

plugins/mcp-actions-backend/src/metrics.ts
kurtaking, line 1: Shared attributes and the sorts could eventually be lifted into a `@backstage/otel` lib or something similar.

packages/backend-plugin-api/package.json
```diff
@@ -68,6 +68,9 @@
     "luxon": "^3.0.0",
     "zod": "^3.25.76"
   },
+  "peerDependencies": {
+    "@opentelemetry/api": "^1.9.0"
+  },
```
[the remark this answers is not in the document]
kurtaking: @Rugvip - do you have a recommendation on how we should handle this?

plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
```diff
@@ -65,6 +96,13 @@ export const createStreamableRouter = ({
...
         });
       }
+
+      const durationSeconds = (performance.now() - sessionStart) / 1000;
+
+      sessionDuration.record(durationSeconds, {
+        ...baseAttributes,
+        'error.type': sessionErrorType,
+      });
     }
```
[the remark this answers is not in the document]
kurtaking: This is wrong.

packages/backend-plugin-api/package.json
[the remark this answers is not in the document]
Sarabadu: I think it should be ok to get the api as non optional dependency... the opentelemetry api get a default noop in case there is no opentelemetry configuration 🤔

benjdlambert:
@kurtaking same for this one here, I didn't wanna just be rude and trample all the work you've done, so we've pulled it into a seperate branch and rebased it ontop of the latest stuff :pray: thanks!

https://github.com/backstage/backstage/pull/32978

kurtaking:
Thank you for the help getting this finished up @benjdlambert. 🙏🏼 🙏🏼 🙏🏼

PR #32360, opened 2026-01-17
Fix silent orphan cleanup failure caused by unhandled Promise rejection

WHOIM1205:
### Summary
This PR fixes a lifecycle bug in the catalog processing engine where orphan cleanup could silently fail during backend startup.

The orphan cleanup task was being scheduled asynchronously without awaiting the result. If task scheduling failed (e.g. due to database unavailability), the Promise rejection was unhandled, backend startup still reported success, and orphan cleanup was permanently disabled with no indication.

This change makes orphan cleanup scheduling part of the startup contract so failures are surfaced immediately and startup fails fast.

---

### Problem
`startOrphanCleanup()` was synchronous but invoked the async `scheduler.scheduleTask()` without awaiting it.

If `scheduleTask()` failed due to:
- database connection failure
- task persistence failure
- invalid task ID validation

the error became an unhandled Promise rejection that:
- was not propagated to `engine.start()`
- did not fail backend startup
- left orphan cleanup permanently disabled
- produced no clear error or signal to operators

This resulted in orphaned catalog entities accumulating indefinitely.

---

### Impact
- Orphaned entities are never cleaned up
- `final_entities` table grows without bound
- Catalog queries return stale / phantom entities
- Search indexes become polluted
- Data integrity degrades over time with no alerting

This affects all production Backstage instances using  
`catalog.orphanStrategy: delete`.

---

### Fix
- Converted `startOrphanCleanup()` to an async method
- Awaited `scheduler.scheduleTask()` during initialization
- Updated `start()` to await orphan cleanup startup

This ensures that:
- Startup fails fast if orphan cleanup cannot be scheduled
- Errors are properly propagated and visible
- Unhandled Promise rejections are eliminated
- Orphan cleanup cannot silently fail

---

### Code Changes
- `startOrphanCleanup()` is now `async`
- `scheduler.scheduleTask()` is awaited
- `engine.start()` awaits orphan cleanup initialization

---

### Steps to Reproduce (Before Fix)
1. Deploy Backstage with PostgreSQL
2. Configure an entity provider (e.g. GitHub)
3. Make the database temporarily unavailable during backend startup
4. Backend reports successful startup
5. `catalog_orphan_cleanup` task is never created
6. Remove entities from the source
7. Wait for multiple refresh cycles
8. Orphaned entities still appear in the catalog

---

### Behavior After Fix
- Backend startup fails immediately if orphan cleanup cannot be scheduled
- Clear error is surfaced to the caller
- Orphan cleanup is guaranteed to be active when startup succeeds
- Catalog data integrity is preserved over time

---

### Why This Matters
Orphan cleanup is a correctness-critical subsystem when `orphanStrategy: delete` is enabled.  
Failing fast is preferable to silently corrupting catalog data and accumulating stale entities in production.

---

### Checklist
- [x] Fix eliminates unhandled Promise rejection
- [x] Startup lifecycle correctly propagates failures
- [x] No behavior change when scheduler succeeds
- [x] Prevents silent catalog corruption

### DIFF
--- a/plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
+++ b/plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
@@ -118,7 +118,7 @@ export class DefaultCatalogProcessingEngine {
     }
 
     const stopPipeline = this.startPipeline();
-    const stopCleanup = this.startOrphanCleanup();
+    const stopCleanup = await this.startOrphanCleanup();
 
     this.stopFunc = () => {
       stopPipeline();
@@ -347,7 +347,7 @@ export class DefaultCatalogProcessingEngine {
     });
   }
 
-  private startOrphanCleanup(): () => void {
+  private async startOrphanCleanup(): Promise<() => void> {
     const orphanStrategy =
       this.config.getOptionalString('catalog.orphanStrategy') ?? 'delete';
     if (orphanStrategy !== 'delete') {
@@ -371,7 +371,7 @@ export class DefaultCatalogProcessingEngine {
     };
 
     const abortController = new AbortController();
-    this.scheduler.scheduleTask({
+    await this.scheduler.scheduleTask({
       id: 'catalog_orphan_cleanup',
       frequency: { milliseconds: this.orphanCleanupIntervalMs },
       timeout: { milliseconds: this.orphanCleanupIntervalMs * 0.8 },


WHOIM1205:
Hi @Naycon @benjdlambert @schultzp2020 👋  

This PR fixes a silent startup failure where orphan cleanup could be permanently disabled due to an unhandled Promise rejection during task scheduling.

I’d really appreciate a review when you have time — especially around the startup lifecycle change and fail-fast behavior.  
Happy to adjust or add tests if needed.

Thanks!

awanlin (changes requested):
Hi @WHOIM1205, there's no need to ping people, they will get notified when they are assigned to the PR upon its creation. I'd highly recommend reading over the Contributing guide, it explains the review process, changesets which are missing, and our AI Policy: https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md

benjdlambert:
@whoim09 this needs a changeset, see the above message: https://github.com/backstage/backstage/pull/32360#issuecomment-3764269757

plugins/catalog-backend/src/processing/DefaultCatalogProcessingEngine.ts
```diff
@@ -118,7 +118,7 @@ export class DefaultCatalogProcessingEngine {
     }
 
     const stopPipeline = this.startPipeline();
-    const stopCleanup = this.startOrphanCleanup();
+    const stopCleanup = await this.startOrphanCleanup();
```
schultzp2020, line 121: If `startOrphanCleanup()` rejects, the already-started pipeline leaks because `stopFunc` is never set. Consider wrapping this in a try/catch so the pipeline is torn down before the error propagates:

```ts
const stopPipeline = this.startPipeline();
let stopCleanup: () => void;
try {
  stopCleanup = await this.startOrphanCleanup();
} catch (error) {
  stopPipeline();
  throw error;
}

this.stopFunc = () => {
  stopPipeline();
  stopCleanup();
};
```

schultzp2020 (commented):
Thanks for the contribution, @WHOIM1205!

### Pipeline cleanup on failure

Now that `startOrphanCleanup()` can throw, there's a subtle resource leak: `startPipeline()` fires up the processing pipeline on line 120, but if the subsequent `await this.startOrphanCleanup()` rejects, `this.stopFunc` is never assigned and the running pipeline has no teardown path. Please wrap the orphan cleanup in a try/catch that calls `stopPipeline()` before re-throwing — see the inline comment.

### Missing changeset

This still needs a changeset as @benjdlambert [mentioned](https://github.com/backstage/backstage/pull/32360#issuecomment-3841635119).

Everything else looks good — the core idea of awaiting `scheduleTask` is the right call.

Issue #31066, opened 2025-09-04
White Noise Warnings Microsoft Catalog Provider - No Photo


matthew-da-clark:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
"Photo" "Warning"
```

### 🗃️ Project Area

Catalog

### 🛠️ Task

White noise warnings for loading users since no picture - ideally if was debug or had option of severity for this in config

Should be debug warning - would put PR but technically not allowed (working on it)

[Log Warning Line](https://github.com/backstage/backstage/blob/42643faff1aea0c579526888f1d622029787e9e5/plugins/catalog-backend-module-msgraph/src/microsoftGraph/read.ts#L507)

### 📊 Priority & Impact

Low Priority

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

No, but I'm happy to collaborate on a PR with someone else

freben:
Yeah that's harsh, agreed.

awanlin:
This is an awesome easy first issue for some one!

b718:
I am happy to work on this, but I want to make sure I am understanding the problem first!

The problem at hand is that the logger should not be logging at the level of `warn` but instead `debug` and the reason we want to down-level it is because this log message can be noisy and it's not something that we need to alert the user on.

If that understanding is correct, happy to put out the PR for it once I can get my local-dev working! Thanks in advance!

Parsifal-M:
> The problem at hand is that the logger should not be logging at the level of `warn` but instead `debug` and the reason we want to down-level it is because this log message can be noisy and it's not something that we need to alert the user on.
> 

Yep that's the idea! 👍

PR #32361, opened 2026-01-18
fix(logger-level): changed logger level of user photo loading to debug

b718:

As described by the original issue, the problem at hand is that the logger should not be logging at the level of `warn` but instead `debug`. The reason for this is because this log message can be noisy and it's not necessary to log at a visible level for something like this.

Closes: https://github.com/backstage/backstage/issues/31066


awanlin (approved):
Many thanks @b718 🚀

PR #32362, opened 2026-01-18
 Fix: Stop execution after OAuth token refresh rejection

WHOIM1205:
### Summary
This PR fixes a control-flow bug in the OAuth token refresh logic where execution continued after calling `reject()`, allowing the code path to reach `resolve()` with potentially invalid data.

While JavaScript Promises only settle once, continuing execution after a rejection is incorrect and can lead to misleading logs, unclear error attribution, and unsafe assumptions in a security-critical authentication flow.

This change ensures execution halts immediately after an error during token refresh.

---

### 📍 Affected Code
- **File:** `plugins/auth-node/src/passport/PassportHelpers.ts`
- **Function:** `executeRefreshTokenStrategy`

---

### ❌ Problem
In the OAuth refresh callback, `reject()` was called without returning:

```ts
if (err) {
  reject(new ForwardedError(`Failed to refresh access token`, err));
}

if (!accessToken) {
  reject(new Error(`Failed to refresh access token, no access token received`));
}

resolve({ accessToken, refreshToken: newRefreshToken, params });

### DIFF
--- a/plugins/auth-node/src/passport/PassportHelpers.ts
+++ b/plugins/auth-node/src/passport/PassportHelpers.ts
@@ -199,6 +199,7 @@ export class PassportHelpers {
         ) => {
           if (err) {
             reject(new ForwardedError(`Failed to refresh access token`, err));
+            return;
           }
           if (!accessToken) {
             reject(
@@ -206,6 +207,7 @@ export class PassportHelpers {
                 `Failed to refresh access token, no access token received`,
               ),
             );
+            return;
           }
 
           resolve({


WHOIM1205:
Hi 

This is a small, focused fix (2-line change) in the OAuth token refresh flow to stop execution after a rejection. It only affects error paths and preserves all existing successful behavior.

Thanks in advance!

awanlin (changes requested):
Hi @WHOIM1205, there's no need to ping people, they will get notified when they are assigned to the PR upon its creation. I'd highly recommend reading over the Contributing guide, it explains the review process, changesets which are missing, and our AI Policy: https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md

freben (changes requested):
Saw this fall through the cracks.

While it's not actually fixing anything, it's still good hygiene for the future.

It needs a changeset though to be shipped.

PR #32372, opened 2026-01-19
feat(search-elasticsearch): add authentication extension point

nojaf:

This PR/RFC proposes adding an authentication extension point to `@backstage/plugin-search-backend-module-elasticsearch` that enables dynamic authentication mechanisms such as bearer tokens with automatic rotation.

### The Problem

The current Elasticsearch module supports several authentication methods:

| Provider     | Supported Auth Methods        |
| ------------ | ----------------------------- |
| `elastic`    | Username/Password             |
| `aws`        | AWS SigV4 (automatic via SDK) |
| `opensearch` | Username/Password             |
| Default      | API Key or Username/Password  |

However, **bearer token authentication is not supported**, which is a common requirement for enterprise environments that:

- Use OAuth2/OIDC identity providers for service authentication
- Require automatic token rotation (tokens expire, typically hourly)
- Need to integrate with internal identity services
- Run OpenSearch/Elasticsearch clusters secured by token-based authentication

### Real-World Use Case

Many enterprises operate internal OpenSearch clusters that are secured using bearer tokens issued by corporate identity providers. These environments typically require:

1. **Dynamic token acquisition** - Tokens must be fetched from an identity service at runtime
2. **Automatic token rotation** - Tokens expire (often hourly) and must be refreshed
3. **Multiple authentication fallbacks** - Different auth methods for different environments:
   - Kubernetes service accounts for cluster deployments
   - Developer tooling integration for local development
   - OAuth device flow as a fallback

### Current Workarounds

Without native support, adopters are forced to:

1. **Monkey-patch the client** - Intercept transport requests to inject tokens (fragile, breaks on upgrades)
2. **Fork the module** - Maintain a separate fork with custom auth logic (maintenance burden)
3. **Use static tokens** - Manually rotate tokens in configuration (operationally risky)

These workarounds are fragile, prevent easy upgrades to newer Backstage versions, and create maintenance burden.

### New Extension Point: `elasticsearchAuthExtensionPoint`

Add a new extension point that allows modules to provide custom authentication:

```typescript
/**
 * A provider that supplies authentication headers for Elasticsearch/OpenSearch requests.
 */
export interface ElasticSearchAuthProvider {
  /**
   * Returns authentication headers to be included in requests.
   * Called before each request, enabling dynamic token refresh and rotation.
   */
  getAuthHeaders(): Promise<Record<string, string>>;
}

/**
 * Extension point for providing custom authentication to the Elasticsearch search engine.
 */
export interface ElasticSearchAuthExtensionPoint {
  setAuthProvider(provider: ElasticSearchAuthProvider): void;
}

export const elasticsearchAuthExtensionPoint =
  createExtensionPoint<ElasticSearchAuthExtensionPoint>({
    id: 'search.elasticsearchEngine.auth',
  });
```

### Implementation Approach

1. **Custom Transport Classes** - Create transport wrappers that call `getAuthHeaders()` before each request
2. **Module Registration** - Register the extension point alongside the existing translator extension point
3. **Backward Compatible** - When no auth provider is set, existing config-based auth works unchanged

### Usage Example

Adopters create a small backend module to wire up their auth provider:

```typescript
import { createBackendModule } from '@backstage/backend-plugin-api';
import { elasticsearchAuthExtensionPoint } from '@backstage/plugin-search-backend-module-elasticsearch';

export default createBackendModule({
  pluginId: 'search',
  moduleId: 'elasticsearch-custom-auth',
  register(env) {
    env.registerInit({
      deps: {
        elasticsearchAuth: elasticsearchAuthExtensionPoint,
      },
      async init({ elasticsearchAuth }) {
        elasticsearchAuth.setAuthProvider({
          async getAuthHeaders() {
            const token = await myTokenService.getToken();
            return { Authorization: `Bearer ${token}` };
          },
        });
      },
    });
  },
});
```

### Why an Extension Point vs. Config?

**Config-based approaches don't work for this use case:**

1. **Config is resolved at startup** - Token rotation requires per-request header injection
2. **No script execution in config** - Backstage config doesn't support `$exec` or similar
3. **Tokens are ephemeral** - Can't be stored in static YAML files

**Extension points are the right fit because:**

1. They follow established Backstage patterns (see `elasticsearchTranslatorExtensionPoint`)
2. They enable runtime behavior customization
3. They keep custom auth logic in adopter codebases, not in the core module

### Why Not Just Add Static Bearer Token Config?

A static `bearerToken` config option would only solve the simplest case:

```yaml
# This doesn't help with token rotation
search:
  elasticsearch:
    auth:
      bearerToken: ${STATIC_TOKEN} # Expires in 1 hour, then what?
```

The extension point approach solves both simple and complex cases.

### Why Per-Request vs. Connection-Level?

The auth provider's `getAuthHeaders()` is called per-request because:

1. **Token rotation** - Tokens may expire between requests
2. **Consistency with AWS pattern** - The existing `AwsSigv4Signer` also signs per-request
3. **Flexibility** - Adopters can implement caching in their provider if appropriate

### Backward Compatibility

- **Fully backward compatible** - Existing configurations work unchanged
- **Additive change** - New extension point, no changes to existing APIs
- **Optional feature** - Auth provider is only used when explicitly set

### Provider Compatibility

| Provider     | Custom Auth Provider Support |
| ------------ | ---------------------------- |
| `elastic`    | Yes                          |
| `opensearch` | Yes                          |
| Default      | Yes                          |
| `aws`        | No (uses SigV4 signing)      |

The AWS provider throws a clear error if a custom auth provider is configured, since AWS authentication uses a fundamentally different mechanism (request signing).

### DIFF
--- a/docs/features/search/search-engines.md
+++ b/docs/features/search/search-engines.md
@@ -291,3 +291,65 @@ search:
       fuzziness: AUTO
       prefixLength: 3;
 ```
+
+### Custom Authentication Extension Point
+
+For enterprise environments that require dynamic authentication mechanisms such as bearer tokens with automatic rotation, the Elasticsearch module provides an authentication extension point. This is useful when:
+
+- Using OAuth2/OIDC identity providers for service authentication
+- Tokens need to be refreshed automatically (e.g., tokens that expire hourly)
+- Integrating with internal identity services
+- Running Elasticsearch/OpenSearch clusters secured by token-based authentication
+
+To use custom authentication, create a backend module that provides an auth provider:
+
+```ts title="packages/backend/src/modules/elasticsearchAuth.ts"
+import { createBackendModule } from '@backstage/backend-plugin-api';
+import { elasticsearchAuthExtensionPoint } from '@backstage/plugin-search-backend-module-elasticsearch';
+
+export default createBackendModule({
+  pluginId: 'search',
+  moduleId: 'elasticsearch-custom-auth',
+  register(env) {
+    env.registerInit({
+      deps: {
+        elasticsearchAuth: elasticsearchAuthExtensionPoint,
+      },
+      async init({ elasticsearchAuth }) {
+        elasticsearchAuth.setAuthProvider({
+          async getAuthHeaders() {
+            // Fetch token from your identity service
+            const token = await myTokenService.getToken();
+            return { Authorization: `Bearer ${token}` };
+          },
+        });
+      },
+    });
+  },
+});
+```
+
+Then register this module in your backend:
+
+```ts title="packages/backend/src/index.ts"
+const backend = createBackend();
+
+// Other plugins...
+
+backend.add(import('@backstage/plugin-search-backend'));
+backend.add(import('@backstage/plugin-search-backend-module-elasticsearch'));
+
+/* highlight-add-start */
+backend.add(import('./modules/elasticsearchAuth'));
+/* highlight-add-end */
+
+backend.start();
+```
+
+The `getAuthHeaders` method is called before each request, allowing for just-in-time token retrieval and automatic rotation. When an auth provider is configured, it takes precedence over any static authentication in `app-config.yaml`.
+
+:::note Note
+
+Custom authentication is supported for the `elastic`, `opensearch`, and default providers. The `aws` provider uses AWS SigV4 request signing and does not support custom auth providers.
+
+:::


nojaf:
Could someone please re-run the E2E build. I don't believe this failed due to my changes but had a setup problem instead.

.changeset/elasticsearch-auth-extension-point.md
```diff
@@ -0,0 +1,40 @@
...
+            return { Authorization: `Bearer ${token}` };
+          },
+        });
+      },
+    });
+  },
+});
+```
+
+The auth provider takes precedence over static auth config when set. Note: AWS provider (SigV4) does not support custom auth providers.
```
awanlin: These will get rolled up into the changelog for the related plugin, they generally should be pretty short and concise. The only case where they will be longer would be a breaking change where we would include the manual steps. I would keep line 5 and remove everything else. The extended content is good and should really be added to the docs, likely somewhere on this page: https://backstage.io/docs/features/search/search-engines#elasticsearch

plugins/search-backend-module-elasticsearch/src/auth.ts
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright 2024 The Backstage Authors
```
awanlin:
```suggestion
 * Copyright 2026 The Backstage Authors
```

Net new file so the year should reflect that, we won't change it on existing files though. Please do the same for any of the other new files 👍

nojaf:
> Please add some to this page

@awanlin done as well here!

PR #32379, opened 2026-01-20
fix(cli): use content-based hashing for CSS module class names

Naycon:

CSS module class names were colliding when multiple versions of the same
package were loaded simultaneously. The hash was based on file path and
class name only, not the actual CSS content.

This replaces the default path-based hash with a content-based hash derived
from the CSS content. This ensures:

- Different CSS content produces different class names
- Identical content produces identical class names (stable across builds)
- No dependency on package version or file path for uniqueness

Uses 10 characters of hex-encoded MD5 hash for safe, readable class name suffixes.

### DIFF
--- a/packages/cli/src/modules/build/lib/builder/config.ts
+++ b/packages/cli/src/modules/build/lib/builder/config.ts
@@ -16,7 +16,9 @@
 
 import chalk from 'chalk';
 import fs from 'fs-extra';
+import { createHash } from 'crypto';
 import {
+  basename,
   extname,
   relative as relativePath,
   resolve as resolvePath,
@@ -239,7 +241,18 @@ export async function makeRollupConfigs(
           include: /node_modules/,
           exclude: [/\/[^/]+\.(?:stories|test)\.[^/]+$/],
         }),
-        postcss(),
+        postcss({
+          modules: {
+            generateScopedName(name: string, filename: string, css: string) {
+              const hash = createHash('md5')
+                .update(css)
+                .digest('hex')
+                .slice(0, 10);
+              const file = basename(filename, '.module.css');
+              return `${file}_${name}__${hash}`;
+            },
+          },
+        }),
         forwardFileImports({
           exclude: /\.icon\.svg$/,
           include: [
--- a/.changeset/orange-hoops-matter.md
+++ b/.changeset/orange-hoops-matter.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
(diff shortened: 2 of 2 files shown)


Naycon:
> Yep!
> 
> Would it also be possible to simply use a content hash? Would be a bit safer due to for example package patches

Yea, let me look into that. This was the first quick stab at a solution, but that'd be more elegant.

packages/cli/src/modules/build/lib/builder/config.ts
```diff
@@ -239,7 +241,19 @@ export async function makeRollupConfigs(
           include: /node_modules/,
           exclude: [/\/[^/]+\.(?:stories|test)\.[^/]+$/],
         }),
-        postcss(),
+        postcss({
+          modules: {
+            generateScopedName(name: string, filename: string, css: string) {
+              const normalizedCss = css.replace(/\s+/g, ' ').trim();
```
Rugvip: Normalized in what way? 😅, likely fine to skip this tbh

packages/cli/src/modules/build/lib/builder/config.ts
```diff
@@ -239,7 +241,19 @@ export async function makeRollupConfigs(
...
+              const hash = createHash('md5')
+                .update(normalizedCss)
+                .digest('base64url')
```
Rugvip: I think 8 characters of hex is fairly standard, although we maybe 10-12 even for safety? any form of base64 is gonna get a bit weird because any of `/+-_` can be a bit confusing depending on where they show up in the filename.

packages/cli/src/modules/build/lib/builder/config.ts
Naycon: Yea, was just ensuring indentation changes or newlines added wouldn't cause hash changes. But might be overkill.

packages/cli/src/modules/build/lib/builder/config.ts
Naycon: I think 5 was what was used out of the box from postcss, but happy to bump to 8, and can swap to hex instead of base64.

packages/cli/src/modules/build/lib/builder/config.ts
Naycon: Removed.

packages/cli/src/modules/build/lib/builder/config.ts
Naycon: Updated.

PR #32386, opened 2026-01-20
fix(search-backend-module-elasticsearch): improve index refresh

nojaf:

This PR fixes an inefficient bulk refresh behavior where `refreshOnCompletion: true` causes a full cluster refresh (`_all` indexes) instead of refreshing only the specific index being populated.

### Problem

When using the Elasticsearch/OpenSearch bulk helper with `refreshOnCompletion: true`, all indexes in the cluster are refreshed instead of just the newly created index. This causes:

- **Performance degradation**: Full-cluster refreshes on every bulk indexing operation
- **Increased CPU and I/O**: Unnecessarily refreshing all indexes
- **Multi-tenant impact**: Other applications affected by Backstage's refreshes
- **Scalability issues**: Performance worsens as the cluster grows

### Root Cause

The Elasticsearch bulk helper's `refreshOnCompletion` parameter accepts three values:
- `false` - no refresh (default)
- `true` - refresh **all indexes** (`_all`)
- `"index-name"` - refresh only the specified index

The current code passes a boolean `true`, which triggers a full cluster refresh.

### Original Implementation (Feb 2022)
The original implementation by Eric Peterson (Spotify) was correct:
```typescript
// Commit 2c171166a07663f9572b7e3914d2dc4a121df10e
refreshOnCompletion: that.indexName  // ✅ Refreshes only the specific index
```

#### AOSS Support Added (Sept 2023)
Andrew Ochsner added `skipRefresh` support for AWS OpenSearch Serverless (which doesn't support refresh):
```typescript
// Commit 006df4a58152be772fbbc9acc312195625fcd83a
refreshOnCompletion: !options.skipRefresh && that.indexName  // ✅ Still correct
```

#### Follow-up Fix (Oct 2023)
```typescript
// Commit 6fc10ed06e820e2ca398d11a90b881b59e149d39
refreshOnCompletion: options.skipRefresh !== true && that.indexName  // ✅ Still correct
```

#### Regression Introduced (Oct 2023)
```typescript
// Commit 8ede5296539ac4371e6e7fe6fdb9ce2b47a6d8b8
refreshOnCompletion: options.skipRefresh !== true  // ❌ Bug - refreshes ALL indexes
```

The commit message was simply "pass true or false not index name" with no explanation.

#### Why the Regression Occurred

Based on the PR discussion (#20212), there was confusion between two different APIs:
- The **low-level bulk API** `refresh` parameter (accepts `true`, `false`, `"wait_for"`)
- The **bulk helper** `refreshOnCompletion` parameter (accepts `true`, `false`, or **a string index name**)

During code review, there was discussion about whether to pass `false` or omit the field entirely. The author referenced OpenSearch bulk API documentation, but this documentation refers to the low-level API, not the bulk helper. The change from index name to boolean was made under the mistaken assumption they were equivalent.

### The Fix

Restore the original behavior by passing the specific index name:

```typescript
// Before (buggy)
refreshOnCompletion: options.skipRefresh !== true,

// After (fixed)
refreshOnCompletion: options.skipRefresh !== true ? that.indexName : false,
```

This ensures:
- When `skipRefresh` is not `true` (default): refresh only the specific index being populated
- When `skipRefresh` is `true` (AOSS mode): no refresh occurs

#### Before Fix
- Every bulk indexing operation triggers `POST /_all/_refresh`
- All indexes in the cluster are refreshed
- Significant performance overhead, especially on large clusters

#### After Fix
- Bulk indexing triggers `POST /{index-name}/_refresh`
- Only the specific index being populated is refreshed
- Minimal performance overhead, targeted refresh

### Testing

The existing tests continue to pass. The test mocks don't verify the exact value passed to `refreshOnCompletion`, only whether refresh is called or skipped based on the `skipRefresh` option.

### References

- [Elasticsearch Bulk Helper Documentation](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/client-helpers.html)
- [OpenSearch Bulk API Documentation](https://docs.opensearch.org/latest/api-reference/document-apis/bulk/)
- Original correct implementation: `2c171166a07663f9572b7e3914d2dc4a121df10e`
- Regression commit: `8ede5296539ac4371e6e7fe6fdb9ce2b47a6d8b8`
- AOSS Support PR: #20212

### DIFF
--- a/.changeset/fix-opensearch-bulk-refresh.md
+++ b/.changeset/fix-opensearch-bulk-refresh.md
@@ -2,8 +2,4 @@
 '@backstage/plugin-search-backend-module-elasticsearch': patch
 ---
 
-Fixed inefficient bulk refresh behavior where `refreshOnCompletion: true` caused a full cluster refresh of all indexes instead of refreshing only the specific index being populated.
-
-When using the Elasticsearch/OpenSearch bulk helper, passing `refreshOnCompletion: true` triggers a refresh of `_all` indexes. This change restores the original behavior of passing the specific index name, ensuring only the target index is refreshed after bulk indexing completes.
-
-This fix improves performance for OpenSearch/Elasticsearch deployments, especially in multi-tenant environments or clusters with many indexes.
+Fixed bulk indexing to refresh only the target index instead of all indexes, improving performance in multi-index deployments.


.changeset/fix-opensearch-bulk-refresh.md
```diff
@@ -0,0 +1,9 @@
+---
+'@backstage/plugin-search-backend-module-elasticsearch': patch
+---
+
+Fixed inefficient bulk refresh behavior where `refreshOnCompletion: true` caused a full cluster refresh of all indexes instead of refreshing only the specific index being populated.
+
+When using the Elasticsearch/OpenSearch bulk helper, passing `refreshOnCompletion: true` triggers a refresh of `_all` indexes. This change restores the original behavior of passing the specific index name, ensuring only the target index is refreshed after bulk indexing completes.
+
+This fix improves performance for OpenSearch/Elasticsearch deployments, especially in multi-tenant environments or clusters with many indexes.
```
awanlin: Can we shrink this down into a one or two concise sentences please? Line 5 does a good job of that. This gets rolled up into the changelog where short and simple is much preferred.

nojaf: @awanlin done, let me know if this works!

nojaf:
Thanks for the reviews and following up on this @awanlin!

> Thank you for contributing to Backstage! The changes in this pull request will be part of the 1.48.0 release, scheduled for Tue, 17 Feb 2026.

This is such an amazing automated message! So wonderful to learn when we reasonably can expect this fix to land in a release. Kudos!

PR #32391, opened 2026-01-20
fix(useEntityListProvider): solves bug related to qs library upgrade breaking when more than 20 values in the same query parameter key

sonikro:

Similar to this [fix](https://github.com/backstage/backstage/commit/5c49a00bbd991c9e1cc6ed5e8476d8bdc4b93ae6)

This bug was causing our `EntityOwnerPicker` to crash because in some situations our users had more than **20** groups associated with then, which would cause the **qs.parse** to return an OBJECT instead of an Array of String, which would break everything else.

### DIFF
--- a/.changeset/big-rabbits-think.md
+++ b/.changeset/big-rabbits-think.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-react': patch
 ---
 
-Fixes a bug where the EntityListProvider would not correctly hydrate queryParameters if more than 20 were provided for the same key.
+Fixes a bug where the `EntityListProvider` would not correctly hydrate query parameters if more than 20 were provided for the same key.


.changeset/big-rabbits-think.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-react': patch
+---
+
+fixes bug related to qs library upgrade. makes the EntityListProvider return the correct queryParameters even when there are more than 20 occurrences of the same value
```
aramissennyeydd:
```suggestion
Fixes a bug where the EntityListProvider only hydrated the first 20 array items from query parameters.
```
Talking more to the end user impact?

sonikro: I changed it slightly according to your comment.

.changeset/big-rabbits-think.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-react': patch
+---
+
+Fixes a bug where the EntityListProvider would not correctly hydrate queryParameters if more than 20 were provided for the same key.
```
freben:
```suggestion
Fixes a bug where the `EntityListProvider` would not correctly hydrate query parameters if more than 20 were provided for the same key.
```

Issue #32315, opened 2026-01-14
catalog: Favorites in different namespaces report invalid count


drodil:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
catalog namespace favorite
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

N/A

### 📝 Description & Context

When there are two entities with the same name, but in different namespaces, using the catalog favorite returns an invalid count and entities for favorites.

<img width="1524" height="555" alt="Image" src="https://github.com/user-attachments/assets/5bacd435-7a48-4765-ab0d-2cacd1cb1eeb" />

<img width="985" height="519" alt="Image" src="https://github.com/user-attachments/assets/11f61758-d362-46dc-9e6f-2e77569ef1ea" />

### 👍 Expected Behavior

The favorite count matches the actual number of favorited items

### 🥾 Reproduction steps

1. Add additional entity with different namespace for example to `artist-lookup-component.yaml`:

```yaml
---
apiVersion: backstage.io/v1alpha1
kind: Component
metadata:
  name: artist-lookup
  description: Artist Lookup
  namespace: test
  tags:
    - java
    - data
  links:
    - url: https://example.com/user
      title: Examples Users
      icon: user
    - url: https://example.com/group
      title: Example Group
      icon: group
    - url: https://example.com/cloud
      title: Link with Cloud Icon
      icon: cloud
    - url: https://example.com/dashboard
      title: Dashboard
      icon: dashboard
    - url: https://example.com/help
      title: Support
      icon: help
    - url: https://example.com/web
      title: Website
      icon: web
    - url: https://example.com/alert
      title: Alerts
      icon: alert
  annotations:
    backstage.io/source-template: template:default/springboot-template
    backstage.io/linguist: 'https://github.com/backstage/backstage/tree/master/plugins/playlist'
spec:
  type: service
  lifecycle: experimental
  owner: team-a
  system: artist-engagement-portal
  dependsOn: ['resource:artists-db']
  apiConsumedBy: ['component:www-artist']
```

2. Start the backstage backend
3. Go to catalog page, open artist-lookup from default nameapce, favorite the entity
4. Go back to catalog and select favorites from the filter panel 

### Have you read the Code of Conduct?

[...]

PR #32411, opened 2026-01-21
Made changes to show correct number of favorite components

AdityaK60:

When there are two entities with the same name, but in different namespaces, using the catalog favorite returns an invalid count for favorites. This change corrects the count of the favorite.

image

Resolves issue: #32315


AdityaK60:
@aramissennyeydd and @camilaibs Please review

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
```diff
@@ -177,14 +172,14 @@ export const CatalogTable = (props: CatalogTableProps) => {
...
-        onClick: () => toggleStarredEntity(entity),
+        onClick: () => toggleStarredEntity(entityRefString),
       };
     },
   ];
 
   const currentKind = filters.kind?.label || '';
   const currentType = filters.type?.value || '';
-  const currentCount = typeof totalItems === 'number' ? `(${totalItems})` : '';
+  const currentCount = `(${entities.length})`;
```
Parsifal-M: Will this still work with pagination? Given you have 100 starred entities and are viewing the first page of 20, the header would incorrectly show 20 instead of 100? Or maybe I am wrong 😅

freben (changes requested):
I think that's a valid question ^ - does this really do the right thing anyway? Could you try it out?

If there are duplicates, we should probably rather fix how the original total is calculated, no?

I believe [this](https://github.com/backstage/backstage/blob/c8f65a1e53e5f641de8efbc99b5b2be426b31d1d/plugins/catalog-react/src/components/UserListPicker/useStarredEntitiesCount.ts#L28) is the hook that computes this, as used [here](https://github.com/backstage/backstage/blob/c8f65a1e53e5f641de8efbc99b5b2be426b31d1d/plugins/catalog-react/src/components/UserListPicker/useStarredEntitiesCount.ts#L28).

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
```diff
@@ -98,15 +101,13 @@ export const CatalogTable = (props: CatalogTableProps) => {
...
-    totalItems,
-    paginationMode,
-  } = entityListContext;
+  const { loading, error, entities, filters, pageInfo, paginationMode } =
+    entityListContext;
+
+  // Get accurate counts from the same hooks used by the sidebar
+  const { count: starredCount } = useStarredEntitiesCount();
+  const { count: ownedCount } = useOwnedEntitiesCount();
+  const { count: allCount } = useAllEntitiesCount();
```
Parsifal-M: I don't quite think this is the direction we want to go in, I'm wondering as was mentioned, can we not handle this in the hook itself? I wonder if we could do something here https://github.com/backstage/backstage/blob/c8f65a1e53e5f641de8efbc99b5b2be426b31d1d/plugins/catalog-react/src/components/UserListPicker/useStarredEntitiesCount.ts#L28

so that it only returns the expected kind:namespace/name entities rather than just by name 🤔  just an idea

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
```diff
@@ -168,7 +193,8 @@ export const CatalogTable = (props: CatalogTableProps) => {
       };
     },
     ({ entity }) => {
-      const isStarred = isStarredEntity(entity);
+      const entityRefString = stringifyEntityRef(entity);
+      const isStarred = isStarredEntity(entityRefString);
```
freben, line 210: This will make everybody's currently starred things disappear, right?

plugins/catalog/src/components/CatalogTable/CatalogTable.tsx
```diff
@@ -78,6 +85,24 @@ const refCompare = (a: Entity, b: Entity) => {
   return toRef(a).localeCompare(toRef(b));
 };
 
+// Small internal components that each call only one count hook.
+// Conditionally mounting these (based on the active user filter) avoids making
+// unnecessary catalog API requests when a given filter type is not active.
```
freben, line 93: Hm maybe I misunderstand below, but it doesn't seem like any less computation is being made in practice.

.changeset/gold-suits-arrive.md
```diff
@@ -0,0 +1,8 @@
+---
+'@backstage/plugin-catalog-react': minor
+'@backstage/plugin-catalog': patch
```
freben: This should be split into two files with their own distinct wording.

freben (changes requested):
Hey sorry, just some questions.

.changeset/gold-suits-arrive.md
AdityaK60: Ok

AdityaK60:
Please provide your comments on the changes

PR #32428, opened 2026-01-21
fix(ui): restore React 17 compatibility for useId

Naycon:

Use `useId` from `react-aria` instead of React's built-in hook,
which is only available in React 18+. This restores compatibility
with React 17.

### DIFF
--- a/.changeset/twelve-dolls-tap.md
+++ b/.changeset/twelve-dolls-tap.md
@@ -1,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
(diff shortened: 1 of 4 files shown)


.changeset/twelve-dolls-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Fixed React 17 compatibility by using `useId` from `react-aria` instead of React's built-in hook which is only available in React 18+.
```
freben, line 5:
```suggestion
Fixed React 17 compatibility by using `useId` from `react-aria` instead of the built-in React hook which is only available in React 18+.
```

Issue #17466, opened 2023-04-20
🐛 Bug Report: on config reload the database connection is not reestablished with new dynamic credentials


jgrumboe:
### 📜 Description

We use Hashicorp Vault to create dynamic database credentials for backstage and consul-template to write the app-config.yaml file containing these.
Whenever consul-template fetches new credentials and writes them to app-config.yaml, backstage catches the changes and reloads the config files, but the database connection is not reestablished with the new credentials. That leads to "error: password authentication failed" errors in processing the entities because the database connection still uses the old credentials (fetched when the process was started).

I found issue #7213, where someone described the same use case, and it was closed by #9608. This is why I thought I should be working.
In our case, the credentials are directly in the main app-config.yaml, not in an $include file.

### 👍 Expected behavior

After reloading config files and recognizing that credentials are changing, I would expect the database connection to be newly initialized.

### 👎 Actual Behavior with Screenshots

I searched for the current database username in the app-config.yaml, once right after startup.
![image](https://user-images.githubusercontent.com/5127047/233397637-425a0a2f-2311-4231-b443-2e8d4611b1bf.png)

So, "v-admin-di-backstag-VkCjzxAQSiymMsuCStdB-1681996840" was the initial username.

Then I waited for the TTL to expire (after 20 minutes for testing purpose) and reread the config file.
![image](https://user-images.githubusercontent.com/5127047/233398037-874e240a-55ad-41da-8652-507a476973cf.png)

The username changed to "v-admin-di-backstag-7O3BEEOx8mzoxSUtAF3h-1681999388".

Then I checked the loglines of the backstage container.
`2023-04-20T15:49:05.822+02:00 | {"dd":{"env":"staging","service":"backstage","version":"20230420-1.12.1-17f53315"},"level":"info","message":"Found 4 new secrets in config that will be redacted","service":"backstage"}`
indicates the reload of the config. (Happens regularly in the logfile.)

[...]

PR #32465, opened 2026-01-22
collect all pg config interpretation in one function

freben:
Just a first simplification step toward https://github.com/backstage/backstage/issues/17466

### DIFF
--- a/.changeset/lazy-groups-hunt.md
+++ b/.changeset/lazy-groups-hunt.md
@@ -1,0 +1,5 @@
+---
+'@backstage/backend-defaults': patch
+---
+
+Minor internal restructure of the postgres config loading code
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -186,4 +186,6 @@ docs.json
 tsconfig.typedoc.tmp.json
 
 # Storybook
-dist-storybook/
\ No newline at end of file
+dist-storybook/
+
+.claude
```
benjdlambert: You can add this to `~/.git/gitignore` to do it for all projects, but fine to let this through.

We do have `.cursor/rules` though so wondering actually if we might wanna commit some skills or what not eventually?

benjdlambert (approved):
Woooo :tada:

.gitignore
freben: Yeah good call, reverted this

PR #32468, opened 2026-01-22
test the pg thing too

freben:


packages/backend-defaults/src/entrypoints/database/connectors/postgres.test.ts
```diff
@@ -394,3 +395,491 @@ describe('postgres', () => {
...
+
+    it('omits database from base connection when pluginDivisionMode is database', () => {
+      const config = new ConfigReader({
+        client: 'pg',
+        connection: { host: 'localhost', database: 'shared_db' },
+      });
+
+      const result = computePgPluginConfig(config, 'catalog', prefix);
+
+      expect((result.connection as any).database).toBeUndefined();
```
benjdlambert: ewww, btu fine i guess?

packages/backend-defaults/src/entrypoints/database/connectors/postgres.test.ts
```diff
@@ -394,3 +395,491 @@ describe('postgres', () => {
...
+            connection: { filename: ':memory:' },
+          },
+        },
+      });
+
+      const result = computePgPluginConfig(config, 'catalog', prefix);
+
+      expect((result.connection as any).host).toBeUndefined();
+      expect((result.connection as any).user).toBeUndefined();
+      expect((result.connection as any).filename).toBe(':memory:');
```
benjdlambert: Same here, is there a union here we're missing?

packages/backend-defaults/src/entrypoints/database/connectors/postgres.test.ts
freben: yeah good catch, there was a better type to use

freben:
There's still the matter of the `Knex.Config` type not being strict so it doesn't know whether it's a pg connection type in it or not - so there's still a little bit of casting in the test

PR #32471, opened 2026-01-22
upgrade to `tar` v7

freben:


packages/cli/src/modules/build/lib/buildBackend.ts
```diff
@@ -78,7 +78,7 @@ export async function buildBackend(options: BuildBackendOptions) {
         portable: true,
         noMtime: true,
         gzip: true,
-      } as CreateOptions & { noMtime: boolean },
```
freben, line 81: Biggest v7 improvement is that it has native typescript support. It seems that the types indeed match our intent here - if you try to change the `noMtime` parameter to something invalid it does complain etc.

packages/backend-defaults/src/entrypoints/urlReader/lib/tree/TarArchiveResponse.ts
```diff
@@ -184,7 +182,7 @@ export class TarArchiveResponse implements UrlReaderServiceReadTreeResponse {
           }
 
           // Block symlinks/hardlinks that escape the extraction directory
-          const entry = stat as FileStat & { type?: string; linkpath?: string };
+          const entry = stat as ReadEntry;
```
benjdlambert, line 185: Does this even need a cast now?

freben: Yeah it's a bit unfortunate of an interface - the doc comment for `filter` says

```ts
    /**
     * A function that is called with `(path, stat)` when creating an archive, or
     * `(path, entry)` when extracting. Return true to process the file/entry, or
     * false to exclude it.
     */
    filter?: (path: string, entry: Stats | ReadEntry) => boolean;
```

and in this context we are extracting - we do need to cast.

PR #32472, opened 2026-01-22
docs(table): rewrite Table component documentation

Naycon:

Comprehensive rewrite of the Table component documentation with new interactive examples covering sorting, search, pagination, selection, custom rows, and primitives.

**Documentation changes:**
- New prop definitions with JSX descriptions
- Expanded code snippets for all major use cases
- Added ReactAriaLink component for linking to React Aria docs
- Made PropsTable columns configurable
- Improved Chip styling to support multiline content

**Bug fix (`@backstage/ui`):**
- Fixed "Invalid prop `id` supplied to `React.Fragment`" error in Table component
- Changed `cell` and `header` return types from `ReactNode` to `ReactElement`

**Screenshots**

image

image


cdedreuille (approved):
🙏

.changeset/quiet-carpets-arrive.md
```diff
@@ -0,0 +1,22 @@
...
+
+```diff
+const columns: ColumnConfig<MyItem>[] = [
+  {
+    id: 'name',
+    label: 'Name',
+-   cell: (item) => item.name,
++   cell: (item) => <CellText title={item.name} />,
+-   header: () => 'Name',
++   header: () => <Column>Name</Column>,
```
Rugvip, line 19: Is this the way we want these to be used? It seems quite convenient to be able to supply a string instead, with us defaulting to these components if that's the case and there are sensible defaults?

Naycon: The type change reflects what React Aria expects for rendering functions, `(item: object) => ReactElement`. 

But what I think you're asking is if you have to do this in order to get a column rendered with `Name` as header, and then the answer is no. Just removing the whole `header` part will make sure `Name` is picked up and used, since it's the label. So `header` is only if you need to do something more complicated/advanced that we don't support out of the box yet.

I was struggling a little here to find a realistic example where you'd have done something that causes a migration to be necessary. I'd imagine most people would return a ReactElement already, so they wouldn't need to migrate.

Rugvip: Ah alright, same with the cell value then? As in you'd typically only defined it if you want to do something more interesting. Is `(item) => <CellText title={item.name} />` the default value for `cell`? It's still somewhat interesting to be able to do something like `item => item.name.toUpperCase()` without needing to manually wrap with `CellText`

Naycon: Type-wise it's the same with cell, but we don't have any automatic out of the box-behavior there. Mostly because we're working with `item`, which has an unknown shape (keep in mind the cell definition is made on the column config). We could cook up some way of saying that `itemField: 'name'` means we'll do `item => <CellText title={item.name} />` for you automatically, but that feels like it's not so useful in many cases, and perhaps confusing as well (e.g. do we support `itemField: 'user.name'`, or `itemField: 'user.profiles[0].name'` as syntax). 

So instead we opted for always having to explicitly specify how you want the cells to render, so there's no doubt what gets rendered. Currently we have `CellText` and `CellProfile` for the common use cases we've seen, and then you can use `Cell` as a wrapper (React Aria needs this) for when you want to do something custom outside of that.

Issue #32314, opened 2026-01-14
MenuItem with content too wide ends in uncaught runtime error


Nek:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
MenuItem, MenuTrigger
```

### 🗃️ Project Area

Design System

### 🔗 External Integration

N/A

### 📝 Description & Context

This code:
```ts
import { Menu, MenuItem, MenuTrigger } from '@backstage/ui';

  <MenuTrigger>
    <Button>Support</Button>
    <Menu>
        <MenuItem>#data-application-portal-support</MenuItem>
    </Menu>
  </MenuTrigger>
```
Will produce:
<img width="177" height="69" alt="Image" src="https://github.com/user-attachments/assets/b42f4ff3-9260-4a10-9e94-25620b75ecad" />

Clicking the button will end with runtime errors:
<img width="1084" height="870" alt="Image" src="https://github.com/user-attachments/assets/a0644184-2c49-4148-b5ad-0af45c6f13d5" />

If I shorten the content of `MenuItem` it works just fine.

### 👍 Expected Behavior

I expect no errors to be thrown and menu to open.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

AmbrishRamachandiran:
Assign to me

PR #32490, opened 2026-01-22
MenuItem fix in the UI

AmbrishRamachandiran:
#https://github.com/backstage/backstage/issues/32314

**Changes**

Added overflow handling to .bui-MenuItemContent in Menu.module.css
Long text now truncates with ellipsis instead of causing layout errors
Added LongContent story to demonstrate the fix

**Technical Details**

Set min-width: 0 on wrapper and content to enable flex shrinking
Added overflow: hidden, text-overflow: ellipsis, and white-space: nowrap to content
Icons and arrows get flex-shrink: 0 to maintain proper sizing

### DIFF
--- a/packages/ui/src/components/Menu/Menu.module.css
+++ b/packages/ui/src/components/Menu/Menu.module.css
@@ -107,6 +107,7 @@
     color: var(--bui-fg-primary);
     font-size: var(--bui-font-size-3);
     gap: var(--bui-space-6);
+    min-width: 0;
   }
 
   .bui-MenuItemListBox {
@@ -144,10 +145,17 @@
     display: flex;
     align-items: center;
     gap: var(--bui-space-2);
+    /* Allow content to shrink and handle overflow */
+    min-width: 0;
+    flex: 1;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
 
     & > svg {
       width: 1rem;
       height: 1rem;
+      flex-shrink: 0;
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Menu/Menu.module.css
```diff
@@ -107,6 +107,7 @@
     color: var(--bui-fg-primary);
     font-size: var(--bui-font-size-3);
     gap: var(--bui-space-6);
+    min-width: 0;
```
Naycon, line 110: Is this necessary? This is not a flex item, just a flex container, right?

packages/ui/src/components/Menu/Menu.module.css
```diff
@@ -144,17 +145,25 @@
     display: flex;
     align-items: center;
     gap: var(--bui-space-2);
+    /* Allow content to shrink and handle overflow */
+    min-width: 0;
+    flex: 1;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
Naycon, line 152: Since the element is `display: flex`, ellipsis won't work here. The text element becomes a flex item and gets truncated instead.

packages/ui/src/components/Menu/Menu.module.css
```diff
@@ -144,17 +145,25 @@
     display: flex;
     align-items: center;
     gap: var(--bui-space-2);
+    /* Allow content to shrink and handle overflow */
+    min-width: 0;
+    flex: 1;
```
Naycon, line 150: Is this necessary?

Naycon (changes requested):
Thanks for trying to improve this! However the fix is only partially working (ellipsis is not applied). 

We also need [a changeset](https://github.com/backstage/backstage/tree/master/.changeset), and I think CI failed because of some issue with code styling (did you run prettier?).

PR #32495, opened 2026-01-23
fix(signals): race condition on multiple subscriptions

drodil:

wrapped the connect functionality to internal promise to prevent multiple connect requests being made when there are a lot of subscriptions happening at the same time.

Before:

image

After (just one request):

image

### DIFF
--- a/plugins/signals/src/api/SignalClient.ts
+++ b/plugins/signals/src/api/SignalClient.ts
@@ -30,10 +30,11 @@ const WS_CLOSE_GOING_AWAY = 1001;
 export class SignalClient implements SignalApi {
   static readonly DEFAULT_CONNECT_TIMEOUT_MS: number = 1000;
   static readonly DEFAULT_RECONNECT_TIMEOUT_MS: number = 5000;
-  private ws: WebSocket | null = null;
+  private wsPromise: Promise<WebSocket> | null = null;
   private subscriptions: Map<string, Subscription> = new Map();
-  private messageQueue: string[] = [];
-  private reconnectTo: any;
+  private readonly subscribedChannels: Set<string> = new Set();
+  private messageQueue: Array<JsonObject> = [];
+  private reconnectTimer: NodeJS.Timeout | null = null;
 
   static create(options: {
     identity: IdentityApi;
@@ -58,7 +59,7 @@ export class SignalClient implements SignalApi {
   private identity: IdentityApi;
   private discoveryApi: DiscoveryApi;
   private connectTimeout: number;
-  private reconnectTimeout: number;
+  private reconnectTimeoutMs: number;
 
   private constructor(
     identity: IdentityApi,
@@ -69,7 +70,7 @@ export class SignalClient implements SignalApi {
     this.identity = identity;
     this.discoveryApi = discoveryApi;
     this.connectTimeout = connectTimeout;
-    this.reconnectTimeout = reconnectTimeout;
+    this.reconnectTimeoutMs = reconnectTimeout;
   }
 
   subscribe<TMessage extends JsonObject = JsonObject>(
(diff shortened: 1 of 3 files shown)


plugins/signals/src/api/SignalClient.ts
```diff
@@ -136,55 +148,92 @@ export class SignalClient implements SignalApi {
...
     const apiUrl = await this.discoveryApi.getBaseUrl('signals');
     const { token } = await this.identity.getCredentials();
 
     const url = new URL(apiUrl);
     url.protocol = url.protocol === 'http:' ? 'ws:' : 'wss:';
-    this.ws = new WebSocket(url.toString(), token);
+    const ws = new WebSocket(url.toString(), token);
+
+    // Store the pending WebSocket immediately so unsubscribe() can close it
+    this.pendingWs = ws;
```
freben: Just checking. Did we really need two ws members? What is the use case of having two live at the same time? It seems to me that when we make a (re)connection attempt, it's fine to just throw away the old `this.ws` potentially after sending a close to it. Right? It can do whatever it needs in the background since it's presumably not in a usable state anyway, while the new one is trying to connect.

It's almost like, when the connectPromise is active, we can just assume that no usable ws exists at all.

As a matter of fact - can all of this basically be replaced by a `#wsPromise?: Promise<WebSocket>` member instead? Like, before subscribing or sending, you call `ensureConnection(): Promise<WebSocket>` and that thing does whatever is needed internally like reconnection loops or whatever that results in a usable socket (or an error thrown).

Or did I misunderstand the flow of events here?

drodil: Yep. Listened too much copilot code review, it was much cleaner in the previous commit. Will tune this a bit more

drodil: It should be working now. Also takes care of subscriptions that are received during the connection establishment correctly now. 

Should not listen to Copilot so much, though some of the comments were on point.

plugins/signals/src/api/SignalClient.ts
```diff
@@ -102,91 +98,152 @@ export class SignalClient implements SignalApi {
...
     let connectSleep = 0;
     while (
-      this.ws &&
-      this.ws.readyState !== WebSocket.OPEN &&
+      ws.readyState !== WebSocket.OPEN &&
+      ws.readyState !== WebSocket.CLOSED &&
       connectSleep < this.connectTimeout
     ) {
       await new Promise(r => setTimeout(r, 100));
       connectSleep += 100;
```
Rugvip: Slightly safer to just store the start time and diff with `Date.now()` instead, timeouts can be a bit wonky.

It is nice and simply this but also strikes me as something that really should be driven by events 😅

plugins/signals/src/api/SignalClient.ts
```diff
@@ -213,16 +270,18 @@ export class SignalClient implements SignalApi {
 
     this.reconnectTo = setTimeout(() => {
       this.reconnectTo = null;
-      if (this.ws) {
-        this.ws.close();
+
+      // Close any existing connection if we have one
+      if (this.wsPromise) {
+        this.wsPromise.then(ws => ws.close(WS_CLOSE_NORMAL)).catch(() => {});
```
Rugvip, line 283: Hmm, we'll process the send queue before this happens it looks like, might be an issue? 🤔

plugins/signals/src/api/SignalClient.ts
Rugvip, line 267: We only need to reconnect if we have any desired subscriptions right?

plugins/signals/src/api/SignalClient.ts
```diff
@@ -30,9 +30,10 @@ const WS_CLOSE_GOING_AWAY = 1001;
 export class SignalClient implements SignalApi {
   static readonly DEFAULT_CONNECT_TIMEOUT_MS: number = 1000;
   static readonly DEFAULT_RECONNECT_TIMEOUT_MS: number = 5000;
-  private ws: WebSocket | null = null;
+  private wsPromise: Promise<WebSocket> | null = null;
   private subscriptions: Map<string, Subscription> = new Map();
-  private messageQueue: string[] = [];
+  private readonly subscribedChannels: Set<string> = new Set();
+  private messageQueue: Array<JsonObject> = [];
   private reconnectTo: any;
```
Rugvip: Hah, was wondering what we were reconnecting to until I also started wondering if we had some backoffs for the reconnects :p

While at it, I think call this one reconnectTimeout and the other reconnectTimeoutMs

plugins/signals/src/api/SignalClient.ts
drodil: Yep changed and added TODO for later :D

plugins/signals/src/api/SignalClient.ts
drodil: Indeed, added check(s)

plugins/signals/src/api/SignalClient.ts
drodil: It might, changed location of queue sending to be after connection is available

plugins/signals/src/api/SignalClient.ts
drodil: Naming for the win (+ any type 😆 ). Renamed properly

drodil:
Closing this in favor of #32692

PR #32496, opened 2026-01-23
frontend-plugin-api: add new internal extension input option, complete app-react deprecations

Rugvip:

This completes the deprecations of the app blueprints, removing them from `frontend-plugin-api`, moving them to `app-react` and removing support for extensions from other plugins in `app`. What I realized as part of these moves is that non-app plugins can still interfere with the app plugin if they can still attach themselves to forbidden inputs. They'll be ignored, but if the consuming input is a singleton, the original input will be ignored. Even worse, if it's a required singleton input, the app will break.

To fix this, I though it's worthwhile to introduce an `internal` option for `createExtensionInput`, which limits the input to only accept attachments from the same plugin. This moves the check to the framework level, so that we can both emit proper app errors, and also ensure that conflicting attachments are completely ignored, without taking space from valid ones.

### DIFF
--- a/packages/frontend-plugin-api/src/wiring/createExtensionInput.ts
+++ b/packages/frontend-plugin-api/src/wiring/createExtensionInput.ts
@@ -60,11 +60,11 @@ export interface ExtensionInput<
  * plugin. The following options are available:
  *
  * - `singleton`: If set to `true`, only one extension can be attached to the
- *   input at a time. Additional extension will trigger an app error and be
+ *   input at a time. Additional extensions will trigger an app error and be
  *   ignored.
  * - `optional`: If set to `true`, the input is optional and can be omitted,
  *   this only has an effect if the `singleton` is set to `true`.
- * - `internal`: If set to `true`, only extensions from the same plugins will be
+ * - `internal`: If set to `true`, only extensions from the same plugin will be
  *   allowed to attach to this input. Other extensions will trigger an app error
  *   and be ignored.
  *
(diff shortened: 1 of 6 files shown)


packages/frontend-plugin-api/src/wiring/createExtensionInput.ts
```diff
@@ -38,10 +43,60 @@ export interface ExtensionInput<
...
+ *
+ * The `config` object can be used to restrict the behavior and shape of the
+ * input. By default an input will accept zero or more extensions from any
+ * plugin. The following options are available:
+ *
+ * - `singleton`: If set to `true`, only one extension can be attached to the
+ *   input at a time. Additional extensions will trigger an app error and be
+ *   ignored.
+ * - `optional`: If set to `true`, the input is optional and can be omitted,
+ *   this only has an effect if the `singleton` is set to `true`.
```
freben, line 66: Orly? I thought it would be useful for "at least _someone_ needs to provide something otherwise the plugin can't start". And that it'd essentially make `inputs.foo` not have `| undefined`

Rugvip: > And that it'd essentially make inputs.foo not have `| undefined`

That's what it does, although that only makes sense for singletons. Otherwise `inputs.foo` is gonna be an array that is empty if there are no inputs. We could have `optional: false` mean `> 0` items in the array, but honestly that doesn't seem super useful.

PR #32498, opened 2026-01-23
app-visualizer: improve rendering performance of details page

Rugvip:

🧹, bit of performance optimization of the details page. It progressively renders the page rather than trying to do it all upfront. The browser used to lock up for quite a while in big apps.

### DIFF
--- a/plugins/app-visualizer/src/components/AppVisualizerPage/DetailedVisualizer.tsx
+++ b/plugins/app-visualizer/src/components/AppVisualizerPage/DetailedVisualizer.tsx
@@ -31,7 +31,7 @@ import {
   RiCloseCircleLine as DisabledIcon,
 } from '@remixicon/react';
 import { Focusable } from 'react-aria-components';
-import { memo, useMemo, useState, useEffect, useRef } from 'react';
+import { memo, useMemo, useState, useEffect, useRef, Fragment } from 'react';
 
 function getContrastColor(bgColor: string): string {
   const hex = bgColor.replace('#', '');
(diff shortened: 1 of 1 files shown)


plugins/app-visualizer/src/components/AppVisualizerPage/DetailedVisualizer.tsx
```diff
@@ -105,131 +106,233 @@ function getFullPath(node?: AppNode): string {
...
-    borderRadius: '10px',
-    color,
-    backgroundColor,
-    display: 'flex',
-    alignItems: 'center',
-    fontWeight:
-      'var(--bui-font-weight-regular)' as React.CSSProperties['fontWeight'],
-  };
+function collectAllNodes(node: AppNode): AppNode[] {
+  const nodes: AppNode[] = [node];
```
freben: Should this be made a set? If you have multiple attachment points, will this lead to dupes?

plugins/app-visualizer/src/components/AppVisualizerPage/DetailedVisualizer.tsx
```diff
@@ -105,131 +106,233 @@ function getFullPath(node?: AppNode): string {
...
+  for (const children of node.edges.attachments.values()) {
+    for (const child of children) {
+      nodes.push(...collectAllNodes(child));
```
freben: probably doesn't matter in practice for any reasonably sized app tree but this recursive re-spread over and over 👀  :)

plugins/app-visualizer/src/components/AppVisualizerPage/DetailedVisualizer.tsx
```diff
@@ -105,131 +106,233 @@ function getFullPath(node?: AppNode): string {
...
+    const rendered = new Set<string>();
+
+    const processBatch = () => {
+      const endIndex = Math.min(currentIndex + batchSize, allNodes.length);
+      for (let i = currentIndex; i < endIndex; i++) {
+        rendered.add(allNodes[i].spec.id);
+      }
+      currentIndex = endIndex;
+
+      setRenderedNodes(new Set(rendered));
```
freben: Can this perhaps be made even simpler? It wondered if this could be a super thin helper style component that accepts an anonymous element array and then has a single state variable.

```tsx
const ProgressiveCollection = (props: {
  items: Element[]
}) => {
  const [maxIndex, setMaxIndex] = useState(0);

  // an effect that ticks the index += 10 if it's less than the items array length

  return (
    <>
      {items.slice(0, maxIndex).map((item, index) => {
        <Fragment key={index}>
          {item}
        </Fragment>
      })}
    </>
  );
}
```

Unless you think that it's even a problem that the parent has to JSX-render all the children upfront - which I suppose should not be the case.

plugins/app-visualizer/src/components/AppVisualizerPage/DetailedVisualizer.tsx
Rugvip: Multiple attachment points still produce duplicates, so that's fine. These are guaranteed to be unique already at the framework level

plugins/app-visualizer/src/components/AppVisualizerPage/DetailedVisualizer.tsx
Rugvip: Yeah not too concerned about this one

plugins/app-visualizer/src/components/AppVisualizerPage/DetailedVisualizer.tsx
Rugvip: Forwarded verbatim 😁 

Didn't get it quite right as the nesting was removed, but using this strategy at every level seems to work just fine :shipit: 👍

plugins/app-visualizer/src/components/AppVisualizerPage/DetailedVisualizer.tsx
```diff
@@ -105,131 +106,149 @@ function getFullPath(node?: AppNode): string {
...
-  const { id } = dataRef;
-  const instance = node?.instance;
+function ProgressiveCollection({
+  items,
+  batchSize = 10,
+}: {
+  items: Array<() => React.ReactElement>;
+  batchSize?: number;
+}) {
+  const [maxIndex, setMaxIndex] = useState(0);
```
freben, line 116: Could start at batchSize to not waste a tick

Rugvip: That still risks a bit of an explosion for broad trees, i.e. 10*10*10 immediate renders

PR #32503, opened 2026-01-23
feat(auth-backend-module-github): export emailMatchingUserEntityProfileEmail resolver

### LINKED ISSUE #32500: feat(auth-backend-module-github): export `emailMatchingUserEntityProfileEmail` resolver
### 📜 Issue Labels

### 🔎 Search Terms

```plain
auth-backend-module-github, github provider, emailMatchingUserEntityProfileEmail, auth resolver
```

### 🗃️ Project Area

Auth

### 🔖 Need

Currently, when configuring the GitHub authentication provider using the New Backend System, users cannot easily use the standard email matching strategy purely through configuration.

The emailMatchingUserEntityProfileEmail resolver—which matches an incoming identity to a User entity in the Catalog via email—is available for other providers (like Google or Okta) or generally via auth-node, but it is not exposed or registered by default in the @backstage/plugin-auth-backend-module-github-provider. [Resolvers Implementation](https://github.com/backstage/backstage/blob/24eb7d793358f95b027164ec739e6cd1ff8b0a69/plugins/auth-backend-module-github-provider/src/resolvers.ts)

This limitation forces adopters to write custom TypeScript code to implement basic email matching, which contradicts the low-code/configuration-first goal of the New Backend System.

### 📝 Proposal

I propose updating @backstage/plugin-auth-backend-module-github-provider to explicitly register and export the emailMatchingUserEntityProfileEmail resolver.

This would allow adopters to enable email-based login simply by updating their app-config.yaml, without modifying source code:

```YAML
auth:
  providers:
    github:
      production:
        signIn:
          resolvers:
            - resolver: emailMatchingUserEntityProfileEmail
```
The implementation should verify that the GitHub profile actually contains an email address (potentially requiring the user:email scope) and then perform the standard catalog lookup.

### 🔄 Alternatives

_No response_

nunoferna:
Closes #32500

I added the `emailMatchingUserEntityProfileEmail` logic that was missing for Github authentication.

### DIFF
--- a/.changeset/five-memes-judge.md
+++ b/.changeset/five-memes-judge.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-auth-backend-module-github-provider': minor
+---
+
+Added the `emailMatchingUserEntityProfileEmail` sign-in resolver that matches users by their email with `spec.profile.email` Catalog Users.


vinzscam (changes requested):
have you tried to use this resolver before this change? It looks like it should be already available for github, see: https://github.com/backstage/backstage/blob/master/plugins/auth-backend-module-github-provider/src/module.ts#L41

nunoferna:
> have you tried to use this resolver before this change? It looks like it should be already available for github, see: https://github.com/backstage/backstage/blob/master/plugins/auth-backend-module-github-provider/src/module.ts#L41

You're right this solution won't fix. The problem may be the scope. I am doing without enterprise I am using a personal GithubApp should go to GithubOAuthApp. Even though I could login with username.

PR #32508, opened 2026-01-23
BUI changelog updates

cdedreuille:

Adds new structure to Backstage UI's changelog leveraging the existing changelog release system with a new template. This allows us to extract both affected components, breaking changes, and migration notes.

This also fixes the indentation that was making the page quite unreadable.

The docs were broken after the refactor of our definition hooks. This PR fixes the client vs server issues as well. https://github.com/backstage/backstage/pull/32308

### Before
CleanShot 2026-01-23 at 22 18 45@2x

### After
CleanShot 2026-01-23 at 22 18 30@2x

### DIFF
--- a/docs-ui/scripts/sync-changelog.mjs
+++ b/docs-ui/scripts/sync-changelog.mjs
@@ -897,6 +897,30 @@ async function main() {
     console.log(`  - ${comp}: ${count} ${count === 1 ? 'entry' : 'entries'}`);
   });
 
+  // Warn about unknown components
+  const unknownComponents = [];
+  allEntries.forEach(entry => {
+    const fullText = entry.description;
+    const componentMatch = fullText.match(
+      /Affected components?:[ \t]*([^\n]+)/i,
+    );
+    if (componentMatch) {
+      const names = componentMatch[1].split(',').map(n => n.trim());
+      names.forEach(name => {
+        if (!mapComponentName(name, validComponents)) {
+          unknownComponents.push(name);
+        }
+      });
+    }
+  });
+
+  if (unknownComponents.length > 0) {
+    console.log('\n⚠️  Unknown components (skipped):');
+    [...new Set(unknownComponents)].forEach(name => {
+      console.log(`  - ${name}`);
+    });
+  }
+
   // Create changelogs directory if it doesn't exist
   const changelogsDir = path.join(__dirname, '../src/utils/changelogs');
   if (!fs.existsSync(changelogsDir) && !dryRun) {


docs-ui/scripts/sync-changelog.mjs
```diff
@@ -187,8 +187,15 @@ async function parseChangelogMd(changelogPath, sinceVersion, validComponents) {
...
-        currentSection = null;
+        // Skip pre-release versions (e.g., 0.11.0-next.1, 0.9.0-alpha.2) to avoid duplicates
+        // Pre-release commits are included in the final release version
+        if (/^\d+\.\d+\.\d+-/.test(versionText)) {
+          currentVersion = null;
+          currentSection = null;
+        } else {
+          currentVersion = versionText;
+          currentSection = null;
+        }
```
Naycon: The `normalizeVersion` function is already taking care of stripping `-next.1` and similar off of the version, so you end up with `0.11.0` for both the examples in the comment above. Is that not working?

docs-ui/scripts/sync-changelog.mjs
```diff
@@ -314,16 +323,54 @@ async function parseListItem(
       .map(name => mapComponentName(name, validComponents))
       .filter(Boolean);
 
-    // Optionally strip "Affected components:" line from description
+    // Strip "**Affected components:**" line from description
+    description = description
+      .replace(/\n*\*\*Affected components:\*\*[ \t]*[^\n]+/g, '')
+      .trim();
+  } else {
+    // Fallback: try old format without bold markers
```
Naycon: Do we need this, or should we simply reformat the changelog? It would simplify maintaining the code.

docs-ui/scripts/sync-changelog.mjs
```diff
@@ -332,34 +379,34 @@ async function parseListItem(
...
+
+  // Version >= 1.0.0: Only Major Changes are breaking
+  if (major >= 1) {
+    return section === 'Major Changes';
   }
 
-  return undefined;
+  // Version < 1.0.0: Both Major and Minor Changes are breaking
+  return section === 'Major Changes' || section === 'Minor Changes';
 }
```
Naycon, line 362: What's the reason for the old logic not working? It looks sane to me.

I think we should always rely on version numbers for this detection and not do string matching. We should follow semver carefully, which means we should _never_ end up with breaking changes that are only described as such in the changelog but not properly marked in in the version number.

docs-ui/scripts/sync-changelog.mjs
cdedreuille: I saw a use case where the next appeared again. I can double check.

PR #32512, opened 2026-01-24
catalog-backend: attribute provider connection failures to modules

Rugvip:
## Summary

This PR adds support for extension point factories and module startup failure reporting, enabling catalog entity provider connection failures to be attributed to module startup results.

## Changes

- **@backstage/backend-plugin-api**: Added support for extension point factories. This makes it possible to call `registerExtensionPoint` with a single options argument and provide a factory for the extension point rather than a direct implementation. The factory is passed a context with a `reportModuleStartupFailure` method that makes it possible for plugins to report and attribute startup errors to the module that consumed the extension point.

- **@backstage/backend-app-api**: Added support for extension point factories, along with the ability to report module startup failures via the extension point factory context.

- **@backstage/plugin-catalog-backend**: Attribute catalog entity provider connect failures to module startup results using module handles. This makes it possible to control backend startup with `onPluginModuleBootFailure` configuration, even if the failure happens when connecting the entity provider, rather than at module init.

- **@backstage/backend-test-utils**: Updated `startTestBackend` to support factory-based extension points (v1.1 format) in addition to the existing direct implementation format.

## Checklist

### DIFF
--- a/.changeset/catalog-provider-module-failures.md
+++ b/.changeset/catalog-provider-module-failures.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-backend': minor
 ---
 
-Failure to connect catalog providers are now attributed to the module that provided the failing provider. This means that such failures will be reported as module startup failures rather than a failure to start the catalog plugin, and will therefore respect `onPluginModuleBootFailure` configuration instead.
+Failures to connect catalog providers are now attributed to the module that provided the failing provider. This means that such failures will be reported as module startup failures rather than a failure to start the catalog plugin, and will therefore respect `onPluginModuleBootFailure` configuration instead.


packages/backend-plugin-api/src/wiring/types.ts
```diff
@@ -119,6 +150,35 @@ export interface InternalBackendModuleRegistration {
...
+/** @internal */
+export type ExtensionPointRegistration = {
+  extensionPoint: ExtensionPoint<unknown>;
+  factory: (context: ExtensionPointFactoryContext) => unknown;
+};
+
+/** @internal */
+export interface InternalBackendPluginRegistrationV1_1 {
+  pluginId: string;
+  type: 'plugin-v1.1';
```
freben, line 162: I presume you're choosing the version string in the "backward compatibility sense", that as long as you obey keeping your backend ahead of all features that you install into it you're fine? Sure, that's probably useful, forever naming these as v2, v3 etc no matter the scale of changes could have worked too but this may be informative

packages/backend-plugin-api/src/wiring/types.ts
```diff
@@ -36,6 +36,20 @@ export type ExtensionPoint<T> = {
...
+ *
+ * @public
+ */
+export interface ExtensionPointFactoryContext {
+  /**
+   * Report a startup failure that happened as part of using an extension that
+   * the module provided. This should be called before the plugin's `init`
+   * function returns.
+   */
+  reportModuleStartupFailure(options: { error: Error }): void;
```
freben, line 50: Hm. Sorry I'm being all over the place but this makes me think. It's a bit unfortunate that this is something you have to learn and remember to call. I foresee a bunch of `if (context)` etc popping up.

If we invert the consumption pattern, can we just use regular throws instead? As in, the provided extension is only available in a callback whose errors the framework can catch.

```ts
for (p of providerProviders /* haha */) {
  await p.consume(async provider => {
    // no need for a try/catch or a context
    await provider.connect(createConnection(provider));
  });
}
```

Not sure. Had to bring it up because this explicit reporting seemed like a potential minor smell. Thoughts?

packages/backend-plugin-api/src/wiring/types.ts
Rugvip: yep I was aiming to keep it aligned with the `v1` on the parent object as well. Didn't really feel like bumping the entire thing to v2 when we're making a non-breaking change on the consumer side.

packages/backend-plugin-api/src/wiring/types.ts
Rugvip: I do think this is a bit of a special case where it's somewhat fine to expect consumers to do the right thing. I'm not sure that level of hand-holding is beneficial tbh.

In this specific case it's not quite as simple as a consume callback either right? we first collect all providers and deduplicate, merging them with the built-in ones. This is why we need the `if (context) {` in this case, because not all providers come from a module. It's not until later on after we've figured out which providers to connect that we actually call `.connect`. That is the initial consumption and the initialization that happens later are logically far enough apart that I don't think a `.consume` pattern would be very pretty?

PR #32517, opened 2026-01-25
fix: Support Prettier v3 in repo-tools (#27559)

### LINKED ISSUE #27559: 🐛 Bug Report: `backstage-repo-tools api-reports` fails to format generated api report files when using Prettier v3
### 📜 Description

Prettier is [used to format generated API reports](https://github.com/backstage/backstage/blob/master/packages/repo-tools/src/commands/api-reports/api-extractor.ts#L198-L208). This code relies on the `resolveConfig.sync` method in Prettier, which [was removed in v3](https://github.com/prettier/prettier/pull/12574). This means the code hits the catch block and silently fails.

Note that even if we move away from using this method, the `format` method itself in prettier has _also_ moved to be async, so we'll likely need a deeper refactoring here if we want to support Prettier v3. It's not clear how we'd actually go about that though, since the curretn approach relies on patching a synchronous method. Perhaps one solution here is a feature request to api-extractor to add support for supplying a custom formatter.

### 👍 Expected behavior

* API report generation should succeed with both prettier v2 and v3
* When formatting fails, we should log an error to make debugging easier

Report generation succeeds but files are not formatted.

### 👟 Reproduction steps

* Update to prettier v3
* Generate api-reports in any Backstage codebase
* Observe that the resulting files are not formatted.

_No response_

### 🖥️ Your Environment

_No response_

None

GabDug:

Attempt to fix #27559, by supporting both Prettier v2 and v3 in repo-tools api-reports command. Also tested against alpha [v4](https://github.com/prettier/prettier/pull/18261).

I have not bumped the required version yet, to allow for a smoother migration (and because we currently still need v2 in the workspace for changeset, see https://github.com/changesets/changesets/pull/1744) 

### Remarks

- Used the [`@prettier/sync`](https://www.npmjs.com/package/@prettier/sync) package where the async version can't be used
  - No noticeable performance regression in my benchmarks (within margin of error)
- Ugly typing to support both v2 and v3, but isolated to a single file to it'll be easy to remove once v3 is required.
- Updated the config resolution to be more specific
  - We provided a directory, it's better to provide a filepath, as Prettier configuration resolution changes depending on other factors (e.g. load the .editorconfig md rules for .md files)
 - Tested that it matches Prettier CLI results
   - `report.api.md` & `report-*.api.md` could be removed from `.prettierignore`
   - No conflict with `yarn build:api-reports` (although we can see some whitespace changes, ignored by `ApiReportGenerator`)
 - Tested with v2 and a forced resolution to v3 and upcoming v4 `4.0.0-alpha.13`

### Alternative explored

- Patch `ApiReportGenerator.areEquivalentApiFileContents` and compare the contents by stripping all whitespace, and other symbols modified by prettier (`,’";`…)
    - Flaky, very lossy
    - Would not work if consumers of repo-tools have some prettier plugins like [prettier-plugin-sort-imports](https://www.npmjs.com/package/prettier-plugin-sort-imports)
- Patch `ApiReportGenerator.areEquivalentApiFileContents` and `ApiReportGenerator.generateReviewFileContent` to add the hash of the unformatted content and compare that (& run prettier outside of ApiReportGenerator)
    - Works well, but api-report could be manually edited and it wouldn’t report it as changed: not reliable enough, too risky
- Upstream changes
  - [PR to add a sync formatting callback](https://github.com/microsoft/rushstack/pull/3871) is opened since 2023, so I doubt a refactor to support an async one is reasonable

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -14,8 +14,10 @@ type-docs
 # old reports
 api-report.md
 api-report-*.md
+# new reports
 report.api.md
 report-*.api.md
+
 knip-report.md
 cli-report.md
 plugins/scaffolder-backend/sample-templates
--- a/packages/repo-tools/src/commands/api-reports/api-reports/patchApiReportGeneration.ts
+++ b/packages/repo-tools/src/commands/api-reports/api-reports/patchApiReportGeneration.ts
@@ -17,7 +17,7 @@
 import { ExtractorMessage } from '@microsoft/api-extractor';
 import { AstDeclaration } from '@microsoft/api-extractor/lib/analyzer/AstDeclaration';
 import { Program } from 'typescript';
-import { tryRunPrettier } from '../common/tryRunPrettier';
+import { tryRunPrettier } from '../common';
 import { paths as cliPaths } from '../../../lib/paths';
 
 let applied = false;
@@ -153,7 +153,6 @@ export function patchApiReportGeneration() {
        * the middle of the process as API Extractor does a comparison of the contents of the old
        * and new files during generation. This inserts the formatting just before that comparison.
        */
-
       const content = originalGenerateReviewFileContent.call(
         this,
         collector,


Rugvip (approved):
Nice, thank you! 😁  🎉 

Obviously a bit sketchy with `make-synchronized` under the hood but not seeing any other great ways forward tbh.

Had to double check if the API had been updated on the API Extractor side to have an async variant... nope :<

Let's  :shipit: 👌

PR #32519, opened 2026-01-25
fix(search): prevent memory leak in LunrSearchEngine docStore

Sanchit2662:
## Summary

This PR fixes a memory leak in `LunrSearchEngine` where indexed documents were retained in memory across periodic reindexing cycles, even after being deleted from their source.

While Lunr indices were correctly replaced on each run, the backing document store (`docStore`) was merged additively, causing unbounded growth in long-running Backstage deployments with scheduled indexing.

---

## Fix

The document store has been made **type-aware**, ensuring that documents for a given search type are **fully replaced** on each reindex rather than merged globally.

### Key changes

* Replaced the single global document store with per-type stores:

```ts
protected docStores: Record<string, Record<string, IndexableDocument>> = {};
```

* Updated indexing behavior to replace documents per type:

```ts
this.docStores[type] = newDocuments;
```

* Preserved backward compatibility via a deprecated accessor:

```ts
/** @deprecated Use docStores instead */
protected get docStore() {
  return Object.values(this.docStores).reduce((a, s) => ({ ...a, ...s }), {});
}
```

This ensures deleted documents are automatically removed from memory on subsequent reindexing runs, while keeping search behavior unchanged.

---

* [x] All your commits have a `Signed-off-by` line in the message. ([more info])

### DIFF
--- a/.changeset/hungry-pens-clap.md
+++ b/.changeset/hungry-pens-clap.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-search-backend-node': patch
+---
+
+Fixed memory leak in `LunrSearchEngine` where the document store grew unbounded across reindexing cycles. The fix replaces the single global `docStore` with per-type document stores that are fully replaced during each indexing cycle, preventing deleted documents from accumulating in memory.
--- a/plugins/search-backend-node/src/engines/LunrSearchEngine.ts
+++ b/plugins/search-backend-node/src/engines/LunrSearchEngine.ts
@@ -66,11 +66,6 @@ export class LunrSearchEngine implements SearchEngine {
     );
   }
 
-  /** @deprecated Use docStores instead */
-  protected set docStore(value: Record<string, IndexableDocument>) {
-    this.docStores.__legacy__ = value;
-  }
-
   constructor(options: { logger: LoggerService }) {
     this.logger = options.logger;
     const uuidTag = uuid();
@@ -261,7 +256,7 @@ export class LunrSearchEngine implements SearchEngine {
     const realResultSet: IndexableResultSet = {
       results: results.slice(offset, offset + pageSize).map((d, index) => ({
         type: d.type,
-        document: this.docStore[d.result.ref],
+        document: this.docStores[d.type]?.[d.result.ref],
         rank: page * pageSize + index + 1,
         highlight: {
           preTag: this.highlightPreTag,
@@ -269,7 +264,7 @@ export class LunrSearchEngine implements SearchEngine {
           fields: parseHighlightFields({
             preTag: this.highlightPreTag,
             postTag: this.highlightPostTag,
-            doc: this.docStore[d.result.ref],
+            doc: this.docStores[d.type]?.[d.result.ref],
             positionMetadata: d.result.matchData.metadata as any,
           }),
         },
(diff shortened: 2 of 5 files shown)


Sanchit2662:
Hi @freben , 
I’ve submitted this fix to prevent a memory leak in the LunrSearchEngine docStore, ensuring long-running Backstage deployments don’t retain deleted documents in memory.
Whenever you have some time, I’d really appreciate any feedback and a review on this PR. Thanks so much!

awanlin:
Hi @Sanchit2662, I haven't had a chance to review this and it's likely good for us to have in place but your last comment really jumped at me - "LunrSearchEngine docStore, ensuring long-running Backstage deployments don’t retain deleted documents in memory." The Lunr Search Engine is intended for local development use only and NOT to be used outside of that context, you should be using the Postgres Search Engine, we do [call this out in the documentation](https://backstage.io/docs/features/search/search-engines#lunr) as well. Have I misunderstood your comment?

Sanchit2662:
Thanks for pointing this out , this is my mistake.

You’re absolutely right: Lunr is intended for local development only, and for long-running / production deployments the Postgres Search Engine should be used. Given that, this PR is addressing a non-issue in practice. Sorry for the confusion

awanlin:
No worries @Sanchit2662, just want to make sure you're following the guidance we've recomended. I still think this PR is worthwhile and we'll work to review it 👍

.changeset/hungry-pens-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-search-backend-node': patch
+---
+
+Fixed memory leak in `LunrSearchEngine` where the document store grew unbounded across reindexing cycles. The fix replaces the single global `docStore` with per-type document stores that are fully replaced during each indexing cycle, preventing deleted documents from accumulating in memory.
```
Rugvip, line 5: Need to add `reindexing` to https://github.com/backstage/backstage/blob/master/.github/vale/config/vocabularies/Backstage/accept.txt to avoid the error here

Rugvip (changes requested):
Thanks, just need a fix for the vale errors

plugins/search-backend-node/src/engines/LunrSearchEngine.ts
```diff
@@ -53,14 +53,21 @@ export type LunrQueryTranslator = (query: SearchQuery) => ConcreteLunrQuery;
...
-  protected docStore: Record<string, IndexableDocument>;
+  protected docStores: Record<string, Record<string, IndexableDocument>> = {};
   protected logger: LoggerService;
   protected highlightPreTag: string;
   protected highlightPostTag: string;
 
+  /** @deprecated Use docStores instead */
+  protected get docStore(): Record<string, IndexableDocument> {
+    return Object.values(this.docStores).reduce(
+      (acc, store) => ({ ...acc, ...store }),
```
vinzscam:
```suggestion
      (acc, store) => Object.assign(acc, store),
```

freben:
@Sanchit2662 are you around to take this across the finish line?

Sanchit2662:
Hey @freben , apologies for the gap here. I got caught up with some other work, but I'm resuming this now and will get it over the line.

Sanchit2662:
Hi @freben , made the required changes . Please tell if any further changes are required

freben:
@Sanchit2662 Do you have any input on the Copilot review comments?

PR #32540, opened 2026-01-27
fix(ui): prevent infinite render loop in Tabs with routed tabs

Naycon:

When navigating to a URL that doesn't match any tab href, the Tabs
component would pass `selectedKey={null}` to React Aria's Tabs,
which triggers an infinite render loop.

This fix uses an empty string instead of null to indicate no
selection, which React Aria handles correctly.

### DIFF
--- a/.changeset/rare-papers-decide.md
+++ b/.changeset/rare-papers-decide.md
@@ -2,4 +2,4 @@
 '@backstage/ui': patch
 ---
 
-Fixed an infinite render loop in Tabs when navigating to a URL that doesn't match any tab href.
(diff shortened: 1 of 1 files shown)


.changeset/rare-papers-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Fixed an infinite render loop in Tabs when navigating to a URL that doesn't match any tab href.
```
freben:
```suggestion
Fixed an infinite render loop in Tabs when navigating to a URL that doesn't match any tab `href`.
```

Naycon: Thanks, bites me every time that one! 😂

PR #32541, opened 2026-01-27
docs(ui): comprehensive documentation improvements for all components

Naycon:

Improves documentation quality and consistency across all UI components in docs-ui.

**Documentation improvements:**
- Added descriptions to all props across all components
- Fixed prop types to match source implementations
- Added missing props (onAction, isInvalid, aria-label, etc.)
- Corrected default values and removed non-existent props
- Added side-by-side layout option to Snippet component
- Added React Aria reference links to all components

**Bug fix:**
- Removed non-existent `large` size variant from Button types

### Screenshot

image

### DIFF
--- a/.changeset/tired-queens-cross.md
+++ b/.changeset/tired-queens-cross.md
@@ -1,0 +1,14 @@
+---
+'@backstage/ui': minor
+---
+
+**BREAKING**: Removed `large` size variant from Button component as it was never implemented.
+
+**Migration:**
+
+```diff
+- <Button size="large">Click me</Button>
++ <Button size="medium">Click me</Button>
+```
+
+**Affected components:** Button
--- a/packages/ui/src/components/Button/types.ts
+++ b/packages/ui/src/components/Button/types.ts
@@ -20,7 +20,7 @@ import type { LeafSurfaceProps, Responsive } from '../../types';
 
 /** @public */
 export type ButtonOwnProps = LeafSurfaceProps & {
-  size?: Responsive<'small' | 'medium' | 'large'>;
+  size?: Responsive<'small' | 'medium'>;
   variant?: Responsive<'primary' | 'secondary' | 'tertiary'>;
   iconStart?: ReactElement;
   iconEnd?: ReactElement;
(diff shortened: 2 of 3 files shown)


packages/ui/src/components/Button/types.ts
```diff
@@ -20,7 +20,7 @@ import type { LeafSurfaceProps, Responsive } from '../../types';
 
 /** @public */
 export type ButtonOwnProps = LeafSurfaceProps & {
-  size?: Responsive<'small' | 'medium' | 'large'>;
+  size?: Responsive<'small' | 'medium'>;
```
cdedreuille, line 23: Why this change here @Naycon? Is this a mistake?

Naycon: Nope, this is on purpose. Docs listed small and medium, types had small, medium and large, but there are only styles for small and medium. 

Made as a separate commit to highlight the change and includes a changeset. This is aligning docs and code, and was the only place where the documentation was actually right and the code wrong. :)

cdedreuille (approved):
Thanks @Naycon. Looks good 🙏

.changeset/tired-queens-cross.md
```diff
@@ -0,0 +1,14 @@
+---
+'@backstage/ui': minor
+---
+
+**BREAKING**: Removed `large` size variant from Button component as it was never implemented.
```
Rugvip, line 5: If this was never functional you don't need to call it a breaking change. No most important thing to consider is breaking behavior. If this broke a typescript type that was already broken, that's fine.

Naycon: I think this is the crux - the typescript type wouldn't have been broken. The only "valid" use case for `size="large"` though is that you bring your own custom styles for the large button. In theory this could have happened (and would've worked), and if you relied on this your code would've compiled fine before this change but won't after. 

Happy to change to patch and remove the breaking label if you still think this should count as non-breaking.

Rugvip: yep I see how that could be a bigger risk in the future, let's opt keep it correct, that is as it is :shipit:

PR #32542, opened 2026-01-27
feat: Introduce ToastApi for rich notifications in the new frontend system

cdedreuille:

This PR introduces a new `ToastApi` for displaying rich toast notifications in the new frontend system, along with a modern `ToastDisplay` component that replaces `AlertDisplay` from `@backstage/core-components`.

### Why a new API?

After discussing with @Rugvip, we decided to introduce a dedicated `ToastApi` rather than extending the existing `AlertApi`. The key differences that warranted a new API:

- **Props structure**: Toast uses `title` + `description` whereas Alert uses a single `message`
- **Status values**: Toast uses `status` ('info' | 'success' | 'warning' | 'danger') vs Alert's `severity`
- **Timeout control**: Toast allows each notification to specify its own `timeout`, whereas Alert uses a global `transient` display mode
- **Additional features**: Toast supports action links, and programmatic dismissal

### Deprecation of AlertApi

`AlertApi` is now deprecated and will be removed in a future release. Deprecation warnings have been added to:

- `alertApiRef` in `@backstage/core-plugin-api`
- `alertApiRef` in `@backstage/frontend-plugin-api`
- `AlertApiForwarder` in `@backstage/core-app-api`

When using `AlertApi`, developers will see console warnings guiding them to migrate to `ToastApi`.

### Migration Path

The `ToastDisplay` component subscribes to **both** `ToastApi` and `AlertApi`, providing a seamless migration path:

- Existing code using `AlertApi` continues to work unchanged
- New code can adopt `ToastApi` for richer notifications
- Both APIs can be used side-by-side until `AlertApi` is fully deprecated

#### Migration Guide

| AlertApi                                     | ToastApi                                    |
| -------------------------------------------- | ------------------------------------------- |
| `message: string`                            | `title: ReactNode`                          |
| `severity: 'error'`                          | `status: 'danger'`                          |
| `severity: 'success' \| 'info' \| 'warning'` | `status: 'success' \| 'info' \| 'warning'`  |
| `display: 'transient'`                       | `timeout: 5000` (or custom ms)              |
| `display: 'permanent'`                       | omit `timeout`                              |
| `post()` returns `void`                      | `post()` returns `string` (key for dismiss) |

#### Example Migration

```typescript
// Before (AlertApi) - DEPRECATED
import { alertApiRef, useApi } from '@backstage/core-plugin-api';

const alertApi = useApi(alertApiRef);
alertApi.post({
  message: 'Entity saved successfully',
  severity: 'success',
  display: 'transient',
});

// After (ToastApi)
import { toastApiRef, useApi } from '@backstage/frontend-plugin-api';

const toastApi = useApi(toastApiRef);
toastApi.post({
  title: 'Entity saved successfully',
  status: 'success',
  timeout: 5000,
});
```

### Usage

```typescript
import { toastApiRef, useApi } from '@backstage/frontend-plugin-api';

const toastApi = useApi(toastApiRef);

// Full-featured toast
toastApi.post({
  title: 'Entity saved',
  description: 'Your changes have been saved successfully.',
  status: 'success',
  timeout: 5000,
  links: [{ label: 'View entity', href: '/catalog/entity' }],
});

// Simple toast
toastApi.post({ title: 'Processing...', status: 'info' });

// Programmatic dismiss
const key = toastApi.post({ title: 'Uploading...', status: 'info' });
// Later...
toastApi.close(key);
```

https://github.com/user-attachments/assets/d5bacf01-cb92-4ced-96cb-73414b4ebccd

https://github.com/user-attachments/assets/e0e95e0c-23a5-4cbc-8325-340a0d07b10c

### DIFF
--- a/packages/frontend-plugin-api/report.api.md
+++ b/packages/frontend-plugin-api/report.api.md
@@ -31,7 +31,7 @@ export type AlertApi = {
   alert$(): Observable<AlertMessage>;
 };
 
-// @public
+// @public @deprecated
 export const alertApiRef: ApiRef_2<AlertApi, 'core.alert'> & {
   readonly $$type: '@backstage/ApiRef';
 };


plugins/catalog-backend-module-gitea/package.json
Rugvip, line 1: best skip this change

packages/ui/report.api.md
```diff
@@ -2007,6 +2008,52 @@ export type TextVariants =
...
+}
+
+// @public
+export interface ToastLink {
+  href: string;
+  label: string;
+}
+
+// @public
+export const toastQueue: ToastQueue<ToastContent>;
```
Rugvip: Let's rely on the existing alert API for this, potentially extending it if needed. Or if we need to create a new toast API and hook things up via that that's fine too. See the API implementation in https://github.com/backstage/backstage/blob/master/packages/core-app-api/src/apis/implementations/AlertApi/AlertApiForwarder.ts, and the way that's picked up by the AlertDisplay and renderer in the current Snackbar: https://github.com/backstage/backstage/blob/master/packages/core-components/src/components/AlertDisplay/AlertDisplay.tsx

packages/ui/src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,309 @@
...
+  Ref,
+  isValidElement,
+  ReactElement,
+  useRef,
+  useLayoutEffect,
+  useState,
+} from 'react';
+import { useToast } from '@react-aria/toast';
+import { useButton } from 'react-aria';
+import { motion } from 'motion/react';
```
Rugvip, line 20: Fairly big library right? Might be a shame to have a lot of copies of it, given that works case every plugin has its own copy of BUI. Not sure if tree shaking is gonna take care of it tbh. This is a bit of a global app concern, so it may be worth exploring shipping this via the `app` plugin only, the entire Toast that is.

plugins/catalog-backend-module-gitea/package.json
cdedreuille: @Rugvip Ok this is now fixed but probably something to fix on `master` as Prettier will fix it on pre-commit.

packages/ui/src/hooks/useInvertedThemeMode.ts
```diff
@@ -0,0 +1,66 @@
...
+
+type ThemeMode = 'light' | 'dark';
+
+/**
+ * Detects the current theme mode from the DOM and returns the inverted value.
+ * Checks both document.documentElement and document.body for the data-theme-mode attribute.
+ *
+ * @returns The inverted theme mode ('light' when app is dark, 'dark' when app is light)
+ */
+export function useInvertedThemeMode(): ThemeMode {
```
Rugvip, line 32: Hmm, this is a bit of a tricky one. It's not particularly safe or at least flexible to assume that these are the elements that contain these attributes.

I wouldn't even say that it's safe to assume that an inverted theme even exists?

Basically, no other way to handle this? 😅

cdedreuille:
After discussing this in more details with @Rugvip, we would like to explore a new approach.

The new Toast notification system we've built has different capabilities than the existing `AlertApi`:

| Feature | AlertApi | ToastApi (new) |
|---------|----------|----------------|
| Content | `message: string` | `title` + `description` |
| Timeout control | Consumer sets global timeout via `transientTimeoutMs` | Plugin sets per-toast `timeout` |
| Links | Not supported | Supported |
| Icons | Not supported | Supported |
| Programmatic dismiss | Not supported | Supported via `close(key)` |

Rather than trying to bridge the `AlertApi` to our new Toast system (which loses features and creates awkward mappings), we would like to explore building a new `ToastApi` that exposes the full capabilities of the Toast component.

**Usage:**

```typescript
import { toastApiRef, useApi } from '@backstage/frontend-plugin-api';

const toastApi = useApi(toastApiRef);

// Full-featured toast
toastApi.post({
  title: 'Entity saved',
  description: 'Your changes have been saved successfully.',
  status: 'success',
  timeout: 5000,
  links: [{ label: 'View entity', href: '/catalog/entity' }],
});

// Programmatic dismiss
const key = toastApi.post({ title: 'Uploading...', status: 'info' });
// Later...
toastApi.close(key);
```

**Migration path:** The `ToastDisplay` component would subscribes to both APIs, allowing existing code using `alertApi.post()` to continue working during the transition period. This provides a smooth migration path where both systems work side by side until `AlertApi` is fully deprecated. Plugins can migrate to the new `ToastApi` at their own pace to take advantage of the richer notification features.

.changeset/deprecate-alert-api.md
```diff
@@ -0,0 +1,56 @@
+---
+'@backstage/frontend-plugin-api': patch
+'@backstage/core-plugin-api': patch
+'@backstage/core-app-api': patch
```
Rugvip: Let's only add it in NFS, don't want to have to deal with support in the reverse direction 😅. i.e. skip the changes in `core-app-api` and `core-plugin-api`

.changeset/toast-api-introduction.md
```diff
@@ -0,0 +1,38 @@
+---
+'@backstage/frontend-plugin-api': minor
+'@backstage/plugin-app': minor
```
Rugvip:
```suggestion
'@backstage/frontend-plugin-api': patch
'@backstage/plugin-app': patch
```

No breaking changes right?

packages/frontend-plugin-api/src/apis/definitions/ToastApi.ts
```diff
@@ -0,0 +1,127 @@
...
+  /** URL the link points to */
+  href: string;
+};
+
+/**
+ * Message handled by the {@link ToastApi}.
+ *
+ * @public
+ */
+export type ToastMessage = {
```
Rugvip: Could you make this one `ToastApiMessage`? We try to keep some structural namespacing on types when possible

packages/frontend-plugin-api/src/apis/definitions/ToastApi.ts
```diff
@@ -0,0 +1,127 @@
...
+import { createApiRef, ApiRef } from '../system';
+import { Observable } from '@backstage/types';
+import { ReactNode } from 'react';
+
+/**
+ * Link item for toast notifications.
+ *
+ * @public
+ */
+export type ToastLink = {
```
Rugvip: Same thing here `ToastApiMessageLink`

packages/frontend-plugin-api/src/apis/definitions/ToastApi.ts
```diff
@@ -0,0 +1,127 @@
...
+ * @public
+ */
+export type ToastApi = {
+  /**
+   * Post a toast notification for display to the user.
+   *
+   * @param toast - The toast message to display
+   * @returns A unique key that can be used to programmatically dismiss the toast
+   */
+  post(toast: ToastMessage): string;
```
Rugvip: Could you make this one return an object instead? want to keep the API evolvable, so at the very least `: { key: string }`

I do wonder if we should make it `: { close(): void }` instead though? I.e. a method you can use to close the toast. It limits the ability to close toasts a bit better to the one that opened it, and it should be relatively simple to hold onto that reference. With that we could also remove the `close(key: string)` and `close$` methods

plugins/app/src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,260 @@
...
+        onAnimationComplete={definition => {
+          // Clean up the manual close tracking after exit animation
+          if (definition === 'exit') {
+            manuallyClosingToasts.delete(toast.key);
+          }
+        }}
+        transition={{ type: 'spring', stiffness: 400, damping: 35 }}
+        data-status={finalStatus}
+      >
+        <div className="toast-wrapper">
```
Rugvip: CSS module for these ye?

Speaking of that, we may still want global classnames for there? This is actually a good first example of making something outside of the core BUI library themeable? Very happy to skip that for the initial release though

plugins/app/src/components/Toast/ToastQueue.ts
```diff
@@ -0,0 +1,47 @@
...
+ * const key = toastQueue.add({ title: 'Processing...' });
+ * // Later...
+ * toastQueue.close(key);
+ * ```
+ *
+ * @internal
+ */
+export const toastQueue = new ToastQueue<ToastContent>({
+  maxVisibleToasts: 4,
+});
```
Rugvip: Could you avoid having this as a global? create it in the display instead

plugins/app/src/components/Toast/types.ts
```diff
@@ -0,0 +1,102 @@
...
+export interface ToastLink {
+  /** Display text for the link */
+  label: string;
+  /** URL the link points to */
+  href: string;
+}
+
+/**
+ * Content for a toast notification
+ * @public
```
Rugvip: No need for these to be public right?

plugins/app/src/hooks/useInvertedThemeMode.ts
```diff
@@ -0,0 +1,70 @@
...
+type ThemeMode = 'light' | 'dark';
+
+/**
+ * Detects the current theme mode from the DOM and returns the inverted value.
+ * Checks both document.documentElement and document.body for the data-theme-mode attribute.
+ *
+ * @returns The inverted theme mode ('light' when app is dark, 'dark' when app is light)
+ * @internal
+ */
+export function useInvertedThemeMode(): ThemeMode {
```
Rugvip, line 32: Hmm, again, this assumes that there's even an opposing light/dark theme. Not sure all apps will even have that? Wondering what other potential options there are here

plugins/app/package.json
```diff
@@ -20,7 +20,9 @@
     "directory": "plugins/app"
   },
   "license": "Apache-2.0",
-  "sideEffects": false,
+  "sideEffects": [
+    "*.css"
```
Rugvip: Nope! not having people import this or installing global CSS :p

Rugvip (changes requested):

plugins/app/package.json
cdedreuille: Ah yes. I brought it back.

plugins/app/src/components/Toast/types.ts
cdedreuille: True. I made it internal.

plugins/app/src/components/Toast/Toast.tsx
cdedreuille: Yes I'm not sure what would be the right naming for that then. Should we keep `.bui` prefix?

packages/frontend-plugin-api/src/apis/definitions/ToastApi.ts
cdedreuille: Yes I renamed it.

packages/frontend-plugin-api/src/apis/definitions/ToastApi.ts
cdedreuille: Yes I renamed it

plugins/app/src/hooks/useInvertedThemeMode.ts
cdedreuille: I reformatted it using the `appThemeApi` instead. Let's see how it goes.

packages/frontend-plugin-api/src/apis/definitions/ToastApi.ts
cdedreuille: Yes, I cleaned it up. Now the API will work as followed:
```
const toast = toastApi.post({ title: 'Uploading...', status: 'info' });
toast.close();
```

.changeset/toast-api-introduction.md
cdedreuille: Done

.changeset/deprecate-alert-api.md
cdedreuille: Done

cdedreuille:
@Rugvip Yes agree with that. Let's ship it. You might have to generate the API report though

Issue #31127, opened 2025-09-12
conditional execution based on step output for "each" keyword items


OSobky:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
each, scaffolder, software templates, template
```

### 🗃️ Project Area

Software Templates

### 🔖 Need

As I am not sure if this is a bug or a change, I am opening this as a change request.

The current behavior in a step which have an if condition based on **"each keyword"** output array is the if condition only evaluate the first item in the output array

for example:
```yaml
- action: debug:log
  id: write-debug-line-if
  each: ${{ parameters.repoUrl }} # Optional iterable, run the same step multiple times
  if: ${{ steps['CheckCiFile'].output.code[each.key] | string === "200" }}
  name: Write "Hello Backstage!" log line if 200
  input:
    message:  "Hey ${{ steps['CheckCiFile'].output.code[each.key] }}"
```

### 📝 Proposal

Only run the task if the condition is true for this specific item otherwise skipped and add a null value in the corresponding step output array. 

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Yes, but I would like some more guidance

camilaibs:
@backstage/scaffolder-maintainers any thoughts 🙂 ?

raihle:
I also hit this. I want to iterate over repositories and create PRs only if we made changes. But it seems `if` is being evaluated outside the context of `each`, because the step is skipped even if every item in the output is `true`.

```yaml
    - id: create-pr
      name: Create Pull Request
      action: publish:github:pull-request
      each: ${{ steps['xxx'].output.changed }}
      if: ${{ each.value }}
      input:
        # ...
```

That's not wrong per se - it could definitely be useful to skip the step entirely and not just on a per-iteration basis. But I think it's _also_ definitely useful to be able to skip specific iterations.

Changing the behavior so that `if` is evaluated "after" `each` probably wouldn't break anything. If you're not using the `each` context, then it will still evaluate to the same value on every iteration. But I'm not familiar enough with the scaffolder (and its use in the wild) to say for certain.

benjdlambert:
@raihle I think that this is an oversight that the `each` context is not available in the `if` parameter.

[...]

PR #32546, opened 2026-01-28
Scaffolder: Provide 'each' context when evaluating 'if'

raihle:

I took a stab at https://github.com/backstage/backstage/issues/31127

This makes the `each` context available when `if` conditions for steps are evaluated during template execution, allowing _some_ iterations to run while others are skipped.

Previously the `each` context was not available and the condition was only evaluated once, so it was not possible to conditionally skip specific iterations.

- [x] ~~Added or updated documentation~~ (not needed, I think)
- [x] ~~Screenshots attached (for UI changes)~~ (no UI changes)

### DIFF
--- a/plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
+++ b/plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
@@ -341,28 +341,7 @@ export class NunjucksWorkflowRunner implements WorkflowRunner {
         }
       }
 
-      const resolvedEach =
-        step.each &&
-        this.render(
-          step.each,
-          {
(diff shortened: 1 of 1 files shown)


raihle:
> @raihle 👋 thanks for this, I think there's a little bit of cleanup we can do part of this PR. I do have a branch locally but I can't push to this PR.
> 
> Can you have a look at the commit I pushed here to bring these changes into this PR?
> 
> [a531000](https://github.com/backstage/backstage/commit/a5310000f4ec6c7c349ec34ed855fb5cba301f56)
> 
> It's just simplifying the logic a little bit to remove duplicates and cleaning up the `if` context so that it's the same for all of the templating stages.

Nice, I found one more duplicate too :)

PR #32548, opened 2026-01-28
.patches: ignore already applied patches, remove broken cleanup workflow

Rugvip:

Updating the patch process a bit. I realized the cleanup workflow doesn't work due to lack of permissions, and tbh that's for the better. This instead changes the patch process to ignore already applied patches, and removes the automatic workflow. It means an occasional cleanup is useful but prioritizes getting new patches out more cleanly.

### DIFF
--- a/scripts/patch-release-for-pr.js
+++ b/scripts/patch-release-for-pr.js
@@ -278,13 +278,16 @@ async function main(args) {
 
     const commitMessage = `Patch from PR #${prNumber}`;
 
+    const commitMessagePattern = `^${commitMessage}$`;
+
     // Check if this patch has already been applied by looking for the commit message
     try {
       const existingCommit = await run(
         'git',
         'log',
+        '--extended-regexp',
         '--grep',
-        commitMessage,
+        commitMessagePattern,
         '--format=%H',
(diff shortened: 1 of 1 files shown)


scripts/patch-release-for-pr.js
```diff
@@ -271,19 +275,75 @@ async function main(args) {
       '--reverse',
       '--pretty=%H',
     );
-    for (const logSha of logLines.split(/\r?\n/)) {
-      await run('git', 'cherry-pick', '-n', logSha);
+
+    const commitMessage = `Patch from PR #${prNumber}`;
+
+    const commitMessagePattern = `^${commitMessage}$`;
```
freben, line 281: In the future maybe we could inject some form of `<!-- applies-backstage-patch-pr 123456 -->` marker or something, to get stricter matching.

freben (approved):
Cool, let's try it!

scripts/patch-release-for-pr.js
Rugvip: yeah idd, @copilot could you open a draft PR that implements that idea?

Issue #31813, opened 2025-11-18
Redis connection issues


Alexandru-Pita:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
cache
redis
connection
time out
reconnect
redis connection
redis time out
```

### 🗃️ Project Area

Unknown

### 🔗 External Integration

N/A

### 📝 Description & Context

Hello!

Me and my team have encountered some issues regarding cache service implementation with Redis.

This is our current `app-config.yaml` file:
```yaml
backend:
  cache:
    store: redis
    connection: rediss://:${REDIS_PASSWORD}@${REDIS_HOST}:${REDIS_PORT}
    redis:
      client:
        namespace: ${REDIS_NAMESPACE}
        useUnlink: true
```

**Our current versions**:
- `backstage.json`: `1.40.2`
- `backend-defaults`: `0.11.1`
- `@keyv/redis`: `4.6.0`
- `keyv`: `5.5.0`

**Current Behavior**:
We have our app deployed in Azure using Kubernetes. Redis is also managed using Azure.

Right now when we deploy our app, for a certain amount of time the connection to Redis is kept alive and we can `set / get` data from Redis. After that time, the connection to Redis is lost and any kind of request to Redis will hang with no errors at all.

We observed these issues first with `TechDocs`. We realized something was wrong when we kept constantly seeing this specific log `Cache miss: ...`, which can be found [here](https://github.com/backstage/backstage/blob/master/plugins/techdocs-backend/src/cache/TechDocsCache.ts#L77).

The next thing we did was to create a back end plugin for testing purposes with some endpoints to `set / get` data to / from Redis. This is how we got even more information about this strange Redis behavior. After some time when the Redis connection was lost, we kept hitting a `504 Gateway Time-out` status code from `Microsoft-Azure-Application-Gateway` through these test endpoints. In our case the `Application Gateway` load balancer has some configuration that doesn't let requests take more than a specified time.

One last thing we did was to implement a `keep-alive` logic in the same plugin we used for testing purposes. Doing this, actually solved our issues, but this feels like a "hackish" solution.

### 👍 Expected Behavior

We expect that:
- The whole `keep-alive` mechanism to be done by default by `@keyv/redis`
- When connection is lost, a log would also be good (right now we did a lot of investigation in order to find these issues)

### Have you read the Code of Conduct?

[...]

Issue #31742, opened 2025-11-13
redis cannot to be reconnection when connection is broken


zhaoqiang1980:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
redis reconnection
redis reconnect
cache redis connection
redis disconnected
```

### 🗃️ Project Area

Unknown

### 🔗 External Integration

N/A

### 📝 Description & Context

I am using Redis for caching. If I stop the Redis server, any page that uses the cache becomes unresponsive for a long time. When I restart Redis, the connection does not automatically recover. Can you help me understand why this happens?  

-- backstage version: 1.42
-- redis: 5.0.14.1

app-config:
backend:      
   cache:
      store: redis
      connection: 'redis://localhost:6379/1'
       defaultTtl: 3600000

### 👍 Expected Behavior

Redis is only used for caching, so the application should not hang when the connection is broken. When Redis becomes available again, it should be able to reconnect automatically.  

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Undecided

aramissennyeydd:
👋 Can you try bumping to the latest version of Backstage and seeing if the behavior remains? If that doesn't help, I'd recommend looking through the `@keyv/redis` docs for relevant Redis config options, that's the library we use for Redis support, https://github.com/backstage/backstage/blob/master/packages/backend-defaults/src/entrypoints/cache/CacheManager.ts#L278-L279

matthewadowns:
There is `defaultReconnectStrategy` https://keyv.org/docs/storage-adapters/redis/#keyv-redis-options-1 with exponential backoff retry, but it's possible the reconnect logic isnt getting hit if the socket is timing out -- stale sockets can hang without error.   One option could be to add a config option for a custom `reconnectConfig`. 

When Redis goes away, the cache client can hang because the underlying socket doesn’t always detect the disconnect. The default reconnect logic only triggers after a socket closes or times out, so if the connection is “half‑open” (idle drop), reconnect never kicks in.

Suggested mitigation:
Add keepalive + ping + socket timeout so the client actively keeps the connection alive and detects broken sockets. This prevents long hangs and allows reconnects to happen.

[...]

PR #32559, opened 2026-01-28
fix: add redis socket keepalive options

matthewadowns:

Adds Redis socket keepalive and reconnect configuration to the backend cache (`backend.cache.redis.client.socket`) and documents the new options. Also adds unit coverage for keepalive/reconnect config parsing and cluster defaults.

This is intended to address https://github.com/backstage/backstage/issues/31813 (and likely https://github.com/backstage/backstage/issues/31742) by exposing node‑redis socket controls (`keepAlive`, `keepAliveInitialDelay`, `pingInterval`, `socketTimeout`) and optional reconnect strategy configuration, mirroring the node‑redis client configuration options  

(See https://github.com/redis/node-redis/blob/master/docs/client-configuration.md)

- [x] A changeset describing the change and affected packages. (/.changeset/redis-keepalive-options.md)
- [x] Added or updated documentation (config schema `packages/backend-defaults/config.d.ts`)
- [ ] Screenshots attached (for UI changes) — __N/A - config example given instead__

#### Tests run

- `yarn test --no-watch packages/backend-defaults/src/entrypoints/cache/CacheManager.test.ts`

#### Example Config Options Usage 

```yaml
backend:
  cache:
    store: redis
    connection: redis://localhost:6379/1
    defaultTtl: 3600000
    redis:
      client:
        namespace: my-backstage
        # Socket options passed through to node-redis:
        socket:
          keepAlive: true
          keepAliveInitialDelay: 10000
          socketTimeout: 20000
          reconnectStrategy:
            baseDelayMs: 100
            maxDelayMs: 2000
            jitterMs: 50
            maxRetries: 20
            stopOnSocketTimeout: true
        pingInterval: 15000
```

## Why This Is Needed

- Backstage cache uses Keyv → node‑redis, but didn’t expose socket controls or reconnect tuning for long‑lived connections.
- In real infra (LB/NAT/managed Redis), idle TCP connections are routinely dropped; without keepalive/ping/timeout or reconnect strategy controls, sockets can go stale and recovery becomes unreliable.
- node‑redis documents these socket options and defaults, but Backstage wasn’t mapping them cleanly.

## What This Changes

- Exposes supported node‑redis socket options (keepalive, delay, ping interval, socket timeout).
- Adds optional reconnect strategy configuration (backoff/jitter, max retries, stop‑on‑timeout).
  - In validating this change in our own internal instance, we found that it was necessary to override the default `keyv` reconnect.  Specifically, this forces stale connections to close to trigger a reconnect (instead of dangling in an unusable state)
- Ensures keepalive is explicitly enabled when a delay is set.
- Aligns config typing with node‑redis: `keepAlive` is boolean; timeouts/intervals accept numbers or human durations.

## Result

- Prevents silent stale sockets.
- Allows node‑redis to detect dead connections and reconnect reliably.
- Lets operators tune Redis connections to match infrastructure timeouts and error handling preferences.

## References
- https://github.com/redis/node-redis/blob/master/docs/client-configuration.md
- https://nodejs.org/api/net.html#socketsetkeepaliveenable-initialdelay

### DIFF
--- a/packages/backend-defaults/config.d.ts
+++ b/packages/backend-defaults/config.d.ts
@@ -766,16 +766,15 @@ export interface Config {
               socket?: {
                 /**
                  * Enables TCP keepalive. When `true`, uses the Redis client default
-                 * delay (5000ms in node-redis). When a number, it is interpreted as
-                 * the initial delay in milliseconds. If unset, Backstage does not
-                 * override the Redis client defaults.
-                 *
-                 * Do not set a numeric value together with `keepAliveInitialDelay`;
-                 * prefer `keepAlive: true` with `keepAliveInitialDelay`. If only
+                 * delay (5000ms in node-redis). If unset, Backstage does not
+                 * override the Redis client defaults. If only
                  * `keepAliveInitialDelay` is set, keepalive is enabled with that
                  * delay.
+                 *
+                 * See https://github.com/redis/node-redis/blob/master/docs/client-configuration.md
+                 * and https://nodejs.org/api/net.html#socketsetkeepaliveenable-initialdelay
                  */
-                keepAlive?: boolean | number;
+                keepAlive?: boolean;
                 /**
                  * Initial delay in milliseconds for TCP keepalive probes.
                  */
--- a/packages/backend-defaults/config.d.ts
+++ b/packages/backend-defaults/config.d.ts
@@ -780,62 +780,62 @@ export interface Config {
                  */
                 keepAliveInitialDelay?: number;
                 /**
-                 * Send `PING` command at interval (in milliseconds). Useful for
-                 * environments with idle connection timeouts.
-                 */
-                pingInterval?: number;
-                /**
                  * The maximum duration (in milliseconds) that the socket can remain
                  * idle before being automatically closed.
                  */
                 socketTimeout?: number;
+              };
+              /**
+               * Send `PING` command at interval (in milliseconds). Useful for
+               * environments with idle connection timeouts.
+               */
+              pingInterval?: number;
+              /**
+               * Optional reconnect strategy configuration. When set, Backstage
+               * creates a reconnect strategy function and passes it to the Redis
+               * client. The strategy is used when node-redis detects a socket
+               * close or error (e.g. the connection was dropped by a load
+               * balancer or NAT).
+               *
+               * Note: half-open connections are not always detected by TCP. If
+               * the socket stays idle and the network silently drops it, the
+               * reconnect strategy will not run until the client observes an
+               * error. Use periodic traffic (pinger) or infrastructure/OS
+               * keepalive settings to avoid long idle periods.
+               *
+               * The strategy implements exponential backoff with jitter:
+               * delay = min(2^retries * baseDelayMs, maxDelayMs) ± jitterMs.
+               *
+               * See
+               * https://keyv.org/docs/storage-adapters/redis/#gracefully-handling-errors-and-timeouts.
+               */
+              reconnectStrategy?: {
                 /**
-                 * Optional reconnect strategy configuration. When set, Backstage
-                 * creates a reconnect strategy function and passes it to the Redis
-                 * client. The strategy is used when node-redis detects a socket
-                 * close or error (e.g. the connection was dropped by a load
-                 * balancer or NAT).
-                 *
-                 * Note: half-open connections are not always detected by TCP. If
-                 * the socket stays idle and the network silently drops it, the
-                 * reconnect strategy will not run until the client observes an
-                 * error. Use periodic traffic (pinger) or infrastructure/OS
-                 * keepalive settings to avoid long idle periods.
-                 *
-                 * The strategy implements exponential backoff with jitter:
-                 * delay = min(2^retries * baseDelayMs, maxDelayMs) ± jitterMs.
-                 *
-                 * See
-                 * https://keyv.org/docs/storage-adapters/redis/#gracefully-handling-errors-and-timeouts.
+                 * Base delay in milliseconds for exponential backoff. Defaults to
+                 * 100ms when reconnectStrategy is configured.
                  */
-                reconnectStrategy?: {
-                  /**
-                   * Base delay in milliseconds for exponential backoff. Defaults to
-                   * 100ms when reconnectStrategy is configured.
-                   */
-                  baseDelayMs?: number;
-                  /**
-                   * Max delay in milliseconds for exponential backoff. Defaults to
-                   * 2000ms when reconnectStrategy is configured.
-                   */
-                  maxDelayMs?: number;
-                  /**
-                   * Random jitter in milliseconds added to each delay (± jitter).
-                   * Defaults to 50ms when reconnectStrategy is configured.
-                   */
-                  jitterMs?: number;
-                  /**
-                   * Maximum number of retries before giving up. When unset, retries
-                   * are unbounded.
-                   */
-                  maxRetries?: number;
-                  /**
-                   * When true, stop reconnecting on socket timeout errors. This is
-                   * useful when you want timeouts to be treated as fatal and rely
-                   * on higher-level retry logic.
-                   */
-                  stopOnSocketTimeout?: boolean;
-                };
+                baseDelayMs?: number;
+                /**
+                 * Max delay in milliseconds for exponential backoff. Defaults to
+                 * 2000ms when reconnectStrategy is configured.
+                 */
+                maxDelayMs?: number;
+                /**
+                 * Random jitter in milliseconds added to each delay (± jitter).
+                 * Defaults to 50ms when reconnectStrategy is configured.
+                 */
+                jitterMs?: number;
+                /**
+                 * Maximum number of retries before giving up. When unset, retries
+                 * are unbounded.
+                 */
+                maxRetries?: number;
+                /**
+                 * When true, stop reconnecting on socket timeout errors. This is
+                 * useful when you want timeouts to be treated as fatal and rely
+                 * on higher-level retry logic.
+                 */
+                stopOnSocketTimeout?: boolean;
               };
             };
             /**
--- a/packages/backend-defaults/config.d.ts
+++ b/packages/backend-defaults/config.d.ts
@@ -792,7 +792,20 @@ export interface Config {
                 /**
                  * Optional reconnect strategy configuration. When set, Backstage
                  * creates a reconnect strategy function and passes it to the Redis
-                 * client. See
+                 * client. The strategy is used when node-redis detects a socket
+                 * close or error (e.g. the connection was dropped by a load
+                 * balancer or NAT).
+                 *
+                 * Note: half-open connections are not always detected by TCP. If
+                 * the socket stays idle and the network silently drops it, the
+                 * reconnect strategy will not run until the client observes an
+                 * error. Use periodic traffic (pinger) or infrastructure/OS
+                 * keepalive settings to avoid long idle periods.
+                 *
+                 * The strategy implements exponential backoff with jitter:
+                 * delay = min(2^retries * baseDelayMs, maxDelayMs) ± jitterMs.
+                 *
+                 * See
                  * https://keyv.org/docs/storage-adapters/redis/#gracefully-handling-errors-and-timeouts.
                  */
                 reconnectStrategy?: {
@@ -817,7 +830,9 @@ export interface Config {
                    */
                   maxRetries?: number;
                   /**
-                   * When true, stop reconnecting on socket timeout errors.
+                   * When true, stop reconnecting on socket timeout errors. This is
+                   * useful when you want timeouts to be treated as fatal and rely
+                   * on higher-level retry logic.
                    */
                   stopOnSocketTimeout?: boolean;
                 };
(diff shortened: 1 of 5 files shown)


packages/backend-defaults/config.d.ts
```diff
@@ -758,6 +758,38 @@ export interface Config {
...
+                 * delay.
+                 *
+                 * See https://github.com/redis/node-redis/blob/master/docs/client-configuration.md
+                 * and https://nodejs.org/api/net.html#socketsetkeepaliveenable-initialdelay
+                 */
+                keepAlive?: boolean;
+                /**
+                 * Initial delay in milliseconds for TCP keepalive probes.
+                 */
+                keepAliveInitialDelay?: number;
```
freben: Can we make all of these three into `number | HumanDuration | string` just like `defaultTtl` above? This lets people say for example `pingInterval: '10s'` and there's no confusion on what units are being used.

Note how it is parsed [here](https://github.com/backstage/backstage/blob/23916a0b5c097b5640288e92acb6a10a511b42ab/packages/backend-defaults/src/entrypoints/cache/CacheManager.ts#L99). You can break that out into a helper function to reuse for reading these.

packages/backend-defaults/src/entrypoints/cache/types.ts
```diff
@@ -28,6 +32,23 @@ export type RedisCacheStoreOptions = {
...
   client?: KeyvRedisOptions;
   cluster?: RedisClusterOptions;
+  socket?: RedisSocketOptions;
+};
+
+type RedisSocketOptions = Omit<
+  NonNullable<RedisClientOptions['socket']>,
+  'keepAlive'
+> & {
+  keepAlive?: boolean;
```
freben: Why did this need an override?

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -183,6 +183,69 @@ export class CacheManager {
       ),
     };
 
+    const clientConfig = redisConfig.getOptionalConfig('client');
+    const socketConfig = clientConfig?.getOptionalConfig('socket');
+    const keepAliveConfig = socketConfig?.getOptional('keepAlive');
+    const keepAlive =
+      typeof keepAliveConfig === 'boolean' ? keepAliveConfig : undefined;
```
freben: this can be simplified to just `const keepAlive = redisConfig.getOptionalBoolean('client.socket.keepAlive')`; it can't be anything else than boolean or undefined then.

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -183,6 +183,69 @@ export class CacheManager {
...
+    const keepAliveInitialDelay = socketConfig?.getOptionalNumber(
+      'keepAliveInitialDelay',
+    );
```
freben: see comment elsewhere about supporting human formats on these three

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -183,6 +183,69 @@ export class CacheManager {
...
+    const pingInterval = socketConfig?.getOptionalNumber('pingInterval');
+    const socketTimeout = socketConfig?.getOptionalNumber('socketTimeout');
+
+    if (typeof keepAliveConfig === 'number') {
```
freben: No need. If you use `getOptionalBoolean` (which matches the config schema too!) then this can never happen

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -183,6 +183,69 @@ export class CacheManager {
...
+      logger?.warn(
+        "Invalid 'client.socket.keepAlive' value. Expected boolean, got number. Ignoring keepAlive.",
+      );
+    }
+
+    if (keepAlive === false && keepAliveInitialDelay !== undefined) {
+      logger?.warn(
```
freben: Optional to check in my mind. If you have consciously set it to false then you're fine with fine-tuning parameters being ignored too.

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -183,6 +183,69 @@ export class CacheManager {
...
+            ...(keepAliveForSocket !== undefined
+              ? { keepAlive: keepAliveForSocket }
+              : {}),
+            ...(keepAliveInitialDelayForSocket !== undefined
+              ? { keepAliveInitialDelay: keepAliveInitialDelayForSocket }
+              : {}),
+            ...(pingInterval !== undefined ? { pingInterval } : {}),
+            ...(socketTimeout !== undefined ? { socketTimeout } : {}),
+          }
+        : undefined;
```
freben: This logic seems complex to me. Could we have only `let socketOptions: RedisSocketOptions | undefined` at the top, and then you make it an empty object and assign things into it inside the initial `if` check above? Like,

```ts
if (keepAlive || keepAliveInitialDelay !== undefined) {
  socketOptions = {};
  // ...
  if (...) { socketOptions.socketTimeout = ...
}
```

Just to make it a bit easier to follow

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -204,6 +267,17 @@ export class CacheManager {
           'maxCommandRedirections',
         ),
       };
+
+      if (redisOptions.socket) {
+        redisOptions.cluster.defaults = {
+          ...(redisOptions.cluster.defaults ?? {}),
+          socket: {
+            ...((redisOptions.cluster.defaults as { socket?: object })
```
freben: Do those types not have the socket field in it?

packages/backend-defaults/config.d.ts
matthewadowns: Thanks for the tip -- I will update to make this more user friendly and align with other such config options.

packages/backend-defaults/src/entrypoints/cache/types.ts
matthewadowns: Great catch -- fixing this.  Thanks!

matthewadowns: removed -- `keepAlive` is only a boolean

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
matthewadowns: Updated, thanks!

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
matthewadowns: updated, thanks

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
matthewadowns: Agreed --removed.  Thanks!

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
matthewadowns: updated, thank you!

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -49,6 +54,20 @@ type StoreFactory = (pluginId: string, defaultTtl: number | undefined) => Keyv;
...
+  ): number | undefined {
+    const value = config.getOptional(key);
+    if (value === undefined) {
+      return undefined;
+    }
+    if (typeof value === 'number') {
+      return value;
+    }
+    return durationToMilliseconds(readDurationFromConfig(config, { key }));
+  }
```
matthewadowns, line 69: Helper method added per @freben feedback.  This could be factored out into a shared method, probably in `@backstage/config/` or `@backstage/backend-plugin-api`, but this refactor would be bigger change that should be done separately.

matthewadowns:
@freben Thanks again for the review!   I've addressed all comments and also updated this to include the reconnect configuration options.  We're running this in our own instance now, and have verified it fixes the issues mentioned.

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -165,7 +186,7 @@ export class CacheManager {
     config: RootConfigService,
     logger?: LoggerService,
   ): RedisCacheStoreOptions {
-    const redisOptions: RedisCacheStoreOptions = {
+    const redisOptions: RedisCacheStoreOptions & { socket?: object } = {
```
freben, line 189: What does this extra socket type change do?

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -183,6 +204,117 @@ export class CacheManager {
...
 
+    const clientConfig = redisConfig.getOptionalConfig('client');
+    const socketConfig = clientConfig?.getOptionalConfig('socket');
+    const keepAlive = redisConfig.getOptionalBoolean('client.socket.keepAlive');
+    const keepAliveInitialDelay = socketConfig
+      ? CacheManager.readOptionalDuration(socketConfig, 'keepAliveInitialDelay')
+      : undefined;
+    const pingInterval = clientConfig
+      ? CacheManager.readOptionalDuration(clientConfig, 'pingInterval')
+      : undefined;
```
freben, line 215: the `getOptional` family of calls work "recursively" - it's safe if _any_ part of the chain is missing. So all of this can be simplified to like

```ts
const keepAliveInitialDelay = CacheManager.readOptionalDuration(redisConfig, 'client.socket.keepAliveInitialDelay')
const pingInterval = CacheManager.readOptionalDuration(redisConfig, 'client.pingInterval')
```

the extra optionality stuff in here makes it harder to read, and I think socketConfig is essentially not needed.

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -183,6 +204,117 @@ export class CacheManager {
...
+    }
+
+    logger?.debug('backend.cache: derived socket options', {
+      keepAliveForSocket,
+      keepAliveInitialDelayForSocket,
+    });
+
+    const reconnectConfig =
+      socketConfig?.getOptionalConfig('reconnectStrategy');
+    const baseDelayMs = reconnectConfig?.getOptionalNumber('baseDelayMs');
```
freben, line 246: These could also have gotten the readOptionalDuration treatment! `baseDelay?: ...`

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -183,6 +204,117 @@ export class CacheManager {
...
+      'stopOnSocketTimeout',
+    );
+    const hasReconnectConfig =
+      baseDelayMs !== undefined ||
+      maxDelayMs !== undefined ||
+      jitterMs !== undefined ||
+      maxRetries !== undefined ||
+      stopOnSocketTimeout !== undefined;
+
+    const reconnectStrategy = hasReconnectConfig
```
freben, line 260: All of this significant config stuff might actually be best shunted into like a new `entrypoints/cache/providers/redis/readReconnectStrategy.ts` or so, with a function that takes a config and returns the strategy.

Then that folder could have a `readOptionalDuration.ts`, a `readSocketOptions.ts`, ... much easier to read and reason about, plus they have the luxury of being able to make an early return, for example

```ts
if (keepAlive === true) {
  return DEFAULT_KEEPALIVE;
```

and stuff like that.

This is all a bit hard to follow.

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -337,23 +504,56 @@ export class CacheManager {
...
+              ? {
+                  url: this.connection,
+                  ...(storeOptions.socket
+                    ? { socket: storeOptions.socket }
+                    : {}),
+                  ...(storeOptions.pingInterval !== undefined
+                    ? { pingInterval: storeOptions.pingInterval }
+                    : {}),
+                }
+              : this.connection;
```
freben, line 529: We have this slightly awkward deep-merging thing in two places. Could that be cleaned up? Broken out into one place perhaps too?

packages/backend-defaults/src/entrypoints/cache/CacheManager.ts
```diff
@@ -337,23 +504,56 @@ export class CacheManager {
...
+          stores[pluginId] = new KeyvRedis(connection, redisOptions);
         }
 
         // Always provide an error handler to avoid stopping the process
         stores[pluginId].on('error', (err: Error) => {
           this.logger?.error('Failed to create redis cache client', err);
           this.errorHandler?.(err);
+
+          // Disconnects the client to force a new connection on the next request after errors.
+          const store = stores[pluginId];
```
freben, line 539: This seems to be a major deviation in behavior - does it have to be done as part of the same PR? Maybe it's a separate concern? I thought the keyv connector handled stuff like this internally, do we need to force close them really, is that a documented thing that one has to?

freben (changes requested):

mjk2007:
Is this being updated? We are using the Backstage version 1.48.5 with Redis cache and we're also facing this issue.

freben:
@matthewadowns are you still around to give the comments a look?

PR #32562, opened 2026-01-29
chore: replace legacy package.json workspaces.packages by workspaces

GabDug:

Small changes to remove legacy way of declaring workspace packages. This was introduced in [2020](https://github.com/backstage/backstage/commit/a3db7f35c2694e82000d3f3432cbdf446b6c5e6e).

This syntax used to allow ["nohoist" in Yarn v.1.4.2](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/), which was replaced by [Yarn v4 `nmHoistingLimits`](https://yarnpkg.com/migration/guide#replace-nohoist-by-nmhoistinglimits), and neither is used within the monorepo.

#### But why?

[Yarn v4 still reads this accepts `workspaces.packages`](https://github.com/yarnpkg/berry/blob/60b83ace11bd36270c1ed117319139eaa6971db3/packages/yarnpkg-core/sources/Manifest.ts#L390), but not [Yarn v6](https://github.com/yarnpkg/zpm). As it's a legacy, (now) poorly documented and unused syntax, I think we should remove it.

This should also prevent future curious contributors doing too much archeology on past workspaces behaviours, which I believe is a net positive.

#### Reference

- JSON Schema documents [both](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/package.json#L1011)
- Yarn v1 document it as an array: https://classic.yarnpkg.com/lang/en/docs/workspaces/
- Yarn v4 document it as an array: https://yarnpkg.com/configuration/manifest#workspaces
- Npm, array as well: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#workspaces 


packages/cli-common/src/paths.ts
```diff
@@ -107,6 +107,25 @@ export function findOwnRootDir(ownDir: string) {
   return resolvePath(ownDir, '../..');
 }
 
+/**
+ * Gets the workspaces pattern from a package.json object.
+ * @param pkgJson - The package.json object to get the workspaces pattern from.
+ * @returns The workspaces patterns (glob not resolved).
+ * @public
+ */
+export function getWorkspacesPatterns(pkgJson: any): string[] {
```
freben: `any` is too lax here - you could make it `unknown` or perhaps `JsonObject`. Either way, you'll find that it forces you to do an `object` type check in there to be safe.

GabDug: Updated to be type-safe, checking both object and the array content. A bit clunky as cli-common is dependency-light.
Also extracted to a dedicated file, it felt off in `paths.ts`

packages/cli-node/src/pacman/PackageManager.ts
```diff
@@ -101,7 +101,7 @@ export async function detectPackageManager(): Promise<PackageManager> {
     const packageJson = await fs.readJson(
       paths.resolveTargetRoot('package.json'),
     );
-    if (packageJson.workspaces) {
+    if (getWorkspacesPatterns(packageJson).length > 0) {
```
GabDug: Some logic change here, as it would have previously matched on `workspaces.packages: []`

Seems more correct but feedback is welcomed!

packages/cli-common/src/paths.ts
```diff
@@ -141,7 +142,7 @@ export function findPaths(searchDir: string): Paths {
           try {
             const content = fs.readFileSync(path, 'utf8');
             const data = JSON.parse(content);
-            return Boolean(data.workspaces);
+            return getWorkspacesPatterns(data).length > 0;
```
Rugvip: Honestly I feel that the existing check is enough and simple enough to be worth using over the common utility. I suggest we just use `Boolean(data.workspaces)` or equivalent for all the monorepo checks that are covered in this PR.

packages/cli-node/src/pacman/yarn/Yarn.ts
Rugvip: As far as I can tell this is unused and I think we can remove it instead or replace it with something that returns that packages rather than the patterns. That's already handled by [PackageGraph.listTargetPackages](https://github.com/backstage/backstage/blob/5e3ef57e4ee3fffc14b0448f9604506cdd0626b5/packages/cli-node/src/monorepo/PackageGraph.ts#L194) though so not that much value in having that here I think.

Rugvip (changes requested):
Thank you! 🎉 

I think we can clean up a bit more to be honest, don't think `getWorkspacesPatterns` is something we really need to add and instead just align the workspace checks on `Boolean(pkg.workspaces)`. Sure you could have `workspaces: []` but arguably that's still a workspace.

GabDug:
> Thank you! 🎉 
> 
> I think we can clean up a bit more to be honest, don't think `getWorkspacesPatterns` is something we really need to add and instead just align the workspace checks on `Boolean(pkg.workspaces)`. Sure you could have `workspaces: []` but arguably that's still a workspace.

Agreed it might be a little much, but not so sure about the empty workspaces being workspaces. I will check Yarn & other tools behaviors, we should try to match that.

Rugvip (changes requested):
It's not super interesting since `workspaces: []` is arguably invalid configuration. `@manypkg` which we use considers `[]` a workspace at least though: https://github.com/Thinkmill/manypkg/blob/327e4ac7673134f4d49b8414679849daa2e2eaef/packages/tools/src/YarnTool.ts#L31-L38

Rugvip (approved):
Let's :shipit:, thank you! 🎉

.changeset/slick-dogs-pay.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/create-app': patch
+'@backstage/cli-node': patch
+'@backstage/cli': patch
+---
+
+Properly support `package.json` `workspaces` field
```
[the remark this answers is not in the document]
Rugvip: That interface is not exported...

.changeset/slow-items-pull.md
```diff
@@ -0,0 +1,18 @@
+---
+'@backstage/create-app': patch
+---
+
+Replace `package.json` deprecated `workspaces.packages` by `workspaces`
```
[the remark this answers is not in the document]
Rugvip: Yep, this ☝️

Rugvip (changes requested):
One of those copilot comments, ignore the rest 😅 

Hit more conflicts though : /

GabDug:
@Rugvip No more conflicts!

Rugvip (approved):

PR #32563, opened 2026-01-29
fix(auth): oidc token expiration

### LINKED ISSUE #32115: Plugin authentication breaks portal after 3 hours of uptime  "Failed to verify incoming user token no applicable key found in the JSON Web Key Set"
### 📜 Issue Labels

### 🔎 Search Terms

```plain
Failed to verify incoming user token - JWKS error - JSON Web Key Set
```

### 🗃️ Project Area

Auth

### 🔗 External Integration

N/A

### 📝 Description & Context

**The issue**
⚠️ Our Backstage powered portal has been down for all of our users  since we upgraded Backstage from version 1.44.2 to 1.45.2. Note my team mate posted a [question on discord](https://discord.com/channels/687207715902193673/978977570852835379/threads/1448691580323106826) but at the time of writing there were no answers yet.

🎉 UPDATE: Read below for more details. Summary: We were able to pinpoint the culprit and solve the issue for now by downgrading @backstage/plugin-auth-backend v0.25.6 to v0.25.5. The implementation for [feat(auth): allow configuring DCR token expiration #31278](https://github.com/backstage/backstage/pull/31278/files) is the cause for this issue.

🐛 This means there is a bug in @backstage/plugin-auth-backend v0.25.6 that needs to be fixed

**What we observed**
- After 3 hours of uptime all users would start seeing 401 errors in the frontend
- No data would load at all
- (Note: we first suspected that our upgrade to Node 22 might also have an effect. We downgraded back to Node 20 and the issue was still there)

**The main error logged in the backend**
```
Failed to verify incoming user token no applicable key found in the JSON Web Key Set

Error code: `ERR_JWKS_NO_MATCHING_KEY`

JWKSNoMatchingKey: no applicable key found in the JSON Web Key Set at LocalJWKSet.getKey
```

**Workaround implemented**
We figured out that restarting the pods would prevent the error from happening, since the keys would not get a chance to expire in the first place. This is how we are going into the weekend

**Suspected cause**
The more we look at the [code changes](https://github.com/backstage/backstage/pull/31278/files) for @backstage/plugin-auth-backend v0.25.6 the more we suspect the issue lies there.

A new experimental setting `dcrTokenExpiration` was introduced however the normal code flow was changed as well. In `plugins/auth-backend/src/service/router.ts` in line 166 a new `oidcTokenIssuer` is created with `expirationSeconds: dcrTokenExpiration`. But then existing code flow is changed in line 175 where this new issuer is passed into the created `oidcRouter`. This could easily break existing logic.

Note we are currently testing downgrading to version v0.25.5 to see if this solves our issue. We wanted to created this issue early to help you if you face similar issues.

[UPDATES after posting the original issue]
- Our workaround (restarting pods) has been helping in a fantastic way - at least our users are not bothered by problems
- However, our deployment runs into a problem when *trying* the downgrade of @backstage/plugin-auth-backend `Unhandled rejection Plugin 'auth' startup failed; caused by Error: The migration directory is corrupt, the following files are missing: 20251118120000_oauth_state_text.js`. This is another class of problems we are running into in a situation like this when trying to rollback faulty updates
- My amazing colleague Rik has manually fixed the error on our dev environment by running `knex migrate` commands. 
- We are keeping an eye on the dev environment to see if errors are popping up after 3 hours
- We fixed the migration script errors by including the following in our database config as per the tip in this [issue](https://github.com/backstage/backstage/issues/11218#issuecomment-3140082587)
```
knexConfig:
      migrations:
        disableMigrationsListValidation: true
```
- This helped downgrading the package to 0.25.5
- 🎉 After getting this through the errors finally disappeared confirming the suspected cause
- Note: We are using auth.providers.microsoft for our user tokens

### 👍 Expected Behavior

There are no errors when you use Backstage

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

_No response_

Undecided

drodil:

this will fix issue with custom `auth.backstageTokenExpiration` being longer than OIDC token expiration causing issues with auth providers not being able to verify user tokens.

closes #32115

### DIFF
--- a/plugins/auth-backend/src/service/readTokenExpiration.ts
+++ b/plugins/auth-backend/src/service/readTokenExpiration.ts
@@ -29,8 +29,11 @@ export function readBackstageTokenExpiration(config: RootConfigService) {
 }
 
 export function readDcrTokenExpiration(config: RootConfigService) {
+  const backstageExpiration = readBackstageTokenExpiration(config);
   return readTokenExpiration(config, {
     configKey: 'auth.experimentalDynamicClientRegistration.tokenExpiration',
+    minExpiration: backstageExpiration,
+    defaultExpiration: backstageExpiration,
   });
 }
 
--- a/.changeset/kind-candles-listen.md
+++ b/.changeset/kind-candles-listen.md
@@ -1,0 +1,8 @@
+---
+'@backstage/plugin-auth-backend': patch
+---
+
+Fixed OIDC token being able to expire before backstage token.
+
+This can cause issue with OIDC removing public keys from the `KeyStore`
+needed by the backstage token issuer.
--- a/plugins/auth-backend/src/service/readTokenExpiration.test.ts
+++ b/plugins/auth-backend/src/service/readTokenExpiration.test.ts
@@ -17,6 +17,7 @@
 import { ConfigReader } from '@backstage/config';
 import {
   readBackstageTokenExpiration,
+  readDcrTokenExpiration,
   readTokenExpiration,
 } from './readTokenExpiration.ts';
 
@@ -100,6 +101,34 @@ describe('Test for default backstage token expiry time', () => {
         defaultExpiration: 1234,
       }),
     ).toBe(1234);
+  });
+
+  it('should not allow dcr token expiration to be lower than backstage token expiration', () => {
+    const config = new ConfigReader({
+      app: {
+        baseUrl: 'http://example.com/extra-path',
+      },
+      auth: {
+        backstageTokenExpiration: { minutes: 30 },
+        experimentalDynamicClientRegistration: {
+          tokenExpiration: { minutes: 10 }, // Cannot be smaller than backstageTokenExpiration
+        },
+      },
+    });
+    expect(readDcrTokenExpiration(config)).toBe(1800);
+  });
+
+  it('should default dcr token expiration to backstage token expiration when not configured', () => {
+    const config = new ConfigReader({
+      app: {
+        baseUrl: 'http://example.com/extra-path',
+      },
+      auth: {
+        backstageTokenExpiration: { hours: 2 }, // 7200 seconds
+        // experimentalDynamicClientRegistration.tokenExpiration not configured
+      },
+    });
+    expect(readDcrTokenExpiration(config)).toBe(7200);
   });
 
   it('will return custom min/max expiration', () => {


benjdlambert:
As an aside for this, it looks like MCP clients like Cursor / Claude, /should/ implement `refresh_tokens` which has been implemented part of #32312 or #31482. So it might be worth recommending that instead of increasing the expiry of said tokens.

Haven't had time to test this yet though locally with clients to see if they actually do refresh the token for you instead of going through the auth flow after 1h every time.

drodil:
@benjdlambert cool! Then we could just revert the custom DCR token expiration. Do you have a plan for moving the DCR out of the experimental phase?

benjdlambert:
Honestly I'd rather just use CIMD instead of DCR. But it's not that widely implemented yet. So that's why they're both in an experimental / discovery phase.

plugins/auth-backend/src/service/readTokenExpiration.ts
```diff
@@ -29,8 +29,11 @@ export function readBackstageTokenExpiration(config: RootConfigService) {
 }
 
 export function readDcrTokenExpiration(config: RootConfigService) {
+  const backstageExpiration = readBackstageTokenExpiration(config);
   return readTokenExpiration(config, {
     configKey: 'auth.experimentalDynamicClientRegistration.tokenExpiration',
+    minExpiration: backstageExpiration,
```
vinzscam, line 35: let's ship this as quick fix for those who haven't specified dcr tokenExpiration but the problem still persist if `experimentalDynamicClientRegistration.tokenExpiration` has been defined, as either dcr token or user token will be wiped

vinzscam: Honestly I think we should revert https://github.com/backstage/backstage/pull/31278.
As with these changes, if `auth.experimentalDynamicClientRegistration.tokenExpiration` is specified:
- if the value is < `backstageExpiration`, it has no effect
- if the value is > `backstageExpiration` the dcr keys will be wiped anyway by `TokenFactory`

drodil: We are using 8 hour dcr token expiration and it really doesn't expire before that. Without the config it expired in just one hour by the default backstage token expiration and made the usage of mcp server hard

drodil: Actually, you might be right that the signing key is cleared based on backstage expiration * 3, so in 3 hours 🤔 The right way to fix this would be to save the expiration time/cleanup time to the database and clean keys based on that.

vinzscam: yeah exactly. But given the comments above, if we are moving away from this `uth.experimentalDynamicClientRegistration.tokenExpiration` it might be worth to remove this feature. But if we plan to keep it then we need to fix this properly somehow

benjdlambert:
@drodil what do you think? Seen as `refresh_tokens` are shipping in this release should we walk back on the DCR token expiration instead and push towards using the `experimentalRefreshToken` config instead?

benjdlambert:
Chatted with @rugvip, we're gonna do that, will whip up a PR for this now :pray:

PR #32566, opened 2026-01-29
fix(storybook): clean up configuration after v10.3 upgrade

cdedreuille:
## Summary

This PR fixes Storybook configuration issues after upgrading to v10.3:

- Removes non-existent `.mdx` file patterns that were causing multiple warnings on startup
- Removes deprecated `showName` property from toolbar configurations that was causing TypeScript errors

### MDX Pattern Removal
The configuration was looking for `.mdx` files in all story directories, but none exist. This was generating 8 warnings on every Storybook launch. The fix simplifies the stories array to only include the actual `.stories.*` files.

### Toolbar Configuration Fix
The `showName` property is no longer supported in Storybook v10.3's `ToolbarConfig` type. The `dynamicTitle: true` property provides similar functionality by showing the selected item's title in the toolbar.

## Files Changed

- `.storybook/main.ts` - Removed MDX pattern and simplified stories array
- `.storybook/preview.tsx` - Removed deprecated `showName` from toolbar configs

## Testing

- [ ] Storybook launches without warnings about missing `.mdx` files
- [ ] No TypeScript errors in Storybook configuration files
- [ ] Theme mode and theme name toolbar items work correctly


Rugvip (approved):
🧹

PR #32575, opened 2026-01-29
Fix CSS Module purity errors for Next.js 16 compatibility

cdedreuille:
## Summary

Fixes build errors and ESLint compatibility issues introduced after upgrading to Next.js 16.1.6 (PR #32565).

### CSS Module Purity Errors
The build was failing with CSS Module purity errors:

```
Selector "[data-theme="dark"]" is not pure. Pure selectors must contain at least one local class or id.
```

Next.js 16 with Turbopack enforces stricter CSS Module validation. Global selectors like `[data-theme='dark']` cannot exist standalone - they must be combined with local classes.

**Changes:**
- Flattened nested dark theme selectors in `Popover.module.css`
- Flattened nested dark theme selectors in `Tooltip.module.css`
- Changed from `[data-theme='dark'] { .class { } }` to `[data-theme='dark'] .class { }`

This matches the pattern already used correctly in `Dialog.module.css`.

### ESLint 9 Upgrade
Next.js 16 requires ESLint 9 for proper flat config support, causing pre-commit hook failures with circular reference errors.

**Changes:**
- Upgraded ESLint from ^8 to ^9
- Migrated from `.eslintrc.json` to `eslint.config.mjs` (flat config format)
- Updated lint script to use `eslint .` directly (`next lint` removed in Next.js 16)
- Added proper tsconfig includes for Next.js 16 dev types
- Fixed layout.tsx to use proper theme CSS link tags

## Test plan

- [x] Run `yarn build` in `docs-ui` directory - build completes successfully
- [x] Verify all 41 static pages generate without errors  
- [x] Run `yarn lint` - passes with only warnings (no errors)
- [x] Pre-commit hooks work without circular structure errors
- [ ] Visual check that dark theme still works correctly for Popover and Tooltip components

### DIFF
--- a/docs-ui/src/components/CustomTheme/customTheme.tsx
+++ b/docs-ui/src/components/CustomTheme/customTheme.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useEffect, useState, useCallback } from 'react';
+import { useEffect, useState, useCallback, useSyncExternalStore } from 'react';
 import CodeMirror from '@uiw/react-codemirror';
 import { sass } from '@codemirror/lang-sass';
 import styles from './styles.module.css';
@@ -13,6 +13,10 @@ import { RiArrowDownSLine } from '@remixicon/react';
 const defaultTheme = `:root {
   --bui-bg-solid: #000;
 }`;
+
+// Stable server snapshots for useSyncExternalStore
+const serverIsClient = false;
+const serverDefaultTheme = defaultTheme;
 
 const myTheme = createTheme({
   theme: 'light',
@@ -46,11 +50,40 @@ const myTheme = createTheme({
 });
 
 export const CustomTheme = () => {
-  const [isClient, setIsClient] = useState(() => typeof window !== 'undefined');
   const [open, setOpen] = useState(true);
-  const [customTheme, setCustomTheme] = useState<string | undefined>(undefined);
   const { selectedThemeName } = usePlayground();
   const [savedMessage, setSavedMessage] = useState<string>('Save');
+
+  // SSR-safe client detection
+  const isClient = useSyncExternalStore(
+    () => () => {},
+    () => true,
+    () => serverIsClient,
+  );
+
+  // SSR-safe localStorage access for custom theme
+  const customThemeFromStorage = useSyncExternalStore(
+    callback => {
+      window.addEventListener('storage', callback);
+      return () => window.removeEventListener('storage', callback);
+    },
+    () => {
+      const stored = localStorage.getItem('customThemeCss');
+      if (!stored) {
+        localStorage.setItem('customThemeCss', defaultTheme);
+        return defaultTheme;
+      }
+      return stored;
+    },
+    () => serverDefaultTheme,
+  );
+
+  const [customTheme, setCustomTheme] = useState(customThemeFromStorage);
+
+  // Sync from storage when it changes
+  useEffect(() => {
+    setCustomTheme(customThemeFromStorage);
+  }, [customThemeFromStorage]);
 
   const updateStyleElement = (theme: string) => {
     let styleElement = document.getElementById(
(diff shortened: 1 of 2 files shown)


.changeset/fix-nextjs16-css-modules.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/ui': patch
+---
+
+Fixed CSS Module syntax to comply with Next.js 16 Turbopack validation by flattening nested dark theme selectors.
+
+**Affected components:** popover, tooltip
```
Naycon:
```suggestion
**Affected components:** Popover, Tooltip
```

These need to match in case for the changelog sync script to work properly I think.

.changeset/fix-nextjs16-css-modules.md
Naycon, line 1: This file name doesn't feel like the usual changeset file names. 🤔. Likely doesn't matter much, but I assume this was an AI not running the changeset commands from CLI. :)

docs-ui/src/app/layout.tsx
```diff
@@ -49,6 +47,10 @@ export default async function RootLayout({
       data-theme-name="backstage"
       suppressHydrationWarning
     >
+      <head>
+        <link rel="stylesheet" href="/theme-backstage.css" />
+        <link rel="stylesheet" href="/theme-spotify.css" />
+      </head>
```
Naycon: Why was this change necessary? We were able to import `globals.css` above, what's different with the theme files?

docs-ui/eslint.config.mjs
```diff
@@ -0,0 +1,18 @@
...
+const eslintConfig = defineConfig([
+  ...nextCoreWebVitals,
+  globalIgnores(['.next/**', 'dist/**', 'node_modules/**']),
+  {
+    rules: {
+      'notice/notice': 'off',
+      'react/forbid-elements': 'off',
+      'jsx-a11y/alt-text': 'off',
+      '@next/next/no-css-tags': 'off',
+      'react-hooks/set-state-in-effect': 'warn',
```
Naycon: Are we sure these should be turned off rather than fixed? Why are they necessary?

packages/ui/src/components/Popover/Popover.module.css
```diff
@@ -122,20 +122,16 @@
     }
   }
 
-  [data-theme='dark'] {
-    .bui-Popover {
-      background: var(--bui-bg-surface-2);
-      border: 1px solid var(--bui-gray-4);
-    }
+  [data-theme='dark'] .bui-Popover {
```
Naycon, line 125: Out of curiosity, would the error be fixed if we'd specify the element the `data-theme` is hosted on? E.g. doing `body[data-theme='dark']`? I think that'd be a better fix in the long run, allowing us to more easily group and nest things that are meant to go together without repeating selectors. But this is fine for now.

.changeset/fix-nextjs16-css-modules.md
cdedreuille: Other changeset did the same. I'm assume this does not matter. All you need is for it to be different from the other ones.

docs-ui/src/app/layout.tsx
cdedreuille: It did not work with Next.js 16 for files placed in the public folder. The global CSS is not in the public folder.

docs-ui/eslint.config.mjs
cdedreuille: You're right. I fix those and remove the new rules

.changeset/fix-nextjs16-css-modules.md
cdedreuille: Fixed

docs-ui/src/app/layout.tsx
cdedreuille: You're right. I moved all of them inside the `/css` folder.

PR #32586, opened 2026-01-29
docs: add backstage.peerModules metadata field

Rugvip:

This proposes a new `backstage.integrationFor` field in package metadata. It closes the gap where modules currently are unable to indicate what package they're really from. For example the scaffolder module for the catalog is only represented as a module for the catalog backend, but in reality it is closely tied to the scaffolder too.

Aiming to start out with simply introducing the field and let people that want to populate it, but down the line I the goal would have to be this more closely tied to package discovery.


packages/cli/src/modules/maintenance/commands/repo/fix.ts
```diff
@@ -430,6 +430,54 @@ export function fixPluginPackages(
...
+  if (integrationFor === undefined) {
+    return;
+  }
+
+  // Validate that integrationFor is only used on module packages
+  if (role !== 'backend-plugin-module' && role !== 'frontend-plugin-module') {
+    const path = relativePath(
+      paths.targetRoot,
+      resolvePath(pkg.dir, 'package.json'),
+    );
```
freben: Just compute this path once on the outer level to avoid repetition

Rugvip:
Hmm, thinking a bit more more about this one, maybe it would be better to invert the relationship? As in it is the plugin's package responsibility to point to integrating modules. That way it's a lot easier to discover those packages without downloading all of NPM.

Rugvip:
Updated to flip this around into `backstage.peerModules` instead.

packages/cli/src/modules/maintenance/commands/repo/fix.ts
```diff
@@ -430,6 +430,47 @@ export function fixPluginPackages(
   }
 }
 
+export function fixPeerModules(pkg: FixablePackage) {
```
freben, line 433: No fix currently though - just validation. Do we want it to be possible to auto generate this?

Rugvip: I think it's a bit sketchy to auto generate tbh, it's hard to identify modules to begin with and even harder to know whether they should be recommended or not.

Rugvip:
Feeling relatively confident about this structure now so gonna :shipit: 

Bit from a separate discussion: there's a question whether we'd want to be able to provide more data for these down the line. I want to try to keep this super simple to start and see if that's enough, but we could potentially add optional formats like `{ name: ..., optional: true }` to the array items if needed, or an additional `optionalPeerModules`. All depends on how we end up tying all of this together

PR #32588, opened 2026-01-30
feat(catalog): add table column customization via app-config

### LINKED ISSUE #32549: feat(catalog): Enable catalog table column customization via app-config
### 📜 Issue Labels

### 🔎 Search Terms

```plain
- catalog table columns configuration
- catalog table customize columns
- catalog column config
- hide catalog columns
- custom catalog columns
- catalog table metadata columns
- CatalogTable columns app-config
```

### 🗃️ Project Area

Catalog

### Rationale

Platform engineers deploying Backstage often need to customize the Software Catalog table to match their organization's needs. Currently, customizing which columns appear in the catalog table requires code changes - either by passing a custom `columns` prop to `CatalogTable`/`DefaultCatalogPage` or by creating a custom columns function.

This presents challenges for:

1. **Platform engineers** who want to hide irrelevant columns (e.g., hiding `namespace` when all entities use `default`, or hiding `tags` if not used)
2. **Organizations** that want to display custom metadata columns (e.g., security tier, cost center, team ownership from annotations/labels)

### Who benefits

- Platform engineers who want to customize the catalog without code changes
- Organizations with custom entity metadata they want to surface in the catalog table

### Related context

- This is a common request in the Backstage community for configuration-driven customization

### 📝 Proposal

Add a new configuration option `catalog.table.columns` that allows customizing the catalog table columns via `app-config.yaml`.

### Configuration Schema

```yaml
catalog:
  table:
    columns:
      # Option 1: Include only specific columns (whitelist)
      include:
        - name
        - owner
        - type
        - lifecycle
      
      # Option 2: Exclude specific columns from defaults (blacklist)
      exclude:
        - tags
        - namespace
        - description
      
      # Option 3: Add custom columns from entity metadata
      custom:
        - title: "Security Tier"
          field: "metadata.annotations['security/tier']"
          width: 120
          sortable: true
          defaultValue: "Not Set"
        - title: "Cost Center"
          field: "metadata.labels['cost-center']"
          kind:
            - Component
            - Resource
```

### Available Built-in Column IDs

- `name` - Entity name/title
- `owner` - Owner relation
- `type` - spec.type
- `lifecycle` - spec.lifecycle
- `description` - metadata.description
- `tags` - metadata.tags
- `namespace` - metadata.namespace
- `system` - System relation
- `targets` - Location targets (for Location kind)

### Configuration Options

- `include` (`string[]`) - When specified, only these columns are shown (whitelist mode)
- `exclude` (`string[]`) - Columns to remove from the defaults (blacklist mode)
- `custom` (`CustomColumn[]`) - Additional columns to add

### Custom Column Properties

- `title` (string, required) - Column header text
- `field` (string, required) - Entity field path (supports dot notation and bracket notation for annotations/labels)
- `width` (number, optional) - Column width in pixels
- `sortable` (boolean, optional) - Enable sorting (default: true)
- `defaultValue` (string, optional) - Value to display when field is empty
- `kind` (string | string[], optional) - Limit column to specific entity kinds

### 🔄 Alternatives

_No response_

samarthsinh2660:
### Summary

This PR adds support for customizing catalog table columns via `app-config.yaml` when using the New Frontend System, as requested in #32549.

Platform engineers can now configure which columns appear in the catalog table without code changes. The configuration supports three modes:

**1. Include mode (whitelist)** - Show only specific columns:
```yaml
app:
  extensions:
    - page:catalog:
        config:
          columns:
            include:
              - name
              - owner
              - type
```

**2. Exclude mode (blacklist)** - Hide specific columns:
```yaml
app:
  extensions:
    - page:catalog:
        config:
          columns:
            exclude:
              - namespace
              - tags
              - description
```

**3. Custom columns** - Add columns from entity metadata:
```yaml
app:
  extensions:
    - page:catalog:
        config:
          columns:
            custom:
              - title: "Security Tier"
                field: "metadata.annotations['security/tier']"
                width: 120
                sortable: true
                defaultValue: "Not Set"
              - title: "Cost Center"
                field: "metadata.labels['cost-center']"
                kind:
                  - Component
                  - Resource
```

### Available Built-in Column IDs
`name`, `owner`, `type`, `lifecycle`, `description`, `tags`, `namespace`, `system`, `targets`

### Custom Column Properties
| Property | Type | Required | Description |
|----------|------|----------|-------------|
| `title` | string | Yes | Column header text |
| `field` | string | Yes | Entity field path (supports dot/bracket notation) |
| `width` | number | No | Column width in pixels |
| `sortable` | boolean | No | Enable sorting (default: true) |
| `defaultValue` | string | No | Value when field is empty |
| `kind` | string \| string[] | No | Limit to specific entity kinds |

### Implementation Details

- **Config Schema**: Added to `config.d.ts` with full TypeScript types
- **Column Utilities**: New `columnConfig.ts` with `resolveFieldPath()`, `createCustomColumn()`, and `applyColumnConfig()` functions
- **New Frontend System**: Integrated via `pages.tsx` with Zod validation schema
- **Component Updates**: `CatalogTable` and `BaseCatalogPage` accept `columnsConfig` prop

Fixes #32549

### DIFF
--- a/.changeset/catalog-table-columns-config.md
+++ b/.changeset/catalog-table-columns-config.md
@@ -22,11 +22,11 @@ app:
               - namespace
               - tags
             custom:
-              - title: "Security Tier"
+              - title: 'Security Tier'
                 field: "metadata.annotations['security/tier']"
                 width: 120
-                defaultValue: "Not Set"
-              - title: "Cost Center"
+                defaultValue: 'Not Set'
+              - title: 'Cost Center'
                 field: "metadata.labels['cost-center']"
                 kind:
                   - Component
@@ -36,6 +36,7 @@ app:
 Available built-in column IDs: `name`, `owner`, `type`, `lifecycle`, `description`, `tags`, `namespace`, `system`, `targets`.
 
 Custom columns support:
+
 - Dot notation for field paths (e.g., `spec.type`)
 - Bracket notation for annotations/labels (e.g., `metadata.annotations['key']`)
 - Optional width, sortable, defaultValue, and kind filtering
--- a/.changeset/catalog-table-columns-config.md
+++ b/.changeset/catalog-table-columns-config.md
@@ -1,0 +1,41 @@
+---
+'@backstage/plugin-catalog': minor
+---
+
+Added support for customizing catalog table columns via `app-config.yaml` in the New Frontend System.
+
+This feature allows platform engineers to configure which columns are displayed in the catalog table without code changes. The configuration supports:
+
+- **Include mode**: Show only specific columns by listing them
+- **Exclude mode**: Hide specific columns from the defaults
+- **Custom columns**: Add columns from entity metadata (annotations, labels, spec fields)
+
+Example configuration:
+
+```yaml
+app:
+  extensions:
+    - page:catalog:
+        config:
+          columns:
+            exclude:
+              - namespace
+              - tags
+            custom:
+              - title: "Security Tier"
+                field: "metadata.annotations['security/tier']"
+                width: 120
+                defaultValue: "Not Set"
+              - title: "Cost Center"
+                field: "metadata.labels['cost-center']"
+                kind:
+                  - Component
+                  - Resource
+```
+
+Available built-in column IDs: `name`, `owner`, `type`, `lifecycle`, `description`, `tags`, `namespace`, `system`, `targets`.
+
+Custom columns support:
+- Dot notation for field paths (e.g., `spec.type`)
+- Bracket notation for annotations/labels (e.g., `metadata.annotations['key']`)
+- Optional width, sortable, defaultValue, and kind filtering
--- a/.changeset/catalog-table-columns-config.md
+++ b/.changeset/catalog-table-columns-config.md
@@ -39,4 +39,4 @@ Custom columns support:
 
 - Dot notation for field paths (e.g., `spec.type`)
 - Bracket notation for annotations/labels (e.g., `metadata.annotations['key']`)
-- Optional width, sortable, defaultValue, and kind filtering
+- Optional `width`, `sortable`, `defaultValue`, and `kind` filtering
--- a/plugins/catalog/config.d.ts
+++ b/plugins/catalog/config.d.ts
@@ -23,5 +23,67 @@ export interface Config {
       | {
           limit?: number;
         };
+
+    /**
+     * Configuration for the catalog table columns.
+     * Allows customizing which columns are displayed and adding custom columns.
+     * @deepVisibility frontend
+     */
+    table?: {
+      /**
+       * Column configuration options.
+       */
+      columns?: {
+        /**
+         * When specified, only these columns are shown (whitelist mode).
+         * Available column IDs: name, owner, type, lifecycle, description, tags, namespace, system, targets
+         */
+        include?: string[];
+
+        /**
+         * Columns to remove from the defaults (blacklist mode).
+         * Available column IDs: name, owner, type, lifecycle, description, tags, namespace, system, targets
+         */
+        exclude?: string[];
+
+        /**
+         * Custom columns to add from entity metadata.
+         */
+        custom?: Array<{
+          /**
+           * Column header text (required).
+           */
+          title: string;
+
+          /**
+           * Entity field path using dot notation.
+           * Supports bracket notation for annotations/labels.
+           * Examples: "metadata.name", "metadata.annotations['backstage.io/techdocs-ref']"
+           */
+          field: string;
+
+          /**
+           * Column width in pixels (optional).
+           */
+          width?: number;
+
+          /**
+           * Enable sorting for this column (default: true).
+           */
+          sortable?: boolean;
+
+          /**
+           * Value to display when field is empty (optional).
+           */
+          defaultValue?: string;
+
+          /**
+           * Limit column to specific entity kinds (optional).
+           * When specified, the column only appears for these kinds.
+           */
+          kind?: string | string[];
+        }>;
+      };
+    };
   };
 }
--- a/plugins/catalog/package.json
+++ b/plugins/catalog/package.json
@@ -64,6 +64,7 @@
     "@backstage/core-components": "workspace:^",
     "@backstage/core-plugin-api": "workspace:^",
     "@backstage/errors": "workspace:^",
+    "@backstage/filter-predicates": "workspace:^",
     "@backstage/frontend-plugin-api": "workspace:^",
     "@backstage/integration-react": "workspace:^",
     "@backstage/plugin-catalog-common": "workspace:^",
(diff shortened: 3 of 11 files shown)


.changeset/catalog-table-columns-config.md
```diff
@@ -0,0 +1,42 @@
...
+            custom:
+              - title: 'Security Tier'
+                field: "metadata.annotations['security/tier']"
+                width: 120
+                defaultValue: 'Not Set'
+              - title: 'Cost Center'
+                field: "metadata.labels['cost-center']"
+                kind:
+                  - Component
+                  - Resource
```
vinzscam, line 33: here perhaps we can use the new catalog predicates
https://backstage.io/docs/features/software-catalog/catalog-customization/#entity-predicate-queries

.changeset/catalog-table-columns-config.md
```diff
@@ -0,0 +1,42 @@
...
+```yaml
+app:
+  extensions:
+    - page:catalog:
+        config:
+          columns:
+            exclude:
+              - namespace
+              - tags
+            custom:
```
vinzscam, line 24: @backstage/catalog-maintainers we need to discuss this contract

.changeset/catalog-table-columns-config.md
```diff
@@ -0,0 +1,42 @@
...
+              - title: 'Cost Center'
+                field: "metadata.labels['cost-center']"
+                kind:
+                  - Component
+                  - Resource
+```
+
+Available built-in column IDs: `name`, `owner`, `type`, `lifecycle`, `description`, `tags`, `namespace`, `system`, `targets`.
+
+Custom columns support:
```
vinzscam, line 38: something we need to consider is how to support ordering when pagination is enabled. Ideally we shouldn't define any `customSort` in the columns but instead we should invoke `updateFilters({ order: new EntityOrderFilter(...) })` via `useEntityList()`

vinzscam (changes requested):
great to see this

.changeset/catalog-table-columns-config.md
samarthsinh2660: That makes sense - I'll look into replacing the custom `kind` filter with the catalog
  predicate queries for consistency with the existing Backstage patterns

.changeset/catalog-table-columns-config.md
samarthsinh2660: Happy to wait for feedback from @backstage/catalog-maintainers on the include/exclude API
  structure. Should I hold off on changes until that discussion concludes, or are there specific alternatives you'd like
   me to explore?

.changeset/catalog-table-columns-config.md
samarthsinh2660: Good catch on the pagination issue. I'll investigate using `EntityOrderFilter` with
  `useEntityList()` instead of `customSort`. However, for custom columns pointing to arbitrary entity fields (like
  `metadata.annotations['security/tier']`), the catalog backend may not support ordering by those fields. Should custom
  columns disable sorting by default, or is there a way to support backend ordering for dynamic fields?

ahamedjobayer551-debug (approved):
@samarthsinh2660

.changeset/catalog-table-columns-config.md
its-mitesh-kumar: @vinzscam if you are particularly talking about `exclude` contract, I think it can be quite useful. For example, if a user wants to remove just one column from the default set, they can simply add that column to `exclude` instead of listing all the remaining columns in `include`.

It also helps with forward compatibility. If new columns are added to the default column set in the future, they would automatically appear unless explicitly excluded. This means users won’t need to keep updating their configuration whenever new default columns are introduced.

@backstage/catalog-maintainers whats your thought on it ?

.changeset/catalog-table-columns-config.md
its-mitesh-kumar: @samarthsinh2660 Please check if the catalog backend support sorting with pagination on dynamic fields, if not please proceed with frontend sorting for custom columns

.changeset/catalog-table-columns-config.md
its-mitesh-kumar: @samarthsinh2660 could you please address this comments.

samarthsinh2660: I'm planning to replace the kind field in custom columns with
  filter using the catalog predicate query system (FilterPredicate from
  @backstage/filter-predicates), so users can write:

  custom:
    - title: "Cost Center"
      field: "metadata.labels['cost-center']"
      filter:
        kind: { $in: [Component, Resource] }

  This is more consistent with the existing Backstage patterns and more flexible
   (supports $all, $any, $not, $in, etc.). I'll update the PR shortly with this
  change.

.changeset/catalog-table-columns-config.md
samarthsinh2660: Hey @its-mitesh-kumar, I checked — the catalog backend only supports ordering
  by a fixed set of standard fields (like metadata.name, metadata.namespace).
  Arbitrary paths such as metadata.annotations['security/tier'] are not
  supported as backend order fields.

  So for custom columns, frontend-only sorting via customSort is the right
  approach. I'll keep customSort for custom columns and leave it disabled by
  default (sortable: false as default) since backend ordering isn't available
  for dynamic fields. Will proceed with this.

jmezach:
Great to see this, looking forward to this getting merged :). One thing I was wondering while looking at this is if this would support custom columns with their own rendering. For example, we've made a custom column that based on a custom annotation renders a logo of the deployment target (ie. Kubernetes, Windows, etc.). I briefly looked through the implementation, but it looks like that won't be supported?

Sarabadu:
Sharing my thought here from the discussion we had in the SIG.

While doing the catalog migration to NFS, missing this configuration was one of the missing pieces that bring me to a customIndexPage, so im looking forward to have a way to do it.

We have some custom components in some colums, so also wondering if we should have a plan for that configuration too. My first expectation was to find a `CatalogColumBlueprint` to work with and been able to add column in the same way we add filters to the page.

If we plan for that i would expect the built in configuration to look more like

```yaml
app:
  extensions:
    - entity-column:catalog/name
    - entity-column:catalog/owner
    - entity-column:catalog/type
       config:
           filter: {kind: component}
    - entity-column:catalog/namespace: false
              
```              

Some limitation would be that for _any_ custom column would need to create an extension so if we want to create custom one only by config maybe we need to have some page configuration anyways 🤔 

i mean there is no way to add 
```yaml
app:
  extensions:
    - entity-column:catalog/non-existent
       config:
           ....
```

The other limitation on extension sorting is that we cannot vary the order of existent extension depending a filter. IIRC we we cannot have more that one configuration per extension id, so if we want different order of the column depending on the entity filter, is not really possible

PR #32614, opened 2026-01-30
feat: use branch name instead of tag for stable docs

aramissennyeydd:

We need a patch process for docs that allows us to update docs in stable releases. The .patches flow should work for the most part, the only additional need is for the docs to use the patch branch and not the latest tag to simplify the workflow (no patch packages automatic PR needed). We can just open PRs against the patch branch with changes, merging those PRs should allow us to see doc updates reflected in the next run of the deploy_microsite in master.

Validation: https://github.com/backstage/backstage/actions/runs/21534050162/job/62055966406?pr=32614, uses patch/v1.47.0 as the ref.

### DIFF
--- a/.github/workflows/deploy_microsite.yml
+++ b/.github/workflows/deploy_microsite.yml
@@ -233,7 +233,7 @@ jobs:
       - name: checkout latest release
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
-          ref: refs/tags/${{ needs.stable.outputs.release }}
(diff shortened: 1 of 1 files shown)


aramissennyeydd:
@awanlin What would be the right place to document that? Not sure I see a good place for this style of documentation.

awanlin:
> @awanlin What would be the right place to document that? Not sure I see a good place for this style of documentation.

Personally I think putting at the bottom of this makes sense, we can always move it later: https://github.com/backstage/backstage/blob/master/microsite/README.md

PR #32616, opened 2026-01-31
frontend-test-utils: add shallowExtensionInstance for lightweight extension testing

Rugvip:

This adds a new `shallowExtensionInstance` utility for lightweight unit testing of extension outputs. It executes an extension's factory in isolation and returns an `ExtensionDataContainer`, without building a full app tree like `createExtensionTester` does.

**When to use it:**
- Testing that an extension produces correct output data given config/inputs
- Fast unit tests that don't need React rendering or multi-extension wiring

**Comparison with `createExtensionTester`:**

| Aspect | `shallowExtensionInstance` | `createExtensionTester` |
|--------|---------------------------|------------------------|
| What it does | Directly calls the extension factory | Builds full app tree |
| Performance | Fast - single function call | Slower - resolves entire tree |
| Extension support | v2 only | v1 and v2 |
| React rendering | No | Yes, via `.reactElement()` |
| Multi-extension | No | Yes, via `.add()` |

Also migrated existing tests that only use `.get()` to the new utility.

### DIFF
--- a/packages/frontend-plugin-api/src/wiring/createExtension.test.ts
+++ b/packages/frontend-plugin-api/src/wiring/createExtension.test.ts
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
 
-import { createExtensionTester } from '@backstage/frontend-test-utils';
+import {
+  shallowExtensionInstance,
+  createExtensionTester,
+} from '@backstage/frontend-test-utils';
 import { createExtension } from './createExtension';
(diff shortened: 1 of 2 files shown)


freben (commented):
Would it be reasonable to do `const { outputs } = createShallowExtensionInstance`? "Create" for consistency, and result room for whatever else we may come up with?

Rugvip:
@freben dropping this idea for new either way, it's too similar to `createExtensionTester` right now

Issue #32584, opened 2026-01-29
Fix Slack DMs suppressed by mixed group + user recipients


kaidubauskas-dd:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
"SlackNotificationProcessor DMs","notifications mixed users and groups","slack DMs not sending for users"
```

### 🗃️ Project Area

Notifications

### 🔖 Need

- Today if any recipient is a group entity, the SlackProcessor returns [early](https://github.com/backstage/backstage/blob/master/plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts#L276) in postProcess and sends no user DMs. If a user is explicitly included in the receipients list, the presence of a group should not prevent this user from getting a DM. 
- This issue was raised in [#29308](https://github.com/backstage/backstage/pull/29308#discussion_r2007032780) before the metadata field was added in [#30028](https://github.com/backstage/backstage/pull/30028), but was never addressed after. 

### 📝 Proposal

I propose we add a 'explicitUserEntityRefs' key to the NotificationPayload.metadata in processOptions, containing all user refs of the original notification. Then in postProcess we only send DMs to users in the metadata array. This way users resolved from groups still get no DMs while users passed in as user refs do. 

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Yes, and I have enough information to get started

vinzscam:
Feel free to open a PR so that we can discuss the implementation there

PR #32619, opened 2026-01-31
fix(notifications-slack): Only DM explicit user recipients

kaidubauskas-dd:

Issue: [#32584 ](https://github.com/backstage/backstage/issues/32584)

Right now if recipients contains a group entity, we return early in postProcess preventing user DMs. Users explicitly listed as a recipient should still receive direct messages. 

This change ensures user recipients still receive Slack DMs by early exiting only if they were not included in the notifications recipients array and a group entity was a recipient

 
### DIFF
--- a/.changeset/curvy-socks-punch.md
+++ b/.changeset/curvy-socks-punch.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-notifications-backend-module-slack': minor
 ---
 
-**BREAKING**: Only send DMs to user entity recipients. Notifications sent to non-user entities no longer send Slack DMs to resolved users.
+**BREAKING**: Only send direct messages to user entity recipients. Notifications sent to non-user entities no longer send Slack direct messages to resolved users.


plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -210,12 +214,14 @@ export class SlackNotificationProcessor implements NotificationProcessor {
...
     const entityRefs = [options.recipients.entityRef].flat();
 
+    const userEntityRefs: string[] = [];
     const outbound: ChatPostMessageArguments[] = [];
     await Promise.all(
       entityRefs.map(async entityRef => {
         const compoundEntityRef = parseEntityRef(entityRef);
         // skip users as they are sent direct messages
         if (compoundEntityRef.kind === 'user') {
+          userEntityRefs.push(stringifyEntityRef(compoundEntityRef));
```
freben: we already have `entityRef` above

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -272,10 +289,25 @@ export class SlackNotificationProcessor implements NotificationProcessor {
...
+      const hasValidUserRefs =
+        Array.isArray(rawUserRefs) &&
+        rawUserRefs.every(ref => typeof ref === 'string');
+      const explicitUserEntityRefs = hasValidUserRefs
+        ? rawUserRefs.map(ref => stringifyEntityRef(parseEntityRef(ref)))
+        : undefined;
+
+      if (!explicitUserEntityRefs?.includes(normalizedUserRef)) {
+        const entityRefs = [options.recipients.entityRef].flat();
+        if (entityRefs.some(e => parseEntityRef(e).kind === 'group')) {
```
freben: Sorry for a silly question. I am actually not sure what the logic is here, and what is being communicated between the two steps. Can you enlighten me?

freben: Mentioning @henrikedegrd here too, maybe you know better than me

kaidubauskas-dd: Sure. postProcess runs on user entities, including entities resolved from groups, components etc. Im keeping track of which user entities were passed into the notification request in processOptions so we can check if the user was resolved from a different entity type in postProcess. This way, a request can contain a mixed list of groups and users, and the users will still get a DM. Before we just returned early if the request contained any group entity (no DMs would be sent at all)

henrikedegrd: I think what @kaidubauskas-dd says is correct.

previously we checked if any recipient was a group, and therefor just returned and skipped any further processing.

However I think to return an enriched payload that gets sent downstream to postProcess might be hard to maintain later in order to track the original users.

I dont really know what is availible in options.recipients.entityRef here.
If all already are present here (as the original notification-option looked like) then we might not need to have an enriched payload? just see if we ended up in postProcess on an entity that was included in the original options?

And another question is regarding the notification-settings, if you pass along entities from the processOptions, are they bypassed the user set notification-settings? (IE: they dont want this type of notification for this channel? )

kaidubauskas-dd: @henrikedegrd This was an oversight on my part. You're right we should compare to options.recipients. Fixed this to be much simpler

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -273,8 +274,18 @@ export class SlackNotificationProcessor implements NotificationProcessor {
...
-      if (entityRefs.some(e => parseEntityRef(e).kind === 'group')) {
-        // We've already dispatched a slack channel message, so let's not send a DM.
+      const explicitUserEntityRefs = entityRefs
+        .filter(entityRef => parseEntityRef(entityRef).kind === 'user')
+        .map(entityRef => stringifyEntityRef(parseEntityRef(entityRef)));
+      const normalizedUserRef = stringifyEntityRef(
+        parseEntityRef(notification.user),
+      );
+
+      if (
```
henrikedegrd: Can this be simplified even more?

Like (pseudo-code):
`if(!entityRefs.contains(normalizedUserRef)) { return; }`

I dont see why we need to check that entityRefs contains a group also? (But i might miss something)

If we end up here with a user (notification.user) that was not included in the original recipients-list, then it must have been derived from a group and shall not receive a DM?

kaidubauskas-dd: It doesn't necessarily have to be derived from a group, it could be from a component or other entity type. Imo, we _should_ only send DMs to original users. I'm willing to change to this if we agree, but the logic before allowed users resolved from non-group entity types to get DMs, so I assumed this was intentional..? Maybe not?

The logic here still isn't perfect. eg. A notification to a component entity by itself will send DMs to resolved users while a notification to a component and group wont, but without access to the recipient resolution step, I see no clear way to get around this.

henrikedegrd: So to clear up the logic here;

In notification.user we get the user that the notification ends up on (either that that user was targeted directly or via lets say a group or a component) ?

And in options.recipients we get the original notifications _original_ targets?

As it is now and if the above is true, if a notification is targeted 1 user, 1 group and one component. Then that explicit user and neither user derived from the component would get a DM. But if the notification targeted 1 user and 1 component, then all of them would get a DM (as you say).

So in order to do an improvement here, we need to know exactly what notification.user and options.recipients contains.

And then probably we need some form of broader decision on how DM:s shall be sent;

Targeted a group == no DM:s to users?
Targeted anything other than a group == DM:s to users?

If thats the case then we need to lookup all the users belonging to the group and make sure they only get a PR if they were explicitly targeted?

kaidubauskas-dd: > In notification.user we get the user that the notification ends up on (either that that user was targeted directly or via lets say a group or a component) ?
> 
> And in options.recipients we get the original notifications original targets?

Yes, these are both correct

> As it is now and if the above is true, if a notification is targeted 1 user, 1 group and one component. Then that > explicit user and neither user derived from the component would get a DM. But if the notification targeted 1 > user and 1 component, then all of them would get a DM (as you say).

Not quite. Right now, if a notification is targeted to 1 group+1 user+one component, no one gets a DM. In fact, any notification that includes a group will block DMs for _all_ resolved users. If the notification does not contain a group, all resolved users _do_ get a DM.

So
Contains a group --> no DMs at all
No group --> everyone gets a DM

> And then probably we need some form of broader decision on how DM:s shall be sent;

I think this is an important point. Happy to take this conversation elsewhere if you want not sure where would be best.

In my mind, a DM is a personal, direct communication and should only go to users who were intentionally and explicitly named as recipients. Groups and components speak through their channel annotations, and the people behind shouldn't be getting individual DMs just because they happen to own or belong to an entity that received a notification.

I would propose only users who are explicitly named as recipients (kind: user in options.recipients.entityRef) should receive a DM. All other entity types ( groups, components etc.) represent indirect or aggregate targets and should be communicated with via their own annotated Slack channel (slack.com/bot-notify), not by fanning out DMs to every transitively resolved user.

If we go with this, we wouldn't have to do any lookups, we would just remove the check for if a group was a target in my current implementation. 

Maybe an ideal situation would be to default to what I talked about here, but allow configuration for what entity types will send DMs to resolved users (eg. opting groups in while keeping components out)? Though this would require additional catalog lookups for groups,componets etc that largely repeat work already done during [recipient resolution](https://github.com/backstage/backstage/blob/master/plugins/notifications-backend/src/service/router.ts#L827). Also I'm not too sure how expensive lookups are. 

Curious what your thoughts would be here?  (@drodil since you also seemed involved in the initial implementation of this module)

freben: What you say resonates with me, as an innocent unknowing bystander :) That is, that there should probably be no "implicit DM" sending - only initial explicit targeting, plus possibly if an entity is explicitly configured to point at a single user, then they already made that choice themselves.

henrikedegrd: I agree, you should only get a DM if you are explicitly targeted.

So that leaves my initial comment valid;

We only need to do this:
`if(!entityRefs.contains(normalizedUserRef)) { return; }
`

then would all derived users from other stuff be ignored, right?

And I would start by doing this change, and then maybe later allow some form of config for implicit DM:s to keep the changes separate.

However this will be a "breaking change" since currently users targeted implicitly via a component (or anything other than a group) gets a DM.

I have no idea how a change like this gets "voted" on? Haven't been in a situation before where a change of behavior is in play. But this change gets my "vote" :)

lucasoares:
Any help needed to make this ready? Really need this one haha

kaidubauskas-dd:
> Any help needed to make this ready? Really need this one haha

Sorry got caught up in other stuff. Should be ready now

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
kaidubauskas-dd: Changed to only DM user entities. Not sure on the breaking change process either. Im assuming a notifications maintainer would also have to weigh in and agree?

.changeset/curvy-socks-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-notifications-backend-module-slack': minor
+---
+
+**BREAKING**: Only send DMs to user entity recipients. Notifications sent to non-user entities no longer send Slack DMs to resolved users.
```
freben:
```suggestion
**BREAKING**: Only send direct messages to user entity recipients. Notifications sent to non-user entities no longer send Slack direct messages to resolved users.
```

freben (approved):

PR #32623, opened 2026-02-02
feat: queue service

### LINKED ISSUE #24116: 🚀 Feature: Core QueueService
### 🔖 Feature description

Backstage core could provide a new service for queuing items and consuming these queues. The service should allow different kinds of queues like FIFO, FILO, etc. 

### 🎤 Context

`QueueService` could be used in multiple places like scheduled tasks to consume for example entities to process. Additional use-cases are welcome in this issue.

The basic implementation could be just an in-memory queue but this could be replaced to use for example Redis, Kafka, database, etc. 

None

drodil:

new core service for queuing jobs. supports multiple adapters; kafka, memory, postgres, redis and AWS SQS. support for priority, delay, pausing and resuming processing and automatic retries.

added docs and mock service for testing.

closes #24116


packages/backend-defaults/src/entrypoints/queue/adapters/PostgresQueue.ts
aramissennyeydd, line 1: fmi: How does pg-boss handle migrations? I see reference to a CLI on their page, https://github.com/timgit/pg-boss#cli - is that the recommended way?

drodil: It does it on `start`: https://github.com/timgit/pg-boss/blob/master/src%2Findex.ts#L106

Have to double-check the args though

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,326 @@
+---
+id: queue
+title: Queue Service
+sidebar_label: Queue
+description: Documentation for the Queue service
+---
+
+This service lets your plugin execute background jobs. It provides a unified interface for different queue backends, allowing you to switch implementations without changing your plugin code.
```
awanlin: Doesn't the Scheduler service do this - "This service lets your plugin execute background jobs." - already? Could we reword this to make the benefits a bit more clear as to why you would use a queue?

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,326 @@
...
+
+### PostgreSQL
+
+The PostgreSQL adapter uses pg-boss for a reliable, database-backed queue with full feature support.
+
+```yaml
+backend:
+  queue:
+    store: postgres
+    # Optional: dedicated connection for queue
```
awanlin: We have a comment here but not for Redis above, we should probably make them the same

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,326 @@
...
+
+```yaml
+backend:
+  queue:
+    store: redis
+    connection: redis://localhost:6379
+```
+
+**Fallback to cache connection:**
+If `backend.queue.connection` is not specified, the queue will use the main database connection from `backend.cache.connection`.
```
awanlin: What is your thinking: is it likely that adopters of this service will run two instances of Redis or for PG two instances or databases? We should probably provide some guidance or recemendations.

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,326 @@
...
+
+```yaml
+backend:
+  queue:
+    defaultConcurrency: 5
+```
+
+## Retry and Error Handling
+
+Jobs automatically retry with exponential backoff when they fail.
```
awanlin: Add "backoff" to this list of words here to make Vale happy: https://github.com/backstage/backstage/blob/master/.github/vale/config/vocabularies/Backstage/accept.txt

awanlin (changes requested):
Awesome work @drodil, I just looked over the docs portion and a bit of code to help with context. Left a few comments on the docs 👍

docs/backend-system/core-services/queue.md
drodil: This is here mainly for flexibility. If users want to use different db or redis, it's possible. I think I saw similar approach in some other service 🤔

awanlin: That's fair, it's just the first thing that came to my mind was: why would I do that, should I do that? Having some sort of answer would help people decide what path to take.

docs/backend-system/core-services/queue.md
awanlin, line 1: You'll need to add this to the sidebar config here: https://github.com/backstage/backstage/blob/5d88631eff41f76d837b8652ba0e0cd5c183a4b5/microsite/sidebars.ts#L508

awanlin (changes requested):
Thanks for the updates @drodil, still have one open comment and one new one. 👍

docs/backend-system/core-services/queue.md
drodil: I removed the fallbacks completely, the connections need to be defined for redis and postgres.

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+processes on-demand jobs that are added dynamically. It provides a unified
+interface for different queue backends, allowing you to switch implementations
+without changing your plugin code.
+
+## Configuration
+
+The queue service can be configured using the `backend.queue` section
+in your `app-config.yaml`.
+
+### In-Memory (default)
```
Rugvip: Hmm, kinda feeling we need the default to be SQLite/PostgreSQL. A little bit because it makes local dev a bit nicer and simpler because we can lean on the SQLite restoration for persistence across reloads also for local dev, although there are other ways to go about that. But the big one is to default to PostgreSQL in production for this, because it's a very good queueing system and we should support it as the default imo.

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+
+```ts
+const workQueue = await queue.getQueue('my-work-queue', {
+  store: 'memory',
+});
+```
+
+**Best for:** Local development, testing
+
+### Redis
```
Rugvip, line 70: Hmm, I'm a bit torn on this tbh, is Redis even really a better queueing system than Postgres? 🤔 Will keep reading x)

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+- **Priorities**: Not supported. Kafka processes messages in order within each partition.
+- **Delays**: Not supported. Jobs are processed immediately.
+- **Concurrency**: Not configurable. Kafka processes messages sequentially per partition. To increase parallelism,
+  increase the number of partitions for the topic.
+- **Retries**: Supported, but retried messages are sent back to the end of the topic. This means retries are not
+  immediate and may be delayed if there are many other messages in the queue.
+
+### PostgreSQL
+
+The PostgreSQL adapter uses pg-boss for a reliable, database-backed queue
```
Rugvip: Hmm, bit torn on bringing in a dep for this honestly, because of the how it might not meld well with knex. Would hate to have this be a completely separately managed connection. Ideally we can lean on the existing per-plugin connections and schemas.

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+          task: 'send_email',
+          recipient: 'user@example.com',
+        });
+
+        // Add a job with options
+        await workQueue.add(
+          { task: 'cleanup' },
+          {
+            delay: 5000, // Wait 5 seconds before processing
+            priority: 50, // Lower priority than default (20)
```
Rugvip, line 179: Quick thought that came up reading this: I do think we should aim for something that could conceivably replace some of the catalog processing queues

.changeset/purple-bugs-strive.md
```diff
@@ -0,0 +1,49 @@
...
+  coreServices,
+  createBackendPlugin,
+} from '@backstage/backend-plugin-api';
+
+createBackendPlugin({
+  pluginId: 'example',
+  register(env) {
+    env.registerInit({
+      deps: {
+        queue: coreServices.queue,
```
Rugvip: Thinking we should start by shipping this as alpha from `@backstage/backend-plugin-api/alpha`

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+        // otherwise it will throw an error.
+        const redisQueue = await queue.getQueue('my-redis-queue', {
+          store: 'redis',
+        });
+
+        // Add a simple job to the queue
+        await workQueue.add({
+          task: 'send_email',
+          recipient: 'user@example.com',
+        });
```
Rugvip, line 172: Thinking there should likely be some way to upsert/remove items as well, i.e. key items in the queue. That does depend on a separate discussion though, so lemme bring that up here:

I'm been pondering what patterns a queue service needs to suport, and kinda landed on that it's either a consuming work item queue, with producers posting items and consumers removing them from the queue. Or it's move of a circular queue, producers ensure that the items they want are in the queue, and then consumers keep processing them indefinitely.

Having thought a little bit more on top of that, I _think_ we can achieve this with a single API. We just need to give producers enough control to be able to mutate existing items, and give consumers control over whether they actually want to remove items from the queue or not.

Qurious what your thoughts are on that whole thing.

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+        // Process jobs (with default concurrency of 1)
+        workQueue.process(async job => {
+          logger.info(`Processing job ${job.id}`, {
+            payload: job.payload,
+            attempt: job.attempt,
+          });
+
+          // Simulate work
+          await new Promise(resolve => setTimeout(resolve, 1000));
+        });
```
Rugvip, line 193: Wondering if we should aim for two levels of API here, one higher-level simple callback, and something lower level with more stateful direct access, more of a `const batch = workerQueue.process()`

I would be very interested in being able to provide relatively advanced options here around how many parallel tasks are allowed in-flight and potentially more global concerns across instances like target throughput rate or whatnot. Some of that definitely not being part of the initial implementation, but I would at the very least assume that we want to support concurrent processing on a single instance.

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+      },
+    });
+  },
+});
+```
+
+The queue will be paused until job handler is attached to the queue
+with `workQueue.process`.
```
Rugvip: This strikes me as obvious but also important, potentially worth reframing around the queue not draining by itself, it needs processors to consume the items in the queue.

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+
+await workQueue.add(
+  { task: 'background_job' },
+  { priority: 100 }, // Low priority (processed last)
+);
+```
+
+### Delay
+
+Delay the processing of a job by a specified number of milliseconds.
```
Rugvip, line 240: Out of curiousity, any specific use-case for this one? I see it fit with a potential catalog processing queue or scheduled notifications

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+
+Process multiple jobs simultaneously to improve throughput.
+
+```ts
+// Process up to 5 jobs concurrently
+workQueue.process(
+  async job => {
+    await processTask(job.payload);
+  },
+  { concurrency: 5 },
```
Rugvip, line 270: Aha, there's the throughput 😁, yep, just wanna make sure we have space for more advanced options down the line as well. Even things like load-based. Main thing to do down is imagine a bit of what those APIs might looks like and make sure it's a clean addition on top of this.

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+);
+
+// Default concurrency is 1 (sequential processing)
```
Rugvip, line 273: Immediate reaction is that we might as well make this higher, but maybe it's fine. It did trigger a thought though: a global concurrency may be an interesting thing, mostly setting it to 1 though ofc. Could imagine a future option that is something like `concurrency: 'globally-sequential'`. Need to think a bit more about whether that'd actually be useful enough to be worth the effort though 😅

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+  await processTask(job.payload);
+});
+```
+
+You can define the default concurrency with config:
+
+```yaml
+backend:
+  queue:
+    defaultConcurrency: 5
```
Rugvip, line 284: Hmm, feels a bit strange to configure the default concurrency given how wildly the reaction to that is going to be from different implementations. I do think being able to configure the queues is an interesting idea though.

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+```
+
+## Retry and Error Handling
+
+Jobs automatically retry with exponential backoff when they fail.
```
Rugvip: So back in the queue but with an increasing delay? I could imagine wanting to be able to configure that behavior although not sure exactly how. Maybe there's a sensible default like this and then there's an `onError` callback that needs to return a decision or such

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+  // If this throws an error, the job will be retried
+  await riskyOperation(job.payload);
+
+  // Retry schedule (with default maxAttempts: 5):
+  // Attempt 1: Immediate
+  // Attempt 2: ~1 second delay
+  // Attempt 3: ~2 seconds delay
+  // Attempt 4: ~4 seconds delay
+  // Attempt 5: ~8 seconds delay
+  // After attempt 5: Send to DLQ handler (if configured)
```
Rugvip, line 343: Ah, yep this. Related to the above though, for something like the catalog use-case we'd just want to bump it back for processing and maybe record the error somewhere, no retries, no, DLQ (unless that's the record)

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+      attempts: job.attempt,
+      failedAt: new Date(),
+    });
+  },
+});
+```
+
+### Configure Maximum Attempts
+
+Set the maximum number of retry attempts in your configuration:
```
Rugvip, line 379: Another one that feels more like a per-queue thing, although maybe some aspects are global, like if you even care to DLQ

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+
+## Queue Management
+
+Control queue processing with pause, resume, and disconnect operations.
+
+### Pause and Resume
+
+```ts
+// Pause processing (workers stop picking up new jobs)
+await workQueue.pause();
```
Rugvip: If nothing else we'll want this for controlled shutdown

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+logger.info('Queue paused');
+
+// Resume processing
+await workQueue.resume();
```
Rugvip: any immediate use-cases for this one? I could imagine a simple drain + stop being enough and then if you want to resume you can just recreate the queue

packages/backend-defaults/src/entrypoints/queue/adapters/BaseQueue.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.logger.debug(`[${this.queueName}] Queue disconnecting`);
+    await this.onDisconnect();
+    this.handler = undefined;
+  }
+
+  /**
+   * Handles a failed job by either retrying or sending to DLQ.
+   * Returns true if the job should be retried, false if it was sent to DLQ.
+   */
+  protected async handleFailedJob(
```
Rugvip, line 107: Generally gonna prefer a composition design with provided callbacks, but can hold off a bit on hat discussion to settle other things first

docs/backend-system/core-services/queue.md
```diff
@@ -0,0 +1,356 @@
...
+    sqs:
+      region: us-east-1
+      # Optional: custom endpoint for LocalStack
+      endpoint: http://localhost:4566
+```
+
+**Limitations:**
+
+- **Priorities**: Not supported. SQS does not natively support message priorities.
+- **Delays**: Supported up to 15 minutes (AWS SQS limitation).
```
Rugvip: Hmm, feeling that this is a pretty big problem actually. It means you can't rely on these primitives in the implementation. Or you do, and something ends up breaking. I do feel fairly strongly that we should make sure all implementations fully implement the service contract, other than shortcuts for local dev ofc.

Rugvip (changes requested):
Thank you! 🎉 This is exciting! 😁

One things to pinpoint is what we're aiming for. Bunch of comments on the docs with all my thoughts, think there's a fair amount to settle there before diving deep into the implementation.

I also think it's worth finding 1-2 initial consumers to include and ship into alpha pretty eagerly.

docs/backend-system/core-services/queue.md
freben: Likely the stitcher 👍

docs/backend-system/core-services/queue.md
drodil: Changed default to database which uses Knex now

docs/backend-system/core-services/queue.md
drodil: Removed the pg-boss dependency

docs/backend-system/core-services/queue.md
drodil: Stitcher now uses the queue service for deferred stitching

.changeset/purple-bugs-strive.md
drodil: Yep, moved

docs/backend-system/core-services/queue.md
drodil: Yep added lower level API as well

docs/backend-system/core-services/queue.md
drodil: Yep, fixed it by having the delay in the wrapper so the message can be sent again to the queue in case the delay > 15mins.

PR #32631, opened 2026-02-02
Introduce the `@backstage/filter-predicates` package

freben:
Filter predicates were first [introduced for entities](https://backstage.io/docs/features/software-catalog/catalog-customization/#entity-predicate-queries) in the new frontend system. That usage has now been cemented and the syntax has [started to spread](https://github.com/backstage/backstage/pull/31776) to more places than the initial one. Therefore, we are breaking out these facilities into a package of its own so that both frontend and backend packages can make use of them should they desire to.

### DIFF
--- a/packages/filter-predicates/src/predicates/evaluate.ts
+++ b/packages/filter-predicates/src/predicates/evaluate.ts
@@ -104,13 +104,13 @@ function evaluateFilterPredicateValue(
     }
     return value === undefined;
   }
-  if ('$startsWith' in filter) {
+  if ('$hasPrefix' in filter) {
     if (typeof value !== 'string') {
       return false;
     }
     return value
       .toLocaleUpperCase('en-US')
-      .startsWith(filter.$startsWith.toLocaleUpperCase('en-US'));
+      .startsWith(filter.$hasPrefix.toLocaleUpperCase('en-US'));
   }
 
   return false;
--- a/packages/filter-predicates/src/predicates/types.ts
+++ b/packages/filter-predicates/src/predicates/types.ts
@@ -153,7 +153,7 @@ export type FilterPredicateExpression = {
 export type UnknownFilterPredicateValueMatcher = {
   [KMatcher in `$${string}`]: JsonValue;
 } & {
-  [KMatcher in '$exists' | '$in' | '$contains' | '$startsWith']: never;
+  [KMatcher in '$exists' | '$in' | '$contains' | '$hasPrefix']: never;
 };
 
 /**
@@ -188,7 +188,7 @@ export type FilterPredicateValue =
       /**
        * Asserts that the property value is string, and that it starts with the given string.
        */
-      $startsWith: string;
+      $hasPrefix: string;
     }
   | UnknownFilterPredicateValueMatcher;
 
--- a/packages/filter-predicates/report.api.md
+++ b/packages/filter-predicates/report.api.md
@@ -61,7 +61,7 @@ export type FilterPredicateValue =
       $contains: FilterPredicate;
     }
   | {
-      $startsWith: string;
+      $hasPrefix: string;
     }
   | UnknownFilterPredicateValueMatcher;
 
@@ -102,6 +102,6 @@ export type UnknownFilterPredicateOperator = {
 export type UnknownFilterPredicateValueMatcher = {
   [KMatcher in `$${string}`]: JsonValue;
 } & {
-  [KMatcher in '$exists' | '$in' | '$contains' | '$startsWith']: never;
+  [KMatcher in '$exists' | '$in' | '$contains' | '$hasPrefix']: never;
 };
 ```
--- a/packages/filter-predicates/src/predicates/evaluate.test.ts
+++ b/packages/filter-predicates/src/predicates/evaluate.test.ts
@@ -237,7 +237,7 @@ describe('evaluate', () => {
         'metadata.annotations.github.com/repo': { $exists: true },
       },
     ],
-    ['a', { 'spec.type': { $startsWith: 'g' } }],
+    ['a', { 'spec.type': { $hasPrefix: 'g' } }],
   ])('filter entry %#', (expected, filter) => {
     it('filterPredicateToFilterFunction', () => {
       const filtered = entities.filter(entity =>


packages/filter-predicates/src/predicates/types.ts
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * @remarks
+ *
+ * A predicate is always an object at the root. The most basic use case is to
+ * declare keys that are dot-separated paths into a structured data value, and
+ * values that all need to match the corresponding property value in the data.
+ *
+ * The equality test is case-insensitive and numbers are converted to strings,
+ * i.e. `"Component"` and `"component"` are considered equal, and so are `7` and
+ * `"7"`.
```
freben, line 47: It can be argued that this is risky to set in stone. It's true for the current code, but is it true for all implementations? The in-memory matcher doesn't coerce booleans to strings, but the catalog backend search table does, for example. Should it all be "undocumented behaviour" for now? Will we end up having to make it settable by the caller as an options object (`caseInsensitive: true`) or is that taking it too far?

packages/filter-predicates/src/predicates/types.ts
```diff
@@ -0,0 +1,156 @@
...
+       * values.
+       */
+      $in: FilterPredicatePrimitive[];
+    }
+  | {
+      /**
+       * Asserts that the property value is an array, and that at least one of
+       * its elements matches the given predicate.
+       */
+      $contains: FilterPredicate;
```
freben, line 185: We already know that some implementations may not be able to support all matchers. The catalog will struggle to support a generic `$contains` - it will probably at most support a couple of specific variations of it (eg for relations matching), if even that. Does that mean that we need to somehow deconstruct these types and maker them composable in some more advanced way?

packages/filter-predicates/src/predicates/types.ts
```diff
@@ -0,0 +1,156 @@
...
+ * @public
+ */
+export type FilterPredicateValue =
+  | FilterPredicatePrimitive
+  | {
+      /**
+       * Asserts that the property exists and has any value (`true`) - or not
+       * (`false`).
+       */
+      $exists: boolean;
```
freben, line 171: Will it be easy enough to add more matchers? It would be very valuable to be able to have a `$startsWith` matcher for URLs in the catalog for example, as well as perhaps `$like` (but that one is more debatable). A consumer of this contract probably needs to be defensive at runtime and be able to gracefully handle unknown dollar-prefixed matchers and return false for them.

packages/predicates/report.api.md
```diff
@@ -0,0 +1,79 @@
...
+
+> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
+
+```ts
+import { Config } from '@backstage/config';
+import type { z } from 'zod/v3';
+import type { ZodType } from 'zod/v3';
+
+// @public
+export function createFilterPredicateSchema(
```
freben: Despite what [the docs say](https://zod.dev/library-authors), this currently only supports v3 zod. Making it support both v3 and v4 AND have a consistent return type is the hard bit. It's here because of the pre-existing `createEntityPredicateSchema`. How do you feel about this one? Should it be named `createZodV3FilterPredicateSchema` or something like that? Should we just leave it out of the contract entirely in the initial release, forcing us to keep this new package and the catalog-react one in sync?

packages/filter-predicates/src/predicates/types.ts
Rugvip: I think the biggest hurdle for that is support for all matches in the evaluation. I think at minimum we need to keep it to things that are supported in standard SQL, but maybe ideally JSON operations in Postgres?

freben: The json/jsonb restriction mainly comes down to the key selection right? Not so much the matchers, which should be the same once the value is extracted so to speak

packages/predicates/report.api.md
Rugvip: Hmm, yeah this is a bit of an issue in NFS as well. We may want to consider options here sooner rather than later. At the same time it's likely fine to keep around for a good while longer and it's designed to be possible to evolve by completely switching out at least.

The main blocker is the json schema handling I think. Speaking of that: https://zod.dev/json-schema?id=ztojsonschema Not entirely obvious to me if that supports both v3 and v4 schemas 🤔

packages/filter-predicates/src/predicates/types.ts
freben: I think my main question is - should we add more such matchers right away while we can and don't have to break the contract in the future

.changeset/wise-rabbits-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/predicates': minor
```
benjdlambert: How do we feel about making this `@backstage/filter-predicates` instead?

packages/filter-predicates/src/predicates/evaluate.ts
```diff
@@ -15,51 +15,48 @@
...
- * Evaluate a entity predicate against a value, typically an entity.
+ * Evaluate a filter predicate against a value.
  *
- * @internal
+ * @public
  */
-function evaluateEntityPredicate(
-  filter: EntityPredicate,
-  value: JsonValue,
+export function evaluateFilterPredicate(
```
benjdlambert, line 26: Should we maybe have a depth limit here? Maybe not blocking

benjdlambert (commented):
Just a question about the package name really? :pray:

.changeset/wise-rabbits-double.md
freben: Done!

packages/filter-predicates/src/predicates/evaluate.ts
freben: You mean if like a malicious actor injects an extremely deep predicate? I think that's the only thing we'd have to defend against then, since the depth of the target data itself doesn't matter I think. 🤔

packages/filter-predicates/src/predicates/types.ts
freben: Fixed by adding `UnknownFilterPredicateValueMatcher` and also `$startsWith`

packages/filter-predicates/src/predicates/types.ts
```diff
@@ -0,0 +1,200 @@
...
+       */
+      $contains: FilterPredicate;
+    }
+  | {
+      /**
+       * Asserts that the property value is string, and that it starts with the given string.
+       */
+      $startsWith: string;
+    }
+  | UnknownFilterPredicateValueMatcher;
```
Rugvip, line 193: A downside of this approach is that you don't get type errors when using invalid matchers. You do at least get correct autocomplete though, so it's not the end of the work. I guess worth a double check if you think there'd be an alright way to split these into input/output types? It needs to be very clear when to use which one though, and ideally you'd probably force producers to always parse predicate predicates they are provided.

packages/filter-predicates/src/predicates/types.ts
```diff
@@ -0,0 +1,200 @@
...
+       * Asserts that the property value is an array, and that at least one of
+       * its elements matches the given predicate.
+       */
+      $contains: FilterPredicate;
+    }
+  | {
+      /**
+       * Asserts that the property value is string, and that it starts with the given string.
+       */
+      $startsWith: string;
```
Rugvip: I wonder if this one should be distinguished from the rest in some way given that it only operates on strings? It's a bit awkward to have `$contains` and `$startsWith` next to each other, but meaning very different things. Was thinking for example the obvious `$stringStartsWith`, opening up for `$stringContains` in the future. Or maybe `$textStartsWith`. Another slightly weirder one would be `$hasPrefix`, `$hasInfix`, `$hasSuffix`.

Rugvip (approved):
Nice 🎉 👌 

A Q around the future compatibility hurting DX a tiny bit, but no biggie, contract is still there. Most important thing stability-wise is also that the input remains stable.

I think my biggest concern is `$startsWith` vs `$contains` and that we might lock ourselves in a weird spot.

packages/filter-predicates/src/predicates/types.ts
freben: Changed to `hasPrefix` - I really like that. Can make add `hasSubstring` down the line if the infix word is too fancy/technical.

This all makes me think that `contains` really should be `containsArrayElement` or `hasArrayItem` or so.

Issue #32595, opened 2026-01-30
CatalogIndexPage errors if filters list for owners is too large


depopry:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
Backstage "Cannot read properties of undefined (reading 'toLocaleLowerCase')"

Backstage CatalogIndexPage Autocomplete errors

ChatGPT prompts on errors
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

N/A

### 📝 Description & Context

We have an organisation hierarchy for a total of 51 teams (defined as kind: Group, type: team) 

In the team entity page, we utilise an `EntityLinksCard` to display how many services a team or hierarchy of teams owns

When clicking on the link to services (which routes through to /catalog (CatalogIndexPage)) parent team for the hierarchy (named all-teams) the UI presents an error

```
Error
TypeError

Message
Cannot read properties of undefined (reading 'toLocaleLowerCase')

Stack Trace
TypeError: Cannot read properties of undefined (reading 'toLocaleLowerCase')
    at humanizeEntityRef (webpack-internal:///../../node_modules/@backstage/plugin-catalog/node_modules/@backstage/plugin-catalog-react/dist/components/EntityRefLink/humanize.esm.js:36:15)
    at getOptionLabel (webpack-internal:///../../node_modules/@backstage/plugin-catalog/node_modules/@backstage/plugin-catalog-react/dist/components/EntityOwnerPicker/EntityOwnerPicker.esm.js:175:162)
    at eval (webpack-internal:///../../node_modules/@material-ui/lab/esm/Autocomplete/Autocomplete.js:446:18)
    at Array.map (<anonymous>)
    at Autocomplete (webpack-internal:///../../node_modules/@material-ui/lab/esm/Autocomplete/Autocomplete.js:444:30)
    at renderWithHooks (webpack-internal:///../../node_modules/react-dom/cjs/react-dom.development.js:15486:18)
    at updateForwardRef (webpack-internal:///../../node_modules/react-dom/cjs/react-dom.development.js:19240:20)
    at beginWork (webpack-internal:///../../node_modules/react-dom/cjs/react-dom.development.js:21670:16)
    at beginWork$1 (webpack-internal:///../../node_modules/react-dom/cjs/react-dom.development.js:27460:14)
[...]

PR #32637, opened 2026-02-03
fix(catalog): resolve owned-by filter using POST endpoint to fix 431 when users have many groups

### LINKED ISSUE #32367: `OwnedEntityPicker` 431 error with many groups and catalogFilter entries in scaffolder template
### 📜 Issue Labels

### 🔎 Search Terms

```plain
OwnedEntityPicker, 431 error, 431 error in scaffolder,
```

### 🗃️ Project Area

Software Templates

### 🔗 External Integration

N/A

### 📝 Description & Context

When using the `OwnedEntityPicker` field in a Backstage scaffolder template, users who belong to many groups (35+ groups with long names) encounter a 431 "Request Header Fields Too Large" HTTP error when the template contains multiple catalogFilter entries.

Image

### 👍 Expected Behavior

The `OwnedEntityPicker` should work seamlessly regardless of the number of groups a user belongs to or the number of catalogFilter entries in the template.

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

**Steps to reproduce:**

1. Create a user that belongs to 35+ groups with long names (e.g., group-001-abcdefghijklmnopqrstuv through group-036-abcdefghijklmnopqrstuv)

2. Configure authentication to include all group membership in the user's ownedEntityRefs:

guest provider:
```yaml

auth:
  providers:
    guest:
      userEntityRef: user:default/testuser
      ownershipEntityRefs:
        - user:default/testuser
        - group:default/group-001-abcdefghijklmnopqrstuv
        - group:default/group-002-abcdefghijklmnopqrstuv
        # include all 35+ groups
```
3. Create a scaffolder template with `OwnedEntityPicker` field containing multiple catalogFilter entries:  
```yaml
properties:
  ownedEntity:
     title: Entities
     type: string
     "ui:field": OwnedEntityPicker
     "ui:options":
       catalogFilter:
       - kind: component
         metadata.namespace: system1
       - kind: component
         metadata.namespace: system2
       - kind: component
         metadata.namespace: system3
         # add more filters if needed
```

4. Access the scaffolder template as the user with many groups

5. The OwnedEntityPicker will fail with a 431 error when trying to fetch entities. (see Network tab).

karthikjeeyar:
## Description

     
Scaffolder entity pickers that load catalog data now call `catalogApi.queryEntities` with a query predicate instead of a traditional filter. That matches how CatalogClient uses the POST `/entities/by-query` path when a predicate query is present, so large template-defined filters are sent in the request body instead of as long GET query strings.

## Screenshot

image

---

## **How to test?:**

1. Create a user that belongs to 35+ groups with long names (e.g., group-001-abcdefghijklmnopqrstuv through group-036-abcdefghijklmnopqrstuv)

2. Configure authentication to include all group membership in the user's ownedEntityRefs:

guest provider:
```yaml

auth:
  providers:
    guest:
      userEntityRef: user:default/testuser
      ownershipEntityRefs:
        - user:default/testuser
        - group:default/group-001-abcdefghijklmnopqrstuv
        - group:default/group-002-abcdefghijklmnopqrstuv
        # include all 35+ groups
```
3. Create a scaffolder template with `OwnedEntityPicker` field containing multiple catalogFilter entries:  
```yaml
properties:
  ownedEntity:
     title: Entities
     type: string
     "ui:field": OwnedEntityPicker
     "ui:options":
       catalogFilter:
          - kind: component
          - kind: api
          - kind: template
          - kind: system
          - kind: template 
          - kind: group
         # add more filters if needed
```

4. Access the scaffolder template as the user with many groups

5. The OwnedEntityPicker should list all the components/api/templates as defined in the catalogFilter

### DIFF
--- a/plugins/scaffolder/src/components/fields/EntityPicker/EntityPicker.tsx
+++ b/plugins/scaffolder/src/components/fields/EntityPicker/EntityPicker.tsx
@@ -90,11 +90,13 @@ export const EntityPicker = (props: EntityPickerProps) => {
       'spec.profile.displayName',
       'spec.type',
     ];
-    const { items } = await catalogApi.getEntities(
-      catalogFilter
-        ? { filter: catalogFilter, fields }
-        : { filter: undefined, fields },
-    );
+    const streamRequest = catalogFilter
+      ? { query: {}, filter: catalogFilter, fields }
+      : { fields };
+    const items: Entity[] = [];
+    for await (const batch of catalogApi.streamEntities(streamRequest)) {
+      items.push(...batch);
+    }
 
     const entityRefToPresentation = new Map<
       string,
--- a/.changeset/catalog-query-post-filters.md
+++ b/.changeset/catalog-query-post-filters.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder': patch
+---
+
+Improves scaffolder entity pickers by using the catalog POST endpoint so large template filters are sent in the request body instead of the URL, helping avoid 431 errors and empty option lists.
--- a/plugins/scaffolder/src/components/fields/EntityPicker/EntityPicker.test.tsx
+++ b/plugins/scaffolder/src/components/fields/EntityPicker/EntityPicker.test.tsx
@@ -52,7 +52,9 @@ describe('<EntityPicker />', () => {
   let props: FieldProps<string>;
 
   const catalogApi = catalogApiMock.mock({
-    getEntities: jest.fn(async () => ({ items: entities })),
+    streamEntities: jest.fn(async function* () {
+      yield entities;
+    }),
   });
 
   let Wrapper: ComponentType<PropsWithChildren<{}>>;
@@ -95,7 +97,7 @@ describe('<EntityPicker />', () => {
         </Wrapper>,
       );
 
-      expect(catalogApi.getEntities).toHaveBeenCalledWith({
+      expect(catalogApi.streamEntities).toHaveBeenCalledWith({
         fields: [
           'kind',
           'metadata.name',
@@ -105,7 +107,6 @@ describe('<EntityPicker />', () => {
           'spec.profile.displayName',
           'spec.type',
         ],
-        filter: undefined,
       });
     });
 
@@ -137,7 +138,9 @@ describe('<EntityPicker />', () => {
         formData,
       } as unknown as FieldProps<any>;
 
-      catalogApi.getEntities.mockResolvedValue({ items: entities });
+      catalogApi.streamEntities.mockImplementation(async function* () {
+        yield entities;
+      });
     });
 
     it('searches for users and groups', async () => {
@@ -147,11 +150,10 @@ describe('<EntityPicker />', () => {
         </Wrapper>,
       );
 
-      expect(catalogApi.getEntities).toHaveBeenCalledWith(
+      expect(catalogApi.streamEntities).toHaveBeenCalledWith(
         expect.objectContaining({
-          filter: {
-            kind: ['User'],
-          },
+          query: {},
+          filter: { kind: ['User'] },
         }),
       );
     });
@@ -182,7 +184,9 @@ describe('<EntityPicker />', () => {
         formData,
       } as unknown as FieldProps<any>;
 
-      catalogApi.getEntities.mockResolvedValue({ items: entities });
+      catalogApi.streamEntities.mockImplementation(async function* () {
+        yield entities;
+      });
     });
 
     it('searches for a specific group entity', async () => {
@@ -192,8 +196,9 @@ describe('<EntityPicker />', () => {
         </Wrapper>,
       );
 
-      expect(catalogApi.getEntities).toHaveBeenCalledWith(
+      expect(catalogApi.streamEntities).toHaveBeenCalledWith(
         expect.objectContaining({
+          query: {},
           filter: [
             {
               kind: ['Group'],
@@ -217,7 +222,9 @@ describe('<EntityPicker />', () => {
         },
       };
 
-      catalogApi.getEntities.mockResolvedValue({ items: entities });
+      catalogApi.streamEntities.mockImplementation(async function* () {
+        yield entities;
+      });
 
       await renderInTestApp(
         <Wrapper>
@@ -225,8 +232,9 @@ describe('<EntityPicker />', () => {
         </Wrapper>,
       );
 
-      expect(catalogApi.getEntities).toHaveBeenCalledWith(
+      expect(catalogApi.streamEntities).toHaveBeenCalledWith(
         expect.objectContaining({
+          query: {},
           filter: {
             kind: ['Group'],
             'metadata.name': 'test-entity',
@@ -253,8 +261,9 @@ describe('<EntityPicker />', () => {
         </Wrapper>,
       );
 
-      expect(catalogApi.getEntities).toHaveBeenCalledWith(
+      expect(catalogApi.streamEntities).toHaveBeenCalledWith(
         expect.objectContaining({
+          query: {},
           filter: [
             {
               kind: ['User'],
@@ -291,7 +300,9 @@ describe('<EntityPicker />', () => {
         formData,
       } as unknown as FieldProps<any>;
 
-      catalogApi.getEntities.mockResolvedValue({ items: entities });
+      catalogApi.streamEntities.mockImplementation(async function* () {
+        yield entities;
+      });
     });
     it('Prevents user from modifying input when ui:disabled is true', async () => {
       props.uiSchema = { 'ui:disabled': true };
@@ -355,7 +366,9 @@ describe('<EntityPicker />', () => {
         formData,
       } as unknown as FieldProps<any>;
 
-      catalogApi.getEntities.mockResolvedValue({ items: entities });
+      catalogApi.streamEntities.mockImplementation(async function* () {
+        yield entities;
+      });
     });
 
     it('searches for a Group entity', async () => {
@@ -365,8 +378,9 @@ describe('<EntityPicker />', () => {
         </Wrapper>,
       );
 
-      expect(catalogApi.getEntities).toHaveBeenCalledWith(
+      expect(catalogApi.streamEntities).toHaveBeenCalledWith(
         expect.objectContaining({
+          query: {},
           filter: [
             {
               kind: ['Group'],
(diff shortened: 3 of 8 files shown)


Rugvip (changes requested):
Nope this is too much code for something that likely has a simpler fix

karthikjeeyar:
@Rugvip I went ahead with the idea of a special catalog filter using UserInfoService in the backend to resolve the user + groups instead of sending them via frontend as discussed in this issue https://github.com/backstage/backstage/issues/32367.  could you briefly describe the simpler fix you’re thinking of? I’m happy to rework the PR in that direction.

Rugvip (changes requested):
Ah sorry I came here from #32595 thinking it was a fix for that. I had a better look and I think this couples the catalog data model and filtering a bit too much. I wonder if a middle ground might be to make it an additional query parameter for the `query-by` endpoint etc.? although I'll admit I haven't thought through the impact of that either

plugins/scaffolder/src/components/fields/EntityPicker/EntityPicker.tsx
```diff
@@ -202,13 +202,15 @@ export const EntityPicker = (props: EntityPickerProps) => {
         loading={loading}
         onChange={onSelect}
         options={entities?.catalogEntities || []}
-        getOptionLabel={option =>
+        getOptionLabel={option => {
+          if (option === undefined || option === null || Array.isArray(option))
+            return '';
```
karthikjeeyar: I have added this check to prevent scaffolder UI from crashing due to `option` resolves here as an array and throwing the following error

```
Cannot read properties of undefined (reading 'toLocaleLowerCase')
ref.ts:157 Uncaught TypeError: Cannot read properties of undefined (reading 'toLocaleLowerCase')
    at stringifyEntityRef (ref.ts:157:1)
    at getOptionLabel (EntityPicker.tsx:207:1)
```


This error is similar to the one described here in this issue https://github.com/backstage/backstage/issues/32595 but this happens on the **scaffolder** plugin.


@Rugvip I see you have closed the PR linked to the above issue, we may want to fix that in the catalog-react as well, right?

packages/catalog-client/src/types/api.ts
```diff
@@ -176,6 +176,12 @@ export interface GetEntitiesRequest {
    * request.
    */
   after?: string;
+  /**
+   * When true, restrict results to entities owned by the current user (user +
+   * groups from the request identity). Avoids sending large ownership ref
+   * lists in the filter.
+   */
+  ownedByCurrentUser?: boolean;
```
vinzscam: I'm not sure about this, as this technically belongs to the filters. I would prefer to define something compatible with `EntityFilterQuery ` that can be embedded inside the existing `filter`
cc @backstage/catalog-maintainers

vinzscam (changes requested):
thanks this is great!

packages/catalog-client/src/types/api.ts
karthikjeeyar: Thanks for the review @vinzscam. My first attempt was to keep everything in the filter by using a reserved value, e.g. `filter: { 'relations.ownedBy': __current__user__ }`, and having the backend replace that with the actual ownershipEntityRefs from the userInfoService. After Patrick’s comment I switched to a query parameter (ownedByCurrentUser) approach.


Currently, Proposed approach is to align with "compatible with `EntityFilterQuery`  that can be embedded inside the existing filter":
- no change to the `EntityFilterQuery` type.
- Treat **ownedByCurrentUser: 'true'** as a reserved key in the same filter object (**string 'true'** so it stays within the existing EntityFilterQuery value types).
- Send it in the same filter query param (e.g. filter=kind=Component,ownedByCurrentUser=true); no separate query param.
- **Backend**: when parsing filter, detect and strip `ownedByCurrentUser=true`, resolve current user’s refs via UserInfoService, and AND a `relations.ownedBy` filter with the rest.

Does this match what you had in mind before I implement it?

drodil: I am in favor to introduce new magic constant for filters instead, just like the `CATALOG_FILTER_EXISTS` (https://github.com/backstage/backstage/blob/master/packages%2Fcatalog-client%2Fsrc%2FtestUtils%2FInMemoryCatalogClient.ts). Or actually two; `CATALOG_FILTER_CURRENT_USER_REF` and `CATALOG_FILTER_CURRENT_USER_OWNERSHIP_REFS`. 

With these you cand send it in the filters like:

```
filters: {
  'spec.owner': CATALOG_FILTER_CURRENT_USER_OWNERSHIP_REFS
}
```

This way it would be possible to filter also based on other fields using these instead of just the owner.

freben: @drodil Remember not to match on spec, but on relations. I know you know, but wanted to mention for those reading along.

Regarding filter syntax - after https://github.com/backstage/backstage/pull/32282 maybe we could introduce `{ $identity: 'ownership' }` (and `'user'`) as a matcher or something :) But at that point it's a bit less of a problem since we aren't bound by url length limits.

karthikjeeyar: I have added the special filter constants which will be resolved in the backend.



I also added a special object schema `{ currentUser: true }` (similar to `{exists: true }` ) for the software template authors. we can now pass this object in the `catalogFilter` and it will be resolved to the filter constants based on the key (`relations.ownedBy -> CATALOG_FILTER_CURRENT_USER_OWNERSHIP_REFS`)

**Filters the entity owned by current user:** 
```yaml
catalogFilter:
  - kind: component
    spec.owner: { currentUser: true } // this will be resolved as single user ref
```

**Filters the entity by ownership:** 

```yaml
catalogFilter:
  - kind: component
    relations.ownedBy: { currentUser: true } // this will be resolved as user + group ref in the backend
```
 the frontend converts this object based on it's key (i.e `relations.ownedBy` to the special filter constant) and the backend resolves it to the current user’s refs.

freben (changes requested):
We're revamping the query format in https://github.com/backstage/backstage/pull/32874 which switches to POST - and I think we'll let that be the solution instead.

karthikjeeyar:
@freben Thanks for the feedback. I noticed `EntityPicker` and `OwnedEntityPicker` use `getEntities()`, which always uses GET with the filter in the URL, so they never hit the POST endpoint. Should I switch them to use `queryEntities` with a predicate to fix the 431?

karthikjeeyar:
@freben @vinzscam I have updated the code to use `queryEntities` with a query predicate, so requests use POST `entities/by-query` instead of long URLs. Please take a look and let me know if there are any changes needed. Thanks.

plugins/scaffolder/src/utils/convertFilterToPredicate.ts
```diff
@@ -0,0 +1,64 @@
...
+} from '@backstage/catalog-client';
+
+/**
+ * Converts an {@link EntityFilterQuery} into a predicate for use with the
+ * `query` option of `CatalogApi.queryEntities` (POST `/entities/by-query`).
+ *
+ * Mirrors the internal `convertFilterToPredicate` in `@backstage/catalog-client`
+ * so the scaffolder can build predicate queries without relying on non-exported APIs.
+ */
+export function convertFilterToPredicate(
```
freben: Not needed; `queryEntities` accepts a `filter` and does any needed conversion internally

plugins/scaffolder/src/utils/queryEntitiesAllPages.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Loads all entities matching an initial {@link CatalogApi.queryEntities} request
+ * by following `pageInfo.nextCursor` until exhausted.
+ *
+ * `queryEntities` applies a default page size when `limit` is omitted (see catalog
+ * backend), unlike `getEntities` which returned the full result set when no limit
+ * was set. Callers that need the full list (for example template pickers) should
+ * use this helper instead of a single `queryEntities` call without pagination.
+ */
+export async function queryEntitiesAllPages(
```
freben: Not needed; the client has a `streamEntities` (and you can use `Array.fromAsync` on that if you need specifically an array)

plugins/scaffolder/package.json
```diff
@@ -63,6 +63,7 @@
     "@backstage/core-components": "workspace:^",
     "@backstage/core-plugin-api": "workspace:^",
     "@backstage/errors": "workspace:^",
+    "@backstage/filter-predicates": "workspace:^",
```
freben: if you don't do the conversion by hand you can remove this too

freben (changes requested):

plugins/scaffolder/src/utils/convertFilterToPredicate.ts
karthikjeeyar: @freben yes , but the catch is, `queryEntities` will route the request to POST endpoint only if the request contains `query` predicate in it [here](https://github.com/backstage/backstage/blob/master/packages/catalog-client/src/CatalogClient.ts#L304-L307). Although adding a dummy query like this `{ query : {}, filter }`  will route to the POST endpoint and filter will be handled, but I wasn't sure if that is acceptable solution, hence added the full conversion function here.

 Let me know your thoughts on routing to POST endpoint by using `query : {}` along with the traditional filter, I am happy to modify the code.

freben: It's worth it I think!

I also started looking into whether this could be automated https://github.com/backstage/backstage/pull/33988 - it won't land until earliest during next week if the other maintainers agree, but maybe that would be the ultimate solution in the end

plugins/scaffolder/src/utils/queryEntitiesAllPages.ts
karthikjeeyar: Switched to use `streamEntities`

Edit: Seems like `Array.fromAsync` isn't included in repo defaults, so I had switched to use `for await...of` for now. 

Refer Github copilot [review](https://github.com/backstage/backstage/pull/32637#discussion_r3108502837):

plugins/scaffolder/src/utils/convertFilterToPredicate.ts
karthikjeeyar: Updated it.

karthikjeeyar:
@freben I have addressed all your comments, please take a look when you find time. let me know if there’s anything else you need from my side to get this ready for merge!

karthikjeeyar:
Hey @freben, just a gentle nudge on this when you have a moment. All the requested changes have been applied. Looking forward to your thoughts!

freben (approved):
Thanks! Let's 🚢 

For anyone reading along: this is essentially a bit of a hack that relies on queryEntities doing a certain type of `if` internally where it chooses the POST path. But that's OK by me - eventually we'll follow up by making the client itself always use the POST path no matter what (to benefit all consumers), and at that point the hack becomes unnecessary but entirely harmless. So I say let's go ahead

PR #32656, opened 2026-02-03
fix: Update UserListPicker to use kind from filters instead of queryParams

sanchez-boop:

The `UserListPicker` has a check to only display the filters that apply to certain kinds. For example, in the User and Group kinds, the 'Owned' filter should not appear as a user cannot own a group or a person. This check works as expected when the page is first loaded. When switching the kind in the same session, the `UserListPicker` does not update to include/exclude 'owned'  depending on the kind. 

##### Reproduction steps: 
1. visit the [backstage demo catalog table](https://demo.backstage.io/catalog?filters%5Bkind%5D=component&filters%5Blifecycles%5D=production&filters%5Blifecycles%5D=experimental&filters%5Buser%5D=all)
2. select either `user` or `group` from the kind picker 
3. The UserListPicker does not update to exclude 'owned' 

note: this also happens the other way around. If the first kind is [user](https://demo.backstage.io/catalog?filters%5Bkind%5D=user&filters%5Buser%5D=all), selecting another kind in the kind picker will not show 'owned' in the UserListPicker

##### The fix:
Read the kind from the filters instead of the queryParams as filters updates when the kind changes in the picker. 

### DIFF
--- a/plugins/catalog-react/src/components/UserListPicker/UserListPicker.test.tsx
+++ b/plugins/catalog-react/src/components/UserListPicker/UserListPicker.test.tsx
@@ -708,7 +708,9 @@ describe('<UserListPicker />', () => {
 
         const updateFilters = (update: any) =>
           _setFiltersState((prev: any) =>
-            typeof update === 'function' ? update(prev) : { ...prev, ...update },
+            typeof update === 'function'
+              ? update(prev)
+              : { ...prev, ...update },
           );
 
         const value = {


sanchez-boop:
> Hmm, I'm following the reproduction steps but I don't really see anything wrong? The owned option is disabled, and even if I select owned while browsing for example components, if I then switch to groups it resets and show all of them?

Thanks for the quick reply! This PR isn't about the 'owned' option being disabled, that works as expected! This PR focuses on the 'owned' option being displayed. 

It is hard to explain in writing without showing the actual flow, let me try explaining it starting at the user page instead...

On intial load of the [User catalog page](https://demo.backstage.io/catalog?filters%5Bkind%5D=user&filters%5Buser%5D=all), the UserListPicker does not show an 'owned' option. This is intended as a user cannot be owned. 

Screenshot 2026-02-03 at 3 37 25 PM

On this same page, if you switch the kind to component, the 'owned' option does not get displayed. This is not intended behavior as a user can own components.
Screenshot 2026-02-03 at 3 39 06 PM

In my PR Summary, I explained it the other way around, starting at the component catalog page with the owned option, and switching the kind to user, where the owned option should not be, but remains. This PR makes sure that the 'owned' option renders or doesn't render when switching kinds, not just on initial load. Hopefully how I explain it in this comment makes more sense and shows the breaking a little more.

Rugvip (approved):
Ah alright, got it, thanks! 👍 

Was missing the context on the owned option not being rendered when first loading the page

PR #32657, opened 2026-02-03
frontend-test-utils: add mountedRoutes option for renderTestApp + docs

Rugvip:

🧹, closing some gaps in the API

### DIFF
--- a/packages/frontend-test-utils/src/app/renderTestApp.tsx
+++ b/packages/frontend-test-utils/src/app/renderTestApp.tsx
@@ -14,14 +14,17 @@
  * limitations under the License.
  */
 
+import { Fragment } from 'react';
 import { createSpecializedApp } from '@backstage/frontend-app-api';
 import {
   coreExtensionData,
   createApiFactory,
+  createExtension,
   createFrontendModule,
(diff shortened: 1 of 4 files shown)


packages/frontend-test-utils/src/app/renderTestApp.tsx
```diff
@@ -100,6 +120,28 @@ export function renderTestApp<TApiPairs extends any[] = any[]>(
...
+          kind: 'test-route',
+          name: path,
+          attachTo: { id: 'app/routes', input: 'routes' },
+          output: [
+            coreExtensionData.reactElement,
+            coreExtensionData.routePath,
+            coreExtensionData.routeRef,
+          ],
+          factory: () => [
+            coreExtensionData.reactElement(<Fragment />),
```
freben, line 136: Is it useful to only ever render something empty? It gets rid of errors, but does it further testability enough?

Rugvip: generally you'd just verify the link `href` rather than actually navigating to it, so I think it's alright

PR #32685, opened 2026-02-04
[BUI] Allow a 'data' prop to be passed inplace to the useTable hook

grantila:

When using the BUI `Table` component with its `useTable` hook in mode `"complete"`, it won't update the table when data changes. It cannot be overridden by changing the `getData` function either, as it's stabilized (basically memoed without dependencies) internally.

This has three drawbacks:
 * It is rather unintuitive, and is a bit against how most React components and hooks work - you pass new data, the component/hook reflects this.
 * For updating on data changes, the user needs a manual `useEffect` depending on the data, calling a `reload`, a bit cluttered.
 * The `useEffect` solution causes an internal async function to update state. In the example of the [Manage page](https://github.com/backstage/community-plugins/tree/main/workspaces/manage), when navigating between different entity kinds (System, Component, API), the page will update properly, but the table remains with the old data, then after a short while updates to the new list of entities for the table. It causes some flickering.

This PR changes the API for `useTable` when using the mode `"complete"` and allows _either_ a `getData` _or_ a `data` prop with the list of items. It's a simpler API, no need for `useEffect`, and updates immediately on `data` changes.

### DIFF
--- a/packages/ui/src/components/Table/hooks/useCompletePagination.ts
+++ b/packages/ui/src/components/Table/hooks/useCompletePagination.ts
@@ -54,7 +54,7 @@ export function useCompletePagination<T extends TableItem, TFilter>(
   // Load data on mount and when loadCount changes (reload trigger)
   useEffect(() => {
     if (data) {
-      return undefined;
+      return;
     }
 
     let cancelled = false;


.changeset/long-months-laugh.md
```diff
@@ -0,0 +1,9 @@
+---
+'@backstage/ui': patch
+---
+
+Allow data to be passed inplace to the `useTable` hook using the property `data` instead of `getData()` for mode `"complete"`.
```
Naycon:
```suggestion
Allow data to be passed directly to the `useTable` hook using the property `data` instead of `getData()` for mode `"complete"`.
```

PR #32686, opened 2026-02-04
[BUI] Allow automatic Table column widths and styling TableRoot

grantila:

This PR:

 * Adds `tableLayout="auto"` to the `Table` component, which makes the column widths adapt to the content. The `"auto"`value also disables the `ResizableTableContainer` wrapper, as this conflicts with `table-layout: auto`.
 * Adds a `styles={{ tableRoot: { /* custom CSS */ }, }}` prop to the `Table` component, to be able to style the inner `TableRoot` component (i.e. the `<table>` HTML element).

Automatic column widths work really well with `react-aria-components`, and is used by the [Manage page](https://github.com/backstage/community-plugins/tree/main/workspaces/manage) (in the [upcoming BUI PR](https://github.com/backstage/community-plugins/pull/6874)).

For big data tables, it's sometimes (certainly not always) useful to have a container `Box` and allow the table to grow wider than the container. Being able to customize the `<table>` CSS attributes is necessary for this. This `styles` prop could be extended with more internal components if necessary.

### DIFF
--- a/packages/ui/src/components/Table/types.ts
+++ b/packages/ui/src/components/Table/types.ts
@@ -134,6 +134,5 @@ export interface TableProps<T extends TableItem> {
   selection?: TableSelection;
   emptyState?: ReactNode;
   className?: string;
-  columnSizing?: 'content' | 'manual';
   style?: React.CSSProperties;
 }
--- a/.changeset/thirty-mirrors-own.md
+++ b/.changeset/thirty-mirrors-own.md
@@ -2,6 +2,6 @@
 '@backstage/ui': patch
 ---
 
-The Table component now defaults to automatically determine the columns widths based on the content, and the prop `columnSizing` can be set to `manual` to disable this. If this prop is unset and any column has a width property set, it turns to `manual` column sizing mode.
+The Table component now wraps the react-aria-components `Table` with a `ResizableTableContainer` only if any column has a width property set. This means that column widths can adapt to the content otherwise (if no width is explicitly set).
 
 Affected components: Table
--- a/docs-ui/src/app/components/table/props-definition.tsx
+++ b/docs-ui/src/app/components/table/props-definition.tsx
@@ -230,13 +230,6 @@ export const tablePropDefs: Record<string, PropDef> = {
     values: ['ReactNode'],
     description: 'Content to display when the table has no data.',
   },
-  columnSizing: {
-    type: 'enum',
-    values: ['content', 'manual'],
-    default: 'content',
-    description:
-      'Set to "manual" (or define any of the width properties on a column) to manually control column widths.',
-  },
   ...classNamePropDefs,
   ...stylePropDefs,
 };
--- a/packages/ui/report.api.md
+++ b/packages/ui/report.api.md
@@ -1700,7 +1700,6 @@ export function Table<T extends TableItem>({
   selection,
   emptyState,
   className,
-  columnSizing,
   style,
 }: TableProps<T>): JSX_2.Element;
 
@@ -1812,8 +1811,6 @@ export interface TableProps<T extends TableItem> {
   className?: string;
   // (undocumented)
   columnConfig: readonly ColumnConfig<T>[];
-  // (undocumented)
-  columnSizing?: 'content' | 'manual';
   // (undocumented)
   data: T[] | undefined;
   // (undocumented)
--- a/packages/ui/src/components/Table/components/Table.tsx
+++ b/packages/ui/src/components/Table/components/Table.tsx
@@ -29,7 +29,7 @@ import type {
   RowRenderFn,
   TablePaginationType,
 } from '../types';
-import { CSSProperties, useMemo } from 'react';
+import { useMemo } from 'react';
 import { VisuallyHidden } from '../../VisuallyHidden';
 import { Flex } from '../../Flex';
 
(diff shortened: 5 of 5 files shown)


grantila:
The CSS styling of the inner `TableRoot` (the HTML `<table>`) could also be done through CSS modules:

```css
.myTableComponent > table {
  // css for the TableRoot goes here
}
```

I'm not sure which one you prefer:
 1. Having CSSProperties as a `styles` objects like in this PR, or
 2. Mimicking the MUI `classes`, which is similar to 1, but requires CSS modules, or
 3. Only having a className for the top-level component and let people do the above CSS module handling with CSS selectors. This depends on the exact implementation of the BUI Table, what if there's a div in-between the Table and TableRoot - the above CSS wouldn't work.

I can adapt the PR to any of the above. The `tableLayout="auto"` is what's most important in this PR.

packages/ui/src/components/Table/types.ts
```diff
@@ -134,5 +134,7 @@ export interface TableProps<T extends TableItem> {
   selection?: TableSelection;
   emptyState?: ReactNode;
   className?: string;
+  tableLayout?: 'auto' | 'fixed';
   style?: React.CSSProperties;
+  styles?: Partial<Record<'tableRoot', React.CSSProperties>>;
```
Naycon: I think we may need to change this, I'm afraid `style` vs `styles` will be confusing to people. 

I think you should be able to style the table inside a `Table` component already via css classes by setting `className` and then having something like `.myClassname .bui-Table { ... }` to target the table element that `TableRoot` renders. It should be equivalent. It would also avoid doing what you wrote in your comment on styling via `.myClassname table`, since that's indeed an internal implementation detail. But the `bui-*` classes are considered part of the public API and are safe to target.

---

`styles` could be an interesting suggestion on how to allow setting style on components that are composed of multiple parts (like Table), but I think we'd maybe need to consider more components and align on that pattern as a replacement for `style` then. Since it's not the main point of this PR, maybe we can leave that for now?

packages/ui/src/components/Table/types.ts
```diff
@@ -134,5 +134,7 @@ export interface TableProps<T extends TableItem> {
   selection?: TableSelection;
   emptyState?: ReactNode;
   className?: string;
+  tableLayout?: 'auto' | 'fixed';
```
Naycon: I wonder if it wouldn't make more sense to have a prop that purely turns on/off the `ResizableTableContainer` wrapping instead of this, and then rely on styling via classname to apply the `table-layout` like you want it (see below). 

The naming is a little confusing to me, and wasn't obvious from the start that this maps partially to the css property with the same name and partially to the resizable wrapping. 🤔

packages/ui/src/components/Table/types.ts
grantila: Yeah, that's fair. This is indeed a bigger topic and out-of-scope for this particular feature.

packages/ui/src/components/Table/types.ts
grantila: Right, this is a bit unfortunate, that you need to set both the `table-layout` CSS property to `auto` _and_ remove the `ResizableTableContainer` to get automatic column widths. On the other hand, you could argue that `ResizableTableContainer` shouldn't be there by default, only if you want to be able to manually specify column widths.

Suggestion:
Perhaps this could be changed to a prop `columnSizing` which can have two values `content` or `manual`. Would that make sense? It would set the `table-layout` CSS property _and_ enable/disable the `ResizableTableContainer` (which is an internal detail the user won't need to know about). I would recommend `content` being default unless some of the columns have `width`, `defaultWidth`, `minWidth` or `maxWidth` set, but that's up to you.

grantila:
Adapted the code to the comment on naming. I think it's clear now what the new prop `columnSizing` means and what it does.

grantila:
Rebased in master, all good IMO

packages/ui/src/components/Table/types.ts
Naycon: Right, so let me give you a little background on how we're reasoning and what we believe is going on:

What we're trying to optimize for here is the most common case, which I believe would be that you want a responsive table that doesn't overflow the page and that truncates the column content. Additionally, the assumption is that you're not happy with giving each column the same width. Happy to have this assumption challenged, but that's where we're coming from at least.

This reflects in the `bui-Table` styles, where we have a fixed layout and width set to 100%. This would give you the responsive table, but equal columns out of the box if you use `TableRoot` to compose your own table.

What `ResizableTableContainer` does is resetting these options as element styles. The table element is fixed layout and width is set to fit-content, while each column gets equal space calculated by RA and set explicitly in pixels. Both results in the same behavior in the end, equal sized responsive columns that fit on screen in the container.

Because of the above, we opted to always render the container, the behavior is the same anyway. Now, this gets in the way of letting you doing what you want, so it makes sense to change it. 

For the `ResizableTableContainer` wrapping, I think we might simply be able to add detection - if any column has set width, defaultWidth, minWidth or maxWidth, then wrap, otherwise don't. This would get rid of any hardcoded styles, since these came from RA.

Regarding the `table-layout` on `bui-Table`, I still believe this is a sane default value (happy to hear counter arguments though). With the container gone, you'd be able to simply apply a css class and style the `bui-Table` the way you like from there. Would that work? I'm a little hesitant to add in props that will set style props in order to undo styles we're setting in our own css.

grantila: Thanks for the explanation! I think you're right that applying `table-layout` could be done manually by css classes, so the `ResizableTableContainer` could be used only if a width is set, and that would basically be it 👍

I have updated the PR.

PR #32692, opened 2026-02-05
update the signals client structure and avoid duplicate registrations

freben:


plugins/signals/src/api/SignalClient.ts
```diff
@@ -25,15 +26,18 @@ type Subscription = {
 
 const WS_CLOSE_NORMAL = 1000;
 const WS_CLOSE_GOING_AWAY = 1001;
+const WS_CLOSE_INTENTIONALLY = 1002;
```
drodil: 1002 is actually "Protocol Error". If you need custom codes, use something from `3000-3999` or `4000-4999` range. See https://websocket.org/reference/close-codes/

plugins/signals/src/api/SignalClient.ts
```diff
@@ -77,156 +81,293 @@ export class SignalClient implements SignalApi {
...
+            try {
+              socket.send(JSON.stringify({ action: 'unsubscribe', channel }));
+            } catch {
+              // intentionally ignored
+            }
+          }
+
+          if (this.subscriptions.size === 0) {
+            this.connectionPromise = undefined;
+            socket.close(WS_CLOSE_INTENTIONALLY);
```
drodil, line 122: I think we can just close with normal exit code

plugins/signals/src/api/SignalClient.ts
```diff
@@ -77,156 +81,293 @@ export class SignalClient implements SignalApi {
...
-    }
-    this.messageQueue = [];
-    if (data) {
-      this.ws!.send(jsonMessage);
+      socket.addEventListener('error', () => {
+        this.reconnect();
+      });
+
+      socket.addEventListener('close', event => {
+        if (event.code !== WS_CLOSE_INTENTIONALLY) {
```
drodil, line 147: I wonder what we do with 1001 "Going Away" messages when the server/browser is shutting down, or page navigation occurs 🤔

plugins/signals/src/api/SignalClient.ts
```diff
@@ -77,156 +81,293 @@ export class SignalClient implements SignalApi {
...
+
+      return socket;
+    })();
+
+    this.connectionPromise = connectionPromise;
+    return await connectionPromise;
+  }
+
+  /**
+   * Throws away the current connection to the signal server (if any) and
```
drodil, line 160:
```suggestion
   * Throws away the current connection to the signal connection (if any) and
```

plugins/signals/src/api/SignalClient.ts
```diff
@@ -77,156 +81,293 @@ export class SignalClient implements SignalApi {
...
+        if (socket.readyState !== WebSocket.OPEN) {
+          throw new Error(
+            'Connection unexpectedly closed while connecting to signal server',
+          );
+        }
+
+        return socket;
+      } catch {
+        try {
+          socket?.close();
```
drodil: Should we be explicit on the close code across the codebase? While `WS_CLOSE_NORMAL` is the default, I think it makes it more readable

plugins/signals/src/api/SignalClient.ts
```diff
@@ -77,156 +81,293 @@ export class SignalClient implements SignalApi {
...
-    this.connect()
-      .then(() => {
-        // Do not subscribe twice to same channel even there is multiple callbacks
-        if (!exists) {
-          this.send({ action: 'subscribe', channel });
-        }
-      })
-      .catch(() => {
-        this.reconnect();
+    this.ensureConnection().then(socket => {
```
drodil: Just wondering, could this actually block the UI in case the connection just cannot be established, and the client is in a reconnect loop? The previous client might have had the same problem while the idea was to keep the `subscribe` unblocking operation

plugins/signals/src/api/SignalClient.ts
```diff
@@ -77,156 +81,293 @@ export class SignalClient implements SignalApi {
...
-    const apiUrl = await this.discoveryApi.getBaseUrl('signals');
-    const { token } = await this.identity.getCredentials();
+    const connectTimeout = this.connectTimeout;
 
-    const url = new URL(apiUrl);
-    url.protocol = url.protocol === 'http:' ? 'ws:' : 'wss:';
-    this.ws = new WebSocket(url.toString(), token);
+    await new Promise<void>((resolve, reject) => {
+      const timeoutHandle = setTimeout(onTimeout, connectTimeout);
+      socket.addEventListener('open', opOpen);
```
drodil, line 238: The event-based connect timeout is nice 👍

plugins/signals/src/api/SignalClient.ts
```diff
@@ -77,156 +81,293 @@ export class SignalClient implements SignalApi {
...
+  private async ensureWebSocketIsOpen(socket: WebSocket) {
+    if (socket.readyState === WebSocket.OPEN) {
       return;
+    } else if (socket.readyState !== WebSocket.CONNECTING) {
+      throw new Error('Web socket was unexpectedly not in a CONNECTING state');
     }
 
-    const apiUrl = await this.discoveryApi.getBaseUrl('signals');
-    const { token } = await this.identity.getCredentials();
+    const connectTimeout = this.connectTimeout;
```
drodil, line 234: Made me think, should we have some backoff period or randomize the reconnect/connect timeouts in case there are issues connecting? Maybe a more advanced thing, but something to keep in mind

drodil:
Overall, good changes, especially the event-based connect/reconnect handling! Maybe there's something in my PR https://github.com/backstage/backstage/pull/32495 to use here as well (tests maybe?)

plugins/signals/src/api/SignalClient.ts
freben: The reason for this, is that I need to capture in the reconnect code that this really is an intentional close - so it doesn't try to undo that immediately with a reconnect. At least that was the original intent 🤔

plugins/signals/src/api/SignalClient.ts
freben: Ah my bad for not being familiar with these - went with a 3000 code then

plugins/signals/src/api/SignalClient.ts
freben: Hm. I think this was probably right already, wasn't it?

plugins/signals/src/api/SignalClient.ts
freben: yeah i agree, added

PR #32695, opened 2026-02-05
auth-backend: add experimental refresh token support

benjdlambert:
This is the `offline_session` support pulled out of #31482 and #32312.

Might make it easier to review,

### DIFF
--- a/plugins/auth-backend/src/authPlugin.ts
+++ b/plugins/auth-backend/src/authPlugin.ts
@@ -26,10 +26,7 @@ import {
 } from '@backstage/plugin-auth-node';
 import { catalogServiceRef } from '@backstage/plugin-catalog-node';
 import { createRouter } from './service/router';
-import { OfflineSessionDatabase } from './database/OfflineSessionDatabase';
 import { OfflineAccessService } from './service/OfflineAccessService';
-import { readDurationFromConfig } from '@backstage/config';
-import { durationToMilliseconds } from '@backstage/types';
 
 /**
  * Auth plugin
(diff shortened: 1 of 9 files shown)


plugins/auth-backend/src/authPlugin.ts
```diff
@@ -77,8 +82,119 @@ export const authPlugin = createBackendPlugin({
...
+          const maxTokensPerUser =
+            config.getOptionalNumber(
+              'auth.experimentalRefreshToken.maxTokensPerUser',
+            ) ?? 20;
+
+          if (maxTokensPerUser <= 0) {
+            throw new Error(
+              'auth.experimentalRefreshToken.maxTokensPerUser must be a positive number',
+            );
+          }
```
Rugvip: Lots of code here that I think would be best to move out

plugins/auth-backend/src/authPlugin.ts
```diff
@@ -77,8 +82,119 @@ export const authPlugin = createBackendPlugin({
...
+            } catch (error) {
+              logger.error('Failed to cleanup expired offline sessions', error);
+            }
+          }, cleanupIntervalMs);
+          cleanupInterval.unref();
+
+          lifecycle.addShutdownHook(() => {
+            clearInterval(cleanupInterval);
+          });
+        }
```
Rugvip: Best move this somewhere else as well

plugins/auth-backend/src/database/OfflineSessionDatabase.ts
```diff
@@ -0,0 +1,264 @@
...
+
+const TABLE_NAME = 'offline_sessions';
+
+type DbOfflineSessionRow = {
+  id: string;
+  user_entity_ref: string;
+  oidc_client_id: string | null;
+  token_hash: string;
+  created_at: Date;
+  last_used_at: Date;
```
freben, line 28: technically should probably be `Date | string` but it's not used externally so it doesn't matter

Rugvip (approved):
🎉

PR #32711, opened 2026-02-06
BUI - Replace Surface system with Bg (neutral background) system

cdedreuille:

This PR replaces the entire `Surface` / `onSurface` system with a new provider/consumer background (`bg`) system built around neutral levels. The old system had separate concepts for surfaces (`'0'`–`'3'`, `'auto'`) and leaf components (`onSurface`), which made it difficult to share background semantics between containers and interactive components like buttons and cards. The new system splits responsibilities clearly: **provider** components (Box, Flex, Grid, Card) establish a background context, and **consumer** components (Button, ButtonIcon, ButtonLink, ToggleButton) read from that context automatically.

### New types: `ContainerBg` and `ProviderBg`

Replace the old `Surface` type. Values are explicit and semantic:

- `ContainerBg` — `'neutral-1'` | `'neutral-2'` | `'neutral-3'` | `'danger'` | `'warning'` | `'success'`
- `ProviderBg` — `ContainerBg | 'neutral-auto'`

| Old (`Surface`)       | New (`ContainerBg`)                            |
| --------------------- | ---------------------------------------------- |
| `'0'`                 | `'neutral-1'`                                  |
| `'1'`                 | `'neutral-2'`                                  |
| `'2'`                 | `'neutral-3'`                                  |
| `'3'`                 | _(no prop equivalent — `neutral-4` is CSS-only)_ |
| `'danger'`            | `'danger'`                                     |
| `'warning'`           | `'warning'`                                    |
| `'success'`           | `'success'`                                    |
| `'auto'`              | `'neutral-auto'` _(must be set explicitly)_    |

> **Note:** There is no `neutral-4` prop value. Provider components cap at `neutral-3`. The `neutral-4` level exists only in consumer component CSS (e.g. a Button on a `neutral-3` surface uses `neutral-4` tokens internally via `data-on-bg`).

### Removed types

- `Surface` — replaced by `ContainerBg` / `ProviderBg`
- `LeafSurfaceProps` — removed entirely (consumer components no longer need an `onSurface` prop)
- `ContainerSurfaceProps` — removed entirely (providers use `bg?: Responsive<ProviderBg>` directly)

---

### New hooks: `useBgProvider` and `useBgConsumer`

Replace the deleted `useSurface` hook. Exported from `@backstage/ui`.

- **`useBgProvider(bg?)`** — for provider components (Box, Flex, Grid, Card). Resolves the `bg` value:
  - `undefined` → transparent, no context change (this is the default for Box, Flex, and Grid — they do **not** auto-increment)
  - A `ContainerBg` value (e.g. `'neutral-1'`) → uses that value directly
  - `'neutral-auto'` → increments from the parent context, capping at `neutral-3`. Only components that explicitly pass `'neutral-auto'` (e.g. Card) will auto-increment; it is never implicit.
- **`useBgConsumer()`** — for consumer components (Button, ButtonIcon, ButtonLink). Returns the parent container's `bg` unchanged. No prop needed on the component.

### New provider: `BgProvider`

Replaces `SurfaceProvider`. Establishes the bg context for child components. Provider components wrap their children in `BgProvider` when a resolved bg is defined.

### Deleted hook: `useSurface`

The entire `useSurface.tsx` file and its `SurfaceProvider` have been deleted. No component references them anymore.

---

### Box

| Before | After |
| --- | --- |
| `<Box surface="1">` | `<Box bg="neutral-1">` |

- **types.ts** — `surface?: Responsive<Surface>` → `bg?: Responsive<ProviderBg>`
- **definition.ts** — `surface: 'container'` → `bg: 'provider'`
- **Box.tsx** — Uses `childrenWithBgProvider` instead of `surfaceChildren`
- **CSS** — `[data-surface]` selectors → `[data-bg]` selectors with `neutral-*` tokens

Box is **transparent by default** — without an explicit `bg` prop, it does not set any background context.

### Button

| Before | After |
| --- | --- |
| `<Button onSurface="1">` | `<Button>` _(reads from parent context automatically)_ |

- **types.ts** — Removed `LeafSurfaceProps` extension (no more `onSurface` prop)
- **definition.ts** — `surface: 'leaf'` → `bg: 'consumer'`, removed `onSurface` from propDefs
- **CSS** — `[data-on-surface='1'/'2'/'3']` → `[data-on-bg='neutral-1'/'neutral-2'/'neutral-3']`

### ButtonIcon

Same pattern as Button:

- **types.ts** — Removed `LeafSurfaceProps` extension
- **definition.ts** — `surface: 'leaf'` → `bg: 'consumer'`, removed `onSurface` from propDefs
- **CSS** — `[data-on-surface]` → `[data-on-bg]` selectors

### ButtonLink

Same pattern as Button:

- **types.ts** — Removed `LeafSurfaceProps` extension
- **definition.ts** — `surface: 'leaf'` → `bg: 'consumer'`, removed `onSurface` from propDefs
- **CSS** — `[data-on-surface]` → `[data-on-bg]` selectors

### ToggleButton

| Before | After |
| --- | --- |
| `<ToggleButton onSurface="1">` | `<ToggleButton>` _(no bg handling)_ |

- **types.ts** — Removed `onSurface?: Responsive<Surface>` prop entirely
- **ToggleButton.tsx** — Removed `useSurface` usage; ToggleButton no longer participates in the bg system directly

### Card

| Before | After |
| --- | --- |
| `<Card surface="1">` | `<Card>` _(auto-increments via inner Box)_ |

- Card does **not** accept a `bg` prop directly. It hardcodes `bg="neutral-auto"` on its inner `Box`, so it always auto-increments from the parent context.
- **Card.tsx** — Renders `<Box bg="neutral-auto" ...>` internally
- **CSS** — `[data-surface='0'/'1'/'2'/'3']` → `[data-bg='neutral-1'/'neutral-2'/'neutral-3']`; `--bui-bg-surface-*` tokens → `--bui-bg-neutral-*` tokens

### Alert

| Before | After |
| --- | --- |
| `<Alert surface="...">` | `<Alert>` _(surface handling removed)_ |

Alert's background is driven by its `status` prop (`data-status` CSS selectors), not the surface/bg system. Surface handling was removed entirely — Alert is now transparent to the bg chain. Children of Alert inherit bg from Alert's parent.

- **types.ts** — Removed `ContainerSurfaceProps` extension
- **definition.ts** — Removed `surface: 'container'` and `surface: {}` from propDefs

### Flex

| Before | After |
| --- | --- |
| `<Flex surface="1">` | `<Flex bg="neutral-1">` |

- **types.ts** — `surface?: Responsive<Surface>` → `bg?: Responsive<ProviderBg>`
- **definition.ts** — `dataAttributes.surface` → `dataAttributes.bg` with `neutral-*` values
- **Flex.tsx** — Replaced `useSurface`/`SurfaceProvider` with `useBgProvider`/`BgProvider`
- **CSS** — `[data-surface]` → `[data-bg]` selectors

Flex is **transparent by default** — without an explicit `bg` prop, it does not set any background context.

### Grid (Root + Item)

| Before | After |
| --- | --- |
| `<Grid.Root surface="1">` | `<Grid.Root bg="neutral-1">` |
| `<Grid.Item surface="1">` | `<Grid.Item bg="neutral-1">` |

- **types.ts** — `surface` → `bg` in both `GridProps` and `GridItemProps`
- **definition.ts** — `dataAttributes.surface` → `dataAttributes.bg` for both definitions
- **Grid.tsx** — Replaced `useSurface`/`SurfaceProvider` with `useBgProvider`/`BgProvider` in both `GridRoot` and `GridItem`
- **CSS** — `[data-surface]` → `[data-bg]` selectors

Grid components are **transparent by default**.

---

### `useDefinition` system

- **useDefinition.tsx** — Removed all `useSurface`/`SurfaceProvider` imports and logic. Now uses `useBgProvider`/`useBgConsumer`/`BgProvider`. Sets `data-bg` for providers and `data-on-bg` for consumers.
- **types.ts** — Removed `surface?: 'container' | 'leaf'` from `ComponentConfig`, replaced with `bg?: 'provider' | 'consumer'`. Removed `SurfacePropsConstraint` type, removed `'data-on-surface'` from `DataAttributes`, added `'data-bg'` and `'data-on-bg'`.
- **defineComponent.ts** — Removed `SurfacePropsConstraint` from the config validation type.

### CSS data attributes

All components moved from `data-surface` / `data-on-surface` to `data-bg` / `data-on-bg`:

| Old attribute | New attribute | Used by | Example value |
| --- | --- | --- | --- |
| `data-surface="1"` | `data-bg="neutral-1"` | Provider components | Container background |
| `data-on-surface="2"` | `data-on-bg="neutral-2"` | Consumer components | Parent's bg, inherited |

---

### Provider components (Box, Flex, Grid)

Provider components are **transparent by default**. Without an explicit `bg` prop, they do not establish any background context and children see through to the nearest ancestor that does.

```tsx
// Explicit background — sets neutral-1 context for children
<Box bg="neutral-1">
  <Button variant="secondary">Sees neutral-1 via data-on-bg</Button>
</Box>

// No bg prop — transparent, no context change
<Box>
  <Button variant="secondary">Inherits from nearest ancestor with bg</Button>
</Box>
```

### Card (auto-incrementing provider)

Card always auto-increments from its parent context because it hardcodes `bg="neutral-auto"` internally. You do not pass a `bg` prop to Card.

```tsx
<Box bg="neutral-1">         {/* data-bg="neutral-1" */}
  <Card>                      {/* auto → data-bg="neutral-2" */}
    <Button variant="secondary">Sees neutral-2 via data-on-bg</Button>
  </Card>
</Box>
```

### Explicit auto-increment with `neutral-auto`

Any provider component can auto-increment by passing `bg="neutral-auto"`:

```tsx
<Box bg="neutral-1">                    {/* data-bg="neutral-1" */}
  <Box bg="neutral-auto">               {/* auto → data-bg="neutral-2" */}
    <Box bg="neutral-auto">             {/* auto → data-bg="neutral-3" */}
      <Button variant="secondary">Sees neutral-3 via data-on-bg</Button>
    </Box>
  </Box>
</Box>
```

### Neutral level capping

Provider components cap at `neutral-3`. Further auto-incrementing stays at `neutral-3`. The `neutral-4` level is not a valid prop value — it exists only in consumer component CSS (e.g. a Button on a `neutral-3` surface renders with `neutral-4` tokens internally via `data-on-bg`).

---

### DIFF
--- a/.changeset/clean-bags-occur.md
+++ b/.changeset/clean-bags-occur.md
@@ -11,19 +11,22 @@ The old `Surface` type (`'0'`–`'3'`, `'auto'`) and its associated props (`surf
 - `ContainerBg` — `'neutral-1'` | `'neutral-2'` | `'neutral-3'` | `'danger'` | `'warning'` | `'success'`
 - `ProviderBg` — `ContainerBg | 'neutral-auto'`
 
-There is no `neutral-4` prop value; containers are capped at `neutral-3`. Consumer components (e.g. Button) inherit the parent's `bg` via `data-on-bg`, and CSS handles the visual step-up.
+Consumer components (e.g. Button) inherit the parent's `bg` via `data-on-bg`, and CSS handles the visual step-up. See "Neutral level capping" below for details on how levels are bounded.
 
 **Hooks:**
 
 - `useBgProvider(bg?)` — for provider components. Returns `{ bg: undefined }` when no `bg` is given (transparent). Supports `'neutral-auto'` to auto-increment from the parent context.
 - `useBgConsumer()` — for consumer components. Returns the parent container's `bg` unchanged.
-- The old `useBg` hook and `UseBgOptions` interface have been removed.
 
 **Component roles:**
 
-- **Provider-only** (Box, Flex, Grid): set `data-bg`, wrap children in `BgProvider`. Transparent by default.
-- **Consumer-only** (Button, ButtonIcon, ButtonLink): set `data-on-bg`, inherit from parent.
-- **Provider + Consumer** (Card): sets both `data-bg` and `data-on-bg`, wraps children. Defaults to `neutral-auto`.
+- **Provider-only** (Box, Flex, Grid): set `data-bg`, wrap children in `BgProvider`. **Transparent by default** — they do _not_ auto-increment; pass `bg="neutral-auto"` explicitly if you want automatic neutral stepping.
+- **Consumer-only** (Button, ButtonIcon, ButtonLink): set `data-on-bg`, inherit the parent container's `bg` unchanged.
+- **Provider + Consumer** (Card): sets both `data-bg` and `data-on-bg`, wraps children. Card passes `bg="neutral-auto"` to its inner Box, so it auto-increments from the parent context.
+
+**Neutral level capping:**
+
+Provider components cap at `neutral-3`. There is no `neutral-4` prop value. The `neutral-4` level exists only in consumer component CSS — for example, a Button sitting on a `neutral-3` surface uses `neutral-4` tokens internally via `data-on-bg`.
 
 **Migration:**
 
@@ -86,6 +89,6 @@ Update CSS selectors targeting surface data attributes:
 + [data-on-bg='neutral-1'] { ... }
 ```
 
-Note: Provider components use `data-bg` (values: `neutral-1` through `neutral-3`, plus intents). Consumer components use `data-on-bg`, which reflects the parent container's `bg` directly.
+Note: Provider components use `data-bg` (values: `neutral-1` through `neutral-3`, plus intent values). Consumer components use `data-on-bg`, which reflects the parent container's `bg` directly. The `neutral-4` level never appears as a prop or `data-bg` value — it is used only in consumer CSS.
 
 **Affected components:** Box, Button, ButtonIcon, ButtonLink, ToggleButton, Card, Flex, Grid
--- a/packages/ui/src/hooks/useBg.tsx
+++ b/packages/ui/src/hooks/useBg.tsx
@@ -86,9 +86,20 @@ export function useBgConsumer(): BgContextValue {
 /**
  * Hook for provider components (e.g. Box, Card) to resolve and provide bg context.
  *
- * - `bg` is `undefined` -- transparent, no context change, returns `{ bg: undefined }`
- * - `bg` is a `ContainerBg` value -- uses that value directly
- * - `bg` is `'neutral-auto'` -- increments from the parent context, capping at `neutral-3`
+ * **Resolution rules:**
+ *
+ * - `bg` is `undefined` -- transparent, no context change, returns `{ bg: undefined }`.
+ *   This is the default for Box, Flex, and Grid (they do **not** auto-increment).
+ * - `bg` is a `ContainerBg` value -- uses that value directly (e.g. `'neutral-1'`).
+ * - `bg` is `'neutral-auto'` -- increments the neutral level from the parent context,
+ *   capping at `neutral-3`. Only components that explicitly pass `'neutral-auto'`
+ *   (e.g. Card) will auto-increment; it is never implicit.
+ *
+ * **Capping:**
+ *
+ * Provider components cap at `neutral-3`. The `neutral-4` level is **not** a valid
+ * prop value -- it exists only in consumer component CSS (e.g. a Button on a
+ * `neutral-3` surface renders with `neutral-4` tokens via `data-on-bg`).
  *
  * The caller is responsible for wrapping children with `BgProvider` when the
  * resolved bg is defined.


packages/ui/src/hooks/useBg.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  // No explicit bg: auto-increment from context if available
+  if (contextBg === undefined) {
+    return undefined;
+  }
+
+  // If context is a responsive object, we can't auto-increment
+  if (typeof contextBg === 'object') {
+    return undefined;
+  }
```
Naycon: Hmm, this feels like a flaw and perhaps a source of bugs/weird behavior. Shouldn't we work with resolved values here instead? Are there any benefits to keeping the value as the responsive object rather than resolving it and letting children (container or leaf) work with that?

packages/ui/src/hooks/useDefinition/types.ts
```diff
@@ -81,12 +75,11 @@ type ResolvedOwnProps<
...
 };
 
-type ChildrenProps<Surface extends 'container' | 'leaf' | undefined> =
-  Surface extends 'container'
-    ? { surfaceChildren: ReactNode; children?: never }
-    : Surface extends 'leaf'
-    ? { children: ReactNode; surfaceChildren?: never }
-    : { children: ReactNode };
+type ChildrenProps<
+  BgMode extends 'container' | 'leaf' | undefined = undefined,
```
Naycon: There's a new default to `undefined` here. Is this intentional and necessary?

packages/ui/src/hooks/useDefinition/index.ts
```diff
@@ -21,4 +21,5 @@ export type {
   PropDefConfig,
   UseDefinitionOptions,
   UseDefinitionResult,
+  BgPropsConstraint,
```
Naycon: Any reason why this type is now exported? Feels like an internal type that we don't need to expose, or am I missing something?

packages/ui/src/components/Box/types.ts
```diff
@@ -15,12 +15,12 @@
...
 
 import type { ReactNode, CSSProperties } from 'react';
-import type { Responsive, Surface, SpaceProps } from '../../types';
+import type { Responsive, Bg, SpaceProps } from '../../types';
 
 /** @public */
 export type BoxOwnProps = {
   as?: keyof JSX.IntrinsicElements;
-  surface?: Responsive<Surface>;
+  bg?: Responsive<Bg>;
```
Naycon: There's a utility type already that I think needs updating and that could be good to use here instead, `ContainerSurfaceProps` in `src/types.ts`. It needs updating ofc, and the `Leaf` type should be removed. 

With that we can do `export type BoxOwnProps = ContainerBgProps & { ... }` in all of these containers, and we don't need to manually say `bg?: Responsive<Bg>` everywhere. We should perhaps apply that to all containers instead of manually hardcoding the type.

.changeset/clean-bags-occur.md
```diff
@@ -0,0 +1,69 @@
...
++ import { useBg, BgProvider } from '@backstage/ui';
+
+- const { surface } = useSurface({ surface: props.surface });
++ const { bg } = useBg({ mode: 'container', bg: props.bg });
+
+- const { surface } = useSurface({ onSurface: props.onSurface });
++ const { bg } = useBg({ mode: 'leaf' });
+```
+
+**Affected components:** box, button, button-icon, button-link, toggle-button, card, alert, flex, grid
```
Naycon: This one is a little off, should be uppercase first letter and not kebab-case I think, or the sync script won't work.

.changeset/clean-bags-occur.md
```diff
@@ -0,0 +1,69 @@
...
++ <Card bg="neutral-2">
+
+- <Flex surface="0">
++ <Flex bg="neutral-1">
+
+- <Grid.Root surface="1">
++ <Grid.Root bg="neutral-1">
+```
+
+Remove `onSurface` from leaf components — it is now fully automatic:
```
Naycon: I think we're missing a section on updating css right? Selectors will change with this, won't they?

.changeset/clean-bags-occur.md
```diff
@@ -0,0 +1,69 @@
+---
+'@backstage/ui': minor
+---
+
+**BREAKING**: Replaced `Surface` / `onSurface` system with new `Bg` background system
+
+The old `Surface` type (`'0'`–`'3'`, `'auto'`) and its associated props (`surface`, `onSurface`) have been replaced by a new `Bg` type with semantic neutral levels (`'neutral-1'` through `'neutral-4'`) and intents (`'danger'`, `'warning'`, `'success'`). Leaf components like Button no longer need an explicit prop — backgrounds auto-increment from parent context.
```
Naycon: Maybe I'm slow, what does "background auto-increment from parent context" mean? What is auto-incrementing and why? I'd expect leaf components like Button to just have to specify what they want to look like when on top of a container. I thought `onSurface` was removed because we don't allow overrides of bg values for leaves anymore. And that auto-increment was the default behavior for container components, i.e. if you don't set a value for `bg`, we autoincrement the value for the nested container. But the text seems to suggest my understanding is wrong.

.changeset/clean-bags-occur.md
Naycon: Another mention of "fully automatic".

packages/ui/src/components/Alert/definition.ts
```diff
@@ -34,15 +34,13 @@ export const AlertDefinition = defineComponent<AlertOwnProps>()({
     spinner: 'bui-AlertSpinner',
     actions: 'bui-AlertActions',
   },
-  surface: 'container',
```
Naycon, line 37: Side note - kind of nice to see that just changing the definition like this creates the desired change in the component without having to update any component code. 🎉

packages/ui/src/hooks/useDefinition/index.ts
cdedreuille: You're right. I removed it.

packages/ui/src/components/Box/types.ts
cdedreuille: I did remove it as it felt quite unnecessary. `bg?: Responsive<Bg>;` is short enough that it doesn't need any more abstraction.

packages/ui/src/hooks/useBg.tsx
cdedreuille: You're right. I updated that.

packages/ui/src/hooks/useDefinition/types.ts
cdedreuille: It seems that we need the `undefined` but I removed `= undefined` as this was not necessary

.changeset/clean-bags-occur.md
cdedreuille: Done

.changeset/clean-bags-occur.md
cdedreuille: You're right. I added an example.

.changeset/clean-bags-occur.md
cdedreuille: The auto increment is done on leaf component. So if you have a container with `neutral-1`, then the leaf `Button` will be `neutral-2`. This process is done automatically now.

.changeset/clean-bags-occur.md
cdedreuille: Yes

Naycon: Should this be rephrased now? It's not more or less automatic than it was before, you just can't override the container value anymore, right?

packages/ui/src/hooks/useBg.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+  // Resolve responsive prop value to a scalar for the current breakpoint
+  const resolvedPropBg =
+    options.bg !== undefined
+      ? resolveResponsiveValue(options.bg, breakpoint)
+      : undefined;
+
+  if (options.mode === 'leaf') {
+    return { bg: resolveBgForLeaf(context.bg) };
```
Naycon: Might be clearer to either just return `context.bg`:

```suggestion
    return { bg: context.bg };
```

or even simplify to:

```suggestion
    return context.bg;
```

packages/ui/src/hooks/useBg.tsx
```diff
@@ -0,0 +1,165 @@
...
+  }
```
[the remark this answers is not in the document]
Naycon: Copilot is on to something here - I'm not so worried about the performance, but I think it might make sense to restructure this hook a little so it becomes clearer what happens for containers and what happens for leaves. 

I'd suggest getting context + breakpoint from the hooks, then splitting the flow and having `useContainerBg` and `useLeafBg` as two separate internal hooks that consume options, context and breakpoint and return `BgContextValue`. That will make it easier to read the code an follow what's going on.

packages/ui/src/hooks/useBg.tsx
```diff
@@ -0,0 +1,171 @@
...
+function resolveBgForContainer(
+  contextBg: ContainerBg | undefined,
+  propBg: ContainerBg | undefined,
+): ContainerBg | undefined {
+  // Explicit bg prop takes priority
+  if (propBg !== undefined) {
+    return propBg;
+  }
+
+  // No explicit bg: auto-increment from context if available
```
Naycon: This comment is confusing, it looks like it's commenting on the if-statement, but I think it's actually meant for the whole rest of the flow. I'd say remove it, the code is quite clear on its own.

packages/ui/src/hooks/useDefinition/useDefinition.tsx
```diff
@@ -77,16 +72,11 @@ export function useDefinition<
...
-        ? resolveResponsiveValue(resolvedSurface as any, breakpoint)
-        : resolvedSurface;
-    if (surfaceValue !== undefined) {
-      dataAttributes['data-on-surface'] = String(surfaceValue);
-    }
+  // Set the bg data attribute from the resolved bg value
+  // Containers use data-bg, leaf components use data-on-bg
+  if (definition.bg && resolvedBg !== undefined) {
+    const attrName = definition.bg === 'leaf' ? 'data-on-bg' : 'data-bg';
+    dataAttributes[attrName] = String(resolvedBg);
```
Naycon: Nice!

.changeset/clean-bags-occur.md
```diff
@@ -0,0 +1,81 @@
...
++ <Button variant="secondary">
+
+- <ButtonIcon onSurface="2" variant="secondary" />
++ <ButtonIcon variant="secondary" />
+
+- <ToggleButton onSurface="1">
++ <ToggleButton>
+```
+
+Alert no longer accepts a `surface` prop (its background is driven by `status`):
```
Naycon: Completely optional - I wonder if it's worth putting this part in its own changeset, to avoid getting all of this changeset reproduced in the Alert components changeset in the docs?

.changeset/clean-bags-occur.md
```diff
@@ -0,0 +1,81 @@
+---
+'@backstage/ui': minor
+---
+
+**BREAKING**: Replaced `Surface` / `onSurface` system with new `ContainerBg` background system
+
+The old `Surface` type (`'0'`–`'3'`, `'auto'`) and its associated props (`surface`, `onSurface`) have been replaced by a new `ContainerBg` type with semantic neutral levels (`'neutral-1'` through `'neutral-3'`) and intents (`'danger'`, `'warning'`, `'success'`). Containers are capped at `neutral-3`; the `neutral-4` level is reserved for leaf component CSS only. Leaf components like Button no longer need an explicit prop — they receive a `data-on-bg` attribute matching the parent container's bg, and CSS handles the visual step-up.
```
Naycon: Do we need to talk about `neutral-4` here at all? `neutral-` is a new set of prop values right, so I don't think anyone would expect `neutral-4` to be available. This text seems to be written in the point of view of this PR more than from the point of view of someone reading this in the changeset perhaps?

Also, there's a small Vale error that needs fixing, `bg` (without code formatting) is not a word/term.

packages/ui/src/hooks/useDefinition/useDefinition.tsx
```diff
@@ -77,16 +72,11 @@ export function useDefinition<
...
-      typeof resolvedSurface === 'object'
-        ? resolveResponsiveValue(resolvedSurface as any, breakpoint)
-        : resolvedSurface;
-    if (surfaceValue !== undefined) {
-      dataAttributes['data-on-surface'] = String(surfaceValue);
-    }
+  // Set the bg data attribute from the resolved bg value
+  // Containers use data-bg, leaf components use data-on-bg
+  if (definition.bg && resolvedBg !== undefined) {
+    const attrName = definition.bg === 'leaf' ? 'data-on-bg' : 'data-bg';
```
Naycon: Technically I don't think it matters at all right now, but it might be worth pointing out that since you've marked the container `bg` as `bg: { dataAttribute: true }` in the component definitions, what you'll do here is to overwrite `data-bg` once more for containers. `data-bg` will already have been created and set by the data attributes handling above if the container has an explicit `bg` prop set.

packages/ui/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -34,17 +34,16 @@ export const ToggleButton = forwardRef(
...
-    const { surface } = useSurface({ onSurface });
+    const { bg } = useBg({ mode: 'leaf' });
 
     return (
       <AriaToggleButton
         className={clsx(classNames.root, styles[classNames.root], className)}
         ref={ref}
         {...dataAttributes}
-        {...(typeof surface === 'string' ? { 'data-on-surface': surface } : {})}
+        {...(bg ? { 'data-on-bg': bg } : {})}
```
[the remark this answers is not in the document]
Naycon: Fair point from Copilot here, do we need `useBg` at all?

packages/ui/src/components/Grid/Grid.stories.tsx
```diff
@@ -106,102 +106,91 @@ export const RowAndColumns = meta.story({
...
+        <Grid.Item>Neutral 1 (Grid.Root)</Grid.Item>
         <Grid.Item>
-          <Grid.Root {...args} surface="auto">
-            <Grid.Item>Nested: Surface auto (becomes 1)</Grid.Item>
-            <Grid.Item surface="auto">
-              Nested: Surface auto (becomes 2)
-            </Grid.Item>
+          <Grid.Root {...args}>
+            <Grid.Item>Nested: Auto (becomes neutral-2)</Grid.Item>
+            <Grid.Item>Nested: Auto (becomes neutral-2)</Grid.Item>
```
Naycon: Copilot commented on this and I think it's on to something. Aren't these labels incorrect? `Root -> Root -> Item` should mean three increments, so these items will be `neutral-3`. Is btw the desired behavior? Should both root and item behave like containers? All grids will increment two levels when used and make a box with boxes visual layout. 

And yes, you can always override this manually, just asking what we believe is the good default behavior, so people don't have to bring overrides to do the common case.

packages/ui/src/hooks/useDefinition/types.ts
Naycon: Yea, we do need `| undefined` as part of the type, but it was the defaulting (`= undefined`) that I was curious to if it was necessary.

packages/ui/src/components/Grid/Grid.stories.tsx
cdedreuille: You're right. `Grid.Root`, `Grid.Item`, and `Flex` now only call `useBg({ mode: 'container', bg: props.bg })` when an explicit `bg` prop is provided. Without it, they pass undefined to useBg, which returns the raw parent context -- no increment, no BgProvider, no `data-bg` attribute. They're fully transparent layout primitives.

packages/ui/src/hooks/useDefinition/useDefinition.tsx
cdedreuille: This part has been refactored.

.changeset/clean-bags-occur.md
cdedreuille: Updated. Thanks

.changeset/clean-bags-occur.md
cdedreuille: You're right. I moved it to a new changeset

packages/ui/src/hooks/useBg.tsx
cdedreuille: Removed.

.changeset/clean-bags-occur.md
```diff
@@ -0,0 +1,91 @@
...
+
+```diff
+- [data-surface='1'] { ... }
++ [data-bg='neutral-1'] { ... }
+
+- [data-on-surface='1'] { ... }
++ [data-on-bg='neutral-1'] { ... }
+```
+
+Note: Provider components use `data-bg` (values: `neutral-1` through `neutral-3`, plus intents). Consumer components use `data-on-bg`, which reflects the parent container's bg directly (no auto-increment).
```
Naycon: Fix Vale error and rephrase slightly:

```suggestion
Note: Provider components use `data-bg` (values: `neutral-1` through `neutral-3`, plus intents). Consumer components use `data-on-bg`, which reflects the closest provider component's `bg`.
```

Naycon (approved):
Looks good! It's a big one, but I think we're ready to :shipit:!

PR #32728, opened 2026-02-06
Integration - Removed long deprecated code

awanlin:

Removed deprecated Azure DevOps, Bitbucket Gerrit and GitHub code.

>[!NOTE]
> Part of this PR deletes the `@backstage/plugin-scaffolder-backend-module-bitbucket` package, we'll need to mark it as deprecated on NPM. The code inside it has long been marked as such.

### DIFF
--- a/packages/integration/src/azure/AzureIntegration.test.ts
+++ b/packages/integration/src/azure/AzureIntegration.test.ts
@@ -25,7 +25,7 @@ describe('AzureIntegration', () => {
           azure: [
             {
               host: 'h.com',
-              token: 'token',
+              credentials: [{ personalAccessToken: 'token' }],
             },
           ],
         },
--- a/plugins/catalog-backend-module-azure/src/processors/AzureDevOpsDiscoveryProcessor.test.ts
+++ b/plugins/catalog-backend-module-azure/src/processors/AzureDevOpsDiscoveryProcessor.test.ts
@@ -103,7 +103,12 @@ describe('AzureDevOpsDiscoveryProcessor', () => {
       const processor = AzureDevOpsDiscoveryProcessor.fromConfig(
         new ConfigReader({
           integrations: {
-            azure: [{ host: 'dev.azure.com', token: 'blob' }],
+            azure: [
+              {
+                host: 'dev.azure.com',
+                credentials: [{ personalAccessToken: 'token' }],
+              },
+            ],
           },
         }),
         { logger: mockServices.logger.mock() },
@@ -123,8 +128,14 @@ describe('AzureDevOpsDiscoveryProcessor', () => {
       new ConfigReader({
         integrations: {
           github: [
-            { host: 'dev.azure.com', token: 'blob' },
-            { host: 'azure.myorg.com', token: 'blob' },
+            {
+              host: 'dev.azure.com',
+              credentials: [{ personalAccessToken: 'token' }],
+            },
+            {
+              host: 'azure.myorg.com',
+              credentials: [{ personalAccessToken: 'token' }],
+            },
           ],
         },
       }),
@@ -145,7 +156,12 @@ describe('AzureDevOpsDiscoveryProcessor', () => {
     const processor = AzureDevOpsDiscoveryProcessor.fromConfig(
       new ConfigReader({
         integrations: {
-          github: [{ host: 'dev.azure.com', token: 'blob' }],
+          github: [
+            {
+              host: 'dev.azure.com',
+              credentials: [{ personalAccessToken: 'token' }],
+            },
+          ],
         },
       }),
       { logger: mockServices.logger.mock() },
--- a/plugins/scaffolder-backend-module-cookiecutter/src/actions/fetch/cookiecutter.examples.test.ts
+++ b/plugins/scaffolder-backend-module-cookiecutter/src/actions/fetch/cookiecutter.examples.test.ts
@@ -14,10 +14,12 @@
  * limitations under the License.
  */
 
-import { ConfigReader } from '@backstage/config';
 import { JsonObject } from '@backstage/types';
 import { ScmIntegrations } from '@backstage/integration';
-import { createMockDirectory } from '@backstage/backend-test-utils';
+import {
+  createMockDirectory,
+  mockServices,
+} from '@backstage/backend-test-utils';
 import { createFetchCookiecutterAction } from './cookiecutter';
 import { join } from 'node:path';
 import type { ActionContext } from '@backstage/plugin-scaffolder-node';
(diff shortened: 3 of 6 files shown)


packages/integration/src/gerrit/core.ts
```diff
@@ -350,7 +330,7 @@ export function getGerritBranchApiUrl(
   config: GerritIntegrationConfig,
   url: string,
 ) {
-  const { branch, project } = parseGerritGitilesUrl(config, url);
+  const { branch, project } = parseGitilesUrlRef(config, url);
```
awanlin: Hi @anicke or @tcardonne, could use your help with this one as people who have worked in this code before. It looks like `parseGitilesUrlRef` does not return a `branch` the way that `parseGerritGitilesUrl` did. I'm honestly at a loss as to how to find that, did give the docs a look over. Any chance you can help me with that? We'd really like to remove this old code and as we are doing a major bump of this package it makes sense to remove it all. 🙏

.changeset/sharp-ravens-shop.md
```diff
@@ -0,0 +1,10 @@
+---
+'@backstage/integration': major
+---
+
+**BREAKING** Removed deprecated Azure DevOps, Bitbucket, Gerrit and GitHub code:
+
+- For Azure Devops the long deprecated `token` string and `credential` object have been removed from the `config.d.ts`, please use the `credentials` array object instead.
+- For Bitbucket the long deprecated `bitbucket` object has been removed from the `config.d.ts`, please use the `bitbucketCloud` or `bitbucketServer` objects instead.
+- For Gerrit the `parseGerritGitilesUrl` function has been removed please use `parseGitilesUrlRef` instead. The `buildGerritGitilesArchiveUrl` function has also been removed, please use `buildGerritGitilesArchiveUrlFromLocation` instead.
+- For GitHub the `getGitHubRequestOptions` function has been removed, there is no replacement for it.
```
aramissennyeydd:
```suggestion
- For Azure Devops, the long deprecated `token` string and `credential` object have been removed from the `config.d.ts`. Use the `credentials` array object instead.
- For Bitbucket, the long deprecated `bitbucket` object has been removed from the `config.d.ts`. Use the `bitbucketCloud` or `bitbucketServer` objects instead.
- For Gerrit, the `parseGerritGitilesUrl` function has been removed, use `parseGitilesUrlRef` instead. The `buildGerritGitilesArchiveUrl` function has also been removed, use `buildGerritGitilesArchiveUrlFromLocation` instead.
- For GitHub, the `getGitHubRequestOptions` function has been removed.
```

.changeset/six-trees-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-defaults': minor
+---
+
+**BREAKING** Removed deprecated `BitbucketUrlReader` please us the `BitbucketCloudUrlReader` or the `BitbucketServerUrlReader` instead.
```
aramissennyeydd:
```suggestion
**BREAKING** Removed deprecated `BitbucketUrlReader`. Use the `BitbucketCloudUrlReader` or the `BitbucketServerUrlReader` instead.
```

.changeset/tiny-zoos-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-common': minor
+---
+
+**BREAKING** Removed deprecated `bitbucket` integration from being registered in the `ScaffolderClient` use the `bitbucketCloud` or `bitbucketServer` integrations instead.
```
aramissennyeydd:
```suggestion
**BREAKING** Removed deprecated `bitbucket` integration from being registered in the `ScaffolderClient`. Use the `bitbucketCloud` or `bitbucketServer` integrations instead.
```

aramissennyeydd (changes requested):
Nice! Couple of comments on the changesets

packages/integration/src/gerrit/core.ts
tcardonne: Hi @awanlin 👋 

IIRC, the `parseGitilesUrlRef` is more generic than the previous `parseGerritGitilesUrl`, as it can parse URLs that points to a specific commit, tag or just head.

Something like this should work and pass tests, as it doesn't make sense to call `getGerritBranchApiUrl` for non-branches refs:

```
const { ref, refType, project } = parseGitilesUrlRef(config, url);
 if (refType !== 'branch') {
     throw new Error(`Unsupported gitiles ref type: ${refType}`);
 }
```

Hope that helps! And thank you for taking care of this!
Some time ago I wanted to work on doing some cleaning on the Gerrit integration. I think it could be simplified a lot and some parts doesn't feel solid. But didn't had the time at work to do so

awanlin: Thanks will give that a try! I mostly wanted to remove the old Azure DevOps items but as that would mean a major version bump I figured I should remove as much of the other deprecated code as I could. More than happy to see improvements contributed for this from those using the tools in their day to day.

awanlin: Implemented your suggestion @tcardonne, will see how the CI like it and refactor from there. 🚀

packages/integration/config.d.ts
```diff
@@ -27,27 +27,6 @@ export interface Config {
        * @visibility frontend
        */
       host: string;
-      /**
-       * Token used to authenticate requests.
-       * @visibility secret
-       * @deprecated Use `credentials` instead.
-       */
-      token?: string;
```
freben, line 35: I wonder - should we be so brash as to actually refuse startup if any of these deleted root keys exist? If we do, we capture problems early. Including leaking sensitive tokens that no longer are marked as secret, AND failures with auth

awanlin: Yes, that's actually a good call. Where should I add that?

freben: Wherever they used to be read I guess!

Also - this needs a rebase now.

awanlin: Found a good spot for the Azure items, and rebased. 👍

docs/backend-system/building-backends/08-migrating.md
```diff
@@ -1340,105 +1340,3 @@ const backend = createBackend();
 backend.add(import('@backstage/plugin-kubernetes-backend'));
 /* highlight-add-end */
 ```
```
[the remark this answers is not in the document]
awanlin: It's related and can be safely removed at this time.

packages/integration/src/azure/config.ts
```diff
@@ -415,3 +357,19 @@ export function readAzureIntegrationConfigs(
...
+ * don't leak sensitive tokens we have this check in place
+ * to throw an error if found
+ *
+ * @internal
+ * @deprecated To be removed at a later date
+ */
+function deprecatedConfigCheck(config: Config) {
+  if (config.getOptional('credential') || config.getOptional('token')) {
+    throw new Error(
+      `Invalid Azure integration config, 'credential' and 'token' have been removed. Use 'credentials' instead.`,
```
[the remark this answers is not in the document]
awanlin: No this is fine, they both have been removed so this makes sense as a general warning.

PR #32734, opened 2026-02-06
frontend-test-utils: more mockApis and utilities

Rugvip:

This adds a lot more `mockApi` implementations in `frontend-test-utils`. It also introduces a new way to provide mock implementations, by being able to pass mock instances like `mockApis.config()` directory to `renderInTestApp({ apis: [...] })` and friends. This utility is also made available for external use and already picked up by the catalog mock.

Some breaking changes are removal of the `.factory` util that is now redundant, and `TestApiRegistry`, which doesn't have a use anymore.

### DIFF
--- a/docs/frontend-system/utility-apis/05-testing.md
+++ b/docs/frontend-system/utility-apis/05-testing.md
@@ -139,7 +139,7 @@ import {
 import { myApiRef, type MyApi } from '@internal/plugin-example-react';
 
 // Fake instance with real behavior
-export function myApiMock(options?: { greeting?: string }): MyApi {
+export function myApiMock(options?: { greeting?: string }) {
   return attachMockApiFactory(myApiRef, {
     greet: async () => options?.greeting ?? 'Hello!',
   });
--- a/packages/frontend-test-utils/report.api.md
+++ b/packages/frontend-test-utils/report.api.md
@@ -234,7 +234,7 @@ export namespace mockApis {
     email?: string;
     displayName?: string;
     picture?: string;
-  }): IdentityApi & MockWithApiFactory<IdentityApi>;
+  }): MockWithApiFactory<IdentityApi>;
   export namespace identity {
     const // (undocumented)
       mock: (
@@ -376,7 +376,7 @@ export class MockPermissionApi implements PermissionApi {
 // @public
 export class MockStorageApi implements StorageApi {
   // (undocumented)
-  static create(data?: MockStorageBucket): MockStorageApi;
+  static create(data?: JsonObject): MockStorageApi;
   // (undocumented)
   forBucket(name: string): StorageApi;
   // (undocumented)
@@ -390,11 +390,6 @@ export class MockStorageApi implements StorageApi {
   // (undocumented)
   snapshot<T extends JsonValue>(key: string): StorageValueSnapshot<T>;
 }
-
-// @public
-export type MockStorageBucket = {
-  [key: string]: any;
-};
 
 // @public
 export class MockTranslationApi implements TranslationApi {
--- a/packages/frontend-test-utils/src/apis/StorageApi/MockStorageApi.ts
+++ b/packages/frontend-test-utils/src/apis/StorageApi/MockStorageApi.ts
@@ -15,15 +15,8 @@
  */
 
 import { StorageApi, StorageValueSnapshot } from '@backstage/core-plugin-api';
-import { JsonValue, Observable } from '@backstage/types';
+import { JsonObject, JsonValue, Observable } from '@backstage/types';
 import ObservableImpl from 'zen-observable';
-
-/**
- * Type for map holding data in {@link MockStorageApi}
- *
- * @public
- */
-export type MockStorageBucket = { [key: string]: any };
 
 /**
  * Mock implementation of the {@link core-plugin-api#StorageApi} to be used in tests
@@ -32,20 +25,20 @@ export type MockStorageBucket = { [key: string]: any };
  */
 export class MockStorageApi implements StorageApi {
   private readonly namespace: string;
-  private readonly data: MockStorageBucket;
+  private readonly data: JsonObject;
   private readonly bucketStorageApis: Map<string, MockStorageApi>;
 
   private constructor(
     namespace: string,
     bucketStorageApis: Map<string, MockStorageApi>,
-    data?: MockStorageBucket,
+    data?: JsonObject,
   ) {
     this.namespace = namespace;
     this.bucketStorageApis = bucketStorageApis;
     this.data = { ...data };
   }
 
-  static create(data?: MockStorageBucket) {
+  static create(data?: JsonObject) {
     // Translate a nested data object structure into a flat object with keys
     // like `/a/b` with their corresponding leaf values
     const keyValues: { [key: string]: any } = {};
@@ -83,7 +76,7 @@ export class MockStorageApi implements StorageApi {
       return {
         key,
         presence: 'present',
-        value: data,
+        value: data as T,
       };
     }
     return {
--- a/packages/frontend-test-utils/src/apis/StorageApi/index.ts
+++ b/packages/frontend-test-utils/src/apis/StorageApi/index.ts
@@ -15,4 +15,3 @@
  */
 
 export { MockStorageApi } from './MockStorageApi';
-export type { MockStorageBucket } from './MockStorageApi';
--- a/packages/frontend-test-utils/src/apis/index.ts
+++ b/packages/frontend-test-utils/src/apis/index.ts
@@ -78,7 +78,7 @@ export type { MockPermissionApi } from './PermissionApi';
 /**
  * @public
  */
-export type { MockStorageApi, MockStorageBucket } from './StorageApi';
+export type { MockStorageApi } from './StorageApi';
 
 /**
  * @public
--- a/packages/frontend-test-utils/src/apis/mockApis.ts
+++ b/packages/frontend-test-utils/src/apis/mockApis.ts
@@ -313,7 +313,7 @@ export namespace mockApis {
     email?: string;
     displayName?: string;
     picture?: string;
-  }): IdentityApi & MockWithApiFactory<IdentityApi> {
+  }): MockWithApiFactory<IdentityApi> {
     const {
       userEntityRef = 'user:default/test',
       ownershipEntityRefs = ['user:default/test'],


docs/frontend-system/utility-apis/05-testing.md
```diff
@@ -0,0 +1,169 @@
...
+| `mockApis.discovery({ baseUrl })` | Inline                | Returns `${baseUrl}/api/${pluginId}`, defaults to `http://example.com` |
+| `mockApis.error(options?)`        | `MockErrorApi`        | Collects errors; has `getErrors()`, `waitForError()`                   |
+| `mockApis.featureFlags(options?)` | `MockFeatureFlagsApi` | In-memory flag state; has `getState()`, `setState()`, `clearState()`   |
+| `mockApis.fetch(options?)`        | `MockFetchApi`        | Wraps `cross-fetch`; supports identity injection and plugin protocol   |
+| `mockApis.identity(options?)`     | Inline                | Configurable user ref, ownership, token, profile                       |
+| `mockApis.permission(options?)`   | `MockPermissionApi`   | Defaults to `ALLOW`; accepts a handler function                        |
+| `mockApis.storage({ data })`      | `MockStorageApi`      | In-memory storage with bucket support                                  |
+| `mockApis.translation()`          | `MockTranslationApi`  | Passthrough returning default messages from translation refs           |
```
freben, line 192: add to vale

docs/frontend-system/utility-apis/05-testing.md
```diff
@@ -0,0 +1,169 @@
...
+
+```ts
+import { attachMockApiFactory } from '@backstage/frontend-test-utils';
+import { myApiRef, type MyApi } from '@internal/plugin-example-react';
+
+export function myApiMock(options?: { greeting?: string }): MyApi {
+  const instance: MyApi = {
+    greet: async () => options?.greeting ?? 'Hello!',
+  };
+  return attachMockApiFactory(myApiRef, instance);
```
freben: One of the things I miss regularly, is the `simpleMock` and corresponding `ServiceMock` type. In order to get proper jest.fn wrappers and not having to do ugly `as Mocked...` or similar casts in the test code.

Doesn't have to happen right now, but this example reinforces that desire, where the ultimate interface instead then might be

```ts
return createMockApitWithFactory(myApiRef, () => ({
  greet: async ...,
});
```

and it properly comes out with a mock-typed `greet` wrapped in a `jest.fn`. Or maybe it'd be a two step thing where you are able to call `simpleMock` first and then this with the current interface, I don't know.

Again, can be done as a separate thing, but those facilities are so good, that we've copied them internally a bunch of times already.

(edit: I see farther down that this is done partially; at least the type is exported, but the helper (and the default behavior of fn-wrapping) is neat)

packages/frontend-test-utils/src/apis/AlertApi/MockAlertApi.ts
```diff
@@ -0,0 +1,102 @@
...
+ * @example
+ * ```ts
+ * const alertApi = new MockAlertApi();
+ * alertApi.post({ message: 'Test alert' });
+ * expect(alertApi.getAlerts()).toHaveLength(1);
+ * ```
+ */
+export class MockAlertApi implements AlertApi {
+  private alerts: AlertMessage[] = [];
+  private observers = new Set<(alert: AlertMessage) => void>();
```
freben: can be readonly

packages/frontend-test-utils/src/apis/AlertApi/MockAlertApi.ts
```diff
@@ -0,0 +1,102 @@
...
+ * @public
+ * @example
+ * ```ts
+ * const alertApi = new MockAlertApi();
+ * alertApi.post({ message: 'Test alert' });
+ * expect(alertApi.getAlerts()).toHaveLength(1);
+ * ```
+ */
+export class MockAlertApi implements AlertApi {
+  private alerts: AlertMessage[] = [];
```
freben: can be readonly if cleared with splice

packages/frontend-test-utils/src/apis/ConfigApi/MockConfigApi.ts
```diff
@@ -0,0 +1,111 @@
...
+ *     <MyTestedComponent />
+ *   </TestApiProvider>,
+ * );
+ * ```
+ */
+export class MockConfigApi implements ConfigApi {
+  private readonly config: ConfigReader;
+
+  // NOTE: not extending in order to avoid inheriting the static `.fromConfigs`
+  constructor(data: JsonObject) {
```
freben: `mockAps.config` (and its service equivalent) accepts `{ data }`, not `data`. Is there any value in keeping these consistent just because? It's not a real problem, we can add more constructor arguments, but just checking since this is a public class

packages/frontend-test-utils/src/apis/FeatureFlagsApi/MockFeatureFlagsApi.ts
```diff
@@ -0,0 +1,102 @@
...
+ * ```ts
+ * const api = new MockFeatureFlagsApi({
+ *   initialStates: { 'my-feature': FeatureFlagState.Active }
+ * });
+ * expect(api.isActive('my-feature')).toBe(true);
+ * ```
+ */
+export class MockFeatureFlagsApi implements FeatureFlagsApi {
+  private registeredFlags: FeatureFlag[] = [];
+  private states: Map<string, FeatureFlagState>;
```
freben:
```suggestion
  private readonly registeredFlags: FeatureFlag[] = [];
  private readonly states: Map<string, FeatureFlagState>;
```

packages/frontend-test-utils/src/apis/FetchApi/MockFetchApi.ts
```diff
@@ -0,0 +1,169 @@
...
+   * Leaving out this parameter or passing `undefined`, makes the API use the
+   * global `fetch` implementation to make real network requests.
+   *
+   * `'none'` swallows all calls and makes no requests at all.
+   *
+   * You can also pass in any `fetch` compatible callback, such as a
+   * `jest.fn()`, if you want to use a custom implementation or to just track
+   * and assert on calls.
+   */
+  baseImplementation?: undefined | 'none' | typeof crossFetch;
```
freben: i wonder, do we really need cross-fetch here 🤔 maybe it's mostly because of the jest test env itself, okok

packages/frontend-test-utils/src/apis/TranslationApi/MockTranslationApi.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    const interpolator = JsxInterpolator.fromI18n(i18n);
+
+    return new MockTranslationApi(i18n, interpolator);
+  }
+
+  #i18n: I18n;
+  #interpolator: JsxInterpolator;
+  #registeredRefs = new Set<string>();
```
freben:
```suggestion
  readonly #i18n: I18n;
  readonly #interpolator: JsxInterpolator;
  readonly #registeredRefs = new Set<string>();
```

.github/vale/config/vocabularies/Backstage/accept.txt
```diff
@@ -584,3 +584,5 @@ zsh
 resizable
 enums
 LLMs
+Passthrough
+passthrough
```
freben:
```suggestion
passthrough
```

if you add a lowercase one it means "any case", so the one above is not needed

plugins/scaffolder-react/src/next/hooks/useTemplateSchema.test.tsx
```diff
@@ -119,7 +122,12 @@ describe('useTemplateSchema', () => {
       const { result } = renderHook(() => useTemplateSchema(manifest), {
         wrapper: ({ children }: PropsWithChildren<{}>) => (
           <TestApiProvider
-            apis={[[featureFlagsApiRef, { isActive: () => false }]]}
+            apis={[
+              [
+                featureFlagsApiRef,
+                mockApis.featureFlags.mock({ isActive: jest.fn(() => false) }),
```
freben: Do these need to be tuples?

docs/frontend-system/utility-apis/05-testing.md
Rugvip: Yep indeed, let's get it in the while at it, update docs and catalog-react and whatnot. Called it `createApiMock`. I avoided naming it `createMockApi` because that's not actually what you get back. Could've been something like `createMockApiFactory` but I feel `createApiMock` kinda captures the intent very well.

packages/frontend-test-utils/src/apis/FetchApi/MockFetchApi.ts
Rugvip: updated

.github/vale/config/vocabularies/Backstage/accept.txt
Rugvip: Ah, we had some existing examples of double casing

plugins/scaffolder-react/src/next/hooks/useTemplateSchema.test.tsx
Rugvip: Kind of, it was using the old provider from `test-utils`, update now though

packages/frontend-test-utils/report.api.md
```diff
@@ -108,36 +166,273 @@ export class MockAnalyticsApi implements AnalyticsApi {
...
+      ) => ApiMock<FetchApi_2>;
+  }
+  export function identity(options?: {
+    userEntityRef?: string;
+    ownershipEntityRefs?: string[];
+    token?: string;
+    email?: string;
+    displayName?: string;
+    picture?: string;
+  }): IdentityApi & MockWithApiFactory<IdentityApi>;
```
freben: `IdentityApi &` stood out to me here

packages/frontend-test-utils/report.api.md
```diff
@@ -108,36 +166,273 @@ export class MockAnalyticsApi implements AnalyticsApi {
...
+  // (undocumented)
+  set<T>(key: string, data: T): Promise<void>;
+  // (undocumented)
+  snapshot<T extends JsonValue>(key: string): StorageValueSnapshot<T>;
+}
 
-export { MockPermissionApi };
+// @public
+export type MockStorageBucket = {
+  [key: string]: any;
```
freben: Could arguably be `JsonValue`

plugins/scaffolder/src/components/TemplateTypePicker/TemplateTypePicker.test.tsx
```diff
@@ -75,12 +76,7 @@ const apis = TestApiRegistry.from(
       } as GetEntityFacetsResponse),
     },
   ],
-  [
-    alertApiRef,
-    {
-      post: jest.fn(),
-    } as unknown as AlertApi,
-  ],
+  [alertApiRef, mockApis.alert()],
```
freben, line 79: some tuples here too :)

packages/frontend-test-utils/report.api.md
```diff
@@ -108,36 +166,273 @@ export class MockAnalyticsApi implements AnalyticsApi {
...
+// @public
+export interface MockFetchApiOptions {
+  baseImplementation?: undefined | 'none' | typeof fetch;
+  injectIdentityAuth?:
+    | undefined
+    | {
+        token: string;
+      }
+    | {
+        identityApi: Pick<IdentityApi_2, 'getCredentials'>;
```
freben, line 354: Just noting that there are a couple of core-plugin-api types in here (`IdentityApi_2` is from there). No huge deal - but thought it might be worth mentioning.

Rugvip: Yep should naturally move over eventually

plugins/scaffolder/src/components/TemplateTypePicker/TemplateTypePicker.test.tsx
Rugvip: can't be arsed moving all of them :p

packages/frontend-test-utils/report.api.md
Rugvip: ah yep because the other ones generally use `Mock* &`, but this one is obviously redundant

packages/frontend-test-utils/report.api.md
Rugvip: Yeah this is old stuff lifted over, switched out the entire thing for `JsonObject`.

PR #32760, opened 2026-02-09
🧹  remove support for `react-router` beta

freben:


Rugvip (commented):
Hmm, is this worth it? 🤔

We'd need to `major` bump `core-app-api`, but things are already set up so that RR beta will disappear with the switch to NFS.

freben:
Oh so the ruckus may not be worth it ultimately? Alright, that probably makes sense. Close?

Rugvip:
Yeah, I think at some point we ship the CLI change, but bit too early for that

PR #32761, opened 2026-02-09
cli: add support for CSS exports in package build

Rugvip:
## Summary

This adds support for CSS exports in package builds. When a package declares a CSS file in its `exports` field, for example:

```json
{
  "exports": {
    ".": "./src/index.ts",
    "./css/styles.css": "./src/css/styles.css"
  }
}
```

The CLI will now:

1. **At build time**: Detect CSS exports and bundle them as part of the Rollup build, resolving any `@import` statements
2. **At publish time**: Rewrite the export path from `src/` to `dist/`

Our Web(Rs)Pack setup already supports `@import` and CSS entry points, so this already works at dev time, including hot reloads.

The `@backstage/ui` package has been migrated to use this new approach, removing the custom `build-css.mjs` script and the separate build steps from CI workflows.

@backstage/design-system-maintainers ptal 🙏 

### DIFF
--- a/packages/cli/src/modules/maintenance/commands/repo/fix.ts
+++ b/packages/cli/src/modules/maintenance/commands/repo/fix.ts
@@ -27,9 +27,12 @@ import {
   resolve as resolvePath,
   posix,
   relative as relativePath,
+  extname,
 } from 'node:path';
 import { paths } from '../../../../lib/paths';
 import { publishPreflightCheck } from '../../lib/publishing';
+
+const SCRIPT_EXTS = ['.js', '.jsx', '.ts', '.tsx', '.json'];
 
 /**
  * A mutable object representing a package.json file with potential fixes.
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/entryPoints.ts
```diff
@@ -33,18 +33,25 @@ const defaultIndex = {
   ext: '.ts',
 };
 
+const SCRIPT_EXTS = ['.js', '.jsx', '.ts', '.tsx'];
```
freben, line 36: will the `.mjs` family of extensions never come into play?

packages/cli/src/modules/build/lib/packager/productionPack.ts
```diff
@@ -148,7 +148,11 @@ async function rewriteEntryPoints(
 
   for (const entryPoint of entryPoints) {
     if (!SCRIPT_EXTS.includes(entryPoint.ext)) {
```
freben, line 150: Note that this uses a different SCRIPT_EXTS - maybe that's a risk, should it be extracted to a shared const?

packages/cli/src/modules/maintenance/commands/repo/fix.ts
```diff
@@ -27,10 +27,13 @@ import {
   resolve as resolvePath,
   posix,
   relative as relativePath,
+  extname,
 } from 'node:path';
 import { paths } from '../../../../lib/paths';
 import { publishPreflightCheck } from '../../lib/publishing';
 
+const SCRIPT_EXTS = ['.js', '.jsx', '.ts', '.tsx', '.json'];
```
freben, line 35: oh one more 🤔

packages/cli/src/lib/entryPoints.ts
Rugvip: At least for now we only support these for the entry points: https://github.com/backstage/backstage/blob/cad0b68d9ccde7b9c15db140864e0da74545cd46/packages/cli/src/modules/build/lib/builder/config.ts#L46

Honestly probably something to adjust, but thinking that's a separate change.

packages/cli/src/modules/maintenance/commands/repo/fix.ts
Rugvip: Yeah striving to keep these decoupled for the CLI modularization, the duplication isn't great though : /

packages/cli/src/modules/build/lib/packager/productionPack.ts
Rugvip: Yep same as https://github.com/backstage/backstage/pull/32761#discussion_r2786956774

Issue #31755, opened 2025-11-13
Provide actions/mcp tools to interact with the scaffolder


johnmcollier:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
mcp, scaffolder, actions
```

### 🗃️ Project Area

Software Templates

### 🔖 Need

It would be great if there were MCP tools that could interact with the scaffolder in Backstage. e.g. discovery of software templates, dry-running through templates with different parameters, executing templates, getting logs for events, etc). This would allow information around the software templates to be exposed to AI clients (and even allow software template execution done via AI clients). I did a basic search through the scaffolder plugins in backstage/backstage and didn't see any (or any open pull requests).

### 📝 Proposal

Provide a basic set of actions that can interact with the scaffolder and software templates, such as:
- Discovery and validation of software software templates
- Dry run testing of software templates
- Execution of software templates
- Getting scaffolder tasks
- Getting scaffolder task logs 

and package the tools as part of the scaffolder-backend plugin (if there's a better location for this, let me know). 

### 🔄 Alternatives

Templates can be more generically queried via the catalog get action/tool currently available (https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/actions/createGetCatalogEntityAction.ts), which may be sufficient for now.

I'm not sure if there's ways to achieve the other tools, but if there are, let me know.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Yes, but I would like some more guidance

johnmcollier:
If there's interest in these tools being provided, myself and @Jdubrick would be interested in contributing them.

benjdlambert:
Hey :wave: so we discussed this a little bit at KubeCon and I think that most of these make sense.

There is however a few nuances here.

- `Discovery and validation of software software templates`: should this be actually more of a catalog action seen as the templates and validation are done over there?
- `Dry run testing of software templates`: this is normally fine, as the template doesn't usually require GitHub Auth, see below:
[...]

PR #32765, opened 2026-02-09
feat(scaffolder): Create scaffolder mcp action to list all installed template actions

yangcao77:

This PR added scaffolder mcp action to list all installed template actions
similiar as `/create/actions` endpoint, it will fetch all installed actions from `templateActionRegistry` and then return the list, with input, output and examples. 

tested with cursor mcp server enabled. able to use the mcp tool to get all actions, and even input from a specific type of actions. see screenshot:

mcp-tool1
mcp-tool2

### DIFF
--- a/plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
+++ b/plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
@@ -36,7 +36,7 @@ Each action includes:
 - id: The action identifier
 - description: What the action does
 - schema: Input and output JSON schemas
-- examples: Usage examples (if available)`,
+- examples: Usage examples when available`,
     schema: {
       input: z => z.object({}).describe('No input is required'),
       output: z =>
--- a/.changeset/tangy-clouds-repeat.md
+++ b/.changeset/tangy-clouds-repeat.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-scaffolder-backend': minor
 ---
 
-create scaffolder MCP action to list all installed scaffolder template actions
+Added a new `list-scaffolder-actions` action that returns all installed scaffolder actions with their schemas and examples
--- a/plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.test.ts
+++ b/plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.test.ts
@@ -16,17 +16,22 @@
 import { createListScaffolderActionsAction } from './createListScaffolderActionsAction';
 import { actionsRegistryServiceMock } from '@backstage/backend-test-utils/alpha';
 import { scaffolderServiceMock } from '@backstage/plugin-scaffolder-node/testUtils';
-import type {
-  Action,
-  ListActionsResponse,
-} from '@backstage/plugin-scaffolder-common';
+import type { ListActionsResponse } from '@backstage/plugin-scaffolder-common';
+
+type ListActionsOutput = {
+  actions: Array<{
+    id: string;
+    description: string;
+    schema: { input: object; output: object };
+    examples: Array<{ description: string; example: string }>;
+  }>;
+};
 
 describe('createListScaffolderActionsAction', () => {
-  it('should list all scaffolder actions successfully', async () => {
+  it('should list all scaffolder actions sorted by id with full properties', async () => {
     const mockActionsRegistry = actionsRegistryServiceMock();
     const mockScaffolderService = scaffolderServiceMock.mock();
-    const mockActions = createMockActions();
-    mockScaffolderService.listActions.mockResolvedValue(mockActions);
+    mockScaffolderService.listActions.mockResolvedValue(createMockActions());
 
     createListScaffolderActionsAction({
       actionsRegistry: mockActionsRegistry,
@@ -38,23 +43,17 @@ describe('createListScaffolderActionsAction', () => {
       input: {},
     });
 
-    // Verify output structure
-    expect(result.output).toHaveProperty('actions');
-    const output = result.output as any;
-    expect(Array.isArray(output.actions)).toBe(true);
+    const output = result.output as ListActionsOutput;
     expect(output.actions).toHaveLength(3);
 
-    // Verify actions are sorted by id
-    const actionIds = output.actions.map((a: any) => a.id);
+    const actionIds = output.actions.map(a => a.id);
     expect(actionIds).toEqual([
       'catalog:register',
       'debug:log',
       'fetch:template',
     ]);
 
-    // Verify action structure
-    const firstAction = output.actions[0];
-    expect(firstAction).toEqual({
+    expect(output.actions[0]).toEqual({
       id: 'catalog:register',
       description: 'Registers entities in the catalog',
       schema: {
@@ -82,13 +81,13 @@ describe('createListScaffolderActionsAction', () => {
       input: {},
     });
 
-    const output = result.output as any;
+    const output = result.output as ListActionsOutput;
     expect(output.actions).toHaveLength(1);
     expect(output.actions[0]).toEqual({
       id: 'minimal-action',
-      description: '', // Empty string when no description
-      schema: { input: {}, output: {} }, // Default empty schema
-      examples: [], // Empty array when no examples
+      description: '',
+      schema: { input: {}, output: {} },
+      examples: [],
     });
   });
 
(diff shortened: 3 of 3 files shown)


plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
```diff
@@ -0,0 +1,95 @@
...
+      if (templateActionRegistry) {
+        const actionsMap = await templateActionRegistry.list({ credentials });
+        actionsList = Array.from(actionsMap.values()).map(action => ({
+          id: action.id,
+          description: action.description || '',
+          schema: action.schema || { input: {}, output: {} },
+          examples: action.examples || [],
+        }));
+      } else {
+        throw new Error('templateActionRegistry must be provided');
```
johnmcollier: new Errors / non-backstage errors typically won't get surfaced by the MCP server when this action is called through it. It's better to use one of the built-in error types https://backstage.io/api/stable/modules/_backstage_errors.html when possible

(`InputError` I suppose?)

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
```diff
@@ -0,0 +1,95 @@
...
+                  .array(z.any())
+                  .optional()
+                  .describe('Usage examples'),
+              }),
+            )
+            .describe('List of all available scaffolder actions'),
+        }),
+    },
+    action: async ({ credentials }) => {
+      let actionsList: Array<{
```
johnmcollier: I saw that the `TemplateAction` type returned by `templateActionRegistry.list` also have a `handler` field? Any reason why we wouldn't want to return it as well?

The reason I'm asking is that if we include that, we can just simplify this to be an `Array<TemplateAction>`.

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
```diff
@@ -0,0 +1,95 @@
...
+    },
+    description: `Lists all installed Scaffolder actions.
+Each action includes:
+- id: The action identifier
+- description: What the action does
+- schema: Input and output JSON schemas
+- examples: Usage examples (if available)`,
+    schema: {
+      input: z => z.object({}),
+      output: z =>
```
johnmcollier, line 42: If we want to avoid duplicating the `TemplateAction` type's fields we could simplify this to just be `z.object` similar to what the [get-catalog-entity](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/actions/createGetCatalogEntityAction.ts#L59) action does.

There's downsides though (as we're no longer restricting the output schema to just the `TemplateAction` type, but I'm not sure how to enforce that using zod schemas with TypeScript types

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
yangcao77: I was trying to return `Array<TemplateAction>` as well, but so I was experiencing error. since the handler field is a Promise function which cannot be serialized to JSON, and therefore can't be sent over MCP protocol

yangcao77: and also, from security perspective, we should not expose internal implementation hanlder functions neither. so only I return the metadata that's useful for discovering and understanding actions

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
yangcao77: updated to use `InputError`

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
yangcao77: good point. I've updated to use `z.object({}).passthrough()`. it can still be used as we defined the action object as well.

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
johnmcollier: Yeah, that makes sense

plugins/scaffolder-backend/src/ScaffolderPlugin.ts
```diff
@@ -159,10 +165,21 @@ export const scaffolderPlugin = createBackendPlugin({
         events,
         auditor,
         actionsRegistry,
+        actionsRegistryService,
       }) {
         const log = loggerToWinstonLogger(logger);
         const integrations = ScmIntegrations.fromConfig(config);
 
+        // Create a scoped logger for scaffolder plugin consistent log formatting
+        const scopedLogger = logger.child({ plugin: 'scaffolder' });
```
benjdlambert: This should already happen in the `LoggerService`.

benjdlambert: https://github.com/backstage/backstage/blob/6c11f6e7e77542a8d9969f0f2fc9b1ffc48eab6b/packages/backend-defaults/src/entrypoints/logger/loggerServiceFactory.ts#L38

plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -164,6 +165,7 @@ export interface RouterOptions {
   auditor?: AuditorService;
   autocompleteHandlers?: Record<string, AutocompleteHandler>;
   actionsRegistry: ActionsService;
```
benjdlambert: I think that `createRouter` is no longer exported, so we can break this API internally.

Let's just change `options.actionsRegistry` to be `TemplateActionRegistry` instead?

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
```diff
@@ -0,0 +1,76 @@
...
+      idempotent: true,
+    },
+    description: `Lists all installed Scaffolder actions.
+Each action includes:
+- id: The action identifier
+- description: What the action does
+- schema: Input and output JSON schemas
+- examples: Usage examples (if available)`,
+    schema: {
+      input: z => z.object({}),
```
benjdlambert: I think it would be smart to describe these fields with `.description('JSON schema representing the input parameter for the action)`. Same for output.

plugins/scaffolder-backend/src/ScaffolderPlugin.ts
yangcao77: good to know that. removed this block of code

plugins/scaffolder-backend/src/service/router.ts
yangcao77: I updated `actionsRegistry` to `TemplateActionRegistry` in router.ts, and also `TaskWorker` as well as `NunjucksWorkflow`
to avoid future confusion between the MCP `actionRegistry` v.s.  scaffolder`templateActionRegistry`

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
yangcao77: updated

plugins/scaffolder-backend/src/service/router.ts
benjdlambert: But why pass through both? 

We should just pass through `templateActionRegistry` and remove the `new DefaultTemplateActionRegistry` from this file as it's now going to be created outside instead and always given to the router?

yangcao77: you are right. the old `actionRegistry` can be removed as `templateActionRegistry` is already created in `ScaffolderPlugin` and being passed in. and it is the only place where router is created & being used. 

I have made the change & committed.

yangcao77:
@benjdlambert I've addressed your review comment. PTAL with you have time.

yangcao77:
@benjdlambert I've addressed your review comment. PTAL

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
```diff
@@ -0,0 +1,76 @@
...
+Each action includes:
+- id: The action identifier
+- description: What the action does
+- schema: Input and output JSON schemas
+- examples: Usage examples (if available)`,
+    schema: {
+      input: z => z.object({}).describe('No input is required'),
+      output: z =>
+        z
+          .object({})
```
benjdlambert: I think that we should look at the below Copilot message.

This should be typed better, we can type `schema.input` and `schema.output` as passthrough, and add a `.describe()` to those fields to state that they are `JSONSchema for input of Action` and `JSONSchema for output of Action`.

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
```diff
@@ -0,0 +1,82 @@
...
+          .describe(
+            'JSON schema representing the output parameter for the action',
+          ),
+    },
+    action: async ({ credentials }) => {
+      let actionsList: Array<{
+        id: string;
+        description: string;
+        schema: any;
+        examples?: any[];
```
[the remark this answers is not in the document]
benjdlambert: I don't actually think we will need this once we type the `output` schema properly above. We should just be able to do 

```ts
const scaffolderActions = await templateActionRegistry.list({ credentials });
return {
  output: {
    actions: scaffolderActions.sort((a,b) => a.id.localeCompare(b.id))
  },
};
```

benjdlambert (changes requested):
Hey :wave:

Thanks for this, just a few comments here :pray:

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
yangcao77: if I understood correctly, updated the action output to define the `actions` list with `id, description, schema and examples`. and passthrough `schema.input` and `schema.output`

plugins/scaffolder-backend/src/actions/createListScaffolderActionsAction.ts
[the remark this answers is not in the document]
yangcao77: `templateActionRegistry.list` returns a map, and cannot be sort. 
I have updated the code as suggested, but still need convert the map to an array before the sort.

yangcao77:
@benjdlambert I've also updated this PR to use the new `scaffolderService` instead.

PR #32771, opened 2026-02-09
backend-test-utils: add `createServiceMock`

Rugvip:

This implements the same pattern as `createApiMock` in #32734 but for the backend system. The internal `simpleMock` utility is now exported as `createServiceMock` from `@backstage/backend-test-utils`, letting plugin authors create standardized mock factories for their own service refs.

Also updates `catalog-node` to use it instead of its own copy of `simpleMock`, and does a small cleanup in `gateway-backend` to use `discovery.factory` directly.

### DIFF
--- a/.changeset/repo-tools-type-deps-ambient-jest.md
+++ b/.changeset/repo-tools-type-deps-ambient-jest.md
@@ -2,4 +2,4 @@
 '@backstage/repo-tools': patch
 ---
 
-The `type-deps` command now detects ambient global types from the `jest` namespace in declaration files, rather than only looking for explicit imports and reference directives.
+The `type-deps` command now follows relative imports and re-exports into declaration chunk files, and detects ambient global types such as the `jest` namespace.
--- a/packages/repo-tools/src/commands/type-deps/type-deps.ts
+++ b/packages/repo-tools/src/commands/type-deps/type-deps.ts
@@ -78,6 +78,50 @@ function shouldCheckTypes(pkg: Package & PackageJsonWithTypes) {
   );
 }
 
+/**
+ * Read a declaration file and recursively follow all relative imports and
+ * re-exports, collecting the source content of each file encountered.
+ */
+function collectDeclSources(entryPath: string): string[] {
+  const visited = new Set<string>();
+  const sources: string[] = [];
+
+  function visit(filePath: string) {
+    const resolved = resolvePath(filePath);
+    if (visited.has(resolved)) {
+      return;
+    }
+    visited.add(resolved);
+
+    // Handle .js -> .d.ts extension mapping used by declaration chunk files
+    let actualPath = resolved;
+    if (!fs.existsSync(actualPath)) {
+      actualPath = resolved.replace(/\.js$/, '.d.ts');
+    }
+    if (!fs.existsSync(actualPath)) {
+      return;
+    }
+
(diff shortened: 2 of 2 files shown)


plugins/gateway-backend/src/plugin.test.ts
freben, line 57: This could be rewritten into `.mock({ getBaseUrl: async ...` now

plugins/catalog-node/src/testUtils/catalogServiceMock.ts
```diff
@@ -17,41 +17,14 @@
 import {
   createServiceFactory,
   ServiceFactory,
-  ServiceRef,
 } from '@backstage/backend-plugin-api';
 import { InMemoryCatalogClient } from '@backstage/catalog-client/testUtils';
 import { Entity } from '@backstage/catalog-model';
 import { catalogServiceRef } from '@backstage/plugin-catalog-node';
 // eslint-disable-next-line @backstage/no-undeclared-imports
```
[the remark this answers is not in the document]
freben: This is correct, right?

packages/backend-test-utils/src/alpha/services/simpleMock.ts
```diff
@@ -13,45 +13,8 @@
...
-      factory: createServiceFactory({
-        service: ref,
-        deps: {},
-        factory: () => mock,
-      }),
-    }) as ServiceMock<TService>;
-  };
-}
+/** @alpha */
+export type { ServiceMock } from '../../services/simpleMock';
```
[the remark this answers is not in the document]
freben: I think this is correct and should be a separate import + export

packages/backend-test-utils/src/services/createServiceMock.ts
freben, line 1: could rename this file to `createServiceMock.ts`

plugins/gateway-backend/src/plugin.test.ts
```diff
@@ -68,13 +67,7 @@ describe('gateway', () => {
...
     backend.add(mockServices.auth.factory());
     backend.add(mockServices.httpAuth.factory());
-    backend.add(
-      createServiceFactory({
-        service: coreServices.discovery,
-        deps: {},
-        factory: () => discovery,
-      }),
-    );
+    backend.add(discovery.factory);
```
freben, line 71: Tempting to make `.factory` unnecessary here too, but at the same time we probably don't want to pollute the regular backend interface with that capability. If we make a dedicated test backend perhaps.

Rugvip: Yep, I'd already be a breaking change given our existing mock APIs, so let's consider that for later. Agree it might be worth for the test utils at least though.

packages/backend-test-utils/src/alpha/services/simpleMock.ts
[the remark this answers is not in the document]
Rugvip: Full import from the package entry at usage is the way to go about this, i.e. `import { createServiceMock } from '@backstage/backend-test-utils'`

Rugvip:
Kept the alpha exports as they were in the end. Using an explicit import from the main entry point via the package name is usually the way to go, but in this case that fails due to a circular dependency. Redeclaring with import/export doesn't work either because the type aliasing gets resolved too aggressively. I don't think the `@public` on the alpha export is a big deal.

PR #32775, opened 2026-02-09
workflows/sync_pull-requests: more reliable triggering

Rugvip:

Running into some issues with PRs from forks. Also hoping that this might fix the need to approve workflows triggered, but less confident about that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,18 @@ jobs:
         with:
           egress-policy: audit
 
+      # Store the PR number for the Sync Pull Requests workflow
+      - name: Save PR context
+        if: matrix.node-version == '22.x'
+        run: |
+          mkdir -p ./context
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -32,6 +32,18 @@ jobs:
         with:
           egress-policy: audit
 
+      # Store the PR number for the Sync Pull Requests workflow
+      - name: Save PR context
+        if: matrix.node-version == '22.x'
```
freben, line 37: Oh we have to remember to update this in the future. Same below. Is it just here to get rid of race conditions or does `upload-artifact` actually refuse overwrites or something?

Rugvip: Indeed. Yep started with just an overwrite but then this 🤖 came along and didn't like that: https://github.com/backstage/backstage/pull/32775#discussion_r2785047643

PR #32777, opened 2026-02-10
Expose group filter extension on scaffolder page

### LINKED ISSUE #32776: Expose grouping of software templates through New Frontend System
### 📜 Issue Labels

### 🔎 Search Terms

```plain
scaffoder grouping groups software tempaltes NFS
```

### 🗃️ Project Area

Software Templates

### 🔖 Need

In the old system it was easy to group software templates by passing a groups prop to the ScaffolderPage component 

```
<ScaffolderPage
  groups={[
    {
      title: 'Recommended',
      filter: entity =>
        entity?.metadata?.tags?.includes('recommended') ?? false,
    },
  ]}
/>
```

In the new frontend system this grouping property is not exposed through any of the plugin extensions on the scaffolder plugin 

### 📝 Proposal

Expose the groups as an extension to the `page:scaffolder` keeping the same signature as the existing prop `TemplateGroupFilter` to make it easy to migrate from the old system. There is still a group prop on the scaffolder router (https://github.com/backstage/backstage/blob/master/plugins/scaffolder/src/components/Router/Router.tsx#L87) so we just need to expose the extension then pass any configured grouping to the router when we make it in https://github.com/backstage/backstage/blob/master/plugins/scaffolder/src/alpha/extensions.tsx#L60

I'm looking at the existing formFields extension on the scaffolderPage as an example on how to accomplish this. https://github.com/backstage/backstage/blob/master/plugins/scaffolder-react/src/next/blueprints/FormFieldBlueprint.tsx 
https://github.com/backstage/backstage/blob/master/plugins/scaffolder/src/alpha/extensions.tsx#L57

### 🔄 Alternatives

Need to expose this functionality somehow to allow users with many templates provide common sense filters. We could change the way we do this to allow for the filtering to be defined in YAML somewhere, when searching around it looks like the catalog might have some filtering options with EntityPredicates (https://github.com/backstage/backstage/blob/master/plugins/catalog-react/src/alpha/predicates/entityPredicateToFilterFunction.ts). 

For now I think it makes the most sense to continue to use the same pattern. 

jChapman:
Updated app-next to use the newly exposed extension so software templates are filter just like on the non new frontend version.

closes #32776 

image

Exposed group via extension so you can set it via NFS. Example from the updated app-next App.tsx:

```
const customizedScaffolder = scaffolderPlugin.withOverrides({
  extensions: [
    TemplateGroupBlueprint.make({
      params: {
        title: 'Recomended',
        filter: entity =>
          entity?.metadata?.tags?.includes('recommended') ?? false,
      },
    }),
  ],
});

const app = createApp({
  features: [
    customizedCatalog,
@@ -156,6 +170,7 @@ const app = createApp({
    customHomePageModule,
    devtoolsPlugin,
    devtoolsPluginUnprocessed,
    customizedScaffolder,
    ...collectedLegacyPlugins,
  ],
  ```

Copied heavily from the FormField extension, so I have some questions and want to learn and make any changes suggested (hence why I made this a draft PR). 

### Questions 
1. Should I prefer making a blueprint like I did (when I copied from FormField or should I do something more like the `plugin.getExtension().override` I see in places like the catalog?

```
const customizedCatalog = catalogPlugin.withOverrides({
  extensions: [
    catalogPlugin.getExtension('entity-content:catalog/overview').override({
      params: {
        icon: <InfoIcon />,
      },
    }),
  ],
});
```

2. Any good examples of how to test something like this? 
3. I don't fully understand the API reports, but I see how to run them and I can add those to a changeset 

    - Draft PR, will do when submitting full PR
    - This is a good idea, could update these docs to include this option (IDK how you're doing new front end vs old in docs): https://backstage.io/docs/features/software-templates/configuration/#customizing-the-scaffolderpage-with-grouping-and-filtering
    - Again, looking for suggestions on how to test this
    - I did do at least this one

### DIFF
--- a/packages/app-next/src/App.tsx
+++ b/packages/app-next/src/App.tsx
@@ -49,6 +49,8 @@ import devtoolsPlugin from '@backstage/plugin-devtools/alpha';
 import { unprocessedEntitiesDevToolsContent } from '@backstage/plugin-catalog-unprocessed-entities/alpha';
 import catalogPlugin from '@backstage/plugin-catalog/alpha';
 import InfoIcon from '@material-ui/icons/Info';
+import scaffolderPlugin from '@backstage/plugin-scaffolder/alpha';
+import { TemplateGroupBlueprint } from '@backstage/plugin-scaffolder-react/alpha';
 
 /*
 
@@ -142,6 +144,18 @@ const collectedLegacyPlugins = convertLegacyAppRoot(
   </FlatRoutes>,
 );
 
+const customizedScaffolder = scaffolderPlugin.withOverrides({
+  extensions: [
+    TemplateGroupBlueprint.make({
+      params: {
+        title: 'Recomended',
+        filter: entity =>
+          entity?.metadata?.tags?.includes('recommended') ?? false,
+      },
+    }),
+  ],
+});
+
 const app = createApp({
   features: [
     customizedCatalog,
@@ -156,6 +170,7 @@ const app = createApp({
     customHomePageModule,
     devtoolsPlugin,
     devtoolsPluginUnprocessed,
+    customizedScaffolder,
     ...collectedLegacyPlugins,
   ],
   advanced: {
--- a/plugins/scaffolder-react/src/next/blueprints/TemplateGroupBlueprint.ts
+++ b/plugins/scaffolder-react/src/next/blueprints/TemplateGroupBlueprint.ts
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2026 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import {
+  createExtensionBlueprint,
+  createExtensionDataRef,
+} from '@backstage/frontend-plugin-api';
+import { TemplateGroupFilter } from '@backstage/plugin-scaffolder-react';
+
+const templateGroupDataRef = createExtensionDataRef<TemplateGroupFilter>().with(
+  {
+    id: 'scaffolder.template-group-filter',
+  },
+);
+
+/**
+ * @alpha
+ * Creates extensions that allow for template grouping within the Scaffolder
+ */
+export const TemplateGroupBlueprint = createExtensionBlueprint({
+  kind: 'scaffolder-template-group-filter',
+  attachTo: { id: 'page:scaffolder', input: 'groups' },
+  dataRefs: {
+    templateGroup: templateGroupDataRef,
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: plugins/scaffolder-react/src/next/blueprints/TemplateGroupBlueprint.ts [hunk 1]
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2026 The Backstage Authors
```
jChapman, line 2: Not sure if you update these as you add new files

packages/app-next/src/App.tsx
```diff
@@ -54,6 +49,8 @@ import devtoolsPlugin from '@backstage/plugin-devtools/alpha';
 import { unprocessedEntitiesDevToolsContent } from '@backstage/plugin-catalog-unprocessed-entities/alpha';
 import catalogPlugin from '@backstage/plugin-catalog/alpha';
 import InfoIcon from '@material-ui/icons/Info';
+import scaffolderPlugin from '@backstage/plugin-scaffolder/alpha';
+import { TemplateGroupBlueprint } from '@backstage/plugin-scaffolder-react/alpha';
```
jChapman, line 53: Is it OK to require plugin-scaffolder -react as well as plugin-scaffolder?

benjdlambert:
Hey :wave: thanks for this!

I'm actually thinking that a good place to implement this would in the `PageExtension` config and using https://github.com/backstage/backstage/pull/32631 to create a filter expression out of the config.

Maybe we /could/ support both, but it's a little easier to update config rather than writing code to provide these groups is my thinking so far?

What do you think?

jChapman:
> Hey 👋 thanks for this!
> 
> I'm actually thinking that a good place to implement this would in the `PageExtension` config and using #32631 to create a filter expression out of the config.
> 
> Maybe we /could/ support both, but it's a little easier to update config rather than writing code to provide these groups is my thinking so far?
> 
> What do you think?

I think we should eventually support using filter predicates for grouping templates for sure, so IMO the question is if we want to still support the old method of defining groups. 

Pros: 
- Ease of transition from old format 
- Easy to understand and use (array of functions that take an entity and return a bool? pretty simple IMO)

Cons:
- Cannot define groups this way in YAML 
  - Assuming we eventually use bring in filter predicates then you just use those instead if you want to do this, however...
- A second way to do the same thing as filter predicates 

IMO we should just go with filter predicates and not allow this old method of grouping. 

While researching this enough to submit a good issue I found what looked like a surprisingly simple path to implementing/exposing the old grouping method so I just went with it. Happy to take a look at how to accomplish this with the new(er) filter-predicates package, let me know if you have any suggestions or resources to look at.

awanlin:
Just want to follow up that I agree with Ben, I'd very much like to see this part of the current [`scaffolderPage`](https://github.com/awanlin/backstage/blob/36573c6dd0aae08267d9bcee2fb71c25d153d8f0/plugins/scaffolder/src/alpha/extensions.tsx#L34) that way we could do something like this in the config:

```yaml
- page:scaffolder:
        config:
          groups:
            - title: 'Recommended'
              filter:
                metadata.tags:
                  $in: ['recommended']
``` 

This section in the docs outlines all the awesome filtering you can do in YAMl using what was mentioned: https://backstage.io/docs/features/software-catalog/catalog-customization#entity-filters

Seems a shame to have to create a new Blueprint for this. One of the core features of the New Frontend System is to help reduce the amount of code you need to write and to do more in the config. 👍

Issue #32764, opened 2026-02-09
🐛 Bug Report: O(n²) performance bottleneck in buildEntitySearch traverse() causes event loop blocking


pedronastasi:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
buildEntitySearch performance — Found 1 issue (#30216) about a 200-char field length limit, not performance-related

catalog entity search O(n²) — Found 3 unrelated issues (JWKS refresh, group leadership, PostgreSQL auth)

traverse toLocaleLowerCase — Found 0 issues

buildEntitySearch Array.some duplicate — Found 0 issues

catalog search performance CPU — Found many results but none about the specific buildEntitySearch traverse() bottleneck
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

N/A

### 📝 Description & Context

📜 Description
The buildEntitySearch function in catalog-backend has a critical O(n²) performance bottleneck that causes severe event loop blocking, especially when processing entities with large arrays (100+ tags or similar fields). This manifests as:

- Event loop delays spiking to 600-700ms (normal: <100ms)
- Health check timeouts causing pod restarts
- 100% CPU utilization on the Node.js event loop thread
- Production errors every 5 hours correlating with entity processing cycles

The root cause is in the `traverse()` method at line [115](https://github.com/backstage/backstage/blob/875123707c25a175fc1dc1828cb77ba405cb5f96/plugins/catalog-backend/src/database/operations/stitcher/buildEntitySearch.ts#L115):

```
...
 if (typeof item === 'string') {
          const pathKey = `${path}.${item}`;
          if (
            !output.some(
              kv =>
                kv.key.toLocaleLowerCase('en-US') ===
                pathKey.toLocaleLowerCase('en-US'),
            )
          ) {
            output.push({ key: pathKey, value: true });
          }
        }
...
```

For each string item, `Array.some()` performs a linear search through the growing output array with repeated `toLocaleLowerCase()` calls, resulting in O(n²) complexity.

### 👍 Expected Behavior

Entity processing should not block the event loop for extended periods. The `buildEntitySearch` function should have O(n) or O(n log n) complexity to handle entities with large arrays efficiently.

👎 Actual Behavior
When processing entities with 100+ tags, similar large arrays, or complex objects:

- Event loop iterations drop from 400/sec to 150/sec
[...]

PR #32786, opened 2026-02-10
Fix O(n²) performance bottleneck in buildEntitySearch traverse()

pedronastasi:

`traverse()` in `buildEntitySearch.ts` uses `Array.some()` to deduplicate synthetic path keys, scanning the entire output array with `toLocaleLowerCase()` on every comparison. This is O(n²) and blocks the event loop for 600-700ms on entities with 100+ tags, causing health check failures and pod restarts when complex entities are processed.

| Before | After |
|--------|-------|
|image  | image |

### Changes

- Replace `Array.some()` linear scan with a `Set<string>` for O(1) duplicate path key detection
- Call `toLocaleLowerCase('en-US')` once per item instead of once per comparison

No behavioral change — same output, ordering, and case-insensitive dedup semantics.

### Related issue
https://github.com/backstage/backstage/issues/32764

### DIFF
--- a/plugins/catalog-backend/src/database/operations/stitcher/buildEntitySearch.test.ts
+++ b/plugins/catalog-backend/src/database/operations/stitcher/buildEntitySearch.test.ts
@@ -63,7 +63,6 @@ describe('buildEntitySearch', () => {
       const output = traverse(input);
       const elapsed = performance.now() - start;
 
-      // Each item produces 2 entries: { key: 'tags', value: 'tag-N' } and { key: 'tags.tag-N', value: true }
       expect(output).toHaveLength(1000);
       // Should complete well under 100ms with O(n); O(n²) would be noticeably slower
       expect(elapsed).toBeLessThan(100);


plugins/catalog-backend/src/database/operations/stitcher/buildEntitySearch.test.ts
```diff
@@ -50,6 +50,23 @@ describe('buildEntitySearch', () => {
...
+      const items = Array.from({ length: 500 }, (_, i) => `tag-${i}`);
+      const input = { tags: items };
+
+      const start = window.performance.now();
+      const output = traverse(input);
+      const elapsed = window.performance.now() - start;
+
+      expect(output).toHaveLength(1000);
+      // Should complete well under 100ms with O(n); O(n²) would be noticeably slower
+      expect(elapsed).toBeLessThan(100);
```
freben, line 68: Definite risk for flakiness here - but let's see how it goes

freben (approved):

plugins/catalog-backend/src/database/operations/stitcher/buildEntitySearch.test.ts
freben: Note that the test uses `window` which is not available in node tests. Did you mean to import from `node:perf_hooks`? https://nodejs.org/api/perf_hooks.html#performancenow

pedronastasi: Thanks for the callout. 😄  I'll fix it 😄

pedronastasi: The elapsed time is `~2` ms locally for this test.. `100`ms seems to have a large error margin, and it  still covers the `1105`ms it takes with  the older logic 🤔  🤞

PR #32812, opened 2026-02-12
frontend-plugin-api: new system for page extensions

Rugvip:

WIP, but getting there

This is a merging of BUI and NFS that brings together the new app header, page headers, a customizable page layout, and the support for extensible sub pages for a plugin. Plugins as well as pages can now also register both titles and icons. Ultimately the goal of this is to deprecate `NavItemBlueprint` and start populating the sidebar and structuring the information architecture based on pages and subpages instead.

The new design is best showcased in the app visualizer in this PR:

image

Each of the Tree / Detailed / Text tabs are subpages of the app visualizer page, while the visualizer page is now an empty placeholder that other plugins/modules can extend as well. The "Copy as JSON" button is one of the new plugin-wide header actions.

This also introduces a pattern to hide the `core-components` page header when not need, so far applied to the API docs in this PR:

image

Some pages still need the header for actions until migrated, like the catalog:

image

All of this is for NFS, old system remains as is

### DIFF
--- a/packages/core-components/src/layout/Page/PageWithHeader.tsx
+++ b/packages/core-components/src/layout/Page/PageWithHeader.tsx
@@ -21,14 +21,13 @@ import { Page } from './Page';
 
 type PageWithHeaderProps = ComponentProps<typeof Header> & {
   themeId: string;
-  noHeader?: boolean;
 };
 
 export function PageWithHeader(props: PropsWithChildren<PageWithHeaderProps>) {
-  const { themeId, children, noHeader, ...restProps } = props;
+  const { themeId, children, ...restProps } = props;
   return (
     <Page themeId={themeId}>
-      {noHeader ? <div /> : <Header {...restProps} />}
+      <Header {...restProps} />
       {children}
     </Page>
   );
--- a/packages/frontend-plugin-api/src/blueprints/PageBlueprint.tsx
+++ b/packages/frontend-plugin-api/src/blueprints/PageBlueprint.tsx
@@ -89,8 +89,8 @@ export const PageBlueprint = createExtensionBlueprint({
 
         return (
           <PageLayout
-            title={title}
-            icon={icon}
+            title={title ?? node.spec.plugin.title ?? node.spec.plugin.pluginId}
+            icon={icon ?? node.spec.plugin.icon}
             noHeader={noHeader}
             headerActions={headerActions}
           >
--- a/packages/frontend-plugin-api/src/blueprints/SubPageBlueprint.tsx
+++ b/packages/frontend-plugin-api/src/blueprints/SubPageBlueprint.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 The Backstage Authors
+ * Copyright 2026 The Backstage Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/packages/frontend-plugin-api/src/apis/definitions/PluginHeaderActionsApi.ts
+++ b/packages/frontend-plugin-api/src/apis/definitions/PluginHeaderActionsApi.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 The Backstage Authors
+ * Copyright 2026 The Backstage Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/plugins/app/src/apis/PluginHeaderActionsApi/DefaultPluginHeaderActionsApi.tsx
+++ b/plugins/app/src/apis/PluginHeaderActionsApi/DefaultPluginHeaderActionsApi.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 The Backstage Authors
+ * Copyright 2026 The Backstage Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


packages/frontend-plugin-api/src/blueprints/HeaderActionBlueprint.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+import {
+  createExtensionBlueprint,
+  createExtensionBlueprintParams,
+  createExtensionDataRef,
+} from '../wiring';
+
+const actionDataRef = createExtensionDataRef<() => Promise<JSX.Element>>().with(
```
freben: The type of this makes me think. Should we either:

- Have a `coreExtensionData.reactElementLoader` (keep it granular, but standardise), or
- Make this more like `createExtensionDataRef<{ loader: ... }>` (a more complex type)

Essentially still unsure about the right granularity for these. Either way, the naming "action" seems to me that it might be much more than just an element loader eventually, for example `placement: right` or `group: personal` or stuff like that being added later. As long as we have sensible room to grow the types somewhere

packages/frontend-plugin-api/src/blueprints/HeaderActionBlueprint.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * and will appear in the header of all pages belonging to that plugin.
+ *
+ * @public
+ */
+export const HeaderActionBlueprint = createExtensionBlueprint({
+  kind: 'header-action',
+  attachTo: { id: 'api:app/header-actions', input: 'actions' },
+  output: [actionDataRef],
+  dataRefs: {
+    action: actionDataRef,
```
freben: So to clarify - should this be named `elementLoader: ....` or should `actionDataRef` be a thicker type

packages/frontend-plugin-api/src/blueprints/HeaderActionBlueprint.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * Creates extensions that provide plugin-scoped header actions.
+ *
+ * @remarks
+ *
+ * These actions are automatically scoped to the plugin that provides them
+ * and will appear in the header of all pages belonging to that plugin.
+ *
+ * @public
+ */
+export const HeaderActionBlueprint = createExtensionBlueprint({
```
freben: Naming wise, it is worth naming this (and the header itself) in a big way, with a prefix word like "top" or "app" or "static" or "global" or something in them?

packages/frontend-plugin-api/src/icons/types.ts
```diff
@@ -31,7 +31,15 @@ import { ComponentType } from 'react';
...
 export type IconComponent = ComponentType<{
   fontSize?: 'medium' | 'large' | 'small' | 'inherit';
 }>;
+
+/**
+ * The type used for icon elements throughout Backstage.
+ *
+ * @public
+ */
+export type IconElement = JSX.Element | null;
```
freben, line 49: Alright so that is the JSX symbol we're meant to use externally

packages/frontend-plugin-api/report.api.md
```diff
@@ -1421,14 +1425,51 @@ export const googleAuthApiRef: ApiRef<
 >;
 
 // @public
+export const HeaderActionBlueprint: ExtensionBlueprint_2<{
+  kind: 'header-action';
```
freben: I wonder if we'll ever annoy someone with kind collisions as we add features ... :)

What if ours were always `backstage-`? What if theirs were requested to always be `x-`? :) Probably not worth it.

packages/frontend-plugin-api/report.api.md
```diff
@@ -1898,6 +2017,38 @@ export type StorageValueSnapshot<TValue extends JsonValue> =
...
+    | ExtensionDataRef_2<string, 'core.routing.path', {}>
+    | ExtensionDataRef_2<
+        RouteRef<AnyRouteRefParams_2>,
+        'core.routing.ref',
+        {
+          optional: true;
+        }
+      >
+    | ExtensionDataRef_2<JSX_2, 'core.reactElement', {}>
+    | ExtensionDataRef_2<string, 'core.title', {}>;
```
freben: Do these not have icons as an output? Not sure exactly what we'd do with it right now but it makes sense to me somehow

packages/frontend-plugin-api/report.api.md
Rugvip: Yeah bit of a tricky one. I think we want core things to be relatively simple but probably also named in a way that reduces risk for conflict with plugins. For other blueprints you'd generally always want to prefix with the plugin name or related resource, like `EntityHeaderAction` or w/e might come up. Still remains to be seen how for the ecosystem takes extensibility and if naming conflicts is gonna turn into a big issue. I'm optimistic enough it can be figured out in other way that there's no need to complicate the system further, including more strict prefixing patterns etc.

packages/frontend-plugin-api/src/icons/types.ts
Rugvip: Yep, it was just that this one used to be a global also I think?

packages/frontend-plugin-api/src/blueprints/HeaderActionBlueprint.tsx
Rugvip: Yep will make this `PluginHeaderActionBlueprint` instead and rename things. That's what the BUI component will be called eventually too.

packages/frontend-plugin-api/src/blueprints/HeaderActionBlueprint.tsx
Rugvip: Yep, switching this one to be a plain element instead and use the extension boundary to load it, like the page blueprint does.

`reactElementLoader` is an option, it just depends on where we want to introduce the extension boundary, i.e. is it in the blueprint or the consuming extension. Both are generally owned by the same plugin so it doesn't really matter much as far as I can tell. Thinking it's fine to stick to what we've been doing.

plugins/app/src/extensions/AppNav.tsx
```diff
@@ -79,9 +141,77 @@ function NavContentRenderer(props: {
...
+        resolvedTitle !== pluginTitle &&
+        resolvedTitle !== pluginId;
+      const title = hasExplicitPageTitle
+        ? resolvedTitle
+        : matchingNavItem?.title ?? pluginTitle ?? pluginId;
+
+      // PageBlueprint resolves icon as: params.icon ?? plugin.icon
+      // We want the priority: page (params) -> nav item -> plugin -> (excluded)
+      const resolvedIcon = node.instance.getData(coreExtensionData.icon);
+      const hasExplicitPageIcon = resolvedIcon && !node.spec.plugin.icon;
```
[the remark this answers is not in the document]
Rugvip: Yep this needed a fix https://github.com/backstage/backstage/pull/32812/commits/f88dabe3c77b4bc3abd2150598ade96ea8eabbc8

packages/frontend-plugin-api/src/blueprints/PageBlueprint.tsx
```diff
@@ -43,17 +73,106 @@ export const PageBlueprint = createExtensionBlueprint({
...
+
+      yield coreExtensionData.reactElement(<PageContent />);
+    } else {
+      const PageContent = () => {
+        const headerActions = usePluginHeaderActions(pluginId);
+        return (
+          <PageLayout title={title} icon={icon} headerActions={headerActions} />
+        );
+      };
+      yield coreExtensionData.reactElement(<PageContent />);
```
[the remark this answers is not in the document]
Rugvip: This is fine

packages/core-components/src/layout/Page/Page.tsx
```diff
@@ -30,12 +30,14 @@ const useStyles = makeStyles(
       gridTemplateColumns: 'auto 1fr auto',
       overflowY: 'auto',
       height: '100vh',
+      maxHeight: '100%',
```
freben: This is always a bit scary.

packages/core-components/src/layout/Page/PageWithHeader.tsx
```diff
@@ -21,13 +21,14 @@ import { Page } from './Page';
...
+  noHeader?: boolean;
 };
 
 export function PageWithHeader(props: PropsWithChildren<PageWithHeaderProps>) {
-  const { themeId, children, ...restProps } = props;
+  const { themeId, children, noHeader, ...restProps } = props;
   return (
     <Page themeId={themeId}>
-      <Header {...restProps} />
+      {noHeader ? <div /> : <Header {...restProps} />}
```
freben: null?

packages/frontend-plugin-api/src/components/PageLayout.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Props for the PageLayout component
+ * @public
+ */
+export interface PageLayoutProps {
+  title?: string;
+  icon?: IconElement;
+  noHeader?: boolean;
+  headerActions?: ReactNode;
```
freben: Slightly worried about this being a naked `ReactNode`. Could something more like at least `Array<{ children: ReactNode }>` let us both iterate more on that in the future, as well as more cleanly laying them out using for example a flex box controlled by the header itself?

packages/frontend-plugin-api/src/components/PageLayout.tsx
```diff
@@ -0,0 +1,142 @@
...
+        >
+          {title && (
+            <div
+              style={{
+                display: 'flex',
+                alignItems: 'center',
+                gap: '8px',
+                padding: '12px 24px 8px',
+                fontSize: '18px',
+                fontWeight: 500,
```
freben, line 77: Is this really all we use for title styling? No css vars, no overridability? The title prop itself is a string and as such can't be tweaked/styled from the outside as well

packages/frontend-plugin-api/src/components/PageLayout.tsx
```diff
@@ -0,0 +1,142 @@
...
+              }}
+            >
+              {icon}
+              {title}
+              {headerActions && (
+                <div style={{ marginLeft: 'auto' }}>{headerActions}</div>
```
freben, line 83: This one can get spicy when having lots of actions or a long title (long complex component names for example). I wonder if there's some fancy css that could be used for the left and right hand sides here, maybe a grid model where the right hand side is limited by minmax or whatever. Can iterate on that of course

packages/frontend-plugin-api/src/components/PageLayout.tsx
```diff
@@ -0,0 +1,142 @@
...
+          {tabs && tabs.length > 0 && (
+            <nav
+              style={{
+                display: 'flex',
+                gap: '4px',
+                padding: '0 24px',
+              }}
+            >
+              {tabs.map(tab => (
+                <a
```
freben, line 96: Oh no tabs component?

plugins/app-visualizer/src/components/CopyTreeButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+      setTimeout(() => setCopied(false), 2000);
+    });
+  };
+
+  return (
+    <Button
+      variant="secondary"
+      size="small"
+      iconStart={copied ? <RiCheckLine /> : <RiFileCopyLine />}
+      onPress={handlePress}
```
freben, line 58: maybe disable while copied

plugins/app/src/extensions/AppNav.tsx
```diff
@@ -20,55 +20,117 @@ import {
...
+  constructor(items: NavContentNavItem[], taken?: Iterable<string>) {
+    this.#items = items;
+    this.#index = new Map(items.map(item => [item.node.spec.id, item]));
+    this.#taken = new Set(taken);
+  }
+
+  take(id: string): NavContentNavItem | undefined {
+    const item = this.#index.get(id);
+    if (item) {
+      this.#taken.add(id);
```
freben, line 54: so it's fine to take twice by definition - ok

plugins/catalog/src/alpha/plugin.tsx
```diff
@@ -15,8 +15,8 @@
  */
 
 import { createFrontendPlugin } from '@backstage/frontend-plugin-api';
-
 import { entityRouteRef } from '@backstage/plugin-catalog-react';
+import CategoryIcon from '@material-ui/icons/Category';
```
freben, line 19: should we already now be picking out of the blessed icon library for bui instead? maybe it doesn't matter much

packages/core-components/src/layout/Page/Page.tsx
Rugvip: Ah good catch, this is no longer needed following the new `FullPage` approach, rolling this back

packages/core-components/src/layout/Page/PageWithHeader.tsx
Rugvip: Rolling this back as well since we're not aiming to handle this this time around

packages/frontend-plugin-api/src/components/PageLayout.tsx
Rugvip: Yep, this is mostly aiming to forward the `PluginHeader` props from BUI: https://github.com/backstage/backstage/blob/0ec3c0e50e9225291f5ba095e8ee0e50f807cf42/packages/ui/src/components/PluginHeader/types.ts#L29

I do think we need to keep this very open, i.e. we can't just have it be data driven and need to support complex interactible elements. I'll make this `Array<JSX.Element>` for now in the page layout.

packages/frontend-plugin-api/src/components/PageLayout.tsx
Rugvip: No, this is the default implementation that is only intended to be rendered in tests, actually I don't really know what it looks like x)

The app plugin brings in the actual plugin header from BUI

plugins/app-visualizer/src/components/CopyTreeButton.tsx
Rugvip: Leaving this as polish, who knows if this starys around :p

plugins/app/src/extensions/AppNav.tsx
Rugvip: yep

plugins/catalog/src/alpha/plugin.tsx
Rugvip: Yep, leaving this as a slighly later problem to solve. It's both possible to override these icons and the remix and MUI ones also match size-wise

packages/frontend-plugin-api/src/blueprints/SubPageBlueprint.tsx
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright 2024 The Backstage Authors
```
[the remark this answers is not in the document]
Rugvip: Almost

Issue #32805, opened 2026-02-12
Regression in `v1.48.0-next.2`: Duplicated Routes Clicking on Tabs in New Frontend System


awanlin:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
duplicate tabs, react router, new frontend system, nfs
```

### 🗃️ Project Area

Core Framework

### 🔗 External Integration

N/A

### 📝 Description & Context

If you create a new Backstage instance with `npx @backstage/create-app@latest --next` then install a few plugins that add tabs to the catalog you'll get duplicate routes in the URL and the tabs won't switch.

This is likely related to https://github.com/backstage/backstage/pull/31818 and perhaps something on the New Frontend System side got missed.

https://github.com/user-attachments/assets/601479b7-3dd6-4141-b4e4-e37d7f213bf2

### 👍 Expected Behavior

Tabs should switch without issue and load the proper content.

### 🥾 Reproduction steps

1. Run `npx @backstage/create-app@latest --next`
2. Then run: `yarn --cwd packages/app add @backstage/plugin-api-docs`
3. Then run: `yarn --cwd packages/app add @backstage/plugin-techdocs`
4. Now run: `yarn start`
5. Next open the example entity
6. Click on the tabs

Notice: the tabs don't switch and the routes are duplicated in the URL

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

awanlin:
Hi @schultzp2020, any chance you can give this one a look? With the release going out next week it would be great to have it fixed before then. Happy to help but right now I'm not sure exactly where the root of the issue is. 🤔

awanlin:
Another bit of detail, I did a quick test of this with the DevTools plugin as it has 3 tabs and I don't see this issue there using the New Frontend System. That might point us in a direction.

awanlin:
But the App Visualizer exhibits this problem and it's a lot weirder! Just clicking on the Detail tab got me this, notice all the trees in the URL:

<img width="1711" height="483" alt="Image" src="https://github.com/user-attachments/assets/4c918d05-f07e-427e-818f-9ce4d9078135" />

[...]

PR #32824, opened 2026-02-12
fix: resolve tab URL duplication with v7_relativeSplatPath

schultzp2020:

fixes https://github.com/backstage/backstage/issues/32805

With `v7_relativeSplatPath` enabled in the New Frontend System, clicking tabs in catalog entity pages, DevTools, Settings, and the App Visualizer caused URL segments to accumulate on each click (e.g., `/devtools/info/config/tasks` instead of `/devtools/tasks`). This happened because relative links inside splat (`*`) child routes resolve from the full matched URL under v7 behavior, rather than from the parent route path.

The fix touches three components. In `EntityTabs` and `AppVisualizerPage` (both NFS-only), `useRoutes` was replaced with `matchRoutes` for direct content rendering, and tab links now use a `../` prefix when inside a splat child route so they navigate up to the parent before resolving the new tab path. In the shared `RoutedTabs` component from `core-components`, the same approach is used but with an additional compatibility layer: `useRoutes` is retained as the primary content renderer for legacy contexts that depend on the nested route context it provides, falling back to `matchRoutes` when `useRoutes` returns null (which happens in the NFS Outlet structure). The `../` prefix is only applied when `useResolvedPath('.')` confirms that v7 splat behavior is active, keeping legacy apps unaffected.

Existing tests for `RoutedTabs` continue to pass unchanged. New regression tests were added to both `RoutedTabs` and `EntityTabs` covering tab selection, link href generation, and multi-tab click-through navigation under `v7_relativeSplatPath` and `v7_startTransition`.

### DIFF
--- a/.changeset/nice-wasps-grin.md
+++ b/.changeset/nice-wasps-grin.md
@@ -1,0 +1,7 @@
+---
+'@backstage/core-components': patch
+'@backstage/plugin-app-visualizer': patch
+'@backstage/plugin-catalog': patch
+---
+
+Fixed tab navigation URL duplication in `RoutedTabs`, `EntityTabs`, and `AppVisualizerPage` caused by the `v7_relativeSplatPath` future flag. Replaced `useRoutes` with `matchRoutes` for direct rendering and added `../` prefix for relative tab links when inside a splat child route.
--- a/packages/core-components/src/components/TabbedLayout/RoutedTabs.test.tsx
+++ b/packages/core-components/src/components/TabbedLayout/RoutedTabs.test.tsx
@@ -15,9 +15,18 @@
  */
 
 import { renderInTestApp } from '@backstage/test-utils';
-import { act, fireEvent } from '@testing-library/react';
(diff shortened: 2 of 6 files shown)


schultzp2020:
@awanlin @Rugvip this PR should fix https://github.com/backstage/backstage/issues/32805

Rugvip (commented):
I'm worried this fix is too local

How would the change affect something like this plugin? https://github.com/backstage/community-plugins/blob/79ebaef65507b95897f84c83f4779bed4d20cf28/workspaces/report-portal/plugins/report-portal/src/components/Router.tsx#L29

schultzp2020:
> I'm worried this fix is too local
> 
> How would the change affect something like this plugin? https://github.com/backstage/community-plugins/blob/79ebaef65507b95897f84c83f4779bed4d20cf28/workspaces/report-portal/plugins/report-portal/src/components/Router.tsx#L29

I'll see if there is a better solution!

schultzp2020:
Closing for now since we are pivoting to a different direction.

Issue #32311, opened 2026-01-13
Bun for package manager and CLI packaging


tplass-ias:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
bun, yarn, package manager, npm
```

### 🗃️ Project Area

CLI Tooling

### 🔖 Need

Our org is starting to normalize on bun rather than yarn for JS projects, and it would be ideal to follow-suite so we can dogfood this tool ourselves.

### 📝 Proposal

very similar to https://github.com/backstage/backstage/issues/20658

Bun claims to have faster package resolution than yarn, which should be helpful for first builds in monorepo's created by backstage-cli w/ many plugins

Bun can also package a standalone executable, which would be great for caching techdocs-cli into a machine image with pinned dependencies (not possible with `npm i -g`, which opened up our CI based docs publishing to sha1-hulud 😅)

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

No, I don't have time to work on this right now

GabDug:
Although switching/supporting bun can be a large endeavor, I've previously tested the `techdocs-cli` with Bun and it seemed to work just fine (and we got a slimmer image). Might be a good candidate for some POCs.

jackmtpt:
Bun - or really anything else other than `npm install` - is really needed for the techdocs-cli. It takes an embarassing amount of time to install on a fresh runner:

`added 1010 packages in 4m`

Even for javascript that's just pathetically slow.

tplass-ias:
>  embarassing amount of time to install on a fresh runner

this is 100% correct. Often times our users whole build+deploy pipeline can complete before the techdocs-cli install completes. It erodes trust in our platform and hinders adoption of Backstage TechDocs across our org

We're going to try using `deno compile --node-modules-dir=auto -A npm:@techdocs/cli@1.10.4` to package+cache techdocs-cli in our CI runner container image for the time being 🤞 hope this idea helps some other backstage instance maintainers!

sgleske-ias:
Adding on to Tyler's comment:

[...]

PR #32836, opened 2026-02-13
Vendor openBrowser from react-dev-utils

GabDug:
### Context

`react-dev-utils` is a [deprecated](https://github.com/facebook/create-react-app?tab=readme-ov-file#create-react-app--) package that hasn't had an official release in [4 years](https://www.npmjs.com/package/react-dev-utils), and the beta release we're currently using is more than [a year old](https://www.npmjs.com/package/react-dev-utils/v/12.1.0-next.26).

`react-dev-utils` has [24 direct dependencies, and many transitive dependencies](https://npmgraph.js.org/?q=react-dev-utils%4012.0.0-next.60#color=outdated&zoom=h). Many direct/transitive dependencies are outdated.

The techdocs-cli depends on it only to open the browser, which should not need to bring `fork-ts-checker-webpack-plugin` and many other dependencies on our end-users machines.

[Some users are also complaining about the install size of `techdocs-cli`](https://github.com/backstage/backstage/issues/32311#issuecomment-3861212788), this should help a little (but the bulk of it is heavy Cloud SDKs for publishing, a way bigger refactor).

### Proposal

Vendor it in `cli-common`, use in in Techdocs & Backstage CLI.

Upstream is licensed under MIT, so we should be good :) 

#### Switch to `open-browsers`

I considered switching to [open-browsers](https://www.npmjs.com/package/open-browsers) package, which already redistribute an updated version of the utility.

However, the package has very few weekly downloads and its organization is [pretty much inactive](https://github.com/react-doc), so I suggest we vendor it instead, to avoid any supply chain risks.

> Note: In my org's custom `techdocs-cli`, we are currently using a pinned version of `open-browsers`, but I do think it's the best/safest way forward to vendor it.

#### Drop it altogether

Compared to raw `open`, the utility has features to re-use Chrome tabs on MacOS. This is a nice QoL for techdocs-cli and Backstage CLI, which imo is worth the ~300 LoC to maintain.

### Maintenance burden

Seeing how [little changes were needed upstream](https://github.com/facebook/create-react-app/commits/main/packages/react-dev-utils/openBrowser.js), and that this utility has been working fine for years, we can only hope we won't have much issues.

### 🤔 But Gabriel we still have `react-dev-utils` in `@backstage/cli`

For now, but we could consider vendoring the 3 usages left, or replacing them by alternatives. Some may not be needed after Webpack support removal. 

## Clanker notice

The utility has been ported in TS with AI for this POC, and only lightly reviewed & tested manually.

If we're okay to vendor-it, I'll go more thoroughly. We might also prefer to keep the AppleScript as a file instead of inlined. We may also keep a plain JS version closer to the original. Lmk!

### DIFF
--- a/.changeset/deep-pigs-drive.md
+++ b/.changeset/deep-pigs-drive.md
@@ -2,4 +2,4 @@
 '@backstage/cli-common': patch
 ---
 
-Added `openBrowser`, which opens a URL in the user's browser and reuses an existing Chromium tab on macOS where possible. Install the optional `open` peer dependency to use it; without it `openBrowser` returns `false`.
+Added `openBrowser`, which opens a URL in the user's browser and reuses an existing Chromium tab on macOS where possible. Only the default-browser fallback requires the optional `open` peer dependency; `openBrowser` warns and returns `false` when that path is reached without it.
--- a/.changeset/deep-pigs-drive.md
+++ b/.changeset/deep-pigs-drive.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli-common': patch
+---
+
+Added `openBrowser`, which opens a URL in the user's browser and reuses an existing Chromium tab on macOS where possible. Install the optional `open` peer dependency to use it; without it `openBrowser` returns `false`.
--- a/packages/cli-common/src/vendored/react-dev-utils/openBrowser.ts
+++ b/packages/cli-common/src/vendored/react-dev-utils/openBrowser.ts
@@ -42,11 +42,12 @@
  * - Replaced chalk with Node built-in styleText (node:util)
  * - Inlined AppleScript: openChrome.applescript is embedded as a string and piped to `osascript -` on stdin; no separate .applescript file. The package build only emits JavaScript into dist, so a sibling script file would be missing from the published package.
  * - osascript is invoked via execFileSync with an argument array rather than a shell string. The URL is therefore passed verbatim instead of through encodeURI, which upstream needed for shell quoting but which double-encodes already-escaped query parameters.
- * - Process check on macOS: original uses "ps cax | grep <browser>"; we use "pgrep -x <browser>", which matches the command name exactly rather than by substring, and without spawning a shell.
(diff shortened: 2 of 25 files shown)


packages/cli/src/types.d.ts
awanlin, line 1: I waffled on this but for the sake of clarity can we add a changeset for this change, please.

packages/cli-common/src/vendored/react-dev-utils/openBrowser.ts
```diff
@@ -0,0 +1,314 @@
...
+      'Vivaldi',
+      'Chromium',
+    ];
+
+    for (const chromiumBrowser of supportedChromiumBrowsers) {
+      try {
+        // Try our best to reuse existing tab
+        // on OSX Chromium-based browser with AppleScript.
+        // Original used: ps cax | grep "<browser>".
+        execFileSync('pgrep', ['-x', chromiumBrowser], { stdio: 'ignore' });
```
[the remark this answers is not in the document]
awanlin: Agree with this, can we look into it, please?

awanlin (changes requested):
Sorry this never got looked at when it was first raised @GabDug, gave it a look over now and there's two comments I've left. I'd also like to get eyes from at least one other maintainer.

packages/cli/src/types.d.ts
freben: Agreed, this should get one

.changeset/deep-pigs-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli-common': patch
+---
+
+Added `openBrowser`, which opens a URL in the user's browser and reuses an existing Chromium tab on macOS where possible. Install the optional `open` peer dependency to use it; without it `openBrowser` returns `false`.
```
freben: `BROWSER=*.js` and the macOS AppleScript path can succeed without that dependency. Maybe can word it as, only the fallback/default-browser path requires `open`.

freben:
Looks ok overall like a good approach though

awanlin (approved):
Thanks for the follow up. Approving but would like approval from @freben as well before we merge. 👍

PR #32844, opened 2026-02-13
docs: use postgres:17-bookworm in database getting-started and remove version key from Compose file

LinusAndersson02:

I updated the database getting-started tutorial for docker compose.

- I changed Postgres image from 17.0-trixie to 17-bookworm since 17.0-trixie is not a published tag and bookworm should be more stable.
- I also removed the top-level "version" key from the compose YAML file since it is obsolete/ignored since Docker Compose V2 that the tutorial assumes; "docker compose" instead of "docker-compose"

#### :heavy_check_mark: 

### DIFF
--- a/docs/getting-started/config/database.md
+++ b/docs/getting-started/config/database.md
@@ -211,13 +211,13 @@ You can run Postgres in a Docker container, this is great for local development 
 First we need to pull down the container image, we'll use Postgres 17, check out the [Postgres Version Policy](../../overview/versioning-policy.md#postgresql-releases) to learn which versions are supported.
 
 ```shell
-docker pull postgres:17.0-trixie
+docker pull postgres:17-bookworm
 ```
 
 Then we just need to start up the container.
 
 ```shell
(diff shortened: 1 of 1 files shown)


awanlin:
Hi @LinusAndersson02, could we have not used `postgres:18-trixie` instead? We support Postgres 18 and use Trixie everywhere else, that would have been a better option. I only saw this after the fact.

LinusAndersson02:
Hi @awanlin ,If backstage supports it then yes, it was me who got confused when quickly reading the postgres version policy.
Is there a reason why we use trixie instead of bookworm?

I can make a new pull request where i update the version policy example and compose tutorial

awanlin:
Yes, it does, the version policy says we support the last 5 versions which is 18-14 and links to the docs if you are unsure. Debian Trixie is the current [stable version](https://www.debian.org/releases/), we don't have a policy on this but it seems reasonable to be on the latest stable.

awanlin:
If you can raise a PR that would be awesome, and you're right we should update the example. I was leaving it as is when I update this back in `1.45.0` but it's more likely people see it and don't realize newer versions are supported. 👍

LinusAndersson02:
Will raise a PR tomorrow! 👍

PR #32854, opened 2026-02-15
catalog: move stitch queue concerns into `final_entities`

freben:
Related to https://github.com/backstage/backstage/issues/32856

Move all stitch queue concerns into the `final_entities` table, which means removing `next_stitch_at` and `next_stitch_ticket` from `refresh_state`.

### DIFF
--- a/plugins/catalog-backend/migrations/20260215000000_move_stitch_queue.js
+++ b/plugins/catalog-backend/migrations/20260215000000_move_stitch_queue.js
@@ -1,0 +1,189 @@
+/*
+ * Copyright 2026 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// @ts-check
+
+/**
+ * Moves the stitch queue columns from refresh_state to final_entities.
+ * The stitch queue semantically belongs to final_entities since stitching
+ * operates on a per-entity-ref basis and produces one output entity.
+ *
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function up(knex) {
+  const isSQLite = knex.client.config.client.includes('sqlite');
+
+  // Step 1: Add next_stitch_at column to final_entities
+  await knex.schema.alterTable('final_entities', table => {
+    table
+      .dateTime('next_stitch_at')
+      .nullable()
+      .comment('Timestamp of when entity should be stitched');
+  });
+
+  // Step 1b: Create partial index separately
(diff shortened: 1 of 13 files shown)


plugins/catalog-backend/migrations/20260215000000_move_stitch_queue.js
```diff
@@ -0,0 +1,189 @@
...
+          entity_id: row.entity_id,
+          entity_ref: row.entity_ref,
+          hash: '', // Empty hash indicates needs stitching
+          stitch_ticket: row.next_stitch_ticket || '',
+          final_entity: null,
+          last_updated_at: null,
+          next_stitch_at: row.next_stitch_at,
+        })),
+      )
+      .onConflict('entity_ref')
```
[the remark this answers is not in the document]
freben: This is actually intentional.

The longer term table shape will no longer have an `entity_id` at all - or at most it'll be "advisory" if we still want to refer back to some "main anchor" entity that the current shape derives from. But even then, it likely won't have a foreign key constraint on it. The primary column will move over to be `entity_ref`, and deletions will start to happen via tombstones - not row deletion.

plugins/catalog-backend/migrations/20260215000000_move_stitch_queue.js
```diff
@@ -0,0 +1,189 @@
...
+ *
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function up(knex) {
+  const isSQLite = knex.client.config.client.includes('sqlite');
+
+  // Step 1: Add next_stitch_at column to final_entities
+  await knex.schema.alterTable('final_entities', table => {
+    table
+      .dateTime('next_stitch_at')
```
Rugvip, line 32: Potentially default to a select from `refresh_state` to do it all in one go, and then remove the default after?

plugins/catalog-backend/src/database/operations/stitcher/markForStitching.ts
```diff
@@ -104,30 +104,56 @@ export async function markForStitching(options: {
...
+              refreshStateRows.map(row => ({
+                entity_id: row.entity_id,
+                entity_ref: row.entity_ref,
+                hash: '',
+                stitch_ticket: ticket,
+                next_stitch_at: knex.fn.now(),
+              })),
+            )
+            .onConflict('entity_id')
+            .merge(['next_stitch_at', 'stitch_ticket']);
```
Rugvip, line 105: wait, what if a row is removed from `refresh_state` in-between?

plugins/catalog-backend/migrations/20260215000000_move_stitch_queue.js
freben: Hm not sure what you mean - this column stays nullable.

plugins/catalog-backend/src/database/operations/stitcher/markForStitching.ts
freben: Yeah you are right, there's a window between the read from `refresh_state` and write to `final_entities` where a foreign key constraint error can be made possible.

This ties into what I meant in the meta issue - I actually want the `final_entities.entity_id` column to go away or at most be advisory. When there are several candidates for the final entity, a 1:1 relation may no longer make sense. I was considering making that change later down the line rather than as part of this PR - but now I may have to reconsider what I should do with this.

freben:
Do you think that it might be worth just making a dedicated `stitch_queue` table instead of having any of this in `final_entities`? I honestly don't mind that idea - it would be kind of nice.

PR #32874, opened 2026-02-17
`feat(catalog)`: support entity predicate filtering

benjdlambert:
This is #32282 with a few bonus commits on top and a rebase.

There is more to be done for sure - implement (at least some subset of) `$contains`, remove the cursor parsing, remove some duplication - but the PR is already large enough and the risk just keeps stacking up. So let's not do more right now.

### DIFF
--- a/plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
+++ b/plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
@@ -175,18 +175,18 @@ function applyFieldCondition(options: {
       //
       // FROM: `{ "a": { "$contains": "b" } }`
       //
-      // TO:   `{ "a.b": "true" }`
+      // TO:   `{ "a": "b" }`
       //
       // The search table does not actually show us that "a" was an array to
       // begin with, so this can mistakenly also match on an object that had a
-      // "b" key with a true value. We'll consider that an acceptable tradeoff
-      // though.
+      // "b" key with a primitive value. We'll consider that an acceptable
+      // tradeoff though.
       if (isPrimitive(target)) {
         const matchQuery = knex<DbSearchRow>('search')
           .select('search.entity_id')
           .where({
-            key: `${key}.${String(target).toLocaleLowerCase('en-US')}`,
-            value: 'true',
+            key,
+            value: String(target).toLocaleLowerCase('en-US'),
           });
         return targetQuery.andWhere(onEntityIdField, 'in', matchQuery);
       }


Nilay1999:
Hey @benjdlambert – thanks for taking this forward with the extra review commits! I have a few more changes I'd like to add. So should i rebase my branch (Nilay1999:feat/advanced-catalog-search) with yours ?

benjdlambert:
Hey @Nilay1999 :wave:

It's entirely up to you. This branch is basically all of your PR with some of @freben's feedback addressed and a bit of cleanup. I didn't wanna just push those commits over the top of your work in case I was heading in the wrong direction.

What were you thinking of doing? Wondering if we should get a review from @freben first and see where we wanna head?

Nilay1999:
Sure @benjdlambert 👍🏻 
Actually i was facing some issues with the openapi models which i generated earlier, and i can see you have removed them, simplified the api schema which solves my compilation issues.
So for now i'm just rebasing my branch to pickup your changes in my local and lets wait for @freben's review.

plugins/catalog-backend/src/schema/openapi.yaml
```diff
@@ -1095,6 +1095,55 @@ paths:
               type: string
           explode: false
           style: form
+    post:
+      operationId: QueryEntitiesByPredicate
```
freben, line 1099: Could also be `GetEntitiesByQueryPredicate` or something like that - for consistency with the above, keeping their type names close.

plugins/catalog-backend/src/schema/openapi.yaml
```diff
@@ -1095,6 +1095,55 @@ paths:
...
+        - {}
+        - JWT: []
+      requestBody:
+        required: false
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                cursor:
```
freben, line 1124: if we wanted to be VERY precise we could split it into two options (initial and cursor) - but honestly I don't mind keeping it flat

plugins/catalog-backend/src/schema/openapi.yaml
```diff
@@ -1095,6 +1095,55 @@ paths:
...
+                  type: string
+                limit:
+                  type: number
+                orderField:
```
freben: Tempting to make this one properly typed now that we're in JSON land already - [the client side already has](https://github.com/backstage/backstage/blob/b1919bdf2a0f4cc587a4adff91d03db229227056/packages/catalog-client/src/types/api.ts#L166) like `order: { field: 'metadata.name', order: 'asc' }` (alternatively an array of those) which then unnecessarily gets encoded into a string and then parsed again in the service.

packages/catalog-client/src/CatalogClient.ts
```diff
@@ -266,6 +271,30 @@ export class CatalogClient implements CatalogApi {
...
+
+    // Route to POST endpoint if query predicate is provided (initial request)
+    if (isQueryEntitiesInitialRequest(request) && request.query) {
+      return this.queryEntitiesByPredicate(request, options);
+    }
+
+    // Route to POST endpoint if cursor contains a query predicate (pagination)
+    if (
+      !isQueryEntitiesInitialRequest(request) &&
+      cursorContainsQuery(request.cursor)
```
freben: It really is a bit unfortunate to have to introspect the cursor. 🤔

We could have something like a prefix on the actual raw cursor that is possible to look at without all of the decoding.

Or make the GET endpoint support the same cursor format and make the client pick between GET and POST _only_ based on whether the token is above a certain number of kilobytes and risks busting URL length limits.

But OK, let's leave it like this for now. We can improve on it separately.

packages/catalog-client/src/CatalogClient.ts
```diff
@@ -320,6 +349,59 @@ export class CatalogClient implements CatalogApi {
...
 
+  /**
+   * Query entities using predicate-based filters (POST endpoint).
+   * @internal
+   */
+  private async queryEntitiesByPredicate(
+    request: QueryEntitiesRequest,
+    options?: CatalogRequestOptions,
+  ): Promise<QueryEntitiesResponse> {
+    const body: Record<string, unknown> = {};
```
freben:
```suggestion
    const body: QueryEntitiesByPredicateRequest = {};
```

I think maybe that makes the code below actually read a bit better since you deal with types at each one explicitly.

packages/catalog-client/src/CatalogClient.ts
```diff
@@ -320,6 +349,59 @@ export class CatalogClient implements CatalogApi {
...
+      }
+      if (fields?.length) {
+        body.fields = fields;
+      }
+    } else {
+      body.cursor = request.cursor;
+      if (request.limit !== undefined) {
+        body.limit = request.limit;
+      }
+      if (request.fields?.length) {
```
freben, line 408: Oh are we still leaving `fields` outside the cursor for this one? I was hoping the cursor would be self descriptive. And it's a bit arbitrary that it's outside either way.

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,245 @@
...
+  );
+}
+
+function isExistsValue(
+  value: FilterPredicateValue,
+): value is { $exists: boolean } {
+  return typeof value === 'object' && value !== null && '$exists' in value;
+}
+
+function isInValue(
```
freben: I'm REALLY hungry for `$hasPrefix` to be implemented, can we do that right away too?

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,245 @@
...
+      onEntityIdField,
+      knex,
+      !negate,
+    );
+  }
+
+  // Handle $all (AND)
+  if (isAllPredicate(filter)) {
+    return targetQuery[negate ? 'andWhereNot' : 'andWhere'](
+      function allFilter() {
```
freben: matter of taste I guess, but I find that arrow functions end up easier to grasp. ESPECIALLY when nesting of course.

```ts
return targetQuery[negate ? 'andWhereNot' : 'andWhere'](inner => {
  for (...) {
    inner.andWhere(...
```

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,245 @@
...
+      targetQuery,
+      onEntityIdField,
+      knex,
+      !negate,
+    );
+  }
+
+  // Handle $all (AND)
+  if (isAllPredicate(filter)) {
+    return targetQuery[negate ? 'andWhereNot' : 'andWhere'](
```
freben: In `locations/by-query` claude/copilot pointed out that we should look for empty arrays for `$all` and `$any`, to avoid malformed SQL being formed. Just returning something that can never match like `1 = 0`

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,245 @@
...
+  negate: boolean,
+): Knex.QueryBuilder {
+  // Handle $not
+  if (isNotPredicate(filter)) {
+    return applyPredicateInStrategy(
+      filter.$not,
+      targetQuery,
+      onEntityIdField,
+      knex,
+      !negate,
```
freben: I don't think we need the negate. It makes the logic harder to follow. You could implement `$not` as

```ts
return targetQuery.whereNot(inner => applyPredicateInStrategy(filter.$not, ...));
```

I know the other strategy uses a bool like this. But I don't recall if I had any good reason to do that. Like is `NOT(a IN b)` really worse than `a NOT IN b`? I would wager no.

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,245 @@
...
+                value,
+              )}`,
+            );
+          }
+        }
+      },
+    );
+  }
+
+  return targetQuery;
```
freben: Do we really just return here, or is it an error condition (found an expression we could not handle)?

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,245 @@
...
+        )} }`,
+    );
+  }
+
+  // Handle field expressions like { "kind": "component" } or { "spec.type": { "$in": ["service", "website"] } }
+  if (isFieldExpression(filter)) {
+    return targetQuery[negate ? 'andWhereNot' : 'andWhere'](
+      function fieldFilter() {
+        for (const [key, value] of Object.entries(filter)) {
+          const normalizedKey = key.toLowerCase();
```
freben: The search builder carefully uses `.toLocaleLowerCase('en-US')`. Same everywhere below too

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,245 @@
...
+          filter,
+        )} }`,
+    );
+  }
+
+  // Handle field expressions like { "kind": "component" } or { "spec.type": { "$in": ["service", "website"] } }
+  if (isFieldExpression(filter)) {
+    return targetQuery[negate ? 'andWhereNot' : 'andWhere'](
+      function fieldFilter() {
+        for (const [key, value] of Object.entries(filter)) {
```
freben: Should we check for empty keyset here too, and handle with a never-matcher?

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,245 @@
...
+            }
+          } else if (isInValue(value)) {
+            // Handle $in
+            const values = value.$in.map(v => String(v).toLowerCase());
+            const matchQuery = knex<DbSearchRow>('search')
+              .select('search.entity_id')
+              .where({ key: normalizedKey })
+              .whereIn('value', values);
+            this.andWhere(onEntityIdField, 'in', matchQuery);
+          } else if (isPrimitive(value)) {
```
freben: Again, `$hasPrefix` would be awesome!

plugins/catalog-backend/src/schema/openapi.yaml
freben: Backtracking on this, was fine as-is

plugins/catalog-backend/src/schema/openapi.yaml
freben: fixed

plugins/catalog-backend/src/schema/openapi.yaml
freben: skipping for now

packages/catalog-client/src/CatalogClient.ts
freben: since we use a lot of the same internals as before, this makes sense to keep until we do some future refactor of it all

packages/catalog-client/src/CatalogClient.ts
freben: keeping, but with a TODO comment

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
freben: done

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
freben: done

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
freben: done

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
freben: removed and flattened

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
Nilay1999: hmm yeah, this should work i guess

```typscript
const entries = Object.entries(filter);
if (entries.length === 0) {
   if (negate) {
      return targetQuery.andWhere(knex.raw('1 = 0'));
   }
   return targetQuery;
}
```

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
Nilay1999: yeah, actually i did add those condition but its in a separate branch 😅

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
freben: done

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
freben: done

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
freben: inverted the checks and made sure to bail properly

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
freben: made it so that an empty object positively matches everything

plugins/catalog-backend/src/service/createRouter.test.ts
```diff
@@ -1120,6 +1177,7 @@ describe('createRouter readonly and raw json enabled', () => {
   beforeAll(async () => {
     entitiesCatalog = {
       entities: jest.fn(),
+
```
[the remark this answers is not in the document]
freben: removed

plugins/catalog-backend/src/service/request/parseEntityQuery.ts
```diff
@@ -0,0 +1,114 @@
...
+    orderFields,
+    fullTextFilter: request.fullTextFilter
+      ? {
+          term: request.fullTextFilter.term ?? '',
+          fields: request.fullTextFilter.fields,
+        }
+      : undefined,
+    fields: request.fields,
+    limit: request.limit,
+  };
```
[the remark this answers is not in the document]
freben: added support for this; it's important that the post endpoint has feature parity with the get endpoint since the client chooses dynamically between them internally

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,266 @@
...
+              .whereIn('value', values);
+            this.andWhere(onEntityIdField, 'in', matchQuery);
+          } else if (isHasPrefixValue(value)) {
+            // Handle $hasPrefix
+            const prefix = value.$hasPrefix.toLocaleLowerCase('en-US');
+            const escaped = prefix.replace(/[%_\\]/g, c => `\\${c}`);
+            const matchQuery = knex<DbSearchRow>('search')
+              .select('search.entity_id')
+              .where({ key: normalizedKey })
+              .andWhereRaw('?? like ? escape ?', [
```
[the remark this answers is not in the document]
freben: added hasPrefix, left out contains

plugins/catalog-backend/src/schema/openapi.yaml
```diff
@@ -1095,6 +1095,66 @@ paths:
...
+                    term:
+                      type: string
+                    fields:
+                      type: array
+                      items:
+                        type: string
+                fields:
+                  type: array
+                  items:
+                    type: string
```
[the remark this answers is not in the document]
freben: i don't think this comment is right

packages/catalog-client/src/CatalogClient.ts
```diff
@@ -320,6 +348,54 @@ export class CatalogClient implements CatalogApi {
...
+    const body: QueryEntitiesByPredicateRequest = {};
+
+    if (isQueryEntitiesInitialRequest(request)) {
+      const { query, limit, orderFields, fullTextFilter, fields } = request;
+      if (query && typeof query === 'object') {
+        body.query = query;
+      }
+      if (limit !== undefined) {
+        body.limit = limit;
+      }
```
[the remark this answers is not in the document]
freben: offset should now be passed through

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    it('filters with $all', async () => {
+      await expect(
+        query({
+          $all: [{ kind: 'component' }, { 'spec.type': 'service' }],
+        }),
+      ).resolves.toEqual(['service-a', 'service-b']);
+    });
+
+    it('returns nothing for empty $all', async () => {
+      await expect(query({ $all: [] })).resolves.toEqual([]);
```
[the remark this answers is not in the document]
freben: fixed

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,219 @@
...
+        value: String(value).toLocaleLowerCase('en-US'),
+      });
+    return targetQuery.andWhere(onEntityIdField, 'in', matchQuery);
+  }
+
+  if (isObject(value)) {
+    if ('$exists' in value) {
+      const existsQuery = knex<DbSearchRow>('search')
+        .select('search.entity_id')
+        .where({ key });
```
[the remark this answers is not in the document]
freben: added explicit condition

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,219 @@
...
+        targetQuery: inner,
+        onEntityIdField,
+        knex,
+      }),
+    );
+  }
+
+  if ('$all' in filter) {
+    if (filter.$all.length === 0) {
+      return targetQuery.andWhereRaw('1 = 0');
```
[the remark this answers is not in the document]
freben: done

packages/catalog-client/src/CatalogClient.ts
```diff
@@ -320,6 +348,54 @@ export class CatalogClient implements CatalogApi {
...
+   */
+  private async queryEntitiesByPredicate(
+    request: QueryEntitiesRequest,
+    options?: CatalogRequestOptions,
+  ): Promise<QueryEntitiesResponse> {
+    const body: QueryEntitiesByPredicateRequest = {};
+
+    if (isQueryEntitiesInitialRequest(request)) {
+      const { query, limit, orderFields, fullTextFilter, fields } = request;
+      if (query && typeof query === 'object') {
```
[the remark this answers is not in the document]
freben: it's ok, the parameter was never in use

plugins/catalog-backend/src/service/request/parseEntityQuery.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (!request.cursor) {
+      throw new InputError('Cursor cannot be empty');
+    }
+
+    const cursor = decodeCursor(request.cursor);
+    return {
+      cursor,
+      fields: request.fields,
+      limit: request.limit,
+    };
```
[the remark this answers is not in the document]
freben: handled by openapi

plugins/catalog-backend/src/service/request/parseEntityQuery.ts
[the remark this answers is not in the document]
freben: handled by openapi

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.test.ts
```diff
@@ -0,0 +1,262 @@
...
+      await expect(
+        query({
+          $all: [{ kind: 'component' }, { 'spec.type': 'service' }],
+        }),
+      ).resolves.toEqual(['service-a', 'service-b']);
+    });
+
+    it('matches everything for empty $all', async () => {
+      await expect(query({ $all: [] })).resolves.toEqual([
+        'api-d',
```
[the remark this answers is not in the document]
freben: already fixed

packages/catalog-client/src/utils.ts
```diff
@@ -26,6 +32,58 @@ export function isQueryEntitiesInitialRequest(
...
+      } else if (strings.length === 1) {
+        parts.push({ [key]: strings[0] } as FilterPredicateExpression);
+      } else if (strings.length > 1) {
+        parts.push({ [key]: { $in: strings } } as FilterPredicateExpression);
+      }
+    }
+
+    return parts.length === 1 ? parts[0] : { $all: parts };
+  });
+
```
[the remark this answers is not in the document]
freben: the type system is good enough of a safeguard here i think

plugins/catalog-backend/src/service/request/applyPredicateEntityFilterToQuery.ts
```diff
@@ -0,0 +1,329 @@
...
+            knex,
+          }),
+        );
+      }
+    });
+  }
+
+  // Treat the filter as a field expression like { "kind": "component" } or { "spec.type": { "$in": ["service", "website"] } }
+  if (Object.keys(filter).length === 0) {
+    return targetQuery.andWhereRaw('1 = 1');
```
Rugvip, line 103: Wait why not just `return targetQuery`? 😅

freben: Because of the recursive nature of the query building, we make sure that we don't end up with empty nodes (like `AND ()`)

PR #32875, opened 2026-02-17
ui: rename Header to PluginHeader

Rugvip:

Had a discussion with @cdedreuille about this one, the overall plan is to rename the current `Header` to `PluginHead` and the current `HeaderPage` to just `Header`. We can't do that all at once since it'd be confusing, so instead we start with an immediate breaking rename of `Header` to `PluginHeader`. This is a component that didn't really have any existing use in Backstage apps just yet, so the impact of the immediate breaking change should be minimal. It is being introduced through #32812 alongside this change though.

This renames the `Header` component in `@backstage/ui` to `PluginHeader`. The following public exports have been renamed:

- `Header` → `PluginHeader`
- `HeaderProps` → `PluginHeaderProps`
- `HeaderDefinition` → `PluginHeaderDefinition`

CSS class names updated from `bui-Header*` to `bui-PluginHeader*` to match. The `HeaderTab` type is unchanged since it's shared with `HeaderPage`.

The docs-ui site has been updated as well — the component page moved from `/components/header` to `/components/plugin-header`, and all snippets, examples, and changelog references have been updated.


Rugvip:
@emmaindal yep the intention is for this header to be at the scope of the plugin rather than individual pages. In general we want to reduce the conflict with that's currently the `HeaderPage` and promote that one as much as possible for use on plugin pages. I think if you use the `PageBlueprint` you'd be more inclined to reach for the `PageHeader`, but that's not what we want.

PR #32890, opened 2026-02-17
feat: Add status check functions for scaffolder steps

Ferin79:

- Introduced `always()` and `failure()` functions to control step execution after failures.
- Updated documentation to explain usage of new status check functions.
- Updated NunjucksWorkflowRunner to process these functions in step conditions.
- Added tests to verify behavior of steps using `always()` and `failure()`.

Screenshot 2026-02-17 at 11 08 17 AM


Ferin79:
Hey @Lucas-Desouza & @jpdeburgo does this PR solves the issue from https://github.com/backstage/backstage/issues/23104 ?

Ferin79:
Tagging @Rugvip & @freben for review

Ferin79:
bump @Rugvip for review

Ferin79:
@benjdlambert can you please review my PR ?

benjdlambert:
Yep will do - it's on the list.

We're in the run up to KubeCon right now so we're currently occupied with that, so bear with us.

Ferin79:
> Yep will do - it's on the list.
> 
> We're in the run up to KubeCon right now so we're currently occupied with that, so bear with us.

Hey @benjdlambert any timeline for this PR?

Ferin79:
Hey @benjdlambert bumping

Ferin79:
Hey @Rugvip can you please review the PR ?

plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
```diff
@@ -543,6 +543,32 @@ export class NunjucksWorkflowRunner implements WorkflowRunner {
     }
   }
 
+  /**
+   * Check if a step's `if` condition uses status check functions (always() or failure())
+   * that should cause the step to be attempted even when a previous step has failed.
+   */
+  private hasStatusCheckFunction(step: TaskStep): boolean {
+    if (typeof step.if !== 'string') return false;
+    return /\b(?:always|failure)\s*\(\s*\)/.test(step.if);
```
benjdlambert: I think that this RegEx is a bit too loose here.

I think a better way forward would be to only accept the form `${{ always() }}` or `${{ failure() }}` rather than also the text form, and that way you can check the rendered result instead.

This also simplifies things a little bit, and just keeps things implemented as `nunjucks` filters instead.

It cleans up a bit of the implementation too.

One thing that we would need to check still is that we only run the step that has `if: ${{ failure() }}` on failure, if the `if` block contained a status check global.

We don't wanna run steps that has `if: ${{ true }}` etc..

benjdlambert (changes requested):
Hey :wave: thanks for this, sorry about the delay in review. Been a very busy few months.

plugins/scaffolder-backend/src/scaffolder/tasks/NunjucksWorkflowRunner.ts
Ferin79: Thanks, that makes sense. I’ve updated the implementation to only support `${{ always() }}` and `${{ failure() }}` in the if field, and removed the looser text-form handling and regex check.

The runner now checks whether the rendered `if` expression actually invoked a status check global before allowing a step to run after a failure. That means steps like `${{ true }}` or other truthy expressions won’t continue execution after a failure unless they explicitly use `always()` or `failure()`.

I also updated the tests and docs to match this behavior.

benjdlambert (approved):
Nice thanks! :tada:

PR #32894, opened 2026-02-17
frontend-plugin-api: stabilize translation API

Rugvip:

This stabilizes all translation API usage across the new frontend system packages. There were a few lingering `@alpha` tags and imports from `@backstage/core-plugin-api/alpha` that should've been using the stable `@backstage/frontend-plugin-api` path instead.

Changes:
- Promoted `BaseOptions` from `@alpha` to `@public` in `@backstage/frontend-plugin-api`
- Switched translation API imports in `@backstage/plugin-app` to use `@backstage/frontend-plugin-api`
- Switched translation API imports and JSDoc links in `@backstage/frontend-test-utils` to use `@backstage/frontend-plugin-api`

### DIFF
--- a/packages/frontend-plugin-api/src/apis/definitions/TranslationApi.ts
+++ b/packages/frontend-plugin-api/src/apis/definitions/TranslationApi.ts
@@ -22,7 +22,7 @@ import { JSX } from 'react';
 /**
  * Base translation options.
  *
- * @alpha
+ * @ignore
  */
 interface BaseOptions {
   interpolation?: {
--- a/.changeset/stable-translation-plugin-app.md
+++ b/.changeset/stable-translation-plugin-app.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-app': patch
+---
+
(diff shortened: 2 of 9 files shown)


packages/frontend-plugin-api/src/apis/definitions/TranslationApi.ts
```diff
@@ -22,7 +22,7 @@ import { JSX } from 'react';
 /**
  * Base translation options.
  *
- * @alpha
+ * @public
```
freben: Hm was this even exported? I don't think so. And it's poorly named for it too :)

Rugvip: it's it's just used in `OptionArgs`, it should have an `@ignore` instead

.changeset/stable-translation-api.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/frontend-plugin-api': patch
+'@backstage/frontend-test-utils': patch
+'@backstage/plugin-app': patch
+---
+
+Stabilized all translation API usage across the new frontend system packages by removing `@alpha` tags and switching imports to stable paths from `@backstage/frontend-plugin-api`.
```
Rugvip: Split into multiple changesets with more information about impact to users

PR #32904, opened 2026-02-18
cli: add translations export and import commands

Rugvip:

This adds `translations export` and `translations import` commands to the Backstage CLI for managing translation files at scale, particularly when integrating with external translation systems.

The `translations export` command discovers all `TranslationRef` definitions across frontend plugin dependencies (including transitive ones) using the TypeScript type system, and exports their default English messages as JSON files along with a manifest. The `translations import` command reads translated JSON files and generates a TypeScript wiring module with `TranslationResource` objects ready to plug into the app.

The commands are app-centric — they're run from within a specific package directory (e.g. `packages/app`) and discover refs from that package's dependency tree. The file layout is fully configurable via a `--pattern` flag (default: `messages/{id}.{lang}.json`).

### DIFF
--- a/docs/tooling/cli/03-commands.md
+++ b/docs/tooling/cli/03-commands.md
@@ -24,7 +24,7 @@ repo [command]                                 Command that run across an entire
 package [command]                              Lifecycle scripts for individual packages
 migrate [command]                              Migration utilities
 versions:bump [options]                        Bump Backstage packages to the latest versions
-translations [command]                          Translation message management
+translations [command]                         Translation message management
 clean                                          Delete cache directories [DEPRECATED]
 build-workspace <workspace-dir> [packages...]  Builds a temporary dist workspace from the provided
                                                 packages
@@ -431,8 +431,8 @@ Usage: backstage-cli create-github-app <github-org>
 
 ## translations export
 
-Export translation messages from all frontend plugin dependencies of the current
-package. This command must be run from within a package directory (e.g.
+Export translation messages from an app and all of its frontend plugins to JSON
+files. This command must be run from within a package directory (e.g.
 `packages/app`), not from the repository root.
 
 The command discovers all `TranslationRef` definitions in the dependency tree,
--- a/packages/cli/src/modules/translations/index.ts
+++ b/packages/cli/src/modules/translations/index.ts
@@ -24,7 +24,7 @@ export default createCliPlugin({
     reg.addCommand({
       path: ['translations', 'export'],
       description:
-        'Export translation messages from all frontend plugins to JSON files',
+        'Export translation messages from an app and all of its frontend plugins to JSON files',
       execute: async ({ args }) => {
         const argv = await yargs()
           .options({


packages/cli/src/modules/translations/lib/messageFilePath.ts
```diff
@@ -0,0 +1,83 @@
...
+  pattern: string,
+): (relativePath: string) => { id: string; lang: string } | undefined {
+  validatePattern(pattern);
+
+  // Build a regex from the pattern by escaping special chars and replacing
+  // {id} and {lang} with named capture groups.
+  const escaped = pattern
+    .replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+    .replace(/\\{id\\}/g, '(?<id>[^/]+)')
+    .replace(/\\{lang\\}/g, '(?<lang>[a-z]{2})');
```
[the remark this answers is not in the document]
Rugvip: Nope, those are reserved for a potential future `{locale}` placeholder

docs/tooling/cli/03-commands.md
```diff
@@ -24,6 +24,7 @@ repo [command]                                 Command that run across an entire
 package [command]                              Lifecycle scripts for individual packages
 migrate [command]                              Migration utilities
 versions:bump [options]                        Bump Backstage packages to the latest versions
+translations [command]                          Translation message management
```
freben:
```suggestion
translations [command]                         Translation message management
```

docs/plugins/internationalization.md
```diff
@@ -256,6 +294,116 @@ In an app you can both override the default messages, as well as register transl
...
+### Using the CLI for full translation workflows
+
+When translating your app to other languages at scale — especially when working with external translation systems — the Backstage CLI provides `translations export` and `translations import` commands that automate the extraction and wiring of translation messages across all your plugin dependencies.
+
+#### Exporting default messages
+
+From your app package directory (e.g. `packages/app`), run:
+
+```bash
+yarn backstage-cli translations export
```
freben, line 308: Kinda curious how we'll handle changes to this structure over time. As plugins add / rename / remove translation keys.

I guess it'll be safe to re-run the export since it only overwrites the defaults. But if you translate english to english (just rewording things or adding company specific nomenclature), is that stored separately from the defaults? So they don't get clobbered

packages/cli/src/modules/translations/index.ts
```diff
@@ -0,0 +1,75 @@
...
+import { lazy } from '../../lib/lazy';
+import { DEFAULT_MESSAGE_PATTERN } from './lib/messageFilePath';
+
+export default createCliPlugin({
+  pluginId: 'translations',
+  init: async reg => {
+    reg.addCommand({
+      path: ['translations', 'export'],
+      description:
+        'Export translation messages from all frontend plugins to JSON files',
```
freben:
```suggestion
        'Export translation messages from an app and all of its frontend plugins to JSON files',
```

packages/cli/src/modules/translations/index.ts
```diff
@@ -0,0 +1,75 @@
...
+            pattern: {
+              type: 'string',
+              default: DEFAULT_MESSAGE_PATTERN,
+              description:
+                'File path pattern for message files, with {id} and {lang} placeholders',
+            },
+          })
+          .help()
+          .parse(args);
+        await lazy(() => import('./commands/export'), 'default')(argv);
```
freben, line 46: I guess I never thought of this before - why is this a lazy, that then immediately gets invoked?

packages/cli/src/modules/translations/index.ts
```diff
@@ -0,0 +1,75 @@
...
+          .options({
+            output: {
+              type: 'string',
+              default: 'translations',
+              description:
+                'Output directory for exported messages and manifest',
+            },
+            pattern: {
+              type: 'string',
+              default: DEFAULT_MESSAGE_PATTERN,
```
freben, line 39: Is it overall an unfortunate pattern (heh) to apply defaults out here, instead of concentrating it inside the module itself?

packages/cli/src/modules/translations/lib/messageFilePath.ts
```diff
@@ -0,0 +1,83 @@
...
+export function createMessagePathParser(
+  pattern: string,
+): (relativePath: string) => { id: string; lang: string } | undefined {
+  validatePattern(pattern);
+
+  // Build a regex from the pattern by escaping special chars and replacing
+  // {id} and {lang} with named capture groups.
+  const escaped = pattern
+    .replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+    .replace(/\\{id\\}/g, '(?<id>[^/]+)')
```
freben, line 43: is `^/` right? The examples have patterns like `messages/{id}.{lang}.json` where there are other separators than slashes

packages/cli/src/modules/translations/lib/discoverPackages.ts
```diff
@@ -0,0 +1,207 @@
...
+    throw new Error(
+      'No package.json found in the current directory. ' +
+        'The translations commands must be run from within a package directory.',
+    );
+  }
+
+  if (resolvePath(packageDir) === resolvePath(repoRoot)) {
+    throw new Error(
+      'The translations commands must be run from within a package directory, ' +
+        'not from the repository root. For example: cd packages/app && backstage-cli translations export',
```
freben, line 55: Can plugins ship with their own builtin translations too, not just defaults?

packages/cli/src/modules/translations/commands/import.ts
```diff
@@ -0,0 +1,214 @@
...
+    ...importLines,
+    '',
+    'export default [',
+    ...resourceLines,
+    '];',
+    '',
+  ].join('\n');
+
+  await fs.ensureDir(resolvePath(outputPath, '..'));
+  await fs.writeFile(outputPath, fileContent, 'utf8');
```
freben, line 179: Should we add the default output dir to the .prettierignore / .eslintignore of the default-app, or run prettier on the output? Maybe it doesn't matter since the user will commit it themselves after, hopefully with precommit hooks.

docs/plugins/internationalization.md
Rugvip: I'm not really expecting engling to english translations to use this systems tbh, for individual overrides I think using the TS API directly is what we should recommend

packages/cli/src/modules/translations/index.ts
Rugvip: The `lazy` ends up being a helper to unwrap the CommonJS translation layer when we use an `import`. The use of `import` is to make sure we don't need to eagerly load the entire implementation for this command when invoking other CLI commands

packages/cli/src/modules/translations/index.ts
Rugvip: Hmm, not sure I have a strong opinion, but having it out here means it can show up in the docs, which is nice

packages/cli/src/modules/translations/lib/discoverPackages.ts
Rugvip: Hmm, true, they can. Leaning towards solving that at the doc/process level rather than for example specifying languages to export for. Overall would leave that as a potential followup

packages/cli/src/modules/translations/commands/import.ts
Rugvip: Yeah would leave this for consumers to figure out for now tbh, maybe providing something opinionated later based on adopter feedback

packages/cli/src/modules/translations/lib/messageFilePath.ts
Rugvip: Yeh, `{id}` needs to match basically everything because they can include anything. Overall the matching is relying on the `{lang}` group being much more strict, so the `{id}` kind of ends up being "the rest"

PR #32912, opened 2026-02-18
feat(scaffolder): add new ui:options autoSelect tag

jsundquist:

This PR adds a new customization to the EntityPicker field to allow teams to disable the `autoSelect` feature of the Autocomplete field.  By providing this field, it will make it so that if a user closes the field with out selecting any entity, the field will keep its original empty value. 

The problem that this is looking to solves is that if a user is looking for their entity, but doesn't select anything and moves on, they don't mistakenly select something they didn't mean to.


benjdlambert (approved):
nice! thanks for this! :pray:

PR #32914, opened 2026-02-18
feat(scaffolder): Create scaffolder mcp action to dry run scaffolder template

yangcao77:

This PR added scaffolder mcp action to dry run a provided scaffolder template
it fetch `/v2/dry-run` endpoint, with input: `templateYaml`, user params `values`, template path`directoryContents`. then return the validation result, including errors and logs. 

tested with cursor mcp server enabled. able to use the mcp tool to dry-run a template. I can see failures and errors when no inputs are given, and with everything set, I also saw a success result.
see screenshot:
failure-no-input
validation-pass

also added tests as part of this PR
tests

### DIFF
--- a/plugins/scaffolder-backend/src/actions/index.ts
+++ b/plugins/scaffolder-backend/src/actions/index.ts
@@ -14,12 +14,12 @@
  * limitations under the License.
  */
 import { ActionsRegistryService } from '@backstage/backend-plugin-api/alpha';
-import { ScaffolderClient } from '@backstage/plugin-scaffolder-common';
+import { ScaffolderService } from '@backstage/plugin-scaffolder-node';
 import { createDryRunTemplateAction } from './createDryRunTemplateAction';
 
 export const createScaffolderActions = (options: {
   actionsRegistry: ActionsRegistryService;
-  scaffolderClient: ScaffolderClient;
+  scaffolderService: ScaffolderService;
 }) => {
   createDryRunTemplateAction(options);
 };
--- a/plugins/scaffolder-backend/src/ScaffolderPlugin.ts
+++ b/plugins/scaffolder-backend/src/ScaffolderPlugin.ts
@@ -21,9 +21,9 @@ import {
 import { ScmIntegrations } from '@backstage/integration';
 import { catalogServiceRef } from '@backstage/plugin-catalog-node';
 import { eventsServiceRef } from '@backstage/plugin-events-node';
-import { ScaffolderClient } from '@backstage/plugin-scaffolder-common';
 import {
   scaffolderActionsExtensionPoint,
+  scaffolderServiceRef,
   TaskBroker,
   TemplateAction,
 } from '@backstage/plugin-scaffolder-node';
@@ -146,11 +146,11 @@ export const scaffolderPlugin = createBackendPlugin({
         httpRouter: coreServices.httpRouter,
         httpAuth: coreServices.httpAuth,
         auditor: coreServices.auditor,
-        discovery: coreServices.discovery,
         catalog: catalogServiceRef,
         events: eventsServiceRef,
         actionsRegistry: actionsServiceRef,
         actionsRegistryService: actionsRegistryServiceRef,
+        scaffolderService: scaffolderServiceRef,
       },
       async init({
         logger,
@@ -161,23 +161,16 @@ export const scaffolderPlugin = createBackendPlugin({
         auth,
         httpRouter,
         httpAuth,
-        discovery,
         catalog,
         permissions,
         events,
         auditor,
         actionsRegistry,
         actionsRegistryService,
+        scaffolderService,
       }) {
         const log = loggerToWinstonLogger(logger);
         const integrations = ScmIntegrations.fromConfig(config);
-
-        // Create ScaffolderClient for MCP actions (isomorphic client)
-        const scaffolderClient = new ScaffolderClient({
-          discoveryApi: discovery,
-          fetchApi: { fetch },
-          scmIntegrationsApi: integrations,
-        });
 
         const templateExtensions = {
           additionalTemplateFilters: convertFiltersToRecord(
@@ -227,10 +220,9 @@ export const scaffolderPlugin = createBackendPlugin({
           `Starting scaffolder with the following actions enabled ${actionIds}`,
         );
 
-        // Register MCP actions for scaffolder
         createScaffolderActions({
           actionsRegistry: actionsRegistryService,
-          scaffolderClient,
+          scaffolderService,
         });
 
         const router = await createRouter({
--- a/plugins/scaffolder-backend/src/actions/createDryRunTemplateAction.test.ts
+++ b/plugins/scaffolder-backend/src/actions/createDryRunTemplateAction.test.ts
@@ -15,9 +15,9 @@
  */
 import { createDryRunTemplateAction } from './createDryRunTemplateAction';
 import { actionsRegistryServiceMock } from '@backstage/backend-test-utils/alpha';
-import { ResponseError } from '@backstage/errors';
+import { scaffolderServiceMock } from '@backstage/plugin-scaffolder-node/testUtils';
 
-type ValidateScaffolderOutput = {
+type DryRunTemplateOutput = {
   valid: boolean;
   message: string;
   errors?: string[];
@@ -55,17 +55,13 @@ spec:
 `;
 
 describe('createDryRunTemplateAction', () => {
-  let mockScaffolderClient: {
-    dryRun: jest.Mock;
-  };
+  const mockScaffolderService = scaffolderServiceMock.mock();
 
   beforeEach(() => {
-    mockScaffolderClient = {
-      dryRun: jest.fn(),
-    };
+    jest.resetAllMocks();
   });
 
-  it('should validate a valid template and return success when dry-run succeeds', async () => {
+  it('should return success with logs when dry-run succeeds', async () => {
     const mockActionsRegistry = actionsRegistryServiceMock();
     const dryRunResult = {
       log: [
@@ -73,23 +69,24 @@ describe('createDryRunTemplateAction', () => {
           body: {
             message: 'Step completed',
             stepId: 'step-1',
-            status: 'completed',
+            status: 'completed' as const,
           },
         },
       ],
       output: { result: 'ok' },
       steps: [{ id: 'step-1', name: 'Step One', action: 'debug:log' }],
+      directoryContents: [],
     };
 
-    mockScaffolderClient.dryRun.mockResolvedValue(dryRunResult);
+    mockScaffolderService.dryRun.mockResolvedValue(dryRunResult);
 
     createDryRunTemplateAction({
       actionsRegistry: mockActionsRegistry,
-      scaffolderClient: mockScaffolderClient as any,
+      scaffolderService: mockScaffolderService,
     });
 
     const result = await mockActionsRegistry.invoke({
-      id: 'test:validate-scaffolder',
+      id: 'test:dry-run-template',
       input: { templateYaml: validTemplateYaml },
     });
 
@@ -107,30 +104,34 @@ describe('createDryRunTemplateAction', () => {
       steps: [{ id: 'step-1', name: 'Step One', action: 'debug:log' }],
     });
 
-    expect(mockScaffolderClient.dryRun).toHaveBeenCalledWith({
-      template: expect.objectContaining({
-        apiVersion: 'scaffolder.backstage.io/v1beta3',
-        kind: 'Template',
-        metadata: expect.objectContaining({
-          name: 'test-template',
+    expect(mockScaffolderService.dryRun).toHaveBeenCalledWith(
+      {
+        template: expect.objectContaining({
+          apiVersion: 'scaffolder.backstage.io/v1beta3',
+          kind: 'Template',
+          metadata: expect.objectContaining({
+            name: 'test-template',
+          }),
         }),
-      }),
-      values: {},
-      directoryContents: [],
-    });
+        values: {},
+        directoryContents: [],
+      },
+      { credentials: expect.anything() },
+    );
   });
 
-  it('should pass values and files to scaffolderClient.dryRun', async () => {
+  it('should pass values and files to the scaffolder service', async () => {
     const mockActionsRegistry = actionsRegistryServiceMock();
-    mockScaffolderClient.dryRun.mockResolvedValue({
+    mockScaffolderService.dryRun.mockResolvedValue({
       log: [],
       output: {},
       steps: [],
+      directoryContents: [],
     });
 
     createDryRunTemplateAction({
       actionsRegistry: mockActionsRegistry,
-      scaffolderClient: mockScaffolderClient as any,
+      scaffolderService: mockScaffolderService,
     });
 
     const values = { name: 'my-app' };
@@ -142,7 +143,7 @@ describe('createDryRunTemplateAction', () => {
     ];
 
     await mockActionsRegistry.invoke({
-      id: 'test:validate-scaffolder',
+      id: 'test:dry-run-template',
       input: {
         templateYaml: validTemplateYaml,
         values,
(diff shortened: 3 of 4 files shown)


plugins/scaffolder-backend/src/actions/createValidateScaffolderAction.ts
```diff
@@ -0,0 +1,242 @@
...
+      idempotent: true,
+    },
+    description: `Validates a scaffolder template by calling the /v2/dry-run endpoint.
+This action:
+- Parses and validates the YAML template structure
+- Validates the template against the template entity schema
+- Calls the /v2/dry-run endpoint to dry-run template steps without side effects
+- Returns success with execution logs or specific validation errors`,
+    schema: {
+      input: (z: any) =>
```
benjdlambert:
```suggestion
      input: (z) =>
```

plugins/scaffolder-backend/src/actions/createValidateScaffolderAction.ts
```diff
@@ -0,0 +1,242 @@
...
+                  .describe('File path relative to template root'),
+                base64Content: z
+                  .string()
+                  .describe('Base64-encoded file content'),
+              }),
+            )
+            .optional()
+            .describe('Optional directory contents for template files'),
+        }),
+      output: (z: any) =>
```
benjdlambert:
```suggestion
      output: (z) =>
```

plugins/scaffolder-backend/src/actions/createValidateScaffolderAction.ts
```diff
@@ -0,0 +1,242 @@
...
+            .describe('Parsed template steps'),
+        }),
+    },
+    action: async ({
+      input,
+      credentials,
+    }: {
+      input: any;
+      credentials: any;
+    }) => {
```
benjdlambert:
```suggestion
    action: async ({
      input,
      credentials,
    }) => {
```

plugins/scaffolder-backend/src/actions/createValidateScaffolderAction.ts
```diff
@@ -0,0 +1,242 @@
...
+                parseError.linePos
+                  ? `At line ${parseError.linePos[0].line}, column ${parseError.linePos[0].col}`
+                  : '',
+              ].filter(Boolean),
+            },
+          };
+        }
+
+        // Validate template structure
+        if (!(await templateEntityV1beta3Validator.check(template))) {
```
benjdlambert: I think that this should happen in the dry-run endpoint already no?

plugins/scaffolder-backend/src/actions/createValidateScaffolderAction.ts
```diff
@@ -0,0 +1,242 @@
...
+            output: {
+              valid: false,
+              message: 'Template has no steps defined',
+              errors: ['Template must have at least one step'],
+            },
+          };
+        }
+
+        // Get a token to call the scaffolder API
+        const { token } = await auth.getPluginRequestToken({
```
benjdlambert: I think that we should just replace all this code with creating a `ScaffolderClient` instead, and calling `dryRun` on that like the frontend code does.

The client should be isomorphic?

Hopefully when using the `ScaffolderClient` that should help us clean up the error handing below, as we should be able to just `throw ForwardedError(ex)` or something similar?

If that doesn't work then we need to look at how errors are being propagated through the actions service properly.

plugins/scaffolder-backend/src/actions/index.ts
```diff
@@ -13,11 +13,14 @@
...
-  CatalogScmEvent,
-  CatalogScmEventContext,
-  CatalogScmEventsService,
-  CatalogScmEventsServiceSubscriber,
-} from './types';
+export const createScaffolderActions = (options: {
+  actionsRegistry: ActionsRegistryService;
+  auth: AuthService;
+  discovery: DiscoveryService;
+}) => {
```
benjdlambert, line 23: Thinking here might be a good place to create a client that gets passed through to `createDryRunTemplateAction`.

plugins/scaffolder-backend/src/actions/index.ts
```diff
@@ -13,11 +13,14 @@
...
+  createValidateScaffolderAction(options);
```
benjdlambert:
```suggestion
  createDryRunTemplateAction(options);
```

plugins/scaffolder-backend/src/actions/createValidateScaffolderAction.ts
benjdlambert: Actually, I think a better solution here might be hit the catalog for this. I think there's a `validateEntity` method on the Catalog which should take an entity. That way we're validating it against the running Catalog and the custom model it might have, rather than the one that's shipped with the scaffolder module.

Although, we could actually just implement that logic in the `dry-run` endpoint entirely to be honest if it's not already doing that today.

This is probably the same to be said for the `yaml.parse` parts above too right? Maybe let's just lean on the Catalog for this?

plugins/scaffolder-backend/src/actions/createValidateScaffolderAction.test.ts
```diff
@@ -0,0 +1,424 @@
...
+    getPluginRequestToken: jest.Mock;
+  };
+  let mockDiscovery: {
+    getBaseUrl: jest.Mock;
+  };
+  let mockFetch: jest.Mock;
+  let originalFetch: typeof global.fetch;
+
+  beforeEach(() => {
+    mockAuth = {
```
benjdlambert: You can use `mockServices.auth()`  and `mockServices.discovery()`.

plugins/scaffolder-backend/src/actions/createValidateScaffolderAction.ts
yangcao77: I don't know we even have a `ScaffolderClient` in place. checked it's under `scaffolder-common` plugin. 
I've updated the code to use `ScaffolderClient` and `throw ForwardedError` as suggested

yangcao77: Actually. Instead of directly throwing the `ForwardedError`. I still handled the error returned from `ScaffolderClient`

I was manually trying some failure cases with the dry-run action, but I can only see the error code (500/403/400), but no detailed errors. and therefore the MCP client cannot give proper suggestions on the validation error neither. 

Backstage `ResponseError` sets its message to that generic Request failed with `${statusCode} ${statusText}`.
Any real error details (e.g. stack trace, validation message) are in:
`error.cause` and/or `error.body`
If the action rethrows with `ForwardedError('Template validation failed', error)`, and the MCP response only shows the string form of that error, which doesn’t include cause or body.

plugins/scaffolder-backend/src/actions/createValidateScaffolderAction.ts
yangcao77: double checked the template validation has already been done in the dry-run endpoint. so directly calling the dry-run endpoint. I will leave the implementation of hitting `validateEntity` in dry-run endpoint as out of scope of this PR. 

in terms of the `yaml.parse`,  I checked even `validateEntityAction` parses the yaml before calling `catalog.validateEntity`, https://github.com/backstage/backstage/blob/c08353e318f74bd4caf12a44100cc5eb026463ee/plugins/catalog-backend/src/actions/createValidateEntityAction.ts#L61-L78

plugins/scaffolder-backend/src/actions/createValidateScaffolderAction.ts
benjdlambert: OK, I think that we should probably try and surface these errors better. I wonder if we should move your logic for extracting the errors messages into the `mcp-actions` backend plugin and when it executes an action as a tool instead to pull out the messages instead and then we get this functionality in all the calls rather than having to implement it always at the action level which can be easily missed.

yangcao77: That is a good point! However, I was trying to implement the error extraction logic in `mcp-actions-backend`, `handleErrors`, and was tracing error flow.

The MCP backend does not run the action in-process. Instead, it calls `DefaultActionsService.invoke()`, which does an HTTP POST to the scaffolder backend (e.g. `POST .../actions/scaffolder/validate-scaffolder/invoke`). 
The action throws on the scaffolder backend, which returns an error code with a serialized error, at this point, the error status got changed. During the cases I tested out, it is changed to `500 Server Internal Error` while the actual original error code was `400`.

for example:
```
  body: {
    error: {
      name: 'ResponseError',
      message: 'Failed execution of action "scaffolder:validate-scaffolder"; caused by ResponseError: Template validation failed; caused by ResponseError: Request failed with 400 Bad Request',
      cause: [Object],
      stack: 'ResponseError: Failed execution of action "scaffolder:validate-scaffolder"; caused by ResponseError: Template validation failed; caused by ResponseError: Request failed with 400 Bad Request\n' +
        '    at <anonymous> (/Users/stephanie/Documents/backstage/packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.ts:154:17)\n' +
        '    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)'
    },
    request: {
      method: 'POST',
      url: '/.backstage/actions/v1/actions/scaffolder%3Avalidate-scaffolder/invoke'
    },
    response: { statusCode: 500 }
  },
  cause: Error [ResponseError]: Failed execution of action "scaffolder:validate-scaffolder"; caused by ResponseError: Template validation failed; caused by ResponseError: Request failed with 400 Bad Request
      at <anonymous> (/Users/stephanie/Documents/backstage/packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.ts:154:17)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
    cause: { cause: [Object] }
  },
  statusCode: 500,
  statusText: 'Internal Server Error'
```

The MCP backend receives a new `ResponseError`, via `ResponseError.fromResponse(response)`,  whose body contains that serialized chain. The real message is nested at `body.error.cause.cause.body.error.message` 

Extracting the nested, innermost error, is complicated and confusing, in addition, even the error status got changed. If we handle it at the action level,it will be more clear & easy to understand, IMO.

What do you think? Do we still want to proceed this change?

benjdlambert: Hey :wave: thanks for the digging here!

It looks like there's something wrong here though. The client seems to be returning 400 but then the `'/.backstage/actions/v1/actions/scaffolder%3Avalidate-scaffolder/invoke` endpoint is returning 500 which isn't great.

The `ForwardedError` should I think deal with this, but it looks like something might be missing perhaps?

@Rugvip do you have any pointers here as to how we should treat these errors? Am I missing something obvious?

benjdlambert: @yangcao77 I put together this: https://github.com/backstage/backstage/pull/33021 which I think might help. Can you try on top of that?

yangcao77: I checkout lastest master branch (HEAD at commit `858eb3755c01bc7504e6d25b3ac0712128488fb6`)

I still see the same behavior:

```
    at Function.fromResponse (/Users/stephanie/Documents/backstage/packages/errors/src/errors/ResponseError.ts:76:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async DefaultActionsService.invoke (/Users/stephanie/Documents/backstage/packages/backend-defaults/src/alpha/entrypoints/actions/DefaultActionsService.ts:116:13)
    at async <anonymous> (/Users/stephanie/Documents/backstage/plugins/mcp-actions-backend/src/services/McpService.ts:124:30)
    at async handleErrors (/Users/stephanie/Documents/backstage/plugins/mcp-actions-backend/src/services/handleErrors.ts:83:12)
    at async <anonymous> (/Users/stephanie/Documents/backstage/plugins/mcp-actions-backend/src/services/McpService.ts:116:24)
    at async wrappedHandler (/Users/stephanie/Documents/backstage/node_modules/@modelcontextprotocol/sdk/src/server/index.ts:262:32) {
  response: Response {
    status: 500,
    statusText: 'Internal Server Error',
    headers: Headers {
      'content-security-policy': "default-src 'self';base-uri 'self';font-src 'self' https: data:;frame-ancestors 'self';img-src 'self' data:;object-src 'none';script-src 'self' 'unsafe-eval';script-src-attr 'none';style-src 'self' https: 'unsafe-inline';upgrade-insecure-requests;connect-src 'self' http: https:",
      'referrer-policy': 'no-referrer',
      'strict-transport-security': 'max-age=15552000; includeSubDomains',
      'x-content-type-options': 'nosniff',
      'x-dns-prefetch-control': 'off',
      'x-download-options': 'noopen',
      'x-frame-options': 'SAMEORIGIN',
      'x-permitted-cross-domain-policies': 'none',
      'x-xss-protection': '0',
      'content-type': 'application/json; charset=utf-8',
      etag: 'W/"3db8-4OmhHmgPfpYQqi6x5WPzcKmfr1I"',
      vary: 'Accept-Encoding',
      'content-encoding': 'gzip',
      date: 'Thu, 26 Feb 2026 22:32:11 GMT',
      connection: 'keep-alive',
      'keep-alive': 'timeout=5',
      'transfer-encoding': 'chunked'
    },
    body: ReadableStream { locked: true, state: 'closed', supportsBYOB: true },
    bodyUsed: true,
    ok: false,
    redirected: false,
    type: 'basic',
    url: 'http://localhost:7007/api/scaffolder/.backstage/actions/v1/actions/scaffolder%3Avalidate-scaffolder/invoke'
  },
  body: {
    error: {
      name: 'ResponseError',
      message: 'Template validation failed; caused by ResponseError: Request failed with 400 Bad Request',
      cause: [Object],
      stack: 'ResponseError: Template validation failed; caused by ResponseError: Request failed with 400 Bad Request\n' +
        '    at Object.action (/Users/stephanie/Documents/backstage/plugins/scaffolder-backend/src/actions/createDryRunTemplateAction.ts:183:15)\n' +
        '    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n' +
        '    at async <anonymous> (/Users/stephanie/Documents/backstage/packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.ts:129:24)'
    },
    request: {
      method: 'POST',
      url: '/.backstage/actions/v1/actions/scaffolder%3Avalidate-scaffolder/invoke'
    },
    response: { statusCode: 500 }
  },
  cause: Error [ResponseError]: Template validation failed; caused by ResponseError: Request failed with 400 Bad Request
      at Object.action (/Users/stephanie/Documents/backstage/plugins/scaffolder-backend/src/actions/createDryRunTemplateAction.ts:183:15)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async <anonymous> (/Users/stephanie/Documents/backstage/packages/backend-defaults/src/alpha/entrypoints/actionsRegistry/DefaultActionsRegistryService.ts:129:24) {
    cause: {
      response: {},
      body: [Object],
      cause: {},
      statusCode: 400,
      statusText: 'Bad Request'
    }
  },
  statusCode: 500,
  statusText: 'Internal Server Error'
}
```

benjdlambert:
Hey @yangcao77, I hope you don't mind I pushed a few updates to this branch now.

We merged in a `scaffolderServiceRef` which is a server side version of the `scaffolderClient` is aware of `BackstageCredentials` which means that it proxies through the user auth correctly.

I rebased this on top of the the MCP changes from yesterday with the error handling and it seems to work for me now when I get an error dry-running.

image


It's no longer embedded and wrapped up.

yangcao77:
@benjdlambert Thanks!
I've checked out the new changes to this branch, and confirm that `handleErrors` in `mcp-actions-backend` can get the error from the dry run with the correct status code now:

Screenshot 2026-02-27 at 6 15 11 PM

PR #32920, opened 2026-02-19
feat(cli): add auth commands for OIDC login

benjdlambert:
This is the last part of the whole `offline_access` + `CIMD` + `CLI Auth` branches.

### DIFF
--- a/packages/cli/src/modules/auth/commands/login.ts
+++ b/packages/cli/src/modules/auth/commands/login.ts
@@ -321,7 +321,7 @@ async function persistInstance(options: {
   const { instanceName, backendBaseUrl, clientId, token } = options;
   const secretStore = await getSecretStore();
   await withMetadataLock(async () => {
-    const service = `backstage-cli:instance:${instanceName}`;
+    const service = `backstage-cli:auth-instance:${instanceName}`;
     await secretStore.set(service, 'accessToken', token.access_token);
     if (token.refresh_token) {
       await secretStore.set(service, 'refreshToken', token.refresh_token);
--- a/packages/cli/src/modules/auth/commands/logout.ts
+++ b/packages/cli/src/modules/auth/commands/logout.ts
@@ -47,7 +47,7 @@ export default async ({ args, info }: CommandContext) => {
   await withMetadataLock(async () => {
     const instance = await getInstanceByName(instanceName);
     const secretStore = await getSecretStore();
-    const service = `backstage-cli:instance:${instanceName}`;
+    const service = `backstage-cli:auth-instance:${instanceName}`;
     const refreshToken = (await secretStore.get(service, 'refreshToken')) ?? '';
 
     if (refreshToken) {
--- a/packages/cli/src/modules/auth/commands/printToken.ts
+++ b/packages/cli/src/modules/auth/commands/printToken.ts
@@ -44,7 +44,7 @@ export default async ({ args, info }: CommandContext) => {
   }
 
   const secretStore = await getSecretStore();
-  const service = `backstage-cli:instance:${instance.name}`;
+  const service = `backstage-cli:auth-instance:${instance.name}`;
   const accessToken = await secretStore.get(service, 'accessToken');
   if (!accessToken) {
     throw new Error('No access token found. Run "auth login" to authenticate.');
--- a/packages/cli/src/modules/auth/commands/show.ts
+++ b/packages/cli/src/modules/auth/commands/show.ts
@@ -49,7 +49,7 @@ export default async ({ args, info }: CommandContext) => {
     .replace(/\/$/, '');
 
   const secretStore = await getSecretStore();
-  const service = `backstage-cli:instance:${instance.name}`;
+  const service = `backstage-cli:auth-instance:${instance.name}`;
   const accessToken = await secretStore.get(service, 'accessToken');
   if (!accessToken) {
     throw new Error('No access token found. Run "auth login" to authenticate.');
--- a/packages/cli/src/modules/auth/lib/auth.ts
+++ b/packages/cli/src/modules/auth/lib/auth.ts
@@ -43,7 +43,7 @@ export async function refreshAccessToken(
   return withMetadataLock(async () => {
     const instance = await getInstanceByName(instanceName);
 
-    const service = `backstage-cli:instance:${instanceName}`;
+    const service = `backstage-cli:auth-instance:${instanceName}`;
     const refreshToken = (await secretStore.get(service, 'refreshToken')) ?? '';
     if (!refreshToken) {
       throw new Error(
--- a/packages/cli/src/modules/auth/lib/auth.test.ts
+++ b/packages/cli/src/modules/auth/lib/auth.test.ts
@@ -135,7 +135,7 @@ describe('auth', () => {
 
       expect(mockStorage.getInstanceByName).toHaveBeenCalledWith('test');
       expect(mockSecretStoreInstance.get).toHaveBeenCalledWith(
-        'backstage-cli:instance:test',
+        'backstage-cli:auth-instance:test',
         'refreshToken',
       );
       expect(mockHttp.httpJson).toHaveBeenCalledWith(
@@ -150,12 +150,12 @@ describe('auth', () => {
         },
       );
       expect(mockSecretStoreInstance.set).toHaveBeenCalledWith(
-        'backstage-cli:instance:test',
+        'backstage-cli:auth-instance:test',
         'accessToken',
         'new-access-token',
       );
       expect(mockSecretStoreInstance.set).toHaveBeenCalledWith(
-        'backstage-cli:instance:test',
+        'backstage-cli:auth-instance:test',
         'refreshToken',
         'new-refresh-token',
       );


packages/cli/src/modules/auth/commands/login.ts
```diff
@@ -0,0 +1,366 @@
...
+
+  const clientId = `${authBaseUrl}/.well-known/oauth-client/cli.json`;
+
+  // Verify the server supports CIMD before starting the auth flow
+  const metadataResponse = await fetch(clientId);
+  if (!metadataResponse.ok) {
+    throw new Error(
+      `Server does not support CLI authentication. Ensure CIMD is enabled on the backend.`,
+    );
+  }
```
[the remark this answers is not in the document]
freben: +1 to this, cleanup handling

packages/cli/src/modules/auth/commands/login.ts
```diff
@@ -0,0 +1,366 @@
...
+  });
+
+  await persistInstance({
+    instanceName,
+    backendBaseUrl,
+    clientId,
+    token,
+  });
+
+  process.stderr.write('Login successful\n');
```
freben: stderr?

packages/cli/src/modules/auth/commands/login.ts
```diff
@@ -0,0 +1,366 @@
...
+
+async function pickBaseUrl() {
+  const cwd = process.cwd();
+  const candidates: Array<{ url: string; file: string }> = [];
+
+  const patterns = [
+    'app-config.yaml',
+    'app-config.*.yaml',
+    'packages/*/app-config.yaml',
+    'packages/*/app-config.*.yaml',
```
freben, line 179: Interesting, why these? And not plugins?

packages/cli/src/modules/auth/commands/login.ts
```diff
@@ -0,0 +1,366 @@
...
+  ];
+  const files = patterns.flatMap(p => glob.sync(p, { cwd, nodir: true }));
+  for (const file of files) {
+    try {
+      const content = await fs.readFile(path.resolve(cwd, file), 'utf8');
+      const doc = YAML.parse(content);
+      const url = doc?.backend?.baseUrl as string | undefined;
+      if (url) {
```
freben, line 187: probably check for dupes here?

packages/cli/src/modules/auth/commands/login.ts
```diff
@@ -0,0 +1,366 @@
...
+  return picked;
+}
+
+function normalizeUrl(u: string): string;
+function normalizeUrl(u: string | undefined): string | undefined;
+function normalizeUrl(u: string | undefined): string | undefined {
+  if (u === undefined) {
+    return undefined;
+  }
+  const url = new URL(u);
```
freben: does this need a try/catch with a human readable "X is not a valid URL" or so?

packages/cli/src/modules/auth/commands/login.ts
```diff
@@ -0,0 +1,366 @@
...
+  authorize.searchParams.set('code_challenge', challenge);
+  authorize.searchParams.set('code_challenge_method', 'S256');
+  return authorize.toString();
+}
+
+async function openBrowserOrPrint(url: string, noBrowser?: boolean) {
+  if (noBrowser) {
+    process.stderr.write(`Open this URL to continue: ${url}\n`);
+  } else {
+    process.stderr.write(`Opening the following URL: ${url}\n`);
```
freben: when do we use stderr vs stdout really? 🤔 what's the strategy

packages/cli/src/modules/auth/commands/login.ts
```diff
@@ -0,0 +1,366 @@
...
+      process.stderr.write(
+        'Warning: No refresh token received. You will need to re-authenticate when the access token expires.\n',
+      );
+    }
+    await upsertInstance({
+      name: instanceName,
+      baseUrl: backendBaseUrl,
+      clientId,
+      issuedAt: Date.now(),
+      accessTokenExpiresAt: Date.now() + token.expires_in * 1000,
```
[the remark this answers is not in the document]
freben: ah this seems to make sense right

packages/cli/src/modules/auth/commands/logout.ts
```diff
@@ -0,0 +1,73 @@
...
+          },
+          signal: AbortSignal.timeout(30_000),
+        });
+      } catch {
+        // ignore errors per RFC 7009
+      }
+    }
+
+    await secretStore.delete(service, 'clientSecret');
+    await secretStore.delete(service, 'refreshToken');
```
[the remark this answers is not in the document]
freben: Makes sense

packages/cli/src/modules/auth/commands/show.ts
```diff
@@ -0,0 +1,61 @@
...
+    .option('instance', {
+      type: 'string',
+      desc: 'Name of the instance to show',
+    })
+    .parse();
+
+  let instance = await getSelectedInstance(parsed.instance);
+
+  if (accessTokenNeedsRefresh(instance)) {
+    instance = await refreshAccessToken(instance.name);
```
freben, line 45: Or should it just show (expired) or whatever? Not sure the show command necessarily should be refreshing. Maybe an extra cmdline option to enable this?

packages/cli/src/modules/auth/lib/localServer.ts
```diff
@@ -0,0 +1,93 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import http from 'node:http';
+import { URL } from 'node:url';
+
+const CALLBACK_PORT = 8055;
```
freben, line 20: Would it be useful for this to be random - or is it part of the contract (allowlisting callbacks remotely for example) that it should be some certain port?

packages/cli/src/modules/auth/lib/localServer.ts
```diff
@@ -0,0 +1,93 @@
...
+      return;
+    }
+    const u = new URL(req.url, 'http://127.0.0.1');
+    if (u.pathname !== '/callback') {
+      res.statusCode = 404;
+      res.end('Not Found');
+      return;
+    }
+    const code = u.searchParams.get('code') ?? undefined;
+    const state = u.searchParams.get('state') ?? undefined;
```
freben, line 51: The caller checks the state?

packages/cli/src/modules/auth/lib/secretStore.ts
```diff
@@ -0,0 +1,109 @@
...
+  async delete(service: string, account: string): Promise<void> {
+    await this.keytar.deletePassword(service, account);
+  }
+}
+
+class FileSecretStore implements SecretStore {
+  private readonly baseDir: string;
+  constructor() {
+    const root =
+      process.env.XDG_DATA_HOME ||
```
freben, line 61: whoa such stuff

packages/cli/src/modules/auth/lib/storage.ts
```diff
@@ -0,0 +1,175 @@
...
+
+import { NotFoundError } from '@backstage/errors';
+import fs from 'fs-extra';
+import os from 'node:os';
+import path from 'node:path';
+import lockfile from 'proper-lockfile';
+import YAML from 'yaml';
+import { z } from 'zod';
+
+const METADATA_FILE = 'instances.yaml';
```
freben: maybe more specific, like `auth-instances.yaml`? since it's directly under `backstage-cli/`

packages/cli/src/modules/auth/lib/secretStore.ts
```diff
@@ -0,0 +1,109 @@
...
+      (process.platform === 'win32'
+        ? process.env.APPDATA || path.join(os.homedir(), 'AppData', 'Roaming')
+        : path.join(os.homedir(), '.local', 'share'));
+    this.baseDir = path.join(root, 'backstage-cli', 'secrets');
```
freben: more specific like `auth-secrets`? or will we generalise this one to potentially be used for general secret storage

freben (changes requested):

packages/cli/src/modules/auth/lib/localServer.ts
benjdlambert: It needs to be fixed for the CIMD document callback URLs.

packages/cli/src/modules/auth/commands/login.ts
Rugvip: happens when converting to the list just below

packages/cli/src/modules/auth/commands/login.ts
Rugvip: 🤷, these should prolly be stdout

packages/cli/src/modules/auth/commands/logout.ts
```diff
@@ -0,0 +1,84 @@
...
+  await withMetadataLock(async () => {
+    const instance = await getInstanceByName(instanceName);
+    const clientId = instance.clientId;
+    const secretStore = await getSecretStore();
+    const service = `backstage-cli:instance:${instanceName}`;
+    const clientSecret = (await secretStore.get(service, 'clientSecret')) ?? '';
+    const refreshToken = (await secretStore.get(service, 'refreshToken')) ?? '';
+
+    if (clientId && clientSecret && refreshToken) {
+      const basic = Buffer.from(`${clientId}:${clientSecret}`).toString(
```
Rugvip: Couple of remnants here right? we no client secret anymore

packages/cli/src/modules/auth/lib/localServer.ts
Rugvip: Which I guess could be configured, but happy to start here

packages/cli/src/modules/auth/commands/show.ts
benjdlambert: Can add this in a followup, wanna get this 🚢

packages/cli/src/modules/auth/lib/auth.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    const parsed = TokenResponseSchema.safeParse(response);
+    if (!parsed.success) {
+      throw new Error(`Invalid token response: ${parsed.error.message}`);
+    }
+    const token = parsed.data;
+
+    await secretStore.set(service, 'accessToken', token.access_token);
+    await secretStore.set(service, 'refreshToken', token.refresh_token);
+    const newInstance = {
```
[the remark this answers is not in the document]
benjdlambert: Fixed. `refresh_token` is optional.

packages/cli/src/modules/auth/lib/localServer.ts
benjdlambert: probably doesn't need to as the server will reject if mismatched right?

packages/cli/src/modules/auth/lib/storage.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+  await writeAll(data);
+}
+
+export async function removeInstance(name: string): Promise<void> {
+  const data = await readAll();
+  const next = data.instances.filter(i => i.name !== name);
+  if (next.length !== data.instances.length) {
+    await writeAll({ instances: next });
+  }
```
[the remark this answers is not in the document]
benjdlambert: It's already called within a lock, can't do it again otherwise it will deadlock.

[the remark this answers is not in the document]
benjdlambert: see above

packages/cli/src/modules/auth/index.ts
```diff
@@ -0,0 +1,73 @@
...
+      },
+    });
+    reg.addCommand({
+      path: ['auth', 'logout'],
+      description: 'Log out the CLI and revoke refresh token',
+      execute: async ({ args }) => {
+        yargs().parse(args);
+        await commands.logout(args);
+      },
+    });
```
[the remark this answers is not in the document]
benjdlambert: Fixed

packages/cli/src/modules/auth/commands/printToken.ts
```diff
@@ -0,0 +1,54 @@
...
+  );
+
+  let instance = await getSelectedInstance(instanceFlag);
+
+  if (accessTokenNeedsRefresh(instance)) {
+    instance = await refreshAccessToken(instance.name);
+  }
+
+  const secretStore = await getSecretStore();
+  const service = `backstage-cli:instance:${instance.name}`;
```
freben: still wondering if these names could be more like `auth-instance` or `cmd-instance` or whatnot

PR #32939, opened 2026-02-20
cli-common: new path utils that simplifies use in dependencies

Rugvip:
## Summary

Splits the `findPaths` API from `@backstage/cli-common` into two focused replacements:

- **`targetPaths`** — a lazy singleton for cwd-based path resolution (no `__dirname` needed). Re-resolves automatically when `process.cwd()` changes.
- **`findOwnPaths(__dirname)`** — for package-relative path resolution. Results are cached per package root with a hierarchical directory cache.

The old `findPaths` is deprecated but preserved for backward compatibility.

All consumers across the repo are migrated to the new API, and `findOwnPaths` is called lazily inside functions rather than at module scope.

### DIFF
--- a/.changeset/fix-prettier-existence-check.md
+++ b/.changeset/fix-prettier-existence-check.md
@@ -1,0 +1,5 @@
+---
+'@backstage/repo-tools': patch
+---
+
+Fixed prettier existence checks in OpenAPI commands to use `fs.pathExists` instead of checking the resolved path string, which was always truthy.
--- a/packages/e2e-test/src/commands/runCommand.ts
+++ b/packages/e2e-test/src/commands/runCommand.ts
@@ -27,11 +27,11 @@ import { waitFor, print } from '../lib/helpers';
 import mysql from 'mysql2/promise';
 import pgtools from 'pgtools';
 
+import { OptionValues } from 'commander';
 import { findOwnPaths, runOutput, run } from '@backstage/cli-common';
 
 /* eslint-disable-next-line no-restricted-syntax */
 const ownPaths = findOwnPaths(__dirname);
-import { OptionValues } from 'commander';
 
 const templatePackagePaths = [
   'packages/cli/templates/frontend-plugin/package.json.hbs',
--- a/packages/repo-tools/src/commands/package/schema/openapi/generate/client.ts
+++ b/packages/repo-tools/src/commands/package/schema/openapi/generate/client.ts
@@ -87,7 +87,7 @@ async function generate(
 
   await fs.writeFile(
     resolve(parentDirectory, 'index.ts'),
-    `// 
+    `//
     export * from './generated';`,
   );
 
@@ -96,7 +96,7 @@ async function generate(
   });
 
   const prettier = targetPaths.resolveRoot('node_modules/.bin/prettier');
-  if (prettier) {
+  if (await fs.pathExists(prettier)) {
     await exec(`${prettier} --write ${parentDirectory}`, [], {
       signal: abortSignal?.signal,
     });
--- a/packages/repo-tools/src/commands/package/schema/openapi/generate/server.ts
+++ b/packages/repo-tools/src/commands/package/schema/openapi/generate/server.ts
@@ -77,7 +77,9 @@ export const createOpenApiRouter = async (
   );
 
   await exec(`yarn backstage-cli package lint`, ['--fix', tsPath, indexFile]);
-  if (await targetPaths.resolveRoot('node_modules/.bin/prettier')) {
+  if (
+    await fs.pathExists(targetPaths.resolveRoot('node_modules/.bin/prettier'))
+  ) {
     await exec(`yarn prettier`, ['--write', tsPath, indexFile], {
       cwd: targetPaths.rootDir,
     });
--- a/packages/repo-tools/src/commands/package/schema/openapi/init.ts
+++ b/packages/repo-tools/src/commands/package/schema/openapi/init.ts
@@ -61,10 +61,12 @@ capture:
                 # 🔧 Specify a command that will generate traffic
                 command: yarn backstage-cli package test --no-watch ${ROUTER_TEST_PATHS.map(
                   e => `"${e}"`,
-                ).join(' ')} 
+                ).join(' ')}
   `,
   );
-  if (await targetPaths.resolveRoot('node_modules/.bin/prettier')) {
+  if (
+    await fs.pathExists(targetPaths.resolveRoot('node_modules/.bin/prettier'))
+  ) {
     await exec(`yarn prettier`, ['--write', opticConfigFilePath]);
   }
 }
--- a/packages/repo-tools/src/commands/repo/schema/openapi/test.ts
+++ b/packages/repo-tools/src/commands/repo/schema/openapi/test.ts
@@ -82,7 +82,9 @@ async function test(
     throw err;
   }
   if (
-    (await targetPaths.resolveRoot('node_modules/.bin/prettier')) &&
+    (await fs.pathExists(
+      targetPaths.resolveRoot('node_modules/.bin/prettier'),
+    )) &&
     options?.update
   ) {
     await exec(`yarn prettier`, ['--write', openapiPath]);


packages/cli/src/lib/version.ts
```diff
@@ -16,7 +16,9 @@
 
 import fs from 'fs-extra';
 import semver from 'semver';
-import { paths } from './paths';
+import { findOwnPaths } from '@backstage/cli-common';
+
+const ownPaths = findOwnPaths(__dirname);
```
Rugvip, line 23: Needs to be below the import

packages/cli/src/lib/yarnPlugin.test.ts
```diff
@@ -19,8 +19,9 @@ import { getHasYarnPlugin } from './yarnPlugin';
 
 const mockDir = createMockDirectory();
 
-jest.mock('./paths', () => ({
-  paths: {
+jest.mock('@backstage/cli-common', () => ({
+  ...jest.requireActual('@backstage/cli-common'),
+  targetPaths: {
```
Rugvip: Use a spy here instead

packages/cli-common/src/paths.ts
```diff
@@ -26,34 +26,62 @@ import { dirname, resolve as resolvePath } from 'node:path';
...
+  resolveTargetRoot: ResolveFunc;
+};
+
+/**
+ * Paths relative to the package that the calling code lives in. Requires
+ * `__dirname` to locate the package root.
+ *
+ * @public
+ */
+export type OwnPaths = {
```
Rugvip, line 54: With the new division into path types we don't need the properties to also include the type in the name. Let's also remove the non-resolve variants too, meaning that this should have just `resolve` and `resolveRoot`. Make the same changes for the target paths too

packages/cli-common/src/paths.ts
```diff
@@ -26,34 +26,62 @@ import { dirname, resolve as resolvePath } from 'node:path';
 export type ResolveFunc = (...paths: string[]) => string;
 
 /**
- * Common paths and resolve functions used by the cli.
- * Currently assumes it is being executed within a monorepo.
+ * Paths relative to the target project that the CLI is operating on, based on
+ * `process.cwd()`. This can be imported directly — no `__dirname` needed.
+ *
+ * Lazily initialized on first access and re-resolved if `process.cwd()` changes.
```
Rugvip: Update these doc comments to be from the point of view of a user seeing them in isolation without the history from this PR

.changeset/cli-common-cached-paths.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli-common': patch
+---
+
+Added hierarchical caching to `findOwnDir`, avoiding redundant filesystem walks when `findPaths` is called from multiple locations within the same package.
```
Rugvip: Focus on the end user impact instead

packages/cli/src/modules/build/lib/bundler/paths.ts
```diff
@@ -16,19 +16,21 @@
 
 import fs from 'fs-extra';
 import { resolve as resolvePath } from 'node:path';
-import { paths } from '../../../../lib/paths';
+import { targetPaths, findOwnPaths } from '@backstage/cli-common';
+
+const ownPaths = findOwnPaths(__dirname);
```
Rugvip: Be lazy about `findOwnPaths`, call only when needed and never at the top-level of a module

packages/backend/src/index.ts
```diff
@@ -32,7 +32,7 @@ const searchLoader = createBackendFeatureLoader({
   *loader({ config }) {
     yield import('@backstage/plugin-search-backend');
     yield import('@backstage/plugin-search-backend-module-catalog');
-    yield import('@backstage/plugin-search-backend-module-explore');
+    yield import('@backstage-community/plugin-search-backend-module-explore');
```
Rugvip: Open a separate PR for this change

packages/cli-common/src/paths.ts
```diff
@@ -131,40 +262,49 @@ export function findPaths(searchDir: string): Paths {
...
 
+  ownPathsCache.set(ownDir, paths);
+  return paths;
+}
+
+/**
+ * Find paths related to a package and its execution context.
+ *
+ * @public
+ * @deprecated Use {@link targetPaths} for cwd-based paths and
```
Rugvip, line 302: This should indeed be depreacated, but this PR really remove all existing usage in the repo? Make sure all usage of this in the repo is removed

Rugvip (commented):
This missing API report updates

packages/backend-dynamic-feature-service/src/manager/plugin-manager.test.ts
```diff
@@ -38,8 +38,33 @@ import { createSpecializedBackend } from '@backstage/backend-app-api';
 import { ConfigSources } from '@backstage/config-loader';
 import { Logs, MockedLogger, LogContent } from '../__testUtils__/testUtils';
 import { PluginScanner } from '../scanner/plugin-scanner';
-import { findPaths } from '@backstage/cli-common';
+import { targetPaths } from '@backstage/cli-common';
 import { createMockDirectory } from '@backstage/backend-test-utils';
+
+jest.mock('@backstage/cli-common', () => {
```
Rugvip: use a spy instead of this excessive mocking

packages/repo-tools/src/lib/paths.ts
```diff
@@ -14,13 +14,26 @@
...
 
-import { findPaths } from '@backstage/cli-common';
+import { targetPaths, findOwnPaths } from '@backstage/cli-common';
 import { PackageGraph } from '@backstage/cli-node';
 import { Minimatch } from 'minimatch';
 import { isAbsolute, relative as relativePath } from 'node:path';
 
 /* eslint-disable-next-line no-restricted-syntax */
-export const paths = findPaths(__dirname);
+export const paths = {
```
Rugvip: refactor usage of this instead

packages/cli-common/src/paths.ts
```diff
@@ -107,22 +151,88 @@ export function findOwnRootDir(ownDir: string) {
...
+        const data = JSON.parse(content);
+        return Boolean(data.workspaces);
+      } catch (error) {
+        throw new Error(
+          `Failed to parse package.json file while searching for root, ${error}`,
+        );
+      }
+    }) ?? targetDir;
+  return cachedTargetRoot;
+}
```
Rugvip: make the implementation of this an internal class instead, but still export a singleton instance of it via `export const targetPaths = new ...`. Make sure the constructor doesn't do any eager work. Do a similar refactor for the own dir, but ofc without the singleton. For both cases, try to unify the caching a bit between the two methods.

packages/cli/src/modules/build/commands/package/build/command.ts
```diff
@@ -41,19 +42,19 @@ export async function command(opts: OptionValues): Promise<void> {
...
         return arg;
       }
-      return paths.resolveTarget(arg);
+      return targetPaths.resolve(arg);
     });
 
     if (role === 'frontend') {
       return buildFrontend({
-        targetDir: paths.targetDir,
+        targetDir: targetPaths.resolve(),
```
Rugvip: Let's add back the props as this is a bit silly, make them `.dir` and `.rootDir`

.changeset/cli-common-cached-paths.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli-common': minor
+---
+
+Added `targetPaths` and `findOwnPaths` as replacements for `findPaths`, with a cleaner separation between target project paths and package-relative paths.
```
Rugvip, line 5: Provide more specific instructions for how to migrate existing usage, with a simple example for each of the old props

packages/cli-common/src/paths.ts
```diff
@@ -107,64 +122,189 @@ export function findOwnRootDir(ownDir: string) {
   return resolvePath(ownDir, '../..');
 }
 
+// Hierarchical directory cache shared across all OwnPathsImpl instances.
+// When we resolve a searchDir to its package root, we also cache every
+// intermediate directory, so sibling directories share work.
+const dirCache = new Map<string, string>();
```
Rugvip, line 128: Keep this as a static on the class too

packages/backend-dynamic-feature-service/src/manager/plugin-manager.test.ts
```diff
@@ -38,14 +38,19 @@ import { createSpecializedBackend } from '@backstage/backend-app-api';
...
 import { createMockDirectory } from '@backstage/backend-test-utils';
 import { rootLifecycleServiceFactory } from '@backstage/backend-defaults/rootLifecycle';
 import { BackstagePackageJson, PackageRole } from '@backstage/cli-node';
 
 describe('backend-dynamic-feature-service', () => {
   const mockDir = createMockDirectory();
 
+  beforeAll(() => {
+    overrideTargetPaths(require('path').resolve(__dirname, '../../../..'));
+  });
```
Rugvip: doesn't look like there's a need to mock here

packages/cli-node/src/paths.ts
```diff
@@ -14,7 +14,27 @@
...
 
-import { findPaths } from '@backstage/cli-common';
+import { targetPaths, findOwnPaths } from '@backstage/cli-common';
 
 /* eslint-disable-next-line no-restricted-syntax */
-export const paths = findPaths(__dirname);
+const ownPaths = findOwnPaths(__dirname);
+
+/* eslint-disable-next-line no-restricted-syntax */
+export const paths = {
```
Rugvip: Migrate this fully, as in refactor all usage within `cli-node` to not need this

packages/cli/src/modules/build/commands/repo/start.test.ts
```diff
@@ -16,8 +16,9 @@
 
 import { PackageGraph } from '@backstage/cli-node';
 import { findTargetPackages } from './start';
-import { posix } from 'node:path';
-import { paths } from '../../../../lib/paths';
+import { overrideTargetPaths } from '@backstage/cli-common/testUtils';
+
+overrideTargetPaths({ dir: '/root', rootDir: '/root' });
```
Rugvip: This should be able to use the single-arg variant `overrideTargetPaths('/root')`

packages/cli/src/modules/migrate/commands/versions/bump.test.ts
```diff
@@ -139,6 +138,7 @@ const expectLogsToMatch = (
 
 describe('bump', () => {
   mockDir = createMockDirectory();
+  beforeAll(() => overrideTargetPaths(mockDir.path));
```
Rugvip: There's no need to wrap this in `beforeAll`

packages/cli/src/modules/new/lib/execution/writeTemplateContents.test.ts
```diff
@@ -26,14 +29,14 @@ const baseConfig = {
...
 
 describe('writeTemplateContents', () => {
-  const mockDir = createMockDirectory();
-
   beforeEach(() => {
     mockDir.clear();
+    mockDir.setContent({
+      'package.json': JSON.stringify({
+        workspaces: { packages: ['packages/*', 'plugins/*'] },
+      }),
```
Rugvip, line 37: this doesn't seem to be needed? If it is, use `workspace: [...]` instead

packages/create-app/src/createApp.test.ts
```diff
@@ -19,12 +19,34 @@ import path from 'node:path';
...
+  const MOCK_CREATE_APP_ROOT = '/mock/create-app-root';
+  const mockOwnPaths = {
+    resolve: (...paths: string[]) =>
+      pathModule.join(MOCK_CREATE_APP_ROOT, ...paths),
+    resolveRoot: (...paths: string[]) =>
+      pathModule.join('/mock/monorepo-root', ...paths),
+  };
+  return {
+    ...actual,
+    findPaths: jest.fn(),
```
Rugvip: There should be no usage of `findPaths` that need mocking

packages/yarn-plugin/src/util/getWorkspaceRoot.test.ts
```diff
@@ -37,19 +37,27 @@ describe('getWorkspaceRoot', () => {
...
+        targetPaths: {
+          get dir() {
+            return mockResolveRoot();
+          },
+          get rootDir() {
+            return mockResolveRoot();
+          },
+          resolveRoot: mockResolveRoot,
+        },
       }));
```
Rugvip: Use `overrideTargetPaths` instead

Rugvip (commented):
Some more stuff

Rugvip:
> The idea is that these entire modules are dynamically loaded only when called upon right? So 500 path resolutions don't fire up all at the same time when the cli boots :) since they all are in the file roots.

Yep, with some built-in caching to avoid duplicate work across callers still. The access to the `.dir`, `.resolve` etc. are lazy as well, so no work done at the point of importing `targetPaths`

packages/cli-common/src/paths.ts
```diff
@@ -107,64 +122,209 @@ export function findOwnRootDir(ownDir: string) {
...
+/**
+ * Find paths relative to the package that the calling code lives in.
+ *
+ * Results are cached per package root, and the package root lookup uses a
+ * hierarchical directory cache so that multiple calls from different
+ * subdirectories within the same package share work.
+ *
+ * @public
+ */
+export function findOwnPaths(searchDir: string): OwnPaths {
```
freben, line 294: could this one have taken `meta: ImportMeta` instead?

packages/cli-common/src/paths.ts
```diff
@@ -107,64 +122,209 @@ export function findOwnRootDir(ownDir: string) {
   return resolvePath(ownDir, '../..');
```
freben, line 122: No double checking of what's in there at all? Thinking is it safe to assume these tools never get used for repos with slightly different shapes that ours?

packages/cli-common/src/paths.ts
```diff
@@ -107,64 +122,209 @@ export function findOwnRootDir(ownDir: string) {
...
+    return resolvePath(this.#dir, ...paths);
+  };
+
+  resolveRoot = (...paths: string[]): string => {
+    return resolvePath(this.rootDir, ...paths);
+  };
+}
+
+// Finds the root of a given package
+export function findOwnDir(searchDir: string) {
```
freben, line 203: is this still needed?

packages/cli-common/src/paths.ts
Rugvip: yeah thinking we'll add an option for that once we support it more universally

packages/cli-common/src/paths.ts
Rugvip: Um, yeah hadn't spotted this, let's fix. It's in the old code as well though so will address in a followup

packages/cli-common/src/paths.ts
Rugvip: definitely not x) Will clean this up together with the other fix in a followup

PR #32944, opened 2026-02-21
refactor(cli): Use commander for CLI args instead of yargs

GabDug:

Move `backstage-cli info` & `backstage-cli config:*` arg parsing from `yargs` to the de-facto standard `commander`.

This makes the CLI behavior more consistent (e.g. around --help), which will be more visible once commander help coloring will be enabled (in a subsequent PR).

This does *not* remove `yargs` from Backstage CLI dependencies, as it is required for Jest args parsing (and Jest brings it as transitive dependency anyway).

Both commands have been tested manually for regressions.

### Example

Before
```
backstage-cli config:check --help  
Options:
  --version     Show version number                                    [boolean]
  --package                                                             [string]
  --lax                                                                [boolean]
  --frontend                                                           [boolean]
  --deprecated                                                         [boolean]
  --strict                                                             [boolean]
  --config                                                 [array] [default: []]
  --help        Show help                                              [boolean]
```

After
```
backstage-cli config:check --help  
Usage: program [options]

Validate that the given configuration loads and matches schema

Options:
  --config <path>   Config files to load instead of app-config.yaml (default: [])
  --package <name>  Package to load config from
  --lax             Load config in lax mode
  --frontend        Load config for frontend
  --deprecated      Allow deprecated config keys
  --strict          Strict validation
  -h, --help        display help for command
```

Note: `Usage: program [options]` is another issue for all commander sub-commands.

### DIFF
--- a/packages/cli/src/modules/config/index.ts
+++ b/packages/cli/src/modules/config/index.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import { createCliPlugin } from '../../wiring/factory';
-import { Command } from 'commander';
+import { Command, Option } from 'commander';
 import { lazy } from '../../lib/lazy';
 
 export const configOption = [
@@ -65,10 +65,22 @@ export default createCliPlugin({
         const defaultCommand = command
           .option(...configOption)
           .option('--package <name>', 'Package to load config from')
-          .option('--lax', 'Load config in lax mode')
-          .option('--frontend', 'Load config for frontend')
-          .option('--with-secrets', 'Include secrets in the output')
-          .option('--format <type>', 'Output format')
+          .option('--lax', 'Assumes all environment variables are set')
+          .option(
+            '--frontend',
+            'Print only configuration with frontend visibility',
+          )
+          .addOption(
+            new Option(
+              '--with-secrets',
+              'Include secrets in the output',
+            ).conflicts(['frontend']),
+          )
+          .addOption(
+            new Option('--format <type>', 'Output format')
+              .choices(['json', 'yaml'])
+              .default('yaml'),
+          )
           .description('Print the app configuration for the current package')
           .action(lazy(() => import('./commands/print'), 'default'));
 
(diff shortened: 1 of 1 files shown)


GabDug:
Thanks for the context @Rugvip!

I'd be interested to have your take regarding commander/the current CLI area of improvements!

While I'm not a fan of commander's way of declaring arguments (vs `yargs`/parseArgs` more data-driven declarative approach), I though it was a pretty good fit for the Backstage CLI: it's maintained, simple enough when needed, but can also accommodate more complex/dynamic patterns and be extended properly.

I agree that type-safety is an issue, and that the current efforts via [`@commander-js/extra-typings`](https://github.com/commander-js/extra-typings) lead to rather noisy types, compared to TS-first approaches.

Whichever technical solution is adopted should imo also have a consistent UX between root/modules (e.g. --help, arg parsing) & allow for first-class auto-documentation (current CLI API reports rely on --help parsing, `techdocs-cli` arg docs are manually copied/pasted). It should also enable further improvements, like completions for users, and static flags for interactive commands. 

Needless to say, I'm interested to contribute on these CLI topics, whether it's with commander, yargs, clipanion, parseArgs, more custom built things, or more probably a combination of these.

Side note: [Commander v15](https://github.com/tj/commander.js/issues/2487) in may 2026 will be ESM-only, commander 14 should be maintained until may 2027. [clipanion](https://github.com/arcanis/clipanion) development seems inactive.

Rugvip:
@GabDug yep I think it's worth reevaluating the `clipanion` direction indeed, based on that lack of activity. The top priority for the CLI right now is to complete the modularization that splits the CLI up into separate modules that can be picked, used and replaced independently. The way we've structured that is to not have any specific parser in the module API but let each module pick their own. That said, it'd still also be very good to settle on a recommended direction for how module should implement their parsing.

I'm curious what you'd go for given the following requirements, not only considering the current options:

### Must have

- Must have good TypeScript support with type-safe flag declarations and parsing.
- Good help page generation that works with the Backstage CLI module API

### Nice to have

- Automatic env var support, where flags can be supplied as vars instead
- Both CJS and ESM support
- Clean, intuitive, and slim API with little boilerplate
- No need to wrap the API with Backstage opinionation, works as is
- Works well with lazy loading, for example with separate shared flag declarations and actual implementation in a separate `import(...)`ed file.

Rugvip:
@GabDug I ended up trying out [cleye](https://github.com/privatenumber/cleye), pretty happy with the result. In case you want to have a look: https://github.com/backstage/backstage/pull/33036

PR #32956, opened 2026-02-22
cli-node: move parallel worker utilities from cli

Rugvip:

This moves the parallel worker utilities (`runParallelWorkers`, `runWorkerQueueThreads`, `runWorkerThreads`, `parseParallelismOption`, `getEnvironmentParallelism`) from the CLI internal lib into `@backstage/cli-node` so they can be shared across CLI modules as they become independent packages.

Part of the ongoing CLI module independence work.

### DIFF
--- a/packages/cli-node/report.api.md
+++ b/packages/cli-node/report.api.md
@@ -216,7 +216,9 @@ export function runConcurrentTasks<TItem>(
 // @public
 export function runWorkerQueueThreads<TItem, TResult, TContext>(
   options: WorkerQueueThreadsOptions<TItem, TResult, TContext>,
-): Promise<TResult[]>;
+): Promise<{
+  results: TResult[];
+}>;
 
 // @public
 export type WorkerQueueThreadsOptions<TItem, TResult, TContext> = {
--- a/packages/cli-node/src/concurrency/runWorkerQueueThreads.test.ts
+++ b/packages/cli-node/src/concurrency/runWorkerQueueThreads.test.ts
@@ -21,7 +21,7 @@ describe('runWorkerQueueThreads', () => {
     const sharedData = new SharedArrayBuffer(10);
     const sharedView = new Uint8Array(sharedData);
 
-    const results = await runWorkerQueueThreads({
+    const { results } = await runWorkerQueueThreads({
       context: sharedData,
       items: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
       workerFactory: (data: SharedArrayBuffer) => {
--- a/packages/cli-node/src/concurrency/runWorkerQueueThreads.ts
+++ b/packages/cli-node/src/concurrency/runWorkerQueueThreads.ts
@@ -78,7 +78,7 @@ export type WorkerQueueThreadsOptions<TItem, TResult, TContext> = {
  */
 export async function runWorkerQueueThreads<TItem, TResult, TContext>(
   options: WorkerQueueThreadsOptions<TItem, TResult, TContext>,
-): Promise<TResult[]> {
+): Promise<{ results: TResult[] }> {
   const items = Array.from(options.items);
   const workerFactory = options.workerFactory;
   const workerData = options.context;
@@ -134,7 +134,7 @@ export async function runWorkerQueueThreads<TItem, TResult, TContext>(
       }),
   );
 
-  return results;
+  return { results };
 }
 
 /* istanbul ignore next */
--- a/packages/cli/src/modules/lint/commands/repo/lint.ts
+++ b/packages/cli/src/modules/lint/commands/repo/lint.ts
@@ -105,7 +105,7 @@ export async function command(opts: OptionValues, cmd: Command): Promise<void> {
     }),
   );
 
-  const resultsList = await runWorkerQueueThreads({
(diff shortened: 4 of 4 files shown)


.changeset/cli-node-parallel-helpers.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli-node': minor
```
Rugvip: make it a patch.

Also need a separate changeset for the CLI that simply notes an internal refactor to use these

packages/cli-node/src/concurrency/index.ts
freben, line 1: I know it's petty buuuut it's tempting to want to name this folder and the functions with "concurrent" rather than "parallel". Happy to live with either though.

packages/cli-node/report.api.md
```diff
@@ -200,4 +200,49 @@ export class PackageRoles {
...
+export type ParallelWorkerOptions<TItem> = {
+  parallelismFactor?: number;
+  items: Iterable<TItem>;
+  worker: (item: TItem) => Promise<void>;
+};
+
+// @public
+export function runParallelWorkers<TItem>(
+  options: ParallelWorkerOptions<TItem>,
+): Promise<void[]>;
```
freben: ...why the array :)

Does this one have some particular reason for not being able to return results?

packages/cli-node/report.api.md
```diff
@@ -200,4 +200,49 @@ export class PackageRoles {
...
+
+// @public
+export type WorkerQueueThreadsOptions<TItem, TResult, TData> = {
+  items: Iterable<TItem>;
+  workerFactory: (
+    data: TData,
+  ) =>
+    | ((item: TItem) => Promise<TResult>)
+    | Promise<(item: TItem) => Promise<TResult>>;
+  workerData?: TData;
```
freben: `context?: TContext`? same below

packages/cli-node/report.api.md
```diff
@@ -200,4 +200,49 @@ export class PackageRoles {
...
+  threadCount?: number;
```
freben: `parallelismFactor` above, literal thread naming here and below. Should they all be `concurrency`, `maxConcurrency`, or `concurrencyLimit` instead?

packages/cli-node/src/concurrency/index.ts
Rugvip: Hmm, yeah bit torn. I like "parallel" because it implies that there's more going on than just organizing tasks in a single thread. Strictly speaking it's not pure parallelism but I think I still prefer communicating that intent.

packages/cli-node/src/parallel/parallel.ts
```diff
@@ -20,7 +20,8 @@ import { Worker } from 'node:worker_threads';
 
 const defaultParallelism = Math.ceil(os.cpus().length / 2);
 
-const PARALLEL_ENV_VAR = 'BACKSTAGE_CLI_BUILD_PARALLEL';
+const PARALLEL_ENV_VAR = 'BACKSTAGE_CLI_PARALLELISM';
```
Rugvip: make this `'BACKSTAGE_CLI_CONCURRENCY'` instead, given that this isn't exactly the number of active threads

packages/cli-node/report.api.md
Rugvip: yep let's do that instead. Worker data is a thing but here's it's already mapped over to a different context.

packages/cli-node/report.api.md
```diff
@@ -200,4 +200,49 @@ export class PackageRoles {
...
+  options: ParallelWorkerOptions<TItem>,
+): Promise<void[]>;
+
+// @public
+export function runWorkerQueueThreads<TItem, TResult, TData>(
+  options: WorkerQueueThreadsOptions<TItem, TResult, TData>,
+): Promise<TResult[]>;
+
+// @public
+export function runWorkerThreads<TResult, TData, TMessage>(
```
Rugvip: Actually let's get rid of this one, it doesn't seem to be used

packages/cli-node/report.api.md
Rugvip: Yeah this isn't really parallel either. To some extent this is just a less fancy `p-limit`. I tihnk we could call this one `runConcurrentTasks`. And let's just have it return `void` for now

packages/cli-node/report.api.md
```diff
@@ -200,4 +200,49 @@ export class PackageRoles {
   static getRoleFromPackage(pkgJson: unknown): PackageRole | undefined;
   static getRoleInfo(role: string): PackageRoleInfo;
 }
+
+// @public
+export type ParallelWorkerOptions<TItem> = {
+  parallelismFactor?: number;
```
Rugvip: `concurrencyFactor` on this one instead

packages/cli-node/report.api.md
Rugvip: They're are a bit different. Actually let's just drop this option for now

packages/cli-node/src/concurrency/index.ts
Rugvip: Switched, since a fair chunk of this is just concurrency and the main threaded one has it in the name

freben (approved):

PR #32965, opened 2026-02-22
catalog: slightly simplify deferred stitching runs

freben:
This is on top of https://github.com/backstage/backstage/pull/32854 .

Review while ignoring whitespace.

This extra insert of the stitch ticket only makes sense in immediate mode, which is going away. There it may happen that stitch calls happen without passing in a ticket at all.

In deferred mode the very act of marking something for deferred stitching already inserts the stitch ticket, which then also always gets passed in explicitly to `performStitching`. And doing it again inside the stitch performer code just leads to more risk for race conditions.

### DIFF
--- a/plugins/catalog-backend/src/database/operations/stitcher/markForStitching.ts
+++ b/plugins/catalog-backend/src/database/operations/stitcher/markForStitching.ts
@@ -56,6 +56,7 @@ export async function markForStitching(options: {
   strategy: StitchingStrategy;
   entityRefs?: Iterable<string>;
   entityIds?: Iterable<string>;
+  stitchTicket?: string;
 }): Promise<void> {
   const entityRefs = sortSplit(options.entityRefs);
   const entityIds = sortSplit(options.entityIds);
@@ -106,7 +107,7 @@ export async function markForStitching(options: {
   } else if (mode === 'deferred') {
     // It's OK that this is shared across final_entities rows; it just needs to
     // be uniquely generated for every new stitch request.
-    const ticket = uuid();
+    const ticket = options.stitchTicket ?? uuid();
 
     // Use a single-pass upsert: look up entity info from refresh_state, then
     // insert into final_entities with ON CONFLICT merge. This handles both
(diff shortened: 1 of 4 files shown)


plugins/catalog-backend/src/database/operations/stitcher/markForStitching.ts
```diff
@@ -56,6 +56,7 @@ export async function markForStitching(options: {
   strategy: StitchingStrategy;
   entityRefs?: Iterable<string>;
   entityIds?: Iterable<string>;
+  stitchTicket?: string;
```
freben, line 59: Only added for testability

plugins/catalog-backend/src/database/operations/stitcher/performStitching.test.ts
```diff
@@ -82,15 +90,18 @@ describe('performStitching', () => {
         },
       ]);
 
+      await markForStitching({
+        knex,
+        strategy,
+        entityRefs: ['k:ns/n'],
+        stitchTicket: 'first-ticket',
+      });
```
freben, line 98: This, as a happy path for deferred stitching, should always have these mark calls in front of the actual stitching, because that's what the flow of events always is anyway.

plugins/catalog-backend/src/database/operations/stitcher/performStitching.ts
```diff
@@ -64,38 +64,40 @@ export async function performStitching(options: {
...
-        .merge(['stitch_ticket']);
-    } catch (error) {
-      // It's possible to hit a race where a refresh_state table delete + insert
-      // is done just after we read the entity_id from it. This conflict is safe
-      // to ignore because the current stitching operation will be triggered by
-      // the old entry, and the new entry will trigger it's own stitching that
-      // will update the entity.
-      if (isDatabaseConflictError(error)) {
-        logger.debug(`Skipping stitching of ${entityRef}, conflict`, error);
+    if (options.strategy.mode !== 'deferred') {
```
freben, line 71: This is the actual interesting part of the change.

PR #32967, opened 2026-02-23
fix(cli): add config schema validation to repo fix command

### LINKED ISSUE #32951: Linting for Config Schema
### 📜 Issue Labels

### 🔎 Search Terms

```plain
lint, check, config, schema, config.d.ts, package.json
```

### 🗃️ Project Area

CLI Tooling

### 🔖 Need

Backstage offers a great configuration system and using it in plugins is very straightforward but one area that you can easily miss and spend time struggling with is properly adding the schema file to the `package.json` file as described in the documentation: https://backstage.io/docs/conf/defining#schema-collection-and-definition.

Often we see plugins what have it missing entirely, or they added it to just the `files` array or just add the `configSchema` property. 

### 📝 Proposal

To help avoid the issues outline above we should introduce a linting check or some other mechanism that will validate this. If a `config.d.ts` file exists then you should have the needed items in the `package.json`, if some is missing it should fail. If you have some in your package.json but no `config.d.ts` file it should also fail.

### 🔄 Alternatives

We continue to rely on the docs but as noted this still comes up as an issue.

bhavyamsharmaa:
## What this PR does

This adds a config schema validation step to the ackstage-cli repo fix command.

When a package contains a config.d.ts file, the fixer ensures that:
- The configSchema property in package.json points to config.d.ts
- The config.d.ts file is included in the iles array

If the configSchema is an inline JSON object (not a file path), validation is skipped since file-level checks don't apply.

This helps prevent common mistakes where plugin authors forget to properly reference their config schema definitions in package.json, which can lead to missing configuration at runtime.

### Changes

- **New**: ixConfigSchema.ts - Fixer function that auto-corrects config schema references
- **New**: ixConfigSchema.test.ts - Unit tests covering all validation scenarios
- **Modified**: ix.ts - Added ixConfigSchema to the list of fixers

### Why 
epo fix instead of lint

Per feedback from @Rugvip, this is better suited as part of the existing 
epo fix infrastructure rather than a standalone lint check, since:
- It executes very fast
- It can auto-fix issues rather than just reporting them
- It aligns with the plan to move non-ESLint checks to unified tooling

Fixes #32951

---

- [x] I have read the [Contributing](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md) document
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have added a changeset describing my changes

### DIFF
--- a/.changeset/fix-config-schema-validation.md
+++ b/.changeset/fix-config-schema-validation.md
@@ -1,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+Added config schema validation to `repo fix` command. When a package contains a `config.d.ts` file, the fixer ensures that `configSchema` in `package.json` references it and that it is included in the `files` array.
--- a/packages/cli/src/modules/maintenance/commands/repo/fix.ts
+++ b/packages/cli/src/modules/maintenance/commands/repo/fix.ts
@@ -31,6 +31,7 @@ import {
 } from 'node:path';
 import { paths } from '../../../../lib/paths';
 import { publishPreflightCheck } from '../../lib/publishing';
+import { fixConfigSchema } from './fixConfigSchema';
 
 const SCRIPT_EXTS = ['.js', '.jsx', '.ts', '.tsx', '.json'];
 
@@ -497,7 +498,11 @@ export async function command(opts: OptionValues): Promise<void> {
   const packages = await readFixablePackages();
   const fixRepositoryField = createRepositoryFieldFixer();
 
-  const fixers: PackageFixer[] = [fixPackageExports, fixSideEffects];
+  const fixers: PackageFixer[] = [
+    fixPackageExports,
+    fixSideEffects,
+    fixConfigSchema,
+  ];
 
   // Fixers that only apply to repos that publish packages
   if (opts.publish) {
--- a/packages/cli/src/modules/maintenance/commands/repo/fixConfigSchema.test.ts
+++ b/packages/cli/src/modules/maintenance/commands/repo/fixConfigSchema.test.ts
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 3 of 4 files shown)


bhavyamsharmaa:
Address review feedback from Copilot for schema lint validation. The logic has been factored into a shared helper \lintConfigSchema.ts\ and is unconditionally invoked by both \package lint\ and \
epo lint\. Any detected schema errors are now converted to ESLint messages, so they are cleanly formatted and picked up by the \--output-file\ json generator.

packages/cli/src/modules/lint/commands/repo/lint.ts
```diff
@@ -43,9 +43,9 @@ export async function command(opts: OptionValues, cmd: Command): Promise<void> {
   const cache = new SuccessCache('lint', opts.successCacheDir);
   const cacheContext = opts.successCache
     ? {
-        entries: await cache.read(),
-        lockfile: await Lockfile.load(paths.resolveTargetRoot('yarn.lock')),
-      }
+      entries: await cache.read(),
+      lockfile: await Lockfile.load(paths.resolveTargetRoot('yarn.lock')),
+    }
```
awanlin: Why to formatting changes here?

packages/cli/src/modules/lint/commands/package/lint.ts
```diff
@@ -33,11 +34,33 @@ export default async (directories: string[], opts: OptionValues) => {
...
+      suppressedMessages: [],
+    });
+  }
+
   const failed =
     results.some(r => r.errorCount > 0) ||
     (!ignoreWarnings &&
       results.reduce((current, next) => current + next.warningCount, 0) >
-        maxWarnings);
+      maxWarnings);
```
awanlin: Odd formatting change for no reason here?

awanlin (changes requested):
Thanks for the contribution @bhavyamsharmaa, can you rebase so we can get the CI running again, please? Left a few comments and would like to see those from copilot address. Also, it would be beneficial to add tests so we can validate these changes.

Finally, please keep the PR checklist and avoid removing it in your PRs, please.

Rugvip (changes requested):
Sorry I think we should hold off a bit on this as we want to explore moving to a more unified lint tooling for these kind of non-ESLint based checks.

In the meantime though, if there's any check here that executes very fast and won't have any false-positives we can add those to the existing `fix` command.

bhavyamsharmaa:
Hi @Rugvip @awanlin, thanks for the feedback! I've reworked this PR based on your suggestions:

- **Moved logic to 
epo fix**: Instead of adding checks to the lint command, the config schema validation is now a fixer in ackstage-cli repo fix that auto-corrects configSchema and iles entries in package.json.
- **Handles inline schemas**: If configSchema is an object (inline JSON schema), validation is skipped since file-level checks don't apply.
- **Added tests**: Added unit tests covering all validation scenarios (missing config, correct config, inline schema, wrong path, missing files entry).
- **Removed formatting changes**: Reverted all unrelated changes to lint command files.
- **Added changeset and PR checklist**.
- **Rebased on latest master**.

PR #32975, opened 2026-02-23
chore: remove applied patch files

Rugvip:
Removes the pending patch files now that they've all been applied in https://github.com/backstage/backstage/pull/32974.


freben (changes requested):
want to get in wait for https://github.com/backstage/backstage/pull/32973 first

freben (approved):

PR #32977, opened 2026-02-24
`feat(metrics)`: Implement `MetricsService`

benjdlambert:
This is a rebase of https://github.com/backstage/backstage/pull/31793 and fixing some of the code review comments from the Framework SIG. Didn't wanna just push this on top of the PR, but think that this superceeds that work.

With many thanks to @kurtaking!


packages/backend-plugin-api/src/alpha/MetricsService.ts
```diff
@@ -0,0 +1,271 @@
...
+   */
+  advice?: MetricAdvice;
+}
+
+/**
+ * A counter metric that only supports non-negative increments.
+ *
+ * @alpha
+ */
+export interface Counter<
```
freben: Hm. The plugin-api namespace is flat and becomes MASSIVE, mingling all kinds of concerns side by side. We've been pretty careful to have quite explicit symbol names for that reason.

Do we have to make these names more explicit?

`CounterMetric`

`CounterMetricInstrument` (arguably the [full name](https://opentelemetry.io/docs/concepts/signals/metrics/#metric-instruments))

We have sometimes used prefixing to make things grouped nicely,

`MetricsServiceCounter`

`MetricsServiceCounterMetric`

packages/backend-defaults/src/alpha/entrypoints/metrics/metricsServiceFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+export const metricsServiceFactory = createServiceFactory({
+  service: metricsServiceRef,
+  deps: {
+    config: coreServices.rootConfig,
+    pluginMetadata: coreServices.pluginMetadata,
+  },
+  factory: ({ config, pluginMetadata }) => {
+    const pluginId = pluginMetadata.getId();
+
+    const pluginConfig = config.getOptionalConfig(`${pluginId}.metrics`);
```
freben: Should probably be `backend.metrics.plugin.${pluginId}` so that it lies next to `backend.database`, `backend.cache` etc like other backend services and can have room for defaults directly under `backend.metrics`?

Not sure if this was ever meant to be spanning frontend and backend, which might change the picture a bit ofc.

Also, this needs a `config.d.ts` change

packages/backend-defaults/src/alpha/entrypoints/metrics/metricsServiceFactory.test.ts
```diff
@@ -0,0 +1,95 @@
...
+          },
+        }),
+        metricsServiceFactory,
+      ],
+    }).getSubject('my-plugin');
+
+    expect(subject).toBeDefined();
+    expect(typeof subject.createCounter).toBe('function');
+    expect(typeof subject.createHistogram).toBe('function');
+    expect(typeof subject.createObservableGauge).toBe('function');
```
freben: the assertions don't do much throughout this file :)

It might make sense to mock DefaultMetricsService and assert that its constructor was called with the right values from config

packages/backend-test-utils/src/alpha/services/MetricsServiceMock.ts
```diff
@@ -0,0 +1,59 @@
...
+import {
+  MetricsService,
+  metricsServiceRef,
+} from '@backstage/backend-plugin-api/alpha';
+import { metricsServiceFactory } from '@backstage/backend-defaults/alpha';
+
+/**
+ * @alpha
+ */
+export namespace metricsServiceMock {
```
freben, line 27: Eventually we want to help people make assertions on metrics being produced. So there may need to be a `metricsServiceMock()` where you can like `expect(s.getCounter('name').add).toHaveBeenCalledWith` or something like that.

But that can be done later down the line

plugins/catalog-backend/src/database/metrics.ts
```diff
@@ -69,7 +69,7 @@ export function initDatabaseMetrics(knex: Knex) {
         this.set(Number(total[0].count));
       },
     }),
-    entities_count: meter
+    entities_count: metrics
       .createObservableGauge('catalog_entities_count', {
```
freben, line 73: Just double checking - we explicitly arrived at NOT performing any automatic metric name munging (adding plugin id as the prefix etc) right

packages/backend-defaults/src/alpha/entrypoints/metrics/DefaultMetricsService.ts
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * This implementation provides a thin wrapper around the OpenTelemetry Meter API.
+ *
+ * @alpha
+ */
+export class DefaultMetricsService implements MetricsService {
+  private readonly meter: Meter;
+
+  private constructor(opts: DefaultMetricsServiceOptions) {
+    this.meter = metrics.getMeter(opts.name, opts.version, {
```
freben, line 54: Does giving these meters a name change anything about the exported metrics at all?

packages/backend-defaults/src/alpha/entrypoints/metrics/metricsServiceFactory.ts
Rugvip: yep, +1 to having this live in one place

.changeset/rare-adults-attack.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/backend-plugin-api': patch
+'@backstage/backend-defaults': patch
+---
+
+Adds an alpha `MetricsService` with minimal implementation to support 1:1 migration of existing catalog metrics. We do not recommend using the service for any other purpose at this time.
```
Rugvip: Hmm, this seems a bit strange tbh, why not?

packages/backend-plugin-api/src/alpha/MetricsService.ts
Rugvip: +1, let's namespace these a bit. I'd lean towards `MetricsServiceCounter` etc.

packages/backend-plugin-api/src/alpha/refs.ts
```diff
@@ -56,3 +56,14 @@ export const rootSystemMetadataServiceRef = createServiceRef<
...
+
+/**
+ * Service for managing metrics.
+ *
+ * @alpha
+ */
+export const metricsServiceRef = createServiceRef<
+  import('./MetricsService').MetricsService
+>({
+  id: 'alpha.core.metrics',
```
Rugvip, line 68: Should be flipped around right? iirc we do `.alpha` at the end

plugins/catalog-backend/src/database/metrics.ts
Rugvip: Feels like a lil bit of a foot-gun, but I also see why we need it. Potential idea would be something like `'plugin:entities_count'` vs `'global:mcp_...'` or such, or in object form. In the end it's probably best to keep like this but document very clearly that plugins are expected to prefix with their own ID unless they're producing some established metric structure

packages/backend-plugin-api/src/alpha/refs.ts
benjdlambert: Hmm, not according to the existing ones today.

`alpha.actions` `alpha.actionsRegistry` and `alpha.pluginMetadata`

.changeset/rare-adults-attack.md
benjdlambert: Yeah, I can tweak.

plugins/catalog-backend/src/database/metrics.ts
benjdlambert: Chatted a bit on discord about this and thinking that we leave it be for now.

Maybe what we do is add the `backstagePluginId={pluginId}` attribute to all metrics, and then recommend people drop the plugin namespacing if they're like this.

`entities_count` would become the name with `backstagePluginId=catalog` then.

benjdlambert: That way we can ship this as non breaking to existing metrics, and then make one bigger change later right?

benjdlambert: I think that's what the intention behind `scopeName` is but not sure if that shows up in the otel report. Need to confirm.

freben (approved):

packages/backend-plugin-api/src/alpha/refs.ts
Rugvip: Well shit, I guess in that case it doesn't matter :p

plugins/catalog-backend/src/database/metrics.ts
kurtaking: @benjdlambert is right here, we should be relying on `scopeName` to uniquely identify the plugin the metrics are coming from. 

Really, we DO NOT want to perform any messing with the metric name. This was the original proposal but conflicts with OpenTelemetry semantic conventions. 

For example, [otel recommends](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#metric-mcpserveroperationduration) using `mcp.server.operation.duration` in the spec. If we prepend `backstage.plugin...` or whatever, then we are immediately going against the standard  the spec.  Instead each plugin should receive a metrics service that is  configured to identify the plugin via the Instrumentation Scope. The scope name follows the pattern `backstage-plugin-{pluginId}`.

packages/backend-defaults/src/alpha/entrypoints/metrics/DefaultMetricsService.ts
kurtaking: Consolidate any future talk to https://github.com/backstage/backstage/pull/32977/changes#r2851154301.

kurtaking: tying up some loose ends - Yes.

The metrics will now be associated with the plugin they came from via the OTEL [instrumentation scope](https://opentelemetry.io/docs/concepts/instrumentation-scope/).

Image

PR #32979, opened 2026-02-24
feat(ui): replace --bui-bg-popover with layered bg approach

Naycon:

Remove the `--bui-bg-popover` CSS token and adopt a two-layer
background pattern across overlay components: outer shell uses
`--bui-bg-app`, inner content uses `Box bg="neutral-1"`. 

This fixes issues where content that is bg-aware would render 
incorrectly based on a previous parents bg instead of what the 
popover is using. 

**Components updated:** Popover, Tooltip, Menu, Dialog

Each component gets a local border-radius CSS variable, a `Box`
content wrapper with `neutral-1` background, and updated arrow
fills where applicable.

Story "is open" variants now use a dot-grid background pattern
to detect background transparency issues in Chromatic snapshots.

##### Before

image

##### After

image

Notice how the `Cancel` button now being aware it's rendered on a `neutral-1` bg.

### DIFF
--- a/packages/ui/src/components/Dialog/Dialog.module.css
+++ b/packages/ui/src/components/Dialog/Dialog.module.css
@@ -43,8 +43,10 @@
   }
 
   .bui-Dialog {
-    background: var(--bui-bg-popover);
-    border-radius: 0.5rem;
+    --dialog-border-radius: 0.5rem;
+    background: var(--bui-bg-app);
+    box-shadow: var(--bui-shadow);
+    border-radius: var(--dialog-border-radius);
     border: 1px solid var(--bui-border-1);
     color: var(--bui-fg-primary);
     position: relative;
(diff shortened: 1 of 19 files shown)


packages/ui/src/components/Dialog/Dialog.module.css
```diff
@@ -43,8 +43,9 @@
   }
 
   .bui-Dialog {
```
Naycon, line 45: Just to add, there's an alternative solution we could take if we want to avoid introducing internal `Box` elements for these components. We would then make all the popover-like elements bg providers, and then via css put a pseudo element with the app bg under each of them. 

I think I prefer this solution, it has less moving parts and is pretty clear to me, but I'm open to changing it if someone feels differently.

cdedreuille: I'm with you on this one.

PR #32986, opened 2026-02-24
run vale directly instead of through the action

freben:
We keep running into the vale limit of not being able to lint more than 300 files, when making huge PRs or releases.

Well, this is what claude suggested to fix it, since the limit is inherent to the inability for the GitHub API to return more than 300 files with the implementation that the action uses internally.

Is it ... is it too wild?

### DIFF
--- a/.github/workflows/verify_docs-quality.yml
+++ b/.github/workflows/verify_docs-quality.yml
@@ -24,20 +24,20 @@ jobs:
       - name: Install vale
         run: |
           VALE_VERSION="3.7.1"
-          VALE_REPO="errata-ai/vale"
+          VALE_CHECKSUM="ba4924bf2c5884499f09b02a6eb3318b29df40a3e81701c0804b9b1aefcd9483"
           VALE_DIST_DIR="/tmp/vale-dist"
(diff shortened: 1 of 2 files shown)


.github/workflows/verify_docs-quality.yml
```diff
@@ -18,18 +21,34 @@ jobs:
...
-        uses: errata-ai/vale-action@d89dee975228ae261d22c15adcd03578634d429c # v2.1.1
-        with:
-          # This also contains --config=.github/vale/config.ini ... :/
-          files: '${{ steps.generate.outputs.args }}'
-          version: latest
+          mkdir -p "$VALE_DIST_DIR"
+
+          # Download pinned Vale binary and corresponding checksums file
+          gh release download "v${VALE_VERSION}" --repo "$VALE_REPO" --pattern "vale_${VALE_VERSION}_Linux_64-bit.tar.gz" --dir "$VALE_DIST_DIR"
+          gh release download "v${VALE_VERSION}" --repo "$VALE_REPO" --pattern "vale_${VALE_VERSION}_checksums.txt" --dir "$VALE_DIST_DIR"
```
Rugvip: If we want maximum value out of this check it'd be better to inline the checksum(s) and bump them along with the vale version

Issue #32980, opened 2026-02-24
Update `backstage-cli repo list-deprecations` to only list deprecations from imported code


awanlin:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
backstage-cli, repo, list-deprecations, deprecate, command, imported, plugin
```

### 🗃️ Project Area

CLI Tooling

### 🔖 Need

In the community plugins repo you can opt-in to having the `yarn backstage-cli repo list-deprecations` ran as part of the CI to help catch usages of deprecated code making it easier to handle those changes early before they become braking changes. Where we are running into an issue is if you deprecate some of your own plugin code the `backstage-cli repo list-deprecations` command is picking that up and failing the CI. In this case we aren't actually consuming deprecated code but notifying people that code has been deprecated. 

More details here: https://github.com/backstage/community-plugins/issues/7496

### 📝 Proposal

It would be great if the `backstage-cli repo list-deprecations` command only looked at imported code or had a flag to handle this in some way.

### 🔄 Alternatives

Right now Plugin Owners are opting out of having `backstage-cli repo list-deprecations` ran in the CI, this does mean that when code is deprecated they need to manually keep that in account.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

No, but I'm happy to collaborate on a PR with someone else

mehrdadbn9:
Hi, I'd like to work on updating backstage-cli repo list-deprecations command to output YAML. The current output format is not machine-friendly. Could a maintainer please assign it to me? Thank you!

awanlin:
Hi @mehrdadbn9, that's not what this issue is asking for and I believe someone already has a PR active for this: https://github.com/backstage/backstage/pull/32988. If you want to log an issue regarding the output feel free 👍

PR #32988, opened 2026-02-24
Update backstage-cli repo list-deprecations to ignore local code

bhavyamsharmaa:
This pull request updates the backstage-cli repo list-deprecations command to ignore deprecation warnings that originate from code within the same package.

### DIFF
--- a/packages/cli/src/modules/maintenance/commands/repo/list-deprecations.ts
+++ b/packages/cli/src/modules/maintenance/commands/repo/list-deprecations.ts
@@ -20,7 +20,7 @@ import { OptionValues } from 'commander';
 import { relative as relativePath } from 'node:path';
 import { PackageGraph } from '@backstage/cli-node';
 import { targetPaths } from '@backstage/cli-common';
-import * as ts from 'typescript';
+import { ts } from 'ts-morph';
 
 // We import the deprecation plugin rule since we need to inject our custom context.report
 const deprecationPlugin = require('eslint-plugin-deprecation');
@@ -55,7 +55,7 @@ export async function command(opts: OptionValues) {
         create(context: any) {
           const customContext = Object.create(context);
           Object.defineProperty(customContext, 'report', {
-            value: function (descriptor: any) {
+            value: function report(descriptor: any) {
               const services = context.parserServices;
               if (services?.program) {
                 const tc = services.program.getTypeChecker();


.changeset/curvy-moments-lose.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+Updated the `repo list-deprecations` command to ignore deprecation warnings orginating from the local package, focusing solely on those stemming from external dependencies.
```
awanlin, line 5:
```suggestion
Updated the `repo list-deprecations` command to ignore deprecation warnings originating from the local package, focusing solely on those stemming from external dependencies.
```

That should address the Vale error.

packages/cli/src/modules/maintenance/commands/repo/list-deprecations.ts
```diff
@@ -20,21 +20,107 @@ import { OptionValues } from 'commander';
...
+
+// We import the deprecation plugin rule since we need to inject our custom context.report
+const deprecationPlugin = require('eslint-plugin-deprecation');
 
 export async function command(opts: OptionValues) {
   const packages = await PackageGraph.listTargetPackages();
 
+  // Create lookup for quickly finding which package a file belongs to
+  // Sort by length to match the most specific package dir first
+  const sortedPackages = [...packages].sort(
```
awanlin, line 33: This seems like a pretty big code addition for this, is there truly no other approach for this? Have you looked at https://typescript-eslint.io/rules/no-deprecated/? Maybe that would have what we need built in.

awanlin (changes requested):
Thanks for starting on this @bhavyamsharmaa, left a pair of comments.

Also, please use the PR checklist instead of removing it. It's handy for us reviewers to see. 🙏

Rugvip (changes requested):
Hmm, could you elaborate a bit more on what issue this solves? 🤔 At an initial glance it seems to me like it's actually desirable to include those deprecations in the listing.

awanlin:
> Hmm, could you elaborate a bit more on what issue this solves? 🤔 At an initial glance it seems to me like it's actually desirable to include those deprecations in the listing.

For context this issue will help: https://github.com/backstage/backstage/issues/32980 and then this upstream issue: https://github.com/backstage/community-plugins/issues/7496

PR #32997, opened 2026-02-25
feat(catalog-backend-module-github): prefer verified org emails in default user transformer

djamaile:

The `defaultUserTransformer` now prefers `organizationVerifiedDomainEmails` over the user's public GitHub email when populating the user entity profile.

### Problem

The GitHub org catalog module fetches `organizationVerifiedDomainEmails` via GraphQL for every user but the `defaultUserTransformer` ignores it entirely — it only uses `item.email` (the user's public GitHub email). This means:

- With GitHub App auth, `email` isn't even requested (conditional `@include`), so users get **no email**
- Many users have private emails, so `email` comes back empty even with PAT auth
- The verified org emails (corporate emails) are fetched but thrown away

### Solution

- Prefer `organizationVerifiedDomainEmails[0]` over `item.email` when available
- Strip plus-addressed routing tags that GitHub adds to verified domain emails (e.g. `user+abc123@example.com` → `user@example.com`)
- Fall back to `item.email` when verified domain emails are empty or undefined

- [ ] All your commits have a `Signed-off-by` line in the message.

---
Generated with Claude Code

### DIFF
--- a/.changeset/github-org-verified-emails.md
+++ b/.changeset/github-org-verified-emails.md
@@ -1,0 +1,34 @@
+---
+'@backstage/plugin-catalog-backend-module-github': minor
+---
+
+The default user transformer now prefers organization verified domain emails over the user's public GitHub email when populating the user entity profile. It also strips plus-addressed routing tags that GitHub adds to these emails.
+
+If you want to retain the old behavior, you can do so with a custom user transformer using the `githubOrgEntityProviderTransformsExtensionPoint`:
+
+```ts
+import { createBackendModule } from '@backstage/backend-plugin-api';
+import { githubOrgEntityProviderTransformsExtensionPoint } from '@backstage/plugin-catalog-backend-module-github-org';
+import { defaultUserTransformer } from '@backstage/plugin-catalog-backend-module-github';
+
+export default createBackendModule({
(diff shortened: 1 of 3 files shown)


.changeset/github-org-verified-emails.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-github': patch
```
freben:
```suggestion
'@backstage/plugin-catalog-backend-module-github': minor
```

While it can be argued that this also fixes bugs, it can be a breaking change for some legitimate users.

.changeset/github-org-verified-emails.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+The default user transformer now prefers organization verified domain emails over the user's public GitHub email when populating the user entity profile.
```
freben: Would it be possible to add a section on "If you want to retain the old behavior, you can do so with a custom user transformer:" and a short snippet of code that inserts such a small transformer, using the `githubOrgEntityProviderTransformsExtensionPoint`. It would probably call `defaultUserTransformer` and then just mutate the email.

PR #33002, opened 2026-02-25
feat(ui): neutral bg system overhaul — single incremental value + overlay reset

cdedreuille:
Two related changes that together complete the neutral background system redesign:

1. **Simplify the public `bg` prop API** — replace explicit `neutral-1/2/3` and `neutral-auto` with a single `'neutral'` value that always auto-increments from the parent context.
2. **Fix neutral bg inheritance in overlay components** — add an internal `BgReset` context cut that prevents popovers, tooltips, dialogs, and menus from inheriting the neutral level of their trigger's surroundings.

### Problem

The old API allowed consumers to pin a container to an explicit neutral level (`bg="neutral-1"`, `bg="neutral-2"`, `bg="neutral-3"`). This made it possible to nest a `neutral-1` box inside a `neutral-3` surface, breaking the visual hierarchy. The system was designed to be strictly incremental, but the types didn't enforce it.

### Change

The public-facing `ProviderBg` type is replaced:

**Before:**
```ts
type ProviderBg = 'neutral-1' | 'neutral-2' | 'neutral-3' | 'neutral-auto' | 'danger' | 'warning' | 'success';
```

**After:**
```ts
type ProviderBg = 'neutral' | 'danger' | 'warning' | 'success';
```

`'neutral'` always auto-increments from the parent context (capping at `neutral-3`). It is no longer possible to skip levels or go backwards.

`ContainerBg` (`neutral-1 | neutral-2 | neutral-3 | ...`) remains unchanged as an `@internal` type — it is still used inside `BgContext` and applied as `data-bg` on DOM elements for CSS.

### Problem

React portals escape the DOM tree, but not the React context tree. Overlay components (Popover, Tooltip, Dialog, Menu) inherit `BgContext` from wherever their trigger is placed in the component tree. A trigger inside a `Box bg="neutral"` (neutral-1) means the overlay's `Box bg="neutral"` sees neutral-1 as its parent and resolves to neutral-2 — visually incorrect, since the overlay always appears at the viewport level. Because of this, Tooltip, Dialog, and Menu were using the now-removed `bg="neutral-1"` explicit value to force a reset for their content, which after part 1 is no longer a viable solution.

### Change

Added an internal `BgReset` component to `useBg.tsx`. It injects `{ bg: undefined }` directly into `BgContext`, cutting the inherited neutral chain. Any `bg="neutral"` inside it restarts from level 1.

```tsx
/** @internal */
export const BgReset = ({ children }: { children: ReactNode }) => (
  <BgContext.Provider value={createVersionedValueMap({ 1: { bg: undefined } })}>
    {children}
  </BgContext.Provider>
);
```

`BgReset` is exported from `useBg.tsx` for use by overlay components, but is **not** re-exported from the package's public `index.ts` — it is an internal implementation detail.

### DIFF
--- a/docs-ui/src/app/snippets.ts
+++ b/docs-ui/src/app/snippets.ts
@@ -1,19 +1,19 @@
 export const surfacesSnippet = `<Flex direction="column" gap="4">
-  <Box bg="neutral-1">
+  <Box bg="neutral">
     <Button variant="secondary">Hello World</Button>
   </Box>
-  <Box bg="neutral-1">
+  <Box bg="neutral">
     <Button variant="secondary">Hello World</Button>
   </Box>
 </Flex>`;
 
-export const adaptiveSnippet = `<Box bg="neutral-1">
+export const adaptiveSnippet = `<Box bg="neutral">
   <Card> {/* automatically set background to neutral-2 */}
     <Button variant="secondary">Button with background set to neutral-3</Button>
   </Card>
 </Box>`;
 
-export const customCardSnippet = `<Box bg="autoIncrement">
+export const customCardSnippet = `<Box bg="neutral">
   <Text>Hello World</Text>
 </Box>`;
 
--- a/packages/ui/report.api.md
+++ b/packages/ui/report.api.md
@@ -1012,14 +1012,7 @@ export const FlexDefinition: {
     'direction',
   ];
   readonly dataAttributes: {
-    readonly bg: readonly [
-      'neutral-1',
-      'neutral-2',
-      'neutral-3',
-      'danger',
-      'warning',
-      'success',
-    ];
+    readonly bg: readonly ['neutral', 'danger', 'warning', 'success'];
   };
 };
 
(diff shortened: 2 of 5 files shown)


cdedreuille:
@Naycon I did change some snippets to match the new `bg` prop but other than that I think we're good. This change is mostly done in the background.

Rugvip (approved):
I wonder if we should keep `neutral-x` around as well? Both because it avoids breaking the existing API, but also because it would allow for explicit selection if you're for example creating a Portal/hovering element that you want to introduce at a certain height. I do see the risk as well though so maybe that'd be more of a separate explicit utility.

cdedreuille:
@Rugvip We did add a `resetBgLevel` to make sure you come back to `ResetBg` in case you need to create popovers. Hopefully this is enough. My suggestion would be that if people want to build custom experiences they would use CSS tokens directly in that case. You can use `--bui-bg-neutral-1` directly.

PR #33007, opened 2026-02-25
cli: split loadCliConfig into separate build and config implementations

Rugvip:

This splits the shared `loadCliConfig` function in the CLI into two separate implementations — one for the build module and one for the config module — to remove the cross-module dependency from `build → config`.

The two use cases have different needs:
- **Config module** — one-shot config loading. Loads config once, processes/validates it, done. The watch/streaming loop is removed since no config commands need it.
- **Build module** — continuous config watching. The dev server needs live config updates when files change, so this version keeps the full watch/streaming capability.

### DIFF
--- a/packages/cli/src/modules/build/lib/config.ts
+++ b/packages/cli/src/modules/build/lib/config.ts
@@ -26,16 +26,13 @@ type Options = {
   args: string[];
   targetDir?: string;
   fromPackage?: string;
-  mockEnv?: boolean;
-  withDeprecatedKeys?: boolean;
-  fullVisibility?: boolean;
-  strict?: boolean;
+  withFilteredKeys?: boolean;
+  watch?: (newFrontendAppConfigs: AppConfig[]) => void;
 };
 
 export async function loadCliConfig(options: Options) {
   const targetDir = options.targetDir ?? targetPaths.dir;
 
-  // Consider all packages in the monorepo when loading in config
   const { packages } = await getPackages(targetDir);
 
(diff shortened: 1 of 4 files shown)


packages/cli/src/modules/config/lib/config.ts
```diff
@@ -73,48 +71,29 @@ export async function loadCliConfig(options: Options) {
...
-            resolve(configs);
-            loaded = true;
-
-            if (!options.watch) {
-              abortController.abort();
-            }
-          }
+          resolve(configs);
+          loaded = true;
+          abortController.abort();
```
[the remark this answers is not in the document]
freben: Agreed - or do you want to exhaust the iterable? If so, maybe add a comment explaining why there is no break / return

packages/cli/src/modules/build/lib/config.ts
```diff
@@ -0,0 +1,128 @@
...
+              withFilteredKeys: options.withFilteredKeys,
+              ignoreSchemaErrors: true,
+            });
+            options.watch?.(newFrontendAppConfigs);
+          } else {
+            resolve(configs);
+            loaded = true;
+
+            if (!options.watch) {
+              abortController.abort();
```
freben, line 91: so i guess same here, should there be a return

Rugvip: yeah probably. All in the existing code so makes sense to do separately either way. @copilot can you open a PR to fix?

PR #33013, opened 2026-02-25
cli: eliminate src/lib/ directory and last cross-module import

Rugvip:

This eliminates the `packages/cli/src/lib/` directory entirely and removes the last cross-module import in the CLI package.

Three changes:

1. **Split `version.ts`** — template version provider (`packageVersions`, `createPackageVersionProvider`) moved to `modules/new/lib/version.ts`, while CLI version info (`version`, `isDev`, `findVersion`) moved to `wiring/version.ts`
2. **Remove `publishPreflightCheck` from `repo fix`** — it was the last cross-module import (from maintenance into build). The preflight check still runs during `package prepack`, so no loss of coverage
3. **Fix build module self-references** — `startFrontend.ts` was importing via `../../../../build/lib/bundler` instead of `../../../lib/bundler`

### DIFF
--- a/.changeset/cli-final-lib-cleanup.md
+++ b/.changeset/cli-final-lib-cleanup.md
@@ -1,5 +1,0 @@
----
-'@backstage/cli': patch
----
-
-Internal refactor to remove the `src/lib/` directory and eliminate the last cross-module import in the CLI package.


.changeset/cli-final-lib-cleanup.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli': patch
+---
+
+Internal refactor to remove the `src/lib/` directory and eliminate the last cross-module import in the CLI package.
```
Rugvip: skip the changeset, there's enough of them already

packages/cli/src/modules/maintenance/commands/repo/fix.ts
```diff
@@ -507,8 +505,6 @@ export async function command(opts: OptionValues): Promise<void> {
       fixPluginId,
       fixPluginPackages,
       fixPeerModules,
-      // Run the publish preflight check too, to make sure we don't uncover errors during publishing
-      publishPreflightCheck,
```
Rugvip, line 511: Figured we could skip this one here for now in order to unblock. It will fail at publishing instead but overall want to aim to move this to separate tooling so feeling it's alright

.changeset/cli-final-lib-cleanup.md
Rugvip: [automated] Fixed in 5411f0a

PR #33019, opened 2026-02-25
catalog: add group aliases and configurable content ordering to entity page

Rugvip:

This adds two new configuration features for entity page groups in the new frontend system:

**Group alias IDs** — Groups can now declare `aliases` so that content extensions targeting an aliased group ID get included in the aliasing group. This makes it easy to rename or merge groups without reconfiguring individual extensions:

```yaml
groups:
  - develop:
      title: Develop
      aliases:
        - development
```

**Configurable content ordering** — A new `contentOrder` option controls how content items are sorted within each group, with support for both a page-level default and per-group overrides:

```yaml
contentOrder: title  # new default
groups:
  - documentation:
      title: Docs
      contentOrder: natural  # override for this group
```

Two sort modes: `title` (alphabetical by title, the new default) and `natural` (preserves extension discovery order, the previous behavior). Ordering only applies within groups — ungrouped tabs retain their natural order.

### DIFF
--- a/plugins/catalog/src/alpha/components/EntityTabs/EntityTabsList.tsx
+++ b/plugins/catalog/src/alpha/components/EntityTabs/EntityTabsList.tsx
@@ -171,6 +171,11 @@ export function EntityTabsList(props: EntityTabsListProps) {
   }, [items, groupDefinitions, aliasToGroup, defaultContentOrder]);
 
   const selectedItem = items[selectedIndex];
+  const selectedGroup = resolveGroupId(
+    selectedItem?.group,
+    groupDefinitions,
+    aliasToGroup,
+  );
   return (
     <Box className={styles.tabsWrapper}>
       <Tabs
@@ -180,7 +185,7 @@ export function EntityTabsList(props: EntityTabsListProps) {
         variant="scrollable"
         scrollButtons="auto"
         aria-label={t('entityTabs.tabsAriaLabel')}
-        value={selectedItem?.group ?? selectedItem?.id}
+        value={selectedGroup ?? selectedItem?.id}
       >
         {groups.map(([id, tabGroup]) => (
           <EntityTabsGroup


plugins/catalog-react/src/alpha/blueprints/extensionData.tsx
```diff
@@ -41,13 +41,20 @@ export const entityFilterExpressionDataRef =
...
   });
 
+/** @alpha */
+export type EntityContentGroupDefinition = {
+  title: string;
+  icon?: string | ReactElement;
+  /** Other group IDs that should be treated as aliases for this group. */
+  aliases?: string[];
+  /** How to sort the content items within this group. Overrides the page-level default. */
+  contentOrder?: 'alpha' | 'natural';
```
Rugvip: Let's make it `'title'` instead of `'alpha'`, in case we want to add ordering by for example ID or something else in the future

plugins/catalog-react/src/alpha/blueprints/extensionData.tsx
```diff
@@ -41,13 +41,20 @@ export const entityFilterExpressionDataRef =
...
+};
+
 /** @alpha */
 export type EntityContentGroupDefinitions = Record<
   string,
-  {
-    title: string;
-    icon?: string | ReactElement;
-  }
+  EntityContentGroupDefinition
```
Rugvip: let's keep the type inlined here instead of declaring `EntityContentGroupDefinition`

plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
```diff
@@ -72,11 +72,12 @@ export function useSelectedSubRoute(subRoutes: SubRoute[]): {
 type EntityTabsProps = {
   routes: SubRoute[];
   groupDefinitions: EntityContentGroupDefinitions;
+  defaultContentOrder?: 'alpha' | 'natural';
```
Rugvip: Call this one `contentOrder` and also apply it to any of the additional content extensions that aren't part of any groups, although do not order the groups themselves by this

plugins/catalog-react/src/alpha/blueprints/extensionData.tsx
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/deb4476

plugins/catalog-react/src/alpha/blueprints/extensionData.tsx
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/deb4476

plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/deb4476

Rugvip: Actually let's roll this back. If one wants to have natural order of rest items and title order inside groups, that becomes very annoying to configure. Let's keep rest items in natural order for now an have "content order" refer to the ordering specifically within the groups. So move this back to `defaultContentOrder` as well.

Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/3eac7e2 — reverted to `defaultContentOrder` and restricted ordering to within groups only

plugins/catalog/src/alpha/components/EntityTabs/EntityTabsList.tsx
```diff
@@ -78,20 +78,59 @@ type TabGroup = {
...
-      const groupOrId = group && tab.group ? tab.group : tab.id;
+      const resolvedGroupId = resolveGroupId(
+        tab.group,
+        groupDefinitions,
+        aliasToGroup,
+      );
+      const group = resolvedGroupId
+        ? groupDefinitions[resolvedGroupId]
+        : undefined;
+      const groupOrId = group && resolvedGroupId ? resolvedGroupId : tab.id;
```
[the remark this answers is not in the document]
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/1b32ec6 — the selected tab value is now resolved through the alias map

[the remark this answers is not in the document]
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/1b32ec6

The `Tabs` `value` prop was using the raw `selectedItem.group` which could be an aliased ID (e.g. `"development"`), while the `EntityTabsGroup` components receive the resolved group key (e.g. `"develop"`) as their `value`. This mismatch would prevent MUI from matching the selected tab.

Fixed by resolving `selectedItem.group` through the same `resolveGroupId()` helper before passing it to the `Tabs` value:

```ts
const selectedGroup = resolveGroupId(
  selectedItem?.group,
  groupDefinitions,
  aliasToGroup,
);
// ...
value={selectedGroup ?? selectedItem?.id}
```

PR #33021, opened 2026-02-26
Fix error forwarding for `actionService` and `mcp-actions-backend`

benjdlambert:
Known errors thrown by actions were wrapped in `ForwardedError`, which coerced 400-level status codes to 500 and buried the original error message. This removes the `ForwardedError` wrapping so errors propagate with their original type and status code. Maybe there's a better fix here?


Rugvip (approved):
👍  to removing the wrapping, but `ForwardedError` should be forwarding the error name as well and allow for correct status codes. Wonder if there's something else going on? 🤔

PR #33022, opened 2026-02-26
`feat(catalog)`: Implement `query-catalog-entities` on top of the filter predicates

benjdlambert:
With #32793 being merged, we now can build a nice action on top of it to forward the predicates through to the Catalog.

This replaces the great work already started by @drodil over in #31380

### DIFF
--- a/plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
+++ b/plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
@@ -16,7 +16,6 @@
 import { ActionsRegistryService } from '@backstage/backend-plugin-api/alpha';
 import { CatalogService } from '@backstage/plugin-catalog-node';
 import { createZodV3FilterPredicateSchema } from '@backstage/filter-predicates';
-import { InputError } from '@backstage/errors';
 
 export const createQueryCatalogEntitiesAction = ({
   catalog,
@@ -74,34 +73,35 @@ Entity references use the format "kind:namespace/name", e.g. "component:default/
 
 Entities have bidirectional relations stored in the "relations" array. Common relation types: ownedBy/ownerOf, dependsOn/dependencyOf, providesApi/apiProvidedBy, consumesApi/apiConsumedBy, parentOf/childOf, memberOf/hasMember, partOf/hasPart.
 
-Relations can be queried via "relations.<type>" e.g. "relations.ownedby".
+Relations can be queried via "relations.<type>" e.g. "relations.ownedby: user:default/jane-doe". The value there must always be a valid entity reference.
 
-## Filter Syntax
+When querying for entity relationships, prefer using relations over spec fields. For example, use "relations.ownedby" instead of "spec.owner" to find entities owned by a particular group or user.
 
-The filter uses predicate expressions with dot-notation field paths.
+## Query Syntax
+
+The query uses predicate expressions with dot-notation field paths.
 
 Simple matching:
-  { filter: { kind: "Component" } }
-  { filter: { kind: "Component", "spec.type": "service" } }
+  { query: { kind: "Component" } }
+  { query: { kind: "Component", "spec.type": "service" } }
 
 Value operators:
-  { filter: { kind: { "$in": ["API", "Component"] } } }
-  { filter: { "metadata.annotations.backstage.io/techdocs-ref": { "$exists": true } } }
-  { filter: { "spec.dependsOn": { "$contains": "component:default/shared-lib" } } }
-  { filter: { "metadata.name": { "$hasPrefix": "team-" } } }
+  { query: { kind: { "$in": ["API", "Component"] } } }
+  { query: { "metadata.annotations.backstage.io/techdocs-ref": { "$exists": true } } }
+  { query: { "metadata.tags": { "$contains": "java" } } }
+  { query: { "metadata.name": { "$hasPrefix": "team-" } } }
 
 Logical operators:
-  { filter: { "$all": [{ kind: "Component" }, { "spec.lifecycle": "production" }] } }
-  { filter: { "$any": [{ "spec.type": "service" }, { "spec.type": "website" }] } }
-  { filter: { "$not": { kind: "Group" } } }
+  { query: { "$all": [{ kind: "Component" }, { "spec.lifecycle": "production" }] } }
+  { query: { "$any": [{ "spec.type": "service" }, { "spec.type": "website" }] } }
+  { query: { "$not": { kind: "Group" } } }
 
 Querying relations - find all entities owned by a specific group:
-  { filter: { "relations.ownedby": "group:default/team-alpha" } }
+  { query: { "relations.ownedby": "group:default/team-alpha" } }
 
 Combined example - find production services or websites with TechDocs:
-  { filter: { "$all": [
-    { kind: "Component" },
-    { "spec.lifecycle": "production" },
+  { query: { "$all": [
+    { kind: "Component", "spec.lifecycle": "production" },
     { "$any": [{ "spec.type": "service" }, { "spec.type": "website" }] },
     { "metadata.annotations.backstage.io/techdocs-ref": { "$exists": true } }
   ] } }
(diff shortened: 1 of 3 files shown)


plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+  spec fields: memberOf (entity refs), profile (displayName, email, picture).
+
+**Location** - A marker that references other catalog descriptor files to be ingested.
+  spec fields: type, target, targets, presence.
+
+### Relations
+
+Entities have bidirectional relations stored in the "relations" array. Common relation types: ownedBy/ownerOf, dependsOn/dependencyOf, providesApi/apiProvidedBy, consumesApi/apiConsumedBy, parentOf/childOf, memberOf/hasMember, partOf/hasPart.
+
+Relations can be queried via "relations.<type>" e.g. "relations.ownedby".
```
freben:
```suggestion
Relations can be queried via "relations.<type>" e.g. "relations.ownedby: user:default/jane-doe". The value there must always be a valid entity reference.
```

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+
+Limit returned fields: { fields: ["kind", "metadata.name", "metadata.namespace"] }
+Sort results: { orderFields: { field: "metadata.name", order: "asc" } }
+Full text search: { fullTextFilter: { term: "auth" } }
+Pagination: Use limit (e.g. 20) and the returned nextPageCursor for subsequent requests via cursor.
+    `,
+    schema: {
+      input: z =>
+        z.object({
+          filter: createZodV3FilterPredicateSchema(z)
```
freben:
```suggestion
          query: createZodV3FilterPredicateSchema(z)
```

just to keep things aligned with the service APIs and the catalog client etc. Filter will be going away as the old stuff.

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+            .optional()
+            .describe(
+              'Entity predicate filter. Supports field matching, $all, $any, $not, $exists, $in, $contains, and $hasPrefix operators.',
+            ),
+          fields: z
+            .array(z.string())
+            .optional()
+            .describe(
+              'Specific fields to include in the response. If not provided, all fields are returned.',
```
freben:
```suggestion
              'Specific fields to include in the response. If not provided, all fields are returned. Each entry is a dot separated path into an entity, e.g. `spec.type`.',
```

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+            .describe(
+              'Ordering criteria for the results. Can be a single order directive or an array for multi-field sorting.',
+            ),
+          fullTextFilter: z
+            .object({
+              term: z.string().describe('Full text search term'),
+              fields: z
+                .array(z.string())
+                .optional()
+                .describe('Fields to search within'),
```
freben:
```suggestion
                .describe('Fields to search within. Each entry is a dot separated path into an entity, e.g. `spec.type`.'),
```

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+            })
+            .optional()
+            .describe('Full text search criteria'),
+          cursor: z
+            .string()
+            .optional()
+            .describe(
+              'Cursor for pagination. This can be used only after first request with response containing a cursor',
```
freben:
```suggestion
              'Cursor for pagination. This can be used only after the first request with a response containing a cursor. If a cursor is given it takes precedence over `offset`',
```

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+          nextPageCursor: z
+            .string()
+            .optional()
+            .describe('Next page cursor used to fetch next page of entities'),
+        }),
+    },
+    action: async ({ input, credentials }) => {
+      if (input.cursor && (input.filter || input.offset !== undefined)) {
+        throw new InputError(
+          'Cannot combine cursor with filter or offset. Use cursor alone for pagination.',
```
freben: I don't think we should worry too much about this. Especially not in the action. It's kind of nice to write a loop that calls with

```
{ ...request, cursor: currentCursor }
```

without worrying about what to hand pick out of the object. Especially since I'm not very fond of the kind of oddball subset of things we put in the cursor and which ones you MUST send with each request.

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+        );
+      }
+
+      const response = await catalog.queryEntities(
+        input.cursor
+          ? {
+              cursor: input.cursor,
+              fields: input.fields,
+              limit: input.limit,
```
freben: yeah this is the thing i'm talking about; forcing every caller everywhere to know about these unique semantics that differ in different endpoints, is unfortunate.

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+          offset: z
+            .number()
+            .int()
+            .min(0)
+            .optional()
+            .describe('Number of entities to skip before returning results.'),
+          orderFields: z
+            .union([
+              z.object({
+                field: z.string().describe('Field to order by'),
```
freben:
```suggestion
                field: z.string().describe('Field to order by. The format is a dot separated path into an entity, e.g. `spec.type`.'),
```

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+            .optional()
+            .describe(
+              'Specific fields to include in the response. If not provided, all fields are returned.',
+            ),
+          limit: z
+            .number()
+            .int()
+            .positive()
+            .optional()
+            .describe('Maximum number of entities to return.'),
```
freben:
```suggestion
            .describe('Maximum number of entities to return at a time.'),
```

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+Entities have bidirectional relations stored in the "relations" array. Common relation types: ownedBy/ownerOf, dependsOn/dependencyOf, providesApi/apiProvidedBy, consumesApi/apiConsumedBy, parentOf/childOf, memberOf/hasMember, partOf/hasPart.
+
+Relations can be queried via "relations.<type>" e.g. "relations.ownedby".
+
+## Filter Syntax
+
+The filter uses predicate expressions with dot-notation field paths.
+
+Simple matching:
+  { filter: { kind: "Component" } }
```
freben: as mentioned in the other comment - probably change all these to "query"

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+  { filter: { kind: "Component", "spec.type": "service" } }
+
+Value operators:
+  { filter: { kind: { "$in": ["API", "Component"] } } }
+  { filter: { "metadata.annotations.backstage.io/techdocs-ref": { "$exists": true } } }
+  { filter: { "spec.dependsOn": { "$contains": "component:default/shared-lib" } } }
```
freben: A bit unfortunate to propose searching in a relation generating spec field, could also use `metadata.tags` if you wanted an array

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+
+Querying relations - find all entities owned by a specific group:
+  { filter: { "relations.ownedby": "group:default/team-alpha" } }
+
+Combined example - find production services or websites with TechDocs:
+  { filter: { "$all": [
+    { kind: "Component" },
+    { "spec.lifecycle": "production" },
+    { "$any": [{ "spec.type": "service" }, { "spec.type": "website" }] },
+    { "metadata.annotations.backstage.io/techdocs-ref": { "$exists": true } }
```
freben, line 106: 1, 2, and 4 could be a single entry but it's ok

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -0,0 +1,226 @@
...
+  ] } }
+
+## Other Options
+
+Limit returned fields: { fields: ["kind", "metadata.name", "metadata.namespace"] }
+Sort results: { orderFields: { field: "metadata.name", order: "asc" } }
+Full text search: { fullTextFilter: { term: "auth" } }
```
freben: i think fields is required in this one

plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts
freben: Thinking about it some more ... we should probably iterate on this to say something about do NOT use the spec if you know there's a relation that fits the purpose (how would it know though).

PR #33030, opened 2026-02-26
fix(openapi-jest-utils): memory leak in proxy clean up

aramissennyeydd:

Fixes a memory leak in proxy cleanup where the module-level cache kept references to old proxies that had already been cleaned up.

### DIFF
--- a/packages/backend-openapi-utils/src/testUtils.ts
+++ b/packages/backend-openapi-utils/src/testUtils.ts
@@ -16,9 +16,8 @@
 import { Express } from 'express';
 import { Server } from 'node:http';
 import { Proxy } from './proxy/setup';
-import { randomUUID } from 'node:crypto';
 
-const proxiesToCleanup: Record<string, Proxy> = {};
+const proxiesToCleanup: Set<Proxy> = new Set();
 
 /**
  * !!! THIS CURRENTLY ONLY SUPPORTS SUPERTEST !!!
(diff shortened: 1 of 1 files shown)


packages/backend-openapi-utils/src/testUtils.ts
```diff
@@ -16,8 +16,9 @@
 import { Express } from 'express';
 import { Server } from 'node:http';
 import { Proxy } from './proxy/setup';
+import { randomUUID } from 'node:crypto';
 
-const proxiesToCleanup: Proxy[] = [];
+const proxiesToCleanup: Record<string, Proxy> = {};
```
freben: This can be `Set<Proxy>` and then use reference equality to `.delete(proxy)` in the finally clause

packages/backend-openapi-utils/src/testUtils.ts
```diff
@@ -48,10 +50,11 @@ function registerHooks() {
...
 
-  afterAll(() => {
-    for (const proxy of proxiesToCleanup) {
-      proxy.stop();
-    }
+  afterAll(async () => {
+    const stopPromises = Object.entries(proxiesToCleanup).map(([key, proxy]) =>
+      proxy.stop().finally(() => delete proxiesToCleanup[key]),
+    );
+    await Promise.allSettled(stopPromises);
```
freben, line 55: But wait - you really didn't need to delete them one by one right? Was there really any risk of more of them being added WHILE `afterAll` was running? At this point you should be able to just await all of the stop calls, and then do `proxiesToCleanup.length = 0` (which looks crazy but works, feel free to do `proxiesToCleanup.splice(0, proxiesToCleanup.length)` instead)

freben:
And one more question - did it really leak? When running afterAll, isn't the worker being shut down anyway?

packages/backend-openapi-utils/src/testUtils.ts
aramissennyeydd: I'm ... not really sure - this is a module level cache which I imagine could be added to across workers?

aramissennyeydd: updated to use .clear here on the set, `.length` I _think_ only works on arrays

PR #1, opened 2020-02-01
Add LICENSE, CoC etc

stefanalund:

PR #33042, opened 2026-02-26
630:P1 Add gateway-backend router and proxy regression tests

salonitilekar:
Closes #1

## Summary
- add an upstream test route (`/echo/:id`) in gateway integration tests to validate forwarded request shape
- add a regression test to ensure installed/local plugins bypass proxy resolution
- add a regression test to ensure proxied requests preserve path and query parameters

## How to run relevant tests
- yarn test --no-watch plugins/gateway-backend/src/plugin.test.ts

## Evidence (behavior now protected)
- local route handling is protected from accidental proxying (`discovery.getBaseUrl` is not called for installed plugin routes)
- proxy forwarding behavior is protected for at least one upstream, including path and query preservation
- all gateway plugin tests pass with the new coverage

### DIFF
--- a/plugins/gateway-backend/src/plugin.test.ts
+++ b/plugins/gateway-backend/src/plugin.test.ts
@@ -105,6 +105,14 @@ describe('gateway', () => {
                 res.json({ bar: true });
               });
 
+              router.get('/echo/:id', (req, res) => {
+                res.json({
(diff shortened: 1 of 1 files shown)


salonitilekar:
Opened in upstream repo by mistake for class assignment; re-opening in team fork.

PR #33045, opened 2026-02-27
perf(ui): share a single set of matchMedia listeners in useBreakpoint

Naycon:

Each call to `useBreakpoint()` previously created 18 independent
`matchMedia` listeners via `useMediaQuery` (6 for current breakpoint,
6 duplicate for `up()`, 6 for `down()`). Since `useBreakpoint` is
called from `useStyles`, `useDefinition`, and `useBgProvider`, a
typical page accumulated hundreds of listeners that all fired on
every resize.

This replaces the per-instance `useState`/`useMediaQuery` pattern
with a module-level singleton using `useSyncExternalStore`. All
callers now share a single set of 6 `matchMedia` listeners regardless
of component count. The `up()` and `down()` helpers are now pure
index comparisons instead of additional media query subscriptions.

The public API is unchanged.

### Performance tests

This is a simple resize from 980ish to 1030ish pixels, crossing the 1024px breakpoint boundary. It's from the docs-ui Table page, simply because it renders a bunch of BUI components in storybook snippets and was convenient for running the tests. The gains are bigger when crossing more breakpoint boundaries.

**Before**

image

**After**

image

### DIFF
--- a/packages/ui/src/hooks/useBreakpoint.ts
+++ b/packages/ui/src/hooks/useBreakpoint.ts
@@ -49,9 +49,12 @@ function computeBreakpoint(): Breakpoint {
   return 'initial';
 }
 
-// --- Singleton store ---
+// --- Module-scoped singleton store ---
+// This is intentionally not a global singleton. Multiple copies of this module
+// (e.g. different package versions or module federation remotes) each get their
+// own store. This avoids cross-version coupling at the cost of a few extra
+// listeners, which is a fine trade-off.
 // `current` is initialized lazily on first client-side access.
-// `listeners` is eagerly initialized since Set is safe to create on the server.
 
 let current: Breakpoint | undefined;
 const listeners = new Set<() => void>();


packages/ui/src/hooks/useBreakpoint.ts
```diff
@@ -25,40 +26,92 @@ export const breakpoints: { name: string; id: Breakpoint; value: number }[] = [
...
-    down: (key: Breakpoint): boolean => {
-      return downMatches.get(key) ?? false;
-    },
+// --- Singleton store ---
+// `current` is initialized lazily on first client-side access.
+// `listeners` is eagerly initialized since Set is safe to create on the server.
+
+let current: Breakpoint | undefined;
+const listeners = new Set<() => void>();
+let initialized = false;
```
Rugvip: Just so there's no surprise: this will still be duplicated across BUI instances, such as different package versions or module federation remotes. From what can tell that's completely fine though.

Maybe make a note that we're not bothering with a global singleton here though. And I don't think we should, because it complicates stability.

Naycon: Yea, fair point. I'll update the comment to point this out. Agree it's fine to duplicate it a few times.

Naycon: Fixed.

PR #33046, opened 2026-02-27
feat(auth-backend): add `who-am-i` action to actions registry

benjdlambert:
A simple action to return the Catalog user that is signed in. Not sure if this is best placed in the `catalog-backend` or in `auth-backend`. 

### DIFF
--- a/plugins/auth-backend/src/actions/createWhoAmIAction.ts
+++ b/plugins/auth-backend/src/actions/createWhoAmIAction.ts
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2025 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


plugins/auth-backend/src/actions/createWhoAmIAction.ts
```diff
@@ -0,0 +1,66 @@
...
+        credentials,
+      });
+
+      if (!entity) {
+        throw new NotFoundError(
+          `User entity not found in the catalog for "${userEntityRef}"`,
+        );
+      }
+
+      return { output: entity };
```
Rugvip: should we also at least prepare for forwarding custom token claims as well? 🤔

Not sure if we want to or not, but figured it can be nice to have the output be `{ entity }`.

Maybe we simply call the user info service and return whatever it provides alongside the entity?

benjdlambert: yeah good shout!

PR #33053, opened 2026-02-27
`feat(scaffolder)`: add BUI theme for scaffolder forms

benjdlambert:
Adds an experimental Backstage UI RJSF Form theme as an alternative to the Material UI theme for scaffolder forms. The form logic is heavily inspired from the Material-UI one.

All field extensions (`EntityPicker`, `RepoUrlPicker`, etc.) render BUI variants when enabled, which is a bit of a mess short term, but easier to have all the business logic in one place rather than duplicating the components I felt.

You can enable it using the following config:

```yaml
app:
  extensions:
    - sub-page:scaffolder/templates:
        config:
          enableBackstageUi: true
```

Or in the legacy frontend system through JSX:

```tsx
<ScaffolderPage formProps={{ EXPERIMENTAL_theme: 'bui' }} />
```

There will be a point in the future where we flip the defaults with an opt out to let people migrate their Field Extensions from MUI to BUI, but you can opt in now and migrate them if you wish!

A kitchen-sink template (`bui-kitchen-sink-demo`) is registered in the example app to exercise every widget and field extension when the BUI theme is enabled.

Inputs use an outlined style (matching BUI's PasswordField) so they sit on the same neutral-1 surface as the form rather than relying on a background step-up that can read as disabled. Form widgets where BUI doesn't yet ship a component (Textarea, date/time pickers) fall back to native HTML inputs styled to match.

image


acierto (approved):

plugins/scaffolder-react/src/next/components/Form/BuiTheme/templates/ArrayFieldDescriptionTemplate.tsx
Sarabadu, line 1: Instead of adding all the components inside the scaffolder, does it make sense to have a new `@backstage/ui-rjsf` packages exporting the theme, and just select the theme inside the scaffolder?

Sarabadu (commented):
Nice to see this is moving, just a comment, I was hoping to see this as a package to use it also outside of scaffolder

plugins/scaffolder-react/src/next/components/Form/BuiTheme/templates/ArrayFieldDescriptionTemplate.tsx
benjdlambert: perhaps eventually yes.

PR #33059, opened 2026-02-27
feat(auth): surface JWT payload claims on BackstageServicePrincipal for jwks tokens

jsundquist:

When a request is authenticated via an external access method of type `jwks`, the verified JWT payload is now preserved and exposed as `tokenData` on the resulting `BackstageServicePrincipal`. This allows action handlers and plugins to read identity claims (e.g. `upn`, `name`, `oid`, `appid`) directly from `context.credentials.principal.tokenData` without re-decoding the token.

The field is optional and only populated for `jwks`-validated tokens; all existing static-token and plugin-token principals are unaffected.


packages/backend-plugin-api/report.api.md
```diff
@@ -171,6 +171,7 @@ export type BackstageServicePrincipal = {
   type: 'service';
   subject: string;
   accessRestrictions?: BackstagePrincipalAccessRestrictions;
+  tokenClaims?: Record<string, unknown>;
```
Rugvip: I think we should view this from the point of view of forwarding arbitrary data from the external token handlers, and with that make it even more general. The name of the key is the main bit, where these wouldn't necessarily always be claims in the JWT sense. Perhaps the "token" prefix is a bit redundant as well, but maybe not. Could also be good to have a name that implies that this is intended for internal customization only, although given the type that should be fairly clear. Some alternatives that come to mind:

- `customData`
- `tokenData`
- `attributes`

Wondering if making this a `JsonObject` from `@backstage/types` makes it easier to consume the data as well? It means you have to pass through raw data from custom token handlers rather than being able to attach functions etc, but I think that's alright. If we make this serializable we'd need this as well.

packages/backend-defaults/src/entrypoints/auth/helpers.ts
```diff
@@ -128,13 +133,20 @@ export function toInternalBackstageCredentials(
...
+  tokenClaims?: Record<string, unknown>,
 ): BackstageServicePrincipal {
   const result = {
     type: 'service',
     subject: sub,
     accessRestrictions,
   } as const;
   Object.defineProperties(result, {
+    tokenClaims: {
+      enumerable: false,
```
Rugvip, line 142: How come you went with making this non-enumerable? don't have a strong opinion at this point, just looking to understand

Rugvip (changes requested):
Thank you! Agree this would be useful and mainly want to make sure we keep the separation between the interface and implementation clean.

Another potential direction to explore is to add the ability to attach context values to the credentials directly (in the Golang sense). It's even the case that you could do that already now afaik using symbols as keys for the credentials object. If we do end up going with that pattern instead I think it could be good to document it to make it an explicitly supported thing that we don't accidentally break in the future,

packages/backend-plugin-api/report.api.md
jsundquist: This has been updated to make use of `tokenData` and updated to also use `JsonObject`.

packages/backend-defaults/src/entrypoints/auth/helpers.ts
jsundquist: The main reason for making it non-enumerable is to prevent accidental leakage of sensitive PII. Since credentials/principals are JSON-serializable in Backstage, an enumerable `tokenData` would silently appear in `JSON.stringify()` output — and by extension in logs, telemetry, and anything else that serializes credentials. Non-enumerable keeps it invisible to serialization while still being fully accessible when you explicitly reach for `credentials.principal.tokenData`. This also follows the same pattern as the `token` property, which is non-enumerable for the same reason.

PR #33072, opened 2026-03-02
feat(catalog): allow clearing kind filter when allFilterEnabled is trued

### LINKED ISSUE #33052: Catalog Filters - Kind filter (All)
### 📜 Issue Labels

### 🔎 Search Terms

```plain
Catalog, Kind Filters, All Kind Filter, Filters
```

### 🗃️ Project Area

Catalog

### 🔖 Need

When searching the catalog for entities. I personally find it confusing that I must select an kind in which I wish to seach and that if I am doing a more general search of the catalog to see what is available to me, that I have to come back to the kind filter to modify and then research. 

Ideally being able to search the entire catalog woudl be useful from within the catalog page. 
The main filter, immho, which doens't seem to have a good default currently is the kind filter. 
I would like the option to have an `All` filter, in the kind dropdown which will allow searching of the catalog by all kinds. 

### 📝 Proposal

I have had a look through the codebase and I think that one way to make this change would be to add the kind `All` to the list of items in the `EntityKindPicker`. This would be similar to how it is handled in the `EntityTypePicker`. 

Then, by modifying the hook to check for specifically an `all` kind filter, this would then update the filters to the undefined for kind. This would then mean when this fitler is applied the url query parameter, it would not include kind and therefore bring back all entity kinds. 

As this may not be a feature wanted by the whole community, I would suggest making this an optional, nullable prop of the EntityKindPicker, something to the effect of `AllFilterEnabled?: Boolean` with the default value set to`false` such that current users of EntityKindPicker would not be effected by the change and instead would have to opt into the change if suitable for their usecase. 

### 🔄 Alternatives

_No response_

NETIZEN-11:
## Summary

This PR updates the Catalog Kind filter behavior to allow clearing the filter when `allFilterEnabled` is set to `true`.

Previously, when the "All" option was enabled, the Kind filter could not be fully cleared. This change ensures that users can properly reset the filter state while preserving the expected behavior of the "All" option.

## Changes

- Updated Kind filter logic in the Catalog
- Enabled clearing/reset behavior when `allFilterEnabled` is true
- Maintained existing filtering functionality

## Notes

Please let me know if any adjustments are needed or if additional test coverage should be added.

Closes #33052


awanlin (changes requested):
Hi @NETIZEN-11, we are still coming to an agreement on a path forward: https://github.com/backstage/backstage/issues/33052

NETIZEN-11:
Thanks for the update! I'll wait for the final decision on the approach. Let me know if any changes are needed from my side.

PR #33079, opened 2026-03-02
feat(ui/Text): add copy-to-clipboard and disabled features

deepthi-28:
Adds copy-to-clipboard functionality to the Text component
Introduces a disabled prop for visually non-interactive text
Updates stories and documentation for new Text features

image
image


Naycon (changes requested):
Can you raise an issue here to discuss what it is you need and why? I'm not sure a general copy button on text makes sense, and I'm also not sure when you'd need a disabled text variant. Disabled is usually tied to an interactive BUI component, and those tend to support disabling already.

deepthi-28:
> Can you raise an issue here to discuss what it is you need and why? I'm not sure a general copy button on text makes sense, and I'm also not sure when you'd need a disabled text variant. Disabled is usually tied to an interactive BUI component, and those tend to support disabling already.

Hi @Naycon, The TextField already supports a disabled state that visually mutes the entire field. When a Text component is used as a <label> alongside a disabled TextField, there was no way to match that same visual state on the label. Adding disabled to Text makes it consistent with how TextField and other form components already behave in BUI.

Another use case: when displaying read-only identifiers (e.g. entity UIDs, pipeline IDs, or system-generated values), the disabled style clearly signals to the user that this value cannot be edited — it's just for reference. Combined with copyable, users can still copy the ID while understanding it's not an editable field.

**Copy-to-clipboard** is useful for displaying read-only values that users often need to copy — such as entity UIDs, API keys, or pipeline IDs — without requiring a separate `ButtonIcon` layout every time.

Happy to raise a GitHub issue to discuss further if that helps!

Naycon:
> Hi @Naycon, The TextField already supports a disabled state that visually mutes the entire field. When a Text component is used as a alongside a disabled TextField, there was no way to match that same visual state on the label. Adding disabled to Text makes it consistent with how TextField and other form components already behave in BUI.
> 

Not sure I follow here - `TextField` is a form component, which can be disabled yes. But that shows the value inside the input in a different way, but the label of the disabled TextField does not change. This is intentional, the label should remain consistent and easily readable/scannable even when the input is disabled.

image

> Another use case: when displaying read-only identifiers (e.g. entity UIDs, pipeline IDs, or system-generated values), the disabled style clearly signals to the user that this value cannot be edited — it's just for reference. Combined with copyable, users can still copy the ID while understanding it's not an editable field.

If you want to display read-only identifiers, I suggest using the same component as you would for the non-readonly version of the identifiers, and then set `isReadOnly` on them. This is available on some form components, but I'm happy to review PRs that enables this for more if it's not already there via React Aria inheritance, or that improves the visual styling of components that do enable it. 

If you don't want to use those, I'd suggest adjusting the design so the read-only identifiers are clearly readable in some other way by e.g. grouping them separately so it becomes clear from the context and semantics that these are key value pairs that are set and can't be changed. This is partially why I said to raise an issue - it would be easier to reason about this if you gave us concrete examples and screenshots of what you're trying to do. :)

> **Copy-to-clipboard** is useful for displaying read-only values that users often need to copy — such as entity UIDs, API keys, or pipeline IDs — without requiring a separate `ButtonIcon` layout every time.

I can see why having a Copy to clipboard component is useful, but I do believe this should be it's own component. You may e.g. want to have multiple `Text` components content copy via one button, to e.g. be able to copy several sections worth of text in one go without having to resort to rendering both of them in the same Text component with "hacks" like `<br />` to line break between them. I actually made a [somewhat accessible copy button recently](https://github.com/backstage/backstage/pull/33725/changes#diff-d77c81da03882b99b0d8aec414abb1df333e1a8113f3e588e92719999d9adfe4R126) for another PR. I've kept it locally there for now while we figure out if this type of functionality should be offered up via BUI or if it should live in some other shared Backstage package.

PR #33092, opened 2026-03-03
fix: OAuth 2.0 Protected Resource Metadata returning local endpoint (fix #33062)

### LINKED ISSUE #33062: .well-known/oauth-protected-resource endpoint only returns localhost URLs
### 📜 Issue Labels

### 🔎 Search Terms

```plain
"mcp", "oauth-protected-resource"
```

### 🗃️ Project Area

Auth

### 🔗 External Integration

N/A

### 📝 Description & Context

On Backstage 1.48.1 (`@backstage/plugin-mcp-actions-backend` version 0.1.9), with Dynamic Client Registration enabled, My MCP client fails to authenticate with any Backstage instance other than those running on `localhost`.

### 👍 Expected Behavior

The authentication flow starting with Claude Code, VS Code, etc. can resolve the correct OAuth authorization code flow URL to send the user to the OAuth consent page in Backstage.

### 🥾 Reproduction steps

1. Start Backstage on custom domain (https://backstage.example.com, not a real site but illustrates the point)
2. Set `auth.experimentalDynamicClientRegistration.enabled` to `true`  
4. Add the Backstage instance's MCP endpoint to Claude Code's MCP settings (https://backstage.example.com/api/mcp-actions/v1)
5. Claude Code will display error: `Error: Unable to connect. Is the computer able to access the url?`
6. Inspect Claude Code's logs and see: `Error: Protected resource http://localhost:3000/api/mcp-actions does not match expected https://backstage.example.com/api/mcp-actions/v1 (or origin)`
7. Browse to https://backstage.example.com/.well-known/oauth-protected-resource which displays:

```json
{
  "resource": "http://localhost:3000/api/mcp-actions",
  "authorization_servers": [
    "http://localhost:3000/api/auth"
  ]
}
```

Downgrading `@backstage/plugin-mcp-actions-backend` to `0.1.9-next.1` allows Claude Code to authorize with an MCP server at https://backstage.example.com.

Notably, everything also works as expected when Backstage is running on `localhost`.

GabDug:

Fix #33062, by returning external plugin URL, like auth-backend well-known OIDC files.

### DIFF
--- a/plugins/mcp-actions-backend/src/plugin.test.ts
+++ b/plugins/mcp-actions-backend/src/plugin.test.ts
@@ -191,10 +191,17 @@ describe('Mcp Backend', () => {
     });
 
     it('should expose oauth-protected-resource when DCR is enabled', async () => {
+      const mockExternalBaseUrl = 'http://external.local:0/api';
+      const mockDiscovery = mockServices.discovery.mock({
+        getExternalBaseUrl: async pluginId =>
+          `${mockExternalBaseUrl}/${pluginId}`,
+      });
+
       const { server } = await startTestBackend({
         features: [
           mcpPlugin,
           mockPluginWithActions,
+          mockDiscovery.factory,
           mockServices.rootConfig.factory({
             data: {
               backend: {
@@ -219,6 +226,10 @@ describe('Mcp Backend', () => {
       expect(response.body.resource).toMatch(/\/api\/mcp-actions$/);
       expect(response.body.authorization_servers).toHaveLength(1);
       expect(response.body.authorization_servers[0]).toMatch(/\/api\/auth$/);
+      expect(response.body.resource).toContain(`${mockExternalBaseUrl}`);
+      expect(response.body.authorization_servers[0]).toContain(
+        `${mockExternalBaseUrl}/`,
+      );
     });
 
     it('should expose oauth-protected-resource when CIMD is enabled', async () => {


plugins/mcp-actions-backend/src/plugin.ts
```diff
@@ -114,8 +114,8 @@ export const mcpPlugin = createBackendPlugin({
             '/.well-known/oauth-protected-resource',
             async (_, res) => {
               const [authBaseUrl, mcpBaseUrl] = await Promise.all([
-                discovery.getBaseUrl('auth'),
-                discovery.getBaseUrl('mcp-actions'),
+                discovery.getExternalBaseUrl('auth'),
+                discovery.getExternalBaseUrl('mcp-actions'),
               ]);
```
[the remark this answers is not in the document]
GabDug: Yeah fair point.

plugins/mcp-actions-backend/src/plugin.ts
```diff
@@ -114,8 +114,8 @@ export const mcpPlugin = createBackendPlugin({
             '/.well-known/oauth-protected-resource',
             async (_, res) => {
               const [authBaseUrl, mcpBaseUrl] = await Promise.all([
-                discovery.getBaseUrl('auth'),
-                discovery.getBaseUrl('mcp-actions'),
+                discovery.getExternalBaseUrl('auth'),
```
benjdlambert, line 117: Hmm yeah, I thought about this a lot before making it the internal baseUrl.

The idea here is that you need to access the `auth` plugin without going through any form of reverse proxy right?

If you had something in front of Backstage which required you to go through authentication steps, then this would no longer be compatible with Claude or MCP clients.

@freben I think we spoke about this too right?

GabDug: @benjdlambert Yeah I was just reading #32307.

I'm not sure I understand the setup you're describing, if you have something in front of Backstage w/ additional auth steps that prevent access to your `.well-known` / `auth` plugin, I don't see how MCP client login would work without an additional token/header to auth?

Afaik, the resource needs to match the external MCP actions path to avoid Claude Code `Error: Protected resource http://<internal>/api/mcp-actions does not match expected https://<external>/api/mcp-actions/v1 (or origin)`, but it _may_ work with an un-reachable auth server.

GabDug: I definitively don't have the full picture here, especially w.r.t. the numerous MCP clients implementations & various ways of deploying Backstage.

This seems to do the trick for our close-to-default deployment setup with Claude Code. VS Code MCP auth does not seem to care.

benjdlambert: Let me explain a little bit more:

So the `external` URL is what you would expect the frontend to call. 

This means that it can include it's session, for instance an SSO session. This `external` URL might point to an SSO load balancer or something which you have to sign into first. 

This external URL is not what you want backends to call, or Claude or MCP clients as they shouldn't have to go through this SSO reverse proxy, it should be an internal URL only. 

So the MCP server would be `http://my-company.internal.com/api/mcp-actions/v1`

What is your `backend.baseUrl` set as in config? That should be your internal domain like `my-company.internal.com`

You can also perhaps fix this with explicit config:

```yaml
discovery:
  endpoints:
    - target: http://my-company.internal.com/api/auth
      plugins: [auth]
```

benjdlambert: Basically all this to say, that I think you should set the MCP Server URL to the internal URL not the external one. 😅

GabDug: @benjdlambert Ha I understand you POV better!

I don't have the same understanding of internal / external: the internal URL is [documented](https://backstage.io/api/stable/interfaces/_backstage_backend-plugin-api.index.DiscoveryService.html) as for "service-to-service" (e.g. plugin-to-plugin via service mesh...) and the external as "reachable from the Backstage frontend and other external services" (a MCP client / CLI should imo fit there).

In our case, the "external" URL is already behind a VPN (which I assume is a common pattern), so reachable by both the frontend and our end-users machines for a MCP/CLI.
I'm not sure how end-users would resolve a K8s service or cluster-network IP, if we expose the internal URL.

With a default (single deployment) Backstage deployment, internal would resolve to localhost:7007/api/...

freben: The instant that you

* scale up to more than one single pod, _or_
* split your backend plugins apart into dedicated deployments, _or_
* start calling your plugins from anything else than within itself

(whichever of them comes first), `backend.baseUrl` should be a routable address. Even if it's just two pods, you want them to talk amongst each other as a mesh to share load. Something like `backstage-backend.zone.ourservicenetwork.company.net` or whatever. Such that resolving it actually hits one of your pods. This is typically an address that requires you to be on the service network or have VPN.

`app.baseUrl` is decidedly outward facing. Some do choose it to be only internally routable, absolutely. But a significant part of adopters have it publicly routable and pointing at their corporate SSO, since it's so much easier to reach it that way while on the go.

The docs saying that it's for "other external services" should probably be clarified. For external services that aren't on your service network at all, and you for example need to receive webhooks for them - then yes, this _could_ be the address you hand out, by sheer necessity. But that comes with the caveat that if you do so, you expect adopters to configure their infra to surgically skip SSO for a subset of routes and maybe add allowlists for remote IP and things like that, to get traffic flowing.

`app.baseUrl` truly is heavily focused on browser traffic. I absolutely cannot curl ours, for example, barring stealing some cookie headers from the browser and going through some hoops.

Yes, handing out the `getBaseUrl` as the original code does, implies that callers probably have to be on VPN to use it too.

GabDug: @freben Thanks for clarification & info. I agree with you that `backend.baseUrl` should be routable.

However what you describe still doesn't match my understanding of the documentation of DiscoveryService 😅 

`getBaseUrl` specifically states "internal endpoint for the plugin, with the shortest route possible. The URL should be used for service-to-service communication within a Backstage backend deployment.". 

Its example is `http://10.1.2.3/api/catalog` which could very well be in a cluster subnet. I understand with `shortest route possible` that a K8s service, or an IP could be a correct implementation, skipping any gateway/reverse-proxy. With a single-pod deployment, serving both frontend and all backend plugins, it is `localhost`.

Expecting the result of `getBaseUrl` to be user-accessible and not only "within a Backstage backend deployment" seems like a significant change, although the use-case you described is relevant.

On top of the documentation clarification you mentioned, I wonder if there's should not be third option to get the right URL depending on its usage (probably with some overlap depending on your deployment pattern):
- Backstage Internal for plugin-to-plugin (`localhost:7007`, `*.svc.cluster.local`)
- Company/user-accessible
- World-accessible for webhooks/callbacks

freben: > With a single-pod deployment, serving both frontend and all backend plugins, it is localhost.

l disagree even with this one. :) A pod would even do well to basically actively _discourage_ speaking to its own instance and prefer a neighbouring one instead where available - which could be the case during a rollout when a new candidate pod is spinning up before tearing down the previous one. Imagine you rolling out a change, and you have a new plugin in there that's a bit CPU heavy and needs to speak with the catalog that was there since before the change. It's beneficial for this new rollout to start speaking to pre-existing, live, hot catalog instances wherever they may be - within reason of course, in the same zone etc. And ideally of course the address it does so through is an API gateway or service mesh or similar so you can get safe routing and retries and load balancing to the least loaded pod at that moment in time and whatnot on top of it all. Localhost is probably _not_ what you should be talking to, with a plethora of plugins all racing to try to start up at the same time and maybe the readiness check isn't even green yet.

Crucially, this new service doesn't even know where the catalog is located, and does not have to care. You can change your deployment topology and do all sorts of wild things, because you know that the target url is set to something stable either way.

And there's a hidden intent with these urls too - that they be shareable in a response to a third party for "you can reach me on this address". We just need to document this better.

GabDug: @freben Agreed, but while it's not a very good setup, but it is the default one, and I assume that other are running naive setups with a couple of replicas.

My point is not to advise anyone to keep using the default, it's just that I don't believe the assumption that the result of  `getBaseUrl` is user-accessible holds true, e.g. if it resolves to ~~localhost~~ a K8s service. 
As you've explained above, my assumption that the result of `getExternalBaseUrl` is accessible on end-users devices, e.g. if behind a SSOed reverse-proxy, does not hold neither.

I'm happy to continue discussing a way forward, whether its DiscoveryService clarifications, potential changes to the DiscoveryServices, config/overrides for this specific issue or just closing the PR :)

modethirteen: > What is your `backend.baseUrl` set as in config? That should be your internal domain like `my-company.internal.com`

Hello, I'm the original reporter. Our `backend.baseUrl` is set to an externally reachable url (e.x. `portal.companyname.com`).

benjdlambert:
Gonna merge this, the failing e2e is unrelated.

PR #33111, opened 2026-03-04
fix(catalog-backend-module-gitlab): correct GitLab API parameter from 'topics' to 'topic'

### LINKED ISSUE #33110: GitlabDiscoveryEntityProvider provides an undocumented filtering option on project topics, but it does not work
### 📜 Issue Labels

### 🔎 Search Terms

```plain
gitlab, entity provider, filtering, topic
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

GitLab

### 📝 Description & Context

The following PR (which has been merged last year) https://github.com/backstage/backstage/pull/29183, proposed a very useful way for the `GitlabDiscoveryEntityProvider` to filter GitLab projects based on **topics**, during discovery.

Unfortunately it does not work, because the provider makes a call to the GitLab API and passes a URL parameter named `topics` (plural), whereas GitLab expects a parameter named `topic` (singular).

### 👍 Expected Behavior

Configuring the provider with `topics` should allow us to filter the candidate projects for the discovery process.

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

_No response_

AmateurMind:
## Summary
Fixes #33110

The `GitlabDiscoveryEntityProvider` passes the query parameter `topics` (plural) to the GitLab Projects API when filtering projects by topic. However, the GitLab API expects the parameter name `topic` (singular), causing the filter to be ignored.

## Changes
- **GitlabDiscoveryEntityProvider.ts**
  - Updated the API query parameter from `topics` to `topic` when calling `listProjects`.

- **client.ts**
  - Updated the `ListProjectOptions` interface to use `topic` instead of `topics` to match the GitLab API parameter.

## Root Cause
The original implementation introduced topic filtering but used `topics` (plural) as the query parameter name.

The GitLab API endpoint `GET /projects` expects `topic` (singular), so the filtering was not applied.

## Notes
This is a **non-breaking change**.

The internal configuration (`topics` in `app-config.yaml`) remains unchanged. Only the outbound API request parameter has been corrected.

---

## Checklist
- [ ] A changeset describing the change and affected packages
- [x] Screenshots attached (for UI changes) — N/A
- [x] All commits include a `Signed-off-by` line

### DIFF
--- a/.changeset/fix-gitlab-topic-filter.md
+++ b/.changeset/fix-gitlab-topic-filter.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-gitlab': patch
+---
+
+Fixed GitLab project topic filtering by using correct API parameter 'topic' instead of 'topics'
--- a/plugins/catalog-backend-module-gitlab/src/lib/client.ts
+++ b/plugins/catalog-backend-module-gitlab/src/lib/client.ts
@@ -41,7 +41,7 @@ interface ListProjectOptions extends CommonListOptions {
   archived?: boolean;
   group?: string;
   membership?: boolean;
-  topics?: string;
+  topic?: string;
   last_activity_after?: string;
 }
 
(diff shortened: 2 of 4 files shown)


freben:
Note the comments about prettier - you may want to enable format on save in your editor (or run `yarn prettier --write <directory>` from the command line)

wasadigi:
This PR is not stale, the proposed fix is correct (and we have an adopter waiting for this).

PR #33112, opened 2026-03-04
BUI - Add new range slider component

AmbrishRamachandiran:
## Summary

Adds a new `Slider` component to `@backstage/ui` for selecting numeric ranges with dual-thumb controls.

## Why

Plugin developers frequently need range selection for filtering (dates, prices, scores) but had to implement custom solutions. This component provides a reusable, accessible solution.

image

## What's Added

- **Slider component** with dual-thumb range selection
- Custom value formatting support (e.g., currency, units)
- Full keyboard navigation and accessibility
- Integration with FieldLabel for consistent form patterns
- 8 Storybook examples showcasing different use cases
- Complete documentation at `/components/range-slider`

## Usage

```tsx
<RangeSlider
  label="Budget Range"
  minValue={0}
  maxValue={10000}
  defaultValue={[2000, 8000]}
  showValueLabel
  formatValue={(v) => `$${v.toLocaleString()}`}
/>

### DIFF
--- a/docs-ui/src/utils/data.ts
+++ b/docs-ui/src/utils/data.ts
@@ -86,10 +86,6 @@ export const components: Page[] = [
     slug: 'radio-group',
   },
   {
-    title: 'SearchAutocomplete',
-    slug: 'search-autocomplete',
-  },
-  {
     title: 'SearchField',
     slug: 'search-field',
   },
--- a/packages/ui/report.api.md
+++ b/packages/ui/report.api.md
@@ -1162,6 +1162,7 @@ export const FieldLabelDefinition: {
     readonly description: {};
     readonly htmlFor: {};
     readonly id: {};
+    readonly descriptionId: {};
     readonly className: {};
   };
 };
@@ -1173,6 +1174,7 @@ export type FieldLabelOwnProps = {
   description?: string | null;
   htmlFor?: string;
   id?: string;
+  descriptionId?: string;
   className?: string;
 };
 
@@ -2287,20 +2289,6 @@ export interface SliderOwnProps {
 // @public (undocumented)
 export interface SliderProps<T extends number | number[]>
   extends Omit<SliderProps_2<T>, 'children' | 'className'>,
-    Omit<
-      FieldLabelProps,
-      | 'htmlFor'
-      | 'id'
-      | 'className'
-      | 'defaultValue'
-      | 'onChange'
-      | 'slot'
-      | 'style'
-      | 'label'
-      | 'secondaryLabel'
-      | 'description'
-      | 'isRequired'
-    >,
     SliderOwnProps {}
 
 // @public (undocumented)
--- a/packages/ui/src/definitions.ts
+++ b/packages/ui/src/definitions.ts
@@ -52,10 +52,6 @@ export { MenuDefinition } from './components/Menu/definition';
 export { PasswordFieldDefinition } from './components/PasswordField/definition';
 export { PopoverDefinition } from './components/Popover/definition';
 export { RadioGroupDefinition } from './components/RadioGroup/definition';
-export {
-  SearchAutocompleteDefinition,
-  SearchAutocompleteItemDefinition,
-} from './components/SearchAutocomplete/definition';
 export { SearchFieldDefinition } from './components/SearchField/definition';
 export { SelectDefinition } from './components/Select/definition';
 export { SkeletonDefinition } from './components/Skeleton/definition';

### CODE REVIEW

**Comment by AmbrishRamachandiran** [comment]
@cdedreuille @Naycon any thoughts? it will be good addition for the cost cloud plugins

**Comment by AmbrishRamachandiran** [comment]
@cdedreuille @Naycon @freben @Rugvip can you review?

File: packages/ui/src/components/RangeSlider/types.ts [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+      | 'defaultValue'
+      | 'onChange'
+      | 'slot'
+      | 'style'
+    > {
+  /**
+   * Whether to show a value label in the header next to the field label
+   * @defaultValue false
+   */
+  showValueLabel?: boolean;
```
cdedreuille: Do we really need that? I don't see any stories that does not have the `showValueLabel` set to true. Perhaps we can start without this prop.

packages/ui/src/components/RangeSlider/types.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * Format the value for display
+   */
+  formatValue?: (value: number) => string;
+}
```
cdedreuille: How is that different from the default formatOptions from React Aria?

packages/ui/src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -0,0 +1,210 @@
...
+import clsx from 'clsx';
+import { FieldLabel } from '../FieldLabel';
+import { FieldError } from '../FieldError';
+import type { RangeSliderProps } from './types';
+import { useDefinition } from '../../hooks/useDefinition';
+import { RangeSliderDefinition } from './definition';
+import styles from './RangeSlider.module.css';
+
+/** @public */
+export const RangeSlider = forwardRef<HTMLDivElement, RangeSliderProps>(
```
cdedreuille: I'm not sure about shipping `RangeSlider` instead of `Slider`. Ideally we would like to stick to React Aria pretty closely for our more atomic component. Could you please reformat this comment to take that into account as well as create stories with single thumb options?

packages/ui/src/components/RangeSlider/types.ts
AmbrishRamachandiran: Agreed and removed. The value label is now always displayed when a label is provided, simplifying the API.

packages/ui/src/components/RangeSlider/types.ts
AmbrishRamachandiran: Good catch! Removed formatValue in favor of React Aria's built-in formatOptions prop, which uses standard Intl.NumberFormat options. This provides better i18n support and follows React Aria's conventions. Updated stories to use formatOptions for currency, percentage, etc.

packages/ui/src/components/RangeSlider/RangeSlider.tsx
AmbrishRamachandiran: changes done updated rangeslider to slider and added example for single slider

packages/ui/src/components/Slider/types.ts
```diff
@@ -0,0 +1,46 @@
...
+  secondaryLabel?: FieldLabelProps['secondaryLabel'];
+  description?: FieldLabelProps['description'];
+  isRequired?: boolean;
+}
+
+/** @public */
+export interface SliderProps<T extends number | number[]>
+  extends Omit<AriaSliderProps<T>, 'children' | 'className'>,
+    Omit<
+      FieldLabelProps,
```
Naycon: This should not be here, since you're not passing any props apart from the ones you've declared in the own props above on to FieldLabel. You'd only add an inheritance here to get props added to `restProps`, and you typically do this when you intend your component to be a RA component or native html element, and you want to enable users to pass on props to that element. You should remove the whole FieldLabelProps omit type here (keep what you have in own props though).

AmbrishRamachandiran:
@cdedreuille can you take a look and let me know any changes required

packages/ui/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,125 @@
...
+      ref={ref}
+    >
+      {label && (
+        <div className={classes.header}>
+          <FieldLabel
+            label={label}
+            secondaryLabel={secondaryLabelText}
+            description={description}
+          />
+          <SliderOutput className={classes.output}>
```
Naycon, line 69: @cdedreuille I'm not sure about this from a UX/design perspective. In my mind, the field labels are just for labelling a field, and should not be used to display the value of the component. This component is not functional without a label in its current state (if you don't provide a label you can't see the current value), but all the rest of our form components work with and without a label. 

I'd suggest we should find a design where the value is displayed independently of the label. Here's a couple of example of this, done with a11y in focus:

https://www.w3.org/WAI/ARIA/apg/patterns/slider/examples/slider-color-viewer/
https://www.w3.org/WAI/ARIA/apg/patterns/slider-multithumb/examples/slider-multithumb/

Displaying the values like this, above the handle(s) gives users using magnification a way to see what value they are inputing, and it'd also let us use the component without a visual label if needed.

WDYT?

packages/ui/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const secondaryLabelText = secondaryLabel || (isRequired ? 'Required' : null);
+
+  return (
+    <AriaSlider
+      className={clsx(classes.root, className)}
+      aria-label={ariaLabel}
+      aria-labelledby={ariaLabelledBy}
+      defaultValue={defaultValue}
+      value={value}
+      {...restProps}
```
Naycon: You can remove all of this apart from the restProps spread - this is the preferred way to pass on variables unless there's a reason you need to change/enhance what you pass on:

```suggestion
      {...restProps}
```

You'll have to remove the `defaultValue` and `value` variables above, since they're not used anywhere.

packages/ui/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,125 @@
...
+      ref={ref}
+    >
+      {label && (
+        <div className={classes.header}>
+          <FieldLabel
```
Naycon, line 62: Does RA automatically associate this label with the slider? I haven't looked into how it behaves, but this should be marked as labelling the slider in one way or another so a11y users understand what the slider is for if `label` is used.

As a bonus, it'd be nice if the description is also picked up (just be aware that aria-labelledby and aria-describedby accepts element id's, not text).

packages/ui/src/components/Slider/types.ts
AmbrishRamachandiran: Fixed - removed the FieldLabelProps inheritance since we're not passing those props through to restProps.

packages/ui/src/components/Slider/Slider.tsx
AmbrishRamachandiran: Fixed - removed the explicit prop spreads and unused variables. Now only using {...restProps}.

packages/ui/src/components/Slider/Slider.tsx
AmbrishRamachandiran: Fixed - added proper label and description association using aria-labelledby and aria-describedby. Generated unique IDs with useId() and extended FieldLabel to support descriptionId prop for the aria-describedby reference.

AmbrishRamachandiran:
@Naycon fixed all review comments other than -> https://github.com/backstage/backstage/pull/33112#discussion_r2917270566

AmbrishRamachandiran:
@cdedreuille @Naycon any changes needed?

packages/ui/src/components/Slider/Slider.tsx
cdedreuille: From a design perspective, even though I agree that it is clearer to see the value next to the thumb I think it looks cleaner in having it on the right side as proposed. The stories don't reflect the reality too much as you will not have sliders that long.

So as a first version I think this version is fine.

cdedreuille:
@AmbrishRamachandiran Thanks a lot for all the updates. On my side I think this component is ready to go. If you could fix the conflicts, I would be happy to approve it.

AmbrishRamachandiran:
> @AmbrishRamachandiran Thanks a lot for all the updates. On my side I think this component is ready to go. If you could fix the conflicts, I would be happy to approve it.

@cdedreuille resolved the merge conflicts any chances to get it approved?

AmbrishRamachandiran:
@cdedreuille any chances to get it reviewed & approved?

AmbrishRamachandiran:
@Naycon can you please take a look?

PR #33124, opened 2026-03-04
feat(scaffolder): create scaffolder mcp action to execute a template

yangcao77:

This PR added scaffolder mcp action to execute a template.
it calls the `scaffolderService.scaffold` API, with provided `templateRef` and `input values`. it will start a task to execute the template, and returns the `taskId`.
`scaffolderService` already receives credentials from the MCP context representing the calling user, so passing it straight through as { credentials } means:
The request is executed on behalf of the user with their identity
The scaffolder backend applies RBAC checks against that identity the same way it would for a direct API call

see the screenshot & repo created when calling the mcp action: https://github.com/yangcao77/test-260304-mcp
Screenshot 2026-03-04 at 5 09 20 PM

Screenshot 2026-03-04 at 5 11 23 PM

### DIFF
--- a/plugins/scaffolder-backend/src/actions/createExecuteTemplateAction.ts
+++ b/plugins/scaffolder-backend/src/actions/createExecuteTemplateAction.ts
@@ -34,7 +34,8 @@ export const createExecuteTemplateAction = ({
     },
     description: `Executes a Scaffolder template with its template ref and input parameter values.
 The template is run using the credentials provided to this action, and respects any RBAC permissions associated with those credentials.
-Returns a taskId that can be used to track execution progress.`,
+Returns a taskId that can be used to track execution progress.
+Use the catalog.get-catalog-entity action to fetch the Template entity and discover its parameter schema and secrets definition before calling this action.`,
     schema: {
       input: z =>
         z.object({
@@ -44,24 +45,16 @@ Returns a taskId that can be used to track execution progress.`,
               'The template entity reference to execute, e.g. "template:default/my-template"',
             ),
           values: z
-            .record(
-              z.string(),
-              z.union([
-                z.string(),
-                z.number(),
-                z.boolean(),
-                z.null(),
-                z.array(z.unknown()),
-                z.record(z.unknown()),
-              ]),
-            )
+            .record(z.unknown())
             .describe(
-              'Input parameter values required by the template. Collect all required template parameters from the user before calling this action. e.g. { "repoUrl": "github.com?owner=my-org&repo=my-repo", "description": "My new service" }',
+              'Input parameter values required by the template. Use catalog.get-catalog-entity to discover the required parameters for the template.',
             ),
           secrets: z
             .record(z.string())
             .optional()
-            .describe('Optional secrets to pass to the template execution.'),
+            .describe(
+              'Optional secrets to pass to the template execution. Use catalog.get-catalog-entity to discover the secrets definition on the Template entity.',
+            ),
         }),
       output: z =>
         z.object({
--- a/.changeset/cold-bikes-beam.md
+++ b/.changeset/cold-bikes-beam.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-scaffolder-backend': minor
 ---
 
-create a new `execute-template` action to start a task that execute a template with provided values.
+Added a new `execute-template` actions registry action that executes a scaffolder template with provided input values and returns a task ID for tracking progress.
--- a/plugins/scaffolder-backend/src/actions/createExecuteTemplateAction.test.ts
+++ b/plugins/scaffolder-backend/src/actions/createExecuteTemplateAction.test.ts
@@ -54,9 +54,42 @@ describe('createExecuteTemplateAction', () => {
     );
   });
 
-  it('should pass the caller credentials to the scaffolder service', async () => {
+  it('should forward empty values and pass secrets to the scaffolder service', async () => {
     const mockActionsRegistry = actionsRegistryServiceMock();
-    mockScaffolderService.scaffold.mockResolvedValue({ taskId: 'task-xyz' });
+    mockScaffolderService.scaffold.mockResolvedValue({
+      taskId: 'task-with-secrets',
+    });
+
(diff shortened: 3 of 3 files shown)


plugins/scaffolder-backend/src/actions/createExecuteTemplateAction.ts
```diff
@@ -0,0 +1,73 @@
...
+        z.object({
+          templateRef: z
+            .string()
+            .describe(
+              'The template entity reference to execute, e.g. "template:default/my-template"',
+            ),
+          values: z
+            .record(z.unknown())
+            .describe(
+              'Input parameter values required by the template. Collect all required template parameters from the user before calling this action.',
```
johnmcollier: It might be good to include the expected format of the input parameter key/values

johnmcollier:
Based on earlier discussions in https://github.com/backstage/backstage/issues/31755#issuecomment-3541599051, we might also need https://github.com/backstage/backstage/pull/32320 since it seems some secrets required by templates are populated by the frontend

plugins/scaffolder-backend/src/actions/createExecuteTemplateAction.ts
yangcao77: the value here is expected to be `Record<string, JsonValue>`.
I've pushed a new commit to enhance the format by making the key type explicit and the value a proper JSON-compatible union instead of z.unknown().

yangcao77:
> Based on earlier discussions in https://github.com/backstage/backstage/issues/31755#issuecomment-3541599051, we might also need https://github.com/backstage/backstage/pull/32320 since it seems some secrets required by templates are populated by the frontend

pushed a new commit to include optional `secret` as this action's input

plugins/scaffolder-backend/src/actions/createExecuteTemplateAction.ts
johnmcollier: Sorry, what I meant was, for users of the action, it's not immediately clear how they need to format the input parameters when calling the action. I think an example might help in this case, like you did for `templateRef`

yangcao77: updated

plugins/scaffolder-backend/src/actions/createExecuteTemplateAction.ts
```diff
@@ -0,0 +1,88 @@
...
+              ]),
+            )
+            .describe(
+              'Input parameter values required by the template. Collect all required template parameters from the user before calling this action. e.g. { "repoUrl": "github.com?owner=my-org&repo=my-repo", "description": "My new service" }',
+            ),
+          secrets: z
+            .record(z.string())
+            .optional()
+            .describe('Optional secrets to pass to the template execution.'),
+        }),
```
drodil, line 58: I think it would add value to be able to wait the task to finish when using this action so that the agent calling this MCP tool would also get the output of the task. This could be optional with additional input `waitForCompletion` as optional boolean. We actually already have this tool in our own instance so I can share the wait part here if you think this is good idea:

```ts
     const backendBaseUrl = await discovery.getBaseUrl('scaffolder');
      const requestUrl = `${backendBaseUrl}/v2/tasks`;

      const { token } = await auth.getPluginRequestToken({
        onBehalfOf: credentials,
        targetPluginId: 'scaffolder',
      });

      const response = await jsonRequest<{ id: string }>(requestUrl, {
        bearerAuth: token,
        method: 'POST',
        throwOnError: true,
        data: {
          templateRef: input.templateRef,
          values: input.params,
        },
      });

      if (!response) {
        throw new Error('Failed to create scaffolder task');
      }

      const link = `${appBaseUrl}/scaffolder/tasks/${response.id}`;
      if (!input.waitForCompletion) {
        return { output: { id: response.id, link } };
      }

      const maxWaitTime = 300_000; // Maximum wait time of 5 minutes
      const sleepInterval = 5000; // Polling interval of 5 seconds
      let maxRetries = Math.floor(maxWaitTime / sleepInterval);

      const taskUrl = `${backendBaseUrl}/v2/tasks/${response.id}`;
      let taskResponse: ScaffolderTask | null = null;

      while (maxRetries--) {
        await sleep(sleepInterval);
        taskResponse = await jsonRequest<ScaffolderTask>(taskUrl, {
          bearerAuth: token,
          method: 'GET',
          throwOnError: true,
        });

        if (
          taskResponse &&
          taskResponse.status !== 'open' &&
          taskResponse.status !== 'processing'
        ) {
          break;
        }
      }

      return {
        output: {
          id: response.id,
          link,
          createdAt: taskResponse?.createdAt,
          spec: taskResponse?.spec,
          status: taskResponse?.status,
        },
      };
    },
```

yangcao77: Agree that the ability to wait for response is going to be helpful!
however, I'm thinking if it is better to introduce the `waitForCompletion` in the endpoint itself.

awanlin:
Hi @yangcao77, can you rebase and add this to the list of well-known actions here, please? https://github.com/backstage/backstage/blob/master/docs/ai/well-known-actions.md

yangcao77:
> Hi @yangcao77, can you rebase and add this to the list of well-known actions here, please? https://github.com/backstage/backstage/blob/master/docs/ai/well-known-actions.md

@awanlin Done. pushed a new commit to include `execute-template` in scaffolder's mcp actions list

benjdlambert:
@yangcao77 sorry this needs another rebase :(

yangcao77:
@benjdlambert Just pushed a new commit to resolve the conflict.

yangcao77:
@benjdlambert I have resolved all conflicts, PTAL when you have time. Thanks.

benjdlambert:
Hey @yangcao77 sorry, been a bit busy with KubeCon this week. Will have a look as soon as I get a minute :pray:

benjdlambert (approved):
Hey @yangcao77!

I went and just tidied this up a little bit, but I think that this is looking good! :tada:

Let's ship this and iterate :tada:

PR #33125, opened 2026-03-04
feat(cli): add `backstage-cli package bundle` command for plugin dynamic loading 

davidfestal:
### Add `backstage-cli package bundle` command for dynamic plugin loading

Adds a new `package bundle` CLI command that produces self-contained plugin bundles deployable independently and loadable dynamically by a Backstage application. The bundles are designed to be consumed at runtime by [`@backstage/backend-dynamic-feature-service`](/backstage/backstage/tree/master/packages/backend-dynamic-feature-service) (backend plugins) and [`@backstage/frontend-dynamic-feature-loader`](/backstage/backstage/tree/master/packages/frontend-dynamic-feature-loader) (frontend plugins via Module Federation).

**Backend plugins** — Builds CJS output via [`createDistWorkspace`](/backstage/backstage/tree/master/packages/cli/src/modules/build/lib/packager/createDistWorkspace.ts), packs the main plugin and its local `workspace:^`/`backstage:^` dependencies into a flat structure (main plugin at root, transitive local deps under `embedded/`), seeds and prunes a lockfile, and installs a private `node_modules`.

**Frontend plugins** — Produces a Module Federation remote build via [`buildFrontend`](/backstage/backstage/tree/master/packages/cli/src/modules/build/lib/buildFrontend.ts). Only the main plugin is packed (no `embedded/`, no lockfile, no `node_modules`). The output `package.json` is adjusted so `main` points to `remoteEntry.js` and `types` to `@mf-types/index.d.ts` when available; misleading fields (`exports`, `module`, `typesVersions`) are removed.

**Batch bundling** — A `--pre-packed-dir` option accepts the output of `backstage-cli build-workspace <dir> --alwaysPack` to avoid redundant `createDistWorkspace` calls when bundling many plugins from the same monorepo.

Key options: `--output-destination`, `--output-name`, `--clean`, `--no-build`, `--no-install`, `--verbose`.

Also includes a companion fix in [`@backstage/backend-dynamic-feature-service`](/backstage/backstage/tree/master/packages/backend-dynamic-feature-service): the [`CommonJSModuleLoader`](/backstage/backstage/tree/master/packages/backend-dynamic-feature-service/src/loader/CommonJSModuleLoader.ts) resolution fallback for `resolvePackagePath` now also handles the case where [`@backstage/backend-plugin-api`](/backstage/backstage/tree/master/packages/backend-plugin-api) is bundled inside a dynamic plugin's own `node_modules`, rather than only when it lives in the host application.

### DIFF
--- a/docs/tooling/cli/03-commands.md
+++ b/docs/tooling/cli/03-commands.md
@@ -280,32 +280,35 @@ Options:
                               plugins, this also enables yarn.lock generation for SBOM.
 ```
 
-### Output Structure
+### Output Contract
 
-The bundle is created in a directory named after the package. For example,
-`@myorg/plugin-foo` creates `myorg-plugin-foo/` with the following structure:
+The bundle output is a directory that can be deployed as a standalone unit.
+Consumers of the bundle (such as `@backstage/backend-dynamic-feature-service`
+or `@backstage/frontend-dynamic-feature-loader`) can rely on the following
+guarantees:
 
-**Backend plugins:**
+**All bundles:**
 
-```text
-myorg-plugin-foo/
-├── package.json              # Customized for dynamic loading
-├── dist/                     # Built plugin code
-│   ├── index.cjs.js
-│   └── .config-schema.json   # Config schema (local packages only)
-├── embedded/                 # Embedded workspace packages (if any)
-│   └── myorg-plugin-foo-common/
-│       ├── package.json
-│       └── dist/
-└── node_modules/             # All production dependencies
-```
+- A `package.json` at the bundle root with entry points configured for dynamic
+  loading. The `backstage.role` and `files` fields are preserved from the source package.
+- A `dist/` directory containing the built plugin code.
+- A `dist/.config-schema.json` file (when any config schemas apply) containing
+  gathered schemas from the plugin, its local workspace dependencies, and
+  third-party dependencies. Schemas from unrelated Backstage packages are excluded.
+- No `scripts` or `devDependencies` in `package.json`.
 
-**Frontend plugins** produce module federation assets at the bundle root (no `embedded/` or `node_modules/` unless `--pre-packed-dir` is used).
+**Backend plugins** (`backend-plugin`, `backend-plugin-module`):
 
-The `.config-schema.json` contains merged config schemas from the plugin and any
-embedded workspace packages. It excludes schemas from npm dependencies, as those
-should be provided by the host application. The `package.json` is updated to
-reference this generated JSON file instead of any original `.d.ts` schema.
+- A `node_modules/` directory with all production dependencies (including local
+  workspace dependencies), pinned to their exact versions from the source lockfile.
+- `bundleDependencies` is set to `true` in `package.json`.
+
+**Frontend plugins** (`frontend-plugin`, `frontend-plugin-module`):
+
+- `main` points to `dist/remoteEntry.js` (the Module Federation remote entry).
+- `types` points to `dist/@mf-types/index.d.ts` when type declarations are
+  available.
+- No embedded `node_modules/` directory.
 
 ### Environment Variables
 


packages/cli/src/modules/build/commands/package/bundle/command.ts
```diff
@@ -0,0 +1,878 @@
...
+  };
+
+  const logRunOutput = (data: Buffer) => {
+    if (prefix) {
+      for (const line of data.toString('utf8').split(/\r?\n/)) {
+        if (line) writeLine(line, 'out');
+      }
+    } else {
+      logStream.write(stripAnsi(data.toString('utf8')));
+      if (verbose) process.stdout.write(chalk.dim(data.toString('utf8')));
```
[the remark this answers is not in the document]
davidfestal: Fixed

packages/cli-module-build/src/commands/package/bundle/command.ts
```diff
@@ -0,0 +1,878 @@
...
+    const yarnrcLines = ['nodeLinker: node-modules'];
+    try {
+      // Include yarnPath so the same Yarn version that created the lockfile
+      // is used for pruning/installing -- lockfile formats differ across
+      // major Yarn versions (e.g. ~builtin vs optional!builtin patches).
+      const resolved = await runOutput(['yarn', 'config', 'get', 'yarnPath'], {
+        cwd: targetPaths.rootDir,
+      });
+      if (resolved) {
+        yarnrcLines.push(`yarnPath: ${resolved}`);
```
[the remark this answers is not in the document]
davidfestal: Fixed

packages/backend-dynamic-feature-service/README.md
```diff
@@ -136,35 +136,47 @@ yarn install  # Installs all the plugin's dependencies
...
 
-- Systematic use of main application's `@backstage/*` packages (no version conflicts), enabling the future implementation of `@backstage` dependency version checking at start time
-- Self-contained packages with only necessary private dependencies
-- No post-installation steps required (extract and run)
-- Consistent dependency structure across all dynamic plugins
-- Production-ready distribution format
+```bash
+cd my-backstage-plugin
+yarn backstage-cli package bundle --output-destination /path/to/dynamic-plugins-root
+# Creates a self-contained bundle in the /path/to/dynamic-plugins-root/my-backstage-plugin/ sub-folder
```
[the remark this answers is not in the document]
davidfestal: No, if the package name is the same as the plugin folder name (without a scope), this example is valid and simple enough. Let's not introduce package name mangling here, and make the example more complicated. details of output naming is already explain in docs and command help.

packages/cli-module-build/src/commands/package/bundle/command.ts
```diff
@@ -0,0 +1,886 @@
...
+  if (needsDependencies) {
+    // ── Step 1: Populate embedded packages ──────────────────────────────
+    const embeddedDir = joinPath(target, 'embedded');
+    let targets: { name: string; dir: string }[];
+
+    if (opts.prePackedDir) {
+      // ── Strategy A: Copy from pre-built dist workspace ───────────────
+      // Reuses output from `backstage-cli build-workspace --alwaysPack`.
+      // Works for both backend (performance optimization) and frontend
+      // (enables lockfile generation for SBOM).
```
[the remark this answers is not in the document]
davidfestal: Fixed

packages/backend-dynamic-feature-service/src/loader/CommonJSModuleLoader.ts
```diff
@@ -99,15 +99,17 @@ export class CommonJSModuleLoader implements ModuleLoader {
...
-      // (this is mostly done by calling `@backstage/backend-plugin-api/resolvePackagePath`).
-      const resolvingPackageJsonFromBackstageApplication =
-        request?.endsWith('/package.json') &&
-        mod?.path &&
-        !dynamicPluginsPaths.some(p => mod.path.startsWith(p));
+      // Is this a `resolvePackagePath` call from `@backstage/backend-plugin-api`?
+      // This covers both the host application's copy and a bundled copy
+      // living inside a dynamic plugin's own node_modules.
+      const isFromBackendPluginApi = mod?.path?.includes(
+        path.join('node_modules', '@backstage', 'backend-plugin-api'),
```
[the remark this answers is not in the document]
davidfestal: Fixed

packages/cli-module-build/src/lib/packager/createDistWorkspace.ts
```diff
@@ -280,11 +284,40 @@ const FAST_PACK_SCRIPTS = [
...
+  await fs.ensureDir(targetDir);
+  await run(['yarn', 'pack', '--filename', archivePath], {
+    cwd: packageDir,
+    onStdout: prefixLogFunc(prefix, logger.log),
+    onStderr: prefixLogFunc(prefix, logger.warn),
+  }).waitForExit();
+
+  await tar.extract({ file: archivePath, cwd: targetDir, strip: 1 });
+  await fs.remove(archivePath);
+}
```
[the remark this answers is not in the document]
davidfestal: The original code on master has the exact same pattern: `tar.extract` followed by `fs.remove(archivePath)` with no `try`/`finally` wrapping. The extracted `packToDirectory` function preserves the same behavior verbatim.

packages/backend-dynamic-feature-service/src/features/features.test.ts
```diff
@@ -465,6 +465,45 @@ Require stack:
     });
   });
 
+  it('should load a backend plugin that bundles its own @backstage/backend-plugin-api', async () => {
+    const dynamicPLuginsLister = new DynamicPluginLister();
+    const dynamicPluginsRootForBundled = resolvePath(
+      __dirname,
+      '__fixtures__/dynamic-plugins-root-for-bundled',
```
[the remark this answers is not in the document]
davidfestal: Fixed

docs/tooling/cli/03-commands.md
```diff
@@ -204,6 +205,121 @@ Options:
...
+  --verbose                   Stream detailed output from internal steps (build, pack,
+                              install) to the console. Without this flag, output is
+                              captured to per-step log files and only shown on error.
+  --pre-packed-dir <dir>      Path to a pre-built dist workspace (from
+                              build-workspace --alwaysPack). Skips local dependency
+                              packing and uses pre-packed packages directly. For frontend
+                              plugins, this also enables yarn.lock generation for SBOM.
+```
+
+### Output Structure
```
Rugvip: note from wg: update this to something like "Output Expectations/Contract" and be specific about what we guarantee but remove some of the implementation details.

davidfestal: Fixed in  commit 335b2caef5b8ab8d5d6a4419050339a3e0c9782e

packages/cli-module-build/src/commands/package/bundle/command.test.ts
```diff
@@ -0,0 +1,1074 @@
...
+  };
+});
+
+jest.mock('@backstage/config-loader', () => ({
+  loadConfigSchema: (...args: unknown[]) => mockLoadConfigSchema(...args),
+}));
+
+jest.mock('node:module', () => ({
+  createRequire: (...args: unknown[]) => mockCreateRequire(...args),
+}));
```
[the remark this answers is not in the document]
davidfestal: Fixed

davidfestal:
@Rugvip In addition to fixing [your comment](https://github.com/backstage/backstage/pull/33125#pullrequestreview-3898910553), I also added the `experimental` flag and updated the docs accordingly in commit c39abaeea1149d4a315b88d2350f7147efadbb2e

packages/cli-module-build/src/commands/package/bundle/command.ts
```diff
@@ -0,0 +1,985 @@
...
+    const mainPluginRelDir = relativePath(targetPaths.rootDir, targetPaths.dir);
+    const mainPluginEmbeddedDir = resolvePath(embeddedDir, mainPluginRelDir);
+
+    console.log(
+      chalk.blue(
+        `Moving main plugin ${chalk.cyan(pkg.name)} to bundle root...`,
+      ),
+    );
+
+    const mainPluginEntries = await fs.readdir(mainPluginEmbeddedDir);
```
schultzp2020, line 332: When the main plugin isn't found in the pre-packed dir, the warning above fires but execution continues here, where `readdir` throws an ENOENT that doesn't explain the real cause.

Would an early check help?

```suggestion

    if (!(await fs.pathExists(mainPluginEmbeddedDir))) {
      throw new Error(
        `Main plugin ${chalk.cyan(pkg.name)} was not found in the ` +
          `embedded workspace at ${chalk.cyan(mainPluginRelDir)}. ` +
          `Ensure the pre-packed workspace includes this plugin.`,
      );
    }

    const mainPluginEntries = await fs.readdir(mainPluginEmbeddedDir);
```

packages/cli-module-build/src/commands/package/bundle/command.ts
```diff
@@ -0,0 +1,985 @@
...
+    targetPkg.bundleDependencies = true;
+    const patchVersionPattern = /^patch:.+@npm%3A([^#]+)#/;
+    const stripped: Record<string, string> = {};
+    if (rootResolutions) {
+      for (const [key, value] of Object.entries(rootResolutions)) {
+        if (typeof value !== 'string') {
+          continue;
+        }
+        const patchMatch = value.match(patchVersionPattern);
+        stripped[key] = patchMatch ? patchMatch[1] : value;
```
schultzp2020, line 619: This silently drops `patch:` protocol resolutions to bare versions. If the monorepo patches a dep for a security fix, the bundle gets the unpatched version with no indication.

Would a warning help operators catch this?

```suggestion
        const patchMatch = value.match(patchVersionPattern);
        if (patchMatch) {
          console.warn(
            chalk.yellow(
              `  Resolution ${chalk.cyan(key)} uses patch: protocol \u2014 ` +
                `stripping to ${chalk.cyan(patchMatch[1])}. ` +
                `Patch files are not included in the bundle.`,
            ),
          );
        }
        stripped[key] = patchMatch ? patchMatch[1] : value;
```

packages/cli/src/modules/build/commands/package/bundle/command.ts
```diff
@@ -0,0 +1,985 @@
...
+  await fs.writeFile(joinPath(target, '.gitignore'), '*\n');
+  await fs.writeFile(joinPath(target, '.bundle-output'), '');
+
+  const rootPkg = await fs.readJson(
+    resolvePath(targetPaths.rootDir, 'package.json'),
+  );
+
+  // Backend plugins always need embedded packages, lockfile, and node_modules.
+  // Frontend plugins need them only when --pre-packed-dir is provided.
+  const needsDependencies = !isFrontend || !!opts.prePackedDir;
```
schultzp2020: `needsDependencies` controls both lockfile operations (which frontend + `--pre-packed-dir` needs for SBOM) and `bundleDependencies: true` in package.json (which is backend-only per the output contract docs).

With `--pre-packed-dir` on a frontend plugin, the resulting package.json gets `bundleDependencies: true` even though there's no `node_modules/`. Is that the intended behavior, or should this be split into two concerns?

docs/tooling/cli/03-commands.md
```diff
@@ -204,6 +204,129 @@ Options:
...
+guarantees:
+
+**All bundles:**
+
+- A `package.json` at the bundle root with entry points configured for dynamic
+  loading. The `backstage.role` and `files` fields are preserved from the source package.
+- A `dist/` directory containing the built plugin code.
+- A `dist/.config-schema.json` file (when any config schemas apply) containing
+  gathered schemas from the plugin and its local workspace dependencies. Schemas
+  from npm dependencies are excluded.
```
schultzp2020: nit: `filterBundleConfigSchemas` actually includes third-party deps without backstage metadata (and the test asserts this). Might be worth aligning:

```suggestion
  gathered schemas from the plugin, its local workspace dependencies, and
  third-party dependencies. Schemas from unrelated Backstage packages are excluded.
```

packages/cli-module-build/src/commands/package/bundle/command.ts
davidfestal: Yes, that's a valuable enhancement. Thanks !

packages/cli-module-build/src/commands/package/bundle/command.ts
davidfestal: The `patch:` stripping is intentional.

The actual patched version should be preserved through the seeded `yarn.lock` (copied from the monorepo), which already contains the resolved patched artifact.
The resolutions copied from the monorepo root `package.json` into the bundle's `package.json` only guides the bundle lockfile pruning, which only removes unused entries without re-resolve existing ones.

So the stripped resolution is purely a hint for the pruning step, and the actual patched version flows through the lockfile. A warning would be misleading -- it would suggest something is lost when nothing actually is.

packages/cli/src/modules/build/commands/package/bundle/command.ts
davidfestal: Fixed by removing the `bundledDependencies: true` for frontend plugins.

docs/tooling/cli/03-commands.md
davidfestal: Fixed

packages/cli-module-build/src/commands/package/bundle/command.ts
davidfestal: Fixed

davidfestal:
@Rugvip Any idea whether this PR could be merged, knowing that the provided command is now marked experimental mode and hidden by default ?

packages/cli-module-build/src/commands/package/bundle/command.ts
```diff
@@ -0,0 +1,979 @@
...
+ * When `--pre-packed-dir` is provided, local dependencies are copied from a
+ * pre-built dist workspace instead of calling `createDistWorkspace`:
+ * - For backend plugins this is a performance optimization when many plugins from the same monorepo are being bundled.
+ * - For frontend plugins it additionally enables lockfile generation (seed + prune) for dependency tracking purposes such as SBOM generation.
+ * - The pre-built dist workspace is produced by
+ *   `backstage-cli build-workspace <output-dir> [packages...] --alwaysPack`
+ *   and `<output-dir>` is then passed as `--pre-packed-dir`.
+ * - The `--alwaysPack` flag is required so that `workspace:^` and `backstage:^`
+ *   dependency specs are resolved to concrete versions in the packed output.
+ */
```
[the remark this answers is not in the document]
davidfestal: Seems already explained in the current comment. We don't need to make the comment any longer.

.changeset/cli-package-bundle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli': minor
```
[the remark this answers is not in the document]
davidfestal: Fixed

Issue #33104, opened 2026-03-03
`github:actions:dispatch` Return workflow run ID from dispatch action


abdellahhn:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
workflow dispatch run id, github actions dispatch return run id, scaffolder action workflow run url
```

### 🗃️ Project Area

Software Templates

### 🔖 Need

Currently, the `github:actions:dispatch` scaffolder action triggers a GitHub Actions workflow but discards the response entirely, the run ID is never surfaced as action outputs. This makes it impossible for downstream template steps to:

- Poll the triggered run for status/completion
- Link users to the workflow run in Backstage UI
- Joi dependent steps that depend on the dispatched run finishing

GitHub's API now supports returning run details directly in the dispatch response by passing `return_run_details: true`. This means we no longer need a separate polling step to discover the run ID, we can get it immediately.

### 📝 Proposal

Modify `createGithubActionsDispatchAction` to:

1. Add an optional boolean input `returnWorkflowRunDetails` (default: `false`) to the action schema.
2. When `true`, pass `return_run_details: true` to `client.rest.actions.createWorkflowDispatch(...)`.
3. Expose the response as action outputs: `workflowRunId`, `workflowRunUrl`.

```Javascript
schema: {
  input: {
    // ...existing inputs...
    returnWorkflowRunDetails: z =>
      z.boolean({
        description: 'If true, returns the workflow run ID and URLs as action outputs.',
      }).optional(),
  },
  output: {
    workflowRunId: z => z.number({ description: 'The triggered workflow run ID' }).optional(),
    workflowRunUrl: z => z.string({ description: 'API URL of the workflow run' }).optional(),
  },
},

// In handler:
const result = await client.rest.actions.createWorkflowDispatch({
  owner,
  repo,
  workflow_id: workflowId,
  ref: branchOrTagName,
  inputs: workflowInputs,
  return_run_details: returnWorkflowRunDetails ?? false,
});

if (returnWorkflowRunDetails && result.data) {
  ctx.output('workflowRunId', result.data.workflow_run_id);
  ctx.output('workflowRunUrl', result.data.run_url);
}
```

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

[...]

PR #33132, opened 2026-03-05
feat(scaffolder): add returnWorkflowRunDetails to github:actions:dispatch action

abdellahhn:

Fixes https://github.com/backstage/backstage/issues/33104

GitHub recently added a r`eturn_run_details` parameter to the [workflow dispatch API](https://github.blog/changelog/2026-02-19-workflow-dispatch-api-now-returns-run-ids/) that returns the run ID and URLs immediately on dispatch. This adds an optional `returnWorkflowRunDetails` input to the `github:actions:dispatch` action to expose this, removing the need for fragile timestamp-based polling workarounds.

When `returnWorkflowRunDetails` is true, three new outputs are available: `workflowRunId`, `workflowRunUrl`, and `workflowRunHtmlUrl`.

On the type cast: `@octokit/openapi-types` still types this endpoint as `OctokitResponse<never, 204>` because it hasn't picked up the updated GitHub OpenAPI spec yet. A cast with as unknown as is used to access the response data at runtime. This is safe since Octokit doesn't enforce or strip responses, it simply wraps the HTTP call. The cast is documented with a comment and can be removed once `@octokit/openapi-types` ships the updated spec.

Tested locally against a real GitHub repository using a Backstage scaffolder template with `returnWorkflowRunDetails: true`, the action successfully returned the run ID and URLs as outputs. ✅

### DIFF
--- a/plugins/scaffolder-backend-module-github/src/actions/githubActionsDispatch.ts
+++ b/plugins/scaffolder-backend-module-github/src/actions/githubActionsDispatch.ts
@@ -168,7 +168,11 @@ export function createGithubActionsDispatchAction(options: {
           ctx.output('workflowRunUrl', runDetails.workflowRunUrl);
           ctx.output('workflowRunHtmlUrl', runDetails.workflowRunHtmlUrl);
 
-          ctx.logger.info(`Workflow run url: ${runDetails.workflowRunHtmlUrl}`);
+          if (runDetails.workflowRunHtmlUrl) {
+            ctx.logger.info(
+              `Workflow run url: ${runDetails.workflowRunHtmlUrl}`,
+            );
+          }
         }
       } catch (e) {
         assertError(e);


plugins/scaffolder-backend-module-github/src/actions/githubActionsDispatch.ts
```diff
@@ -106,20 +122,59 @@ export function createGithubActionsDispatchAction(options: {
...
+            }
+            return null;
+          },
+        });
+
+        if (runDetails) {
+          ctx.output('workflowRunId', runDetails.workflowRunId);
+          ctx.output('workflowRunUrl', runDetails.workflowRunUrl);
+          ctx.output('workflowRunHtmlUrl', runDetails.workflowRunHtmlUrl);
+        }
```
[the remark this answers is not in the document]
abdellahhn: The `if (returnWorkflowRunDetails && response.data)` guard ensures null is only returned when no details are expected. The checkpoint key collision concern doesn't apply here since each workflow dispatch produces a unique run ID, so the same key will never be reused across different invocations.

.changeset/few-clouds-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend-module-github': patch
+---
+
+Added optional returnWorkflowRunDetails input to github:actions:dispatch action. When true, exposes workflowRunId, workflowRunUrl, and workflowRunHtmlUrl as outputs using the GitHub API return_run_details parameter.
```
awanlin:
```suggestion
Added optional `returnWorkflowRunDetails` input to `github:actions:dispatch` action. When true, exposes `workflowRunId`, `workflowRunUrl`, and `workflowRunHtmlUrl` as outputs using the GitHub API `return_run_details` parameter.
```

.github/vale/config/vocabularies/Backstage/accept.txt
```diff
@@ -581,3 +581,4 @@ resizable
 enums
 LLMs
 cleye
+return_run_details
```
awanlin: Remove this please, using backticks in the changeset is the better solution in this case 👍

plugins/scaffolder-backend-module-github/src/actions/githubActionsDispatch.ts
```diff
@@ -62,7 +62,14 @@ export function createGithubActionsDispatchAction(options: {
           z
             .record(z.string(), {
               description:
-                'Inputs keys and values to send to GitHub Action configured on the workflow file. The maximum number of properties is 10.',
+                'Inputs keys and values to send to GitHub Action configured on the workflow file. The maximum number of properties is 25.',
```
awanlin, line 65: Can you share docs confirming this changes from 10 to 25, please? Seems unrelated to your change.

abdellahhn: Of course🙂, here is the doc for it https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event (you can see it under the body parameters)

abdellahhn:
Thanks for the review @awanlin! Applied the comments.

plugins/scaffolder-backend-module-github/src/actions/githubActionsDispatch.ts
awanlin: Thanks for confirming this.

PR #33150, opened 2026-03-05
feat(ui): add analytics event tracking to BUI navigation components

Naycon:

Adds analytics event tracking to `@backstage/ui` components that have navigation behavior. When a user clicks a Link, ButtonLink, Tab, MenuItem, Tag, or Row that navigates, a `click` event is fired through Backstage's analytics system.

### What's included

**`@backstage/ui` — BUIProvider and analytics module**
- `BUIProvider` — a generic integration provider that accepts a `useAnalytics` hook (and can be expanded with additional capabilities in the future), backed by a single versioned `'bui'` context
- `useAnalytics` hook with noop default and dev-mode swap guard
- `getNodeText` utility for extracting event subjects from React children
- `defineComponent` / `useDefinition` integration — components opt in with `analytics: true` and get a tracker via the hook

**`@backstage/ui` — component integration**
- Link, ButtonLink, Tab, MenuItem, Tag, and Row fire `click` events with text subject and `{ to }` attribute
- All components accept a `noTrack` prop to suppress tracking
- Tab/MenuItem/Tag/Row only fire when the item has an `href`

**`@backstage/core-app-api` — old frontend system wiring**
- `BUIProvider` added to `AppManager`'s provider tree

**`@backstage/plugin-app` — new frontend system wiring**
- `BUIProvider` added directly in `AppRoot`

**`@backstage/ui` — bugfix**
- Fixed MenuItem `onAction` prop ordering so user-provided handlers are chained rather than overwritten

**Documentation**
- Analytics setup and `noTrack` usage documented in the installation guide
- `noTrack` prop added to all affected component prop tables

### DIFF
--- a/docs-ui/src/app/components/button-link/props-definition.tsx
+++ b/docs-ui/src/app/components/button-link/props-definition.tsx
@@ -65,7 +65,7 @@ export const buttonLinkPropDefs: Record<string, PropDef> = {
   noTrack: {
     type: 'boolean',
     description:
-      'Suppresses analytics event tracking for this component (if configured).',
+      'Suppresses the automatic analytics click event, e.g. if you already have custom tracking.',
   },
   onSurface: {
     type: 'enum',
--- a/docs-ui/src/app/components/link/props-definition.tsx
+++ b/docs-ui/src/app/components/link/props-definition.tsx
@@ -71,7 +71,7 @@ export const linkPropDefs: Record<string, PropDef> = {
   noTrack: {
     type: 'boolean',
     description:
-      'Suppresses analytics event tracking for this component (if configured).',
+      'Suppresses the automatic analytics click event, e.g. if you already have custom tracking.',
   },
   standalone: {
     type: 'boolean',
--- a/docs-ui/src/app/components/menu/props-definition.tsx
+++ b/docs-ui/src/app/components/menu/props-definition.tsx
@@ -312,7 +312,7 @@ export const menuItemPropDefs: Record<string, PropDef> = {
   noTrack: {
     type: 'boolean',
     description:
-      'Suppresses analytics event tracking for this component (if configured).',
+      'Suppresses the automatic analytics click event, e.g. if you already have custom tracking.',
   },
   textValue: {
     type: 'string',
--- a/docs-ui/src/app/components/table/props-definition.tsx
+++ b/docs-ui/src/app/components/table/props-definition.tsx
@@ -459,7 +459,7 @@ export const rowPropDefs: Record<string, PropDef> = {
   noTrack: {
     type: 'boolean',
     description:
-      'Suppresses analytics event tracking for this component (if configured).',
+      'Suppresses the automatic analytics click event, e.g. if you already have custom tracking.',
   },
   ...classNamePropDefs,
   ...stylePropDefs,
--- a/docs-ui/src/app/components/tabs/props-definition.ts
+++ b/docs-ui/src/app/components/tabs/props-definition.ts
@@ -71,7 +71,7 @@ export const tabPropDefs: Record<string, PropDef> = {
   noTrack: {
     type: 'boolean',
     description:
-      'Suppresses analytics event tracking for this component (if configured).',
+      'Suppresses the automatic analytics click event, e.g. if you already have custom tracking.',
   },
   ...childrenPropDefs,
   ...classNamePropDefs,
--- a/docs-ui/src/app/components/tag-group/props-definition.tsx
+++ b/docs-ui/src/app/components/tag-group/props-definition.tsx
@@ -87,7 +87,7 @@ export const tagPropDefs: Record<string, PropDef> = {
   noTrack: {
     type: 'boolean',
     description:
-      'Suppresses analytics event tracking for this component (if configured).',
+      'Suppresses the automatic analytics click event, e.g. if you already have custom tracking.',
   },
   ...childrenPropDefs,
   ...classNamePropDefs,


plugins/app/src/extensions/AnalyticsWrapper.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { AppRootWrapperBlueprint } from '@backstage/plugin-app-react';
+import { AnalyticsProvider as BUIAnalyticsProvider } from '@backstage/ui';
+import { useAnalytics } from '@backstage/core-plugin-api';
+
+export const analyticsWrapperExtension = AppRootWrapperBlueprint.make({
```
Rugvip: Thinking we want to add this to the app root instead: https://github.com/backstage/backstage/blob/d56d9eae3ca228aead82ceecae46b917a8adbc37/plugins/app/src/extensions/AppRoot.tsx#L117

both because I think this is a bit on unnecessary noise, and to make sure it's available pretty much everywhere

packages/core-components/report-alpha.api.md
```diff
@@ -21,6 +21,9 @@ export const coreComponentsTranslationRef: TranslationRef<
     readonly 'table.pagination.lastTooltip': 'Last Page';
     readonly 'table.pagination.nextTooltip': 'Next Page';
     readonly 'table.pagination.previousTooltip': 'Previous Page';
+    readonly 'emptyState.missingAnnotation.title': 'Missing Annotation';
+    readonly 'emptyState.missingAnnotation.actionTitle': 'Add the annotation to your component YAML as shown in the highlighted example below:';
+    readonly 'emptyState.missingAnnotation.readMore': 'Read more';
```
Rugvip, line 26: if these cause trouble in CI, just roll them back

plugins/app/src/extensions/AnalyticsWrapper.tsx
Naycon: Fixed.

packages/ui/report.api.md
```diff
@@ -254,6 +254,32 @@ export interface AlertProps
 // @public (undocumented)
 export type AlignItems = 'stretch' | 'start' | 'center' | 'end';
 
+// @public
+export type AnalyticsEventAttributes = {
+  [key: string]: string | boolean | number;
+};
+
+// @public
+export function AnalyticsProvider(props: AnalyticsProviderProps): JSX_2.Element;
```
Rugvip: Ah sorry, I realized another thing: Maybe we should make this a more generic integration provider that can be expanded on in the future, a more generic "BUI Provider" that has `useAnalytics` as its first input, put we might add additional ones down the line. Feeling that's a lot cleaner to manage than a bunch of stacked contexts.

Naycon: Yea, makes sense. Good thinking. Will adapt to that.

docs-ui/src/app/components/button-link/props-definition.tsx
```diff
@@ -62,6 +62,11 @@ export const buttonLinkPropDefs: Record<string, PropDef> = {
       </>
     ),
   },
+  noTrack: {
+    type: 'boolean',
+    description:
+      'Suppresses analytics event tracking for this component (if configured).',
```
iamEAP: Useful to indicate why you might use it, or would that be too verbose? e.g. maybe instead of `(if configured)`, something like `(e.g. if you already have custom tracking in a click handler)`.

Perhaps if not here, then in a convenient spot in documentation.

plugins/app/src/extensions/AppRoot.tsx
```diff
@@ -35,6 +35,8 @@ import {
   RouterBlueprint,
   SignInPageBlueprint,
 } from '@backstage/plugin-app-react';
+import { BUIProvider } from '@backstage/ui';
+import { useAnalytics } from '@backstage/core-plugin-api';
```
iamEAP: Hmm... Should this pull from `@backstage/frontend-plugin-api` instead?

packages/ui/src/analytics/BUIProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+ *       <AppContent />
+ *     </BUIProvider>
+ *   );
+ * }
+ * ```
+ *
+ * @public
+ */
+export function BUIProvider(props: BUIProviderProps) {
+  const { useAnalytics, children } = props;
```
iamEAP, line 48: Ah, is the reason for this just that we don't want to have a direct dependency on the package that exposes `useAnalytics()` from BUI?

Don't know if it would be a good idea, but did you explore other ways to get around that? Maybe peer dependencies or something? Thought we tried to keep the `useAnalytics()` hook as safe as possible to use, even outside any real API/analytics context, precisely for this kind of situation. (Also so it wouldn't fail in tests where components weren't fully wrapped in the right context)

docs-ui/src/app/components/button-link/props-definition.tsx
Naycon: Yea, good thinking. This will be in the props table, so space is a little limited. I can try and see if it looks all right.

packages/ui/src/analytics/BUIProvider.tsx
Rugvip: Yep and that safety is helping out here 👍 

We're aiming to avoid a `@backstage/frontend-plugin-api` dependency in BUI. A peer dep I don't think helps much in this case, would need the actual dep in the end anyway right? Unless we have separate entry points or smth for with/without Backstage usage. Since `useAnalytics` is based on the Utility API wiring in the end we can't really just move that out either.

I think the main competing option is a duplication of the `useAnalytics` implementation into BUI using a relative import like we do in places like https://github.com/backstage/backstage/blob/b9d90a7140cccf4c47168d954212133bbd2326fd/packages/frontend-test-utils/src/app/createExtensionTester.tsx#L28-L39

The downside of that is that BUI gets tied to a specific version of that implementation, also creates an implicit dependency on the package during dev time.

Thinking that this is the cleanest way to wire things up in the end.

plugins/app/src/extensions/AppRoot.tsx
Naycon: Right you are, fixing!

docs-ui/src/app/components/button-link/props-definition.tsx
Naycon: It worked out, thanks for the suggestion!

PR #33158, opened 2026-03-05
catalog: move stitch queue into dedicated table

freben:
```
                        Table "public.stitch_queue"
     Column     |           Type           | Collation | Nullable | Default 
----------------+--------------------------+-----------+----------+---------
 entity_ref     | character varying(255)   |           | not null | 
 stitch_ticket  | character varying(255)   |           | not null | 
 next_stitch_at | timestamp with time zone |           | not null | 
Indexes:
    "stitch_queue_pkey" PRIMARY KEY, btree (entity_ref)
    "stitch_queue_next_stitch_at_idx" btree (next_stitch_at)
```

## Summary

- Creates a dedicated `stitch_queue` table instead of embedding stitch columns in `final_entities`
- `entity_ref` is the primary key, with `stitch_ticket` and `next_stitch_at` columns
- On stitch completion, the row is deleted (if the ticket is unchanged) rather than nulling a timestamp
- Migration moves data from `refresh_state` stitch columns into the new table

Alternative approach to #32854 — compare the two for design review.

## Test plan

- All existing stitcher tests updated and passing across SQLite, Postgres 14/18, MySQL 8
- Migration up/down test covers data migration and schema changes


freben:
Tested locally - should be shipped to our beta instance to exercise

PR #33173, opened 2026-03-06
docs(bep): define shared AI assets repo and cli support

drodil:

this PR introduces BEP for backstage managed AI assets in the backstage
monorepo that can be shared with backstage contributors and adaptors.

added first skill to the docs to test this out. locally it's possible to
run: npx skills add ./docs which adds the mui-to-bui skill to be used.

### DIFF
--- a/beps/0013-backstage-ai/README.md
+++ b/beps/0013-backstage-ai/README.md
@@ -27,7 +27,7 @@ creation-date: 2026-03-06
 This BEP proposes that curated Backstage AI skills are authored in the main Backstage
 repository and published separately to a well-known endpoint on `backstage.io`. Instead
 of creating a new Backstage-owned repository and custom discovery protocol, Backstage
-will publish skills using the well-known layout already supported by `skills.sh`.
+will publish skills using the well-known layout already supported by [`skills.sh`](https://skills.sh/).
 
 The published output can be released on its own publishing cadence even though the source
 content lives in the main monorepo.


beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+This preserves autonomy but does not solve discovery, reuse, or consistency. It also increases duplicated maintenance effort across adopters.
+
+2. Publish assets only as documentation in the Backstage docs site.
+
+This improves discoverability but provides no versioned install/update mechanism and no ergonomic CLI integration.
+
+3. Ship shared assets directly in the Backstage monorepo.
+
+This couples fast-moving AI content to core release cadence and governance. A dedicated repository provides cleaner ownership and iteration speed.
```
Rugvip: Wanna challenge this a little bit. I think there could be ways to ship this work that doesn't necessarily slow things down. For example we could go for a more direct publish to `backstage.io` flow as a way of shipping this (as separate structured data), have it in an independently released package that doesn't follow the main release process, or structure it so that it can be grabbed straight from the repo via raw requests.

I do see a fairly big benefit to keep in the main repo. Not least because we likely want to use a fair number of those skills here 😁, but also to keep governance simple and have it be more visible.

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,252 @@
...
+- Semantic version.
+- Description and tags.
+- Compatibility hints (for example minimum Backstage version, CLI version).
+- Target compatibility hints (for example `copilot`, `claude`, or `generic`).
+- Maintainer/owner.
+- Optional security or compliance notes.
+
+Repository CI will validate metadata schema, lint content, and verify bundle references.
+
+### CLI
```
Rugvip: This is not a solution that's unique to Backstage at all really. I'd prefer leaning on existing tooling for this rather than our own if possible. Immediate search finding is for example https://github.com/tech-leads-club/agent-skills. ⚠️ I have no idea what that is or if it even fits our use-case or whatever else might be a reason not to use that project ⚠️, but seems like something in the direction we want. Same thing for the skills and distribution honestly.

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,252 @@
...
+modified files silently and will provide clear conflict guidance.
+
+## Design Details
+
+### Shared repository design
+
+The repository will use a predictable layout:
+
+- `skills/` for reusable skill packages.
+- `prompts/` for prompt templates.
```
Rugvip: honestly at this point I feel this is less important, or is that just me? 😅

Rugvip (changes requested):
Love it! And yes let's get this going, very much time to do so.

Kept lightweight on comments for now. The main thing I'd like to explore more at this point is what existing tooling and solutions are there in this space. Both the end user tooling and installation management, but also the way things are structured.

beps/0013-backstage-ai/README.md
drodil: Yeah, that was an alternative I had in mind. I am just concerned how much extra work and ci checks are needed for someone to contribute new skills.

beps/0013-backstage-ai/README.md
drodil: That's an issue we've encountered internally as well; there's really not tooling for this - or at least something we're aware of. Maybe worth to do some more digging.

beps/0013-backstage-ai/README.md
drodil: Might also remove this, it's something from GHCP world (.github/prompts).

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,243 @@
...
+- [Release Plan](#release-plan)
+- [Dependencies](#dependencies)
+- [Alternatives](#alternatives)
+
+## Summary
+
+This BEP proposes that curated Backstage AI skills are authored in the main Backstage
+repository and published separately to a well-known endpoint on `backstage.io`. Instead
+of creating a new Backstage-owned repository and custom discovery protocol, Backstage
+will publish skills using the well-known layout already supported by `skills.sh`.
```
awanlin:
```suggestion
will publish skills using the well-known layout already supported by [`skills.sh`](https://skills.sh/).
```

.github/vale/config/vocabularies/Backstage/accept.txt
```diff
@@ -261,6 +262,7 @@ mailto
 maintainer's
 maintainership
 makefile
+makeStyles
```
awanlin: Would should just use backticks (`) around where ever this word is being used instead.

beps/0013-backstage-ai/README.md
awanlin: Would be great to include a bit of background on skills.sh, it looks to be part of Vercel, and I can't see a GitHub repo for it, easily at least, it's not a link on the main skills.sh page. Could this get converted to a commercial offering where the free aspects go away?

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,243 @@
...
+- Evaluate whether an optional thin `backstage-cli` wrapper adds enough value to justify maintenance.
+
+## Dependencies
+
+- `skills.sh` continuing to support the current well-known endpoint format.
+- A Backstage publishing workflow that can emit static files to `backstage.io/.well-known/skills/`.
+- Agreement on long-term ownership and governance model for Backstage-authored skills.
+- Documentation updates describing authoring, publishing, and adopter usage patterns.
+
+## Alternatives
```
awanlin, line 215: These alternatives make sense but they pre-assume that skills.sh will always be around. Given how fast things are changing is that safe to assume? It feels like it will but it also feels like a like move to a commercial path could come very quickly. History has shown that the free tier always gets cut when that happens.

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,243 @@
...
+- It should default to the Backstage well-known endpoint.
+- It should delegate to `skills.sh` semantics rather than defining its own registry or manifest.
+- It should not introduce separate bundle, version, or update concepts.
+
+This approach keeps discovery and installation interoperable with the broader ecosystem,
+while still letting Backstage publish curated guidance from an official source.
+
+## Design Details
+
+### Source and publishing design
```
awanlin, line 121: What is the flow for experimental or draft skills? In the case of the MUI to BUI migration skill you included here we were using the Community Plugin repo to test and validate it before potentially moving/adding it to the main Backstage repo. This same pattern could come up with other skills and it would be good to have an way to handle this.

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+The source of truth for Backstage skills will live in the Backstage monorepo. The exact
```
awanlin, line 123: Does this include skills for the Community Plugin repo? We (Community Plugin Maintainers) may want to add some skills to help us with various maintainers tasks - reporting, bulk changes, etc. - would those need to be added here?

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+Phase 2: Publishing rollout
+
+- Implement a publishing workflow that generates `/.well-known/skills/` content for `backstage.io`.
+- Publish `index.json` and the initial skill directories.
+- Validate the published endpoint against `skills.sh` consumption behavior.
+
+Phase 3: General availability
+
+- Publish adoption documentation centered on `npx skills add https://backstage.io`.
```
awanlin, line 204: I think this part is going to be crucial. Reading this over one thing comes to mind that I think will come up that we experience with the Backstage.io docs: why did you use Docusaurus instead of using TechDocs. I can see the same here: why did you use skills.sh instead of using the Catalog. I think setting the context right here will make that clear but it's something we need to take into account.

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,243 @@
...
+title: Backstage AI Skills Publishing and Discovery
+status: implementable
+authors:
+  - '@drodil'
+owners:
+  - '@backstage/maintainers'
+creation-date: 2026-03-06
+---
+
+# BEP: Backstage AI skills publishing & discovery
```
awanlin: As part of this BEP we should keep in mind how this is different from Skills that you might have at your own organization and how Backstage can be used to help with discovery. I can see skills.sh being useful in the context of development for Backstage. But I think the Backstage Catalog is a better potential model for internal Skills and that we should be clear around what we are recommending for these two use cases.

.github/vale/config/vocabularies/Backstage/accept.txt
drodil: They were inside backticks in the mui-to-bui skill but vale still complained aboth that 🤔

beps/0013-backstage-ai/README.md
drodil: It's here https://github.com/vercel-labs/skills - what comes to conversion to a commercial offering, that's a risk in every tool. At least it's MIT licensed.

beps/0013-backstage-ai/README.md
drodil: Well, there is no support for that or even versioning as we speak with `skills`. It could be handled with naming convention 🤔

beps/0013-backstage-ai/README.md
drodil: Yes, this would be the place for all the skills.

beps/0013-backstage-ai/README.md
drodil: Good questions, but good to keep in mind the separation of backstage docs and runtime. If backstage.io were just a backstage instance itself, the catalog and techdocs would make sense.

beps/0013-backstage-ai/README.md
drodil: Yep, catalog is an option, and could also support skills.sh out of the box, even the skills would be catalog entities.

beps/0013-backstage-ai/README.md
drodil: Answered already a bit above. But apart from that, it's always possible to do own tooling for this (like this BEP originally proposed until discussion with @Rugvip ) but then we are drifting away from something that's already used by many people.

.github/vale/config/vocabularies/Backstage/accept.txt
Rugvip: They're not currently in backticks, could you try adding adding it back everywhere since we'd want to use that anyway? and skip it here as well. Same thing with Flex

Could be that you forgot to use backticks in the headings?

beps/0013-backstage-ai/README.md
Rugvip: I see this as being completely focused around Backstage development and how we ship skills for backstage, rather than skills in Backstage

beps/0013-backstage-ai/README.md
Rugvip: I like the focus here that we're using a well-known layout supported by an existing tool, rather than the tool specifically being the way we deliver skills

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,243 @@
...
+- Building a Backstage-specific discovery index, manifest, or asset registry.
+- Supporting bundles, lockfiles, version pinning, or custom update flows in this first iteration.
+
+## Proposal
+
+Introduce three deliverables:
+
+1. A Backstage monorepo location that serves as the source of truth for Backstage-authored AI skills.
+1. A publishing workflow that exposes approved skills at `https://backstage.io/.well-known/skills/`.
+1. Documentation for consuming those skills through `skills.sh`, with an optional thin `backstage-cli` convenience wrapper.
```
Rugvip: What would `backstage-cli` add here? 🤔

beps/0013-backstage-ai/README.md
Rugvip: I view skills pretty much as living documents tbh. Easy to get immediate value from but with a healthy amount of iteration to refine them over time. Would probably lean towards user control of what skills to bring in rather than any explicit experimental skills or such

beps/0013-backstage-ai/README.md
Rugvip: Yep, and might even lean towards just documenting useful skills in consuming repos rather than putting them directly in the repo too, just because of the variance across where they need to be installed for different clients. Up to each consuming repo to decide though imo.

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+- `SKILL.md` as the primary skill entry point.
+- Supporting prompt or instruction files referenced by the skill.
+- Optional metadata files used by Backstage publishing workflows.
+
+Published output on `backstage.io` will follow the `skills.sh` well-known layout:
+
+- `https://backstage.io/.well-known/skills/index.json`
+- `https://backstage.io/.well-known/skills/<skill-name>/SKILL.md`
+- Additional files under `https://backstage.io/.well-known/skills/<skill-name>/...`
```
Rugvip, line 137: Are there external docs for this to reference?

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+The main adoption path is documentation, not new tooling:
+
+- Users install Backstage skills with `npx skills add https://backstage.io`.
+- Users can install a specific published skill by pointing at its well-known skill URL.
+- Any install behavior, target path decisions, or local merge semantics remain the responsibility of `skills.sh`.
+
+An optional Backstage CLI command could be added later for convenience, for example:
+
+- `backstage-cli ai add [skill]`
```
Rugvip: might lean towards `skills add` or `ai skills add`. I think something like `list` might be useful too.

Either way, I do view this as something to add only if we see significant value in it or if existing tooling heads in a direction that doesn't work for us

beps/0013-backstage-ai/README.md
Rugvip: Yeah not too worried about this right now tbh, we have plenty of options for moving away from `skills.sh` if needed in the future

beps/0013-backstage-ai/README.md
drodil: https://github.com/vercel-labs/skills/pull/573/changes - just because the information about that lives only in the source code for now.

beps/0013-backstage-ai/README.md
drodil: Just leaving the whole `backstage-cli` wrapper out for now. If there ever is a need to get rid of `skill` we might consider that or something else at that point. Just to get the ball rolling now.

beps/0013-backstage-ai/README.md
drodil: Removed.

.github/vale/config/vocabularies/Backstage/accept.txt
drodil: Yep, fixed. Hopefully 👍

beps/0013-backstage-ai/README.md
awanlin: I agree on the focus. The reason that I left this comment is the related PR to add a plugin that really made the intention unclear and I think including clarity in this BEP around that is important.

drodil: Yeah, that other PR mixed things 😆 But that one needs more discussion, hopefully @benjdlambert has some good ideas how to move forward.

beps/0013-backstage-ai/README.md
```diff
@@ -0,0 +1,234 @@
+---
+title: Backstage AI Skills Publishing and Discovery
+status: implementable
+authors:
+  - '@drodil'
+owners:
+  - '@backstage/maintainers'
```
Rugvip, line 7:
```suggestion
owners:
  - '@backstage/maintainers'
project-areas:
  - core
```

Rugvip (approved):
I'm happy with the BEP in its current form 👍 💯 

From my point of view we can :shipit: and start building

Rugvip:
:tada:

```
npx skills add https://backstage.io
```

PR #33177, opened 2026-03-06
feat(org,catalog-graph)!: migrate MembersListCard, OwnershipCard, and CatalogGraphCard to BUI

Naycon:

Second batch of entity page card migrations from MUI/InfoCard to BUI card primitives via `EntityInfoCard`.

### Changes

**MembersListCard:**
- Card shell migrated to `EntityInfoCard`
- Member items redesigned using BUI `Card`, `Avatar`, `Flex`, and `Text` in a horizontal layout (avatar left, name/email/description right)
- Pagination replaced with BUI `TablePagination` (shows "1-50 of 123" instead of page numbers)
- Search field replaced with BUI `SearchField`
- Aggregate members toggle replaced with BUI `Switch`, moved to header actions
- Exported `useEntityRoute` from `@backstage/plugin-catalog-react`

**OwnershipCard:**
- Card shell migrated to `EntityInfoCard`
- Relations toggle replaced with BUI `Switch`, moved to header actions
- Removed `variant`, `maxScrollHeight` props
- Removed unused MUI imports (`List`, `ListItem`, `Tooltip`, etc.)

**CatalogGraphCard:**
- Card shell migrated to `EntityInfoCard`
- Deep link replaced with BUI `ButtonLink` in card footer
- Removed `variant` prop
- Added `@backstage/ui` and `@remixicon/react` dependencies

### Before / After

| Before | After |
|--------|-------|
| image | image |
| image | image |
| image | image |

### DIFF
--- a/packages/cli/src/modules/translations/lib/extractTranslations.test.ts
+++ b/packages/cli/src/modules/translations/lib/extractTranslations.test.ts
@@ -52,7 +52,7 @@ describe('extractTranslations', () => {
     expect(refs[0].messages).toHaveProperty(['membersListCard.title']);
 
     // Verify interpolation placeholders are preserved
-    expect(refs[0].messages['membersListCard.subtitle']).toContain(
+    expect(refs[0].messages['membersListCard.title']).toContain(
       '{{groupName}}',
     );
   });


plugins/catalog-react/src/components/EntityRefLink/EntityRefLink.tsx
```diff
@@ -78,7 +78,8 @@ export const EntityRefLink = forwardRef<any, EntityRefLinkProps>(
 // contrived, because it tries to retain the casing of the entity name if
 // present, but not of other parts. This is in an attempt to make slightly more
 // nice-looking URLs.
-function useEntityRoute(
+/** @public */
+export function useEntityRoute(
```
Rugvip: Hmm, now that you want to export this we gotta bike-shed a bit 😉

We already have `EntityRefLink`, so I'm kinda thinking we make this simple and call it `useEntityRefLink`? Now one question is whether we should use the same pattern as `useRouteRef`, i.e. returning a link generation function rather than the link itself. That works a lot better for populating a large or variable number of links. So all in all, perhaps this?

```tsx
function MyComponent(props) {
  const entityLink = useEntityRefLink()
  return <Link href={entityLink(props.entity)} />
```

plugins/org/report-alpha.api.md
```diff
@@ -214,17 +214,13 @@ export const orgTranslationRef: TranslationRef<
...
-    readonly 'membersListCard.aggregateMembersToggle.ariaLabel': 'Users Type Switch';
-    readonly 'membersListCard.aggregateMembersToggle.directMembers': 'Direct Members';
-    readonly 'membersListCard.aggregateMembersToggle.aggregatedMembers': 'Aggregated Members';
+    readonly 'membersListCard.noSearchResult': 'Found no members matching "{{searchTerm}}".';
+    readonly 'membersListCard.aggregateMembersToggle.label': 'Include subgroups';
     readonly 'ownershipCard.title': 'Ownership';
-    readonly 'ownershipCard.aggregateRelationsToggle.ariaLabel': 'Ownership Type Switch';
-    readonly 'ownershipCard.aggregateRelationsToggle.directRelations': 'Direct Relations';
-    readonly 'ownershipCard.aggregateRelationsToggle.aggregatedRelations': 'Aggregated Relations';
+    readonly 'ownershipCard.aggregateRelationsToggle.label': 'Include indirect ownership';
```
freben, line 223: Are the translation changes (like no longer having `subtitle` and title having `{{groupName}}`) worth calling out in the changeset too?

plugins/org/src/components/Cards/OwnershipCard/OwnershipCard.test.tsx
```diff
@@ -326,13 +328,15 @@ describe('OwnershipCard', () => {
         },
       );
 
-      expect(getByTitle('Aggregated Relations')).toBeInTheDocument();
+      await waitFor(() => {
+        expect(screen.getByRole('switch')).toBeChecked();
```
freben: as a general pattern, prefer `await expect(screen.findByX).resolves.toY` over using `findBy` (considering enabling that lint rule some time down the line). But not blocking on this

plugins/catalog-react/src/components/EntityRefLink/EntityRefLink.tsx
Naycon: Sure, I'm still learning to navigate the thinking here and the aspect so I don't have much of an opinion. Certainly seems nicer if you need more than one, just as you say, so happy to update to go with that!

plugins/org/report-alpha.api.md
Naycon: Wasn't sure how you do this normally, but will call them out. Won't hurt!

plugins/org/src/components/Cards/OwnershipCard/OwnershipCard.test.tsx
Naycon: Doing other changes, might as well fix this too. :)

plugins/catalog-react/src/components/EntityRefLink/EntityRefLink.tsx
Naycon: Updated now.

PR #33184, opened 2026-03-06
feat: add oxlint as alternative lint engine

schultzp2020:

This adds oxlint as an alternative lint engine that can be adopted incrementally, package by package, alongside the existing ESLint setup. Packages opt in by passing `--engine oxlint` to the `backstage-cli package lint` command.

The core of the change is a new lint dispatcher that checks the `--engine` flag and routes to either the existing ESLint runner or a new oxlint runner. The oxlint runner loads a shared base configuration from `packages/cli/config/oxlint.json` containing ecosystem-wide rules that apply to any Backstage project. The base config enables 9 plugins (typescript, import, jest, react, jsx-a11y, unicorn, react-perf, node, and promise), resolves JS plugin paths to absolute, injects Node.js builtin module restrictions for frontend packages, and applies role-based rule adjustments (e.g. allowing `console` in backend packages, enabling `no-dirname-in-src`). Three custom JS plugins port Backstage-specific ESLint rules to oxlint's plugin API: `backstage-plugin` (monorepo import rules, Material UI import rewriting), `notice-plugin` (Apache 2.0 copyright header enforcement), and `restricted-syntax-plugin` (locale-safe casing, React/Winston import restrictions, `__dirname` usage). All three plugins have test coverage.

Monorepo-specific rules (notice header enforcement, Material UI 4 import restrictions) are separated into a root-level `.oxlintrc.json` rather than the shared base config. The config system walks from the repo root down to each package directory, merging any `.oxlintrc.json` files found along the way in outermost-first order — so a `plugins/.oxlintrc.json` could apply shared overrides to all plugins. Arrays use replacement semantics so overrides can fully replace items like `jsPlugins` or `ignorePatterns`. Bare specifier paths in `jsPlugins` (e.g. `@backstage/cli/config/oxlint-plugins/notice-plugin.js`) are resolved via `require.resolve()`.

The repo-level `backstage-cli repo lint` command also gained `--engine` support, with each package's engine determined from its lint script. The success cache hash was bumped to `v2` to account for the engine dimension. A new `scripts/lint-staged-fix.js` replaces the previous direct `eslint --fix` call in `lint-staged`, grouping staged files by their package's engine and running the appropriate linter with role-aware configuration. `oxlint` and `oxlint-tsgolint` (for type-aware linting) are added as optional peer dependencies of `@backstage/cli` and as root devDependencies. `packages/types` serves as the pilot package with oxlint enabled by default.


schultzp2020:
Just a heads up, I'm trying a new AI workflow.

packages/cli/package.json
```diff
@@ -207,6 +207,8 @@
     "jest-environment-jsdom": "*",
     "jsdom": "^27.1.0",
     "mini-css-extract-plugin": "^2.4.2",
+    "oxlint": ">=1.0.0",
```
drodil: Should we also move eslint to peerDependency like jest was moved some time ago? What way the users could get rid of eslint alltogether if they want to. Can be done later as well.

drodil (commented):
Will give it a spin next week!

packages/cli/package.json
schultzp2020: That's a good idea! I'll see what others think about moving eslint to a peerDependency in this PR. 👀

packages/cli/package.json
```diff
@@ -44,6 +44,10 @@
   "jest": {
     "coveragePathIgnorePatterns": [
       "/__fixtures__/"
+    ],
+    "roots": [
+      "<rootDir>",
+      "<rootDir>/../config/oxlint-plugins"
```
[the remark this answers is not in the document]
schultzp2020: Not applicable, the current configuration is correct.

packages/cli-module-lint/src/commands/package/lint.ts
```diff
@@ -54,47 +65,40 @@ export default async ({ args, info }: CommandContext) => {
...
 
-  if (failed) {
-    process.exit(1);
-  }
+  const eslintMod = await import('./eslint');
+  const runEslint =
+    typeof eslintMod.default === 'function'
+      ? eslintMod.default
+      : (eslintMod.default as any).default;
+  await runEslint(opts);
```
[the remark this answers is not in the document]
planeiii-te: I would simplify that first function to remove the useless repitition:

```js
export const resolveDefaultExport = <T>(mod: any): T =>
  typeof mod?.default === 'function'
    ? mod.default as T
		: mod as T;
```

.oxlintrc.json
```diff
@@ -0,0 +1,11 @@
+{
+  "jsPlugins": [
+    "@backstage/cli/config/oxlint-plugins/restricted-syntax-plugin.js",
+    "@backstage/cli/config/oxlint-plugins/backstage-plugin.js",
+    "@backstage/cli/config/oxlint-plugins/notice-plugin.js"
+  ],
```
schultzp2020, line 6: Lists all three jsPlugins even though `restricted-syntax-plugin.js` and `backstage-plugin.js` are already in the base config. `mergeOverrideFile` uses array **replacement** semantics (not concatenation), so the root override fully replaces the base config's `jsPlugins` list. This trades a bit of duplication for the ability to opt out of plugins at any override level.

packages/cli-module-lint/src/lib/oxlintConfig.ts
```diff
@@ -0,0 +1,276 @@
...
+      typeof local[key] === 'object' &&
+      local[key] !== null &&
+      !Array.isArray(local[key]) &&
+      typeof config[key] === 'object' &&
+      config[key] !== null
+    ) {
+      config[key] = { ...config[key], ...local[key] };
+    } else {
+      config[key] = local[key];
+    }
```
schultzp2020, line 174: Objects are shallow-merged (override keys win), but arrays and scalars are fully replaced — see comment on `.oxlintrc.json` for why.

packages/cli-module-lint/src/lib/oxlintConfig.ts
```diff
@@ -0,0 +1,276 @@
...
+  packageDir: string,
+  rootDir: string,
+): string[] {
+  const normalizedRoot = resolve(rootDir);
+  const normalizedPkg = resolve(packageDir);
+
+  // Guard: if packageDir is not under rootDir (or equal), only check packageDir itself
+  if (
+    normalizedPkg !== normalizedRoot &&
+    !normalizedPkg.startsWith(normalizedRoot + sep)
```
schultzp2020, line 195: `startsWith(normalizedRoot + sep)` rather than just `startsWith(normalizedRoot)` — without the `sep` suffix, a rootDir of `/tmp/foo` would incorrectly match `/tmp/foobar`. Covered by the `does not false-positive when packageDir shares a prefix with rootDir` test.

packages/cli-module-lint/src/lib/oxlintConfig.ts
```diff
@@ -0,0 +1,276 @@
...
+      ...entries.map((entry: string | { name: string }) => {
+        if (typeof entry === 'string') {
+          return {
+            name: entry,
+            message: `Avoid using implicitly global variables. Use e.g. window.${entry} instead if this was your intent.`,
+          };
+        }
+        return entry;
+      }),
+    ];
```
schultzp2020, line 89: The config file stores `no-restricted-globals` entries as plain strings to keep it compact (~170 lines instead of ~680). At runtime each string is expanded into `{name, message}` with a templated message matching the ESLint config's `.map()` pattern. Entries already in object form pass through unchanged.

schultzp2020 (commented):
Design decisions...

Rugvip:
ah, didn't realize you where also looking at this concretely 😅. Vibed up https://github.com/backstage/backstage/pull/33427, happy for you to take over but I think it's good to bring in a fair number of things from over there, especially the config helper and shipping it as a separate module

schultzp2020:
not stale

schultzp2020:
Replaced in favor of https://github.com/backstage/backstage/pull/34132

PR #33204, opened 2026-03-08
feat: allow dynamically enabling and disabling extensions

aramissennyeydd:

Built on top of #33154 - goal here is to allow users to specify entity predicates to enable or disable extensions before rendering using feature flags or permissions (to start). See `pagesPlugin.tsx` for examples of what this integration looks like.

### DIFF
--- a/plugins/app/package.json
+++ b/plugins/app/package.json
@@ -53,6 +53,7 @@
   "dependencies": {
     "@backstage/core-components": "workspace:^",
     "@backstage/core-plugin-api": "workspace:^",
+    "@backstage/filter-predicates": "workspace:^",
     "@backstage/frontend-plugin-api": "workspace:^",
     "@backstage/integration-react": "workspace:^",
     "@backstage/plugin-app-react": "workspace:^",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4295,6 +4295,7 @@ __metadata:
     "@backstage/core-components": "workspace:^"
     "@backstage/core-plugin-api": "workspace:^"
     "@backstage/dev-utils": "workspace:^"
+    "@backstage/filter-predicates": "workspace:^"
     "@backstage/frontend-defaults": "workspace:^"
     "@backstage/frontend-plugin-api": "workspace:^"
     "@backstage/frontend-test-utils": "workspace:^"


.changeset/ff-perm-enabled-frontend-app-api.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/frontend-app-api': patch
```
Rugvip: Skip changesets for this imo, we'll just lean on the parent PR

.changeset/ff-perm-enabled-frontend-defaults.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/frontend-defaults': patch
+---
+
+Updated `createApp` to await the now-async `finalize()` call from `prepareSpecializedApp`, enabling permission-gated extensions to be resolved before the app tree is rendered.
```
Rugvip: Skip this one too

.changeset/ff-perm-enabled-frontend-plugin-api.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/frontend-plugin-api': patch
+---
+
+Added `enabled` option to `createExtension`, `createExtensionBlueprint`, and `AppNodeSpec`, accepting a `FilterPredicate` from `@backstage/filter-predicates` to conditionally enable extensions based on feature flags or other runtime conditions.
+
+Added `permissions` option to `createFrontendPlugin` and `createFrontendModule`, allowing plugins and modules to declare which permissions they use. These permissions are checked at app startup so that extensions can conditionally enable themselves using a `permissions` predicate, e.g. `enabled: { permissions: { $contains: 'catalog.entity.create' } }`.
```
Rugvip: Hmm, this seems like a bit of a foot-gun, but can't think of any nicer way either. Wonder if there's any form of check we could add to make it less likely that users forget to declare permissions? 🤔

packages/frontend-app-api/src/tree/instantiateAppNodeTree.ts
```diff
@@ -516,6 +518,13 @@ export function instantiateAppNodeTree(
     if (node.spec.disabled) {
       return undefined;
     }
+    if (
+      predicateContext !== undefined &&
+      node.spec.enabled !== undefined &&
+      !evaluateFilterPredicate(node.spec.enabled, predicateContext)
+    ) {
+      return undefined;
```
Rugvip, line 526: This made me wonder if the reason the extension was not instantiated should be provided somewhere. Clear enough if it's disabled, and only slightly trickier if it is orphaned vs not matching the predicate. I don't think we should do anything right now, but could be needed in the future.

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
```diff
@@ -320,13 +336,30 @@ export type PreparedSpecializedApp = {
...
 };
 
+// Minimal local permission API interface to avoid a dependency on @backstage/plugin-permission-react
+type MinimalPermissionApi = {
+  authorize(
+    requests: EvaluatePermissionRequest[],
+  ): Promise<EvaluatePermissionResponse[]>;
+};
+
+const localPermissionApiRef = createApiRef<MinimalPermissionApi>({
```
Rugvip, line 359: Ah, yeah we prolly wanna move this into frontend-plugin-api, but fine to keep like this for now

packages/frontend-app-api/package.json
```diff
@@ -36,8 +36,10 @@
     "@backstage/core-app-api": "workspace:^",
     "@backstage/core-plugin-api": "workspace:^",
     "@backstage/errors": "workspace:^",
+    "@backstage/filter-predicates": "workspace:^",
     "@backstage/frontend-defaults": "workspace:^",
     "@backstage/frontend-plugin-api": "workspace:^",
+    "@backstage/plugin-permission-common": "workspace:^",
```
Rugvip: make this a dev dep by using an explicit `import type ...`

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
```diff
@@ -727,3 +808,53 @@ function mergeExtensionFactoryMiddleware(
...
+/**
+ * Recursively walks a FilterPredicate and returns all string values referenced
+ * by `permissions: { $contains: '...' }` expressions. This lets us issue a
+ * single batched authorize call for only the permissions actually referenced.
+ */
+function extractPermissionNames(predicate: FilterPredicate): string[] {
+  return extractPredicateKeyNames(predicate, 'permissions');
+}
+
+function extractPredicateKeyNames(
```
Rugvip, line 831: huh, this seems to go against the changeset description? I we can do this reliably I think I prefer this over having to register the permissions in the plugin

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
```diff
@@ -446,11 +479,7 @@ export function prepareSpecializedApp(
     | undefined;
   return {
     signIn,
-    finalize() {
-      if (finalized) {
-        return finalized;
-      }
-
+    async buildPredicateContext(): Promise<ExtensionPredicateContext> {
```
Rugvip, line 482: I think this should be something that happens under the hood, so that after the sign-in is complete this part has also completed and the `finalize()` call can do the rest synchronously.

packages/frontend-defaults/report.api.md
```diff
@@ -23,6 +23,7 @@ export function createApp(options?: CreateAppOptions): {
 // @public
 export interface CreateAppOptions {
   advanced?: {
+    allowUnknownExtensionConfig?: boolean;
```
Rugvip: merge conflict? 😅

packages/frontend-plugin-api/report.api.md
```diff
@@ -247,6 +248,8 @@ export interface AppNodeSpec {
   // (undocumented)
   readonly disabled: boolean;
   // (undocumented)
+  readonly enabled?: FilterPredicate;
```
Rugvip, line 251: Don't quite like `disabled` + `enabled` tbh, thinking that might risk confusion about when to use either. What do you think about `predicate` instead of we end up with that structure? 🤔 Some other options:

- `if`
- `condition`
- `test`
- `filter`
- `guard`
- `check`
- `switch`

plugins/permission-react/src/apis/IdentityPermissionApi.ts
```diff
@@ -52,11 +52,18 @@ export class IdentityPermissionApi implements PermissionApi {
 
   async authorize(
     request: AuthorizePermissionRequest,
-  ): Promise<AuthorizePermissionResponse> {
-    const response = await this.permissionClient.authorize(
-      [request],
+  ): Promise<AuthorizePermissionResponse>;
+  async authorize(
+    requests: AuthorizePermissionRequest[],
+  ): Promise<AuthorizePermissionResponse[]>;
```
Rugvip, line 66: Hmm how do you feel about batching requests sent within the same tick instead? Think it would generally be good to have that and that doesn't touch the external API or slow things down.

packages/frontend-defaults/report.api.md
aramissennyeydd: not entirely sure 🤔 - reset to base and it went away

Issue #33060, opened 2026-02-27
RFC: Modeling/Importing of AI Models/ModelServers in the Catalog


gabemontero:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
ai models, ai model servers, rfc, catalog model, plugins
```

### 🗃️ Project Area

Catalog

### 🔖 Need

The need articulated by @benjdlambert in [RFC: Modeling of MCP Servers in the Catalog](https://github.com/backstage/backstage/issues/32062) for a consistent catalog representation of AI's MCP certainly applies to other elements of AI, as he in fact called out in that RFC in a couple of places (the `Needs` section, and in the details of his `Option 3` in the `Proposal`).

This RFC intends to build off of  [RFC: Modeling of MCP Servers in the Catalog](https://github.com/backstage/backstage/issues/32062)  and start the conversation for modeling of AI Models and Model Servers into the catalog.  This RFC also intends to start the conversation on how the importing of the agreed upon model into the catalog could be structured, especially in the context of the disparate repositories and registries for AI Models and Model Servers, could be done most efficiently.

Before diving into the proposal, allow for the acknowledgement that a BEP is most likely the next logical step in getting community consensus on these points.  A partial draft for a BEP does exist at the time this RFC was opened.  However, given the potential complexities around this subject, the intent of opening this RFC prior to opening a PR for the BEP is to socialize enough IP here to both

- solicit essential feedback that could get the BEP off to a good start, or
- discover there is enough pushback from the community that fully investing in a BEP on this subject, at least at this time, is not warranted.

### 📝 Proposal

First, some goals, and non-goals, for the proposal (in line with the BEP template)

# Goals

- Provide a normalization layer that allows for a consistent expression of AI Model 
metadata in the Backstage catalog
  - Unlike MCP, singular specifications do not exist for expressing AI Model metadata, services, or hosting repositories /
    registries of AI Model metadata
- Minimize the number of EntityProviders needed for importing AI Models and Model Servers into the 
catalog
- Allow for an opt-in approach for enablement
- Allow for implementation of AI Model provider specific plugins via either community plugins or core plugins
- Decouple the modeling of AI Model metadata into the catalog from implementing the import of that metadata from
a given AI Model provider

# Non-Goals

[...]

PR #33207, opened 2026-03-09
`feat(catalog-model)`: Implement `type: mcp-server` with `kind: API`

benjdlambert:
Implementation of #32062.

Not sure of the best way to ship this, so this is a first pass. It's a bit of a breaking change to `kind: API`. Wondering if there's a better way or if we just handle it like this?

### DIFF
--- a/.changeset/mcp-server-api-entity.md
+++ b/.changeset/mcp-server-api-entity.md
@@ -1,0 +1,9 @@
+---
+'@backstage/catalog-model': major
+---
+
+Added `backstage.io/v1alpha2` API entity schema with support for MCP server entities. MCP server entities on v1alpha2 use a structured `remotes` array instead of a `definition` string.
+
+`ApiEntity` is now a union of `ApiEntityV1alpha1 | ApiEntityV1alpha2`. Consumers should narrow on `apiVersion` before accessing version-specific fields like `spec.definition` or `spec.remotes`.
+
+New exports: `ApiEntityV1alpha2`, `ApiEntityV1alpha2Spec`, `apiEntityV1alpha2Validator`.
--- a/packages/catalog-model/report.api.md
+++ b/packages/catalog-model/report.api.md
@@ -33,7 +33,10 @@ export const ANNOTATION_SOURCE_LOCATION = 'backstage.io/source-location';
 export const ANNOTATION_VIEW_URL = 'backstage.io/view-url';
 
 // @public
-interface ApiEntityV1alpha1 extends Entity {
+export type ApiEntity = ApiEntityV1alpha1 | ApiEntityV1alpha2;
+
+// @public
+export interface ApiEntityV1alpha1 extends Entity {
   // (undocumented)
   apiVersion: 'backstage.io/v1alpha1' | 'backstage.io/v1beta1';
   // (undocumented)
@@ -47,11 +50,40 @@ interface ApiEntityV1alpha1 extends Entity {
     system?: string;
   };
 }
-export { ApiEntityV1alpha1 as ApiEntity };
-export { ApiEntityV1alpha1 };
 
 // @public
 export const apiEntityV1alpha1Validator: KindValidator;
+
+// @public
+export interface ApiEntityV1alpha2 extends Entity {
+  // (undocumented)
+  apiVersion: 'backstage.io/v1alpha2';
+  // (undocumented)
+  kind: 'API';
+  // (undocumented)
+  spec: {
+    lifecycle: string;
+    owner: string;
+    system?: string;
+  } & ApiEntityV1alpha2Spec;
+}
+
+// @public
+export type ApiEntityV1alpha2Spec =
+  | {
+      type: 'mcp-server';
+      remotes: {
+        type: string;
+        url: string;
+      }[];
+    }
+  | {
+      type: string;
+      definition: string;
+    };
+
+// @public
+export const apiEntityV1alpha2Validator: KindValidator;
 
 // @public
 export class CommonValidatorFunctions {
@@ -255,7 +287,7 @@ export { GroupEntityV1alpha1 };
 export const groupEntityV1alpha1Validator: KindValidator;
 
 // @public (undocumented)
-export function isApiEntity(entity: Entity): entity is ApiEntityV1alpha1;
+export function isApiEntity(entity: Entity): entity is ApiEntity;
 
 // @public (undocumented)
 export function isComponentEntity(
--- a/packages/catalog-model/src/kinds/ApiEntityV1alpha2.ts
+++ b/packages/catalog-model/src/kinds/ApiEntityV1alpha2.ts
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2025 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { Entity } from '../entity/Entity';
+import schema from '../schema/kinds/API.v1alpha2.schema.json';
+import { ajvCompiledJsonSchemaValidator } from './util';
+
+/**
+ * The specification fields that vary based on the API type.
+ *
+ * @public
+ */
+export type ApiEntityV1alpha2Spec =
+  | {
+      type: 'mcp-server';
+      remotes: { type: string; url: string }[];
+    }
+  | { type: string; definition: string };
+
+/**
+ * Backstage API kind Entity using the v1alpha2 schema.
+ *
+ * @remarks
+ *
+ * This version supports structured spec fields that vary by API type.
+ * MCP server entities use a remotes array instead of a definition string.
+ * The JSON schema enforces which fields are required based on spec.type.
+ *
+ * See {@link https://backstage.io/docs/features/software-catalog/system-model}
+ *
+ * @public
+ */
+export interface ApiEntityV1alpha2 extends Entity {
+  apiVersion: 'backstage.io/v1alpha2';
+  kind: 'API';
+  spec: {
+    lifecycle: string;
+    owner: string;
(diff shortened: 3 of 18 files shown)


.changeset/mcp-server-api-entity.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/catalog-model': minor
```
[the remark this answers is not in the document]
clairep94: is this accurate @benjdlambert ?

plugins/api-docs/report.api.md
```diff
@@ -127,7 +127,7 @@ export type DefaultApiExplorerPageProps = {
 export function defaultDefinitionWidgets(): ApiDefinitionWidget[];
 
 // @public (undocumented)
-export const EntityApiDefinitionCard: () => JSX_2.Element;
+export const EntityApiDefinitionCard: () => JSX_2.Element | null;
```
freben: It would probably be better to revert these two to return the card but with an empty marker. Than in NFS we add a selector to not draw the extension if there's no definition. Maybe split out an inner component that either renders the empty state or the `<Tabs>` stuff

plugins/api-docs/src/components/ApiDefinitionCard/ApiDefinitionCard.tsx
```diff
@@ -35,19 +35,24 @@ export const ApiDefinitionCard = () => {
     return <Alert severity="error">{t('apiDefinitionCard.error.title')}</Alert>;
   }
 
+  const { spec } = entity;
   const definitionWidget = getApiDefinitionWidget(entity);
   const entityTitle = entity.metadata.title ?? entity.metadata.name;
 
+  if (!('definition' in spec)) {
```
freben: i'm tempted to add an empty check as well. Ie empty string renders empty state

packages/catalog-model/src/kinds/ApiEntityV1alpha1.ts
```diff
@@ -18,6 +18,18 @@ import type { Entity } from '../entity/Entity';
...
+ * The specification fields that vary based on the API type.
+ *
+ * @public
+ */
+export type ApiEntityV1alpha1Spec =
+  | {
+      type: 'mcp-server';
+      remotes: { type: string; url: string }[];
+    }
+  | { type: string; definition: string };
```
freben: Hm. While we're here and ripping this type apart anyway, does it make sense to even put a `definition?: string` in the "base" type? Then this one becomes just a type and allowing extra fields.

packages/catalog-model/src/schema/kinds/API.v1alpha1.schema.json
```diff
@@ -23,6 +23,25 @@
...
+      "kind": "API",
+      "metadata": {
+        "name": "backstage-mcp-actions",
+        "description": "Exposes tools related to the Backstage Ecosystem such as the Software Catalog"
+      },
+      "spec": {
+        "type": "mcp-server",
+        "lifecycle": "experimental",
+        "owner": "backstage",
+        "remotes": [
```
freben: Man AT LEAST this should ideally have been a new `apiVersion` one would think ... But given what calling code looks like, there's not much point.

packages/catalog-model/src/schema/kinds/API.v1alpha1.schema.json
```diff
@@ -65,11 +91,49 @@
...
               "description": "An entity reference to the system that the API belongs to.",
               "minLength": 1
-            },
-            "definition": {
-              "type": "string",
-              "description": "The definition of the API, based on the format defined by the type.",
-              "minLength": 1
+            }
+          },
+          "if": {
```
freben: We've written these using `oneOf` before. But I presume the `if` is because of the free form `type: string` greedily matching everything, making the validator too lax?

freben (changes requested):

packages/catalog-model/src/kinds/ApiEntityV1alpha2.ts
```diff
@@ -0,0 +1,62 @@
...
+import { ajvCompiledJsonSchemaValidator } from './util';
+
+/**
+ * The specification fields that vary based on the API type.
+ *
+ * @public
+ */
+export type ApiEntityV1alpha2Spec =
+  | {
+      type: 'mcp-server';
```
gabemontero, line 28: if we later agreed to add say add a `model-server` type, `agent` type, `prompt` type, would that necessitate more version bumps to v1alpha3, etc ?

benjdlambert: I don't think thats the intended behavior no. I think that I might hard code the additional API types for now, and just say `unknown` if it doesn't match any of the existing ones, so you have to explicitly cast or use type narrowing to select the right type from the union.

gabemontero: ok thanks for confirming/clarifying that the version won't need to be bumped as new v1alpha2 "remote" types are defined in addition to `mcp-server` @benjdlambert 

the notion of `just say 'unknown' if it doesn't match any of the existing ones, so you have to explicitly cast or use type narrowing to select the right type from the union` makes sense / seems like the correct way forward IIUC ..... I think it lines up the historical notion of the `type` field for `API` not being free form, like it is say with `Component` and `Resource`, because `API` primarily enables specific, contextual functionality.

one quick f/up question:  would examples of explicitly casting and the use of type narrowing you alluded to in one of the markdowns under backstage/docs/features/software-catalog to illustrate what consumers need to do make sense?

thanks again

gabemontero: > ok thanks for confirming/clarifying that the version won't need to be bumped as new v1alpha2 "remote" types are defined in addition to `mcp-server` @benjdlambert
> 
> the notion of `just say 'unknown' if it doesn't match any of the existing ones, so you have to explicitly cast or use type narrowing to select the right type from the union` makes sense / seems like the correct way forward IIUC ..... I think it lines up the historical notion of the `type` field for `API` not being free form, like it is say with `Component` and `Resource`, because `API` primarily enables specific, contextual functionality.
> 
> one quick f/up question: would examples of explicitly casting and the use of type narrowing you alluded to in one of the markdowns under backstage/docs/features/software-catalog to illustrate what consumers need to do make sense?

cool https://github.com/backstage/backstage/pull/33207/changes/bc12abc074a63a0b9f8bf3d9ae17ece96f4b8ad6 answers my question ^^ @benjdlambert thanks 

> 
> thanks again

packages/catalog-model/report.api.md
```diff
@@ -33,7 +33,10 @@ export const ANNOTATION_SOURCE_LOCATION = 'backstage.io/source-location';
 export const ANNOTATION_VIEW_URL = 'backstage.io/view-url';
 
 // @public
-interface ApiEntityV1alpha1 extends Entity {
+export type ApiEntity = ApiEntityV1alpha1 | ApiEntityV1alpha2;
```
Rugvip, line 36: Hmm, it's a bit of a problem that we need to treat this as breaking. Honestly I'm thinking it might be better to set up some policy that says that these are always discriminated by the `apiVersion` or something like that, so that we can add new versions without considering it a breaking change? 🤔

benjdlambert: Yeah I had that originally and have flip flopped about this in the first place. 

Happy to add documentation piece about working with the catalog, but is it worth us doing this once for all types and then moving forward it's fine to add more versions without breaking the API or thinking that it's not worth it?

Rugvip: Yeah that's what I was thinking. Establish something that let's us do this in the future. Perhaps we can now ship with with @freben's work?

gabemontero: hello @Rugvip @benjdlambert - a couple of questions on this thread, given a few weeks have passed, and Backstage/Kube Con are now behind us:

- am I correct in interpreting that the breaking change aspect is now being treated as a blocker for this item?
- was the work from @freben that was alluded to a few weeks ago the mini rfc opened yesterday (as of this comment's posting), https://github.com/backstage/backstage/issues/33770 and the associated draft PR (again as of this comment's posting) https://github.com/backstage/backstage/pull/33663 ?
- and if so, is that work essentially a dependency for this item in that it allows for avoiding the breaking change aspect?

fwiw, aside from interest in this PR and its associated RFC, I'm also interested in the context of the RFC https://github.com/backstage/backstage/issues/33060 that I opened, which attempts to build off of this work.

thanks

abhyudayareddy:
Thanks @benjdlambert — glad to see this moving forward, and the core question you've raised is exactly the right one to resolve before merging.

On the breaking-change concern: the problem is that making `spec.definition` optional on all `kind: API` entities is too broad a change. Any existing tooling (catalog processors, Backstage plugins, external importers) that reads `spec.definition` and assumes it's always a string will silently break or require defensive guards everywhere.

The fix I'd propose is to treat `spec.type` as a proper JSON Schema discriminant for `kind: API`:

```json
{
  "if": {
    "properties": { "type": { "not": { "const": "mcp-server" } } }
  },
  "then": {
    "required": ["definition"]
  }
}
```

This keeps `definition` **required** for all existing API kinds (`openapi`, `grpc`, `graphql`, `asyncapi`) and makes it optional **only** when `type: mcp-server`, where `remotes` carries the structured transport config instead. Existing entities and validators are unaffected — they never see `type: mcp-server` unless they opt in.

The schema for the MCP-specific fields would then be:

```yaml
spec:
  type: mcp-server
  lifecycle: production
  owner: group:payments-platform
  remotes:
    - name: payments-mcp
      transport: sse
      url: https://payments.internal/mcp
      protocolVersion: "2025-11-05"
```

On the companion pieces I mentioned in #32062 — I'm happy to take the `McpServerDiscoveryProcessor` as a separate, non-breaking PR. It would:
- Read a `claude_desktop_config.json`-style JSON (or a custom annotation on a `kind: Location` entity)
- Emit `kind: API, type: mcp-server` entities into the catalog
- Wire into the existing `CatalogBuilder.addEntityProvider()` pattern behind a feature flag

That keeps this PR focused on the model change and the processor as a follow-on. Let me know if the discriminated-union approach addresses your concern and I'll get started.

benjdlambert:
There's another attempt at this using the catalog layers system #34016

packages/catalog-model/report.api.md
gabemontero: I'm taking https://github.com/backstage/backstage/pull/33207#issuecomment-4289549140 as answering as "yes" minimally my questions in the second and third bullet

benjdlambert:
Superceded by layers implementation

PR #33214, opened 2026-03-09
Fix flaky createBarrier test in catalog-backend

freben:

The `createBarrier` test "abandons a wait after the timeout expires" was flaky on CI. It used real `setTimeout` calls with tight 50ms margins between barrier timeouts and tick delays. On loaded CI machines, timer coalescing could fire both in the same event loop turn, causing `fn2` to be called before the assertion.

Switched to `jest.useFakeTimers()` with `jest.advanceTimersByTime()` for deterministic control, and added `afterEach` cleanup to restore real timers.

### DIFF
--- a/plugins/catalog-backend/src/processing/TaskPipeline.test.ts
+++ b/plugins/catalog-backend/src/processing/TaskPipeline.test.ts
@@ -121,8 +121,13 @@ describe('startTaskPipeline', () => {
 });
 
 describe('createBarrier', () => {
-  const tick = (millis: number) =>
-    new Promise(resolve => setTimeout(resolve, millis));
+  beforeEach(() => {
+    jest.useFakeTimers();
+  });
+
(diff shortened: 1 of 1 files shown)


plugins/catalog-backend/src/processing/TaskPipeline.test.ts
```diff
@@ -121,36 +121,44 @@ describe('startTaskPipeline', () => {
 });
 
 describe('createBarrier', () => {
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
   const tick = (millis: number) =>
```
benjdlambert: Do we still need this then? 😅

freben: yes because other tests were using it - but i told it to do the same for the other tests too (flaky or not) and remove this one, good call

PR #33224, opened 2026-03-09
catalog-backend: fix catalog multi-ancestry

### LINKED ISSUE #31315: 🐛 Bug Report: Entities cannot have multiple ancestors anymore?
### 📜 Issue Labels

### 🔎 Search Terms

```plain
catalog ancestors refresh_state_references
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

GitHub

### 📝 Description & Context

When multiple locations point to the same entity file the location that is registered last becomes that entity sole ancestor. This behavior used to be different.

### 👍 Expected Behavior

Previously, if multiple locations pointed to the same entity definition file both of them would become ancestors of that entity. Quick code research indicated a [change in the DefaultProcessingDatabase.ts](https://github.com/backstage/backstage/commit/11d0b7526afc737a977fc3a882e751e37350ddd9#diff-111bd797b4de91143a5d51e3fbf24b88d4a9417357b22db7ecd333067389ef89), where previously only existing source references were replaced but now even existing target references are removed too.

From what I understand this makes multi-ancestry impossible now, since [`EntitiesCatalog.entityAncestry` relies on the `refresh_state_references` table](https://github.com/backstage/backstage/blob/2aaf01a0fa38224bc8ce3ace7552cdaa7b0f434a/plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts#L642) to find ancestors and now there can only be one.

This might be intentional, of course, but it might as well be a bug, since it makes it harder to keep track of the origins of an entity when its ancestor can change based on what location was added last. Also this can limit certain ancestry-based implementations.

If it is not a bug, and single ancestry is now the way, could you please explain this shift?

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

_No response_

stanislav-c:
This pull request fixes a bug in the catalog backend where multiple locations pointing to the same entity file could incorrectly result in only the last processed location being recorded as the entity's ancestor. Now, when multiple parents with null location keys reference the same child entity, all are preserved as ancestors, and exclusive ownership is only taken when a specific location key is set. The changes include updates to both the processing and provider database logic, as well as new tests to verify the correct ancestry behavior.

fixes https://github.com/backstage/backstage/issues/31315

### DIFF
--- a/plugins/catalog-backend/src/database/DefaultProcessingDatabase.ts
+++ b/plugins/catalog-backend/src/database/DefaultProcessingDatabase.ts
@@ -320,6 +320,10 @@ export class DefaultProcessingDatabase implements ProcessingDatabase {
 
     // Keeps track of the entities that we end up inserting to update refresh_state_references afterwards
     const stateReferences = new Array<string>();
+    // Tracks entities that are transitioning from null to non-null location key.
+    // These are entities being "claimed" by a specific location for the first time,
+    // so we need to remove any existing weak (null-key) references from other parents.
+    const stealableReferences = new Array<string>();
 
     // Upsert all of the unprocessed entities into the refresh_state table, by
     // their entity ref.
@@ -327,14 +331,18 @@ export class DefaultProcessingDatabase implements ProcessingDatabase {
       const entityRef = stringifyEntityRef(entity);
       const hash = generateStableHash(entity);
 
-      const updated = await updateUnprocessedEntity({
-        tx,
-        entity,
-        hash,
-        locationKey,
-      });
+      const { updated, claimedFromNullLocationKey } =
+        await updateUnprocessedEntity({
+          tx,
+          entity,
+          hash,
+          locationKey,
+        });
       if (updated) {
         stateReferences.push(entityRef);
+        if (claimedFromNullLocationKey) {
+          stealableReferences.push(entityRef);
+        }
         continue;
       }
 
@@ -378,13 +386,21 @@ export class DefaultProcessingDatabase implements ProcessingDatabase {
       }
     }
 
-    // Lastly, replace refresh state references for the originating entity and any successfully added entities
-    await tx<DbRefreshStateReferencesRow>('refresh_state_references')
+    // Replace refresh state references for the originating entity and any
+    // successfully added entities that are transitioning ownership.
+    const deletionQuery = tx<DbRefreshStateReferencesRow>(
+      'refresh_state_references',
+    )
       // Remove all existing references from the originating entity
-      .where({ source_entity_ref: options.sourceEntityRef })
-      // And remove any existing references to entities that we're inserting new references for
-      .orWhereIn('target_entity_ref', stateReferences)
-      .delete();
+      .where({ source_entity_ref: options.sourceEntityRef });
+    if (stealableReferences.length > 0) {
+      // Also remove references to entities that are being claimed from null
+      // location key (weak ownership) by a specific location key (strong
+      // ownership). This ensures that previously weak references from other
+      // parents are replaced when a location takes ownership of an entity.
+      deletionQuery.orWhereIn('target_entity_ref', stealableReferences);
+    }
+    await deletionQuery.delete();
     await tx.batchInsert(
       'refresh_state_references',
       stateReferences.map(entityRef => ({
--- a/plugins/catalog-backend/src/database/DefaultProcessingDatabase.ts
+++ b/plugins/catalog-backend/src/database/DefaultProcessingDatabase.ts
@@ -388,19 +388,19 @@ export class DefaultProcessingDatabase implements ProcessingDatabase {
 
     // Replace refresh state references for the originating entity and any
     // successfully added entities that are transitioning ownership.
-    const deletionQuery = tx<DbRefreshStateReferencesRow>(
-      'refresh_state_references',
-    )
+    await tx<DbRefreshStateReferencesRow>('refresh_state_references')
       // Remove all existing references from the originating entity
-      .where({ source_entity_ref: options.sourceEntityRef });
-    if (stealableReferences.length > 0) {
-      // Also remove references to entities that are being claimed from null
-      // location key (weak ownership) by a specific location key (strong
-      // ownership). This ensures that previously weak references from other
-      // parents are replaced when a location takes ownership of an entity.
-      deletionQuery.orWhereIn('target_entity_ref', stealableReferences);
-    }
-    await deletionQuery.delete();
+      .where({ source_entity_ref: options.sourceEntityRef })
+      .modify(qb => {
+        if (stealableReferences.length > 0) {
+          // Also remove references to entities that are being claimed from null
+          // location key (weak ownership) by a specific location key (strong
+          // ownership). This ensures that previously weak references from other
+          // parents are replaced when a location takes ownership of an entity.
+          qb.orWhereIn('target_entity_ref', stealableReferences);
+        }
+      })
+      .delete();
     await tx.batchInsert(
       'refresh_state_references',
       stateReferences.map(entityRef => ({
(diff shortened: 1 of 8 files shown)


stanislav-c:
@freben, did I [understand you](https://github.com/backstage/backstage/issues/31315#issuecomment-3364767732) correctly? 🙂

OSfrog:
Nice work, this has caused some issue for us

schultzp2020 (approved):
Solid fix for the multi-ancestry regression.

plugins/catalog-backend/src/database/DefaultProviderDatabase.ts
```diff
@@ -167,6 +170,11 @@ export class DefaultProviderDatabase implements ProviderDatabase {
           if (ok) {
             await tx<DbRefreshStateReferencesRow>('refresh_state_references')
               .where('target_entity_ref', entityRef)
+              .modify(qb => {
```
freben: Ooh, this is a nice way of embedding optional modifications to the query without having to use extra variables and stuff. 👍

plugins/catalog-backend/src/database/DefaultProcessingDatabase.ts
```diff
@@ -388,13 +396,21 @@ export class DefaultProcessingDatabase implements ProcessingDatabase {
...
+    const deletionQuery = tx<DbRefreshStateReferencesRow>(
+      'refresh_state_references',
+    )
       // Remove all existing references from the originating entity
-      .where({ source_entity_ref: options.sourceEntityRef })
-      // And remove any existing references to entities that we're inserting new references for
-      .orWhereIn('target_entity_ref', stateReferences)
-      .delete();
+      .where({ source_entity_ref: options.sourceEntityRef });
+    if (stealableReferences.length > 0) {
```
freben: this could also have been wrapped in a `.modify` right, leading to less code rewrite around it?

plugins/catalog-backend/src/database/operations/refreshState/updateUnprocessedEntity.ts
```diff
@@ -29,12 +33,26 @@ export async function updateUnprocessedEntity(options: {
...
+  // references from other parents that only had weak ownership.
+  let claimedFromNullLocationKey = false;
+  if (locationKey) {
+    const existingNullKeyRow = await tx<DbRefreshStateRow>('refresh_state')
+      .where({ entity_ref: entityRef })
+      .whereNull('location_key')
+      .select('entity_id')
+      .first();
+    claimedFromNullLocationKey = existingNullKeyRow !== undefined;
+  }
```
freben: For bonus points - the code you have here could be for mysql and sqlite (because of their lesser capabilities), but for pg you could have a single statement with `.returning('OLD.location_key')` to get transactional safety

plugins/catalog-backend/src/database/DefaultProviderDatabase.ts
```diff
@@ -167,6 +170,11 @@ export class DefaultProviderDatabase implements ProviderDatabase {
...
+                if (!claimedFromNullLocationKey) {
+                  qb.andWhere({ source_key: options.sourceKey });
```
freben: Hm. To me this makes it look like you'll be deleting `source_key: options.sourceKey, target_entity_ref: entityRef` and then immediately inserting it again on the next statement. Or am I misreading that?

Something is off here.

Should it just do _nothing_ if `!claimedFromNullLocationKey`?

Or should this actually be split into two cases, based on whether it was the `updateUnprocessedEntity` or `insertUnprocessedEntity` path that succeeded because they have different needs in terms of what references need to be inserted?

freben (dismissed):

plugins/catalog-backend/src/database/operations/refreshState/updateUnprocessedEntity.ts
stanislav-c: After reading on the topic, it looks like Knex does not support `OLD.*` inside `.returning()` and would require using a raw query.
Would using [a sub-transaction like this](https://github.com/backstage/backstage/pull/33224/changes/6570f4f4390371ca092d2239c0e6d00bd6f1a0de#diff-c6a27b4502a0a8fd95248dcafe7c8ddc544c423edf785fa194c7ec21968345cb) be sufficient here or would it impact the performance (write concurrency) too much?

freben:
Hey! I rudely rebased and implemented the change I was looking for - let me ask for a second opinion from another maintainer before merging.

plugins/catalog-backend/src/database/operations/refreshState/updateUnprocessedEntity.ts
```diff
@@ -29,32 +33,67 @@ export async function updateUnprocessedEntity(options: {
...
-        return inner.whereNull('location_key');
+  return tx.transaction(
+    async subTx => {
+      // When claiming an entity with a non-null location key, check whether it
+      // currently has a null location key. If so, the update will transition it
+      // from weak (null) to strong (keyed) ownership, and we'll need to steal
+      // references from other parents that only had weak ownership.
+      let claimedFromNullLocationKey = false;
+      if (locationKey) {
+        const existingNullKeyRow = await subTx<DbRefreshStateRow>(
```
Rugvip: It's a bit unfortunate that we need to do this on essentially every update, while the case of claiming a null key is actually very rare. Tricky to avoid though, but one potential idea that came to mind was to optimistically do the existing update, but without `.whereNull('location_key')` for a provided key, but if that fails then fall back to also trying an insert with the override. That means we'll hit that branch whenever there are location key conflicts or first inserts, but over the lifetime of the catalog I believe those cases should be a lot more rare. With that structure we might also be able to lower the isolation level? 🤔

Rugvip (approved):
Thank you! 👍 

Looks good, but had a potential idea for a tweak to the update logic. Didn't dig deep enough to know for sure that it makes sense, but perhaps you can glance at that as well @freben and decide on a merge

stanislav-c:
> Hey! I rudely rebased and implemented the change I was looking for - let me ask for a second opinion from another maintainer before merging.

No worries! Thanks for pushing this over the finish line!

stanislav-c:
Guys, when do you think this can be released?

plugins/catalog-backend/src/database/operations/refreshState/updateUnprocessedEntity.ts
stanislav-c: Should this be included in this PR, or would it be better handled separately? Multi-ancestry support has been broken for a while, and our org would be very grateful if this fix could be released soon. 🙂

stanislav-c:
Not stale - still relevant.

freben (changes requested):
Thanks for working through this. The underlying fix is still worthwhile, and the weak-vs-strong ownership semantics here match the intended behavior: preserve multiple parents while `location_key` is null, and remove the weak references only when an entity transitions from null to a concrete location key.

I do not think the current transition-detection implementation is safe to merge yet, though:

1. `updateUnprocessedEntity` now opens a nested transaction for every deferred entity, including ordinary unkeyed updates. In production it is already called from an outer processing/provider transaction, so this adds a `SAVEPOINT` / `RELEASE` round trip per entity and can become significant for large locations.
2. The read-before-write sequence is not atomic on SQLite. Two concurrent keyed claims can both observe the null key, after which one can fail with `SQLITE_BUSY`. That error is not classified as a retryable deadlock and the provider path may log and swallow it as an individual entity failure.

Could we make the transition detection write-atomic without a nested transaction? For a keyed update, first issue the update with `WHERE entity_ref = ? AND location_key IS NULL`; if it affects one row, it was the strong claim. Otherwise issue the normal update with `WHERE entity_ref = ? AND location_key = ?`. The unkeyed path can retain its existing single conditional update. This preserves the semantics of this PR while making the claim result correspond to the write that actually succeeded on every supported database.

I prefer this focused semantic fix over folding in the broader reference-sync and migration work from #34606. That optimization can be evaluated separately once takeover semantics and migration recovery are settled.

stanislav-c:
> Could we make the transition detection write-atomic without a nested transaction? For a keyed update, first issue the update with `WHERE entity_ref = ? AND location_key IS NULL`; if it affects one row, it was the strong claim. Otherwise issue the normal update with `WHERE entity_ref = ? AND location_key = ?`. The unkeyed path can retain its existing single conditional update. This preserves the semantics of this PR while making the claim result correspond to the write that actually succeeded on every supported database.

@freben thanks for pointing this out. I've now made the conditional write atomic and removed nested transactions.

PR #33227, opened 2026-03-09
BUI - Add StatCard component for displaying metrics and KPIs

AmbrishRamachandiran:
## Summary

Adds `StatCard` component for displaying dashboard metrics and KPIs with status indicators and trends.

## Why

Teams frequently need to show service health, deployment counts, and other metrics in their Backstage portals. This component provides a consistent, accessible way to display these key statistics with visual status indicators (success/warning/error) and trend data, eliminating the need for custom implementations.
image
image


cdedreuille:
@AmbrishRamachandiran Thanks for contributing to BUI. While this pattern is indeed something we are exploring, I'm not sure we are considering it as a component but most likely a recipe or guideline instead. It is too specific to be an actually component right now.

PR #33253, opened 2026-03-10
feat(skills): add agent skills plugins and install flow

drodil:

add plugin for registering, listing and using agent skills from backstage instance using `npx skills add` cli functionality.

relates a bit to #33173

image
image


drodil:
@awanlin what are the plans to make catalog support agent skills? And how would it work? Do you need to introduce a new catalog-info file for each skill, or is it handled by some custom entity provider scanning for skills in repositories? How would the installation of these skills work through catalog entities? I mean, it would be theoretically possible to also add support for `skills.sh` for the catalog 🤔 

And yes, I was wondering if this would be better placed in the community plugins, but there is a big BUT; I cannot be the maintainer for that, so someone would have to take over the ownership (which I am happy to give away).

awanlin:
I was thinking that this would potentially set some of the ground work for skills: https://github.com/backstage/backstage/pull/33207

awanlin:
I'd say for now let's just keep things as they are, maybe put as Draft, and work towards a consensus about the direction. I don't like to make what I feel is a large decision like this without some general agreement from the other maintainers. They will likely have their own additional context to share which could also help my thinking too. 👍

drodil:
> I was thinking that this would potentially set some of the ground work for skills: #33207

Yep, been following this, and it makes sense to have the MCP servers in the catalog. For skills, I am not so sure. Should we try to find some alignment for this in some of the SIG meetings?

benjdlambert:
If you can hold off a little bit on storage parts until KubeCon next week, we've got some things in this space with an upcoming RFC which are pertinent to this.

plugins/skills-node/src/service/DefaultSkillsRegistryService.ts
```diff
@@ -0,0 +1,236 @@
...
+  private constructor(
+    client: SkillsClient,
+    auth: AuthService,
+    logger: LoggerService,
+    urlReader: UrlReaderService,
+  ) {
+    this.client = client;
+    this.auth = auth;
+    this.logger = logger;
+    this.urlReader = urlReader;
```
drodil, line 66: We could also support skills from app-conflg:

```yaml
skills:
  - https://github.com/backstage/community-plugins/blob/main/.claude/skills/mui-to-bui-migration/SKILL.md
```
and register those automatically.

drodil:
Closing this in favor of AIContext in catalog

hardik-id:
Hello, how do I follow this topic? I am also curious to see how we manage agents/skills catalog?

drodil:
@hardik-id please see https://github.com/backstage/backstage/issues/33575

PR #33255, opened 2026-03-10
feat: add support for wildcard patterns in package discovery configuration

stanislav-c:
Added support for wildcard patterns using `*` in `backend.packages.include` and `backend.packages.exclude` as well as `app.packages.include` and `app.packages.exclude`. For example, `@backstage/plugin-catalog-backend-module-*` will now match all packages starting with
that prefix, and `@backstage/plugin-*-backend-module-*` will match all module packages from all backend plugins. This allows for more flexible and concise package discovery configurations, especially in cases where there are multiple related packages that share a common naming pattern.


packages/backend-defaults/src/PackageDiscoveryService.test.ts
```diff
@@ -0,0 +1,241 @@
...
+    expect(matcher.matches('', '*')).toBe(true);
+  });
+
+  it('matches empty pattern', () => {
+    expect(matcher.matches('', '')).toBe(true);
+    expect(matcher.matches('foo', '')).toBe(false);
+  });
+
+  it('caches regex objects', () => {
+    // This test checks that cache is used, but cannot directly inspect WeakRef internals
```
schultzp2020: The implementation was updated from `WeakRef` to a regular `Map`, but this still references WeakRef internals.

```suggestion
    // This test verifies that repeated calls with the same pattern work correctly
```

schultzp2020 (approved):
Nice, clean addition. One stale comment inline, otherwise looks good.

packages/backend-defaults/src/PackageDiscoveryService.ts
```diff
@@ -78,10 +78,31 @@ async function findClosestPackageDir(
   );
 }
 
+/** @internal */
+export class PatternMatcher {
```
Rugvip: I think the pattern of a stateful matcher is maybe a little bit overkill, you could also map over the input filters once and convert them to functions that do the matching:

```ts
filters.map(filter => filter.includes('*') ? {const pattern = new RegExp(...); return name => pattern.test(name) } : name => name === filter
```

Rugvip (changes requested):
Yep makes sense 👍 

Wondering if you could add it for `app.packages` as well? I think it's gonna be a bit too confusing to have it supported in one but not the other

packages/backend-defaults/src/PackageDiscoveryService.ts
stanislav-c: Good point. I replaced PatternMatcher with pure functions capturing unique regular expressions in closures. Also implemented the same behavior in FE. This leads to duplicate code however. I could not find an appropriate shared isomorphic package for this. Introducing a new one just for a couple of functions seems excessive. What would you suggest?

stanislav-c: @Rugvip what would be the best way to proceed here?

PR #33260, opened 2026-03-10
refactor(catalog,api-docs): migrate entity table cards to BUI

Naycon:

Migrates 13 entity relation table cards from MUI/InfoCard + core-components Table to BUI. All 13 cards follow the same pattern: fetch related entities via a relation type, display them in a table inside a card. Rather than migrating each card individually, this PR introduces two new shared components in `@backstage/plugin-catalog-react/alpha`:

- **`EntityDataTable`** — BUI Table wrapper for entity data with column factories and presets for common entity types (component, resource, system, domain). Uses `useTable` for client-side sorting.
- **`EntityRelationCard`** — card shell with built-in data fetching via `useRelatedEntities`, error handling via BUI `Alert`, and empty state rendering. Displays results in `EntityDataTable` inside an `EntityInfoCard`.
- **`entityColumnPresets`** — grouped column presets and help links for component, resource, system, and domain entity types.

These are exported under `/alpha` to allow iteration before stabilizing.

Each of the 13 existing cards becomes a thin config wrapper around `EntityRelationCard`, passing only its relation type, entity kind, column preset, and translation keys.

**Catalog plugin (8 cards)** — previously used `RelatedEntitiesCard` as a shared base:
- `HasComponentsCard`, `HasResourcesCard`, `HasSubcomponentsCard`, `HasSubdomainsCard`, `HasSystemsCard`
- `DependsOnComponentsCard`, `DependsOnResourcesCard`, `DependencyOfComponentsCard`

**Api-docs plugin (5 cards)** — previously duplicated the loading/error/table pattern inline:
- `ConsumedApisCard`, `ProvidedApisCard`, `HasApisCard`
- `ConsumingComponentsCard`, `ProvidingComponentsCard`

**Backwards compatibility:**

All 13 cards accept both old and new props via an inline union type (`FooCardProps | FooCardLegacyProps`). When legacy props (`variant`, `columns`, `tableOptions`) are passed, the old MUI-based implementation is used. When no legacy props are passed (default), the new BUI-based `EntityRelationCard` renders. The old props are deprecated.

- `RelatedEntitiesCard` in `@backstage/plugin-catalog` is deprecated in favor of `EntityRelationCard` from `@backstage/plugin-catalog-react/alpha`
- Old api-docs preset functions (`createSpecApiTypeColumn`, `getApiEntityColumns`) are deprecated

**Other changes:**
- Column sorting support via `EntityColumnConfig.sortValue` — all built-in columns are sortable, matching the old material-table behavior
- Error states use BUI `Alert` instead of core-components `ResponseErrorPanel`
- Empty state help links use BUI `Link` instead of core-components `Link`
- API definition button uses BUI `ButtonIcon` with Remix icon instead of MUI `ToggleButton`
- Translation messages now have consistent trailing periods

### Screenshots

| Before | After |
|--------|-------|
| image | image |
| image | image |
| image | image |

With an error:
image

### DIFF
--- a/plugins/catalog/src/components/RelatedEntitiesCard/RelatedEntitiesCard.tsx
+++ b/plugins/catalog/src/components/RelatedEntitiesCard/RelatedEntitiesCard.tsx
@@ -70,7 +70,7 @@ export type RelatedEntitiesCardProps<T extends Entity> = {
  * this card as its implementation with some of the props set to the appropriate
  * values.
  *
- * @deprecated Use {@link @backstage/plugin-catalog-react#EntityRelationCard} instead.
+ * @deprecated Use {@link @backstage/plugin-catalog-react/alpha#EntityRelationCard} instead.
  * @public
  */
 export const RelatedEntitiesCard = <T extends Entity>(
--- a/.changeset/migrate-catalog-entity-cards-to-bui.md
+++ b/.changeset/migrate-catalog-entity-cards-to-bui.md
@@ -4,4 +4,4 @@
 
 Migrated entity table cards (`HasComponentsCard`, `HasResourcesCard`, `HasSubcomponentsCard`, `HasSubdomainsCard`, `HasSystemsCard`, `DependsOnComponentsCard`, `DependsOnResourcesCard`, `DependencyOfComponentsCard`) to use BUI when no legacy props are passed. The old `variant`, `columns`, and `tableOptions` props are deprecated but still supported — passing any of them triggers the legacy MUI-based rendering. The new `columnConfig` prop accepts `EntityColumnConfig[]` for BUI-based rendering.
 
-`RelatedEntitiesCard` is deprecated — use `EntityRelationCard` from `@backstage/plugin-catalog-react` instead.
+`RelatedEntitiesCard` is deprecated — use `EntityRelationCard` from `@backstage/plugin-catalog-react/alpha` instead.
--- a/plugins/api-docs/src/components/ApisCards/presets.tsx
+++ b/plugins/api-docs/src/components/ApisCards/presets.tsx
@@ -96,11 +96,14 @@ function createSpecApiTypeColumnConfig(
   return {
     id: 'apiType',
     label: t('apiEntityColumns.typeTitle'),
+    isSortable: true,
     cell: entity => (
       <Cell>
         <ApiTypeTitle apiEntity={entity as unknown as ApiEntity} />
       </Cell>
     ),
+    sortValue: entity =>
+      (entity.spec as Record<string, string> | undefined)?.type ?? '',
   };
 }
 


plugins/catalog/src/components/RelatedEntitiesCard/RelatedEntitiesCard.tsx
```diff
@@ -70,6 +70,7 @@ export type RelatedEntitiesCardProps<T extends Entity> = {
  * this card as its implementation with some of the props set to the appropriate
  * values.
  *
+ * @deprecated Use {@link @backstage/plugin-catalog-react#EntityRelationCard} instead.
```
Naycon: Not entirely sure if this is how we want to go about this, or if we'd rather remove this directly.

schultzp2020 (commented):
The PR checklist notes tests aren't added yet. Since `EntityDataTable` and `EntityRelationCard` are new public API surface in `catalog-react`, basic rendering and error-state coverage would be good to have before merging. Thoughts?

Naycon:
> The PR checklist notes tests aren't added yet. Since `EntityDataTable` and `EntityRelationCard` are new public API surface in `catalog-react`, basic rendering and error-state coverage would be good to have before merging. Thoughts?

Oh yes, right you are! Will add tests for those. Good catch, thanks @schultzp2020!

Edit: Added tests for them now.

plugins/api-docs/src/components/ApisCards/ConsumedApisCard.tsx
```diff
@@ -35,15 +37,39 @@ import {
...
+  variant?: InfoCardVariants;
+  /** @deprecated Use `columnConfig` instead. */
   columns?: TableColumn<ApiEntity>[];
+  /** @deprecated Use `columnConfig` instead. */
   tableOptions?: TableOptions;
-}) => {
+}
+
+/** @public */
+export type ConsumedApisCardProps =
```
Rugvip: Hmm, is it possible to declare these as an inline union instead? It's a bit unfortunate to have to add the legacy props type, but it's especially unfortunate to end up with `ConsumedApisCardProps = `ConsumedApisCardBaseProps` in the future

.changeset/add-entity-data-table-and-relation-card.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-react': minor
+---
+
+Added `EntityDataTable` component for rendering entity data in a BUI Table, and `EntityRelationCard` component for displaying related entities in a card with built-in data fetching. These replace `EntityTable` and `RelatedEntitiesCard` (from `@backstage/plugin-catalog`) respectively, providing a unified BUI-based pattern for entity table cards. Column presets (`componentColumnConfig`, `systemColumnConfig`, etc.) are included for common entity types.
```
Rugvip: Hmm, should we perhaps add these under `/alpha` to start?

plugins/catalog-react/report.api.md
```diff
@@ -187,6 +189,13 @@ export const columnFactories: Readonly<{
   createSpecTypeColumn<T extends Entity>(): TableColumn<T>;
 }>;
 
+// @public (undocumented)
+export const componentColumnConfig: EntityColumnConfig[];
```
Rugvip: I'd like to avoid all of these separate exports, using a pattern similar to the existing one instead

plugins/catalog-react/report.api.md
```diff
@@ -187,6 +189,13 @@ export const columnFactories: Readonly<{
...
+
+// @public (undocumented)
+export const componentEntityHelpLink =
```
Rugvip: Bit of an odd one, do we really need an export for this? 🤔 

If we really want to keep them around, I think it's best to group them up somehow

Rugvip (changes requested):

plugins/catalog/report.api.md
```diff
@@ -525,54 +558,88 @@ export function hasRelationWarnings(
...
   tableOptions?: TableOptions;
   // (undocumented)
   title?: string;
-  // (undocumented)
+  // @deprecated (undocumented)
   variant?: InfoCardVariants;
 }
 
 // @public (undocumented)
 export interface HasSystemsCardProps {
```
Rugvip, line 638: Hmm, this ends up being a breaking change now technically. Meh, I think it's fine to argue that these are input types that are only consumed via the component which has been updated in a non-breaking way. Can expect basically no impact from this and that pattern is supported by our versioning policy.

PR #33267, opened 2026-03-10
feat(ui): centralize routing in BUIProvider

Naycon:

`BUIProvider` now auto-detects if it's inside a React Router context and
provides React Aria's `RouterProvider` for client-side navigation. This
removes the per-component routing wrappers (`InternalLinkProvider`,
`createRoutingRegistration`) that each BUI component previously needed.

Changes:
- **`@backstage/ui`**: Added `BUIRouterProvider` export. `BUIProvider`
  auto-detects Router context via `useInRouterContext()`. Removed
  `InternalLinkProvider` directory. Updated Link, ButtonLink, Tabs,
  Menu, TagGroup, and Table.Row to rely on the central provider.
- **`@backstage/plugin-app`**: Moved `BUIProvider` from outside
  `AppRouter` to inside it, so it detects the Router context.
- **`@backstage/core-app-api`**: Added `BUIRouterProvider` inside the
  legacy `AppRouter` to provide routing for BUI components.
- **docs-ui**: Updated BUIProvider documentation to cover routing and
  the requirement to render inside a React Router context.


Naycon:
Thanks @cdedreuille! I'll respond in part here:

---

> * Should we move BUIProvider outside of the analytics folder since it is not tied to analytics only anymore? If we still want to have both then we could perhaps have `BUIAnalyticsProvider` as well and unify both into `BUIProvider`

Ah yes, good points. I'll look into this, but in general we should see the BUIRouterProvider as the exception here (see below). But you're right we should move the general provider to someplace not called `analytics`, since it's doing more than that.

---

> * Why do we need to have both `BUIProvider` and `BUIRouterProvider`? Can't we have both in the same provider?

We need BUIRouterProvider separately because of how the legacy app is setup and work. We need to add the router provider in a place where a React Router router is provided (e.g. BrowserRouter). In the legacy app, this is done in the AppRouter, which does not feel like the right place for a general BUIProvider, and downstream from the AppRouter, there's no good place to put it. I asked Claude to visualize the tree for us, this is what it looks like for the legacy app:

```
  Provider (AppManager.getProvider())
    → ApiProvider
      → BUIProvider          ← outside Router
        → AppContextProvider
          → ...
            → AppRouter (AppManager.getRouter())
              → RouterComponent (BrowserRouter)  ← Router starts here
                → RouteTracker
                → children
```

Once the legacy app isn't supported anymore, we can get rid of the BUIRouterProvider, assuming there's no custom app shell case we want to support.

---

> * How about Storybook? Should we add this provider globally?

Right now I don't think it needs to be added there no. We don't do tracking, and we don't provide a React Router router anywhere central inside of which we could place it either. 

But you're right in that we should add the BUIProvider to the stories where we use the memory router. Good catch, I'll fix that!

packages/ui/src/routing/BUIRouterProvider.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { RouterProvider } from 'react-aria-components';
+import { useNavigate, useHref } from 'react-router-dom';
+
+/**
+ * Provides React Aria's RouterProvider for client-side navigation.
+ * Must be rendered within a React Router context.
+ *
+ * @public
+ */
+export function BUIRouterProvider({
```
Rugvip: How come we need both this and the BUIProvider? It seems we'd be able to move it down the same way in `core-app-api`

If we really need both, perhaps `BUIProvider.Router` to avoid multiple top-level exports? 🤔

Naycon: Swapped to providing BUIProvider instead of the router one, and removed that from the public API.

PR #33280, opened 2026-03-11
fix(catalog): make catalog entity deletion transactional

drodil:

additionally, make the catalog build respect the deferred stitching mode after deleting entities.

### DIFF
--- a/plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
+++ b/plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
@@ -525,94 +525,96 @@ export class DefaultEntitiesCatalog implements EntitiesCatalog {
   }
 
   async removeEntityByUid(uid: string): Promise<void> {
-    const dbConfig = this.database.client.config;
+    const relationPeerRefs = await this.database.transaction(async tx => {
+      const dbConfig = tx.client.config;
 
-    // Clear the hashed state of the immediate parents of the deleted entity.
-    // This makes sure that when they get reprocessed, their output is written
-    // down again. The reason for wanting to do this, is that if the user
-    // deletes entities that ARE still emitted by the parent, the parent
-    // processing will still generate the same output hash as always, which
-    // means it'll never try to write down the children again (it assumes that
-    // they already exist). This means that without the code below, the database
-    // never "heals" from accidental deletes.
-    if (dbConfig.client.includes('mysql')) {
-      // MySQL doesn't support the syntax we need to do this in a single query,
-      // http://dev.mysql.com/doc/refman/5.6/en/update.html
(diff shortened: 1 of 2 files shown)


plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
```diff
@@ -525,94 +529,106 @@ export class DefaultEntitiesCatalog implements EntitiesCatalog {
...
+
+      // Clear the hashed state of the immediate parents of the deleted entity.
+      // This makes sure that when they get reprocessed, their output is written
+      // down again. The reason for wanting to do this, is that if the user
+      // deletes entities that ARE still emitted by the parent, the parent
+      // processing will still generate the same output hash as always, which
+      // means it'll never try to write down the children again (it assumes that
+      // they already exist). This means that without the code below, the database
+      // never "heals" from accidental deletes.
+      if (dbConfig.client.includes('mysql')) {
```
schultzp2020, line 539: The old code had a helpful comment here explaining *why* MySQL needs a separate code path (`// MySQL doesn't support the syntax we need to do this in a single query, http://dev.mysql.com/doc/refman/5.6/en/update.html`). Might be worth keeping for future readers who wonder about the branching.

schultzp2020 (approved):
Nice work coordinating the deferred stitching path. One small observation inline, nothing blocking.

plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
drodil: Yep, I add it back here.

plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
```diff
@@ -525,94 +529,108 @@ export class DefaultEntitiesCatalog implements EntitiesCatalog {
...
-      );
+      await tx<DbRefreshStateRow>('refresh_state')
+        .where('entity_id', uid)
+        .delete();
 
-    await this.database<DbRefreshStateRow>('refresh_state')
-      .where('entity_id', uid)
-      .delete();
+      const refs = new Set(relationPeers.map(p => p.ref));
+      if (refs.size > 0 && this.stitchingStrategy.mode === 'deferred') {
```
freben: DefaultStitcher [already deals with](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/stitching/DefaultStitcher.ts#L92) all of this. It's meant to work as a facade, hiding the fact whether you are in deferred or immediate mode. I think you can revert everything except the transaction wrapping

drodil: Oh yeah, missed that! Will roll back!

PR #33282, opened 2026-03-11
`feat(auth)`: Migrate `ConsentPage` to `@backstage/ui`

benjdlambert:
Migrating the Auth plugin to Backstage UI instead :tada:


plugins/auth/src/components/ConsentPage/ConsentPage.tsx
```diff
@@ -179,70 +133,67 @@ export const ConsentPage = () => {
...
+            <Alert
+              status="warning"
+              icon
+              title="Security Notice"
+              description={
+                <>
+                  By authorizing this application, you are granting it access to
+                  your {appTitle} account. The application will receive an
+                  access token that allows it to act on your behalf.
+                  <div className={styles.callbackUrl}>
```
Naycon, line 163: Just checking, there's no good corresponding [`Text` variant, color and weight](https://ui.backstage.io/components/text#variants) that'd get you to where you want here? What you have will work, but it might be safer to keep in line with BUI when you can.

plugins/auth/src/components/ConsentPage/ConsentPage.tsx
```diff
@@ -16,99 +16,48 @@
...
-  title: string;
-  children: React.ReactNode;
-}) => (
-  <Page themeId="tool">
-    <Header title={title} />
-    <Content>{children}</Content>
-  </Page>
+const ConsentPageLayout = ({ children }: { children: React.ReactNode }) => (
+  <FullPage>
+    <VisuallyHidden>
```
Naycon, line 42: ❤️

plugins/auth/src/components/ConsentPage/ConsentPage.tsx
benjdlambert: I don't think there is one for monospace. Also wanna be clear that this entire plugin frontend is pure vibes and had no design input. So happy to use alternative patterns if there's something else 😅.

Image

plugins/auth/src/components/ConsentPage/ConsentPage.tsx
```diff
@@ -16,99 +16,44 @@
...
-  children: React.ReactNode;
-}) => (
-  <Page themeId="tool">
-    <Header title={title} />
-    <Content>{children}</Content>
-  </Page>
+const ConsentPageLayout = ({ children }: { children: React.ReactNode }) => (
+  <FullPage>
+    <Container py="8">{children}</Container>
+  </FullPage>
```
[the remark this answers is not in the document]
freben: I think the comment about helmet is interesting - having a useful tab title here is probably a good thing

benjdlambert:
@freben I think that this will be fixed with `PluginHeader` eventually, which will be rendered part of NFS shortly right?

freben:
@benjdlambert well unless this thing disables it right? Honestly it feels like the type of pretty special thing that might want to remove ALL chrome. But then again, we could still have helmet be rendered even when there's no chrome

benjdlambert:
@freben maybe yeah? I'm gonna merge this as is for now, and followup with @Rugvip next week on the plan for the header.

PR #33298, opened 2026-03-11
devtools: Add cancel task operation to scheduled tasks

freben:
Screenshot 2026-03-11 at 20 28 47

## Summary

- Renames `useTriggerScheduledTask` to `useScheduledTasksOperations` with shared `isLoading`/`error` state across both operations
- Adds `cancelTask` alongside `triggerTask`, backed by `cancelScheduledTask` on the `DevToolsApi` interface and client (hitting `POST /.backstage/scheduler/v1/tasks/:id/cancel`)
- Adds a cancel button (stop icon) next to the existing trigger button in the scheduled tasks table UI
- Makes the table a bit more dense

There's more to do here. Refreshes for one, and I don't think the error handling is correct still. But we can do it in stages.

## Test plan

- [ ] Verify trigger task still works as before via the refresh icon button
- [ ] Verify cancel task works on a running task via the new stop icon button
- [ ] Verify both buttons are disabled while an operation is in progress
- [ ] Verify alert notifications appear on success/failure for both operations

### DIFF
--- a/.changeset/devtools-cancel-scheduled-task.md
+++ b/.changeset/devtools-cancel-scheduled-task.md
@@ -3,4 +3,4 @@
 '@backstage/plugin-devtools': patch
 ---
 
-Added `cancelScheduledTask` to the DevTools API and a cancel button to the scheduled tasks UI. The `useTriggerScheduledTask` hook has been renamed to `useScheduledTasksOperations`, now exposing both `triggerTask` and `cancelTask` with shared loading and error state.
+Added `cancelScheduledTask` to the DevTools API and a cancel button to the scheduled tasks UI.


plugins/devtools/src/api/DevToolsClient.ts
```diff
@@ -82,7 +82,31 @@ export class DevToolsClient implements DevToolsApi {
       throw await ResponseError.fromResponse(response);
     }
 
-    return response.json() as Promise<TriggerScheduledTask>;
+    return {};
```
freben, line 85: When successful, there just isn't any json body at all, so this would fail

plugins/devtools/src/components/Content/ScheduledTasksContent/ScheduledTasksContent.tsx
```diff
@@ -292,14 +317,15 @@ export const ScheduledTasksContent = () => {
         </ErrorPanel>
       )}
 
-      {!loading && !error && (
+      {scheduledTasks && (
```
freben, line 320: This would flicker when loading otherwise

awanlin (approved):
Thanks @freben, wasn't expecting this nice addition!  🚀

.changeset/devtools-cancel-scheduled-task.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-devtools-common': patch
+'@backstage/plugin-devtools': patch
+---
+
+Added `cancelScheduledTask` to the DevTools API and a cancel button to the scheduled tasks UI. The `useTriggerScheduledTask` hook has been renamed to `useScheduledTasksOperations`, now exposing both `triggerTask` and `cancelTask` with shared loading and error state.
```
Rugvip: The `useScheduledTasksOperations` change is internal only right? could probably skip

plugins/devtools/src/components/Content/ScheduledTasksContent/ScheduledTasksContent.tsx
```diff
@@ -209,28 +210,52 @@ export const ScheduledTasksContent = () => {
...
-              <RefreshIcon />
-            </IconButton>
-          </Tooltip>
+          <Box display="flex" justifyContent="center">
+            <Tooltip title="Run Task">
+              <IconButton
+                aria-label="Trigger"
+                disabled={isLoading}
+                onClick={async () => {
+                  try {
```
drodil, line 219: Extracting these handlers to functions outside the rendering block would make it readable. Nit-picking here 😆

freben: Yeh I agree. This PR was made as a background task, for context. I noticed a few standout things when submitting and started looking into cleaning up some concepts around all this but there was too much to address right away. There's a couple of responsibility things that need to be cleaned up, the table isn't reloading / reflecting actions, there's a few pre existing bugs - in short, I backed out of trying to make things better in this PR beyond the most important direct error in the `.json()` response parsing.

.changeset/devtools-cancel-scheduled-task.md
freben: Correct, will cut down

freben:
```suggestion
Added `cancelScheduledTask` to the DevTools API and a cancel button to the scheduled tasks UI.
```

PR #33321, opened 2026-03-12
BUI - Add Autocomplete Component with Multiple Display Modes

AmbrishRamachandiran:
## Why
Backstage UI needed a flexible autocomplete/combobox component for scenarios where users need to search and select from a list of options with auto-filtering. While Select handles simple dropdowns, many use cases require:
- Type-to-filter behavior for large option lists
- Visual display modes beyond simple lists (grids, tables, tags)
- Rich option content with icons, descriptions, and structured data
- Custom value support for user-entered options
## What's Added
**New `Autocomplete` component** built on React Aria's ComboBox with 5 display modes:
1. **ListBox** (default) - Simple filtered list
2. **Menu** - Rich menu items with icons and descriptions  
3. **Grid** - Visual card-based selection with configurable columns
4. **Tags** - Compact pill-style options for categories
5. **Table** - Multi-column tabular data with custom rendering

image


AmbrishRamachandiran:
@cdedreuille @Naycon i think this component will add more value for software templates kind of features what do you think?

Naycon:
Hey @AmbrishRamachandiran! Any chance you could raise an issue and provide some more info on why the existing components (Menu, Select, etc.) don't cover what you need, and why covering these needs in a separate component is the solution? Some of the existing components do provide type-to-filter solutions (e.g. Menu and Select), and I recently added a SearchAutocomplete component intended to cover cases where you're not working with a set of items that is known before the user types. 

Having a catch-all component like this might prove difficult to manage, both from the perspective maintaining prop combinations or compositions, and to properly mark it up for accessibility. But I think we need to understand the use case you're trying to cover better before we can decide what a good path forward is.

Issue #32104, opened 2025-12-11
MCP Actions Backend Audit Logging


thepetk:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
- `audit-logging`
- `auditorService`
- `MCP`
- `mcp-actions-backend`
```

### 🗃️ Project Area

Unknown

### 🔖 Need

This issue suggest an enhancement, to add audit logging capabilities to the `mcp-actions-backend` plugin, allowing users & organizations to monitor the MCP server operations within Backstage. I consider this an important topic for security, compliance, and observability reasons.

Currently, the `mcp-actions-backend` plugin does not create audit events for MCP server operations. This creates a gap in visibility for:
- When MCP tools are executed and their status of execution.
- When MCP authentication succeeds or fails
- Which MCP servers are being accessed

**Background**
Backstage already provides an [Auditor Service docs](https://backstage.io/docs/backend-system/core-services/auditor) used already by other backend plugins like `catalog-backend`.

**NOTE**
That's my first enhancement issue here, I wasn't sure if this qualifies to be a "BEP", happy to re-open it if you think it should be :)

### 📝 Proposal

The initial approach of MCP Actions Backend plugin audit events could include the following types:
- `mcp-connection-established`
- `mcp-connection-closed`
- `mcp-connection-error`
- `mcp-auth-attempt`
- `mcp-auth-success`
- `mcp-auth-failure`
- `mcp-http-error`
- `mcp-method-not-allowed`
- `mcp-invalid-session`
- `mcp-tool-discovery`
- `mcp-tool-execution-request`
- `mcp-tool-execution-success`
- `mcp-tool-execution-failure`
- `mcp-tool-not-found`
- `mcp-tool-oidc-discovery`

I have tried to capture what I'm envisioning in my fork here: https://github.com/backstage/backstage/compare/master...thepetk:backstage:mcp-audit-logging?expand=1

### 🔄 Alternatives

There are for sure alternatives, in the end is up to maintainers to decide how/if this should be approached!

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Yes, but I would like some more guidance

thepetk:
[...]

PR #32751, opened 2026-02-08
docs: add mcp-actions-backend to main docs

tesla39:
Fixes #32717

Adds a main docs page for mcp-actions-backend and links it in the docs sidebar.

PR #33339, opened 2026-03-13
`feat(mcp-actions-backend)`: add audit logging for MCP server operations

thepetk:

This PR enhances the `mcp-actions-backend` plugin with audit logging via the Backstage Auditor Service, giving operators visibility into MCP server operations. Connections, tool discovery, and tool execution are now tracked across both `SSE` and `streamable HTTP` transports.

We have now three new event IDs:
- `connection`: when a transport session is established or closed.
- `tool-discovery`: when a client lists available tools
- `tool-execution`: when a tool is invoked

`connection` and `tool-discovery` are using the `low` (default) severity while the `tool-execution` uses `medium`.

The approach in the PR tries to follow what exists already in `catalog-backend` and `scaffolder-backend`, where a single event per operation with `.success()` / `.fail()` captures the outcome, and meta fields are used for differentiation of events. For more info you can check the discussions in the related issue: https://github.com/backstage/backstage/issues/32104

**Note:** This PR was developed with AI assistance (Cursor). However, the implementation, design decisions, etc. were reviewed and verified by the author.

### DIFF
--- a/plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
+++ b/plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
@@ -25,7 +25,6 @@ import {
 } from '@backstage/backend-plugin-api/alpha';
 import {
   AuditorService,
-  AuditorServiceEvent,
   HttpAuthService,
   LoggerService,
 } from '@backstage/backend-plugin-api';
@@ -71,7 +70,7 @@ export const createStreamableRouter = ({
       'network.protocol.name': 'http',
     };
 
-    let connectionEvent: AuditorServiceEvent;
+    let connectionEvent;
     try {
       connectionEvent = await auditor.createEvent({
         eventId: 'connection',
@@ -79,7 +78,7 @@ export const createStreamableRouter = ({
         meta: { transport: 'streamable', actionType: 'established' },
       });
     } catch {
-      // Make audit logging best-effort: fall back to a no-op event if auditing is unavailable.
+      // best-effort
       connectionEvent = { success: async () => {}, fail: async () => {} };
     }
 
@@ -106,9 +105,13 @@ export const createStreamableRouter = ({
         transport.handleRequest(req, res, req.body),
       );
 
-      await connectionEvent.success().catch(() => {});
+      try {
+        await connectionEvent.success();
+      } catch {
+        // best-effort
+      }
 
-      res.on('close', () => {
+      res.on('close', async () => {
         transport.close();
         server.close();
 
@@ -116,14 +119,16 @@ export const createStreamableRouter = ({
 
         sessionDuration.record(durationSeconds, baseAttributes);
 
-        auditor
-          .createEvent({
+        try {
+          const e = await auditor.createEvent({
             eventId: 'connection',
             request: req,
             meta: { transport: 'streamable', actionType: 'closed' },
-          })
-          .then(e => e.success())
-          .catch(() => {});
+          });
+          await e.success();
+        } catch {
+          // best-effort
+        }
       });
     } catch (error) {
       const err = toError(error);
@@ -149,11 +154,13 @@ export const createStreamableRouter = ({
         'error.type': errorType,
       });
 
-      await connectionEvent
-        .fail({
+      try {
+        await connectionEvent.fail({
           error: error instanceof Error ? error : new Error(String(error)),
-        })
-        .catch(() => {});
+        });
+      } catch {
+        // best-effort
+      }
     }
   });
 
--- a/plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
+++ b/plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
@@ -19,12 +19,16 @@ import { performance } from 'node:perf_hooks';
 import { McpService } from '../services/McpService';
 import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js';
 import { LATEST_PROTOCOL_VERSION } from '@modelcontextprotocol/sdk/types.js';
-import { HttpAuthService, LoggerService } from '@backstage/backend-plugin-api';
 import { toError } from '@backstage/errors';
 import {
-  MetricsService,
   TracingService,
 } from '@backstage/backend-plugin-api/alpha';
+import {
+  HttpAuthService,
+  LoggerService,
+  AuditorService,
+} from '@backstage/backend-plugin-api';
+import { MetricsService } from '@backstage/backend-plugin-api/alpha';
 import { bucketBoundaries, McpServerSessionAttributes } from '../metrics';
 import { McpServerConfig } from '../config';
 
@@ -34,6 +38,7 @@ export const createStreamableRouter = ({
   logger,
   metrics,
   tracing,
+  auditor,
   serverConfig,
 }: {
   mcpService: McpService;
@@ -41,6 +46,7 @@ export const createStreamableRouter = ({
   httpAuth: HttpAuthService;
   metrics: MetricsService;
   tracing: TracingService;
+  auditor: AuditorService;
   serverConfig?: McpServerConfig;
 }): Router => {
   const router = PromiseRouter();
@@ -64,10 +70,18 @@ export const createStreamableRouter = ({
       'network.protocol.name': 'http',
     };
 
+    const connectionEvent = await auditor.createEvent({
+      eventId: 'connection',
+      severityLevel: 'medium',
+      request: req,
+      meta: { transport: 'streamable', actionType: 'established' },
+    });
+
     try {
       const server = mcpService.getServer({
         credentials: await httpAuth.credentials(req),
         serverConfig,
+        req,
       });
 
       const transport = new StreamableHTTPServerTransport({
@@ -81,6 +95,7 @@ export const createStreamableRouter = ({
         tracing.context.active(),
         req.headers,
       );
+      await connectionEvent.success();
       await tracing.context.with(ctx, () =>
         transport.handleRequest(req, res, req.body),
       );
@@ -92,6 +107,14 @@ export const createStreamableRouter = ({
         const durationSeconds = (performance.now() - sessionStart) / 1000;
 
         sessionDuration.record(durationSeconds, baseAttributes);
+
+        auditor
+          .createEvent({
+            eventId: 'connection',
+            request: req,
+            meta: { transport: 'streamable', actionType: 'closed' },
+          })
+          .then(e => e.success());
       });
     } catch (error) {
       const err = toError(error);
@@ -116,6 +139,8 @@ export const createStreamableRouter = ({
         ...baseAttributes,
         'error.type': errorType,
       });
+
+      await connectionEvent.fail({ error: error as Error });
     }
   });
 
--- a/plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
+++ b/plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
@@ -90,7 +90,6 @@ export const createStreamableRouter = ({
       });
 
       await server.connect(transport);
-      await connectionEvent.success();
 
       const ctx = tracing.propagation.extract(
         tracing.context.active(),
@@ -99,6 +98,7 @@ export const createStreamableRouter = ({
       await tracing.context.with(ctx, () =>
         transport.handleRequest(req, res, req.body),
       );
+      await connectionEvent.success();
 
       res.on('close', () => {
         transport.close();
@@ -141,7 +141,11 @@ export const createStreamableRouter = ({
         'error.type': errorType,
       });
 
-      await connectionEvent.fail({ error: error as Error });
+      await connectionEvent
+        .fail({
+          error: error instanceof Error ? error : new Error(String(error)),
+        })
+        .catch(() => {});
     }
   });
 
--- a/plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
+++ b/plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
@@ -72,7 +72,6 @@ export const createStreamableRouter = ({
 
     const connectionEvent = await auditor.createEvent({
       eventId: 'connection',
-      severityLevel: 'medium',
       request: req,
       meta: { transport: 'streamable', actionType: 'established' },
     });
@@ -114,7 +113,8 @@ export const createStreamableRouter = ({
             request: req,
             meta: { transport: 'streamable', actionType: 'closed' },
           })
-          .then(e => e.success());
+          .then(e => e.success())
+          .catch(() => {});
       });
     } catch (error) {
       const err = toError(error);
--- a/plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
+++ b/plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
@@ -25,6 +25,7 @@ import {
 } from '@backstage/backend-plugin-api/alpha';
 import {
   AuditorService,
+  AuditorServiceEvent,
   HttpAuthService,
   LoggerService,
 } from '@backstage/backend-plugin-api';
@@ -70,7 +71,7 @@ export const createStreamableRouter = ({
       'network.protocol.name': 'http',
     };
 
-    let connectionEvent;
+    let connectionEvent: AuditorServiceEvent;
     try {
       connectionEvent = await auditor.createEvent({
         eventId: 'connection',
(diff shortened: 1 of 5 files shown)


plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
```diff
@@ -72,6 +85,7 @@ export const createStreamableRouter = ({
       });
 
       await server.connect(transport);
+      await connectionEvent.success();
```
schultzp2020, line 100: `connectionEvent.success()` fires here before `handleRequest()` on line 89. If `handleRequest` throws, the catch block at line 133 also calls `connectionEvent.fail()` — both outcomes recorded on the same event.

The router avoids this by placing `connectionEvent.success()` as the last statement in its try block. Consider moving this after `handleRequest`, or narrowing the try/catch scope to only cover connection establishment.

```suggestion
      await server.connect(transport);
      await transport.handleRequest(req, res, req.body);
      await connectionEvent.success();
```

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -171,9 +197,11 @@ export class McpService {
         });
 
         isError = !!(result as { isError?: boolean })?.isError;
+        await auditorEvent.success();
```
schultzp2020: +1 to Copilot's comment here. The metrics code already distinguishes `isError` (lines 212-215 set `errorAttribute = 'tool_error'`), so the audit event should match.

```suggestion
        isError = !!(result as { isError?: boolean })?.isError;
        if (isError) {
          await auditorEvent.fail({ meta: { toolError: true } });
        } else {
          await auditorEvent.success();
        }
```

schultzp2020 (commented):
Two items inline — the streamable router can double-fire success/fail on the same connection event, and the `isError` case should be reflected in the audit event (as Copilot also noted). Neither is blocking, but both worth addressing before merge.

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -134,6 +153,13 @@ export class McpService {
       let errorType: string | undefined;
       let isError = false;
 
+      const auditorEvent = await this.auditor.createEvent({
```
benjdlambert, line 241: Hmm, I'm a little torn on this.

I'm thinking that it would be smart to push this to the actions registry and let the audit events there handle it due to how it's all setup.

`mcp-actions` -> `actions-registry` -> `upstream service (scaffolder)`

However, I think actually it makes sense to have both. The naming here has made me change my mind, that `tool-execution` is well scoped enough to keep this as is, and we also add audit logging the `actionsRegistry` too. Yes there's gonna be duplicate events with slightly different event names, but I'm thinking that it's similar to observability that more information is better than too little.

johnmcollier (commented):
Might be good to update the `mcp-actions-backend` readme with the expected audit log events?

(Or is that better for the soon to be created [mcp-actions-backend docs](https://github.com/backstage/backstage/pull/32751) @schultzp2020 ?)

plugins/mcp-actions-backend/src/services/McpService.ts
thepetk: @schultzp2020 I've commited Copilot's suggestion!

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -170,10 +196,25 @@ export class McpService {
...
         });
 
-        isError = !!(result as { isError?: boolean })?.isError;
+        const resultIsError = (result as { isError?: boolean })?.isError;
+        isError = !!resultIsError;
+
+        if (isError) {
+          const errorDescription =
+            (result as { errorDescription?: string })?.errorDescription ??
+            `Tool "${params.name}" reported isError=true`;
```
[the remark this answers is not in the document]
thepetk: I'm not sure if I like this approach tbh, I feel what we have now is sufficient to begin with.

thepetk:
Switching back to draft cause obviously something went wrong with the commits merged from copilot + fixes pushed from local.

Will try to address the issue and mark it as ready for review.

thepetk:
Rebased to latest `main` -> switched back to `ready for review`

thepetk:
@schultzp2020 @benjdlambert & @johnmcollier happy to hear your thoughts after the latest updates for this PR!

thepetk:
I'll wait for further input before addressing the PR comments from copilot.

Just IMHO the amount of input coming from the bot on the review it feels a bit overwhelming after a point and I'm not sure how good is balancing between helping this process or making it over-complicated.

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -171,10 +218,31 @@ export class McpService {
...
```
schultzp2020: `handleErrors` returns `{ content, isError }` — it never sets an `errorDescription` field, so the fallback message is always used. The actual error description (e.g. `"NotFoundError: ..."`) lives in `content[0].text`. Is this intentional, or should the audit event pull from there instead?

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -92,6 +104,20 @@ export class McpService {
...
+          eventId: 'tool-discovery',
+          ...(req && { request: req }),
+        });
+      } catch {
+        // Make audit logging best-effort: fall back to a no-op event if auditing is unavailable.
+        auditorEvent = {
+          success: async () => {},
+          fail: async () => {},
+        };
+      }
```
schultzp2020: `McpService` wraps every `auditor.createEvent()` in try/catch with a no-op fallback, but the routers call it directly (letting failures propagate). Since `auditor` is a required core service dependency, is the defensive wrapping intentional? If so, it might be worth being consistent across routers too — or removing it here to match.

schultzp2020 (changes requested):
Please add a changeset before merge. Two questions inline.

plugins/mcp-actions-backend/src/services/McpService.ts
thepetk: That's true, my intention was to have more of a "best-effort" auditing thus the wrappers. I've added the same logic in the routers -> here: https://github.com/backstage/backstage/pull/33339/commits/b221bfcbe31d19da830cbe5f8abe677a35ddb967

plugins/mcp-actions-backend/src/services/McpService.ts
thepetk: Nope, it's not intentional I think is nice to get a more specific error there. Fixed in: https://github.com/backstage/backstage/pull/33339/commits/b221bfcbe31d19da830cbe5f8abe677a35ddb967

thepetk:
> Please add a changeset before merge.

@schultzp2020 done!

thepetk:
@schultzp2020 just checking in case you might have any updated review from your side for this PR! :)

thepetk:
Just a comment not to be closed as stale.

@schultzp2020 let me know if you have any further additional comments

gabemontero:
> Just a comment not to be closed as stale.
> 
> @schultzp2020 let me know if you have any further additional comments

@thepetk - have you ever added this PR to the bi-weekly Framework SIG PR review section, and join that call to facilitate?  If not, that could be another avenue for completing review, etc.

awanlin:
> > Just a comment not to be closed as stale.
> > @schultzp2020 let me know if you have any further additional comments
> 
> @thepetk - have you ever added this PR to the bi-weekly Framework SIG PR review section, and join that call to facilitate? If not, that could be another avenue for completing review, etc.

Hi @gabemontero, appreciate you trying to help out but we'd really like to avoid the various SIGs turning into PR reviews. If there's large architectural concerns or a PR isn't being reviewed at all then noting it in a SIG is reasonable. In this case @schultzp2020 has helped with the approach in the related issue for this and we really just need the review to be completed. 👍

gabemontero:
> > > Just a comment not to be closed as stale.
> > > @schultzp2020 let me know if you have any further additional comments
> > 
> > 
> > @thepetk - have you ever added this PR to the bi-weekly Framework SIG PR review section, and join that call to facilitate? If not, that could be another avenue for completing review, etc.
> 
> Hi @gabemontero, appreciate you trying to help out but we'd really like to avoid the various SIGs turning into PR reviews. If there's large architectural concerns or a PR isn't being reviewed at all then noting it in a SIG is reasonable. In this case @schultzp2020 has helped with the approach in the related issue for this and we really just need the review to be completed. 👍

Understood @awanlin 

And cross referencing your comment with the history, it has received an initial review (vs. not being reviewed at all)

fwiw I've pinged @thepetk separately as well; he'll pursue other avenues vs. employing the framework SIG

thanks

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -92,6 +104,20 @@ export class McpService {
       const startTime = performance.now();
       let errorType: string | undefined;
 
+      let auditorEvent: AuditorServiceEvent;
+      try {
+        auditorEvent = await this.auditor.createEvent({
```
schultzp2020: It may be better to wrap everything in a try catch block rather than having multiple try catch blocks with a no op if it fails. https://github.com/backstage/backstage/blob/34abc1d0eef444525aa1d9a3520027156ce8edd4/plugins/catalog-backend/src/service/createRouter.ts#L128

thepetk: @schultzp2020 tried to simplify the logic of the `try`/`catch` blocks in `McpService.ts` -> see [34635a7](https://github.com/backstage/backstage/pull/33339/commits/34635a78fab1d5fed94f8384e2b1eea7fb48d8f4)

plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
```diff
@@ -59,10 +66,23 @@ export const createStreamableRouter = ({
...
+      connectionEvent = await auditor.createEvent({
+        eventId: 'connection',
+        request: req,
+        meta: { transport: 'streamable', actionType: 'established' },
+      });
+    } catch {
+      // Make audit logging best-effort: fall back to a no-op event if auditing is unavailable.
+      connectionEvent = { success: async () => {}, fail: async () => {} };
+    }
+
```
schultzp2020: I believe you can also simplify this block as well.

plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
```diff
@@ -105,6 +135,12 @@ export const createStreamableRouter = ({
         ...baseAttributes,
         'error.type': errorType,
       });
+
+      await connectionEvent
+        .fail({
+          error: error instanceof Error ? error : new Error(String(error)),
+        })
+        .catch(() => {});
```
schultzp2020: Try to stay away from using both async/await and then/catch. If possible lean on the more modern async/await pattern.

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -92,6 +103,11 @@ export class McpService {
       const startTime = performance.now();
       let errorType: string | undefined;
 
+      const auditorEvent = await this.auditor.createEvent({
```
schultzp2020, line 172: This is much cleaner! Nice job!

schultzp2020 (changes requested):

plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
thepetk: @schultzp2020 I've tried to address your comments both on `createStreamableRouter` but also on `createSSERouter` here: [570c5dc](https://github.com/backstage/backstage/pull/33339/commits/570c5dc166b9d35eb96cb750e58b3ce074535fb3)

plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
thepetk: Simplified in: [570c5dc](https://github.com/backstage/backstage/pull/33339/commits/570c5dc166b9d35eb96cb750e58b3ce074535fb3)

schultzp2020 (approved):
🚣

thepetk:
Hey folks! This one is now approved, but I'm not sure if there's anything required from my side to get it merged.

@schultzp2020 @gabemontero @johnmcollier as I'm sure you might have more context :)

plugins/mcp-actions-backend/src/routers/createSseRouter.ts
```diff
@@ -28,34 +32,77 @@ export const createSseRouter = ({
...
   const transportsToSessionId = new Map<string, SSEServerTransport>();
 
   router.get('/', async (req, res) => {
-    const server = mcpService.getServer({
-      credentials: await httpAuth.credentials(req),
-      serverConfig,
-    });
+    let connectionEvent: AuditorServiceEvent;
+    try {
+      connectionEvent = await auditor.createEvent({
```
benjdlambert: Hmm, this is a bit unfortunate, I would expect that we should move this to the auditor to return a shim if something fails rather than consumers doing this? 

What's the case where this would fail?

plugins/mcp-actions-backend/src/routers/createSseRouter.ts
```diff
@@ -28,34 +32,77 @@ export const createSseRouter = ({
...
+      const normalizedError =
+        error instanceof Error
+          ? error
+          : new Error(
+              typeof error === 'string'
+                ? error
+                : 'Unknown error during SSE connection',
+            );
+      try {
+        await connectionEvent.fail({ error: normalizedError });
```
benjdlambert: Same again here with the `try/catch`.

Also I think you can do something like `toError(error)` instead here? `import { toError } from '@backstage/errors'`

plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
```diff
@@ -116,6 +152,14 @@ export const createStreamableRouter = ({
         ...baseAttributes,
         'error.type': errorType,
       });
+
+      try {
+        await connectionEvent.fail({
+          error: error instanceof Error ? error : new Error(String(error)),
```
benjdlambert: Same here with `toError`

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -155,6 +172,24 @@ export class McpService {
           ? this.filterActions(allActions, serverConfig)
           : allActions;
 
+        const tools = actions.map(action => ({
```
benjdlambert: I don't think we need this at all anymore? 

`await auditorEvent.success({ meta: { toolCount: actions.length } });`

benjdlambert (commented):
Hey :wave: thanks for this! Just wanna clarify a few things here before we move forward :pray:

plugins/mcp-actions-backend/src/routers/createSseRouter.ts
thepetk: Tried to address this in: `234e349`

Pretty much this is the way we started but after copilot review I decided to add those best effort try/catch blocks. I have removed them for now.

All I see is there's not really a case where `createEvent` would fail as it returns a `Promise<AuditorServiceEvent>`.

plugins/mcp-actions-backend/src/routers/createSseRouter.ts
thepetk: > Same again here with the try/catch.

Done! removed the try/catch from both places, and switched to toError(error).

plugins/mcp-actions-backend/src/routers/createStreamableRouter.ts
thepetk: Done!

plugins/mcp-actions-backend/src/services/McpService.ts
thepetk: I've switched now to `await auditorEvent.success({ meta: { toolCount: tools.length } });` . I've rebased it too so I think now it makes more sense

thepetk:
@benjdlambert I tried to address your input, switched to `toError` but also removed the copilot suggestions for base effort try/catch blocks. I do agree to your point it's something better to be handled on the auditor side.

Let me know wdyt :)

thepetk:
I'm also resolving all copilot comments for now. I do feel it has over-complicated the review process here.

awanlin (changes requested):
Thanks for the contribution @thepetk, can you rebase to resolve the conflicts? Once that's done we can merge this in. 👍

thepetk:
> Thanks for the contribution @thepetk, can you rebase to resolve the conflicts? Once that's done we can merge this in. 👍

@awanlin I've rebased my PR :)

benjdlambert (approved):
Thanks for this! Apologies on the delay here getting this in! :tada:

awanlin (approved):
🚢

PR #33344, opened 2026-03-13
`feat(mcp-actions)`: Added the ability to add server name and description to `mcp-actions`

benjdlambert:
Can add name and description for servers, even when they've been split into plugins too to help guide an LLM.

Also a free bug fix here that `:` are not really allowed for namespacing the MCP spec, so changing that to `.` instead.


plugins/mcp-actions-backend/src/plugin.ts
```diff
@@ -87,16 +87,25 @@ export const mcpPlugin = createBackendPlugin({
             router.use(`/v1/${key}`, streamableRouter);
           }
         } else {
+          const defaultServerConfig = {
```
freben: any particular reason to name it "default"?

PR #33346, opened 2026-03-13
create-app - Added the `mcp-actions-backend` and the `plugin-auth` plugins

awanlin:

Added the `mcp-actions-backend` plugin to `create-app`. This also includes the following:

- Documentation for the Experimental Client ID Metadata Documents
- Both templates include config to make the auth, catalog, and scaffolder actions available
- The New Frontend System template includes the `auth-plugin` needed by the Experimental Dynamic Client Registration auth method
- The New Frontend System template includes config for both Experimental Client Id Metadata Documents and Experimental Dynamic Client Registration auth methods. Experimental Dynamic Client Registration is disabled
- The Legacy Frontend system template includes config for just the Experimental Client Id Metadata Documents auth method
- Both templates include `mcpActions` config


Rugvip (commented):
👍  to the addition, I think there may be a few more things in config we need to add to make it function well in practice though. Dunno if we need to list a few initial actions, and then we'd want to ship with CIMD enabled by default prolly. Maybe it's good to let all of that stabilize a bit before we add it? Not only on our side but a lot of clients don't implement CIMD quite yet either.

awanlin:
Let's wait for the next release for this, I'll make the needed changes post BackstageCon/KubeCon 👍

awanlin:
Hi @Rugvip and @johnmcollier, I've updated this with the needed config and some examples with links to the docs as well. Can you give it another look, please? 🙏

packages/create-app/templates/default-app/app-config.yaml.hbs
```diff
@@ -67,6 +73,18 @@ auth:
   providers:
     # See https://backstage.io/docs/auth/guest/provider
     guest: {}
+  # see https://backstage.io/docs/ai/mcp-actions#experimental-client-id-metadata-documents
+  # to learn more about client id metadata documents
+  experimentalClientIdMetadataDocuments:
+    enabled: false
```
Rugvip, line 79: I'm thinking we should ship with this enabled tbh, it's the much better option at this point and doesn't expose nearly as much risk as the DCR approach

packages/create-app/templates/next-app/app-config.yaml.hbs
```diff
@@ -120,3 +145,8 @@ kubernetes:
 permission:
   # setting this to `false` will disable permissions
   enabled: true
+
+# see https://backstage.io/docs/ai/mcp-actions for more details
+mcpActions:
+  name: 'My MCP Server' # defaults to "backstage"
+  description: 'Tools for interacting with My MCP Server' # optional
```
Rugvip: Thinking we should perhaps keep it oriented towards Backstage, thinking "My Backstage Actions" or something like that? Thinking both "MCP" and "Server" are a bit redundant in the name

packages/create-app/templates/default-app/app-config.yaml.hbs
```diff
@@ -67,6 +73,18 @@ auth:
   providers:
     # See https://backstage.io/docs/auth/guest/provider
     guest: {}
+  # see https://backstage.io/docs/ai/mcp-actions#experimental-client-id-metadata-documents
+  # to learn more about client id metadata documents
+  experimentalClientIdMetadataDocuments:
+    enabled: true
```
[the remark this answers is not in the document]
Rugvip: Oh yep this is true ☝️, makes sense to have it disabled by default in the legacy template imo

docs/ai/mcp-actions.md
```diff
@@ -192,6 +196,30 @@ You can configure the auth-backend and install the auth frontend plugin to enabl
...
+### Experimental: Client ID Metadata Documents
+
+:::warning
+This feature is highly experimental; proceed with caution.
+:::
+
+In the [November 2025 MCP specification](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization), it outlined a new authorization method will to replace Dynamic Client Registration called [Client ID Metadata Documents (CIMD)](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#client-id-metadata-documents).
+
+This can be enabled in the `auth-backend` plugin by using the `auth.experimentalClientIdMetadataDocuments.enabled` flag in config:
+
```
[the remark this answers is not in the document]
Rugvip: Yep, this ☝️, although with the context that it's installed by default (and add that 😁)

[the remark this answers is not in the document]
awanlin: Revised the related documentation to make this more clear.

docs/ai/mcp-actions.md
```diff
@@ -180,17 +184,49 @@ You can configure the auth-backend and install the auth frontend plugin to enabl
    });
    ```
 
-3. Enable the feature:
+#### Dynamic Client Registration
```
Rugvip: Let's strongly recommend CIMD, so switching the order of the two and adding a note to DCR that it should only be used if clients do not support CIMD, and that it'll likely be removed in the future

awanlin: For sure, changed the order 👍

Rugvip (approved):
👌, let's :shipit:

PR #33353, opened 2026-03-14
fix(ui): open external Table row hrefs in a new tab

cdedreuille:

When a `Table` row has an external `href` (e.g. `https://...`), clicking it navigated in the same tab. The fix automatically applies `target="_blank"` and `rel="noopener noreferrer"` on the `Row` component when the `href` is detected as an external link.

Fixes [BACUI-272](https://linear.app/spotify/issue/BACUI-272).

### DIFF
--- a/packages/ui/src/components/Table/components/Row.tsx
+++ b/packages/ui/src/components/Table/components/Row.tsx
@@ -91,7 +91,7 @@ export function Row<T extends object>(props: RowProps<T>) {
         {...restProps}
         target={effectiveTarget}
         rel={effectiveRel}
-        className={classes.root}
+        className={clsx(classes.root, restProps.className)}
         data-react-aria-pressable={hasInternalHref ? 'true' : undefined}
         onAction={handlePress}
       >


packages/ui/src/components/Table/components/Row.tsx
```diff
@@ -36,9 +36,26 @@ export function Row<T extends object>(props: RowProps<T>) {
...
+  const effectiveRel =
+    effectiveTarget === '_blank'
+      ? [
+          ...new Set([
+            'noopener',
+            'noreferrer',
+            ...(restProps.rel?.split(/\s+/).filter(Boolean) ?? []),
+          ]),
+        ].join(' ')
+      : restProps.rel;
```
Naycon, line 57: Isn't this a little overkill? I'd say we should just leave this as an option that you can steer from the outside. If the link is external, we set `rel` to the safe option explicitly down below (`rel="noopener noreferrer"`), and then allow the `...restProps` spread to override it at will. I don't think we need this merge stuff.

packages/ui/src/components/Table/components/Row.tsx
```diff
@@ -71,9 +88,11 @@ export function Row<T extends object>(props: RowProps<T>) {
     <InternalLinkProvider href={href}>
       <ReactAriaRow
         href={href}
-        className={classes.root}
-        data-react-aria-pressable={hasInternalHref ? 'true' : undefined}
         {...restProps}
+        target={effectiveTarget}
+        rel={effectiveRel}
```
Naycon, line 93: If you decide to simplify the above, these should go before the `...restProps` spread.

packages/ui/src/components/Table/components/Row.tsx
Rugvip: hmm, yeah at the same time the override becomes a little bit pointless just because you'd have to litter it everywhere if you wanted to remove the originals. I think this is alright tbh, avoids the mistake of unintentionally overriding

cdedreuille: Yes I would agree with @Rugvip on this one.

PR #33355, opened 2026-03-14
cli: add CLI module plugin API and auto-discovery

Rugvip:

This introduces a modular CLI architecture where CLI commands are provided by independent `@backstage/cli-module-*` packages instead of being bundled inside `@backstage/cli`. The CLI now auto-discovers modules from the project root's dependencies, making it fully composable.

**What's in here:**

- A `cli-module` package role and the `createCliModule` / `CliCommand` / `CliCommandContext` API in `@backstage/cli-node`
- Internal opaque type wiring via `@internal/cli` (mirroring the `@internal/frontend` pattern)
- All existing CLI command groups split into 11 separate `@backstage/cli-module-*` packages
- A `runCliModule` helper in `@backstage/cli-node` so each module can be executed standalone (`npx @backstage/cli-module-info info`)
- Auto-discovery of CLI modules from the project root's `dependencies`/`devDependencies` — any package with the `cli-module` role is loaded automatically
- Fallback to the built-in set with a deprecation warning when no modules are found
- Updated `create-app` templates to include all standard CLI modules


.changeset/cli-discover-modules.md
```diff
@@ -0,0 +1,31 @@
...
+---
+
+The CLI now automatically discovers CLI modules from the project root's `dependencies` and `devDependencies`. Any installed package with the `cli-module` Backstage role will be loaded automatically without needing to be hardcoded in the CLI itself.
+
+If no CLI modules are found in the project dependencies, the CLI falls back to the built-in set of modules and prints a deprecation warning. This fallback will be removed in a future release. To prepare for this, add the following CLI modules as `devDependencies` in your root `package.json`:
+
+```json
+{
+  "devDependencies": {
+    "@backstage/cli-module-auth": "backstage:^",
```
freben, line 28: This makes it seem like we need a `cli-defaults`.

Rugvip: Was thinking about that, is it worth it? Could be something like a `cli-preset-*` or smth too. I guess it's especially good for our ability to ship new modules with minor impact. We could also exclude certain modules from the defaults as well, for example if we end up with `*-test-jest` and `*-test-vitest`

freben: Unless we end up with particularly heavy and specialised modules - something that pulls in the entire AWS library for example - a big `-defaults` or preset seems decent as a platform for us to ship things over time. I'm not sure that adding quite this number of individual modules is convenient enough especially when we want to expand on the offering in the future right?

Could even have config support for

```yaml
cli:
  modules:
    exclude:
      - jest
```

(analogous to `app.packages.exclude` etc) if you as an adopter want to not expose your users to a builtin module at all.

Rugvip: Added, without the config for now, let's see where we end up with a need for that. I made the specific modules take precedence over collections, so if you install `cli-defaults` + e.g. `cli-module-test-vitest`, it would just override all commands in the existing `cli-module-test-jest` ones. The conflict resolution is done at the command level, overriding all existing commands from any conflicting modules.

packages/cli-module-test-jest/src/commands/repo/test.ts
```diff
@@ -182,7 +181,7 @@ export default async ({ args, info }: CommandContext) => {
   // Only include our config if caller isn't passing their own config
   if (!hasFlags('-c', '--config')) {
     /* eslint-disable-next-line no-restricted-syntax */
-    args.push('--config', findOwnPaths(__dirname).resolve('config/jest.js'));
+    args.push('--config', require.resolve('@backstage/cli/config/jest'));
```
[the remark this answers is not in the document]
Rugvip: Fixed — the jest config files have been moved into `@backstage/cli-module-test-jest/config/` and are now resolved via `findOwnPaths` instead of `require.resolve`.

packages/cli-module-test-jest/src/commands/package/test.ts
```diff
@@ -26,11 +26,11 @@ function includesAnyOf(hayStack: string[], ...needles: string[]) {
   return false;
 }
 
-export default async ({ args }: CommandContext) => {
+export default async ({ args }: CliCommandContext) => {
   // Only include our config if caller isn't passing their own config
   if (!includesAnyOf(args, '-c', '--config')) {
     /* eslint-disable-next-line no-restricted-syntax */
-    args.push('--config', findOwnPaths(__dirname).resolve('config/jest.js'));
+    args.push('--config', require.resolve('@backstage/cli/config/jest'));
```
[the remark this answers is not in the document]
Rugvip: Fixed — same as above, resolved via `findOwnPaths` now.

packages/cli-module-build/src/lib/bundler/paths.ts
```diff
@@ -50,9 +50,7 @@ export function resolveBundlingPaths(options: BundlingPathsOptions) {
     targetHtml = resolvePath(targetDir, `${entry}.html`);
     if (!fs.pathExistsSync(targetHtml)) {
       /* eslint-disable-next-line no-restricted-syntax */
-      targetHtml = findOwnPaths(__dirname).resolve(
-        'templates/serve_index.html',
-      );
+      targetHtml = require.resolve('@backstage/cli/templates/serve_index.html');
```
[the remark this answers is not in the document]
Rugvip: Fixed — the template has been moved to `@backstage/cli-module-build/templates/` and is resolved via `findOwnPaths`.

packages/cli-module-build/src/index.ts
```diff
@@ -14,10 +14,11 @@
  * limitations under the License.
  */
 
-import { createCliPlugin } from '../../wiring/factory';
+import { createCliModule } from '@backstage/cli-node';
+import packageJson from '../package.json';
```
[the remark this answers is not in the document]
Rugvip: Fixed — the named `buildPlugin` export has been inlined into `export default createCliModule(...)`.

packages/cli-module-test-jest/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "dependencies": {
+    "@backstage/cli-common": "workspace:^",
+    "@backstage/cli-node": "workspace:^",
+    "cleye": "^2.3.0",
+    "yargs": "^16.2.0"
+  },
+  "devDependencies": {
+    "@backstage/cli": "workspace:^"
+  },
+  "peerDependencies": {
```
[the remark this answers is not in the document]
Rugvip: Fixed — the jest config now lives in `cli-module-test-jest/config/` and is resolved locally.

packages/cli-module-build/src/lib/bundler/paths.ts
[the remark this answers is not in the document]
Rugvip: Fixed — the template is now shipped with `cli-module-build` and resolved via `findOwnPaths`.

packages/cli-module-config/report.api.md
```diff
@@ -0,0 +1,23 @@
...
+
+// Warning: (ae-missing-release-tag) "configOption" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
+//
+// @public (undocumented)
+export const configOption: readonly [
+  '--config <path>',
+  'Config files to load instead of app-config.yaml',
+  (opt: string, opts: string[]) => string[],
+  string[],
+];
```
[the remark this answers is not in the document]
Rugvip: Fixed — the `configOption` export has been removed entirely.

packages/cli-defaults/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@backstage/cli-defaults",
+  "version": "0.0.0",
+  "description": "Default set of CLI modules for the Backstage CLI",
+  "backstage": {
+    "role": "cli-module"
```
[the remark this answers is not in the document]
Rugvip: Fixed — the CLI loader now supports array exports from cli-module packages, so this is handled correctly.

packages/cli-defaults/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  config,
+  createGithubApp,
+  info,
+  lint,
+  maintenance,
+  migrate,
+  newModule,
+  testJest,
+  translations,
+];
```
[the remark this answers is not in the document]
Rugvip: Fixed — the CLI loader supports both single module and array-of-modules default exports from cli-module packages.

packages/cli-module-build/src/index.ts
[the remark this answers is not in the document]
Rugvip: Fixed — the export has been inlined into `export default createCliModule(...)`.

packages/cli-module-info/src/commands/info.ts
```diff
@@ -99,12 +99,27 @@ export default async ({ args, info }: CommandContext) => {
...
+        require(require.resolve('@backstage/cli/package.json', {
+          paths: [process.cwd()],
+        })) as { version: string }
+      ).version;
+    } catch {
+      /* not available */
+    }
+
     const systemInfo = {
       os: `${os.type} ${os.release} - ${os.platform}/${os.arch}`,
```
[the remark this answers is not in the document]
Rugvip: Pre-existing issue on master, not introduced by this PR.

packages/cli-internal/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "types": "src/index.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "lint": "backstage-cli package lint",
+    "test": "backstage-cli package test"
+  },
+  "dependencies": {
+    "@backstage/cli-node": "workspace:^"
```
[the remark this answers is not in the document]
Rugvip: This is a private inlined workspace package — same pattern used by `@internal/opaque` in `frontend-internal`. Dependencies are resolved via the workspace, not published.

packages/cli-module-migrate/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Commands
+
+| Command                        | Description                                                       |
+| :----------------------------- | :---------------------------------------------------------------- |
+| `versions:bump`                | Bump Backstage packages to the latest versions                    |
+| `versions:migrate`             | Migrate plugins moved to the @backstage-community namespace       |
+| `migrate package-roles`        | Add package role field to packages that don't have it             |
+| `migrate package-scripts`      | Set package scripts according to each package role                |
+| `migrate package-exports`      | Synchronize package subpath export definitions                    |
```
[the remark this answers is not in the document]
Rugvip: Pre-existing on master — the command already throws a removal error.

packages/cli-node/src/cli-module/createCliModule.ts
```diff
@@ -0,0 +1,67 @@
...
+   * Called immediately when the plugin is created.
+   */
+  init: (registry: {
+    /** Registers a new command with the CLI. */
+    addCommand: (command: CliCommand) => void;
+  }) => Promise<void>;
+}): CliModule {
+  const commands: CliCommand[] = [];
+  const commandsPromise = options
+    .init({ addCommand: command => commands.push(command) })
```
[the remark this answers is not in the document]
Rugvip: Good suggestion — validation for the command path has been added in `createCliModule`, which already throws on missing `packageJson.name`.

packages/cli-node/src/cli-module/runCliModule.ts
```diff
@@ -0,0 +1,232 @@
...
+ * @public
+ */
+export async function runCliModule(options: {
+  /** The CLI module to run. */
+  module: CliModule;
+  /** The program name shown in help output and usage strings. */
+  name: string;
+  /** The version string shown when `--version` is passed. */
+  version?: string;
+}): Promise<void> {
```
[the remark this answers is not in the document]
Rugvip: Noted — test coverage for `runCliModule` can be added as a follow-up.

packages/cli-node/src/cli-module/runCliModule.ts
```diff
@@ -0,0 +1,232 @@
...
+            };
+
+            if (typeof internal.command.execute === 'function') {
+              await internal.command.execute(context);
+            } else {
+              const mod = await internal.command.execute.loader();
+              const fn =
+                typeof mod.default === 'function'
+                  ? mod.default
+                  : (mod.default as any).default;
```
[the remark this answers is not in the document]
Rugvip: The nested `default.default` handling is needed for CJS/ESM interop with dynamic imports. This is a common pattern in the codebase.

packages/cli-module-test-jest/src/commands/package/test.ts
```diff
@@ -96,5 +96,6 @@ export default async ({ args }: CommandContext) => {
     process.exit(1);
   }
 
+  // eslint-disable-next-line @backstage/no-undeclared-imports
   await require('jest').run(args);
```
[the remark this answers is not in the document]
Rugvip: Fixed — the peer dependency has been changed from `jest-cli` to `jest` to match the runtime `require('jest')` usage.

packages/cli-module-test-jest/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "jest-cli": "^29.0.0 || ^30.0.0"
```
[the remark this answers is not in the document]
Rugvip: Fixed — peer dependency changed from `jest-cli` to `jest`.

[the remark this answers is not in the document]
Rugvip: Fixed — peer dependency changed from `jest-cli` to `jest`.

packages/create-app/templates/next-app/package.json.hbs
```diff
@@ -50,6 +50,17 @@
   ],
   "devDependencies": {
     "@backstage/cli": "^{{version '@backstage/cli'}}",
+    "@backstage/cli-module-auth": "^{{version '@backstage/cli-module-auth'}}",
```
freben: Not using `cli-defaults` here?

packages/repo-tools/src/commands/api-reports/cli-reports/runCliExtraction.ts
```diff
@@ -132,7 +132,7 @@ export async function runCliExtraction({
     const pkgJson = await fs.readJson(resolvePath(fullDir, 'package.json'));
 
     if (!pkgJson.bin) {
-      throw new Error(`CLI Package in ${packageDir} has no bin field`);
+      continue;
```
freben, line 140: Isn't this still desired for the `cli` role?

package.json
```diff
@@ -125,6 +125,17 @@
   },
   "devDependencies": {
     "@backstage/cli": "workspace:*",
+    "@backstage/cli-module-auth": "workspace:*",
```
freben: Not using `cli-defaults` here?

packages/cli/src/wiring/discoverCliModules.ts
```diff
@@ -0,0 +1,69 @@
...
+    dependencies?: Record<string, string>;
+    devDependencies?: Record<string, string>;
+  };
+  try {
+    projectPkg = JSON.parse(fs.readFileSync(pkgJsonPath, 'utf8'));
+  } catch {
+    return [];
+  }
+
+  const allDeps = {
```
freben, line 46: Maybe it's done elsewhere, but immediate reaction when seeing this - this requires you to add a dep even to a local, private, nonpublished package. Should we be looking through workspaces too?

packages/cli/config/nodeTransform.cjs
```diff
@@ -14,74 +14,14 @@
...
-      paths: [joinPath(linkedRoot.dir, 'node_modules')],
-    });
-  };
+try {
+  require('@backstage/cli-module-build/config/nodeTransform.cjs');
+} catch (e) {
+  if (e.code === 'MODULE_NOT_FOUND') {
+    throw new Error(
+      '@backstage/cli-module-build is required to use the node transform. ' +
+        'Please install it as a dependency.',
```
freben, line 23: Right, so at least getting the defaults package in there will basically be a requirement for everyone, makes sense

packages/cli-defaults/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+import testJest from '@backstage/cli-module-test-jest';
+import translations from '@backstage/cli-module-translations';
+
+/**
+ * The default set of CLI modules for the Backstage CLI.
+ *
+ * @public
+ */
+export default [
+  auth,
```
freben, line 34: Makes it tempting to support promises too, allowing `import(...)` here

packages/cli-internal/src/InternalCommandNode.ts
```diff
@@ -0,0 +1,55 @@
...
+export interface CommandTreeNode {
+  readonly $$type: '@backstage/CommandTreeNode';
+}
+
+/** @internal */
+export interface CommandLeafNode {
+  readonly $$type: '@backstage/CommandLeafNode';
+}
+
+export type CommandNode = CommandTreeNode | CommandLeafNode;
```
freben, line 30: Nice to finally not use deep commander intrinsics. 👍

packages/cli-module-auth/bin/backstage-cli-module-auth
```diff
@@ -0,0 +1,33 @@
...
+  path.resolve(__dirname, '../src'),
+);
+
+if (!isLocal) {
+  const { runCliModule } = require('@backstage/cli-node');
+  const cliModule = require('..').default;
+  const pkg = require('../package.json');
+  runCliModule({ module: cliModule, name: pkg.name, version: pkg.version });
+} else {
+  require('@backstage/cli/config/nodeTransform.cjs');
```
freben: Will this continue to be a main, stable entrypoint of the cli itself? Or should these in general refer to the new `@backstage/cli-module-build` instead?

Side question: We don't do replacement at a package level, but at a command level. So you don't really replace the build module, you just replace build commands - but in some cases these direct imports are used instead as the contract. Is that ever a problem? I mean for the node transform you could argue that it's part of our module's own internals. But let's say the lint transform. But then again ... maybe that only ever occurs in .eslintrc where you just point it to your own directly anyway.

packages/cli-module-auth/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "devDependencies": {
+    "@backstage/backend-test-utils": "workspace:^",
+    "@backstage/cli": "workspace:^",
+    "@types/fs-extra": "^11.0.0",
+    "@types/proper-lockfile": "^4"
+  },
+  "optionalDependencies": {
+    "keytar": "^7.9.0"
+  },
+  "bin": "bin/cli-module-auth"
```
freben: Let's prefix these with `backstage-` as you mentioned

packages/cli-module-build/bin/backstage-cli-module-build
```diff
@@ -0,0 +1,33 @@
...
+);
+
+if (!isLocal) {
+  const { runCliModule } = require('@backstage/cli-node');
+  const cliModule = require('..').default;
+  const pkg = require('../package.json');
+  runCliModule({ module: cliModule, name: pkg.name, version: pkg.version });
+} else {
+  require('@backstage/cli/config/nodeTransform.cjs');
+  require('../src/cli');
```
freben: wouldn't mind calling all these `run.ts` or `invoke.ts` or whatnot, but don't really have all that strong of an opinion

packages/cli-module-config/cli-report.md
```diff
@@ -0,0 +1,109 @@
...
+Options:
+  -V, --version
+  -h, --help
+
+Commands:
+  config [command]
+  config:check
+  config:docs
+  config:print
+  config:schema
```
freben, line 19: will we ever be deprecating these colon-separated ones and making them multi-word instead?

packages/cli-module-create-github-app/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@backstage/cli-module-create-github-app",
```
freben: Would this one make sense as `@backstage/cli-module-github` and the command `backstage-cli github app create`?

packages/cli-node/src/cli-module/createCliModule.ts
```diff
@@ -0,0 +1,73 @@
...
+  }) => Promise<void>;
+}): CliModule {
+  if (!options.packageJson.name) {
+    throw new Error(
+      'The packageJson provided to createCliModule must have a name',
+    );
+  }
+
+  const commands: CliCommand[] = [];
+  const commandsPromise = options
```
freben: could start this chain with a `Promise.resolve()` to ensure any problems actually happen async.

packages/cli-node/src/cli-module/runCliModule.ts
```diff
@@ -0,0 +1,232 @@
...
+      }),
+    );
+  }
+
+  return graph;
+}
+
+function exitWithError(error: unknown): never {
+  if (!isError(error)) {
+    process.stderr.write(`\n${chalk.red(stringifyError(error))}\n\n`);
```
freben: same code inside the if and below - move up outside of the if?

packages/cli/src/wiring/CliInitializer.ts
```diff
@@ -15,55 +15,98 @@
...
 import { isPromise } from 'node:util/types';
 
-function isNodeHidden(
-  node:
-    | { $$type: '@tree/leaf'; command: BackstageCommand }
-    | { $$type: '@tree/root'; children: unknown[] },
-): boolean {
-  if (node.$$type === '@tree/leaf') {
-    return !!node.command.deprecated || !!node.command.experimental;
+function isNodeHidden(node: CommandNode): boolean {
```
freben: this implementation occurs twice, but no biggie really

packages/cli/src/wiring/CliInitializer.ts
```diff
@@ -15,55 +15,98 @@
...
 
-type UninitializedFeature = CliFeature | Promise<{ default: CliFeature }>;
+type UninitializedFeature =
+  | CliModule
+  | CliModule[]
+  | Promise<{ default: CliModule | CliModule[] }>;
+
+interface TaggedFeature {
+  feature: CliModule;
+  fromArray: boolean;
```
freben, line 45: I don't think I'm entirely clear on what this does below 🤔 Like why isn't it uniform. Maybe a comment on the field that explains the logic behind it

packages/cli/src/wiring/discoverCliModules.ts
Rugvip: It's kinda how it works anyway, you need to add a local dep even for local workspace packages to show up as executable

packages/cli-defaults/src/index.ts
Rugvip: then you need the `.default.default` unwrapping and whatnot, figured this is fine for now. Something we can evolve it becomes a more commonly used pattern

packages/cli/src/wiring/discoverCliModules.ts
freben: Sure but the method of executing local ones could be completely different to dep ones

packages/cli-module-auth/bin/backstage-cli-module-auth
Rugvip: Come to think of it, this transform should be in `cli-node` instead, that's a much better home for it.

I'd like to avoid any cross-module direct imports, which I think solves the side question?

packages/cli-module-config/cli-report.md
Rugvip: yeah let's do that, not now though x)

packages/cli-module-create-github-app/package.json
Rugvip: It's easier to keep the command as is for now. About the module name, yeah let's go with just `github`, feels a bit lower risk

packages/cli-module-auth/bin/backstage-cli-module-auth
freben: I think my question is, "is `require('@backstage/cli/config/...` itself as a pattern a problem, and will people ever want or need to dynamically replace those things".

Like, replacement of the jest transform. But at that point - as long as the reference to it is "visible" and under your own control - as is the case inside an .eslintrc - then the method of "dynamic" replacement is static instead - you just change the .eslintrc to point to your file.

But there are some such contracts that get case specific. To replace ts parameters you change tsconfig for example. So there's a level of indirection happening and it varies depending on which config thing we're talking about. I just recall some support questions asking "how do I tweak X" and it turns out it's tricky because you have to entirely break glass and leave our tooling to do it.

What if we never invented the overrides parameter to the jest transform, and instead had a "module plugin pattern" of sorts that _injects `/config/` layer behaviors_? As a separate initiative to the commands

packages/cli-module-config/cli-report.md
freben: I'm asking in part because it sometimes affects how I think about the right module structure and granularity. But marking as resolved!

packages/create-app/templates/next-app/package.json.hbs
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/4d081452b1 — now uses cli-defaults instead of listing individual modules

package.json
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/4d081452b1 — now uses cli-defaults

packages/cli-module-auth/bin/backstage-cli-module-auth
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/4d081452b1 — moved nodeTransform to cli-node. Also made @swc/core an optional peer dep in https://github.com/backstage/backstage/commit/92b0f79d9c

packages/cli-module-create-github-app/package.json
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/4d081452b1 — renamed to @backstage/cli-module-github

packages/cli-node/src/cli-module/createCliModule.ts
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/4d081452b1

packages/cli-node/src/cli-module/runCliModule.ts
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/4d081452b1

packages/cli/src/wiring/CliInitializer.ts
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/4d081452b1 — extracted isCommandNodeHidden to @internal/cli

packages/cli/src/wiring/CliInitializer.ts
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/4d081452b1 — added doc comment explaining the deduplication logic

packages/repo-tools/src/commands/api-reports/cli-reports/runCliExtraction.ts
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/4d081452b1 — now only skips cli-module packages, throws for cli role

packages/cli-module-auth/package.json
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/26eab3bf83

packages/cli-module-auth/bin/backstage-cli-module-auth
Rugvip: Hmm, generally `config` is very tied to the module implementation, so I'm leaning towards keeping that pattern but shipping through modules. Of course aiming to deprecate the `@backstage/cli/config/*` entry points over time as well. Honestly not 100% sure where to go with this long term but feeling the state in the PR now is a good point to iterate from.

packages/cli-module-info/src/commands/info.ts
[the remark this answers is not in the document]
freben: wow this is actually correct

packages/cli-module-build/bin/backstage-cli-module-build
Rugvip: You mean the file itself? Think it needs to be named as is

PR #33356, opened 2026-03-14
frontend-app-api: add phased app preparation

Rugvip:

Built on top of #33154 and #33204, bringing those together and updating especially the implementation in the `frontend-app-api`.

This adds phased frontend app preparation through `prepareSpecializedApp`, so apps can render a bootstrap (sign-in page) tree first and hand off to a finalized app once sign-in finishes. We realized this change was needed in order to be able to support features in the frontend system that depend on users' identity.

It also changes conditional frontend wiring a fair bit by:
- adding feature-level `if` support for plugins and modules
- letting plugin and extension overrides replace or remove existing `if` predicates
- deferring supported conditional APIs and `app/root.elements` until finalization, while warning for unsupported bootstrap-visible predicates and missing bootstrap API accesses

There are a couple of related follow-ups in here too:
- `plugin-app` is adjusted for the phased bootstrap/finalize flow
- permission predicate evaluation now makes separate `authorize()` calls and relies on internal batching instead of a public array overload
- `docs/frontend-system` and the relevant changesets are updated to match the new behavior


packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
Rugvip, line 1: there's definitely an opportunity to split this up a bit, but figured I'd leave that for a followup refactor

plugins/permission-react/src/apis/IdentityPermissionApi.ts
```diff
@@ -50,13 +58,12 @@ export class IdentityPermissionApi implements PermissionApi {
...
+  ): Promise<AuthorizePermissionResponse>;
   async authorize(
     request: AuthorizePermissionRequest,
   ): Promise<AuthorizePermissionResponse> {
-    const response = await this.permissionClient.authorize(
-      [request],
-      await this.identityApi.getCredentials(),
-    );
-    return response[0];
+    return this.loader.load(request);
```
freben: ❤️  but also, just for stack traces,

```suggestion
    return await this.loader.load(request);
```

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/61765e183fc1e9594748c7d2a22f76f6e86d5e0e

packages/frontend-app-api/report.api.md
```diff
@@ -204,4 +236,37 @@ export type FrontendPluginInfoResolver = (ctx: {
...
   info: FrontendPluginInfo;
 }>;
+
+// @public
+export type PreparedSpecializedApp = {
+  getBootstrapApp(): BootstrapSpecializedApp;
+  onFinalized(callback: (app: FinalizedSpecializedApp) => void): () => void;
+  finalize(options?: {
+    sessionState?: SpecializedAppSessionState;
+  }): FinalizedSpecializedApp;
```
Rugvip: Only one of these two can be called. Either you let the user handle the bootstrap with `.onFinalized`, or you immediately initialize the app with `.finalize`

packages/frontend-defaults/src/createApp.tsx
Rugvip, line 1: Good overview of how this affects the public API of `frontend-app-api`

docs/frontend-system/architecture/10-app.md
```diff
@@ -72,6 +72,38 @@ app:
...
+const bootstrapApp = preparedApp.getBootstrapApp();
+
+const unsubscribe = preparedApp.onFinalized(
+  (finalizedApp: FinalizedSpecializedApp) => {
+    console.log(finalizedApp.sessionState);
+  },
+);
+```
+
+The `getBootstrapApp()` method exposes the partial app tree that is available during bootstrap. The `onFinalized()` method notifies you once the full app tree has been finalized, and `finalize(sessionState?)` can be used when you already have a reusable session state or when you want to bypass the asynchronous bootstrap flow in tests.
```
aramissennyeydd: From https://github.com/backstage/backstage/pull/33356#discussion_r2939807537, should we make it more clear that the user _needs_ to handle finalizing themselves if they call `onFinalized`?

Rugvip: I guess we can say that it's delegated to the sign-in page and finalization happens after the sign-in page has provided a result? 🤔

packages/frontend-app-api/report.api.md
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/2ccc1e3d234264af8cecb2ea075589bc9775910b

packages/frontend-app-api/src/wiring/FrontendApiRegistry.ts
```diff
@@ -0,0 +1,139 @@
...
+    for (const apiRefId of apiRefIds) {
+      this.apis.delete(apiRefId);
+    }
+  }
+
+  private load<T>(ref: ApiRef<T>, loading: AnyApiRef[] = []): T | undefined {
+    const existing = this.apis.get(ref.id);
+    if (existing) {
+      return existing as T;
+    }
```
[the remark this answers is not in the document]
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/2ccc1e3d234264af8cecb2ea075589bc9775910b

packages/frontend-app-api/src/wiring/FrontendApiRegistry.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    const deps = {} as { [name: string]: unknown };
+    for (const [key, depRef] of Object.entries(factory.deps)) {
+      const dep = this.load(depRef, [...loading, factory.api]);
+      if (!dep) {
+        throw new Error(
+          `No API factory available for dependency ${depRef} of dependent ${factory.api}`,
+        );
+      }
```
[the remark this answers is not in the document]
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/2ccc1e3d234264af8cecb2ea075589bc9775910b

docs/frontend-system/architecture/10-app.md
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/2ccc1e3d234264af8cecb2ea075589bc9775910b

plugins/permission-react/src/apis/IdentityPermissionApi.ts
```diff
@@ -24,20 +25,27 @@ import {
...
     identityApi: IdentityApi,
   ) {
-    this.permissionClient = permissionClient;
-    this.identityApi = identityApi;
+    this.loader = new DataLoader(
+      async (requests: readonly AuthorizePermissionRequest[]) => {
+        const credentials = await identityApi.getCredentials();
+        return permissionClient.authorize([...requests], credentials);
+      },
+    );
```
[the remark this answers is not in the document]
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/77b6834d3b5f49788bf01d67ab027ce338828073

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
```diff
@@ -241,20 +59,15 @@ export type CreateSpecializedAppOptions = {
    */
   bindRoutes?(context: { bind: CreateAppRouteBinder }): void;
 
+  /**
+   * APIs to expose to the app during startup.
+   */
+  apis?: ApiHolder;
+
```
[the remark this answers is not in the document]
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/77b6834d3b5f49788bf01d67ab027ce338828073

packages/frontend-app-api/src/wiring/predicates.ts
```diff
@@ -0,0 +1,185 @@
...
+
+// Minimal local permission API interface to avoid a dependency on @backstage/plugin-permission-react
+type MinimalPermissionApi = {
+  authorize(
+    request: EvaluatePermissionRequest,
+  ): Promise<EvaluatePermissionResponse>;
+};
+
+export const localPermissionApiRef = createApiRef<MinimalPermissionApi>({
+  id: 'plugin.permission.api',
```
freben, line 46: wouldn't mind a local cross-module import to just get the id out of the real one

packages/frontend-app-api/src/wiring/predicates.ts
```diff
@@ -0,0 +1,185 @@
...
+}
+
+function extractPredicateKeyNames(
+  predicate: FilterPredicate,
+  key: string,
+): string[] {
+  if (typeof predicate !== 'object' || predicate === null) {
+    return [];
+  }
+  const obj = predicate as Record<string, unknown>;
```
freben, line 163: wouldn't it be better to cast it to FilterPredicate here, to get the type helps in the code below? `if ('$all' in predicate)` will do the right thing inside that statement. Sure, maybe this is even more defensive in a way but 🤷

packages/frontend-app-api/src/wiring/predicates.ts
```diff
@@ -0,0 +1,185 @@
...
+} from '@backstage/frontend-plugin-api';
+import { FilterPredicate } from '@backstage/filter-predicates';
+import type {
+  EvaluatePermissionRequest,
+  EvaluatePermissionResponse,
+} from '@backstage/plugin-permission-common';
+
+export type ExtensionPredicateContext = {
+  featureFlags: string[];
+  permissions: string[];
```
freben, line 30: I was slightly worried about the string array types here, both because it doesn't open up for more fields in the future (not sure what they would be though honestly), and because it requires the use of `$contains` which is clunky.

Was hoping for something more like `if: { featureFlags.myFlag.enabled: true }` but i know that the drawback then is that there are dashes and dots in the names.

Then again we don't have to abide by any set of rules _at all_ really in terms of the context contents - we could have `if: { featureFlag.enabled: 'my-ugly-flag-name' }` if we wanted. We can evaluate these in any way we see fit; don't have to be a static data object being operated on.

packages/frontend-app-api/src/wiring/prepareSpecializedApp.tsx
```diff
@@ -0,0 +1,926 @@
...
+
+      let subscribed = true;
+
+      if (finalized) {
+        const finalizedApp = finalized;
+        Promise.resolve().then(() => {
+          if (subscribed) {
+            callback(finalizedApp);
+          }
+        });
```
freben, line 568:
```suggestion
        }).catch(() => {});
```

perhaps, to avoid unhandled rejections?

packages/frontend-app-api/src/wiring/treeLifecycle.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+  function ThrowBootstrapError(): never {
+    throw error;
+  }
+
+  // This synthetic child gives the finalized tree a stable place to rethrow
+  // bootstrap failures through the normal extension boundary stack.
+  (bootstrapChildNode as AppNode & { instance?: AppNodeInstance }).instance = {
+    getDataRefs() {
+      return [coreExtensionData.reactElement].values();
```
freben:
```suggestion
      return [coreExtensionData.reactElement];
```

right...?

packages/frontend-app-api/src/wiring/treeLifecycle.tsx
freben, line 1: It's honestly pretty confusing - but let's go

packages/frontend-app-api/src/wiring/predicates.ts
Rugvip: Actually was thinking we'll move this to `frontend-plugin-api`, just wanted to do it in a followup. Kinda needs https://github.com/backstage/backstage/pull/33365 as well

packages/frontend-app-api/src/wiring/predicates.ts
Rugvip: Hmm, I think we should stick to the static data as much as possible tbh, gets a bit too arcane otherwise? 🤔

Honestly I don't find it to be that bad, although I agree something like `featureFlags.myFlag.enabled: true` could be nice, but like you say that has a bunch of problems.

plugins/permission-react/src/apis/IdentityPermissionApi.ts
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/e10f0d85b41ad6e0317b2d9d59ff2531f0b98f6a

packages/frontend-app-api/src/wiring/predicates.ts
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/e10f0d85b41ad6e0317b2d9d59ff2531f0b98f6a

packages/frontend-app-api/src/wiring/prepareSpecializedApp.tsx
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/e10f0d85b41ad6e0317b2d9d59ff2531f0b98f6a

packages/frontend-app-api/src/wiring/treeLifecycle.tsx
Rugvip: [automated] Fixed in https://github.com/backstage/backstage/commit/e10f0d85b41ad6e0317b2d9d59ff2531f0b98f6a

packages/frontend-app-api/src/wiring/apiFactories.ts
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// TODO(Rugvip): It would be good if this was more explicit, but I think that
+//               might need to wait for some future update for API factories.
+function getApiOwnerId(apiRefId: string): string {
+  const [prefix, ...rest] = apiRefId.split('.');
+  if (!prefix) {
+    return apiRefId;
+  }
+  if (prefix === 'core') {
```
freben, line 297: hey i've seen this elsewhere :D

packages/frontend-app-api/src/wiring/predicates.ts
freben: I kinda like how the last one reads though :) And it sort of matches how the search table works in the catalog (flattening arrays into `key.value`)

But OK, not pushing for it. Just feels like an easier way to write the conditions for the user too. `if: { featureFlag: 'my-flag' }` is pretty compelling UX, possibly with the addition of `.disabled`/`.enabled` if we need to check for both (we could support all three, with just `featureFlag` essentially meaning `featureFlag.enabled`)

PR #33358, opened 2026-03-14
feat(ui): add List and ListRow components

cdedreuille:

Adds `List` and `ListRow` components to `@backstage/ui`, built on top of React Aria's `GridList`/`GridListItem` primitives with Backstage UI styles.

React Aria's `ListBox` primitive doesn't support interactive children (buttons, menus), which is a hard requirement for our use case. `GridList` is the correct primitive for rows that contain actions, menus, and tags — it has full keyboard and screen reader support for interactive children.

Features:

- Standalone list for displaying interactive rows directly on the page (not in a dropdown)
- `ListRow` supports icons, descriptions, menus (menuItems), and custom actions (customActions)
- Single and multiple selection modes via selectionMode prop
- Disabled rows via disabledKeys
- Follows the same component pattern as `TagGroup`, `Menu`, and `Select`

Closes [BACUI-42](https://linear.app/spotify/issue/BACUI-42/code-listboxlistrow-component)

### DIFF
--- a/.changeset/fix-focus-visible-data-attr.md
+++ b/.changeset/fix-focus-visible-data-attr.md
@@ -1,0 +1,7 @@
+---
+'@backstage/ui': patch
+---
+
+Fixed focus ring styles to use React Aria's `[data-focus-visible]` data attribute instead of the native CSS `:focus-visible` pseudo-class. This ensures keyboard focus rings render reliably when focus is managed programmatically by React Aria (e.g. inside a GridList, Menu, or Select).
+
+**Affected components:** Accordion, Button, ButtonIcon, ButtonLink, Card, List, Menu, Select, ToggleButtonGroup
--- a/packages/ui/src/components/Accordion/Accordion.module.css
+++ b/packages/ui/src/components/Accordion/Accordion.module.css
@@ -53,7 +53,7 @@
     cursor: pointer;
     text-align: left;
 
-    &:focus-visible {
+    &[data-focus-visible] {
       outline: none;
       transition: none;
       box-shadow: inset 0 0 0 2px var(--bui-ring);
--- a/packages/ui/src/components/Button/Button.module.css
+++ b/packages/ui/src/components/Button/Button.module.css
@@ -73,7 +73,7 @@
       --fg: var(--bui-fg-solid-disabled);
     }
 
-    &:focus-visible {
+    &[data-focus-visible] {
       outline: 2px solid var(--bui-ring);
       outline-offset: 2px;
     }
@@ -106,7 +106,7 @@
       --bg-active: var(--bg-solid-danger-disabled);
     }
 
-    &:focus-visible {
+    &[data-focus-visible] {
       outline: 2px solid var(--bui-border-danger);
       outline-offset: 2px;
     }
@@ -143,7 +143,7 @@
       --fg: var(--bui-fg-disabled);
     }
 
-    &:focus-visible {
+    &[data-focus-visible] {
       outline: none;
       transition: none;
       box-shadow: inset 0 0 0 2px var(--bui-ring);
@@ -172,7 +172,7 @@
       --fg: var(--bui-fg-disabled);
     }
 
-    &:focus-visible {
+    &[data-focus-visible] {
       box-shadow: inset 0 0 0 2px var(--bui-border-danger);
     }
   }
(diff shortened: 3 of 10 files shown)


.changeset/add-listbox-component.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/ui': patch
+---
+
+Added `ListBox` and `ListBoxItem` components. These provide a standalone, accessible list of selectable options built on top of React Aria's `ListBox` and `ListBoxItem` primitives. Items support icons, descriptions, and single or multiple selection modes.
+
+**Affected components:** ListBox, ListBoxItem
```
Naycon: I think you need to add these to the docs-ui types for them to be picked up by the changelog script. Here's what I'm doing for the SearchAutocomplete component:
https://github.com/backstage/backstage/pull/33340/changes#diff-ae453e85ba59a4f967743be203fbd45f2ac1ba4429118c3207593ca1dbf47e57

packages/ui/src/components/ListBox/types.ts
```diff
@@ -0,0 +1,82 @@
...
+   */
+  children?: React.ReactNode;
+  /**
+   * Optional secondary description text.
+   */
+  description?: string;
+  /**
+   * Optional icon displayed before the label, rendered in a 32×32px box.
+   */
+  icon?: React.ReactNode;
```
Rugvip: Bit of a mix for what we use here, I think `ReactElement` is best in order to avoid `string`, and then add `boolean`/`null` if needed

cdedreuille: Works for me. I updated it to use `ReactElement`

packages/ui/src/components/List/List.tsx
```diff
@@ -0,0 +1,127 @@
...
+          {customActions && (
+            <div
+              className={classes.actions}
+              onClick={e => e.stopPropagation()}
+              onKeyDown={e => e.stopPropagation()}
+            >
+              {customActions}
+            </div>
+          )}
+          {menuItems && (
```
Naycon, line 101: React Aria doesn't allow for interactive elements inside ListItems unfortunately:
https://react-aria.adobe.com/ListBox#text-slots
The keyboard navigation is broken on the stories where there are e.g. menu items.

.changeset/add-listbox-component.md
cdedreuille: Ah you're right. I just added it.

packages/ui/src/components/List/List.tsx
cdedreuille: Yes you're right. Good call. I just refactored it to use GridList instead and renamed the components `List` and `ListRow`

packages/ui/src/components/List/List.tsx
```diff
@@ -0,0 +1,118 @@
...
+    <RAGridListItem
+      textValue={textValue}
+      className={classes.root}
+      {...dataAttributes}
+      {...restProps}
+    >
+      {({ isSelected }) => (
+        <>
+          {isSelected && (
+            <div className={classes.check}>
```
Naycon, line 81: Should this be `slot="description"`? Here's what RA has in their example for their checkbox:
```
{selectionMode === 'multiple' && selectionBehavior === 'toggle' && (
  <Checkbox slot="selection" />
)}
```
They only show the checkbox when you can toggle multiple items.

packages/ui/src/components/List/List.tsx
```diff
@@ -0,0 +1,118 @@
...
+              <RiCheckLine />
+            </div>
+          )}
+          {icon && (
+            <Box bg="neutral" className={classes.icon}>
+              {icon}
+            </Box>
+          )}
+          <div className={classes.label}>
+            <span>{children}</span>
```
Naycon, line 91: Just calling out, wrapping in a span means we only permit phrasing content here. Might be exactly what we intend, but using children that contains e.g. a div would mean broken html.

docs-ui/src/app/components/list/page.mdx
```diff
@@ -0,0 +1,103 @@
...
+  open
+  preview={<SelectionModeMultiple />}
+  code={selectionModeMultiple}
+/>
+
+### Disabled items
+
+<Snippet align="center" py={4} open preview={<Disabled />} code={disabled} />
+
+<Theming definition={ListDefinition} />
```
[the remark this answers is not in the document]
Naycon: This might be worth considering. 

(This is a response to Copilots comment about adding the ListItemDefinition here to show its theming option)

PR #33364, opened 2026-03-16
frontend-plugin-api: accept IconElement in AuthProviderInfo

Rugvip:

This adds `IconElement` support to the `AuthProviderInfo.icon` field, which previously only accepted `IconComponent`. With `IconComponent` already deprecated in favor of `IconElement` in the new frontend system, this was a gap where auth provider metadata still required the old type.

The `icon` field now accepts `IconComponent | IconElement`, so existing code continues to work while new code can pass `<MyIcon />` directly. The consumers in `LoginRequestListItem` (core-components) and `ProviderSettingsItem` (user-settings) are updated to handle both types at runtime using `isValidElement`.

### DIFF
--- a/plugins/user-settings/src/components/AuthProviders/ProviderSettingsItem.tsx
+++ b/plugins/user-settings/src/components/AuthProviders/ProviderSettingsItem.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useEffect, useState } from 'react';
+import { createElement, isValidElement, useEffect, useState } from 'react';
 import Button from '@material-ui/core/Button';
 import Grid from '@material-ui/core/Grid';
 import ListItem from '@material-ui/core/ListItem';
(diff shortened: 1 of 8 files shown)


plugins/user-settings/src/components/AuthProviders/ProviderSettingsItem.tsx
```diff
@@ -86,11 +87,14 @@ export const ProviderSettingsItem = (props: {
...
+    icon === null || isValidElement(icon)
+      ? icon
+      : createElement(icon as IconComponent);
+
   return (
     <ListItem>
-      <ListItemIcon>
-        <Icon />
-      </ListItemIcon>
+      <ListItemIcon>{iconElement}</ListItemIcon>
```
freben, line 97: doesn't this one need `?? <></>`?

Rugvip: Hmm, good point, actually having `| null` on `IconElement` might be wrong to begin with? 🤔

Gonna ship and have a think because this seems like a potentially broader issue. Reminds me of `children?: ReactNode` on all `ComponentType`s

PR #33365, opened 2026-03-16
frontend-plugin-api: convert ApiRef to an opaque type

Rugvip:

This converts `ApiRef` in the new frontend system to an opaque type with a `$$type: '@backstage/ApiRef'` discriminator, matching the pattern already used by route refs and extension data refs. The change makes it impossible to create `ApiRef` instances as plain object literals — you must use `createApiRef`.

Also adds a builder-pattern creation overload similar to extension data refs:

```ts
const myApiRef = createApiRef<MyApi>().with({ id: 'plugin.my.api' });
```

The existing `createApiRef<MyApi>({ id: 'plugin.my.api' })` pattern continues to work.

The resulting API ref object shape remains compatible with old-system expectations (still has `.id` for lookups), and old-system API refs created via `createApiRef` from `@backstage/core-plugin-api` automatically get the new opaque type since that package re-exports from `@backstage/frontend-plugin-api`.

An `OpaqueApiRef` helper is added to `@backstage/frontend-internal` for internal type validation, following the same pattern as `OpaqueRouteRef`.

The idea is to roll this out as a non-breaking change in the old system as well well, so that we can make `$$type` required in NFS on release.

### DIFF
--- a/plugins/app-react/report.api.md
+++ b/plugins/app-react/report.api.md
@@ -86,7 +86,7 @@ export const IconBundleBlueprint: ExtensionBlueprint<{
   };
   output: ExtensionDataRef<
     {
-      [x: string]: IconComponent | IconElement;
+      [x: string]: IconElement | IconComponent;
     },
     'core.icons',
     {}
@@ -97,7 +97,7 @@ export const IconBundleBlueprint: ExtensionBlueprint<{
   dataRefs: {
     icons: ConfigurableExtensionDataRef<
       {
-        [x: string]: IconComponent | IconElement;
+        [x: string]: IconElement | IconComponent;
       },
       'core.icons',
       {}
--- a/plugins/app/report.api.md
+++ b/plugins/app/report.api.md
@@ -478,7 +478,7 @@ const appPlugin: OverridableFrontendPlugin<
         icons: ExtensionInput<
           ConfigurableExtensionDataRef<
             {
-              [x: string]: IconComponent | IconElement;
+              [x: string]: IconElement | IconComponent;
             },
             'core.icons',
             {}


packages/core-plugin-api/src/apis/system/ApiRef.test.ts
```diff
@@ -19,9 +19,15 @@ import { createApiRef } from './ApiRef';
 describe('ApiRef', () => {
   it('should be created', () => {
     const ref = createApiRef({ id: 'abc' });
+    expect(ref.$$type).toBe('@backstage/ApiRef');
     expect(ref.id).toBe('abc');
     expect(String(ref)).toBe('apiRef{abc}');
-    expect(() => ref.T).toThrow('tried to read ApiRef.T of apiRef{abc}');
+    expect(ref.T).toBeNull();
```
freben, line 25: Why did this change?

packages/frontend-plugin-api/src/apis/system/types.ts
```diff
@@ -19,9 +19,10 @@
  *
  * @public
  */
-export type ApiRef<T> = {
-  id: string;
-  T: T;
+export type ApiRef<T, TId extends string = string> = {
+  readonly $$type?: '@backstage/ApiRef';
```
freben, line 23: It is a bit unfortunate that now all plugin API reports below have this & { $$type } bloat declaring that they have the field present. Do we need this question mark here until we stop supporting the old frontend system and kill the core libraries basically? Or is this needed forever due to the design? Is there any way that we can use the opposite `Omit` in some cases or something like that?

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
```diff
@@ -455,14 +456,19 @@ function createApiFactories(options: {
...
+  }
+
+  const apiRefId = apiRef.id;
   const [prefix, ...rest] = apiRefId.split('.');
   if (!prefix) {
     return apiRefId;
   }
-  if (prefix === 'core') {
-    return 'app';
-  }
```
freben, line 465: can it be the case that we keep seeing these for a while for old plugin-api deps?

packages/core-plugin-api/src/apis/system/ApiRef.test.ts
Rugvip: Figured we should do it while at it. We stopped having `T` throw in newer implementations and just have it be `null` instead

packages/frontend-plugin-api/src/apis/system/types.ts
Rugvip: Mentioned in the PR body :p, idea is to have this be here for 1-2 release just so that things are more uniformly bumped in the old system, then we remove the optionality of `$$type`. We could do it straight away too but that'll cause some type issues for transitive `*-plugin-api` dependencies that aren't bumped alongside the main Backstage versions. Honestly it should be fairly low impact so we could consider shipping with `$$type` required, as long as we support the old format at runtime ofc.

packages/frontend-app-api/src/wiring/createSpecializedApp.tsx
Rugvip: Don't think so, because these refs are determined by the app, so the fact that this logic is being brought in means you'd have the version of things at the app level that also has the explicit plugin IDs

PR #33377, opened 2026-03-16
feat(frontend-plugin-api): add `titleRouteRef` to PageBlueprint

Naycon:

Add `titleLink` to `PageLayoutProps` so the plugin header title links back to the plugin root. `PageBlueprint` resolves it from `plugin.routes.root` (with fallback to `params.routeRef`) via `RouteResolutionApi`.

- PageLayout swap passes the resolved `titleLink` string directly to `PluginHeader`
- Header actions get stable React keys via `cloneElement`

### DIFF
--- a/.patches/pr-33377.txt
+++ b/.patches/pr-33377.txt
@@ -1,0 +1,1 @@
+Added titleLink prop to PageLayoutProps so the plugin header title can link back to the plugin root


packages/frontend-plugin-api/src/blueprints/PageBlueprint.tsx
```diff
@@ -68,13 +68,21 @@ export const PageBlueprint = createExtensionBlueprint({
        * Hide the default plugin page header, making the page fill up all available space.
        */
       noHeader?: boolean;
+      /**
+       * Show the plugin page header even when the page has no tabs.
+       */
+      showHeader?: boolean;
```
Rugvip: I want to make this the default actually, so would rather skip adding this for now if we can

Naycon: Got it! Removed that part of the PR.

packages/frontend-plugin-api/src/components/PageLayout.tsx
```diff
@@ -43,6 +44,7 @@ export interface PageLayoutProps {
   title?: string;
   icon?: IconElement;
   noHeader?: boolean;
+  titleRouteRef?: RouteRef;
```
Rugvip: Should we make this `titleHref` or `titleLink` instead? 🤔 Should be possible to grab pretty easily directly via `RouteResolutionApi`

Rugvip (commented):
🎉 , could you add a `.patches` entry as well? I assume we want that too

packages/frontend-plugin-api/src/components/PageLayout.tsx
Naycon: Done!

PR #33380, opened 2026-03-16
fix(backend-defaults): preserve executable bit when extracting tar archives

omerazmon-idme:
## Summary

- `TarArchiveResponse.dir()` called `tar.extract()` without `chmod: true`, causing the OS umask to silently strip the execute permission from files that should be executable (e.g. shell scripts with mode `0755`).
- This was especially noticeable with `fetch:template` on GitHub-hosted Backstage templates: GitHub tarballs correctly record mode bits (`100755`) in their tar entries, but those bits were discarded during extraction.
- The fix adds `chmod: true` and `processUmask: 0o22` to `tar.extract()`. `chmod: true` forces an explicit `chmod()` call on each extracted file so the mode matches the archive entry exactly, regardless of process umask. `processUmask: 0o22` avoids the deprecated/thread-unsafe `process.umask()` call that `chmod: true` would otherwise require.

## Test plan

- [ ] Added regression test in `TarArchiveResponse.test.ts`: creates a tar archive containing a `0755` shell script, extracts it via `TarArchiveResponse.dir()`, and asserts the executable bit is preserved
- [ ] All existing `TarArchiveResponse` tests continue to pass
- [ ] Changeset added for `@backstage/backend-defaults` (patch)

> **Note:** This PR is largely AI-assisted. The change has been personally reviewed, understood, and the test verified locally.


omerazmon-idme:
Your AI needs an UPGRADE or new brains. First it asks for a change, which I accepted, then it asks for a reverse change, then it summarily closes the PR. This is an actual BUG.

omerazmon-idme:
Please review

PR #33381, opened 2026-03-16
feat(cli): add actions CLI module for distributed actions registry

benjdlambert:
OK, so this is a bit.

This adds `@backstage/cli-module-actions` with commands for listing and executing actions from the distributed actions registry. Exports auth helpers from `cli-module-auth` for cross-module reuse. and relaxes the actions registry auth check to allow direct user invocations from the CLI, but wondering if this is the right choice?


packages/cli-module-actions/src/commands/sourcesList.ts
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { cli } from 'cleye';
+import type { CliCommandContext } from '@backstage/cli-node';
+import {
+  getSelectedInstance,
+  getInstanceConfig,
+} from '@backstage/cli-module-auth';
```
Rugvip, line 22: Ah, this is not the way I envision this to happen 😁, will refactor a bit after

PR #33389, opened 2026-03-17
plugins: migrate NFS pages to HeaderPage

Rugvip:

This changes the new frontend system plugin pages over to the Backstage UI header components across the supported plugins, so the plugin header and page-level header patterns line up better across NFS pages.

It switched the extensibility model for settings and devtools to use subpages instead.

The header is introduced everywhere but not all pages are updates to use the new layout. There are some like the entity page that need a bit more work, but I don't think the header is that disruptive to the UX and provides a good global hook for additional functionality. Making it opt-in at this point I think would be more disruptive.

### DIFF
--- a/packages/frontend-plugin-api/src/blueprints/PageBlueprint.test.tsx
+++ b/packages/frontend-plugin-api/src/blueprints/PageBlueprint.test.tsx
@@ -24,7 +24,7 @@ import {
   createExtensionBlueprint,
   createExtensionInput,
 } from '../wiring';
-import { waitFor, screen } from '@testing-library/react';
+import { waitFor } from '@testing-library/react';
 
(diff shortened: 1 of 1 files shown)


plugins/catalog-unprocessed-entities/src/components/UnprocessedEntities.tsx
```diff
@@ -66,3 +67,14 @@ export const UnprocessedEntities = () => {
     </Page>
   );
 };
+
+export const NfsUnprocessedEntities = () => {
+  return (
+    <>
+      <HeaderPage title="Unprocessed Entities" />
```
freben, line 74: Yeah it does also look a bit wonky to render a standalone thing named `HeaderPage` without children.

PR #33396, opened 2026-03-17
cli-node: add CliAuth class for shared CLI authentication

Rugvip:

This refactors the authentication exports and shared behavior out of `@backstage/cli-module-auth` and into `@backstage/cli-node`, replacing the current ad-hoc usage with a class-based `CliAuth` API that follows the local coding conventions (including a `static create` method).

The split uses filesystem state as the communication path between the two packages:
- **`cli-module-auth`** retains ownership of _writing_ auth instances (login, logout, select, config updates)
- **`cli-node`** owns _reading_ the current instance and providing it to consumers via `CliAuth`

The `CliAuth` class resolves the currently selected instance, transparently refreshes expired access tokens, and exposes `getAccessToken()` and `getConfig()` for other CLI modules. This makes it easy for any CLI module to pick up and use the current auth with just `CliAuth.create()`.

Also migrates `cli-module-actions` to the new API and deprecates the function-level exports from `cli-module-auth`.

### DIFF
--- a/plugins/app-react/report.api.md
+++ b/plugins/app-react/report.api.md
@@ -86,7 +86,7 @@ export const IconBundleBlueprint: ExtensionBlueprint<{
   };
   output: ExtensionDataRef<
     {
-      [x: string]: IconElement | IconComponent;
+      [x: string]: IconComponent | IconElement;
     },
     'core.icons',
     {}
@@ -97,7 +97,7 @@ export const IconBundleBlueprint: ExtensionBlueprint<{
   dataRefs: {
     icons: ConfigurableExtensionDataRef<
       {
-        [x: string]: IconElement | IconComponent;
+        [x: string]: IconComponent | IconElement;
       },
       'core.icons',
       {}
--- a/plugins/app/report.api.md
+++ b/plugins/app/report.api.md
@@ -478,7 +478,7 @@ const appPlugin: OverridableFrontendPlugin<
         icons: ExtensionInput<
           ConfigurableExtensionDataRef<
             {
-              [x: string]: IconElement | IconComponent;
+              [x: string]: IconComponent | IconElement;
             },
             'core.icons',
             {}


.changeset/cli-module-auth-deprecate-exports.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli-module-auth': patch
```
Rugvip: Skip this changeset and remove any existing one related to these exports, it's not released yet

.changeset/cli-module-actions-use-cli-auth.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/cli-module-actions': patch
+---
+
+Migrated to use `CliAuth` from `@backstage/cli-node` for authentication instead of importing individual functions from `@backstage/cli-module-auth`.
```
Rugvip: Skip this changeset and change any existing one, this is not released yet

packages/cli-node/report.api.md
```diff
@@ -142,6 +160,17 @@ export class GitUtils {
...
+// @public (undocumented)
+export type HttpInit = {
+  headers?: Record<string, string>;
+  method?: string;
+  body?: any;
+  signal?: AbortSignal;
+};
+
+// @public (undocumented)
+export function httpJson<T>(url: string, init?: HttpInit): Promise<T>;
```
Rugvip: Don't export this, just duplicate it, it's just a small wrapper

packages/cli-module-auth/report.api.md
```diff
@@ -4,15 +4,19 @@
 
 ```ts
 import { CliModule } from '@backstage/cli-node';
+import { HttpInit } from '@backstage/cli-node';
+import { httpJson } from '@backstage/cli-node';
+import { SecretStore } from '@backstage/cli-node';
+import { StoredInstance } from '@backstage/cli-node';
 
-// @public (undocumented)
```
Rugvip: No need to deprecate anything here, just get rid of it all, it's not released

packages/cli-node/report.api.md
```diff
@@ -133,6 +148,9 @@ export function createCliModule(options: {
   }) => Promise<void>;
 }): CliModule;
 
+// @public (undocumented)
+export function getSecretStore(): Promise<SecretStore>;
```
Rugvip: I don't want to be exporting this, but that probably goes away when cleaning up the deprecations

packages/cli-node/src/auth/CliAuth.ts
```diff
@@ -0,0 +1,156 @@
...
+    return new CliAuth(instance, secretStore);
+  }
+
+  private constructor(instance: StoredInstance, secretStore: SecretStore) {
+    this.#instance = instance;
+    this.#secretStore = secretStore;
+  }
+
+  /** The resolved instance metadata. */
+  get instance(): StoredInstance {
```
Rugvip: Avoid getters, make these methods instead so options can be added in the future

packages/cli-node/src/auth/CliAuth.ts
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * Reads a per-instance configuration value previously stored by the
+   * auth module (e.g. `pluginSources`).
+   */
+  async getConfig<T = unknown>(key: string): Promise<T | undefined> {
+    return getInstanceConfig<T>(this.#instance.name, key);
+  }
+
+  async #refreshAccessToken(): Promise<void> {
+    const service = `backstage-cli:auth-instance:${this.#instance.name}`;
```
Rugvip: Move a shared helper for creating this and any other low-level identifiers to `cli-internal`. Don't move anything that would require extra dependencies

packages/cli-node/src/auth/secretStore.ts
```diff
@@ -0,0 +1,112 @@
...
+  async delete(service: string, account: string): Promise<void> {
+    const file = this.filePath(service, account);
+    await fs.remove(file);
+  }
+}
+
+let singleton: SecretStore | undefined;
+
+/** @public */
+export async function getSecretStore(): Promise<SecretStore> {
```
Rugvip, line 111: Avoid using `fs-extra` as a dependency, refactoring to `node:fs` instead. Then move the secret store to `cli-internal` so that it can be re-used without being exported here.

packages/cli-node/report.api.md
```diff
@@ -272,6 +287,17 @@ export function runWorkerQueueThreads<TItem, TResult, TContext>(
   results: TResult[];
 }>;
 
+// @public (undocumented)
+export type StoredInstance = {
```
Rugvip: Let's avoid exporting the full instance on `CliAuth` and hide this type. We can just rely on the existing instance name and base URL

packages/cli-module-auth/report.api.md
```diff
@@ -5,61 +5,10 @@
...
-  baseUrl: string;
-  clientId: string;
-  issuedAt: number;
-  accessTokenExpiresAt: number;
-  selected?: boolean;
-  config?: Record<string, unknown>;
-};
-
 // @public (undocumented)
 export function updateInstanceConfig(
```
Rugvip: This export should also be gone after this change, move it to `CliAuth` as well. Let's make it `getMetadata` and `setMetadata`

packages/cli-node/src/auth/CliAuth.ts
```diff
@@ -0,0 +1,152 @@
...
+      );
+    }
+    return token;
+  }
+
+  /**
+   * Reads a per-instance configuration value previously stored by the
+   * auth module (e.g. `pluginSources`).
+   */
+  async getConfig<T = unknown>(key: string): Promise<T | undefined> {
```
Rugvip: Don't allow this type parameter, it's not safe. Instead always return `unknown` and refactor the other code to use `zod` to ensure the schema.

packages/cli-node/src/auth/storage.ts
```diff
@@ -0,0 +1,144 @@
...
+  name: z
+    .string()
+    .min(1)
+    .regex(INSTANCE_NAME_PATTERN, 'Instance name contains invalid characters'),
+  baseUrl: z.string().url(),
+  clientId: z.string().min(1),
+  issuedAt: z.number().int().nonnegative(),
+  accessTokenExpiresAt: z.number().int().nonnegative(),
+  selected: z.boolean().optional(),
+  config: z.record(z.string(), z.unknown()).optional(),
```
Rugvip: Let's make this `metadata` instead

.changeset/auth-module-exports.md
```diff
@@ -2,4 +2,4 @@
 '@backstage/cli-module-auth': patch
 ---
 
-Export auth helper utilities for use by other CLI modules. Added per-instance config storage with `getInstanceConfig` and `updateInstanceConfig`.
+Export auth helper utilities for use by other CLI modules.
```
Rugvip: Drop this changeset

packages/cli-module-actions/src/lib/resolveAuth.ts
```diff
@@ -14,35 +14,27 @@
...
-    throw new Error('No access token found. Run "auth login" to authenticate.');
-  }
-
-  const pluginSources =
-    (await getInstanceConfig<string[]>(instance.name, 'pluginSources')) ?? [];
+  const auth = await CliAuth.create({ instanceName: instanceFlag });
+  const accessToken = await auth.getAccessToken();
+  const pluginSources = pluginSourcesSchema.parse(
+    await auth.getMetadata('pluginSources'),
+  );
```
freben, line 32: This guy and the schema are repeated a ton. Not that it matters all that much.

packages/cli-node/src/auth/authIdentifiers.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** @internal */
+export function getAuthInstanceService(instanceName: string): string {
```
freben, line 18: should it import the one in cli-internal instead?

packages/cli-module-actions/src/lib/httpJson.ts
```diff
@@ -0,0 +1,39 @@
...
+import { ResponseError } from '@backstage/errors';
+
+type HttpInit = {
+  headers?: Record<string, string>;
+  method?: string;
+  body?: any;
+  signal?: AbortSignal;
+};
+
+export async function httpJson<T>(url: string, init?: HttpInit): Promise<T> {
```
freben, line 26: should use the one from cli-common right?

packages/cli-module-auth/src/lib/http.ts
```diff
@@ -16,15 +16,13 @@
...
-/** @public */
 export type HttpInit = {
   headers?: Record<string, string>;
   method?: string;
   body?: any;
   signal?: AbortSignal;
 };
 
-/** @public */
 export async function httpJson<T>(url: string, init?: HttpInit): Promise<T> {
```
freben, line 26: there's maybe a bunch of these around, right?

packages/cli-module-actions/src/lib/httpJson.ts
Rugvip: got rid of the export, such a thin silly wrapper

packages/cli-module-auth/src/lib/http.ts
Rugvip: yeah, feeling it's fine to duplicate for now

PR #33400, opened 2026-03-17
fix(core-components): resolve active state routing bug in SidebarItem

### LINKED ISSUE #33302: Sidebar highlights multiple items when one route is a sub-path of another (e.g. /test and /test/again)
### 📜 Issue Labels

### 🔎 Search Terms

```plain
sidebar active item
sidebar route highlight
sidebar multiple items highlighted
sidebar subpath active state
backstage sidebar route match
```

### 🗃️ Project Area

Unknown

### 🔗 External Integration

N/A

### 📝 Description & Context

When two sidebar items have paths where one is a prefix of the other (for example `/test` and `/test/again`), both items appear highlighted when navigating to the nested route.

This happens because the active route matching logic appears to match based on **prefix/path inclusion**, causing both the parent path and the nested path to be considered active.

You can observe a similar issue in the [Backstage demo instance](https://demo.backstage.io/catalog/default/Group/team-a) here for the below routes:

 | Current URL | sidebar Item |
|-------------|------------------|
| `/catalog` | `catalog` |
| `/catalog/default/Group/team-a` | `My Group` |

Clicking on My Group higlights/active both catalog and myGroup 

### Example sidebar routes:
- /test
- /test/again

When navigating to:
- /test/again

both sidebar items become highlighted.
This creates confusion in the navigation because only the exact matching route should be considered active.

### 👍 Expected Behavior

Only the sidebar item corresponding to the **current route** should be highlighted.
And for 

| Current URL | Highlighted Item |
|-------------|------------------|
| `/test` | `/test` |
| `/test/again` | `/test/again` |

- The `/test` item should **not remain highlighted** when the user navigates to `/test/again`.
- Additionally, if **no route exactly matches the current location**, the sidebar should highlight the **best matching path (the closest parent route)** instead of highlighting multiple items.

Image

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

1. Add two sidebar routes in the Backstage sidebar configuration:
    - /test
    - /test/again
2. Start the Backstage application.
3. Navigate to: **/test/again**
4. Observe the sidebar.
    - **Observed Result:**  
    - Both `/test` and `/test/again` sidebar items are highlighted.

or Simply go [https://demo.backstage.io](https://demo.backstage.io) go to this and click on the myGroup
 

Undecided

Deepesh123455:
Pull Request Description
 Problem
Currently, the Backstage sidebar incorrectly highlights multiple items when a user navigates to a nested route where one path is a prefix of another (e.g., /catalog and /catalog/default/...). This happens because of a legacy matching logic that uses .startsWith(), causing both parent and child routes to appear "active" simultaneously.

Solution
Since Backstage has migrated to React Router v6, the previous WorkaroundNavLink is no longer necessary. This PR:

Removes the custom WorkaroundNavLink and its manual path-matching logic.

Refactors SidebarItemBase to use the native NavLink from react-router-dom.

Leverages React Router v6's built-in isActive state for class application.

Enables the end prop support, allowing the sidebar to correctly distinguish between exact and partial matches.

How to verify
Add two sidebar items in Root.tsx: one for /test and one for /test/again.

Navigate to /test/again.

Observed Result: Only the "Test Again" item is highlighted (Previously, both would be highlighted).

Checklist
[x] I have followed the [Contributing Guidelines](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md).

[x] I have ran yarn lint and yarn tsc and they pass.

[x] I have added a changeset using yarn changeset.

Fixes: #33302

### DIFF
--- a/.changeset/metal-teams-push.md
+++ b/.changeset/metal-teams-push.md
@@ -1,0 +1,5 @@
+---
+'@backstage/core-components': patch
+---
+
+This PR resolves an issue where multiple sidebar items were incorrectly highlighted when navigating to nested routes (e.g., /test and /test/again both appearing active).
--- a/packages/core-components/src/layout/Sidebar/Items.test.tsx
+++ b/packages/core-components/src/layout/Sidebar/Items.test.tsx
@@ -19,13 +19,17 @@ import {
   TestApiProvider,
   renderInTestApp,
 } from '@backstage/test-utils';
-import { createEvent, fireEvent, screen } from '@testing-library/react';
+import {
+  createEvent,
+  fireEvent,
+  screen,
+  renderHook,
+} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import HomeIcon from '@material-ui/icons/Home';
 import CreateComponentIcon from '@material-ui/icons/AddCircleOutline';
 import { Sidebar } from './Bar';
 import { SidebarItem, SidebarSearchField, SidebarExpandButton } from './Items';
-import { renderHook } from '@testing-library/react';
 import { makeStyles } from '@material-ui/core/styles';
 import { analyticsApiRef } from '@backstage/core-plugin-api';
 
--- a/packages/core-components/src/layout/Sidebar/Items.test.tsx
+++ b/packages/core-components/src/layout/Sidebar/Items.test.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 The Backstage Authors
+ * Copyright 2026 The Backstage Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


its-mitesh-kumar:
@rohitratannagar PTAL on this PR

rohitratannagar:
@Deepesh123455 Thanks for the PR.

I feel Exact path matching is not the correct solution here, as it breaks the expected behavior of sidebar navigation.

The intended behavior is hierarchical:

1. When a sidebar item represents a parent route, it should remain active for all of its subroutes.
   For example, if you have a sidebar item for `/catalog`, then it should be active for:

   * `/catalog`
   * `/catalog/default`
   * `/catalog/default/component`
   * `/catalog/default/component/my-service`

2. However, if a subroute also has its own dedicated sidebar item, then that item should take precedence and be marked as active instead of the parent.

   For example:

   * Sidebar items:

     * `/catalog`
     * `/catalog/import`

   * Behavior:

     * Visiting `/catalog/import` → only **Import** should be active
     * **Catalog** should NOT remain active in this case

So the rule is:

* Parent stays active for subroutes 
* But if a more specific sidebar item exists for that route, it overrides the parent 

This ensures correct highlighting and avoids multiple items appearing active at the same time.

### Attaching a  video of the current behavior from this PR for better context.


https://github.com/user-attachments/assets/aac51a03-7784-460b-b009-482cb417240a

Deepesh123455:
Hi @rohitratannagar, thanks for the detailed explanation and the video! That makes total sense. My initial approach with exact matching (end={true}) was definitely too strict and broke the hierarchical active state for general sub-routes.

To implement the 'Best Match Wins' / Longest Prefix Match logic (where /catalog stays active for /catalog/default but yields to /catalog/import if that specific item exists), we run into a limitation with React Router v6's standalone <NavLink>. Since each <NavLink> evaluates its active state independently, /catalog doesn't inherently know that a 'better' match like /catalog/import exists in the sidebar.

To solve this correctly, we would need a way to evaluate all rendered SidebarItem paths against the current location.pathname to find the longest matching prefix, and only apply the active class to that specific item.

Do we currently have a SidebarContext (or similar) where child SidebarItem components register their paths? If so, we could calculate the 'active' path centrally and pass it down. If not, would you be open to introducing a lightweight context in the Sidebar component to track the registered paths and determine the best match?

Let me know your thoughts on the architectural approach, and I'd be happy to rework the PR to implement it!

Issue #33120, opened 2026-03-04
Add a retrieve-techdocs-content action to the TechDocs backend


thomvaill:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
techdocs action mcp
```

### 🗃️ Project Area

TechDocs

### 🔖 Need

When using the search action proposed in #31010, an agent can find TechDocs pages by querying the search engine. But the search results only return a short excerpt (`text` field) and a URL (`location`), not the full document content.

For catalog entities, the agent can call the `get-catalog-entity` action to get the full entity details. But there is no equivalent for TechDocs: there is no action to retrieve the actual content of a documentation page.

The agent could try to fetch the URL directly using a web tool, but this would fail in case of authentication and would return raw HTML instead of clean text.

A `retrieve-techdocs-content` action would complete the search action by letting agents actually read the documentation they find.

The two actions combined would make Backstage a "[context7](https://context7.com/) like"

### 📝 Proposal

This proposal is heavily inspired by the great work from @johnmcollier in [backstage/community-plugins#6013](https://github.com/backstage/community-plugins/pull/6013) (specifically the [`plugin.ts`](https://github.com/backstage/community-plugins/pull/6013/files#diff-6480ec7a596f7fbbfbf381f5b25700e7e0587ffbbb2eea170852be846760a777R354) file), which already implemented a `retrieve-techdocs-content` action as a community plugin. Thanks for paving the way!

The idea here is to bring this into the core TechDocs backend plugin, with some adjustments to make it work well with the proposed search action in #31010.

**Input (two ways to call it):**

1. Using `location` from a search result (easiest after a search query):

   - `location` (string): the TechDocs URL path as returned by the search action, e.g. `/docs/default/component/my-service/guides/deployment`. The action parses it internally to extract the entity and page.

2. Using explicit entity reference (when you already know the entity):
   - `entityRef` (string): entity reference in the format `kind:namespace/name` (e.g. `component:default/my-service`)
   - `pagePath` (optional, string): path to a specific page (defaults to `index.html`)

One of `location` or `entityRef` must be provided. The `location` input is important because the search action (#31010) returns a `location` field in each result. Without it, the agent would have to parse the TechDocs URL structure to reconstruct an `entityRef` and `pagePath`, which is fragile and error-prone.

**Output:**

[...]

PR #33412, opened 2026-03-18
feat(techdocs): add get-techdocs-metadata MCP action

thenmozhi-loganathan:
Adds a get-techdocs-metadata action to the TechDocs backend plugin, registered with the Actions Registry Service (alpha), so AI assistants can discover what documentation exists before fetching it.

The action returns a TechDocs site's name, description, navigation structure, and page list for a given entity. A typical flow is metadata first to locate the relevant page, then a targeted content fetch — rather than blindly pulling index.html and burning tokens on the wrong page.

Responses are capped at 5MB with a 15s request timeout to prevent resource exhaustion.

**Scope note:** this PR originally also added a get-techdocs-content action that converted generated HTML to Markdown. Per the Framework SIG discussion in #35038, that approach is superseded by reading preserved source Markdown. The content action has been removed here and will return in a follow-up PR once #35039 lands. This PR is now independent of that work.


thomvaill:
Hey! Thank you for your PR.
However as discussed in https://github.com/backstage/backstage/issues/33120, I think it is really important that the MCP returns pages as Markdown and not HTML (because better understanding of LLMs and less token consumption).
Also, I don't think the metadata action is that useful. Do you have a usecase in mind?

thenmozhi-loganathan:
> Hey! Thank you for your PR. However as discussed in #33120, I think it is really important that the MCP returns pages as Markdown and not HTML (because better understanding of LLMs and less token consumption). Also, I don't think the metadata action is that useful. Do you have a usecase in mind?

Thanks for the feedback @thomvaill!

**Markdown Output:** Updated `get-techdocs-content` to return Markdown instead of HTML using `node-html-markdown` (already used in the confluence-to-markdown scaffolder module).

**Metadata Action Use Cases:**

1. **Navigation Discovery** — AI can see what pages exist via the `nav` structure and guide users to the right content
2. **Context Building** — Site name/description helps AI understand docs before fetching content
3. **Efficient Multi-page Retrieval** — AI can selectively fetch relevant pages instead of blindly fetching index.html
4. **Search Optimization** — Knowing structure helps AI pick the right page for user questions
5. **Reduced Token Waste** — Lightweight metadata response vs fetching large pages to discover structure

**Example workflow:**
User: "How do I configure authentication?"
AI: 1. get-techdocs-metadata → sees nav has "Configuration" > "Authentication"
    2. get-techdocs-content with path "configuration/authentication/index.html"
    3. Returns focused answer

Let me know if you'd like any adjustments!

thenmozhi-loganathan:
Please re-open this PR, it is waiting for review.

pshenbagadelip:
We are looking for this feature and when are we planning to review and release this mcp tool?

josejobin45:
Is there any other PR's related to this already in work? Why isn't this merged already.

egiddy:
Any update on this?

0x2b3bfa0:
@thenmozhi-loganathan & @thomvaill, any additional feedback / updates on this? This feature would be extremely useful for a lot of people.

awanlin:
Hi @thenmozhi-loganathan, trying to follow up here as my message from a few weeks ago in didn't get a reply: https://github.com/backstage/backstage/issues/33120#issuecomment-4927696796

For everyone else wanting this feature that's the current blocker. I can't make the Framework SIG myself so if any of you can raise this during the next one that would really help move this forward. 🙏

awanlin (changes requested):
Hi @thenmozhi-loganathan, sorry for the extra ping, want to make sure this gets in my queue to be reviewed. Until the path forward is agreed there's not much more to be done here. If you could clean up all the comments - this will make it a lot easier to review as I'll know what has been actioned, right now there are many that are open so I don't know if you have addressed them or they are on longer applicable.

I'd also give a look into [Turndown](https://github.com/mixmark-io/turndown) over `node-html-makrdown`, if we go the HTML conversion path I feel that Turndown has better options longterm.

awanlin:
Hi @thenmozhi-loganathan, is it safe to assume that you won't be working on this? That's fair as this has been a slow process but I want to make sure I don't move on with a different contributor without giving you the chance to pick this back up. 🙏

thenmozhi-loganathan:
> Hi @thenmozhi-loganathan, is it safe to assume that you won't be working on this? That's fair as this has been a slow process but I want to make sure I don't move on with a different contributor without giving you the chance to pick this back up. 🙏

Hi @awanlin, Apologise for the late response. I am happy to contribute here, can you please provide some brief about SIG meeting and When is the next one? Is walking through the design and PR changes sufficient? would like to know the meeting expectation.

awanlin:
Hi @thenmozhi-loganathan, thanks for following up. So I share this in the SIG, though I could not attend, and we got feedback that what I proposed in https://github.com/backstage/backstage/issues/35038 makes sense and converting from HTML to Markdown is not ideal. Given that this is blocked till https://github.com/backstage/backstage/pull/35039 gets merged. 

What we could do it break out `get-techdocs-metadata` into a dedicated PR as that is something we could ship faster.

What do you think?

thenmozhi-loganathan:
> Hi @thenmozhi-loganathan, thanks for following up. So I share this in the SIG, though I could not attend, and we got feedback that what I proposed in #35038 makes sense and converting from HTML to Markdown is not ideal. Given that this is blocked till #35039 gets merged.
> 
> What we could do it break out `get-techdocs-metadata` into a dedicated PR as that is something we could ship faster.
> 
> What do you think?

Hi @awanlin, thanks for the SIG follow-up and for the pointer to #35038 / #35039 — that rationale makes sense to me. Converting MkDocs HTML back to Markdown was always going to be lossy for admonitions, tabbed content, and build-time-rendered diagrams, so reading preserved sources is clearly the better long-term shape.

Happy to go with the split, and I've gone ahead and done it — this PR is now get-techdocs-metadata only:

Removed the content action and its tests
Dropped the node-html-markdown dependency (the PR no longer touches package.json or yarn.lock at all)
Trimmed the README and changeset to match
That leaves no HTML-conversion surface, so it's unblocked and shippable independently. I've kept the createTechdocsActions wrapper in place so the content action can slot straight back in later.

Once #35039 lands I'll open a follow-up PR for get-techdocs-content that reads the _sources/ Markdown directly. That also makes the Turndown question moot, since there'd be no HTML conversion at all.

I've reused this PR rather than opening a new one so the discussion history and signed-off commits carry over — happy to open a clean one instead if you'd prefer.

I've also resolved all the older Copilot threads — they were actioned a while back, sorry for leaving them open and making this harder to review.

.changeset/techdocs-metadata-action.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/plugin-techdocs-backend': minor
+---
+
+Added a `get-techdocs-metadata` action to the TechDocs backend plugin, registered with the Actions Registry Service (alpha).
```
awanlin, line 5: Keep this and remove the rest. When you are creating changesets and you feel they need to be longer that should be an indicator that you should be adding it to the docs. So please add some docs about this action here: https://backstage.io/docs/ai/well-known-actions

plugins/techdocs-backend/src/actions/createGetTechdocsMetadataAction.test.ts
awanlin: New files should have the current year, existing files don't get updated. In this case it should be 2026. Make sure to do that for all the new files in this PR 👍 

FYI - AI tools for some reason trip up on this and we updated our `AGENTS.md` file to take this into account but that was done after your PR was raised.

plugins/techdocs-backend/src/actions/createGetTechdocsMetadataAction.ts
awanlin, line 39: Should these 3 `const`s be configurable? I'm not sure we have patterns for that related to Actions but would like to at least consider this.

plugins/techdocs-backend/README.md
awanlin: We prefer not to have documentation added here, this is not where people should be looking for this, especially for core features. I've noted where in a different comment. Revert this file changes, please. 🙏

plugins/techdocs-backend/src/actions/createGetTechdocsMetadataAction.ts
awanlin, line 63: Please add permissions to this as well: https://backstage.io/docs/backend-system/core-services/actions-registry#adding-a-permission-to-an-action. I think Catalog Entity Read is what's being used by TechDocs, please double check: https://github.com/backstage/backstage/blob/7c1019c4020ccc07119bdc1f839b35508e1260f2/plugins/catalog-common/src/permissions.ts#L47

awanlin: Should this be something more generic?

awanlin: Is this actually needed given the `...` (spread) done above? I might be reading this wrong though.

awanlin (changes requested):
Thanks for the follow up @thenmozhi-loganathan, and for moving this to focus on the action we should be able to ship sooner. Gave this a full review and left some comments. 👍

plugins/techdocs-backend/src/actions/createGetTechdocsMetadataAction.ts
thenmozhi-loganathan: I had a proper dig and I don't think catalogEntityReadPermission will work here. The actions-registry docs say visibilityPermission must be a BasicPermission, not a resource permission, but catalogEntityReadPermission is created with resourceType: RESOURCE_TYPE_CATALOG_ENTITY and is typed as CatalogEntityPermission = ResourcePermission<...>, so it's exactly the shape that isn't allowed.

The only precedent I could find in the repo is createDryRunTemplateAction, which uses templateDryRunPermission — a purpose-built basic permission with attributes: {} and no resourceType. So the equivalent here would be defining a new one, something like:


export const techdocsMetadataReadPermission = createPermission({
  name: 'techdocs.metadata.read',
  attributes: { action: 'read' },
});
The wrinkle is that techdocs-common doesn't define any permissions today, so this would mean adding a new public export to a second package — which felt like a bigger call than I wanted to make unilaterally on your behalf, especially with naming that we'd then be stuck with.

Worth noting the action already authorizes each request through auth.getPluginRequestToken, so this is really about visibility in listings rather than access control on the data itself.

How would you like to play it? Happy to add the new permission in this PR if you're comfortable with the name, or split it out if you'd rather it got its own discussion. Just let me know which you prefer.

thenmozhi-loganathan: I went looking for a precedent before answering. I couldn't find one — none of the existing actions in the repo (catalog.*, scaffolder.*, search.query) pull limits from config, so there doesn't seem to be an established pattern for this yet.

My instinct is to leave them hard-coded for now. They're internal safety valves rather than tuning knobs, and techdocs_metadata.json is realistically a few KB — 5MB is already a very generous ceiling that exists to stop a pathological response eating memory, not something an adopter would sensibly want to raise. Same for the 15s timeout.

TEXT_PAGE_EXTENSIONS I'd argue is different again — that one's really "what counts as a documentation page", which feels like it should stay consistent rather than vary per install.

That said, I don't feel strongly, and you have far better context on where this is heading. Happy to move any or all of them into config if you'd prefer

awanlin: Yeah, we can leave this for a follow up based on community feedback.

awanlin: This one I'm strongly opinionated, we need some sort of permission. Feel like this will catch us if we don't have something. ~~There is a permission being added here: https://github.com/backstage/backstage/pull/34437. I wonder if we could copy that in this PR? Which ever PR gets in first just needs to rebase. If not then just a new basic permission could work as well.~~ I should have read more, this only works with a basic permission so that other work won't help. So lets add a basic permission, I'd call it `techdocs.actions.read` as we know we want to add other actions right just waiting for various parts to fall in place first.

awanlin (changes requested):
Thanks for the follow up @thenmozhi-loganathan, just the permission item left: https://github.com/backstage/backstage/pull/33412/changes#r3993235218

thenmozhi-loganathan:
> Thanks for the follow up @thenmozhi-loganathan, just the permission item left: https://github.com/backstage/backstage/pull/33412/changes#r3993235218

Done — added techdocs.actions.read as you suggested.

I put it in techdocs-common behind the /alpha entry point, following the same pattern as catalog-common and scaffolder-common. That package didn't have any permissions before, so it also picks up an exports/typesVersions entry and a dependency on @backstage/plugin-permission-common. The API report confirms it resolves as a BasicPermission, so it's the right shape for visibilityPermission.

Also exported a techdocsPermissions array alongside it, matching catalogPermissions — so future TechDocs actions have somewhere obvious to land, as you mentioned.

Separate changeset for techdocs-common since it's a new public API.

PR #33417, opened 2026-03-18
Optimize MakeSortedExtensionsMap to avoid UnionToArray for tuples

freben:

Optimizes `MakeSortedExtensionsMap` in `@backstage/frontend-plugin-api` to accept extension tuples directly from `createFrontendPlugin`, avoiding an expensive `UnionToArray` round-trip that caused O(n²) type instantiations.

### Problem

`createFrontendPlugin` receives extensions as a tuple (`TExtensions extends readonly ExtensionDefinition[]`), but `MakeSortedExtensionsMap` accepted a union (`UExtensions extends ExtensionDefinition`). This meant the tuple was converted to a union via `TExtensions[number]`, then `UnionToArray` (using `PopUnion` + recursive `Exclude`) converted it back to an array for sorting. Each `Exclude<U, R>` step requires structural comparison of all remaining union members — O(n²) in type instantiations, and very expensive when extension types are structurally complex (custom blueprints with zod config schemas, complex params, deeply nested outputs).

### Fix

Changed `MakeSortedExtensionsMap` to accept the tuple directly:

```typescript
export type MakeSortedExtensionsMap<
  TExtensions extends readonly ExtensionDefinition[],
  TId extends string,
> = number extends TExtensions['length']
  ? ExtensionArrayToMap<SortExtensions<UnionToArray<TExtensions[number]>>, TId>
  : ExtensionArrayToMap<SortExtensions<[...TExtensions]>, TId>;
```

- **Tuple path** (common case — inline extensions): `[...TExtensions]` goes directly to `SortExtensions`, skipping `UnionToArray` entirely
- **Generic array fallback** (e.g. spread variables): falls back to `UnionToArray`, which is cheap since the union collapses to a single base type

The `number extends TExtensions['length']` check distinguishes tuples (literal length) from generic arrays (length = `number`).

### Benchmark (isolated, 25 tagged `ExtensionDefinition` variants)

| Metric | Before (UnionToArray) | After (direct tuple) | Delta |
|---|---|---|---|
| Instantiations | 135,153 | 117,295 | **-13.2%** |
| Check time | ~1.03s | ~0.79s | **-23%** |

The improvement scales with both the number of extensions and the structural complexity of each extension type — plugins with many complex custom blueprint extensions will see the largest gains.

### DIFF
--- a/.changeset/tough-vans-take.md
+++ b/.changeset/tough-vans-take.md
@@ -2,4 +2,4 @@
 '@backstage/frontend-plugin-api': patch
 ---
 
-Simplify types for MakeSortedExtensionsMap
+Optimized `MakeSortedExtensionsMap` to accept extension tuples directly, avoiding an expensive `UnionToArray` conversion that caused O(n²) type instantiations when `createFrontendPlugin` was called with many inline extensions.


freben:
@Rugvip you are right that the main repo isn't affected much. 

```
  What we measured (isolated UnionToArray on 25-member union of ExtensionDefinition variants):

  ┌────────────────┬───────────────────┬────────────────────┬────────┐
  │     Metric     │ With UnionToArray │ Direct tuple (fix) │ Delta  │
  ├────────────────┼───────────────────┼────────────────────┼────────┤
  │ Instantiations │ 135,153           │ 117,295            │ -13.2% │
  ├────────────────┼───────────────────┼────────────────────┼────────┤
  │ Check time     │ ~1.03s            │ ~0.79s             │ -23%   │
  ├────────────────┼───────────────────┼────────────────────┼────────┤
  │ Types          │ 36,061            │ 35,089             │ -2.7%  │
  └────────────────┴───────────────────┴────────────────────┴────────┘

  Why the OSS repo doesn't show dramatic improvement: The open-source extension types (PageBlueprint, ApiBlueprint, etc.) are
  structurally simple. The O(n²) UnionToArray is cheap per step when each Exclude<U, R> operates on simple types. The 700s blowup
   you observed synthetically is because complex extension types (custom blueprints with zod config schemas, complex params, deeply
  nested outputs) make each Exclude step enormously expensive — the structural comparison dominates.

  The fix eliminates UnionToArray entirely for the tuple path, which removes the O(n²) Exclude steps. The improvement scales with
   both the number of extensions and the structural complexity of each extension type.
```

freben:
Regression test for `tsc:full` before/after:

```
  ┌─────┬────────────┬────────┐
  │ Run │ Fix branch │ Master │
  ├─────┼────────────┼────────┤
  │ 1   │ 164s       │ 205s   │
  ├─────┼────────────┼────────┤
  │ 2   │ 184s       │ 153s   │
  ├─────┼────────────┼────────┤
  │ 3   │ 164s       │ 144s   │
  ├─────┼────────────┼────────┤
  │ Avg │ 170.7s     │ 167.3s │
  └─────┴────────────┴────────┘

  The averages are essentially identical (170.7s vs 167.3s, ~2% difference well within noise). Both branches had one outlier run
  (184s fix, 205s master) — likely background system load. The fix branch does not make yarn tsc:full slower.
```

packages/frontend-plugin-api/src/wiring/MakeSortedExtensionsMap.ts
```diff
@@ -126,6 +126,8 @@ type ExtensionArrayToMap<
 // correctness or API stability point of view.
 /** @ignore */
 export type MakeSortedExtensionsMap<
-  UExtensions extends ExtensionDefinition,
+  TExtensions extends readonly ExtensionDefinition[],
   TId extends string,
-> = ExtensionArrayToMap<SortExtensions<UnionToArray<UExtensions>>, TId>;
+> = number extends TExtensions['length']
+  ? ExtensionArrayToMap<SortExtensions<UnionToArray<TExtensions[number]>>, TId>
+  : ExtensionArrayToMap<SortExtensions<[...TExtensions]>, TId>;
```
Rugvip, line 133: Bit confused about this ternary, it seems like if it's an array we convert it to a union and then back to an array? Basically shouldn't we just leave `ExtensionArrayToMap<SortExtensions<[...TExtensions]>, TId>` and get rid of the rest? 🤔

freben: Claude did skip the ternary in the first iteration, but realised when running tsc that it fails - and added it in as a fix.

You get a lot of errors like

```
packages/frontend-test-utils/src/app/renderTestApp.tsx:107:28 - error TS2345: Argument of type '"sign-in-page:app"' is not assignable to parameter of type 'never'.

107     appPlugin.getExtension('sign-in-page:app').override({
                               ~~~~~~~~~~~~~~~~~~

packages/frontend-test-utils/src/app/renderTestApp.tsx:107:57 - error TS2345: Argument of type '{ disabled: true; }' is not assignable to parameter of type '{ attachTo?: { id: string; input: string; relative?: undefined; } | { relative: { kind?: string | undefined; name?: string | undefined; }; input: string; id?: undefined; } | ExtensionInput<ExtensionDataRef, { ...; }> | undefined; ... 5 more ...; factory?: ((originalFactory: <TFactoryParamsReturn extends never>(conte...'.
  Type '{ disabled: true; }' is not assignable to type '`Error: The extension factory has undeclared output(s): ${any}`'.

107     appPlugin.getExtension('sign-in-page:app').override({
                                                            ~
108       disabled: true,
    ~~~~~~~~~~~~~~~~~~~~~
109     }),
    ~~~~~
```

Rugvip: What about dropping the ternary and leaving just `ExtensionArrayToMap<SortExtensions<UnionToArray<TExtensions[number]>>, TId>`? do we still get the benefits from that?

freben: The array -> union -> array step was what was causing the issues in the first place. The fix explicitly tries to fall back to that path only for small corner cases. I could try it but I don't have much hope for it becoming better than what we had before 🤔

freben:
Verified to NOT be a sufficient fix for an internal case

PR #33455, opened 2026-03-20
scaffolder: migrate NFS plugin to use SubPageBlueprint page layout

Rugvip:

This migrates the scaffolder plugin's new frontend system (NFS) definition to use the `SubPageBlueprint` pattern with tabbed page layout, mirroring the approach used by the Settings and DevTools plugins.

Previously the NFS scaffolder page used a single `PageBlueprint` with a `loader` that rendered the entire `InternalRouter` — a flat React Router `<Routes>` where every page rendered its own `<Page>` + `<Header>`. This refactor splits the plugin into proper sub-pages:

- **Templates** — default tab, with a nested route for the template wizard (`/:namespace/:templateName`)
- **Tasks** — task list tab, with a nested route for the ongoing task detail (`/:taskId`)  
- **Actions** — installed actions documentation
- **Template Editor** — editor intro, with nested routes for the full editor, form previewer, and custom fields explorer
- **Templating Extensions** — template filters, functions, and values

The `PageBlueprint` now has no `loader` and instead receives sub-pages as inputs, letting the framework's `PageLayout` render the header and tab navigation. Content-only variants were extracted from `TemplateWizardPage` and `OngoingTask` for use in the NFS path, keeping the legacy frontend system components unchanged.

image

### DIFF
--- a/.changeset/scaffolder-nfs-page-layout.md
+++ b/.changeset/scaffolder-nfs-page-layout.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-scaffolder': patch
 ---
 
-Fixed the layout of the scaffolder plugin in the new frontend system to use the new page layout. Direct `FormFieldBlueprint` attachments to the scaffolder page are no longer consumed, custom form fields should instead be provided through the form fields API.
+Fixed the layout of the scaffolder plugin in the new frontend system to use the new page layout.
--- a/.changeset/scaffolder-nfs-task-header-menu.md
+++ b/.changeset/scaffolder-nfs-task-header-menu.md
@@ -1,5 +1,0 @@
----
-'@backstage/plugin-scaffolder': patch
----
-
-Added a header with a kebab context menu to the scaffolder task page in the new frontend system, restoring the ability to show and hide page elements like logs and the button bar.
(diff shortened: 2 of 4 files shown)


plugins/scaffolder/src/alpha/formFieldsApi.ts
```diff
@@ -39,25 +40,43 @@ export const formFieldsApi = ApiBlueprint.makeWithOverrides({
...
-            return internalFormFields;
+            return formFields.map(OpaqueFormField.toInternal);
           },
         }),
       }),
     );
   },
 });
 
+function getPageFormFieldLoaders(appTreeApi: typeof appTreeApiRef.T) {
```
Rugvip, line 62: This is the new solution for handling the old formFields input for now. Since I'd like to ship a patch for this we can't break things. Basically we just reach over to the other part of the tree and extract the output, rather than having the scaffolder page handle it, since it's now broken up

plugins/scaffolder/src/alpha/extensions.tsx
```diff
@@ -39,33 +40,88 @@ export const scaffolderPage = PageBlueprint.makeWithOverrides({
...
+
+export const scaffolderActionsSubPage = SubPageBlueprint.make({
+  name: 'actions',
+  params: {
+    path: 'actions',
+    title: 'Actions',
+    loader: () =>
+      Promise.all([
+        import('../components/ActionsPage'),
+        import('@backstage/core-components'),
```
freben, line 89: well this is fancy

Rugvip: oh, interesting 😅

Rugvip:
Alright, thoroughly tested in both NFS and OFS locally, let's :shipit:

PR #33510, opened 2026-03-22
docs: use generic examples in backend migration guide to avoid deprecated APIs

LoveChauhan-18:
Hey, I just made a Pull Request!
This PR replaces references to the deprecated and now removed KubernetesBuilder in the backend migration documentation with a generic ExampleBuilder.

Using specific plugin implementations in general migration guides leads to stale documentation when those plugins are refactored. Moving to a generic example ensures the guide remains accurate and useful for all plugin developers.

Changes:
Updated 

08-migrating.md
 to use a generic ExampleBuilder and ExamplePlugin pattern.
Cleaned up 

authenticationstrategy.md
 by removing the obsolete legacy backend system section and fixing broken links to the deleted KubernetesBuilder.ts file.
Formatted modified files with Prettier to match project standards.

### DIFF
--- a/docs/backend-system/building-plugins-and-modules/08-migrating.md
+++ b/docs/backend-system/building-plugins-and-modules/08-migrating.md
@@ -7,31 +7,31 @@ description: How to migrate existing backend plugins to the new backend system
 
 Migrating an existing backend plugin to the new backend system is fairly straightforward. The process is similar across the majority of plugins which just return a `Router` that is then wired up in the `index.ts` file of your backend. The primary thing that we need to do is to make sure that the dependencies that are required by the plugin are available, and then registering the router with the HTTP router service.
 
-Let's look at an example of migrating the Kubernetes backend plugin. In the existing (old) system, the kubernetes backend is structured like this:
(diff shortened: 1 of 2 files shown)


LoveChauhan-18:
Thanks for the detailed review!

I've addressed the suggested changes:

* Fixed broken markdown references
* Aligned file naming for consistency
* Corrected grammar and formatting issues
* Updated type definitions to match the intended usage

Please let me know if any further improvements are needed. I appreciate your feedback!

awanlin:
Hi @LoveChauhan-18, appreciate the contribution but in this case this isn't a change we want. At this point we likely want to simply delete this page as we removed support for the legacy backend nearly a year ago. This needs to be decided by the Docs Maintainers.

Some feedback about this PR more generally: 

1. If you ask to work on an issue it's best to wait for a reply so you can get some feedback on direction, could have avoided this scenario were we close your PR.
2. Simply running this through an AI tool doesn't mean that you should not take time to think through if the change itself makes sense. In this case simply renaming things to example is not terribly helpful for someone trying to use the guide so a more concrete example would have been better. This likely wasn't a good issue for someone to start off with as you need a lot of prior Backstage history.
3. Don't blow away the PR check list, it's there for a reason and helps us confirm you read it over and considered the items it asks about.

PR #33516, opened 2026-03-22
fix(auth): improve SignInResolverFactory Zod schema usage

Teo-ShaoWei:
## Context

Contributes to #30607

When the Zod version in a user's project does not align with the one in Backstage core, it can result in "Type instantiation is excessively deep and possibly infinite" errors, `yarn lint:all` running out of memory, and similar issues. This can break Backstage apps due to Backstage upgrades.

The root cause: `SignInResolverFactory` previously used `TOptionsInput` and `TOptionsOutput` as generic type parameters, constructing `ZodSchema<TOptionsOutput, ZodTypeDef, TOptionsInput>` from them. `ZodSchema` resolves to `z.ZodType` where `Input` defaults to `Output`, tangling the derivation of `TOptionsInput` and `TOptionsOutput` together as basis types. TypeScript struggles to resolve this in our current usage.

The fix follows the Zod recommendation for writing generic functions:
- https://v3.zod.dev/?id=writing-generic-functions
- https://zod.dev/library-authors?id=how-to-accept-user-defined-schemas

Instead of deriving the schema type from input/output types via the general `ZodSchema`, use the schema itself (`TSchema extends z.ZodType` for v4) as the generic basis and derive input/output types via `z.input<TSchema>` and `z.output<TSchema>`. While Zod v4 has improved the generics (https://zod.dev/v4/changelog?id=updates-generics), it is still cleaner and more natural to have the schema itself as the sole basis.

Additionally, the resolver option schemas across all auth providers were using `.optional()` with JavaScript-level parameter defaults (`create(options = {})`). This reduces the impact of runtime validation, since `create` never receives well-defined, complete objects. This is replaced with `.default()` and `.prefault()` (Zod v4's pre-parse default).

## Changes

- Migrate `auth-node` and all `auth-backend-module-*` packages from `zod/v3` to `zod/v4`
- Replace `zod-to-json-schema` dependency with the built-in `z.toJSONSchema` in `zod/v4`
- Replace `zod-validation-error` dependency with the built-in `z.prettifyError`
- Use `.safeParse` instead of `.parse` so that the error is well-typed
- Refactor `SignInResolverFactoryOptions` and `createSignInResolverFactory` to use `TSchema extends z.ZodType` as the generic basis instead of `TOptionsInput`/`TOptionsOutput`
- Change `SignInResolverFactory` call signature to accept `unknown` options, since input is always validated by Zod at runtime; default type parameters changed from `any` to `unknown`
- Default to `z.undefined()` when no `optionsSchema` is provided, removing the separate `!optionsSchema` code path
- Update all 16 auth-backend-module resolver files to use `.prefault()` instead of `.optional()` with parameter defaults, letting Zod handle default values at the schema level
- Implement the missing `allowedDomains` enforcement in `emailMatchingUserEntityProfileEmail`
  - The option was introduced in #28967 but was never wired up, which is considered a bug, as users would pass in allowed domains without getting the intended restriction behaviour
  - This is valuable for larger organisations that work with external vendors and want to restrict sign-in to specific email domains

### DIFF
--- a/plugins/auth-node/report.api.md
+++ b/plugins/auth-node/report.api.md
@@ -16,8 +16,7 @@ import { Profile } from 'passport';
 import { Request as Request_2 } from 'express';
 import { Response as Response_2 } from 'express';
 import { Strategy } from 'passport';
-import type { ZodSchema } from 'zod/v3';
-import type { ZodTypeDef } from 'zod/v3';
+import * as z from 'zod/v4';
 
 // @public (undocumented)
 export interface AuthOwnershipResolutionExtensionPoint {
@@ -196,9 +195,10 @@ export function createOAuthProviderFactory<TProfile>(options: {
   stateTransform?: OAuthStateTransform;
   profileTransform?: ProfileTransform<OAuthAuthenticatorResult<TProfile>>;
   signInResolver?: SignInResolver<OAuthAuthenticatorResult<TProfile>>;
-  signInResolverFactories?: {
-    [name in string]: SignInResolverFactory;
-  };
+  signInResolverFactories?: Record<
+    string,
+    SignInResolverFactory<OAuthAuthenticatorResult<TProfile>>
+  >;
 }): AuthProviderFactory;
 
 // @public (undocumented)
@@ -216,7 +216,7 @@ export function createProxyAuthProviderFactory<TResult>(options: {
   authenticator: ProxyAuthenticator<unknown, TResult, unknown>;
   profileTransform?: ProfileTransform<TResult>;
   signInResolver?: SignInResolver<TResult>;
-  signInResolverFactories?: Record<string, SignInResolverFactory>;
+  signInResolverFactories?: Record<string, SignInResolverFactory<TResult>>;
 }): AuthProviderFactory;
 
 // @public (undocumented)
@@ -227,15 +227,10 @@ export function createProxyAuthRouteHandlers<TResult>(
 // @public (undocumented)
 export function createSignInResolverFactory<
   TAuthResult,
-  TOptionsOutput,
-  TOptionsInput,
+  TSchema extends z.ZodType = z.ZodUndefined,
 >(
-  options: SignInResolverFactoryOptions<
-    TAuthResult,
-    TOptionsOutput,
-    TOptionsInput
-  >,
-): SignInResolverFactory<TAuthResult, TOptionsInput>;
+  input: SignInResolverFactoryOptions<TAuthResult, TSchema>,
+): SignInResolverFactory<TAuthResult, z.input<TSchema>>;
 
 // @public (undocumented)
 export function decodeOAuthState(encodedState: string): OAuthState;
@@ -637,9 +632,7 @@ export interface ReadDeclarativeSignInResolverOptions<TAuthResult> {
   // (undocumented)
   config: Config;
   // (undocumented)
-  signInResolverFactories: {
-    [name in string]: SignInResolverFactory<TAuthResult, unknown>;
-  };
+  signInResolverFactories: Record<string, SignInResolverFactory<TAuthResult>>;
 }
 
 // @public (undocumented)
@@ -662,13 +655,12 @@ export type SignInResolver<TAuthResult> = (
 ) => Promise<BackstageSignInResult>;
 
 // @public (undocumented)
-export interface SignInResolverFactory<TAuthResult = any, TOptions = any> {
+export interface SignInResolverFactory<
+  TAuthResult = unknown,
+  _TOptions = unknown,
+> {
   // (undocumented)
-  (
-    ...options: undefined extends TOptions
-      ? [options?: TOptions]
-      : [options: TOptions]
-  ): SignInResolver<TAuthResult>;
+  (options?: unknown): SignInResolver<TAuthResult>;
   // (undocumented)
   optionsJsonSchema?: JsonObject;
 }
@@ -676,13 +668,12 @@ export interface SignInResolverFactory<TAuthResult = any, TOptions = any> {
 // @public (undocumented)
 export interface SignInResolverFactoryOptions<
   TAuthResult,
-  TOptionsOutput,
-  TOptionsInput,
+  TSchema extends z.ZodType = z.ZodUndefined,
 > {
   // (undocumented)
-  create(options: TOptionsOutput): SignInResolver<TAuthResult>;
+  create: (options: z.output<TSchema>) => SignInResolver<TAuthResult>;
   // (undocumented)
-  optionsSchema?: ZodSchema<TOptionsOutput, ZodTypeDef, TOptionsInput>;
+  optionsSchema?: TSchema;
 }
 
 // @public
--- a/plugins/auth-node/src/oauth/createOAuthProviderFactory.ts
+++ b/plugins/auth-node/src/oauth/createOAuthProviderFactory.ts
@@ -33,9 +33,10 @@ export function createOAuthProviderFactory<TProfile>(options: {
   stateTransform?: OAuthStateTransform;
   profileTransform?: ProfileTransform<OAuthAuthenticatorResult<TProfile>>;
   signInResolver?: SignInResolver<OAuthAuthenticatorResult<TProfile>>;
-  signInResolverFactories?: {
-    [name in string]: SignInResolverFactory;
-  };
+  signInResolverFactories?: Record<
+    string,
+    SignInResolverFactory<OAuthAuthenticatorResult<TProfile>>
+  >;
 }): AuthProviderFactory {
   return ctx => {
     return OAuthEnvironmentHandler.mapConfig(ctx.config, envConfig => {
--- a/plugins/auth-node/src/proxy/createProxyAuthProviderFactory.ts
+++ b/plugins/auth-node/src/proxy/createProxyAuthProviderFactory.ts
@@ -31,7 +31,7 @@ export function createProxyAuthProviderFactory<TResult>(options: {
   authenticator: ProxyAuthenticator<unknown, TResult, unknown>;
   profileTransform?: ProfileTransform<TResult>;
   signInResolver?: SignInResolver<TResult>;
-  signInResolverFactories?: Record<string, SignInResolverFactory>;
+  signInResolverFactories?: Record<string, SignInResolverFactory<TResult>>;
 }): AuthProviderFactory {
   return ctx => {
     const signInResolver =
--- a/plugins/auth-node/src/sign-in/commonSignInResolvers.ts
+++ b/plugins/auth-node/src/sign-in/commonSignInResolvers.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { z } from 'zod/v3';
+import * as z from 'zod/v4';
 import { createSignInResolverFactory } from './createSignInResolverFactory';
 import { NotAllowedError } from '@backstage/errors';
 
@@ -38,16 +38,27 @@ export namespace commonSignInResolvers {
       optionsSchema: z
         .object({
           allowedDomains: z.array(z.string()).optional(),
-          dangerouslyAllowSignInWithoutUserInCatalog: z.boolean().optional(),
+          dangerouslyAllowSignInWithoutUserInCatalog: z
+            .boolean()
+            .prefault(false),
         })
-        .optional(),
-      create(options = {}) {
+        .prefault({}),
+      create({ allowedDomains, dangerouslyAllowSignInWithoutUserInCatalog }) {
         return async (info, ctx) => {
           const { profile } = info;
 
           if (!profile.email) {
             throw new Error(
               'Login failed, user profile does not contain an email',
+            );
+          }
+
+          const [localPart] = profile.email.split('@');
+          const emailDomain = profile.email.slice(localPart.length + 1);
+
+          if (allowedDomains && !allowedDomains.includes(emailDomain)) {
+            throw new NotAllowedError(
+              'Sign-in user email is not from an allowed domain',
             );
           }
 
@@ -73,7 +84,7 @@ export namespace commonSignInResolvers {
                   },
                   {
                     dangerousEntityRefFallback:
-                      options?.dangerouslyAllowSignInWithoutUserInCatalog
+                      dangerouslyAllowSignInWithoutUserInCatalog
                         ? { entityRef: { name: noPlusEmail } }
                         : undefined,
                   },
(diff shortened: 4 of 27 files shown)


Teo-ShaoWei:
I made the fix in `zod/v3` first, which requires minor changes to change it to `zod/v4` (interface-wise), so I just go ahead.

This shouldn't cause issues for users as we qualify the import, i.e. `zod/v4` instead of `zod`. The same code should be applicable for users who are still on Zod v3. Though with v1.49 we should encourage all users to go to Zod v4 as it can still use `zod/v3`.

plugins/auth-node/src/sign-in/commonSignInResolvers.ts
```diff
@@ -51,6 +53,15 @@ export namespace commonSignInResolvers {
...
           }
 
+          const [localPart] = profile.email.split('@');
+          const emailDomain = profile.email.slice(localPart.length + 1);
+
+          if (allowedDomains && !allowedDomains.includes(emailDomain)) {
+            throw new NotAllowedError(
+              'Sign-in user email is not from an allowed domain',
+            );
+          }
```
Teo-ShaoWei, line 63: This is not directly related to the topic of this MR: improving the Zod typings. Let me know if we should raise a separate PR, or is it ok to leave it in 🙏

Some context:
The `allowedDomains` option was initially added [here](https://github.com/backstage/backstage/pull/28967/changes#diff-c226aa28d27bc5a8a3997cdebdad07e8dbb75b10243cd6c9cdc2660bb38e14a6R40). It looks like an unintended addition, so the naive bugfix should be to remove it. However, it's a useful behaviour as we might have vendors/consultants whom we wish to add user entities for, but might not want to grant them access to the Backstage instance.

Teo-ShaoWei:
@freben would like to seek your help please... 🙇 I can't figure out why the E2E Linux is failing from the log. It seems not related to this merge request, but this is the only commit on top of a merged `master` 🤔

EDIT: I saw that you are onto the issue, e.g. with #33529. Trouble you with it 🙇

plugins/auth-node/report.api.md
```diff
@@ -662,27 +655,25 @@ export type SignInResolver<TAuthResult> = (
...
+  TAuthResult = unknown,
+  _TOptions = unknown,
+> {
   // (undocumented)
-  (
-    ...options: undefined extends TOptions
-      ? [options?: TOptions]
-      : [options: TOptions]
-  ): SignInResolver<TAuthResult>;
+  (options?: unknown): SignInResolver<TAuthResult>;
```
Teo-ShaoWei, line 663: This is counterintuitive, but it makes a lot of sense in hindsight. The input to sign-in resolver factories is intended to be parsed immediately by a Zod schema. Currently, the only usage in this codebase assumes the input is some sort of unknown too:

https://github.com/backstage/backstage/blob/03ce639a9885873893da684a05a86763f91fcd74/plugins/auth-node/src/sign-in/readDeclarativeSignInResolver.ts#L49

On the other hand, `unknown` makes it much easier to write `createSignInResolverFactory` in a clean and correct manner.

The UX during tests improved too. I frequently want to write failure tests with invalid input, since the input to runtime validation can always be dirty, to verify that the handling is correct. This is not possible currently, as `yarn tsc` would complain that the input option is not of the correct type.

CleanShot 2026-03-23 at 22 53 02

plugins/auth-node/src/oauth/createOAuthProviderFactory.ts
```diff
@@ -33,9 +33,10 @@ export function createOAuthProviderFactory<TProfile>(options: {
   stateTransform?: OAuthStateTransform;
   profileTransform?: ProfileTransform<OAuthAuthenticatorResult<TProfile>>;
   signInResolver?: SignInResolver<OAuthAuthenticatorResult<TProfile>>;
-  signInResolverFactories?: {
-    [name in string]: SignInResolverFactory;
-  };
+  signInResolverFactories?: Record<
+    string,
+    SignInResolverFactory<OAuthAuthenticatorResult<TProfile>, any>
```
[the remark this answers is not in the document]
Teo-ShaoWei: This was the best I could have done with `unknown`, due to the fact that function arguments in TypeScript are contravariant.  This is likely why the original code uses `any` instead of `unknown`, and only use `unknown` at the final usage site:

https://github.com/backstage/backstage/blob/03ce639a9885873893da684a05a86763f91fcd74/plugins/auth-node/src/sign-in/readDeclarativeSignInResolver.ts#L25-L27

And then it occured to me that maybe TypeScript is right, we shouldn't be using a specific input argument just to expand into `unknown` later! This lead to https://github.com/backstage/backstage/pull/33516#discussion_r2975707608.

plugins/auth-node/src/proxy/createProxyAuthProviderFactory.ts
```diff
@@ -31,7 +31,7 @@ export function createProxyAuthProviderFactory<TResult>(options: {
   authenticator: ProxyAuthenticator<unknown, TResult, unknown>;
   profileTransform?: ProfileTransform<TResult>;
   signInResolver?: SignInResolver<TResult>;
-  signInResolverFactories?: Record<string, SignInResolverFactory>;
+  signInResolverFactories?: Record<string, SignInResolverFactory<TResult, any>>;
```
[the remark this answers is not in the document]
Teo-ShaoWei: Same as above.

plugins/auth-node/src/sign-in/createSignInResolverFactory.ts
```diff
@@ -14,70 +14,51 @@
...
-        `Invalid sign-in resolver options, ${fromError(error)}`,
+        `Invalid sign-in resolver options:\n${z.prettifyError(result.error)}`,
       );
     }
-    return options.create(parsedOptions);
+    return create(result.data);
   };
-
-  factory.optionsJsonSchema = zodToJsonSchema(optionsSchema) as JsonObject;
+  factory.optionsJsonSchema = z.toJSONSchema(schema) as JsonObject;
```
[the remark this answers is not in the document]
Teo-ShaoWei: What we have is a nett positive. There may be value in checking for `optionsJsonSchema` existence only if the user needs to differentiate between the `undefined` case and the Zod schema case. With the default behaviour sorted out and correctly blended into the Zod schema case with `z.undefined()`, there's no longer a need to do so.

plugins/auth-node/report.api.md
freben: But, I mean, is that really a positive argument though? :) The type safety was very much a feature after all. Tests can `as any` at will; other code shouldn't have to.

freben (changes requested):
This needs a rebase now.

plugins/auth-node/report.api.md
Rugvip: The option type is there to help guide consumers of the resolver, I can't see why we would want to remove it. For tests, `as any` as @freben suggested or `//@ts-expect-error`

Rugvip (changes requested):
Thanks! 👍 

Generally the change looks good, but it needs to be structured in a way that is non-breaking to users. A potential strategy for that is to introduce an overload that uses the new types, and deprecate the existing form. Important part of that would be to make sure that the new overload is sufficiently different from the existing one, such that the type system doesn't have any trouble selecting the correct one.

When moving one of the sign-in resolvers back to v3, which is what all consumers outside of this repo would currently be using, I get the following errors:

<details>
<summary>TS Errors</summary>

```
plugins/auth-backend-module-github-provider/src/resolvers.ts:36:5 - error TS2740: Type 'ZodOptional<ZodObject<{ dangerouslyAllowSignInWithoutUserInCatalog: ZodOptional<ZodBoolean>; }, "strip", ZodTypeAny, { dangerouslyAllowSignInWithoutUserInCatalog?: boolean | undefined; }, { ...; }>>' is missing the following properties from type 'ZodType<unknown, unknown, $ZodTypeInternals<unknown, unknown>>': def, type, toJSONSchema, check, and 18 more.

36     optionsSchema: z
       ~~~~~~~~~~~~~

  plugins/auth-node/src/sign-in/createSignInResolverFactory.ts:36:3
    36   optionsSchema?: TSchema;
         ~~~~~~~~~~~~~
    The expected type comes from property 'optionsSchema' which is declared here on type 'SignInResolverFactoryOptions<unknown, ZodType<unknown, unknown, $ZodTypeInternals<unknown, unknown>>>'

plugins/auth-backend-module-github-provider/src/resolvers.ts:41:5 - error TS2322: Type '(options?: unknown) => (info: SignInInfo<OAuthAuthenticatorResult<GithubProfile>>, ctx: AuthResolverContext) => Promise<...>' is not assignable to type '(options: unknown) => SignInResolver<unknown>'.
  Type '(info: SignInInfo<OAuthAuthenticatorResult<GithubProfile>>, ctx: AuthResolverContext) => Promise<BackstageSignInResult>' is not assignable to type 'SignInResolver<unknown>'.
    Types of parameters 'info' and 'info' are incompatible.
      Type 'SignInInfo<unknown>' is not assignable to type 'SignInInfo<OAuthAuthenticatorResult<GithubProfile>>'.
        Type 'unknown' is not assignable to type 'OAuthAuthenticatorResult<GithubProfile>'.

41     create(options = {}) {
       ~~~~~~

  plugins/auth-node/src/sign-in/createSignInResolverFactory.ts:37:3
    37   create: (options: z.output<TSchema>) => SignInResolver<TAuthResult>;
         ~~~~~~
    The expected type comes from property 'create' which is declared here on type 'SignInResolverFactoryOptions<unknown, ZodType<unknown, unknown, $ZodTypeInternals<unknown, unknown>>>'

plugins/auth-backend-module-github-provider/src/resolvers.ts:59:24 - error TS2339: Property 'dangerouslyAllowSignInWithoutUserInCatalog' does not exist on type '{}'.

59               options?.dangerouslyAllowSignInWithoutUserInCatalog
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

plugins/auth-backend-module-github-provider/src/resolvers.ts:73:7 - error TS2740: Type 'ZodOptional<ZodObject<{ dangerouslyAllowSignInWithoutUserInCatalog: ZodOptional<ZodBoolean>; }, "strip", ZodTypeAny, { dangerouslyAllowSignInWithoutUserInCatalog?: boolean | undefined; }, { ...; }>>' is missing the following properties from type 'ZodType<unknown, unknown, $ZodTypeInternals<unknown, unknown>>': def, type, toJSONSchema, check, and 18 more.

73       optionsSchema: z
         ~~~~~~~~~~~~~

  plugins/auth-node/src/sign-in/createSignInResolverFactory.ts:36:3
    36   optionsSchema?: TSchema;
         ~~~~~~~~~~~~~
    The expected type comes from property 'optionsSchema' which is declared here on type 'SignInResolverFactoryOptions<unknown, ZodType<unknown, unknown, $ZodTypeInternals<unknown, unknown>>>'

plugins/auth-backend-module-github-provider/src/resolvers.ts:78:7 - error TS2322: Type '(options?: unknown) => (info: SignInInfo<OAuthAuthenticatorResult<GithubProfile>>, ctx: AuthResolverContext) => Promise<...>' is not assignable to type '(options: unknown) => SignInResolver<unknown>'.
  Type '(info: SignInInfo<OAuthAuthenticatorResult<GithubProfile>>, ctx: AuthResolverContext) => Promise<BackstageSignInResult>' is not assignable to type 'SignInResolver<unknown>'.
    Types of parameters 'info' and 'info' are incompatible.
      Type 'SignInInfo<unknown>' is not assignable to type 'SignInInfo<OAuthAuthenticatorResult<GithubProfile>>'.
        Type 'unknown' is not assignable to type 'OAuthAuthenticatorResult<GithubProfile>'.

78       create(options = {}) {
         ~~~~~~

  plugins/auth-node/src/sign-in/createSignInResolverFactory.ts:37:3
    37   create: (options: z.output<TSchema>) => SignInResolver<TAuthResult>;
         ~~~~~~
    The expected type comes from property 'create' which is declared here on type 'SignInResolverFactoryOptions<unknown, ZodType<unknown, unknown, $ZodTypeInternals<unknown, unknown>>>'

plugins/auth-backend-module-github-provider/src/resolvers.ts:98:26 - error TS2339: Property 'dangerouslyAllowSignInWithoutUserInCatalog' does not exist on type '{}'.

98                 options?.dangerouslyAllowSignInWithoutUserInCatalog
                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 6 errors in the same file, starting at: plugins/auth-backend-module-github-provider/src/resolvers.ts:36
```

</details>

plugins/auth-node/report.api.md
Teo-ShaoWei: Thanks @freben and @Rugvip for giving this a thorough review. I'll properly investigate the other suggestions, I'll need some time 🙇

Specifically for this one hope to receive some guidance 🙏

This `unknown` here doesn't affect the `createSignInResolverFactory` that are written, but the created resolvers. The is only one usage of the resolver in the codebase here:

https://github.com/jonkoops/backstage/blob/f700a1ba8214c6e49ee811fab998eb2782f314b0/plugins/auth-node/src/sign-in/readDeclarativeSignInResolver.ts#L48-L50

That single usage does the correct thing: provide a runtime type-unknown value, and have it passes through Zod the first thing within the resolver.

In that case I'm not sure how we are removing type safety from this switch to `unknown`, on the other hand it looks more precise to me imo... 🤔

Sorry if I derived this wrongly... In that case could I find out where in the code we might rely on the original definition of:
```ts
...options: undefined extends TOptions
      ? [options?: TOptions]
      : [options: TOptions]
  ): SignInResolver<TAuthResult>;
```
to guide the consumers of the resolver on the typing?

Teo-ShaoWei:
@Rugvip Just to confirm the details on [this comment](https://github.com/backstage/backstage/pull/33516#pullrequestreview-4067309827).

You meant that existing users would be using Zod v3/v4 with `zod/v3` imports, is that right? I can agree with your suggestion that we introduce an overload that uses the new types, and deprecate the existing form. It's similar to what's proposed here: https://github.com/backstage/backstage/issues/30607#issuecomment-4137300815. Let's figure out how to get that to work 💪

Specifically for this MR, I had it implemented in `zod/v3` initially already, so I can try to revert to that. That would be non-breaking for users, and potentially make migrating to `zod/v4` for us easier 👍 I'll get to it when I've bandwidth 🙇

---

Just to put a pin here for our future reference: Usage of Zod v3 with `zod/v4` imports in the current state would work correctly, here's a screenshot of what I have:

CleanShot 2026-04-07 at 22 15 01

This verified that:
1. I'm on the `fix-auth-zod-typing` branch (this PR branch) on this repo.
2. All `package.json`'s Zod has been overwritten with `^3.25`.
3. The `yarn.lock` only has `zod@npm:^3.25.76` and `zod@npm:^3.25 || ^4.0`, but the latter is only used by `@modelcontextprotocol/sdk`. That doesn't affect our use here.
4. `yarn tsc` is all good.
5. The `optionsSchema` can be resolved correctly to the correct `zod/v4` types. If I mouse over `create`, the type is well-defined too.

Rugvip (changes requested):
@Teo-ShaoWei I've ended up exploring this area myself the last few days, and the direction we want to head in is likely to move towards https://standardschema.dev/. Initial move towards that and example for the frontend system in https://github.com/backstage/backstage/pull/33836.

Is this something you'd want to explore for this case as well? That means moving to standard schema rather than zod v4, which should achieve the original goal.

Teo-ShaoWei:
@Rugvip Ok understood. I agree that the standard schema is a good one-time solution. I'll refer and align to #33516 and migrate this PR.

While I'm interested in wrapping this up properly, don't wait for me, though, as I can't promise my bandwidth 🙇 If you're already working on it, just message me here to let me know so I don't double-work 👍

PR #33552, opened 2026-03-24
feat(ui): add showPaginationLabel prop to TablePagination

jroebu14:

Add a `showPaginationLabel` boolean prop to `TablePagination` and `useTable` pagination options. When set to `false`, the pagination label (e.g., "1 - 20 of 150") is hidden while navigation controls remain visible. Defaults to `true`.

- The prop originates in `TablePaginationOwnProps` and is picked through into `PaginationOptions`, following the established pattern for `showPageSizeOptions`
- When hidden, `aria-describedby` references to the label are also removed; the `VisuallyHidden` live region for screen readers is unaffected
- Works in all three pagination modes (complete, offset, cursor)

### DIFF
--- a/.changeset/show-pagination-label.md
+++ b/.changeset/show-pagination-label.md
@@ -4,4 +4,4 @@
 
 Added `showPaginationLabel` prop to `TablePagination` and `useTable` pagination options. When set to `false`, the pagination label (e.g., "1 - 20 of 150") is hidden while navigation controls remain visible. Defaults to `true`.
 
-**Affected components:** TablePagination, useTable
+**Affected components:** `TablePagination`, `useTable`


.changeset/show-pagination-label.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Added `showPaginationLabel` prop to `TablePagination` and `useTable` pagination options. When set to `false`, the pagination label (e.g., "1 - 20 of 150") is hidden while navigation controls remain visible. Defaults to `true`.
```
Naycon, line 5: By adding this, we can have the changelog script automatically pick up these changes:

```suggestion
Added `showPaginationLabel` prop to `TablePagination` and `useTable` pagination options. When set to `false`, the pagination label (e.g., "1 - 20 of 150") is hidden while navigation controls remain visible. Defaults to `true`.

**Affected components:** TablePagination, useTable
```

Naycon, line 5: Can we document this in docs-ui/ as well? There's documentation for both `useTable` and `TablePagination` props there.

jroebu14: Done — updated in 8d5a75b.

jroebu14: Added in a47f2b7 — documented in both the `useTable` `paginationOptions` description and the `TablePagination` props table.

packages/ui/src/components/Table/hooks/useTable.ts
```diff
@@ -74,6 +75,7 @@ function useTableProps<T extends TableItem>(
       },
       showPageSizeOptions,
       getLabel,
+      showPaginationLabel,
     }),
     [
```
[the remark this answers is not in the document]
jroebu14: This is a pre-existing pattern — `showPageSizeOptions` and `getLabel` are also excluded from the inner `useMemo` dependency array and only appear in the outer one (lines 109-110). I followed that convention for consistency. Happy to fix all three in a follow-up if you'd prefer, but didn't want to mix that in with this feature PR.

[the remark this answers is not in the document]
jroebu14: Fixed in da1d5c7 — moved all three (`showPageSizeOptions`, `getLabel`, `showPaginationLabel`) from the outer `useMemo` dependency array to the inner pagination `useMemo` dependency array so changes correctly trigger a new pagination object reference.

docs-ui/src/app/components/table/props-definition.tsx
```diff
@@ -413,6 +414,12 @@ export const tablePaginationPropDefs: Record<string, PropDef> = {
     values: ['(props) => string'],
     description: 'Custom function to generate the pagination label text.',
   },
+  showPaginationLabel: {
+    type: 'boolean',
+    default: 'true',
+    description:
+      'Whether to display the pagination label (e.g., "1 - 20 of 150"). When false, only navigation controls are shown.',
```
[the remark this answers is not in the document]
jroebu14: page size dropdown is sort of a navigation control but I can make this change if you think @Naycon but it feels more confusing and wordy to me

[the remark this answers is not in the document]
Naycon: Nah, all good, I like the original better.

.changeset/show-pagination-label.md
```diff
@@ -0,0 +1,7 @@
...
+
+**Affected components:** `TablePagination`, `useTable`
```
Naycon, line 7: Actually, we don't want backticks here.
```suggestion
**Affected components:** TablePagination, useTable
```

PR #33553, opened 2026-03-24
docs: Add section to the migration docs for internal only plugins

Parsifal-M:

Adds a section to the Frontend System migration docs that is more specific to internal only plugins (that don't need to support legacy)

### DIFF
--- a/docs/frontend-system/building-plugins/05-migrating.md
+++ b/docs/frontend-system/building-plugins/05-migrating.md
@@ -50,9 +50,9 @@ export { myPlugin as default } from './plugin';
 
 ### Migrating pages
 
-Replace `createRoutableExtension` with `PageBlueprint`:
+Replace `createRoutableExtension` with `PageBlueprint`. Because the `loader` returns a JSX element, any file that uses `PageBlueprint` (or any other blueprint with a JSX loader) must use a `.tsx` extension. Rename `src/plugin.ts` to `src/plugin.tsx` and update the import in `src/index.ts` to match.
 
-```tsx title="src/plugin.ts"
+```tsx title="src/plugin.tsx"
 import {
   createFrontendPlugin,
   PageBlueprint,
@@ -212,7 +212,7 @@ and the following instruction in the plugin README:
 <Route path="/foo" element={<FooPage />} />
 ```
 
-it can be migrated as the following, keeping in mind that you may need to switch from `.ts` to `.tsx`:
+it can be migrated as the following. Because the `loader` returns a JSX element, any file that uses `PageBlueprint` (or any other blueprint with a JSX loader) must use a `.tsx` extension. Rename the file from `.ts` to `.tsx` if you have not done so already:
 
 ```tsx
 import { PageBlueprint } from '@backstage/frontend-plugin-api';
--- a/docs/frontend-system/building-plugins/05-migrating.md
+++ b/docs/frontend-system/building-plugins/05-migrating.md
@@ -9,22 +9,24 @@ This guide covers how to migrate a frontend plugin and its components, routes, a
 
 The main concept is that routes, components, and APIs are now extensions. You can use the appropriate [extension blueprints](../architecture/23-extension-blueprints.md) to migrate all of them to extensions.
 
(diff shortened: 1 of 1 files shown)


Parsifal-M:
@Rugvip This is kinda what I meant when we discussed it the other day, WDYT? Not sure if it needs to be its own page or just a section in the migrating page

docs/frontend-system/building-plugins/05-migrating.md
```diff
@@ -5,9 +5,119 @@ sidebar_label: Migration Guide
...
+
+const fooPage = PageBlueprint.make({
+  params: {
+    path: '/foo',
+    routeRef: rootRouteRef,
+    loader: () => import('./components/FooPage').then(m => <m.FooPage />),
+  },
+});
+
+export const myPlugin = createFrontendPlugin({
```
Parsifal-M, line 72: I noticed looking at the new skill to migrate a plugin this is missing this part:

```tsx
  title: 'My Plugin',
  icon: <RiToolsLine />,
  info: {
    packageJson: () => import('../package.json'),
  },
```

docs/frontend-system/building-plugins/05-migrating.md
```diff
@@ -5,9 +5,119 @@ sidebar_label: Migration Guide
...
-This guide allows you to migrate a frontend plugin and its own components, routes, apis to the new frontend system.
+This guide covers how to migrate a frontend plugin and its components, routes, and APIs to the new frontend system (NFS).
 
-The main concept is that routes, components, apis are now extensions. You can use the appropriate [extension blueprints](../architecture/23-extension-blueprints.md) to migrate all of them to extensions.
+The main concept is that routes, components, and APIs are now extensions. You can use the appropriate [extension blueprints](../architecture/23-extension-blueprints.md) to migrate all of them to extensions.
+
+There are two migration paths depending on whether your plugin needs to remain compatible with the legacy frontend system:
+
+- **Internal plugins** — plugins used only within your own project. You can replace the legacy code directly without maintaining a compatibility layer.
+- **Published plugins** — plugins shared with or consumed by external Backstage apps. You must keep the legacy system support intact alongside the new system.
```
awanlin: Can we rephrase this? There are valid reasons for you to migrate using the `/alpha` export internally where you don't publish your plugin. I would call this Phased when we use the `/alpha` export and Direct when we go directly to the New Frontend System.

We'll then need to adjust the rest of these changes to take that into account.

docs/frontend-system/building-plugins/05-migrating.md
```diff
@@ -5,9 +5,119 @@ sidebar_label: Migration Guide
...
+
+This page covers both paths. If you are migrating a published plugin, skip ahead to [Migrating published plugins](#migrating-the-plugin).
+
+## Migrating internal plugins
+
+For plugins that are only used within your own project, you can migrate directly by replacing the legacy plugin code in place. There is no need to create a separate `src/alpha.tsx` entry point or maintain dual exports.
+
+### Updating the plugin definition
+
+Replace the `createPlugin` call from `@backstage/core-plugin-api` with `createFrontendPlugin` from `@backstage/frontend-plugin-api` directly in your existing `src/plugin.ts` file:
```
awanlin: We should enforce changing to `plugin.tsx` here, that's what we [ship in the template](https://github.com/backstage/backstage/blob/master/packages/cli-module-new/templates/frontend-plugin/src/plugin.tsx.hbs) and will avoid issues down the road where you might add something that requires it and have really odd errors. That will also help us avoid repeating this is later blocks as it can be assumed or we can link back to this to reinforce the fact.

awanlin (changes requested):
Thanks for this @Parsifal-M, agree that this needs to be added but I'm not sure how we are splitting these is the right approach. Happy to iterate on this if you feel my approach could be improved/changed.

docs/frontend-system/building-plugins/05-migrating.md
Parsifal-M: WDYT? A more phased/direct approach [fd7a4fb](https://github.com/backstage/backstage/pull/33553/commits/fd7a4fb1c3c2341882a6677767eb9b59c63d41f2)

LMK!

PR #33560, opened 2026-03-24
fix(home): update widget layout properties when entering edit mode

0x46616c6b:
### Bug

`CustomHomepageGrid`'s `changeEditMode` function does not update `isResizable` and `isDraggable` on existing widget layouts when entering edit mode. It only calls `setEditMode(true)` without touching the layout objects.

This means that after a user saves their homepage layout and later re-enters edit mode, all existing widgets have `isResizable: false` / `isDraggable: false` persisted in their layouts and cannot be resized or moved. Only newly added widgets (via `handleAdd`) correctly receive `isResizable: editMode`.

### Fix

Move the widget layout property update outside the `if (!mode)` guard in `changeEditMode` so it runs for both entering and exiting edit mode:

- **Entering edit mode** (`mode = true`): Sets `isResizable` / `isDraggable` to `true` on all widgets (respecting `w.resizable === false`) and calls `setWidgets(newWidgets)` to update the React state.
- **Exiting edit mode** (`mode = false`): Sets them to `false` and calls `storeWidgets(newWidgets)` to persist, same as before.

---
<sub>This fix was identified and verified in production. The changes and the PR were generated with assistance from OpenCode.</sub>

### DIFF
--- a/.changeset/fix-home-widget-resize-edit-mode.md
+++ b/.changeset/fix-home-widget-resize-edit-mode.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-home': patch
+---
+
+Fixed a bug where existing homepage widgets could not be resized or moved after re-entering edit mode on a previously saved layout. The `changeEditMode` function now updates `isResizable` and `isDraggable` on all widget layouts when entering edit mode, not only when exiting it.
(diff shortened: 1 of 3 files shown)


0x46616c6b:
I still wait for review from maintainers

Issue #20955, opened 2023-11-03
🚀 Feature: Use the Catalog presentation API


freben:
### 🔖 Feature description

Now that #17474 is merged, let's use the presentation APIs all over the place! This should help with ensuring that entities are presented uniformly with their titles / display names / etc, and with the same iconography everywhere if applicable.

### 🎤 Context

See #17474

### ✌️ Possible Implementation

Places in the code that use `EntityRefLink`/`EntityRefLinks` are already covered.

But there's still a fair number of places that just plain render `metadata.name`, or use `humanizeEntityRef`, or use some custom logic like that. Those should generally try to use the `EntityDisplayName` React component if possible, or use `EntityPresentation.promise` when in non-React code (which becomes available after #24595 is merged).

We'd appreciate the help with digging through the Backstage core and plugins, finding places where the presentation API should be used.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

No, but I'm happy to collaborate on a PR with someone else

hainenber:
Hi there, could I take up this one? Thanks!

freben:
It's free for anyone to chip away at! Several small PRs is probably better than one large - easier to review. And it's probably going to take a few stabs before it's complete so may be worth not assigning to a single person perhaps 🤔

hainenber:
I see, I was acquainted with other projects to ask for assignment firstly and not aware of contribution etiquette here

Let me take a stab at it. Thanks!

freben:
No worries, glad you asked! A lot of times we do assign to avoid trampling on each others' feet, but this one may be diverse enough to try leaving it out to begin with. Let's try and then reevaluate if needed!

handsamtw:
Hi @freben , I would love to help out this issue. After reviewing the codebase, I have a couple of questions.

Since humanizeEntityRef is a function returning a string, how is it suggested to replace EntityDisplayName, which is a React component?
I have lookup the place using EntityDisplayName, but I can't figure out how it related to the usage of humanizeEntityRef.

Thank you for the patience!

freben:
Hi!

[...]

PR #33576, opened 2026-03-25
Replace deprecated humanizeEntityRef with Catalog Presentation API

dyatko:
Migrate all `humanizeEntityRef` and `humanizeEntity` usages to the Catalog Presentation API across catalog, catalog-react, org-react, catalog-import, scaffolder, and techdocs plugins.

## Summary

- **React components**: Use `useEntityPresentation` hook (AncestryPage) or `EntityDisplayName` component
- **Async loaders**: Use `entityPresentationApiRef` with `.promise` (StepPrepareCreatePullRequest, TemplateFormPreviewer)
- **Sync contexts** (sort/filter/column factories): Accept optional `entityPresentation?: EntityPresentationApi` param and use `.snapshot` when available, falling back to `defaultEntityPresentation`
- **Deprecation**: Add `@deprecated` tags to `humanizeEntityRef` and `humanizeEntity`, pointing to `useEntityPresentation`, `EntityDisplayName`, and `entityPresentationApiRef`
- **TSDoc**: Improve documentation on `entityPresentationApiRef`, `EntityPresentationApi`, `useEntityPresentation`, `EntityDisplayName`, and `defaultEntityPresentation` with guidance on which to use when
- **Docs**: Add Entity Presentation docs page with usage examples and migration guide from `humanizeEntityRef`
- **Sidebar**: Add entry to both `mkdocs.yml` and `microsite/sidebars.ts`
- **Changesets**: One for catalog-react (deprecation), one combined for the remaining plugins

Related to https://github.com/backstage/backstage/issues/20955

### DIFF
--- a/docs/features/software-catalog/entity-presentation.md
+++ b/docs/features/software-catalog/entity-presentation.md
@@ -52,24 +52,12 @@ function MyComponent({ entityRef }: { entityRef: string }) {
 The hook subscribes to the `EntityPresentationApi` and returns a snapshot
 that may update over time as additional data is fetched in the background.
 
-### Using the API directly
+### Using the API directly (async, preferred for non-React)
 
-In contexts where hooks are not available, you can use the
-`entityPresentationApiRef` API directly. The API provides two access
-patterns:
-
-- **`.snapshot`** for synchronous access (for example in sort comparators or
-  filter callbacks):
-
-```ts
-import { entityPresentationApiRef } from '@backstage/plugin-catalog-react';
-
-const title = entityPresentationApi.forEntity(entity, {
-  defaultKind: 'Component',
-}).snapshot.primaryTitle;
-```
-
-- **`.promise`** for async contexts (for example inside data loaders):
+In non-React **async** contexts where you can `await` -- such as data
+loaders, `useAsync` callbacks, or event handlers -- use the
+`entityPresentationApiRef` API directly with `.promise` for the richest
+possible presentation:
 
 ```ts
 const presentation = await entityPresentationApi.forEntity(entity, {
@@ -78,9 +66,39 @@ const presentation = await entityPresentationApi.forEntity(entity, {
 const title = presentation.primaryTitle;
 ```
 
-The `.snapshot` path uses cached data when available, so it performs well
-even in tight loops like sorting. The `.promise` path resolves to a richer
-presentation that may include data fetched from the catalog.
+The `.promise` path resolves to a full presentation that may include data
+fetched from the catalog. **This is the preferred approach whenever an
+async context is available.**
+
+### `entityPresentationSnapshot` helper (synchronous fallback)
+
+When a synchronous return value is required and `await` is not possible --
+such as in sort comparators, column factories, or filter callbacks -- use
+`entityPresentationSnapshot` as a fallback. It accepts `Entity`,
+`CompoundEntityRef`, or string ref inputs and uses the presentation API
+when available, falling back to `defaultEntityPresentation` otherwise:
+
+```ts
+import {
+  entityPresentationSnapshot,
+  entityPresentationApiRef,
+} from '@backstage/plugin-catalog-react';
+
+// In a column factory or sort comparator where you have
+// the API instance (or undefined if not registered):
+const title = entityPresentationSnapshot(
+  entity,
+  {
+    defaultKind: 'Component',
+  },
+  entityPresentationApi,
+).primaryTitle;
+```
+
+Because this function is synchronous, it uses cached data from the
+presentation API. If the entity has been seen before, the snapshot will
+contain the full resolved title; otherwise it falls back to what can be
+extracted from the ref alone.
 
 ## Customizing entity presentation
 
@@ -124,6 +142,6 @@ Replace them as follows:
 | :---------------------------------------------------- | :--------------------------------------------------------------------- |
 | `humanizeEntityRef(entity)` in JSX                    | `<EntityDisplayName entityRef={entity} />`                             |
 | `humanizeEntityRef(entity)` in a React component      | `useEntityPresentation(entity).primaryTitle`                           |
-| `humanizeEntityRef(entity)` in a sort/filter callback | `entityPresentationApi.forEntity(entity).snapshot.primaryTitle`        |
 | `humanizeEntityRef(entity)` in an async loader        | `(await entityPresentationApi.forEntity(entity).promise).primaryTitle` |
+| `humanizeEntityRef(entity)` in a sort/filter callback | `entityPresentationSnapshot(entity, ctx, api).primaryTitle`            |
 | `humanizeEntity(entity, fallback)`                    | `useEntityPresentation(entity).primaryTitle`                           |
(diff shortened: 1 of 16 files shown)


.changeset/replace-humanize-entity-ref-catalog-import.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-import': patch
+---
+
+Replaced `humanizeEntityRef` with `defaultEntityPresentation` from the Catalog Presentation API in `StepPrepareCreatePullRequest`.
```
freben: You can probably merge all of these changesets into one, with a generic statement of implementing the new method instead of the old. Possibly a separate one for `catalog-react` saying that `humanizeEntityRef` was marked deprecated, but the rest can be a single one I think.

mkdocs.yml
```diff
@@ -62,6 +62,7 @@ nav:
           - Extending the model: 'features/software-catalog/extending-the-model.md'
           - External integrations: 'features/software-catalog/external-integrations.md'
           - Catalog Customization: 'features/software-catalog/catalog-customization.md'
+          - Entity Presentation: 'features/software-catalog/entity-presentation.md'
```
freben, line 65: It's nice to add here, but for the microsite backstage.io the important file is `microsite/sidebars.ts` actually

plugins/catalog-react/src/components/EntityRefLink/humanize.ts
```diff
@@ -76,6 +81,9 @@ export function humanizeEntityRef(
  *
  * If neither of those are found or populated, fallback to `defaultName`.
  *
+ * @deprecated Use {@link defaultEntityPresentation} instead, which provides
+ * the same resolution logic via `primaryTitle`.
```
freben: Should probably point to `useEntityPresentation` and `entityPresentationApiRef` only. Using the default is mostly meant to be something that you would do _in your own override_ of the `entityPresentationApiRef`, as like a try/catch fallback or stuff like that - not calling it directly in code.

plugins/catalog-react/src/components/EntityRefLink/humanize.ts
```diff
@@ -25,6 +25,11 @@ import get from 'lodash/get';
  * @param defaultNamespace - if set to false then namespace is never omitted,
  * if set to string which matches namespace of entity then omitted
  *
+ * @deprecated Use {@link defaultEntityPresentation} for non-React contexts,
```
freben: See the other comment about avoiding direct use of `defaultEntityPresentation`

docs/features/software-catalog/entity-presentation.md
freben, line 1: Yay this is fantastic

docs/features/software-catalog/entity-presentation.md
```diff
@@ -0,0 +1,117 @@
...
+  );
+}
+```
+
+The hook subscribes to the `EntityPresentationApi` and returns a snapshot
+that may update over time as additional data is fetched in the background.
+If no presentation API is registered, it falls back to
+`defaultEntityPresentation`.
+
+### `defaultEntityPresentation` function
```
freben: As I allude to in other comments - this was never meant to be called directly from the frontend. Whatever comes out of this one, won't match a customised presentation API - and not even match what comes out of a NOT-customised one since it won't have access to the entity body.

It's tough though. It's still better than the old `humanize` ones. I hesitate to say it, but maybe we can keep this as-is for a bit while we try to figure out what to do instead.

plugins/catalog-import/src/components/StepPrepareCreatePullRequest/StepPrepareCreatePullRequest.tsx
```diff
@@ -162,7 +162,10 @@ export const StepPrepareCreatePullRequest = (
     });
 
     return groupEntities.items
-      .map(e => humanizeEntityRef(e, { defaultKind: 'group' }))
+      .map(
+        e =>
+          defaultEntityPresentation(e, { defaultKind: 'group' }).primaryTitle,
```
freben: This is already inside an async loader. This one could use the actual real presentation API and use `.promise` to get an actual good presentation.

plugins/catalog-react/src/components/EntityDataTable/columnFactories.tsx
```diff
@@ -63,8 +60,8 @@ export const columnFactories = Object.freeze({
         </Cell>
       ),
       sortValue: entity =>
-        entity.metadata?.title ||
-        humanizeEntityRef(entity, { defaultKind: options.defaultKind }),
+        defaultEntityPresentation(entity, { defaultKind: options.defaultKind })
```
freben: could we add `entityPresentation?: EntityPresentationApi` as a parameter to `createEntityRefColumn`? And of course passing it in wherever we can. Even if the implementation uses only `.snapshot` here (due to not being in an async context), it's still better because the API pretty heavily tries to cache entities so there's a chance it's in-memory already. And if it isn't ... it'll use `defaultEntityPresentation` anyway.

plugins/catalog-react/src/components/EntityOwnerPicker/EntityOwnerPicker.tsx
```diff
@@ -203,7 +203,7 @@ export const EntityOwnerPicker = (props?: EntityOwnerPickerProps) => {
                   defaultNamespace: 'default',
                 })
               : o;
-          return humanizeEntity(entity, humanizeEntityRef(entity));
+          return defaultEntityPresentation(entity).primaryTitle;
```
freben: same here, using `.snapshot`?

freben (dismissed):

mitrioss:
Thanks for the update and for migrating `EntityOwnedPicker` to use `EntityPresentationRef`.

While testing the changes in the catalog filters (Owner picker), I noticed a small inconsistency in how the entity name is displayed.

Currently:

* When opening **Catalog → Filters → Owner**, the dropdown list still shows `metadata.name`.
* If I select an owner and then **reload the page**, the selected value is rendered using `spec.profile.displayName`.
* However, the **items in the dropdown list itself still display `metadata.name`**, not the presentation fields.

So the behavior ends up being mixed:

* **Dropdown list:** `metadata.name`
* **Selected value after reload:** `displayName`

My question is:
Is it expected that the dropdown list will also switch to using the presentation from `EntityPresentationRef` (e.g. `displayName`, then `title`, then `name` as fallback)? Or is that not planned for this component?

It would be great if the list and the selected value used the same presentation logic for consistency.

freben:
Oh, that sounds like maybe it could be made better. Any tweaks and improvements are welcome!

It's a bit tricky - since proper rendering may require async action, and SOME entities might actually already be cached from previous lookups, there can be cases where the sync lookup even results in a mixture of proper and not-proper naming.

Ideally we can make it so that we still use the sync method, BUT trigger a fetch in the background of any entities that weren't already cached, and then cause a repaint when they are all fetched. But that'll require a bit of rewiring I bet 🤔 

This system is still under development - if you have ideas for making it better, we'd appreciate hearing about it.

PR #33595, opened 2026-03-26
fix(catalog-react): fix component list truncation in EntityRelationCard due to incorrect pageSize

### LINKED ISSUE #33583: System Page : "Has components" card only shows one component instead of whole list
### 📜 Issue Labels

### 🔎 Search Terms

```plain
system components list only showing one component
components list not showing all components
system page not showing all components
```

### 🗃️ Project Area

Unknown

### 🔗 External Integration

N/A

### 📝 Description & Context

When I open the page of a system (eg. https://demo.backstage.io/catalog/default/system/artist-engagement-portal), i should see all components in "Has components" card.
However, the latest version of backstage only shows one component instead of the complete list.

https://demo.backstage.io/catalog/default/system/artist-engagement-portal
Image

After sorting the column  by "Name", the other component is shown instead.

Image

### 👍 Expected Behavior

I should see the complete list of components, and not just the first component of the system.

### 📦 Reproduction Repo

https://github.com/backstage/backstage/tree/master/packages/catalog-model/examples/systems/

### 🥾 Reproduction steps

1. Go to https://demo.backstage.io/catalog/default/system/artist-engagement-portal
2. In the card entitled "has components", only "artist-lookup" is displayed.
3. Order the column "Name" in reverse by clicking on the column "Name" twice.
4. Now, "www-artist" is shown. However, "artist-lookup" isn't.
5. From the demo data set, we can confirm that artist-engagement-portal has two components. Thus, both should be displayed in the list.

Deepesh123455:
## Description

This PR fixes an issue where the `EntityRelationCard` (specifically the "Has components" card) only displayed a single component initially, even if multiple components were associated with the system. 

### Root Cause
The issue originated in `EntityDataTable.tsx`. During the initial component mount, the data fetch is still in progress, meaning the `tableData` array is empty. 
The pagination hook was previously configured as:
`paginationOptions: { pageSize: tableData.length || 1 }`

Because `tableData.length` evaluated to `0`, the OR `||` operator forced the `pageSize` to fall back to `1`. The `useTable` hook then locked onto this internal state. Once the API call finished and hydrated the data array with multiple items, the table was restricted to rendering only the first row due to the locked `pageSize` of `1`.

### The Fix
Updated the fallback logic in `paginationOptions` to use `Number.MAX_SAFE_INTEGER` when the array is empty:
`paginationOptions: { pageSize: tableData.length > 0 ? tableData.length : Number.MAX_SAFE_INTEGER }`

This prevents the initial render from truncating the UI and allows the table to display all dynamically loaded rows correctly since `pagination={{ type: 'none' }}` is being used.

## Resolves
Fixes #33583

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/.changeset/tall-insects-mix.md
+++ b/.changeset/tall-insects-mix.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-react': patch
+---
+
+Fix component list truncation in EntityRelationCard by setting a large default pageSize
--- a/.changeset/tall-insects-mix.md
+++ b/.changeset/tall-insects-mix.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-react': patch
 ---
 
-Fix component list truncation in EntityRelationCard by setting a large default pageSize
+Fix component list truncation in EntityRelationCard by setting a large default `pageSize`
--- a/plugins/catalog-react/src/components/EntityDataTable/EntityDataTable.tsx
+++ b/plugins/catalog-react/src/components/EntityDataTable/EntityDataTable.tsx
@@ -61,7 +61,12 @@ export function EntityDataTable(props: EntityDataTableProps) {
     mode: 'complete',
     data: tableData,
     sortFn,
-    paginationOptions: { pageSize: tableData.length || 1 },
+    // FIX: Using Number.MAX_SAFE_INTEGER prevents the table pageSize from
+    // locking to 1 during the initial render when tableData is empty.
+    paginationOptions: {
+      pageSize:
+        tableData.length > 0 ? tableData.length : Number.MAX_SAFE_INTEGER,
+    },
   });
 
   return (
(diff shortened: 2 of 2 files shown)


Naycon:
Sorry for the competing PR @Deepesh123455! I think I caused this when I updated the cards the other week, and I saw this yesterday as well when working on something else. I added a PR in parallel to you where I add an option to have no pagination in `useTable` when in `complete` mode. This feels a little more idiomatic than having to pass infinity as page size. This is what I originally intended to do when doing the card PR, but it grew a little too big and then other things came in between.

Deepesh123455:
Hi @Naycon! No worries at all! I'm glad to see a more idiomatic solution being integrated into useTable. My goal was to fix the truncation issue, and your approach of adding a 'no pagination' option in complete mode definitely feels like the right long-term fix for the infrastructure.

Thank you for the detailed explanation! I've learned a lot about how Backstage handles table states through this process. I'm happy to close my PR once yours is merged, or let me know if there's anything else I can help with!

PR #33614, opened 2026-03-27
fix(catalog-react): fix relation cards showing only one entity

### LINKED ISSUE #33583: System Page : "Has components" card only shows one component instead of whole list

Naycon:

Fixes #33583 — entity relation cards (e.g., "Has components" on System pages) only displayed one entity at a time.

The root cause was that `EntityDataTable` derived `pageSize` from `data.length`, which was `0` during the initial loading render. Since `useTable` captured that value in `useState`, it stayed at `1` even after data loaded — causing `slice(0, 1)` to show only one row.

This PR adds native support for disabling pagination in `useTable`'s `complete` mode via `paginationOptions: { type: 'none' }`, which skips data slicing entirely and produces the correct `pagination` type in `tableProps`. `EntityDataTable` now uses this instead of the `pageSize: data.length` workaround.

#33595 is fixing the same thing by just passing a much larger page size, which could be an alternative, but I think the approach in this PR might be more idiomatic. 

#### Before

image

#### After

image

### DIFF
--- a/.patches/pr-33614.txt
+++ b/.patches/pr-33614.txt
@@ -1,0 +1,1 @@
+Fix entity relation cards showing only one entity


gusevda:
This fix was released in Backstage 1.49.3, but it seems like there is a package versioning issue. Without a clean dependency install, it leaves behind a stale 2.1.0 nested copy of plugin-catalog-react because plugin-catalog doesn't require ^2.1.1.

freben:
@gusevda How did you perform the update? Did you use `yarn backstage-cli versions:bump`?

gusevda:
I think I did. But never mind, if no one else reported this, let's skip it. Thanks.

PR #33616, opened 2026-03-27
feat(frontend-plugin-api): add titleRouteRef to createFrontendPlugin

Naycon:

Adds a `titleRouteRef` option to `createFrontendPlugin` that controls the plugin header title link. When set, all pages in the plugin render the title as a link to this route.

Resolution order in `PageBlueprint`: `plugin.titleRouteRef` → `plugin.routes.root` → page's own `routeRef`.

This is needed for plugins like catalog where the root route key isn't named `root`, causing the PluginHeader title to not be a link on sub-pages.

### DIFF
--- a/packages/frontend-plugin-api/src/blueprints/PageBlueprint.tsx
+++ b/packages/frontend-plugin-api/src/blueprints/PageBlueprint.tsx
@@ -96,7 +96,9 @@ export const PageBlueprint = createExtensionBlueprint({
       title ?? node.spec.plugin.title ?? node.spec.plugin.pluginId;
     const resolvedIcon = icon ?? node.spec.plugin.icon;
     const titleRouteRef =
-      (node.spec.plugin.routes as { root?: RouteRef }).root ?? params.routeRef;
+      node.spec.plugin.titleRouteRef ??
+      (node.spec.plugin.routes as { root?: RouteRef }).root ??
+      params.routeRef;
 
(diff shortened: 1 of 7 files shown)


packages/frontend-plugin-api/src/blueprints/PageBlueprint.tsx
```diff
@@ -96,7 +96,9 @@ export const PageBlueprint = createExtensionBlueprint({
       title ?? node.spec.plugin.title ?? node.spec.plugin.pluginId;
     const resolvedIcon = icon ?? node.spec.plugin.icon;
     const titleRouteRef =
-      (node.spec.plugin.routes as { root?: RouteRef }).root ?? params.routeRef;
+      node.spec.plugin.titleRouteRef ??
+      (node.spec.plugin.routes as { root?: RouteRef }).root ??
```
Rugvip, line 100: Oh sorry I totally missed this as part of #33377. Let's not do this, it's hidden behavior without any explicit contracts. I'm not sure about adding `titleRouteRef` either tbh. Is this linking functionality something we need at all honestly? 🤔, it hasn't existed in the past and we already have breadcrumbs in the page content as the main pattern right?

Rugvip (changes requested):

Rugvip (changes requested):
Given the breadcrumbs work I think we should hold this one a bit right?

PR #33618, opened 2026-03-27
feat(catalog): add Kind field to About Card

Naycon:

Adds a Kind field to the About Card. Tags moved before Type and Lifecycle, Kind placed after them. A new `aboutCard.kindField.label` translation key was added.

This is in preparation for migrating the catalog entity page header to BUI, where Kind will no longer be displayed in the header.

### Screenshots

image

### DIFF
--- a/plugins/catalog/src/components/AboutCard/AboutContent.tsx
+++ b/plugins/catalog/src/components/AboutCard/AboutContent.tsx
@@ -178,6 +178,15 @@ export function AboutContent(props: AboutContentProps) {
           />
         </AboutField>
       )}
+      <AboutField
+        label={t('aboutCard.tagsField.label')}
+        value={t('aboutCard.tagsField.value')}
+      >
+        {(entity?.metadata?.tags || []).map(tag => (
+          <Chip key={tag} size="small" label={tag} />
+        ))}
+      </AboutField>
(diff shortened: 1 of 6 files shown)


plugins/catalog/src/components/AboutCard/AboutContent.tsx
```diff
@@ -198,14 +206,7 @@ export function AboutContent(props: AboutContentProps) {
...
       )}
-      <AboutField
-        label={t('aboutCard.tagsField.label')}
-        value={t('aboutCard.tagsField.value')}
-      >
-        {(entity?.metadata?.tags || []).map(tag => (
-          <Chip key={tag} size="small" label={tag} />
-        ))}
-      </AboutField>
+      <AboutField label={t('aboutCard.kindField.label')} value={entity.kind} />
```
Rugvip: Hmm, thinking kind should probably be listed before type? Just because of the relation between the two being the same, there's a set of types for any given kind (or none)

Rugvip (approved):
Yep! 👍 

Just a thought on the order, but happy to ship

plugins/catalog/src/components/AboutCard/AboutContent.tsx
Naycon: Oh, yea that'd make sense. Fixing!

PR #33629, opened 2026-03-27
refactor(backend-defaults): migrate internal Zod usage from v3 to v4

jonkoops:
The auditor's severity log level mappings previously used a `zod/v3` `z.record()` schema with manual fallbacks for defaults and relied on casting into Zod error internals (`.received`, `.options`) that changed between v3 and v4. This replaces it with a `z.object()` schema using `.default()` so that Zod owns the default values and type inference, and derives the valid values and received input without reaching into undocumented error properties.

This does not migrate all `zod/v3` imports in the package, as the remaining usages are tied to public API types (e.g. `AnyZodObject` from `@backstage/backend-plugin-api`).

See #30607


packages/backend-defaults/src/entrypoints/auditor/utils.ts
```diff
@@ -16,8 +16,20 @@
 
 import type { Config } from '@backstage/config';
 import { InputError } from '@backstage/errors';
-import { z } from 'zod/v3';
-import { CONFIG_ROOT_KEY, severityLogLevelMappingsSchema } from './types';
+import { z } from 'zod/v4';
```
Teo-ShaoWei, line 19: Will we want to go with namespace import e.g. `import * as z from "zod/v4";` directly? It's seems to be the de facto for Zod 4 (they claim might have tree shaking improvements but I'm not sure why...). This is supported in Zod 3 too, so no worries with backward compat.

packages/backend-defaults/src/entrypoints/auditor/utils.ts
```diff
@@ -26,41 +38,32 @@ import { CONFIG_ROOT_KEY, severityLogLevelMappingsSchema } from './types';
...
-    const validKeys = (
-      res.error.issues.at(0) as unknown as Record<PropertyKey, unknown>
-    ).options as string[];
+  if (!parsed.success) {
+    const issue = parsed.error.issues[0];
+    const key = issue.path[0] as keyof typeof input;
+    const receivedValue = input[key];
     throw new InputError(
-      `The configuration value for 'backend.auditor.severityLogLevelMappings.${key}' was given an invalid value: '${value}'. Expected one of the following valid values: '${validKeys.join(
+      `The configuration value for '${CONFIG_ROOT_KEY}.severityLogLevelMappings.${key}' was given an invalid value: '${receivedValue}'. Expected one of the following valid values: '${logLevel.options.join(
```
Teo-ShaoWei, line 62: Understood that we are keeping to the current implementation for now, so we can leave it be.

We might want to make use of one of the standard formatted Zod errors instead of creating our own. E.g. this one only return error for one of the keys even if more are incorrect.

jonkoops: Yes, this is one of the considerations I made working though this, I decided to keep things as they are for now.

There are actually several places where Zod errors are handled, which can be improved (see also `zod-validation-error`). I think we can take this as a future enhancement, and apply it everywhere by migratiing to default Zod built-ins.

packages/backend-defaults/src/entrypoints/auditor/utils.ts
jonkoops: I guess we could, in this case since the package is a backend, it would not really make any difference. Personally, I have no preference in this matter.

Issue #33532, opened 2026-03-24
🪲`CatalogIndexPage` still exporting the legacy core page header


kimsagro1:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
"Catalog index page"
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

N/A

### 📝 Description & Context

I scaffold a new  backstage `1.49.2` instance using `npx @backstage/create-app@latest`

You can see the default catalog page using the new header

<img width="1915" height="899" alt="Image" src="https://github.com/user-attachments/assets/bf031500-1852-48dd-b6b3-4972c665ea8e" />

I want to have a new page at `/systems` that shows the catalog but with the kind defaulted to `system` so I added a new `PageBlueprint`

```typescript
import { createApp } from "@backstage/frontend-defaults";
import catalogPlugin from "@backstage/plugin-catalog/alpha";
import { navModule } from "./modules/nav";
import { AppIcon } from "@backstage/core-components";
import {
	createRouteRef,
	createFrontendPlugin,
	PageBlueprint,
} from "@backstage/frontend-plugin-api";
import { CatalogIndexPage } from "@backstage/plugin-catalog";

export default createApp({
	features: [catalogPlugin, navModule, createSystemsCatalogPlugin()],
});

export function createSystemsCatalogPlugin() {
	const rootRouteRef = createRouteRef();

	return createFrontendPlugin({
		pluginId: `catalog-systems`,
		extensions: [
			PageBlueprint.make({
				params: {
					title: "Systems",
					icon: <AppIcon id="kind:system" />,
					path: "/systems",
					routeRef: rootRouteRef,
					loader: async () => <CatalogIndexPage initialKind="system" />,
				},
			}),
		],
	});
}
```

However this used the legacy page header

<img width="1898" height="899" alt="Image" src="https://github.com/user-attachments/assets/44b4f73e-74ff-4803-add0-2894b532bae4" />

I believe the issue is that `CatalogIndexPage` exports `./components/CatalogPage`

https://github.com/backstage/backstage/blob/6bfcd9967b08717dd0da067bc3e3246e8afba0ba/plugins/catalog/src/plugin.ts#L101-L103

But the page itself uses `NfsDefaultCatalogPage`

[...]

PR #33630, opened 2026-03-27
feat(catalog): migrate NFS catalog pages to BUI header system

Naycon:

Migrates the catalog plugin's NFS entity page and index page from MUI Header to the BUI two-tier header system (PluginHeader + Header).

**Entity Page:**
- EntityHeader rewritten with BUI Header, `useEntityPresentation` for title
- New `useEntityTabs` hook replaces EntityTabs component tree, producing `HeaderNavTabItem[]` for grouped tab navigation
- New BUI FavoriteEntity toggle and EntityContextMenu in header actions
- Context menu icons migrated from MUI to Remixicon, with MUI icon compat styles for adopters
- CreateButton registered as `PluginHeaderActionBlueprint`
- InspectEntityDialog controlled via `?inspect` search param
- WarningPanel and MUI Alert replaced with BUI Alert
- EntityTabs/ and EntityLabels/ directories removed

**Index Page:**
- `HeaderPage` → BUI `Header`, `Content` → BUI `Container`
- CreateButton moved from inline to `PluginHeaderActionBlueprint`

**Accessibility:**
- VisuallyHidden live region on FavoriteEntity for screen reader state announcements
- `role="alert"` on error and not-found states
- Improved context menu trigger aria-label

**Breaking changes:**
- Custom headers via `EntityHeaderBlueprint` now replace the entire BUI Header including tab navigation
- `EntityContextMenuItemBlueprint` renders BUI `MenuItem` — custom icons may render differently (MUI icons are auto-sized via compat styles, but other icon types should be verified)

### Screenshots

| Before | After |
|---|---|
| image | image |
| image | image |


.changeset/cyan-ads-beg.md
```diff
@@ -0,0 +1,9 @@
+---
+'@backstage/plugin-catalog': major
+---
+
+**BREAKING**: Migrated the NFS entity page header from MUI to BUI. Custom headers provided via `EntityHeaderBlueprint` now replace the entire BUI Header including tab navigation. Previously, custom headers only replaced the header area above the tabs.
```
Rugvip: This part is still in alpha right? So we can make it a patch or minor bump and mark it as breaking alpha. Oh this also seems like an improvement for the customizability 👍

.changeset/young-ants-rhyme.md
```diff
@@ -0,0 +1,18 @@
+---
+'@backstage/plugin-catalog-react': major
```
Rugvip: Same thing here, shouldn't need a major bump

Rugvip (changes requested):
Nice! 😁 🎉 

One extra thing: the content jumpinging around depending on whether a scrollbar is needed is quite jarring. Do you think we could fix that in the BUI `Container` somehow? Feeling we ideally fix that before shipping this to adopters. Another tiny similar thing is that whe the tab menu is opened, the scrollbar disappears 🤔, that isn't quite as bad though.

Sidenote: kinda coming around to the centered container being good in some places in places like the entity that generally aren't made for wide content anyway. Feeling it's a bit of a waste on the catalog index page though, where the table and filters could've been given more space.

kimsagro1:
@Naycon will this also resolve issue #33532 - `CatalogIndexPage` still exporting the legacy core page header?

Rugvip (approved):
Nice! 🎉 

Happy to ship, perhaps this could be rebased on top of master instead and handle the header link separately?

Parsifal-M:
Just my two cents, we actually really like the current header i.e:
image

is there no way we can preserve this somehow? We're not a huge fan of the minimalist approach, it would be a shame to push us to use it or do our own custom magic instead

Or maybe a middle ground would be to at least let us style the new minimalist header with similar patterns/colors?

cdedreuille (changes requested):
I just want to pause a little bit on this update. From a MUI to BUI perspective, this update is a great addition but there's some UI updates that I would love to explore a bit more before we can confidently defend this design.
- The create button at the top does not page sense on the entity page. We should have it only on the catalog landing page
- We are missing the entity type
- The infos on the header are actually important at a glance. If you are on the API's page, we loose these infos.
- We should think about extensibility to let anyone change it to allow for more colors potentially.

@Parsifal-M We know this is a big lift to go from a big splash of colours to a more minimalist approach but as a default for Backstage we do not believe that this is a good representation and use of space. We should have better extensibility in mind to allow anyone to either theme it or replace it though.

PR #33649, opened 2026-03-28
feat(notifications-slack): add scope-based message update support

erikmiller-gusto:

When a Backstage notification is re-sent with the same scope, the notifications backend updates the existing DB record and sets notification.updated. Previously, the SlackNotificationProcessor always called chat.postMessage(), creating duplicate Slack messages.

This adds scope-based update support with a pluggable persistence backend:

- New `TimestampStore` interface abstracting timestamp persistence for scope-based message updates
- Default `CacheTimestampStore` implementation using the Backstage `CacheService` with 24-hour TTL auto-expiry — requires no database setup, migrations, or scheduled cleanup
- New `notificationsSlackTimestampStoreExtensionPoint` allowing custom `TimestampStore` implementations to be injected (e.g., for database-backed persistence)
- After each `chat.postMessage()`, store the response `ts` via the `TimestampStore`
- When `postProcess` receives a notification with `updated` set and a matching stored `ts`, use `chat.update()` instead of `chat.postMessage()`
- Scope context includes `origin` to prevent cross-origin collisions
- New `messagesUpdated` metrics counter for observability
- Graceful degradation: store lookup/write failures log warnings and fall through to `postMessage`

### Persistence Options

**Default (CacheService):** No configuration needed. Timestamps are stored in the Backstage CacheService (in-memory by default, or backed by Redis/Memcache if configured via `backend.cache.store`). Entries automatically expire after 24 hours.

**Custom store via extension point:** Users can inject any `TimestampStore` implementation by creating a backend module that registers with `notificationsSlackTimestampStoreExtensionPoint`. See the README for an example.

### DIFF
--- a/plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
+++ b/plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
@@ -547,10 +547,11 @@ export class SlackNotificationProcessor implements NotificationProcessor {
     ts: string,
   ): Promise<void> {
     try {
+      const now = this.db!.fn.now();
       await this.db!('slack_message_timestamps')
-        .insert({ origin, scope, channel, ts, created_at: new Date() })
+        .insert({ origin, scope, channel, ts, created_at: now })
         .onConflict(['origin', 'scope', 'channel'])
-        .merge({ ts, created_at: new Date() });
+        .merge({ ts, created_at: now });
     } catch (error) {
       this.logger.warn('Failed to persist Slack message timestamp', {
         origin,
--- a/plugins/notifications-backend-module-slack/src/module.ts
+++ b/plugins/notifications-backend-module-slack/src/module.ts
@@ -18,6 +18,7 @@ import {
   createBackendModule,
   resolvePackagePath,
 } from '@backstage/backend-plugin-api';
+import { Knex } from 'knex';
 import { metricsServiceRef } from '@backstage/backend-plugin-api/alpha';
 import { notificationsProcessingExtensionPoint } from '@backstage/plugin-notifications-node';
 import { SlackNotificationProcessor } from './lib/SlackNotificationProcessor';
@@ -32,7 +33,17 @@ const MIGRATIONS_DIR = resolvePackagePath(
   'migrations',
 );
 
-const CLEANUP_RETENTION_MS = 24 * 60 * 60 * 1000; // 24 hours
+const DB_MIGRATIONS_TABLE = 'notifications_module_slack__knex_migrations';
+const CLEANUP_RETENTION_SECONDS = 24 * 60 * 60; // 24 hours
+
+function nowMinus(knex: Knex, seconds: number): Knex.Raw {
+  if (knex.client.config.client.includes('sqlite3')) {
+    return knex.raw(`datetime('now', ?)`, [`-${seconds} seconds`]);
+  } else if (knex.client.config.client.includes('mysql')) {
+    return knex.raw(`now() - interval ${seconds} second`);
+  }
+  return knex.raw(`now() - interval '${seconds} seconds'`);
+}
 
 /**
  * The Slack notification processor for use with the notifications plugin.
@@ -92,6 +103,7 @@ export const notificationsModuleSlack = createBackendModule({
         if (!database.migrations?.skip) {
           await db.migrate.latest({
             directory: MIGRATIONS_DIR,
+            tableName: DB_MIGRATIONS_TABLE,
           });
         }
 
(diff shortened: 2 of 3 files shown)


kurtaking (commented):
Thanks for tackling this — my main concern is adding a persistent layer and a scheduled cleanup task to a module that's currently stateless. 

What do you think of using Backstage's built-in [CacheService](https://backstage.io/docs/backend-system/core-services/cache/) with a TTL, which gives you optional persistence without owning a DB? The core logic is all great — I'd just decouple it from the persistence mechanism and start with the simplest thing that works.

erikmiller-gusto:
My thinking had been that the base notification module already includes database persistence, so this isn't introducing something new.  Whereas using the cache service could provide a similar functionality... with limitations.

The default is an in-memory cache, which would end up providing little benefit in actively developed backstage instance that gets a lot of deployments throughout the workday.  In order to achieve parity with the frontend UI's notification behaviour, it would require folks to configure a persistent caching service (redis, memcache, etc).  At which point, we're kind of back to: why not just use a database, users are almost certainly already using so it's not introducing any real additional overhead...?

Thoughts?

erikmiller-gusto:
Alternatively, could make it optional to either use a database or the caching service for persistence

erikmiller-gusto:
Alright, so I think I've found a better middle ground and still lets me achieve my original intentions.  Use the cache by default but allow overriding it via extension point so could use a database backed store.

kurtaking:
Hey @erikmiller-gusto, I see you've made quite a few changes since your original response to me. I'm sorry it's taken me a few days to get back with you and to be honest, I will admit that my initial feedback missed the mark.. Every instance running this module already has a DB because of the backend plugin (like you said). You also brought up a great point around the in-memory cache + frequent deploys. That was a blind spot in my recommendation and the more I think about this, the more I think your original instinct was right. 

Had I seen that sooner, I would have stopped you from switching to cache service + added extension. The straightforward DB approach is probably the simplest thing that actually works.

erikmiller-gusto:
no prob, the good news is i can just back out the second commit :)

erikmiller-gusto:
@kurtaking are things in an OK state to accept this PR now?  Anything else I need to address?  Appreciate it :)

kurtaking (approved):
@erikmiller-gusto - approving on my end, but the PR will still require an additional +1 from core notification maintainers. 

-- 

For @backstage/notifications-maintainers - I first recommended maybe using the `CacheService`, but we landed on the DB-backed persistence being the right approach. Every instance running this module already has a database via the notifications backend whereas adding the `CacheService` would require something like redis to truly solve the problem (in-memory breaks down on actively deployed instances where restarts would wipe the `ts` mappings).

I tried looking for alternatives that would not require persistence, but you need the original timestamp from the Slack message in order to update it. That must be kept somewhere.

plugins/notifications-backend-module-slack/src/module.ts
```diff
@@ -54,17 +62,65 @@ export const notificationsModuleSlack = createBackendModule({
...
+
+        if (processors.length === 0) {
+          return;
+        }
+
+        const db = await database.getClient();
+
+        if (!database.migrations?.skip) {
+          await db.migrate.latest({
+            directory: MIGRATIONS_DIR,
```
Rugvip, line 105: Since this is a module it'll need a separate migrations table to not conflict with the plugin one, see for example https://github.com/backstage/backstage/blob/d229dc49adbb3059153e14cf312a4889c8a4c66e/plugins/catalog-backend-module-incremental-ingestion/src/database/migrations.ts#L29

plugins/notifications-backend-module-slack/src/module.ts
```diff
@@ -54,17 +62,65 @@ export const notificationsModuleSlack = createBackendModule({
...
+        await scheduler.scheduleTask({
+          id: 'slack-message-timestamps-cleanup',
+          frequency: { hours: 24 },
+          timeout: { minutes: 5 },
+          initialDelay: { hours: 2 },
+          scope: 'global',
+          fn: async () => {
+            const cutoff = new Date(Date.now() - CLEANUP_RETENTION_MS);
+            const deleted = await db('slack_message_timestamps')
+              .where('created_at', '<=', cutoff)
```
Rugvip: Best rely on the database clock instead of the local one, see for example https://github.com/backstage/backstage/blob/874456a2493bfcaa264e54517c903ee17df211e1/plugins/catalog-backend/src/database/operations/stitcher/getDeferredStitchableEntities.ts#L83

plugins/notifications-backend-module-slack/src/lib/SlackNotificationProcessor.ts
```diff
@@ -434,12 +473,92 @@ export class SlackNotificationProcessor implements NotificationProcessor {
...
+
+  private async saveTimestamp(
+    origin: string,
+    scope: string,
+    channel: string,
+    ts: string,
+  ): Promise<void> {
+    try {
+      await this.db!('slack_message_timestamps')
+        .insert({ origin, scope, channel, ts, created_at: new Date() })
```
Rugvip: Same here, `this.db.fn.now()`

erikmiller-gusto:
Hi @Rugvip anything else I need to take care of for this?

PR #33656, opened 2026-03-28
fix(auth-node): omit granted-scope cookie from new OAuth authorization request

### LINKED ISSUE #33099: 🐛 Bug Report: OAuth provider can request invalid previously-granted scopes
### 📜 Issue Labels

### 🔎 Search Terms

```plain
oauth provider cookie granted scope
```

### 🗃️ Project Area

Auth

### 🔗 External Integration

N/A

### 📝 Description & Context

When `plugins/auth-node/src/oauth/CookieScopeManager` is asked which scopes to request from the OAuth provider, it combines:
- the explicitly requested scopes for this request
- the required and additional scopes configured globally for the provider
- *and the previously granted scopes from the `<provider>-granted-scope` cookie (if available)*

This will break if a previously granted scope is now not granted anymore (for example, the user has been removed from some permission group).

I believe this is a bug because my expectation would be that the only scopes the application cares about are the required and additional scopes configured globally for the provider, along with any explicitly requested scopes required by the current request.

We should not request a scope we don't care about from the OAuth provider just because we were granted it in the past.

I'll say upfront I'm not an expert in the OAuth specification, but I believe it's valid for an OAuth server to return additional scopes on top of those explicitly requested (in our setup we include lists of group memberships for a user as separate scopes for example). But even if that's _not_ officially supported, there are cases where you might want to retire or rename a scope, and in which case requesting a renamed/deleted scope will always fail, even though the user is authorised to the app in question.

---

The specific case that made me notice this was as part of some work integrating our Backstage stack with an internal API I'm writing. As part of this, in my dev environment I sometimes switch from running the API on localhost and pointing Backstage at it, or pointing Backstage at our QA instance of the API.

We use scopes AND client ids to distinguish different applications, and different deployments of those applications. So my internal api will have scopes like `my-api-prod`, `my-api-qa` and `my-api-dev` (and a client id like `my-api-prod.backstage` etc). And so switching between different backends means switching the client id I use to auth, and the scopes I request. I cannot get a dev scope for a QA client id and vice versa.

I've set up an auth provider for this API, which looks like this (anonymised):

```yaml
    my-api-provider:
      development:
        metadataUrl: https://<our-internal-oauth-authorization-server>/.well-known/openid-configuration
        clientId: my-api-dev.backstage
        clientSecret: XXXXXXXXXXXXXXX
        additionalScopes:
          - 'my-api-dev ${DEFAULT_ENV_SCOPES} email'
```

By first authenticating to the dev instance I get a `my-api-provider-granted-scope` of `openid profile email my-api-dev offline_access my-api-dev offline_access openid email`.

If I then later on change my setup to point at QA instead by changing `clientId`, `clientSecret` and `additionalScopes` accordingly, the auth fails because it tries to request the previously granted `my-api-dev` scope as the `my-api-qa` clientId and gets denied.

### 👍 Expected Behavior

`<provider>-granted-scope` should be ignored when requesting new auth from an OAuth provider.

I'm not sure exactly what the fix should be in code for this (which is why I don't feel comfortable submitting a PR to fix), as I don't know all the places that cookie is used, or how many code paths use `plugins/auth-node/src/oauth/CookieScopeManager`. Naively I believe https://github.com/backstage/backstage/blob/6c6a41c4d168d9e82e01fe99f29e86540995877c/plugins/auth-node/src/oauth/CookieScopeManager.ts#L104 should be changed to only use a combination of `requestScope`, `required` and `additional` and should ignore `grantedScope` completely, but I don't know if that breaks other workflows.

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

1. Configure an OAuth provider to grant a scope - `scope-a` - to your user for some client
2. Configure a Backstage auth provider using the `OAuthApi` type, and add `scope-a` to your `additionalScopes`
Code extracts:
```typescript
export const myApiAuthApiRef: ApiRef<OAuthApi> = createApiRef({
  id: 'auth.my-api-provider',
});
```

app-config.yaml extract:
```yaml
    my-api-provider:
      development:
        metadataUrl: https://<oauth server>/.well-known/openid-configuration
        clientId: <clientId>
        clientSecret: <clientSecret>
        additionalScopes:
          - 'scope-a'
```
3. Request a token for that provider on the frontend somewhere:
```typescript
const myApiAuthApi = useApi(myApiAuthApiRef);
const token = await myApiAuthApi.getAccessToken(); // Note, we don't request any explicit scope here, we just want a token
```
4. In the network tab, observe a call to `<BACKEND_URL>/api/auth/my-api-provider/start?scope=&origin=http%3A%2F%2Flocalhost%3A3000&redirectUrl=<REDACTED>&flow=redirect&env=development` - notice no scope has been requested explicitly
5. It should return a 302 redirect to the OAuth auth server with a url like:
```
https://<OAUTH SERVER>/connect/authorize?client_id=<clientId>&scope=scope-a&response_type=code&redirect_uri=<REDACTED>& (rest of oauth state)
```
Note that it is now requesting the `scope-a` scope we added to `additionalScopes`
6. Auth should succeed and you get a valid token
7. You also get a cookie `my-api-provider-granted-scope` with `scope-a` included
8. *NOW* change your OAuth provider to return `scope-b` _instead of_ `scope-a`
9. Make a similar change to `app-config.yaml` to change `additionalScopes` to `['scope-b']`. This should be fine - we're requesting a scope the OAuth server will give us
10. Restart the Backstage app
11. Request the token in the same way again
12. This time in the network tab, you should see the same request to `<BACKEND_URL>/api/auth/my-api-provider/start` BUT this time the 302 redirect is to
```
https://<OAUTH SERVER>/connect/authorize?client_id=<clientId>&scope=scope-a%20scope-b&response_type=code&redirect_uri=<REDACTED>& (rest of oauth state)
```
Note that it's trying to request both the old and the new scope. This will fail as the OAuth server no longer grants the original scope to the user.

walsm232:

Fixes #33099

For OAuth providers that persist scopes, Backstage was reusing previously granted scopes from a session cookie when building new authorization requests. That could send stale scopes to the identity provider after a change to configuration, OAuth client, or which scopes are valid. Authorization can then fail or be denied when the request still includes scopes that are no longer valid for the user or client (even though current configuration only needs a new valid set (see #33099)).

This change no longer applies that persisted scope when starting a new authorization, only what the app is configured to need now (and what the client explicitly requests) is used. Refreshing tokens still uses persisted scope information as it did before.

### Why the old behavior existed

In #24743, scope handling was unified so required, additional, requested, and persisted (granted) scopes are merged in one place. That fixed refresh flows and inconsistent behaviour between refresh and full auth, and aligned merging of client- and server-defined scope (see the PR description in there).

I think that persisting granted scopes in a cookie was meant to support refresh when an IdP does not return or normalizes `scope`, and to keep one consistent merge model.

### How that led to this issue

I think that applying the same merge on every new `/authorize` means the cookie can outlive a config or client change. The next login then asks the IdP for scopes that are no longer valid for this app or user (e.g. renamed scope, different OAuth client, revoked access), as described in #33099.

### DIFF
--- a/.changeset/oauth-scope-cookie-authorize.md
+++ b/.changeset/oauth-scope-cookie-authorize.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-auth-node': patch
+---
+
+Fixed OAuth authorization requests incorrectly including previously granted scopes from the `<provider>-granted-scope` cookie, which could cause sign-in to fail when provider configuration changes or the identity provider no longer allows those scopes. New authorization requests now only include configured and explicitly requested scopes; persisted scopes are still used when refreshing tokens.
--- a/plugins/auth-node/src/oauth/CookieScopeManager.test.ts
+++ b/plugins/auth-node/src/oauth/CookieScopeManager.test.ts
@@ -142,15 +142,15 @@ describe('CookieScopeManager', () => {
     });
 
     await expect(manager.start(makeReq())).resolves.toEqual({
-      scope: 'g',
+      scope: '',
       scopeState: {
-        scope: 'g',
+        scope: '',
       },
     });
     await expect(manager.start(makeReq('x'))).resolves.toEqual({
-      scope: 'x g',
+      scope: 'x',
       scopeState: {
-        scope: 'x g',
+        scope: 'x',
       },
     });
 
@@ -159,13 +159,13 @@ describe('CookieScopeManager', () => {
       manager.handleCallback(makeReq(), {
         // The state is prioritized even if scope is present in the result
         result: { session: { scope: 'x,y' } } as OAuthAuthenticatorResult<any>,
-        state: { scope: 'x g' } as OAuthState,
+        state: { scope: 'x' } as OAuthState,
         origin: 'https://other.example.com',
       }),
-    ).resolves.toEqual('x g');
+    ).resolves.toEqual('x');
     expect(setGrantedScopes).toHaveBeenCalledWith(
       expect.anything(),
-      'x g',
+      'x',
       'https://other.example.com',
     );
     setGrantedScopes.mockClear();
@@ -265,15 +265,15 @@ describe('CookieScopeManager', () => {
     });
 
     await expect(manager.start(makeReq())).resolves.toEqual({
-      scope: 'granted-g required-a additional-b',
+      scope: 'required-a additional-b',
       scopeState: {
-        scope: 'granted-g required-a additional-b',
+        scope: 'required-a additional-b',
       },
     });
     await expect(manager.start(makeReq('x'))).resolves.toEqual({
-      scope: 'requested-x granted-g required-a additional-b',
+      scope: 'requested-x required-a additional-b',
       scopeState: {
-        scope: 'requested-x granted-g required-a additional-b',
+        scope: 'requested-x required-a additional-b',
       },
     });
 
--- a/plugins/auth-node/src/oauth/CookieScopeManager.ts
+++ b/plugins/auth-node/src/oauth/CookieScopeManager.ts
@@ -99,9 +99,7 @@ export class CookieScopeManager {
     req: express.Request,
   ): Promise<{ scopeState?: Partial<OAuthState>; scope: string }> {
     const requestScope = splitScope(req.query.scope?.toString());
-    const grantedScope = splitScope(this.cookieManager?.getGrantedScopes(req));
-
-    const scope = this.scopeTransform(requestScope, grantedScope);
+    const scope = this.scopeTransform(requestScope, []);
 
     if (this.cookieManager) {
       // If scopes are persisted then we pass them through the state so that we
--- a/plugins/auth-node/src/oauth/createOAuthRouteHandlers.test.ts
+++ b/plugins/auth-node/src/oauth/createOAuthRouteHandlers.test.ts
@@ -222,6 +222,51 @@ describe('createOAuthRouteHandlers', () => {
       );
     });
 
+    it('should not merge stale granted-scope cookie into authorization request', async () => {
+      const agent = request.agent(
+        wrapInApp(
+          createOAuthRouteHandlers({
+            ...baseConfig,
+            authenticator: {
+              ...mockAuthenticator,
+              shouldPersistScopes: true,
+            },
+          }),
+        ),
+      );
+
+      agent.jar.setCookie(
+        'my-provider-granted-scope=stale-scope',
+        '127.0.0.1',
+        '/my-provider',
+      );
+
+      mockAuthenticator.start.mockResolvedValue({
+        url: 'https://example.com/redirect',
+      });
+
+      const res = await agent.get('/my-provider/start').query({
+        env: 'development',
+        scope: 'my-scope',
+      });
+
+      const { value: nonce } = getNonceCookie(agent);
+
+      expect(res.status).toBe(302);
+      expect(mockAuthenticator.start).toHaveBeenCalledWith(
+        {
+          req: expect.anything(),
+          scope: 'my-scope',
+          state: encodeOAuthState({
+            nonce: decodeURIComponent(nonce),
+            env: 'development',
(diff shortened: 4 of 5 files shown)


Rugvip (changes requested):
Thank you for looking into this! 👍 

I think we need to explore other options. Including the already granted scopes is something that's required for a lot of providers and I think this will break more than it fixes unfortunately. I might be wrong though, but that'd require a bit more digging to prove that ;)

Some other potential options could be some way to more explicitly create a completely fresh session, maybe specifically if a previous attempt fails. There could also something built into the scope transforms that aim to remove invalid scopes, although that's ofc kinda tricky, but could also be reserved for providers where requesting invalid scopes leads to an error.

walsm232:
Thanks @Rugvip for taking a look and for your review + feedback! Ah so it sounds like this might be a case of _"it looks like it would be a small fix but then it's not"_ ;)

I dug a bit deeper into this:
- I re-checked the flow and provider usage, and I think there are two competing concerns:
  - **Real bug case:** stale scopes in `<provider>-granted-scope` can poison a new `/start` request and cause `invalid_scope` when provider config changes.
  - **Compatibility concern:** `plugin-auth-node` previously documented that persisted providers merge requested + granted even for full auth flow, so changing `/start` to ignore granted scopes may break expected behaviour for some adopters/custom providers, does that sound correct?

From what I understand (and I think you might be able to help confirm some of this as I'm looking at it for the first time):
- The change here only affects `/start`, `/refresh` still merges granted scopes and still uses `scopeAlreadyGranted`.
- In Backstage’s built-in OAuth providers, only `github`, `bitbucket`, and `oidc` set `scopes.persist: true`.
- Google’s `includeGrantedScopes` is separate/provider-side and not tied to Backstage’s persisted-scope cookie.

So it sort of feels like the safest conclusion is:
- My current patch fixes a legitimate stale-cookie failure mode.
- But it may also be a behaviour regression vs. existing documented `scopes.persist` semantics, especially for external/custom providers right?

and based on what you've said it sounds like the right direction / approach would be one of:
- Explicit fresh auth path (opt-in) that skips persisted granted scope or
- Failure-triggered fallback: if `/start` fails with provider `invalid_scope`, clear granted-scope cookie and retry once or
- Config/option-controlled behaviour so existing adopters can keep current merge semantics.

If you’re open to it, I can pivot the PR in that direction instead of changing default `/start` behaviour globally. Do you think that one of these would be a good / suitable option for handling this?

walsm232:
> There could also something built into the scope transforms that aim to remove invalid scopes

I've thought about this a bit and I wonder if we can add something like a `newAuthorization` boolean (or something similar) to the scope transform options. This could be `true` when resolving scopes for a new `/start` request and false on `/refresh`. The granted scopes from the cookie are still passed to the transform in both cases but the flag gives the transform context to decide whether to include them.

In this scenario the default transform is unchanged (it still merges granted scopes on both `/start` and `/refresh`, so any existing usage is unaffected).

To handle this scenario, it could be opted-in by providing a custom transform that skips granted scopes on new authorizations:

```ts
scopes: {
  persist: true,
  required: ['openid', 'profile'],
  transform({ requested, granted, required, additional, newAuthorization }) {
    if (newAuthorization) {
      return [...requested, ...required, ...additional];
    }
    return [...requested, ...granted, ...required, ...additional];
  },
},
```

I guess that the trade-off is that the stale-scope issue isn't solved by default and it requires opt-in. But it would avoid any risk of breaking existing scope accumulation behaviour and gives control over how granted scopes are handled in each flow I think.

Curious to know what you think!

Rugvip:
Hmm, I'm thinking something about handling the `invalid_scope` error might be the best approach, possibly withing the providers themselves because I don't trust them to all have the same error handling 😅. Maybe both standard + enabling custom error handling is the safest.

The tricky bit about clearing scopes on new sessions is that you might have distinct scopes at different parts of the app. One page might want scope A only, and another one wants scope B. Navigating back and forth between those pages you'd then be asked to re-authenticate every single time.

PR #33658, opened 2026-03-28
Add FetchMiddlewares.clarifyFailures and improve permission error handling

freben:
## Summary

- Added `FetchMiddlewares.clarifyFailures()` to `@backstage/core-app-api`, which replaces the generic "TypeError: Failed to fetch" with a message including the request method and URL (e.g. "Failed to fetch: GET https://example.com/api/catalog")
- Wired it into the default fetch middleware stack in `@backstage/app-defaults`
- Wrapped extension permission authorization in `@backstage/frontend-app-api` with a try/catch that surfaces errors as `ForwardedError`

### DIFF
--- a/packages/core-app-api/src/apis/implementations/FetchApi/ClarifyFailuresFetchMiddleware.ts
+++ b/packages/core-app-api/src/apis/implementations/FetchApi/ClarifyFailuresFetchMiddleware.ts
@@ -29,22 +29,36 @@ export class ClarifyFailuresFetchMiddleware implements FetchMiddleware {
         return await next(input as any, init);
       } catch (e) {
         if (e instanceof TypeError && e.message === 'Failed to fetch') {
-          let method: string;
-          let url: string;
+          try {
+            let method: string;
+            let url: string;
 
-          if (input instanceof Request) {
-            method = input.method;
-            url = input.url;
-          } else {
-            method = init?.method || 'GET';
-            url = String(input);
+            if (isRequestLike(input)) {
+              method = input.method;
+              url = input.url;
+            } else {
+              method = init?.method || 'GET';
+              url = String(input);
+            }
+
+            e.message = `Failed to fetch: ${method} ${url}`;
+          } catch {
+            // intentionally ignored - clarification is best-effort
           }
-
-          e.message = `Failed to fetch: ${method} ${url}`;
-          throw e;
         }
         throw e;
       }
     };
   }
+}
+
+function isRequestLike(input: unknown): input is Request {
+  return (
+    input !== null &&
+    typeof input === 'object' &&
+    'method' in input &&
+    'url' in input &&
+    typeof input.method === 'string' &&
+    typeof input.url === 'string'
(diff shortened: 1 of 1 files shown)


packages/core-app-api/src/apis/implementations/FetchApi/ClarifyFailuresFetchMiddleware.ts
```diff
@@ -0,0 +1,50 @@
...
+      try {
+        // NOTE: The "as any" cast is because of subtle undici type differences
+        // that happened in a node types bump. Immaterial at runtime.
+        return await next(input as any, init);
+      } catch (e) {
+        if (e instanceof TypeError && e.message === 'Failed to fetch') {
+          let method: string;
+          let url: string;
+
+          if (input instanceof Request) {
```
Rugvip: `instanceof` is a bit sketchy due to the risk of duplication from other contexts. Might be better to check for those fields more directly instead.

packages/core-app-api/src/apis/implementations/FetchApi/ClarifyFailuresFetchMiddleware.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+export class ClarifyFailuresFetchMiddleware implements FetchMiddleware {
+  apply(next: typeof fetch): typeof fetch {
+    return async (input, init) => {
+      try {
+        // NOTE: The "as any" cast is because of subtle undici type differences
+        // that happened in a node types bump. Immaterial at runtime.
+        return await next(input as any, init);
+      } catch (e) {
+        if (e instanceof TypeError && e.message === 'Failed to fetch') {
```
Rugvip, line 31: Wondering if this should go even a bit further, potentially using ResponseError too? 🤔

packages/core-app-api/src/apis/implementations/FetchApi/ClarifyFailuresFetchMiddleware.ts
freben: Ah right, that one is not a builtin like TypeError. Fixed

packages/core-app-api/src/apis/implementations/FetchApi/ClarifyFailuresFetchMiddleware.ts
freben: Hmmmmm _maybe_, it does break a native fetch contract though - and we don't have an actual response to base it on so we would have to "fake" it by hand. Hmmmm leaving as is I think?

Rugvip: Yeah was thinking a bit more while we're at it. Prolly by adding a new option that helps convert `!ok` responses automatically since that's a wrapping you just always have to do otherwise more or less. Very different change though so let's ship this

Rugvip (approved):

PR #33660, opened 2026-03-28
POC: In-process CLI report extraction

freben:
## Summary

Alternative to #33659 — instead of parallelizing subprocess spawns, this eliminates them entirely by loading CLI modules in-process.

- Loads each `CliModule` via `require()`, builds the Commander program tree in-process, and walks it to extract help pages
- For the main `cli` package (role: `cli`), loads `@backstage/cli-defaults` to get all modules
- For `cli-module-*` packages, loads their `src/index` directly
- **~50x faster**: ~1s vs ~50s for all 14 CLI packages

### Trade-off

Leaf command options/arguments are **not visible** to this approach. Each leaf command uses `helpOption(false)` + `allowUnknownOption(true)` in Commander, and parses its own flags inside its `execute` function (often using cleye). The in-process extraction only sees what's registered in Commander, so leaf commands lose their `--format`, `--config`, etc. options in the generated reports.

This could be addressed by:
1. Having commands declare their options in the `CliCommand` type (would require changes across all commands)
2. A hybrid approach: in-process for tree structure, subprocess only for leaf commands
3. Accepting the simpler reports as sufficient

### Comparison with #33659

| | #33659 (parallel spawn) | This PR (in-process) |
|---|---|---|
| Speed | ~12s (4x faster) | ~1s (50x faster) |
| Report fidelity | Identical to current | Missing leaf command options |
| Complexity | Low (async spawn + p-limit) | Medium (loads internal CLI APIs) |
| New dependencies | None | `@internal/cli`, `@backstage/cli-defaults` |

### DIFF
--- a/packages/cli-module-actions/cli-report.md
+++ b/packages/cli-module-actions/cli-report.md
@@ -5,10 +5,9 @@
 ### `backstage-cli-module-actions`
 
 ```
-Usage: @backstage/cli-module-actions [options] [command]
+Usage: backstage-cli-module-actions [options] [command]
 
 Options:
-  -V, --version
   -h, --help
 
 Commands:
@@ -19,7 +18,7 @@ Commands:
 ### `backstage-cli-module-actions actions`
 
 ```
-Usage: @backstage/cli-module-actions actions [options] [command] [command]
+Usage: backstage-cli-module-actions actions [options] [command] [command]
 
 Options:
   -h, --help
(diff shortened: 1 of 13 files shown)


freben:
Ye that's what I meant that this is only vaguely illustrating the direction. Calling it with that explicit arg sounds good, I can explore that

freben:
Dude, it got VERY hairy because of deep things inside how these CLIs execute. Somewhere in the guts of execution there's some exit-x package or whatnot that messes with the whole exit handling and eats streams and shit. This all turned out to be not worth spending time on now since in-process handling had so many tricky side effects.

PR #33663, opened 2026-03-29
Add catalog model layer system with JSON Schema based kind declarations

freben:

Adds a catalog model extensibility system that allows plugins to declare and extend catalog entity kinds, annotations, labels, tags, and relations using composable JSON Schema based layers.

### Key changes:

- **`@backstage/catalog-model`**: New `model/` module with a layer builder API (`createCatalogModelLayer`) for declaring kinds, annotations, labels, tags, and relations. Includes JSON Schema validation, merging, and compilation via `compileCatalogModel`. The compiled `CatalogModel` provides `listKinds()`, `listRelations()`, `getMetadata()` for summaries, and `getKind()`/`getRelations()` for detailed lookups. Built-in entity kinds (Component, API, Group, etc.) now include model layer definitions.
- **`@backstage/plugin-catalog-node`**: New `provideStaticCatalogModel` helper that provides a static catalog model layer at startup.
- **`@backstage/plugin-catalog-backend`**: New `ModelProcessor` that validates entities against compiled model schemas, emits relations from declared relation fields, and a `SchemaValidator` utility. New `createGetCatalogModelDescriptionAction` that returns a markdown-formatted description of the catalog model (useful for LLM integrations). Integrated into `CatalogBuilder` and `CatalogPlugin`.
- **`@backstage/plugin-scaffolder-common`**: Template kind model layer definition, serving as an example of how plugins can declare their own catalog entity kinds.
- **`@backstage/plugin-scaffolder-backend`**: Integrates the Template kind model layer with the catalog model system.

### DIFF
--- a/.changeset/scaffolder-backend-model-registration.md
+++ b/.changeset/scaffolder-backend-model-registration.md
@@ -1,5 +1,0 @@
----
-'@backstage/plugin-scaffolder-backend': patch
----
-
-Integrated the Template kind model layer with the catalog model registry service.
--- a/packages/catalog-model/src/kinds/ApiEntityV1alpha1.ts
+++ b/packages/catalog-model/src/kinds/ApiEntityV1alpha1.ts
@@ -16,7 +16,7 @@
 
 import { createCatalogModelLayer } from '../model/createCatalogModelLayer';
 import type { Entity } from '../entity/Entity';
-import schema from '../schema/kinds/API.v1alpha1.schema.json';
+import jsonSchema from '../schema/kinds/API.v1alpha1.schema.json';
 import { ajvCompiledJsonSchemaValidator } from './util';
 
 /**
@@ -46,7 +46,7 @@ export interface ApiEntityV1alpha1 extends Entity {
  * @public
  */
 export const apiEntityV1alpha1Validator =
-  ajvCompiledJsonSchemaValidator(schema);
+  ajvCompiledJsonSchemaValidator(jsonSchema);
 
 /**
  * Extends the catalog model with the API kind.
@@ -84,7 +84,7 @@ export const apiEntityModel = createCatalogModelLayer({
             },
           ],
           schema: {
-            jsonSchema: schema,
+            jsonSchema,
           },
         },
       ],
--- a/packages/catalog-model/src/kinds/DomainEntityV1alpha1.ts
+++ b/packages/catalog-model/src/kinds/DomainEntityV1alpha1.ts
@@ -16,7 +16,7 @@
 
 import { createCatalogModelLayer } from '../model/createCatalogModelLayer';
 import type { Entity } from '../entity/Entity';
-import schema from '../schema/kinds/Domain.v1alpha1.schema.json';
+import jsonSchema from '../schema/kinds/Domain.v1alpha1.schema.json';
 import { ajvCompiledJsonSchemaValidator } from './util';
 
 /**
@@ -44,7 +44,7 @@ export interface DomainEntityV1alpha1 extends Entity {
  * @public
  */
 export const domainEntityV1alpha1Validator =
-  ajvCompiledJsonSchemaValidator(schema);
+  ajvCompiledJsonSchemaValidator(jsonSchema);
 
 /**
  * Extends the catalog model with the Domain kind.
@@ -82,7 +82,7 @@ export const domainEntityModel = createCatalogModelLayer({
             },
           ],
           schema: {
-            jsonSchema: schema,
+            jsonSchema,
           },
         },
       ],
--- a/packages/catalog-model/src/kinds/ComponentEntityV1alpha1.ts
+++ b/packages/catalog-model/src/kinds/ComponentEntityV1alpha1.ts
@@ -16,7 +16,7 @@
 
 import { createCatalogModelLayer } from '../model/createCatalogModelLayer';
 import type { Entity } from '../entity/Entity';
-import schema from '../schema/kinds/Component.v1alpha1.schema.json';
+import jsonSchema from '../schema/kinds/Component.v1alpha1.schema.json';
 import { ajvCompiledJsonSchemaValidator } from './util';
 
 /**
@@ -50,7 +50,7 @@ export interface ComponentEntityV1alpha1 extends Entity {
  * @public
  */
 export const componentEntityV1alpha1Validator =
-  ajvCompiledJsonSchemaValidator(schema);
+  ajvCompiledJsonSchemaValidator(jsonSchema);
 
 /**
  * Extends the catalog model with the Component kind.
@@ -116,7 +116,7 @@ export const componentEntityModel = createCatalogModelLayer({
             },
           ],
           schema: {
-            jsonSchema: schema,
+            jsonSchema,
           },
         },
       ],
--- a/packages/catalog-model/src/kinds/GroupEntityV1alpha1.ts
+++ b/packages/catalog-model/src/kinds/GroupEntityV1alpha1.ts
@@ -16,7 +16,7 @@
 
 import { createCatalogModelLayer } from '../model/createCatalogModelLayer';
 import type { Entity } from '../entity/Entity';
-import schema from '../schema/kinds/Group.v1alpha1.schema.json';
+import jsonSchema from '../schema/kinds/Group.v1alpha1.schema.json';
 import { ajvCompiledJsonSchemaValidator } from './util';
 
 /**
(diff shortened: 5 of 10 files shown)


packages/catalog-model/src/model/sources/CatalogModelSources.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Provides a static catalog model on top of the default one (which is
+   * included automatically). User-provided layers take precedence over the
+   * default model when layer IDs overlap.
+   */
+  static static(layers: CatalogModelLayer[]): CatalogModelSource {
+    return new StaticCatalogModelSource(
+      uniqBy([...layers, defaultCatalogEntityModel], 'layerId'),
+    );
```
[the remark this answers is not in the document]
freben: It no longer throws.

packages/catalog-model/src/kinds/annotations.ts
```diff
@@ -0,0 +1,59 @@
...
+import { createCatalogModelLayer } from '../model/createCatalogModelLayer';
+import { ANNOTATION_EDIT_URL, ANNOTATION_VIEW_URL } from '../entity';
+import {
+  ANNOTATION_LOCATION,
+  ANNOTATION_ORIGIN_LOCATION,
+  ANNOTATION_SOURCE_LOCATION,
+} from '../location';
+
+export const wellKnownAnnotationsModel = createCatalogModelLayer({
+  layerId: 'catalog.backstage.io/well-known-annotations',
```
Rugvip, line 26: 👌

packages/catalog-model/src/kinds/ApiEntityV1alpha1.ts
```diff
@@ -46,3 +47,47 @@ export interface ApiEntityV1alpha1 extends Entity {
...
+      },
+      description:
+        'An API describes an interface that can be exposed by a component.',
+      versions: [
+        {
+          name: ['v1alpha1', 'v1beta1'],
+          relationFields: [
+            {
+              selector: { path: 'spec.owner' },
+              relation: 'ownedBy',
```
Rugvip, line 74: What happens if it doesn't exist? Mostly curious

packages/catalog-model/src/kinds/DomainEntityV1alpha1.ts
```diff
@@ -44,3 +45,47 @@ export interface DomainEntityV1alpha1 extends Entity {
...
+            },
+            {
+              selector: { path: 'spec.subdomainOf' },
+              relation: 'partOf',
+              defaultKind: 'Domain',
+              defaultNamespace: 'inherit',
+            },
+          ],
+          schema: {
+            jsonSchema: schema,
```
benjdlambert: Small nit, but not important.

I would prefer if the import was called `jsonSchema`.


```suggestion
            jsonSchema,
```

packages/catalog-model/src/model/jsonSchema/validateMetaSchema.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Validates that the schema is valid according to known JSON Schema meta
+ * schemas - does not perform any semantic validation beyond that.
+ */
+export function validateMetaSchema(schema: unknown): schema is SchemaObject {
+  if (!isJsonObject(schema) || typeof schema.then === 'function') {
+    throw new InputError('Invalid JSON schema: must be an object');
+  }
+
+  const ajv = getAjv();
+  try {
```
benjdlambert, line 32: kinda wanna get rid of `ajv` in the future, but this is fine for now as we use it in a lot of places. Wanna maybe replace it with something like https://github.com/cfworker/cfworker/tree/main/packages/json-schema instead. Less `eval`y.

packages/catalog-model/src/model/modelActions/addAnnotation.ts
```diff
@@ -0,0 +1,79 @@
...
+    jsonSchema: JsonObject;
+  };
+}
+
+export function opsFromCatalogModelAnnotation(
+  annotation: CatalogModelAnnotationDefinition,
+): CatalogModelOp[] {
+  if (annotation.schema?.jsonSchema) {
+    validateMetaSchema(annotation.schema.jsonSchema);
+    if (annotation.schema.jsonSchema.type !== 'string') {
```
Rugvip, line 63: Weird edge-case to potentially allow: `anyOf: [{ type: string }, { type: string }]`. I say leave for later though

packages/catalog-model/src/model/modelActions/removeKind.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * The definition of a kind removal from the catalog model.
+ *
+ * @alpha
+ */
+export interface CatalogModelRemoveKindDefinition {
+  /**
+   * The kind to remove, e.g. "Component".
+   */
+  kind: string;
```
Rugvip, line 29: Potentially the ability to remove a specific version? 🤔

.changeset/scaffolder-backend-model-registration.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend': patch
```
benjdlambert: Don't think we need this changeset?

plugins/catalog-backend/src/actions/createGetCatalogModelDescriptionAction.ts
```diff
@@ -0,0 +1,197 @@
...
+  actionsRegistry,
+}: {
+  modelHolder: ModelHolder | undefined;
+  actionsRegistry: ActionsRegistryService;
+}) => {
+  actionsRegistry.register({
+    name: 'get-catalog-model-description',
+    title: 'Get a Catalog Model Description',
+    description:
+      'Returns a markdown formatted description of the current catalog model, including all registered entity kinds, their spec fields, and available relations.',
```
benjdlambert: If we're not gonna include spec fields for now then we should update this.

benjdlambert: Probably do include `annotations` though.

packages/catalog-model/src/model/defaultCatalogEntityModel.ts
```diff
@@ -0,0 +1,49 @@
...
+import { wellKnownRelationsModel } from '../kinds/relations';
+import { createCatalogModelLayer } from './createCatalogModelLayer';
+
+/**
+ * The default catalog entity model, containing all built-in Backstage entity
+ * kinds, relations, and annotations.
+ *
+ * @alpha
+ */
+export const defaultCatalogEntityModel = createCatalogModelLayer({
```
Rugvip, line 35: ah, hadn't expected aggregation at this level, but doesn't hurt

packages/catalog-model/src/kinds/DomainEntityV1alpha1.ts
freben: updating!

packages/catalog-model/src/model/jsonSchema/validateMetaSchema.ts
freben: agreed, just did this basically to not mess with the performance profile, but we should definitely evaluate that as a better alternative if it isn't wildly slow

.changeset/scaffolder-backend-model-registration.md
freben: correct, deleting; this was a remnant from when i had the catalogModelRegistryService in place. Deleted it since it was a bit premature and just caused confusion

plugins/catalog-backend/src/actions/createGetCatalogModelDescriptionAction.ts
freben: Correct, updating

packages/catalog-model/src/model/modelActions/removeKind.ts
freben: Good call, yes. As a matter of fact the high level `updateKind` may be a bit insufficient here ... unless you can do like

```ts
model.updateKind({
  names: {
    kind: 'Component'
  },
  versions: [
    { name: 'v1beta1', schema: false }
```

or something like that. Feels wonky.

packages/catalog-model/src/kinds/ApiEntityV1alpha1.ts
freben: It emits the forward one, but skips the reverse. So what you specify here does get honored.

packages/catalog-model/src/model/defaultCatalogEntityModel.ts
freben: Yeh it felt useful (especially in this large model) to _somehow_ be able to break apart into separate declarations. It becomes unwieldy otherwise.

But as we discussed, when doing aggregation we don't want to lose track of origins of individual things either. But we have room to work with that inside the opaque type if needed.

PR #33680, opened 2026-03-30
feat(backend-defaults): add AWS RDS IAM authentication support for PostgreSQL

### LINKED ISSUE #16980: 🚀 Feature: Add IAM authentication support for PostgresDB Aurora Serverless
### 🔖 Feature description

aws Aurora Serverless supports IAM authentication  via roles and it would be nice to extend backstage to allow authenticating optionally with IAM roles towards an aurora based postgres DB. Additionally, the configuration can also take the access key ID and secret as an input, these would be generated by IRSA and injected in runtime for k8s based workloads.

reference: https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.html

related: https://github.com/backstage/backstage/issues/12844

### 🎤 Context

At the moment backstage supports only password authentication (afaik). 

_No response_

rolandfuszenecker-seon:
Backstage supports Azure (`type: azure`) and Google Cloud SQL (`type: cloudsql`) IAM authentication for PostgreSQL out of the box, but AWS RDS IAM authentication was missing. This adds `type: rds` support following the exact same pattern as the existing Azure implementation — a knex connection factory function that calls `@aws-sdk/rds-signer` to generate a short-lived IAM token on every new connection, so tokens auto-refresh before the 15-minute expiry.

I tested this against a real AWS RDS PostgreSQL instance using assumed-role credentials and confirmed all Backstage plugins connect and run migrations successfully.

Closes  #16980

- [ ] Added or updated documentation — config schema updated in config.d.ts
- [ ] Screenshots attached — not applicable, backend-only change

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -538,53 +538,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@aws-sdk/client-cognito-identity@npm:3.1020.0":
-  version: 3.1020.0
-  resolution: "@aws-sdk/client-cognito-identity@npm:3.1020.0"
-  dependencies:
-    "@aws-crypto/sha256-browser": "npm:5.2.0"
-    "@aws-crypto/sha256-js": "npm:5.2.0"
-    "@aws-sdk/core": "npm:^3.973.26"
-    "@aws-sdk/credential-provider-node": "npm:^3.972.28"
-    "@aws-sdk/middleware-host-header": "npm:^3.972.8"
-    "@aws-sdk/middleware-logger": "npm:^3.972.8"
-    "@aws-sdk/middleware-recursion-detection": "npm:^3.972.9"
-    "@aws-sdk/middleware-user-agent": "npm:^3.972.27"
-    "@aws-sdk/region-config-resolver": "npm:^3.972.10"
-    "@aws-sdk/types": "npm:^3.973.6"
-    "@aws-sdk/util-endpoints": "npm:^3.996.5"
-    "@aws-sdk/util-user-agent-browser": "npm:^3.972.8"
-    "@aws-sdk/util-user-agent-node": "npm:^3.973.13"
-    "@smithy/config-resolver": "npm:^4.4.13"
-    "@smithy/core": "npm:^3.23.13"
-    "@smithy/fetch-http-handler": "npm:^5.3.15"
-    "@smithy/hash-node": "npm:^4.2.12"
-    "@smithy/invalid-dependency": "npm:^4.2.12"
-    "@smithy/middleware-content-length": "npm:^4.2.12"
-    "@smithy/middleware-endpoint": "npm:^4.4.28"
-    "@smithy/middleware-retry": "npm:^4.4.45"
-    "@smithy/middleware-serde": "npm:^4.2.16"
-    "@smithy/middleware-stack": "npm:^4.2.12"
-    "@smithy/node-config-provider": "npm:^4.3.12"
-    "@smithy/node-http-handler": "npm:^4.5.1"
-    "@smithy/protocol-http": "npm:^5.3.12"
-    "@smithy/smithy-client": "npm:^4.12.8"
-    "@smithy/types": "npm:^4.13.1"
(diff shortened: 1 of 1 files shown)


packages/backend-defaults/config.d.ts
```diff
@@ -632,6 +632,35 @@ export interface Config {
...
+             * Requires the `@aws-sdk/rds-signer` package to be installed.
+             * The IAM role or user must have the `rds-db:connect` permission for the database user.
+             */
+            type: 'rds';
+            /**
+             * The hostname of the RDS instance.
+             */
+            host: string;
+            /**
+             * The port number the database is listening on. Defaults to 5432.
```
freben: Does that default make sense, given that it only matches one possible RDS type? Should it be dependent on the `client` value instead, or autodetect based on the remote somehow, or be a required field?

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -271,6 +273,56 @@ export async function buildCloudSqlConfig(
   };
 }
 
+export async function buildRdsPgConfig(config: Config): Promise<Knex.Config> {
+  const { Signer } =
+    require('@aws-sdk/rds-signer') as typeof import('@aws-sdk/rds-signer');
+
+  const rawConfig = config.get() as Record<string, unknown>;
+  const normalized = normalizeConnection(rawConfig.connection as any);
```
freben: Note that `normalizeConnection` is a bit unfortunately named; it is specialised for PG connection strings only. I'm not sure it's safe to call unconditionally.

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -271,6 +273,56 @@ export async function buildCloudSqlConfig(
   };
 }
 
+export async function buildRdsPgConfig(config: Config): Promise<Knex.Config> {
+  const { Signer } =
+    require('@aws-sdk/rds-signer') as typeof import('@aws-sdk/rds-signer');
+
+  const rawConfig = config.get() as Record<string, unknown>;
```
freben: We try to avoid `.get` / `.getOptional` as much as possible. It throws away valuable information and ruins the ability to get proper type checking.

In the code below where you access the raw config and do casting, use `config.get[Optional][String|Number|etc](key)` instead. When the user has made a mistake somewhere, the error messages will be precise and detailed and point them at the exact field and file that they need to fix.

At that point you can throw away `if (!username)` checks etc too because you just use `getString` which guarantees a value.

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
```diff
@@ -271,6 +273,56 @@ export async function buildCloudSqlConfig(
...
+    process.env.AWS_DEFAULT_REGION;
+  if (!region) {
+    throw new Error(
+      'Missing region for AWS RDS IAM auth: set connection.region or the AWS_REGION environment variable',
+    );
+  }
+
+  const signer = new Signer({ hostname, port, username, region });
+
+  async function getConnectionConfig() {
```
freben, line 316: You will need an expiration checker to be returned; otherwise the returned connection will be reused forever.

https://knexjs.org/guide/#configuration-options

Image

packages/backend-defaults/config.d.ts
rolandfuszenecker-seon: At first from backstage docs my understanding was that backstage is only compatible with postgres, that's why I set it to default and kept the possibility to override it from the config file if someone using different port. Thanks for pointing it out, modifying it to make port required.

packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts
rolandfuszenecker-seon: This follows the same pattern used in buildAzurePgConfig, but I modified the other parts. I kept the region check, as it's not obvious that AWS_REGION or AWS_DEFAULT_REGION is set by default.

PR #33682, opened 2026-03-30
Fix GitHub app credentials for bare host URLs

vinzscam:

When calling GitHub `getCredentials()` with just a host URL (like `https://github.com`), the provider was falling back to a PAT instead of using the configured GitHub App. This broke endpoints like `listInstallations` that need app-level auth.

Now it generates an app-level JWT when no org/repo is in the URL.

### DIFF
--- a/packages/integration/src/github/SingleInstanceGithubCredentialsProvider.ts
+++ b/packages/integration/src/github/SingleInstanceGithubCredentialsProvider.ts
@@ -121,7 +121,7 @@ class GithubAppManager {
   }
 
   async getInstallationCredentials(
-    owner: string,
+    owner?: string,
     repo?: string,
   ): Promise<{ accessToken: string | undefined }> {
     // No owner means a bare host URL (e.g. https://github.com) — return an
@@ -267,7 +267,10 @@ export class GithubAppCredentialsMux {
     return installs.flat();
   }
 
-  async getAppToken(owner: string, repo?: string): Promise<string | undefined> {
+  async getAppToken(
+    owner?: string,
+    repo?: string,
+  ): Promise<string | undefined> {
     if (this.apps.length === 0) {
       return undefined;
     }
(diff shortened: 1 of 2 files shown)


packages/integration/src/github/SingleInstanceGithubCredentialsProvider.ts
```diff
@@ -124,6 +124,17 @@ class GithubAppManager {
     owner: string,
```
Rugvip: Think we should make it explicitly optional here as well, rather than relying on an empty string or it being undefined anyway

packages/integration/src/github/SingleInstanceGithubCredentialsProvider.ts
```diff
@@ -124,6 +124,17 @@ class GithubAppManager {
...
   ): Promise<{ accessToken: string | undefined }> {
+    // No owner means a bare host URL (e.g. https://github.com) — return an
+    // app-level JWT rather than an installation token.
+    if (!owner) {
+      const auth = createAppAuth({
+        appId: this.baseAuthConfig.appId,
+        privateKey: this.baseAuthConfig.privateKey,
+      });
+      const { token } = await auth({ type: 'app' });
+      return { accessToken: token };
```
Rugvip, line 135: Should probably cache this part as well right? 🤔 in general it seems it would be good if this could take the path as the other cases, just without an owner 🤷

vinzscam: the issue is that this particular token is valid for 10 minutes and the grace period of the existing cache is...10 minutes 😅
Since the token is generated client side I'm not sure caching adds anything here

packages/integration/src/github/SingleInstanceGithubCredentialsProvider.ts
vinzscam: makes sense

PR #33691, opened 2026-03-30
Add OpenTelemetry metrics instrumentation to events service

harleyk-gusto:

Add OpenTelemetry metrics instrumentation to `@backstage/plugin-events-node` via the `MetricsService`, recording histogram metrics for event publish and subscribe operations.

### Changes

- Record `events.publish.duration` histogram for publish operations, with attributes for topic, plugin ID, and status
- Record `events.subscribe.process.duration` histogram for individual subscriber handler invocations, with attributes for topic, subscriber ID, plugin ID, and status
- Wire `MetricsService` through `DefaultEventsService.forPlugin()` and the events service factory
- All metrics are optional — the `MetricsService` dependency is not required, preserving full backward compatibility

This resolves the existing TODO comment in the codebase: `TODO(pjungermann): add opentelemetry?`

### DIFF
--- a/plugins/events-node/src/api/DefaultEventsService.ts
+++ b/plugins/events-node/src/api/DefaultEventsService.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { performance } from 'node:perf_hooks';
 import {
   AuthService,
   DiscoveryService,
@@ -21,6 +22,10 @@ import {
   LoggerService,
   RootConfigService,
 } from '@backstage/backend-plugin-api';
+import {
+  MetricsService,
+  MetricsServiceHistogram,
+} from '@backstage/backend-plugin-api/alpha';
 import { EventParams } from './EventParams';
 import {
   EVENTS_NOTIFY_TIMEOUT_HEADER,
@@ -28,7 +33,13 @@ import {
   EventsServiceSubscribeOptions,
 } from './EventsService';
 import { DefaultApiClient } from '../generated';
-import { ResponseError } from '@backstage/errors';
+import { isError, ResponseError } from '@backstage/errors';
+
+/** @internal */
+type EventsDurationAttributes = {
+  'events.topic'?: string;
+  'error.type'?: string;
+};
 
 const POLL_BACKOFF_START_MS = 1_000;
 const POLL_BACKOFF_MAX_MS = 60_000;
@@ -121,14 +132,19 @@ class PluginEventsService implements EventsService {
   private readonly mode: EventBusMode;
   private client?: DefaultApiClient;
   private readonly auth?: AuthService;
+  private readonly publishDuration?: MetricsServiceHistogram<EventsDurationAttributes>;
+  private readonly subscribeDuration?: MetricsServiceHistogram<EventsDurationAttributes>;
+  private readonly reportTopics: boolean;
 
   constructor(
     pluginId: string,
     localBus: LocalEventBus,
     logger: LoggerService,
     mode: EventBusMode,
+    reportTopics: boolean,
     client?: DefaultApiClient,
     auth?: AuthService,
+    metrics?: MetricsService,
   ) {
     this.pluginId = pluginId;
     this.localBus = localBus;
@@ -136,6 +152,26 @@ class PluginEventsService implements EventsService {
     this.mode = mode;
     this.client = client;
     this.auth = auth;
+    this.reportTopics = reportTopics;
+
+    if (metrics) {
+      this.publishDuration = metrics.createHistogram<EventsDurationAttributes>(
+        'events.publish.duration',
+        {
+          description: 'Duration of event publish operations',
+          unit: 's',
+        },
+      );
+      this.subscribeDuration =
+        metrics.createHistogram<EventsDurationAttributes>(
+          'events.subscribe.process.duration',
+          {
+            description:
+              'Duration of individual subscriber event handler invocations',
+            unit: 's',
+          },
+        );
+    }
   }
 
   async publish(params: EventParams): Promise<void> {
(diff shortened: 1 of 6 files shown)


harleyk-gusto:
I saw the TODO comment when I was making this improvement in our own instance. searched for any existing work here and didn't find any. Hopefully this contribution is helpful.

plugins/events-node/src/api/DefaultEventsService.ts
```diff
@@ -30,6 +35,23 @@ import {
 import { DefaultApiClient } from '../generated';
 import { ResponseError } from '@backstage/errors';
 
+/** @internal */
+type EventsPublishAttributes = {
+  'events.topic': string;
+  'events.plugin_id': string;
```
freben: The plugin ID is already recorded [as part of the meter scope](https://github.com/backstage/backstage/blob/42d0a2749a3d1eb9e9364957e5a2aac52a0a87ad/packages/backend-defaults/src/alpha/entrypoints/metrics/metricsServiceFactory.ts#L41) and does not need to be an attribute.

Same for the subscribe attributes below

plugins/events-node/src/api/DefaultEventsService.ts
```diff
@@ -30,6 +35,23 @@ import {
 import { DefaultApiClient } from '../generated';
 import { ResponseError } from '@backstage/errors';
 
+/** @internal */
+type EventsPublishAttributes = {
+  'events.topic': string;
```
freben: Is it useful to record the topic? This risks becoming a _very_ high cardinality attribute in some installations which messes with things a bit. Depends a lot on the topic structure chosen.

For publishing I don't foresee the topic ever affecting the timing (they'll all be the same). For subscription processing - sure, it could vary per topic. But maybe it's enough to just have these grouped by plugin ID and then if it spikes you'll have to dig deeper by other means.

Same for the subscribe attributes below

plugins/events-node/src/api/DefaultEventsService.ts
```diff
@@ -30,6 +35,23 @@ import {
...
+  'events.plugin_id': string;
+  'events.status': 'success' | 'error';
+  'error.type'?: string;
```
freben: This is good 👍 Let's just ensure it stays low cardinality.

plugins/events-node/src/api/DefaultEventsService.ts
```diff
@@ -30,6 +35,23 @@ import {
 import { DefaultApiClient } from '../generated';
 import { ResponseError } from '@backstage/errors';
 
+/** @internal */
+type EventsPublishAttributes = {
+  'events.topic': string;
+  'events.plugin_id': string;
+  'events.status': 'success' | 'error';
```
freben: No need for this; the presence / non-presence of `error.type` is enough.

Same for the subscribe attributes below

plugins/events-node/src/api/DefaultEventsService.ts
```diff
@@ -30,6 +35,23 @@ import {
...
+  'error.type'?: string;
+};
+
+/** @internal */
+type EventsSubscribeAttributes = {
+  'events.topic': string;
+  'events.subscriber.id': string;
```
freben: Aren't these often just like a UUID? Should probably be skipped too.

plugins/events-node/src/api/DefaultEventsService.ts
```diff
@@ -129,70 +153,146 @@ class PluginEventsService implements EventsService {
...
     this.mode = mode;
     this.client = client;
     this.auth = auth;
+
+    if (metrics) {
+      this.publishDuration = metrics.createHistogram<EventsPublishAttributes>(
+        'events.publish.duration',
+        {
+          description: 'Duration of event publish operations',
+          unit: 'seconds',
```
freben: these should both be `'s'` https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units

plugins/events-node/src/api/DefaultEventsService.ts
```diff
@@ -129,70 +153,146 @@ class PluginEventsService implements EventsService {
...
+      return;
+    }
+    const attrs: EventsPublishAttributes = {
+      'events.topic': topic,
+      'events.plugin_id': this.pluginId,
+      'events.status': status,
+    };
+    if (error) {
+      attrs['error.type'] =
+        error instanceof Error ? error.name : 'UnknownError';
```
freben: there's an `isError` that is more thorough and accepts custom error types better, in `©backstage/errors`

same below

freben (changes requested):
Hey thanks! Some best-practices tweaks I think. Ping @kurtaking here too, keep me accountable, I'm no expert here :)

harleyk-gusto:
Thanks for the feedback, this simplified this quite a bit. I made topics one you could opt into through a config option as well to help for those without an unbounded number of topics.

harleyk-gusto:
Any chance this could get another review?

kurtaking:
> Hey thanks! Some best-practices tweaks I think. Ping @kurtaking here too, keep me accountable, I'm no expert here :)

Whew, my apologies. I just came across the alert for being tagged here. Taking a look.

plugins/events-node/src/service.ts
```diff
@@ -43,19 +44,21 @@ export const eventsServiceFactory = createServiceFactory({
...
-    { pluginMetadata, discovery, logger, lifecycle, auth },
+    { pluginMetadata, discovery, logger, lifecycle, auth, metrics },
     eventsService,
   ) {
     return eventsService.forPlugin(pluginMetadata.getId(), {
       discovery,
       logger,
       lifecycle,
       auth,
+      metrics,
```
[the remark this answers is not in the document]
kurtaking: You'd have to be doing something pretty unusual with a custom backend to run into this issue. The default backend implementation will wire the factory when installing `plugin-events-backend`. I don't know how much of a concern this actually is.

plugins/events-node/src/api/DefaultEventsService.ts
```diff
@@ -14,21 +14,32 @@
...
 } from './EventsService';
 import { DefaultApiClient } from '../generated';
-import { ResponseError } from '@backstage/errors';
+import { isError, ResponseError } from '@backstage/errors';
+
+/** @internal */
+type EventsDurationAttributes = {
+  'events.topic'?: string;
+  'error.type'?: string;
+};
```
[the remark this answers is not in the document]
kurtaking: @harleyk-gusto - the PR description is stale. Please consider updating it.

kurtaking (commented):
What do we think of leaning on the [Semantic conventions for messaging systems](https://opentelemetry.io/docs/specs/semconv/messaging/)? 

I'm leaning towards _no_ because Backstage's events service is an internal pub/sub, but curious what others think

PR #33696, opened 2026-03-31
fix(ui): fix Header nav hover indicator covering tab text

Naycon:

The `HeaderNav` hover indicator uses `position: absolute` and paints
on top of tab text. This works with the default BUI theme because
`--bui-bg-neutral-2` is semi-transparent, but any theme using an
opaque value for that token fully obscures the tab labels on hover.

This adds `position: relative` and `z-index: 2` to nav items (matching
the `Tabs` component pattern) so they paint above the indicators.

Also fixes a reference to the non-existent `--bui-font-family` CSS
variable — the correct token is `--bui-font-regular`.

### DIFF
--- a/packages/ui/src/components/Header/HeaderNav.module.css
+++ b/packages/ui/src/components/Header/HeaderNav.module.css
@@ -50,7 +50,7 @@
 
   .bui-HeaderNavItem,
   .bui-HeaderNavGroup {
-    font-family: var(--bui-font-family);
+    font-family: var(--bui-font-regular);
     font-size: var(--bui-font-size-3);
     font-weight: var(--bui-font-weight-regular);
     color: var(--bui-fg-secondary);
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Header/HeaderNav.module.css
```diff
@@ -61,6 +61,8 @@
     padding-inline: var(--bui-space-2);
     text-decoration: none;
     cursor: pointer;
+    position: relative;
+    z-index: 2;
```
Rugvip, line 65: Definitely needed? 🤔, thinking if we could reorder elements or smth instead

Naycon: The indicator is absolute positioned, so reordering won't be enough. It actually works as-is out of the box in Backstage, but some themes might set a non-transparent neutral color, which would cause issues. This fix was/is actually already in the Tabs component, I removed it when I created this one because I thought it would be enough without. :)

PR #33700, opened 2026-03-31
docs: Update FE migration skill

Parsifal-M:

Some of the instructions are not valid on older versions of backstage, this is an attempt to at least prompt the AI agent in the right direction before starting, i.e. "Hey, you are not on the right version to do this some of it may not work"

### DIFF
--- a/docs/.well-known/skills/plugin-full-frontend-system-migration/SKILL.md
+++ b/docs/.well-known/skills/plugin-full-frontend-system-migration/SKILL.md
@@ -9,7 +9,7 @@ This skill helps fully migrate an existing Backstage plugin from the old fronten
 
 This is the preferred approach for internal plugins that are only used in a single app, since there is no need to maintain backward compatibility. It can also be used for published plugins when you're ready to drop old system support entirely.
 
-It is highly recommended to be on Backstage version 1.49.x or above. This can be verified by looking in the `backstage.json` file in the root of the repository.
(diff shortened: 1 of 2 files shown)


Parsifal-M:
Was checking with @drodil should we use https://agentskills.io/specification#frontmatter the compatibility field instead?

docs/.well-known/skills/plugin-full-frontend-system-migration/SKILL.md
Parsifal-M, line 229: I'm also not sure about this part btw, because if you do this it removes the nice header and replaces it with the BUI one which is quite basic, and I don't know if anyone would want to do that (currently)

Like replacing this

Image

with the below for example I cannot imagine many people wanting that

Image

Rugvip: That is the idea though, to use a more minimal header with more space for content. Still something for @backstage/design-system-maintainers to consider though

Parsifal-M:
Will merge now, maybe later we can come back and discuss the compatibility frontmatter

docs/.well-known/skills/plugin-full-frontend-system-migration/SKILL.md
Parsifal-M: Of course its also a bit of personal taste 😄  but would be a shame to not be able to keep it similar

PR #33703, opened 2026-03-31
feat(auth): support provider logout redirects, implement Auth0 federated logout

jroebu14:
## Summary

- Added `OAuthAuthenticatorLogoutResult` type to `@backstage/plugin-auth-node` — the `logout` method on `OAuthAuthenticator` can now optionally return `{ logoutUrl }` to trigger a browser redirect after sign-out
- Implemented `logout` in the Auth0 authenticator that returns Auth0's `/v2/logout?federated` URL, clearing both the Auth0 session and any upstream IdP session
- Updated `DefaultAuthConnector.removeSession()` in `@backstage/core-app-api` to check for a `logoutUrl` in the logout response and redirect the browser

## Problem

After signing out of Backstage, clicking "Sign in" does not require re-authentication when using Auth0. This happens because:

1. The Auth0 authenticator has no `logout` method — the backend only clears the Backstage refresh token cookie
2. The Auth0 session cookie persists in the browser
3. For SSO users, the upstream IdP session also persists

Users can sign back in without entering credentials (email/password) or re-authenticating with their identity provider.

## Solution

Three backward-compatible changes across the auth stack:

**`@backstage/plugin-auth-node`** (patch) — The `OAuthAuthenticator.logout()` return type changes from `Promise<void>` to `Promise<void | OAuthAuthenticatorLogoutResult>`. The route handler returns the URL as JSON if provided, otherwise sends the existing empty 200. Providers that don't implement `logout` (or return `void`) are completely unaffected.

**`@backstage/plugin-auth-backend-module-auth0-provider`** (minor) — Adds a `logout` method that constructs Auth0's `/v2/logout?federated` URL from existing config (`domain`, `clientId`). The `federated` parameter tells Auth0 to also terminate the upstream IdP session.

**`@backstage/core-app-api`** (patch) — `DefaultAuthConnector.removeSession()` checks if the logout response contains JSON with a `logoutUrl`. If present, redirects the browser. If not (empty 200), behaves exactly as before.

## Extensibility

The framework changes are generic — any OAuth provider (Google, Okta, GitHub, etc.) can adopt the same pattern by adding a `logout` method that returns `{ logoutUrl }`. This PR only implements it for Auth0.

## Auth0 Configuration

The `returnTo` URL in the logout redirect must be whitelisted in the Auth0 tenant's **Allowed Logout URLs** setting (Application → Settings). Without this, Auth0 will reject the redirect.

### DIFF
--- a/plugins/auth-backend-module-auth0-provider/src/authenticator.ts
+++ b/plugins/auth-backend-module-auth0-provider/src/authenticator.ts
@@ -126,15 +126,15 @@ export const auth0Authenticator = createOAuthAuthenticator({
   },
 
   async logout(input, { domain, clientID, federated }) {
+    const logoutUrl = new URL(`https://${domain}/v2/logout`);
+    if (federated) {
+      logoutUrl.searchParams.set('federated', '');
+    }
+    logoutUrl.searchParams.set('client_id', clientID);
     const origin = input.req.get('origin');
-    const federatedParam = federated ? 'federated&' : '';
-    const returnToParam = origin
-      ? `&returnTo=${encodeURIComponent(origin)}`
-      : '';
-    return {
-      logoutUrl: `https://${domain}/v2/logout?${federatedParam}client_id=${encodeURIComponent(
-        clientID,
-      )}${returnToParam}`,
-    };
+    if (origin) {
+      logoutUrl.searchParams.set('returnTo', origin);
+    }
+    return { logoutUrl: logoutUrl.toString() };
   },
 });


packages/core-app-api/src/lib/AuthConnector/DefaultAuthConnector.ts
```diff
@@ -194,6 +194,28 @@ export class DefaultAuthConnector<AuthSession>
...
+
+    // If the auth provider returned a logout URL (e.g. for Auth0 federated
+    // logout), redirect the browser to clear the provider's session cookies.
+    try {
+      const contentType = res.headers.get('content-type');
+      if (contentType?.includes('application/json')) {
+        const body = await res.json();
+        if (body.logoutUrl) {
+          const url = new URL(body.logoutUrl);
+          if (url.protocol === 'https:' || url.hostname === 'localhost') {
```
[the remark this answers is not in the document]
Rugvip: Bit curious why this check is here at all actually? 🤔

plugins/auth-node/report.api.md
```diff
@@ -329,6 +332,11 @@ export interface OAuthAuthenticatorLogoutInput {
   req: Request_2;
 }
 
+// @public (undocumented)
+export interface OAuthAuthenticatorLogoutResult {
+  logoutUrl?: string;
```
Rugvip, line 337: Wonder if this should be called `redirectUrl` instead? 🤔 No strong opinion at this point though

plugins/auth-backend-module-auth0-provider/src/authenticator.ts
```diff
@@ -115,4 +115,13 @@ export const auth0Authenticator = createOAuthAuthenticator({
   async refresh(input, { helper }) {
     return helper.refresh(input);
   },
+
+  async logout(input, { domain, clientID }) {
+    const origin = input.req.get('origin') ?? '';
```
Rugvip: I could imagine other return URLs being desirable too, but I think we can leave that as a potential future improvement

plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
```diff
@@ -280,18 +284,31 @@ export function createOAuthRouteHandlers<TProfile>(
         throw new AuthenticationError('Invalid X-Requested-With header');
       }
 
+      const origin = req.get('origin');
+      if (origin && !isOriginAllowed(origin)) {
+        throw new NotAllowedError(`Origin '${origin}' is not allowed`);
+      }
```
Rugvip, line 290: Not sure why we'd need to check the origin here really? 🤔, at the same time it shouldn't hurt

.changeset/core-app-api-logout-redirect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/core-app-api': patch
```
Rugvip, line 2: Could you add `@backstage/plugin-app` here as well? I know it's a bit weird but there are some internal imports that make it make sense :p

.changeset/core-app-api-logout-redirect.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+The `DefaultAuthConnector` now checks for a `logoutUrl` in the logout response body. If the auth provider returns one (e.g. Auth0 federated logout), the browser is redirected to that URL to clear the provider's session cookies. This is backward compatible — providers that return an empty response are unaffected.
```
Rugvip: `DefaultAuthConnector` isn't an external export, so avoid mentioning it in the changeset. Can keep it more general like "the default auth implementation"

.changeset/auth-node-logout-result.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-auth-node': minor
```
Rugvip: not a breaking change

plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
UsainBloot: Consistency really as its called in start

packages/core-app-api/src/lib/AuthConnector/DefaultAuthConnector.ts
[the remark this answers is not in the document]
UsainBloot: Prevents the logout url being hijacked as a `javascript:` url and invoked in the browser.

plugins/auth-node/report.api.md
jroebu14: Went with `logoutUrl` since it describes the intent of ending the provider session rather than a generic redirect. Happy to rename if you prefer though\!

plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts
Rugvip: Yep let's go with the cautious route and lift it in the future if needed

PR #33713, opened 2026-04-01
verify-links: catch broken anchors, directory links, and invisible characters

Rugvip:

Follow-up to #33704 — that PR fixed ~30 broken links and anchors that CI didn't catch because `verify-links.js` was only checking whether link targets existed as files, completely ignoring anchors, directory links, and invisible characters in URLs.

This enhances the link verification script to catch those categories of issues as errors:

- **Broken anchors** — extracts heading slugs from target markdown files and verifies that `#anchor` references actually resolve, both for same-file references (e.g. `#some-heading`) and cross-file references (e.g. `./other.md#some-heading`)
- **Duplicate heading deduplication** — handles GitHub/Docusaurus behavior of appending `-1`, `-2`, etc. to duplicate heading slugs
- **Explicit heading IDs** — supports `{#custom-id}` syntax used by Docusaurus
- **Directory links** — flags relative links within `docs/` that point to directories instead of files (should use `index.md` suffix)
- **Invisible characters** — detects zero-width spaces and other invisible Unicode characters in URLs
- **Code block stripping** — skips links inside fenced code blocks to avoid false positives

Also fixes a handful of newly-caught broken links in existing docs (broken anchors in glossary, BEP, and REVIEWING.md, plus a directory link in the frontend system docs).

- [x] All your commits have a `Signed-off-by` line in the message. ([more info](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md#creating-changesets))

### DIFF
--- a/scripts/verify-links.js
+++ b/scripts/verify-links.js
@@ -36,17 +36,24 @@ function headingToSlug(headingText) {
     return explicitId[1];
   }
 
+  let slug = headingText
+    .toLowerCase()
+    // Remove inline code backticks
+    .replace(/`/g, '')
+    // Remove markdown bold/italic markers
+    .replace(/[*_]/g, '')
+    // Remove markdown links, keep link text
+    .replace(/\[([^\]]*)\]\([^)]*\)/g, '$1');
+
+  // Remove HTML tags in a loop to handle nested fragments like <scr<script>ipt>
+  let previous;
+  do {
+    previous = slug;
+    slug = slug.replace(/<[^>]+>/g, '');
+  } while (slug !== previous);
+
   return (
-    headingText
-      .toLowerCase()
-      // Remove inline code backticks
-      .replace(/`/g, '')
-      // Remove markdown bold/italic markers
-      .replace(/[*_]/g, '')
-      // Remove markdown links, keep link text
-      .replace(/\[([^\]]*)\]\([^)]*\)/g, '$1')
-      // Remove HTML tags
-      .replace(/<[^>]+>/g, '')
+    slug
       // Replace special characters with hyphens (keeping alphanumeric, hyphens, spaces)
       .replace(/[^\w\s-]/g, '')
       .trim()


scripts/verify-links.js
```diff
@@ -19,9 +19,94 @@
...
+      // Replace special characters with hyphens (keeping alphanumeric, hyphens, spaces)
+      .replace(/[^\w\s-]/g, '')
+      .trim()
+      .replace(/\s+/g, '-')
+  );
+}
+
+// Extracts all heading anchors from a markdown file's content,
+// handling duplicate headings with -1, -2, etc. suffixes (GitHub/Docusaurus behavior)
+function extractHeadingAnchors(content) {
```
aramissennyeydd, line 66: Does Docusaurus offer this validation already? Could we block on it there instead?

scripts/verify-links.js
```diff
@@ -40,15 +125,23 @@ async function listFiles(dir) {
...
-const projectRoot = resolvePath(__dirname, '..');
-
 async function verifyUrl(basePath, absUrl, docPages) {
-  const url = absUrl
-    .replace(/#.*$/, '')
-    .replace(
-      /https:\/\/github.com\/backstage\/backstage\/(tree|blob)\/master/,
-      '',
+  // Check for invisible/zero-width characters in the URL
+  if (INVISIBLE_CHAR_PATTERN.test(absUrl)) {
```
aramissennyeydd, line 130: Should we be reaching for a tool like https://github.com/markdownlint/markdownlint here instead?

scripts/verify-links.js
Rugvip: Doh, missed the comments :p

Yeah seems like there's something that's potentially possible to use instead: https://docusaurus.io/docs/api/docusaurus-config#onBrokenAnchors

Will try a followup to switch to that, although maybe we just run both, because the link validation script is broader than just the docs

scripts/verify-links.js
Rugvip: Hmm, yep, or potentially custom rules for ESLint / future oxlint

PR #33714, opened 2026-04-01
cli: experimental embedded-postgres support for local dev

Rugvip:

This adds support for using an embedded PostgreSQL server as the database for local development, as an alternative to SQLite. This is driven by config, set `backend.database.client` to `embedded-postgres` in app config and the CLI will automatically start a local PG server and inject the connection details.

This is an alternative solution to #32079. Both are aiming to solve the same problem: to enable the ability to just have a single implementation of the database code, rather than needing to deal with both SQLite for local dev and PostgreSQL for production. This is still experimental, but I think it's time that we enabled this to try it out, and see if it's a direction to take the project. While SQLite is conventient, I do think this could significantly simlify DB access code.

To use it, install `embedded-postgres` as a dev dependency and set this in your dev config:

```yaml
backend:
  database:
    client: embedded-postgres
```

### DIFF
--- a/packages/cli-module-build/src/lib/runner/startEmbeddedDb.ts
+++ b/packages/cli-module-build/src/lib/runner/startEmbeddedDb.ts
@@ -92,11 +92,14 @@ export async function startEmbeddedDb() {
     onLog() {},
   });
 
-  // Create the cluster config files
-  await pg.initialise();
-
-  // Start the server
-  await pg.start();
+  try {
+    await pg.initialise();
+    await pg.start();
+  } catch (error) {
+    await pg.stop().catch(() => {});
+    await fs.remove(tmpDir).catch(() => {});
+    throw error;
+  }
 
   return {
     connection: {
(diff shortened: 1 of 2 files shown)


packages/cli-module-build/src/lib/runner/startEmbeddedDb.ts
```diff
@@ -0,0 +1,116 @@
...
+  const tmpDir = await fs.mkdtemp(resolvePath(os.tmpdir(), TEMP_DIR_PREFIX));
+
+  await fs.writeFile(resolvePath(tmpDir, PID_FILE), String(process.pid));
+
+  const pg = new EmbeddedPostgres({
+    databaseDir: tmpDir,
+    user,
+    password,
+    port,
+    persistent: false,
```
freben, line 88: So it still needs a path when persistent is false?

Rugvip: yeah I think it's just automatic delete under the hood, but that's only as reliably as the close, so therefore the extra cleanup code for stale DBs

PR #33718, opened 2026-04-01
auth0: cache profile API responses during token refresh

UsainBloot:

Every page refresh triggers the auth0 authenticator to call Auth0's `/userinfo` API to fetch the user profile. Auth0 enforces strict rate limits on this endpoint, causing request failures at scale when users frequently refresh pages.

This adds a `createAuth0Authenticator` factory that accepts an optional `CacheService`. When provided, profile responses are cached with a 1-minute TTL so repeated refreshes within that window skip the API call. The default module now uses the cached variant. The existing `auth0Authenticator` export remains available for use without caching.

### DIFF
--- a/plugins/auth-backend-module-auth0-provider/src/authenticator.ts
+++ b/plugins/auth-backend-module-auth0-provider/src/authenticator.ts
@@ -140,17 +140,20 @@ export function createAuth0Authenticator(options?: { cache?: CacheService }) {
       );
 
       const { sub } = decodeJwt(result.params.id_token);
-      const cacheKey = `auth0-profile:${sub}`;
-      let fullProfile = (await profileCache?.get(cacheKey)) as
-        | PassportProfile
-        | undefined;
+      const cacheKey = sub ? `auth0-profile:${sub}` : undefined;
+
+      let fullProfile = cacheKey
+        ? ((await profileCache?.get(cacheKey)) as PassportProfile | undefined)
+        : undefined;
 
       if (!fullProfile) {
         fullProfile = await helper.fetchProfile(result.accessToken);
-        await profileCache?.set(
-          cacheKey,
-          JSON.parse(JSON.stringify(fullProfile)),
-        );
+        if (cacheKey) {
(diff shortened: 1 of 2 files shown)


plugins/auth-backend-module-auth0-provider/package.json
```diff
@@ -37,6 +37,7 @@
     "@backstage/backend-plugin-api": "workspace:^",
     "@backstage/errors": "workspace:^",
     "@backstage/plugin-auth-node": "workspace:^",
+    "@types/passport": "^1.0.3",
```
UsainBloot, line 40: This type dep has moved from `devDependencies` to `dependencies` because the type Strategy is now part of the public api and `yarn lint:type-deps` was failing

plugins/auth-backend-module-auth0-provider/src/authenticator.ts
```diff
@@ -138,7 +138,10 @@ export function createAuth0Authenticator(options?: { cache?: CacheService }) {
         input.scope,
       );
 
-      const cacheKey = `auth0-profile:${input.refreshToken}`;
+      const sub = JSON.parse(
```
UsainBloot: Lets use jose to decode this

plugins/auth-backend-module-auth0-provider/src/authenticator.ts
```diff
@@ -83,58 +92,94 @@ export const auth0Authenticator = createOAuthAuthenticator({
...
+      const result = await PassportHelpers.executeRefreshTokenStrategy(
+        strategy,
+        input.refreshToken,
+        input.scope,
+      );
 
-  async refresh(input, { helper }) {
-    return helper.refresh(input);
-  },
+      const { sub } = decodeJwt(result.params.id_token);
```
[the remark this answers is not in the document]
UsainBloot: Addressed in [dfde6be](https://github.com/backstage/backstage/pull/33718/commits/dfde6becee383b07cebd5890f485b512c17b7418)

PR #33720, opened 2026-04-01
errors: add `toError` utility and migrate usages

Rugvip:

This adds a `toError` utility function to `@backstage/errors`, alongside the existing `isError` and `assertError`. It converts an unknown value to an `ErrorLike` object — if the value is already error-like it's returned as-is, otherwise it's wrapped in a new `Error` with the value stringified as the message.

The PR also migrates all existing `assertError` usages across the codebase to use `toError` instead, along with `isError(e) ? e.message : fallback` ternaries and unsafe `(error as Error).message` casts. The main benefit is that `toError` doesn't throw on non-error values (unlike `assertError`), and it's more concise — a two-line `assertError(e); e.message` pattern becomes just `toError(e).message`.

### DIFF
--- a/plugins/catalog-backend/src/processing/DefaultCatalogProcessingOrchestrator.test.ts
+++ b/plugins/catalog-backend/src/processing/DefaultCatalogProcessingOrchestrator.test.ts
@@ -338,7 +338,7 @@ describe('DefaultCatalogProcessingOrchestrator', () => {
           ok: false,
           errors: [
             new InputError(
-              "Policy check failed for location:default/l; caused by unknown error 'boom'",
+              'Policy check failed for location:default/l; caused by Error: boom',
             ),
           ],
         }),


Rugvip:
@freben in that case it should already be error-like and we just return as is right? But yeah either way I think we can evolve the non-error case conversion over time. Main thing was just that I felt we've been ending up with a ton of workarounds due to the lack of this utility. Either `assertError` that really needed to continue execution, or `isError(e) ? e : String(e)` or even just `String(e)` in places like logging. Actually lemme update this PR with fixing some of those cases

Rugvip:
@freben alright worth another look, added a bunch of usage and update the implementation a bit. Only thing I'm unsure of is whether it's worth handling the `[object Object]` case for an error-like object passed to `toError`, e.g. if you pass `{ name: 'Error', message: 'fail' }`. Feeling we should pass that through as is since making and changes or clones of the object is a bit risky. Throwing something like that is a pretty big mistake in the first place too.

Rugvip:
Gonna merge this one to get it in the `next` release today to have more chance of catching any issues.

PR #33727, opened 2026-04-02
fix(plugin-mcp-actions-backend): fix /.well-known/oauth-protected-resource for external MCP clients

jsundquist:
fix OAuth protected resource metadata for external deployments

Fixes two bugs in the `/.well-known/oauth-protected-resource` endpoint (RFC 9728) that cause
OAuth authentication to fail for MCP clients connecting to Backstage from outside the backend
container (e.g. cloud deployments behind a reverse proxy).

### Bug 1 — Internal URLs returned to external clients

The handler was calling `discovery.getBaseUrl()` for both `auth` and `mcp-actions`. In
production deployments, `getBaseUrl()` resolves to internal container addresses
(e.g. `http://localhost:7007/api/auth`) that are unreachable from external MCP clients.
Both URLs must use `getExternalBaseUrl()` so they resolve using `backend.baseUrl` — the
publicly accessible address.

### Bug 2 — Missing `/v1` path suffix on `resource` field

The Streamable HTTP transport is mounted at `/v1` within the plugin router
(`router.use('/v1', streamableRouter)`), so MCP clients connect to `.../api/mcp-actions/v1`.
RFC 9728 requires the `resource` field to exactly match that URL (or its origin). The
previous value of `resource` was `.../api/mcp-actions` (no `/v1`), causing clients that
validate this field to reject the response with an error such as:

> Protected resource `.../api/mcp-actions` does not match expected `.../api/mcp-actions/v1` (or origin)

### DIFF
--- a/plugins/mcp-actions-backend/src/plugin.test.ts
+++ b/plugins/mcp-actions-backend/src/plugin.test.ts
@@ -367,5 +367,48 @@ describe('Mcp Backend', () => {
       expect(response.body.authorization_servers).toHaveLength(1);
       expect(response.body.authorization_servers[0]).toMatch(/\/api\/auth$/);
     });
+
+    it('should use external base URLs in oauth-protected-resource for external deployments', async () => {
+      const discoveryMock = mockServices.discovery.mock({
+        getBaseUrl: async (plugin: string) =>
+          `http://internal-host:7007/api/${plugin}`,
+        getExternalBaseUrl: async (plugin: string) =>
+          `https://external.example.com/api/${plugin}`,
+      });
+
+      const { server } = await startTestBackend({
+        features: [
+          mcpPlugin,
+          mockPluginWithActions,
+          discoveryMock.factory,
+          mockServices.rootConfig.factory({
+            data: {
+              backend: {
+                actions: {
+                  pluginSources: ['local'],
+                },
+              },
+              auth: {
+                experimentalDynamicClientRegistration: {
+                  enabled: true,
+                },
+              },
+            },
+          }),
+        ],
+      });
+
+      const response = await request(server).get(
+        '/.well-known/oauth-protected-resource',
+      );
+      expect(response.status).toBe(200);
+      expect(response.body.resource).toBe(
+        'https://external.example.com/api/mcp-actions/v1',
+      );
+      expect(response.body.authorization_servers).toHaveLength(1);
+      expect(response.body.authorization_servers[0]).toBe(
+        'https://external.example.com/api/auth',
+      );
+    });
   });
 });


plugins/mcp-actions-backend/src/plugin.ts
```diff
@@ -144,11 +144,11 @@ export const mcpPlugin = createBackendPlugin({
...
               const [authBaseUrl, mcpBaseUrl] = await Promise.all([
-                discovery.getBaseUrl('auth'),
-                discovery.getBaseUrl('mcp-actions'),
+                discovery.getExternalBaseUrl('auth'),
+                discovery.getExternalBaseUrl('mcp-actions'),
               ]);
               res.json({
-                resource: mcpBaseUrl,
+                resource: `${mcpBaseUrl}/v1`,
                 authorization_servers: [authBaseUrl],
```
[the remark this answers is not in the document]
jsundquist: Added additional tests to satisfy request above.

plugins/mcp-actions-backend/src/plugin.ts
```diff
@@ -144,11 +144,11 @@ export const mcpPlugin = createBackendPlugin({
...
             async (_, res) => {
               const [authBaseUrl, mcpBaseUrl] = await Promise.all([
-                discovery.getBaseUrl('auth'),
-                discovery.getBaseUrl('mcp-actions'),
+                discovery.getExternalBaseUrl('auth'),
+                discovery.getExternalBaseUrl('mcp-actions'),
               ]);
               res.json({
-                resource: mcpBaseUrl,
+                resource: `${mcpBaseUrl}/v1`,
```
Rugvip, line 151: Do we need `v1` here? Seems like we might as well stick to the entire router?

jsundquist: The `/v1` is needed because that's the actual endpoint URL clients connect to (`/v1/sse` for SSE, `/v1` for Streamable HTTP). RFC 9728 requires `resource` to be the canonical URL of the protected resource — using just the base URL would be a mismatch since nothing is served there.

freben:
Just noting that if merged, it replaces https://github.com/backstage/backstage/pull/33092

plugins/mcp-actions-backend/src/plugin.ts
benjdlambert: If `/v1/sse` works because `/v1` is in it, then I'd assume that removing the `/v1` should make this work too.

This would be ideal if that's the case because of the server split that's also possible and this "just working" for those servers too.

Can you confirm that it is 100% needed?

benjdlambert: @jsundquist we tested this internally and it works without the `/v1` suffix. Looks like the origin is good enough here and we should stick to that providing it's compliant.

Could you update the PR to reflect these changes? Can ship this in a patch release.

Rugvip:
@freben ah yep, let's go with https://github.com/backstage/backstage/pull/33092

plugins/mcp-actions-backend/src/plugin.ts
jsundquist: Previously I tested both ways and was only able to get it to work within Claude Code to require the `/v1`, suffix. Looks like it is working today with or without it in the code. I can revert the change so that it is just including the `getExternalBaseUrl()` change.

benjdlambert:
Sorry @jsundquist, thanks for the work here, but we just merged the other one as that one came first. Gonna close this but thanks for pushing us in the right direction here :pray:

PR #33731, opened 2026-04-02
feat(ui): export TableBodySkeleton as public API

jroebu14:

Export the `TableBodySkeleton` component from `@backstage/ui` so it can be used independently of the built-in `Table` component — for example, in custom table implementations that use `TableRoot`, `TableHeader`, and `TableBody` directly.

The column type constraint is relaxed from `ColumnConfig<T>` to `{ id: string }` so it works with custom column types.

### DIFF
--- a/docs-ui/src/app/components/table/page.mdx
+++ b/docs-ui/src/app/components/table/page.mdx
@@ -24,6 +24,7 @@ import {
   tableRootPropDefs,
   columnPropDefs,
   rowPropDefs,
+  tableBodySkeletonPropDefs,
 } from './props-definition';
 import {
   tableUsageSnippet,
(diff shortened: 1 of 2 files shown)


.changeset/export-table-body-skeleton.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/ui': minor
```
Naycon: Since we're pre-1.0 and this is not breaking, this should be a patch change I think.

```suggestion
'@backstage/ui': patch
```

jroebu14: Already fixed in b197deecee00.

Issue #33328, opened 2026-03-13
Metrics Service Alpha [meta]


kurtaking:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
metrics-service metrics otel telemetry
```

## 🗃️ Project Area

Metrics

### 🔖 Need

The new alpha metrics service  wraps the OpenTelemetry Meter API and provides plugin-scoped meters. This tracks the migration of existing metrics to the new alpha Metrics Service. 

### Related

- [Service Documentation](https://github.com/backstage/backstage/pull/33020)
- [Metrics BEP](https://github.com/backstage/backstage/blob/master/beps/0012-metrics-service/README.md)

### Existing OTEL metrics

The following work tracks existing OTEL metrics which need to be migrated to the new MetricsService.

| Package | PR Link |
|---|---|
| `@backstage/plugin-catalog-backend` | https://github.com/backstage/backstage/pull/32977 |
| `@backstage/plugin-catalog-backend-module-incremental-ingestion` | https://github.com/backstage/backstage/pull/33349 |
| `@backstage/plugin-catalog-node` | https://github.com/backstage/backstage/pull/33737 |
| `@backstage/plugin-mcp-actions-backend` | https://github.com/backstage/backstage/pull/32978 |
| `@backstage/backend-defaults` scheduler | https://github.com/backstage/backstage/pull/32992 |
| `@backstage/plugin-scaffolder-backend` | https://github.com/backstage/backstage/pull/33534 |
| `@backstage/plugin-notifications-backend-module-slack` | https://github.com/backstage/backstage/pull/33357 |
| `@backstage/plugin-events-backend-module-google-pubsub` | |

---

## Prom-client metric deletion Status

All prom-client metrics currently marked deprecated need to be removed from the codebase.

| Package | PR Link |
|---|---|
| `@backstage/plugin-catalog-backend` | |
| `@backstage/plugin-scaffolder-backend` | |

## Fix existing metrics not following OTEL spec

The following section highlights existing metrics that do not follow the OTEL spec.

## @backstage/plugin-notifications-backend-module-slack

[...]

PR #33737, opened 2026-04-03
Migrate Catalog Scm Events to use MetricsService

kurtaking:

Related to https://github.com/backstage/backstage/issues/33328

Migrates the internal SCM events metrics in the `@backstage/plugin-catalog-node` package to use the new alpha `MetricsService` API, replacing the previous use of OpenTelemetry's metrics API. 

### DIFF
--- a/plugins/catalog-node/src/scmEvents/DefaultCatalogScmEventsService.test.ts
+++ b/plugins/catalog-node/src/scmEvents/DefaultCatalogScmEventsService.test.ts
@@ -15,18 +15,15 @@
  */
 
 import { createDeferred } from '@backstage/types';
-import { MetricsAPI } from '@opentelemetry/api';
+import { metricsServiceMock } from '@backstage/backend-test-utils/alpha';
 import { DefaultCatalogScmEventsService } from './DefaultCatalogScmEventsService';
 
 describe('DefaultCatalogScmEventsService', () => {
-  const counterAdd = jest.fn();
-  const mockMetrics = {
-    getMeter: () => ({
-      createCounter: () => ({
-        add: counterAdd,
-      }),
-    }),
-  } as unknown as MetricsAPI;
+  let mockMetrics: ReturnType<typeof metricsServiceMock.mock>;
+
+  beforeEach(() => {
+    mockMetrics = metricsServiceMock.mock();
+  });
 
   it('should publish and subscribe to events', async () => {
     const service = new DefaultCatalogScmEventsService(mockMetrics);
@@ -118,6 +115,7 @@ describe('DefaultCatalogScmEventsService', () => {
 
     service.markEventActionTaken({ action: 'refresh' });
 
-    expect(counterAdd).toHaveBeenCalledWith(1, { action: 'refresh' });
+    const counter = mockMetrics.createCounter.mock.results[0].value;
+    expect(counter.add).toHaveBeenCalledWith(1, { action: 'refresh' });
   });
 });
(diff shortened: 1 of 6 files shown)


plugins/catalog-node/src/scmEvents/catalogScmEventsServiceRef.ts
kurtaking, line 1: Is this approach an acceptable way to keep this a singleton?

kurtaking: Otherwise, we may need to look into the root-scoped metrics service.

plugins/catalog-node/src/scmEvents/DefaultCatalogScmEventsService.ts
```diff
@@ -38,18 +41,17 @@ import {
...
 
-    const meter = metrics.getMeter('default');
     this.#metrics = {
-      actions: meter.createCounter('catalog.events.scm.actions', {
+      actions: metrics.createCounter('catalog.events.scm.actions', {
         description:
           'Number of actions taken as a result of SCM event messages',
-        unit: 'short',
+        unit: '{action}',
       }),
```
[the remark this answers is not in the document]
kurtaking: This has been updated to use `{action}`.

plugins/catalog-node/src/scmEvents/DefaultCatalogScmEventsService.test.ts
```diff
@@ -118,6 +115,7 @@ describe('DefaultCatalogScmEventsService', () => {
 
     service.markEventActionTaken({ action: 'refresh' });
 
-    expect(counterAdd).toHaveBeenCalledWith(1, { action: 'refresh' });
+    const counter = mockMetrics.createCounter.mock.results[0].value;
```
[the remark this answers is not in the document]
kurtaking: Think this is overkill for the test itself. I recall a conversation somewhere about adding additional utilities to the service mock to help with this. That would be a better route, but still not certain we are ready to add that.

plugins/catalog-node/src/scmEvents/catalogScmEventsServiceRef.ts
aramissennyeydd: Where should these metrics sit namespace wise?

plugins/catalog-node/src/scmEvents/catalogScmEventsServiceRef.ts
```diff
@@ -35,15 +35,17 @@ import { DefaultCatalogScmEventsService } from './DefaultCatalogScmEventsService
...
+      let sharedInstance: DefaultCatalogScmEventsService | undefined;
+      return createServiceFactory({
         service,
-        deps: {},
-        createRootContext() {
-          return new DefaultCatalogScmEventsService(metrics);
+        deps: { metrics: metricsServiceRef },
+        factory({ metrics }) {
+          if (!sharedInstance) {
+            sharedInstance = new DefaultCatalogScmEventsService(metrics);
```
freben, line 45: No I don't think this works out. If we have several plugins on the same machine using this ref, it'll emit metrics just as belonging to the first-across-the-post plugin.

Would it make sense to instead make a single `new DefaultCatalogScmEventsServiceFactory()` in the root context, and giving it a `.instance(options: { pluginMetadata: PluginMetadataService, metrics: MetricsService }): CatalogScmEventsService` method that returns a thing that can both do the "shared" functionality as needed, but at the same time emit plugin-scoped metrics?

freben (changes requested):
Sorry for pushing this for so long - finally getting around to reviewing

Issue #33514, opened 2026-03-22
integrate the permissionsRegistry in the scaffolder-backend plugin


vrabbi:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
permissionsRegistry, scaffolder, permissions
```

### 🗃️ Project Area

Software Templates

### 🔖 Need

The scaffolder plugin currently does not use the permissionsRegistry which is the recommended approach absed on the docs. by not doing so the community RBAC plugin does not allow UI management of the RBAC roles from this plugin as it pulls the data from the registry.

### 📝 Proposal

add the permissionsRegistry integration in the scaffolder backend plugin

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

No, I don't have time to work on this right now

PR #33740, opened 2026-04-03
`feat(scaffolder)`: Migrate scaffolder to use permissions registry

benjdlambert:
With the addition of the `actionsRegistry` to the `scaffolder-backend` plugin, that pulls in the `PermissionsRegistry` service which would override the old legacy permissions router and end up returning zero permissions from the `/.well-known/` endpoint.

This migrates the scaffolder to use the new `PermissionRegistry` service instead.

Fixes https://github.com/backstage/backstage/issues/33514

### DIFF
--- a/plugins/scaffolder-node/report-alpha.api.md
+++ b/plugins/scaffolder-node/report-alpha.api.md
@@ -4,10 +4,16 @@
 
 ```ts
 import { ExtensionPoint } from '@backstage/backend-plugin-api';
+import type { JsonObject } from '@backstage/types';
 import { JsonValue } from '@backstage/types';
+import { PermissionResourceRef } from '@backstage/plugin-permission-node';
+import type { SerializedTask } from '@backstage/plugin-scaffolder-node';
 import { TaskBroker } from '@backstage/plugin-scaffolder-node';
+import type { TaskFilter } from '@backstage/plugin-scaffolder-node';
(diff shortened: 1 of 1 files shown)


plugins/scaffolder-node/src/alpha/index.ts
Rugvip, line 1: These are well established existing permissions right? 🤔, thinking they'd be a stable export straight away

.changeset/scaffolder-backend-permissions-registry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend': minor
```
Rugvip, line 2: If we want to patch we should make it a patch bump as well. Dunno if this is a great patch candidate though. Definitely something where release process separation would help 🫠

plugins/scaffolder-node/src/alpha/index.ts
benjdlambert: Hmm all the existing ones are alpha currently, so not sure if we should do that yet?

Rugvip: oh, the resource refs? yeah in that case we can leave as is. Guess we should get that to stable, only think to consider is whether we want to figure out how to have default decisions be part of the system first, but that might be a separate thing. Have a BEP for that laying around but it ran into a design roadblock that I can't remember right now 😅

PR #33753, opened 2026-04-04
catalog-backend: persist location_entity_ref in locations table

freben:

This is step 1 of a larger effort to migrate \`kind: Location\` entity names from \`generated-<sha1-of-type+target>\` to \`generated-<some-fixed-value>\`, so that a location's identity in the catalog is decoupled from its mutable target URL. This will allow us to implement atomic move operations on locations, as opposed to the past structure that required a delete+add since the name was coupled to the URL.

Note that it _happens_ to pick the same naming pattern still for new locations. But it doesn't have to. It just has to be unique, and when it's persisted, it never changes even if the type or target is updated.

### What this PR does

Adds a DB migration that introduces a \`location_entity_ref\` column on the \`locations\` table. The column stores the full entity ref of the corresponding \`kind: Location\` catalog entity (e.g. \`location:default/generated-<sha1hex>\`).

**Migration (\`20260403000000_add_location_entity_ref.js\`)**:
- Adds the column as \`varchar(255)\` (nullable initially to avoid needing a DEFAULT)
- Backfills all existing non-bootstrap rows by computing the sha1 hash in JS:
  - **Postgres**: single \`UPDATE … FROM unnest(ids::uuid[], refs::text[])\` for all rows at once — the SQL text is constant-size and the planner uses an index nested-loop (O(N)), so one round trip is optimal
  - **MySQL**: 1 000-row batches using \`UPDATE … INNER JOIN (SELECT … UNION ALL …)\` — the SQL text grows linearly with batch size so smaller batches keep packet sizes manageable
  - **SQLite**: per-row updates inside a single transaction
- The internal bootstrap row gets an empty-string placeholder (\`''\`); it will be removed in a future migration
- The column is then tightened to \`NOT NULL\`

**\`DefaultLocationStore\`**: all code paths that previously recomputed the entity ref from type+target now read \`location_entity_ref\` from the DB row instead. New rows inserted by \`createLocation\` and \`#createLocationsByExactUrl\` populate the column at insert time.

The value never changes for a given row, so this is a safe one-time migration. The column gives us a stable handle to reference the entity ref without recomputing it, and is the necessary foundation for the next step (switching to UUID-based names).

- [x] All your commits have a \`Signed-off-by\` line in the message. ([more info](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md#developer-certificate-of-origin))

### DIFF
--- a/plugins/catalog-backend/migrations/20260403000000_add_location_entity_ref.js
+++ b/plugins/catalog-backend/migrations/20260403000000_add_location_entity_ref.js
@@ -33,8 +33,10 @@ const MYSQL_BATCH_SIZE = 1000;
  * removed in a future migration) gets an empty string as a placeholder value.
  *
  * The migration adds the column as nullable first, fills every row, then
- * tightens it to NOT NULL. This avoids the MySQL strict-mode restriction that
- * TEXT columns cannot have DEFAULT values.
+ * tightens it to NOT NULL. Adding a NOT NULL column to a non-empty table
+ * requires a DEFAULT so the database can back-fill existing rows; supplying a
+ * sentinel default here would be misleading, so instead we add the column as
+ * nullable, fill every row with its real value, then tighten the constraint.
  *
  * Postgres:  single `UPDATE … FROM unnest(ids::uuid[], refs::text[])` for all
  *            rows — the SQL text is fixed-size and the planner does an index


freben:
## Migration performance test on a real database

Tested the Postgres backfill against a production-sized database (~50k location rows) via a Cloud SQL proxy. The migration was run inside a transaction that was rolled back, with `EXPLAIN (ANALYZE, BUFFERS)` on the UPDATE step.

**Database:** ~49,623 non-bootstrap rows in the `locations` table.

```
Update on locations  (cost=3733.26..3817.27) (actual time=480.369..480.372 rows=0 loops=1)
  Buffers: shared hit=601829 dirtied=2691 written=1304, temp read=425 written=425
  InitPlan 1 (build id array)
    -> Seq Scan on locations  (actual time=0.008..8.986 rows=49623 loops=1)
         Buffers: shared hit=998
  InitPlan 2 (build ref array)
    -> Seq Scan on locations  (actual time=0.046..34.966 rows=49623 loops=1)
         Buffers: shared hit=998
  -> Nested Loop  (actual time=78.772..233.201 rows=49623 loops=1)
       -> Function Scan on unnest data  (actual time=78.647..99.055 rows=49623 loops=1)
            Buffers: temp read=425 written=425
       -> Index Scan using locations_pkey  (actual time=0.002..0.002 rows=1 loops=49623)
            Index Cond: (id = data.id)
            Buffers: shared hit=198492
Planning Time: 0.875 ms
Execution Time: 481.730 ms
```

**~482 ms total for 49,623 rows.** The plan is clean:

- Two sequential scans to materialise the `id` and `ref` arrays into InitPlans (~9 ms + ~35 ms)
- The `unnest` expands both arrays in lockstep and feeds a tight index scan on `locations_pkey` (0.002 ms per row)
- The `temp read/write=425` is Postgres spilling ~3.4 MB of the two 50k-element arrays to disk during materialisation — expected and harmless for a one-time migration

At this rate the migration would comfortably finish in a few seconds even for a catalog ten times this size.

plugins/catalog-backend/src/util/conversion.ts
```diff
@@ -32,13 +33,35 @@ export function locationSpecToMetadataName(location: LocationSpec) {
...
+   * is taken from the ref instead of being recomputed from the location spec.
+   */
+  locationEntityRef?: string;
 }): LocationEntityV1alpha1 {
   const location = opts.location;
   const parentEntity = opts.parentEntity;
 
+  const name = opts.locationEntityRef
+    ? parseEntityRef(opts.locationEntityRef).name
+    : locationSpecToMetadataName(location);
```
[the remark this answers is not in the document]
freben: No code change needed here: the existing `opts.locationEntityRef ?` ternary already guards against the empty-string case — in JavaScript `''` is falsy, so the bootstrap row's empty-string placeholder falls straight through to `locationSpecToMetadataName(location)` and `parseEntityRef` is never called with an empty string.

plugins/catalog-backend/src/tests/migrations.test.ts
```diff
@@ -1092,4 +1093,85 @@ describe('migrations', () => {
...
+    async databaseId => {
+      const knex = await databases.init(databaseId);
+
+      await migrateUntilBefore(
+        knex,
+        '20260403000000_add_location_entity_ref.js',
+      );
+
+      // The bootstrap location row was added by an earlier migration; it should
+      // NOT receive a location_entity_ref value.
```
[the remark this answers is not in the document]
freben: Fixed — updated the comment to say the bootstrap row gets an empty-string placeholder rather than being skipped.

plugins/catalog-backend/migrations/20260403000000_add_location_entity_ref.js
```diff
@@ -0,0 +1,146 @@
...
+      .comment(
+        'The entity ref of the corresponding Location kind entity, e.g. location:default/generated-<sha1hex>',
+      );
+  });
+
+  // Step 2: Bootstrap row gets an empty string placeholder.
+  await knex('locations')
+    .where('type', 'bootstrap')
+    .update({ location_entity_ref: '' });
+
```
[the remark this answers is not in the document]
freben: Fixed — updated the PR description to reflect the actual behaviour: the column is `varchar(255)`, the bootstrap row gets `''` as a placeholder, and the column is tightened to `NOT NULL` after all rows are filled.

plugins/catalog-backend/migrations/20260403000000_add_location_entity_ref.js
```diff
@@ -0,0 +1,146 @@
...
+  if (rows.length > 0) {
+    /** @type {Array<{ id: string; location_entity_ref: string }>} */
+    const computed = rows.map(row => ({
+      id: row.id,
+      location_entity_ref: `location:default/generated-${createHash('sha1')
+        .update(`${row.type}:${row.target}`)
+        .digest('hex')}`.toLocaleLowerCase('en-US'),
+    }));
+
```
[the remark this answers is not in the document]
freben: The current approach (load all rows, compute hashes in JS, single unnest UPDATE) is intentional. Postgres handles this in one round-trip with an index nested-loop — per a real EXPLAIN ANALYZE on a ~50k-row database it takes ~480 ms total. Streaming page-by-page would trade memory for many more round-trips and a longer open transaction, which is worse in practice for a one-time migration. Leaving as-is.

plugins/catalog-backend/migrations/20260403000000_add_location_entity_ref.js
```diff
@@ -0,0 +1,151 @@
...
+      .nullable()
+      .comment(
+        'The entity ref of the corresponding Location kind entity, e.g. location:default/generated-<sha1hex>',
+      );
+  });
+
+  // Step 2: Bootstrap row gets an empty string placeholder.
+  await knex('locations')
+    .where('type', 'bootstrap')
+    .update({ location_entity_ref: '' });
```
Rugvip, line 70: Hmm, old catalog stuff only right and filtered out of any NBS catalog implementation iirc? 🤔, maybe not the right time to change this, but wondering if it should simply be deleted instead

plugins/catalog-backend/migrations/20260403000000_add_location_entity_ref.js
```diff
@@ -0,0 +1,151 @@
...
+
+    if (client === 'pg') {
+      // One round trip for all rows: the SQL text is constant-size, and
+      // Postgres executes the unnest join with an index nested-loop (O(N)).
+      await knex.raw(
+        `UPDATE locations
+         SET location_entity_ref = data.ref
+         FROM unnest(?::uuid[], ?::text[]) AS data(id, ref)
+         WHERE locations.id = data.id`,
+        [computed.map(r => r.id), computed.map(r => r.location_entity_ref)],
```
Rugvip, line 94: Hmm, gonna hit the parameter limit with this, no? 🤔

plugins/catalog-backend/src/providers/DefaultLocationStore.ts
```diff
@@ -107,7 +111,10 @@ export class DefaultLocationStore implements LocationStore, EntityProvider {
 
     // Always upsert the entity, even if the location already existed, to
     // recover from cases where the entity was inadvertently deleted.
-    const entity = locationSpecToLocationEntity({ location });
+    const entity = locationSpecToLocationEntity({
+      location,
+      locationEntityRef: location.location_entity_ref,
```
Rugvip, line 116: is this duplicate here for future iteration?

Rugvip (approved):
Yep seems like a good direction 👍

Pre-approved in case of minor fixes

plugins/catalog-backend/migrations/20260403000000_add_location_entity_ref.js
freben: Either way it's still in there - as in, there is no migration that deletes it so the code has to deal with it. We can fix that in a separate PR at this point I'm sure, but it's but a minor nuisance in here

plugins/catalog-backend/migrations/20260403000000_add_location_entity_ref.js
freben: They are just two parameters, that happen to be array shaped!

plugins/catalog-backend/src/providers/DefaultLocationStore.ts
freben: It's just that the naming is a bit confusing - the top one is a `LocationSpec` (the input type), and that doesn't have the new field in it - that only exists in the database now

plugins/catalog-backend/migrations/20260403000000_add_location_entity_ref.js
Rugvip: yep np to leave, was just thinking this could be the change that deletes it

plugins/catalog-backend/migrations/20260403000000_add_location_entity_ref.js
Rugvip: ah

PR #33756, opened 2026-04-05
fix(scaffolder): add strict zod validation to catalog:write action

JAi-SATHVIK:
### Problem
Before this change, the `catalog:write` built-in action in the Scaffolder was using a very loose `z.record(z.any()) `schema. This allowed Software Templates to generate invalid `catalog-info.yaml `files (e.g., missing critical fields like `apiVersion `or `kind`). These errors would only be discovered much later when the Catalog failed to register the entity, leading to silent failures and a frustrating debugging experience for users.

### Solution
I implemented Zod schema for the `entity `input. It now enforces the existence of the three fundamental Backstage entity requirements:

1. `apiVersion`
2. `kind`
3. `metadata.name`

I used `.passthrough() `on both the `metadata `and the root object to ensure that all valid extended data (like `spec`, `annotations`, `links`, etc.) is preserved while still guaranteeing that the core file structure is valid.

- [X] A changeset describing the change and affected packages.
- [X] All your commits have a `Signed-off-by` line in the message.

### DIFF
--- a/plugins/scaffolder-backend/src/scaffolder/actions/builtin/catalog/write.ts
+++ b/plugins/scaffolder-backend/src/scaffolder/actions/builtin/catalog/write.ts
@@ -35,13 +35,32 @@ export function createCatalogWriteAction() {
       input: {
         filePath: z =>
           z.string().optional().describe('Defaults to catalog-info.yaml'),
-        // TODO: this should reference an zod entity validator if it existed.
         entity: z =>
           z
-            .record(z.any())
+            .object({
+              apiVersion: z
+                .string()
+                .describe('The version of the catalog API.'),
+              kind: z.string().describe('The kind of the catalog entity.'),
+              metadata: z
+                .object({
+                  name: z.string().describe('The name of the catalog entity.'),
+                  annotations: z
+                    .record(z.string())
+                    .optional()
+                    .describe('Annotations related to the catalog entity.'),
+                })
+                .catchall(z.any())
+                .describe('Metadata related to the catalog entity.'),
+              spec: z
+                .record(z.any())
+                .optional()
+                .describe('The specification of the catalog entity.'),
+            })
+            .catchall(z.any())
             .describe(
               'You can provide the same values used in the Entity schema.',
-            ),
+            ) as any,
       },
     },
     examples,
@@ -61,7 +80,7 @@ export function createCatalogWriteAction() {
             ...(entityRef
               ? {
                   annotations: {
-                    ...entity.metadata.annotations,
+                    ...(entity.metadata as any).annotations,
                     'backstage.io/source-template': entityRef,
                   },
                 }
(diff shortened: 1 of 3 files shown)


plugins/scaffolder-backend/src/scaffolder/actions/builtin/catalog/write.ts
```diff
@@ -35,13 +35,32 @@ export function createCatalogWriteAction() {
...
+              spec: z
+                .record(z.any())
+                .optional()
+                .describe('The specification of the catalog entity.'),
+            })
+            .catchall(z.any())
             .describe(
               'You can provide the same values used in the Entity schema.',
-            ),
+            ) as any,
```
benjdlambert: This makes the API report worse. Let's remove this.

plugins/scaffolder-backend/src/scaffolder/actions/builtin/catalog/write.ts
```diff
@@ -35,13 +35,32 @@ export function createCatalogWriteAction() {
...
+                    .describe('Annotations related to the catalog entity.'),
+                })
+                .catchall(z.any())
+                .describe('Metadata related to the catalog entity.'),
+              spec: z
+                .record(z.any())
+                .optional()
+                .describe('The specification of the catalog entity.'),
+            })
+            .catchall(z.any())
```
benjdlambert: Do we need this here?

I don't think there should be anything extra at the top level envelope

plugins/scaffolder-backend/src/scaffolder/actions/builtin/catalog/write.ts
```diff
@@ -61,7 +80,7 @@ export function createCatalogWriteAction() {
             ...(entityRef
               ? {
                   annotations: {
-                    ...entity.metadata.annotations,
+                    ...(entity.metadata as any).annotations,
```
benjdlambert: This also shouldn't be needed to cast either

benjdlambert (changes requested):
Hey!

Thanks for this, few little things to clean up first I think 🙏

plugins/scaffolder-backend/src/scaffolder/actions/builtin/catalog/write.ts
JAi-SATHVIK: Removed `as any` from the bottom of the schema.

plugins/scaffolder-backend/src/scaffolder/actions/builtin/catalog/write.ts
JAi-SATHVIK: Removed `.catchall(z.any())` from the root schema.

plugins/scaffolder-backend/src/scaffolder/actions/builtin/catalog/write.ts
JAi-SATHVIK: Replaced `...(entity.metadata as any).annotations` with `...(entity.metadata.annotations || {})`

JAi-SATHVIK:
Hi @benjdlambert, just checking back in on this. I’ve addressed the requested changes:

- Removed the as any casts from the schema and the metadata spread.
- Removed .catchall(z.any()) from the root schema.
- Verified that the API report remains clean with these changes.


Thanks,

PR #33761, opened 2026-04-05
cli: add cli-module-vitest and begin Jest to Vitest migration

Rugvip:
Experimenting a bit with Vitest


Rugvip:
Perhaps obvious conclusion from this so far: it's not doable without using an incremental approach

PR #33782, opened 2026-04-07
feat(backend): add extensionPointFactoryMiddleware to createBackend

UsainBloot:
## Context

There is a need to reimplement the behavior of backend extension points without modifying plugin or module code directly. Common use cases include injecting cross-cutting concerns like logging, metrics, or observability into extension point interactions, enforcing organizational policies at extension point boundaries, and transforming or filtering data as it flows through the extension point API.

Today, there is no generic mechanism for this in the backend system. To be able to alter extension points behavior, you must choose between forking plugins, duplicating module configuration, or making yarn patches. All of these approaches are fragile, hard to maintain, and don't scale across multiple extension points.

The frontend system already solves this problem through `extensionFactoryMiddleware`, which allows the frontend to alter extension factory outputs before they reach consumers. The backend system lacks an equivalent pattern.

### Type Definition

`ExtensionPointFactoryMiddleware` is an opaque type, constructed only via the `createExtensionPointFactoryMiddleware` helper. Internally backed by `OpaqueType` from `@internal/opaque` in a new `@internal/backend` package.

```typescript
// Public interface (opaque)
export interface ExtensionPointFactoryMiddleware {
  $$type: '@backstage/ExtensionPointFactoryMiddleware';
}

// Factory function (preserves type inference)
export function createExtensionPointFactoryMiddleware<T>(options: {
  extensionPoint: ExtensionPoint<T>;
  middleware: (original: T) => Promise<T>;
}): ExtensionPointFactoryMiddleware;
```

### Usage

Available on `createSpecializedBackend` only (not `createBackend`), since this is an advanced use case. `defaultServiceFactories` is now exported from `@backstage/backend-defaults` to support this.

```typescript
import {
  createSpecializedBackend,
  createExtensionPointFactoryMiddleware,
} from '@backstage/backend-app-api';
import { defaultServiceFactories } from '@backstage/backend-defaults';
import { myExtensionPoint } from '@internal/my-plugin-node';

const backend = createSpecializedBackend({
  defaultServiceFactories,
  extensionPointFactoryMiddleware: [
    createExtensionPointFactoryMiddleware({
      extensionPoint: myExtensionPoint,
      middleware: async original => ({
        ...original,
        addHandler(...handlers) {
          // Reimplement each handler with logging before passing through
          const enhanced = handlers.flat().map(h => withLogging(h));
          original.addHandler(...enhanced);
        },
      }),
    }),
  ],
});
```

### Key Features

1. **Immutable** - Set once at backend creation time, fixed after creation. Mirrors the frontend's `extensionFactoryMiddleware`.
2. **Declarative** - Target extension points by reference, not string IDs. The framework handles matching and ignores middleware for unregistered extension points.
3. **Type-safe** - `middleware` callback is fully typed via inference from the extension point.
4. **Backend-init only** - Only the backend entry point can set middleware. Plugins and modules have no access.
5. **Composable** - Multiple middlewares can target the same extension point, chaining in declaration order.

### DIFF
--- a/packages/backend-internal/.eslintrc.js
+++ b/packages/backend-internal/.eslintrc.js
@@ -1,5 +1,1 @@
-module.exports = require('@backstage/cli/config/eslint-factory')(__dirname, {
-  rules: {
-    '@backstage/no-top-level-material-ui-4-imports': 'error',
-  },
-});
+module.exports = require('@backstage/cli/config/eslint-factory')(__dirname);


packages/backend-app-api/src/wiring/types.ts
```diff
@@ -21,6 +21,34 @@ import {
...
 } from '@backstage/backend-plugin-api';
 
+/**
+ * A middleware entry that reimplements a specific extension point's output.
+ * The framework matches by extension point ID and passes through all
+ * non-matching extension points automatically.
+ *
+ * @public
+ */
+export interface ExtensionPointFactoryMiddleware<T = unknown> {
```
Rugvip: It's a lil bit trivial, but I think it's worth making this an opaque type tbh. Would be the first one in a `backend-internal` package

packages/backend-app-api/src/wiring/types.ts
```diff
@@ -21,6 +21,34 @@ import {
...
+  extensionPoint: ExtensionPoint<T>;
+  middleware: (original: T) => T;
```
Rugvip: What's your thought on the interaction with the factory variant of extension points?

https://github.com/backstage/backstage/blob/f1d29b4d4d5191e5c0c0eb98f797324d703133b9/packages/backend-plugin-api/src/wiring/types.ts#L73-L83

Could be fine to just call this multiple times simply, but wondering if there are other options to consider

packages/backend-app-api/src/wiring/types.ts
```diff
@@ -21,6 +21,34 @@ import {
...
+}
+
+/**
+ * Creates a typed middleware entry that reimplements a specific extension point.
+ * Use this helper to preserve type inference for the middleware callback.
+ *
+ * @public
+ */
+export function createExtensionPointFactoryMiddleware<T>(
```
Rugvip: Options object 😁

packages/backend-defaults/report.api.md
```diff
@@ -5,10 +5,18 @@
 ```ts
 import { Backend } from '@backstage/backend-app-api';
 import { BackendFeature } from '@backstage/backend-plugin-api';
+import { createExtensionPointFactoryMiddleware } from '@backstage/backend-app-api';
+import { ExtensionPointFactoryMiddleware } from '@backstage/backend-app-api';
 
 // @public (undocumented)
-export function createBackend(): Backend;
+export function createBackend(options?: {
+  extensionPointFactoryMiddleware?: ExtensionPointFactoryMiddleware[];
```
Rugvip: Maybe we could keep this only on `createSpecializedBackend`? Just to have it a bit more hidden given the fairly advanced use-case. Main thing we need to enable that with minimal overhead is exporting `defaultServiceFactories`, which I think is fine

packages/backend-app-api/src/wiring/types.ts
```diff
@@ -21,6 +21,34 @@ import {
...
+  extensionPoint: ExtensionPoint<T>,
+  middleware: (original: T) => T,
```
Rugvip: Thinking we might want this to be async actually? 🤔

We'd need to force it to always be, as in we can't do `T | Promise<T>` because `T` could be anything, including a promise.

packages/backend-app-api/src/wiring/types.ts
```diff
@@ -20,6 +20,34 @@ import {
...
+import { OpaqueExtensionPointFactoryMiddleware } from '@internal/backend';
+
+/**
+ * A middleware entry that reimplements a specific extension point's output.
+ * The framework matches by extension point ID and passes through all
+ * non-matching extension points automatically.
+ *
+ * @public
+ */
+export interface ExtensionPointFactoryMiddleware<_T = unknown> {
```
Rugvip: I guess we don't really need the type parameter on this type anymore? It's more of a declaration-time thing

UsainBloot: Addressed in [28a362f](https://github.com/backstage/backstage/pull/33782/commits/28a362f2f769e73e3ab749a5c195a2eff9a58407)

packages/backend-internal/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = require('@backstage/cli/config/eslint-factory')(__dirname, {
+  rules: {
+    '@backstage/no-top-level-material-ui-4-imports': 'error',
```
benjdlambert: hmm wat?

.changeset/extension-point-middleware-backend-app-api.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-app-api': minor
+---
+
+Added `ExtensionPointFactoryMiddleware` type and `createExtensionPointFactoryMiddleware` helper to reimplement extension point outputs at backend creation time.
```
benjdlambert, line 5: Hmm `Middleware` is perhaps the wrong name here? It's not really middleware in the traditional sense. Some other potential names, but honestly not gonna die on this hill either.

`ExtensionPointInterceptor`, `ExtensionPointDecorator`, `ExtensionPointOverride`?

.changeset/extension-point-middleware-backend-defaults.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/backend-defaults': patch
```
benjdlambert, line 2: Think that this should be `minor` rather than `patch` as it's 1.x and it's a new feature.

.changeset/extension-point-middleware-backend-app-api.md
freben: It likely will exclusively be used in a middleware like manner right? Teeing into calls, metrics etc, MAYBE filtering - but ultimately generally calling the original one. Outright replacing isn't really an option since the target plugin holds an instance of the leaf implementation no matter what - it's not being replaced.

.changeset/extension-point-middleware-backend-defaults.md
UsainBloot: In agents.md
> For non-breaking changes that introduce new APIs or features, use minor for packages at version 1.0.0 or higher, and patch for packages below 1.0.0

Should this package be released as 1.0.0? `@internal/frontend` currently is only on `0.0.x`

packages/backend-internal/.eslintrc.js
UsainBloot: yep fair

.changeset/extension-point-middleware-backend-defaults.md
freben: The `@internal` ones aren't released themselves though, so their versions effectively don't matter.

.changeset/extension-point-middleware-backend-app-api.md
UsainBloot: The types of the factory should enforce that the middleware has the same base signature as the original one.

I'm not against it being called override, but I tried to align the name with the frontend version of this.

.changeset/extension-point-middleware-backend-defaults.md
benjdlambert: Oh wait, sorry, I thought `backend-defaults` was 1.x. Ignore.

Rugvip (approved):
:shipit:

PR #33783, opened 2026-04-07
feat(ui): add automatic active tab detection to Header

Naycon:

When `activeTabId` is omitted, `HeaderNav` now auto-detects the active
tab from the current route using `matchRoutes`. This eliminates the
boilerplate `useActiveTabId` pattern that every consumer had to
reimplement.

The `activeTabId` prop accepts three states:
- **`undefined`** (omitted) — auto-detect from current route
- **`string`** — explicit tab ID (existing behavior, unchanged)
- **`null`** — explicitly no active tab

Also refactors `HeaderNavLink` to properly use the definition system
(`resolveHref`, `analytics`) and resolves relative tab hrefs correctly.

### DIFF
--- a/.changeset/header-nav-resolve-href.md
+++ b/.changeset/header-nav-resolve-href.md
@@ -1,0 +1,11 @@
+---
+'@backstage/ui': minor
+---
+
+**BREAKING**: Tab `href` values in the Header component are now resolved through the router context instead of being passed raw to the `<a>` tag. This means relative `href` values (e.g. `sub3`, `./sub4`, `../catalog`) are now resolved against the current route, and absolute `href` values may be affected by the router's `basename` configuration.
+
+**Migration:**
+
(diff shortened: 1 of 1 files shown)


packages/ui/report.api.md
```diff
@@ -1560,7 +1477,16 @@ export const HeaderNavItemDefinition: {
   readonly classNames: {
     readonly root: 'bui-HeaderNavItem';
   };
+  readonly analytics: true;
+  readonly resolveHref: true;
```
Rugvip, line 1481: Is setting this a breaking change?

Naycon: Hmm, maybe it depends on how you reason about it? 🤔 It's breaking in that it changes the behavior, but the previous behavior was broken and not following any reasonable patterns. But yes, can mark it as a breaking change just in case someone relied on it.

Rugvip: I was thinking more generally actually, what is the impact on behavior from setting or clearing this, should we consider both to be breaking, neither, or just one direction? Not blocking this PR btw, just realized it's something that's useful to settle

Issue #33695, opened 2026-03-31
Support spec.owners (plural) as an alternative to spec.owner in catalog entity model


SaachiNayyer:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
catalog, owner, ownedby
```

### 🗃️ Project Area

Catalog

### 🔖 Need

Currently, catalog entity model only supports spec.owner as a single string value. Many adopters need to assign multiple owners to a single entity (e.g., shared ownership across teams or individuals).

Today, the only way to achieve this is by building a custom entity processor that:
Defines an extended entity type with a spec.owners array field
Emits RELATION_OWNED_BY / RELATION_OWNER_OF for each entry in the array
Maps spec.owners[0] back to spec.owner for backward compatibility

This is a common enough need that native support would benefit the broader Backstage community. Notably, the IS_ENTITY_OWNER permission rule already works correctly with multiple ownedBy relations — the gap is only in the catalog model and the built-in entity processor.

### 📝 Proposal

Add optional spec.owners (string array) support to the catalog entity descriptor format for Component, System, Domain, API, and Template kinds:

When spec.owners is provided, the BuiltinKindsEntityProcessor emits ownedBy/ownerOf relations for all entries
spec.owner (singular) continues to work as-is — no breaking change

If both spec.owner and spec.owners are provided, spec.owners takes precedence (with spec.owner treated as a fallback or ignored)
Validation updated to accept either field

This is a non-breaking, additive change. Existing entities using spec.owner are unaffected.

### 🔄 Alternatives

Custom entity processor (current workaround): Each adopter builds their own processor to handle spec.owners. This works but duplicates effort across the community.

Using group ownership only: Assigning a group as owner and adding individual users to that group. This doesn't always fit organizational structures where ownership is ad-hoc or cross-team.

### Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit a PR?

Yes, and I have enough information to get started

awanlin:
Hi @SaachiNayyer, I've left comments on your PR for this but missed this linked issue. 

[...]

Issue #17017, opened 2023-03-22
scaffolder/next: `async` Validators should show a loading form


benjdlambert:
Currently, when the `async` validators are run, there's no indication to the user that they are running, so with longer running validators you don't see any loading screen or what not.

Maybe we can grey out the form, or at least the the submit button so that the user is aware that the form is loading.

benjdlambert:
Closed by https://github.com/backstage/backstage/pull/17316

Issue #1329, opened 2020-06-17
Allow multiple owners for components


JordiPolo:
Currently a `kind: Component` needs to specify an `owner` inside `spec`. But this information may depend of who you ask! 
Is the owner the lead engineer, the team, the guy paying , the manager.... Depends!

So I propose that:
```yaml
spec:
  owner: team@company.com
```

Becomes:
```yaml
spec:
  owners: 
    team: team@company.com
```

Which now allows:
```yaml
spec:
  owners: 
    team: team@company.com
    manager: xx
    architecture: xx
    documentation: yyy
    quality: zzzz
    sales: yyy
    consumer_service: ddd
```

The change would not limit the owner keys to the above, it is just an example.

Rugvip:
Good feedback! 👍 

For now we've just gone with a simple dumb string as the owner, as we haven't really tied it to any kind of identities yet, and we still need to figure out exactly how we want to do that. I do think we're going to have to extend the schema to allow for different types of owners later on.

In the internal version we do have a rule that some team should be **the** owner of a service, and it might be a model that we would want to encourage through the schema, but not enforce.

Fox32:
We are also interested into having multiple owners. Here is an inspiration from one of the Demo videos:

![image](https://user-images.githubusercontent.com/648527/101139048-3ff07800-3611-11eb-934d-fd628f8edf7f.png)

Now that the relationships and basic ownership is in place, we should have a look at this again. I wonder how such a feature can be implemented, some ideas:

### Having multiple distinct owner fields or a complex structure in the entity as Jordi suggested:

```yaml
spec:
  owners: 
    technical: team@company.com
    product: xx
    ...
```

or

```yaml
spec:
  technicalOwner: team@company.com
  productOwner: xx
```

In that case we have multiple options for storing it internally:
* Having a new `technicalOwnerOf` relationship and some more relationships following the same pattern. This might get out of hand fast. But would be extensible by following [Adding a new relation](https://backstage.io/docs/features/software-catalog/extending-the-model#adding-a-new-relation-type). One advantage is, that this might be introduced as a non breaking change as the ownership relationship isn't polluted with additional relationships. Code that can't handle other ownership relationships just ignores them.
[...]

PR #33786, opened 2026-04-07
Support spec.owners as an alternative to spec.owner in catalog entity…

SaachiNayyer:
… model #33695

Added optional `spec.owners` (string array) field to Component, API, System, Domain, and Resource kinds. Either `spec.owner` or `spec.owners` is accepted. Existing `spec.owner` usage is unaffected.

### DIFF
--- a/.changeset/weak-breads-relax.md
+++ b/.changeset/weak-breads-relax.md
@@ -1,0 +1,8 @@
+---
+'@backstage/plugin-catalog-backend-module-scaffolder-entity-model': minor
+'@backstage/plugin-scaffolder-common': minor
+'@backstage/plugin-catalog-backend': minor
+'@backstage/catalog-model': minor
+---
+
+Added optional `spec.owners` (string array) field to Component, API, System, Domain, Resource, and Template kinds. When `spec.owners` is provided, ownership relations are emitted for all entries. Existing `spec.owner` usage is unaffected.
--- a/packages/catalog-model/report.api.md
+++ b/packages/catalog-model/report.api.md
@@ -42,7 +42,8 @@ interface ApiEntityV1alpha1 extends Entity {
   spec: {
     type: string;
     lifecycle: string;
-    owner: string;
+    owner?: string;
+    owners?: string[];
     definition: string;
     system?: string;
   };
@@ -82,7 +83,8 @@ interface ComponentEntityV1alpha1 extends Entity {
   spec: {
     type: string;
     lifecycle: string;
-    owner: string;
+    owner?: string;
+    owners?: string[];
     subcomponentOf?: string;
     providesApis?: string[];
     consumesApis?: string[];
@@ -122,7 +124,8 @@ interface DomainEntityV1alpha1 extends Entity {
   kind: 'Domain';
   // (undocumented)
   spec: {
-    owner: string;
+    owner?: string;
+    owners?: string[];
     subdomainOf?: string;
     type?: string;
   };
@@ -411,7 +414,8 @@ interface ResourceEntityV1alpha1 extends Entity {
   // (undocumented)
   spec: {
     type: string;
-    owner: string;
+    owner?: string;
+    owners?: string[];
     dependsOn?: string[];
     dependencyOf?: string[];
     system?: string;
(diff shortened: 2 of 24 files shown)


.changeset/weak-breads-relax.md
```diff
@@ -0,0 +1,8 @@
+---
+'@backstage/plugin-catalog-backend-module-scaffolder-entity-model': minor
+'@backstage/plugin-scaffolder-common': minor
+'@backstage/plugin-catalog-backend': minor
+'@backstage/catalog-model': minor
+---
+
+Added optional `spec.owners` (string array) field to Template kind.
```
awanlin: Why is this change needed, would like to better understand this?

My first thought is why not simply use a group which is supported but as there is lack of details around the need for this change it's hard to offer better solutions or if this might be the right approach.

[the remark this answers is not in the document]
SaachiNayyer: done

SaachiNayyer: Hi @awanlin 
Today, spec.owner accepts a single entity reference. If two teams (e.g., team-platform and team-security) jointly own a service, the only workaround is to create a synthetic group (e.g., group:platform-and-security-owners) solely to represent that shared ownership. This leads to:
Phantom groups in the catalog that don't map to real org structure

Permission drift — the IS_ENTITY_OWNER rule already evaluates all ownedBy relations, but there's no first-class way to populate multiple ownedBy relations without a custom processor

Adoption friction — this is one of the most common custom processors adopters build (see [#17017](https://github.com/backstage/backstage/issues/17017))
spec:
  owners:
    - group:team-platform
    - group:team-security

awanlin: Thanks for the follow up. To be clear Ownership should be assigned to the team that owned the code and is tasked with bug fixes and feature additions. We clearly outline this in the documentation here: https://backstage.io/docs/features/software-catalog/descriptor-format#specowner-required.

Do these two teams truly own it? Or is it more likely that they each own a distinct aspect? Creating custom relation for this case would be the right approach: https://backstage.io/docs/features/software-catalog/extending-the-model#adding-a-new-relation-type 

As for adoption friction would be great to get actual data on this comment. My anecdotal experience working with Backstage adopters as part of my day to day and on Discord doesn't point in that direction. Having data around this would help better understand the need and what the friction points are. 

FYI the issue your linked to is about the Scaffolder, maybe there was a typo?

SaachiNayyer: Apologies for the wrong issue link earlier. The closest prior discussion is [#1329](https://github.com/backstage/backstage/issues/1329), though that was about typed ownership roles (tech owner, product owner) — a much bigger model change. Our proposal is narrower: just multiple entries for the same ownedBy relation.

We agree with the single-ownership philosophy for the common case — clear accountability matters. But there are scenarios where two teams genuinely co-own the same codebase in the same capacity, not distinct aspects:

A service being migrated between teams — during transition, both teams are equally responsible for bugs, on-call, and releases. This can last weeks or months.

An inner-source library where contributions and maintenance are shared across teams by design, with no single team being "the" owner.

In these cases, creating a synthetic group doesn't reflect reality, and custom relations like coOwnedBy wouldn't integrate with IS_ENTITY_OWNER permission checks or catalog ownership filters without additional work. The permission framework is already built to evaluate multiple ownedBy relations — this change just lets the descriptor format express that.

spec.owner continues to work exactly as-is — this is purely additive. Adopters who follow the single-owner model are completely unaffected

awanlin: Having a single owner is a forcing function for both examples you shared: it forces a decision to be made to provide clear ownership. For a migration the new team should take ownership and then fallback to the previous team should they need help; having two owners creates confusion for those needing support. For an inner-source library there will be a distinct owner for it and that generally will come out naturally over it's lifecycle based on those contributing; it will generally start with the team that the initial contributor is a member of.

Taking a step back: say there are two, or three, or four teams that are the owner of a Component, how do I know which team I should ask for help? How more likely is it that I get bounced from team to team to team? This waters down one of the core the benefits that Backstage provides via the Catalog.

You can create custom permissions rules so that's not really a blocker from my point of view. To be clear my issue isn't the technical merits of this change - it's should we even do this at all.

PR #33789, opened 2026-04-07
feat(catalog): expose `entityRef` on Location type and add `PUT /locations/:id`

freben:

This PR is a follow-up to #33753 (which added `location_entity_ref` to the `locations` DB table). It exposes that stable value through the API and adds a new endpoint for updating existing locations in-place.

### Changes

**`entityRef` on `Location`:**
- All location API responses (`GET /locations`, `GET /locations/:id`, `GET /locations/by-entity/...`, `POST /locations/by-query`) now include an `entityRef` field with the stable entity reference (e.g. `location:default/generated-<sha1hex>`) for each location.
- `entityRef` is filterable via `POST /locations/by-query` alongside the existing `id`, `type`, and `target` filters.
- The `Location` type in `catalog-client`, `catalog-node`, and the generated OpenAPI models all include the new required `entityRef` field.

**`PUT /locations/:id`:**
- New endpoint that accepts the same `{ type, target }` body as `POST /locations` and updates the type/target of an existing location without changing its entity ref.
- Issues an `added`-only delta mutation so the catalog entity is updated in-place — the entity ref stays stable while the spec changes. Since `applyMutation` treats `added` as an upsert, no explicit `removed` entry is needed.
- The underlying DB update uses `UPDATE ... RETURNING` on Postgres/SQLite (single round-trip) and `UPDATE` + row-count check + `SELECT` inside a transaction on MySQL.
- Uses `catalogLocationCreatePermission` for authorization.
- Wired through `LocationStore`, `LocationService`, `AuthorizedLocationService`, `CatalogApi`, `CatalogService`, and `CatalogClient`.


packages/catalog-client/src/types/api.ts
```diff
@@ -372,6 +372,8 @@ export type Location = {
   id: string;
   type: string;
   target: string;
+  /** The entity ref of the corresponding Location kind entity, e.g. `location:default/generated-<sha1hex>`. */
+  entityRef: string;
```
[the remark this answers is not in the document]
freben: added a `**BREAKING PRODUCERS**` instead since this is not a type that anyone typically creates, except as noted, perhaps in tests

plugins/catalog-backend/src/service/AuthorizedLocationService.ts
```diff
@@ -123,6 +123,25 @@ export class AuthorizedLocationService implements LocationService {
...
   }
 
+  async updateLocation(
+    id: string,
+    location: LocationInput,
+    options: { credentials: BackstageCredentials },
+  ): Promise<Location> {
+    const authorizationResponse = (
+      await this.permissionApi.authorize(
+        [{ permission: catalogLocationCreatePermission }],
```
Rugvip, line 133: Hmm, is this really the correct behavior? 🤔, I somewhat expected a more strict check here. Seems a bit tricky that anyone that can register a new location can also update someone else's

Rugvip (approved):
Yep! 👍 

Just a Q on permissions

plugins/catalog-backend/src/service/AuthorizedLocationService.ts
freben: Tricky thing is, we don't have resource permissions to lean upon here. We don't have any ownership of neither the location nor the Location entity. Which, arguably, we should fix some day of course.

freben: We could check for BOTH delete and create if we wanted to? If that's what we feel that it effectively amounts to. But it would still be the basic global permission, not per location.

PR #33804, opened 2026-04-08
feat(ui): resolve route-relative hrefs by default in useDefinition

Naycon:

Made href resolution always-on in `useDefinition` instead of requiring each component to opt in via `resolveHref: true`. The hook now always calls `useInRouterContext`/`useHref` and only applies the resolved value when an `href` prop is actually provided, preventing href from leaking into components that don't accept it.

- Removed `resolveHref` from `ComponentConfig` and all 14 component definitions
- Removed `ResolveHrefConstraint` type (no longer needed)
- Removed a now-unnecessary type cast in the `Text` component

### DIFF
--- a/.changeset/bumpy-boxes-chew.md
+++ b/.changeset/bumpy-boxes-chew.md
@@ -1,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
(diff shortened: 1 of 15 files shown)


Rugvip (approved):
Yep, makes sense 👍

PR #33835, opened 2026-04-10
fix(ui): preserve active tab indicator for non-routed tabs

jroebu14:

Fixes the active tab indicator disappearing on initial render when using non-routed (uncontrolled) `Tabs` — i.e. `<Tab id="foo">` without an `href` prop. Also ensures the indicator correctly hides when transitioning from a selected tab back to no selection in controlled mode.

### Root Cause

PR #33784 changed `.bui-TabActive` from `opacity: 1` (hardcoded) to `opacity: var(--active-tab-opacity)`, initialized to `0`. The JS in `TabsIndicators` sets it to `1` when `state.selectedKey` is a valid key, and `0` otherwise.

For **routed tabs** (with `href`), the `Tabs` component passes `selectedKey=""` when no tab matches the current route — correctly hiding the indicator.

For **non-routed tabs** (no `href`), `selectedKey={undefined}` is passed to React Aria (uncontrolled mode). React Aria manages selection internally, but `state.selectedKey` from `TabListStateContext` can be `null` briefly before the state settles. The `else` branch in `updateCSSVariables` was treating this the same as `""`, setting opacity to `0` and hiding the indicator.

### Fix

The `else` branch is replaced with an `else if` that hides the indicator only when:

1. `selectedKey === ''` — explicitly deselected (routed tabs, no URL match)
2. `selectedKey` is `null`/`undefined` **and** a tab was previously selected — transitioning back to no selection (controlled mode deselection)

When `selectedKey` is `null`/`undefined` and no tab was previously selected (uncontrolled mode initial render), the indicator state is left unchanged so it doesn't flash away before React Aria's state settles.

### How to reproduce

Use `Tabs` without `href` on the `Tab` components:

```tsx
<Tabs>
  <TabList>
    <Tab id="overview">Overview</Tab>
    <Tab id="review">Review</Tab>
  </TabList>
  <TabPanel id="overview">...</TabPanel>
  <TabPanel id="review">...</TabPanel>
</Tabs>
```

The active indicator appears momentarily on page load, then disappears. It reappears when hovering any tab.

#### ✅ Checklist

### DIFF
--- a/.changeset/fix-tabs-indicator-uncontrolled.md
+++ b/.changeset/fix-tabs-indicator-uncontrolled.md
@@ -1,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Fixed the active tab indicator disappearing in non-routed (uncontrolled) Tabs. The indicator opacity is now only hidden when `selectedKey` is explicitly empty (routed tabs with no matching route), rather than when it is `null` or `undefined` (uncontrolled mode where React Aria manages selection internally).
--- a/.changeset/fix-tabs-indicator-uncontrolled.md
+++ b/.changeset/fix-tabs-indicator-uncontrolled.md
@@ -2,4 +2,4 @@
 '@backstage/ui': patch
 ---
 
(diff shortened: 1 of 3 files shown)


Naycon:
Closing this, since #33984 has been merged and addresses the same issue.

PR #33842, opened 2026-04-10
fix: handle limit=0 in getNotifications query

suhr25:
### **SUMMARY**

This PR fixes a pagination bug where `limit=0` returned all notifications instead of an empty list.
The issue was in `DatabaseNotificationsStore.getNotifications()` due to incorrect checks for `limit` and `offset`.

---

#### **Before**

```ts
if (options.limit) {
  query.limit(options.limit);
}
```

#### **After**

```ts
if (options.limit !== undefined) {
  query.limit(options.limit);
}
```

---

### **VERIFICATION**

Calling the API with `limit=0` previously returned all notifications.
Now it correctly returns an empty array.
Other values behave exactly the same.

### DIFF
--- a/.changeset/fix-notifications-limit-zero.md
+++ b/.changeset/fix-notifications-limit-zero.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-notifications-backend': patch
+---
+
+Fix handling of `limit=0` in `getNotifications` query to return empty results instead of all notifications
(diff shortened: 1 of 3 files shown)


freben (changes requested):
Hi! Thanks. This also needs a changeset in order to be shipped.

freben (approved):

PR #33846, opened 2026-04-11
feat(ui): add Storybook MCP setup and UI eval system for @backstage/ui

cdedreuille:

This PR sets up the Storybook MCP server integration for `@backstage/ui` and introduces a TypeScript eval harness to benchmark LLM-generated UI code quality with and without MCP context.

### What's included

**MCP registration** (`.cursor/mcp.json`)
Registers `backstage-ui-mcp` pointing at `http://localhost:6006/mcp` — the Storybook MCP server exposed by the already-installed `@storybook/addon-mcp`. Cursor picks this up automatically so agents have access to `@backstage/ui` component docs in every session.

**AGENTS.md update**
Adds a `## UI Development` section instructing agents to always query MCP before generating any `@backstage/ui` code, and to treat `packages/ui/src/recipes/` and `packages/ui/src/guidelines/` as the golden-path composition reference.

**`packages/ui/evals/` — eval harness**
A plain TypeScript subfolder (no separate `package.json`) that runs via `tsx` from the repo root:

```bash
# start Storybook first (MCP mode)
yarn storybook &
ANTHROPIC_API_KEY=sk-ant-... yarn eval:ui
# baseline only (no Storybook needed)
ANTHROPIC_API_KEY=sk-ant-... yarn eval:ui --mode baseline
```

Six eval tasks in two tiers:

- **Tier 1 — Recipe/Guideline tasks** (grounded in the real golden-path stories): `cards-with-list`, `cards-with-table`, `header-with-actions`
- **Tier 2 — Component API tasks** (individual prop/API knowledge): `login-form`, `dialog-form`, `data-table`

Each task runs in **baseline** mode (LLM gets only the prompt) and **MCP** mode (LLM gets real component docs fetched from the Storybook MCP via raw `fetch` — no SDK). All LLM calls use native Node.js `fetch` against the Anthropic API; no new SDK dependencies are introduced.

Scoring is entirely AST-based and deterministic (no LLM at judge time):
- **Import correctness** — all `@backstage/ui` components imported from the right package
- **Component selection** — required components present in JSX
- **Prop accuracy** — no hallucinated props (vs MCP-documented props)
- **Recipe conformance** (Tier 1 only) — JSX composition pattern matches the reference recipe story

Results are printed as a Markdown comparison table (baseline vs MCP, with delta) and saved to `packages/ui/evals/results/` (gitignored).


.github/vale/config/vocabularies/Backstage/accept.txt
```diff
@@ -588,3 +588,8 @@ zod
 Zolotusky
 zoomable
 zsh
+devDependencies
+eval
+Evals
+gitignored
+useTable
```
Rugvip, line 595: Only eval(s) is needed here really, gitignored is probably better written as "ignored by git", "listed in `.gitignore`", or such. Other should be using inline codeblocks.

.cursor/mcp.json
```diff
@@ -0,0 +1,8 @@
+{
+  "mcpServers": {
+    "backstage-ui-mcp": {
+      "type": "http",
+      "url": "http://localhost:6006/mcp"
```
Rugvip, line 5: Hmm, would prolly be better to have this be a `local` server at this point, as in MCP over stdio? Gets a bit tedious the have to run this manually

package.json
```diff
@@ -57,6 +57,7 @@
     "start:microsite": "cd microsite/ && yarn start",
     "start-backend": "echo \"Use 'yarn start example-backend' instead\"",
     "storybook": "storybook dev -p 6006",
+    "eval:ui": "tsx packages/ui/evals/run.ts",
```
Rugvip, line 60: Is this intended for consumption by consumers of the package as well? wonder if `evals` should be an entry point instead, or a binary, or even a separate package?

Also use `node --require @backstage/cli-node/config/nodeTransform.cjs`  rather than `tsx`

packages/ui/evals/.eslintrc.js
```diff
@@ -0,0 +1,14 @@
+// Standalone ESLint config for the evals subfolder.
+// root: true prevents ESLint from merging with packages/ui/.eslintrc.js,
+// which requires plugins that are not needed here.
+module.exports = {
```
Rugvip, line 4: Seems a bit strange that this would be needed, would expect to be able to bake it in or use a lower level parser API.

packages/ui/evals/modes/with-mcp.ts
```diff
@@ -0,0 +1,50 @@
...
+Return only the TSX code block, no explanation or markdown outside the code fence.
+
+You have access to the following @backstage/ui component documentation fetched from the Storybook MCP server.
+Use ONLY the props and patterns documented below — never invent or assume undocumented props.
+
+<component-documentation>
+${componentDocsSection}
+</component-documentation>`;
+
+  const code = await callLlm(systemPrompt, task.prompt);
```
Rugvip, line 48: Hmm, it looks to me like this might be possible to implement using an inverse approach, where the agent is directed via a skill or tool return/description to provide the code to an eval tool instead?

Needing to have the model call like this is troublesome in a bunch of different ways. One obvious bit is that the anthropic key requirement is quite tricky, especially since you don't get a key with an enterprise account. Another part is that the generated code won't adhere to whatever local instructions might exist. Both from a code style perspective, although some of that can probably be mitigated by further processing the generated code, but especially opinionated UI instructions like "prefer tables over card views" or "only use dialogs for confirmation, use a drawer for other scenarios".

packages/ui/evals/judge/import-check.ts
```diff
@@ -0,0 +1,51 @@
+import { parseCode, collectImports, collectJsxElements } from './ast';
+
+/**
+ * Checks that all @backstage/ui components used as JSX elements are imported
+ * from "@backstage/ui" and not from some other path.
+ *
+ * Score: proportion of @backstage/ui components correctly imported (0–100).
+ * Returns 100 if parsing fails (we penalise this elsewhere).
+ */
+export function scoreImportCorrectness(code: string): number {
```
Rugvip, line 10: This seems like something that's potentially better handled through specific instructions rather than a score? 🤔 Actually having that thought for more of these

packages/ui/evals/judge/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  code: string;
+  /** Raw MCP doc text per component name, for prop accuracy scoring. */
+  mcpDocs?: Map<string, string>;
+}
+
+/**
+ * Runs all applicable judge scorers against the generated code and returns
+ * a JudgeScores object with individual dimension scores and a composite.
+ */
+export function judge(input: JudgeInput): JudgeScores {
```
Rugvip, line 19: Something I've seen a lot in generation of BUI code is usage of intrinsic elements with custom styling using BUI tokens. This seems like something that'd help a lot in aligning that kind of behavior better.

Rugvip (changes requested):
Could of early Qs, still need to understand better where this is going 😁

cdedreuille:
I'll close this one for now and revisit it how to do better evals later on.

PR #33847, opened 2026-04-11
DialogApi: Add open() method, deprecate show/showModal

Rugvip:
The existing `show()` and `showModal()` methods on `DialogApi` render dialog chrome (a Material UI `<Dialog>`) as part of the implementation. This causes focus trap conflicts when the caller's content uses components from a different design library — for example, Backstage UI components inside the MUI dialog wrapper, or MUI components like `Select` inside a BUI dialog wrapper.

The new `open()` method renders the caller's content as-is, without any dialog chrome. The caller provides the full dialog component (overlay, backdrop, surface), making the API fully design-library-agnostic.

### Release plan

Usage of this API is still very low, so the plan is an aggressive rollout:

1. **This PR**: Deprecate `show`/`showModal` at the TypeScript level (`@deprecated`) and with runtime `console.warn` messages. Add `open()` as the replacement.
2. **Soon after**: Remove `show`/`showModal` from the `DialogApi` TypeScript interface, so callers get compile errors guiding them to `open()`.
3. **For a while longer**: Keep `show`/`showModal` functional in the `@backstage/plugin-app` implementation (they are re-implemented on top of `open()` with a MUI Dialog wrapper), so that any remaining JavaScript callers or late migrators continue to work at runtime.


Naycon (approved):
Awesome, thanks for this! 🙇

freben (approved):

Issue #33750, opened 2026-04-04
MCP OAuth discovery: authorization_servers URL, plugin-path .well-known 401, and missing WWW-Authenticate header


Thetanner:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
mcp
oauth-protected-resource
oauth
authorization_servers
well-known
401
WWW-Authenticate
dynamic
```

### 🗃️ Project Area

Auth

### 🔗 External Integration

N/A

### 📝 Description & Context

`@backstage/plugin-mcp-actions-backend` v0.1.9 has several OAuth discovery issues that prevent MCP clients (Claude Desktop, Cursor) from completing the RFC 9728 / RFC 8414 OAuth flow. These are in addition to the `getBaseUrl` bug reported in #33062.

> **Note:** The `getBaseUrl` → `getExternalBaseUrl` fix shown in our patch overlaps with #33092. This issue focuses on three *additional* bugs not addressed by that PR: the `authorization_servers` origin mismatch, plugin-path `.well-known` 401s, and missing `WWW-Authenticate` headers.

We are running Backstage 1.48.4 with `auth.experimentalDynamicClientRegistration.enabled: true` behind a reverse proxy. We have applied local yarn patches and middleware workarounds for all three issues described below.

**Bug 1: `authorization_servers` uses the full auth plugin URL instead of the origin**

The `/.well-known/oauth-protected-resource` handler returns:

```json
{
  "resource": "https://example.com/api/mcp-actions/v1",
  "authorization_servers": ["https://example.com/api/auth"]
}
```

Per [RFC 8414 Section 3](https://www.rfc-editor.org/rfc/rfc8414#section-3), clients construct the metadata URL by appending `/.well-known/oauth-authorization-server` to the authorization server value. This produces:

```
https://example.com/api/auth/.well-known/oauth-authorization-server  ->  404
```

The actual OAuth Authorization Server metadata is served at the root:

```
https://example.com/.well-known/oauth-authorization-server  ->  200
```

Claude Desktop fails with "Couldn't reach the MCP server" because it cannot discover the authorize, token, or registration endpoints.

**Fix:** `authorization_servers` should return the origin URL (`new URL(authBaseUrl).origin`) rather than the full plugin path.

**Our patch (applied to `dist/plugin.cjs.js`):**

```diff
[...]

PR #33855, opened 2026-04-13
feat: change resource url to comply with RFC 9728 Section 7.3

vincentrit:

Fixed the `.well-known/oauth-protected-resource` resource URL to comply with
[RFC 9728 Section 7.3](https://datatracker.ietf.org/doc/html/rfc9728#name-impersonation-attacks).

Partially addresses #33750

### DIFF
--- a/.changeset/shaky-years-doubt.md
+++ b/.changeset/shaky-years-doubt.md
@@ -1,0 +1,6 @@
+---
+'@backstage/plugin-mcp-actions-backend': minor
+---
+
+Fixed the `.well-known/oauth-protected-resource` resource URL to comply with
+[RFC 9728 Section 7.3](https://datatracker.ietf.org/doc/html/rfc9728#name-impersonation-attacks).
--- a/.changeset/shaky-years-doubt.md
+++ b/.changeset/shaky-years-doubt.md
@@ -3,4 +3,4 @@
 ---
 
 Fixed the `.well-known/oauth-protected-resource` resource URL to comply with
-[RFC 9728 Section 7.3](https://datatracker.ietf.org/doc/html/rfc9728#name-impersonation-attacks).
+[RFC 9728 Section 7.3](https://datatracker.ietf.org/doc/html/rfc9728#name-impersonation-attacks). Enabling dynamic resource paths.
--- a/.changeset/shaky-years-doubt.md
+++ b/.changeset/shaky-years-doubt.md
@@ -1,5 +1,5 @@
 ---
-'@backstage/plugin-mcp-actions-backend': minor
+'@backstage/plugin-mcp-actions-backend': patch
 ---
 
 Fixed the `.well-known/oauth-protected-resource` resource URL to comply with
--- a/plugins/mcp-actions-backend/src/plugin.ts
+++ b/plugins/mcp-actions-backend/src/plugin.ts
@@ -140,36 +140,42 @@ export const mcpPlugin = createBackendPlugin({
           // Protected Resource Metadata (RFC 9728)
           // https://datatracker.ietf.org/doc/html/rfc9728
           // This allows MCP clients to discover the authorization server for this resource
-          const serverSuffixes: string[] = [];
+          rootRouter.use(
+            '/.well-known/oauth-protected-resource/*',
+            async (req, res) => {
+              const requestedResourcePath = `/${req.params[0]}`;
 
-          if (serverConfigs && serverConfigs.size > 0) {
-            for (const key of serverConfigs.keys()) {
-              serverSuffixes.push(`/v1/${key}`);
-            }
-          } else {
-            serverSuffixes.push('/v1');
-          }
+              const serverSuffixes =
+                serverConfigs && serverConfigs.size > 0
+                  ? [...serverConfigs.keys()].map(key => `/v1/${key}`)
+                  : ['/v1'];
 
-          const [authBaseUrl, mcpBaseUrl] = await Promise.all([
-            discovery.getExternalBaseUrl('auth'),
-            discovery.getExternalBaseUrl('mcp-actions'),
-          ]);
+              const [authBaseUrl, mcpBaseUrlString] = await Promise.all([
+                discovery.getExternalBaseUrl('auth'),
+                discovery.getExternalBaseUrl('mcp-actions'),
+              ]);
 
-          const mcpBasePath = new URL(mcpBaseUrl).pathname;
+              const mcpBaseUrl = new URL(mcpBaseUrlString);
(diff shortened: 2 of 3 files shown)


benjdlambert:
Hey :wave: thanks for this!

We've been back and forth a bit on this, but from what I understand the `/v1` suffix isn't required. The spec only mandates that the `resource` field is an HTTPS URL with no fragment, and that it matches the identifier used to construct the `.well-known` URL.                                                                                                                                                                                          

`/v1` is an internal routing concern of the MCP server, not part of the resource identifier from what I understand from the spec.

I think that this only applies if the `/.well-known/oauth-protected-resource` is actually being used from `/api/mcp-actions/v1/.well-known/oauth-protected-resource` which it is not.

Are there clients that are not working without this change?

vincentrit:
Hey Ben, thanks a lot for the fast reply.

Valid point regarding the resource identifier. It ideally should not be tied to the current API version.

Unfortunately this currently does not work with the GeminiCLI. Their current approach is to treat the full URL path as the resource identifier:  
https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/mcp/oauth-utils.ts#L71-L88

> I think that this only applies if the /.well-known/oauth-protected-resource is actually being used from /api/mcp-actions/v1/.well-known/oauth-protected-resource which it is not.

In this case it is not about the prefix but rather the suffix. GeminiCLI appends the entire request path to `/.well-known/oauth-protected-resource` when constructing the discovery URL. From the client's perspective this effectively makes the versioned path part of the resource identifier.

One thing I am currently curious about is how clients are expected to deterministically determine the resource identifier in practice. The RFC explains how to construct the metadata URL from a resource identifier, but it seems less explicit about how clients are supposed to determine that identifier in the first place.

benjdlambert:
OK, so upon reading the spec again, I think that the real fix here is that we should update the routes from `/.well-known/oauth-protected-resource` to `/.well-known/oauth-protected-resource/api/mcp-actions/v1` and any another server that might exist on the `mcp-actions` backend plugin.

I think that should cover all use cases here, maybe perhaps something like this:

```ts                                                                                                                                                                                                                                         
  const serverSuffixes: string[] = [];                          
                                                                                                                                                                                                                                                
  if (serverConfigs && serverConfigs.size > 0) {                
    for (const key of serverConfigs.keys()) {                                                                                                                                                                                                   
      serverSuffixes.push(`/v1/${key}`);                                                                                                                                                                                                        
    }                                                                                                                                                                                                                                           
  } else {                            
    serverSuffixes.push('/v1');                                                                                                                                                                                                                 
  }                                                             
                         
  for (const suffix of serverSuffixes) {
    const mcpBasePath = `/api/mcp-actions${suffix}`;

    rootRouter.use(
      `/.well-known/oauth-protected-resource${mcpBasePath}`,
      async (_req, res) => {                                                                                                                                                                                                                    
          discovery.getExternalBaseUrl('auth'),                                                                                                                                                                                                 
          discovery.getExternalBaseUrl('mcp-actions'),          
        ]);                                                                                                                                                                                                                                     
                                      
        res.json({                                                                                                                                                                                                                              
          resource: `${mcpBaseUrl}${suffix}`,                   
          authorization_servers: [authBaseUrl],
        });                           
      },
    );
  }
  ```

That way it supports `/v1` and `/v1/catalog` if such a split was configured.

**Comment by vincentrit** [comment]
Sounds like a plan. I'll implement this tomorrow.

**Comment by vincentrit** [comment]
Hey, I just pushed an update. 

I moved the discovery call outside the loop to prevent unnecessary lookups and to enable dynamic `mcpBasePath` resolution. I checked if this causes any unpredictable behavior (like the auth URL not being available when the router is built), but I haven't noticed any issues. 

Is there a specific reason why the discovery call originally had to be inside the route handler?

File: plugins/mcp-actions-backend/src/plugin.ts [hunk 1]
```diff
@@ -140,19 +140,36 @@ export const mcpPlugin = createBackendPlugin({
...
+
+          if (serverConfigs && serverConfigs.size > 0) {
+            for (const key of serverConfigs.keys()) {
+              serverSuffixes.push(`/v1/${key}`);
+            }
+          } else {
+            serverSuffixes.push('/v1');
+          }
+
+          const [authBaseUrl, mcpBaseUrl] = await Promise.all([
```
benjdlambert: I think that we should move this inside that handler.

The reason is that this /could/ change over the lifetime of a single deployment. If these plugins are deployed separately and not on the same instance, then the other auth plugin could change it's `externalBaseUrl` and we'd want to resolve that every time.

It's not a big deal do this more often that what might be required, we're leaning on the fact that these lookups should be pretty cheap and even cached in some cases too.

.changeset/shaky-years-doubt.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-mcp-actions-backend': minor
```
benjdlambert:
```suggestion
'@backstage/plugin-mcp-actions-backend': patch
```

Would probably just make this a `patch` as the `mcp-actions-backend` plugin is `<1.x` and it's a non-breaking change.

benjdlambert (changes requested):
Thanks for updating this!

Just a few small comments and I think we're ready to 🚢

plugins/mcp-actions-backend/src/plugin.ts
vincentrit: Understood. One edge case to keep in mind is that if the mcp-actions discovery URL changes at runtime, our Express router won't automatically adapt since the path is bound during startup. This would lead to a mismatch between the URLs. Resolving the auth URL, however, could be moved inside the route handler without any issues.

One potential workaround would be to mount a wildcard route for /.well-known/oauth-protected-resource/* and handle the path validation and 404 logic dynamically inside the handler itself.

benjdlambert: Hmm, I think I prefer the other solution though of binding the routes up front though. 

The glob matching feels a little unfortunate to have to wildcard that and run additional logic for anything that matches that. 

It could also potentially cause conflicts with other routers, which I'd prefer not to potentially run into down the line.

Could we go back to defining the routes up front? :pray:

vincentrit: Sure, but keep in mind, that if the URL changes at runtime the URLs won't match anymore.
I am going to revert this for now.

benjdlambert (approved):
Nice, let's go with this! :pray:

vincentrit:
Thats cool, thanks!

PR #33867, opened 2026-04-13
backend-plugin-api: tighten internal any usage

Rugvip:

🧹

Replaces internal `any` type annotations with more precise types in the `backend-plugin-api` package:

- **`DepsToInstances` conditional type** — `ServiceRef<unknown, any, 'multiton'>` → `ServiceRef<unknown, 'root' | 'plugin', 'multiton'>`
- **`createServiceRef` implementation signature** — `any` scope param → `'root' | 'plugin'`
- **`isDatabaseConflictError`** — `(e as any)?.message` → `instanceof Error` check with `{ message?: unknown }` cast fallback

No public API changes — the remaining `any` usages in public API signatures (`Request<any, ...>`, `PermissionRule<any, ...>`) are intentional since they need to accept any concrete type parameter due to variance constraints.

### DIFF
--- a/packages/backend-plugin-api/src/services/utilities/database.ts
+++ b/packages/backend-plugin-api/src/services/utilities/database.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { isError } from '@backstage/errors';
+
 /**
  * Tries to deduce whether a thrown error is a database conflict.
  *
@@ -23,8 +25,7 @@
  *          known database engine
  */
 export function isDatabaseConflictError(e: unknown) {
-  const message =
-    e instanceof Error ? e.message : (e as { message?: unknown })?.message;
+  const message = isError(e) ? e.message : undefined;
 
   return (
     typeof message === 'string' &&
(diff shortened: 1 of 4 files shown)


packages/backend-plugin-api/src/services/utilities/database.ts
```diff
@@ -23,7 +23,8 @@
  *          known database engine
  */
 export function isDatabaseConflictError(e: unknown) {
-  const message = (e as any)?.message;
+  const message =
+    e instanceof Error ? e.message : (e as { message?: unknown })?.message;
```
freben: using `isError` resulting in an `ErrorLike` is fine here right?

freben (approved):

packages/backend-plugin-api/src/services/utilities/database.ts
Rugvip: yeah true let's switch this to isError

Rugvip: [automated] Addressed in 3b8c6d4

PR #33885, opened 2026-04-14
MembersListCard displayName -> title -> name

Parsifal-M:

Everywhere else on the group entity page we prefer the displayName, then the title, then the name.

But in the MembersLIstCard we do displayName or name.

Found this while doing something internally.

Of course the correct way is to not use `metadata.title` for groups and prefer the `profile.displayName` but we might as well add the fallback here as its everywhere else on the page.

### DIFF
--- a/.changeset/clean-weeks-beg.md
+++ b/.changeset/clean-weeks-beg.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-org': patch
+---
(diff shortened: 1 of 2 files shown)


Rugvip (approved):
Yep, thanks! 👍 

Probably needs entity presentation API, but that's for some other time

PR #33895, opened 2026-04-14
fix: use iovalkey Cluster for Valkey cluster mode

ganievs:

Fixed Valkey cluster mode to use `iovalkey`'s `Cluster` class instead of
`createCluster` from `@keyv/redis`. The previous implementation passed a
`@redis/client` `RedisCluster` instance to `@keyv/valkey`, which expects an
`iovalkey` `Cluster` instance. This caused the cluster client to not be
recognized correctly, as the two libraries have incompatible object models.

### DIFF
--- a/.changeset/fix-valkey-cluster-client.md
+++ b/.changeset/fix-valkey-cluster-client.md
@@ -1,0 +1,9 @@
+---
+'@backstage/backend-defaults': patch
+---
+
+Fixed Valkey cluster mode to use `iovalkey`'s `Cluster` class instead of
+`createCluster` from `@keyv/redis`. The previous implementation passed a
+`@redis/client` `RedisCluster` instance to `@keyv/valkey`, which expects an
+`iovalkey` `Cluster` instance. This caused the cluster client to not be
+recognized correctly, as the two libraries have incompatible object models.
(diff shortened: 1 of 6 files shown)


Teo-ShaoWei:
Kudos much @ganievs 🚀

I'm setting up a Valkey cluster and am encountering the same mismatch issue. The MR looks legit. @Rugvip sorry to summon you in person 🙇 Is there an issue with getting this merged?

For me, I can reproduce the issue via a Backstage app as a AWS ECS Task, and try to connect it to a AWS ElastiCache serverless Valkey cache. I got a `127.0.0.1:6379` → `ECONNREFUSED` which is very weird due to the localhost instead of AWS resource URI.

So the issue is that in `@keyv/valkey@^1.0.3`, the constructor checks for a pre-configured client like this:

```ts
if (
    typeof uri !== "string" &&
    "options" in uri && uri.options &&
    ("family" in uri.options || uri.isCluster)  // ← checks iovalkey's isCluster property
) {
    this._client = uri;   // recognized as pre-configured client
} else {
    // node-redis clusters fall here — uri.isCluster doesn't exist on node-redis
    this._client = new Redis(options.uri!, options);  // uri = undefined → localhost:6379
}
```

**Result:** `uri.isCluster` is `undefined` on node-redis clusters (it's an iovalkey property). The constructor falls to `else`, creates a standalone iovalkey `Redis` with `uri = undefined`, and iovalkey defaults to `127.0.0.1:6379` → `ECONNREFUSED`.

Changing the dependencies to a fully Valkey one is the right way to go 👍

Rugvip (approved):
Thank you! 🎉

Will trust that this is solid :shipit:

PR #33900, opened 2026-04-14
feat: move to scoped plugin routing

schultzp2020:

This PR introduces a router-agnostic routing architecture for Backstage's new frontend system, allowing individual plugins to use their preferred routing framework (React Router v6, React Router v7, TanStack Router, or others) while coexisting within a single Backstage application.

Today, all Backstage plugins are coupled to a single global `BrowserRouter` from `react-router-dom`. This makes it impossible for plugins to adopt newer routing frameworks independently, and creates a hard upgrade boundary where the entire application must migrate at once. This PR decouples plugins from that shared router by introducing a `RoutingContract` abstraction -- an observable-based contract that provides each plugin with its own scoped view of the URL.

### Design

The architecture has three layers:

**NavigationController** owns `window.history` and acts as the single source of truth for the browser URL. It provides scoped `RoutingContract` instances to plugins, each with a `basePath`, a `location$` observable (filtered to the plugin's URL prefix), and a `navigate()` function that translates relative plugin paths into absolute URLs before updating the browser.

**AppRouteSwitch** subscribes to `NavigationController.location$`, matches the current URL against a `RouteTable` using longest-prefix matching, and renders the matched plugin's page extension. The `PageBlueprint` delivers each plugin's `RoutingContract` via React context.

**Router adapters** bridge the `RoutingContract` into a specific router framework's internal context. The React Router v6 adapter, for example, provides `UNSAFE_LocationContext` and `UNSAFE_NavigationContext` so that existing `<Routes>`, `<Route>`, and `useNavigate()` calls continue to work without any changes to plugin code. Three adapter packages are included: `@backstage/plugin-react-router-v6-adapter`, `@backstage/plugin-react-router-v7-adapter`, and `@backstage/plugin-tanstack-router-adapter`.

### What changed

The core routing internals in `frontend-app-api` no longer depend on `react-router-dom` for route resolution. `RouteResolver`, `RouteTracker`, and `useRouteRef` now use standalone `generatePath` and `matchRouteRefs` implementations. The `BrowserRouter` in `AppRoot` has been replaced with a root-level `ScopedRouterProvider` backed by the `NavigationController`, so app chrome (sidebar, route tracker) and plugins all read from the same source of truth.

The `Link` component in `core-components` has been updated to detect cross-plugin navigation. When a link target is outside the current plugin's `basePath`, the component delegates to the framework-level `NavigationControllerApi` instead of `react-router-dom`, ensuring that navigation between plugins using different routers works correctly.

`useRouteRefParams` has been deprecated with a runtime warning in development mode, guiding developers toward their router's native `useParams` hook instead.

Two existing plugins (catalog and user-settings) have been wired to explicit router adapters as proof-of-concept migrations. A multi-router validation test suite confirms that plugins using different router frameworks can coexist, navigate between each other, and handle browser back/forward correctly.

### New packages

- `@backstage/plugin-react-router-v6-adapter` -- bridges RoutingContract to React Router v6
- `@backstage/plugin-react-router-v7-adapter` -- bridges RoutingContract to React Router v7
- `@backstage/plugin-tanstack-router-adapter` -- bridges RoutingContract to TanStack Router

### Backward compatibility

Existing plugins require no changes. The `ScopedRouterProvider` automatically sets up a React Router v6 context from the `RoutingContract`, so existing `<Routes>`, `<Route>`, `useNavigate()`, and `useLocation()` calls continue to work as before. Plugins can opt in to a different router at their own pace by swapping the adapter.

### DIFF
--- a/docs/superpowers/specs/plugin-migration-guide.md
+++ b/docs/superpowers/specs/plugin-migration-guide.md
@@ -1,0 +1,222 @@
+# Plugin Migration Guide: Router-Agnostic Routing
+
+This guide walks through the minimal changes required to migrate a Backstage plugin from the global React Router context to the new router-agnostic `RoutingContract` system. The scaffolder plugin is used as a representative example.
+
+## Overview
+
+The migration touches three areas:
+
+1. **Page extension** -- wrap the plugin's page component in a `ScopedRouterProvider` so it receives its own scoped routing context.
+2. **Parameter access** -- replace `useRouteRefParams` with `useParams` from `react-router-dom` (or equivalent), since route parameters are now resolved within the plugin's scoped router.
+3. **Tests** -- replace `MemoryRouter` / `renderInTestApp` router setup with `createMockContract` and `NestedRoutingContractProvider`.
+
+---
+
+## 1. Page Extension (Plugin Entry Point)
+
+The plugin's page extension currently mounts via `createRoutableExtension`, which relies on the global `BrowserRouter`. After migration, the `PageBlueprint` automatically provides a `RoutingContract` through `ScopedRouterProvider`, so the plugin's internal `<Routes>` and `<Route>` elements work within a scoped context.
+
+### Before
+
+```tsx
+// plugin.tsx
+import { createRoutableExtension } from '@backstage/core-plugin-api';
+import { rootRouteRef } from './routes';
+
+export const ScaffolderPage = scaffolderPlugin.provide(
+  createRoutableExtension({
+    name: 'ScaffolderPage',
+    component: () => import('./components/Router').then(m => m.Router),
+    mountPoint: rootRouteRef,
+  }),
+);
+```
+
+The `Router` component uses `<Routes>` from `react-router-dom` directly, relying on the global browser router:
+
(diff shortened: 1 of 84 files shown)


schultzp2020:
Note that this is most definitely a PoC still lol.

Rugvip (commented):
Just to capture the separate discussion here too: the main very high level question on this right now is whether we actually need the routing contract and different adapters, or whether it would be possible to lean more heavily on the browser built-ins for the bridging, especially the `history` object?

docs/superpowers/specs/plugin-migration-guide.md
```diff
@@ -0,0 +1,222 @@
+# Plugin Migration Guide: Router-Agnostic Routing
+
+This guide walks through the minimal changes required to migrate a Backstage plugin from the global React Router context to the new router-agnostic `RoutingContract` system. The scaffolder plugin is used as a representative example.
+
+## Overview
+
+The migration touches three areas:
```
milliehartnt123, line 7: Is this the actual order in which these changes need to be made?

PR #33905, opened 2026-04-14
Patch release: upgrade `tar` to v7 for v1.46.x

jonkoops:
Backports https://github.com/backstage/backstage/pull/32471

This release fixes a dependency on the deprecated `tar` v6 by upgrading to `tar` v7.

This is a partial patch — only the packages whose latest published patch version falls within the v1.46.x release line are bumped here:

- `@backstage/backend-defaults` `0.14.1` → `0.14.2`

The remaining affected packages (`@backstage/cli`, `@backstage/repo-tools`, `@backstage/plugin-scaffolder-backend`, `@backstage/plugin-scaffolder-node`) have their latest patch versions published in other Backstage releases and are covered by separate patch PRs against those branches.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "root",
-  "version": "1.46.5",
+  "version": "1.46.6",
   "backstage": {
     "cli": {
       "new": {
--- a/packages/backend-defaults/CHANGELOG.md
+++ b/packages/backend-defaults/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @backstage/backend-defaults
 
+## 0.14.2
+
+### Patch Changes
+
+- Upgraded `tar` dependency to v7.
+
 ## 0.14.1
 
 ### Patch Changes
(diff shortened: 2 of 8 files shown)


jonkoops:
The test failures on this PR are pre-existing on the `patch/v1.46.0` branch and unrelated to the `tar` upgrade. The same failures can be seen on the [base branch CI runs](https://github.com/backstage/backstage/actions/runs/21530321596/job/62044388451).

The `MockDirectory.test.ts` failure occurs because `resolveRealPath` in `isChildPath.ts` only catches `ENOENT`, not `EACCES`. On the CI runner `/root` exists but is inaccessible, so `realpathSync('/root/a.txt')` throws `EACCES` instead of falling through to the parent-walking logic.

The `githubPullRequest.test.ts` failure occurs because the "with broken symlink" test creates a symlink pointing to `../../nothing/yet`, which escapes the workspace. After the security hardening backport (`66e08b08f94a`), `isChildPath` now follows symlinks to their real targets, causing `resolveSafeChildPath` to correctly reject the escaping symlink. The test was not updated to account for this.

Both issues were introduced by the security fix backported to `patch/v1.46.0` in https://github.com/backstage/backstage/commit/66e08b08f94a and are reproducible on the base branch without any of our changes.

Rugvip (approved):
Thanks! 👍

PR #33908, opened 2026-04-14
feat(techdocs): make sidebar positioning configurable via CSS custom properties

Naycon:

The TechDocs sidebar layout at tablet breakpoints hardcodes the Backstage app sidebar widths (`224px` pinned, `72px` collapsed) to calculate its positioning. Apps with custom sidebar widths have no way to adjust this, causing the TechDocs navigation to render on top of or behind other content.

This replaces the hardcoded values with CSS custom properties that inherit through the shadow DOM boundary:

- `--techdocs-sidebar-closed-offset-pinned` (default: `224px`) — closed position offset when app sidebar is pinned
- `--techdocs-sidebar-closed-offset-collapsed` (default: `72px`) — closed position offset when app sidebar is collapsed
- `--techdocs-sidebar-open-translate` (default: `16rem`) — slide distance when the drawer is opened

Example usage for an app with a wider sidebar:
```css
:root {
  --techdocs-sidebar-closed-offset-pinned: 300px;
  --techdocs-sidebar-closed-offset-collapsed: 100px;
}
```

Or to fully hide the sidebar off-screen and have it overlay when opened:
```css
:root {
  --techdocs-sidebar-closed-offset-pinned: 0px;
  --techdocs-sidebar-closed-offset-collapsed: 0px;
}
```

### DIFF
--- a/.changeset/free-ways-flow.md
+++ b/.changeset/free-ways-flow.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-techdocs': patch
+---
+
+Made the TechDocs sidebar positioning at tablet breakpoints configurable via CSS custom properties, allowing apps with custom sidebar widths to override the defaults. The available properties are `--techdocs-sidebar-closed-offset-pinned`, `--techdocs-sidebar-closed-offset-collapsed`, and `--techdocs-sidebar-open-translate`.
--- a/.patches/pr-33908.txt
+++ b/.patches/pr-33908.txt
@@ -1,0 +1,1 @@
+Make TechDocs sidebar positioning configurable via CSS custom properties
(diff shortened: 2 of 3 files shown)


Rugvip (approved):
Alright, although is it worth exploring ways to make this unnecessary in the first place? 🤔 For example we may want the full page wrapper from BUI? Mostly thinking something more generic would be possible because no other plugin needs this special handling, but at the same time I know techdocs is very special 😅

Naycon:
> Alright, although is it worth exploring ways to make this unnecessary in the first place? 🤔 For example we may want the full page wrapper from BUI? Mostly thinking something more generic would be possible because no other plugin needs this special handling, but at the same time I know techdocs is very special 😅

@Rugvip yes, most definitely! This is just a stopgap solution to give some degree of control. The real fix is looking into both design and code and restructuring it so it either doesn't need this behavior at all, or that it positions inside of its own content area. Just don't have the time to do that proper fix right now unfortunately.

PeaWarrior (approved):

PR #33920, opened 2026-04-15
ui: inline `ColumnSize` and `ColumnStaticSize` types from `@react-types/table`

Rugvip:
The `ColumnSize` and `ColumnStaticSize` types were imported from `@react-types/table`, which is not a direct dependency of `@backstage/ui` — it was only available as a transitive dependency that can break across version bumps.

Since these are simple type aliases (`ColumnStaticSize = number | `${number}` | `${number}%`` and `ColumnSize = ColumnStaticSize | `${number}fr``), this replaces the import with locally defined types. They are also now explicitly exported from the package.

### DIFF
--- a/.changeset/fix-react-types-table-imports.md
+++ b/.changeset/fix-react-types-table-imports.md
@@ -1,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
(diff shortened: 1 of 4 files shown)


benjdlambert (approved):

Rugvip:
https://github.com/backstage/backstage/pull/33918 instead

PR #33921, opened 2026-04-15
beps: add BEP-0014 Connection Service

Rugvip:

This is a new BEP proposing a **Connection Service** for Backstage - a centralized system for configuring and querying connections to external services. It replaces the current `integrations` configuration and provides a shared foundation that the entire plugin ecosystem can build on.

The core problem: every plugin that talks to an external service defines its own config for hosts, credentials, and endpoints. There's no shared registry, no cross-plugin reuse, and no single place for adopters to manage their connections. The existing `integrations` config only covers a handful of SCM providers.

The proposal introduces a `backend.connections` config key, a `coreServices.connections` backend service, and a `@backstage/connections-node` package where all built-in connection types are defined centrally. Connections are pure static data - no API clients, no token exchange. Existing credential providers continue to exist as a separate layer on top.

Key design decisions:
- A centralized registry of all connections known to the plugin ecosystem, maintained collaboratively in `@backstage/connections-node`
- Connections are pure data defined with Zod schemas, each with one or more auth methods as first-class primitives with their own schemas and selection logic
- Plugins must declare consumed connection types via `registerConnection`
- The `find` method enforces that callers handle the auth methods they encounter

Looking for feedback on the overall direction and any blind spots in the design.


beps/0014-connection-service/README.md
```diff
@@ -0,0 +1,603 @@
...
+backend:
+  connections:
+    - type: github
+      host: github.com
+      auth:
+        # Only the catalog plugin can select this auth method
+        - method: token
+          token: ${GITHUB_CATALOG_TOKEN}
+          match:
+            plugins: [catalog]
```
freben, line 140:
```yaml
match:
  plugin: { $in: ['catalog'] }
```
perhaps

beps/0014-connection-service/README.md
```diff
@@ -0,0 +1,603 @@
...
+
+### Configuration Format
+
+Connections are configured as a flat array under `backend.connections` in `app-config.yaml`. The `backend` namespace is the natural home since connections are a backend concern, and avoids adding another top-level configuration key. Plugins should always consume connections through the connection service rather than reading the config directly. To enforce this, the framework may introduce restrictions in the config reading service that prevent plugins from accessing `backend.connections` directly, ensuring that all connection access goes through the connection service and its validation, matching, and scoping logic.
+
+Each entry has a `type` field that identifies the kind of external service:
+
+```yaml
+backend:
+  connections:
```
freben, line 79: This clearly only deals with auth at least at this point. Not sure the generic naming and hierarchy placement is appropriate then. `auth.connections` or `backend.auth.connections` or even `backend.connections.auth` might be considered if this holds.

When seeing the heading and start of the BEP I thought this was gunning for a wider reusability concept. Something that could for example cover databases too, so you could define three different db connections and then just reference those in other parts of the config. "These are the connection params to use for admin level CREATE DATABASE, here's what the service should connect as, ..."

Or maybe pointing to LDAP _both_ for the purposes of auth as well as catalog ingestion. Well OK that last thing might actually pan out - but then you'd maybe start considering naming of connections, to be able to refer to them.

```yaml
connections:
  ldap:   # here's where the typing happens
    - name: production
      target: ldaps://ds.example.net
      bind:
        dn: software-catalog-bot
        secret: ${SOFTWARE_CATALOG_BOT_SECRET}

catalog:
  providers:
    ldap:
      - connection: ldap:production
        moreProviderSpecificProperties: here
```

at that point you also start thinking about how universally applicable the `match` concept would be. Does it always make sense to put on the "target object"? Or sometimes on the "source", or even in a "join table" so to speak?

To be clear, this is mostly me thinking about a bigger picture, and if the exact scope presented in the BEP is the intended one, then the above probably doesn't really apply.

beps/0014-connection-service/README.md
```diff
@@ -0,0 +1,603 @@
...
+
+The `createConnectionType` helper captures the full definition: base config validation, auth methods, output shape, and URL matching. It uses Zod schemas as the source of truth for both base fields and auth methods, and produces a `ConnectionType` object that the registry uses internally.
+
+**What each piece does:**
+
+- **`type`** - the discriminator string used in config and queries. The framework adds `type` to the output automatically.
+- **`configSchema`** - a Zod schema for the base connection fields shared across all auth methods (e.g. `host`, `apiBaseUrl`). The pre-transform shape defines the config input; `.transform()` derives computed defaults. The `host` field is required on input but is not included in the output type, it is used internally for matching only. Consumers should use base URL fields like `apiBaseUrl` instead.
```
aramissennyeydd, line 279: Will this inform the inline json schema we use today for config validation or is this primarily runtime focused?

beps/0014-connection-service/README.md
```diff
@@ -0,0 +1,603 @@
...
+
+The `match` block is designed to evolve. Future selection criteria can be added alongside `plugins` without changing auth method schemas or the overall config structure.
+
+**Config merging.** Arrays in Backstage's config system are replaced wholesale when merging across config sources, they do not merge element-by-element. This means that if `backend.connections` is defined in multiple config files, the last source wins. In practice this is acceptable because different environments (local dev, staging, production) typically define entirely different sets of connections. For the case where base config and secrets need to be split, the recommended approach is to use environment variable substitution within a single config file, or to use the `$include` directive.
```
aramissennyeydd, line 147: Should we support some form of array merging or multi-item `$include`? Trying to think through a couple of our use cases that would be impacted here.

1. User wants to provide a localhost version of a proxy.
2. User wants to use PAT for local dev GH connection but default to other `app-config.yaml` credentials.

aramissennyeydd (commented):
Nice! Couple of questions

beps/0014-connection-service/README.md
Rugvip: Idea is for this to be the source of truth at least. Remains to be seen whether this would be fed into config schema as well or not, we could certainly generate it from these definitions at least.

beps/0014-connection-service/README.md
Rugvip: > default to other `app-config.yaml` credentials.

This is a tricky one for me to gauge how important it actually is. You're of course already able to do this at some level using env var interpolation, and we could have connections gracefully degrade on missing config in development.

Still though, agree with the overall concern. I could definitely see `backend.connections: { [type]: config[] }` as a route too

freben (approved):
Approving, leaving some old comments from before the design was actually settled just for context if perhaps interesting

beps/0014-connection-service/README.md
Rugvip: Maaaaaaaaybe :p, let's see

PR #33945, opened 2026-04-16
chore: add missing transitive dependencies

freben:

Add dependencies that were only available transitively through other packages but were being imported directly. These are phantom dependencies that happen to work because another declared dependency pulls them in, but should be explicitly declared for correctness and resilience against upstream changes.

**Changes:**

- `csstype` added to `@backstage/core-components` dependencies — value import of `CSS.Properties` in `Lifecycle.tsx`
- `copy-to-clipboard` added to `@backstage/core-components` devDependencies — imported and mocked in LogViewer test files
- `@storybook/react-vite` added to `@backstage/ui` devDependencies — type imports (`Meta`, `StoryObj`, `StoryFn`) in story files
- `react-aria-components` added to `@backstage/plugin-notifications` dependencies — type import of `Key` in `NotificationsFilters.tsx`, transitive via `@backstage/ui`
- `@octokit/core` added to `@backstage/plugin-scaffolder-backend-module-github` dependencies — `OctokitOptions` type import used in public API, transitive via `octokit`
- `@rjsf/utils` added to `example-app-legacy` dependencies — type import of `FieldValidation` in scaffolder extensions


packages/app-legacy/package.json
```diff
@@ -77,6 +77,7 @@
     "@material-ui/icons": "^4.9.1",
     "@material-ui/lab": "4.0.0-alpha.61",
     "@octokit/rest": "^19.0.3",
+    "@rjsf/utils": "5.24.13",
```
Rugvip: Seems unfortunate to pin it?

Rugvip (approved):
Although also yes 👍

packages/app-legacy/package.json
freben: Interestingly, everybody else also pins it. But that's still a good insight, let me make them into ranges starting from the pinned one

freben: No i changed my mind -we have already kept rjsf intentionally pinned (not just utils, but the whole library) because it was tricky in terms of staying stable over bumps. So this may not be the time to change that

PR #33970, opened 2026-04-17
ui, app-visualizer: pin react-aria-components for v1.46

Rugvip:


freben (approved):

Rugvip:
Failing tests are unrelated

PR #33979, opened 2026-04-17
ui: pin react-aria-components for 1.49

Rugvip:


yarn.lock
```diff
@@ -17112,7 +17117,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@react-types/calendar@npm:^3.8.3":
+"@react-types/calendar@npm:^3.8.3, @react-types/calendar@npm:~3.8.3":
```
freben, line 17120: more of these remaining carets

freben (approved):

yarn.lock
Rugvip: yep this time it's transitive deps

Naycon (approved):
Looks all right I think! There are a couple of places where we import directly from `'react-stately'` in app and ui, but I believe those might be covered by the included dependency on `'react-stately/*'`...

PR #33983, opened 2026-04-17
ui: pin react-aria-components for 1.47

Rugvip:


Rugvip:
Redo of https://github.com/backstage/backstage/pull/33972, got trigger happy with that one

freben (approved):

freben (approved):

PR #33984, opened 2026-04-17
fix(ui): keep Tabs active indicator visible on uncontrolled tabs

Naycon:

Fixes a bug where the active tab indicator on uncontrolled `<Tabs>` (no `href` props) briefly appears on page load and then disappears, only coming back on hover.

React Aria's `CollectionBuilder` renders `TabList`'s children into both a hidden collection-building tree and the real DOM. The hidden `TabsIndicators` instance sits outside the `TabListStateContext` provider, so its `state` is `null` — its `updateCSSVariables` effect takes the `else` branch and writes `--active-tab-opacity: 0` to the `tabsRef` DOM element that the real instance also writes to. Under the right render ordering, the hidden write lands after the real instance's `opacity: 1` and hides the indicator.

Fix: guard `updateCSSVariables` with an early return when `state == null` so the hidden instance never writes to the shared DOM element.

### DIFF
--- a/.changeset/fix-tabs-active-indicator-disappearing.md
+++ b/.changeset/fix-tabs-active-indicator-disappearing.md
@@ -1,0 +1,7 @@
+---
+'@backstage/ui': patch
+---
+
+Fixed an issue where the active tab indicator would disappear shortly after page load for uncontrolled Tabs.
+
+**Affected components:** Tabs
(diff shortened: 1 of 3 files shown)


Rugvip (approved):
patch?

Naycon:
This should replace the need for #33835.

Naycon:
> patch?

@Rugvip if we're creating one then maybe yes. I think some special circumstances are need in order to trigger this, did not manage to reproduce it in storybook, but we have seen it in the wild. Can list it for addition in the next one.

Naycon:
Added a patch entry for it now.

PR #33985, opened 2026-04-17
fix(scaffolder): migrate EntityPicker to use queryEntities API

elaine-mattos:

Hi! Today we had a user that, while trying to rollout a template with the OwnedEntityPicker, bumped into a "414 Request-URI Too Large" error. 
I decided to switch from `getEntities` to `queryEntities` in the EntityPicker.  I also added a `convertFilterToPredicate` function to transform the filter format to the predicate format that `queryEntities` expects.

Tested it and works fine now! This is an internal implementation change and the component's public API is unchanged. All existing filter formats (single/array, exists, allowedKinds) are supported via the new `convertFilterToPredicate` function. I also updated the tests.

Please let me know your thoughts! :)


freben (changes requested):
Hey! Check out the `queryEntitiesByPredicate` contract - it supports both `filter` AND the new `query` syntax, and translates internally. So you don't have to do it at all.

elaine-mattos:
Hi @freben , thanks for the feedback, I have completely missed that. 

I had to add an empty `query` object to force POST routing (to avoid URL length limits), since `queryEntities` only routes to POST when `request.query` is there...

What do you think?

elaine-mattos:
Hi @freben,

What do you think about it now? We’ve been running this as a patch in our instance because some members of the enabling team belong to a large number of groups by design, so…

elaine-mattos:
not stale...

elaine-mattos:
not stale, will take another look at it soon. (rebase, etc)

elaine-mattos:
Closing this one since https://github.com/backstage/backstage/pull/32637 fixes the issue

PR #33986, opened 2026-04-17
feat(ui): add `isPending` prop and deprecate `loading`

Naycon:

Add `isPending` prop to Alert, Button, ButtonIcon, Table, and TableRoot, aligning with React Aria naming conventions. The `loading` prop is deprecated but still supported as an alias.

**What changed:**

- Public component types gain `isPending?: boolean`; `loading` marked `@deprecated`
- CSS selectors moved from `data-loading` to `data-ispending`; `data-loading` still emitted for backward compat
- Internal Table hooks renamed `loading` → `isPending`
- `useTable` returns both `isPending` and `loading` on `tableProps` for backward compat
- Stories and docs-ui updated to use `isPending`

### DIFF
--- a/docs-ui/src/app/components/alert/props-definition.ts
+++ b/docs-ui/src/app/components/alert/props-definition.ts
@@ -10,31 +10,45 @@ export const alertPropDefs: Record<string, PropDef> = {
     values: ['info', 'success', 'warning', 'danger'],
     responsive: true,
     default: 'info',
+    description:
+      'Visual status of the alert, which determines color and default icon.',
   },
   icon: {
     type: 'enum',
     values: ['boolean', 'React.ReactElement'],
     responsive: false,
+    description:
+      'Set to true to show the default status icon, or pass a custom icon element. Set to false to hide the icon.',
+  },
+  isPending: {
+    type: 'boolean',
+    default: 'false',
+    description:
+      'Replaces the icon with a spinner to indicate a pending operation.',
   },
   loading: {
-    type: 'enum',
-    values: ['boolean'],
-    responsive: false,
+    type: 'boolean',
+    default: 'false',
+    description: 'Deprecated. Use `isPending` instead.',
   },
   title: {
     type: 'enum',
     values: ['React.ReactNode'],
     responsive: false,
+    description: 'Primary message displayed in the alert.',
   },
   description: {
     type: 'enum',
     values: ['React.ReactNode'],
     responsive: false,
+    description: 'Additional detail shown below the title.',
   },
   customActions: {
     type: 'enum',
     values: ['React.ReactNode'],
     responsive: false,
+    description:
+      'Custom action buttons displayed on the right side of the alert.',
   },
   m: {
     type: 'enum',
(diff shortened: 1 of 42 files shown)


docs-ui/src/app/components/alert/props-definition.ts
```diff
@@ -10,31 +10,45 @@ export const alertPropDefs: Record<string, PropDef> = {
...
+    description:
+      'Set to true to show the default status icon, or pass a custom icon element. Set to false to hide the icon.',
+  },
+  isPending: {
+    type: 'boolean',
+    default: 'false',
+    description:
+      'Replaces the icon with a spinner to indicate a pending operation.',
   },
   loading: {
```
Rugvip, line 29: might be nice to be able to mark as deprecated here and utilize somehow, just a really quite unrelated thought 😅

packages/ui/src/components/Button/Button.tsx
```diff
@@ -55,15 +55,17 @@ export const Button = forwardRef(
   (props: ButtonProps, ref: Ref<HTMLButtonElement>) => {
     const { ownProps, restProps, dataAttributes } = useDefinition(
       ButtonDefinition,
-      props,
+      // Merge deprecated `loading` into `isPending` so data attributes and
+      // internal logic only need to check a single prop.
+      { ...props, isPending: props.isPending || props.loading },
```
Rugvip: Prolly wanna `??` here instead I'm thinking? `isPending={false} loading={true}` should probably still be `false`? 🤔

Rugvip (approved):
👍, a thought, but pre-approved

packages/ui/src/components/Button/Button.tsx
Naycon: I was thinking we should just let whichever is true win, and that we should never expect both to be set at the same time.

My reasoning for letting `true` win is that if either prop is indicating something is loading/pending, you're probably also assuming interaction is blocked and you're not ready to deal with it (stale/no data, async op in progress, e.g.). Defaulting to `true` felt like a safer choice. But really, don't use both at the same time! 😅

docs-ui/src/app/components/alert/props-definition.ts
Naycon: Yes, that's a good idea. Something to improve down the line for sure!

packages/ui/src/components/Table/components/TableRoot.tsx
```diff
@@ -28,14 +28,16 @@ import { TableRootProps } from '../types';
 export const TableRoot = (props: TableRootProps) => {
   const { ownProps, restProps, dataAttributes } = useDefinition(
     TableDefinition,
-    props,
+    // Merge deprecated `loading` into `isPending` so data attributes and
+    // internal logic only need to check a single prop.
+    { ...props, isPending: props.isPending || props.loading },
   );
```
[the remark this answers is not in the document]
Naycon: By design.

packages/ui/src/components/Button/Button.tsx
Naycon: Opted for something slightly fancier, to make sure the case where you pass `isPending={false}` results in `data-ispending="false"`. Before the change it would've been removed because of the result of the `||` operation. Truthy value still wins out though.

cdedreuille (approved):

PR #33988, opened 2026-04-17
catalog-client: use POST endpoint for getEntities

freben:

Reimplement `CatalogClient.getEntities()` to use `POST /entities/by-query` instead of `GET /entities`, avoiding request URL size limits when filters or other parameters are large.

The problem with `getEntities` is that it uses the GET verb and hits request size limits for large requests. The `POST /entities/by-query` endpoint already exists and supports the same semantics (filtering, ordering, pagination, field selection), so this change delegates to it internally via the existing private `queryEntitiesByPredicate()` method.

**Key details:**

- The public `GetEntitiesRequest` / `GetEntitiesResponse` types and behavior are completely unchanged
- The `EntityFilterQuery` key-value filter format is converted to `FilterPredicate` format using the existing `convertFilterToPredicate` utility
- `order` maps to `orderFields`, `fields`/`offset`/`limit` map directly
- When the legacy `after` cursor parameter is provided, the method falls back to the original `GET /entities` endpoint since the cursor formats are incompatible
- API reports show no changes, confirming the public API surface is untouched

### DIFF
--- a/packages/catalog-client/src/CatalogClient.ts
+++ b/packages/catalog-client/src/CatalogClient.ts
@@ -195,39 +195,27 @@ export class CatalogClient implements CatalogApi {
     request?: GetEntitiesRequest,
     options?: CatalogRequestOptions,
   ): Promise<GetEntitiesResponse> {
-    const {
-      filter = [],
-      fields = [],
-      order,
-      offset,
-      limit,
-      after,
-    } = request ?? {};
-    const encodedOrder = [];
-    if (order) {
-      for (const directive of [order].flat()) {
-        if (directive) {
-          encodedOrder.push(`${directive.order}:${directive.field}`);
-        }
-      }
+    const { filter, fields, order, offset, limit, after } = request ?? {};
+
+    // The `after` cursor is specific to the legacy GET /entities pagination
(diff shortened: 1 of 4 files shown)


packages/catalog-client/src/CatalogClient.ts
```diff
@@ -195,39 +195,27 @@ export class CatalogClient implements CatalogApi {
...
-      after,
-    } = request ?? {};
-    const encodedOrder = [];
-    if (order) {
-      for (const directive of [order].flat()) {
-        if (directive) {
-          encodedOrder.push(`${directive.order}:${directive.field}`);
-        }
-      }
+    const { filter, fields, order, offset, limit, after } = request ?? {};
```
aramissennyeydd, line 198: Is `limit` still different between the 2 endpoints by default? iirc getEntities doesn't have a default limit which has been an issue for a long time and [queryEntities enforces a default of 200](https://github.com/backstage/backstage/blob/2ce3506e93a12284d31fd0c0ba6e07758dafa618/plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts#L250).

PR #33991, opened 2026-04-17
fix(SecureTemplater): return dispose function to clean up secure temp…

jtbry:

Added a `dispose` method to the return of `SecureTemplater.loadRenderer` to allow for the explicit disposal of the related isolated-vm and freeing of VM context. Currently observing a problem where Scaffolder Task execution causes memory usage increases that are never reclaimed. Theory is these VMs are being missed/taking too long to be caught by GC so explicit freeing has solved the observed problem in my team's lower environments with a node_modules patch.


freben (approved):
Looks good!

PR #33993, opened 2026-04-18
fix(scheduler): handle setTimeout overflow for long sleep durations

freben:

Node.js `setTimeout` uses a 32-bit signed integer for the delay parameter. Values larger than 2^31-1 ms (~24.8 days) cause the callback to fire immediately instead of after the intended delay. This was reported as causing scheduler tasks with long intervals to start early.

The fix chunks the wait into segments of at most 2^30 ms (~12.4 days), looping until the full requested duration has elapsed. The abort signal is checked so that cancellation still works promptly.

### DIFF
--- a/packages/backend-defaults/src/entrypoints/scheduler/lib/PluginTaskSchedulerImpl.test.ts
+++ b/packages/backend-defaults/src/entrypoints/scheduler/lib/PluginTaskSchedulerImpl.test.ts
@@ -150,13 +150,13 @@ describe('PluginTaskManagerImpl', () => {
           id: 'task1',
           timeout: Duration.fromMillis(5000),
           frequency: Duration.fromObject({ years: 1 }),
-          initialDelay: Duration.fromObject({ years: 1 }),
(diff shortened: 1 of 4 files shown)


packages/backend-defaults/src/entrypoints/scheduler/lib/PluginTaskSchedulerImpl.test.ts
```diff
@@ -150,13 +150,13 @@ describe('PluginTaskManagerImpl', () => {
           id: 'task1',
           timeout: Duration.fromMillis(5000),
           frequency: Duration.fromObject({ years: 1 }),
-          initialDelay: Duration.fromObject({ years: 1 }),
+          initialDelay: Duration.fromObject({ seconds: 60 }),
```
freben, line 153: For the reviewer: Interestingly, this test just accidentally worked before, due to the bug that's fixed in here.

Rugvip (approved):
:shipit:

PR #33997, opened 2026-04-19
feat(ui): add description, tags, and metadata props to Header

cdedreuille:
Improves the `Header` component in `@backstage/ui` as part of [ACT-355](https://linear.app/spotify/issue/ACT-355/header-improvements).

### New props

- **`tags`** — renders a row of text or link items above the title, separated by 3×3px circle dividers. Tags with an `href` render as a secondary-coloured `Link`; tags without render as plain `Text`.
- **`description`** — renders a markdown string below the title. Only plain text and inline links are supported (bold, italic, and block-level markdown are intentionally excluded). Anchors use BUI's `Link` component.
- **`metadata`** — renders key-value pairs below the description with 20px gaps between items. The label is secondary-coloured, the value accepts `string | ReactNode`.
- **`breadcrumbs`** — deprecated (functionality preserved, will be removed in a future release, no breaking change).

### New components

- **`HeaderMetadataUsers`** — renders a list of users as a metadata value. A single user shows an avatar with their name beside it. Multiple users show a row of avatars with names revealed on hover via tooltip. Each user optionally accepts an `href` — when provided the avatar becomes a link and the name renders as a primary `Link` (no underline at rest, underline on hover via `standalone`).
- **`HeaderMetadataStatus`** — renders a single status indicator as a metadata value. An 8×8 coloured circle (mapped to BUI status tokens: `danger`, `warning`, `success`, `info`) is shown alongside a label. Optionally accepts an `href` to make the label a link.

### Other changes

- Header root replaced from `<Container>` to a plain `<div>` — the component now spans its full parent width with no max-width constraint.
- Header internal `gap` updated to `var(--bui-space-3)`.
- `HeaderMetadataUsers` and `HeaderMetadataStatus` exported from `@backstage/ui` alongside the new `HeaderMetadataStatusItem` type.

### Stories added

`WithDescription`, `WithTags`, `WithMetadata`, `WithMetadataUsers`, `WithMetadataUsersNoLinks`, `WithMetadataStatus`, `WithDescriptionTagsAndMetadata`, and `WithEverything` updated to include all new props.

### Docs updated

- New examples for Tags, Description, Metadata with users, and Metadata with status.
- `HeaderMetadataUsers` API reference table added to the docs page.
- Deprecated `breadcrumbs` prop marked in the API reference table.


packages/ui/src/components/Header/Header.tsx
```diff
@@ -24,18 +24,96 @@ import { Container } from '../Container';
...
+    breadcrumbs,
+    description,
+    tags,
+    metadata,
+  } = ownProps;
 
   return (
     <Container className={classes.root}>
+      {tags && tags.length > 0 && (
+        <div className={classes.tags}>
```
Naycon: This whole section needs better semantic markup for a11y. You need associate the key-value pairs together in markup. I think a dl/dt/dd set might be what you want, but maybe have a think if that corresponds well to what you're doing here or if something else is better suited.

packages/ui/src/components/Header/Header.tsx
```diff
@@ -24,18 +24,96 @@ import { Container } from '../Container';
...
+ * form `[label](href)` as an array of React nodes (strings and Link elements).
+ * Links with unsafe URL schemes (javascript:, vbscript:, data:) are rendered
+ * as plain text instead.
+ *
+ * We intentionally avoid `react-markdown` here: that package is ESM-only
+ * (v8+), which breaks Jest in Node-role packages that transitively import
+ * `@backstage/ui` (e.g. via `core-app-api`). Since the Header description only
+ * needs inline link support, a small regex-based parser is sufficient and keeps
+ * this package free of ESM dependencies.
+ */
```
Naycon, line 33: This is slightly scary. I don't feel versed enough in xss attacks to judge if this is good enough coverage or if there are cases/quirks that'd let you work around this setup. Just calling it out, I'd rely on a well known and battle hardened lib to be sure. If you forgive my assumption, the last comment feels a little like an AI agent trying to convince itself it made the right choice, and I can't really say I'm confident it (or you in case you did this by hand) is right. :)

packages/ui/src/components/Header/Header.tsx
```diff
@@ -61,6 +139,31 @@ export const Header = (props: HeaderProps) => {
...
+        <Text
+          variant="body-medium"
+          color="secondary"
+          className={classes.description}
+        >
+          {renderInlineMarkdown(description)}
+        </Text>
+      )}
+      {metadata && metadata.length > 0 && (
+        <div className={classes.metaRow}>
```
Naycon: Same as above, this needs better markup to semantically group the content in a way that makes it understandable for a screen reader. Currently, the value of one part of the meta data is read continuously together with the label of the next without pause, making this a pretty confusing experience.

packages/ui/src/components/Header/HeaderMetadataStatus.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { Link } from '../Link';
+import styles from './HeaderMetadataStatus.module.css';
+
+/**
+ * Displays a single status indicator as a coloured dot with a label inside a
+ * Header metadata value. Optionally renders the label as a link when href is provided.
+ *
+ * @public
+ */
+export const HeaderMetadataStatus = ({
```
Naycon, line 28: This component could also use better semantic markup I think, and perhaps some labels as well.

packages/ui/src/components/Header/HeaderMetadataUsers.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <Avatar
+        src={user.src ?? 'data:,'}
+        name={user.name}
+        size="small"
+        purpose="decoration"
+      />
+    );
+    return (
+      <div className={styles.single}>
+        {user.href ? (
```
Naycon: This might read easier if you just combine this shorthand with the one below into one that just return a fragment containing both avatar + link.

packages/ui/src/components/Header/HeaderMetadataUsers.tsx
```diff
@@ -0,0 +1,106 @@
...
+          </Link>
+        ) : (
+          <Text variant="body-medium">{user.name}</Text>
+        )}
+      </div>
+    );
+  }
+
+  return (
+    <div className={styles.stack}>
```
Naycon: Since this is a list, it should likely be marked up as such as well, using ul/li. Just check that those elements are allowed as children of whatever you end up putting this list inside of.

Naycon (commented):
Nice, great improvement! 

Some a11y comments to take a look at, mainly around better semantic markup. I tried the stories with a screen reader, and it needs some more structure to better convey the content. I haven't deeply considered what the correct markup/representation is in each comment, so don't take them as the correct answer.

Also, regex parsing for security always feels scary to me. :)

packages/ui/src/components/Header/Header.tsx
cdedreuille: You're right. I initially tried to avoid using react-markdown because it doesn't work well with Jest and since our use case is quite small I wanted a simple solution but I found another package that do the work in a simpler way and the library is also present in Backstage already.

packages/ui/src/components/Header/Header.tsx
cdedreuille: You're right. I should use `<ul>/<li>` for tags and perhaps `<dl>/<dt>/<dd>` for metadata.

packages/ui/src/components/Header/Header.tsx
cdedreuille: Yes I'll use `<dl>/<dt>/<dd>` for that.

packages/ui/src/components/Header/HeaderMetadataUsers.tsx
cdedreuille: You're right. Much cleaner

packages/ui/src/components/Header/HeaderMetadataUsers.tsx
cdedreuille: You're right. I'll adjust that.

packages/ui/src/components/Header/HeaderMetadataStatus.tsx
cdedreuille: Here's a pass at it:
- The dot <span> gets role="img" with aria-label="{color} status" (e.g. "success status") — it now carries meaning for screen readers rather than being a silent empty element
- The <Text> and any <Link> inside remain fully accessible and untouched — keyboard and screen reader navigation to the link still works
- A screen reader will announce: "success status, Passing" (or "success status, link: Passing" when it has an href) which is clear and non-redundant

packages/ui/src/components/Header/HeaderMetadataUsers.tsx
```diff
@@ -0,0 +1,111 @@
...
+              <Avatar
+                src={user.src ?? 'data:,'}
+                name={user.name}
+                size="small"
+                purpose="decoration"
+              />
+            </Link>
+            <Link href={user.href} variant="body-medium" standalone>
+              {user.name}
+            </Link>
```
Naycon: Nit - this will lead to the link being read twice for a11y users, and it's two focusable elements beside each other linking to the same place. Maybe just wrapping in one link would work?
```suggestion
            <Link
              href={user.href}
              variant="body-medium" 
              standalone
              className={styles.avatarLink}
            >
              <Avatar
                src={user.src ?? 'data:,'}
                name={user.name}
                size="small"
                purpose="decoration"
              />
              {user.name}
            </Link>
```

packages/ui/src/components/Header/Header.tsx
```diff
@@ -21,8 +21,40 @@ import { HeaderNav } from './HeaderNav';
 import { useDefinition } from '../../hooks/useDefinition';
 import { HeaderDefinition } from './definition';
 import { Container } from '../Container';
+import { Lexer } from 'marked';
 import { Link } from '../Link';
-import { Fragment } from 'react/jsx-runtime';
+import { Fragment } from 'react';
+
+// Reject javascript:/vbscript:/data: URIs to prevent XSS via description links.
+const UNSAFE_HREF_RE = /^(javascript:|vbscript:|data:)/i;
```
Naycon: We should replace this with a lib as well, this is vulnerable as-is (e.g. browsers ignore `\t` in urls). `@braintree/sanitize-url` might be a good option.

Naycon (approved):
Looks (and sounds in the screen reader) great! 🚢 🇮🇹

packages/ui/src/components/Header/Header.tsx
```diff
@@ -20,9 +20,31 @@ import { RiArrowRightSLine } from '@remixicon/react';
...
+
+/**
+ * Parses inline Markdown links in a string and returns an array of React nodes.
+ * URLs are sanitized via `@braintree/sanitize-url`; unsafe URLs are rendered as
+ * plain text. Uses `marked` instead of `react-markdown` to avoid ESM issues.
+ */
+function renderInlineMarkdown(text: string): React.ReactNode[] {
+  return Lexer.lexInline(text).map((token, i) => {
+    if (token.type === 'link') {
+      const href = sanitizeUrl(token.href);
```
Rugvip, line 37: 👍

packages/ui/src/components/Header/Header.tsx
```diff
@@ -31,11 +53,45 @@ import { Fragment } from 'react/jsx-runtime';
...
+                </Link>
+              ) : (
+                <Text variant="body-medium" color="secondary">
+                  {tag.label}
+                </Text>
+              )}
+            </li>
+          ))}
+        </ul>
+      )}
```
Rugvip, line 99: Seems a bit tricky to have the tags show up on top of the header? 🤔 It'll create a lot of variance between different pages based on whether they have tags or not. I wonder if showing the tags just to the right of the title works?

Given this is purely about the visuals, my opinion here is very weak 😂

packages/ui/src/components/Header/Header.tsx
```diff
@@ -61,6 +117,35 @@ export const Header = (props: HeaderProps) => {
...
+          color="secondary"
+          className={classes.description}
+        >
+          {renderInlineMarkdown(description)}
+        </Text>
+      )}
+      {metadata && metadata.length > 0 && (
+        <dl className={classes.metaRow}>
+          {metadata.map((item, i) => (
+            <div key={`${i}:${item.label}`} className={classes.metaItem}>
```
Rugvip, line 137: hmm, why key on both index and label? Seems like one or the other would be enough

packages/ui/src/components/Header/Header.tsx
```diff
@@ -61,6 +117,35 @@ export const Header = (props: HeaderProps) => {
         </div>
         <div className={classes.controls}>{customActions}</div>
       </div>
+      {description && (
+        <Text
+          variant="body-medium"
+          color="secondary"
+          className={classes.description}
+        >
+          {renderInlineMarkdown(description)}
```
Rugvip: Thinking it might be worth a `useMemo` for this one?

Rugvip (approved):
Love it! 🎉 

Some optional things

packages/ui/report.api.md
```diff
@@ -1585,17 +1585,63 @@ export const HeaderDefinition: {
...
+  // (undocumented)
+  color: 'danger' | 'warning' | 'success' | 'info';
+  // (undocumented)
+  href?: string;
+  // (undocumented)
+  label: string;
+}
+
+// @public
+export interface HeaderMetadataUser {
```
Rugvip, line 1631: And one extra thing 😁, perhaps this should be `HeaderMetadataUserItem` for consistency? Or maybe the other way around `HeaderMetadataStatusItem` -> `HeaderMetadataStatus`?

packages/ui/src/components/Header/Header.tsx
cdedreuille: It is quite bad practice to use the index alone and we could have duplicated labels. That's why I combined these. This was a recommended solution from Copilot.

packages/ui/src/components/Header/Header.tsx
cdedreuille: Done

packages/ui/src/components/Header/Header.tsx
Rugvip: yeah I guess it's the best we can do really

PR #33998, opened 2026-04-19
Log debug warning when a GitHub App installation is missing for an org in GithubMultiOrgEntityProvider

### LINKED ISSUE #32972: Add a logged warning to GithubMultiorgEntityProvider that says when a Github App is not installed
### 📜 Issue Labels

### 🔎 Search Terms

```plain
"github warning", "github app not working", "github app authorization"
```

### 🗃️ Project Area

Auth

### 🔖 Need

This would make it easier for new users to figure out why the GitHub app doesn't work if they're not a GitHub Organization owner.

### 📝 Proposal

User's may not know (if they have Github Apps role, but not the `Owner` role) that they don't have the app installed. This may be noisy for environments that don't use the apps filter option, so maybe limit it to debug logging.

### 🔄 Alternatives

While there is a warning for this in the [troubleshooting guide](https://backstage.io/docs/integrations/github/github-apps/#troubleshooting), it doesn't say that the `NotFoundError: No app installation found for ${org} for ${appID}` doesn't bubble up to the logs. In my case, the error I got was a rate limiting error that said "You won't get this limited if you're authenticated."

mvanhorn:

Closes #32972

When `GithubMultiOrgEntityProvider.read` iterates over the configured orgs, `GithubCredentialsProvider.getCredentials` throws a `NotFoundError: No app installation found for <org> in <appId>` if the GitHub App is not installed on that org, or if the authenticated user does not have the Organization Owner role required to see the installation. Previously this error bubbled up with no additional context, and the common root cause was only documented in the [GitHub Apps troubleshooting guide](https://backstage.io/docs/integrations/github/github-apps/#troubleshooting).

This change wraps the `getCredentials` call in a `try/catch`. On a `NotFoundError` we emit a `logger.debug` message that names the org, explains the two common causes, and links to the troubleshooting docs. The error is then re-thrown so existing behavior is preserved. Debug level keeps this quiet for deployments that intentionally run the provider against a superset of orgs and rely on `orgs` filtering, while still making the hint discoverable by anyone turning on debug logs to diagnose the original opaque error.

### DIFF
--- a/.changeset/github-multiorg-not-found-debug-log.md
+++ b/.changeset/github-multiorg-not-found-debug-log.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-github': patch
+---
+
+`GithubMultiOrgEntityProvider` now emits a debug-level log when a GitHub App installation is not found for an organization during a read. The message points to the GitHub Apps troubleshooting docs, making it easier to diagnose cases where the app is not installed on the org or where the authenticated user lacks the Organization Owner role required to see the installation.
--- a/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.test.ts
+++ b/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.test.ts
@@ -969,6 +969,36 @@ describe('GithubMultiOrgEntityProvider', () => {
 
       expect(entityProviderConnection.applyMutation).not.toHaveBeenCalled();
     });
+
+    it('should log a debug hint and rethrow when getCredentials fails with NotFoundError', async () => {
+      entityProvider = new GithubMultiOrgEntityProvider({
+        id: 'my-id',
+        gitHubConfig,
+        githubCredentialsProvider: {
+          getCredentials: mockGetCredentials,
+        },
+        githubUrl: 'https://github.com',
+        logger,
+        orgs: ['orgA'],
+      });
+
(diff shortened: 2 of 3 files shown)


mvanhorn:
Addressed points 1 and 3 in e162159:

- Implicit `any`: typed the outer `let credentials` as `GithubCredentials` and added the type to the existing `@backstage/integration` import block.
- Regression test: added `should log a debug hint and rethrow when getCredentials fails with NotFoundError` in `GithubMultiOrgEntityProvider.test.ts`. It mocks `getCredentials` to reject with a `NotFoundError`-named error, asserts `logger.debug` is called with the troubleshooting message, and confirms the error still rethrows so `applyMutation` is not invoked.

On point 2 (`GithubAppCredentialsMux.getAppToken` swallowing `NotFoundError` internally and returning `undefined`): you're right that in the common default-credentials path this catch doesn't fire, so the hint is limited to the direct-`getCredentials`-throws path. Detecting the silent-`undefined`-token case from `GithubMultiOrgEntityProvider` would mean inspecting the `type` / `token` fields of the returned credentials, which crosses the integration package's abstraction. Happy to follow that up in a separate PR if the direction is welcome, but wanted to flag it as a scope change rather than fold it in here.

Verified: `CI=1 yarn test plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.test.ts` - 19 of 19 pass. `yarn tsc` clean.

PR #34016, opened 2026-04-21
feat(catalog-model): add mcp-server spec type for API kind

### LINKED ISSUE #32062: RFC: Modeling of MCP Servers in the Catalog

benjdlambert:
## Summary

Implements [RFC #32062](https://github.com/backstage/backstage/issues/32062) by adding `spec.type: 'mcp-server'` as a structured subtype of the existing `API` kind under `v1alpha1`/`v1beta1`.

- MCP server entities carry a `spec.remotes: { type, url }[]` list instead of a stringly-typed `definition`, for representing Model Context Protocol servers in the catalog.
- Existing API entities (`openapi`, `asyncapi`, `graphql`, etc.) are completely unaffected.
- Registration is opt-in via the `@backstage/plugin-catalog-backend-module-ai-model` backend module, which activates the catalog model layer for specType-based schema dispatch.
- Adds `addKindVersion` to `CatalogModelLayerBuilder` (alpha) so separate layers can add new versions or spec types to existing kinds without re-declaring kind metadata.

## New alpha exports from `@backstage/catalog-model`

- `McpServerApiEntity` — entity type extending `ApiEntityV1alpha1`
- `McpServerRemote` — transport endpoint type
- `mcpServerApiEntityValidator` — `KindValidator` for the mcp-server specType
- `isMcpServerApiEntity` — type guard
- `mcpServerApiEntityModel` — catalog model layer for registration
- `CatalogModelAddKindVersionDefinition` — definition type for `addKindVersion`
- `CatalogModelLayerBuilder.addKindVersion()` — adds versions to an already-declared kind

## Example

```yaml
kind: API
metadata:
  name: backstage-mcp-server
  description: An MCP server that exposes tools related to the Backstage ecosystem
spec:
  type: mcp-server
  lifecycle: experimental
  owner: team-a
  remotes:
    - type: streamable-http
      url: http://localhost:7007/api/mcp/v1
```

## Test plan

- [x] New unit tests for the `mcpServerApiEntityValidator` and `isMcpServerApiEntity` type guard.
- [x] New integration test that compiles the catalog model with the MCP server layer and asserts specType-based schema dispatch routes `mcp-server` and non-mcp entities to different schemas.
- [x] New unit + integration tests for `addKindVersion` (ops generation, specType dispatch, schema validation, and a round-trip through `compileCatalogModel`).
- [x] Existing v1alpha1/v1beta1 tests untouched and still pass.
- [x] `yarn tsc` and tests all clean.

## Checklist

- [x] Changeset for `@backstage/catalog-model` describing the new spec type.
- [x] Tests for new functionality and regression tests for existing versions.
- [x] All commits signed off.

Closes #32062.


gabemontero:
hey @benjdlambert - curious what the thoughts are about building on this change after it merges.

Context: 
- your point in #32062 about 
```
his structure also creates room for future extensions. MCP and the AI space includes additional concepts beyond just MCP Servers, such as Agents, Models and Prompts. Some of these may eventually warrant structured representation in the Software Catalog as well. If we choose to model additional MCP components in a similar manner, those changes should also go through their own RFC’s so that the shared Catalog model can evolve in a predictable and well-considered manner.
```
- my request for comments / interest with #33060 for bringing in AI Model Servers in a similar fashion

So how would you and the maintainers suggest submitting a PR to add other choices for `spec.type` besides `mcp-server`:
- since v1alpha2 is alpha, we can change v1alpha2, and simply add say `ai-model-servers` as an allowed choice for `spec.types` ?
- or is defining a v1alpha3 required to add 'ai-model-servers' as a allowed choice, where replicating the same shape wrt using `addKind` would be what you all proscribe ?
-- also with this, I'm assuming v1alpha3 should be cumulative, and have v1alpha3-version exports of the public exports captured in the report-api.md ... is that a correct assumption?
-- or are you all inclined to have users choose between a single choice for what is allowed for `spec.type`, where subsequent alpha versions of the `API` entity and distinct and not cumulative? 

thanks for any time / insights

packages/catalog-model/src/kinds/McpServerApiEntity.ts
```diff
@@ -0,0 +1,103 @@
...
+ */
+export const mcpServerApiEntityValidator =
+  ajvCompiledJsonSchemaValidator(mcpServerSchema);
+
+/**
+ * Type guard: narrows an entity to the MCP server API subtype.
+ *
+ * @alpha
+ */
+export function isMcpServerApiEntity(
```
Rugvip, line 61: Ah right, this one is actually useful because of the separate type, same with https://github.com/backstage/backstage/pull/34261#discussion_r3265378104

packages/catalog-model/src/model/createCatalogModelLayerBuilder.ts
```diff
@@ -87,6 +91,10 @@ export interface CatalogModelLayerBuilder {
    * Adds a new kind to the model.
    */
   addKind(kind: CatalogModelKindDefinition): void;
+  /**
+   * Adds one or more versions to an already-declared kind.
+   */
+  addKindVersion(definition: CatalogModelAddKindVersionDefinition): void;
```
Rugvip, line 97: Hmm, would've expected this to be covered by `updateKind` or `addKind` with a merge? 🤔

Rugvip (commented):
Yep, just the model API thing

packages/catalog-model/src/model/createCatalogModelLayerBuilder.ts
benjdlambert: Yeah me and @freben spoke about this yesterday and went with this.

Happy to this though in favor of it:

https://github.com/backstage/backstage/blob/c35e9682549269cf05cb4a3679a7feaee5e9cf48/packages/catalog-model/src/model/compileCatalogModel.ts#L222-L237

benjdlambert: `updateKind` doesn't allow you to update a non-existing `kind/spec.type` version.

packages/catalog-model/report-alpha.api.md
```diff
@@ -70,6 +88,12 @@ export interface CatalogModel {
   listRelations(): CatalogModelRelationSummary[];
 }
 
+// @alpha
+export interface CatalogModelAddKindVersionDefinition {
+  kind: string;
+  versions: CatalogModelKindVersionDefinition[];
```
freben, line 94: Possible nit: it's a bit unnecessary to have an array here. I see that it may be part of aligning with addKind, but it's not necessarily the most convenient contract. We can change it later if we want though, or add support for both variants, in a separate release

freben (approved):

PR #34026, opened 2026-04-22
feat(connections): implementing bep-0014

neoreddog:
In progress PR implementing [BEP-0014](https://github.com/backstage/backstage/pull/33921) in its ongoing state. 

Expecting a lot of this to evolve as the BEP is modified, but its a starting point. 

So far I've:
- Created the service definition for the connections package
- Created an example backend app that consumes this
- Basic implementations of functions for creating and consuming connection types

Not going to go into too much detail here, the changes are in the code :^) 


plugins/connections-example-backend/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "express": "^4.14.0",
+    "express-promise-router": "^4.1.0",
+    "zod": "^3.25.76 || ^4.0.0"
+  },
+  "devDependencies": {
+    "@backstage/backend-test-utils": "workspace:^",
+    "@backstage/cli": "workspace:^",
+    "@types/express": "^4.17.6",
+    "@types/supertest": "^2.0.8",
+    "supertest": "^6.2.4"
```
Rugvip: drop these, they don't line up with the rest and won't really be needed

packages/connections/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@backstage/connections",
+  "version": "0.1",
```
Rugvip:
```suggestion
  "version": "0.0.0",
```

For new packages

packages/connections/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "description": "Connections framework used by Backstage to better manage external connections used by integrations and plugins",
+  "backstage": {
+    "role": "node-library",
+    "pluginId": "connections"
+  },
+  "publishConfig": {
+    "access": "restricted"
+  },
```
Rugvip, line 17: Missing `dist` entires for the publish config, see some of the other packages that are set up in a similar way. If a package doesn't use `exports` it needs this instead.

packages/connections/src/schema/harness.ts
```diff
@@ -0,0 +1,42 @@
...
+export const HarnessConnectionType = createConnectionType({
+  type: 'harness',
+  configSchema: z.object({
+    host: z.string(),
+  }),
+  authMethods: [
+    {
+      method: 'none',
+      configSchema: z.object({}),
+    },
```
Rugvip: Doesn't harness allow unauthenticated access? otherwise we can skip this one

packages/connections/src/schema/harness.ts
```diff
@@ -0,0 +1,42 @@
...
+    {
+      method: 'apiKey',
```
Rugvip: Hmm, not sure harness actually has concrete API keys as opposed to always using tokens. Might be that we want to map both over to tokens instead? 🤔

packages/connections/src/schema/bitbucketCloud.ts
```diff
@@ -0,0 +1,51 @@
...
+        appPassword: z.string(),
+      }),
+    },
+    {
+      method: 'oauth',
+      configSchema: z.object({
+        clientId: z.string(),
+        clientSecret: z.string(),
+      }),
+    },
```
Rugvip, line 50: Does this allow API access? otherwise I think we should exclude it for now

packages/connections/src/schema/azureBlobStorage.ts
```diff
@@ -0,0 +1,59 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { createConnectionType } from '../system/createConnectionType';
+import { z } from 'zod/v4';
+
+export const AzureBlobStorageConnectionType = createConnectionType({
+  type: 'azure-blob-storage',
```
Rugvip, line 21: I'm not sure this should be a thing actually. It might be that we instead want to have a more generic azure connection config, and then potentially look into ways to select the desired connection instead. Fine to leave for experimental phase though

packages/connections/src/api/DefaultConnectionService.ts
```diff
@@ -0,0 +1,275 @@
...
+import { combineConnectionSources } from '../system/combineConnectionSources';
+
+function describeError(error: unknown): string {
+  if (error instanceof z.ZodError) {
+    return z.prettifyError(error);
+  }
+  if (error instanceof Error) {
+    return error.message;
+  }
+  return String(error);
```
Rugvip: Think this could be `return String(toError(error))`

packages/connections/src/api/DefaultConnectionService.ts
```diff
@@ -0,0 +1,275 @@
...
+import { ConnectionAuthMethodKey } from './ConnectionType';
+import { Connection, RootConnection } from './Connection';
+import { JsonObject } from '@backstage/types';
+import { InputError, NotAllowedError, NotFoundError } from '@backstage/errors';
+import { z } from 'zod/v4';
+import { getLegacyIntegrations } from '../system/getLegacyIntegrations';
+import { combineConnectionSources } from '../system/combineConnectionSources';
+
+function describeError(error: unknown): string {
+  if (error instanceof z.ZodError) {
```
Rugvip: I think it's best to have a habit of avoiding `instanceof`, especially for `Error` checks. It's not safe when dealing with cross-realm errors. `error.name === 'ZodError'`

packages/connections/src/api/ConnectionsService.ts
```diff
@@ -0,0 +1,38 @@
...
+  }): Promise<Connection<TType, TAuthMethod>>;
+
+  findOptional<
+    TType extends ConnectionTypeKey,
+    TAuthMethod extends ConnectionAuthMethodKey<TType>,
+  >(options: {
+    type: TType;
+    url: string;
+    authMethods: readonly [TAuthMethod, ...TAuthMethod[]];
+  }): Promise<Connection<TType, TAuthMethod> | undefined>;
```
Rugvip: Not 100% sure about the `find` / `findOptional` split. Could be an overload, or we only have `find`. Wonder if we could make `findOptional` hidden for now? To make in a minimal amount of work to bring it back, but also challenge consumption a bit to see if it's really needed?

packages/backend-plugin-api/src/wiring/types.ts
```diff
@@ -65,6 +65,30 @@ type DepsToInstances<
     : TDeps[key]['T'];
 };
 
+/**
+ * Declares that a plugin or module consumes a particular connection type.
+ *
+ * @public
+ */
+export interface ConnectionRegistration {
```
Rugvip, line 73: and let's hide this as well or make it an alpha export

packages/backend-plugin-api/src/wiring/types.ts
```diff
@@ -85,6 +109,14 @@ export interface BackendPluginRegistrationPoints {
...
     factory: (context: ExtensionPointFactoryContext) => TExtensionPoint;
   }): void;
+  /**
+   * Declares a connection type that this plugin consumes at runtime.
+   *
+   * Plugins that depend on the connections service must declare every
+   * connection type they query via `find` / `findOptional`. The framework
+   * verifies declarations at startup and at runtime.
+   */
+  registerConnection(registration: ConnectionRegistration): void;
```
Rugvip: Let's not make this public yet. Keep it private and then cast or use interface augmentation for our test code.

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -171,6 +216,29 @@ export class BackendInitializer {
...
+    moduleId?: string,
+  ): ConnectionRegistration[] {
+    if (moduleId) {
+      return (
+        this.#callerConnectionRegistrations.get(
+          callerKey(pluginId, moduleId),
+        ) ?? []
+      );
+    }
+    // For plugins, aggregate registrations from the plugin itself and all its modules
```
Rugvip: Hmm, don't think so, should still just be the plugin itself? 🤔

Rugvip (changes requested):
Nice! 🎉 

Couple of fixes here and there, most important ones being the ones that enable us to ship, i.e. making sure we don't change any existing API surfaces and getting the package in shape

packages/connections/src/schema/harness.ts
neoreddog: You're right. Seems like you create an API key -> token under that API key -> use token in API requests

I've changed to:
```typescript
authMethods: [
    {
      method: 'token',
      configSchema: z.object({
        token: z.string(),
        apiKey: z.string().optional(),
      }),
    },
  ],
```

packages/connections/src/schema/bitbucketCloud.ts
neoreddog: It seems like it does:
https://developer.atlassian.com/cloud/bitbucket/rest/intro/#making-requests

packages/connections/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@backstage/connections",
+  "version": "0.0.0",
+  "description": "Connections framework used by Backstage to better manage external connections used by integrations and plugins",
+  "backstage": {
+    "role": "node-library",
+    "pluginId": "connections"
+  },
+  "publishConfig": {
+    "access": "restricted",
```
Rugvip: Should be public

packages/connections/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "clean": "backstage-cli package clean",
+    "lint": "backstage-cli package lint",
+    "prepack": "backstage-cli package prepack",
+    "postpack": "backstage-cli package postpack",
+    "start": "backstage-cli package start",
+    "test": "backstage-cli package test"
+  },
+  "dependencies": {
+    "@backstage/backend-plugin-api": "workspace:^",
+    "@backstage/cli-common": "workspace:^",
```
Rugvip: prolly don't need this

packages/connections/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "@backstage/config": "workspace:^",
+    "@backstage/connections": "workspace:^",
```
Rugvip: wut? :D

packages/connections/src/schema/harness.ts
```diff
@@ -0,0 +1,34 @@
...
+  }),
+  authMethods: [
+    {
+      method: 'token',
+      configSchema: z.object({
+        token: z.string(),
+        apiKey: z.string().optional(),
+      }),
+    },
+  ],
```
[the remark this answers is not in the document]
neoreddog: @Rugvip this ends up going against the change we made to harness removing the apiKey only method. what do you suggest? 

I think we're right in that we can only auth with a token & apiKey, not just apiKey.

plugins/connections-example-backend/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "@backstage/backend-defaults": "workspace:^",
+    "@backstage/backend-plugin-api": "workspace:^",
+    "@backstage/catalog-client": "workspace:^",
+    "@backstage/connections": "workspace:^",
+    "@backstage/errors": "workspace:^",
+    "@backstage/plugin-catalog-node": "workspace:^",
+    "@backstage/types": "workspace:^",
+    "express": "^4.22.0",
+    "express-promise-router": "^4.1.0",
+    "zod": "^3.25.76 || ^4.0.0"
```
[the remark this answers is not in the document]
Rugvip: this is wrong

packages/connections/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "postpack": "backstage-cli package postpack",
+    "start": "backstage-cli package start",
+    "test": "backstage-cli package test"
+  },
+  "dependencies": {
+    "@backstage/backend-plugin-api": "workspace:^",
+    "@backstage/config": "workspace:^",
+    "@backstage/errors": "workspace:^",
+    "@backstage/types": "workspace:^",
+    "zod": "^3.25.76 || ^4.0.0"
```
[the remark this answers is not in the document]
Rugvip: this is wrong

plugins/connections-example-backend/src/router.ts
```diff
@@ -0,0 +1,84 @@
...
+        res.status(400).json(e.message);
+        return;
+      }
+      if (e instanceof Error && e.name === 'NotAllowedError') {
+        res.status(403).json(e.message);
+        return;
+      }
+      throw e;
+    }
+    res.status(200).json(connection);
```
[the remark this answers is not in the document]
neoreddog: this is wrong

packages/backend-app-api/src/wiring/BackendInitializer.ts
```diff
@@ -171,6 +216,29 @@ export class BackendInitializer {
...
+    const result: ConnectionRegistration[] = [];
+    for (const [key, registrations] of this.#callerConnectionRegistrations) {
+      if (key === pluginId) {
+        result.push(...registrations);
+      }
+    }
+    return result;
```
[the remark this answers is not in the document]
neoreddog: Fixed in 
a915c6686dd6e6d3290c2a7d4c3b3bd516c54679

Rugvip (approved):
:shipit: 😎

Issue #33809, opened 2026-04-08
publish:github action is failing to push files to newly created repos on GitHub Enterprise Cloud (GHEC)


adobejmong:
### 📝 Acknowledgement

- [x] I acknowledge that issues using this template may be closed without further explanation.

### Content

* Backstage version: 1.44.2
* isomorphic-git@npm:1.25.7
* @backstage/plugin-scaffolder-node@npm:0.12.0
* @backstage/plugin-scaffolder-backend-module-github@npm:0.9.1

Recently, we've noticed that pushing to GitHub Enterprise Cloud (GHEC) using the built-in Backstage action `publish:github` would randomly fail similar to this exception:

```shell
2026-04-02T09:50:37.522Z info: Init git repository {dir=/tmp/edc5eb80-5cb4-44ff-99b4-bd6bec3554b9/deployRepo}
2026-04-02T09:50:37.525Z info: Adding file {dir=/tmp/edc5eb80-5cb4-44ff-99b4-bd6bec3554b9/deployRepo,filepath=.}
2026-04-02T09:50:37.596Z info: Committing file to repo {dir=/tmp/edc5eb80-5cb4-44ff-99b4-bd6bec3554b9/deployRepo,message=initial commit}
2026-04-02T09:50:37.613Z info: Pushing directory to remote {dir=/tmp/edc5eb80-5cb4-44ff-99b4-bd6bec3554b9/deployRepo,remote=origin}
2026-04-02T09:50:37.810Z error: Failed to push to repo {dir=/tmp/edc5eb80-5cb4-44ff-99b4-bd6bec3554b9/deployRepo, remote=origin}
2026-04-02T09:50:37.810Z Error: HTTP Error: 404 Not Found {data={"statusCode":404,"statusMessage":"Not Found","response":"Repository not found."}}
    at Git.push (/app/packages/backend/node_modules/@backstage/plugin-scaffolder-node/dist/scm/git.cjs.js:177:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Object.initRepoAndPush (/app/packages/backend/node_modules/@backstage/plugin-scaffolder-node/dist/actions/gitHelpers.cjs.js:36:3)
    at async Object.initRepoPushAndProtect (/app/packages/backend/node_modules/@backstage/plugin-scaffolder-backend-module-github/dist/actions/helpers.cjs.js:194:24)
[...]

PR #34027, opened 2026-04-22
feat: add util to get octokit client which support retries via octokit/plugin-retry

adobejmong:

This is part 1 of 2.

We've been working with @awanlin on https://github.com/backstage/backstage/issues/33809 as well as with GitHub support.

tl;dr; the recommendation has been to try to retry/add some delay to the requests to see if that will help alleviate some of the issues we sometimes run into when using the built-in Backstage `publish:github` action.

Based on my last chat with @awanlin, it was also recommended that we use [@octokit/plugin-retry](https://github.com/octokit/plugin-retry.js) plugin. https://actions-cool.github.io/octokit-rest/en-US/guide/10_retries/

Since this is my first contribution, I wanted to purposely keep the changes small to get a sense of what the next PR might look like.

This PR introduces the following:

* add `@octokit/plugin-retry` dependency with the same version used in `catalog-backend-module-github` (`^6.0.0`)
* add util called `getOctokitClient` which will return an Octokit client using the `retry` plugin. Default behavior is to retry. To get the original client, you can set either `retries` or `retryAfter` to `0` e.g.
  ```
  const retryOptions = { retries: 0 };
  const client = getOctokitClient(octokitOptions, logger, retryOptions);
  ```
* add unit tests for the changes introduced in `utils.ts`

Part 2 is to replace the calls in `github.ts` and expose additional parameters to the built-in action. For example:

```
      const client = new Octokit({
        ...octokitOptions,
        log: ctx.logger,
      });
```

to

```
      const retryOptions = { retries: 10, retryAfter: 2000 };
      const client = getOctokitClient(octokitOptions, ctx.logger, retryOptions)
```

Thank you in advance for your help.

### DIFF
--- a/plugins/scaffolder-backend-module-github/src/util.test.ts
+++ b/plugins/scaffolder-backend-module-github/src/util.test.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 The Backstage Authors
+ * Copyright 2026 The Backstage Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -39,15 +39,15 @@ describe('isRetryEnabled', () => {
     expect(isRetryEnabled()).toBe(true);
   });
 
-  it('returns true when retries is positive and retryDelay is undefined', () => {
+  it('returns true when retries is positive and retryAfter is undefined', () => {
     expect(isRetryEnabled(3)).toBe(true);
   });
 
-  it('returns true when retries is undefined and retryDelay is positive', () => {
+  it('returns true when retries is undefined and retryAfter is positive', () => {
     expect(isRetryEnabled(undefined, 1000)).toBe(true);
   });
 
-  it('returns true when both retries and retryDelay are positive', () => {
+  it('returns true when both retries and retryAfter are positive', () => {
     expect(isRetryEnabled(5, 2000)).toBe(true);
   });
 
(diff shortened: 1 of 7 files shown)


plugins/scaffolder-backend-module-github/src/util.ts
```diff
@@ -22,9 +22,90 @@ import {
...
+  // explicitly setting retries or retryDelay to 0
+  if (!isRetryEnabled(retries, retryDelay)) {
+    return new Octokit({
+      ...octokitOptions,
+      log: logger,
+    });
+  }
+
+  // Update the octokit options to include retry configuration with logging
+  const MyOctokit = Octokit.plugin(retry);
```
awanlin:
```suggestion
  const OctokitClient = Octokit.plugin(retry);
```

plugins/scaffolder-backend-module-github/src/util.ts
```diff
@@ -22,9 +22,90 @@ import {
...
+  return new MyOctokit({
```
awanlin:
```suggestion
  return new OctokitClient({
```

plugins/scaffolder-backend-module-github/src/util.test.ts
```diff
@@ -0,0 +1,154 @@
+/*
+ * Copyright 2025 The Backstage Authors
```
awanlin:
```suggestion
 * Copyright 2026 The Backstage Authors
```

Net new files should use the current year.

awanlin (changes requested):
Thanks @adobejmong, left a few tiny comments and please follow up on the Copilot comments as well, overall happy with this though! 🙏

plugins/scaffolder-backend-module-github/src/util.ts
```diff
@@ -22,9 +22,91 @@ import {
...
+    return new Octokit({
+      ...octokitOptions,
+      log: logger,
+    });
+  }
+
+  // Update the octokit options to include retry configuration with logging
+  const OctokitClient = Octokit.plugin(retry);
+  return new OctokitClient({
+    ...octokitOptions,
```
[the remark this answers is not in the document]
adobejmong: This is expected since there could be situations where you wouldn't want the client to retry.

awanlin (approved):
Thanks for the follow up @adobejmong 🚀

PR #34035, opened 2026-04-23
Remove uuid dependency in favor of crypto.randomUUID()

freben:

The `uuid` package dropped its CommonJS entry point in v14 (see #34033), making it incompatible with Backstage's CJS build output and Jest test runner. Rather than working around the ESM-only issue, this PR removes the `uuid` dependency entirely and replaces all usage with the built-in `crypto.randomUUID()`.

`crypto.randomUUID()` has been available in Node.js since v16.7 and in all major browsers since March 2022 (Chrome 92, Firefox 95, Safari 15.4, Edge 92), so it's a safe drop-in replacement for `uuid.v4()`.

**Changes:**
- Replace `import { v4 as uuid } from 'uuid'` with `import { randomUUID as uuid } from 'node:crypto'` in all backend packages
- Replace `uuid.v4()` / `uuid()` with `globalThis.crypto.randomUUID()` in the three frontend/common packages (`permission-common`, `search-react`, `signals`)
- Replace `require('uuid').v4()` with `require('node:crypto').randomUUID()` in the catalog-backend bootstrap migration
- Update `jest.mock('uuid')` calls to mock `node:crypto` instead
- Update the `catalog-provider-module` scaffolding template
- Remove `uuid` from `dependencies` in all 29 affected `package.json` files

**Note:** A single transitive `uuid@^11` reference remains in the lockfile via `@snyk/error-catalog-nodejs-public` (pulled in by `snyk-nodejs-lockfile-parser` in `packages/yarn-plugin`). This is outside our control and resolves to uuid 11.1.0 which still ships CJS — no action needed.

Supersedes #34033.


dyatko (approved):
Nice!

PR #34036, opened 2026-04-23
fix(scaffolder-github): add GraphQL fallback when git push is blocked

### LINKED ISSUE #34074: publish:github and github:repo:push fail with ECONNRESET behind network proxies with deep packet inspection
## Description

The `publish:github` and `github:repo:push` scaffolder actions fail with `ECONNRESET` when Backstage runs behind a network proxy that performs deep packet inspection on outbound HTTPS traffic.

The repo creation step succeeds, but the git push step fails immediately:

```
info: Init git repository
info: Adding file
info: Committing file to repo
info: Pushing directory to remote
error: Failed to push to repo
Error: read ECONNRESET
    at TLSWrap.onStreamRead (node:internal/stream_base_commons:216:20)
```

## Root cause

The scaffolder uses isomorphic-git to push files to GitHub. The push sends a `POST` to `/git-receive-pack` with binary git pack data. Network proxies that perform deep packet inspection terminate TLS, inspect the content, and re-encrypt. When the inspection engine encounters the binary pack data, it cannot parse it and drops the connection.

All other scaffolder operations use Octokit API calls which send standard JSON. The proxy can inspect JSON without issues, so repo creation, catalog registration, webhook setup, and branch protection all work. The git push is the only operation that sends binary data via the git smart HTTP protocol.

This cannot be resolved with environment variables (`NODE_USE_ENV_PROXY`, `NODE_TLS_REJECT_UNAUTHORIZED`), proxy agent libraries, or Node.js flags. The proxy routes traffic correctly — the CONNECT tunnel is established, TLS handshake succeeds, and GET requests to `/info/refs` return 200. It is specifically the POST with binary content that gets dropped by the proxy.

## How to reproduce

1. Deploy mitmproxy with a deep packet inspection simulation addon:

```python
# proxy_simulator.py
from mitmproxy import http, ctx

class ProxySimulator:
    def request(self, flow: http.HTTPFlow) -> None:
        if (flow.request.method == "POST"
            and "git-receive-pack" in flow.request.path):
            ctx.log.warn(f"Killing POST to {flow.request.url}")
            flow.kill()

addons = [ProxySimulator()]
```

```bash
mitmdump --listen-port 3128 --ssl-insecure -s proxy_simulator.py
```

2. Install isomorphic-git at the same version Backstage uses:

```bash
npm install isomorphic-git@^1.23.0
```

3. Run the reproduction script (replace `<owner>/<repo>` with any repo you have push access to):

```javascript
// HTTPS_PROXY=http://localhost:3128 NODE_USE_ENV_PROXY=1 \
// NODE_TLS_REJECT_UNAUTHORIZED=0 GITHUB_TOKEN=<token> \
// node reproduce.js

const git = require('isomorphic-git');
const http = require('isomorphic-git/http/node');
const fs = require('fs');
const TOKEN = process.env.GITHUB_TOKEN;
const dir = '/tmp/repro-' + Date.now();
fs.mkdirSync(dir, { recursive: true });

async function main() {
  console.log('git.clone() ...');
  await git.clone({
    fs, http, dir,
    url: 'https://github.com/<owner>/<repo>.git',
    singleBranch: true, depth: 1,
    onAuth: () => ({ username: 'x-access-token', password: TOKEN }),
  });
  console.log('Result: OK');

  console.log('git.add() + git.commit() ...');
  fs.writeFileSync(dir + '/test.txt', 'test');
  await git.add({ fs, dir, filepath: 'test.txt' });
  await git.commit({
    fs, dir, message: 'initial commit',
    author: { name: 'Scaffolder', email: 'scaffolder@backstage.io' },
  });
  console.log('Result: OK');

  console.log('git.push() ...');
  await git.push({
    fs, http, dir, remote: 'origin', ref: 'main',
    onAuth: () => ({ username: 'x-access-token', password: TOKEN }),
  });
  console.log('Result: OK');
}

main().catch(err => {
  console.log('Result: FAILED');
  console.log('Error:', err.code, '-', err.message);
});
```

**Expected output:**
```
git.clone() ...
Result: OK
git.add() + git.commit() ...
Result: OK
git.push() ...
Result: FAILED
Error: ECONNRESET - socket hang up
```

## Who is affected

Anyone running Backstage behind a proxy that performs deep packet inspection. This is common in environments where network security appliances inspect outbound HTTPS traffic.

## Proposed fix

PR #34036 adds a transparent fallback that catches the `ECONNRESET` and retries using GitHub's GraphQL `createCommitOnBranch` mutation. This pushes all files in a single API call using standard JSON which is not affected by packet inspection. The fallback only triggers on connection reset errors, so there is no behavior change for deployments not behind inspecting proxies.

## Related issues

- #29060 — `github:repo:push` fails with `ENOTFOUND` behind proxy
- #28475 — `publish:github` push timeout via simple-get
- #28467 — RFC for running Backstage behind a corporate proxy
- #33746 — retry for `Git.push()`

polasudo:
Fixes #34074

The proxy logs tell the whole story:
                                                                                                                                                                          
  [DPI] Killing POST to .../git-receive-pack    ← git push killed                                                                                                         
  POST https://api.github.com/user/repos        ← REST API works                                                                                                          
  POST https://api.github.com/repos/.../blobs   ← REST API works                                                                                                          
  POST https://api.github.com/repos/.../trees   ← REST API works                                                                                                          
  POST https://api.github.com/repos/.../commits ← REST API works                                                                                                          
  PATCH https://api.github.com/repos/.../refs   ← REST API works 
  
Reproduction and proof

This issue was reproduced on upstream Backstage using isomorphic-git v1.27.1 on Node v22.22.2, the same versions used in the Backstage scaffolder. A mitmproxy instance was deployed with a custom addon that simulates deep packet inspection by dropping POST requests to /git-receive-pack. This replicates the behavior observed in production environments where network proxies inspect TLS (transport layer security) intercepted traffic and drop requests containing binary payloads they cannot parse.

The publish:github and github:repo:push actions both call initRepoAndPush in @backstage/plugin-scaffolder-node, which uses isomorphic-git to push files to GitHub. The push flow makes two HTTP requests through the proxy: a GET to /info/refs?service=git-receive-pack for capability negotiation, and a POST to /git-receive-pack containing the binary git pack data. The GET succeeds because it returns plain text. The POST fails because the binary pack payload triggers the proxy's content inspection engine, which cannot parse it and drops the connection, resulting in **ECONNRESET**.

Every other operation in the scaffolder workflow goes through Octokit's REST API, which sends standard JSON over HTTPS. Proxies can inspect JSON without issues, so repo creation, catalog imports, webhook configuration, and branch protection all work. The git push is the only operation that sends binary data, and it is the only one that breaks.                                                                                                                                                            

This fix adds a transparent fallback in initRepoPushAndProtect. When initRepoAndPush fails with ECONNRESET, the code retries using GitHub's GraphQl API to create blobs, trees, commits, and refs. These are all standard JSON requests to api.github.com that pass through packet-inspecting proxies without issues. The fallback only triggers on connection reset errors, so there is no behavior change for users not behind inspecting proxies. Both publish:github and github:repo:push are covered since they both call initRepoPushAndProtect. All 52 existing tests pass without modification.

How to reproduce                                                                                                                                                        

You need a forward proxy that performs TLS interception and drops POST requests to /git-receive-pack. This simulates the behavior of network proxies that inspect HTTPS traffic and block requests containing binary payloads they cannot parse
1. Set up the proxy simulator                                                                                                                                           
Install mitmproxy (pip install mitmproxy or brew install mitmproxy) and create this addon file:                                                                         

# proxy_simulator.py                                                                                                                                                    
```
from mitmproxy import http, ctx                                                                                                                                         

class ProxySimulator:                                                                                                                                                   
  def request(self, flow: http.HTTPFlow) -> None:                                                                                                                     
    if (flow.request.method == "POST"
      and "git-receive-pack" in flow.request.path):                                                                                                               
      ctx.log.warn(f"Killing POST to {flow.request.url}")                                                                                                         
      flow.kill()                                                                                                                                                 
                                                                                                                                                                          
addons = [ProxySimulator()]
```
Start the proxy:                                                                                                                                                        

mitmdump --listen-port 3128 --ssl-insecure -s proxy_simulator.py                                                                                                        
                  
2. Reproduce the failure                                                                                                                                                

Install isomorphic-git at the same version Backstage uses (^1.23.0) and run the following script. This simulates the exact code path that publish:github and            
github:repo:push take when pushing scaffolded files to GitHub:
                                                                                                                                                                          
npm install isomorphic-git@1.27.1
                                                                                                                                                                          
// reproduce.js                                                                                                                                                         
```
// Run with:                                                                                                                                                            
//   HTTPS_PROXY=http://localhost:3128 NODE_USE_ENV_PROXY=1 \                                                                                                           
//   NODE_TLS_REJECT_UNAUTHORIZED=0 GITHUB_TOKEN=<your-token> \                                                                                                         
//   node reproduce.js                                                                                                                                                  
                                                                                                                                                                          
const git = require('isomorphic-git');                                                                                                                                  
const http = require('isomorphic-git/http/node');                                                                                                                       
const fs = require('fs');                                                                                                                                               
                                                                                                                                                                          
const TOKEN = process.env.GITHUB_TOKEN;                                                                                                                                 
const dir = '/tmp/repro-' + Date.now();                                                                                                                                 
fs.mkdirSync(dir, { recursive: true });                                                                                                                                 
                  
async function main() {                                                                                                                                                 
  console.log('Proxy:', process.env.HTTPS_PROXY);
  console.log('NODE_USE_ENV_PROXY:', process.env.NODE_USE_ENV_PROXY);                                                                                                   
  console.log('');                                                                                                                                                      
                                                                                                                                                                          
  // Step 1: Clone (GET /info/refs passes through the proxy)                                                                                                            
  console.log('git.clone() ...');
  await git.clone({                                                                                                                                                     
      fs, http, dir,                                                                                                                                                      
      url: 'https://github.com/<owner>/<any-repo>.git',                                                                                                                   
      singleBranch: true, depth: 1,                                                                                                                                       
      onAuth: () => ({ username: 'x-access-token', password: TOKEN }),                                                                                                    
    });
    console.log('Result: OK');                                                                                                                                            
                                                                                                                                                                          
    // Step 2: Create a commit locally                                                                                                                                    
    console.log('git.add() + git.commit() ...');                                                                                                                          
    fs.writeFileSync(dir + '/test.txt', 'test');                                                                                                                          
    await git.add({ fs, dir, filepath: 'test.txt' });                                                                                                                     
    await git.commit({                                                                                                                                                    
      fs, dir,                                                                                                                                                            
      message: 'initial commit',                                                                                                                                          
      author: { name: 'Scaffolder', email: 'scaffolder@backstage.io' },                                                                                                   
    });                                                                                                                                                                   
    console.log('Result: OK');                                                                                                                                            
                                                                                                                                                                          
    // Step 3: Push (POST /git-receive-pack with binary pack data — killed by proxy)                                                                                      
    console.log('git.push() ...');
    await git.push({                                                                                                                                                      
      fs, http, dir,
      remote: 'origin', ref: 'main',                                                                                                                                      
      onAuth: () => ({ username: 'x-access-token', password: TOKEN }),
    });                                                                                                                                                                   
    console.log('Result: OK');
  }                                                                                                                                                                       
                  
  main().catch(err => {                                                                                                                                                   
    console.log('Result: FAILED');
    console.log('Error:', err.code, '-', err.message);                                                                                                                    
  });                                                                                                                                                                     
                                                                                                                                                               
 
  Expected output                                                                                                                                                         
                  
  git.clone() ...
  Result: OK
  git.add() + git.commit() ...                                                                                                                                            
  Result: OK
  git.push() ...                                                                                                                                                          
  Result: FAILED  
  Error: ECONNRESET - socket hang up
 ```      


<!--- Please include the following in your Pull Request when applicable: -->

### DIFF
--- a/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
+++ b/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
@@ -430,7 +430,7 @@ async function collectFilesFromDir(
       results.push(
         ...(await collectFilesFromDir(fullPath, relativePath, root)),
       );
-    } else {
+    } else if (entry.isFile()) {
       results.push({
         filePath: relativePath,
         content: await fsPromises.readFile(fullPath),
--- a/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
+++ b/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
@@ -465,16 +465,16 @@ async function pushFilesViaGitHubApi(input: {
     );
   }
 
-  const MAX_PAYLOAD_BYTES = 50 * 1024 * 1024; // 50 MB
-  const totalBytes = files.reduce((sum, f) => sum + f.content.length, 0);
-  if (totalBytes > MAX_PAYLOAD_BYTES) {
+  const totalRawBytes = files.reduce((sum, f) => sum + f.content.length, 0);
+  // base64 expands content by ~4/3; add 1 KB per file for JSON overhead
+  const estimatedPayload =
+    Math.ceil(totalRawBytes * (4 / 3)) + files.length * 1024;
+  const MAX_PAYLOAD_BYTES = 30 * 1024 * 1024; // 30 MB after encoding
+  if (estimatedPayload > MAX_PAYLOAD_BYTES) {
     throw new Error(
-      `GraphQL API fallback payload too large (${(
-        totalBytes /
-        1024 /
-        1024
-      ).toFixed(1)} MB). ` +
-        'The GitHub GraphQL API limits request size. ' +
+      `GraphQL API fallback payload too large ` +
+        `(${(totalRawBytes / 1024 / 1024).toFixed(1)} MB raw, ` +
+        `~${(estimatedPayload / 1024 / 1024).toFixed(1)} MB encoded). ` +
         'Consider reducing template size or resolving the network issue preventing git push.',
     );
   }
@@ -494,7 +494,13 @@ async function pushFilesViaGitHubApi(input: {
       throw refError;
     }
     if (status === 404) {
-      await client.rest.repos.get({ owner, repo });
+      const { data: repoData } = await client.rest.repos.get({ owner, repo });
+      if (repoData.size > 0) {
+        throw new Error(
+          `Branch '${defaultBranch}' not found in ${owner}/${repo}. ` +
+            `The repository exists and its default branch is '${repoData.default_branch}'.`,
+        );
+      }
     }
     logger.info(
       `No existing HEAD found for ${owner}/${repo}#${defaultBranch} ` +
@@ -527,19 +533,37 @@ async function pushFilesViaGitHubApi(input: {
     fileChanges.deletions = [{ path: '.gitkeep' }];
   }
 
-  const result: {
-    createCommitOnBranch: { commit: { oid: string } };
-  } = await client.graphql(CREATE_COMMIT_ON_BRANCH_MUTATION, {
-    input: {
-      branch: {
-        repositoryNameWithOwner: `${owner}/${repo}`,
-        branchName: defaultBranch,
-      },
-      message: { headline: commitMessage },
-      fileChanges,
-      expectedHeadOid: headOid,
+  const commitInput = {
+    branch: {
+      repositoryNameWithOwner: `${owner}/${repo}`,
+      branchName: defaultBranch,
     },
-  });
+    message: { headline: commitMessage },
+    fileChanges,
+    expectedHeadOid: headOid,
+  };
+
+  let result: { createCommitOnBranch: { commit: { oid: string } } };
+  try {
+    result = await client.graphql(CREATE_COMMIT_ON_BRANCH_MUTATION, {
+      input: commitInput,
+    });
+  } catch (commitError: unknown) {
+    const msg = String((commitError as Error).message ?? '');
+    if (!msg.includes('expectedHeadOid')) {
+      throw commitError;
+    }
+    logger.warn('HEAD OID changed since read, retrying once');
+    const { data: freshRef } = await client.rest.git.getRef({
+      owner,
+      repo,
+      ref: `heads/${defaultBranch}`,
+    });
+    commitInput.expectedHeadOid = freshRef.object.sha;
+    result = await client.graphql(CREATE_COMMIT_ON_BRANCH_MUTATION, {
+      input: commitInput,
+    });
+  }
 
   const commitHash = result.createCommitOnBranch.commit.oid;
   logger.info(
--- a/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
+++ b/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
@@ -440,16 +440,8 @@ async function pushFilesViaGitHubApi(input: {
   gitAuthorInfo?: { name?: string; email?: string };
   logger: LoggerService;
 }): Promise<{ commitHash: string }> {
-  const {
-    dir,
-    owner,
-    repo,
-    client,
-    defaultBranch,
-    commitMessage,
-    gitAuthorInfo,
-    logger,
-  } = input;
+  const { dir, owner, repo, client, defaultBranch, commitMessage, logger } =
+    input;
 
   const files = await collectFilesFromDir(dir);
   logger.info(`Collected ${files.length} files for API push`);
@@ -475,7 +467,7 @@ async function pushFilesViaGitHubApi(input: {
       content: '',
       branch: defaultBranch,
     });
-    headOid = init.commit.sha;
+    headOid = init.commit.sha!;
   }
 
   const additions = files.map(file => ({
--- a/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
+++ b/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
@@ -324,64 +324,41 @@ export async function initRepoPushAndProtect(
   const commitMessage =
     getGitCommitMessage(gitCommitMessage, config) || 'initial commit';
 
-  const useApiPush =
-    process.env.HTTP_PROXY ||
-    process.env.HTTPS_PROXY ||
-    process.env.http_proxy ||
-    process.env.https_proxy;
-
   let commitResult: { commitHash: string };
 
-  if (useApiPush) {
-    logger.info(
-      'HTTP proxy detected, pushing via GitHub API to avoid ' +
-        'potential issues with the git smart HTTP protocol.',
-    );
-    commitResult = await pushFilesViaGitHubApi({
+  try {
+    commitResult = await initRepoAndPush({
       dir: getRepoSourceDirectory(workspacePath, sourcePath),
-      owner,
-      repo,
-      client,
+      remoteUrl,
       defaultBranch,
+      auth: {
+        username: 'x-access-token',
+        password,
+      },
+      logger,
       commitMessage,
       gitAuthorInfo,
-      logger,
     });
-  } else {
-    try {
-      commitResult = await initRepoAndPush({
+  } catch (error) {
+    const code = (error as NodeJS.ErrnoException).code;
+    if (code === 'ECONNRESET' || code === 'HttpError') {
+      logger.warn(
+        `Git push failed with ${code}, retrying via GitHub API. ` +
+          'This can happen when a network proxy blocks the binary payload ' +
+          'in the git smart HTTP protocol.',
+      );
+      commitResult = await pushFilesViaGitHubApi({
         dir: getRepoSourceDirectory(workspacePath, sourcePath),
-        remoteUrl,
+        owner,
+        repo,
+        client,
         defaultBranch,
-        auth: {
-          username: 'x-access-token',
-          password,
-        },
-        logger,
         commitMessage,
         gitAuthorInfo,
+        logger,
       });
-    } catch (error) {
-      const code = (error as NodeJS.ErrnoException).code;
-      if (code === 'ECONNRESET' || code === 'HttpError') {
-        logger.warn(
-          `Git push failed with ${code}, retrying via GitHub API. ` +
-            'This can happen when a network proxy blocks the binary payload ' +
-            'in the git smart HTTP protocol.',
-        );
-        commitResult = await pushFilesViaGitHubApi({
-          dir: getRepoSourceDirectory(workspacePath, sourcePath),
-          owner,
-          repo,
-          client,
-          defaultBranch,
-          commitMessage,
-          gitAuthorInfo,
-          logger,
-        });
-      } else {
-        throw error;
-      }
+    } else {
+      throw error;
     }
   }
 
--- a/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
+++ b/plugins/scaffolder-backend-module-github/src/actions/helpers.ts
@@ -17,6 +17,8 @@
 import { Config } from '@backstage/config';
 import { NotFoundError, toError } from '@backstage/errors';
 import { Octokit } from 'octokit';
+import { promises as fsPromises, Dirent } from 'node:fs';
+import { join } from 'node:path';
 
 import {
   getRepoSourceDirectory,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by polasudo** [comment]
reason of switch from rest to graphql is that from graphql we can commit multiple files with 1 request while REST API is just 1 file per 1 request

File: plugins/scaffolder-backend-module-github/src/actions/helpers.ts [hunk 1]
```diff
@@ -322,18 +324,42 @@ export async function initRepoPushAndProtect(
...
+        username: 'x-access-token',
+        password,
+      },
+      logger,
+      commitMessage,
+      gitAuthorInfo,
+    });
+  } catch (error) {
+    const code = (error as NodeJS.ErrnoException).code;
+    if (code === 'ECONNRESET' || code === 'HttpError') {
```
Rugvip: If we can reliably determine if we are in an environment where we know that this will fail, it'd be good to never try the `initRepoAndPush` path at all and instead go straight for `pushFilesViaGitHubApi`, so that we avoid the extra work.

Rugvip (changes requested):
Thank you! 👍 

As discussed during framework SIG the best path forward would be to fix this upstream in isomorphic git. If that fails, lemme know and we'll continue this thread.

polasudo:
> Thank you! 👍
> 
> As discussed during framework SIG the best path forward would be to fix this upstream in isomorphic git. If that fails, lemme know and we'll continue this thread.

Hey, @Rugvip I am back here with the results that this solution is more specific for this project and not for them 
https://github.com/isomorphic-git/isomorphic-git/issues/2308
https://github.com/isomorphic-git/isomorphic-git/pull/2307

is still left my crumbs there as a backup so if possible I will have the solution here in this repo, and I will implement your suggestion also

Rugvip:
@polasudo oh that's a much more higher level contribution upstream than what I had in mind actually. What I expected to solve upstream was the `ECONNRESET` error, which isn't really something we should expect to be happening even over a proxy, no?

polasudo:
> @polasudo oh that's a much more higher level contribution upstream than what I had in mind actually. What I expected to solve upstream was the `ECONNRESET` error, which isn't really something we should expect to be happening even over a proxy, no?

sorry I missunderstood that, yeah will take a look on that and try to fix the ECON there if possible will update it here after I try it.
Thanks

polasudo:
after trying it out the problem is 

The ECONNRESET happens because the proxy's content inspection engine kills the connection when it sees binary git pack data in the POST body. That binary data is the same whether we send it through simple-get or native http.request(). The proxy doesn't care which  HTTP client made the request it inspects the content and drops it either way.                                                                                                                                                                                                                                                       
The truth is that isomorphic-git can't fix this. The binary pack is the git protocol. Every git client in existence sends the same binary POST. If the proxy kills binary POSTs, git push will fail — whether it's isomorphic-git, canonical git, or anything else. The fix should then be here and we can leave out the upstream from it. 

I went upstream to isomorphic-git and tried two approaches:                                                                     
1. Adding a prepareCommit() API that returns staged files as JSON so consumers can push via provider APIs the maintainer said this is out of scope, that isomorphic-git is a git client and should know nothing about git providers                                            (https://github.com/isomorphic-git/isomorphic-git/issues/2308)         
2. Replacing simple-get with native Node HTTP modules for better proxy handling this removes a dependency but doesn't fix the ECONNRESET because the issue isn't in the HTTP client layer                                                                                             
  
The ECONNRESET can't be fixed in isomorphic-git because it's not a bug in isomorphic-git. sadly it lives inside of a protocol for git which is probably too far for me, so I would suggest to land the fix here if possible,
Thank you! @Rugvip

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
polasudo: updated

polasudo:
@Rugvip Hey, after last framework sig we had a conversation about it with @freben and we both think that it would be better to keep the default behaviour and just keep the fallback as I did pushed now, the reason is that it will not be changing any behaviour and if there will be proxy blocking the push it will fallback to the graphQl api and everything should be fine. 
I think we can have a final review and then merge it if it would be fine and good to go 
Thank you

polasudo:
Added 409 handling on `getRef` when the repo has zero commits and (`auto_init: false`). Also added symlink skip and payload size guard.

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
```diff
@@ -368,6 +404,174 @@ export async function initRepoPushAndProtect(
   return { commitHash: commitResult.commitHash };
 }
 
+async function collectFilesFromDir(
```
Rugvip, line 407: Need to work in `resolveSafeChildPath` from `@backstage/backend-plugin-api` to make sure we don't end up with any unsafe symlink resolution.

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
```diff
@@ -368,6 +404,174 @@ export async function initRepoPushAndProtect(
...
+  client: Octokit;
+  defaultBranch: string;
+  commitMessage: string;
+  logger: LoggerService;
+}): Promise<{ commitHash: string }> {
+  const { dir, owner, repo, client, defaultBranch, commitMessage, logger } =
+    input;
+
+  const files = await collectFilesFromDir(dir);
+  logger.info(`Collected ${files.length} files for API push`);
```
Rugvip: Either add a bit more context to this message or drop it imo. For example what repos we're pushing to and so on

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
```diff
@@ -368,6 +404,174 @@ export async function initRepoPushAndProtect(
...
+  try {
+    result = await client.graphql(CREATE_COMMIT_ON_BRANCH_MUTATION, {
+      input: commitInput,
+    });
+  } catch (commitError: unknown) {
+    const msg = String((commitError as Error).message ?? '');
+    if (!msg.includes('expectedHeadOid')) {
+      throw commitError;
+    }
+    logger.warn('HEAD OID changed since read, retrying once');
```
Rugvip: drop or add more context to this one too

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
```diff
@@ -368,6 +404,174 @@ export async function initRepoPushAndProtect(
...
+    commitInput.expectedHeadOid = freshRef.object.sha;
+    result = await client.graphql(CREATE_COMMIT_ON_BRANCH_MUTATION, {
+      input: commitInput,
+    });
+  }
+
+  const commitHash = result.createCommitOnBranch.commit.oid;
+  logger.info(
+    `Pushed ${files.length} files via GitHub GraphQL API: ${commitHash}`,
+  );
```
Rugvip, line 582: and here :D

Rugvip (changes requested):

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
polasudo: done

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
```diff
@@ -368,6 +404,185 @@ export async function initRepoPushAndProtect(
...
+      branch: defaultBranch,
+    });
+    headOid = init.commit.sha!;
+    needsCleanup = true;
+  }
+
+  const additions = files.map(file => ({
+    path: file.filePath,
+    contents: file.content.toString('base64'),
+  }));
```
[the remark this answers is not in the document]
polasudo: Adding file-count chunking would introduce significant complexity (multiple sequential commits, race conditions with concurrent pushes, partial commit cleanup on failure) for a limit that doesn't actually exist.
2nd point is that we also have limitation for file sizes. 
[Doc](https://docs.github.com/en/graphql/reference/commits#createcommitonbranch)
[Doc2](https://docs.github.com/en/graphql/reference/git#filechanges)
[Doc3](https://docs.github.com/en/graphql/reference/git#fileaddition)

plugins/scaffolder-backend-module-github/src/actions/helpers.ts
```diff
@@ -368,6 +404,185 @@ export async function initRepoPushAndProtect(
...
+  let result: { createCommitOnBranch: { commit: { oid: string } } };
+  try {
+    result = await client.graphql(CREATE_COMMIT_ON_BRANCH_MUTATION, {
+      input: commitInput,
+    });
+  } catch (commitError: unknown) {
+    const msg = String((commitError as Error).message ?? '');
+    if (!msg.includes('expectedHeadOid')) {
+      throw commitError;
+    }
```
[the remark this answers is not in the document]
polasudo: The GraphQL errors[] payload for OID mismatches only carries type "UNPROCESSABLE" with no dedicated error code field name expectedHeadOid in the message is specific identifier available. Switching to structured inspection would still mean substring matching the same errors[].message, just with extra nesting. The matched string is the literal GraphQL input field name, so it's stable, and if it ever stops matching we simply don't retry and throw no silent failure.
to help our bot friend (https://github.com/graphql/graphql-spec/blob/main/spec/Section%207%20--%20Response.md)

Rugvip (approved):
Nice, I think we're good to :shipit:! 🎉 

Thank you for enduring throughout the journey 😅

polasudo:
> Nice, I think we're good to :shipit:! 🎉
> 
> Thank you for enduring throughout the journey 😅

it was my pleasure ✌️

PR #34041, opened 2026-04-23
Fix circular self-imports and add no-self-package-imports lint rule

dyatko:

Self-referential imports (a package importing itself by its own name) can surface as runtime errors such as `Cannot access 'X' before initialization` when the package is loaded via `jest.requireActual` or when an ESM consumer follows the cycle. We hit this specifically on `@backstage/frontend-plugin-api`:

> This explodes with `Cannot access '_AppRootElementBlueprintesm' before initialization` because the packaged ESM of `@backstage/frontend-plugin-api` has a self-referential import (`blueprints/AppRootElementBlueprint.esm.js` imports back from the package root). `jest.requireActual` triggers the circular init.

This PR does three things:

1. **Fixes the original crash** by replacing the self-import in `AppRootElementBlueprint.tsx` with a relative one to `ExtensionBoundary`.
2. **Cleans up similar self-imports** in a handful of other internal files. Value imports are switched to relative paths; type-only imports are switched to `import type` so they're erased at runtime and can't cycle. Affected packages: `@backstage/catalog-model`, `@backstage/core-plugin-api`, `@backstage/plugin-catalog-node`, `@backstage/plugin-kubernetes-common`, `@backstage/plugin-kubernetes-node`.
3. **Adds a new `@backstage/no-self-package-imports` ESLint rule** (`error` in the recommended config) to prevent the pattern from coming back.

### How the rule works

The rule is deliberately "smart" about entry points rather than simply banning every self-import. For each package it:

1. Reads the `package.json` `exports` map to discover every published entry and its source file.
2. Walks the relative import/export graph from each entry's source file and records which files end up in each entry's bundle.
3. For any self-import (`import ... from '@your/package[/subpath]'`), it compares the entries the current file belongs to with the entry the import targets:
   - **Same-entry** (the file is in the target entry's bundle) → error. This is a real cycle.
   - **Cross-entry** (the file is only in a different entry's bundle) → allowed by default; opt in to reporting with `allowCrossEntry: false`.
4. Files that aren't reachable from any entry (tests, scripts, orphans) are skipped, since self-imports from them can't affect a published bundle.
5. `import type` imports, `package.json` imports, and imports of other packages are always allowed.

The escape hatch is called out in the error message: if the import is only used as a type, switch it to `import type { … }` and the rule will stop reporting it.

### Packages affected

- `@backstage/eslint-plugin` (minor): adds the new rule and enables it in the recommended config.
- `@backstage/frontend-plugin-api` (patch): fixes the crash.
- `@backstage/catalog-model`, `@backstage/core-plugin-api`, `@backstage/plugin-catalog-node`, `@backstage/plugin-kubernetes-common`, `@backstage/plugin-kubernetes-node` (patch): remove self-imports. No user-facing API changes.


packages/eslint-plugin/index.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
         '@backstage/no-undeclared-imports': 'error',
         '@backstage/no-mixed-plugin-imports': 'warn',
         '@backstage/no-ui-css-imports-in-non-frontend': 'error',
+        '@backstage/no-self-package-imports': 'error',
```
freben, line 27: Just mentioning that this immediately starts affecting all users after upgrading. It's probably a good thing - but worth keeping in mind.

Rugvip (approved):
Nice! 🎉 

Yeah makes sense to force these to be `type` imports 👍

freben (approved):

mclarke47 (approved):

PR #34044, opened 2026-04-23
Simplify href resolution in `useDefinition` with `useResolvedPath`

Naycon:

Replaces `useHref` with `useResolvedPath` + `createPath` for href
resolution in the `useDefinition` hook. `useResolvedPath` returns the
resolved path without the router basename, eliminating the manual
basename detection and stripping logic.


Naycon:
Skipped changeset, because there's no net change here (the shiny new tests confirm).

Rugvip (approved):
Does this work for apps mounted on a path as well? could be good to have a few tests to cover the basic cases

Naycon:
> Does this work for apps mounted on a path as well? could be good to have a few tests to cover the basic cases

Yes, it does. The new test suite for the hook is covering this already, and passed during CI, so I think we're good to go. :)

PR #34047, opened 2026-04-23
fix(ui): fix Table width in Firefox while preserving virtualized scroll

harleyk-gusto:

Fix the `Table` component rendering too narrow in Firefox. The root cause is `overflow: auto` being set directly on the `<table>` HTML element — per the CSS spec, Firefox (correctly) applies this to the anonymous *table wrapper box*, which causes it to shrink-wrap to content width. Chrome deviates from spec and was unaffected.

The fix conditionally applies `overflow: auto` only when the table is virtualized (via a `data-virtualized` attribute). Non-virtualized tables get the Firefox width fix; virtualized tables retain `overflow: auto` so the `Virtualizer`'s scroll container continues to function.

**Before:** Table shrinks to content width in Firefox
**After:** Table fills the available container width in Firefox (matching Chrome behavior), and virtualized tables continue to scroll correctly

### DIFF
--- a/.changeset/fix-table-width-firefox.md
+++ b/.changeset/fix-table-width-firefox.md
@@ -1,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Fixed the Table component not filling its container width in Firefox. The `overflow: auto` property was incorrectly set on the `<table>` element — per the CSS spec, Firefox applies this to the anonymous table wrapper box, causing it to shrink-wrap to content width. The property has been removed; horizontal scrolling for resizable tables continues to work correctly via the `ResizableTableContainer` wrapper.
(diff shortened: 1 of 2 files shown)


harleyk-gusto:
Before:
image
After:
image

cdedreuille:
Thanks a lot for looking into this @harleyk-gusto. I believe the `overflow: auto` is needed in some cases where the table is virtualised. Here's a story example of that:
https://67584b7e8c2eb09c0422c27e-ursktqvxqz.chromatic.com/?path=/story/backstage-ui-table-dev--virtualized-table

I just made some quick tests and changing the width to 100% on the table fixes it.

@harleyk-gusto do you mind testing this a bit more?

harleyk-gusto:
I took another swing at this and made the auto a bit just for virtualized tables and validated this in Firefox myself. Hopefully this works. 🤞

clairep94:
Hi @harleyk-gusto, it seems you may have made some updates locally and forgotten to push?

cdedreuille:
@harleyk-gusto Do you have an update on this one?

Naycon:
Going to close this since we got another PR merged with a fix for this. Thanks for looking into it Harley!

PR #34067, opened 2026-04-26
BUI - Add sticky Header support

cdedreuille:

Adds sticky Header support for `@backstage/ui`, keeping the title/actions row pinned at the top of the scroll container while the rest of the header content scrolls away. The sticky row uses the bg-consumer system for surface-aware backgrounds, shows a smooth stuck-state separator, and includes stories for sticky, non-sticky, and long-title behavior.

https://github.com/user-attachments/assets/7c2c2076-b719-4e2f-aef1-95b20ae03bff

### Storybook improvements

To improve testing, this also improved our global decorators for stories to make sure we leverage our themes and not our neutral background setup. Further improvements would have to be made as separate PRs to make sure we have visual regression testing for the necessary backgrounds in key components.


packages/ui/src/components/Header/Header.module.css
```diff
@@ -18,16 +18,87 @@
...
   .bui-Header {
+    display: flex;
+    flex-direction: column;
+    padding-top: var(--bui-space-2);
+    width: 100%;
+    padding-inline: var(--bui-space-5);
+  }
+
+  .bui-Header[data-sticky] {
+    display: contents;
```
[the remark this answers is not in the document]
cdedreuille: Great call. I reformatted the content to have the header only on the sticky part which include the title and main action.

.changeset/header-sticky-prop.md
```diff
@@ -0,0 +1,7 @@
+---
+'@backstage/ui': patch
+---
+
+Added a `sticky` prop to the `Header` component. When `true`, the title-and-actions bar stays fixed to the top of its scroll container while the rest of the header (tags, description, metadata) scrolls away. The sticky bar background color automatically matches the container surface using the bg-consumer system.
```
Rugvip, line 5: How pervasive do we want this to be? Would it actually me sense to inverse this to opt-out of stickiness?

packages/ui/src/components/Header/definition.ts
```diff
@@ -24,10 +24,20 @@ import styles from './Header.module.css';
  */
 export const HeaderDefinition = defineComponent<HeaderOwnProps>()({
   styles,
+  bg: 'consumer',
   classNames: {
-    root: 'bui-Header',
```
Rugvip, line 28: Removing this is a breaking change, unless it was unused

Rugvip (commented):
Fancy 😁 

Just some small bits

.changeset/header-sticky-prop.md
cdedreuille: From a design perspective we still believe that the default should be that everything scrolls automatically. Only in a few places we might want to have these actions sticky. That said, this is a safe approach that we might update later on as this header get more visible.

packages/ui/src/components/Header/definition.ts
cdedreuille: You're right. I can consider it a breaking change from a theming perspective.

Rugvip: Gotta call out the braking change as well :p

cdedreuille: Done

Rugvip (approved):

PR #34077, opened 2026-04-27
workspaces: introduce isolated ui workspace

Rugvip:

This is the start of a new setup where we can have completely separate, standalone Yarn workspaces under a top-level `workspaces/` folder. The first one is `workspaces/ui`, which will eventually house our UI library packages.

This is intended to largely mirror the setup in the community-plugins repo, with more packages being moved in the future. Very much WIP, this just gets the basic things in place


workspaces/ui/.yarnrc.yml
```diff
@@ -0,0 +1,8 @@
+nodeLinker: node-modules
+
+plugins:
+  - checksum: a6895be0b30a75db6fddafedff143f82ccbd74656bad97775c6a0b9eba6e6ea6b7d66a8e9d0c8505755a981e86b7c6c1c9334ead299a857cf8d6e243957b0bab
+    path: .yarn/plugins/@yarnpkg/plugin-backstage.cjs
+    spec: "https://versions.backstage.io/v1/tags/main/yarn-plugin"
+
+yarnPath: ../../.yarn/releases/yarn-4.8.1.cjs
```
freben, line 16: Just a heads up that this release is like a year old and newer ones have fixes for modern node versions etc - we should bump yarn repo-wide some time soon

workspaces/ui/.yarnrc.yml
freben, line 1: Just mentioning that our root yarnrc has a couple more settings in it, not sure if it's worth unifying somehow (in either direction)

freben (approved):

workspaces/ui/.yarnrc.yml
Rugvip: updated 👍

PR #34078, opened 2026-04-27
feat(devtools): migrate InfoContent to Backstage UI

### LINKED ISSUE #32747: [DevTools] Migrate the `InfoContent` component to Backstage UI (BUI)
### 📝 Acknowledgement

### Content

The `InfoContent` component can be migrated to BUI. 

The component code is here: https://github.com/backstage/backstage/blob/master/plugins/devtools/src/components/Content/InfoContent/InfoContent.tsx

The list of available Backstage UI components is here: https://ui.backstage.io/

The process would be to just switch out the Material UI (MUI) components with their Backstage UI (BUI) versions and visually check to make sure they match.

alien2003:

Migrates the `InfoContent` component (the "About" panel on the DevTools Info page) from Material UI to Backstage UI. This continues the wider DevTools migration started in #33621 (`ExternalDependenciesContent`) and uses the same swap pattern.

Fixes #32747.

### What changed
- `Paper` + `makeStyles` row → BUI `Card` + `CardBody` with prop-based spacing.
- `List` / `ListItem` / `ListItemAvatar` / `ListItemText` row → BUI `Flex` rows of icon + bold label / secondary value, factored into a small local `InfoCell` helper.
- `Avatar` (used as a circular icon wrapper) → dropped. BUI `Avatar` is image / initials only and doesn't accept icon children, so the icons now sit inline next to their labels — same approach the other migrated DevTools content takes for non-status icons.
- `Divider` → dropped; `Flex justify="between"` separates the cells from the copy button.
- MUI `Button` (`onClick`, `startIcon`) → BUI `Button` (`onPress`, `iconStart`).
- MUI lab `Alert` for the error path → BUI `Alert status="danger" icon role="alert"`, mirroring the reference PR.
- The four cells now stack vertically below the `md` breakpoint (responsive `direction` on the outer / inner `Flex`), so the panel no longer crams onto one row on narrow viewports.

`@material-ui/icons/*` and the local `BackstageLogoIcon` (which wraps `@material-ui/core/SvgIcon`) are kept — same scope choice as #33621 which also kept its `Status*` icons. No public API change for `@backstage/plugin-devtools`.

### Screenshots

Wide viewport, light theme:

DevTools Info page, wide light theme

Wide viewport, dark theme:

DevTools Info page, wide dark theme

Narrow viewport (≤ md breakpoint), dark theme — cells stack vertically and the copy button drops to its own row:

DevTools Info page, narrow dark theme

### Verification
- `yarn tsc` clean.
- `yarn backstage-cli repo lint --since master --fix` clean.
- `CI=1 yarn test plugins/devtools/src` — both existing suites pass.
- `yarn build:api-reports` — no public-API diff for `@backstage/plugin-devtools`.
- `yarn start` — verified the four info cells render in light + dark + narrow themes, "Copy Info to Clipboard" button writes the expected `OS / Resources / node / backstage / Dependencies` block to the clipboard, and the `Package Dependencies` table below is unaffected.

- [ ] Added or updated documentation (n/a — internal component refactor).
- [ ] Tests for new functionality and regression tests for bug fixes (n/a — no behaviour change; existing devtools suites still pass).
- [x] Screenshots attached (for UI changes).

### DIFF
--- a/.changeset/devtools-info-content-bui.md
+++ b/.changeset/devtools-info-content-bui.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-devtools': patch
+---
+
+The DevTools Info page now uses Backstage UI components for its layout and the "Copy Info to Clipboard" button, bringing it in line with the rest of the migrated DevTools content.
--- a/.changeset/devtools-info-content-bui.md
+++ b/.changeset/devtools-info-content-bui.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-devtools': patch
 ---
 
-The DevTools Info page now uses Backstage UI components for its layout and the "Copy Info to Clipboard" button, bringing it in line with the rest of the migrated DevTools content.
+Migrated `InfoContent` component from Material UI to Backstage UI (BUI).
--- a/plugins/devtools/src/components/Content/InfoContent/InfoContent.tsx
+++ b/plugins/devtools/src/components/Content/InfoContent/InfoContent.tsx
@@ -110,6 +110,7 @@ export const InfoContent = () => {
                 />
               </Flex>
               <Button
+                variant="secondary"
                 onPress={() => copyToClipboard({ about })}
                 iconStart={<RiFileCopyLine />}
               >
--- a/plugins/devtools/src/components/Content/InfoContent/InfoContent.tsx
+++ b/plugins/devtools/src/components/Content/InfoContent/InfoContent.tsx
@@ -14,18 +14,9 @@
  * limitations under the License.
  */
 
+import type { ReactElement } from 'react';
 import { Progress } from '@backstage/core-components';
-import Avatar from '@material-ui/core/Avatar';
-import Box from '@material-ui/core/Box';
-import Divider from '@material-ui/core/Divider';
-import List from '@material-ui/core/List';
-import ListItem from '@material-ui/core/ListItem';
-import ListItemAvatar from '@material-ui/core/ListItemAvatar';
-import ListItemText from '@material-ui/core/ListItemText';
-import Paper from '@material-ui/core/Paper';
-import Button from '@material-ui/core/Button';
-import { createStyles, makeStyles, Theme } from '@material-ui/core/styles';
-import Alert from '@material-ui/lab/Alert';
+import { Alert, Box, Button, Card, CardBody, Flex, Text } from '@backstage/ui';
 import { useInfo } from '../../../hooks';
 import { InfoDependenciesTable } from './InfoDependenciesTable';
 import DescriptionIcon from '@material-ui/icons/Description';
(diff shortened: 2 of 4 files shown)


.changeset/devtools-info-content-bui.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-devtools': patch
+---
+
+The DevTools Info page now uses Backstage UI components for its layout and the "Copy Info to Clipboard" button, bringing it in line with the rest of the migrated DevTools content.
```
awanlin:
```suggestion
Migrated `InfoContent` component from Material UI to Backstage UI (BUI).
```

Let's keep this short and simple like we are doing for the other migrations.

plugins/devtools/src/components/Content/InfoContent/InfoContent.tsx
```diff
@@ -35,24 +26,6 @@ import { BackstageLogoIcon } from './BackstageLogoIcon';
 import FileCopyIcon from '@material-ui/icons/FileCopy';
```
awanlin: We want to replace the icons with those from React Remix Icon, can you find matching versions to use: https://remixicon.com/

awanlin (changes requested):
Thanks for the contributions @alien2003, left a few comments. I'm not sure I like how the "Copy Info to Clipboard" looks but I don't have a alternative right now. 🤔

.changeset/devtools-info-content-bui.md
alien2003: done, thanks

plugins/devtools/src/components/Content/InfoContent/InfoContent.tsx
alien2003: swapped them:

- DeveloperBoard → RiComputerLine
- Memory → RiCpuLine
- Description → RiNodejsLine
- FileCopy → RiFileCopyLine

kept BackstageLogoIcon since it's the local SVG. added @remixicon/react to the plugin's deps using the same range as @backstage/ui.

alien2003:
thanks! pushed c5d90b6 with both fixes. no strong opinion on the copy button placement — let me know if you want it moved.

plugins/devtools/src/components/Content/InfoContent/InfoContent.tsx
```diff
@@ -66,76 +41,84 @@ const copyToClipboard = ({ about }: { about: DevToolsInfo | undefined }) => {
...
+                  label="NodeJS Version"
+                  value={about?.nodeJsVersion}
+                />
+                <InfoCell
+                  icon={<BackstageLogoIcon />}
+                  label="Backstage Version"
+                  value={about?.backstageVersion}
+                />
+              </Flex>
+              <Button
```
awanlin, line 112:
```suggestion
              <Button variant="secondary"
```

That will be the better option, it really jumps out the with the default of primary. https://ui.backstage.io/components/button#examples

awanlin (changes requested):
Awesome, just one last comment and we should be good to merge this in @alien2003 👍

plugins/devtools/src/components/Content/InfoContent/InfoContent.tsx
alien2003: done

alien2003:
@awanlin thanks! pushed in 5cc971b

PR #34081, opened 2026-04-28
fix(catalog): show not-found instead of falling back to first route on unknown entity sub-paths

benjdlambert:

Navigating to an unknown sub-path on a new-frontend-system entity page (for example `/catalog/default/component/foo/blob`) currently silently renders the first available content tab instead of returning a not-found state. This makes broken or stale links look fine, just on the wrong tab.

Two compounding causes inside `useSelectedSubRoute`:

1. The default overview content registers `path: '/'`. The hook builds a route pattern as `${path}/*`, which produced `//*` and effectively acted as a catch-all for any sub-path.
2. When `useRoutes` returned `null`, the hook fell back to `subRoutes[0]?.children`, masking the miss.

This PR:

- Normalizes paths so `/` becomes an index-style route (matches root only) and other paths keep their `/*` suffix for nested matching.
- Removes the silent fallback to the first sub-route's element.
- Renders a "Route not found" `WarningPanel` inside the entity panel when nothing matches, so the entity header and tabs remain available.
- Updates `EntityTabsList` to handle `selectedIndex = -1` cleanly (no MUI value warnings).

The legacy `RoutedTabs` in `core-components` has the same logic but is not touched here — it's used much more broadly and changing it would be a wider behavior change.

### DIFF
--- a/plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
+++ b/plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
@@ -16,12 +16,10 @@
 import { ReactElement, useMemo } from 'react';
 import { Helmet } from 'react-helmet';
 import { matchRoutes, useParams, useRoutes } from 'react-router-dom';
-import { WarningPanel, Content } from '@backstage/core-components';
-import { useTranslationRef } from '@backstage/core-plugin-api/alpha';
+import { NotFoundErrorPage } from '@backstage/frontend-plugin-api';
 import { EntityTabsPanel } from './EntityTabsPanel';
 import { EntityTabsList } from './EntityTabsList';
 import { EntityContentGroupDefinitions } from '@backstage/plugin-catalog-react/alpha';
-import { catalogTranslationRef } from '../../translation';
 
 type SubRoute = {
   group?: string;
@@ -49,23 +47,14 @@ export function useSelectedSubRoute(subRoutes: SubRoute[]): {
     element: children,
   }));
 
-  // Sort by descending path length so more specific paths match before shorter
-  // prefixes; fall back to lexicographic order for stable output when lengths
-  // are equal.
-  const sortedRoutes = [...routes].sort((a, b) => {
-    const aPath = a.path.replace(/\/\*$/, '');
-    const bPath = b.path.replace(/\/\*$/, '');
-    return bPath.length - aPath.length || aPath.localeCompare(bPath);
-  });
-
-  const element = useRoutes(sortedRoutes) ?? undefined;
+  const element = useRoutes(routes) ?? undefined;
 
   let currentRoute = params['*'] ?? '';
   if (!currentRoute.startsWith('/')) {
     currentRoute = `/${currentRoute}`;
   }
 
-  const [matchedRoute] = matchRoutes(sortedRoutes, currentRoute) ?? [];
+  const [matchedRoute] = matchRoutes(routes, currentRoute) ?? [];
   const foundIndex = matchedRoute
     ? subRoutes.findIndex(
         t => normalizeRoutePath(t.path) === matchedRoute.route.path,
@@ -88,7 +77,6 @@ type EntityTabsProps = {
 
 export function EntityTabs(props: EntityTabsProps) {
   const { routes, groupDefinitions, defaultContentOrder, showIcons } = props;
-  const { t } = useTranslationRef(catalogTranslationRef);
 
   const { index, route, element } = useSelectedSubRoute(routes);
 
@@ -123,14 +111,7 @@ export function EntityTabs(props: EntityTabsProps) {
       />
       <EntityTabsPanel>
         <Helmet title={route?.title} />
-        {element ?? (
-          <Content>
-            <WarningPanel
-              title={t('entityTabs.notFoundTitle')}
-              message={t('entityTabs.notFoundMessage')}
-            />
-          </Content>
-        )}
+        {element ?? <NotFoundErrorPage />}
       </EntityTabsPanel>
     </>
   );
--- a/plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
+++ b/plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
@@ -30,7 +30,7 @@ type SubRoute = {
 };
 
 function normalizeRoutePath(path: string): string {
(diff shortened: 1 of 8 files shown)


plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
```diff
@@ -112,7 +119,14 @@ export function EntityTabs(props: EntityTabsProps) {
...
         <Helmet title={route?.title} />
-        {element}
+        {element ?? (
+          <Content>
+            <WarningPanel
+              title={t('entityTabs.notFoundTitle')}
+              message={t('entityTabs.notFoundMessage')}
+            />
+          </Content>
+        )}
```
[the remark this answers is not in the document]
Rugvip: yep ☝️

plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
```diff
@@ -112,7 +119,14 @@ export function EntityTabs(props: EntityTabsProps) {
       />
       <EntityTabsPanel>
         <Helmet title={route?.title} />
-        {element}
+        {element ?? (
+          <Content>
+            <WarningPanel
+              title={t('entityTabs.notFoundTitle')}
+              message={t('entityTabs.notFoundMessage')}
+            />
```
Rugvip: Not the 404 page? 🤔

plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
```diff
@@ -37,35 +45,37 @@ export function useSelectedSubRoute(subRoutes: SubRoute[]): {
...
   }));
 
-  // TODO: remove once react-router updated
-  const sortedRoutes = routes.sort((a, b) =>
-    // remove "/*" symbols from path end before comparing
-    b.path.replace(/\/\*$/, '').localeCompare(a.path.replace(/\/\*$/, '')),
-  );
+  // Sort by descending path length so more specific paths match before shorter
+  // prefixes; fall back to lexicographic order for stable output when lengths
+  // are equal.
```
Rugvip: I think react-router already does this by now? 🤔 in `matchRoutes`

plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
```diff
@@ -37,35 +45,37 @@ export function useSelectedSubRoute(subRoutes: SubRoute[]): {
...
     currentRoute = `/${currentRoute}`;
   }
 
   const [matchedRoute] = matchRoutes(sortedRoutes, currentRoute) ?? [];
   const foundIndex = matchedRoute
-    ? subRoutes.findIndex(t => `${t.path}/*` === matchedRoute.route.path)
-    : 0;
+    ? subRoutes.findIndex(
+        t => normalizeRoutePath(t.path) === matchedRoute.route.path,
+      )
```
Rugvip: hold up, is this going to handle subpaths within the entity page tabs correctly?

plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
```diff
@@ -37,35 +45,33 @@ export function useSelectedSubRoute(subRoutes: SubRoute[]): {
...
     element: children,
   }));
 
-  // TODO: remove once react-router updated
-  const sortedRoutes = routes.sort((a, b) =>
-    // remove "/*" symbols from path end before comparing
+  // Sort longer paths first so more specific matches win.
+  const sortedRoutes = [...routes].sort((a, b) =>
     b.path.replace(/\/\*$/, '').localeCompare(a.path.replace(/\/\*$/, '')),
   );
```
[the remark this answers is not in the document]
benjdlambert: Done — switched to length-based sort... actually went further per your other comment and removed the manual sort entirely.

plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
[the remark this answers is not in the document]
benjdlambert: Added — there are now `describe('rendering')` tests in EntityTabs.test.tsx that render the actual EntityTabs component via renderInTestApp and assert the not-found page renders for unknown sub-paths.

plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
benjdlambert: Switched to `<NotFoundErrorPage />` from `@backstage/frontend-plugin-api` — no more in-panel WarningPanel, just the standard 404 page rendered inside the entity panel.

plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
benjdlambert: You're right — dropped the manual sort. `useRoutes`/`matchRoutes` rank by specificity in v6, and tests confirm the correct route still wins for nested paths.

plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
benjdlambert: Yep, looks like it does with this test right? https://github.com/backstage/backstage/pull/34081/changes#diff-0fb7966bd0d7db0e85446478ff13a370a3f95a1dcba4f4402044f782aedc0af2R417-R449

plugins/catalog/src/alpha/components/EntityTabs/EntityTabs.tsx
```diff
@@ -28,44 +29,47 @@ type SubRoute = {
   children: JSX.Element;
 };
 
+// Strip leading and trailing slashes, then append `/*` for nested matching.
+// A bare `/` becomes the empty string so it acts as an index route instead of
+// a wildcard that would catch every sub-path.
+function normalizeRoutePath(path: string): string {
+  const trimmed = path.replace(/^\/+/, '').replace(/\/+$/, '');
+  return trimmed ? `${trimmed}/*` : '';
```
freben, line 45: I just realised - if the input was `/*`, the output would be `*/*`. Is that OK? Do people not intentionally put explicit glob matchers at the end sometimes? Maybe they arguably shouldn't for tabs but ... should we handle it gracefully

freben (approved):

PR #34087, opened 2026-04-28
feat: `TracingService` to match `MetricService`

iamEAP:
## What / Why

I'd like to start introducing some more span instrumentation throughout Backstage. Before adding yet more direct dependencies on otel APIs and more ad-hoc span instantiation, I think it probably makes sense to centralize this into a service, similar to the `MetricsService`. Just a thin wrapper around OTel.

This puts the framework in place. I believe it would be immediately possible to migrate the spans instrumented in the Catalog (but will defer that for later). The spans instrumented in the task scheduler would likely require a root equivalent of this service, with more flexible tracer naming (also deferred for later).

### DIFF
--- a/docs/backend-system/core-services/tracing.md
+++ b/docs/backend-system/core-services/tracing.md
@@ -1,0 +1,161 @@
+---
+id: tracing
+title: Tracing Service (alpha)
+sidebar_label: Tracing Service (alpha)
+description: Documentation for the Tracing service
+---
+
+The Tracing Service provides a unified interface for emitting application-level [OpenTelemetry](https://opentelemetry.io/) trace spans from Backstage backend plugins. It scopes each plugin's spans automatically using the OpenTelemetry [Instrumentation Scope](https://opentelemetry.io/docs/concepts/instrumentation-scope/), wraps span lifecycle (auto-end, exception recording, error status) so plugins don't need to write that boilerplate, and transparently enriches spans with the authenticated principal's identity when an HTTP request or `BackstageCredentials` is supplied.
+
+:::note
+This service is currently in **alpha** and is imported from `@backstage/backend-plugin-api/alpha`. The API may change in future releases.
+:::
+
+## Setting up OpenTelemetry
+
+The Tracing Service does **not** configure the OpenTelemetry SDK itself. You are responsible for initializing the OpenTelemetry Node SDK — including exporters, samplers, and resource attributes — before starting the Backstage backend. Follow the [tutorial](../../tutorials/setup-opentelemetry.md) for more information.
+
+## How it Relates to OpenTelemetry Auto-Instrumentation
+
+The Tracing Service **complements** auto-instrumentation rather than replacing it. Auto-instrumentation captures infrastructure-level spans like inbound HTTP requests, outbound HTTP calls, and database queries automatically — including all the standard HTTP / DB attributes. The Tracing Service is for **application-level spans** that only your plugin can produce, and child spans you want to attach to that infrastructure work.
+
+Because HTTP spans are auto-instrumented, you typically should **not** set `http.*` attributes on Tracing Service spans yourself — the parent HTTP span already carries them. Spans you create are children of that HTTP span, in the same trace.
+
+## Using the Service
+
+Since the Tracing Service is an alpha API, the service reference is imported from `@backstage/backend-plugin-api/alpha` instead of `coreServices`.
+
+```ts
+import { createBackendPlugin } from '@backstage/backend-plugin-api';
+import { tracingServiceRef } from '@backstage/backend-plugin-api/alpha';
+
+createBackendPlugin({
+  pluginId: 'todos',
+  register(env) {
+    env.registerInit({
+      deps: { tracing: tracingServiceRef },
+      async init({ tracing }) {
+        // ... wire up your routes/handlers, holding onto `tracing` ...
+
+        const result = await tracing.startActiveSpan(
+          'process-todo',
+          async span => {
+            span.setAttribute('todo.category', 'personal');
+            // ...do the work...
+            return computeResult();
+          },
+        );
+      },
+    });
+  },
+});
+```
+
+`startActiveSpan(name, fn, options?)` runs `fn` inside a new active span. The span is finished automatically when `fn` resolves, and on a thrown error the exception is recorded, `error.type` is set from the error's `name`, and the span status is set to `ERROR` — you do not need to write a `try/catch/finally` for that.
+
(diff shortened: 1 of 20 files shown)


packages/backend-plugin-api/src/alpha/TracingService.ts
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @alpha
+ */
+export type TracingServiceAttributeValue =
+  | string
+  | number
+  | boolean
+  | Array<null | undefined | string>
+  | Array<null | undefined | number>
+  | Array<null | undefined | boolean>;
```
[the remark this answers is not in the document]
iamEAP: This is the exact shape OTel itself uses from `@opentelemetry/api`.

packages/backend-plugin-api/src/alpha/TracingService.ts
```diff
@@ -0,0 +1,109 @@
...
+  attributes?: TracingServiceAttributes;
+  /** The kind of span. */
+  kind?: TracingServiceSpanKind;
+  /**
+   * Authenticated principal source for span enrichment. Preferred when
+   * both `credentials` and `request` are supplied.
+   */
+  credentials?: BackstageCredentials;
+  /** HTTP request to extract credentials from for span enrichment. */
+  request?: Request<any, any, any, any, any>;
```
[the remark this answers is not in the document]
iamEAP: Suggestion would be more restrictive. This is how other services declare similar dependencies.

packages/backend-test-utils/src/alpha/services/TracingServiceMock.ts
```diff
@@ -0,0 +1,89 @@
...
+ * methods called on the span inside the callback.
+ *
+ * @alpha
+ */
+export interface TracingServiceMock extends TracingService {
+  startActiveSpan: jest.MockedFunction<TracingService['startActiveSpan']>;
+  /** Spans created by `startActiveSpan` calls, in order. */
+  spans: MockedTracingServiceSpan[];
+  factory: ServiceFactory<TracingService>;
+}
```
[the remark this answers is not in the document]
iamEAP: startActiveSpan is a mocked jest function and can be asserted on as described.

awanlin (approved):
Thanks @iamEAP, this is a solid addition, thanks for the well thought out docs as well! 🚀

docs/backend-system/core-services/tracing.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Span Options
+
+The third argument to `startActiveSpan` is an optional options object:
+
+| Property      | Type                       | Description                                                                                          |
+| ------------- | -------------------------- | ---------------------------------------------------------------------------------------------------- |
+| `attributes`  | `TracingServiceAttributes` | Attributes to attach to the span at creation time.                                                   |
+| `kind`        | `TracingServiceSpanKind`   | Span kind. Defaults to OpenTelemetry's `internal`. See [Span Kinds](#span-kinds).                    |
+| `credentials` | `BackstageCredentials`     | Authenticated principal source — adds principal-derived attributes to the span.                      |
```
freben, line 66: But maybe it's fair to say that just passing in `request` is the norm?

packages/backend-defaults/src/alpha/entrypoints/tracing/tracingServiceFactory.ts
```diff
@@ -0,0 +1,59 @@
...
+    const tracerConfig = config.getOptionalConfig(
+      `backend.tracing.plugin.${pluginId}.tracer`,
+    );
+    const scopeName = `backstage-plugin-${pluginId}`;
+    const name = tracerConfig?.getOptionalString('name') ?? scopeName;
+    const version = tracerConfig?.getOptionalString('version');
+    const schemaUrl = tracerConfig?.getOptionalString('schemaUrl');
+
+    const captureEndUser =
+      config.getOptionalBoolean('backend.tracing.capture.endUser') ?? false;
```
benjdlambert, line 48: should this just be `captureEndUser`? or do you see a need for a `capture` block?

benjdlambert (approved):
Looks like solid stuff to me :tada:

Nice work - just a small optional comment about the config.

packages/backend-defaults/src/alpha/entrypoints/tracing/DefaultTracingService.ts
```diff
@@ -0,0 +1,171 @@
...
+                code: toSpanStatusCode(status.code),
+                message: status.message,
+              });
+            },
+          };
+          const result = await fn(wrapped);
+          span.end();
+          return result;
+        } catch (err) {
+          const error = err as Error;
```
freben, line 106: This could be improved by branching on `isError` from `@backstage/errors` and carefully using `message` or `String(error)` as needed based on that.

freben (approved):
Nice!

packages/backend-plugin-api/report-alpha.api.md
```diff
@@ -10,6 +10,7 @@ import { JsonObject } from '@backstage/types';
 import { JSONSchema7 } from 'json-schema';
 import { JsonValue } from '@backstage/types';
 import { LoggerService } from '@backstage/backend-plugin-api';
+import type { Request as Request_2 } from 'express';
```
freben, line 13: One more thought - I wonder if this could even be reduced down to being a Node request instead of express. But maybe we can't.

freben:
Merging this to get into the release, this is all alpha so can be tweaked as followups. Thanks for this!

packages/backend-defaults/src/alpha/entrypoints/tracing/tracingServiceFactory.ts
iamEAP: Yeah, good question. I add `capture.toolPayloads` in a follow-up specific to MCP, so it made sense to me to nest it. No strong opinion though.

packages/backend-plugin-api/src/alpha/TracingService.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+export interface TracingService {
+  /**
+   * Runs `fn` inside a new active span. The span is finished when `fn`
+   * resolves or throws.
+   */
+  startActiveSpan<T>(
+    name: string,
+    fn: (span: TracingServiceSpan) => T | Promise<T>,
+    options?: TracingServiceSpanOptions,
```
Rugvip, line 107: Was the switch here to avoid having the overload that moves the param? If we're gonna change it, another option would be `name | { name, ...options }`

packages/backend-plugin-api/report-alpha.api.md
Rugvip: Tried that at some point for the auth service iirc, without success. Might misremember though. Could definitely be reduced down to the minimal methods/props needed though. Either way best to stick to what we do elsewhere for now

packages/backend-plugin-api/src/alpha/TracingService.ts
iamEAP: Yeah, pretty much. Would be fine with that too.

PR #34089, opened 2026-04-28
feat: Instrument MCP tool calls with semantically appropriate span

iamEAP:
## What / Why

Adds OTel trace spans for MCP tools/call invocations in `@backstage/plugin-mcp-actions-backend`, following [relevant semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#server). Built on the new `TracingService` so principal enrichment, span lifecycle, and error handling are inherited rather than reimplemented here.

Each `tools/call` produces a `SERVER` span named `tools/call <toolname>` with the spec attributes (`mcp.method.name`, `gen_ai.tool.name`, `gen_ai.operation.name`). When a tool result returns `isError: true`, the span is marked with `error.type=tool_error` and `ERROR` status; thrown handler errors are auto-recorded by the service.

A new for-purpose flag controls capture of potentially-sensitive and high-cardinality payloads:

```yaml
backend:                                                                           
  tracing:                                             
    capture:
      toolPayloads: true # defaults to false
```

When enabled, `gen_ai.tool.call.arguments` and `gen_ai.tool.call.result` are added to the span. These attributes are marked Opt-In by the OpenTelemetry GenAI semconv because they may contain entity payloads, scaffolder inputs, or other sensitive content. End-user identity (`enduser.id`) is gated by the cross-cutting `backend.tracing.capture.endUser` flag, honored by the new service.

`backstage.plugin.id` overrides the default `mcp-actions` in favor of the plugin that actually owns the action.

The span also propagates and merges in `gen_ai*` attributes from callers, if they provide them via baggage headers.

Developed on top of https://github.com/backstage/backstage/pull/34087


benjdlambert:
@kurtaking I know you've been over a little bit the instrumentation with the `MetricsService`, no pressure, but wanted to bring this to you too. Any thoughts? :pray:

iamEAP:
Not necessary for today's release. ~~Note: I'm rebasing and also looking at adding baggage propagation so spans can pick up stuff like `gen_ai.conversation.id` and `gen_ai.agent.id` from callers, if they supply it.~~ **edit**: Updated with this logic!

packages/backend-plugin-api/report-alpha.api.md
```diff
@@ -292,11 +292,16 @@ export const rootSystemMetadataServiceRef: ServiceRef<
...
+  getActiveBaggage(): TracingServiceBaggage | undefined;
   startActiveSpan<T>(
     name: string,
     fn: (span: TracingServiceSpan) => T | Promise<T>,
     options?: TracingServiceSpanOptions,
   ): Promise<T>;
+  withPropagatedContext<T>(
+    headers: Record<string, string | string[] | undefined>,
+    fn: () => T | Promise<T>,
+  ): Promise<T>;
```
Rugvip: Are these based on the otel SDK? can't find the counterparts in that case. Would like to keep this as close to otel as possible

packages/backend-plugin-api/report-alpha.api.md
```diff
@@ -314,6 +319,20 @@ export type TracingServiceAttributeValue =
   | Array<null | undefined | number>
   | Array<null | undefined | boolean>;
 
+// @alpha
+export interface TracingServiceBaggage {
+  // (undocumented)
+  getAllEntries(): Array<[string, TracingServiceBaggageEntry]>;
+  // (undocumented)
+  getEntry(key: string): TracingServiceBaggageEntry | undefined;
```
Rugvip: This seems to expose too much information? 🤔 what doesn't the upstream equivalent look like?

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -136,43 +187,82 @@ export class McpService {
...
+            credentials,
+            attributes: {
+              ...baggageAttributes(this.tracingService),
+              'mcp.method.name': 'tools/call',
+              'gen_ai.tool.name': params.name,
+              'gen_ai.operation.name': 'execute_tool',
+              ...(this.captureToolPayloads && {
+                'gen_ai.tool.call.arguments': safeStringify(params.arguments),
+              }),
+            },
```
Rugvip: Hmm, I'm kinda feeling the reordering of the params here makes it a lot harder to read. What do you think about switching to one of https://github.com/backstage/backstage/pull/34087#discussion_r3226224109?

docs/backend-system/core-services/tracing.md
```diff
@@ -100,6 +100,41 @@ The span object exposes:
 | `setAttribute(key, value)`     | Set a single attribute. Value is a primitive or array of primitives. |
 | `setStatus({ code, message })` | Set the span status. `code` is `'ok'`, `'error'`, or `'unset'`.      |
 
+## Context Propagation
+
+When your plugin receives requests through a protocol layer that breaks automatic OpenTelemetry context propagation (e.g. a WebSocket handler), use `withPropagatedContext` to extract the trace parent and baggage from the incoming HTTP headers and run the handler within that context:
```
Rugvip: This example seems a bit off given that web-WebSockets can't have custom headers? 🤔 That also makes me a bit worried about the interface and assumption that headers are being passed, although not quite sure what the alternatives are. Would seem as though you'd at least want `{ headers: ... }` or something to allow for more explicit options. Again, whatever otel does takes prio though

Rugvip (changes requested):
Highest prio bits for now

packages/backend-plugin-api/report-alpha.api.md
iamEAP: Yeah, admittedly it's a bit of a blend. The API surface might get a bit more complicated if we break it out, but maybe not so bad. Like:

```typescript
interface TracingService {
  startActiveSpan<T>(name, fn, options?): Promise<T>;
  readonly context: TracingServiceContextAPI;
  readonly propagation: TracingServicePropagationAPI;
}

interface TracingServiceContextAPI {
  active(): TracingServiceContext;
  with<T>(ctx: TracingServiceContext, fn: () => T | Promise<T>): Promise<T>;
}

interface TracingServicePropagationAPI {
  extract(
    ctx: TracingServiceContext,
    carrier: Record<string, string | string[] | undefined>,
  ): TracingServiceContext;
  getActiveBaggage(): TracingServiceBaggage | undefined;
  getBaggage(ctx: TracingServiceContext): TracingServiceBaggage | undefined;
}

interface TracingServiceContext { /* etc */ }
```

packages/backend-plugin-api/report-alpha.api.md
iamEAP: Pretty much identical.

```ts
export interface Baggage {
  getEntry(key: string): BaggageEntry | undefined;
  getAllEntries(): [string, BaggageEntry][];
  // etc.
}

export interface BaggageEntry {
  /** `String` value of the `BaggageEntry`. */
  value: string;
}
```

docs/backend-system/core-services/tracing.md
iamEAP: I think the upstream `propagation.extract` API allows you to pass an optional function that knows how to extract whatever you pass it (defaulting to something that understands the shape given today). ...I'd say we avoid exposing it until we need it?  The definition there is flexible enough to handle a variety of common cases (including headers), and it doesn't seem so bad for callers to do the translation themselves before they pass it in.

iamEAP: Also updated docs with more realistic examples.

docs/ai/mcp-actions.md
```diff
@@ -292,3 +292,51 @@ The MCP Actions Backend emits metrics for the following operations:
...
+
+### Capturing tool arguments and results
+
+When `backend.tracing.capture.toolPayloads` is enabled, the tool's input arguments and output result are recorded on the span as `gen_ai.tool.call.arguments` and `gen_ai.tool.call.result`.
+
+```yaml title="app-config.yaml"
+backend:
+  tracing:
+    capture:
+      toolPayloads: true # defaults to false
```
Rugvip: Ah, should've mentioned this in the first round: this doesn't seem like a great pattern at scale? 🤔 Feeling that it's lacking context and would either need to be a lot more verbose at this level, or be rebased. i.e. either `backend.tracing.capture.mcpActionsToolPayloads` or `mcpActions.tracing.capture.toolPayloads` / `mcpActions.tracing.captureToolPayloads`

packages/backend-plugin-api/src/alpha/TracingService.ts
```diff
@@ -104,6 +104,107 @@ export interface TracingService {
...
+/**
+ * Opaque handle representing a tracing context. Consumers receive
+ * these from {@link TracingServiceContextAPI.active} or
+ * {@link TracingServicePropagationAPI.extract} and pass them back into
+ * the API; the type carries no inspectable fields.
+ *
+ * @alpha
+ */
+export interface TracingServiceContext {
+  readonly [tracingServiceContextBrand]: never;
```
Rugvip: I think it's alright to use our own variant for opaque types here: `$$type: '@backstage/TracingServiceContext'`. It's more resilient and I'm a bit worried we'll run into package duplication issues or other fun errors if we end up using a symbol here.

packages/backend-plugin-api/src/alpha/TracingService.ts
```diff
@@ -104,6 +104,107 @@ export interface TracingService {
...
+}
+
+/**
+ * A read-only view of propagated baggage entries.
+ *
+ * @alpha
+ */
+export interface TracingServiceBaggage {
+  getEntry(key: string): TracingServiceBaggageEntry | undefined;
```
Rugvip: Do you feel we need `getEntry`? I'm kinda leaning towards only implementing `getAllEntries` for now to nudge towards this only being used for bridging and no other silly custom context shenanigans 😁

Rugvip (changes requested):
Nice! 👍 

Few more bits, but feels safer to mirror otel, even if doesn't align quite as well with our other API design style

docs/ai/mcp-actions.md
iamEAP: I'm down with making this super verbose. Thinking tracing configs are going to be a concern of a single team in most cases, so co-locating them feels best ergonomically.

packages/backend-plugin-api/src/alpha/TracingService.ts
iamEAP: Yeah, good call.

docs/ai/mcp-actions.md
Rugvip: So thinking a bit more about this, I think we should either keep it generic and in the `backend` section, and kind of aim for a collection of well-known toggles, or we move it to be plugin configuration. I don't like the idea of having very explicit per plug-in configuration live within the backend space. Realistically I don't think it's really worth aiming for a set of common configs here though, and a lot easier to keep it plugin-specific. This is with the context that we'd move to better isolating plugin specific config in the future too.

Rugvip (approved):
:shipit:

plugins/mcp-actions-backend/src/services/McpService.ts
```diff
@@ -34,18 +35,63 @@ import { handleErrors } from './handleErrors';
...
+}
+
 export class McpService {
   private readonly actions: ActionsService;
   private readonly namespacedToolNames: boolean;
+  private readonly tracingService: TracingService;
+  private readonly captureToolPayloads: boolean;
   private readonly operationDuration: MetricsServiceHistogram<McpServerOperationAttributes>;
 
   constructor(
```
benjdlambert, line 84: This should really be `private`, but I think we can fix that in a followup.

benjdlambert (approved):
Just one thing, but it's not really related to this change so much.

Nice stuff :tada:

kurtaking:
> @kurtaking I know you've been over a little bit the instrumentation with the `MetricsService`, no pressure, but wanted to bring this to you too. Any thoughts? 🙏

I was out last week. It was on my todo list to look at this this morning. Sorry about the delay on that.

packages/backend-plugin-api/src/alpha/refs.ts
```diff
@@ -71,7 +71,7 @@ export const metricsServiceRef = createServiceRef<
 /**
  * Service for managing trace spans.
  *
- * See {@link TracingService} for the API surface.
+ * See `TracingService` for the API surface.
```
kurtaking, line 74: Why was this removed? 🤔

kurtaking (commented):
I don't have as much familiarity with tracing, but the service and attribution alignment with OTel looks good. 

Solid feature, nice work @iamEAP 🚀 The combination of metrics and tracing across the framework will be huge. 🤝

PR #34096, opened 2026-04-29
Fixed a heading case inconsistency

AmbrishRamachandiran:
Fixed a heading case inconsistency in two `docs-ui` component pages. The `container` and `table` pages used `## API Reference` (capital R) while all 37 other component pages use `## API reference`.


Rugvip (changes requested):
Doesn't seem like these are inconsistent at all? The rest of them also use title case

Rugvip (commented):
ah yep there's a bigger mix, we'd need to align all the headings on these pages too imo, but it does seem like lowercase or more commonly used in docs-ui too

PR #34108, opened 2026-04-30
cli-module-build: forward user config to embedded Postgres

Rugvip:

This updates the embedded Postgres setup used during `yarn start` so that user-provided connection configuration is forwarded to the embedded instance. Previously, the CLI would unconditionally override all connection values (host, port, user, password) with its own defaults. Now, if you've configured any of these values under `backend.database.connection`, the embedded Postgres instance will start using those values, and only the ones you haven't configured will be filled in with defaults.

This makes it possible to run the embedded database on a specific host/port and connect to it externally (e.g. with `psql`) while it's running.

### DIFF
--- a/packages/cli-module-build/src/lib/runner/startEmbeddedDb.ts
+++ b/packages/cli-module-build/src/lib/runner/startEmbeddedDb.ts
@@ -16,9 +16,14 @@
 
 import os from 'node:os';
 import fs from 'fs-extra';
-import { resolve as resolvePath } from 'node:path';
+import {
+  isAbsolute as isAbsolutePath,
+  resolve as resolvePath,
+} from 'node:path';
 import { getPortPromise } from 'portfinder';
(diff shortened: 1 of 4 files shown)


packages/cli-module-build/src/lib/runner/startEmbeddedDb.ts
```diff
@@ -72,10 +115,10 @@ export async function startEmbeddedDb() {
 
   await cleanStaleDatabases();
 
-  const host = 'localhost';
-  const user = 'postgres';
-  const password = 'password';
-  const port = await getPortPromise();
+  const host = userConfig?.host ?? 'localhost';
+  const user = userConfig?.user ?? 'postgres';
+  const password = userConfig?.password ?? 'password';
```
Rugvip:
```suggestion
  const password = userConfig?.password ?? 'postgres';
```

freben (approved):

PR #34118, opened 2026-05-04
Add Combobox component to Backstage UI

jabrks:

Introduces a Combobox component to Backstage UI for times when you want to allow users to choose from a list of values but also specify their own in some cases.

### DIFF
--- a/packages/ui/src/components/Combobox/Combobox.tsx
+++ b/packages/ui/src/components/Combobox/Combobox.tsx
@@ -15,13 +15,12 @@
  */
 
 import { forwardRef, useEffect } from 'react';
-import { ComboBox as AriaComboBox, Popover } from 'react-aria-components';
+import { ComboBox as AriaComboBox } from 'react-aria-components';
 import { useFilter } from 'react-aria';
-import clsx from 'clsx';
 import { ComboboxProps } from './types';
 import { useDefinition } from '../../hooks/useDefinition';
 import { ComboboxDefinition } from './definition';
-import { PopoverDefinition } from '../Popover/definition';
+import { Popover } from '../Popover';
 import { FieldLabel } from '../FieldLabel';
 import { FieldError } from '../FieldError';
 import { ComboboxInput } from './ComboboxInput';
@@ -41,8 +40,6 @@ export const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
       ComboboxDefinition,
       props,
     );
-    const { ownProps: popoverOwnProps } = useDefinition(PopoverDefinition, {});
-
     const {
       classes,
       label,
@@ -84,10 +81,7 @@ export const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
         />
         <ComboboxInput icon={icon} placeholder={placeholder} />
         <FieldError />
-        <Popover
-          className={clsx(popoverOwnProps.classes.root, classes.popover)}
-          {...dataAttributes}
-        >
+        <Popover className={classes.popover} hideArrow {...dataAttributes}>
           <ComboboxListBox options={options} />
         </Popover>
       </AriaComboBox>
--- a/packages/ui/src/components/Combobox/Combobox.tsx
+++ b/packages/ui/src/components/Combobox/Combobox.tsx
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2026 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { forwardRef, useEffect } from 'react';
+import { ComboBox as AriaComboBox, Popover } from 'react-aria-components';
+import { useFilter } from 'react-aria';
+import clsx from 'clsx';
+import { ComboboxProps } from './types';
+import { useDefinition } from '../../hooks/useDefinition';
+import { ComboboxDefinition } from './definition';
+import { PopoverDefinition } from '../Popover/definition';
+import { FieldLabel } from '../FieldLabel';
+import { FieldError } from '../FieldError';
+import { ComboboxInput } from './ComboboxInput';
+import { ComboboxListBox } from './ComboboxListBox';
+
+/**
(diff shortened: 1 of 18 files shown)


packages/ui/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,98 @@
...
+      }
+    }, [label, ariaLabel, ariaLabelledBy]);
+
+    const secondaryLabelText =
+      secondaryLabel || (isRequired ? 'Required' : null);
+
+    return (
+      <AriaComboBox
+        className={classes.root}
+        defaultFilter={contains}
```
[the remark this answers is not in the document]
jabrks: Matches the Select component, if it needs fixing it should be addressed in both

packages/ui/src/components/Combobox/types.ts
```diff
@@ -0,0 +1,76 @@
...
+  secondaryLabel?: FieldLabelProps['secondaryLabel'];
+  description?: FieldLabelProps['description'];
+  isRequired?: boolean;
+  className?: string;
+};
+
+/** @public */
+export interface ComboboxProps
+  extends ComboboxOwnProps,
+    Omit<AriaComboBoxProps<Option>, keyof ComboboxOwnProps> {}
```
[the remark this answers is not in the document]
jabrks: Matches the Select component

packages/ui/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,98 @@
...
+      >
+        <FieldLabel
+          label={label}
+          secondaryLabel={secondaryLabelText}
+          description={description}
+          descriptionSlot="description"
+        />
+        <ComboboxInput icon={icon} placeholder={placeholder} />
+        <FieldError />
+        <Popover
```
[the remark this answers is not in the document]
jabrks: Provided by the Combobox component through context

cdedreuille (approved):
Thanks a lot @jabrks! Looks great!

packages/ui/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,98 @@
...
+ * @public
+ */
+export const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
+  (props, ref) => {
+    const { contains } = useFilter({ sensitivity: 'base' });
+    const { ownProps, restProps, dataAttributes } = useDefinition(
+      ComboboxDefinition,
+      props,
+    );
+    const { ownProps: popoverOwnProps } = useDefinition(PopoverDefinition, {});
```
Naycon: Hmm, I don't think we want the definitions to be spread into other components. They're meant as an internal implementation detail.

packages/ui/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,98 @@
...
+        <FieldLabel
+          label={label}
+          secondaryLabel={secondaryLabelText}
+          description={description}
+          descriptionSlot="description"
+        />
+        <ComboboxInput icon={icon} placeholder={placeholder} />
+        <FieldError />
+        <Popover
+          className={clsx(popoverOwnProps.classes.root, classes.popover)}
```
Naycon: I don't think you should need this, the Popover component should do that merging by it's own. Or am I missing something?

Naycon (changes requested):
Sorry but can you take a look at the Popover definition import? I think we may not want to use the definitions across components like that.

packages/ui/src/components/Combobox/ComboboxInput.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <Group className={classes.root} {...dataAttributes}>
+      {icon ? (
+        <div className={classes.icon} aria-hidden="true">
+          {icon}
+        </div>
+      ) : null}
+      <Input className={classes.input} placeholder={placeholder} />
+      <Button className={classes.chevron}>
```
Naycon, line 38: Should this button have a label? Or does RA take care of labelling it automatically?

packages/ui/src/components/Combobox/ComboboxListBox.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  return (
+    <ListBoxItem
+      id={option.value}
+      textValue={option.label}
+      className={classes.root}
+      isDisabled={option.disabled}
+    >
+      <div className={classes.indicator}>
+        <RiCheckLine />
```
Naycon: Should this be hidden from screen readers? I assume RA takes care of announcing properly the state of the listbox item, perhaps that also includes automatically announcing what's in the label slot as the content of the item (and therefore hiding this)?

packages/ui/src/components/Combobox/definition.ts
```diff
@@ -0,0 +1,114 @@
...
+    secondaryLabel: {},
+    description: {},
+    isRequired: {},
+    className: {},
+  },
+});
+
+/**
+ * Component definition for ComboboxInput
+ * @internal
```
Naycon: I believe all the definitions that include class names should be marked as public, to signal the css class names are part of the public API and allow us to detect breaking changes there.

jabrks:
@Naycon This is copied from the Select component. If it's not intended usage then could we fix both together in a follow up?

packages/ui/src/components/Combobox/ComboboxInput.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { ReactNode } from 'react';
+import { Button, Group, Input } from 'react-aria-components';
+import { RiArrowDownSLine } from '@remixicon/react';
+import { useDefinition } from '../../hooks/useDefinition';
+import { ComboboxInputDefinition } from './definition';
+
+interface ComboboxInputProps {
+  icon?: ReactNode;
+  placeholder?: string;
+}
```
Naycon: This should be moved to types, right? You already have `ComboboxOwnInputProps` there, which overlaps with this. The idea behind "own" props is to signal which props are being consumed by the component itself (vs inherited from elsewhere and forwarded). If you only have own props, it's fine to only have that type (and drop the "own" part if you like). 

There may be another instance or so in the other files that need the same treatment.

packages/ui/src/components/Combobox/ComboboxInput.tsx
jabrks: From what I can see, React Aria handles it by passing both a `aria-label` and `aria-labelledby` itself

awanlin (approved):

Naycon (approved):
Looks great, thanks for adding this in!

saeedseyfi:
Regarding `allowsCustomValue` prop: shouldn't this be the default behaviour (no optionality)? I mean with `allowsCustomValue=false` we are getting a `Select` component.

jabrks:
@saeedseyfi I would say not. We've stayed aligned with the React Aria API for the most part so this is consistent with that approach

PR #34124, opened 2026-05-05
feat(ui): add --bui-bg-inherit CSS variable

Naycon:

Adds a public `--bui-bg-inherit` CSS variable that resolves to the
background color of the nearest enclosing `data-bg` ancestor (Box,
Flex, Grid, Card, Accordion), falling back to `--bui-bg-app`. This
lets consumers match their surrounding surface from CSS — useful for
sticky headers, fixed overlays, and scroll-shadow gradients that need
to blend with their container without hardcoding a specific level.

The variable is declared alongside the `background-color` painting
rules in a new `@layer base` block in `core.css`, so the two cannot
drift. The duplicated per-component `[data-bg]` painting rules in
Box, Flex, Grid, Accordion, and Card are removed in favor of the
centralized rules. Card's internal `--bui-card-bg` indirection is
replaced with `--bui-bg-inherit` for its scroll-shadow gradients.

One behavioral expansion: any element with a `data-bg` attribute is
now painted (not only BUI layout components). Accordion retains its
current behavior of not painting intent surfaces via an explicit
`@layer components` override.


cdedreuille (approved):
Beautiful. Much more elegant solution.

PR #34127, opened 2026-05-05
feat(ui): add searchDebounceMs/filterDebounceMs to useTable complete mode

Naycon:

Added opt-in `searchDebounceMs` and `filterDebounceMs` options to `useTable` in `complete` mode so consumers can defer the client-side filter/search/sort pipeline on large datasets without reimplementing input-layer debouncing themselves.

Both options default to `0`, which is a true zero-cost bypass — no timer is armed and no extra render occurs, so existing consumers see no behavior change. The controlled `search` / `onSearchChange` and `filter` / `onFilterChange` callbacks continue to fire on every change; only the internal pipeline (and the offset-reset effect) keys off the debounced surrogates. Sort remains immediate. `offset` and `cursor` modes are untouched.

The offset-reset effect was refactored to key off the debounced query rather than the live query — at default this is behavior-equivalent (referentially equal), and at `delayMs > 0` it prevents a flash of unfiltered page-1 data while typing settles.

Also fixes an inaccurate claim in the docs (`page.mdx`) that complete-mode search was "debounced internally" — it was not.

Verified visually via the new `SearchWithDebounce` dev Storybook story (~5,000 amplified rows, immediate vs. 200ms side-by-side).

- [x] All your commits have a \`Signed-off-by\` line in the message. ([more info](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md#developer-certificate-of-origin))

### DIFF
--- a/.changeset/usetable-complete-debounce.md
+++ b/.changeset/usetable-complete-debounce.md
@@ -1,0 +1,7 @@
+---
+'@backstage/ui': patch
+---
+
+Added `searchDebounceMs` and `filterDebounceMs` options to `useTable` in `complete` mode. Both default to `0` (no debounce, no observable change for existing consumers); set them to defer the client-side filter/search/sort pipeline on large datasets without reimplementing input-layer debouncing. The controlled `search` / `onSearchChange` and `filter` / `onFilterChange` callbacks continue to fire on every change.
+
+**Affected components:** Table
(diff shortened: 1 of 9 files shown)


cdedreuille (approved):
Great improvement 🙏

awanlin (approved):

PR #34142, opened 2026-05-06
feat(auth-backend): validate catalog user existence on refresh token usage

mtlewis:

- Adds a new `auth.experimentalRefreshToken.validateCatalogUserExistence` config option (default `false`)
- When enabled, refresh token usage checks whether the user's catalog entity still exists before issuing a new access token
- If the user has been removed from the catalog, the refresh is rejected and the session is revoked
- Transient catalog errors (e.g. catalog temporarily unavailable) reject the refresh but preserve the session for retry

### DIFF
--- a/plugins/auth-backend/config.d.ts
+++ b/plugins/auth-backend/config.d.ts
@@ -133,6 +133,14 @@ export interface Config {
        * @visibility backend
        */
       maxTokensPerUser?: number;
+      /**
+       * Whether to validate that the user's catalog entity exists when
+       * refreshing a token. When enabled, tokens for users removed from
+       * the catalog will be rejected and revoked.
+       * @default false
+       * @visibility backend
+       */
+      validateCatalogUserExistence?: boolean;
     };
 
     /**
--- a/plugins/auth-backend/config.d.ts
+++ b/plugins/auth-backend/config.d.ts
@@ -134,13 +134,16 @@ export interface Config {
        */
       maxTokensPerUser?: number;
       /**
-       * Whether to validate that the user's catalog entity exists when
-       * refreshing a token. When enabled, tokens for users removed from
-       * the catalog will be rejected and revoked.
+       * Disables the check that verifies the user's catalog entity still
+       * exists when refreshing a token. This is an escape hatch for
+       * Backstage instances that allow sign-in without a corresponding
+       * catalog user entity. Without the check, refresh tokens for
+       * removed or offboarded users remain valid until they naturally
+       * expire.
        * @default false
        * @visibility backend
        */
-      validateCatalogUserExistence?: boolean;
+      dangerouslyDisableCatalogPresenceCheck?: boolean;
     };
 
     /**
--- a/plugins/auth-backend/src/service/OfflineAccessService.ts
+++ b/plugins/auth-backend/src/service/OfflineAccessService.ts
@@ -15,6 +15,7 @@
  */
 
 import {
+  AuthService,
   DatabaseService,
   LifecycleService,
   LoggerService,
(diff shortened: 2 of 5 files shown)


plugins/auth-backend/src/service/OfflineAccessService.ts
```diff
@@ -212,6 +236,33 @@ export class OfflineAccessService {
       throw new AuthenticationError('Invalid refresh token');
     }
 
+    if (this.#validateCatalogUserExistence) {
```
Rugvip: Hmm, I'd like this to be the default honestly, but at the same time we can't assume that the session is actually derived from a user in the catalog in the first place. What would it take to wire up the original sign-in resolved to be called here again at this point? I'm guessing that's not gonna be possible because it may rely on temporary credentials that are no longer valid at refresh.

Perhaps the catalog sign-in helpers could forward enough data to indicate the the resolution that should be re-run or smth? 🤔 or we just bite the bullet and essentially have a secondary sign-in resolved for refreshes.

If you think about this current implementation it doesn't quite cover what we might want either, because the ownership references aren't updated.

All that said I'm kinda happy to ship this as is and as an option though, because any of the other options might be too tricky to implement in a generic way in practice without extra burden on adopter.s

plugins/auth-backend/src/service/OfflineAccessService.ts
```diff
@@ -212,6 +236,33 @@ export class OfflineAccessService {
...
+          );
+          await this.#offlineSessionDb.deleteSession(sessionId);
+          throw new AuthenticationError(
+            'User entity no longer exists in the catalog',
+          );
+        }
+      } catch (error) {
+        if (error instanceof AuthenticationError) {
+          throw error;
+        }
```
Rugvip:
```suggestion
        if (error.name === 'AuthenticationError') {
          throw error;
        }
```

Safest way to do error type checks, works for cross-realm errors

plugins/auth-backend/config.d.ts
```diff
@@ -133,6 +133,14 @@ export interface Config {
...
        */
       maxTokensPerUser?: number;
+      /**
+       * Whether to validate that the user's catalog entity exists when
+       * refreshing a token. When enabled, tokens for users removed from
+       * the catalog will be rejected and revoked.
+       * @default false
+       * @visibility backend
+       */
+      validateCatalogUserExistence?: boolean;
```
Rugvip: Naming feels slightly awkward, possibly "presence" instead?

Rugvip (changes requested):
Good catch 👍

plugins/auth-backend/config.d.ts
mtlewis: :+1: discussed offline and agreed on flipping this on by default and renaming to `dangerouslyDisableCatalogPresenceCheck`. [ed4ffaa](https://github.com/backstage/backstage/pull/34142/commits/ed4ffaa3b39b615b725fef06606e1d5709319313)

plugins/auth-backend/src/service/OfflineAccessService.ts
mtlewis: :+1: [ed4ffaa](https://github.com/backstage/backstage/pull/34142/commits/ed4ffaa3b39b615b725fef06606e1d5709319313)

plugins/auth-backend/src/service/OfflineAccessService.ts
mtlewis: On by default as of [ed4ffaa](https://github.com/backstage/backstage/pull/34142/commits/ed4ffaa3b39b615b725fef06606e1d5709319313). I agree we should go further long-term here. Next short-term step in my view is working towards excluding inactive users for all org data providers in OSS to reduce the gap here, but longer term I do think it'd be great to go further.

Rugvip (approved):
Thank you! Let's :shipit: 👍

PR #34147, opened 2026-05-07
feat(scaffolder): config-driven template groups and swappable TemplateCard

benjdlambert:

In the new frontend system there was no way to define template groups on the scaffolder list page or replace the default `TemplateCard` — both were JSX props on the legacy `<ScaffolderPage>` and have no equivalent today. This PR adds both, and demos them in the example app.

### What changed

- **`@backstage/plugin-scaffolder`**: `subPage:scaffolder/templates` extension gains a `groups` config field. Each group has a `title` and a `filter` predicate (using `@backstage/filter-predicates`, same machinery as `EntityCardBlueprint`). Templates not matched by any group fall into an automatically appended "Other" group; with no groups configured the page renders a single "All Templates" group as before.
- **`@backstage/plugin-scaffolder-react`**: `TemplateCard` is now a swappable component. Apps can replace it via `SwappableComponentBlueprint` against `TemplateCard.ref`. The default implementation is unchanged.
- **`packages/app`**: Demo `BuiTemplateCard` (Backstage UI) registered through a new `appModuleScaffolder` frontend module, plus an example `groups` block in `app-config.yaml` showing two groups + auto-appended "Other".
- **Docs**: New section in `docs/features/software-templates/configuration.md` covering both customisations for the new frontend system.

### Why this approach

- Groups go on the sub-page extension because that's where the list actually renders. The parent `page:scaffolder` is just a tab shell.
- `TemplateCard` swap uses `createSwappableComponent` rather than a blueprint input slot — the swap concern is component-shape, not page-level extension wiring.
- Legacy router path is untouched: `TemplateGroups` still accepts a `CardComponent` prop, so old apps using `<ScaffolderPage components={{ TemplateCardComponent }} />` keep working. The wrapped `TemplateCard` only resolves the swap when no prop is passed.

### Out of scope

- `templateFilter`, the other `components.*` slots (`ReviewStepComponent`, `TaskPageComponent`, etc.), `headerOptions`, `contextMenu`. Can be added later when there is demand.
- Imperative/code-defined groups via blueprint params. Adding them later is non-breaking.

- [x] All your commits have a `Signed-off-by` line in the message


docs/features/software-templates/configuration.md
```diff
@@ -149,3 +153,73 @@ You can have several use cases for that:
...
+## Customizing the templates page in the new frontend system
+
+In the new frontend system the templates page is built from extensions, so
+customisations are configured rather than passed as JSX props.
+
+### Defining template groups in `app-config.yaml`
+
+The `sub-page:scaffolder/templates` extension accepts a `groups` config field.
+Each group has a `title` and a `filter` predicate (using
+[`@backstage/filter-predicates`](https://www.npmjs.com/package/@backstage/filter-predicates)).
```
freben: alternative link: https://backstage.io/docs/features/software-catalog/catalog-customization#entity-predicate-queries

docs/features/software-templates/configuration.md
```diff
@@ -149,3 +153,73 @@ You can have several use cases for that:
...
+                spec.type: service
+            - title: Documentation
+              filter:
+                spec.type: documentation
+```
+
+Predicate values are matched case-insensitively. The matchers `$exists`,
+`$in`, `$contains`, `$hasPrefix` and the logical operators `$all`, `$any`, `$not`
+are also supported — see the
+[filter predicates reference](https://github.com/backstage/backstage/blob/master/packages/filter-predicates/README.md)
```
freben: same here, https://backstage.io/docs/features/software-catalog/catalog-customization#entity-predicate-queries may be more canonical

docs/features/software-templates/configuration.md
```diff
@@ -149,3 +153,73 @@ You can have several use cases for that:
...
+
+export const appModuleScaffolder = createFrontendModule({
+  pluginId: 'app',
+  extensions: [
+    SwappableComponentBlueprint.make({
+      name: 'scaffolder-template-card',
+      params: defineParams =>
+        defineParams({
+          component: TemplateCard,
+          loader: async () => MyTemplateCard,
```
freben: best practise is to make it a lazy import, not sure if worth worrying about here

packages/app/src/modules/appModuleScaffolder.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const appModuleScaffolder = createFrontendModule({
+  pluginId: 'app',
+  extensions: [
+    SwappableComponentBlueprint.make({
+      name: 'scaffolder-template-card',
+      params: defineParams =>
+        defineParams({
+          component: TemplateCard,
+          loader: async () => BuiTemplateCard,
```
freben: maybe we make it lazy here though :)

packages/app/src/modules/BuiTemplateCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <Button size="small" variant="secondary" onPress={handleRun}>
+            Run
+          </Button>
+          {owner && (
+            <Flex gap="0" direction="column" align="end">
+              <Text variant="body-x-small" color="secondary">
+                Created by
+              </Text>
+              <Text variant="body-x-small" color="primary">
+                <EntityRefLink entityRef={owner} hideIcon disableTooltip />
```
freben: tooltip is probably ok to have

plugins/scaffolder-react/src/next/components/TemplateCard/TemplateCardImpl.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * @alpha
+ */
+export interface TemplateCardProps {
+  template: TemplateEntityV1beta3;
+  additionalLinks?: {
+    icon: IconComponent;
+    text: string;
+    url: string;
+  }[];
+  onSelected?: (template: TemplateEntityV1beta3) => void;
```
freben, line 51: the interfaces here are so precise - Oof I wish we had a better strategy for how to evolve these. Leave as is, just worrying

plugins/scaffolder-react/src/next/components/TemplateCard/TemplateCardImpl.tsx
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/**
+ * Default implementation of the `TemplateCard`. The exported `TemplateCard`
+ * is a swappable wrapper around this component.
+ *
+ * @alpha
+ */
+export const TemplateCardImpl = (props: TemplateCardProps) => {
```
freben: Maybe don't visibly export an `*Impl` name - `Default*` instead perhaps?

benjdlambert: It's actually not exported, can drop the `@alpha`

plugins/scaffolder-react/src/next/components/TemplateCard/TemplateCardImpl.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const hasTags = !!template.metadata.tags?.length;
+  const hasLinks =
+    !!additionalLinks?.length || !!template.metadata.links?.length;
+  const displayDefaultDivider = !hasTags && !hasLinks;
+
+  const { allowed: canCreateTask } = usePermission({
+    permission: taskCreatePermission,
+  });
+  const handleChoose = useCallback(() => {
+    analytics.captureEvent('click', 'Template has been opened');
```
acierto, line 88: Is here any magic which sends the information which exactly template was opened?

benjdlambert: There's an `AnalyticsContext` which wraps the cards already with `entityRef: stringifyEntityRef(template)`

freben (approved):

PR #34149, opened 2026-05-07
feat(integration-aws-node): add per-account webIdentityTokenFile config

hudsonb:
Adds an optional `webIdentityTokenFile` field to
`AwsIntegrationAccountConfig` and `AwsIntegrationDefaultAccountConfig`. When set on a per-account config along with a `roleName` and no static credentials, `DefaultAwsCredentialsManager` now retrieves credentials by calling `AssumeRoleWithWebIdentity` directly using the file's contents as the web identity token (via `fromTokenFile`). The token file is re-read on each refresh, so an external process can rotate it in place. Same `fromTokenFile` the AWS SDK already uses for IRSA's env-var path and the AWS CLI's web_identity_token_file = profile setting but exposed at per-account granularity through `aws.accounts[]` config so backends with multiple AWS accounts can use one token file across many role ARNs.

This unlocks multi-account `AssumeRoleWithWebIdentity` for backends running outside AWS (GKE, Cloud Run, Vault sidecars, etc.) without requiring every plugin to construct a custom `AwsCredentialsManager`. Existing call sites and configurations are unaffected; the new path is opt-in via the new optional field.

- [x] All your commits have a `Signed-off-by` line in the message ([more info](https://github.com/backstage/backstage/blob/master/CONTRIBUTING.md#developer-certificate-of-origin))

### DIFF
--- a/.changeset/aws-wif-token-file.md
+++ b/.changeset/aws-wif-token-file.md
@@ -1,0 +1,14 @@
+---
+'@backstage/integration-aws-node': minor
+---
+
+Added `webIdentityTokenFile` to `AwsIntegrationAccountConfig` and
+`AwsIntegrationDefaultAccountConfig`. When set along with a `roleName`,
+`DefaultAwsCredentialsManager` retrieves credentials by calling
+`AssumeRoleWithWebIdentity` (via `fromTokenFile`) using the file's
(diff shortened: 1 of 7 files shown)


freben (approved):
Nice!

Issue #34154, opened 2026-05-07
catalog: entity-facets performance issues with multiple filters


koalaty-code:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
catalog facets database query performance
```

### 🗃️ Project Area

**Catalog**

### 🔗 External Integration

**N/A**

### 📝 Description & Context

The `/api/catalog/entity-facets` endpoint with complex filters generates a database query that performs a full table scan on the `search` table, even for small result sets. This causes severe database CPU spikes when multiple complex filter conditions are applied concurrently.

Observed behavior: When the facets endpoint is called with multiple filter conditions, database CPU spikes to 100% with a single request. The generated SQL performs a full sequential scan on the `search` table (millions of rows) even though the filtered entity set is small.

This appears related to how the query builder applies filters using `WHERE IN (subquery)` pattern instead of `INNER JOIN`.

### 👍 Expected Behavior

The facets endpoint should return results efficiently, even with multiple complex filters:
- Filter entities via `final_entities` (small set)
- Join to `search` table to get facet counts
- Minimal table scans on large tables
- Query completes in <100ms for typical result sets

### 🥾 Reproduction steps

1. Set up Backstage with a production-scale catalog (1M+ search table rows)
2. Call the facets endpoint with multiple filter conditions:

```
GET /api/catalog/entity-facets?facet=spec.lifecycle&filter=kind%3Dcomponent%2Cmetadata.tags%3Dfoo%2Cspec.lifecycle%3Dproduction&filter=kind%3Dresource%2Cspec.type%3Dbar
```

3. Monitor database CPU usage while making a few concurrent requests
4. Expected: CPU usage remains low; Actual: CPU spikes to 100%

#### Generated SQL (problematic pattern)

```sql
SELECT search.key, search.original_value, COUNT(DISTINCT search.entity_id)
FROM search
WHERE search.key IN (?)
  AND search.original_value IS NOT NULL
  AND search.entity_id IN (
    SELECT final_entities.entity_id FROM final_entities
    WHERE final_entities.final_entity IS NOT NULL
    AND (complex filter conditions with multiple EXISTS clauses)
  )
GROUP BY search.key, search.original_value
```

The `WHERE IN (subquery)` causes PostgreSQL to:
1. Perform full table scan on `search` (millions of rows scanned)
[...]

PR #34158, opened 2026-05-08
fix(catalog-backend): use INNER JOIN for filtered entity facets

freben:

Replace the `WHERE search.entity_id IN (...)` form with an INNER JOIN against the filtered `final_entities` subquery in `DefaultEntitiesCatalog#facets`. The results are unchanged; the change just gives the query planner more freedom in join shape and ordering, which on large catalogs leads to dramatically cheaper plans.

Originally proposed in #34154.

- [x] Tests for new functionality and regression tests for bug fixes _(existing facets tests pass on MySQL/PG18/PG14/SQLite)_

### DIFF
--- a/.changeset/facets-inner-join.md
+++ b/.changeset/facets-inner-join.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Improved the performance of the entity facets endpoint when filters are applied. The filtered entity set is now combined with the search table through an inner join rather than a `WHERE entity_id IN (subquery)`. Results are unchanged; on large catalogs the query planner is able to choose dramatically cheaper plans, with measured improvements ranging from roughly 1.2× on already-fast cases to 7× or more on high-cardinality facets.
--- a/plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
+++ b/plugins/catalog-backend/src/service/DefaultEntitiesCatalog.ts
@@ -707,7 +707,17 @@ export class DefaultEntitiesCatalog implements EntitiesCatalog {
         knex: this.database,
       });
 
-      query.whereIn('search.entity_id', entityIdSubquery);
+      // Use INNER JOIN rather than `WHERE search.entity_id IN (...)`. The
+      // results are the same but the JOIN form gives the planner more
+      // freedom in join shape and ordering. On PostgreSQL with large
+      // search tables, the IN form tends to materialize the full filtered
+      // entity set up front and spill to temp; the JOIN form lets the
+      // planner pick a much cheaper plan based on actual selectivities.
+      query.innerJoin(
+        entityIdSubquery.as('filtered_entities'),
+        'search.entity_id',
+        'filtered_entities.entity_id',
+      );
     }
 
     const rows = await query;


freben:
## Adversarial performance testing

Tested on a production-like staging database (~13.8M rows in \`search\`) using \`EXPLAIN (ANALYZE, BUFFERS)\` on representative facets queries. For each scenario the same query was run as a \`WHERE entity_id IN (...)\` (today) and as an \`INNER JOIN\` (this PR). Results are byte-identical between the two.

### Summary: no regressions found

| # | Scenario | A: WHERE IN | B: INNER JOIN | Winner |
|---|---|---:|---:|---|
| 1 | Narrow AND filter (\`kind=component\` AND \`spec.lifecycle=production\`), facet=\`spec.lifecycle\` | 2,397 ms | **1,943 ms** | JOIN (1.2×) |
| 2 | Single broad filter (\`kind=component\`), facet=\`spec.lifecycle\` | 1,907 ms | **1,360 ms** | JOIN (1.4×) |
| 3 | Very narrow filter (\`kind=template\`), facet=\`spec.lifecycle\` | 14 ms | **1 ms** | JOIN (14×) |
| 4 | OR'd filter groups (issue #34154's example) | >60 s | >60 s | tied — catastrophic both ways |
| 5 | High-cardinality facet (\`metadata.name\`), filter \`kind=component\` | 13,961 ms | **1,924 ms** | JOIN (**7.3×**) |
| 6 | Multi-facet \`{spec.lifecycle, spec.type}\`, narrow filter | 3,628 ms | **2,289 ms** | JOIN (1.6×) |
| 7 | Very narrow result via rare tag (\`kind=component\` AND \`metadata.tags=hackweek\`) | 100 ms | 105 ms | tied (within noise) |

The very first cold-cache run of scenario 1 was even more dramatic: 14,688 ms → 3,027 ms (~5×). Subsequent warm-cache runs converged to the ~1.5× shown above. Production traffic across multiple pods and contention for buffers is closer to the cold case than to warm-cache micro-benchmarks, so the real-world impact should be larger than what the warm numbers suggest.

### Why the JOIN is faster

The \`WHERE IN (subquery)\` form forces the planner to materialize the full filtered entity set up front (HashAggregate) and probe \`search\` from that side. On the staging plan that step spilled ~15 MB to temp files. The INNER JOIN form lets the planner reorder, drive from either side based on actual selectivities, and use Memoize for the \`final_entities\` lookup with very high cache hit rates (~70-90% in the observed plans). Buffer hits on the dominant scenarios drop ~45-50%.

### Caveat: scenario 4 (OR'd filter groups)

Both variants time out at 60s on the issue's exact example shape. This is a separate planner pathology around OR-of-conjunctions defeating semi-join optimizations and is **not** addressed by either this PR or by the indices in #34015. It needs its own ticket — likely a query rewrite (e.g., \`UNION ALL\` of the two filter branches inside the derived table). I did not investigate further as part of this work.

### How this composes with #34015

The big remaining bottleneck after this change is a Parallel Bitmap Heap Scan on \`search\` filtered by \`(key=…, original_value IS NOT NULL)\`. On the staging plan that step alone accounts for ~1.2s of the remaining ~3s, fetching ~315k heap rows with heavy lossy bitmaps. That is exactly what the \`search_facets_covering_idx (key, original_value, entity_id) WHERE original_value IS NOT NULL\` index from #34015 attacks — it converts that step into an index-only scan, no heap pages.

So:

- **This PR alone:** brings the worst cases (high-cardinality facets, big result sets) from tens of seconds down to ~1-3s, with no migration risk and a one-line code change.
- **This PR + #34015:** lands in the hundreds-of-ms regime, since the heap-scan floor is removed.
- **Neither addresses scenario 4** — that's a separate ticket.

The index story doesn't materially change in the presence of this PR; if anything, the INNER JOIN gives the planner more room to actually pick plans that take advantage of the new \`(entity_id, key, value)\` index in #34015 (drive from the small filtered set, probe search per entity_id), where the IN form was more constrained.

Recommended order: land this first (cheap, no migration), then #34015.

benjdlambert (approved):

PR #34165, opened 2026-05-08
feat(catalog-backend-module-msgraph): filter out disabled users by default

mtlewis:

The Microsoft Graph (EntraID) catalog provider now always applies an `accountEnabled eq true` base filter when fetching users. This ensures that any disabled/inactive accounts never make it into the Backstage catalog. from polluting the Backstage catalog.

- Any custom `user.filter` is combined with the base filter using `and`, so adopters can layer additional filters on top without needing to manually include `accountEnabled eq true`.
- Removes the mutual exclusivity check between `userFilter` and `userGroupMemberFilter` — these two seem to be compatible from looking at the code. I'll test manually before opening this for review.

### DIFF
--- a/.changeset/msgraph-filter-disabled-users.md
+++ b/.changeset/msgraph-filter-disabled-users.md
@@ -3,4 +3,4 @@
 '@backstage/plugin-catalog-backend-module-msgraph-incremental': minor
 ---
 
-Disabled user accounts are now always filtered out. The provider applies an `accountEnabled eq true` filter automatically and combines it with any custom `user.filter` you provide.
+**BREAKING**: Disabled user accounts are now filtered out by default. The provider automatically applies an `accountEnabled eq true` filter, combining it with any custom `user.filter` you provide. If you previously included `accountEnabled eq true` in your user filter, it is safe to remove it, but leaving it in will not cause any issues.
--- a/.changeset/msgraph-filter-disabled-users.md
+++ b/.changeset/msgraph-filter-disabled-users.md
@@ -1,0 +1,6 @@
+---
+'@backstage/plugin-catalog-backend-module-msgraph': minor
+'@backstage/plugin-catalog-backend-module-msgraph-incremental': minor
+---
+
+Disabled user accounts are now always filtered out. The provider applies an `accountEnabled eq true` filter automatically and combines it with any custom `user.filter` you provide.
--- a/plugins/catalog-backend-module-msgraph/README.md
+++ b/plugins/catalog-backend-module-msgraph/README.md
@@ -52,8 +52,8 @@ catalog:
           # Optional filter for user, see Microsoft Graph API for the syntax
           # See https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0#properties
           # and for the syntax https://docs.microsoft.com/en-us/graph/query-parameters#filter-parameter
-          # This and userGroupMemberFilter are mutually exclusive, only one can be specified
-          filter: accountEnabled eq true and userType eq 'member'
+          # This filter is combined with the base `accountEnabled eq true` filter.
+          filter: userType eq 'member'
           # Set to false to not load user photos.
           loadPhotos: true
           # See  https://docs.microsoft.com/en-us/graph/api/resources/schemaextension?view=graph-rest-1.0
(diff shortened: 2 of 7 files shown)


.changeset/msgraph-filter-disabled-users.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-catalog-backend-module-msgraph': minor
+'@backstage/plugin-catalog-backend-module-msgraph-incremental': minor
+---
+
+Disabled user accounts are now always filtered out. The provider applies an `accountEnabled eq true` filter automatically and combines it with any custom `user.filter` you provide.
```
mtlewis: For discussion: should we communicate this as a breaking change?

plugins/catalog-backend-module-msgraph/README.md
```diff
@@ -52,8 +52,8 @@ catalog:
           # Optional filter for user, see Microsoft Graph API for the syntax
           # See https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0#properties
           # and for the syntax https://docs.microsoft.com/en-us/graph/query-parameters#filter-parameter
-          # This and userGroupMemberFilter are mutually exclusive, only one can be specified
-          filter: accountEnabled eq true and userType eq 'member'
+          # This filter is combined with the base `accountEnabled eq true` filter.
+          filter: userType eq 'member'
```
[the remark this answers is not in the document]
mtlewis: :+1:

plugins/catalog-backend-module-msgraph/src/microsoftGraph/config.ts
```diff
@@ -393,3 +373,11 @@ export function readProviderConfig(
     schedule,
   };
 }
+
+function buildUserFilter(rawFilter: string | undefined): string {
+  const base = 'accountEnabled eq true';
+  if (rawFilter) {
+    return `${base} and (${rawFilter})`;
```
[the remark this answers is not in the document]
mtlewis: I don't think this is really worth it - there's lots of ways to supply an invalid OData expression, and I don't think it particularly makes sense to pull on the thread. It seems safe to let it fail in this situation.

.changeset/msgraph-filter-disabled-users.md
benjdlambert: Yeah I think so :pray:

benjdlambert: You can probably call out that consumers can remove the `accountEnabled eq true` everywhere and let the system do it.

Should we also have an opt out here similar to other providers?

Rugvip (approved):
Yep, makes sense to enforce this for safety 👍

.changeset/msgraph-filter-disabled-users.md
Rugvip: Makes sense to mark as a breaking change, if not just for the visibility

mtlewis: > Should we also have an opt out here similar to other providers? 

I'd propose not. The only reason I kept the opt-out in https://github.com/backstage/backstage/pull/34170 is the extra rate-limit consumption. If our current thinking is that it's definitively incorrect for org data providers to return suspended users, it seems to me we should enforce that behavior without the ability to override back to the wrong thing.

mtlewis: > Yeah I think so 🙏

:+1:. It's already marked as a minor bump, I'll update the wording to match.

> You can probably call out that consumers can remove the accountEnabled eq true everywhere and let the system do it.

Will do!

mtlewis: 00173de655ff2

plugins/catalog-backend-module-msgraph/README.md
```diff
@@ -102,8 +100,6 @@ catalog:
           initialDelay: { seconds: 15},
 ```
 
-`user.filter` and `userGroupMember.filter` are mutually exclusive, only one can be provided. If both are provided, an error will be thrown.
-
 By default, all users are loaded. If you want to filter users based on their attributes, use `user.filter`. `userGroupMember.filter` can be used if you want to load users based on their group membership.
```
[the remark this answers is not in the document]
mtlewis: :+1: 28267e6f4c68a

plugins/catalog-backend-module-msgraph/src/microsoftGraph/config.ts
```diff
@@ -393,3 +373,11 @@ export function readProviderConfig(
...
+  }
+  return base;
```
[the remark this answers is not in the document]
mtlewis: :+1: 28267e6f4c68a

benjdlambert (approved):

PR #34170, opened 2026-05-08
feat(catalog-backend-module-github): auto-exclude suspended users on GHE

mtlewis:

**Suspended users are now automatically excluded** on GitHub Enterprise instances, and `site_admin` scope is no longer required to check for user suspension. To encourage adoption, the previous `excludeSuspendedUsers` option (default false) has been replaced with `dangerouslySkipSuspendedUserCheck` (opposite behavior, also default false).

Since this new behavior makes much heavier use of rate limits, this PR also introduces conditional requests for the REST API consumption in the providers. Response data, headers, etags and last-modified dates are stored in the cacheService cache with a long TTL and included with future requests, meaning that only the first load of user data will typically result in a big rate limit hit. The cache TTL can be really long, because we never return the data in the cache without getting a 304 back from GitHub to indicate it's up to date.

Without this caching system in place, this change would add 2 REST API rate-limit points _per user_ to every provider run when suspended user checking is enabled, so it seems unavoidable that we add it.

### How it works

- Uses the REST API (`GET /users/{username}` and `GET /orgs/{org}/memberships/{username}`) to detect both account-level suspension and suspension of organization membership.
- Automatically detects whether the target is a GitHub Enterprise instance via `GET /versions` and skips the check on github.com where suspension isn't supported
- Uses Octokit's throttling and retry plugins to handle rate limiting
- Filters suspended users during GraphQL pagination via an async filter in `queryWithPaging`

### Breaking change

`excludeSuspendedUsers` config and option is removed. The suspended user check now runs by default. To opt out, set `dangerouslySkipSuspendedUserCheck: true` in config or the provider options.

### DIFF
--- a/.changeset/suspended-users-rest-api.md
+++ b/.changeset/suspended-users-rest-api.md
@@ -1,0 +1,6 @@
+---
+'@backstage/plugin-catalog-backend-module-github': minor
+'@backstage/plugin-catalog-backend-module-github-org': minor
+---
+
+**BREAKING**: Suspended users are now excluded during ingestion by default. User suspended state is now checked using the REST API, which avoids the need for the `site_admin` scope. Both account-level suspension and org-membership suspension are detected. This replaces the previous `excludeSuspendedUsers` option (which defaulted to off) with `dangerouslySkipSuspendedUserCheck` (which defaults to off, meaning the check runs by default). Set `dangerouslySkipSuspendedUserCheck: true` to disable the check if needed.
--- a/plugins/catalog-backend-module-github/src/lib/github.ts
+++ b/plugins/catalog-backend-module-github/src/lib/github.ts
@@ -18,7 +18,6 @@ import { Entity } from '@backstage/catalog-model';
 import {
   GithubCredentials,
   GithubCredentialsProvider,
-  GithubCredentialType,
 } from '@backstage/integration';
 import { graphql } from '@octokit/graphql';
 import { OctokitResponse, RequestParameters } from '@octokit/types';
@@ -125,7 +124,6 @@ export type GithubUser = {
   email?: string;
   name?: string;
   organizationVerifiedDomainEmails?: string[];
-  suspendedAt?: string;
 };
 
 /**
@@ -185,26 +183,19 @@ export type Connection<T> = {
  *
  * Note that the users will not have their memberships filled in.
  *
- * @param client - An octokit graphql client
+ * @param octokit - An octokit client
  * @param org - The slug of the org to read
- * @param tokenType - The type of GitHub credential
  * @param userTransformer - Optional transformer for user entities
  * @param pageSizes - Optional page sizes configuration
- * @param excludeSuspendedUsers - Optional flag to exclude suspended users (only for GitHub Enterprise instances)
- * @param restClient - Optional Octokit REST client; when provided alongside excludeSuspendedUsers, the REST API is used instead of the GraphQL suspendedAt field
+ * @param filterSuspendedUsers - When true, suspended users are excluded using the REST API. Callers should resolve this by checking isGitHubEnterprise() once at the start of a provider run.
  */
 export async function getOrganizationUsers(
-  client: typeof graphql,
+  octokit: Octokit,
   org: string,
-  tokenType: GithubCredentialType,
   userTransformer: UserTransformer = defaultUserTransformer,
   pageSizes: GithubPageSizes = DEFAULT_PAGE_SIZES,
-  excludeSuspendedUsers: boolean = false,
-  restClient?: Octokit,
+  filterSuspendedUsers: boolean = false,
 ): Promise<{ users: Entity[] }> {
-  const useRestForSuspension = excludeSuspendedUsers && !!restClient;
-  const suspendedAtField =
-    excludeSuspendedUsers && !useRestForSuspension ? 'suspendedAt,' : '';
   const query = `
     query users($org: String!, $email: Boolean!, $cursor: String, $organizationMembersPageSize: Int!) {
       organization(login: $org) {
@@ -217,42 +208,30 @@ export async function getOrganizationUsers(
             id,
             login,
             name,
-            ${suspendedAtField}
             organizationVerifiedDomainEmails(login: $org)
           }
         }
       }
     }`;
 
-  let restFilter:
-    | ((user: GithubUser) => Promise<boolean> | boolean)
-    | undefined;
-
-  if (useRestForSuspension) {
-    const isEnterprise = await isGitHubEnterprise(restClient);
-    if (isEnterprise) {
-      restFilter = async (user: GithubUser) =>
-        !(await isSuspended(user.login, restClient, { org }));
-    }
-  }
+  const email = ((await octokit.auth()) as GithubCredentials).type === 'token';
 
   // There is no user -> teams edge, so we leave the memberships empty for
   // now and let the team iteration handle it instead
-
   const users = await queryWithPaging({
-    client,
+    client: octokit.graphql,
     query,
     org,
     connection: r => r.organization?.membersWithRole,
     transformer: userTransformer,
     variables: {
       org,
-      email: tokenType === 'token',
+      email,
       organizationMembersPageSize: pageSizes.organizationMembers,
     },
-    filter: useRestForSuspension
-      ? restFilter
-      : u => (excludeSuspendedUsers ? !u.suspendedAt : true),
+    filter: filterSuspendedUsers
+      ? async user => !(await isSuspended(user.login, octokit, { org }))
+      : undefined,
   });
 
   return { users };
@@ -263,13 +242,13 @@ export async function getOrganizationUsers(
  *
  * Note that the teams will not have any relations apart from parent filled in.
  *
- * @param client - An octokit graphql client
+ * @param octokit - An octokit client
  * @param org - The slug of the org to read
  * @param teamTransformer - Optional transformer for team entities
  * @param pageSizes - Optional page sizes configuration
  */
 export async function getOrganizationTeams(
-  client: typeof graphql,
+  octokit: Octokit,
   org: string,
   teamTransformer: TeamTransformer = defaultOrganizationTeamTransformer,
   pageSizes: GithubPageSizes = DEFAULT_PAGE_SIZES,
@@ -340,7 +319,7 @@ export async function getOrganizationTeams(
   };
 
   const teams = await queryWithPaging({
-    client,
+    client: octokit.graphql,
     query,
     org,
     connection: r => r.organization?.teams,
@@ -356,7 +335,7 @@ export async function getOrganizationTeams(
 }
 
 export async function getOrganizationTeamsFromUsers(
-  client: typeof graphql,
+  octokit: Octokit,
   org: string,
   userLogins: string[],
   teamTransformer: TeamTransformer = defaultOrganizationTeamTransformer,
@@ -435,7 +414,7 @@ export async function getOrganizationTeamsFromUsers(
   };
 
   const teams = await queryWithPaging({
-    client,
+    client: octokit.graphql,
     query,
     org,
     connection: r => r.organization?.teams,
@@ -452,7 +431,7 @@ export async function getOrganizationTeamsFromUsers(
 }
 
 export async function getOrganizationTeamsForUser(
-  client: typeof graphql,
+  octokit: Octokit,
   org: string,
   userLogin: string,
   teamTransformer: TeamTransformer,
@@ -494,7 +473,7 @@ export async function getOrganizationTeamsForUser(
   };
 
   const teams = await queryWithPaging({
-    client,
+    client: octokit.graphql,
     query,
     org,
     connection: r => r.organization?.teams,
@@ -506,7 +485,7 @@ export async function getOrganizationTeamsForUser(
 }
 
 export async function getOrganizationsFromUser(
-  client: typeof graphql,
+  octokit: Octokit,
   user: string,
 ): Promise<{
   orgs: string[];
@@ -522,7 +501,7 @@ export async function getOrganizationsFromUser(
   }`;
 
   const orgs = await queryWithPaging({
-    client,
+    client: octokit.graphql,
     query,
     org: '',
     connection: r => r.user?.organizations,
@@ -534,7 +513,7 @@ export async function getOrganizationsFromUser(
 }
 
 export async function getOrganizationTeam(
-  client: typeof graphql,
+  octokit: Octokit,
   org: string,
   teamSlug: string,
   teamTransformer: TeamTransformer = defaultOrganizationTeamTransformer,
@@ -594,7 +573,7 @@ export async function getOrganizationTeam(
     return await teamTransformer(team, ctx);
   };
 
-  const response: QueryResponse = await client(query, {
+  const response: QueryResponse = await octokit.graphql(query, {
     org,
     teamSlug,
     membersPageSize: pageSizes.teamMembers,
@@ -605,7 +584,7 @@ export async function getOrganizationTeam(
 
   const team = await materialisedTeam(response.organization?.team, {
     query,
-    client,
+    client: octokit.graphql,
     org,
   });
 
@@ -615,7 +594,7 @@ export async function getOrganizationTeam(
 }
 
 export async function getOrganizationRepositories(
-  client: typeof graphql,
+  octokit: Octokit,
   org: string,
   catalogPath: string,
   pageSizes: GithubPageSizes = DEFAULT_PAGE_SIZES,
@@ -670,7 +649,7 @@ export async function getOrganizationRepositories(
     }`;
 
   const repositories = await queryWithPaging({
-    client,
+    client: octokit.graphql,
     query,
     org,
     connection: r => r.repositoryOwner?.repositories,
@@ -686,7 +665,7 @@ export async function getOrganizationRepositories(
 }
 
 export async function getOrganizationRepository(
-  client: typeof graphql,
+  octokit: Octokit,
   org: string,
   repoName: string,
   catalogPath: string,
@@ -733,7 +712,7 @@ export async function getOrganizationRepository(
       }
     }`;
 
-  const response: QueryResponse = await client(query, {
+  const response: QueryResponse = await octokit.graphql(query, {
     org,
     repoName,
     catalogPathRef,
@@ -836,17 +815,18 @@ export async function queryWithPaging<
       throw new Error(`Found no match for ${JSON.stringify(variables)}`);
     }
 
-    for (const node of conn.nodes) {
-      if (filter && !(await filter(node))) {
-        continue;
-      }
-      const transformedNode = await transformer(node, {
-        client,
-        query,
-        org,
-      });
-      if (transformedNode) {
-        result.push(transformedNode);
+    const transformed = await Promise.all(
+      conn.nodes.map(async node => {
+        if (filter && !(await filter(node))) {
+          return undefined;
+        }
+        return transformer(node, { client, query, org });
+      }),
+    );
+
+    for (const node of transformed) {
+      if (node) {
+        result.push(node);
       }
     }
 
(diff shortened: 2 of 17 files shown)


plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -812,18 +818,18 @@ export async function queryWithPaging<
...
+        if (filter && !(await filter(node))) {
+          return undefined;
+        }
+        return transformer(node, { client, query, org });
+      }),
+    );
+
+    for (const node of transformed) {
+      if (node) {
+        result.push(node);
```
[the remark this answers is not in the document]
mtlewis: This isn't correct, because we're using the octokit throttling plugin to control the rate at which we make requests overall.

plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts
```diff
@@ -250,26 +266,37 @@ export class GithubOrgEntityProvider implements EntityProvider {
...
       baseUrl: this.options.gitHubConfig.apiBaseUrl!,
       logger,
     });
 
+    const restClient = createRestClient({
+      token: token!,
+      baseUrl: this.options.gitHubConfig.apiBaseUrl!,
+      logger,
+      cache: this.options.cache,
+    });
```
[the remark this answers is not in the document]
mtlewis: Made token optional. We're not explicitly handling the no-token case elsewhere, and it's possible for things like authenticating proxies to change the behavior for requests that might appear to be unauthenticated, so I think it's best for us to stay unopinionated.

plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts
```diff
@@ -374,14 +401,23 @@ export class GithubOrgEntityProvider implements EntityProvider {
...
     const client = graphql.defaults({
       baseUrl: this.options.gitHubConfig.apiBaseUrl,
       headers,
     });
+    const restClient = createRestClient({
+      token: token!,
+      baseUrl: this.options.gitHubConfig.apiBaseUrl!,
+      logger: this.options.logger,
+      cache: this.options.cache,
+    });
```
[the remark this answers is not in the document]
mtlewis: https://github.com/backstage/backstage/pull/34170#discussion_r3220975979

plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts
```diff
@@ -297,26 +313,36 @@ export class GithubMultiOrgEntityProvider implements EntityProvider {
...
       const client = graphql.defaults({
         baseUrl: this.options.gitHubConfig.apiBaseUrl,
         headers,
       });
+      const restClient = createRestClient({
+        token: token!,
+        baseUrl: this.options.gitHubConfig.apiBaseUrl!,
+        logger,
+        cache: this.options.cache,
+      });
```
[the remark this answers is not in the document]
mtlewis: https://github.com/backstage/backstage/pull/34170#discussion_r3220975979

plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts
```diff
@@ -686,14 +722,23 @@ export class GithubMultiOrgEntityProvider implements EntityProvider {
...
     const client = graphql.defaults({
       baseUrl: this.options.gitHubConfig.apiBaseUrl,
       headers,
     });
+    const restClient = createRestClient({
+      token: token!,
+      baseUrl: this.options.gitHubConfig.apiBaseUrl!,
+      logger: this.options.logger,
+      cache: this.options.cache,
+    });
```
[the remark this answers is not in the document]
mtlewis: https://github.com/backstage/backstage/pull/34170#discussion_r3220975979

plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts
```diff
@@ -452,24 +478,34 @@ export class GithubMultiOrgEntityProvider implements EntityProvider {
...
```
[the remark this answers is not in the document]
mtlewis: https://github.com/backstage/backstage/pull/34170#discussion_r3220975979

plugins/catalog-backend-module-github/src/processors/GithubOrgReaderProcessor.ts
```diff
@@ -138,16 +148,25 @@ export class GithubOrgReaderProcessor implements CatalogProcessor {
...
+    });
 
     const client = graphql.defaults({
       baseUrl: gitHubConfig.apiBaseUrl,
       headers,
     });
 
-    return { client, tokenType };
+    const restClient = createRestClient({
+      token: token!,
```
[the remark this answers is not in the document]
mtlewis: https://github.com/backstage/backstage/pull/34170#discussion_r3220975979

plugins/catalog-backend-module-github/src/processors/GithubMultiOrgReaderProcessor.ts
```diff
@@ -131,21 +132,30 @@ export class GithubMultiOrgReaderProcessor implements CatalogProcessor {
...
+          token,
+        } = await this.githubCredentialsProvider.getCredentials({
+          url: `${baseUrl}/${orgConfig.name}`,
+        });
         const client = graphql.defaults({
           baseUrl: gitHubConfig.apiBaseUrl,
           headers,
         });
+        const restClient = createRestClient({
+          token: token!,
```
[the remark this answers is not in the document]
mtlewis: https://github.com/backstage/backstage/pull/34170#discussion_r3220975979

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -216,9 +216,12 @@ export async function getOrganizationUsers(
 
   // There is no user -> teams edge, so we leave the memberships empty for
   // now and let the team iteration handle it instead
+  const isGitHubEnterprise = await restClient
+    .request('GET /versions')
+    .then(response => !!response.headers['x-github-enterprise-version']);
```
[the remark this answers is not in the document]
mtlewis: We're not expecting many people to set `dangerouslySkipSuspendedUserCheck` to true, but does seem fine to skip this check when it is enabled, especially since that branch is often going to be enabled because of some other issue like rate limits. Done in 00e689a.

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -920,11 +917,145 @@ export const createGraphqlClient = (args: {
...
+  octokit: Octokit,
+  cache: CacheService,
+): void {
+  octokit.hook.wrap('request', async (request, options) => {
+    const resolvedUrl = (options.url || '').replace(/\{([^}]+)\}/g, (_, key) =>
+      encodeURIComponent((options as any)[key]),
+    );
+    const cacheKey = `${options.method}:${options.baseUrl}${resolvedUrl}`;
+    const cached = await cache
+      .get<CachedGitHubResponse>(cacheKey)
```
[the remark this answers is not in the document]
mtlewis: 3c3eff61dfd8b5032c2e0543e0d15a892c952f31 adds the package name to the cache key. Given we're already including the baseURL in the key, credentials feels like overkill.

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -920,11 +917,145 @@ export const createGraphqlClient = (args: {
...
+  const userSuspended = !!userSuspendedAt;
+  const orgMembershipSuspended = membershipRole === 'suspended';
+
+  return userSuspended || orgMembershipSuspended;
+}
+
+async function isGitHubEnterprise(restClient: Octokit): Promise<boolean> {
+  return restClient
+    .request('GET /versions')
+    .then(response => !!response.headers['x-github-enterprise-version']);
```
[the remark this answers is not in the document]
mtlewis: c06a67c077e1fca1e8c862dd0fdc9d9709db236c

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -920,11 +917,145 @@ export const createGraphqlClient = (args: {
...
+export function createRestClient(options: {
+  token: string | undefined;
+  baseUrl: string;
+  logger: LoggerService;
+  cache?: CacheService;
+}): Octokit {
+  const octokit = createThrottledOctokit(options.logger, {
+    auth: options.token,
+    baseUrl: options.baseUrl,
   });
```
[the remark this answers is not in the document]
mtlewis: 65945c71f06b4b7297158dedb6abf0c7695a636e

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -920,11 +917,157 @@ export const createGraphqlClient = (args: {
...
+  cache?: CacheService;
+}): Octokit {
+  const octokit = createThrottledOctokit(options.logger, {
+    baseUrl: options.baseUrl,
   });
 
-  return client;
+  if (options.headers) {
+    const defaultHeaders = options.headers;
+    octokit.hook.wrap('request', (request, requestOptions) => {
```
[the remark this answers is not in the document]
mtlewis: Any particular reason that we shouldn't trust the octokit types here? They assert that `headers` is always defined.

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -920,11 +917,157 @@ export const createGraphqlClient = (args: {
...
+        }
+  ).role;
+
+  const userSuspended = !!userSuspendedAt;
+  const orgMembershipSuspended = membershipRole === 'suspended';
+
+  return userSuspended || orgMembershipSuspended;
+}
+
+export async function isGitHubEnterprise(
```
deiga: question: I think this will return false for GitHub Enterprise Cloud and GitHub Enterprise Cloud with Data Governance instances, can you confirm?

We're on GHEC with EMU, so we have suspended users. They don't have `suspended_at` field, but they do have `"role": "suspended",`.
Should this code support those instances as well?

Obfuscated response from org memberships:
```
{
  "url": "https://api.github.com/orgs/GHECOrg/memberships/SuspendedUserLogin",
  "state": "active",
  "role": "suspended",
  "organization_url": "https://api.github.com/orgs/GHECOrg",
  "user": {
    "login": "SuspendedUserLogin",
    "id": 217711563,
    "node_id": "U_kgDODPoDyw",
    "avatar_url": "https://avatars.githubusercontent.com/u/217711563?v=4",
    "gravatar_id": "",
    "url": "https://api.github.com/users/SuspendedUserLogin",
    "html_url": "https://github.com/SuspendedUserLogin",
    "followers_url": "https://api.github.com/users/SuspendedUserLogin/followers",
    "following_url": "https://api.github.com/users/SuspendedUserLogin/following{/other_user}",
    "gists_url": "https://api.github.com/users/SuspendedUserLogin/gists{/gist_id}",
    "starred_url": "https://api.github.com/users/SuspendedUserLogin/starred{/owner}{/repo}",
    "subscriptions_url": "https://api.github.com/users/SuspendedUserLogin/subscriptions",
    "organizations_url": "https://api.github.com/users/SuspendedUserLogin/orgs",
    "repos_url": "https://api.github.com/users/SuspendedUserLogin/repos",
    "events_url": "https://api.github.com/users/SuspendedUserLogin/events{/privacy}",
    "received_events_url": "https://api.github.com/users/SuspendedUserLogin/received_events",
    "type": "User",
    "user_view_type": "public",
    "site_admin": false
  },
  "direct_membership": true,
  "enterprise_teams_providing_indirect_membership": [],
  "organization": {
    "login": "GHECOrg",
    "id": 123456678,
    "node_id": "...",
    "url": "https://api.github.com/orgs/GHECOrg",
    "repos_url": "https://api.github.com/orgs/GHECOrg/repos",
    "events_url": "https://api.github.com/orgs/GHECOrg/events",
    "hooks_url": "https://api.github.com/orgs/GHECOrg/hooks",
    "issues_url": "https://api.github.com/orgs/GHECOrg/issues",
    "members_url": "https://api.github.com/orgs/GHECOrg/members{/member}",
    "public_members_url": "https://api.github.com/orgs/GHECOrg/public_members{/member}",
    "avatar_url": "https://avatars.githubusercontent.com/u/123456678?v=4",
    "description": "..."
  }
}

```

deiga (commented):
Should this change also support GHEC instances?

plugins/catalog-backend-module-github/src/lib/github.ts
mtlewis: Hey @deiga, thanks for the review. I think this currently supports both GHE and GHEC. In both cases, an `x-github-enterprise-version` header gets returned on the response, so this method will return `true` in both cases.

As far as checking for the suspended role goes - again I think we're already good, unless I'm missing something. If you check out `isSuspended` above, it hits both `/users/{username}` and `orgs/<org>/memberships/{username}`. The former returns a response with `suspended_at` on it if the user is suspended, and the latter returns `role: 'suspended'` if the user's membership from the org has been suspended. We consider the user suspended and exclude them if either is true.

.changeset/suspended-users-rest-api.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-catalog-backend-module-github': minor
+'@backstage/plugin-catalog-backend-module-github-org': minor
+---
+
```
benjdlambert, line 5: Think we should prefix this with: `**BREAKING**` like previous breaking changes too.

I think we can skip the mentioning about configuring the cache right? Should we just make it required in the interface? It's already a breaking change, and it could be argued that those providers shouldn't be exported out of the public API anyways and should just be driven by config and installed modules? 

Maybe I'm misunderstanding some use-case though for them being exported?

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -920,11 +917,157 @@ export const createGraphqlClient = (args: {
...
+      const etag = response.headers.etag;
+
+      if (lastModified || etag) {
+        cache
+          .set(
+            cacheKey,
+            {
+              lastModified,
+              etag,
+              headers: response.headers as unknown as JsonValue,
```
benjdlambert: Is it better to do `JSON.parse(JSON.stringify(...))` here to remove anything that's non-serializable? 

Also we drop the cast at that point too.

.changeset/suspended-users-rest-api.md
mtlewis: I suppose they're probably mostly exported by the `-github` module so they can be re-exported by the `-org` module. But yes, if we're marking this as a breaking change anyway, and if that breaking change will actually be handled in reality by DI most of the time, I think it probably makes sense to just make the cache required.

plugins/catalog-backend-module-github/src/lib/github.ts
mtlewis: Good idea! 70d4296422cb7

plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts
```diff
@@ -220,6 +238,25 @@ export class GithubOrgEntityProvider implements EntityProvider {
...
+    }
+    const { headers } = await this.credentialsProvider.getCredentials({
+      url: this.options.orgUrl,
+    });
+    const restClient = createRestClient({
+      headers,
+      baseUrl: this.options.gitHubConfig.apiBaseUrl!,
+      logger: this.options.logger,
+      cache: this.options.cache,
+    });
```
Rugvip: Hmm, this is gonna bypass the throttling? 🤔, perhaps worth keeping clients around per org at least? Ideally we forward headers per request and re-use the same client though.

plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts
```diff
@@ -261,15 +298,23 @@ export class GithubOrgEntityProvider implements EntityProvider {
       logger,
     });
 
+    const restClient = createRestClient({
```
Rugvip: Same here

plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts
```diff
@@ -382,6 +427,12 @@ export class GithubOrgEntityProvider implements EntityProvider {
       baseUrl: this.options.gitHubConfig.apiBaseUrl,
       headers,
     });
+    const restClient = createRestClient({
```
Rugvip: and here 😁

plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts
```diff
@@ -514,7 +566,7 @@ export class GithubOrgEntityProvider implements EntityProvider {
 
     const mutationEntities: Entity[] = [team];
 
-    if (user && isUserEntity(user)) {
+    if (user && isUserEntity(user) && !(await this.shouldExclude(login, org))) {
```
Rugvip: It seems wrong to simply ignore the exclude case. I'm thinking that should result in a mutation to remove the user? Right now we might actually be delaying the removal of the user to the next refresh right? We can an event to suspend the user but then ignore it because they're suspended :p

plugins/catalog-backend-module-github/src/lib/github.ts
```diff
@@ -920,11 +917,157 @@ export const createGraphqlClient = (args: {
...
+    const cacheKey = `catalog-backend-module-github:${options.method}:${options.baseUrl}${resolvedUrl}`;
+    const cached = await cache
+      .get<CachedGitHubResponse>(cacheKey)
+      .catch(() => undefined);
+
+    if (cached?.lastModified) {
+      options.headers['if-modified-since'] = cached.lastModified;
+    } else if (cached?.etag) {
+      options.headers['if-none-match'] = cached.etag;
+    }
```
Rugvip, line 987: Bit risky this if we're relying on the cache to avoid hammering the upstream GitHub instance, because some GHE instances basically ignore the cache. As in it's just as expensive to handle a request regardless of whether there's a cache hit or not.

Can we expect this to function reasonably well even without the cache? I think one of the important bits is to make sure the throttling is active to make sure we don't happen the instance too hard.

Rugvip (changes requested):
Nice! 👍 

Couple of things, also thinking we want to ship this after mainline

.changeset/suspended-users-rest-api.md
mtlewis: 760563ec00211

plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts
mtlewis: Hadn't looked at the throttling too closely - from the parameters I'd assumed that it was mostly handling hitting rate limits which of course is stateless on the client side, but on further investigation I see that it does do some client-side throttling to avoid making requests in parallel, so yeah, I think we should do some stuff to avoid having a bunch of independent throttles scattered around. Will explore a couple of options - agree it would be ideal to just pass clients around if that refactoring doesn't end up too painful. Alternative is exploring using the built-in support for state sharing in the Octokit plugin itself, but that seems needlessly complicated if the refactoring to share clients is doable, especially since it would require building a shim around cacheService to make it match the redis cache type they're expecting.

plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts
mtlewis: :+1: will cover all of these above, so gonna resolve them just to make it clear what needs working on here.

plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts
mtlewis: :+1: will cover all of these above, so gonna resolve them just to make it clear what needs working on here.

plugins/catalog-backend-module-github/src/lib/github.ts
deiga: Neither `/users/{username}` not `/orgs/<org>/memberships/{username}` returns  `x-github-enterprise-version` for GHEC on github.com

plugins/catalog-backend-module-github/src/lib/github.ts
mtlewis: Without the cache, it's definitely going to be expensive in rate limit points - 2 additional points _per user_ per ingestion, where we're currently spending only a few GQL points per ~hundred users.

The [octokit throttling plugin](https://github.com/octokit/plugin-throttling.js/) is pretty vague about exactly how requests are throttled, but looking at the code it's making heavy use of [bottleneck](https://www.npmjs.com/package/bottleneck) to control concurrency across a few different classes of request. No configurability, but it at least means that we won't routinely make a bunch of massively parallel requests - there are some groups though where concurrency is permitted (https://github.com/octokit/plugin-throttling.js/blob/main/src/index.ts#L24-L56).

I'll take a look at addressing the comment above and then perhaps we can revisit this.

plugins/catalog-backend-module-github/src/providers/GithubOrgEntityProvider.ts
mtlewis: Yeah, good point. I suppose we could also remove this check entirely since this event is really nothing to do with the user's suspended state, but seems like a good moment to recheck.

cbbd66e2bf306

plugins/catalog-backend-module-github/src/lib/github.ts
deiga: @mtlewis what do you think about this? Should GHEC also be supported?

deiga:
This needs to have an explicit mention that the change doesn't work for GHEC if support is not planned to be added

PR #34176, opened 2026-05-09
draft(catalog-backend): deduplicate search rows and add UNIQUE constraint

freben:

> Draft. Prerequisite for the `/entities/by-query` ordering optimization in #34175. Can be reviewed and landed independently.

### Problem

The `search` table has ~434k duplicate rows on production (~3% of entities affected). These duplicates originate from two sources:

1. **Concurrent stitching races** — the old DELETE-all + INSERT-all code path could race between pods, and even the newer `syncSearchRows` upsert has a narrow window where two pods inserting the same entity produce duplicates.
2. **Duplicate array values in entity data** — `buildEntitySearch` didn't deduplicate array-derived entries, so `tags: ['java', 'java']` produced two identical search rows.

These duplicates force `queryEntities` to use `DISTINCT` in its CTE, which prevents the planner from using the `(key, value, entity_id)` index in sort order and short-circuiting on `LIMIT`. Removing `DISTINCT` is the key prerequisite for the ~100× ordering optimization in #34175.

### Solution

Three coordinated changes:

**1. Migration (`20260509000000_search_deduplicate.js`)**

Batch-deletes existing duplicates using a window-function CTE (`PARTITION BY entity_id, key, value`), then creates a `UNIQUE INDEX CONCURRENTLY` on `(entity_id, key, value)`. Handles all three engines:

- **PostgreSQL**: Window-function CTE delete + `CREATE UNIQUE INDEX CONCURRENTLY`. If the non-unique `search_entity_key_value_idx` from #34015 already exists, it's dropped and recreated as unique.
- **MySQL**: Temp table merge (keeps one row per group) + unique index.
- **SQLite**: `DELETE WHERE rowid NOT IN (SELECT MIN(rowid) ...)` + unique index.

Non-transactional migration (`exports.config = { transaction: false }`) to support `CONCURRENTLY` on PG.

**2. `buildEntitySearch` dedup**

Filters the output by `(key, value)` so that duplicate array values (e.g., `tags: ['java', 'java']`) produce one search row, not two. Prevents the source of duplicates.

**3. `syncSearchRows` hardening**

- Deduplicates input entries at the function entry point (all engines).
- Adds `ON CONFLICT (entity_id, key, value) DO UPDATE SET original_value = EXCLUDED.original_value` to the PostgreSQL INSERT path. If a concurrent stitching race produces a conflict, the latest `original_value` wins silently rather than erroring.

### Production data

Measured on the production replica (~462k entities, ~13.5M search rows):

- **434k extra duplicate rows** across 431k groups
- **8,189 NULL-value duplicate groups** (also cleaned by the migration — `PARTITION BY` treats NULLs as equal unlike `UNIQUE`)
- **0 cases** where `(entity_id, key, value)` matched but `original_value` differed — no information loss from deduplication
- **Dedup CTE scan**: ~7 seconds over the full table (acceptable at startup)

### Rolling deploy safety

During a rolling deploy where old and new pods coexist:

- **Old pods** (DELETE-all + INSERT-all): If `buildEntitySearch` produces duplicate array values, the INSERT hits the UNIQUE constraint and fails. The entity's stitch is retried on the next cycle. Once all pods are updated, the retry succeeds (deduplicated input). This window is brief and self-healing.
- **New pods**: `buildEntitySearch` dedup + `ON CONFLICT DO UPDATE` = no constraint violations.

### Relationship to other PRs

```
This PR (dedup + UNIQUE)
  ↓ enables
#34175 (queryEntities INNER JOIN, removes DISTINCT)
  ↓ which enables
~100× speedup on paginated entity listing
```

### DIFF
--- a/.changeset/search-dedup-unique.md
+++ b/.changeset/search-dedup-unique.md
@@ -1,0 +1,7 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Added a migration that removes duplicate rows from the `search` table and adds a `UNIQUE` constraint on `(entity_id, key, value)`. This prevents future duplicates from concurrent stitching races and enables removing the defensive `DISTINCT` in entity listing queries, which is a prerequisite for the planned query ordering optimizations.
+
+Also fixed `buildEntitySearch` to deduplicate its output (e.g. when an entity has duplicate array values like `tags: ['java', 'java']`), and added `ON CONFLICT DO NOTHING` to the PostgreSQL write path in `syncSearchRows` so that any remaining race-condition duplicates are silently skipped rather than causing errors.
--- a/plugins/catalog-backend/migrations/20260509000000_search_deduplicate.js
+++ b/plugins/catalog-backend/migrations/20260509000000_search_deduplicate.js
@@ -1,0 +1,253 @@
+/*
+ * Copyright 2026 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


freben:
closing in favor of https://github.com/backstage/backstage/pull/34177 and https://github.com/backstage/backstage/pull/34175

PR #34178, opened 2026-05-10
fix(scaffolder): skip empty GitLab commits when no actions exist

BhagyashreeKarale:
## Summary

Fixes an issue in `gitlab:repo:push` where GitLab commit creation fails with a 400 error when no file actions are generated.

Previously, scenarios such as:
- empty workspace/sourcePath
- `commitAction: auto` with no effective file changes

would still call the GitLab commit API with an empty actions array, resulting in:

`Provide at least one action, or set allow_empty to true`

This change adds an early no-op return when no commit actions exist, avoiding unnecessary API calls and treating the workflow as a successful no-op.

## Changes
- Added guard clause for empty commit actions
- Added warning log for skipped commit operations
- Preserved existing action outputs with empty `commitHash`

## Validation
Validated locally against the `gitlabRepoPush` action flow and confirmed empty action sets no longer attempt GitLab commit creation.

### DIFF
--- a/.changeset/mighty-keys-join.md
+++ b/.changeset/mighty-keys-join.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend-module-gitlab': patch
+---
+
+Skip GitLab commit creation when no file actions are generated.
--- a/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabRepoPush.examples.test.ts
+++ b/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabRepoPush.examples.test.ts
@@ -119,7 +119,7 @@ describe('gitlab:repo:push', () => {
       mockDir.setContent({
         [workspacePath]: {
           'abcd.txt': 'Test message',
-          source: {
+          src: {
             'abcd.txt': 'Test message',
           },
         },
@@ -135,7 +135,7 @@ describe('gitlab:repo:push', () => {
         [
           {
             action: 'create',
-            filePath: 'abcd.txt',
+            filePath: 'dest/abcd.txt',
             content: 'VGVzdCBtZXNzYWdl',
             encoding: 'base64',
             execute_filemode: false,
--- a/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabRepoPush.ts
+++ b/plugins/scaffolder-backend-module-gitlab/src/actions/gitlabRepoPush.ts
@@ -174,6 +174,18 @@ export const createGitlabRepoPushAction = (options: {
           execute_filemode: file.executable,
         }));
 
+      if (actions.length === 0) {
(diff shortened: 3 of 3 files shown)


elaine-mattos (commented):
Hey @BhagyashreeKarale! I tackled the same issue in #33602 but took a different approach by exposing `allow_empty` as an explicit option instead of skipping the call. Silent no-op can break workflows that intentionally push empty commits to trigger CI. You could take a look if you want, would it also solve your issue?

BhagyashreeKarale:
> Hey @BhagyashreeKarale! I tackled the same issue in #33602 but took a different approach by exposing `allow_empty` as an explicit option instead of skipping the call. Silent no-op can break workflows that intentionally push empty commits to trigger CI. You could take a look if you want, would it also solve your issue?

@elaine-mattos Thanks for pointing this out... that’s a good consideration regarding workflows that intentionally rely on empty commits for CI triggers.

My initial intent here was to prevent the GitLab API failure path for cases where no effective file actions are generated, especially for empty workspace/sourcePath scenarios.

I’ll take a closer look at #33602 and see whether aligning with an explicit allow_empty option would be a better long-term approach here while still addressing the failing no-op cases.

BhagyashreeKarale:
> > Hey @BhagyashreeKarale! I tackled the same issue in #33602 but took a different approach by exposing `allow_empty` as an explicit option instead of skipping the call. Silent no-op can break workflows that intentionally push empty commits to trigger CI. You could take a look if you want, would it also solve your issue?
> 
> @elaine-mattos Thanks for pointing this out... that’s a good consideration regarding workflows that intentionally rely on empty commits for CI triggers.
> 
> My initial intent here was to prevent the GitLab API failure path for cases where no effective file actions are generated, especially for empty workspace/sourcePath scenarios.
> 
> I’ll take a closer look at #33602 and see whether aligning with an explicit allow_empty option would be a better long-term approach here while still addressing the failing no-op cases.

@elaine-mattos  I explored the explicit allow_empty approach as well, but the current @gitbeaker/rest typings used in Backstage don’t expose allow_empty in Commits.create, which would require either type assertions or a larger upstream typing change.

Given that, and since the original issue also mentioned graceful no-op behavior as a valid resolution path, I kept the implementation focused on safely skipping commit creation when no effective actions are generated, while preserving checkpoint/output behavior. Would love to get your opinion on this or suggestions if any.

elaine-mattos (changes requested):
hey @BhagyashreeKarale , thanks for working on this. I think this has the same issue discussed in #34100.

Because `allowEmpty` now exists on current `master`, an unconditional early return when `actions.length === 0` would regress intentional empty-commit workflows as well. Could you please rebase on latest `master` and align with the approach discussed there before we move this forward?

elaine-mattos:
Also, looks like #34480 has addressed the issue....

PR #34185, opened 2026-05-11
feat(scaffolder): add parameters to form decorator context

minusinus:

This adds a further context to scaffolder decorators by providing the original parameters schema.

This would help me with my use-case. I am curious to hear if broadening the context would make sense as well (e.g. exposing the entire `manifest` or `manifest.steps` instead)

### DIFF
--- a/.changeset/scaffolder-form-decorator-parameters.md
+++ b/.changeset/scaffolder-form-decorator-parameters.md
@@ -1,0 +1,6 @@
+---
+'@backstage/plugin-scaffolder-react': minor
+'@backstage/plugin-scaffolder': minor
+---
+
+Form decorators now receive the template's parameter schemas via a new `parameters` property on the decorator context. The value is the array of parameter step schemas as written in the template's `spec.parameters`, allowing decorators to inspect the form's structure when deciding how to mutate state or secrets.
--- a/plugins/scaffolder-react/report-alpha.api.md
+++ b/plugins/scaffolder-react/report-alpha.api.md
@@ -270,6 +270,7 @@ export type ScaffolderFormDecoratorContext<
 > = {
   input: TInput;
   formState: Record<string, JsonValue>;
+  parameters: JsonObject[];
   setFormState: (
     fn: (currentState: Record<string, JsonValue>) => Record<string, JsonValue>,
   ) => void;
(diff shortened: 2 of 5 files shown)


minusinus:
Alternatively, I think a `steps: Record<string, JsonValue>` could be good - to match the `formState`.

benjdlambert:
:wave: Hey!

First off, thanks for this! I can see why this might be something that would be wanted in decorators, but I wanna take a step back a little bit first and understand what's the use case that you have in mind? Theres a few little thoughts from me below here as to why I'd be hesitant to add this, but please reply with more of your use cases and how you plan to use this :pray: 

That said, this feels a little bit like a pretty old statement I wrote in some scaffolder docs [way back when](https://backstage.io/docs/next/features/software-templates/writing-custom-field-extensions/#access-data-from-other-fields) with coupling `FieldExtensions` to the `parameters` schema, or the model / context that was available.

It makes `FieldExtensions` or even in this case `Decorators` very coupled to the form and the parameters schema that it creates (and maybe even other `Decorators` and that order), and I'm not sure that's a great fit. My vision for decorators is that they would be very much plug and play, and you would write them once and they work for all templates to attach additional secrets, or maybe even to help with migrations and mutating the values that get sent over the wire.

I could maybe see a path forward here where we provide something like an `{ entityRef: .... }` that you could use to go an introspect the schema and make decisions based of that, but I think that allowing the raw `parameters` block in here is maybe a little bit too brittle from a consumption point of view. 

One pattern we've seen before is that the `formDecorator` takes a dot notation argument to pull things from the values right also if that's what you're looking for?

Something like:

```yaml
formDecorators:
 - id: test
    input:
      testKey: thing.dot.lol
```

Where you might do something like `formState[input.testKey]` to grab that value, whilst being pretty flexible that the decorator can work on a lot of base templates?

Another thing that I'm reasoning about is that it's really hard to type the incoming parameters in a safe way, we can basically resolve it to being a `JSONSchema` and that's about it, the rest is purely business + custom logic, so it makes these decorators pretty brittle. If they're that coupled to some templates, how do we plan to re-use them, what's the need to pull in the parameters from the template? Would this logic be better placed as a `FieldExtension` itself which is 100% coupled to the input field at least?

minusinus:
Hey! Thanks, your reasoning lands.

I do not think schema access actually couples a decorator to templates, in the way you may be implying.JSONSchema (with rjsf) is generic enough. A decorator walking it for, say, fields with a specific `ui:field` works on every template that uses that field.

For rough context, we have a custom field whose output _includes_ an `entityRef`, this needs a small post-submit cleanup plus a bit of bookkeeping to track the JSONPath of said field and that then ends up downstream. Today we handle it with template-side filter chains, which works but arguably could be better.

That said, I am open to making this work on my end by simply emitting sentinel values from the fields that I am interested in, so I likely have a workable design without this PR.

Thanks again, feel free to close.

PR #34190, opened 2026-05-11
fix(catalog-backend): simplify facets COUNT(DISTINCT) to COUNT(*)

freben:

The UNIQUE constraint on `(entity_id, key, value)` added in #34177 guarantees each entity appears at most once per `(key, original_value)` group in the search table. This makes the `DISTINCT` in `COUNT(DISTINCT search.entity_id)` unnecessary — `COUNT(*)` produces the same result with a simpler aggregation plan.

**Why this works:** For any `(key, original_value)` group, each entity can contribute at most one row because:
1. `original_value` is derived from `value` (case-preserved form of the lowercased `value`)
2. The UNIQUE constraint on `(entity_id, key, value)` prevents duplicate `(entity_id, key, value)` rows
3. Therefore each entity appears at most once per `(key, value)` — and thus at most once per `(key, original_value)` group

**Depends on:** #34177 (must be merged first — without the UNIQUE constraint, `COUNT(DISTINCT)` is still needed)

### DIFF
--- a/.changeset/facets-count-optimization.md
+++ b/.changeset/facets-count-optimization.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Simplified the entity facets aggregation from `COUNT(DISTINCT entity_id)` to `COUNT(*)`. The unique constraint on `(entity_id, key, value)` guarantees each entity appears at most once per search row group, making the `DISTINCT` unnecessary. This allows the database to use a simpler aggregation plan.
--- a/plugins/catalog-backend/src/service/DefaultEntitiesCatalog.test.ts
+++ b/plugins/catalog-backend/src/service/DefaultEntitiesCatalog.test.ts
@@ -2548,13 +2548,13 @@ describe('DefaultEntitiesCatalog', () => {
           },
         ]);
 
-        await expect(
-          catalog.facets({
-            facets: ['spec.type'],
-            query: { kind: 'component' },
-            credentials: mockCredentials.none(),
-          }),
-        ).resolves.toEqual({
+        const result = await catalog.facets({
+          facets: ['spec.type'],
+          query: { kind: 'component' },
+          credentials: mockCredentials.none(),
+        });
+        expect(result.facets['spec.type']).toHaveLength(2);
+        expect(result).toEqual({
           facets: {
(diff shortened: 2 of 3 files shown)


freben:
## EXPLAIN ANALYZE: `COUNT(DISTINCT entity_id)` vs `COUNT(*)`

Benchmarked on a read-only replica of production (13M search rows, 54K components). All timings are warm-cache, best of two runs.

### Filtered facets (`filter=kind%3Dcomponent`)

This is the real-world query pattern for `/entity-facets`.

| Facet | `COUNT(DISTINCT entity_id)` | `COUNT(*)` | Delta | Sort method |
|---|---|---|---|---|
| `spec.type` | 963 ms | 837 ms | **−13%** | disk → memory |
| `spec.lifecycle` | 975 ms | 969 ms | **−1%** | disk → memory |
| `kind` | 988 ms | 814 ms | **−18%** | disk → memory |

The structural change: `COUNT(DISTINCT entity_id)` sorts by `(original_value, entity_id)` to feed the GroupAggregate dedup — this exceeds `work_mem` and spills to disk (external merge, ~3.9 MB temp). `COUNT(*)` sorts by `(original_value)` only, which fits in memory (quicksort, ~3.9 MB). The disk I/O elimination is the main win.

### Unfiltered facets (no filter)

Shows the aggregate improvement in isolation, without the join dominating runtime.

| Facet | `COUNT(DISTINCT entity_id)` | `COUNT(*)` | Delta |
|---|---|---|---|
| `spec.type` | 144 ms | 82 ms | **−43%** |
| `spec.lifecycle` | 135 ms | 64 ms | **−52%** |

The larger improvement here is partly because the planner picks the smaller 2-column `search_key_original_value_idx` when `entity_id` isn't needed (45K buffer hits vs 428K). After #34177 drops that index, both variants would use `search_facets_covering_idx` — but the aggregate step is still simpler (no dedup sorting).

### Summary

The improvement is modest in the filtered case (10-18%) because the nested-loop join dominates the runtime (~1000 ms). The sort/aggregate phase goes from ~160 ms (disk) to ~20 ms (memory). For unfiltered facets it's ~2× faster.

The bigger payoff may be under memory pressure: `work_mem` exhaustion causes the disk spill, and eliminating it removes a source of latency variance.

benjdlambert (approved):

PR #34191, opened 2026-05-11
docs: fill out remaining backend-plugin golden path

aramissennyeydd:

Final TODO removals for the backend-plugin golden path.

### DIFF
--- a/docs/golden-path/plugins/backend/004-reading-from-source.md
+++ b/docs/golden-path/plugins/backend/004-reading-from-source.md
@@ -5,16 +5,182 @@ title: 004 - Git-tracked TODOs
 description: How to ingest TODOs from source code repositories into your plugin
 ---
 
-Problem: You have TODOs in your source code that you want to ingest with your plugin.
+Problem: Your users have a lot of `// TODO:` comments scattered across their repositories and would love to see them surfaced alongside the TODOs they create manually. Backstage already knows where each component lives — through the `backstage.io/source-location` annotation on catalog entities — so we can fetch the source for any component the user owns and harvest those comments.
+
+To do this we need three things:
+
+1. **Credentials** to talk to the SCM provider (GitHub, GitLab, …).
+2. A way to **find** the repositories we care about.
+3. A way to **fetch** their contents.
+
+Backstage gives us all three through the `integrations` config and the `urlReader` core service.
 
 ## Authenticating
 
-<!--TODO-->
+The `urlReader` service uses the `integrations` block of your config to figure out how to authenticate against each SCM. You don't need to write any auth code yourself — once a host is registered there, every call you make through `urlReader` will pick up the right token.
+
+For local development, define the integration in `app-config.local.yaml`. That file is ignored by Git, so it's the right home for development-only config that you don't want to force on every other environment:
+
+```yaml title="app-config.local.yaml"
+integrations:
+  github:
+    - host: github.com
+      token: ${GITHUB_TOKEN}
+```
+
+The token itself should still come from an environment variable rather than being pasted into the yaml — set `GITHUB_TOKEN` in a `.env` file (or your shell) so it never lands in any committed or shared file. A Personal Access Token with `repo` scope is enough for private repositories; public-only setups can omit the token entirely.
+
+We deliberately leave the committed `app-config.yaml` alone here. Production deployments often want a [GitHub App](https://backstage.io/docs/integrations/github/github-apps) instead of a PAT — the credentials rotate automatically and aren't tied to a single user — and putting the dev shape in the committed file would push a PAT-shaped config onto every environment.
+
+The same pattern works for `gitlab`, `bitbucketCloud`, `bitbucketServer`, and `azure`. See the [integrations reference](https://backstage.io/docs/integrations/) for the full list.
 
 ## Querying
 
-<!--TODO-->
+Now that we can authenticate, we need to know _which_ repositories to scan. We'll lean on the catalog: when the user asks for their TODOs, we'll look up the components they own and read the `backstage.io/source-location` annotation from each one.
+
+First, plumb the `urlReader` service into `TodoListService` alongside the catalog client we already have:
+
+```diff title="src/services/TodoListService.ts"
+ import {
+   coreServices,
+   createServiceFactory,
+   createServiceRef,
+   LoggerService,
+   DatabaseService,
++  UrlReaderService,
+ } from '@backstage/backend-plugin-api';
+
+ export const todoListServiceRef = createServiceRef<Expand<TodoListService>>({
+   id: 'todo.list',
+   defaultFactory: async service =>
+     createServiceFactory({
+       service,
+       deps: {
+         logger: coreServices.logger,
+         catalog: catalogServiceRef,
+         database: coreServices.database,
++        urlReader: coreServices.urlReader,
+       },
+       async factory(deps) {
+         return TodoListService.create(deps);
+       },
+     }),
+ });
+```
+
+Wire it through the constructor the same way you did for `database` in the previous step.
+
+Next, add a method that asks the catalog for the components owned by the calling user and returns their source locations:
+
+```ts title="src/services/TodoListService.ts"
+import { getEntitySourceLocation } from '@backstage/catalog-model';
+
+async listOwnedSources(options: {
+  credentials: BackstageCredentials;
+}): Promise<{ entityRef: string; url: string }[]> {
+  const { items } = await this.#catalog.getEntities(
+    {
+      filter: {
+        kind: 'Component',
+        'relations.ownedBy': options.credentials.principal.userEntityRef,
+      },
+      fields: ['kind', 'metadata', 'spec'],
+    },
+    { credentials: options.credentials },
+  );
+
+  return items
+    .map(entity => {
+      try {
+        const { type, target } = getEntitySourceLocation(entity);
+        if (type !== 'url') return undefined;
+        return {
+          entityRef: stringifyEntityRef(entity),
+          url: target,
+        };
+      } catch {
+        return undefined;
+      }
+    })
+    .filter((s): s is { entityRef: string; url: string } => Boolean(s));
+}
+```
+
+A couple of things to note:
+
+1. `getEntitySourceLocation` returns the URL pointing at the entity's source on the SCM (resolved from `backstage.io/source-location`, falling back to `backstage.io/managed-by-location`).
+2. We pass the caller's credentials through to the catalog so authorization is enforced — you should never run catalog reads with anonymous credentials inside a request handler.
 
 ## Fetching
 
-<!--TODO-->
+`urlReader` exposes three methods — `readUrl`, `readTree`, and `search` — and the right one depends on what you're after:
+
+| Method     | When to use it                                              |
+| ---------- | ----------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


docs/golden-path/plugins/backend/004-reading-from-source.md
```diff
@@ -5,16 +5,199 @@ title: 004 - Git-tracked TODOs
...
+For local development, define the integration in `app-config.local.yaml`. That file is ignored by Git, so it's the right home for development-only config that you don't want to force on every other environment:
+
+```yaml title="app-config.local.yaml"
+integrations:
+  github:
+    - host: github.com
+      token: ${GITHUB_TOKEN}
+```
+
+The token itself should still come from an environment variable rather than being pasted into the yaml — set `GITHUB_TOKEN` in a `.env` file (or your shell) so it never lands in any committed or shared file. A Personal Access Token with `repo` scope is enough for private repositories; public-only setups can omit the token entirely.
```
Parsifal-M: IDK about making an `.env` file, i'd either export it to the terminal or just put it in plain text in the local yaml and don't commit it 😅

docs/golden-path/plugins/backend/004-reading-from-source.md
```diff
@@ -5,16 +5,199 @@ title: 004 - Git-tracked TODOs
...
+
+We deliberately leave the committed `app-config.yaml` alone here. Production deployments often want a [GitHub App](https://backstage.io/docs/integrations/github/github-apps) instead of a PAT — the credentials rotate automatically and aren't tied to a single user — and putting the dev shape in the committed file would push a PAT-shaped config onto every environment.
+
+The same pattern works for `gitlab`, `bitbucketCloud`, `bitbucketServer`, and `azure`. See the [integrations reference](https://backstage.io/docs/integrations/) for the full list.
```
Parsifal-M: I mean sure, but you will still need a token eventually added no?

docs/golden-path/plugins/backend/004-reading-from-source.md
```diff
@@ -5,16 +5,199 @@ title: 004 - Git-tracked TODOs
...
+
+  return { items: discovered };
+}
+```
+
+A few things worth highlighting:
+
+1. **Globs are scoped to the source URL.** `urlReader.search` translates `https://github.com/org/repo/tree/main/**/*.ts` into the right per-provider API call — you don't need to know whether the backend is the GitHub Trees API or the GitLab Repository API.
+2. **Authentication is implicit.** Because `github.com` is configured under `integrations`, the search request goes out with the token your env var resolved to. If a user owns a repo your backend doesn't have credentials for, the call will fail with a clear error rather than silently returning nothing.
+3. **Persistence stays in the service layer.** Rather than reaching into `this.#database` here, we delegate to a `createTodos` method that lives next to `createTodo`. The sync flow doesn't need to know how rows are shaped or how conflicts are resolved — that stays the database layer's job.
```
Parsifal-M: I wonder for both of these (also the notes below) add them as short(er) comments inline on the code examples? WDYT

Parsifal-M (commented):
Some small comments! I didn't go over the copilot stuff, once they are resolved will do another quick check! 🙏 

Awesome stuff! 🥳

docs/golden-path/plugins/backend/004-reading-from-source.md
aramissennyeydd: yeah I like it

docs/golden-path/plugins/backend/005-testing.md
```diff
@@ -9,22 +9,136 @@ description: How to write unit tests for your Backstage backend plugin
...
+
+    expect(response.status).toEqual(200);
+    expect(response.body).toEqual({ items: [] });
+  });
+});
+```
+
+A few patterns worth internalising here:
+
+1. **`mockServices` gives you fakes for every core service.** `mockServices.httpAuth()` returns a credential resolver that always succeeds — perfect for tests that aren't actually about auth. There's a corresponding `mockServices.<service>.mock()` that returns a Jest mock you can assert on if you _do_ want to verify auth behaviour.
```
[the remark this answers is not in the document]
aramissennyeydd: Good catch — reworded to describe the actual mock-user-principal default and called out `mockCredentials.none.header()` for explicitly unauthenticated requests.

docs/golden-path/plugins/backend/005-testing.md
```diff
@@ -9,22 +9,136 @@ description: How to write unit tests for your Backstage backend plugin
...
+    // The plugin's `init` ran migrations against an in-memory SQLite DB
+    // before the server started accepting requests.
+    const listResponse = await request(server).get('/api/todo/todos');
+    expect(listResponse.status).toEqual(200);
+    expect(listResponse.body).toEqual({ items: [] });
+
+    const createResponse = await request(server)
+      .post('/api/todo/todos')
+      .set('Authorization', 'Bearer mock-id-token')
+      .send({ title: 'A new todo' });
```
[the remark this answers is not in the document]
aramissennyeydd: Fixed — switched to `mockCredentials.user.header()` from `@backstage/backend-test-utils` so the test doesn't rely on a non-existent mock token.

docs/golden-path/plugins/backend/005-testing.md
```diff
@@ -9,22 +9,136 @@ description: How to write unit tests for your Backstage backend plugin
...
+-     app = express().use(router);
++     app = wrapInOpenApiTestServer(express().use(router));
+    });
+```
 
 ### Integration with Jest tests
 
-<!--TODO-->
+You don't need a separate test file or a separate Jest configuration. With the wrapper in place, every `request(app)` call your existing router tests already make is checked against `src/schema/openapi.yaml`: the route has to exist in the spec, the status code has to be declared, and the response body has to satisfy the declared schema — including any `required` properties. A mismatch surfaces as a normal Jest failure on the test case that produced the bad response.
+
```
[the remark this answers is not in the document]
aramissennyeydd: Switched the snippet to `wrapServer` instead of `wrapInOpenApiTestServer`. `wrapServer` is the newer helper — it unconditionally boots an in-process mockttp proxy that validates against the spec during a normal `yarn test` run (the Optic-based `wrapInOpenApiTestServer` path is being phased out). Added a note flagging `wrapInOpenApiTestServer` as deprecated.

docs/golden-path/plugins/backend/004-reading-from-source.md
```diff
@@ -5,16 +5,182 @@ title: 004 - Git-tracked TODOs
...
+
+async listOwnedSources(options: {
+  credentials: BackstageCredentials;
+}): Promise<{ entityRef: string; url: string }[]> {
+  const { items } = await this.#catalog.getEntities(
+    {
+      filter: {
+        kind: 'Component',
+        'relations.ownedBy': options.credentials.principal.userEntityRef,
+      },
```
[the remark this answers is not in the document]
aramissennyeydd: Typed the parameter as `BackstageCredentials<BackstageUserPrincipal>` (and added a comment explaining the router enforces `allow: ['user']`).

docs/golden-path/plugins/backend/004-reading-from-source.md
```diff
@@ -5,16 +5,182 @@ title: 004 - Git-tracked TODOs
...
+
+  return items
+    .map(entity => {
+      try {
+        const { type, target } = getEntitySourceLocation(entity);
+        if (type !== 'url') return undefined;
+        return {
+          entityRef: stringifyEntityRef(entity),
+          url: target,
+        };
```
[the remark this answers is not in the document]
aramissennyeydd: Added the `stringifyEntityRef` import from `@backstage/catalog-model` to the snippet.

docs/golden-path/plugins/backend/004-reading-from-source.md
aramissennyeydd: agree - dropped the `.env` indirection and recommended just pasting the PAT directly into `app-config.local.yaml`

docs/golden-path/plugins/backend/004-reading-from-source.md
aramissennyeydd: Yep - tweaked the production paragraph to acknowledge a GitHub App still ships credentials, the difference is they rotate automatically and aren't tied to a single user.

docs/golden-path/plugins/backend/004-reading-from-source.md
aramissennyeydd: Done — converted the highlights under "Querying", "Fetching", and "Persisting in bulk" into shorter inline comments on the code blocks; kept one short prose callout where the point was about ops/config rather than the code itself.

docs/golden-path/plugins/backend/004-reading-from-source.md
```diff
@@ -5,16 +5,207 @@ title: 004 - Git-tracked TODOs
...
+3. A way to **fetch** their contents.
+
+Backstage gives us all three through the `integrations` config and the `urlReader` core service.
 
 ## Authenticating
 
-<!--TODO-->
+The `urlReader` service uses the `integrations` block of your config to figure out how to authenticate against each SCM. You don't need to write any auth code yourself — once a host is registered there, every call you make through `urlReader` will pick up the right token.
+
+For local development, define the integration in `app-config.local.yaml`. That file is ignored by Git, so you can paste a Personal Access Token straight in without it landing in any committed or shared file:
```
milliehartnt123, line 22:
```suggestion
For local development, define the integration in `app-config.local.yaml`. That file is ignored by Git, so you can paste a Personal Access Token (PAT) straight in without it landing in any committed or shared file:
```

docs/golden-path/plugins/backend/004-reading-from-source.md
```diff
@@ -5,16 +5,207 @@ title: 004 - Git-tracked TODOs
...
+        return {
+          entityRef: stringifyEntityRef(entity),
+          url: target,
+        };
+      } catch {
+        return undefined;
+      }
+    })
+    .filter((s): s is { entityRef: string; url: string } => Boolean(s));
+}
```
milliehartnt123, line 117: Are these heading twos a series of steps?  Do they need to be done in this order?

docs/golden-path/plugins/backend/004-reading-from-source.md
```diff
@@ -5,16 +5,207 @@ title: 004 - Git-tracked TODOs
...
+      token: ghp_yourPersonalAccessTokenHere
+```
+
+A PAT with `repo` scope is enough for private repositories; public-only setups can omit the token entirely.
+
+We deliberately leave the committed `app-config.yaml` alone here. Production deployments often want a [GitHub App](https://backstage.io/docs/integrations/github/github-apps) instead of a PAT — you'll still ship credentials with the app, but they rotate automatically and aren't tied to a single user — and putting the dev shape in the committed file would push a PAT-shaped config onto every environment.
+
+The same pattern works for `gitlab`, `bitbucketCloud`, `bitbucketServer`, and `azure`. See the [integrations reference](https://backstage.io/docs/integrations/) for the full list.
 
 ## Querying
```
milliehartnt123, line 37: I think the indentation is off here because Querying is not showing as a heading 2.

aramissennyeydd:
closing in favor of a longer lived branch - will figure out how to ship in a bit.

PR #34225, opened 2026-05-13
fix(filter-predicates): reject operator keys mixed with other keys

freben:
## Summary
- A predicate like `{ kind: 'API', $not: { 'spec.type': 'dataset' } }` was silently accepted — the `kind` check was dropped entirely because `evaluateFilterPredicate` short-circuited on the `$not` operator key
- This made the filter match any non-dataset entity regardless of kind; the correct form requires wrapping in `$all`:
  ```json
  { "$all": [{ "kind": "API" }, { "$not": { "spec.type": "dataset" } }] }
  ```
- **Runtime fix**: `evaluateFilterPredicate` now throws `InputError` when `$all`, `$any`, or `$not` are mixed with other keys
- **Schema fix**: Zod v3 and v4 schemas now use `.strict()` on operator object schemas to reject extra keys at parse time
- Added tests for all three operators (`$not`, `$all`, `$any`) in both `evaluate.test.ts` and `schema.test.ts`

## Test plan
- [x] All 183 filter-predicates tests pass
- [x] API reports unchanged (no public API surface change)

### DIFF
--- a/.changeset/fix-filter-predicate-mixed-operator-keys.md
+++ b/.changeset/fix-filter-predicate-mixed-operator-keys.md
@@ -1,0 +1,5 @@
+---
+'@backstage/filter-predicates': patch
+---
+
+Filter predicates that mix operator keys (`$all`, `$any`, `$not`) with other keys are now rejected. Previously, a predicate like `{ kind: 'API', $not: { 'spec.type': 'dataset' } }` would silently drop the `kind` check. The correct form wraps conditions in `$all`.
--- a/packages/filter-predicates/src/predicates/evaluate.test.ts
+++ b/packages/filter-predicates/src/predicates/evaluate.test.ts
@@ -258,6 +258,28 @@ describe('evaluate', () => {
     });
   });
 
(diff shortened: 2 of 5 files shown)


.changeset/fix-filter-predicate-mixed-operator-keys.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/filter-predicates': patch
```
benjdlambert, line 2: `minor`? breaky breaky?

benjdlambert: I mean the behavior was also brokey brokey before, so happy either way, just a quick check what's your thoughts?

freben: Should probably stay patch I think ... even though it may lead to new errors being thrown. This mirrors how typescript refused to let you pass in objects of this form too

benjdlambert (approved):

PR #34252, opened 2026-05-14
permission: cache user info, clean up PolicyQueryUser, drop getPluginRequestToken

freben:
### Problem

The permission backend resolves `userInfo.getUserInfo()` and `auth.getPluginRequestToken()` sequentially for every authorize request with user credentials — even when the same user makes many requests in quick succession. This adds unnecessary latency to every permissioned endpoint.

### Changes

**1. Cache user info lookups** (`@backstage/backend-defaults`)

New `CachedUserInfoService` decorator with a 5-second TTL cache keyed on the raw token. Concurrent calls for the same token coalesce into a single in-flight request. Rejections evict the cache entry so the next call retries. A periodic sweep (every 30s) removes expired entries from tokens that are never requested again. The cache is created once at root level and shared across all plugin instances via `createRootContext`. Note that custom `UserInfoService` implementations registered via their own factory will not benefit from this cache automatically.

**2. Clean up `PolicyQueryUser`** (`@backstage/plugin-permission-node`)

**BREAKING** — removed previously-deprecated fields:

- `token` — **Removed.** Was deprecated in favor of `credentials` with `coreServices.auth`.
- `expiresInSeconds` — **Removed.** Was deprecated.
- `identity` — **Removed.** Was deprecated in favor of `info`.
- `info` — **Deprecated.** Still required and populated for now; will be made optional and then removed in a future release.
- `credentials` — Unchanged.

**3. Drop `getPluginRequestToken` call** (`@backstage/plugin-permission-backend`)

The permission backend no longer calls `auth.getPluginRequestToken()` during policy evaluation (the `token` field it populated was removed). This eliminates one internal round-trip per authorize request.

### Test plan

- [x] Permission-backend router tests pass (25 tests)
- [x] DefaultUserInfoService tests pass (3 tests)
- [x] CachedUserInfoService tests pass (9 tests) — covers caching, TTL expiry, request coalescing, error eviction + retry, shared entries across instances
- [x] Type checker passes
- [x] API reports updated

### DIFF
--- a/packages/backend-defaults/src/entrypoints/userInfo/CachedUserInfoService.ts
+++ b/packages/backend-defaults/src/entrypoints/userInfo/CachedUserInfoService.ts
@@ -58,14 +58,6 @@ export class CachedUserInfoService implements UserInfoService {
 
     const now = Date.now();
 
-    const cached = this.#entries.get(token);
-    if (cached) {
-      if (cached.expiresAt > now) {
-        return cached.promise;
-      }
-      this.#entries.delete(token);
-    }
-
     if (now - this.#lastSweep > SWEEP_INTERVAL_MS) {
       this.#lastSweep = now;
       for (const [key, entry] of this.#entries) {
@@ -73,6 +65,11 @@ export class CachedUserInfoService implements UserInfoService {
           this.#entries.delete(key);
         }
       }
+    }
+
+    const cached = this.#entries.get(token);
+    if (cached && cached.expiresAt > now) {
+      return cached.promise;
     }
 
     const promise = this.#delegate.getUserInfo(credentials).catch(error => {
--- a/packages/backend-defaults/src/entrypoints/userInfo/CachedUserInfoService.ts
+++ b/packages/backend-defaults/src/entrypoints/userInfo/CachedUserInfoService.ts
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2026 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  BackstageCredentials,
+  BackstageUserInfo,
+  UserInfoService,
+} from '@backstage/backend-plugin-api';
+import { toInternalBackstageCredentials } from '../auth/helpers';
+
+const DEFAULT_TTL_MS = 5_000;
+const SWEEP_INTERVAL_MS = 30_000;
+
+export type UserInfoCacheEntry = {
+  promise: Promise<BackstageUserInfo>;
+  expiresAt: number;
+};
+
+export class CachedUserInfoService implements UserInfoService {
+  readonly #delegate: UserInfoService;
+  readonly #entries: Map<string, UserInfoCacheEntry>;
+  readonly #ttlMs: number;
+  #lastSweep: number = Date.now();
+
+  constructor(
+    delegate: UserInfoService,
+    options?: {
+      entries?: Map<string, UserInfoCacheEntry>;
+      ttlMs?: number;
+    },
+  ) {
+    this.#delegate = delegate;
+    this.#entries = options?.entries ?? new Map();
+    this.#ttlMs = options?.ttlMs ?? DEFAULT_TTL_MS;
+  }
+
+  async getUserInfo(
+    credentials: BackstageCredentials,
+  ): Promise<BackstageUserInfo> {
+    const internalCredentials = toInternalBackstageCredentials(credentials);
+    const token = internalCredentials.token;
+    if (!token) {
+      return this.#delegate.getUserInfo(credentials);
+    }
+
+    const now = Date.now();
+
+    const cached = this.#entries.get(token);
+    if (cached) {
+      if (cached.expiresAt > now) {
+        return cached.promise;
+      }
+      this.#entries.delete(token);
+    }
+
+    if (now - this.#lastSweep > SWEEP_INTERVAL_MS) {
+      this.#lastSweep = now;
+      for (const [key, entry] of this.#entries) {
+        if (entry.expiresAt <= now) {
+          this.#entries.delete(key);
+        }
+      }
+    }
+
+    const promise = this.#delegate.getUserInfo(credentials).catch(error => {
+      this.#entries.delete(token);
+      throw error;
+    });
+
+    this.#entries.set(token, {
+      promise,
+      expiresAt: now + this.#ttlMs,
+    });
+
+    return promise;
+  }
+}
--- a/.changeset/permission-backend-drop-token.md
+++ b/.changeset/permission-backend-drop-token.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-permission-backend': patch
+---
+
+The permission backend no longer populates the removed `token` and `identity` fields on `PolicyQueryUser`, and no longer calls `auth.getPluginRequestToken()` during policy evaluation. This removes one internal round-trip per authorize request.
--- a/.changeset/permission-node-optional-deprecated-fields.md
+++ b/.changeset/permission-node-optional-deprecated-fields.md
@@ -1,0 +1,11 @@
+---
+'@backstage/plugin-permission-node': minor
+---
+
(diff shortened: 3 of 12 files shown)


Rugvip (changes requested):
Hold up, very much a breaking change? Would need a deprecation for this first. The existing deprecated fields could be removed if marked as a breaking change

packages/backend-defaults/src/entrypoints/userInfo/CachedUserInfoService.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    if (now - this.#lastSweep > SWEEP_INTERVAL_MS) {
+      this.#lastSweep = now;
+      for (const [key, entry] of this.#entries) {
+        if (entry.expiresAt <= now) {
+          this.#entries.delete(key);
+        }
+      }
+    }
```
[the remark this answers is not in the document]
freben: This doesn't feel worth addressing I think.

packages/backend-defaults/src/entrypoints/userInfo/CachedUserInfoService.ts
```diff
@@ -0,0 +1,90 @@
...
+      return this.#delegate.getUserInfo(credentials);
+    }
+
+    const now = Date.now();
+
+    const cached = this.#entries.get(token);
+    if (cached) {
+      if (cached.expiresAt > now) {
+        return cached.promise;
+      }
```
Rugvip: Could move this block after the sweep and then remove this right?

Rugvip (approved):
Nice 🎉

packages/backend-defaults/src/entrypoints/userInfo/CachedUserInfoService.ts
freben: ah yeah agree, let's do that

packages/backend-defaults/src/entrypoints/userInfo/CachedUserInfoService.ts
```diff
@@ -0,0 +1,87 @@
...
+        if (entry.expiresAt <= now) {
+          this.#entries.delete(key);
+        }
+      }
+    }
+
+    const cached = this.#entries.get(token);
+    if (cached && cached.expiresAt > now) {
+      return cached.promise;
+    }
```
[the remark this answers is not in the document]
freben: Not sure about this one. There would have to be SOME recovery process for things. And the TTL isn't so short that the upstream should reasonably be able to respond much slower than that. Hm. If we were to allow for what is suggested here, there would have to still be a separate time threshold for giving up, that's longer than the regular refresh freq. Either way it'll be no worse than what things were like before - so maybe it's better that this just stays as it currently stands in the PR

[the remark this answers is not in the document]
Rugvip: Yeah, think it's fine to send the second request after expiration

packages/backend-defaults/src/entrypoints/userInfo/CachedUserInfoService.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    const promise = this.#delegate.getUserInfo(credentials).catch(error => {
+      this.#entries.delete(token);
```
Rugvip, line 76: This becomes a bit weird with multiple in-flight requests though, because request A that took 6s might clean out the cache entry for request B start started 5s later. Doesn't have huge impact though

freben: Agreed, will make a followup (no not wait for more builds) with an `if` check that the promise is still the exact same one.

theNailz:
@freben this change has broken the Spotify Backstage RBAC plugin.

RBAC reads the removed PolicyQueryUser.identity field (gone in
@backstage/plugin-permission-node 0.11.0) instead of .info ->
TypeError: Cannot read properties of undefined (reading 'ownershipEntityRefs')
-> HTTP 500 on /api/permission/authorize for any :backstageUser condition

ENVIRONMENT
-----------
@spotify/backstage-plugin-rbac-node ............................ 0.2.17
@spotify/backstage-plugin-rbac-backend ......................... 0.9.14
@spotify/backstage-plugin-permission-backend-module-rbac ....... 0.4.14
@backstage/plugin-permission-node (pinned by the RBAC plugins) . ^0.10.12 -> 0.10.12
@backstage/plugin-permission-node (used by the host backend) ... 0.11.0
@backstage/plugin-permission-backend (the caller) .............. 0.7.12
New backend system.


SUMMARY
-------
Any conditional decision whose params contain the :backstageUser placeholder
- e.g. the IS_ENTITY_OWNER example from your foundational permission policy
tutorial on catalog-entity read - throws a TypeError and returns HTTP 500 on
POST /api/permission/authorize for non-admin users. This cascades into the
Catalog and Search pages failing to load and the Settings page showing
"No Backstage Identity". Admin users covered by an unconditional match: '*'
allow are unaffected, because that decision never resolves the placeholder.


ROOT CAUSE
----------
@spotify/backstage-plugin-rbac-node/dist/resolution.cjs.js resolves the
:backstageUser placeholder from the user's identity field:

    params: mapParams(n, r =>
      Array.isArray(r) && r.includes(BackstageUserPlaceholder)
        ? s?.identity.ownershipEntityRefs ?? []    // reads PolicyQueryUser.identity
        : r)

PolicyQueryUser.identity was REMOVED in @backstage/plugin-permission-node
0.11.0. The host backend builds the user object with that version, so at
runtime the object is { credentials, info } - there is no identity. s.identity
is therefore undefined, and the unguarded .ownershipEntityRefs access throws.
The supported ownership data is present under info.ownershipEntityRefs, which
the plugin does not read.

(The optional chain is also placed at the wrong level: s?.identity.x guards s,
not s.identity. When s is undefined it short-circuits to []; the crash
specifically requires s present and s.identity absent - exactly the 0.11.0
object shape.)

PolicyQueryUser across the two versions:

  0.10.12 (what the RBAC plugins compile against):
    type PolicyQueryUser = {
      token: string;                   // deprecated
      expiresInSeconds?: number;       // deprecated
      identity: BackstageUserIdentity; // deprecated -> "Use the info field instead."
      info: BackstageUserInfo;         // { userEntityRef, ownershipEntityRefs }
    };

  0.11.0 (what the host backend constructs the object with):
    type PolicyQueryUser = {
      info: BackstageUserInfo;         // deprecated, will be removed
    };

The breaking change in 0.11.0 dropped token, expiresInSeconds, and identity.

Upstream reference: backstage/backstage PR #34252 - "permission: cache user
info, clean up PolicyQueryUser, drop getPluginRequestToken"
(https://github.com/backstage/backstage/pull/34252). That PR removed the
already-deprecated token, expiresInSeconds, and identity fields from
PolicyQueryUser and shipped as the @backstage/plugin-permission-node 0.11.0
minor release. Note it also marks the replacement info field as deprecated
("still required and populated for now; will be made optional and then removed
in a future release"), so the long-term migration for the plugin is to resolve
ownership via credentials + the user-info service, not to depend on either
identity or info long-term.

REPRODUCTION
------------
1. Configure a foundational/fallback policy with a conditional read decision on
   catalog-entity that uses IS_ENTITY_OWNER with params.claims: [':backstageUser']
   (verbatim from the foundational-permission-policy tutorial).
2. Run on a backend whose @backstage/plugin-permission-node is 0.11.0
   (current Backstage backends).
3. Sign in as a non-admin user (not covered by an unconditional match: '*' allow).
4. Trigger a conditional catalog read, e.g.
   GET /api/catalog/entity-facets?facet=relations.ownedBy, or open the Catalog page.
5. Observe HTTP 500 on POST /api/permission/authorize with the TypeError above;
   Catalog and Search fail to load.


EXPECTED BEHAVIOUR
------------------
When resolving :backstageUser, the plugin should read the caller's ownership
from the supported info.ownershipEntityRefs and return a normal decision -
never throw.


PROPOSED FIX
------------
Read from the non-deprecated field, with a safe fallback, and bump the
dependency to ^0.11.0:

    ? (s?.info?.ownershipEntityRefs ?? s?.identity?.ownershipEntityRefs ?? [])

  - Prefer info.ownershipEntityRefs (present in both 0.10.x and 0.11.x).
  - Keep the deprecated identity as a fallback for older hosts.
  - Default to [].

A guard-only fix (s?.identity?.ownershipEntityRefs ?? []) stops the crash but
makes IS_ENTITY_OWNER evaluate against an empty ownership set, silently denying
owner-based reads - so reading from info is required for correct behaviour.
Recommend updating the RBAC packages' @backstage/plugin-permission-node
dependency to ^0.11.0 so the plugin builds against the contract its host
actually uses.


IMPACT
------
High for any tenant using owner-based conditions from the documented policy
example: a single internal facets/batch authorize call without identity takes
down Catalog and Search for all non-admin users.


CURRENT WORKAROUND
------------------
Remove the :backstageUser condition from the read decision so the crashing
branch is never reached. delete/update conditions, which are only invoked with
a full user context, can keep it. This avoids the crash at the cost of
owner-based read visibility, so a real upstream fix is still needed.

PR #34257, opened 2026-05-14
permissions: cache permission decisions at the caller side

freben:

> Draft. Companion to #34252 — that PR caches `userInfo` lookups and cleans up `PolicyQueryUser`; this one caches the permission decisions themselves.

### Problem

Every plugin that checks permissions makes an HTTP round-trip to the permission backend (`POST /api/permission/authorize`) for every authorize request. On a catalog page load, the same user triggers many permission checks for the same permission (`catalog.entity.read`) in quick succession. Each one pays the full round-trip cost even though the decision is the same.

From instrumentation, the permission check accounts for 2.6–3.8ms locally and an estimated 60–80ms on production pods — the single biggest latency contributor for fast catalog queries.

### Approach

New `CachedPermissionsService` decorator (`@backstage/backend-defaults`) that wraps `ServerPermissionClient`:

- **5-second TTL cache** keyed on `(method prefix):(permission names):(raw token)`
- **In-flight request coalescing** — concurrent calls for the same key share a single Promise
- **Error eviction** — rejections clear the cache entry so the next call retries
- **Periodic sweep** — expired entries cleaned up every 30 seconds
- **Shared root-level cache** — created via `createRootContext` in `permissionsServiceFactory`, shared across all plugin instances

Both `authorize()` and `authorizeConditional()` are cached, with separate key prefixes so they don't collide.

### Impact

| Scenario | Before | After |
|---|---|---|
| First request for a user+permission | 1 HTTP round-trip | 1 HTTP round-trip (cache miss) |
| Subsequent requests within 5s (same token+permission) | 1 HTTP round-trip each | **0** (cache hit) |
| Concurrent burst (e.g. page load) | N HTTP round-trips | **1** (coalesced) |

### Test plan

- [x] 10 tests covering: delegation, TTL caching, request coalescing, separate tokens, separate permissions, TTL expiry, error eviction + retry, no-token bypass, authorize vs authorizeConditional separation, shared entries across instances
- [x] Type checker passes

### DIFF
--- a/.changeset/cached-permissions-service.md
+++ b/.changeset/cached-permissions-service.md
@@ -1,0 +1,5 @@
+---
+'@backstage/backend-defaults': patch
+---
+
+Added a new `CachedPermissionsService` decorator that wraps the default `ServerPermissionClient` with a 5-second TTL cache and in-flight request coalescing. Permission decisions for the same token and permission set within the TTL window are returned from cache without an HTTP round-trip to the permission backend. The decorator is wired in via `permissionsServiceFactory` using a shared root-level cache. Note that custom `PermissionsService` implementations registered via their own factory will not benefit from this cache automatically.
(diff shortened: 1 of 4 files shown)


freben:
for posterity: some previous reasoning here https://github.com/backstage/backstage/pull/23088

Issue #33575, opened 2026-03-25
RFC: Introduce a new `AIContext` `kind` in the Software Catalog


benjdlambert:
### Need

AI coding agents like Claude Code, GitHub Copilot, and Cursor are rapidly becoming standard developer tooling. These tools rely on contextual files, for example what we commonly call **`rules`** and **`skills`** to encode reusable knowledge \- such as coding patterns, technology choices, framework-specific workflows, and more.

As adoption grows within organizations, the number of context files like skills being created across teams increases quickly. Without any centralized way to manage them, several problems emerge.

**Discoverability.** Skills are scattered across repositories with no central index. Developers have no way to know what skills exist (or even skill marketplaces), which ones are actively maintained, or which are relevant to their discipline or project. The result is that developers either work without context that already exists, spend time searching across repos trying to find something that might help or re-implementing this context.

**Governance.** There is no mechanism to track ownership, lifecycle status, or quality of skills. A deprecated skill looks identical to a production-grade one. Teams cannot enforce standards, review quality, or manage deprecation policies. As the number of skills grows, this lack of oversight becomes a real operational risk, particularly when skills encode architectural decisions or security-sensitive patterns.

**Duplication.** Without visibility into what already exists, teams independently create overlapping skills that cover the same ground. Maintenance burden grows as these duplicates drift out of sync over time. There is no single source of truth, and no way to converge on a canonical version of a skill.

**Consistency across tooling.** Different AI coding agents and coding tools have slightly different conventions for rules and skills formats and discovery. Organizations that use multiple tools, or that are evaluating which tool to standardize on, have no unified way to describe and catalog the contextual knowledge they are building. This creates vendor lock-in at the content layer, even when the underlying knowledge is tool-agnostic.

The Backstage software catalog already solves these exact problems for services, APIs, and infrastructure. It provides a centralized, governed, searchable catalog with ownership tracking, lifecycle management, and entity relationships. Extending the catalog to cover AI context is a natural fit. It brings skills into the same governance and discoverability model that organizations already use for the rest of their software ecosystem, rather than requiring a parallel system to be built and maintained.

### Proposal

Add a new catalog kind, `AIContext`, to represent contextual information consumed by AI coding tools. The initial scope is limited to a few types: **rules** and **skills.**

#### Entity Shape

```yaml
apiVersion: backstage.io/v1alpha1
kind: AIContext
metadata:
  name: frontend-design
  description: Skill for creating production-grade frontend interfaces
spec:
[...]

PR #34261, opened 2026-05-15
feat(catalog): add AiResource entity kind as opt-in module

benjdlambert:
## Summary

Implements the first step of [RFC #33575](https://github.com/backstage/backstage/issues/33575) by introducing a new `AiResource` catalog entity kind.

- Entity types, validators, type guards, and model layer exported from `@backstage/catalog-model/alpha`
- Backend module `@backstage/plugin-catalog-backend-module-ai-model` registers the kind with the catalog
- Uses `backstage.io/v1alpha1` as the apiVersion, consistent with other built-in catalog kinds
- Opt-in: not part of `defaultCatalogEntityModel`, adopters install the module explicitly
- Uses `specType` dispatch for discriminated union schemas per type

### Spec types

**`skill`** — reusable contextual knowledge for AI coding tools
- `disciplines?: string[]`, `categories?: string[]`, `agents?: string[]`, `dependsOn?: string[]`

**`rule`** — governance rules and constraints for AI coding tools
- `disciplines?: string[]`, `category: string` (required), `rationale: string` (required)

**default** — any other type value, base fields only: `type`, `lifecycle`, `owner`, `system?`

### Usage

```ts
// Backend
backend.add(import('@backstage/plugin-catalog-backend-module-ai-model'));

// Types
import {
  type AiResourceEntityV1alpha1,
  isAiResourceEntity,
  isSkillAiResourceEntity,
  isRuleAiResourceEntity,
} from '@backstage/catalog-model/alpha';
```

### Example entities

```yaml
kind: AiResource
metadata:
  name: frontend-design
  description: Skill for creating production-grade frontend interfaces
spec:
  type: skill
  lifecycle: production
  owner: ai-platform-team
  disciplines:
    - web
  categories:
    - framework
  agents:
    - claude-code
```

```yaml
kind: AiResource
metadata:
  name: use-internal-apis
  description: Agents must use internal API clients instead of raw HTTP calls
spec:
  type: rule
  lifecycle: production
  owner: frontend-platform
  disciplines:
    - web
    - backend
  category: architecture
  rationale: Ensures consistent error handling, authentication, and observability across all service calls
```

## Test plan

- [x] Validator tests for default, skill, and rule types
- [x] Type guard tests for all guards
- [x] Backend module registration test
- [x] `yarn tsc`, `yarn lint`, `yarn prettier`, `yarn build:api-reports` all clean
- [x] Example entities ingest successfully via `yarn start`

## Checklist

- [x] Changeset included
- [x] Tests added
- [x] All commits signed off

Relates to #33575.


plugins/catalog-backend-module-ai-resource-entity-model/src/module.ts
```diff
@@ -0,0 +1,43 @@
...
+    reg.registerInit({
+      deps: {
+        model: catalogModelExtensionPoint,
+      },
+      async init({ model }) {
+        model.addModelSource({
+          async *read() {
+            yield { data: [{ layer: aiResourceEntityModel }] };
+          },
+        });
```
freben: this can be

```suggestion
        model.addModelSource(
          CatalogModelSources.static([
            aiResourceEntityModel
          ]),
        );
```

packages/catalog-model/src/kinds/AiResourceEntityV1alpha1.ts
```diff
@@ -0,0 +1,231 @@
...
+  kind: 'AiResource';
+  spec: {
+    type: 'skill';
+    lifecycle: string;
+    owner: string;
+    system?: string;
+    disciplines?: string[];
+    categories?: string[];
+    agents?: string[];
+    dependsOn?: string[];
```
Rugvip, line 65: Wait, how do you get the actual skill? 😅

packages/catalog-model/src/kinds/AiResourceEntityV1alpha1.ts
```diff
@@ -0,0 +1,231 @@
...
+  };
+}
+
+/**
+ * AiResource entity with spec.type 'skill'. Represents reusable contextual
+ * knowledge consumed by AI coding tools.
+ *
+ * @alpha
+ */
+export interface SkillAiResourceEntityV1alpha1 extends Entity {
```
Rugvip: Should these `extend AiResourceEntityV1alpha1Default`?

packages/catalog-model/src/kinds/AiResourceEntityV1alpha1.ts
```diff
@@ -0,0 +1,231 @@
...
+  apiVersion: 'backstage.io/v1alpha1';
+  kind: 'AiResource';
+  spec: {
+    type: 'rule';
+    lifecycle: string;
+    owner: string;
+    system?: string;
+    disciplines?: string[];
+    category: string;
+    rationale: string;
```
Rugvip, line 84: Same thing here, how does one access the concrete rule?

packages/catalog-model/src/kinds/AiResourceEntityV1alpha1.ts
```diff
@@ -0,0 +1,231 @@
...
+
+/**
+ * Type guard for {@link SkillAiResourceEntityV1alpha1}.
+ *
+ * @alpha
+ */
+export const isSkillAiResourceEntity = (
+  entity: Entity,
+): entity is SkillAiResourceEntityV1alpha1 =>
+  isAiResourceEntity(entity) && entity.spec?.type === 'skill';
```
Rugvip, line 143: Hmm, we haven't gone into this detail for the other kinds, feeling a manual `isAiResourceEntity(entity)  && entity.spec.type === 'skill'` might be alright? Otherwise we set a precedent for adding a ton of `isGrpcApiEntity` etc 😅

plugins/catalog-backend-module-ai-model/src/module.ts
```diff
@@ -0,0 +1,44 @@
...
+  moduleId: 'ai-model',
+  register(reg) {
+    reg.registerInit({
+      deps: {
+        model: catalogModelExtensionPoint,
+      },
+      async init({ model }) {
+        model.addModelSource(
+          CatalogModelSources.static([aiResourceEntityModel]),
+        );
```
Rugvip, line 40: 🎉

Rugvip (commented):
Mostly Qs

packages/catalog-model/src/kinds/AiResourceEntityV1alpha1.ts
benjdlambert: It's the source location document. Not currently added to the entity. These entities are designed to be generated from the actual skill definition, parsing the frontmatter, but it explicitly leaves the skill definition out for now, given the fact that there could be accompanying files etc with the skill that don't really fit in the catalog right now either.

Happy to revisit this in the future though.

packages/catalog-model/src/kinds/AiResourceEntityV1alpha1.ts
benjdlambert: See above.

packages/catalog-model/src/kinds/AiResourceEntityV1alpha1.ts
benjdlambert: The provider is also omitted for the first pass of this too, but can probably iterate on that seperately?

Rugvip: Yep, thinking we might want to at least document that expectation? 🤔

plugins/catalog-backend-module-ai-model/README.md
```diff
@@ -36,5 +36,10 @@ spec:
 The `type` field determines which spec fields are available. Currently supported types:
 
 - **`skill`** — reusable contextual knowledge for AI coding tools. Supports additional fields: `disciplines`, `categories`, `agents`, `dependsOn`.
+- **`rule`** — governance rules and constraints for AI coding tools. Supports additional fields: `disciplines`, `category` (required), `rationale` (required).
 
 Any other `type` value is accepted with the base spec fields: `type`, `lifecycle`, `owner`, and optionally `system`.
+
+## Accessing skill and rule content
+
+The actual content of skills and rules is not stored in the entity spec. Instead, the source file is referenced via the standard `backstage.io/source-location` annotation, consistent with how other Backstage entities reference their source files. Entity providers that generate `AiResource` entities from `SKILL.md` or rule files should set this annotation to point to the source file.
```
Rugvip, line 45: So source-location would point to the `SKILL.md`?

Rugvip (approved):

plugins/catalog-backend-module-ai-model/README.md
benjdlambert: Yep that is the way that it's proposed in the current RFC. 

Again happy to revisit this later with some more learnings perhaps.

PR #34265, opened 2026-05-15
github: stabilize entity order in multi-org provider

### LINKED ISSUE #34263: GithubMultiOrgEntityProvider: Teams with identical slugs across orgs overwrite each other
### 📜 Issue Labels

### 🔎 Search Terms

```plain
github multi org team overwrite catalog
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

GitHub

### 📝 Description & Context

When using `GithubMultiOrgEntityProvider `to ingest teams from multiple GitHub organizations, teams that share the same slug across different orgs produce entities with identical `metadata.name` values. Since the `defaultOrganizationTeamTransformer` uses only `item.slug` as the entity name (without any org qualifier), the last-processed org's team overwrites all others.

Because entity providers process orgs in non-deterministic order on each refresh cycle, the managed-by-location annotation and entity content flip randomly every procesing cycle. For example, a team`team-a` existing in both `org1` and `org2` produces a single entity group:default/team-a that alternates between the two orgs' data every processing interval.

### 👍 Expected Behavior

Each org's team should produce a distinct, stable entity. The entity name should include the org to avoid collisions — e.g., `org1-team-a` and `org2-team-a`. The managed-by-location should never flip between orgs across refresh cycles. Another option is to distinguish the groups by using the org name in `metadata.namespace`.

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

1. Configure GithubMultiOrgEntityProvider with 2+ GitHub orgs (e.g., org-a, org-b)
2. Ensure at least one team slug exists in both orgs (e.g., shared-team)
3. Wait for two refresh cycles
4. Observe that group:default/shared-team entity flips its managed-by-location annotation between `url:https://github.com/orgs/org-a/teams/shared-team` and `url:https://github.com/orgs/org-b/teams/shared-team`
5. Only one org's team data survives; the other is silently lost

Workaround:

Provide a custom `TeamTransformer` that prefixes `ctx.org` to the entity name:

```ts
const teamTransformer: TeamTransformer = async (item, ctx) => {
  const entity = await defaultOrganizationTeamTransformer(item, ctx);
  if (!entity) return undefined;
  const orgPrefix = ctx.org ? `${ctx.org}-` : '';
  return {
    ...entity,
    metadata: {
      ...entity.metadata,
      name: `${orgPrefix}${entity.metadata.name}`,
      title: item.name ? `${item.name} (${ctx.org})` : undefined,
    },
  };
};
```

Undecided

freben:
### Problem

When using `GithubMultiOrgEntityProvider` with `alwaysUseDefaultNamespace: true` and teams with identical slugs across orgs, the emitted entity order is non-deterministic (depends on org processing order, which varies each cycle). Since the catalog's upsert path uses last-write-wins for duplicate entity refs, the winning org flips randomly on every refresh cycle. This causes constant unnecessary stitching and flickering entity data.

### Fix

Sort the emitted entities by entity ref (primary) and location annotation (tiebreaker) before applying the full mutation. Non-colliding entities get a stable natural order. Colliding entities (same ref, different orgs) get a deterministic winner — alphabetically first location always wins.

This is a band-aid — the underlying issue is that `alwaysUseDefaultNamespace: true` collapses distinct teams from different orgs into the same entity ref. The proper fix is to either use org-based namespaces (the default behavior) or prefix entity names by org. But this at least eliminates the flickering.

Fixes #34263

### DIFF
--- a/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts
+++ b/plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts
@@ -351,15 +351,32 @@ export class GithubMultiOrgEntityProvider implements EntityProvider {
 
     const { markCommitComplete } = markReadComplete({ allUsers, allTeams });
 
+    const allEntities = [...allUsers, ...allTeams].map(entity => ({
+      locationKey: `github-multi-org-provider:${this.options.id}`,
+      entity: withLocations(
+        `https://${this.options.gitHubConfig.host}`,
+        entity,
+      ),
+    }));
+    allEntities.sort((a, b) => {
+      const am = a.entity.metadata;
+      const bm = b.entity.metadata;
+      if (am.name !== bm.name) return am.name < bm.name ? -1 : 1;
+      const al = am.annotations?.[ANNOTATION_LOCATION] ?? '';
+      const bl = bm.annotations?.[ANNOTATION_LOCATION] ?? '';
+      if (al !== bl) return al < bl ? -1 : 1;
+      if (a.entity.kind !== b.entity.kind) {
+        return a.entity.kind < b.entity.kind ? -1 : 1;
+      }
+      const ans = am.namespace ?? '';
+      const bns = bm.namespace ?? '';
+      if (ans !== bns) return ans < bns ? -1 : 1;
+      return 0;
+    });
+
     await this.connection.applyMutation({
       type: 'full',
-      entities: [...allUsers, ...allTeams].map(entity => ({
-        locationKey: `github-multi-org-provider:${this.options.id}`,
-        entity: withLocations(
-          `https://${this.options.gitHubConfig.host}`,
-          entity,
-        ),
-      })),
+      entities: allEntities,
     });
 
     markCommitComplete();
--- a/.changeset/stable-multi-org-order.md
+++ b/.changeset/stable-multi-org-order.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend-module-github': patch
+---
+
+The `GithubMultiOrgEntityProvider` now emits entities in a stable order during full mutations. Entities are sorted by entity ref, with the location annotation as a tiebreaker for entities that share the same ref. This prevents entity data from flickering between different GitHub orgs across refresh cycles when `alwaysUseDefaultNamespace` is enabled and teams with identical slugs exist in multiple orgs.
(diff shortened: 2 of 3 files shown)


drodil (approved):
Fixes this specific issue, so approving this. Discussion continues whether this should be handled actually on the catalog backend level for all providers, so there would be a possibility to log a warning of duplicate entities and still keep the stable order where the last entity in the list wins.

plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts
```diff
@@ -51,7 +51,7 @@ import {
   TeamEditedEvent,
   TeamEvent,
 } from '@octokit/webhooks-types';
-import { merge } from 'lodash';
+import { merge, sortBy } from 'lodash';
```
deiga: question: lodash isn't the most performant library for sorting. 
Would it make sense to consider (in a separate PR most likely) to migrate to use something like lodash-es, rambda or es-toolkit?

freben: Oh measurably bad? Is there some writeup somewhere or so?

I suggested using this instead of native Array.sort just because it doesn't memoise keys, but that could be another alternative too.

Note that the number of items here isn't extreme; say in the tens or handful of hundreds of thousands range in large orgs.

deiga: There are multiple medium and dev.to posts, none of which I found good enough to share 😬 

I rand a few jsbench.me tests and there native sorting is by far faster than lodash: https://jsbench.me/48mpa7t1h3/1

freben: Alright, updated.

plugins/catalog-backend-module-github/src/providers/GithubMultiOrgEntityProvider.ts
```diff
@@ -351,15 +351,32 @@ export class GithubMultiOrgEntityProvider implements EntityProvider {
...
 
+    const allEntities = [...allUsers, ...allTeams].map(entity => ({
+      locationKey: `github-multi-org-provider:${this.options.id}`,
+      entity: withLocations(
+        `https://${this.options.gitHubConfig.host}`,
+        entity,
+      ),
+    }));
+    allEntities.sort((a, b) => {
+      const am = a.entity.metadata;
```
drodil, line 362: Hm, why not just sort by:
1) stringifyEntityRef
2) LOCATION annotation
3) return 0

freben: I used to. This was a micro optimisation because the repeated string concat and lowercasing just wasn't helping anything :) Maybe I'm silly but ... if there's gonna be like NlogN or N^2 stringifications of hundreds of thousands of things ...

benjdlambert (approved):

PR #34272, opened 2026-05-16
catalog-backend: add missing index on relations.target_entity_ref

freben:
### Problem

The `relations` table has indexes on `originating_entity_id` and `source_entity_ref` but none on `target_entity_ref`. Several query paths join or filter on this column, but the biggest impact is on the **orphan cleanup job** which runs every 30 seconds:

`deleteOrphanedEntities` (called by the processing engine on a 30-second interval) does a `LEFT JOIN relations ON relations.target_entity_ref = orphans.entity_ref` to find entities whose relation targets should also be cleaned up. Without an index, this seq-scans the full relations table (~3.5M rows, 714 MB heap) on every invocation. With up to 100 loop iterations per run, this is a continuous background drain on the database.

The index usage stats from production confirm this: the `relations` table has **2.6M sequential scans reading 728K tuples each** (1.9 trillion tuples read total), with only 203K index scans. The table is almost entirely seq-scanned.

Other affected query paths:
- **Entity ancestry** (`DefaultEntitiesCatalog.ts`): `INNER JOIN relations ON relations.target_entity_ref = ...`
- **Eager pruning** (`deleteWithEagerPruningOfChildren.ts`): `whereIn('relations.target_entity_ref', entityRefs)`

### Measured impact

On a production catalog replica (~3.5M relation rows, 714 MB heap):

| Query | Without index | With index |
|---|---:|---:|
| Point lookup (`WHERE target_entity_ref = ?`) | **122ms** (seq scan, 1.4M rows filtered) | **<1ms** (index scan) |
| Orphan cleanup per iteration | **seq scans 714 MB** | **index lookups only** |

Index creation time: **24 seconds** (tested on staging master with `CREATE INDEX CONCURRENTLY`). Well within pod startup timeout limits.

### Fix

Adds `CREATE INDEX CONCURRENTLY relations_target_entity_ref_idx ON relations (target_entity_ref)`. Estimated index size ~141 MB (same as the existing `source_entity_ref` index — same column type, same cardinality of ~424K distinct values).

The migration is idempotent: it checks for existing/invalid indexes before creating, and handles the interrupted-build case by dropping invalid indexes and retrying.

### Test plan

- [x] All 56 migration tests pass across all database engines (SQLite, MySQL 8, Postgres 14, Postgres 18)
- [x] Type checker passes
- [x] Index creation timed at 24 seconds on staging master

### DIFF
--- a/.changeset/relations-target-index.md
+++ b/.changeset/relations-target-index.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Added a missing index on `relations.target_entity_ref`. Several query paths (orphan deletion, entity ancestry, eager pruning) join or filter on this column, but no index existed — causing full sequential scans of the relations table on every invocation. On a production catalog with ~3.5M relation rows, individual lookups were taking ~122ms (full table scan) instead of <1ms (index scan).
--- a/plugins/catalog-backend/migrations/20260516000000_relations_target_index.js
+++ b/plugins/catalog-backend/migrations/20260516000000_relations_target_index.js
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2026 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


freben:
### Index creation timing

Tested on the staging master database (production-scale data: ~3.5M rows, 714 MB heap):

```
CREATE INDEX CONCURRENTLY relations_target_entity_ref_idx ON relations (target_entity_ref);
-- 23.8 seconds
```

Well within pod startup timeout limits. No risk of the liveness probe issue that affects the larger `search` table index builds.

benjdlambert (approved):

PR #34277, opened 2026-05-18
feat(techdocs): add TechDocs generator pullOptions support for private registry image pull

KarimSeoudy:

Adds support for passing `techdocs.generator.pullOptions` from `app-config.yaml` to the TechDocs Docker image pull.

This makes it possible to authenticate against private registries when using a custom TechDocs generator image with `techdocs.generator.runIn: docker`, `dockerImage: <custom-image>`, and `pullImage: true` (or unset).

### What changed
- Added support for `techdocs.generator.pullOptions` in config
- Passed `pullOptions` through to `dockerode` `dockerClient.pull`
- Updated TechDocs backend config schema to reflect the new option in config
- Updated docs, added config examples for private registry authentication
- Added tests covering config parsing and Docker pull behavior

### Example
```yaml
techdocs:
  generator:
    runIn: 'docker'
    dockerImage: 'custom-registry/techdocs'
    pullImage: true
    pullOptions:
      authconfig:
        username: ${REGISTRY_USERNAME}
        password: ${REGISTRY_PASSWORD}
```

### DIFF
--- a/.changeset/techdocs-docker-private-registry.md
+++ b/.changeset/techdocs-docker-private-registry.md
@@ -1,0 +1,6 @@
+---
+'@backstage/plugin-techdocs-backend': minor
+'@backstage/plugin-techdocs-node': minor
+---
+
+Added support for `techdocs.generator.pullOptions` when pulling the TechDocs generator Docker image from private registries that require authentication.
--- a/plugins/techdocs-backend/config.d.ts
+++ b/plugins/techdocs-backend/config.d.ts
@@ -16,6 +16,8 @@
 
 import { HumanDuration } from '@backstage/types';
 
+import { ContainerRunnerPullOptions } from '@backstage/plugin-techdocs-node';
+
 export interface Config {
   /**
    * Configuration options for the techdocs-backend plugin
@@ -46,6 +48,12 @@ export interface Config {
        * Pull the latest docker image
        */
       pullImage?: boolean;
+
+      /**
+       * Credentials for private registries
+       * @visibility secret
+       */
+      pullOptions?: ContainerRunnerPullOptions;
 
       /**
        * Override behavior specific to mkdocs.
--- a/plugins/techdocs-backend/config.d.ts
+++ b/plugins/techdocs-backend/config.d.ts
@@ -16,6 +16,8 @@
 
 import { HumanDuration } from '@backstage/types';
 
+import { ContainerRunnerPullOptions } from '@backstage/plugin-techdocs-node';
+
 export interface Config {
   /**
    * Configuration options for the techdocs-backend plugin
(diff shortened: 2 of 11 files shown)


KarimSeoudy:
It would be nice if someone can take a look

plugins/techdocs-backend/config.d.ts
```diff
@@ -47,6 +49,11 @@ export interface Config {
        */
       pullImage?: boolean;
 
+      /**
+       * Credentials for private registries
+       */
+      pullOptions?: ContainerRunnerPullOptions;
```
awanlin, line 56: I'm worried this will leak secretes like the password, I think the types will need to be inline and not using the imported types for this to work. Related docs: https://backstage.io/docs/conf/defining#visibility

.changeset/techdocs-docker-private-registry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@backstage/plugin-techdocs-backend': minor
+'@backstage/plugin-techdocs-node': minor
+---
+
+Added support for `techdocs.generator.pullOptions` when pulling the TechDocs generator Docker image from private registries that require authentication.
```
awanlin, line 6: TechDocs get generated in two ways: local build and using the CLI. I don't think this fully covers the CLI pattern where this is more likely to be used. Is that a wrong assessment or is the plan to contribute a follow up for that?

awanlin (changes requested):
Thanks for the contribution @KarimSeoudy, can you please rebase, that should get the CI passing. The failing CI might be way this has been overlooked, we won't review PRs with failing builds but given the amount of time it's hard to say it was always failing or not. 

Over all this looks good, left two comments that need to be addressed/answered. 👍

KarimSeoudy:
Reopen this PR, i'm planning to resolve comments and rebase

.changeset/techdocs-docker-private-registry.md
KarimSeoudy: this covers only docker generator, see the change [here](https://github.com/backstage/backstage/pull/34277/changes#diff-f21421000cf30c2e5550c84c1eb44c491b73158077bd11fbc99732877d7e8f1bR195)

```ts
        case 'docker': {
          const containerRunner =
            this.containerRunner || new DockerContainerRunner();
          await containerRunner.runContainer({
            imageName:
              this.options.dockerImage ?? TechdocsGenerator.defaultDockerImage,
            args: ['build', '-d', '/output'],
            logStream,
            mountDirs,
            workingDir: '/input',
            // Set the home directory inside the container as something that applications can
            // write to, otherwise they will just fail trying to write to /
            envVars: { HOME: '/tmp' },
            pullImage: this.options.pullImage,
            pullOptions: this.options.pullOptions,
            defaultUser: runAsDefaultUser,
          });
          childLogger.info(
            `Successfully generated docs from ${inputDir} into ${outputDir} using techdocs-container`,
          );
          break;
        }
```

plugins/techdocs-backend/config.d.ts
KarimSeoudy: great catch, i set the property visibility to secret since i want to use the type

awanlin (approved):
Thanks for the follow up and sorry I didn't review this sooner @KarimSeoudy 🙏

PR #34282, opened 2026-05-18
catalog-model: add addKindVersion and updateKindVersion to builder

freben:
### Problem

`CatalogModelLayerBuilder` had no way to add a single new version to an existing kind without re-declaring the kind itself. The `updateKind` method accepted a `versions` array, but it could only modify versions that already existed — it couldn't add new ones.

### Changes

**New methods on `CatalogModelLayerBuilder`:**

- **`addKindVersion`** — declares a new (version, specType) slot on an existing kind. Creates the version name if it doesn't exist yet. Fails if the (version, specType) combination is already declared.

- **`updateKindVersion`** — modifies an existing (version, specType) slot (description, relation fields, schema merge). Strict: both the version name and the specType must already exist.

**Deprecation:**

The `versions` array on `updateKind` is deprecated in favor of `addKindVersion` and `updateKindVersion`. It continues to work for backward compatibility.

### Test coverage

- Unit tests for `opsFromCatalogModelKindVersion` (4 tests): single version, multiple names, multiple spec types, relation fields
- Unit tests for `opsFromCatalogModelUpdateKindVersion` (4 tests): single version, multiple names, multiple spec types, schema-less update
- Compiler flow tests (3 tests):
  - Adding a new version (`v1beta1`) to an existing kind, validating the new schema
  - Updating an existing version with a schema merge, verifying both original and merged fields
  - Adding a new spec type variant (`fancy`) alongside an existing one (`basic`)

### DIFF
--- a/.changeset/catalog-model-kind-version-methods.md
+++ b/.changeset/catalog-model-kind-version-methods.md
@@ -1,0 +1,5 @@
+---
+'@backstage/catalog-model': minor
+---
+
+Added `addKindVersion` and `updateKindVersion` methods to `CatalogModelLayerBuilder`, allowing plugins to declare new versions or modify existing versions of a kind without having to re-declare the kind itself. The `versions` array on `updateKind` is now deprecated in favor of these dedicated methods.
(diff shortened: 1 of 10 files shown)


freben:
Closing in favor of #34016 by @benjdlambert which adds the same `addKindVersion` capability along with the API v1alpha2 feature that motivated it. We can follow up with `updateKindVersion` and the deprecation of `updateKind.versions` on top of that PR once it lands.

PR #34284, opened 2026-05-18
fix(catalog-backend): move generateStableHash out of shared util to fix Storybook build

benjdlambert:
## Summary

- Moved `generateStableHash` (and its `node:crypto` dependency) from `util.ts` into `performStitching.ts`, its only consumer
- `util.ts` now only contains pure constants (`BATCH_SIZE`, `NULL_SENTINEL`) with no Node.js dependencies

The Storybook build was failing because `InMemoryCatalogClient` in `catalog-client` reaches into `catalog-backend`'s `buildEntitySearch.ts` via a relative cross-package import. That file imports `NULL_SENTINEL` from `util.ts`, which also contained `generateStableHash` using `createHash` from `node:crypto`. Vite couldn't resolve `node:crypto` for the browser bundle, causing the build to fail.

## Test plan

- [x] `yarn build-storybook` completes successfully
- [x] `yarn tsc` passes
- [x] All 94 stitcher tests pass (`CI=1 yarn test plugins/catalog-backend/src/database/operations/stitcher/`)

### DIFF
--- a/.changeset/fix-storybook-crypto-import.md
+++ b/.changeset/fix-storybook-crypto-import.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
+Moved `generateStableHash` out of shared utility file to avoid pulling `node:crypto` into browser bundles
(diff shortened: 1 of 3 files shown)


Rugvip (approved):
Ah 👍

freben (approved):

awanlin (approved):
Nice! 🚢

PR #34292, opened 2026-05-18
Scaffolder - Added support for `when` conditions on parameter step entries

awanlin:

Added support for `when` conditions on parameter step entries. This allows template authors to conditionally show or hide wizard steps based on answers from earlier steps using the `${{ parameters.fieldName === 'value' }}` expression syntax.

The `when` keyword is used instead of `if` to avoid colliding with JSON Schema's `if`/`then`/`else` keywords, allowing both to be used freely within the same parameter step.


awanlin:
Hi @acierto, any chance you can give this a look? I've had some conversation with @benjdlambert a few weeks ago when I changes this from `if` to `when`. I'd really like to get this into the September release and ideally a few `next` release before then. 🤞

plugins/scaffolder-backend/src/service/router.ts
```diff
@@ -210,6 +211,95 @@ function formatSecretsValidationErrors(result: ValidatorResult) {
   });
 }
 
+const conditionRenderer = createTemplateRenderer();
+
+// The frontend wizard uses a lighter evaluator from @backstage/plugin-scaffolder-common
```
benjdlambert, line 216: Hmm this is unfortunate.

@rugvip what's the limitation here? Will there be a time that this is possible?

benjdlambert (commented):
Hmm it's unfortunate that we have to duplicate the templating implementation, I wonder if we can try and unblock that.

plugins/scaffolder-react/src/next/components/Stepper/Stepper.tsx
```diff
@@ -266,10 +337,10 @@ export const Stepper = (stepperProps: StepperProps) => {
         variant="elevation"
         style={{ overflowX: 'auto' }}
       >
-        {steps.map((step, index) => {
+        {visibleSteps.map((step, index) => {
           const isAllowedLabelClick = activeStep > index;
           return (
-            <MuiStep key={index}>
+            <MuiStep key={step.title}>
```
acierto, line 343: still keyed by title - router.ts defaults untitled steps to the same string, so two hidden/shown untitled steps will duplicate-key here

acierto (changes requested):
+1 on @benjdlambert's point re: duplicating the condition evaluator front/back - would like to see that resolved before merge. One more blocking issue inline.

PR #34296, opened 2026-05-19
fix(create-app): remove disabled nav-item config that breaks custom sidebar

benjdlambert:
The `nav-item` extensions were disabled in the template config to prevent duplicate rendering, but the custom sidebar already handles this via `nav.take()`. After #33788 added filtering of disabled nav items from page discovery, disabling them causes `nav.take('page:catalog')` to return nothing, breaking the sidebar navigation.


Rugvip (approved):
Needs more fixing I think, because we don't want this crash to happen in the first place imo. We'd want this fix in either way though 👍

PR #34316, opened 2026-05-19
fix: correct postgres env var casing in deploy_packages workflow

freben:

The `deploy_packages` workflow used lowercase env var names (`BACKSTAGE_TEST_DATABASE_postgres18_CONNECTION_STRING` / `postgres14`) while `TestDatabases` expects uppercase (`POSTGRES18` / `POSTGRES14`). This meant postgres was never available in that workflow's test runs, causing 4 postgres-only test suites to crash with `describe.each` receiving an empty array.

The vars were originally misspelled as `postrges` (commit 1368b85d3acf fixed the spelling but kept lowercase). The CI workflow (`ci.yml`) has had the correct uppercase casing all along, which is why PR checks passed but the deploy workflow failed.

### DIFF
--- a/packages/backend-test-utils/src/cache/TestCaches.ts
+++ b/packages/backend-test-utils/src/cache/TestCaches.ts
@@ -108,9 +108,6 @@ export class TestCaches {
   }
 
   eachSupportedId(): [TestCacheId][] {
-    if (this.supportedIds.length === 0) {
-      return [['MISSING' as TestCacheId]];
-    }
     return this.supportedIds.map(id => [id]);
   }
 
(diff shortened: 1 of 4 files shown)


packages/backend-test-utils/src/cache/TestCaches.ts
```diff
@@ -108,6 +108,9 @@ export class TestCaches {
   }
 
   eachSupportedId(): [TestCacheId][] {
+    if (this.supportedIds.length === 0) {
+      return [['MISSING' as TestCacheId]];
+    }
```
Rugvip: Hmm, don't think we should do this tbh? seems correct to just fix the env vars. Maybe we can throw a more informative error message though

Rugvip (changes requested):

Rugvip (approved):
:shipit:

packages/backend-test-utils/src/cache/TestCaches.ts
freben: Yeah i marked it as draft because the agent just did stupid things without supervision. Check now again

PR #34317, opened 2026-05-19
fix(catalog): fix flaky entity page test

freben:
## Summary

The `pages.test.tsx` entity page test was flaky because `renderInTestApp` never set up React Router `<Route>` matching for the test element. `useParams()` always returned `{}`, so `useEntityFromUrl` got no route params and posted a "No name provided!" error. The test passed or failed depending on a race between the error effect and the tab render.

### Fix

**`renderInTestApp`**: Add a `mountPath` option that wraps the test element in a `<Route>` with the given path pattern, enabling `useParams()` to extract parameters from the URL. This mirrors what the real `AppRoutes` extension does via `useRoutes()`. Use together with `initialRouteEntries` to set a concrete URL matching the pattern.

**`pages.test.tsx`**: Use `mountPath` + `initialRouteEntries` + the in-memory `catalogApiMock({ entities: [...] })` so the test does real end-to-end routing: URL → params → entity lookup. Href expectations updated to match production behavior (sub-route links resolve relative to the matched parent Route).

### Test plan

- [x] All 20 entity page tests pass
- [x] All `renderInTestApp` tests pass
- [x] All `convertLegacyPageExtension` and `catalog-react/alpha/converters` tests pass
- [x] API report updated

### DIFF
--- a/packages/frontend-test-utils/src/app/renderInTestApp.tsx
+++ b/packages/frontend-test-utils/src/app/renderInTestApp.tsx
@@ -15,7 +15,7 @@
  */
 
 import { Fragment } from 'react';
-import { MemoryRouter } from 'react-router-dom';
+import { MemoryRouter, Route, Routes } from 'react-router-dom';
 import { prepareSpecializedApp } from '@backstage/frontend-app-api';
 import { RenderResult, render } from '@testing-library/react';
 import { ConfigReader } from '@backstage/config';
@@ -125,11 +125,31 @@ export function renderInTestApp<const TApiPairs extends any[] = any[]>(
   element: JSX.Element,
   options?: TestAppOptions<TApiPairs>,
 ): RenderResult {
+  const mountedPaths = options?.mountedRoutes
+    ? Object.keys(options.mountedRoutes)
+    : [];
+
   const extensions: Array<ExtensionDefinition> = [
     createExtension({
       attachTo: { id: 'app/root', input: 'children' },
       output: [coreExtensionData.reactElement],
       factory: () => {
+        if (mountedPaths.length > 0) {
+          return [
+            coreExtensionData.reactElement(
+              <Routes>
+                {mountedPaths.map(path => (
+                  <Route
+                    key={path}
+                    path={path === '/' ? path : `${path.replace(/\/$/, '')}/*`}
+                    element={element}
+                  />
+                ))}
+                <Route path="*" element={element} />
+              </Routes>,
+            ),
+          ];
+        }
         return [coreExtensionData.reactElement(element)];
       },
     }),
--- a/packages/frontend-test-utils/src/app/renderInTestApp.tsx
+++ b/packages/frontend-test-utils/src/app/renderInTestApp.tsx
@@ -135,6 +135,7 @@ export function renderInTestApp<const TApiPairs extends any[] = any[]>(
       output: [coreExtensionData.reactElement],
       factory: () => {
         let content: JSX.Element = element;
+
         if (mountedPaths.length > 0) {
           content = (
             <Routes>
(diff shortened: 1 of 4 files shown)


packages/frontend-test-utils/src/app/renderInTestApp.tsx
```diff
@@ -125,12 +125,33 @@ export function renderInTestApp<const TApiPairs extends any[] = any[]>(
...
+
+        if (mountedPaths.length > 0) {
+          content = (
+            <Routes>
+              {mountedPaths.map(path => (
+                <Route
+                  key={path}
+                  path={path === '/' ? path : `${path.replace(/\/$/, '')}/*`}
+                  element={content}
+                />
```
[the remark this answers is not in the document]
freben: heh, maybe a matter of taste, i asked it to do this way

packages/frontend-test-utils/src/app/renderInTestApp.tsx
```diff
@@ -125,12 +125,33 @@ export function renderInTestApp<const TApiPairs extends any[] = any[]>(
...
+        let content: JSX.Element = element;
+
+        if (mountedPaths.length > 0) {
+          content = (
+            <Routes>
+              {mountedPaths.map(path => (
+                <Route
+                  key={path}
+                  path={path === '/' ? path : `${path.replace(/\/$/, '')}/*`}
+                  element={content}
```
Rugvip: Hmm, I don't think it's right to mount the content on each of these paths. They're not meant for the test subject, but rather abstract route targets that the subject would be linking to.

I'm thinking we might simply need to add a `mountPath?: string` option? 🤔, which ends up being the path that the content is rendered at, with `/` being the default. We'd also have `initialRouteEntries` default to `[mountPath]`. Dunno about the exact naming, just the first thing that came to mind 🤷

Would be nice to avoid new API, but not really seeing any other way around it that isn't too magical.

Rugvip (changes requested):

packages/frontend-test-utils/src/app/renderInTestApp.tsx
freben: Alright, fixed. Still a bit unfortunate that there's a need to balance both `mountPath` (for the pattern) and `initialRouteEntries` in parameterised cases; the api is not super convenient. But it works now I think.

packages/frontend-test-utils/src/app/renderInTestApp.tsx
```diff
@@ -80,7 +80,28 @@ export type TestAppOptions<TApiPairs extends any[] = any[]> = {
...
   /**
-   * Initial route entries to use for the router.
+   * The route path pattern that the test element is rendered at. When set,
+   * the element is wrapped in a `<Route>` with this path, enabling
+   * `useParams()` to extract parameters from the URL. Defaults to `'/'`.
+   *
+   * When `mountPath` is set and `initialRouteEntries` is not, the
+   * initial route entry defaults to `mountPath` (which works for paths
+   * without parameters). For parameterized paths you must also set
+   * `initialRouteEntries` to a concrete URL that matches the pattern.
```
Rugvip: Not seeing this being enforced. What happens if not set? Does the router complain about `:derp` entries or do you actually just end up with `":derp"` params?

Rugvip (approved):
👍, can't think of any more fun names or find any other precedent, I think this is safe enough to ship. Just a minor Q on the behavior

packages/frontend-test-utils/src/app/renderInTestApp.tsx
freben: Mm that's a good point, it would actually set the location history to be the parameterised path string. You'd get something like `name=":name"` literally like you say and that's not very cute. Alright I will update it to stop with the defaulting behavior, that's less magic than like trying to invent some valid entry or so. Yes it's still a bit "loose" (you are EXPECTED to use the parameters together in a wise way) but asserting that there's a matching location path doesn't for example allow for fewer silent errors or so - you'll immediately note it doesn't render anything. Let me know if you feel it's too loose and I can fix it up

PR #34323, opened 2026-05-20
fix(catalog-backend): split queryEntities list and count queries

freben:
## Summary

- Split the `queryEntities` list and count into two separate queries run via `Promise.all`, instead of the current multi-reference CTE that forces full materialization before `LIMIT` can short-circuit.

This is a minimal cherry-pick of the core performance fix from #34214, isolated so it can be merged and deployed as a patch without the larger API surface changes in that PR.

### Problem

The current query wraps the filtered set in a CTE (`filtered`) that is referenced **twice** — once by `filtered_count` and once by the outer `SELECT`. PostgreSQL 12+ refuses to inline a multi-referenced CTE, so it materializes the entire filtered set (e.g. all ~46K components), spills to temp, and only *then* applies `LIMIT 21`. The covering index can't help — the query shape itself prevents short-circuiting.

### Fix

Always reference the `filtered` CTE exactly once (for the list query). Run the count as a standalone query via `Promise.all`. Since the CTE is now single-referenced, Postgres inlines it and the planner can drive directly from the index with `LIMIT` short-circuit.

### Evidence

_All timings from `EXPLAIN ANALYZE` on a staging replica (~462K entities, ~13.2M search rows)._

| Scenario | Before (fused CTE) | After: list query | After: count query | After: wall clock (`Promise.all`) |
|---|---:|---:|---:|---:|
| `kind=component`, page 1, LIMIT 21 | **2,660 ms** | **1.6 ms** | 1,048 ms | **~1,048 ms** |
| No filter, page 1, LIMIT 21 | **7,715 ms** | **0.5 ms** | 5,471 ms | **~5,471 ms** |
| `kind=template` (narrow), page 1 | 1.1 ms | 0.6 ms | 0.4 ms | ~0.6 ms |
| Cursor page 2+ (`kind=component`) | 23 ms | 23 ms | n/a (skipped) | ~23 ms |

**Key takeaways:**

- **The common first-page load (`kind=component`)** drops from **2.7s to ~1s** wall clock — the list query returns in 1.6 ms instead of waiting for 2.7s of CTE materialization. The wall clock is bounded by the count (1s), but row data is available almost immediately.
- **No-filter first page** drops from **7.7s to ~5.5s** — the list returns in 0.5 ms. The count is expensive (5.5s scanning all entities) but runs concurrently with the now-instant list.
- **Narrow filters and cursor pages** are already fast and stay fast — no regression.
- A follow-up (`totalItems: 'exclude'` from #34214) can skip the count entirely, bringing page 1 to **<2 ms** for all filter shapes.

## Test plan

- [x] All 172 `DefaultEntitiesCatalog` tests pass
- [x] Type check passes (no new errors)
- [x] `EXPLAIN ANALYZE` verified on staging replica (timings above)

### DIFF
--- a/.changeset/split-query-entities-count.md
+++ b/.changeset/split-query-entities-count.md
@@ -3,3 +3,5 @@
 ---
 
 Split the `queryEntities` list and count into separate queries instead of a multi-reference CTE. When the `filtered` CTE was referenced twice (once for the count, once for the data), PostgreSQL refused to inline it, forcing full materialization of the filtered set before applying `LIMIT`. By running the count as a standalone query, the list CTE is only referenced once, allowing the planner to short-circuit on `LIMIT` and return the first page in milliseconds instead of waiting for the full filtered set to materialize.
+
+The standalone count query also fixes a pre-existing bug where `totalItems` was inflated for entities with multi-valued sort fields (e.g. tags). The old CTE-based count counted search rows, so an entity with 3 tags would be counted 3 times. The new count uses `EXISTS` to count distinct entities, aligning `totalItems` with the number of entities actually reachable through cursor pagination.
--- a/.changeset/split-query-entities-count.md
+++ b/.changeset/split-query-entities-count.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': patch
+---
+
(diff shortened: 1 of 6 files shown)


freben (commented):
Copilot raised two related points about the count query semantics change — both are intentional:

The old CTE-based count used `COUNT(*)` over search rows, which **inflated** `totalItems` for entities with multi-valued sort fields (e.g. tags). An entity with 3 tags would be counted 3 times. The new EXISTS-based count correctly counts distinct entities, aligning `totalItems` with the number of entities actually reachable through cursor pagination.

Updated the changeset to call this out.

benjdlambert (approved):

PR #34328, opened 2026-05-21
BEP: Workspaces and Reworked Releases

Rugvip:

Adds BEP 0015, which proposes restructuring `backstage/backstage` into per-area workspaces (modeled on `backstage/community-plugins`), introduces an opt-in "staged change" mechanism for breaking changes that can live in `main` and preview through `@next` without holding back `@latest`, and reworks how the `framework` workspace is released and versioned (a date-based `YYNN` release identifier, two releases a year around the KubeCon events, with per-package semver kept independent of the identifier).

The proposal also covers:

- A new `tooling` workspace for release automation that runs dependency-free in-repo and ships a standalone CLI to other adopters (community-plugins and anyone else picking up the same structure).
- A Backstage release manifest hosted in `backstage/versions` that adopters pin via the Backstage Yarn plugin, with floating and frozen pin shapes.
- A folded-in `demo` workspace replacing the legacy example app/backend at the repo root.
- A manifest-driven docs build that pulls each workspace's docs from its latest published commit, with a scheduled rebuild that short-circuits cheaply via git tree-hash fingerprinting.
- A staged migration plan that takes the new structure out one workspace at a time, with `framework` last and `demo` folded in alongside it.

Lots of discussion happened to land on this shape — looking for feedback on the overall direction and on any specific subsection. Happy to iterate.


beps/0015-workspace-release-restructure/README.md
```diff
@@ -0,0 +1,1855 @@
...
+  # once the referenced entry has been merged into main.
+  staged:
+    - framework/20260301-091500-remove-config-mode-flag
+    - auth/20260415-152230-rotate-token-format
+---
+
+Removed the deprecated `EntityRefLink` props `defaultKind` and `defaultNamespace`.
+Pass these as part of the `entityRef` instead.
+
+​```patch
```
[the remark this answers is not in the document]
Rugvip: [automated] Fixed in 849f05a73d8c — removed the U+200B characters before both the opening `​```patch` and the closing `​```` fences in the nested example.

beps/0015-workspace-release-restructure/README.md
```diff
@@ -0,0 +1,1855 @@
...
+diff --git a/plugins/catalog-react/src/components/EntityRefLink/EntityRefLink.tsx b/plugins/catalog-react/src/components/EntityRefLink/EntityRefLink.tsx
+--- a/plugins/catalog-react/src/components/EntityRefLink/EntityRefLink.tsx
++++ b/plugins/catalog-react/src/components/EntityRefLink/EntityRefLink.tsx
+@@ ... @@
+
+- defaultKind, defaultNamespace,
+- ...
+  ​```
```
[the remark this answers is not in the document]
Rugvip: [automated] Fixed in 849f05a73d8c alongside the opening fence — both U+200B characters are gone.

beps/0015-workspace-release-restructure/README.md
```diff
@@ -0,0 +1,1855 @@
...
+
+3. **Updating an existing staged change.** When a PR conflicts with a queued staged
+   change, CI fails with a pointer to the failing entry. The author runs
+   `yarn release stage refresh <slug>`, which re-runs the
+   apply/edit/capture loop and produces an updated staged change. The PR is required
+   to include the refreshed entry.
+
+4. **Promoting staged changes to a major.** No author action required. The
+   bot-maintained `Promote staged (<workspace>)` PR already contains the result of
+   applying every queued staged change (see
```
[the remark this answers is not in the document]
Rugvip: [automated] Good catch. The canonical format is the single `.staged/<slug>.md` file described in "Staged change file format" — the references to `description.md` and `change.patch` here and in four other spots (Promoting staged changes, the `promote-pr` and `dispatch-next` job descriptions, and the `@next` base-version section) were leftovers from an earlier draft. Reconciled in 42abc972ef64 to consistently describe synthesizing a changeset from each entry's front-matter and markdown body.

beps/0015-workspace-release-restructure/README.md
```diff
@@ -0,0 +1,1855 @@
...
+## Release Plan
+
+The migration is staged so that no single PR has to move the entire repository.
+
+1. **BEP approval & tool scaffolding.** Land the BEP, then carve out the new
+   `workspaces/tooling/` workspace (see
+   [Repository tooling](#repository-tooling)) and bootstrap the standalone
+   release-automation CLI inside it with the existing community-plugins commands in
+   their current form. Vendor the new `stage` and `fix` commands behind a flag while
+   the format stabilizes.
```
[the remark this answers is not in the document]
Rugvip: [automated] Correct — there is no `fix` subcommand; patch releases reuse the existing branch-based flow. Aligned the wording with the canonical `stage create|refresh|apply` subcommands from "Tooling consolidation" in 6a2cd47d848a.

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1859 @@
...
+[Framework packages keep independent semver](#framework-packages-keep-independent-semver)
+below.
+
+#### Expected cadence
+
+The project targets two framework releases per year, in April and October. Both
+windows are timed to coincide with the KubeCon events: KubeCon EU lands in April
+most years (occasionally in March), and KubeCon NA lands in November. Under the
+`YYNN` scheme that produces `<YY>04` and `<YY>10` as the cadence slot numbers —
+for example `2604` and `2610` in 2026, then `2704` and `2710` in 2027, and so on.
```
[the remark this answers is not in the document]
Rugvip: [automated] Good catch — the "coincide with KubeCon" framing was sloppy. KubeCon NA is in November, so an October release is *near* it, not coincident with it. Reworded in 04dba025be2a to say the windows are picked so each release lands shortly before a KubeCon, with the most recent named release fresh when people meet there.

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1859 @@
...
+
+There is no fixed calendar for when framework releases ship. The trigger is the same
+as for every other workspace: a maintainer reviewing and merging the
+`Promote staged (framework)` PR (see
+[Promoting staged changes](#promoting-staged-changes)). The
+calendar-driven behavior is in the identifier itself: `YY` is the current two-digit
+year and `NN` is whichever is larger of the current calendar month and one greater
+than the highest `NN` shipped so far this year. The counter restarts at the new
+year. The CLI computes the next identifier deterministically from the current date
+and the highest `NN` already published in the year.
```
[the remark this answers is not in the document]
Rugvip: [automated] Agreed — the promotion-trigger paragraph was overriding the cadence-slot semantics defined a few sections above. The slot is meant to track the planned cadence, not the ship month: a planned April release that slips to May still ships as `2604`. Reconciled in c569e557f9fe by deferring to the Identifier format rules instead of restating a different rule here, and calling out the slip case explicitly.

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1859 @@
...
+staged changes change.
+
+```
+on push to main:
+
+  find-changed-workspaces
+        │
+        ├── matrix per workspace
+        │
+        └── release-workspace.yml
```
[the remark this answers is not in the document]
Rugvip: [automated] Fixed in ca2040a536a6 — standardized on the hyphenated `release-workspace.yml` everywhere; the underscore variant on line 973 was the outlier.

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1865 @@
...
+
+#### How versions evolve through a release cycle
+
+A framework release like `2604` is the named state of the workspace at a moment in
+time. Between one framework release and the next, the workspace continues to ship
+non-breaking changes through the Version Packages PR — patches and minors per the
+usual semver rules:
+
+- **Patch** bumps when a release contains only bug fixes. Example:
+  `core-plugin-api` goes from `2.4.7` to `2.4.8`.
```
freben: Could use the NFS packages as examples to keep them long term viable. The `core-` series appears in several places throughout the document.

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1865 @@
...
+## Summary
+
+This BEP proposes restructuring the `backstage/backstage` repository from a single
+Yarn workspace into a set of independent workspaces under a top-level `workspaces/`
+directory, modeled on `backstage/community-plugins`. Each workspace owns its own
+packages, changesets, version lines, and release cadence, so a bug fix in one area
+no longer drags the rest of the repository through a release.
+
+It also introduces an opt-in "staged change" mechanism: a breaking (or otherwise
+held-back) change is committed to `main` as a description plus a patch file under
```
freben: The BEP keeps saying `main` rather than `master`. Is that meant "symbolically" or did you intend for a default branch change too? :)

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1865 @@
...
+      removal, then captures the diff and prompts for a description, writing both
+      into a single `.staged/<timestamp>-<slug>.md` file with the description as
+      the markdown body and the diff as a trailing `patch` fenced code block.
+   4. Commits. CI verifies the staged change applies cleanly on top of `main`.
+
+3. **Updating an existing staged change.** When a PR conflicts with a queued staged
+   change, CI fails with a pointer to the failing entry. The author runs
+   `yarn release stage refresh <slug>`, which re-runs the
+   apply/edit/capture loop and produces an updated staged change. The PR is required
+   to include the refreshed entry.
```
freben, line 818: I do wonder how hard this will be in practice. We've occasionally made changes that span a lot of files / packages. And we won't have plain git rebase etc to help us out. LLMs can probably help a lot - for those who have them, and sufficient token budgets. But we'll have to carry along the knowledge that every such stage we line up, we incur a potential headache for all contributors that come after (until they get integrated), not the original author.

There are probably some interesting challenges that can be had when several staged changes overlap or even interact indirectly, if one in the middle has a more relaxed `notBefore` than ones that follow etc.

freben (commented):
Cool!

Parsifal-M:
Awesome stuff! And I think a very good step in the right direction 👍

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: [automated] Good call. Swapped the illustrative `core-plugin-api` examples to `@backstage/frontend-plugin-api` in 6b506ddc1f48 so the examples stay relevant past the NFS migration. The workspace map listing in "Workspace: `framework`" still mentions the real `core-*` packages because those are the actual packages that live in the workspace, but the version-evolution / manifest / worked-example snippets are now NFS-based.

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: [automated] Symbolic only — habit, not intent. This BEP does not propose renaming the default branch. Substituted `main` → `master` throughout in 8e3e3ee9c51c to match the actual repo state.

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: > And we won't have plain git rebase etc to help us out

Depends on what tooling we make available 😉. I'm envisioning a couple of tools to help bridge between stages changes and git, essentially being able to enter rebase mode in git with all stages changes, and them take resulting branch and extract it back into staged changes. In the end a git branch is just a series of patches, and so are these staged changes, so it's possible to move back and forth quite easily.

In addition to that though, I also expect us to structure breaking changes in such a way that they minimize merge conflicts as much as possible, for example it might just be flipping a flag default, removing an export, etc. The leaner the patches the simpler the maintenance. I'm not expecting us to be maintaining a series of 1000+ line patches for half a year, then we're effectively doing it wrong. The cost of that is that we end up with a bit more dead code cleanup after breaking releases compared to what we would otherwise, but that tradeoff is prolly worth it, and we can try to structure things in folders etc. to make it simpler to delete. Maybe we'd even add some metadata down the line of the form "also delete this folder".

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1865 @@
...
+    {
+      "name": "@backstage/frontend-plugin-api",
+      "workspace": "framework",
+      "version": "2.4.7"
+    },
+    {
+      "name": "@backstage/plugin-catalog",
+      "workspace": "catalog",
+      "version": "2.3.0"
+    },
```
aramissennyeydd: What does it mean for the catalog plugin to be part of this release version? With separate lockfile and (I'm assuming) no use of workspace:* across workspaces, does the catalog workspace have to have been bumped to this version of the framework to show up here? If yes, that's a little weird as it would break pinning unless patch versions were incremented per workspace coming in.

Rugvip: Yep there would be no strict pinning of anything but the framework packages. The effect of declaring a dependency on a particular Backstage release for the rest of the packages would be to select the most recent version of the plugin that was not built towards a future release of the framework compared to the pinned version.

In practice the plugin releases become a lot more decoupled and managed individually. There is perhaps an alternative here where we aim to do overarching releases for the other workspaces too, for example after `2610` of the framework is out, the catalog updates to be compatible with that release and then itself releases `2610`. These separate workspace releases would generally not be picking up any new features from the framework releases, but rather just handle any breaking changes, which optimally is a no-op with deprecations already addressed. Any new framework features from for example `2610` that a plugin wants to use can be adopted in followup minor `2610` releases of the plugin.

Thinking about this, it might actually be a good pattern to roll out very broadly, where each plugin is able to declare what framework version it was built towards. I like the idea that the plugin workspaces in the main repo wouldn't be using any bespoke tooling, but instead manage framework releases the same way as any other plugin workspace would.

I'll explore a bit what this would mean in practice and what tooling might be needed to support this, and potentially update the BEP.

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1865 @@
...
+The PR is created and kept open even when the workspace has no staged changes. In
+that state it carries a single empty changeset and the description "no staged
+changes queued"; this avoids the noise of creating and closing the PR repeatedly as
+staged changes come and go.
+
+The Promote staged PR is opened as a regular pull request — the same shape as a Version
+Packages PR — and the regular review-and-merge process is the one that applies.
+Merging the PR is the signal to ship: it triggers the release through the same
+mainline flow the workspace uses for every other release, with no separate publish
+path. In the typical case the staged set contains breaking changes that bump the
```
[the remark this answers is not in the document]
Rugvip: [automated] Good catch — the wording read like the Promote staged PR was itself the publishing step, but in reality it just lands the synthesized changesets on `master` and the regular Version Packages PR flow takes it from there. Reworded in 0d1b295f2038 to spell that out: option (b) in your framing.

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1865 @@
...
+  and a future staged change targeted `plugin-foo@2.0.0` again, the next
+  publish would land at `2.0.0-next.0`, which collides with what's already
+  published. Letting the counter grow monotonically across rounds avoids the
+  entire class of collisions. Adopters do not see a clean `-next.0` for a new
+  round, but they keep the snapshot identifier property (all packages with the
+  same `-next.<N>` came from the same publish), which is what they actually
+  rely on.
+
+The updated `package.json` is committed to `master` by the same workflow that runs the
+publish, so the counter survives across runners.
```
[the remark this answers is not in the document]
Rugvip: [automated] Good catch — pushing the counter to `master` was not implementable as written. Reworked the design in 34a4a28c740a to drop the stored counter entirely: every `@next` dispatch fetches the published versions of each workspace package from the npm registry, finds the highest existing `-next.<N>` across all of them, and uses `<N>+1`. No counter file, no protected-branch push, no bot permissions required, and the same flow runs unchanged in any consumer repository.

The monotonic-counter and shared-snapshot-identity properties fall out of the registry lookup by construction. The BEP also now calls out the small wrapping step in the standalone release CLI that sits between `yarn changeset version` (which always emits `-next.0`) and the dispatched `npm publish` to rewrite the version strings to the workspace-wide `<N>`.

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1895 @@
...
+- This BEP does not propose changes to the public package names or import paths.
+- This BEP does not change how adopters consume Backstage releases, beyond introducing
+  more frequent `latest` releases and a more meaningful version scheme for the core
+  framework.
+- This BEP does not pre-commit to a specific set of breaking changes for any workspace.
+  It only describes the mechanism by which they will be authored and released.
+
+## Proposal
+
+### Workspace layout
```
awanlin, line 136: In Community Plugins, we introduced the `bcp.json` file that adds flags to be able to opt-in to new CI checks, would we do something like that here but maybe using the `backstage.json` file instead? Examples here: https://github.com/backstage/community-plugins/blob/main/docs/plugin-maintainers-guide.md#opt-in-to-automatic-version-bump-prs. Ot do we just assume that all workspaces will be able to adopt all CI checks now and into the future?

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1895 @@
...
+##### Workspace: `cli`
+
+All CLI and developer-tooling packages. Independent from `framework` so that CLI
+improvements can ship continuously, and the first non-`framework` workspace to
+migrate so it can exercise the new release tooling end to end.
+
+- `@backstage/cli`, `@backstage/cli-common`, `@backstage/cli-defaults`, `@backstage/cli-internal` (private), `@backstage/cli-node`
+- All `@backstage/cli-module-*` packages. The shared release-automation CLI lives
+  separately in the `tooling` workspace below — it is not a CLI module.
+- `@backstage/codemods`
```
awanlin, line 255: This is likely to be deprecated in favor of the new dedicated repo for this type of tooling. I think 🤔

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1895 @@
...
+#### Private workspaces
+
+##### Workspace: `microsite`
+
+The Docusaurus site that powers `backstage.io`. The workspace is private — it does
+not publish any packages — and is included in the workspace map for completeness.
+See [Documentation and microsite](#documentation-and-microsite) for how plugin
+docs are pulled in at build time.
+
+##### Workspace: `demo`
```
awanlin, line 380: This one is interesting, I'd say the biggest challenge, and is true of the current demo repo, is to balance between being a good reference example for Backstage adopters new to the platform while also showcasing what can be done with it. We want to make sure that what's in the demo matches fairly close to what you get form `create-app`. In this new workspace is the intention it would have unreleased features turned on? Or event test example data?

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1895 @@
...
+[Framework packages keep independent semver](#framework-packages-keep-independent-semver)
+below.
+
+#### Expected cadence
+
+The project targets two framework releases per year, in April and October. Both
+windows are chosen to land shortly before a KubeCon event so the most recent
+named release is fresh when adopters and maintainers meet there: KubeCon EU
+typically falls in April (occasionally March) and KubeCon NA typically falls in
+November, so an April release lines up with KubeCon EU and an October release
```
awanlin, line 1067: This is slightly confusing to me, KubeCon EU can fall on March or April but we use April. KubeCon NA falls on November so we use October. Why not use March for the KubeCon EU slot?

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1895 @@
...
+this rollout.
+
+### Framework versioning and the release lifecycle
+
+The way the framework workspace versions its packages is one of the more
+consequential decisions in this BEP, both for adopters (who pin Backstage releases by
+the framework identifier) and for maintainers (who reason about what's safe to ship
+between releases). This section spells out the scheme end to end.
+
+#### Identifier format
```
awanlin, line 1005: I think this is going to be confusing. If we always stick to the schedule it will end up that the cadence slot lines up with a month and that will become the de facto meaning of this value. With the existence of [CalVer](https://calver.org/) the assumption of it being a month could be compounded.

I understand the intent and that makes sense but I don't have any solutions to suggest right now. Will give it more thinking 🤔

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1895 @@
...
+
+For adopters who use the Backstage Yarn plugin to pin a Backstage release in
+`backstage.json` (see [Backstage release manifest](#backstage-release-manifest)
+for the plugin's resolution model), this means three useful properties:
+
+- A floating pin (`release: "2604"`) keeps them on the current line. When they
+  next run `yarn install` and let the lockfile update, the plugin picks up new
+  minor and patch versions of Backstage packages within the line. There is no
+  possibility of an unexpected breaking change.
+- A frozen pin (`release: "2604.1.3"`) locks resolution to a specific manifest;
```
awanlin: Ah, I don't see this syntax explained anywhere? It's not in the Identifier format section where I would expect to see it. What does the 1 and the 3 represent here?

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1895 @@
...
+3. If the cache hit indicates a successful previous build at this fingerprint,
+   exits immediately — there is nothing new to publish.
+4. Otherwise, runs the full build and publish, then writes the fingerprint to
+   the cache on success.
+
+Step 1 is cheap (it only needs `git ls-tree` reads at the `<workspace>@latest`
+tags on a shallow clone, no checkout of the workspaces themselves), so a no-op
+scheduled run finishes in seconds.
+
+#### Trade-off: cross-workspace doc edits
```
awanlin, line 1866: Just adding on to this:

- Doc maintainers have considered a dedicated repo in the past and have decided against this. It may be worth giving it another look but I still think keeping it with the code is better.
- Data from Backstage TechDocs use shows us that having the docs with the code means that more docs will actually be added/updated/corrected/etc.

beps/0015-workspaces-and-reworked-releases/README.md
awanlin: It's in a later section: Data shape under the Backstage release manifest heading.

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1895 @@
...
+module ecosystem.
+
+- `@backstage/plugin-notifications`, `plugin-notifications-backend`, `plugin-notifications-common`, `plugin-notifications-node`
+- All `@backstage/plugin-notifications-backend-module-*` packages
+
+##### Workspace: `kubernetes`
+
+- `@backstage/plugin-kubernetes`, `plugin-kubernetes-backend`, `plugin-kubernetes-cluster`, `plugin-kubernetes-common`, `plugin-kubernetes-node`, `plugin-kubernetes-react`
+
+##### Workspace: `api-docs`
```
awanlin, line 341: Should we consider folding this into the catalog itself? Does it make sense having a dedicated table just for APIs? The seems like a better fit being built into the catalog itself.

Also the underlying package for `plugin-api-docs-module-protoc-gen-doc` is deprecated and not well maintained, we have talked about deprecated this plugin many times. Maybe now is the time to do it?

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1895 @@
...
+- This BEP does not propose to physically split `backstage/backstage` into multiple git
+  repositories. Everything still lives in one repo.
+- This BEP does not propose changes to the public package names or import paths.
+- This BEP does not change how adopters consume Backstage releases, beyond introducing
+  more frequent `latest` releases and a more meaningful version scheme for the core
+  framework.
+- This BEP does not pre-commit to a specific set of breaking changes for any workspace.
+  It only describes the mechanism by which they will be authored and released.
+
+## Proposal
```
awanlin, line 134: One item I don't see mentioned or may have missed is dependancy management. One of our current and ongoing challenges in the Community Plugins repo is keeping up with the Renovate/Dependabot PRs. Often they sit for weeks looking to get approval from the various owners and then we often end up forcing them in. The Backstage repo has less owners to coordinate but that could still create a bottleneck. Worth having a section in this BEP on this topic.

awanlin (commented):
Really awesome work @Rugvip, looking forward to seeing this iterate and move forward! 🚀

schultzp2020:
This will be an amazing change!

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: [automated] The BEP has been updated in d78dc51a01d1 to reflect the direction the thread converged on. A new "[Workspace target framework version](https://github.com/backstage/backstage/blob/rugvip/workspace-release-restructure-bep/beps/0015-workspaces-and-reworked-releases/README.md#workspace-target-framework-version)" section spells out the model in detail, and the "[Backstage release manifest](https://github.com/backstage/backstage/blob/rugvip/workspace-release-restructure-bep/beps/0015-workspaces-and-reworked-releases/README.md#backstage-release-manifest)" section has been rewritten to scope the manifest to framework packages only.

Short version of the model:

- Every non-framework workspace declares the framework version it was built against in its `backstage.json` (`version: "2610.3.5"`). That is a specific `<line>.<minor>.<patch>`, not a line.
- The pack-time hook copies that value into every published `package.json` as `backstage.version`.
- The Backstage Yarn plugin filters non-framework packages on the npm registry by `backstage.version <= adopters framework version` and picks the latest compatible version. The manifest no longer enumerates non-framework packages.

This is a strict superset of how the catalog plugin would appear in your example: the resolution does not depend on `2604.1.3` knowing anything about catalog, and catalog does not need to update the manifest when it publishes. Adopters on `2604.x.y` see the most recent catalog version tagged with `<= 2604.x.y` at install time. Plugins lag behind the framework by exactly as much as their maintainers bump cadence dictates.

The model is the same for `backstage/backstage`, `backstage/community-plugins`, third-party open-source plugin repositories, and internal corporate plugin repositories — none of them need any special integration with the manifest.

The Alternatives section also now documents why we dont enumerate every published package in the manifest (cross-repo write boundary), why the metadata is a specific framework version rather than a line (honest support window by construction), and why we dont accept range declarations (out of scope; back-port pattern handles multi-line).

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: Yeah went a bit back and forth on this. I think my preference is to keep integration very simple, as in just Yarn scripts, and then potentially have more configuration through the tooling in the workspace. That reduced coupling a lot but of course limits our ability to do smarter things at the top level. Will very likely keep iterating on this until it feels right

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: Yep 👍, would still move if it exists at the time of migration though

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: Yep. My thinking here is that in the end we'll have the best result by pooling resources for the two rather than having one of each. Some things required by the demo site will mean that it's not the optimal example, but my gut feel is that it's worth the tradeoff

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: Just to keep it at an even cadence and not too close to EOY. Don't wanna force it to happen at the same time as KubeCon either, gets stressful :grin:. In the end the KubeCon timing was just something that I felt was the main thing that it makes sense to align with somewhat, but also not a strict requirement either

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: I wanted a proper Denver number this time around, which CalVer doesn't give us. Skriver requires a single integer basically, so anything like 2026-10, 2026-oct or 2026.10 doesn't work. Happy to consider option here, but gonna request concrete ones that feel moves :p. I arrived at this format after a lot of back and forth. What I like about it:

- it works as a semver major version
- it's fully calendar based
- it leaves space for extra releases or switching the cadence
- it's relatively short
- it's reasonably intuitive and informative at a glance
- it scales very far in the future and allows for many extra releases in a year

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: Indeed, incremental releases throughout the half year where the release is active, signaling whether the release contained patch changes only or new features as well

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: The main bits that this provides are the definition widgets. Agree it's prolly worth having this in the catalog workspace though

beps/0015-workspaces-and-reworked-releases/README.md
Rugvip: Hmm, yep worth thinking a bit more about this upfront for sure :thinking:

beps/0015-workspaces-and-reworked-releases/README.md
aramissennyeydd: dumb question, why does this new framework version need to be semver compatible?

beps/0015-workspaces-and-reworked-releases/README.md
```diff
@@ -0,0 +1,1895 @@
...
+to ship in. The project's cadence is "two releases per year, in April and
+October" (see [Expected cadence](#expected-cadence) below); under that cadence
+the slot numbers are `04` and `10`, so the regular releases are `2604`, `2610`,
+`2704`, `2710`, and so on. Picking slot numbers that match the planned calendar
+months means adopters can read the version at a glance and know roughly when it
+shipped, without having to chase exactly which week it landed.
+
+Two situations move `NN` away from a cadence slot:
+
+- **Extra releases off the cadence.** If we ever need to ship more than the
```
Parsifal-M, line 1012: > For
  example, with a `2604`/`2610` cadence, an extra release in July 2026 is `2607`;
  if a second extra release happens later that same month, it is `2608`.

I'm a bit worried this could lead to confusion or collision in some unlucky events, what about something like:

July: `2607`, `2607.1`, `2607.2`  
August: `2608`, `2608.1`

Just a random thought tho 😅

Rugvip: Yeah was leaning towards that earlier but it can't be combined with the number being a semver major. Those type of extra releases should be extremely rare either way, if they ever happen at all. This is more of an escape hatch in case we need it for some reason. Basically there's some flexibility built-in but we don't expect to ever need to use it.

PR #34331, opened 2026-05-21
feat(create-app): add CI workflow to default app template

awanlin:

Added a GitHub Actions CI workflow to the default app template. New Backstage instances created with `create-app` now include a `.github/workflows/ci.yml` that runs lint, type checking, tests, configuration validation, and a Docker image build on every pull request.

Also adds a getting-started docs page covering CI setup (GitHub Actions, GitLab CI, Jenkins) and documents the `repo list-deprecations` CLI command.

### DIFF
--- a/packages/create-app/templates/default-app/.github/workflows/ci.yml
+++ b/packages/create-app/templates/default-app/.github/workflows/ci.yml
@@ -1,0 +1,69 @@
+# Documentation: https://backstage.io/docs/getting-started/ci
+name: PR CI build
+
+on:
+  pull_request: {}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    env:
+      CI: true
+      NODE_OPTIONS: --max-old-space-size=4096
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      # Beginning of yarn setup
+      - name: use node.js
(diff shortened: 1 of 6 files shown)


packages/create-app/templates/default-app/.github/workflows/ci.yml
aramissennyeydd, line 1: Should we make this opt-in during create-app execution?

awanlin: My personal opinion is no. We don't want to have flags so people can get up and running quickly, having this adds value right out of the box, a large portion of our adopters are using GitHub so this feels reasonable to ship form the start. Deleting it is pretty easy. I'm open to hearing opinions going a different direction though.

Parsifal-M:
Hmm I like the idea but if you don't use GitHub it's just extra stuff you then have to delete when you create the app 🤔

I wonder if alternatively we could document these ci jobs and reference them somewhere so you can just copy and paste your flavour of CI 💭

packages/create-app/templates/default-app/.github/workflows/ci.yml
```diff
@@ -0,0 +1,69 @@
+# Documentation: https://backstage.io/docs/getting-started/ci
```
aramissennyeydd, line 1: Chatted in the Docs SIG - Let's add links for where to get templates for the other CI systems we're adding here.

aramissennyeydd (approved):
Nice! I think this makes sense for a sane default and gives us some leverage to recommend new commands to run in CI.

PR #34336, opened 2026-05-22
feat(catalog): add skills endpoint

drodil:

add well-known endpoint to fetch skills from the catalog using `npx skills` cli tool the same way backstage.io provides it's skills.

### DIFF
--- a/plugins/catalog-backend-module-ai-model/src/module.ts
+++ b/plugins/catalog-backend-module-ai-model/src/module.ts
@@ -14,13 +14,18 @@
  * limitations under the License.
  */
 
-import { createBackendModule } from '@backstage/backend-plugin-api';
 import {
-  CatalogModelSources,
+  coreServices,
+  createBackendModule,
+} from '@backstage/backend-plugin-api';
+import {
   aiResourceEntityModel,
+  CatalogModelSources,
   mcpServerApiEntityModel,
 } from '@backstage/catalog-model/alpha';
 import { catalogModelExtensionPoint } from '@backstage/plugin-catalog-node/alpha';
+import { catalogServiceRef } from '@backstage/plugin-catalog-node';
+import { createSkillsRouter } from './SkillsRouter';
 
 /**
  * Registers support for the AiResource entity kind in the catalog.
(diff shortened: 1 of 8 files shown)


plugins/catalog-backend-module-ai-model/src/module.ts
```diff
@@ -34,14 +39,38 @@ export const catalogModuleAiResourceEntityModel = createBackendModule({
...
+
+        const skillsRouter = createSkillsRouter({
+          catalog,
+          auth,
+          reader,
+          cache,
+          logger,
+        });
+
+        rootHttpRouter.use('/.well-known/skills', skillsRouter);
```
Rugvip, line 73: Hmm, is this relying on the catalog being unauthenticated? 🤔

drodil: Nope, it uses the catalog's own service token to fetch the entities. One thing I was wondering is that there is no `addAuthPolicy` in the `RootHttpRouterService` so do all those actually require authentication or not 🤔

drodil: Ah, actually no: Routes registered via rootHttpRouter are unauthenticated by default. The root HTTP router has no auth middleware — it only applies standard middleware (helmet, cors, compression, logging, rate limiting).

Rugvip: Ah yep, didn't spot that it was the root http router. So indeed, it gives unauthenticated access to alls skills in the catalog to the world 😅. I think we need to address that in some way to be able to ship this tbh. Initial questions though is whether there's any possibility to have authentication still be enabled when using the `skills` CLI? 🤔

pguppy: sadly [npx skills ](https://github.com/vercel-labs/skills) tool doesn't support passing thru any sort of authentication at all for our use case, its not actually possible at all to even use this with `npx skills` if we were to add auth. I agree that an unauthenticated endpoint is def a big risk here, giving direct access into the catalog, though currently there isnt' any other option if we want use `npx skills` sadly.

I def do want this feature really badly so I think a good trade-off is defaulting this to be disabled behind a features flag in the app-config though still allowing it to be a thing for users who run backstage internally and are happy to expose the endpoint with no authentication. 

There are alternative tools such as [agent-skills-cli](https://github.com/Karanjot786/agent-skills-cli) which you can use with support authenticated endpoints like

```
# Using token via flag
npx agent-skills-cli install https://your-backstage.example.com/.well-known/skills --token=your-backstage-token

# Or via environment variable (recommended for security)
BACKSTAGE_TOKEN=your-token npx agent-skills-cli install https://your-backstage.example.com/.well-known/skills
```

though they aren't as well supported as the Vercel version. Or we could also ig make our own version of skills download cli tool using the backstage-cli which is kinda of whole different thing though I would image might be the end goal here.


https://github.com/vercel-labs/skills/issues/1176

PR #34338, opened 2026-05-22
chore: group @opentelemetry/* packages in Renovate

freben:
## Summary

- Group all `@opentelemetry/*` packages so they get upgraded together in a single PR, matching how other monorepo package families (Rush Stack, Module Federation, etc.) are already grouped.


awanlin (approved):
Good call, think we should do this for the Demo site too! 🚀

Rugvip (approved):

PR #34342, opened 2026-05-22
golden-path: add docs for integrations

aramissennyeydd:

Replaces the TODO stubs in the catalog, search, and notifications chapters of the plugins golden path with full walkthroughs:

- Catalog: explains the catalog as a typed entity graph and walks through the three main integration points (annotations + processor, custom kind via the catalog model, and entity provider sync).
- Search: explains collators/decorators and walks through building a custom todo collator with `searchIndexRegistryExtensionPoint`, permission-aware indexing, and a frontend result item.
- Notifications: explains the in-app + external delivery model and walks through sending notifications from a scheduled task (alarm fanout) and from the create handler (assign-on-create), with notes on topics, scope, and group recipients.

### DIFF
--- a/docs/golden-path/plugins/integrations/001-catalog.md
+++ b/docs/golden-path/plugins/integrations/001-catalog.md
@@ -9,16 +9,402 @@ description: How to integrate your plugin with the Backstage Software Catalog
 
 ### What is the Software Catalog?
 
-<!--TODO-->
+The [Software Catalog](../../../features/software-catalog/index.md) is the
+graph of typed entities that sits at the center of every Backstage instance.
+Components, APIs, Resources, Systems, Users, Groups — anything an organization
+wants to model and reason about — live in the catalog and are related to each
+other through a small set of well-known relations.
+
+Each entity is identified by its `apiVersion`, `kind`, and `metadata.name`
+(scoped by `metadata.namespace`). The catalog backend ingests entity
+descriptors from one or more _sources_, validates them against a schema, runs
+them through a pipeline of _processors_, stitches the resulting relations
+together, and exposes the final state through a queryable API that the
+frontend and other plugins consume.
+
+For your plugin, the catalog is the place you go when you want a stable,
+shared answer to "which thing is this, who owns it, and how does it relate to
+everything else?" — rather than maintaining your own list of services, owners,
+or resources.
 
 ### Integration Points
 
-<!--TODO-->
+There are three main places where a plugin can plug into the catalog. Pick the
+one that matches the question you are trying to answer.
+
+**Entity providers** push entities into the catalog. They run on a schedule,
+fetch data from an external system, and emit a full or delta set of entities
+that the catalog should know about. Use a provider when your plugin owns a
+source of truth that should appear in the catalog — todos discovered in a
+backing store, for example.
+
+**Catalog processors** transform entities as they flow through the ingestion
+pipeline. They can read annotations, mutate spec fields, emit additional
+entities, or attach extra relations. Use a processor when you want to react to
+existing entities or annotations rather than provide a brand-new source.
+
+**The catalog model** controls which kinds, versions, and spec types are
+considered valid. Extending the model lets you introduce a new first-class
+kind — `Todo`, for example — with its own schema, relations, and type guards.
+Use a model extension when your data does not fit comfortably into any of the
+built-in kinds.
+
+A typical plugin uses one or two of these together: a provider plus a model
+extension to introduce a new kind, or a processor that reads an annotation off
+existing entities to expose plugin-specific data.
 
 ## Adding a new `backstage.io/todo` annotation
 
-<!--TODO-->
+The lightest-weight integration is to attach to entities that already live in
+the catalog. Adopters add an annotation to their `catalog-info.yaml`, and your
+plugin reacts to it. There is no new kind, no new provider, and the source of
+truth stays in the adopter's repository.
+
+### Reserve the annotation name
+
+Annotations are namespaced strings, and the `backstage.io/` prefix is
+reserved for annotations defined by the project. For a plugin you ship
+yourself, pick a namespace you control — for example
+`todo.backstage.io/source` — and document it. For the rest of this guide we
+will pretend our plugin is shipped by the Backstage project and use
+`backstage.io/todo-source`.
+
+Add a constant in your common package so backend, frontend, and any policies
+reference the same string:
+
+```ts
+// plugins/todo-common/src/annotations.ts
+export const TODO_SOURCE_ANNOTATION = 'backstage.io/todo-source';
+```
+
+### Read the annotation in the backend
+
+Inside a route handler, look the entity up through the catalog client and
+read the annotation off `metadata.annotations`:
+
+```ts
+// plugins/todo-backend/src/service/router.ts
+import { CatalogService } from '@backstage/plugin-catalog-node';
+import { TODO_SOURCE_ANNOTATION } from '@internal/plugin-todo-common';
+
+router.get('/todos/by-entity/:ref', async (req, res) => {
+  const credentials = await httpAuth.credentials(req, { allow: ['user'] });
+  const entity = await catalog.getEntityByRef(req.params.ref, { credentials });
+
+  const source = entity?.metadata.annotations?.[TODO_SOURCE_ANNOTATION];
+  if (!source) {
+    res.json({ items: [] });
+    return;
+  }
+
+  const items = await todoList.listTodosForSource(source);
+  res.json({ items });
+});
+```
+
+The annotation is just a hint; the actual todo data still comes from your
+plugin's own store. The catalog is the index, not the database.
+
+### Surface the annotation on the entity page
+
+On the frontend, use the standard helpers to check for the annotation before
+showing your card. This keeps the UI clean for entities that have not opted
+in:
+
+```tsx
+// plugins/todo/src/components/EntityTodoContent.tsx
+import { useEntity } from '@backstage/plugin-catalog-react';
+import { TODO_SOURCE_ANNOTATION } from '@internal/plugin-todo-common';
+
+export const isTodoAvailable = (entity: Entity) =>
+  Boolean(entity.metadata.annotations?.[TODO_SOURCE_ANNOTATION]);
+
+export const EntityTodoContent = () => {
+  const { entity } = useEntity();
+  if (!isTodoAvailable(entity)) {
+    return <MissingAnnotationEmptyState annotation={TODO_SOURCE_ANNOTATION} />;
+  }
+  return <TodoListForEntity entityRef={stringifyEntityRef(entity)} />;
+};
+```
+
+`MissingAnnotationEmptyState` is the conventional way to nudge adopters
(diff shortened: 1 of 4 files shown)


docs/golden-path/plugins/integrations/001-catalog.md
```diff
@@ -9,16 +9,402 @@ description: How to integrate your plugin with the Backstage Software Catalog
...
 ### What is the Software Catalog?
 
-<!--TODO-->
+The [Software Catalog](../../../features/software-catalog/index.md) is the
+graph of typed entities that sits at the center of every Backstage instance.
+Components, APIs, Resources, Systems, Users, Groups — anything an organization
+wants to model and reason about — live in the catalog and are related to each
+other through a small set of well-known relations.
+
+Each entity is identified by its `apiVersion`, `kind`, and `metadata.name`
```
Rugvip, line 18: Not really true, the `apiVersion` isn't part of identifying it as such? 🤔 at least I think it's confusing to equate to the kind + namespace + name tuple, that's a more common and central concept. The `apiVersion` is tells you what the data means

docs/golden-path/plugins/integrations/001-catalog.md
```diff
@@ -9,16 +9,402 @@ description: How to integrate your plugin with the Backstage Software Catalog
 
 ### What is the Software Catalog?
 
-<!--TODO-->
+The [Software Catalog](../../../features/software-catalog/index.md) is the
+graph of typed entities that sits at the center of every Backstage instance.
+Components, APIs, Resources, Systems, Users, Groups — anything an organization
+wants to model and reason about — live in the catalog and are related to each
+other through a small set of well-known relations.
```
Rugvip, line 16:
```suggestion
other through a set of well-known relations.
```

maybe? 🤔

docs/golden-path/plugins/integrations/001-catalog.md
```diff
@@ -9,16 +9,402 @@ description: How to integrate your plugin with the Backstage Software Catalog
 
 ### What is the Software Catalog?
 
-<!--TODO-->
+The [Software Catalog](../../../features/software-catalog/index.md) is the
```
Rugvip, line 12: Thinking there's a need for a fair bit more deep links to specific docs for the different concepts introduced here

docs/golden-path/plugins/integrations/001-catalog.md
```diff
@@ -9,16 +9,402 @@ description: How to integrate your plugin with the Backstage Software Catalog
...
+the catalog. Adopters add an annotation to their `catalog-info.yaml`, and your
+plugin reacts to it. There is no new kind, no new provider, and the source of
+truth stays in the adopter's repository.
+
+### Reserve the annotation name
+
+Annotations are namespaced strings, and the `backstage.io/` prefix is
+reserved for annotations defined by the project. For a plugin you ship
+yourself, pick a namespace you control — for example
+`todo.backstage.io/source` — and document it. For the rest of this guide we
```
Rugvip, line 68:
```suggestion
`todo.acme.org/source` — and document it. For the rest of this guide we
```

Perhaps? Just to avoid confusion. Or `acme.org/todo-source`

docs/golden-path/plugins/integrations/001-catalog.md
```diff
@@ -9,16 +9,402 @@ description: How to integrate your plugin with the Backstage Software Catalog
...
+Inside a route handler, look the entity up through the catalog client and
+read the annotation off `metadata.annotations`:
+
+```ts
+// plugins/todo-backend/src/service/router.ts
+import { CatalogService } from '@backstage/plugin-catalog-node';
+import { TODO_SOURCE_ANNOTATION } from '@internal/plugin-todo-common';
+
+router.get('/todos/by-entity/:ref', async (req, res) => {
+  const credentials = await httpAuth.credentials(req, { allow: ['user'] });
```
Rugvip, line 91:
```suggestion
  const credentials = await httpAuth.credentials(req);
```

Typically you don't want to do this and just go with the default policy, I think it's probably best to avoid in this example. That's unless we specifically want to list todos for a user, but feeling that would complicate the example

docs/golden-path/plugins/integrations/001-catalog.md
```diff
@@ -9,16 +9,402 @@ description: How to integrate your plugin with the Backstage Software Catalog
...
+```
+
+### Read the annotation in the backend
+
+Inside a route handler, look the entity up through the catalog client and
+read the annotation off `metadata.annotations`:
+
+```ts
+// plugins/todo-backend/src/service/router.ts
+import { CatalogService } from '@backstage/plugin-catalog-node';
```
Rugvip, line 87: wonder if this needs a bit more context, maybe wrapping the example in a `createRouter` or such

docs/golden-path/plugins/integrations/001-catalog.md
```diff
@@ -9,16 +9,402 @@ description: How to integrate your plugin with the Backstage Software Catalog
...
+
+`MissingAnnotationEmptyState` is the conventional way to nudge adopters
+towards adding the annotation, and it links straight to your plugin's docs.
+
+### Validate the annotation with a processor
+
+If you want to fail fast when somebody types the annotation wrong, register
+a catalog processor that validates it during ingestion. Processors run as
+part of the catalog pipeline, so a bad annotation surfaces as a refresh error
+on the entity rather than as a confusing empty state in the UI:
```
Rugvip, line 139: Thinking it's worth mentioning that this is highly optional, and most plugins skip this due to the installation complexity and instead fail within the plugin instead

docs/golden-path/plugins/integrations/001-catalog.md
```diff
@@ -9,16 +9,402 @@ description: How to integrate your plugin with the Backstage Software Catalog
...
+      },
+    });
+  },
+});
+```
+
+Adopters install the module alongside the catalog backend, and the annotation
+is now part of their ingestion pipeline.
 
 ## Custom TODO Entity Kind
```
Rugvip, line 188: I think this detracts from the golden path concept quite a bit, in that it's more of a collection of guides rather than a single one if we add this. In general we'd really want to golden path to be a guide for how to concretely build an actual full example project, so that in following the path your end result is a functional thing. It's fine to have a lot of it be explainers for code in the template, especially early on. But beyond that we'd really want the path to tell you exactly what code additions to make, how to verify that they work in the UI, etc.

Then again, maybe this is exactly what we want to build in the guide, given the search integration and so on. Bit weary that it might be too large to also introduce the kinds and so on. This is also the golden path to building plugins, and it's very rare that you are in this are as part of plugin development. I think this is probably better reserved for an app golden path.

Rugvip (commented):
Nice! 🎉 😁 

Couple of more detailed comments on the catalog section, skipped a deep dive into search and notifications since I think we want to settle https://github.com/backstage/backstage/pull/34342/changes#r3311158809 first. I do think it's important for golden paths to also teach the user how to verify what they're building along the way as well.

docs/golden-path/plugins/integrations/001-catalog.md
aramissennyeydd: yeah bad claude moment - will update

docs/golden-path/plugins/integrations/001-catalog.md
aramissennyeydd: Happy to remove this, I was trying to strike a balance between things that may be useful now vs useful in the future and trying to figure out when is the right time to introduce them. I don't know how useful this really would be for end users especially early adopters.

Rugvip: I think it could be useful, but for an app golden path rather than plugin building. It's very rare that you encounter custom kinds for plugins, but definitely something I think worth introducing for apps.

docs/golden-path/plugins/integrations/001-catalog.md
milliehartnt123, line 16: It would be helpful to provide a link to page that describes the set of relationships: Well-known Relations between Catalog Entities (docs/features/software-catalog/well-known-relations)

docs/golden-path/plugins/integrations/001-catalog.md
```diff
@@ -9,16 +9,402 @@ description: How to integrate your plugin with the Backstage Software Catalog
...
+plugin reacts to it. There is no new kind, no new provider, and the source of
+truth stays in the adopter's repository.
+
+### Reserve the annotation name
+
+Annotations are namespaced strings, and the `backstage.io/` prefix is
+reserved for annotations defined by the project. For a plugin you ship
+yourself, pick a namespace you control — for example
+`todo.backstage.io/source` — and document it. For the rest of this guide we
+will pretend our plugin is shipped by the Backstage project and use
```
milliehartnt123, line 69: _and use_ should be _and uses_

docs/golden-path/plugins/integrations/001-catalog.md
```diff
@@ -9,16 +9,402 @@ description: How to integrate your plugin with the Backstage Software Catalog
...
 
 ## Adding a new `backstage.io/todo` annotation
 
-<!--TODO-->
+The lightest-weight integration is to attach to entities that already live in
+the catalog. Adopters add an annotation to their `catalog-info.yaml`, and your
+plugin reacts to it. There is no new kind, no new provider, and the source of
+truth stays in the adopter's repository.
+
+### Reserve the annotation name
```
milliehartnt123, line 63: This looks like a series of steps. The CNCF wants procedural information as a set of steps, instead of a series of paragraphs or subheadings. For example:

1. Reserve the annotation name.
2. Read the annotation in the backend.
3. Surface the annotation on the entity page.
4. Validate the annotation with a processor.
5. Wire the processor into a catalog backend module

docs/golden-path/plugins/integrations/002-search.md
```diff
@@ -9,12 +9,239 @@ description: How to integrate your plugin with Backstage Search
...
+ownership data to documents from another collator — but a collator plus a
+result item is the common case.
 
 ## Creating a custom TODO collator
 
-<!--TODO-->
+The goal is to make every todo in your plugin's store appear in global
+search, scored alongside catalog entities and TechDocs pages.
+
+### Define the document type
```
milliehartnt123, line 57: All these subheadings should be numbered steps for creating a custom TODO collator. CNCF wants procedural information to be numbered steps, not subheadings or paragraphs.

1. Define the document type
2. Build the collator factory
3. Register the collator with the search backend
....... and so on

docs/golden-path/plugins/integrations/004-notifications.md
```diff
@@ -9,16 +9,206 @@ description: How to integrate your plugin with Backstage Notifications
...
 
 ## TODO with an alarm
 
-<!--TODO-->
+The goal is to let users set a due time on a todo and receive a notification
+when the alarm fires. The alarm is just a timestamp; the scheduler service
+fires us up to look for due todos, and the notifications service delivers
+the message.
+
+### Persist the alarm
```
milliehartnt123, line 55: Again these subheadings should be a set of numbered steps to setup TODO with an alarm.

aramissennyeydd:
closing in favor of a longer-lived branch with all golden paths work

PR #34349, opened 2026-05-22
fix(ui): keep BUI Header title and actions aligned on wide viewports

### LINKED ISSUE #34346: BUI Header: title and customActions shift inward
### 📜 Issue Labels

### 🔎 Search Terms

```plain
bui, mui to bui, migration, header, title, customActions, component, backstage, ui
```

### 🗃️ Project Area

Design System

### 🔗 External Integration

N/A

### 📝 Description & Context

The header component has a layout issue when the browser window is very wide.

Component: https://ui.backstage.io/components/header
Package: @backstage/ui: "0.15.0"

### 👍 Expected Behavior

- The title should remain fully left-aligned
- The customActions should remain fully right-aligned

Image

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

It's reproduceable on the demo catalog page: https://demo.backstage.io/catalog

1. Open a page that uses the header component
2. Increase the browser width significantly
3. Observe the position of the title and customActions in the header

Undecided

pen-pal:

Fixes #34346

The BUI `Header` wraps each of its three sections (`headerTop`, `content`, `headerBottom`) in the shared `Container` component, which has `max-width: 120rem` (1920px) and `margin-inline: auto`. On viewports wider than 1920px this centers the header content and leaves empty gutters on either side, while the page body below it (e.g. the catalog table) extends to the full available width. The result is that the title and `customActions` visibly shift inward relative to the rest of the page.

This change adds a scoped override in `Header.module.css` that drops `max-width` and `margin-inline` for the three Header wrappers only, using the same doubled-class specificity pattern already in that file. The shared `Container` cap is unchanged, so other consumers (`SearchField`, `Table`, `PluginHeader`, etc.) keep the readability constraint.

Padding-inline from `Container` is preserved, so the header still has consistent page-level side padding — it just spans the full available width now, with the title flush to the left and `customActions` flush to the right.

  

## Screenshot
**Before** 
before-ultrawide-3440
before-narrow-1440
before-wide-2400

**After**
after-narrow-1440
after-ultrawide-3440
after-wide-2400

### DIFF
--- a/.changeset/catalog-page-container-wrap.md
+++ b/.changeset/catalog-page-container-wrap.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog': patch
+---
+
+Constrained the catalog index page body to the standard page max-width so it stays aligned with the page header on very wide viewports.
--- a/.changeset/fix-bui-header-wide-viewport-alignment.md
+++ b/.changeset/fix-bui-header-wide-viewport-alignment.md
@@ -1,5 +1,0 @@
----
-'@backstage/ui': patch
----
-
-Fixed `Header` so its title and `customActions` stay aligned with the page edges on very wide viewports. Previously, on screens wider than 1920px the header content was centered with empty gutters on either side, making it appear shifted inward relative to the page body below.
--- a/.changeset/fix-bui-header-wide-viewport-alignment.md
+++ b/.changeset/fix-bui-header-wide-viewport-alignment.md
@@ -1,0 +1,5 @@
+---
+'@backstage/ui': patch
+---
+
+Fixed `Header` so its title and `customActions` stay aligned with the page edges on very wide viewports. Previously, on screens wider than 1920px the header content was centered with empty gutters on either side, making it appear shifted inward relative to the page body below.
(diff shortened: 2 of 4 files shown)


patrickwyler:
Is it just a problem because not all plugin content like the one from the catalog page is wrapped with a  [Container](https://ui.backstage.io/components/container)?

```
<Container>
  {/* Your plugin's main content */}
</Container>
```

_Container provides the standard page layout for plugin content. It constrains content to a maximum width, centers it horizontally, and adds consistent horizontal gutters. Use it once per page to wrap your main content area._
https://ui.backstage.io/components/container#core-concepts

Naycon:
I think `Content` is meant to be the main page container element, and so is `Container`, so I don't think we should nest them. It's an either/or kind of situation.

I recently created #34487 that will hopefully help align the Catalog index page better with the BUI PageHeader and Header, including swapping to `Container`. I think that should take care of fixing this as well so it behaves as intended in BUI. Here's what it looks like for a simulated wide screen on that PR:

image

If this is not what you expected, maybe you can create an issue describing what you would like to change in terms of the BUI page layout?

patrickwyler:
@Naycon I closed my issue. I think your MR will solve the problem.

Naycon:
I'm going to close this, as we have another fix in place now for the Catalog index page. Feel free to reopen if I've missed something this achieves that the merged fix does not.

PR #34355, opened 2026-05-25
fix(frontend-plugin-api): honor noHeader on all PageBlueprint pages

### LINKED ISSUE #34051: PageBlueprint "noHeader" override is not applied in certain scenarios
### 📜 Issue Labels

### 🔎 Search Terms

```plain
PageBlueprint
noHeader
Header
```

### 🗃️ Project Area

Unknown

### 🔗 External Integration

N/A

### 📝 Description & Context

I upgraded my Backstage app recently to version `1.50.1` and I wanted my frontend plugins to take the full screen, so I applied the `noHeader` override option.

The issue is that some plugins don't apply this `noHeader` option because of the code that's written here: [PageBlueprint](https://github.com/backstage/backstage/blob/master/packages/frontend-plugin-api/src/blueprints/PageBlueprint.tsx)

There is only one case where this `noHeader` is applied, which is if `params` has this `loader` function passed to it. Otherwise, all other pages won't apply this `noHeader` option.

For example look at these 2 if / else statement blocks:
- [If statement - block 1](https://github.com/backstage/backstage/blob/master/packages/frontend-plugin-api/src/blueprints/PageBlueprint.tsx#L122)
- [Else statement - block 2](https://github.com/backstage/backstage/blob/master/packages/frontend-plugin-api/src/blueprints/PageBlueprint.tsx#L172)

### 👍 Expected Behavior

I expect no header visible on the page when I apply `noHeader` override option to any frontend plugin

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

Here is a block of code that helps with the reproduction:

```ts
import scaffolderPluginBase from '@backstage/plugin-scaffolder/alpha';

export const scaffolderPlugin = scaffolderPluginBase.withOverrides({
  extensions: [
    scaffolderPluginBase.getExtension('page:scaffolder').override({
      params: {
        noHeader: true,
      },
    }),
  ],
});
```

The `noHeader` override option does not work in this scenario because of this `else if` statement:
https://github.com/backstage/backstage/blob/master/packages/frontend-plugin-api/src/blueprints/PageBlueprint.tsx#L122

officialasishkumar:

The `noHeader` option on pages created with `PageBlueprint` only took effect for pages defined with a `loader`. Pages that render sub-pages with tabs, and pages without a loader, ignored the option and always rendered the plugin header, so overriding a page with `noHeader: true` had no effect for many plugins.

This forwards `noHeader` to the page layout in those cases as well, and makes the default page layout skip rendering its header when the option is set, so the page can fill all available space.

The change is covered by tests for both `PageBlueprint` and the default `PageLayout`.

Fixes #34051

### DIFF
--- a/.changeset/page-blueprint-no-header.md
+++ b/.changeset/page-blueprint-no-header.md
@@ -1,0 +1,5 @@
+---
+'@backstage/frontend-plugin-api': patch
+---
+
+Fixed the `noHeader` option on pages created with `PageBlueprint` only taking effect for pages that use a `loader`. The option is now also honored by pages that render sub-pages with tabs and by pages without a loader, and the default page layout now hides its header when the option is set.
--- a/packages/frontend-plugin-api/src/blueprints/PageBlueprint.test.tsx
+++ b/packages/frontend-plugin-api/src/blueprints/PageBlueprint.test.tsx
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import { ReactNode } from 'react';
 import { createRouteRef } from '../routing';
 import { PageBlueprint } from './PageBlueprint';
 import {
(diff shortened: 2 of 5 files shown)


Rugvip (changes requested):
Thank you!

I'm not too sure about this one to be honest. If find it a bit questionable to remove the header but keep the tabs 🤔. I'd probably expect the tabs to go away as well if one wants to remove the header? That will effectively break the page if one doesn't supply the sub page routing in a custom way instead, but the `noHeader` is already only intended as an escape hatch to do more custom things, rather than something to apply broadly.

If one wants to remove or change the header globally, then overriding the `PageLayout` is the way to go instead.

catz:
I’m a bit confused about the new headers.

For example, if I take pages like Catalog Graph and Register Existing Component, both of them show a mix of the new and old header styles.

I see the same behavior when I install the Community RBAC plugin. Is it a bug or not ?

Screenshot 2026-06-22 at 15 56 44

Issue #2254, opened 2020-09-03
Unable to ingest from Azure git repos


mfrinnstrom:
I tried ingesting a template from a private Azure git repo but got an error. After some investigation it seems that the generated URL doesn't correspond with what is in the Azure docs.

## Expected Behavior
As per [this](https://docs.microsoft.com/en-us/rest/api/azure/devops/git/items/get?view=azure-devops-rest-6.0) documentation the URL should be in the following format:

`GET https://dev.azure.com/{organization}/{project}/_apis/git/repositories/{repositoryId}/items?path={path}&api-version=6.0`

## Current Behavior
The currently generated URL looks like the following:

`GET https://dev.azure.com/{organization}/{project}/_apis/sourceProviders/TfsGit/filecontents?repository={repositoryId}&commitOrBranch=undefined&path={path}&api-version=6.0-preview.1`

I have not tried with a Team Foundation Version Control (TFVC) repo so I'm not sure if that works and what would happen if the URL is changed.

## Possible Solution
Change AzureApiReaderProcessor.ts to the documented URL.

## Steps to Reproduce

1. Go to register an existing component
2. Supply a URL for a private Azure repo

## Your Environment

- NodeJS Version (v12): v12.18.3
- Operating System and Version (e.g. Ubuntu 14.04): VS Code Node.js devcontainer
- Browser Information: Chrome 85.0.4183.83

freben:
Hi, and thanks for reporting.

I am not familiar with the Azure APIs and offerings myself, but [commented on the original PR](https://github.com/spotify/backstage/pull/2081#discussion_r476437869) for this that it looked like it implemented one of several providers. And indeed, the [tfvc API docs](https://docs.microsoft.com/en-us/rest/api/azure/devops/tfvc/items/get?view=azure-devops-rest-6.0) do list a different URL for items.get.

There are at least two possible paths forward:

- Give them a distinct location type each. It currently is `azure/api`, but could be split into e.g. `azure/tf` and `azure/git` or similar.
- Make the existing processor support both formats automatically.

The latter would be the most convenient from a user standpoint.

Since I do not have an Azure environment to play with, could you @mfrinnstrom and/or @o-farooq check if the input URLs as seen in the browser location bar are the same no matter what the repo type is? The plugin currently shows

`https://dev.azure.com/{organization}/{project}/_git/reponame?path={path}&version=GB{commitOrBranch}&_a=contents`

[...]

PR #34393, opened 2026-05-26
try to remove protojs pins

freben:
## Summary

Removes the protobufjs/inquire version pins from the seed lockfile and suppresses the resulting bundler warning.

### Background

`@protobufjs/inquire@1.1.0` used `eval("require")(moduleName)` to hide a dynamic require from bundlers. When we pinned it, that was the workaround. The maintainer has since published `1.1.1` and `1.1.2` which remove the eval, but replace it with a bare `require(moduleName)` — still a dynamic require that webpack/rspack flags as "Critical dependency: the request of a dependency is an expression".

However, **protobufjs 7.5.9** (2026-05-17, [PR #2254](https://github.com/protobufjs/protobuf.js/pull/2254)) backported bundler-safe optional module lookups: all call sites that previously used `util.inquire("fs")`, `util.inquire("buffer")`, etc. were replaced with direct requires using webpack magic comments. The `inquire` utility is still in the dependency tree but is never exercised with dynamic module names.

### What this PR does

1. **Removes the version pins** for `protobufjs` and `@protobufjs/inquire` from `packages/create-app/seed-yarn.lock`, letting them resolve naturally (protobufjs 7.6.0, @protobufjs/inquire 1.1.2).

2. **Adds an `ignoreWarnings` entry** in the bundler config (`packages/cli-module-build`) to suppress the false-positive "Critical dependency" warning from `@protobufjs/inquire`. The comment in the source explains the reasoning and links to the upstream issue.

### Why `ignoreWarnings` is safe for all end users

The suppression ships to end users via `@backstage/cli-module-build`. This is safe regardless of which protobufjs version they have resolved in their lockfile:

- `@protobufjs/inquire` is designed for optional module lookups — it returns `null` when a require fails, never crashes. The "Critical dependency" is a bundler static analysis warning, not a runtime error.
- protobufjs is a server-side dependency (gRPC, Firestore, OpenTelemetry). If it ends up in a frontend bundle, the dynamic require was always non-functional regardless of version.
- For protobufjs >= 7.5.9, the dynamic require path in inquire is dead code (never called).
- For protobufjs < 7.5.9, the inquire calls degrade gracefully at runtime. We're not masking a real failure.

We considered adding `"protobufjs": ">=7.5.9"` as a dependency floor in a published package like `@backstage/backend-defaults` to force end users onto the fixed version. We decided against it — the `ignoreWarnings` is sufficient, and promoting a transitive dependency to a direct one just to set a floor could cause resolution conflicts for adopters with other constraints.

See also: [protobufjs/protobuf.js#2057](https://github.com/protobufjs/protobuf.js/issues/2057)

## Test plan

- [ ] E2E `create-app` test passes (yarn install + tsc + build)
- [ ] No protobufjs/inquire warnings in build output


Rugvip (approved):
Alright, seems like an exceptional enough breakage of an existing dependency to make this worthwhile

PR #34394, opened 2026-05-26
Add config to disable experimental public entry point

benjdlambert:
Adds an `app.disablePublicEntryPoint` config option to the app-backend plugin. When set to `true`, the backend skips serving the public entry point to unauthenticated users even if the app bundle includes a `dist/public` directory from an `index-public-experimental` entry point.

This lets adopters explicitly opt out of the automatic public sign-in page behavior.

### DIFF
--- a/plugins/app-backend/config.d.ts
+++ b/plugins/app-backend/config.d.ts
@@ -44,12 +44,12 @@ export interface Config {
     disableStaticFallbackCache?: boolean;
 
     /**
-     * Disables the experimental public entry point used for unauthenticated
-     * sign-in pages. When the app is bundled with an
-     * `index-public-experimental` entry point, the app backend will
-     * automatically serve it to unauthenticated users. Set this to `true` to
-     * skip that behavior and always serve the main entry point instead.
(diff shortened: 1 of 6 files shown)


freben (approved):

plugins/app-backend/config.d.ts
```diff
@@ -42,5 +42,14 @@ export interface Config {
...
     disableStaticFallbackCache?: boolean;
+
+    /**
+     * Disables the experimental public entry point used for unauthenticated
+     * sign-in pages. When the app is bundled with an
+     * `index-public-experimental` entry point, the app backend will
+     * automatically serve it to unauthenticated users. Set this to `true` to
+     * skip that behavior and always serve the main entry point instead.
+     */
+    disableExperimentalPublicEntryPoint?: boolean;
```
Rugvip: Thinking we can perhaps call this simple `disablePublicEntryPoint` after all? Given that the build artifact is the `public` directory, it kinda matches that naming, and there's no experimental in there. The main experimental bit is the entry point picked up by the CLI, rather than the build output

Rugvip (approved):
:shipit:

Issue #34395, opened 2026-05-26
Bug: `@backstage/integration@2.0.2` causes GitLab requests to fail with 406 Not Acceptable on self-hosted GitLab


agayev169:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
gitlab integration, techdocs issue, 406, upgrade, v1.51.0
```

### 🗃️ Project Area

TechDocs

### 🔗 External Integration

GitLab

### Summary

In `@backstage/integration@2.0.2`, the `fetch` call inside `GitLabIntegration.createFetchStrategy()` had `mode: "same-origin"` removed. On Node 22, removing this option causes `undici` to set request headers (likely `Sec-Fetch-Mode: cors` and/or `Origin`) that some GitLab deployments — specifically self-hosted GitLab behind nginx — reject with **HTTP 406 Not Acceptable**.

This breaks TechDocs, `UrlReader.readTree`, and any other code path that goes through `GitLabIntegration.fetch`.

### Affected versions

- `@backstage/integration` **2.0.2** (broken)
- `@backstage/integration` **2.0.1** (works)

Pulled in transitively by `@backstage/backend-defaults@0.17.1` (which bumped its `@backstage/integration` peer from `^2.0.1` to `^2.0.2`).

### The change

In `dist/gitlab/GitLabIntegration.cjs.js`:

```diff
   createFetchStrategy() {
-    let fetchFn = async (url, options) => {
-      return fetch(url, { ...options, mode: "same-origin" });
-    };
+    let fetchFn = (url, options) => fetch(url, options);
```

This is the only GitLab-relevant behavioral change between 2.0.1 and 2.0.2.

### Reproduction

Self-hosted GitLab (any version recent enough to surface this; ours is behind nginx). Node 22.

```js
// verify-gitlab-fetch.mjs
const URL_TO_TEST = 'https://gitlab.example.com/api/v4/projects/group%2Frepo/repository/archive?sha=<branch>&path=<dir>';
const TOKEN = '<gitlab-token>';

async function hit(label, init) {
  const res = await fetch(URL_TO_TEST, { ...init, headers: { Authorization: `Bearer ${TOKEN}` } });
  console.log(label, '->', res.status, res.statusText);
}

await hit('WITH    mode:"same-origin"', { mode: 'same-origin' });
[...]

PR #34415, opened 2026-05-27
bugfix(gitlabUrlReader): Fix issue with repository archive retrieval

stijnbrouwers:

This pull requests solves #34395 

### DIFF
--- a/packages/backend-defaults/src/entrypoints/urlReader/lib/GitlabUrlReader.ts
+++ b/packages/backend-defaults/src/entrypoints/urlReader/lib/GitlabUrlReader.ts
@@ -211,7 +211,7 @@ export class GitlabUrlReader implements UrlReaderService {
     )}/repository/archive?${archiveReqParams.toString()}`;
     const archiveGitLabResponse = await this.integration.fetch(reqUrl, {
       ...getGitLabRequestOptions(this.integration.config, token),
-      mode: `same-origin`,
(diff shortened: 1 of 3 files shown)


packages/backend-defaults/src/entrypoints/urlReader/lib/GitlabUrlReader.ts
```diff
@@ -211,6 +211,7 @@ export class GitlabUrlReader implements UrlReaderService {
     )}/repository/archive?${archiveReqParams.toString()}`;
     const archiveGitLabResponse = await this.integration.fetch(reqUrl, {
       ...getGitLabRequestOptions(this.integration.config, token),
+      mode: 'same-origin',
```
freben, line 217: Could you add a comment above this line, saying why it's here and linking the issue? This will be easy to think that it should be removed again, if you don't know of the context.

freben (commented):
Thanks. If you make a file in the .patches directory, I can have it shipped in a patch release sooner than if we'd wait for mainline.

packages/backend-defaults/src/entrypoints/urlReader/lib/GitlabUrlReader.ts
stijnbrouwers: Both the patch file is created and comment added.

freben (approved):

PR #34416, opened 2026-05-27
fix: restore runtime dependencies incorrectly demoted to devDependencies

freben:
## Summary

PR #33936 removed duplicate `dependencies`/`devDependencies` entries, but two packages had dependencies incorrectly left only in `devDependencies` when they should be runtime dependencies. These packages re-export types from the demoted deps in their published API surface, making them required at build time for consumers.

**Fixed packages:**

| Package | Dependency | Why it's needed at runtime |
|---------|-----------|---------------------------|
| `@backstage/catalog-client` | `@backstage/plugin-catalog-common` | `AnalyzeLocationRequest` and `AnalyzeLocationResponse` types are re-exported in both `report.api.md` and `report-testUtils.api.md` |
| `@backstage/frontend-plugin-api` | `@backstage/config` | `Config` type is re-exported in `report.api.md` |

**How this was found:** scanned all `report*.api.md` files for import statements referencing packages that are only in `devDependencies` (not in `dependencies` or `peerDependencies`).

## Test plan

- [x] `yarn tsc` clean
- [x] No API report changes needed (the types were already in the reports)

### DIFF
--- a/.changeset/fix-missing-runtime-deps.md
+++ b/.changeset/fix-missing-runtime-deps.md
@@ -1,0 +1,6 @@
+---
+'@backstage/catalog-client': patch
+'@backstage/frontend-plugin-api': patch
+---
+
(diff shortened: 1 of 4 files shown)


benjdlambert (approved):

awanlin (approved):
Thanks!

PR #34418, opened 2026-05-27
Redesign CatalogFilterBlueprint to model-based API

benjdlambert:

Redesigns the `CatalogFilterBlueprint` alpha API from a component-rendering model to a declarative, model-based approach. Instead of each filter extension providing its own React component via a `loader`, filters now describe themselves as data and the catalog page controller handles rendering using Backstage UI components.

### Three filter param variants

**Facet-based** — provide a label, entity path, and selection mode. The catalog page fetches available values via catalog facets and renders a BUI `Select`:
```ts
CatalogFilterBlueprint.make({
  name: 'tag',
  params: { label: 'Tags', path: 'metadata.tags', mode: 'multi' },
});
```

**Options-based** — provide static options with a `toFilter` function and optional API dependency injection:
```ts
CatalogFilterBlueprint.make({
  name: 'processing-status',
  params: {
    label: 'Processing Status',
    mode: 'multi',
    options: [
      { label: 'Is Orphan', value: 'orphan' },
      { label: 'Has Error', value: 'error' },
    ],
    toFilter(selected) {
      return {
        getCatalogFilters() { return { ... }; },
        filterEntity(entity) { return ...; },
      };
    },
  },
});
```

**Deprecated loader** — the old component-based escape hatch for filters that don't fit a declarative model (e.g. `UserListPicker`, `EntityOwnerPicker`):
```ts
CatalogFilterBlueprint.make({
  name: 'list',
  params: { loader: async () => <UserListPicker /> },
});
```

### What changed

- `@backstage/plugin-catalog-react`: New blueprint API with `filterDescriptor` data ref and exported descriptor types
- `@backstage/plugin-catalog`: Catalog page reads descriptors and renders `FacetFilterPicker` / `OptionsFilterPicker` using BUI `Select`, or passes through custom elements
- 5 filters converted to facet model (kind, type, tags, lifecycle, namespace)
- 1 filter converted to options model (processing-status)
- 2 filters remain as deprecated loader (owner, user-list)

### DIFF
--- a/plugins/catalog-react/src/alpha/blueprints/CatalogFilterBlueprint.ts
+++ b/plugins/catalog-react/src/alpha/blueprints/CatalogFilterBlueprint.ts
@@ -31,6 +31,7 @@ import { EntityFilter } from '../../types';
 export interface CatalogFacetFilterDescriptor {
   type: 'facet';
   label: string;
+  filterKey: string;
   path: string;
   mode: 'single' | 'multi';
   defaultValue?: string | string[];
@@ -45,6 +46,7 @@ export interface CatalogFacetFilterDescriptor {
 export interface CatalogOptionsFilterDescriptor {
   type: 'options';
   label: string;
+  filterKey: string;
   mode: 'single' | 'multi';
   defaultValue?: string | string[];
   options: Array<{ label: string; value: string }>;
@@ -100,12 +102,14 @@ export const CatalogFilterBlueprint = createExtensionBlueprint({
     params:
       | {
           label: string;
+          filterKey?: string;
           path: string;
           mode: 'single' | 'multi';
           defaultValue?: string | string[];
         }
       | {
           label: string;
+          filterKey?: string;
           mode: 'single' | 'multi';
           defaultValue?: string | string[];
           options: Array<{ label: string; value: string }>;
@@ -134,6 +138,7 @@ export const CatalogFilterBlueprint = createExtensionBlueprint({
         catalogFilterDescriptorDataRef({
           type: 'options',
           label: params.label,
+          filterKey: params.filterKey ?? params.label,
           mode: params.mode,
           defaultValue: params.defaultValue,
           options: params.options,
@@ -146,6 +151,7 @@ export const CatalogFilterBlueprint = createExtensionBlueprint({
       catalogFilterDescriptorDataRef({
         type: 'facet',
         label: params.label,
+        filterKey: params.filterKey ?? params.path,
         path: params.path,
         mode: params.mode,
         defaultValue: params.defaultValue,
--- a/plugins/catalog-react/src/alpha/blueprints/CatalogFilterBlueprint.ts
+++ b/plugins/catalog-react/src/alpha/blueprints/CatalogFilterBlueprint.ts
@@ -16,23 +16,140 @@
 
 import {
   ExtensionBoundary,
-  coreExtensionData,
   createExtensionBlueprint,
+  createExtensionDataRef,
 } from '@backstage/frontend-plugin-api';
+import { ApiRef } from '@backstage/core-plugin-api';
+import { EntityFilter } from '../../types';
 
 /**
- * Creates Catalog Filter Extensions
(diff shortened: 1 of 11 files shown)


plugins/catalog/src/alpha/filters.tsx
```diff
@@ -75,38 +85,40 @@ const catalogModeCatalogFilter = CatalogFilterBlueprint.makeWithOverrides({
...
+          if (selected.includes('orphan')) {
+            const orphan = entity.metadata.annotations?.['backstage.io/orphan'];
+            if (orphan !== 'true') return false;
+          }
+          if (selected.includes('error')) {
+            const status = (entity as any)?.status?.items;
+            if (!status || status.length === 0) return false;
+          }
+          return true;
+        },
```
Rugvip, line 123: Is it realistic to skip the non-paginated approach? 🤔

plugins/catalog/src/alpha/filters.tsx
```diff
@@ -75,38 +85,40 @@ const catalogModeCatalogFilter = CatalogFilterBlueprint.makeWithOverrides({
...
+        toQueryValue() {
+          return selected;
+        },
```
Rugvip: This one confuses me a lil bit, seems like something that best to have hardcoded? 🤔, would need a way to parse it otherwise as well?

plugins/catalog/src/alpha/pages.tsx
```diff
@@ -69,7 +70,9 @@ export const CatalogExportConfigBlueprint = createExtensionBlueprint({
 
 export const catalogPage = PageBlueprint.makeWithOverrides({
   inputs: {
-    filters: createExtensionInput([coreExtensionData.reactElement]),
+    filters: createExtensionInput([
+      CatalogFilterBlueprint.dataRefs.filterDescriptor,
+    ]),
```
Rugvip, line 77: We'll want to keep this one backwards compatible in some way x)

plugins/catalog/src/alpha/pages.tsx
```diff
@@ -105,10 +108,42 @@ export const catalogPage = PageBlueprint.makeWithOverrides({
...
+                />
+              );
+            case 'facet':
+              return (
+                <FacetFilterPicker
+                  key={descriptor.path}
+                  filterKey={descriptor.path}
+                  {...descriptor}
+                />
+              );
```
Rugvip: Hold up, none of these are used in this PR right?

plugins/catalog-react/src/alpha/blueprints/CatalogFilterBlueprint.ts
```diff
@@ -16,23 +16,140 @@
...
+}
+
+/**
+ * Describes a catalog filter that provides its own rendering.
+ * @alpha
+ */
+export interface CatalogCustomFilterDescriptor {
+  type: 'custom';
+  element: JSX.Element;
+}
```
Rugvip, line 67: Yep seems useful to have but also unfortunate at the same time. Happy to keep but a potential option would be to force an override of the entire page intead.

plugins/catalog-react/src/alpha/blueprints/CatalogFilterBlueprint.ts
```diff
@@ -16,23 +16,140 @@
...
+          options: Array<{ label: string; value: string }>;
+          deps?: Record<string, ApiRef<unknown>>;
+          toFilter(
+            selected: string[],
+            deps: Record<string, unknown>,
+          ): EntityFilter | undefined | Promise<EntityFilter | undefined>;
+        }
+      | {
+          /** @deprecated Use the model-based params instead. */
+          loader: () => Promise<JSX.Element>;
```
Rugvip, line 124: Oh wait, so it _is_ backwards compatible 😁, nice. Just still need to have the page itself still accept the old filters for things built with the old blueprint.

Also still worth having the discussion around whether we want this at all

Rugvip (changes requested):
💪

PR #34419, opened 2026-05-27
feat(scaffolder): add template output redirect functionality to the n…

jkomendant:
…ew frontend system

Previously in the old frontend system, there was a way to attach custom template outputs components via the EXPERIMENTAL_TemplateOutputsComponent flag in the ScaffolderPage. The behind the scenes code for TemplateOutputsComponent is still there in the TasksSubPage in the new frontend system. This is just exposing a way to attach a custom TemplateOutputsComponent to the TasksSubPage in the new frontend system using a blueprint.


jkomendant:
@benjdlambert I know the contributing guidelines said not to ping someone directly for a merge review, but for some reason the github copilot review bot is still in awaiting requested review status 2 weeks later. I don't have an option to ping it or retrigger it. Any help to do so would be appreciated.

oliverdirks:
hi @jkomendant, can u maybe try to resolve the conflicts and give the review process another ping?

jkomendant:
@oliverdirks I rebased from master and resolved the merge conflicts. I am not sure what I should type to ping the review bot, because when doing @copilot, nothing shows up. If you have any tips on what the correct ping should be, it'd be greatly appreciated.

oliverdirks:
@jkomendant  Thanks! I think they need to approve that the "copilot-review" starts :)

jkomendant:
@oliverdirks thanks! I had raised another PR a day before this one where copilot had immediately started reviewing it, so had assumed that it maybe had glitched out on this PR. I am hoping that they'll eventually come around to this PR and kick it off and review it :)

awanlin (changes requested):
Thanks for the contribution @jkomendant, and sorry for the delay on the review. While this looks technically sound I'm not sure this is a pattern we want to encourage people to follow. It's a jarring experience being taken away unexpectedly and you lose out on potentially a lot of important information with this. 

I'm not against having the primitives to do this and I don't feel this pattern should be encourage in the docs explicitly. I'd also like to get feedback from the Scaffolder maintainers, please bring this to the next [Scaffolder SIG](https://github.com/backstage/community/tree/main/sigs/sig-scaffolder) to help move things along 👍

jkomendant:
Hi @awanlin, no problem at all. I don't mind removing the documentation for this. I know that it wasn't documented in the legacy frontend system either. We do have one very popular template that uses the redirect functionality via the legacy frontend system and losing the redirect hinders the user experience for it.

jkomendant:
I have removed the documentation in my latest commit

plugins/scaffolder/src/alpha/extensions.tsx
```diff
@@ -100,13 +101,28 @@ export const scaffolderTemplatesSubPage = SubPageBlueprint.makeWithOverrides({
...
-      import('./components/TasksSubPage').then(m => <m.TasksSubPage />),
+  inputs: {
+    templateOutputsComponents: createExtensionInput(
+      [scaffolderTemplateOutputsComponentRef],
+      { optional: true },
+    ),
+  },
+  factory(originalFactory, { inputs }) {
+    return originalFactory({
+      path: '/tasks',
```
awanlin: This was just `task` before do we need the leading `/`?

plugins/scaffolder/src/alpha/extensions.tsx
```diff
@@ -100,13 +101,28 @@ export const scaffolderTemplatesSubPage = SubPageBlueprint.makeWithOverrides({
...
+      title: 'Tasks',
+      loader: async () => {
+        const TemplateOutputsComponent =
+          inputs.templateOutputsComponents?.[0].get(
```
awanlin: Not sure how we call this out but if move than one gets registered only the first gets used here. Maybe a TSDoc comment somewhere could help with that? Or should the blueprint be changes to only support one?

awanlin (changes requested):
Thanks for the follow up @jkomendant, got a pair of comments to help move this forward.

plugins/scaffolder/src/alpha/extensions.tsx
jkomendant: I did not see any behavior changes with or without the leading /, so removed it to keep it consistent with how it was prior and with the other scaffolder extensions

plugins/scaffolder/src/alpha/extensions.tsx
jkomendant: @awanlin this is a good catch! I have changed it from only recognizing the first redirect, to now allowing multiple redirects using the blueprint. I have tested this locally as well with 2 templates using 2 redirects. the rewrite is also cleaner now and you can also use the same redirect for multiple templates if so desired.

The cleaner version of using the blueprint now which has the redirect being passed as the component:

Image

and being able to register it in the app:

Image

awanlin: Nice, thanks for the refactor!

awanlin (approved):
Nice work @jkomendant, just need approve from the Scaffolder Maintainers now, will try and get them to give this a look 👍

PR #34425, opened 2026-05-28
fix(catalog-backend-module-msgraph): filter disabled group members client-side

### LINKED ISSUE #34422: Msgraph plugin userGroupMember filter causes error on graph api with user filter
### 📜 Issue Labels

### 🔎 Search Terms

```plain
msgraph catalog entity provider userGroupMember filter accountEnabled eq true
```

### 🗃️ Project Area

Catalog

### 🔗 External Integration

Azure

### 📝 Description & Context

With 1.51.0 the `catalog-backend-module-msgraph` plugin always sets the `user.filter` option to `accountEnabled eq true`, introduces with https://github.com/backstage/backstage/pull/34165. When using `userGroupMember.filter` to load users based on group membership, this causes an error on the graph API: `The specified filter to the reference property query is currently not supported`. The implicit user filter is now always appended to userGroupMember requests like this `https://graph.microsoft.com/v1.0/groups/${groupId}/members/microsoft.graph.user?$filter=accountEnabled eq true`.

Previously there was also a check so that the `user.filter` and `userGroupMember.filter` were mutually exclusive that is now removed, maybe it was because of this issue?

Example app-config we are using for the provider:
```yaml
    microsoftGraphOrg:
      default:
        tenantId: someTenantId
        user:
          select:
            [
              "id",
              "displayName",
              "mail",
              "jobTitle",
              "onPremisesExtensionAttributes",
            ]
        userGroupMember:
          filter: "displayName in ('AllProductDevelopmentMembers', 'SomeOtherGroup2', 'SomeOtherGroup3')"
        group:
          filter: >-
            securityEnabled eq true and mailEnabled eq true and
            (description eq 'DescriptionIdentifyingProductDevelopmentGroups'
            OR displayName eq 'SomeOtherGroup2'
            OR displayName eq 'SomeOtherGroup3'
          search: '"displayName:SomeIdentifier"'
```

### 👍 Expected Behavior

The msgraph entity provider can be used with `userGroupMember.filter` with no error.

### 📦 Reproduction Repo

_No response_

### 🥾 Reproduction steps

With Microsoft Graph Explorer:
1. Go to https://developer.microsoft.com/en-us/graph/graph-explorer
2. Log in with your Microsoft account with sufficient permissions on Entra ID
3. Enter the following URL with a valid groupId: `https://graph.microsoft.com/v1.0/groups/${groupId}/members/microsoft.graph.user?$filter=accountEnabled eq true`

In Backstage:
1. Setup msgraph catalog plugin module
2. Set a `userGroupMember.filter` (example app-config as in Description & Context above)
3. Try to run the catalog entity provider

freben:
## Summary

Fixes #34422

Reverts most of #34165 and replaces the server-side `accountEnabled eq true` filter with client-side filtering that works across all code paths.

### Problem

PR #34165 added a `buildUserFilter` function that always prepended `accountEnabled eq true` to the user filter. This filter was then passed to both:
- `GET /users?$filter=...` — works fine
- `GET /groups/{id}/members/microsoft.graph.user?$filter=...` — fails with "The specified filter to the reference property query is currently not supported"

The PR also removed the mutual exclusivity check between `userFilter` and `userGroupMemberFilter`, which existed precisely because these target different endpoints with different filter capabilities.

### Fix

1. **Revert `buildUserFilter`** — read `user.filter` / `userFilter` as-is from config, no automatic base filter
2. **Restore mutual exclusivity checks** between `userFilter` and `userGroupMemberFilter`/`userGroupMemberSearch`
3. **Filter disabled users client-side** via `filterDisabledUsers()` in both the `/users` and group members paths — skips users with `accountEnabled === false`
4. **Ensure `accountEnabled` is in `$select`** when a custom `user.select` is configured, so the field is available for client-side filtering
5. **Revert config docs** to show `accountEnabled eq true` as a user-provided filter example, not an automatic base

### Test plan

- [x] All 119 msgraph tests pass (2 new tests for disabled user filtering)
- [x] Config tests updated to reflect reverted behavior

### DIFF
--- a/plugins/catalog-backend-module-msgraph/config.d.ts
+++ b/plugins/catalog-backend-module-msgraph/config.d.ts
@@ -59,9 +59,11 @@ export interface Config {
           // they could also be configured "in code".
 
           /**
-           * The filter to apply to extract users.
+           * The filter to apply to extract users. Disabled users
+           * (accountEnabled !== true) are always filtered out client-side
+           * regardless of this setting.
            *
-           * E.g. "accountEnabled eq true and userType eq 'member'"
+           * E.g. "userType eq 'member'"
            */
           userFilter?: string;
           /**
--- a/plugins/catalog-backend-module-msgraph/config.d.ts
+++ b/plugins/catalog-backend-module-msgraph/config.d.ts
@@ -60,9 +60,8 @@ export interface Config {
 
           /**
            * The filter to apply to extract users.
-           * Combined with the base `accountEnabled eq true` filter.
            *
-           * E.g. "userType eq 'member'"
+           * E.g. "accountEnabled eq true and userType eq 'member'"
            */
           userFilter?: string;
           /**
--- a/plugins/catalog-backend-module-msgraph/config.d.ts
+++ b/plugins/catalog-backend-module-msgraph/config.d.ts
@@ -164,7 +164,8 @@ export interface Config {
               expand?: string;
               /**
                * The filter to apply to extract users.
-               * Combined with the base `accountEnabled eq true` filter
+               * Disabled users (accountEnabled === false) are always
+               * filtered out client-side regardless of this setting.
                *
                * E.g. "userType eq 'member'"
                */
@@ -298,7 +299,8 @@ export interface Config {
                 expand?: string;
                 /**
                  * The filter to apply to extract users.
-                 * Combined with the base `accountEnabled eq true` filter
+                 * Disabled users (accountEnabled === false) are always
+                 * filtered out client-side regardless of this setting.
                  *
                  * E.g. "userType eq 'member'"
                  */
--- a/plugins/catalog-backend-module-msgraph/config.d.ts
+++ b/plugins/catalog-backend-module-msgraph/config.d.ts
@@ -60,8 +60,8 @@ export interface Config {
 
           /**
            * The filter to apply to extract users. Disabled users
-           * (accountEnabled !== true) are always filtered out client-side
-           * regardless of this setting.
+           * (`accountEnabled === false`) are always filtered out
+           * client-side regardless of this setting.
            *
            * E.g. "userType eq 'member'"
            */
@@ -164,7 +164,7 @@ export interface Config {
               expand?: string;
               /**
                * The filter to apply to extract users.
-               * Disabled users (accountEnabled === false) are always
+               * Disabled users (`accountEnabled === false`) are always
                * filtered out client-side regardless of this setting.
                *
                * E.g. "userType eq 'member'"
@@ -299,7 +299,7 @@ export interface Config {
                 expand?: string;
                 /**
                  * The filter to apply to extract users.
-                 * Disabled users (accountEnabled === false) are always
+                 * Disabled users (`accountEnabled === false`) are always
                  * filtered out client-side regardless of this setting.
                  *
                  * E.g. "userType eq 'member'"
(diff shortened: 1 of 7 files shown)


mtlewis:
Seems like maybe we should go further than this and revert most of #34165 in favour of client-side filtering. Seems it was wrong to remove the mutual exclusivity in that PR, and I'd like to get it back in ASAP.

plugins/catalog-backend-module-msgraph/src/microsoftGraph/config.ts
```diff
@@ -61,10 +61,9 @@ export type MicrosoftGraphProviderConfig = {
    */
   clientSecret?: string;
   /**
-   * The filter to apply to extract users. This is combined with the base
-   * `accountEnabled eq true` filter that is always applied automatically.
+   * The filter to apply to extract users.
    *
-   * E.g. "userType eq 'member'"
+   * E.g. "accountEnabled eq true and userType eq 'member'"
```
mtlewis:
```suggestion
   * E.g. userType eq 'member'"
```

plugins/catalog-backend-module-msgraph/src/microsoftGraph/config.ts
```diff
@@ -61,10 +61,9 @@ export type MicrosoftGraphProviderConfig = {
    */
   clientSecret?: string;
   /**
-   * The filter to apply to extract users. This is combined with the base
-   * `accountEnabled eq true` filter that is always applied automatically.
+   * The filter to apply to extract users.
```
mtlewis: Let's mention the client-side filtering here and leave it out of the example maybe? There's no point passing that filter now.

plugins/catalog-backend-module-msgraph/src/microsoftGraph/read.ts
```diff
@@ -42,6 +42,27 @@ import { LoggerService } from '@backstage/backend-plugin-api';
...
+    return select;
+  }
+  return [...select, field];
+}
+
+async function* filterDisabledUsers(
+  users: AsyncIterable<MicrosoftGraph.User>,
+): AsyncIterable<MicrosoftGraph.User> {
+  for await (const user of users) {
+    if (user.accountEnabled !== false) {
```
mtlewis, line 80:
```suggestion
    if (user.accountEnabled === true) {
```
nit, but perhaps better to match the filtering logic that was recommended/defaulted before, and also be picky about what users we include in case of surprising stuff (thinking of examples like `user.accountEnabled` turning out to be `"false"` or something).

plugins/catalog-backend-module-msgraph/config.d.ts
```diff
@@ -60,9 +60,8 @@ export interface Config {
 
           /**
            * The filter to apply to extract users.
-           * Combined with the base `accountEnabled eq true` filter.
```
mtlewis, line 63: Ditto

plugins/catalog-backend-module-msgraph/src/microsoftGraph/read.ts
```diff
@@ -58,16 +79,21 @@ export async function readMicrosoftGraphUsers(
...
-    {
-      filter: options.userFilter,
-      expand: options.userExpand,
-      select: options.userSelect,
-      top: PAGE_SIZE,
-    },
-    options.queryMode,
-    options.userPath,
-    options.signal,
+  const select = options.userSelect
```
mtlewis: do we not need to specify `accountEnabled` even if `options.userSelect` is undefined? Also wonder if we can tweak the params and return value of this function so that we can inline it into the getUsers call, simpler repeated code:
```
client.getUsers(
  {
    // ...
    select: ensureSelectContains(options.userSelect, 'accountEnabled'),
```

mtlewis (approved):

vinzscam (approved):

Issue #34211, opened 2026-05-13
[TechDocs] Add permission framework integration for documentation access control


nandni-g:
### 📜 Issue Labels

- [x] Please familiarize yourself with the issue labels used in this project: [LABELS.md](https://github.com/backstage/backstage/blob/master/LABELS.md)

### 🔎 Search Terms

```plain
techdocs permissions, techdocs access control, techdocs permission framework, techdocs entity read, docs tab restrict access, techdocs backend authorize
```

### 🗃️ Project Area

TechDocs

### 🔖 Need

Currently the TechDocs backend serves static documentation assets (HTML, JS, assets) directly from object storage (S3, GCS, Azure Blob) without consulting the Backstage permission framework. This means it is **impossible** to restrict access to documentation content independently from catalog entity metadata visibility.

This gap affects any organization that needs to host documentation for private or sensitive repositories in a shared Backstage deployment. Use cases include:

- **Private repository docs**: Teams with internal-only source code want TechDocs visible only to entity owners, while still allowing the catalog entry (name, description, owner, tags) to be discoverable by everyone.
- **Compliance-sensitive content**: Security runbooks, architecture decision records, or regulated-product documentation should be readable only by authorized groups.
- **Multi-tenant deployments**: Organizations running a single Backstage instance for many teams need per-entity content gating, not just catalog-metadata gating.

Today the only workaround is a **frontend-only gate** — a component that hides the Docs tab in the UI. This does not protect the underlying docs API endpoint (`/api/techdocs/...`) or the direct S3/storage asset URLs, so it is not a genuine access control mechanism.

`catalog.entity.read` already exists and gates the catalog API correctly. TechDocs should offer an equivalent so that documentation and metadata can have independent visibility policies.

### 📝 Proposal

Add a `techdocs.entity.read` resource permission to the TechDocs plugin, using `catalog-entity` as the resource type (consistent with how `catalog.entity.read` works).

### Permission definition

In `@backstage/plugin-techdocs-common` (or a new `@backstage/plugin-techdocs-node`), export:

```
import { createPermission } from '@backstage/plugin-permission-common';

export const techDocsEntityReadPermission = createPermission({
  name: 'techdocs.entity.read',
  attributes: { action: 'read' },
  resourceType: 'catalog-entity',
});

export const techDocsPermissions = [techDocsEntityReadPermission];
```

Backend integration
In @backstage/plugin-techdocs-backend, before serving any documentation content (HTML, assets, search index), call:

```
[...]

PR #34437, opened 2026-05-29
Add techdocs.entity.read permission for access control

nandni-g:
### Description

Hey! 👋 

I've been working on adding permission support to TechDocs https://github.com/backstage/backstage/issues/34211 so teams can control who can access their documentation.

**The problem we're solving:**
Right now, TechDocs serves docs directly from storage without checking permissions. This means if someone can see an entity in the catalog, they can also read all its docs — even if those docs contain sensitive information like security runbooks or internal architecture decisions.

**What this PR does:**
Introduces a new `techdocs.entity.read` permission that lets you gate access to documentation separately from catalog visibility. When permissions are enabled in your Backstage config, all TechDocs endpoints will check this permission before serving any content.

This is useful for:
- Private repo docs that should only be visible to entity owners
- Compliance-sensitive documentation (security runbooks, regulated-product docs)
- Multi-tenant Backstage deployments where teams need isolated doc access

The permission uses `catalog-entity` as the resource type, so it works seamlessly with existing permission policies.

**Changes:**
- New `techDocsEntityReadPermission` exported from `@backstage/plugin-techdocs-common`
- Permission checks added to all TechDocs backend routes (metadata, sync, and static assets)
- Tests covering the permission flow
- Documentation added to TechDocs how-to guides


kunickiaj (commented):
First off, this is a great feature and one we have wanted to add in our own organization. Separating TechDocs visibility from catalog entity visibility fills a real gap, and the overall direction looks good.

I found two concerns worth addressing:

1. TechDocs search results are still authorized with `catalog.entity.read`, despite containing indexed TechDocs page text. A user allowed to view an entity but denied `techdocs.entity.read` could therefore still retrieve restricted documentation through search.
2. I saw the resolved Copilot thread about static-asset authorization overhead. After tracing the current implementation, I believe the concern still applies: the middleware authorizes every asset independently. A page with N assets causes N permission-backend requests; with a conditional policy, it also causes N catalog `apply-conditions` requests. Existing batching operates only within one authorization request and does not combine separate browser requests.

The second issue should not be solved by authorizing only the HTML entry point because assets, including the search index, are directly addressable. A short-lived authorization cache scoped by credentials, entity reference, and permission appears consistent with the existing `CachedEntityLoader` pattern.

plugins/techdocs-backend/src/service/router.test.ts
```diff
@@ -348,15 +424,41 @@ data: {"updated":true}
 
       expect(response.status).toBe(200);
       expect(MockCachedEntityLoader.prototype.load).toHaveBeenCalled();
+      expect(permissions.authorize).toHaveBeenCalled();
```
kunickiaj, line 427: Could this assert the exact authorization request rather than only that the mock was called? In particular, it would be useful to pin `techDocsEntityReadPermission`, the canonical `component:default/test` resource reference, and the request credentials. Those are the security-relevant parts of this behavior.

plugins/techdocs-common/src/constants.ts
```diff
@@ -21,3 +21,17 @@ export const TECHDOCS_EXTERNAL_ANNOTATION = 'backstage.io/techdocs-entity';
...
+ * should be restricted. A permission policy can read this annotation to decide
+ * whether to apply additional access control (for example, limiting access to
+ * entity owners) via the `techdocs.entity.read` permission.
+ *
+ * The annotation is only a signal; the actual authorization decision is made by
+ * the permission policy. Entities without this annotation are unaffected.
+ *
+ * @public
+ */
+export const TECHDOCS_VISIBILITY_ANNOTATION =
```
kunickiaj, line 36: Since this introduces a public, well-known annotation, could it also be added to `docs/features/software-catalog/well-known-annotations.md` alongside the existing TechDocs annotations?

plugins/techdocs-backend/src/service/router.test.ts
```diff
@@ -327,16 +398,21 @@ data: {"updated":true}
       expect(MockTechDocsCache.get).toHaveBeenCalled();
     });
 
-    it('should check entity access when permissions are enabled', async () => {
+    it('should check entity access and TechDocs permission', async () => {
```
kunickiaj, line 401: Could this test group also cover the compatibility path where `permission.enabled` is absent or false and assert that `permissions.authorize` is not called? That is an important behavior promised by the changeset.

docs/features/techdocs/architecture.md
```diff
@@ -136,6 +136,7 @@ process, making sure it's scalable. It is responsible for securely communicating
 with the cloud storage provider, for both fetching static generated sites and
 publishing the updates. There are other planned features like an authentication
 layer for users to determine whether they have the permission to view a
-particular docs site. There are a handful of features which are extremely hard
+particular docs site. See the [how-to guide on restricting access](./how-to-guides.md#how-to-restrict-access-to-techdocs-using-permissions) for details.
```
kunickiaj, line 139: This still describes per-site authorization as a planned feature, but this PR implements it. Could we rewrite the sentence rather than appending the how-to link to the existing planned-features wording?

.changeset/techdocs-permission-access-control.md
```diff
@@ -0,0 +1,10 @@
+---
+'@backstage/plugin-techdocs-common': minor
+'@backstage/plugin-techdocs-backend': minor
+---
+
+Added `techdocs.entity.read` permission to control access to TechDocs documentation independently from catalog entity visibility. When `permission.enabled` is true, all TechDocs endpoints now authorize requests using this permission before serving content.
```
kunickiaj, line 6: Could this PR also update `DefaultTechDocsCollatorFactory` to use `techDocsEntityReadPermission` instead of `catalogEntityReadPermission`?

The indexed search documents include TechDocs page text, so the current permission creates an access-control gap: a user allowed `catalog.entity.read` but denied `techdocs.entity.read` can still receive restricted documentation through search.

`AuthorizedSearchEngine` already evaluates the collator visibility permission against each document entity resource reference, so switching the permission there looks like the smallest consistent fix. It would also be useful to cover that a TechDocs search result is excluded when the new permission is denied.

awanlin:
Hi @nandni-g, sorry for the miscommunication but we are waiting for the feature flag agreed to in the issue to be added here. Details start at this comment: https://github.com/backstage/backstage/issues/34211#issuecomment-5485580893. I've reopened the PR, if you can rebase and then make the needed changes that would be great. When you have time 🙏

nandni-g:
Hii @awanlin, no worries - flag's in now and I've rebased onto master, conflicts sorted.

It's all behind techdocs.experimentalTechdocsPermissions, off by default, and it drives the search collator too so nothing changes for existing adopters until they opt in. No reindex needed either, both permissions share the catalog-entity resource type so only the query-time check changes.

One thing though - #33412 now adds permissions to plugin-techdocs-common as well, and its techdocsPermissions array differs from mine (techDocsPermissions) by a single capital letter. Theirs is behind /alpha, mine's on the root export. Happy to move mine and merge the arrays, just let me know which way you'd like it.

PR #34446, opened 2026-05-30
refactor(devtools): migrate InfoDependenciesTable to Backstage UI

### LINKED ISSUE #32748: [DevTools] Migrate the `InfoDependenciesTable` component to Backstage Ui (BUI)
### 📝 Acknowledgement

### Content

The `InfoDependenciesTable` component can be migrated to BUI. 

The component code is here: https://github.com/backstage/backstage/blob/master/plugins/devtools/src/components/Content/InfoContent/InfoDependenciesTable.tsx

The list of available Backstage UI components is here: https://ui.backstage.io/

The process would be to just switch out the Table component from `@backstage/core-components` to use the Table from Backstage UI (BUI)

Naga15:
Fixes #32748

## Summary

Migrates `InfoDependenciesTable` (the Package Dependencies table on the DevTools `/devtools/info` page) from Material UI + `@backstage/core-components` `Table` to the new Backstage UI library.

## What's changing

- **Imports**: `Table` / `TableColumn` from `@backstage/core-components` → `Table`, `CellText`, `useTable`, `ColumnConfig` from `@backstage/ui`; added `Box`, `Flex`, `SearchField`, `Text` for the heading row.
- **Column config**: legacy `TableColumn[]` with `title` / `field` → `ColumnConfig<Row>[]` with `id`, `label`, `isRowHeader`, `isSortable`, and a `cell` renderer using `<CellText title={...} />`.
- **Row identity**: introduced a local `Row = PackageDependency & { id: string }` and a memoized `rows` array that maps `{ ...d, id: d.name }`, satisfying BUI's `TableItem` constraint (package names are unique).
- **Pagination, sort, search**: handled via `useTable({ mode: 'complete', data, paginationOptions, initialSort, sortFn, searchFn })` instead of the legacy table's built-in options.
- **Heading + filter**: rendered as a `<Flex justify="between" align="center">` row above the table — `<Text variant="title-medium" weight="bold">Package Dependencies</Text>` on the left, `<SearchField {...search} />` on the right.

## Behavior preserved

- Name + Versions columns
- Initial sort: name ascending
- Page size 15 (default), with 15 / 30 / 100 options
- Substring filter across both name and versions

`InfoDependenciesTable`'s public prop signature is unchanged, so `InfoContent` does not need any edits.

## Test plan

- [x] Local `yarn start` smoke test on `/devtools/info`: Package Dependencies table renders all rows, sorts by name ascending, filter narrows results live, pagination cycles correctly with 15/30/100 page sizes
- [ ] CI: lint, typecheck, tests, changeset verification

No tests exist for `InfoDependenciesTable` today; happy to add one if maintainers prefer.

### DIFF
--- a/.changeset/devtools-info-dependencies-table-bui.md
+++ b/.changeset/devtools-info-dependencies-table-bui.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-devtools': patch
+---
+
+Migrated the `InfoDependenciesTable` component from Material UI / `@backstage/core-components` `Table` to the new Backstage UI library. Behavior is preserved: name/versions columns, name-ascending initial sort, page sizes of 15/30/100, and substring filtering across both columns.
--- a/plugins/devtools/src/components/Content/InfoContent/InfoDependenciesTable.tsx
+++ b/plugins/devtools/src/components/Content/InfoContent/InfoDependenciesTable.tsx
@@ -50,7 +50,7 @@ export const InfoDependenciesTable = ({
   infoDependencies: PackageDependency[] | undefined;
 }) => {
   const rows = useMemo<Row[] | undefined>(
-    () => infoDependencies?.map(d => ({ ...d, id: d.name })),
+    () => infoDependencies?.map((d, i) => ({ ...d, id: `${d.name}-${i}` })),
     [infoDependencies],
   );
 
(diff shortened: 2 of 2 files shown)


Naga15:
Addressed the review feedback:

1. **Row id uniqueness** (line 53): agreed — switched to `${d.name}-${i}` in b0d4e43ba6 as defence in depth.
2. **`versions` field shape** (line 69): no change needed here. `PackageDependency.versions` is declared as `string` in [plugins/devtools-common/src/types.ts#L47-L50](https://github.com/backstage/backstage/blob/master/plugins/devtools-common/src/types.ts#L47-L50), and renders as a plain string (e.g. `0.0.0-use.local`) at runtime in the existing `/devtools/info` page. `.toLowerCase()` / `.localeCompare()` / `<CellText title={item.versions} />` are correct for that contract.

PR #34447, opened 2026-05-30
feat(catalog-backend): add refresh-catalog-entity action

### LINKED ISSUE #33729: mcp-action to refresh catalog entities
### 📜 Issue Labels

### 🔎 Search Terms

```plain
catalog, mcp, actions, scheduler
```

### 🗃️ Project Area

Catalog

### 🔖 Need

Add an action to the `mcp-actions-backend` that allows to schedule the refresh of an entity.
This is particularly useful when creating entities using MCP Server and wanting to access the new entity in the catalog.
For example, I generate entities using an MCP client by executing Backstage Templates. I'd like my agent to automatically refresh the entity to then access it in the catalog.

### 📝 Proposal

The new action can be registered in the `ActionsRegistryService`. 
This is very similar to the other [catalog actions](https://github.com/backstage/backstage/tree/master/plugins/catalog-backend/src/actions):

```
import { ActionsRegistryService } from '@backstage/backend-plugin-api/alpha';
import { stringifyEntityRef } from '@backstage/catalog-model';
import { CatalogService } from '@backstage/plugin-catalog-node';
import { InputError } from '@backstage/errors';

export const createRefreshCatalogEntityAction = ({
  catalog,
  actionsRegistry,
}: {
  catalog: CatalogService;
  actionsRegistry: ActionsRegistryService;
}) => {
  actionsRegistry.register({
    name: 'refresh-catalog-entity',
    title: 'Refresh Catalog Entity',
    attributes: {
      destructive: false,
      readOnly: false,
      idempotent: true,
    },
    description: `
This allows you to trigger a refresh of a single entity in the software catalog.
Each entity is identified by a unique entity reference of the form "kind:namespace/name".
The default namespace is "default".
    `,
    schema: {
      input: z =>
        z.object({
          kind: z
            .string()
            .describe(
              'The kind of the entity to refresh. If unknown it can be omitted.',
            )
            .optional(),
          namespace: z
            .string()
            .describe(
              'The namespace of the entity to query. If the namespace is unknown it can be omitted.',
            )
            .optional(),
          name: z.string().describe('The name of the entity to refresh'),
        }),
      output: z =>
        z.object({
          entityRef: z.string(),
          refreshed: z.boolean(),
        }),
    },
    action: async ({ input, credentials }) => {
      if (!input.name) {
        throw new InputError('Entity name is required');
      }

      const entityRef = stringifyEntityRef({
        kind: input.kind ?? 'Component',
        namespace: input.namespace ?? 'default',
        name: input.name,
      });

      await catalog.refreshEntity(entityRef, { credentials });

      return {
        output: {
          entityRef,
          refreshed: true,
        },
      };
    },
  });
};
```

### 🔄 Alternatives

The refresh could also happen on the catalog level, triggering the scheduler that process all entities. 
Nevertheless, an entity-level refresh should be needed.

Naga15:
Fixes #33729

## Summary

Registers a new `refresh-catalog-entity` action with the `ActionsRegistryService` in `plugins/catalog-backend`. `@backstage/plugin-mcp-actions-backend` auto-exposes any registered action as an MCP tool, so this becomes callable by MCP clients — e.g. an LLM agent invoking a Backstage scaffolder template can now requeue the resulting entity for processing before reading it back.

## Resolution model

Mirrors the existing `get-catalog-entity` action so LLM clients can move between them without kind gymnastics:

- `name` is required; `kind` and `namespace` are optional with no defaults.
- `CatalogService.queryEntities` filters on `metadata.name` plus any `kind`/`namespace` the caller supplied.
- `InputError` when no entity matches; `ConflictError` listing candidate `entityRef`s when a name matches multiple entities.
- `catalog.refreshEntity(stringifyEntityRef(items[0]))` is then called for the resolved entity.

The output drops the always-true `refreshed: boolean` from the original proposal — success is signalled by the promise resolving, and the returned canonical `entityRef` is the useful piece for the caller.

## Surface area

| File | Change |
|---|---|
| `plugins/catalog-backend/src/actions/createRefreshCatalogEntityAction.ts` | New action — `{ name, kind?, namespace? }` → `{ entityRef }`. |
| `plugins/catalog-backend/src/actions/index.ts` | Registers the new action in `createCatalogActions`. |
| `plugins/catalog-backend/src/actions/createRefreshCatalogEntityAction.test.ts` | Six unit tests: single-match, disambiguation by kind, disambiguation by kind+namespace, no match, multiple matches, error propagation. |
| `.changeset/catalog-refresh-entity-mcp-action.md` | `@backstage/plugin-catalog-backend` — minor. |

## Action attributes

- `destructive: false` — refresh does not delete data
- `readOnly: false` — it mutates catalog state (requeues processing)
- `idempotent: true` — refreshing twice is safe

## Test plan

- [x] Unit tests covering happy path, name-only resolution, disambiguation, no match, ambiguous match, and downstream error propagation.
- [ ] CI: lint, typecheck, tests, changeset verification.

### DIFF
--- a/.changeset/catalog-refresh-entity-mcp-action.md
+++ b/.changeset/catalog-refresh-entity-mcp-action.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': minor
+---
+
+Added a new `refresh-catalog-entity` action to the catalog backend's `ActionsRegistryService`. The action triggers a refresh of a single entity identified by `kind` / `namespace` / `name` (kind defaults to `Component`, namespace defaults to `default`). It performs an existence check via `getEntityByRef` first and throws `NotFoundError` for missing entities. Exposed automatically as an MCP tool by `@backstage/plugin-mcp-actions-backend`.
--- a/.changeset/catalog-refresh-entity-mcp-action.md
+++ b/.changeset/catalog-refresh-entity-mcp-action.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-backend': minor
 ---
 
-Added a `refresh-catalog-entity` action so agents and MCP clients can requeue a single entity for processing after creating or updating it — useful for reading back fresh data immediately after a scaffolder run without waiting for the next scheduled processing loop.
+Added a `refresh-catalog-entity` action so agents and MCP clients can re-queue a single entity for processing after creating or updating it — useful for reading back fresh data immediately after a scaffolder run without waiting for the next scheduled processing loop.
--- a/.changeset/catalog-refresh-entity-mcp-action.md
+++ b/.changeset/catalog-refresh-entity-mcp-action.md
@@ -2,4 +2,4 @@
 '@backstage/plugin-catalog-backend': minor
 ---
 
-Added a new `refresh-catalog-entity` action to the catalog backend's `ActionsRegistryService`. The action triggers a refresh of a single entity identified by `kind` / `namespace` / `name` (kind defaults to `Component`, namespace defaults to `default`). It performs an existence check via `getEntityByRef` first and throws `NotFoundError` for missing entities. Exposed automatically as an MCP tool by `@backstage/plugin-mcp-actions-backend`.
+Added a `refresh-catalog-entity` action so agents and MCP clients can requeue a single entity for processing after creating or updating it — useful for reading back fresh data immediately after a scaffolder run without waiting for the next scheduled processing loop.
--- a/plugins/catalog-backend/src/actions/createRefreshCatalogEntityAction.ts
+++ b/plugins/catalog-backend/src/actions/createRefreshCatalogEntityAction.ts
@@ -15,7 +15,7 @@
  */
 import type { ActionsRegistryService } from '@backstage/backend-plugin-api/alpha';
 import { stringifyEntityRef } from '@backstage/catalog-model';
-import { NotFoundError } from '@backstage/errors';
+import { ConflictError, InputError } from '@backstage/errors';
 import type { CatalogService } from '@backstage/plugin-catalog-node';
 
 export const createRefreshCatalogEntityAction = ({
@@ -33,11 +33,11 @@ export const createRefreshCatalogEntityAction = ({
       readOnly: false,
       idempotent: true,
     },
-    description: `Triggers a refresh of a single entity in the Backstage software catalog, requeuing it for processing.
-
-This is useful immediately after creating or updating an entity (for example, via a scaffolder template invoked by an MCP client) when the new data must be visible in the catalog before subsequent actions can read it.
-
-Each entity is identified by its kind, namespace, and name. The default kind is "Component" and the default namespace is "default".`,
+    description: `
+This allows you to trigger a refresh of a single entity in the software catalog, requeuing it for processing.
+This is useful immediately after creating or updating an entity (for example via a scaffolder template invoked by an MCP client) so the new data becomes visible before subsequent actions read it.
+Each entity in the software catalog has a unique name, kind, and namespace. If the name alone matches multiple entities, provide the kind (and namespace) to disambiguate.
+    `,
     schema: {
       input: z =>
         z.object({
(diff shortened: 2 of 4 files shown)


Naga15:
Pushed 469290b4ba addressing the review feedback:

1. **Spell check** (`requeueing` → `requeuing`): fixed in the action description.
2. **Type-only imports** (line 19): switched both `ActionsRegistryService` and `CatalogService` to `import type { ... }`.
3. **Empty / whitespace inputs** (line 57): added `.min(1)` to `kind`, `namespace`, and `name`, plus `.trim()` on `name` so whitespace-only values are rejected at the Zod layer with a clear error before `stringifyEntityRef` can produce a malformed ref.
4. **`getEntityByRef` call assertion**: added `expect(mockCatalog.getEntityByRef).toHaveBeenCalledWith(expectedRef, …)` to the two happy-path tests; the NotFoundError test already covered the missing-entity contract.

CI should re-run shortly.

plugins/catalog-backend/src/actions/createRefreshCatalogEntityAction.ts
```diff
@@ -0,0 +1,90 @@
...
+      output: z =>
+        z.object({
+          entityRef: z
+            .string()
+            .describe('The canonical entity reference that was refreshed.'),
+        }),
+    },
+    action: async ({ input, credentials }) => {
+      const entityRef = stringifyEntityRef({
+        kind: input.kind ?? 'Component',
```
freben: The existing `get-catalog-entity` action intentionally does not default kind or namespace. Instead it queries by name and surfaces ambiguity:

```
// get-catalog-entity: no defaults, handles multiple matches
const filter: Record<string, string> = { 'metadata.name': input.name };
if (input.kind) filter.kind = input.kind;
// ... queryEntities → ConflictError if >1 match
```

The new action takes the opposite approach; it silently defaults kind to `Component` and namespace to `default`, then does an exact `getEntityByRef` lookup. This creates a UX problem for MCP/LLM clients:

- Agent calls `get-catalog-entity({ name: 'orders-api' })` → gets back `api:default/orders-api`
- Agent calls `refresh-catalog-entity({ name: 'orders-api' })` → tries `component:default/orders-api` → `NotFoundError`

The caller gets a confusing failure with no hint that the entity exists under a different kind. The `get-catalog-entity` approach is better for LLM consumers — query without assuming kind, handle ambiguity explicitly. At minimum, the refresh action should follow the same resolution

.changeset/catalog-refresh-entity-mcp-action.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': minor
+---
+
+Added a new `refresh-catalog-entity` action to the catalog backend's `ActionsRegistryService`. The action triggers a refresh of a single entity identified by `kind` / `namespace` / `name` (kind defaults to `Component`, namespace defaults to `default`). It performs an existence check via `getEntityByRef` first and throws `NotFoundError` for missing entities. Exposed automatically as an MCP tool by `@backstage/plugin-mcp-actions-backend`.
```
freben: You can shorten this to be more end user impact focused and have a bit less implementation detail.

freben (changes requested):

Naga15:
Thanks for the review @freben — pushed 61bf0e90d4 addressing the resolution feedback.

**Resolution now mirrors `get-catalog-entity`:**
- `name` is required; `kind` and `namespace` are optional with no defaults.
- `catalog.queryEntities` filters on `metadata.name` plus whatever the caller supplied for `kind` / `namespace`.
- `InputError` on no match; `ConflictError` listing the candidate `entityRef`s on multiple matches.
- `refreshEntity` is then called against `stringifyEntityRef(items[0])`.

This closes the UX gap you flagged: an agent that finds `api:default/orders-api` via `get-catalog-entity` can now refresh it with just `{ name: 'orders-api' }` without knowing or re-specifying the kind.

**Tests refreshed:**
- single-match by name only
- disambiguation by kind
- disambiguation by kind + namespace
- no-match → `InputError`
- multiple-match → `ConflictError` (with candidate-list assertion)
- `refreshEntity` error propagation

**Changeset shortened** per your other note — focuses on end-user impact rather than implementation detail.

I'll update the PR description shortly to reflect the new model.

.changeset/catalog-refresh-entity-mcp-action.md
```diff
@@ -0,0 +1,5 @@
+---
+'@backstage/plugin-catalog-backend': minor
+---
+
+Added a `refresh-catalog-entity` action so agents and MCP clients can requeue a single entity for processing after creating or updating it — useful for reading back fresh data immediately after a scaffolder run without waiting for the next scheduled processing loop.
```
freben:
```suggestion
Added a `refresh-catalog-entity` action so agents and MCP clients can re-queue a single entity for processing after creating or updating it — useful for reading back fresh data immediately after a scaffolder run without waiting for the next scheduled processing loop.
```

freben (approved):

PR #34451, opened 2026-06-01
fix(core-components): lazy-load react-syntax-highlighter and dagre

freben:

The `@backstage/core-components` barrel (`export *`) eagerly loads every component and its transitive dependencies when any single export is imported. The two largest offenders are:

| Component | Dependency | Size |
|---|---|---|
| `CodeSnippet` | `react-syntax-highlighter` (192 highlight.js languages) | **~8.7 MB** |
| `DependencyGraph` | `@dagrejs/dagre` + d3-zoom/selection/shape | **~1.1 MB** |

This is particularly painful in Jest where `import()` is transpiled to synchronous `require()` — the entire module graph loads during React's render phase, blocking for hundreds of milliseconds and causing flaky test timeouts.

### Changes

Split both components into a thin `React.lazy` wrapper + the actual implementation:

- `CodeSnippet.tsx` → wrapper with `Suspense` + `lazy(() => import('./CodeSnippetContent'))`
- `DependencyGraph.tsx` → wrapper with `Suspense` + `lazy(() => import('./DependencyGraphContent'))`

The barrel export is unchanged. `import { CodeSnippet } from '@backstage/core-components'` still works identically — the heavy deps just don't load until the component actually renders. API report is unchanged.


packages/core-components/src/components/CodeSnippet/CodeSnippetContent.tsx
```diff
@@ -0,0 +1,127 @@
...
+    customStyle,
+    showCopyCodeButton = false,
+  } = props;
+  const theme = useTheme();
+  const mode = theme.palette.type === 'dark' ? dark : docco;
+  const highlightColor = theme.palette.type === 'dark' ? '#256bf3' : '#e6ffed';
+
+  return (
+    <Box position="relative">
+      <LightAsync
```
Rugvip, line 100: With the async wrapper we might not want ot use the async version of this anymore

Rugvip (approved):
👍

packages/core-components/src/components/CodeSnippet/CodeSnippetContent.tsx
Rugvip: Although not sure if that starts eagerly loading different syntax options, wouldn't want that either

freben: Both probably work - the existing choice is perhaps the conservative one, i think it's fine to keep

PR #34455, opened 2026-06-01
fix(ui): update Switch to respond to parent bg context

cdedreuille:
## Summary

- Switch now reads its parent background context (`bg: 'consumer'`) and sets `data-on-bg` on the root element
- Indicator track and disabled thumb colors vary per neutral bg level in light mode (gray 4/3/4/3 and 3/2/3/2)
- Dark mode uses gray-6 for the track and gray-5 for the disabled thumb across all contexts
- Selected state migrated from deprecated `--bui-bg-solid`/`--bui-fg-solid` to `--bui-accent-bg`/`--bui-accent-fg`
- Fixed dark mode `--bui-fg-disabled` token value
- Added `AutoBg` and `AutoBgDisabled` stories to visually verify bg context behaviour
- Fixed Storybook preview overflow/scroll in non-fullscreen mode

## Test plan

- [ ] Open the `AutoBg` story and verify the Switch indicator changes shade at each nesting level
- [ ] Open the `AutoBgDisabled` story and verify the thumb color changes shade at each nesting level
- [ ] Toggle dark mode and verify all contexts show gray-6 track and gray-5 disabled thumb
- [ ] Verify the selected state uses the accent colour in both light and dark mode

### DIFF
--- a/packages/ui/src/components/Switch/Switch.module.css
+++ b/packages/ui/src/components/Switch/Switch.module.css
@@ -18,6 +18,24 @@
 
 @layer components {
   .bui-Switch {
+    --switch-indicator-bg: var(--bui-gray-4);
+
+    &[data-on-bg='neutral-1'] {
+      --switch-indicator-bg: var(--bui-gray-3);
+    }
+
+    &[data-on-bg='neutral-2'] {
+      --switch-indicator-bg: var(--bui-gray-4);
+    }
+
+    &[data-on-bg='neutral-3'] {
+      --switch-indicator-bg: var(--bui-gray-3);
+    }
+
+    [data-theme-mode='dark'] & {
+      --switch-indicator-bg: var(--bui-gray-6);
+    }
+
     display: flex;
     /* This is needed so the HiddenInput is positioned correctly */
     position: relative;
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/Switch/definition.ts
```diff
@@ -28,6 +28,7 @@ export const SwitchDefinition = defineComponent<SwitchOwnProps>()({
     root: 'bui-Switch',
     indicator: 'bui-SwitchIndicator',
   },
+  bg: 'consumer',
```
Naycon, line 31: ❤️

packages/ui/src/components/Switch/Switch.module.css
```diff
@@ -18,6 +18,12 @@
 
 @layer components {
   .bui-Switch {
+    --switch-indicator-bg: var(--bui-gray-4);
+
+    [data-theme-mode='dark'] & {
+      --switch-indicator-bg: var(--bui-gray-6);
+    }
+
```
Naycon, line 26: These fail 3:1 contrast vs all neutral bgs, and vs the handle. We need better contrast for track vs background and handle vs track for the component to meet WCAG AA requirements for non-text graphical interactive elements. Can address that in a followup PR, but we have some ways to go since we're around 1.50:1 right now.

Naycon (approved):
Nice! Fixes the bug and is a definite step forward. We're still failing the contrast checks, but we can address that in a separate PR.

PR #34458, opened 2026-06-02
ci: add yarn lock changes workflow

jonkoops:
Adds a workflow that posts a comment on pull requests summarizing changes to `yarn.lock`, including added, removed, and updated packages (both direct and transitive).

Large lockfile diffs are difficult to review manually. This makes it easier for reviewers to understand the dependency impact of a PR at a glance -- for example, whether a seemingly small change pulls in dozens of new transitive dependencies, or whether a dependency removal actually cleans up the tree.

Uses [Simek/yarn-lock-changes](https://github.com/Simek/yarn-lock-changes). Example output: https://github.com/jonkoops/backstage/pull/1.

### DIFF
--- a/.github/workflows/automate_yarn-lock-changes.yml
+++ b/.github/workflows/automate_yarn-lock-changes.yml
@@ -5,6 +5,7 @@ jobs:
   yarn-lock-changes:
     runs-on: ubuntu-latest
     permissions:
+      contents: read
       pull-requests: write
     steps:
       - name: Harden Runner
--- a/.github/workflows/automate_yarn-lock-changes.yml
+++ b/.github/workflows/automate_yarn-lock-changes.yml
@@ -1,0 +1,21 @@
+name: Yarn Lock Changes
+on: [pull_request]
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/automate_yarn-lock-changes.yml
```diff
@@ -0,0 +1,21 @@
+name: Yarn Lock Changes
+on: [pull_request]
+
+jobs:
+  yarn-lock-changes:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
```
Rugvip, line 9: Think it'd be cleanest to set `contents: read` here as well.

Rugvip (changes requested):
Yep happy to try this out, just a minor thing

.github/workflows/automate_yarn-lock-changes.yml
jonkoops: Done :+1:

Rugvip (approved):

PR #34465, opened 2026-06-03
fix(core-components): honor closeDelayMs for sidebar submenus

kunickiaj:

A sidebar submenu closes the instant the pointer leaves the parent item. If the parent sits low in the rail and the expanded menu opens upward, you have to move almost perfectly horizontally to reach an entry — drift diagonally and the menu disappears before you can click it. I kept losing the menu while dogfooding our internal portal, so I went looking for a knob and found that there mostly already is one.

`defaultCloseDelayMs` already exists and is wired to the main rail's hover-collapse via the `closeDelayMs` prop on `<Sidebar>`. It just never reached the submenu. This change resolves the close delay once in `makeSidebarConfig` and surfaces it through `SidebarConfigContext`, so `SidebarItemWithSubmenu` can honor the same value: the menu stays open for `closeDelayMs` after the pointer leaves, and the pending close is cancelled if the pointer comes back.

The default is unchanged (`0` = close immediately), so nobody who hasn't set `closeDelayMs` sees any difference.

#### Why reuse `closeDelayMs` rather than add a new option

It's already documented as the "delay until the sidebar closes on hover," and a submenu is part of the sidebar, so honoring it there reads to me as fixing an inconsistency rather than adding surface area.

Demo (`closeDelayMs={500}`, exaggerated so it's visible):
sidebar-after

No docs change — this reuses the existing public `closeDelayMs` prop and adds no new public API.

### DIFF
--- a/.changeset/sidebar-submenu-close-delay.md
+++ b/.changeset/sidebar-submenu-close-delay.md
@@ -1,0 +1,5 @@
+---
+'@backstage/core-components': patch
+---
+
+A `SidebarItem` that contains a `SidebarSubmenu` now honors the `closeDelayMs` prop on `Sidebar`. Previously the expanded menu closed immediately when the pointer left the parent item, so moving the pointer diagonally toward one of its entries often dismissed it before it could be clicked. When `closeDelayMs` is set, the expanded menu stays open for the configured delay after the pointer leaves, and the close is cancelled if the pointer returns. The default is unchanged (`0`, close immediately), so existing behavior is preserved unless `closeDelayMs` is set.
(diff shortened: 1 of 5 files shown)


freben (commented):
Thanks! Just double checking: does this work well still with serveral popout parents in the sidebar that you flip between? Like having foo and bar in the sidebar, and moving the pointer between them their flyouts toggle instantly and don't overlap? :) Honestly it's probably a tricky thing because you'd have to choose between quick response time, or getting this "forgiving" ability to move over to the flyout without accidentally triggering its neighbor.

kunickiaj:
> Thanks! Just double checking: does this work well still with serveral popout parents in the sidebar that you flip between? Like having foo and bar in the sidebar, and moving the pointer between them their flyouts toggle instantly and don't overlap? :) Honestly it's probably a tricky thing because you'd have to choose between quick response time, or getting this "forgiving" ability to move over to the flyout without accidentally triggering its neighbor.

valid point, i added a repro case to story book to test with and its not handled as is, trying to see if i can update it to handle this better.

PR #34472, opened 2026-06-03
fix(scaffolder): coerce DatabaseTaskStore totalTasks count to a number

### LINKED ISSUE #34471: 🐛 Bug Report: scaffolder `list-scaffolder-tasks` action fails on PostgreSQL (totalTasks is a string, fails z.number() validation)
## 📜 Description

The `scaffolder:list-scaffolder-tasks` action (exposed to AI agents via the MCP actions backend) fails on PostgreSQL with:

```
InputError: Invalid output from action "scaffolder:list-scaffolder-tasks"; caused by [
  {
    "code": "invalid_type",
    "expected": "number",
    "received": "string",
    "path": [ "totalTasks" ],
    "message": "Expected number, received string"
  }
]
```

## 👍 Expected behavior

`list-scaffolder-tasks` returns a numeric `totalTasks`, regardless of the configured database engine.

## 👎 Actual Behavior

On PostgreSQL the output validation fails because `totalTasks` is a **string**.

## 🤖 Root cause

`DatabaseTaskStore.list()` (`plugins/scaffolder-backend/src/scaffolder/tasks/DatabaseTaskStore.ts`) builds a count query:

```ts
countQuery.count('tasks.id', { as: 'count' });
const [results, [{ count }]] = await Promise.all([queryBuilder.select(), countQuery]);
return { tasks, totalTasks: count };
```

knex returns a `COUNT(*)` aggregate as a **string** on PostgreSQL (the underlying column is a `bigint`), but as a **number** on `better-sqlite3`. The count is returned to `totalTasks` without coercion, so `totalTasks` is a string in production.

`createListScaffolderTasksAction` then assigns this value to an output declared as `totalTasks: z.number()`, so the MCP actions registry rejects the output on PostgreSQL.

Why it has gone unnoticed: `DatabaseTaskStore.test.ts` only runs against `better-sqlite3` (`client: 'better-sqlite3'`), where the count is already a number — so the PostgreSQL behaviour is never exercised.

## 👟 Reproduction steps

1. Run a backend configured with `database.client: pg`.
2. Enable the MCP actions backend exposing scaffolder actions.
3. Call `scaffolder:list-scaffolder-tasks` (e.g. `{ "limit": 20 }`).

(Equivalently: knex `.count('id', { as: 'count' })` returns `typeof count === 'string'` on pg vs `'number'` on better-sqlite3.)

## 📃 Provide the context for the Bug.

Surfaces when the scaffolder actions are consumed via the MCP actions backend on a PostgreSQL-backed deployment. Other scaffolder MCP actions are unaffected.

## 🖥️ Your Environment

- `@backstage/plugin-scaffolder-backend` 4.0.0 (also present on `master`)
- Database: PostgreSQL

A fix (coerce the count to a number in `DatabaseTaskStore.list`, plus a database-matrix regression test) is in the PR linked below.

mitsukado-shinagawa:

`DatabaseTaskStore.list()` returns the `COUNT(*)` aggregate directly as `totalTasks`. knex returns that aggregate as a **string** on PostgreSQL (the column is a `bigint`) but as a **number** on `better-sqlite3`, so in production `totalTasks` is a string.

This breaks the `list-scaffolder-tasks` action, whose output schema declares `totalTasks: z.number()`, with:

```
Invalid output from action "scaffolder:list-scaffolder-tasks"; caused by [
  { "code": "invalid_type", "expected": "number", "received": "string", "path": ["totalTasks"] }
]
```

Fixes #34471.

### Changes

- `DatabaseTaskStore.list()` coerces the count with `Number(...)` (with an explanatory comment).
- Added a regression test that runs the store against **every supported database**. The existing `DatabaseTaskStore.test.ts` only used `better-sqlite3` (where the count is already a number), which is why this was never caught — the new `describe.each(databases.eachSupportedId())` case exercises PostgreSQL and asserts `typeof totalTasks === 'number'`.
- Added a changeset (`patch` for `@backstage/plugin-scaffolder-backend`).

### Notes

Opened as a **draft** pending the EasyCLA sign-off. Happy to adjust the test approach (this introduces the first `TestDatabases` usage in `scaffolder-backend`) if maintainers prefer a different placement.

✔️ Checklist
- [x] Added or updated documentation (n/a — internal fix)
- [x] Screenshots attached (n/a)

### DIFF
--- a/.changeset/late-snakes-coerce-count.md
+++ b/.changeset/late-snakes-coerce-count.md
@@ -1,0 +1,5 @@
+---
+'@backstage/plugin-scaffolder-backend': patch
+---
+
+Fixed `DatabaseTaskStore.list` returning `totalTasks` as a string on PostgreSQL. knex returns a `COUNT(*)` aggregate as a string on PostgreSQL (the column is a bigint) while better-sqlite3 returns a number, so the count is now coerced with `Number(...)`. This in turn fixes the `list-scaffolder-tasks` action, whose output schema declares `totalTasks: z.number()` and previously failed validation in production with `Invalid output ... totalTasks: Expected number, received string`.
--- a/plugins/scaffolder-backend/src/scaffolder/tasks/DatabaseTaskStore.test.ts
+++ b/plugins/scaffolder-backend/src/scaffolder/tasks/DatabaseTaskStore.test.ts
@@ -22,6 +22,7 @@ import { ConflictError } from '@backstage/errors';
 import {
   mockServices,
   createMockDirectory,
+  TestDatabases,
 } from '@backstage/backend-test-utils';
 import fs from 'fs-extra';
 import { EventsService } from '@backstage/plugin-events-node';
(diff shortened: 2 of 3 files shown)


mitsukado-shinagawa:
Reworked this as #34479 (the fork/branch behind this PR was deleted, so it couldn't be reopened). The new PR rebases the same fix on current `master` and addresses the automated review feedback here: an `Number.isSafeInteger` guard on the coerced count, a per-test timeout instead of the file-wide `jest.setTimeout`, and a note that `TestDatabases.create()` already registers its own `afterAll` teardown.

PR #34478, opened 2026-06-04
Add support for MS Graph events

stanislav-c:
This pull request introduces a new Microsoft Graph backend module for the Backstage events plugin. The module enables ingestion of Microsoft Graph change notifications, manages webhook subscriptions, and republishes relevant events to Backstage topics. The implementation includes configuration, database schema, migration scripts, and tests to ensure robust handling of subscriptions and event processing.

Events published through this module can later be used both in `@backstage/plugin-catalog-backend-module-msgraph` and `@backstage/plugin-catalog-backend-module-msgraph-incremental` to reflect EntraID User and Group changes as soon as they occur.

## How it works

Unlike GitHub webhook events, MS Graph events cannot be configured in the portal. Client Apps have to create new subscriptions themselves, specifying the webhook URLs, verification secrets/tokens and all other params [through an endpoint](https://learn.microsoft.com/en-us/graph/change-notifications-delivery-webhooks?tabs=javascript#create-a-subscription). The new module `@backstage/plugin-catalog-backend-module-msgraph` looks up deleted entityRefs in the Catalog and re-publishes events to more specific `upsert`/`delete` topics.

```mermaid
sequenceDiagram
    participant eventsModuleMicrosoftGraphWebhook
    participant EventsPlugin
    participant WebhookRequestValidator
    participant MicrosoftGraphEventRouter
    participant CatalogService
    participant MicrosoftGraphIncrementalEntityProvider

    eventsModuleMicrosoftGraphWebhook->>EventsPlugin: eventsExtensionPoint.addHttpPostIngress
    EventsPlugin->>WebhookRequestValidator: createWebhookRequestValidator
    loop on schedule
        Note over eventsModuleMicrosoftGraphWebhook: after lifecycle.addStartupHook
        eventsModuleMicrosoftGraphWebhook->>eventsModuleMicrosoftGraphWebhook: MicrosoftGraphSubscriptionManager.ensureActiveSubscriptions
    end
    EventsPlugin->>WebhookRequestValidator: webhook event request
    WebhookRequestValidator->>WebhookRequestValidator: validate request
    WebhookRequestValidator->>EventsPlugin: request secret valid
    EventsPlugin->>MicrosoftGraphEventRouter: MS Graph change/delete events<br/>published via "msgraph" topic
    MicrosoftGraphEventRouter->>MicrosoftGraphEventRouter: validate payloads
    MicrosoftGraphEventRouter->>EventsPlugin: change/add events re-published via "msgraph/upsert" topic
    MicrosoftGraphEventRouter->>CatalogService: find deleted entities by MS Graph IDs
    CatalogService->>MicrosoftGraphEventRouter: deleted entityRefs
    MicrosoftGraphEventRouter->>EventsPlugin: delete events re-published via "msgraph/delete" topic
    EventsPlugin->>MicrosoftGraphIncrementalEntityProvider: "msgraph/upsert" and "msgraph/delete" topic events
```

All incoming event requests are validated on client secret.

```mermaid
sequenceDiagram
    participant WebhookRequestValidator
    participant MicrosoftGraphSubscriptionManager
    participant MicrosoftGraphSubscriptionsDatabaseClient
    participant MicrosoftGraphClient

    loop on schedule
        MicrosoftGraphSubscriptionManager->>MicrosoftGraphSubscriptionManager: ensureActiveSubscriptions
        MicrosoftGraphSubscriptionManager->>MicrosoftGraphSubscriptionsDatabaseClient: get current subscription IDs
        MicrosoftGraphSubscriptionsDatabaseClient->>MicrosoftGraphSubscriptionManager: subscription IDs
        MicrosoftGraphSubscriptionManager->>MicrosoftGraphClient: validateActiveSubscription
        MicrosoftGraphClient->>MicrosoftGraphSubscriptionManager: expired/invalid subscriptions
        MicrosoftGraphSubscriptionManager->>MicrosoftGraphClient: createSubscription
        MicrosoftGraphSubscriptionManager->>MicrosoftGraphSubscriptionsDatabaseClient: insert subscription record with secret hash and salt
    end

    Note over WebhookRequestValidator: on new webhook event
    WebhookRequestValidator->>MicrosoftGraphSubscriptionsDatabaseClient: databaseClient.getById(subscriptionId)
    MicrosoftGraphSubscriptionsDatabaseClient->>WebhookRequestValidator: token_hash, token_salt
    WebhookRequestValidator->>WebhookRequestValidator: reject request if token_hash !== hashValidationToken(clientState, token_salt)
```

Upsert/Delete events can later be received by `MicrosoftGraphOrgEntityProvider` or `MicrosoftGraphIncrementalEntityProvider`. Group/User IDs from the Upsert events can be used to query MS Graph APIs for all data necessary to build Group/User entities. Delete events already contain `CompoundEntityRefs` at this point. Catalog entities can be deleted/added/updated via `connection.applyMutation({..., type: 'delta'})`.

## How to test locally

LocalTunnel can be used to receive webhook events in local environments.

1. `brew install localtunnel`
2. `lt --port 7007`
3. Copy the URL printed and add it to your events config:
  ```yaml
  events:
    modules:
      msgraph:
        notificationUrl: https://shaggy-mirrors-admire.loca.lt/api/events/http/msgraph # append the URL here
        resources:
          - groups
          - users
        clientId: <...>
        clientSecret: <...>
        tenantId: <...>
  ```
4. Run with DEBUG log level if you want to see all messages `LOG_LEVEL=DEBUG yarn start backend`


plugins/events-backend-module-msgraph/src/service/MicrosoftGraphClient.ts
```diff
@@ -0,0 +1,185 @@
...
+ * Client for interacting with Microsoft Graph Subscriptions API.
+ * @internal
+ */
+export class MicrosoftGraphClient {
+  static fromConfig(
+    config: Config,
+    clientWrapper?: ClientWrapper,
+  ): MicrosoftGraphClient {
+    // Fail fast if the configuration is missing or incomplete
+    config.getConfig('events.modules.msgraph.notificationUrl');
```
freben, line 96: `notificationUrl` is defined and documented as a string, but `getConfig` requires an object-valued config node. A valid documented configuration therefore throws here before the module can initialize. This should be `getString(...)`, ideally covered by a startup/integration test that uses the README configuration without mocking `MicrosoftGraphClient.fromConfig`.

plugins/events-backend-module-msgraph/src/service/MicrosoftGraphSubscriptionManager.ts
```diff
@@ -0,0 +1,276 @@
...
+export class MicrosoftGraphSubscriptionManager {
+  static async fromConfig(
+    config: Config,
+    {
+      logger,
+      databaseClient,
+      msGraphClient,
+    }: MicrosoftGraphSubscriptionManagerOptions,
+  ): Promise<MicrosoftGraphSubscriptionManager> {
+    const resources = config.getStringArray('events.modules.msgraph.resources');
```
freben, line 101: The public schema and README expose this as `events.modules.msgraph.subscriptionResources`, while runtime reads `resources`. Users following the schema therefore fail here even after the `notificationUrl` issue is fixed. Please standardize on the public key and make the full module-startup test use that same spelling.

plugins/events-backend-module-msgraph/src/service/createWebhookRequestValidator.ts
```diff
@@ -0,0 +1,128 @@
...
+    context: RequestValidationContext,
+  ): Promise<void> {
+    if (request.query.validationToken) {
+      logger.info('Received validation request from MS Graph.');
+      logger.debug('Responding with URL-decoded token through context.reject.');
+
+      // This is a validation request, the token needs to be decoded and sent back
+      context.reject({
+        status: 200,
+        payload: decodeURIComponent(request.query.validationToken as string),
```
freben, line 54: Express has already URL-decoded `request.query`, so this decodes Graph's opaque token a second time. For example, a token containing a literal percent escape can be changed or can throw here, causing endpoint validation to fail. The query type also permits arrays/objects, so the cast is unsafe. Please require a single string and return that parsed value directly; an HTTP-level test with encoded percent characters would catch this.

plugins/events-backend-module-msgraph/src/service/createWebhookRequestValidator.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    if (validValues === 0) {
+      logger.warn(
+        `No valid notification items found among ${value.length} items`,
+      );
+      context.reject();
+      return;
+    }
+
+    if (validValues < value.length) {
```
freben, line 120: This warning still accepts the request, after which the ingress publishes the original unfiltered `value` array. Consequently, one valid notification authenticates every invalid/unknown item alongside it. A replayed valid item could be combined with forged user/group changes and those changes would reach the router. Since this validator API cannot replace the payload with a filtered copy, please reject unless every notification item validates (and update the test that currently enshrines the mixed-validity acceptance).

plugins/events-backend-module-msgraph/src/service/MicrosoftGraphSubscriptionManager.ts
```diff
@@ -0,0 +1,276 @@
...
+    if (!schedule && !scheduler) {
+      throw new Error('Either schedule or scheduler must be provided.');
+    }
+
+    const taskRunner =
+      schedule ??
+      scheduler!.createScheduledTaskRunner(subscriptionRefreshSchedule);
+
+    // Ensure we have active subscriptions right away
+    await this.ensureActiveSubscriptions();
```
freben, line 131: This initial reconciliation runs outside the globally scoped scheduled task, so every backend replica executes it concurrently during startup. Multiple replicas can all observe no row and race into Graph subscription creation; Graph rejects duplicate subscriptions for the same resource/change type, potentially failing startup. Please perform the initial run under the global scheduler lock (or another durable per-resource lease) as well.

plugins/events-backend-module-msgraph/src/router/MicrosoftGraphEventRouter.ts
```diff
@@ -0,0 +1,195 @@
...
+    const { upsertedUsers, upsertedGroups, deletedUsers, deletedGroups } =
+      parsedEvents;
+
+    if (upsertedUsers.length > 0 || upsertedGroups.length > 0) {
+      this.logger.debug(
+        `Forwarding ${upsertedUsers.length} modified users and ${upsertedGroups.length} modified groups`,
+      );
+
+      // Republish payload as is - MS Graph provider will fetch the details by IDs
+      this.events
```
freben, line 104: Please await this publication instead of starting it in the background and swallowing rejection. `onEvent` currently reports success before the routed event is published, so a publish failure is merely logged and the input event is permanently considered handled. The standard `EventRouter` awaits `events.publish`, allowing failure semantics to propagate. The delete publication below has the same issue.

freben (changes requested):
The overall module shape fits the existing events modules reasonably well: the feature loader, split webhook/router modules, extension point, and events service usage are familiar. The additional state and scheduling also make sense for Microsoft Graph.

I do not think this is ready to merge yet, though. The inline comments include two independent configuration problems that prevent a documented setup from starting, an authentication issue where invalid notification items can ride along in a valid batch, webhook handshake handling, and an HA startup race.

A few non-blocking design points are worth considering while addressing those:

- This is currently single-tenant, whereas the catalog Microsoft Graph modules support multiple named providers; subscription storage and emitted events carry no provider/tenant identity. If single-tenant is intentional, it should be explicit.
- User/group subscriptions support lifetimes of almost 29 days and Graph supports renewal with `PATCH /subscriptions/{id}`. Creating a minimum-lifetime subscription and replacing it roughly every 35 minutes produces avoidable churn.
- The README diagram currently depicts the incremental provider consuming `msgraph/upsert` and `msgraph/delete`, but neither catalog Microsoft Graph module subscribes to those topics yet. Describing them as topics that future consumers can use would be more accurate.

The direction looks promising, but the startup, validation, and HA issues need fixing before merge.

